PR #1972, opened 2022-02-17
Avoid AdminUpdate panic when Nodes are down

UlrichSchlueter:
### Which issue this PR addresses:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_boards/board/t/Azure%20Red%20Hat%20OpenShift/Stories/?workitem=12478329 Investigate AdminUpdate panic RP: runtime error: index out of range [0] with length 0

Refactors the cluster deploy manager to allow for unit testing of the arooperator createorupdate action

Adds a new mock object for aro operator deployment (and the generate commands)

### What this PR does / why we need it:

This PR does three things:
1) Skip the ensureAroOperator action  and aroDeploymentReady condition if the default IngressProfile information is not available (like it is the case in the work item when the nodes are down at adminupdate time)
2) Refactor the cluster deploy manager for testability. The change decouples the create of the aro operator deployer from the ensureAroOperator action by moving into a new initialization action. This change affects the AdminUpdate, update and install path.
3) Adds a unit test and a new mock implemenation for the aro operator deployment (and the generate commands)

### Test plan for issue:

Manual test for admin update , and install done
Unit tests for ensureAroOperator and aroDeploymentReady added
adminupdate_test update with the new action

### Is there any documentation that needs to be updated for this PR?

Don't think so, this is a fix and internal refactoring

PR #2008, opened 2022-03-16
Make createOrUpdateRouterIPFromCluster reuse existing code + extra case for TestAroDeploymentReady

m1kola:

Just a follow up for #1972:

* Makes `createOrUpdateRouterIPFromCluster` reuse `isIngressProfileAvailable`
* And after merge I noticed that we should probably add an extra test case into `TestAroDeploymentReady`.

### Test plan for issue:

Unit tests should cover the changes

No, just refactoring

### DIFF
--- a/pkg/cluster/ipaddresses.go
+++ b/pkg/cluster/ipaddresses.go
@@ -48,18 +48,7 @@ func (m *manager) updateClusterData(ctx context.Context) error {
 }
 
 func (m *manager) createOrUpdateRouterIPFromCluster(ctx context.Context) error {
-	// check if ingress profile contains default profile we intend to use.
-	// It might not exist if customer updated the profile or api is down.
-	// in both cases we can't do much so return early. Ingress profile can
-	// be set to nil by enricher
-	var found bool
-	for _, ip := range m.doc.OpenShiftCluster.Properties.IngressProfiles {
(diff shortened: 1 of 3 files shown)


bennerv (dismissed):

pkg/cluster/ipaddresses.go
```diff
@@ -52,14 +52,7 @@ func (m *manager) createOrUpdateRouterIPFromCluster(ctx context.Context) error {
 	// It might not exist if customer updated the profile or api is down.
 	// in both cases we can't do much so return early. Ingress profile can
```
s-amann, line 53: Can we remove or clean up this comment now that we are using a well named function I think it explains itself.

s-amann (changes requested):
lgtm, can we remove or clean up the comment?

m1kola:
@bennerv @s-amann I cleaned up the comment. Could you please take another look?

bennerv (approved):
LGTM. 

We rely heavily on this ingress IP address due to the integrations with the dnsmasq controller.  If for some reason the ingress IP address will not populate within the cluster doc, we're going to have issues deploying the operator, DNS for the *.apps domain may go sideways (if the ingress service was deleted or recreated with a new IP address).

PR #2011, opened 2022-03-17
Update OB Build Pipeline to Pass Build Tag as Var

mikeandescavage:

Fixes
https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/13675519

OneBranch pipelines do not support building from a release tag (fails pre-build validation steps).  This is a workaround to support the RH team to still use release tags.

So long a commitid has a release tag associated with it, we will be able to derive the tag and attach it to the container at build time.  Otherwise, we will attach the commitid.

### Test plan for issue:

https://dev.azure.com/msazure/AzureRedHatOpenShift/_build/results?buildId=53181311&view=results

https://dev.azure.com/msazure/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/233405/Performing-Release

### DIFF
--- a/.pipelines/onebranch/templates/template-buildrp-buildaro.yml
+++ b/.pipelines/onebranch/templates/template-buildrp-buildaro.yml
@@ -5,6 +5,16 @@ steps:
     targetType: inline
     script: |
       export GOPATH=$(Agent.TempDirectory)
+      export TAG=$(git describe --exact-match 2>/dev/null)
+      export COMMIT=$(git rev-parse --short=7 HEAD)$([[ $(git status --porcelain) = "" ]] || echo -dirty)
+      if [ -z "$TAG" ];
+      then
+        export VERSION=${COMMIT}
+      else
+        export VERSION=${TAG}
+      fi
+      echo "Version: ${VERSION}"
+      echo "##vso[task.setvariable variable=releasetag;isOutput=true]${VERSION}"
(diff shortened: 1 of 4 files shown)


bryanro92 (dismissed):

petrkotas:
@mikeandescavage I need to clarify how the trigger works. Is the pipeline triggered for each commit or it is triggered manually? I cannot find it in the templates and I see trigger none. 
Maybe we can trigger the pipeline after our successful checks?

mikeandescavage:
@petrkotas the trigger is defined in the UI of ADO for the pipeline.  I'll send a screenshot via Teams

.pipelines/onebranch/templates/template-buildrp-buildaro.yml
```diff
@@ -5,13 +5,24 @@ steps:
...
+      export TAG=$(git describe --exact-match 2>/dev/null)
+      export COMMIT=$(git rev-parse --short=7 HEAD)$([[ $(git status --porcelain) = "" ]] || echo -dirty)
+      if [ -z "$TAG" ];
+      then
+        export VERSION=${COMMIT}
+      else
+        export VERSION=${TAG}
+      fi
+      echo "Version: ${VERSION}"
+      echo "##vso[task.setvariable variable=releasetag;isOutput=true]${VERSION}"
```
bryanro92, line 17: interesting trick to export this

bryanro92 (approved):

PR #2014, opened 2022-03-18
modified pipeline doc folder exclude rule

bizz001:

Fixes CI not ignoring commits containing only doc changes.

Prevents CI from running on doc only commits.

### DIFF
--- a/.pipelines/ci.yml
+++ b/.pipelines/ci.yml
@@ -5,7 +5,7 @@ trigger:
     - master
   paths:
     exclude:
-    - docs/*
+    - docs/
   tags:
     include:
     - v2*
@@ -16,7 +16,7 @@ pr:
     - master
   paths:
     exclude:
-    - docs/*
+    - docs/
 
 variables:
 - template: vars.yml


petrkotas (approved):
LGTM nice catch

bryanro92 (changes requested):
I'm not sure why - but this isn't working. CI is still running, and the pipelines run off your PR branch. I tested a similar change for e2e testing.

Could you implement what I did here for `e2e.yml`? https://github.com/bryanro92/ARO-RP/commit/d99a7a15aedacda3602d6f23d0c2b0c8d30c0bc0

I copied exact pattern from ci.yml to e2e.yml and it circumvented e2e from running, but CI is still running :thinking: 

EDIT: CI actually blew up on my suggested changes with
```
ruamel.yaml.scanner.ScannerError: mapping values are not allowed here
  in ".pipelines/e2e.yml", line 2, column 8
```
However, if we get CI to exclude properly then it wont fail :smile:

.pipelines/ci.yml
```diff
@@ -5,7 +5,7 @@ trigger:
     - master
   paths:
     exclude:
-    - docs/*
+    - docs/
```
m1kola, line 8: It should probably be a recursive wildcard instead: `**`. Found some docs on it:

* https://docs.microsoft.com/en-us/azure/devops/release-notes/2021/sprint-192-update#support-for-wild-cards-in-path-filters
* https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml
* https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/file-matching-patterns?view=azure-devops

m1kola:
@bizz001 is it something we still need? If so, could you please rebase and make sure that CI is green?

cadenmarchese:
@bizz001 bump to address comments above (if we still need this PR).

Issue #2010, opened 2022-03-17
ARO operator retrieves default ingress IP in wrong way


slawekwww:
ARO may contains many Ingress Controllers in ingressProfiles:

```json
  "ingressProfiles": [
    {
      "ip": "10.68.31.30",
      "name": "aaa",
      "visibility": "Private"
    },
    {
      "ip": "10.68.31.254",
      "name": "default",
      "visibility": "Private"
    },
    {
      "ip": "10.68.16.7",
      "name": "fo-eng-eastus",
      "visibility": "Private"
    }
```

Once aro operator retrieves it, it does not check that name is "default" but just take the 1st element from list sorted alphabetically.

https://github.com/Azure/ARO-RP/blob/94c46fc48225a3debfd5844b2d6baf16a69747a1/pkg/operator/deploy/deploy.go#L183

It caused that CRD clusters.aro.openshift.io/v1alpha1 is wrong populated with IngressIP and it impacts ARO cluster workloads - apps accessible via default router is not reachable within ARO.

Please fix this critical bug.

PR #2021, opened 2022-03-22
check for ingressIP when ingressProfiles > 1

karanmagdani:
Signed-off-by: Karan.Magdani <kmagdani@redhat.com>

Fixes https://github.com/Azure/ARO-RP/issues/2010
Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13948245

Bug where we end up selecting wrong ingress when there are multiple ingressProfiles.

### DIFF
--- a/pkg/operator/deploy/deploy.go
+++ b/pkg/operator/deploy/deploy.go
@@ -5,6 +5,7 @@ package deploy
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -138,6 +139,11 @@ func (o *operator) resources() ([]kruntime.Object, error) {
 		domain += "." + o.env.Domain()
 	}
 
+	ingressIP, err := checkIngressIP(o.oc.Properties.IngressProfiles)
+	if err != nil {
+		return nil, err
+	}
+
 	serviceSubnets := []string{
 		"/subscriptions/" + o.env.SubscriptionID() + "/resourceGroups/" + o.env.ResourceGroup() + "/providers/Microsoft.Network/virtualNetworks/rp-pe-vnet-001/subnets/rp-pe-subnet",
 		"/subscriptions/" + o.env.SubscriptionID() + "/resourceGroups/" + o.env.ResourceGroup() + "/providers/Microsoft.Network/virtualNetworks/rp-vnet/subnets/rp-subnet",
@@ -180,7 +186,7 @@ func (o *operator) resources() ([]kruntime.Object, error) {
 			},
 
 			APIIntIP:                 o.oc.Properties.APIServerProfile.IntIP,
-			IngressIP:                o.oc.Properties.IngressProfiles[0].IP,
+			IngressIP:                ingressIP,
 			GatewayPrivateEndpointIP: o.oc.Properties.NetworkProfile.GatewayPrivateEndpointIP,
 			// Update the OperatorFlags from the version in the RP
 			OperatorFlags: arov1alpha1.OperatorFlags(o.oc.Properties.OperatorFlags),
@@ -312,6 +318,21 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
 	}
 
 	return true, nil
+}
+
+func checkIngressIP(ingressProfiles []api.IngressProfile) (string, error) {
+	if ingressProfiles == nil || len(ingressProfiles) < 1 {
+		return "", errors.New("No Ingress Profiles found")
+	}
+	ingressIP := ingressProfiles[0].IP
+	if len(ingressProfiles) > 1 {
+		for _, p := range ingressProfiles {
+			if p.Name == "default" {
+				return p.IP, nil
+			}
+		}
+	}
+	return ingressIP, nil
 }
 
 func isCRDEstablished(crd *extensionsv1.CustomResourceDefinition) bool {
--- a/pkg/operator/deploy/deploy_test.go
+++ b/pkg/operator/deploy/deploy_test.go
@@ -1,0 +1,127 @@
+package deploy
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/util/cmp"
+)
+
+func TestCheckIngressIP(t *testing.T) {
+
+	type test struct {
+		name    string
+		oc      func() *api.OpenShiftClusterProperties
+		want    string
+		wantErr error
+	}
+
+	for _, tt := range []*test{
+		{
+			name: "default IngressProfile",
+			oc: func() *api.OpenShiftClusterProperties {
+				return &api.OpenShiftClusterProperties{
+					IngressProfiles: []api.IngressProfile{
+						{
+							Name: "default",
+							IP:   "1.2.3.4",
+						},
+					},
+				}
+			},
+			want:    "1.2.3.4",
+			wantErr: nil,
+		},
+		{
+			name: "Multiple IngressProfiles, pick default",
+			oc: func() *api.OpenShiftClusterProperties {
+				return &api.OpenShiftClusterProperties{
+					IngressProfiles: []api.IngressProfile{
+						{
+							Name: "custom-ingress",
+							IP:   "1.1.1.1",
+						},
+						{
+							Name: "default",
+							IP:   "1.2.3.4",
+						},
+						{
+							Name: "not-default",
+							IP:   "1.1.2.2",
+						},
+					},
+				}
+			},
+			want:    "1.2.3.4",
+			wantErr: nil,
+		},
+		{
+			name: "Single Ingress Profile, No Default",
+			oc: func() *api.OpenShiftClusterProperties {
+				return &api.OpenShiftClusterProperties{
+					IngressProfiles: []api.IngressProfile{
+						{
+							Name: "custom-ingress",
+							IP:   "1.1.1.1",
+						},
+					},
+				}
+			},
+			want:    "1.1.1.1",
+			wantErr: nil,
+		},
+		{
+			name: "Multiple Ingress Profiles, No Default",
+			oc: func() *api.OpenShiftClusterProperties {
+				return &api.OpenShiftClusterProperties{
+					IngressProfiles: []api.IngressProfile{
+						{
+							Name: "custom-ingress",
+							IP:   "1.1.1.1",
+						},
+						{
+							Name: "not-default",
+							IP:   "1.1.2.2",
+						},
+					},
+				}
+			},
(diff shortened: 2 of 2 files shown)


pkg/operator/deploy/deploy.go
```diff
@@ -138,6 +138,15 @@ func (o *operator) resources() ([]kruntime.Object, error) {
 		domain += "." + o.env.Domain()
 	}
 
+	ingressIP := o.oc.Properties.IngressProfiles[0].IP
+	if o.oc.Properties.IngressProfiles != nil && len(o.oc.Properties.IngressProfiles) > 1 {
```
s-amann, line 142: At this point we would have already blown up if `o.oc.Properties.IngressProfiles` was nil, perhaps we should invert the condition and return an error if it is nil and then proceed with the logic of using the first one or the one named `router-default`

pkg/operator/deploy/deploy.go
```diff
@@ -138,6 +138,15 @@ func (o *operator) resources() ([]kruntime.Object, error) {
...
+		for _, p := range o.oc.Properties.IngressProfiles {
+			if p.Name == "router-default" {
```
bennerv, line 144: I think the name is `default` not `router-default`

```json
    "ingressProfiles": [
      {
        "ip": "1.2.3.4",
        "name": "default",
        "visibility": "Public"
      }
    ],
```

pkg/operator/deploy/deploy.go
karanmagdani: You're right lol! thanks, fixed

pkg/operator/deploy/deploy.go
karanmagdani: Fixed, Thanks!!

s-amann (dismissed):

bryanro92 (commented):
lgtm, but I wonder if we can use this as an opportunity to write unit test for `deploy.go`. even if the only coverage would be for this added block of code which could easily be refactored into a function

pkg/operator/deploy/deploy.go
```diff
@@ -138,6 +138,15 @@ func (o *operator) resources() ([]kruntime.Object, error) {
 		domain += "." + o.env.Domain()
 	}
 
+	ingressIP := o.oc.Properties.IngressProfiles[0].IP
```
m1kola, line 141: `o.oc.Properties.IngressProfiles` can be `nil`, I think.

pkg/operator/deploy/deploy.go
```diff
@@ -138,6 +138,15 @@ func (o *operator) resources() ([]kruntime.Object, error) {
...
+	if len(o.oc.Properties.IngressProfiles) > 1 {
+		for _, p := range o.oc.Properties.IngressProfiles {
+			if p.Name == "default" {
+				ingressIP = p.IP
+			}
```
m1kola, line 146: Ingress profile with name `"default"` is not guaranteed to exist. I think customers can create an ingress controller `"foo"` and delete `"default"`. What happens in this case?

m1kola (dismissed):

pkg/operator/deploy/deploy.go
bryanro92: @karanmagdani can we combine my suggestion of adding some unit testing here with @m1kola's points about edge cases and make sure we are properly handling all cases?

karanmagdani: Added test cases and covered in function. 
For case when: there is no default, it just picks the first ingress profile available (what we currently do). IMO this is fine because i don't think we have a way to find the "non-default" default

pkg/operator/deploy/deploy.go
karanmagdani: Fixed, Thanks

pkg/operator/deploy/deploy.go
```diff
@@ -314,6 +319,23 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
...
+func GetDefaultImageIP(ingressProfiles []api.IngressProfile) (string, error) {
+	if ingressProfiles == nil || len(ingressProfiles) < 1 {
+		return "", fmt.Errorf("No Ingress Profiles found")
+	}
+	ingressIP := ingressProfiles[0].IP
+	if len(ingressProfiles) > 1 {
+		for _, p := range ingressProfiles {
+			if p.Name == "default" {
+				ingressIP = p.IP
+				return ingressIP, nil
```
s-amann, line 332: nit: we can skip assignment and return `p.IP`
```suggestion
				return p.IP, nil
```

pkg/operator/deploy/deploy_test.go
```diff
@@ -0,0 +1,118 @@
...
+				}
+			},
+			want:    "1.1.1.1",
+			wantErr: ``,
+		},
+		{
+			name: "No IngressProfiles, Error",
+			oc: func() *api.OpenShiftClusterProperties {
+				return &api.OpenShiftClusterProperties{
+					IngressProfiles: []api.IngressProfile{},
```
s-amann, line 99: nit: should be super easy to add the final case of `nil` `IngressProfiles` :)

s-amann (approved):
lgtm, my nits are just nits :)

pkg/operator/deploy/deploy.go
karanmagdani: Good point, thanks!

pkg/operator/deploy/deploy_test.go
karanmagdani: fixed, thank you

pkg/operator/deploy/deploy.go
karanmagdani: fixed, thank you!

s-amann (approved):
lgtm

pkg/operator/deploy/deploy.go
```diff
@@ -138,6 +138,11 @@ func (o *operator) resources() ([]kruntime.Object, error) {
 		domain += "." + o.env.Domain()
 	}
 
+	ingressIP, err := GetDefaultImageIP(o.oc.Properties.IngressProfiles)
```
bryanro92, line 141: Why is this function called GetDefaultImageIP? Shouldn't it be IngressIP?

bryanro92 (changes requested):
I think function has confusing name

s-amann (approved):
lgtm

pkg/operator/deploy/deploy.go
```diff
@@ -314,6 +319,22 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
 	return true, nil
 }
 
+// o.oc.Properties.IngressProfile
+func GetIngressIP(ingressProfiles []api.IngressProfile) (string, error) {
```
cadenmarchese, line 323:
```suggestion
func getIngressIP(ingressProfiles []api.IngressProfile) (string, error) {
```

pkg/operator/deploy/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+// Licensed under the Apache License 2.0.
+
+import (
+	"testing"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/util/cmp"
+)
+
+func TestIngressProfiles(t *testing.T) {
```
cadenmarchese, line 13: Maybe this naming makes more sense?
```suggestion
func TestGetIngressIP(t *testing.T) {
```

pkg/operator/deploy/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	type test struct {
+		name    string
+		oc      func() *api.OpenShiftClusterProperties
+		want    string
+		wantErr string
```
cadenmarchese, line 19:
```suggestion
		wantErr error
```

pkg/operator/deploy/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+					IngressProfiles: []api.IngressProfile{
+						{
+							Name: "default",
+							IP:   "1.2.3.4",
+						},
+					},
+				}
+			},
+			want:    "1.2.3.4",
+			wantErr: ``,
```
cadenmarchese, line 36:
```suggestion
			wantErr: nil,
```
Same suggestion for below cases - your function returns `error` type so we should probably use the same type in the test.

pkg/operator/deploy/deploy_test.go
```diff
@@ -0,0 +1,126 @@
...
+			want:    "",
+			wantErr: "No Ingress Profiles found",
+		},
+		{
+			name: "Nil IngressProfiles, Error",
+			oc: func() *api.OpenShiftClusterProperties {
+				return &api.OpenShiftClusterProperties{}
+			},
+			want:    "",
+			wantErr: "No Ingress Profiles found",
```
cadenmarchese, line 111:
```suggestion
			wantErr: errors.New("No Ingress Profiles found"),
```

pkg/operator/deploy/deploy.go
```diff
@@ -138,6 +138,11 @@ func (o *operator) resources() ([]kruntime.Object, error) {
 		domain += "." + o.env.Domain()
 	}
 
+	ingressIP, err := GetIngressIP(o.oc.Properties.IngressProfiles)
```
cadenmarchese, line 141:
```suggestion
	ingressIP, err := getIngressIP(o.oc.Properties.IngressProfiles)
```

cadenmarchese (changes requested):
Nice work on this, just some minor suggestions / nits for consistency.

pkg/operator/deploy/deploy.go
karanmagdani: Fixed

pkg/operator/deploy/deploy_test.go
karanmagdani: Fixed

pkg/operator/deploy/deploy.go
karanmagdani: Thank you! Fixed as per suggestions

s-amann (approved):

s-amann (dismissed):

bryanro92 (changes requested):
nit: name of function

pkg/operator/deploy/deploy.go
```diff
@@ -314,6 +320,21 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
 	return true, nil
 }
 
+func getIngressIP(ingressProfiles []api.IngressProfile) (string, error) {
```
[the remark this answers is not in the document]
karanmagdani: Fixed, changed to checkIngressIP

bryanro92 (approved):
lgtm, we may need a controller portion of this to watch the ingress profile from within the cluster, but this is at least an iteration in a better direction

PR #2022, opened 2022-03-22
Add node-selector to MUO Namespace

bennerv:

Fixes a case where a customer sets a default node selector on their scheduler and pods cannot schedule on master nodes. 

Ensures that the namespace node selector overrides any custom scheduler in place causing MUO openshift workloads from scheduling.  

### Test plan for issue:

Tested on a cluster experiencing the issue and it resolved it.  

no


s-amann (approved):
lgtm

bryanro92 (approved):

PR #2024, opened 2022-03-23
disable ipv6 router advertisements on rp/gateway vmss

bryanro92:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13911582
Fixes https://access.redhat.com/security/cve/cve-2020-13401 

Disables ipv6 router advertisements as docker has a vulnerability related to them. Originally we had just dropped cap_net_raw, but it was insisted we disable ipv6 router advertisements in addition to the previous mitigation.

### Test plan for issue:

CI/e2e's

no, fixes issue with docker


pkg/deploy/generator/resources_rp.go
```diff
@@ -548,6 +548,12 @@ yum --enablerepo=rhui-rhel-7-server-rhui-optional-rpms -y install clamav azsec-c
 done
 
 rpm -e $(rpm -qa | grep ^abrt-)
+
+# https://access.redhat.com/security/cve/cve-2020-13401
+cat >/etc/sysctl.d/02-disable-accept-ra.conf <<'EOF'
```
bennerv, line 553: There aren't any other configs present by default in these files correct?  Otherwise we'd overwrite them.

bennerv (approved):
lgtm - just one question otherwise.

pkg/deploy/generator/resources_rp.go
bryanro92: we are creating the file, it does not exist prior to this pr

SudoBrendan (approved):
LGTM

PR #2025, opened 2022-03-24
Enable Managed Upgrade Operator on new clusters

hawkowl:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13776942

Enables MUO on new clusters with a default pullspec.

### Test plan for issue:

Unit tests + an E2E test

N/A

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -77,5 +77,7 @@ func DefaultOperatorFlags() OperatorFlags {
 		"aro.storageaccounts.enabled":              flagTrue,
 		"aro.workaround.enabled":                   flagTrue,
 		"aro.autosizednodes.enable":                flagFalse,
+		"rh.srep.muo.enabled":                      flagTrue,
+		"rh.srep.muo.managed":                      flagTrue,
 	}
 }
--- a/pkg/util/version/const.go
+++ b/pkg/util/version/const.go
@@ -84,3 +84,8 @@ func MdsdImage(acrDomain string) string {
 
 	return acrDomain + "/genevamdsd:master_20211223.1"
 }
+
+// MUOImage contains the location of the Managed Upgrade Operator container image
+func MUOImage(acrDomain string) string {
+	return acrDomain + "/managed-upgrade-operator:aro-b1"
+}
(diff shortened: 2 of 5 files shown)


pkg/api/defaults.go
```diff
@@ -77,5 +77,7 @@ func DefaultOperatorFlags() OperatorFlags {
 		"aro.storageaccounts.enabled":              flagTrue,
 		"aro.workaround.enabled":                   flagTrue,
 		"aro.autosizednodes.enable":                flagFalse,
+		"rh.srep.muo.enabled":                      flagTrue,
+		"rh.srep.muo.managed":                      flagTrue,
```
s-amann, line 81: This may have been answered in a different PR, but why do these two flags start with `rh.srep` instead of `aro`?

s-amann (approved):
lgtm, just one small question.

bryanro92 (approved):
overall lgtm, a couple of questions, clicking approve, but I don't want this approval to deter others from reviewing this PR as well...

pkg/util/version/const.go
```diff
@@ -84,3 +84,8 @@ func MdsdImage(acrDomain string) string {
 
 	return acrDomain + "/genevamdsd:master_20211223.1"
 }
+
+// MUOImage contains the location of the Managed Upgrade Operator container image
+func MUOImage(acrDomain string) string {
+	return acrDomain + "/managed-upgrade-operator:aro-b1"
```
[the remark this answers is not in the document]
hawkowl: This is a current yolo tag, since it's built with changes that have since been upstreamed. I felt like using the image we'd already deployed and know works was a lower risk than building what will end up being the future one (which will be a tag matching the commit hash of the MUO repo, or MUO tag).

pkg/api/defaults.go
hawkowl: Yeah, that's pretty much it. If we're adding new components from not-ARO (e.g. a new controller that deploys a Microsoft thing) then we can prefix it to differentiate it to the SRE. Although one can look at the code to find the origin of things, if a SRE only has the flag name (e.g `rh.srep.muo`, and one is in an escort on govcloud) then the prefix gives a bit of extra information to help identify where it comes from.

PR #2027, opened 2022-03-25
Commit to allow password auth for VMSS jit access

jhoreman:

13498318 SPIKE: Change the way in which we SSH to RP/Gateway VM Instances in Prod
Fixes

This will allow ARO SRE to properly control SSH access to VMSS (Gateway and RP) through approved JIT processes and forego the use of a shared private key. 

### Test plan for issue:

Request VMSS jit according to SOP (see below) and access VM from SAW VPN. 

Yes, SOP and documentation pending.


bennerv (commented):
/hold on documentation for accessing VMs via existing JIT action.  

Couple questions though
1. Can we remove the SSH public keys now too (so we don't have a 'backdoor' to get the RP)?
2. How will we access INT/Dev VMs?  Can we still use some form of JIT?

jhoreman:
> /hold on documentation for accessing VMs via existing JIT action.
> 
> Couple questions though
> 
>     1. Can we remove the SSH public keys now too (so we don't have a 'backdoor' to get the RP)?
> 
>     2. How will we access INT/Dev VMs?  Can we still use some form of JIT?

SOP/Docs here:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/235246/JIT-VMSS-SSH-Access

>     1. Can we remove the SSH public keys now too (so we don't have a 'backdoor' to get the RP)?

A: This would be preferred. I would allow for some overlap between the two access methods for some time, but ultimately once this process has been accepted we would no longer have a need for a shared key pair. 

>     2. How will we access INT/Dev VMs? Can we still use some form of JIT?

I do not believe DEV/INT have JIT access portals. We may need to leverage some type of direct access through the Azure portal for these.

bryanro92 (commented):
LGTM, straight forward change, i'd just like to touch base with our microsoft friends and make sure they don't have any issues with this architectural change

s-amann (approved):
lgtm, ssh all the RP!

jhoreman:
> LGTM, straight forward change, i'd just like to touch base with our microsoft friends and make sure they don't have any issues with this architectural change

Additional confirmation would be good. FWIW I ran this by the MSFT JIT team and this logically follows their design philosophy around this specific JIT action.

karanmagdani (commented):
Thanks for this PR. just a comment, i noticed there is a deny policy on such actions, as far as i understand, it is not applied by default and we actually have to go in and create it. Can we confirm this with MSFT if possible? https://techcommunity.microsoft.com/t5/microsoft-defender-for-cloud/eliminate-password-based-attacks-on-azure-linux-vms/ba-p/2271139

jhoreman:
> Thanks for this PR. just a comment, i noticed there is a deny policy on such actions, as far as i understand, it is not applied by default and we actually have to go in and create it. Should we confirm this with MSFT ? https://techcommunity.microsoft.com/t5/microsoft-defender-for-cloud/eliminate-password-based-attacks-on-azure-linux-vms/ba-p/2271139

Hey Karan, I believe this policy is not applied by default and we do not leverage it. This account creation happens through the waagent backplane and was successful in testing only by allowing password authentication.

bryanro92 (approved):

bryanro92:
approved, but please rebase

SudoBrendan (approved):
LGTM

bennerv (approved):

PR #2035, opened 2022-04-01
Cluster create after api 2022-04-01

25region:

Cluster creation process after the API release

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -27,8 +27,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/Azure/ARO-RP/pkg/api"
-	"github.com/Azure/ARO-RP/pkg/api/v20210901preview"
-	mgmtredhatopenshift20210901preview "github.com/Azure/ARO-RP/pkg/client/services/redhatopenshift/mgmt/2021-09-01-preview/redhatopenshift"
+	v20220401 "github.com/Azure/ARO-RP/pkg/api/v20220401"
+	mgmtredhatopenshift20220401 "github.com/Azure/ARO-RP/pkg/client/services/redhatopenshift/mgmt/2022-04-01/redhatopenshift"
 	"github.com/Azure/ARO-RP/pkg/deploy"
 	"github.com/Azure/ARO-RP/pkg/deploy/generator"
 	"github.com/Azure/ARO-RP/pkg/env"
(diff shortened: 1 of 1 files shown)


bennerv (dismissed):
This will pass in PR e2e because it runs a local RP and doesn't talk to upstream.  It won't pass in an INT deploy on master because the ARM manifest has yet to be updated and deployed hence the hold label.

s-amann (approved):
lgtm

bennerv:
@25region can you rebase and we can merge this?

pkg/util/cluster/cluster.go
```diff
@@ -448,21 +448,22 @@ func (c *Cluster) createCluster(ctx context.Context, vnetResourceGroup, clusterN
 		}
 
 		oc.Properties.WorkerProfiles[0].VMSize = api.VMSizeStandardD2sV3
+		oc.Properties.ClusterProfile.FipsValidatedModules = api.FipsValidatedModulesDisabled
```
bennerv, line 451: This should be removed.  We can create FIPs clusters anywhere.  

Also, can you set FIPs to enabled by default?  It should be in the api doc above on line 403

bennerv (changes requested):
One change we didn't catch before.

pkg/util/cluster/cluster.go
```diff
@@ -399,8 +399,9 @@ func (c *Cluster) createCluster(ctx context.Context, vnetResourceGroup, clusterN
 	oc := api.OpenShiftCluster{
 		Properties: api.OpenShiftClusterProperties{
 			ClusterProfile: api.ClusterProfile{
-				Domain:          strings.ToLower(clusterName),
-				ResourceGroupID: fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", c.env.SubscriptionID(), "aro-"+clusterName),
+				Domain:               strings.ToLower(clusterName),
+				ResourceGroupID:      fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", c.env.SubscriptionID(), "aro-"+clusterName),
+				FipsValidatedModules: "Enabled",
```
bennerv, line 404: nit  for consistency.  Should be the last one.  
```suggestion
				FipsValidatedModules: api.FipsValidatedModulesEnabled,
```

bennerv (changes requested):
just one final change and LGTM

bennerv (approved):
lgtm

PR #2040, opened 2022-04-05
Development Subscription Migration 

karanmagdani:

Fixes


karanmagdani:
A few things around keyvault needed changing in RP. This is due to the fact that we cannot have two keyvaults with the same name. our current subscription (while in a different tenant) conflicts with the new subscription.

karanmagdani:
Closed in favor of https://github.com/Azure/ARO-RP/pull/2072 due to some issues, i was not able to comment on my own PR

PR #2041, opened 2022-04-05
Skip Linux AZ Sec Pack policies from running on VMSS creation

bennerv:

Fixes VM provisioning issues with MSFT policies in place conflicting with our existing installation.  

The policy in place is not yet available in Gov cloud so we have to revisit this come end of April.

### Test plan for issue:

deploy to canary

Probably. Ross might have more information on this. 

https://aka.ms/AzSecPackAutoConfig


bryanro92 (approved):
LGTM - this is a new "feature". Our requirements are malware scanning runninig, so we still meet these requirements.

PR #2044, opened 2022-04-08
Keynotfound empty subnet sp rotation

SrinivasAtmakuri:
https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/12670363/

### DIFF
--- a/python/az/aro/azext_aro/custom.py
+++ b/python/az/aro/azext_aro/custom.py
@@ -12,7 +12,7 @@ from azure.cli.core.commands.client_factory import get_mgmt_service_client
 from azure.cli.core.commands.client_factory import get_subscription_id
 from azure.cli.core.profiles import ResourceType
 from azure.cli.core.util import sdk_no_wait
-from azure.cli.core.azclierror import FileOperationError, ResourceNotFoundError, UnauthorizedError
+from azure.cli.core.azclierror import FileOperationError, ResourceNotFoundError, UnauthorizedError, ValidationError
 from azext_aro._aad import AADManager
 from azext_aro._rbac import assign_role_to_resource, \
     has_role_assignment_on_resource
@@ -278,12 +278,18 @@ def generate_random_id():
     return random_id
 
 
-def get_network_resources_from_subnets(cli_ctx, subnets):
+def get_network_resources_from_subnets(cli_ctx, subnets, fail):
     network_client = get_mgmt_service_client(cli_ctx, ResourceType.MGMT_NETWORK)
 
     subnet_resources = set()
     for sn in subnets:
         sid = parse_resource_id(sn)
+
+        if 'resource_group' not in sid or 'name' not in sid or 'resource_name' not in sid:
+            if fail:
+                raise ValidationError(f"""(ValidationError) Failed to validate subnet '{sn}'.
+                    Please retry, if issue persists: raise azure support ticket""")
+            logger.info("Failed to validate subnet '%s'", sn)
 
         subnet = network_client.subnets.get(resource_group_name=sid['resource_group'],
                                             virtual_network_name=sid['name'],
@@ -298,7 +304,7 @@ def get_network_resources_from_subnets(cli_ctx, subnets):
     return subnet_resources
 
 
-def get_cluster_network_resources(cli_ctx, oc):
+def get_cluster_network_resources(cli_ctx, oc, fail):
     master_subnet = oc.master_profile.subnet_id
     worker_subnets = set()
 
@@ -316,11 +322,11 @@ def get_cluster_network_resources(cli_ctx, oc):
         name=master_parts['name'],
     )
 
-    return get_network_resources(cli_ctx, worker_subnets | {master_subnet}, vnet)
+    return get_network_resources(cli_ctx, worker_subnets | {master_subnet}, vnet, fail)
 
 
-def get_network_resources(cli_ctx, subnets, vnet):
-    subnet_resources = get_network_resources_from_subnets(cli_ctx, subnets)
+def get_network_resources(cli_ctx, subnets, vnet, fail):
+    subnet_resources = get_network_resources_from_subnets(cli_ctx, subnets, fail)
 
     resources = set()
     resources.add(vnet)
@@ -421,7 +427,7 @@ def resolve_rp_client_id():
 def ensure_resource_permissions(cli_ctx, oc, fail, sp_obj_ids):
     try:
         # Get cluster resources we need to assign permissions on, sort to ensure the same order of operations
-        resources = {ROLE_NETWORK_CONTRIBUTOR: sorted(get_cluster_network_resources(cli_ctx, oc)),
+        resources = {ROLE_NETWORK_CONTRIBUTOR: sorted(get_cluster_network_resources(cli_ctx, oc, fail)),
                      ROLE_READER: sorted(get_disk_encryption_resources(oc))}
     except (CloudError, HttpOperationError) as e:
         if fail:


python/az/aro/azext_aro/custom.py
```diff
@@ -271,9 +271,12 @@ def get_route_tables_from_subnets(cli_ctx, subnets):
     for sn in subnets:
         sid = parse_resource_id(sn)
 
-        subnet = network_client.subnets.get(resource_group_name=sid['resource_group'],
-                                            virtual_network_name=sid['name'],
-                                            subnet_name=sid['resource_name'])
+        if not all(key in sid.keys() for key in ('resource_group','name','resource_name')):
+            raise ResourceNotFoundError("subnet is empty")
```
bennerv, line 275: I think we need to be more descriptive here.  `subnet is empty` will confuse end users running these `az aro` commands receiving the output. Additionally, we can set recommendations in errors we raise which I think we should be doing in this specific instance.

The reasoning behind more specificity would be along the lines of customers questioning the issue with:
* What does "subnet is empty" mean? 
* Why is the subnet empty?  
* How do I fix this issue?

bennerv: Adding a bunch of context here for other reviewers regarding enrichment (which I think may be helpful for review of this)

We have an enricher factory which runs on the RP frontend process.  This runs for 10 seconds max during any create/read/update/delete operation, and queries the cluster's kube-api server to attempt to enrich the cluster doc with worker profiles, cluster version information, cluster service principal, and ingress profiles.  It is a best-effort enricher and if it fails the frontend process will continue on normally.

For our write operations in the `az aro` command module, we attempt to fetch the cluster first, which will invoke the enricher on the frontend for 10 seconds to attempt to fetch this information from the cluster directly by talking to the kube-apiserver. If it fails, it will not affect any other frontend operations from continuing, but may result in failures later down the pipeline (such as one described in the situation in this user story)

If the enricher fails to populate any worker profiles, the python code cannot ensure the required permissions on route tables and other network resources, resulting in a failure from our side and poor customer experience on a python error. 

@SrinivasAtmakuri I think we should do the following in this situation:
1. Update the error message within the `ResourceNotFoundError` to something along the lines of `Failed to validate subnet '_subetName_' 
2. Provide a recommendation of retrying, and if continuous failures, open up an Azure support ticket
3. We might want to rethink the error we're throwing here.  A `ResourceNotFound` doesn't 100% fit the bill.  I'm leaning towards a `ValidationError`, but can be swayed on this.

bennerv (dismissed):
Great work on the logic portion of this.  I think it is spot on, but we should rethink how we're communicating the failure directly to customers (via the `az aro` command module).

We may need to involve PM input on the error message (Jerome, Rahul).  

Also needs python unit tests passing.

python/az/aro/azext_aro/custom.py
bryanro92: :+1:

hawkowl: IIRC we saw this happen when a customer had poorly configured machinesets, so I think that warning and continuing is the better approach, since they couldn't get in to *fix* said machinesets.

python/az/aro/azext_aro/custom.py
```diff
@@ -273,9 +273,12 @@ def get_route_tables_from_subnets(cli_ctx, subnets):
     for sn in subnets:
         sid = parse_resource_id(sn)
 
-        subnet = network_client.subnets.get(resource_group_name=sid['resource_group'],
-                                            virtual_network_name=sid['name'],
-                                            subnet_name=sid['resource_name'])
+        if not all(key in sid.keys() for key in ('resource_group', 'name', 'resource_name')):
+            logger.warning("Failed to validate subnet %s", sn)
```
petrkotas, line 277: I see the coments have been addressed only partially. I believe the error still should be returned, as it happened. 
I also further log should be provided to offer additional guide for customer.

So I would like to see Bens comment here.
1. Log warning with additional info about retries and support tickets, also log the error
2. Raise an error ValidationError, the error happened and it damages the deployment

petrkotas (changes requested):
Hi @SrinivasAtmakuri I still see that the comments has not been addressed. Would you please look at them?

python/az/aro/azext_aro/custom.py
SrinivasAtmakuri: Added as suggested

python/az/aro/azext_aro/custom.py
```diff
@@ -274,9 +274,14 @@ def get_route_tables_from_subnets(cli_ctx, subnets):
     for sn in subnets:
         sid = parse_resource_id(sn)
 
-        subnet = network_client.subnets.get(resource_group_name=sid['resource_group'],
-                                            virtual_network_name=sid['name'],
-                                            subnet_name=sid['resource_name'])
+        if not all(key in sid.keys() for key in ('resource_group', 'name', 'resource_name')):
+            logger.error("Failed to validate subnet '%s'", sn )
```
s-amann, line 278: This extra space should be removed 
```suggestion
            logger.error("Failed to validate subnet '%s'", sn)
```

s-amann (changes requested):
python linting is complaining about the below issues

SrinivasAtmakuri:
> python linting is complaining about the below issues

Made the needed changes, thanks!

s-amann (dismissed):
lgtm

hawkowl (changes requested):
I think that erroring out is perhaps the wrong option here, or if we do, we need some way to override it. This is because this issue originated from a situation where the cluster had many problems, including broken subnets on clusters, but they could not fix that as the SP needed to be rotated. This code needs to not error out when called from https://github.com/Azure/ARO-RP/blob/075795235dfcf384876012e668c62717cf794d9c/python/az/aro/azext_aro/custom.py#L331-L341 somehow.

SrinivasAtmakuri:
@hawkowl Added an additional argument Fail, when set, the code doesn't exit on error, but prints as info and proceeds.

python/az/aro/azext_aro/custom.py
```diff
@@ -273,13 +273,20 @@ def generate_random_id():
...
 
     route_tables = set()
     for sn in subnets:
         sid = parse_resource_id(sn)
 
+        if not all(key in sid.keys() for key in ('resource_group', 'name', 'resource_name')):
+            if fail:
+                logger.error("Failed to validate subnet '%s'", sn)
+                logger.warning("Please retry, if issue presists: raise azure support ticket")
+                raise ValidationError(f"(ValidationError) Failed to validate subnet '{sn}'")
```
s-amann, line 287: nit: it is generally an anti pattern to have multiple log statements for the same 'error' as well as to log and raise an exception, can we combine this? Especially since the error log is the same text as the exception.

typo: presists -> persists
```suggestion
            if fail:
                raise ValidationError(f"(ValidationError) Failed to validate subnet '{sn}'. Please retry, if issue persists: raise azure support ticket")
```

hawkowl:
@SrinivasAtmakuri I think this is good. I will hold on merging, though, since I don't have a huge amount of experience with the CLI, and I'll poke some other FLs.

hawkowl (dismissed):

s-amann (dismissed):

petrkotas (dismissed):
Thanks @SrinivasAtmakuri for doing this PR. LGTM

python/az/aro/azext_aro/custom.py
```diff
@@ -273,13 +273,19 @@ def generate_random_id():
...
 
-def get_route_tables_from_subnets(cli_ctx, subnets):
+def get_route_tables_from_subnets(cli_ctx, subnets, fail):
     network_client = get_mgmt_service_client(cli_ctx, ResourceType.MGMT_NETWORK)
 
     route_tables = set()
     for sn in subnets:
         sid = parse_resource_id(sn)
 
+        if not all(key in sid.keys() for key in ('resource_group', 'name', 'resource_name')):
```
ventifus, line 283: I don't understand what this test is looking for. Is it checking that resource_group, name and resource_name are keys in sid? How about something like this?
```python3
        if 'resource_group' not in sid or 'name' not in sid or 'resource_name' not in sid:
```

s-amann (approved):

s-amann (approved):

hawkowl (approved):
I'm happy and this has been sitting for long enough, so, will merge.

PR #2047, opened 2022-04-08
Fix off by one error when truncating name

### LINKED ISSUE #2046: Off by one error resource group name truncation during shared dev provisioning
Found this when following [prepare a shared rp development](https://github.com/Azure/ARO-RP/blob/master/docs/prepare-a-shared-rp-development-environment.md)

When I have $RESOURCEGROUP that's longer than 15 characters, it's truncated to 15 in deploy/env-development.json when provisioning the KVs.  Below is an example from deploy/env-development.json:

```json
        {
            "name": "[concat(concat(take(resourceGroup().name,15), '-sharedKV'), '/add')]",
            "type": "Microsoft.KeyVault/vaults/accessPolicies",
```

But when running `make runlocal-rp`, I'd get errors saying it couldn't find the provisioned KVs.  This is because the code currently truncates the name to just 14 characters.  Here is the location in the code https://github.com/Azure/ARO-RP/blob/4dbe573e026a736eaa39281f1a115f1d7a3b0a68/pkg/util/cluster/cluster.go#L184

nontw:
Now it truncates to 14 instead of 15. the corresponding arm templates
truncate to 15.

Fixes #2046 

To fix the off by one error during running local-rp.  See the issue for more details.

### Test plan for issue:

Tested my local rp with the provisioned resources in Azure and it worked as expected.

No.  Just a minor bug fix.

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -181,7 +181,7 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 	if len(vnetResourceGroup) > 15 {
 		// keyvault names need to have a maximum length of 24,
 		// so we need to cut off some chars if the resource group name is too long
-		kvName = vnetResourceGroup[:14] + generator.SharedKeyVaultNameSuffix
+		kvName = vnetResourceGroup[:15] + generator.SharedKeyVaultNameSuffix
 	} else {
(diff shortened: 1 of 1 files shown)


bennerv (approved):
LGTM - checks out and remains under the max length: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftkeyvault

PR #2049, opened 2022-04-11
renamed oddly named metrics.Interface to MetricsEmitter

facchettos:
Calling an interface Interface is questionable (to say the least)

Fixes

Renames metrics.Interface to metrics.MetricsEmitter .

### Test plan for issue:
Not needed


pkg/backend/backend.go
```diff
@@ -35,7 +35,7 @@ type backend struct {
 	dbSubscriptions     database.Subscriptions
 
 	aead    encryption.AEAD
-	m       metrics.Interface
+	m       metrics.MetricsEmitter
```
m1kola, line 38: I also don't like calling an interface `Interface`, but if it is the only exported interface in a package - IMO, it is ok. It reads well: `metrics.Interface`.

We have quite a few interfaces called like that across the codebase.

I don't mind renaming them, but if we do that - I would like to avoid repetition: `metrics.MetricsEmitter` - says "metrics" twice.

m1kola (changes requested):
I think we are missing mock generation for this interface.

pkg/backend/backend.go
facchettos: And actually, I think we should split this interface into GaugeEmitter and FloatEmitter interfaces. The two different functions are never used together except in a few test files and `pkg/cosmosdb/metrics.go`  for which it would be a trivial change to update. I think that is a side effect of the weird naming.

m1kola: > If you give the package an alias when you import it

Good point.

> or say you import it with `.` which we could do for this kind of package

We only allow `.` imports of `github.com/onsi/ginkgo` and `github.com/onsi/gomega` at the moment for readability of e2e tests:

https://github.com/Azure/ARO-RP/blob/2c2c257b31dc4d2db79e942da138289cddcb2564/hack/validate-imports/validate-imports.go#L18-L26

We should avoid dot imports when possible to avoid confusion.

I think it is not a concern and we can drop duplicating "metrics".

s-amann (approved):
lgtm, I like the new name and agree moving away from `Interface` is a good move.

bryanro92 (approved):

PR #2051, opened 2022-04-11
Update the secret rotate time to 7 days during RP deploy

bennerv:

Fixes early rotation of secrets when deploying the RP

If the RP rotates secrets too early, there exists a chance where an RP instance restarts and reads in the new secret which the other instances do not have.  This will result in issues between the instances.  

This doesn't solve the issue, but decreases the likelyhood it will happen.

### DIFF
--- a/pkg/deploy/predeploy.go
+++ b/pkg/deploy/predeploy.go
@@ -25,9 +25,9 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/keyvault"
 )
 
-// Rotate the secret on every deploy of the RP iff the most recent
-// secret is less than 3 days old
-const rotateSecretAfter = time.Hour * 72
+// Rotate the secret on every deploy of the RP if the most recent
+// secret is greater than 7 days old
+const rotateSecretAfter = time.Hour * 168
 
 // PreDeploy deploys managed identity, NSGs and keyvaults, needed for main
 // deployment


s-amann (dismissed):
lgtm

pkg/deploy/predeploy.go
```diff
@@ -26,8 +26,8 @@ import (
 )
 
 // Rotate the secret on every deploy of the RP iff the most recent
```
ross-bryan, line 28: nit: while we are here, mind fixing the typo in the word `iff`?

ross-bryan: secondly - isn't the comment backwards
"rotate if most recent is less than 7 days" -> don't we only rotate if most recent is greater than 7 days?

ross-bryan (changes requested):
I think the comment for the const has logic backwards. thoughts?

SudoBrendan (approved):
LGTM

pkg/deploy/predeploy.go
facchettos: Is `iff` a typo or is it just not used properly ?

gvanderpotte: > secondly - isn't the comment backwards "rotate if most recent is less than 7 days" -> don't we only rotate if most recent is greater than 7 days?

Yes, exactly. If the delay between the now() of the predeploy and the previous pre-deploy which rotated the secret, then rotate.

gvanderpotte: > Is `iff` a typo or is it just not used properly ?

I agree, this is not an "iff" but an "if". The condition is necessary but not sufficient :)

gvanderpotte (approved):
some comment on the comment, but overall, this should help us decrease the occurences of issues

pkg/deploy/predeploy.go
bennerv: iff -> if and only if

But the above isn't necessarily true in this case as we still rotate if a secret isn't present.  Fixed.

SudoBrendan (approved):

s-amann (dismissed):
lgtm

bennerv:
Rebased to fix linter issues.  Should be good now

s-amann (approved):

jhoreman (approved):
Should drastically decrease chances of conflict, LGTM

ross-bryan (approved):
lgtm

PR #2052, opened 2022-04-12
[bugfix] give /tmp a bit more room for when the CI VM gets busy

s-amann:

[ADO WI](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13983121)

/tmp fills up on some VMs during builds/tests

### Test plan for issue:

n/a

n/a


pkg/deploy/generator/resources_dev.go
```diff
@@ -306,7 +306,10 @@ for attempt in {1..5}; do
   if [[ ${attempt} -lt 5 ]]; then sleep 10; else exit 1; fi
 done
 
-lvextend -l +50%FREE /dev/rootvg/varlv
+lvextend -l +25%FREE /dev/rootvg/tmplv
+xfs_growfs /tmp
+
+lvextend -l +25%FREE /dev/rootvg/varlv
```
bennerv, line 312: I'm not sure if this is doing what you expect.  

Let's assume there's 1GB of free space:
1. `lvextend -l +25%FREE /dev/rootvg/tmplv` -> adds 0.25GB of space to the /tmp filesystem
2. `lvextend -l +25%FREE /dev/rootvg/varlv` -> adds 25% of remaining space (0.75GB * 25% = .1875GB) to /var
3. `lvextend -l +100%FREE /dev/rootvg/homelv` -> adds all remaining free space to the /home directory `1 - 0.25 - .1875)

I think this is how it works.  If this is the case, you're going to want to do this in reverse order with the following
1. `lvextend -l +50%FREE /dev/rootvg/homelv`
2. `lvextend -l +50%FREE /dev/rootvg/tmplv`
3. `lvextend -l +100%FREE /dev/rootvg/varlv`

bennerv (commented):
Question on this.  not sure how `lvextend` works and I think the logic here might be wrong.

pkg/deploy/generator/resources_dev.go
s-amann: Ah, right. Thank you, I'll update.

bennerv: Yeah from the [man page](https://linux.die.net/man/8/lvextend)

The number can also be expressed as a percentage of the remaining free space in the Volume Group with the suffix %FREE

cadenmarchese (approved):
I like that we're using percent free so that we're not tied to a specific VM storage size. Thanks!

bryanro92 (approved):

ellis-johnson (approved):
lgtm!

bennerv (approved):

PR #1889, opened 2022-01-04
Geneva master resize

jhoreman:
### Which issue this PR addresses:

Net new creation of ARO cluster master resize geneva action

### What this PR does / why we need it:

This work will allow us to leverage geneva actions to resize the compute assigned to ARO master nodes. 

### Test plan for issue:

Functionality can be tested through both direct calls to the RP as well as the pre-approved Geneva environment mock setup

### Is there any documentation that needs to be updated for this PR?

Documentation and SOP will be provided once process (and PR) has been solidified.

PR #2053, opened 2022-04-13
Start/Restart VM Admin Endpoints For Geneva Actions

zacharyljones:
This PR created admin endpoints for starting and restarting cluster nodes. Endpoints will be targeted by new Geneva Actions (Phase 3).

[USER STORY 13911996 - Create Start VM Geneva Action](https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/13911996)
[USER STORY 13912077 - Create Restart Node Geneva Action](https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/13912077)

Note that the "Restart Node" endpoint wasn't explicitly requested, but it will be necessary for the "Graceful Restart/Redeploy Node" action that has been requested (Cordon, drain, restart, uncordon, etc.)

This PR adds admin endpoints that can be used to start or restart cluster nodes without redeploying them. An endpoint for starting a node has been requested from support, while restarting a node without deploying will be use  

### Test plan for issue:

Unit tests have been added for each new endpoint. There are currently no e2e tests for these new endpoints, but tests modeled after the "redeploy vm" e2e test could be added if deemed necessary. A separate "stop node" endpoint would likely need to be added in order to create an e2e test for starting cluster nodes.

I also ran the RP locally, created a cluster. and manually curled the startvm, restartvm, and redeployvm endpoints. I confirmed that all 3 endpoints functioned as intended by monitoring the activity log for target VMs in the Azure portal. 

No - once the Geneva Action frontends have been created, that documentation will be updated to include the functionality of these new endpoints.

### DIFF
--- a/pkg/frontend/admin_openshiftcluster_vm.go
+++ b/pkg/frontend/admin_openshiftcluster_vm.go
@@ -1,0 +1,63 @@
+package frontend
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"net/http"
+	"strings"
+
+	"github.com/gorilla/mux"
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
+)
+
+const (
+	redeployVM = iota
+	restartVM
+	startVM
+)
+
+func (f *frontend) _postAdminOpenShiftClusterVM(ctx context.Context, r *http.Request, log *logrus.Entry, vmAction int) error {
+	vars := mux.Vars(r)
+
+	vmName := r.URL.Query().Get("vmName")
+	err := validateAdminVMName(vmName)
+	if err != nil {
+		return err
+	}
+
+	resourceID := strings.TrimPrefix(r.URL.Path, "/admin")
+
+	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	switch {
+	case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
+		return api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, "", "The Resource '%s/%s' under resource group '%s' was not found.", vars["resourceType"], vars["resourceName"], vars["resourceGroupName"])
+	case err != nil:
+		return err
+	}
+
+	subscriptionDoc, err := f.getSubscriptionDocument(ctx, doc.Key)
+	if err != nil {
+		return err
+	}
+
+	a, err := f.azureActionsFactory(log, f.env, doc.OpenShiftCluster, subscriptionDoc)
+	if err != nil {
+		return err
+	}
+
+	switch {
+	case vmAction == redeployVM:
+		return a.VMRedeployAndWait(ctx, vmName)
+	case vmAction == restartVM:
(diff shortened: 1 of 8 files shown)


pkg/frontend/admin_openshiftcluster_vm.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
+)
+
+const (
+	redeployVM = iota
+	restartVM
+	startVM
+)
+
+func (f *frontend) _postAdminOpenShiftClusterVM(ctx context.Context, r *http.Request, log *logrus.Entry, vmAction int) error {
```
s-amann, line 24: Since this frontend method is in its own file now we should rename it to remove the `_` prefix. In addition we can make the name a bit clearer on its intent. What do you think about `postAdminOpenShiftClusterVMAction`?

pkg/frontend/adminactions/azureactions.go
```diff
@@ -69,3 +71,13 @@ func (a *azureActions) VMRedeployAndWait(ctx context.Context, vmName string) err
 	clusterRGName := stringutils.LastTokenByte(a.oc.Properties.ClusterProfile.ResourceGroupID, '/')
 	return a.virtualMachines.RedeployAndWait(ctx, clusterRGName, vmName)
 }
+
+func (a *azureActions) VMRestartAndWait(ctx context.Context, vmName string) error {
+	clusterRGName := stringutils.LastTokenByte(a.oc.Properties.ClusterProfile.ResourceGroupID, '/')
+	return a.virtualMachines.StartAndWait(ctx, clusterRGName, vmName)
```
s-amann, line 77: Why do we need both `VMRestartAndWait` and `VMStartAndWait` when both methods end up calling `StartAndWait`?

s-amann (changes requested):
Overall I like the direction this is going. Please see my question and refactor suggestion below.

jhoreman:
@zacharyljones Hey Zach. We have separate Start/Stop actions already defined in this PR that's close to being merged:

https://github.com/Azure/ARO-RP/pull/1889

Mind having a look to see where we may be overlapping here? I think a drain/cordon action may still be relevant.

zacharyljones:
> @zacharyljones Hey Zach. We have separate Start/Stop actions already defined in this PR that's close to being merged:
> 
> #1889
> 
> Mind having a look to see where we may be overlapping here? I think a drain/cordon action may still be relevant.

@jhoreman Ah - Thank you for pointing this out. There is no need for my PR - The restart functionality I was adding could be achieved by using the stop and start endpoints together. I am going to abandon this PR.

Does Red Hat plan on adding corresponding Geneva Actions as well? I was planning on doing that once these new admin endpoints were merged/released.

I do have a separate PR for cordon, uncordon, and drain in the works - trying to square away the e2e tests before submitting it. Is Red Hat working on any similar admin endpoints? Vikas does know that I am working on some new admin endpoints for Phase 3 Geneva Actions, but it sounds like some other people may need to be notified.

jhoreman:
> 

Yes we intend to create the associated GA endpoints as well, as soon as the aforementioned PR is merged :smiley:

zacharyljones:
Cool - I'll let Rahul and Vikas know. Thanks Jory!

zacharyljones:
Closed because of overlap with [#1889](https://github.com/Azure/ARO-RP/pull/1889)

PR #2054, opened 2022-04-13
ensure apiserverready check

kartikaysharma27:
https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14058347/

Fixes

The ensureAROOperator call in pkg/cluster/install.go (see https://github.com/Azure/ARO-RP/blob/master/pkg/cluster/install.go#L170 ) sometimes encounters an error because the API Server is not available between when the configmap is found and it runs. We should perform an APIServer Ready check, like in post-bootstrap, before we ensure the ARO Operator.

### DIFF
--- a/pkg/cluster/install.go
+++ b/pkg/cluster/install.go
@@ -167,6 +167,7 @@ func (m *manager) Install(ctx context.Context) error {
 			steps.Action(m.initializeKubernetesClients),
 			steps.Action(m.initializeOperatorDeployer), // depends on kube clients
 			steps.Condition(m.bootstrapConfigMapReady, 30*time.Minute, true),
+			steps.Condition(m.apiServersReady, 30*time.Minute, true),
 			steps.Action(m.ensureAROOperator),
 			steps.Action(m.incrInstallPhase),
 		},


SrinivasAtmakuri:
LGTM!

s-amann (approved):
lgtm

gvanderpotte (approved):
LGTM - I share the concerns by Ben. I think this is some kind of race condition with bootstrap api server being teared down along with the bootstrap node. As a followup, I would love to see metrics of how each step of the cluster install takes, so that we can build dashboards on the performance of cluster step and see if we can measure how often we hit important delays, in particular in this step which should be "immediate" in theory as Ben said "it should work"

hawkowl (approved):
The reason why this is here is that the bootstrap config being here doesn't mean that the API server is not at its proper revision, so when we go to apply the ARO Operator we can hit it being down and upgrading.

This might add up to 30 minutes in the bootstrap phase, but the API Server Ready check in post-bootstrap should correspondingly always be fast.

bennerv:
> The reason why this is here is that the bootstrap config being here doesn't mean that the API server is not at its proper revision, so when we go to apply the ARO Operator we can hit it being down and upgrading.

@hawkowl 
Why would the kube-apiserver ever be completely inaccessible or down?  Only a single kube-apiserver instance should be affected and going through revisions, so two of the three should be healthy and the api-server should be responsive.  (which would be a requirement when we check the bootstrap configmap ready since we're querying the kube-apiserver.)

PR #2059, opened 2022-04-14
redesigned the quota computation to something understandable

facchettos:
This is a redesign of the quota computations, which was absolutely horrendous. 

Fixes

Was not a really good design 

### Test plan for issue:
Should be covered by existing tests. If tests pass it should have kept the same behavior. 

No

### DIFF
--- a/pkg/api/validate/dynamic/quota.go
+++ b/pkg/api/validate/dynamic/quota.go
@@ -11,115 +11,65 @@ import (
 	"github.com/Azure/ARO-RP/pkg/api"
 )
 
-func addRequiredResources(requiredResources map[string]int, vmSize api.VMSize, count int) error {
+type azureVM struct {
+	Size      api.VMSize
+	CoreCount int
+	Family    string
+}
+
+func vmFromSize(vmType api.VMSize) (azureVM, bool) {
+	vm, ok := _vmTypesMap[vmType]
+	return vm, ok
+}
+
+// this should be considered as a const and only accessed via vmFromSize
+var _vmTypesMap map[api.VMSize]azureVM = map[api.VMSize]azureVM{
+	api.VMSizeStandardD2sV3: {Size: api.VMSizeStandardD2sV3, CoreCount: 2, Family: "standardDSv3Family"},
+
+	api.VMSizeStandardD4asV4:  {Size: api.VMSizeStandardD4asV4, CoreCount: 4, Family: "standardDASv4Family"},
+	api.VMSizeStandardD8asV4:  {Size: api.VMSizeStandardD8asV4, CoreCount: 8, Family: "standardDASv4Family"},
+	api.VMSizeStandardD16asV4: {Size: api.VMSizeStandardD16asV4, CoreCount: 16, Family: "standardDASv4Family"},
+	api.VMSizeStandardD32asV4: {Size: api.VMSizeStandardD32asV4, CoreCount: 32, Family: "standardDASv4Family"},
+
+	api.VMSizeStandardD4sV3:  {Size: api.VMSizeStandardD4sV3, CoreCount: 4, Family: "standardDSv3Family"},
+	api.VMSizeStandardD8sV3:  {Size: api.VMSizeStandardD8sV3, CoreCount: 8, Family: "standardDSv3Family"},
+	api.VMSizeStandardD16sV3: {Size: api.VMSizeStandardD16sV3, CoreCount: 16, Family: "standardDSv3Family"},
+	api.VMSizeStandardD32sV3: {Size: api.VMSizeStandardD32sV3, CoreCount: 32, Family: "standardDSv3Family"},
+
+	api.VMSizeStandardE4sV3:   {Size: api.VMSizeStandardE4sV3, CoreCount: 4, Family: "standardESv3Family"},
+	api.VMSizeStandardE8sV3:   {Size: api.VMSizeStandardE8sV3, CoreCount: 8, Family: "standardESv3Family"},
+	api.VMSizeStandardE16sV3:  {Size: api.VMSizeStandardE16sV3, CoreCount: 16, Family: "standardESv3Family"},
+	api.VMSizeStandardE32sV3:  {Size: api.VMSizeStandardE32sV3, CoreCount: 32, Family: "standardESv3Family"},
+	api.VMSizeStandardE64isV3: {Size: api.VMSizeStandardE64isV3, CoreCount: 64, Family: "standardESv3Family"},
+	api.VMSizeStandardE64iV3:  {Size: api.VMSizeStandardE64iV3, CoreCount: 64, Family: "standardESv3Family"},
+
+	api.VMSizeStandardF4sV2:  {Size: api.VMSizeStandardF4sV2, CoreCount: 4, Family: "standardFSv2Family"},
+	api.VMSizeStandardF8sV2:  {Size: api.VMSizeStandardF8sV2, CoreCount: 8, Family: "standardFSv2Family"},
+	api.VMSizeStandardF16sV2: {Size: api.VMSizeStandardF16sV2, CoreCount: 16, Family: "standardFSv2Family"},
+	api.VMSizeStandardF32sV2: {Size: api.VMSizeStandardF32sV2, CoreCount: 32, Family: "standardFSv2Family"},
+	api.VMSizeStandardF72sV2: {Size: api.VMSizeStandardF72sV2, CoreCount: 72, Family: "standardFSv2Family"},
+
+	api.VMSizeStandardM128ms: {Size: api.VMSizeStandardM128ms, CoreCount: 128, Family: "standardMSFamily"},
+	api.VMSizeStandardG5:     {Size: api.VMSizeStandardG5, CoreCount: 32, Family: "standardGFamily"},
+	api.VMSizeStandardGS5:    {Size: api.VMSizeStandardGS5, CoreCount: 32, Family: "standardGFamily"},
+
+	api.VMSizeStandardL4s:    {Size: api.VMSizeStandardL4s, CoreCount: 4, Family: "standardLsv2Family"},
+	api.VMSizeStandardL8s:    {Size: api.VMSizeStandardL8s, CoreCount: 8, Family: "standardLsv2Family"},
+	api.VMSizeStandardL16s:   {Size: api.VMSizeStandardL16s, CoreCount: 16, Family: "standardLsv2Family"},
+	api.VMSizeStandardL32s:   {Size: api.VMSizeStandardL32s, CoreCount: 32, Family: "standardLsv2Family"},
+	api.VMSizeStandardL8sV2:  {Size: api.VMSizeStandardL8sV2, CoreCount: 8, Family: "standardLsv2Family"},
+	api.VMSizeStandardL16sV2: {Size: api.VMSizeStandardL16sV2, CoreCount: 16, Family: "standardLsv2Family"},
+	api.VMSizeStandardL32sV2: {Size: api.VMSizeStandardL32sV2, CoreCount: 32, Family: "standardLsv2Family"},
+	api.VMSizeStandardL48sV2: {Size: api.VMSizeStandardL48sV2, CoreCount: 48, Family: "standardLsv2Family"},
+	api.VMSizeStandardL64sV2: {Size: api.VMSizeStandardL64sV2, CoreCount: 64, Family: "standardLsv2Family"},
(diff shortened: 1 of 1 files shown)


pkg/api/openshiftcluster.go
```diff
@@ -312,6 +312,57 @@ const (
 	VMSizeStandardL64sV2 VMSize = "Standard_L64s_v2"
 )
 
+type AzureVM struct {
+	Type      VMSize
+	CoreCount int
+	Family    string
+}
+
+var vmTypesMap map[VMSize]AzureVM = map[VMSize]AzureVM{
```
m1kola, line 321: In this package we define internal representation of our objects. This code is validation-specific. It should not be here.

pkg/api/validate/dynamic/quota.go
```diff
@@ -11,114 +11,15 @@ import (
 	"github.com/Azure/ARO-RP/pkg/api"
 )
 
-func addRequiredResources(requiredResources map[string]int, vmSize api.VMSize, count int) error {
+func addRequiredResources(requiredResources map[string]int, vm api.AzureVM, count int) error {
+	if vm.Type == "" || vm.CoreCount < 1 || vm.Family == "" {
+		return fmt.Errorf("unexpected node VMSize %s", vm.Type)
```
m1kola, line 16: I believe at this point we should already know that VM size is already valid so there is no need in this check.

Instead we can create an unit test to check that `vmTypesMap` contains all the supported VM sizes to make sure that we do not forget to udpate it when we add new VM sizes.

pkg/api/openshiftcluster.go
```diff
@@ -312,6 +312,57 @@ const (
...
+	VMSizeStandardL8sV2:  {Type: VMSizeStandardL8sV2, CoreCount: 8, Family: "standardLsv2Family"},
+	VMSizeStandardL16sV2: {Type: VMSizeStandardL16sV2, CoreCount: 16, Family: "standardLsv2Family"},
+	VMSizeStandardL32sV2: {Type: VMSizeStandardL32sV2, CoreCount: 32, Family: "standardLsv2Family"},
+	VMSizeStandardL48sV2: {Type: VMSizeStandardL48sV2, CoreCount: 48, Family: "standardLsv2Family"},
+	VMSizeStandardL64sV2: {Type: VMSizeStandardL64sV2, CoreCount: 64, Family: "standardLsv2Family"},
+}
+
+func VMFromTypeName(vmType VMSize) AzureVM {
+	return vmTypesMap[vmType]
+}
```
m1kola, line 364: Not sure that I see a point in this wrapper. Especially if we move the map to the validation code.

m1kola (dismissed):

pkg/api/openshiftcluster.go
facchettos: This wrapper is to simulate a const, so we can't modify the map.

s-amann (dismissed):
lgtm, easier to understand!

pkg/api/validate/dynamic/quota.go
```diff
@@ -133,16 +83,18 @@ func (dv *dynamic) ValidateQuota(ctx context.Context, oc *api.OpenShiftCluster)
 	}
 
 	requiredResources := map[string]int{}
-	err := addRequiredResources(requiredResources, oc.Properties.MasterProfile.VMSize, 3)
-	if err != nil {
-		return err
+	vm, ok := vmFromTypeName(oc.Properties.MasterProfile.VMSize)
+	if !ok {
+		return fmt.Errorf("unexpected node VMSize %s", oc.Properties.MasterProfile.VMSize)
```
cadenmarchese, line 88: Can we return the specific profile that is unexpected?
```suggestion
		return fmt.Errorf("unexpected MasterProfile VMSize %s", oc.Properties.MasterProfile.VMSize)
```

pkg/api/validate/dynamic/quota.go
```diff
@@ -133,16 +83,18 @@ func (dv *dynamic) ValidateQuota(ctx context.Context, oc *api.OpenShiftCluster)
...
 	}
+	addRequiredResources(requiredResources, vm, 3)
 	//worker node resource calculation
 	for _, w := range oc.Properties.WorkerProfiles {
-		err = addRequiredResources(requiredResources, w.VMSize, w.Count)
-		if err != nil {
-			return err
+		vm, ok := vmFromTypeName(w.VMSize)
+		if !ok {
+			return fmt.Errorf("unexpected node VMSize %s", w.VMSize)
```
cadenmarchese, line 95:
```suggestion
			return fmt.Errorf("unexpected WorkerProfile VMSize %s", w.VMSize)
```
Also, should we use `unsupported` here?

```suggestion
			return fmt.Errorf("unsupported WorkerProfile VMSize %s", w.VMSize)
```

s-amann (dismissed):

SudoBrendan (approved):
LGTM

pkg/api/validate/dynamic/quota.go
```diff
@@ -11,115 +11,65 @@ import (
...
-func addRequiredResources(requiredResources map[string]int, vmSize api.VMSize, count int) error {
+type azureVM struct {
+	Size      api.VMSize
+	CoreCount int
+	Family    string
+}
+
+func vmFromSize(vmType api.VMSize) (azureVM, bool) {
+	vm, ok := _vmTypesMap[vmType]
+	return vm, ok
```
darthhexx, line 22: This function adds no value beyond what the map lookup does and adds unnecessary complexity, reduces readability, and adds overhead. We should change `var _vmTypesMap` to `var vmTypesMap` and replace both
```sh
vm, ok := vmFromSize(oc.Properties.MasterProfile.VMSize)
```
with
```sh
vm, ok := vmTypesMap[vmType]
```

darthhexx (changes requested):
This is a much better approach, bar the one nit on unnecessary complexity.

pkg/api/validate/dynamic/quota.go
facchettos: Yeah I thought that would make it more obvious that it was to be considered as a const, but it hasn't had the effect I was hoping for, removing it :)

darthhexx (approved):
lgtm

petrkotas (approved):
@facchettos thanks for the cleanup.

pkg/api/validate/dynamic/quota.go
```diff
@@ -11,115 +11,59 @@ import (
...
-		requiredResources["cores"] += (count * 32)
-	case api.VMSizeStandardL48sV2:
-		requiredResources["standardLsv2Family"] += (count * 48)
-		requiredResources["cores"] += (count * 48)
-	case api.VMSizeStandardL64sV2:
-		requiredResources["standardLsv2Family"] += (count * 64)
-		requiredResources["cores"] += (count * 64)
-	default:
-		//will only happen if pkg/api verification allows new VMSizes
-		return fmt.Errorf("unexpected node VMSize %s", vmSize)
```
m1kola, line 120: > where is the source of truth for the vms we do support ? Is it the list of const in the api package ?

[Here](https://github.com/Azure/ARO-RP/blob/5668a316afae97b690871edac2c59eb8f84efacb/pkg/api/validate/vm.go#L14-L72) is the API version agnostic validation. I would say this is source of truth: if a VM size you specify is missing from here - you won't be able to create a cluster using this vm size no mater what API version you are using.

We also have lists of constants:
* In our internal representation [in `pkg/api`](https://github.com/Azure/ARO-RP/blob/5668a316afae97b690871edac2c59eb8f84efacb/pkg/api/openshiftcluster.go#L274-L313) which should cointain all the supported VMs as well.
* In each external representation/API version in `pkg/api/v*` pacakges. These lists can be shorter. [Example](https://github.com/Azure/ARO-RP/blob/5668a316afae97b690871edac2c59eb8f84efacb/pkg/api/v20200430/openshiftcluster.go#L134-L156) of a shorter list.

> Also the original function had the same kind of checking anyways, in the default case.

I'm sorry, I think I made a wrong suggestion here. We do need this check. I still think that an unit test can be useful here and will help us to catch this issue at testing time instead of runtime. But this is out of scope for your refactoring. If you like, you can look into adding an unit test in a separate PR ;)

pkg/api/validate/dynamic/quota.go
```diff
@@ -133,16 +77,18 @@ func (dv *dynamic) ValidateQuota(ctx context.Context, oc *api.OpenShiftCluster)
...
 	//worker node resource calculation
 	for _, w := range oc.Properties.WorkerProfiles {
-		err = addRequiredResources(requiredResources, w.VMSize, w.Count)
-		if err != nil {
-			return err
+		vm, ok := vmTypesMap[w.VMSize]
+		if !ok {
+			return fmt.Errorf("unsupported WorkerProfile VMSize %s", w.VMSize)
 		}
+		addRequiredResources(requiredResources, vm, w.Count)
```
m1kola, line 91: Can we encapsulate `vmTypesMap` lookups inside `addRequiredResources`? Basically revert changes to `ValidateQuota` and only modify `addRequiredResources`.

This way we can move instantiation of `vmTypesMap` into `addRequiredResources` (currently at the package level). You raised a good point [here](https://github.com/Azure/ARO-RP/pull/2059#discussion_r850515468) on mutability of `vmTypesMap`. This change will reduce the context of this map and as a result will reduce a chance of accidentally mutating it.

The slight downside is that we will have to have a more generic "unexpected node VMSize" (as currently in master) error instead of error pointing to `WorkerProfile` or `MasterProfile`. But I think it is a minor inconvinience: these errors are not propogated to end customer and SREs will still be easily undderstand the issue. Ideally we should catch it at unit testing time (out of scope of this refactoring) instead of runtime as discussed earlier.

s-amann (approved):
lgtm

pkg/api/validate/dynamic/quota.go
```diff
@@ -11,114 +11,64 @@ import (
...
 
-func addRequiredResources(requiredResources map[string]int, vmSize api.VMSize, count int) error {
+type azureVM struct {
+	Size      api.VMSize
+	CoreCount int
+	Family    string
+}
+
+func addRequiredResources(requiredResources map[string]int, vmSize api.VMSize, role string, count int) error {
+	var vmTypesMap map[api.VMSize]azureVM = map[api.VMSize]azureVM{
```
m1kola, line 21:
```suggestion
	vmTypesMap := map[api.VMSize]azureVM{
```

or even this so we don't have to defien a single use struct

```suggestion
	vmTypesMap := map[api.VMSize]struct {
		CoreCount int
		Family    string
	}{
```

pkg/api/validate/dynamic/quota.go
```diff
@@ -11,114 +11,64 @@ import (
 	"github.com/Azure/ARO-RP/pkg/api"
 )
 
-func addRequiredResources(requiredResources map[string]int, vmSize api.VMSize, count int) error {
+type azureVM struct {
+	Size      api.VMSize
```
m1kola, line 15: I think this field is unused.

pkg/api/validate/dynamic/quota.go
```diff
@@ -133,13 +83,14 @@ func (dv *dynamic) ValidateQuota(ctx context.Context, oc *api.OpenShiftCluster)
...
-	err := addRequiredResources(requiredResources, oc.Properties.MasterProfile.VMSize, 3)
+	err := addRequiredResources(requiredResources, oc.Properties.MasterProfile.VMSize, "MasterProfile", 3)
 	if err != nil {
 		return err
 	}
+
 	//worker node resource calculation
 	for _, w := range oc.Properties.WorkerProfiles {
-		err = addRequiredResources(requiredResources, w.VMSize, w.Count)
+		err = addRequiredResources(requiredResources, w.VMSize, "WorkerProfile", w.Count)
```
m1kola, line 93: As I mentioned ealier - this error is not for customers and it is not a big deal if we just have `unexpected node VMSize %s"`: the issue will be very clear - we miss vm type in our code.

But if we want to have more specificity, we should probably call the argument `path` instead of `role` and do `fmt.Sprintf("properties.workerProfiles[%d].vmSize", i)` for workers and `"properties.masterProfile.vmSize"` for master.

This is what we do for customer facing errors: we include path to the field which has an invalid value.

But I would just drop it. No need for more specificity here, IMO.

m1kola (commented):
Almost good to go. There are few small clean ups we can do before we merge.

m1kola (dismissed):
Looks good to me. Let's merge once CI is green.

PR #2060, opened 2022-04-17
Enable first basic linters in ARO

devamanv:
The PR contains changes to enable a few basic linters and fixes for the issues that were reported after enabling those. 
**_Note: Although the PR contains a lot of files, but those are just simple changes to remove unnecessary empty lines and more linters will enabled in a follow-up PR._** 

Fixes a part of [Translate go code tips document to code linters](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/9816834/)

The idea is to expedite the feedback loop on the reviews to some extent so that Code style items are checked automatically.

N/A

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,8 +1,40 @@
 run:
   timeout: 5m
+  # Allowed values: readonly|vendor|mod
+  # By default, it isn't set.
+  modules-download-mode: readonly
+  skip-dirs:
+    - vendor/portal
 
 issues:
   exclude-rules:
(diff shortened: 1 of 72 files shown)


s-amann (approved):
lgtm. It might be cool in the future to customize the initialisms to include things like ARM, ARO, MSFT, IcM etc. 

https://staticcheck.io/docs/configuration/options/#initialisms

ross-bryan (approved):

PR #2062, opened 2022-04-19
Additional gateway tests

s-amann:

[ADO WI](https://msazure.visualstudio.com/AzureRedHatOpenShift/_boards/board/t/Azure%20Red%20Hat%20OpenShift/Stories/?workitem=13703173)

Increases coverage for pkg/gateway

### Test plan for issue:
n/a

n/a


pkg/gateway/gateway.go
```diff
@@ -102,7 +102,9 @@ func NewGateway(ctx context.Context, env env.Core, baseLog, accessLog *logrus.En
 			return nil, err
 		}
 
-		domains = append(domains, u.Hostname())
+		if u.Hostname() != "" {
```
ross-bryan, line 105: I like this conditional, but i'm thinking lets revisit the logic. we should probably log or return an error if our hostname is empty, it is a critical error that would result in the gateway not functioning in a way that is expected

ross-bryan: I think it is hard requirement both ActiveDirectoryEndpoint and ResourceManagerEndpoint hostname is not empty

ross-bryan: and then adjust tests accordingly, imo gateway is not functioning properly if those fields are missing

pkg/gateway/gateway_test.go
```diff
@@ -4,11 +4,177 @@ package gateway
...
+	for _, tt := range []struct {
+		name              string
+		gatewayDomains    string
+		acrResourceID     string
+		mocks             func(env *mock_env.MockCore)
+		expectedAllowList []string
+		wantErr           error
+	}{
+		{
+			name: "Empty gatewayDomains, empty Environment, no acrResourceID so the allowList is empty",
```
ross-bryan, line 43: this would be updated per my previous comment, we'll want an err because our gateway is missing important domains

ross-bryan (changes requested):
initial glance is looking good. submitting some feedback now, will do another round of review later

pkg/gateway/gateway_test.go
```diff
@@ -4,11 +4,177 @@ package gateway
...
+
+			if err != nil {
+				if tt.wantErr == nil {
+					t.Errorf("Got an unexpected error: %s", err.Error())
+				}
+				if tt.wantErr.Error() != err.Error() {
+					t.Errorf("The expected error: %s did not match the actual error: %s", tt.wantErr.Error(), err.Error())
+				}
+				return
+			}
```
bennerv, line 126: This doesn't cover the case where we don't get an error but expect one.  I suggest that you change the type of `wantErr` to a `string` and do soemthing like:

```suggestion
			gtwy, err := NewGateway(ctx, env, baseLog, baseLog, db, httpsl, httpl, tt.acrResourceID, tt.gatewayDomains, metrics)
				if err != nil && err.Error() != tt.wantErr {
				t.Errorf("got error '%v', wanted error '%v'", err, tt.wantErr)
			}

			if err == nil && tt.wantErr != "" {
				t.Errorf("did not get an error, but wanted error '%v'", tt.wantErr)
			}
```

bennerv: forgive the indentation issues here.

bennerv (changes requested):

pkg/gateway/gateway.go
s-amann: Going through `environments.go` I see that we have it configured so I also agree if it is missing we should error out. I will make the required changes to the gateway.

pkg/gateway/gateway_test.go
```diff
@@ -4,11 +4,175 @@ package gateway
...
+
+			gtwyValue, _ := gtwy.(*gateway)
+
+			if len(gtwyValue.allowList) != len(tt.expectedAllowList) {
+				t.Errorf("expected %d but got %d domains in the allow list", len(tt.expectedAllowList), len(gtwyValue.allowList))
+			}
+			for _, expectedDomain := range tt.expectedAllowList {
+
```
bennerv, line 132: nit: remove new line at beginning of block statements (if, for, switch, etc.)

bennerv (changes requested):

bennerv (approved):
lgtm - fixes linting issues with the files as well

pkg/gateway/gateway_test.go
```diff
@@ -4,11 +4,174 @@ package gateway
...
+			httpsl, _ := utilnet.Listen("tcp", ":8443", SocketSize)
+
+			env := mock_env.NewMockCore(controller)
+			tt.mocks(env)
+
+			gtwy, err := NewGateway(ctx, env, baseLog, baseLog, db, httpsl, httpl, tt.acrResourceID, tt.gatewayDomains, metrics)
+
+			if tt.wantErr != "" {
+				if err == nil {
+					t.Errorf("Expected error: '%s' but did not receive any errors", tt.wantErr)
```
SudoBrendan, line 115: nit: `%q` == `'%s'` - this happens in a few places. Feel free to leave as-is.

pkg/gateway/gateway_test.go
```diff
@@ -4,11 +4,174 @@ package gateway
...
+			}
+			if len(gtwyValue.gateways) != 0 {
+				t.Error("gateways should initially be empty")
+			}
+			if gtwyValue.m != metrics {
+				t.Error("metrics was not set properly")
+			}
+			if gtwyValue.ready.Load() != true {
+				t.Error("a new gateway should start as ready")
+			}
```
SudoBrendan, line 170: These feel like they should be individual, named tests instead of part of the `for` cases, or is it important to test these default behaviors on every iteration?

SudoBrendan (approved):
LGTM! I left a few notes you can use or ignore.

pkg/gateway/gateway_test.go
s-amann: nice! I will update to use this, thank you.

pkg/gateway/gateway_test.go
s-amann: I don't have any objection to making these individual tests, I'll work on a refactor.

SudoBrendan (dismissed):
LGTM

pkg/gateway/gateway_test.go
```diff
@@ -4,11 +4,238 @@ package gateway
...
 
 	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+	mock_database "github.com/Azure/ARO-RP/pkg/util/mocks/database"
+	mock_env "github.com/Azure/ARO-RP/pkg/util/mocks/env"
+	mock_metrics "github.com/Azure/ARO-RP/pkg/util/mocks/metrics"
+	utilnet "github.com/Azure/ARO-RP/pkg/util/net"
 )
 
+func TestNewGatway(t *testing.T) {
```
jhoreman, line 24: The smallest nit in the world. Needs an "e". Otherwise lgtm and thanks for adding all these tests.

jhoreman (approved):
LGTM, thanks

ross-bryan (approved):
lgtm

bennerv (approved):
lgtm

PR #2063, opened 2022-04-19
Pulled client id, client secret, and service principal from env vars for doing local dev rp deployments

coreyperkins:
We added a conditional statement @line 141 to use env.IsLocalDevelopmentMode() and pulled AZURE_CLIENT_ID, AZURE_CLIENT_SECRET, and AZURE_SERVICE_PRINCIPAL_ID from local environment variables that are setup when setting up the shared RP env. 

https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14178980

Without this adjustment we were unable to create a aro cluster via the RP running locally. We have used it many times in the last couple weeks to create clusters for development.

Documentation for this item and many others related to setting up the shared RP and running the local RP for dev are incoming on another PR.

@karanmagdani -- could you give me a second look here, any advice, and provide any intel from your own experiences for other reviews?

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -134,15 +134,29 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 		return fmt.Errorf("fp service principal id is not found")
 	}
 
-	c.log.Infof("creating AAD application")
-	appID, appSecret, err := c.createApplication(ctx, "aro-"+clusterName)
-	if err != nil {
-		return err
+	appID := ""
+	appSecret := ""
+	spID := ""
+
+	if env.IsLocalDevelopmentMode() {
+		// for local dev the AAD application is setup via the ./docs/prepare-a-shared-rp-development-environment.md instructions, simply retrieve the values from the env
+		appID = os.Getenv("AZURE_CLIENT_ID")
+		appSecret = os.Getenv("AZURE_CLIENT_SECRET")
+		spID = os.Getenv("AZURE_SERVICE_PRINCIPAL_ID")
+	} else {
+		appID, appSecret, err = c.createApplication(ctx, "aro-"+clusterName)
+		if err != nil {
+			return err
+		}
+
+		spID, err = c.createServicePrincipal(ctx, appID)
+		if err != nil {
+			return err
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


karanmagdani:
Hey @coreyperkins ! Thanks for the PR. Though i'm not sure that we want to merge these changes. While we made those changes to workaround the consent issues, the flow where development clusters create a new AAD application is a requirement. Here are some of the reasons below:

1. One user can create only one cluster. It will fail with an error if user tries to create a cluster while one already exists in their documentdb. I don't exactly remember where that code snippet is but there is a check on the db for unique service principal.
2. In general, we don't want to be creating development clusters with the same SP that is used to create the RP. This can cause some unnecessary side effects. 

These would be some of the permissions we would need to add to the tooling AAD Application to allow it to create new applications:

- Azure Active Directory Graph (2) [Currently used AAD Graph, deprecates in June 2022]
   - Application.Read.All
   - Application.ReadWrite.OwnedBy
- Microsoft Graph (2) [for use once we migrate to microsoft graph]
   - Application.Read.All
   - Application.ReadWrite.OwnedBy

CC @ross-bryan @bennerv in case i'm missing something

coreyperkins:
> Azure Active Directory Graph (2) [Currently used AAD Graph, deprecates in June 2022]
>   
>   * Application.Read.All
>   * Application.ReadWrite.OwnedBy
> 
> * Microsoft Graph (2) [for use once we migrate to microsoft graph]
>   
>   * Application.Read.All
>   * Application.ReadWrite.OwnedBy

I now understand that the creation of the AAD application in development is a requirement when deploying a cluster via the RP. When going through the documentation to create the shared RP env there is a step to create the AAD application manually (as well as to add graph permissions). It looks like this is necessary for role definitions and assignments step. But, per @karanmagdani above we should not be using the same SP to create the development cluster. I also understand that now as well.

The problem for us is that due our lack of permissions on the tenant we share with MS & RH we are unable to create an AAD application as part of creating the development cluster. How should we handle this scenario without adjusting code each time? Should there be some option to bypass creation if necessary? Please advise.

Thanks for the feedback!

coreyperkins:
> 

We are currently using this branch for local dev and just merging in changes from upstream daily (or as needed). I am good with creating a new branch on our end specifically for local dev and documenting which changes are specific to our scenario. Closing PR.

PR #2067, opened 2022-04-19
Added serviceEndpoints to rp-development-predeploy to match what is in rp-production-predeploy.

coreyperkins:

https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14178476

This gets us passed this error when deploying the shared RP env and gets the template updated to match production.

{"statusCode":"BadRequest","serviceRequestId":null,"statusMessage":"{\"error\":{\"code\":\"SubnetsHaveNoServiceEndpointsConfigured\",\"message\":\"Subnets rp-subnet of virtual network /subscriptions/26c7e39e-2dfa-4854-90f0-6bc88f7a0fb8/resourceGroups/v5-eastus/providers/Microsoft.Network/virtualNetworks/rp-vnet do not have ServiceEndpoints for Microsoft.Storage resources configured...

### Test plan for issue:

This has been tested multiple times doing shared RP setups from localhost.

Documentation for this and other items related to the shared RP will be included in a separate PR.

@karanmagdani -- we ran into this a while back, please advise. Thanks!


karanmagdani:
Thanks for the PR @coreyperkins ! This is super useful because this was previously an undocumented step. 

If i remember correctly, we only added `Microsoft.Storage` service endpoint, why are we adding the other two here?

coreyperkins:
> Thanks for the PR @coreyperkins ! This is super useful because this was previously an undocumented step.
> 
> If i remember correctly, we only added `Microsoft.Storage` service endpoint, why are we adding the other two here?

@karanmagdani If I recall correctly we were thinking we'd be better off matching the production code, but I don't see where we did any tests on the additional two service endpoints being added. Would you prefer to go with the minimalist approach (and adjust in the future when/if needed) or go ahead and match production. We are indifferent but will heed your advice!

coreyperkins:
@karanmagdani took the ServiceEndpoints down to just storage. Thanks!

m1kola (changes requested):
I did not look into why we need this, but wanted draw attention to the fact that `deploy/rp-development-predeploy.json` is generated and `make generate` will override manual changes.

See this for details:

https://github.com/Azure/ARO-RP/blob/5668a316afae97b690871edac2c59eb8f84efacb/pkg/deploy/generator/templates_rp.go#L241-L308

coreyperkins:
Closing in favor of a fresh PR: https://github.com/Azure/ARO-RP/pull/2214. This one stayed out so long that squashing became tedious.

PR #2068, opened 2022-04-19
Update StorageAccount Controller To Manage Default Action

bennerv:

- The storage account controller manages the default action on storage accounts
- PUCM no longer sets the default action to Deny for storage accounts 
- Refactor and expand test coverage of storage account + storage account controller

Now operator flags correctly map to storage account controller management

### Test plan for issue:

Work in progress.  Going to deploy a full-service RP and confirm that it successfully manages the storage account after updating the operator.  

No - present in `doc.go` in the PR.


pkg/operator/controllers/storageaccounts/storageaccount_controller.go
```diff
@@ -76,8 +79,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return reconcile.Result{}, err
 	}
 
-	if !instance.Spec.OperatorFlags.GetSimpleBoolean(controllerEnabled) {
-		// controller is disabled
+	// If disabled or managed is missing, do nothing
+	var managed *bool
+	switch strings.ToUpper(instance.Spec.OperatorFlags.GetWithDefault(controllerManaged, "")) {
```
bennerv, line 84: we should treat the `managed` operator flag as a tri-state bool.  essentially if it is set to true or false, it represents those values.  Else it's `nil` and we don't do anything.

pkg/operator/controllers/storageaccounts/storageaccount_controller_test.go
```diff
@@ -0,0 +1,132 @@
...
+						controllerEnabled: "true",
+						controllerManaged: "true",
+					},
+					AZEnvironment: "AzurePublicCloud",
+					ResourceID:    resourceId,
+				},
+			},
+			wantErr: `secrets "azure-credentials" not found`,
+		},
+		// TODO - finish unit tests once we can mock clusterauthorizer.NewAzRefreshableAuthorizer
```
bennerv, line 111: We don't have a good way to currently mock the `clusterauthorizer.NewAzRefreshableAuthorizer` so I skipped additional tests here.  It is still an increase in coverage though.

pkg/operator/controllers/storageaccounts/storageaccounts.go
```diff
@@ -37,6 +37,13 @@ func (r *reconcileManager) reconcileAccounts(ctx context.Context) error {
 		rc.Spec.Storage.Azure.AccountName,
 	}
 
+	// If managed=true, set default action to Deny
+	// If managed=false, set default action to Allow
+	defaultAction := mgmtstorage.DefaultActionAllow
+	if managed {
```
bennerv, line 43: If we don't manage the storage account (which means the default action is Allow), we still set the `VirtualNetworkRules` as it simplifies the logic and won't matter as it's allowed anyway.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -201,6 +201,13 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 	// SetDefaults will set defaults on cluster document
 	api.SetDefaults(doc)
 
+	// TODO: remove once we develop using a VPN and traffic comes from a known location
+	// Storage account network lockdown should be unmanaged in local dev mode
+	// so the defaultAction is allow on the storage account
+	if f.env.IsLocalDevelopmentMode() {
+		doc.OpenShiftCluster.Properties.OperatorFlags["aro.storageaccounts.managed"] = "false"
+	}
```
bennerv, line 209: This is dirty and I don't like it.  managed has to be false so the defaultAction is Allow though and I don't know where else to put it without adding some extra indirection.

PR #2073, opened 2022-04-20
Development subscription migration

karanmagdani:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/10695313

This PR covers changes that were made to facilitate migration to the new subscription. the changes in RP are due to the fact that we cannot have conflicting keyvault names. A -dev prefix was added to all keyvaults which required a change in the RP.

Changes made where required and build out doc available here https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/241796/-WIP-Development-Environment


deploy/env-development.json
```diff
@@ -241,7 +241,7 @@
             ]
         },
         {
-            "name": "[concat(take(resourceGroup().name,15), '-sharedKV')]",
+            "name": "[concat(take(resourceGroup().name,10), '-dev-sharedKV')]",
```
rolandmkunkel, line 244: I think going to 11 would also work here

pkg/deploy/generator/resources_dev.go
```diff
@@ -509,11 +509,11 @@ rm cron
 }
 
 const (
-	sharedKeyVaultName          = "concat(take(resourceGroup().name,15), '" + SharedKeyVaultNameSuffix + "')"
+	sharedKeyVaultName          = "concat(take(resourceGroup().name,10), '" + SharedKeyVaultNameSuffix + "')"
 	sharedDiskEncryptionSetName = "concat(resourceGroup().name, '" + SharedDiskEncryptionSetNameSuffix + "')"
 	sharedDiskEncryptionKeyName = "concat(resourceGroup().name, '-disk-encryption-key')"
-
-	SharedKeyVaultNameSuffix          = "-sharedKV"
+	// Conflicts with current development subscription. cannot have two keyvaults with same name
```
rolandmkunkel, line 515: Have you looked into moving the keyvault instead? 
https://docs.microsoft.com/en-us/azure/key-vault/general/move-subscription

karanmagdani: Thanks for the response @rolandmkunkel ! We cannot move it right now because that would impact our current workflow. If i move keyvault from current subscription, developers will be blocked from cluster creation

deploy/env-development.json
karanmagdani: Yep i agree, to be exact, it is 11, but i like round numbers :D i can change it if we want to make it 11

pkg/deploy/generator/resources_dev.go
rolandmkunkel: How often do we expect to move it in future? I think it might be worth thinking about moving it instead, with a heads-up notice for the developers

karanmagdani: If i move the keyvaults today, developers cannot create clusters anymore. after looking at the portal, i noticed three dev clusters were created today. 

My attempt is to do a 0 impact migration, and switching to old subscription in case something goes wrong is a safeguard protecting it

Makefile
```diff
@@ -179,6 +179,14 @@ test-python: pyenv az
 
 admin.kubeconfig:
 	hack/get-admin-kubeconfig.sh /subscriptions/${AZURE_SUBSCRIPTION_ID}/resourceGroups/${RESOURCEGROUP}/providers/Microsoft.RedHatOpenShift/openShiftClusters/${CLUSTER} >admin.kubeconfig
+	$(MAKE) update-admin.kubeconfig
+
+# ! Temporary solution till dns can be migrated to new subscription
+PUBLIC_IP_ADDR = $(shell az network dns record-set a show -g ${RESOURCEGROUP} --zone-name ${LOCATION}.${PARENT_DOMAIN_NAME} -n api.${CLUSTER} --query "aRecords[0].ipv4Address" -o tsv)
+PUBLIC_IP_FQDN = $(shell az network dns record-set a show -g ${RESOURCEGROUP} --zone-name ${LOCATION}.${PARENT_DOMAIN_NAME} -n api.${CLUSTER} --query fqdn -o tsv)
+update-admin.kubeconfig:
```
ross-bryan, line 187: we can merge PR after we remove this portion post DNS migration

ross-bryan (changes requested):
lgtm, lets get dns migrated, update the makefile, and we can merge this in

Makefile
karanmagdani: Thanks Ross! that's the plan

ross-bryan (approved):

PR #2074, opened 2022-04-21
Add an option to send metrics via UDP instead of Unix Domain Sockets

UlrichSchlueter:
Add a ARO_STATSD_SOCKET variable as option to send metrics into the Genava mdm container via UDP instead of Unix Domain Sockets
Add unit tests
Update related documentation and test scripts

https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14082355

This PR adds an env variable ARO_STATSD_SOCKET that can be used to override the default behavior of the RP and Monitor to send metrics data via a hardcoded UNIX domain socket path. Being able to send metrics via UDP is a pre-requisite for separating the RP/monitor from mdm container deployment on kubernetes for the Hive/OCM project. It is also much more convenient for integration testing of the monitoring code. 

### Test plan for issue:

Unit test added for the handling of the ARO_STATSD_SOCKET content (or lack of)

Integration tested the three cases with data generated from a local rp and monitor (observing a dev cluster)  sent into Geneva INT via local mdm instance:

- original behaviour (ARO_STATSD_SOCKET  not set)
- ARO_STATSD_SOCKET  set to unix:/<somepath>
- ARO_STATSD_SOCKET  set  udp:127.0.0.1:8125

yes, documentation for the local monitor testing and related scripts
Not sure if there is any other documentation for env variables in general.


docs/unit-testing-for-monitoring-metrics.md
```diff
@@ -10,42 +10,56 @@ The ARO monitor component (the part of the aro binary you activate when you exec
...
 
 There are two ways to set to acchieve this: 
-- Run the Geneva MDM container locally (won't work on macOS, see Remote Container section below)
+- Run the Geneva MDM container locally
 - Spawn a VM, start the Geneva container there and connect/tunnel to it.
 
-### Local Container Setup
+and two protocolls to chose from:
+- Unix Domain Sockets, which is the way production is currently (April 2022) is run
+- or UDP, which is much easier to use and is way it will be used to allow for deploying the RP and related components on kubernetes clusters
```
facchettos, line 28: I think there is a typo here ,

pkg/metrics/statsd/statsd_test.go
```diff
@@ -83,3 +86,161 @@ func TestEmitFloat(t *testing.T) {
...
+				}
+			} else {
+				if err != nil && tt.wantError == false {
+					fmt.Println(err.Error())
+					t.Fail()
+				} else if p != tt.protocol {
+					t.Fail()
+				} else if c != tt.connectionstring {
+					t.Fail()
+				}
```
facchettos, line 239: why not use `p != tt.protocol || c != tt.connectionstring` here , since there is no difference in handling ?

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +108,42 @@ func (s *statsd) run() {
 	}
 }
 
+func (s *statsd) getConnectionDetails() (protocol string, connectionstring string, err error) {
+	// allow the socket connection to be overriden via ENV Variable
+	o, isset := os.LookupEnv(STATSD_SOCKET_ENV)
```
facchettos, line 113: That would be great to have more expressive names

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +108,42 @@ func (s *statsd) run() {
...
+		connectionstring = "/var/etw/mdm_statsd.socket"
+		if s.env.IsLocalDevelopmentMode() {
+			connectionstring = "mdm_statsd.socket"
+		}
+
+		return
+	}
+
+	// Assume protocol:connectionstring format
+	a := strings.SplitN(o, ":", 2)
```
facchettos, line 125: Same than above, expressive names go a long way to have clear code

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +108,42 @@ func (s *statsd) run() {
...
+	return
+
+}
+
 func (s *statsd) dial() (err error) {
-	path := "/var/etw/mdm_statsd.socket"
-	if s.env.IsLocalDevelopmentMode() {
-		path = "mdm_statsd.socket"
+
+	p, c, err := s.getConnectionDetails()
```
facchettos, line 140: naming

facchettos (changes requested):

docs/unit-testing-for-monitoring-metrics.md
UlrichSchlueter: yes, will fix.

pkg/metrics/statsd/statsd.go
UlrichSchlueter: “There are only two hard things in Computer Science: cache invalidation and naming things.” . In the past when I got comments on my variables  naming it was to about names being to long and too python-style.
Any general guidelines you could point me to ? Any concrete suggestions here ?

pkg/metrics/statsd/statsd_test.go
UlrichSchlueter: My first version of the code actually combined all three cases in one if statement, before I changed it to this, which I feel is more legible and easier to troubleshoot/debug/set breakpoint. What u think ?

pkg/metrics/statsd/statsd.go
facchettos: I don't know what _too python-style_ is, but I disagree with whoever said that. Names have to be descriptive. I'd rather have a name that is a little long but I know by just looking at it what the variable holds, rather than having to go back and forth to figure it out  

here I'd say maybe something like `socketEnv` could do, it's not perfect but at least we know what the variable holds. WDYT?

pkg/metrics/statsd/statsd.go
facchettos: here I'd say `protocol , conn, err ` would do

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +108,42 @@ func (s *statsd) run() {
...
+	a := strings.SplitN(o, ":", 2)
+	if len(a) != 2 {
+		err = errors.New("malformated ENV Variable ARO_STATSD_SOCKET. Expecting udp:host:port or unix:path-to-socket format")
+
+		return
+	}
+	protocol = strings.ToLower(a[0])
+	connectionstring = a[1]
+
+	return
```
facchettos, line 134: This could probably its own function so we can test it separately. 

Also, why do you do `a := strings.SplitN(o, ":", 2)` if you check afterwards again if its 2 ? Something like udp:a:a:a:a:a pass your checks but would not be valid ?

pkg/metrics/statsd/statsd_test.go
facchettos: ok , that's fair.

pkg/metrics/statsd/statsd.go
facchettos: `Malformated` should be `malformatted` I think. I personally would use `malformed` but I don't really know if there is an actual difference of meaning, so not asking to change :)

UlrichSchlueter: I need the part before the first separator as protocol and then rest of the string as connectionstring to pass to the net.Dial call. I let the evaluation of the actual values of protocol  and connectionstring to the Dial call, I don't want to duplicate that logic and probably get it wrong. I could use two environment variables and save the parsing here, but I opted to mimic how socat is handling these parameters.

Now, SplitN(o,":",2) could deliver 0,1 or 2 items in the slice it returns, so I need to check if I have enough items to go forward. I'll could change it to <2, if that is clearer.

So yes, your example would pass my test, but fail in the dial call.

facchettos: I understand what the logic is , but it's not valid, since it accepts both wrong protocols / malformed urls . 

BTW I don't know what the policy is around here, but I think it would be clearer if you used `return value1, value2,value3` instead of just `return`

facchettos: what do you think about that ?
```go
func validateEnv(env string) error {
	if strings.HasPrefix(env, "udp:") && len(strings.Split(env, ":")) == 3 {
		return nil
	}
	if strings.HasPrefix(env, "unix:") && len(strings.Split(env, ":")) >= 2 {
		return nil
	}

	return fmt.Errorf("invalid env value")
}
```

pkg/metrics/statsd/statsd_test.go
facchettos, line 239: I think we can drop the `else`s 
```suggestion
				if err != nil && tt.wantError == false {
					fmt.Println(err.Error())
					t.Fail()
				} 
				if p != tt.protocol {
					t.Fail()
				} 
				if c != tt.connectionstring {
					t.Fail()
				}
```

facchettos: I think we can drop the `else`s 
```suggestion
				if err != nil && tt.wantError == false {
					fmt.Println(err.Error())
					t.Fail()
				} 
				if p != tt.protocol {
					t.Fail()
				} 
				if c != tt.connectionstring {
					t.Fail()
				}
```

docs/unit-testing-for-monitoring-metrics.md
UlrichSchlueter: Should be better now

pkg/metrics/statsd/statsd_test.go
UlrichSchlueter: Dropped

pkg/metrics/statsd/statsd.go
UlrichSchlueter: Updated the variable names accordingly

pkg/metrics/statsd/statsd.go
UlrichSchlueter: Added additional input validation, split the logic up a bit and updated the unit tests accordingly

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +107,67 @@ func (s *statsd) run() {
...
+		return "", "", fmt.Errorf(unsupportedProtocol, protocol, env)
+	}
+
+	//UDP address check, no such (meaningful) thing for unix:
+	if protocol == "udp" {
+		_, err := net.ResolveUDPAddr(protocol, connectionstring)
+		if err != nil {
+			return "", "", fmt.Errorf(invalidUDPAddress, env, err)
+		}
+	}
```
m1kola, line 136: Is it necessary? Won't `net.Dial` return the same error? Any value in having a separate check here?

m1kola (commented):
I haven't looked at doc changes and the test yet. In general it looks good. So far few small suggestions.

pkg/metrics/statsd/statsd.go
facchettos: for the error, maybe yes, but for the first part of what you selected, we have to check that it is not a weird protocol like tcp , ssh , telnet or whatever. This PR says that it enables things over UDP , and it should only allow UDP

facchettos: also, input sanitization is always a good practice.

pkg/metrics/statsd/statsd.go
```diff
@@ -30,7 +32,9 @@ type statsd struct {
 	now func() time.Time
 }
 
-// New returns a new metrics.Emitter
+const statsdSocketEnv = "ARO_STATSD_SOCKET"
```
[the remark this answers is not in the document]
UlrichSchlueter: Renamed to MDM_STATSD_SOCKET. I'd prefer to keep an indication of the purpose in the name

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +107,67 @@ func (s *statsd) run() {
...
 }
 
-func (s *statsd) dial() (err error) {
-	path := "/var/etw/mdm_statsd.socket"
+func (s *statsd) parseSocketEnv(env string) (string, string, error) {
+	const (
+		malformed           = "malformed ENV variable ARO_STATSD_SOCKET. Expecting udp:<hostname>:<port> or unix:<path-to-socket> format. Got: %s"
+		unsupportedProtocol = "unsupported protocol in ENV variable ARO_STATSD_SOCKET. Expecting  'udp:' or 'unix:'. Got: %s in %s"
+		invalidUDPAddress   = "invalid UDP address in ENV variable ARO_STATSD_SOCKET %s. Error: %s "
+	)
```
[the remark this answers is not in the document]
UlrichSchlueter: Changed as suggested.

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +107,67 @@ func (s *statsd) run() {
...
+	return s.parseSocketEnv(socketEnv)
+}
+
+func (s *statsd) dial() (err error) {
+	protocol, connectionstring, err := s.getConnectionDetails()
+	if err != nil {
+		return
+	}
+
+	s.conn, err = net.Dial(protocol, connectionstring)
```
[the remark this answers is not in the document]
UlrichSchlueter: Changed to follow net package naming

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +107,67 @@ func (s *statsd) run() {
...
+		invalidUDPAddress   = "invalid UDP address in ENV variable ARO_STATSD_SOCKET %s. Error: %s "
+	)
+
+	// Verify protocol:connectionstring format
+	parameters := strings.SplitN(env, ":", 2)
+	if len(parameters) != 2 {
+		return "", "", fmt.Errorf(malformed, env)
+	}
+	protocol := parameters[0]
+	connectionstring := parameters[1]
```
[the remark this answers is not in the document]
UlrichSchlueter: Tested with the url package and decided to leave as is.  The url.Parse call returns a struct that has the address part in different fields depending on protocol/network and address format. The code to go hunting for the values is probably more confusing and more effort to maintain than the code above.

pkg/metrics/statsd/statsd.go
UlrichSchlueter: I think we can make good cases for f both approaches. I'm suggesting to leave it as is now.

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +107,60 @@ func (s *statsd) run() {
...
 	}
 
-	s.conn, err = net.Dial("unix", path)
+	return protocol, connectionstring
+}
+
+func (s *statsd) getConnectionDetails() (string, string, error) {
+	// allow the socket connection to be overriden via ENV Variable
+	socketEnv, isset := os.LookupEnv(statsdSocketEnv)
+	if !isset { //original behaviour
```
s-amann, line 149: nit:
```suggestion
	socketEnv, isSet := os.LookupEnv(statsdSocketEnv)
	if !isSet { //original behaviour
```

s-amann (commented):
lgtm, I plan on trying it out today.

docs/unit-testing-for-monitoring-metrics.md
```diff
@@ -10,42 +10,56 @@ The ARO monitor component (the part of the aro binary you activate when you exec
...
 
 
-
-
-## Unit Testing Setup
+# Unit Testing Setup
 
 If you work on monitor metrics in local dev mode (RP_MODE=Development) you most likely want to see your data somewhere in Geneva INT (https://jarvis-west-int.cloudapp.net/) before you ship your code.
 
 There are two ways to set to acchieve this:
```
s-amann, line 22: nit:
```suggestion
There are two ways to set to achieve this: 
```

docs/unit-testing-for-monitoring-metrics.md
```diff
@@ -10,42 +10,56 @@ The ARO monitor component (the part of the aro binary you activate when you exec
...
+
+> :warning For macOS users: Neither Docker For Desktop nor podman machine currently support Unix Domain Sockets, so this will not work for you. Refer to the Remote Container Setup instead or use the UDP way.
+
+### The UDP way
+
 
-  variables before you start the monitor. 
+An example docker command to start the container locally using UDP is here (you may need to adapt some parameters):
+[Example](../hack/local-monitor-testing/samples/dockerStartCommand-UDP.sh). This will start an mdm container on `0.0.0.0:8125`. 
+Please note that you need to set the `ARP_STATDS_SOCKET` environment variable at monitor start up time so that the monitor finds that socket.
```
s-amann, line 56: I don't see references to `ARP_STATDS_SOCKET` in the codebase. What should this be set to for the "UDP way".

docs/unit-testing-for-monitoring-metrics.md
```diff
@@ -84,12 +98,17 @@ socat -v UNIX-LISTEN:$SOCKETFILE,fork TCP-CONNECT:127.0.0.1:12345
...
 
 
-### Stopping the Network script
+#### Stopping the Network script
 
 Stop the script with Ctrl-C. The script then will do its best to stop the ssh and socal processes it spawned.
 
+### The UDP way
+
+If required you can adapt the [deploy script for the Unix domain socket way](../hack/local-monitor-testing/deploy_MDM_VM.sh) deploys such a VM called $USER-mdm-link on Azure, configures it and installs the mdm container.
```
s-amann, line 107: what would make it required to adapt/change the script? what are the firewall rules mentioned below that need to be in place?

docs/unit-testing-for-monitoring-metrics.md
```diff
@@ -115,17 +146,17 @@ A VS Code launch config that does the same would look like.
...
 
-
-### Finding your data
+## Finding your data
 
 If all goes well, you should see your metric data  in the Jarvis metrics list (Geneva INT (https://jarvis-west-int.cloudapp.net/) -> Manage ->  Metrics) under the account and namespace you specified in CLUSTER_MDM_ACCOUNT and CLUSTER_MDM_NAMESPACE and also be available is the dashboard settings.
 
 
-### Injecting Test Data into Geneva INT
+## Injecting Test Data into Geneva INT
```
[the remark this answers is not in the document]
UlrichSchlueter: Yes, you want to  send the traffic into the UDP socket the mdm container opens, laster commit should have the instructions

docs/unit-testing-for-monitoring-metrics.md
UlrichSchlueter: That's a typo combined with a failed find/replace. The name is now MDM_STATSD_SOCKET.  Commit is on it's way

docs/unit-testing-for-monitoring-metrics.md
UlrichSchlueter: Corrected.

pkg/metrics/statsd/statsd.go
UlrichSchlueter: Changed accordingly

facchettos (dismissed):

docs/unit-testing-for-monitoring-metrics.md
```diff
@@ -10,42 +10,56 @@ The ARO monitor component (the part of the aro binary you activate when you exec
...
+
+### The Unix Domain Socket way
+
+An example docker command to start the container locally using UNIX domain sockets is here (you may need to adapt some parameters):
+[Example](../hack/local-monitor-testing/samples/dockerStartCommand-UDS.sh). As mentioned this is the way it's run in production. The script will configure the mdm container to connect to Geneva INT.  The sample script tries to copy the certificate file to /etc/mdm.pem to mimic production. You can change that by changing the corresponding volume definition. In that case you'll also need to set the `ARP_STATDS_SOCKET` environment variable to `"unix:/<path to where the socket file should be created>"` to point the monitor (or RP for that matter) to the same place/file.
+
+When you start the montitor locally in local dev mode, by default he monitor looks for a Unix Socket file called `mdm_statsd.socket` in the current directory from where the cmd/aro command is executed, typically either the base folder of your repo or `./cmd/aro` is started through VSCode (for example)
```
s-amann, line 47:
```suggestion
When you start the monitor locally in local dev mode, by default he monitor looks for a Unix Socket file called `mdm_statsd.socket` in the current directory from where the cmd/aro command is executed, typically either the base folder of your repo or `./cmd/aro` is started through VSCode (for example)
```

s-amann (dismissed):

pkg/metrics/statsd/statsd_test.go
```diff
@@ -83,3 +86,247 @@ func TestEmitFloat(t *testing.T) {
...
+	s.EmitGauge("tests.test_key", 42, map[string]string{"key": "value"})
+	m := <-c
+	if m != `{"Metric":"tests.test_key","Account":"*","Namespace":"*","Dims":{"hostname":"test-host","key":"value","location":"eastus"},"TS":"0001-01-01T00:00:00.000"}:42|g`+"\n" {
+		t.Error(errors.New(m))
+	}
+}
+
+func TestGetConnectionDetails(t *testing.T) {
+	controller := gomock.NewController(t)
+	defer controller.Finish()
```
m1kola, line 137: This must be inside `t.Run` otherwise failure reporting will be messed up. This can also lead to state sharing/flakes between test cases.

pkg/metrics/statsd/statsd_test.go
```diff
@@ -83,3 +86,247 @@ func TestEmitFloat(t *testing.T) {
...
+
+	env := mock_env.NewMockInterface(controller)
+	env.EXPECT().Location().AnyTimes().Return("eastus")
+	env.EXPECT().Hostname().AnyTimes().Return("test-host")
+
+	socket := "127.0.0.1:8001"
+
+	os.Setenv(statsdSocketEnv, "udp:"+socket)
```
m1kola, line 100: IMO - env manipulations in tests is usually a bad idea. This can lead to flakes: we set env in this test, but we might rely on this env in a different test. Depending on the order of execution we might see failure/success.

I think we can solve this by removing `os.LookupEnv` call from `getConnectionDetails()` and passing a value into `New` similar to `account` and `namespace`.

UPD: In a different comment I say that we can drop this test, but I still think we should remove `os.LookupEnv` from `statsd`.

pkg/metrics/statsd/statsd_test.go
```diff
@@ -83,3 +86,247 @@ func TestEmitFloat(t *testing.T) {
 		t.Error(m)
 	}
 }
+
+func TestEmitGaugeViaUDP(t *testing.T) {
```
m1kola, line 90: I think we can drop this test: we are basically testing Go's standard library here. The only difference between unix socet and UDP from our point of view is what we pass into `net.Dial`. Everything else is `net` package implementation which is tested in it's own.

What we might want to do instead is this: make `net.Dial` mockable and check what we pass into into. In `statsd` struct we are already making `time.Now()` mockable here:

https://github.com/Azure/ARO-RP/blob/1ae6115606b60c4f530ae4154cae8f734e161cc2/pkg/metrics/statsd/statsd.go#L44

It will be the same appraoch.

But the value of this test is not very high given that we are testing `getConnectionDetails()` separately and there is no extra logic in `dial()`. So IMO we can just drop this test.

pkg/metrics/statsd/statsd_test.go
```diff
@@ -83,3 +86,247 @@ func TestEmitFloat(t *testing.T) {
...
+			wantError:  false,
+		},
+		{
+			name:         "Valid UDP env variable",
+			set:          true,
+			socketstring: "udp:127.0.0.1:9000",
+			network:      "udp",
+			address:      "127.0.0.1:9000",
+			wantError:    false,
+		},
```
m1kola, line 170: I would also add a case when `isLocalDev: true`, but `socketstring` is set to make sure that we do not override in dev.

pkg/metrics/statsd/statsd_test.go
```diff
@@ -83,3 +86,247 @@ func TestEmitFloat(t *testing.T) {
...
+func TestParseSocketEnv(t *testing.T) {
+	controller := gomock.NewController(t)
+	defer controller.Finish()
+
+	for _, tt := range []struct {
+		name      string
+		socketEnv string
+		network   string
+		address   string
+		wantError bool
```
m1kola, line 226: I think this needs to be a string with and we must check for a specific error. Given that we return multiple errors we need to make sure that we are hitting the case we intend the case to hit.

Similar to this:

https://github.com/Azure/ARO-RP/blob/1ae6115606b60c4f530ae4154cae8f734e161cc2/pkg/api/admin/openshiftcluster_validatestatic_test.go#L58-L65

pkg/metrics/statsd/statsd_test.go
```diff
@@ -83,3 +86,247 @@ func TestEmitFloat(t *testing.T) {
...
+					t.Fail()
+				}
+			}
+		})
+	}
+}
+
+func TestParseSocketEnv(t *testing.T) {
+	controller := gomock.NewController(t)
+	defer controller.Finish()
```
m1kola, line 219: Same as above: needs to be in `t.Run`

m1kola (changes requested):
I left a lot of comments but most of them are about docs (I also see some unresolved comments from @s-amann).

The biggest change I think we should make is getting rid of `os.LookupEnv` in `statsd` struct.

docs/unit-testing-for-monitoring-metrics.md
```diff
@@ -10,43 +10,38 @@ The ARO monitor component (the part of the aro binary you activate when you exec
...
   - CLUSTER_MDM_ACCOUNT
-  - CLUSTER_MDM_NAMESPACE
-   
-
-  variables before you start the monitor. 
+  - CLUSTER_MDM_NAMESPACE variables before you start the monitor. 
 
-An example docker command to start the container locally is here (you may need to adapt some parameters):
-[Example](../hack/local-monitor-testing/sample/dockerStartCommand.sh). The script will configure the mdm container to connect to Geneva INT
+ The container needs to be provided with the Geneva key and certificate. For the INT instance that is the rp-metrics-int.pem you find in the secrets folder after running the `make secrets` command above.
```
s-amann, line 40: I don't believe this key/cert exist yet in the new subscription. Do you know what needs to be done to create/copy them?

pkg/metrics/statsd/statsd.go
```diff
@@ -21,23 +22,25 @@ type statsd struct {
...
 
 	conn net.Conn
 	ch   chan *metric
 
 	now func() time.Time
 }
 
-// New returns a new metrics.Emitter
-func New(ctx context.Context, log *logrus.Entry, env env.Core, account, namespace string) metrics.Emitter {
+// New returns a new metrics.Interface
```
s-amann, line 35:
```suggestion
// New returns a new metrics.Emitter
```
This function still returns a `metrics.Emitter`

pkg/metrics/statsd/statsd_test.go
UlrichSchlueter: Dropped the test.

pkg/metrics/statsd/statsd_test.go
UlrichSchlueter: Changed to strings and specific errors

pkg/metrics/statsd/statsd_test.go
UlrichSchlueter: Fixed

docs/unit-testing-for-monitoring-metrics.md
UlrichSchlueter: I used the ones from the old subscription, which is not an option anymore. If they are missing in the new storage account we'll need to get them added.

I effectively reverted the documentation to the state before the UDP option because MSFT introduced new security rules that block most of the local testing. Will need to look at that some other time in another PR.

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +106,65 @@ func (s *statsd) run() {
...
+	network := "unix"
+	address := "/var/etw/mdm_statsd.socket"
+
 	if s.env.IsLocalDevelopmentMode() {
-		path = "mdm_statsd.socket"
+		address = "mdm_statsd.socket"
+	}
+
+	return network, address
+}
```
facchettos, line 138: A little nit here, couldn't this be a little simpler ? 
```go
if s.env.IsLocalDevelopmentMode() {
    return "unix", "mdm_statsd.socket"
}
return "unix", "/var/etw/mdm_statsd.socket"
``` for example ?

File: pkg/metrics/statsd/statsd_test.go [hunk 27]
```diff
@@ -83,3 +85,240 @@ func TestEmitFloat(t *testing.T) {
...
+			isLocalDev:           true,
+			mdmsocketstring:      "",
+			network:              "unix",
+			address:              "mdm_statsd.socket",
+			wantErrorToStartWith: "",
+		},
+		{
+			name:                 "Valid UDP env variable",
+			isLocalDev:           false,
+			mdmsocketstring:      "udp:127.0.0.1:9000",
```
facchettos, line 259: This is not standard URL format is it ? To me it would be `udp://127.0.0.1:9000`

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +106,65 @@ func (s *statsd) run() {
...
+
+func (s *statsd) validateSocketDefinition(network string, address string) (bool, error) {
+	//Verify supported protocol provided. TCP might just work as well, but this was never tested
+	if network != "udp" && network != "unix" {
+		return false, fmt.Errorf("unsupported protocol for the mdm statds socket. Expecting  'udp:' or 'unix:'. Got: %q", network)
+	}
+
+	return true, nil
+}
```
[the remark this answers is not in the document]
UlrichSchlueter: I took the addressing scheme idea from the socat tool. Thing is we don't deal with URLs and I don't want to give the impression that we do. Mikalai suggested to use url.Parse in the past, but the parsing for files (emulating UNIX socket format) turned out to not be fitting very well: dependent on the format (relative or absolute paths for example) the address part is returned in different fields of the url struct. Trying to go hunt for the value or determine the exact case and field is not simpler than the 'manual' split and the static validation above. If we decide that the formatting scheme is not acceptable, I'd prefer to go to a two ENV variable format, one for the network and for the address part.

pkg/metrics/statsd/statsd.go
UlrichSchlueter: It would be, but we don't have any hint in the code anymore what the function deals with (after removing the name from return parameter list. (We could add comments, but ...... -8-)

pkg/metrics/statsd/statsd.go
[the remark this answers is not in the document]
facchettos: Sounds good !

s-amann (approved):

pkg/metrics/statsd/statsd_test.go
```diff
@@ -83,3 +85,240 @@ func TestEmitFloat(t *testing.T) {
...
+				if err != nil {
+					t.Error(fmt.Errorf("unexpected error received: %q ", err.Error()))
+				}
+				if part1 != tt.part1 {
+					t.Error(fmt.Errorf("part1 does not match: Wanted: %q but got %q ", tt.part1, part1))
+				}
+				if part2 != tt.part2 {
+					t.Error(fmt.Errorf("part2 does not match: Wanted: %q but got %q ", tt.part2, part2))
+				}
+			}
```
m1kola, line 168: 1. I would drop error formatting. `t.Error(actualValue)` will print test file name and line number as well as name of the test case from `t.Run(...)`). This will be enough to indetify what when what's wrong.
2. IMO depsite the fact that we no longer return multiple errors from the function I still think that `tt.wantError` should be a string as we have multiple different cases where we hit an error and we might add more erorrs later into this function. Will be easier to extend the test.
3. IMO, `part1` and `part2` should be tested independently from an error. This will help us to ensure that when we return an error we return an empty parts
4. `part1` & `part2`/`tt.part1` & `tt.part2` need better name. Maybe `network` & `address` / `tt.wantNetwork` & `tt.wantAddress`?

m1kola: This is what I came up with while reviewing, if you are interested. This is the above feedback applied to all the new tests in `pkg/metrics/statsd/statsd_test.go` file.

<details>
<summary>Spoiler</summary>


```go
func TestParseSocketEnv(t *testing.T) {
	for _, tt := range []struct {
		name        string
		teststring  string
		wantNetwork string
		wantAddress string
		wantError   string
	}{
		{
			name:        "Valid string",
			teststring:  "part1:part2",
			wantNetwork: "part1",
			wantAddress: "part2",
		},
		{
			name:        "Empty first part-regarded valid",
			teststring:  ":part2",
			wantNetwork: "",
			wantAddress: "part2",
		},
		{
			name:        "Empty second part-regarded valid",
			teststring:  "part1:",
			wantNetwork: "part1",
			wantAddress: "",
		},
		{
			name:       "No separator",
			teststring: "somerandommtext",
			wantError:  "malformed definition for the mdm statds socket. Expecting udp:<hostname>:<port> or unix:<path-to-socket> format. Got: \"somerandommtext\"",
		},
		{
			name:       "Empty string",
			teststring: "",
			wantError:  "malformed definition for the mdm statds socket. Expecting udp:<hostname>:<port> or unix:<path-to-socket> format. Got: \"\"",
		},
		{
			name:        "More than one separator",
			teststring:  "a:b:c",
			wantNetwork: "a",
			wantAddress: "b:c",
		},
		{
			name:        "Many separators",
			teststring:  ":::::",
			wantNetwork: "",
			wantAddress: "::::",
		},
		{
			name:        "One separator only",
			teststring:  ":",
			wantNetwork: "",
			wantAddress: "",
		},
	} {
			s := &statsd{}
			network, address, err := s.parseSocketEnv(tt.teststring)

			if err != nil && err.Error() != tt.wantError ||
				err == nil && tt.wantError != "" {
				t.Fatal(err)
			}

			if network != tt.wantNetwork {
				t.Error(network)
			}
			if address != tt.wantAddress {
				t.Error(address)
			}
		})
	}
}

func TestValidateSocketDefinition(t *testing.T) {
	for _, tt := range []struct {
		name         string
		network      string
		address      string
		expectToPass bool
		wantError    string
	}{
		{
			name:         "Valid UDP case",
			network:      "udp",
			address:      "127.0.0.1:9000",
			expectToPass: true,
		},
		{
			name:         "Valid Unix Domain Socket case",
			network:      "unix",
			address:      "/var/something/or/another",
			expectToPass: true,
		},
		{
			name:         "Invalid protocoll",
			network:      "tcp",
			address:      "127.0.0.1:12",
			expectToPass: false,
			wantError:    "unsupported protocol for the mdm statds socket. Expecting  'udp:' or 'unix:'. Got: \"tcp\"",
		},
		{
			name:         "Valid protocol, random invalid address - is good here",
			network:      "udp",
			address:      "somerandomtext149203#$%",
			expectToPass: true,
		},
	} {
			s := &statsd{}

			ok, err := s.validateSocketDefinition(tt.network, tt.address)

			if err != nil && err.Error() != tt.wantError ||
				err == nil && tt.wantError != "" {
				t.Fatal(err)
			}

			if ok != tt.expectToPass {
				t.Error(ok)
			}
		})
	}
}

func TestConnectionDetails(t *testing.T) {
	for _, tt := range []struct {
		name            string
		isLocalDev      bool
		mdmsocketstring string
		network         string
		address         string
		wantError       string
	}{
		{
			name:            "Old behaviour / production mode",
			isLocalDev:      false,
			mdmsocketstring: "",
			network:         "unix",
			address:         "/var/etw/mdm_statsd.socket",
			wantError:       "",
		},
		{
			name:            "Old Behaviour / localdev mode",
			isLocalDev:      true,
			mdmsocketstring: "",
			network:         "unix",
			address:         "mdm_statsd.socket",
			wantError:       "",
		},
		{
			name:            "Valid UDP variable",
			isLocalDev:      false,
			mdmsocketstring: "udp:127.0.0.1:9000",
			network:         "udp",
			address:         "127.0.0.1:9000",
			wantError:       "",
		},
		{
			name:            "Don't override default in localdev ",
			isLocalDev:      true,
			mdmsocketstring: "udp:127.0.0.1:9000",
			network:         "udp",
			address:         "127.0.0.1:9000",
			wantError:       "",
		},
		{
			name:            "Random string without separator",
			isLocalDev:      true,
			mdmsocketstring: "another random string without separator",
			wantError:       "malformed definition for the mdm statds socket. Expecting udp:<hostname>:<port> or unix:<path-to-socket> format. Got: \"another random string without separator\"",
		},
		{
			name:            "Invalid Protocol",
			isLocalDev:      true,
			mdmsocketstring: "foo:bar",
			wantError:       "unsupported protocol for the mdm statds socket. Expecting  'udp:' or 'unix:'. Got: \"foo\"",
		},
	} {
			controller := gomock.NewController(t)
			defer controller.Finish()
			env := mock_env.NewMockInterface(controller)

			if tt.mdmsocketstring == "" {
				if tt.isLocalDev {
					env.EXPECT().IsLocalDevelopmentMode().Return(true)
				} else {
					env.EXPECT().IsLocalDevelopmentMode().Return(false)
				}
			}

			s := &statsd{
				env:          env,
				mdmsocketEnv: tt.mdmsocketstring,
			}

			network, address, err := s.getConnectionDetails()

			if err != nil && err.Error() != tt.wantError ||
				err == nil && tt.wantError != "" {
				t.Fatal(err)
			}

			if network != tt.network {
				t.Error(network)
			}
			if address != tt.address {
				t.Error(network)
			}
		})
	}
}

```
  
</details>

pkg/metrics/statsd/statsd_test.go
m1kola: @facchettos related comment: https://github.com/Azure/ARO-RP/pull/2074#discussion_r856261859

m1kola (commented):
A bunch of nits and bumps of older comments. Nothing critical. I included a refactored unit tests (I did it while reviewing) - feel free to use them if a) you agree with the comments and b) it can seve you a bit of time.

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +106,65 @@ func (s *statsd) run() {
...
+
 	if s.env.IsLocalDevelopmentMode() {
-		path = "mdm_statsd.socket"
+		address = "mdm_statsd.socket"
+	}
+
+	return network, address
+}
+
+func (s *statsd) getConnectionDetails() (string, string, error) {
```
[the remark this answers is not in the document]
UlrichSchlueter: renamed accordingly

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +106,65 @@ func (s *statsd) run() {
...
+func (s *statsd) validateSocketDefinition(network string, address string) (bool, error) {
+	//Verify supported protocol provided. TCP might just work as well, but this was never tested
+	if network != "udp" && network != "unix" {
+		return false, fmt.Errorf("unsupported protocol for the mdm statds socket. Expecting  'udp:' or 'unix:'. Got: %q", network)
+	}
+
+	return true, nil
+}
+
+func (s *statsd) getDefaultSocketValues() (string, string) {
```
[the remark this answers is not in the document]
UlrichSchlueter: renamed accordingly

pkg/metrics/statsd/statsd.go
```diff
@@ -103,13 +106,65 @@ func (s *statsd) run() {
...
-		path = "mdm_statsd.socket"
+		address = "mdm_statsd.socket"
+	}
+
+	return network, address
+}
+
+func (s *statsd) getConnectionDetails() (string, string, error) {
+	// allow the default socket connection to be overriden
+	if s.mdmsocketEnv == "" { //original behaviour
```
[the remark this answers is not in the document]
UlrichSchlueter: removed the comment

AldoFusterTurpin:
Is there any reason to do both t.Error() and t.Fail() for failing in the tests?
If there is no special reason I think it would be great to unify the failing method (choose one or the other).

petrkotas (changes requested):
Hi @UlrichSchlueter I have one comment, beside that I see that the change is transparent. 

I understand it will not be used right now, but it will be used in the upcomming AKS.

hack/local-monitor-testing/samples/dockerStartCommand.sh
```diff
@@ -27,7 +43,7 @@ echo "MDMSOURCEENV  = $MDMSOURCEENVIRONMENT"
 echo "MDMSOURCEROLE  = $MDMSOURCEROLE"
 echo "MDMSOURCEROLEINSTANCE  = $MDMSOURCEROLEINSTANCE"
 
-cp $BASE/secrets/rp-metrics-int.pem /etc/mdm.pem
+#cp $BASE/secrets/rp-metrics-int.pem /etc/mdm.pem
```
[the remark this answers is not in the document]
UlrichSchlueter: I just removed it.

petrkotas (dismissed):
Thanks @UlrichSchlueter for the update. 🎉 🌮

pkg/metrics/statsd/statsd_test.go
```diff
@@ -83,3 +83,220 @@ func TestEmitFloat(t *testing.T) {
...
+			teststring:  ":",
+			wantNetwork: "",
+			wantAddress: "",
+		},
+		{
+			name:        "Convert Upper Case network to lower case but leave address alone",
+			teststring:  "FOO:BAR",
+			wantNetwork: "foo",
+			wantAddress: "BAR",
+		},
```
m1kola, line 146: For visibility: it is a bit weird that we do not error on these cases, but we discussed it with @UlrichSchlueter and decided - it is not worth putting a lot of effort into validating something which is provided by us (SREs configuring the env).

m1kola (dismissed):
Looks good to me. Let's merge once we fix our e2e (it doesn't look related to this PR).

m1kola (approved):

PR #2075, opened 2022-04-21
Dropping ipRules parameter; it is not used in this template (causes "unable to parse template on line 1" error)

coreyperkins:
It is worth noting that it is used in the rp-production.json template but leaving it in place in the rp-development.json template unused results in a syntax error that states "unable to parse template on line 1". This causes the process to hang when settings up role assignments for the shared rp.

https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14203980/

Allows us to setup the shared RP.

### Test plan for issue:

This has been tested by multiple parties via shared RP setup.

Documentation for this and other issues related to setting up the shared RP env and running the RP process locally is coming in another PR.

### DIFF
--- a/pkg/deploy/assets/rp-development.json
+++ b/pkg/deploy/assets/rp-development.json
@@ -11,9 +11,6 @@
         "fpServicePrincipalId": {
             "type": "string"
         },
-        "ipRules": {
-            "type": "array"
-        },
         "rpServicePrincipalId": {
             "type": "string"
         }
--- a/pkg/deploy/generator/templates_rp.go
+++ b/pkg/deploy/generator/templates_rp.go
@@ -26,7 +26,6 @@ func (g *generator) rpTemplate() *arm.Template {
 		"databaseAccountName",
 		"fpServicePrincipalId",
 		"rpServicePrincipalId",
-		"ipRules",
 	}
 	if g.production {
 		params = append(params,
@@ -50,6 +49,7 @@ func (g *generator) rpTemplate() *arm.Template {
 			"fluentbitImage",
 			"fpClientId",
 			"fpServicePrincipalId",
+			"ipRules",
 			"keyvaultPrefix",
 			"keyvaultDNSSuffix",
 			"gatewayDomains",


coreyperkins:
@karanmagdani

karanmagdani (dismissed):
LGTM

m1kola (changes requested):
Same as [here](https://github.com/Azure/ARO-RP/pull/2067#pullrequestreview-949852577): I did not look into why we need this, but `deploy/rp-development.json` is generated and `make generate` will override manual changes.

https://github.com/Azure/ARO-RP/blob/5668a316afae97b690871edac2c59eb8f84efacb/pkg/deploy/generator/templates_rp.go#L21-L151

bennerv:
@coreyperkins mind updating the `pkg/deploy/generator/templates_rp.go` and removing all instances of this.  Then a `make generate` should resolve everything here and we can merge.

coreyperkins:
@bennerv @m1kola -- thanks for the tips!

Ben, I did not remove all instances because I noticed when I ran `make generate` the deploy/rp-production.json still had references to ipRules on line 720: 

`"ipRules": "[if(parameters('disableCosmosDBFirewall'), createArray(), concat(parameters('ipRules'),createArray(createObject('ipAddressOrRange', '104.42.195.92'),createObject('ipAddressOrRange','40.76.54.131'),createObject('ipAddressOrRange','52.176.6.30'),createObject('ipAddressOrRange','52.169.50.45'),createObject('ipAddressOrRange','52.187.184.26'))))]",`

I did, however, move the `ipRules` parameter in `pkg/generator/templates_rp.go` to the array of params that gets appended if the g.production check succeeds on line 30.

Is this correct? If not, how should I proceed? Thanks!

bennerv:
@coreyperkins this needs a `make generate`.  

On a side note, do you know what the `ipRules` parameter gets parameterized into when it fails?  Is it an empty array, or something else?  If it's an empty array, it should still work so I'm a bit confused on why it's failing as the logic within the ARM template should be sufficient that it works for dev and prod without _special_ modification.

coreyperkins:
@bennerv Hi Ben! I am going to reach out to you direct this morning for guidance and then I will document here what we discuss. Thanks!

coreyperkins:
Per @bennerv request I ran `make generate` and submitted the `pkg/deploy/bindata.go` file. Also, we were running into the error because ipRules had no default value (and was thus required). As a reference we were running the command from the shell helper file (https://github.com/Azure/ARO-RP/blob/master/hack/devtools/deploy-shared-env.sh) per the instructions at https://github.com/Azure/ARO-RP/blob/master/docs/prepare-a-shared-rp-development-environment.md. With ipRules now gone this is no longer presenting the error. Thanks!

bennerv (approved):
lgtm - shouldn't need to use ipRules in dev anyway.

coreyperkins:
@bennerv @m1kola -- I believe we are good here now, but if not please advise!

ross-bryan (approved):

hawkowl (approved):

PR #2076, opened 2022-04-21
Changed priority of cluster nsg to 120

coreyperkins:

During our working session @karanmagdani adjust the cluster nsg priority to 120 to eliminate an ordering issue. @karanmagdani I do not have anymore details recorded in our working session documents, will you please review and attach if you do?

Prevents errors in Azure that hinder creating a cluster.

### Test plan for issue:

We ran several deployments from the local dev env.

Documentation related to working sessions with RH will be included in a PR today.


karanmagdani:
@coreyperkins Thanks for the PR. This is another change that is specific to CloudFit only. Some context:

When creating a cluster in SimplySecure world, our set NSG priority `101` conflicts with SimplySecure Policies. According to the docs, the highest priority we can set is 120. Any number lower than that is reserved for Azure stuff. 

Given this information, I believe we don't need to merge this change in.

coreyperkins:
@karanmagdani pulling back to the cfs base, thanks!

PR #2077, opened 2022-04-21
Added the gwy kv template to the rp-development-prepdeploy

coreyperkins:
It was missing from the dev predeploy. For production there is a separate set of templates for the gateway, but this seemed to suffice for dev purposes.

https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14204117

Allows continuing to setup the PR for development.

### Test plan for issue:

This has been manually tested via local dev RP deployments.

Documentation coming in the next PR.

### DIFF
--- a/deploy/rp-development-predeploy.json
+++ b/deploy/rp-development-predeploy.json
@@ -290,6 +290,49 @@
                 "enableSoftDelete": true
             },
             "apiVersion": "2019-09-01"
+        },
+        {
+            "name": "[concat(parameters('keyvaultPrefix'), '-gwy')]",
+            "type": "Microsoft.KeyVault/vaults",
+            "location": "[resourceGroup().location]",
+            "properties": {
+                "tenantId": "[subscription().tenantId]",
+                "sku": {
+                    "family": "A",
+                    "name": "standard"
+                },
+                "accessPolicies": [
+                    {
+                        "tenantId": "[subscription().tenantId]",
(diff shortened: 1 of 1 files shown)


coreyperkins:
@karanmagdani

karanmagdani:
AFAIK we don't need gwy keyvault resource in our development clusters. 

But since i don't know enough about this, i'll defer to @ross-bryan @bennerv in case they have any more thoughts

coreyperkins:
> AFAIK we don't need gwy keyvault resource in our development clusters.
> 
> But since i don't know enough about this, i'll defer to @ross-bryan @bennerv in case they have any more thoughts

Including @zacharyljones. He ran into this and made the adjustment, he may have ref to the specific error reported.

zacharyljones:
@coreyperkins I think @karanmagdani is correct. We created the gwy key vault so we could load the certificates in [this step](https://github.com/Azure/ARO-RP/blob/17ff475d640ea74eb2be80198ab99e4f00dbabdf/docs/prepare-a-shared-rp-development-environment.md?plain=1#L477) of the documentation, but no gwy resources actually get created.

I think we should revise that section in the documentation - the context switching between what is needed in different environments is difficult to parse.

coreyperkins:
@karanmagdani @zacharyljones per the commentary closing this commit. I will make a PR to update the documentation.

PR #2080, opened 2022-04-21
Exclude logs from openshift-logging/gitops namespace

devamanv:
The PR is to exclude logs from the openshift-logging namespace. 

Fixes [Exclude openshift-logging logs from being ingested into Azure log infrastructure](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13146970/)

This is to exclude logs from openshift-logging and openshift-gitops logs going into the Azure Geneva logging infrastructure due to the fact that incorrectly configured customer log ingestion can result in customer workload logs being ingested


pkg/operator/controllers/genevalogging/const.go
```diff
@@ -73,7 +73,7 @@ const (
 [FILTER]
 	Name grep
 	Match containers
-	Regex NAMESPACE ^(?:default|kube-.*|openshift|openshift-.*)$
+	Regex NAMESPACE ^(?:default|kube-.*|openshift|^(?!openshift-(logging|gitops))(openshift-.*))$
```
darthhexx, line 76: There is no need for the second ^ because the first one is outside the non-capturing group.

darthhexx (changes requested):
This is working well. Just a nit to remove the extra ^.

PR #2081, opened 2022-04-22
Enables go fmt simplify in golangci config

m1kola:

I noticed that in our golangci config we have [go fmt simplify](https://pkg.go.dev/cmd/gofmt#hdr-The_simplify_command) disabled. I see no reason to disable it. So this PR enables go fmt simplify.

### Test plan for issue:

CI run

No, only changes linters

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -13,11 +13,8 @@ issues:
   exclude-use-default: false
 
 linters-settings:
-  gofmt:
-    simplify: false
-  
   stylecheck:
-    # added additional checks for comments in Go. 
+    # added additional checks for comments in Go.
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -13,11 +13,8 @@ issues:
   exclude-use-default: false
 
 linters-settings:
-  gofmt:
-    simplify: false
```
m1kola, line 17: [`true` is a default value](https://golangci-lint.run/usage/linters/#gofmt) for golangci

s-amann (approved):
lgtm, so we must already be good to go with this lint as this was the only change!

facchettos (approved):

ross-bryan (approved):

PR #2083, opened 2022-04-22
added stylecheck and moved golangci-lint to a github action

facchettos:
Adds stylecheck as TODO was suggesting. Fixed code which did not comply.

I ran into problems updating the golangci-lint in vendor. It was required because the (very) outdated version did not recognize the options in the config file. I moved linting to a GitHub action which 
1. runs it in parallel
2. is much easier to keep up to date
3. has a better UX as linting errors will show up in github directly in the files tab. 
image

### DIFF
--- a/.github/workflows/golint.yml
+++ b/.github/workflows/golint.yml
@@ -1,0 +1,49 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - main
+  pull_request:
+permissions:
+  contents: read
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  # pull-requests: read
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - run: | 
+          sudo apt-get update 
+          sudo apt-get install libgpgme-dev libgpgme11
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.17
+      - uses: actions/checkout@v3
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.45.2
+
+          # Optional: working directory, useful for monorepos
+          #working-directory: pkg
+
+          # Optional: golangci-lint command line arguments.
+          #args: -c .golangci.yml
+
+          # Optional: show only new issues if it's a pull request. The default value is `false`.
+          # only-new-issues: true
+
+          # Optional: if set to true then the all caching functionality will be complete disabled,
+          #           takes precedence over all other caching options.
+          # skip-cache: true
+
+          # Optional: if set to true then the action don't cache or restore ~/go/pkg.
+          # skip-pkg-cache: true
+
+          # Optional: if set to true then the action don't cache or restore ~/.cache/go-build.
+          # skip-build-cache: true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,7 +16,10 @@ linters-settings:
   stylecheck:
     # added additional checks for comments in Go.
     # Refer https://staticcheck.io/docs/options#checks for details
-    checks: ["all", "-ST1000", "ST1020", "ST1021", "ST1022"]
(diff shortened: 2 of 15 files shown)


m1kola (changes requested):
IMO - we should just update files which do not match stylecheck rules and enable it in `.golangci.yml`.

facchettos:
I changed the style check options slight because : 

1. they were not correct (did not do what they were supposed to)
2. these checks are according too opinionated according to staticcheck.io . So I set it back to the default 

> This option sets which [checks](https://staticcheck.io/docs/checks/) should be enabled. By default, most checks will be enabled, except for those that are too opinionated or that only apply to packages in certain domains.

Makefile
```diff
@@ -166,6 +166,7 @@ unit-test-go:
 
 lint-go:
 	go run ./vendor/github.com/golangci/golangci-lint/cmd/golangci-lint run
+	go run ./vendor/github.com/golangci/golangci-lint/cmd/golangci-lint -c ./.golangci-stylecheck.yml run
```
m1kola, line 169: Looks like a leftover.

.golangci.yml
```diff
@@ -19,7 +19,10 @@ linters-settings:
   stylecheck:
     # added additional checks for comments in Go. 
     # Refer https://staticcheck.io/docs/options#checks for details
-    checks: ["all", "-ST1000", "ST1020", "ST1021", "ST1022"]
+    checks: ["all", "-ST1000", "-ST1003", "-ST1016", "-ST1020", "-ST1021", "-ST1022"]
+    dot-import-whitelist:
+      - github.com/onsi/ginkgo
+      - github.com/onsi/gomega
```
m1kola, line 25: We are already restricting dot imports here: 

https://github.com/Azure/ARO-RP/blob/5668a316afae97b690871edac2c59eb8f84efacb/hack/validate-imports/validate-imports.go#L18-L26

Probably makes sense to drop one of the checks: either in golangci or in our hack script. I would drop hack script validation (unless there is some extra value over golangci?)

m1kola (changes requested):

.golangci.yml
facchettos: I think that we should remove the one from hack.

facchettos:
There is something fishy with `golangci-lint` in the pipeline. It does not seem to take the config file into account. It works fine on my laptop with my own version of `golangci-lint` (v 1.42 while the one vendored is 1.32) though. I think we should update it

s-amann (dismissed):
lgtm

.github/workflows/golint.yml
```diff
@@ -0,0 +1,49 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - main
```
s-amann, line 8: while we don't have a `main` branch, perhaps we should have an item to migrate to this naming strategy

s-amann (approved):
lgtm

.github/workflows/golint.yml
facchettos: I have no opinion on this :) , either is fine with me 
It's in here because I shamelessly stole the config file from the official golangci-lint github action and did not notice it

petrkotas (approved):

.github/workflows/golint.yml
```diff
@@ -0,0 +1,49 @@
...
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - run: | 
+          sudo apt-get update 
+          sudo apt-get install libgpgme-dev libgpgme11
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.17
```
bennerv, line 24: We'll have to add this as part of the revendoring process so we don't miss this.

bennerv (approved):
lgtm

test/e2e/adminapi_cluster_update.go
```diff
@@ -21,7 +21,7 @@ var _ = Describe("[Admin API] Cluster admin update action", func() {
 	BeforeEach(skipIfNotInDevelopmentEnv)
 
 	It("should be able to run cluster update operation on a cluster", func() {
-		var oc *admin.OpenShiftCluster = &admin.OpenShiftCluster{}
+		var oc = &admin.OpenShiftCluster{}
```
ross-bryan, line 24: nice

ross-bryan (approved):

PR #2086, opened 2022-04-25
Add IL5 instance types, improve static validator, add tests

cadenmarchese:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14180703/

- Adds IL5 instances that weren't already added (a few were already there)
- Adds some improvements and unit testing to the static validator
- Adds some more test cases to the dynamic validator

### Test plan for issue:

- Tested an all-IL5 instance install in our gratis dev subscription

Yes, we'll need to update our support policy with the new sizes


pkg/api/validate/vm.go
```diff
@@ -7,66 +7,97 @@ import (
...
+	api.VMSizeStandardL8s,
+	api.VMSizeStandardL16s,
+	api.VMSizeStandardL32s,
+	api.VMSizeStandardL8sV2,
+	api.VMSizeStandardL16sV2,
+	api.VMSizeStandardL32sV2,
+	api.VMSizeStandardL48sV2,
+	api.VMSizeStandardL64sV2,
+}
+
```
facchettos, line 78: Why not use a `map[api.VMSize]struct{}` or `map[api.VMSize]bool` instead of doing a lookup with a loop?

pkg/api/validate/vm.go
```diff
@@ -7,66 +7,97 @@ import (
...
-				api.VMSizeStandardL16sV2,
-				api.VMSizeStandardL32sV2,
-				api.VMSizeStandardL48sV2,
-				api.VMSizeStandardL64sV2:
+		for _, supportedSize := range supportedMasterVMSizes {
+			if supportedSize == vmSize {
 				return true
 			}
 		}
 	}
```
facchettos, line 90: same as before, a map seems better here

pkg/api/validate/vm.go
```diff
@@ -7,66 +7,97 @@ import (
...
+	api.VMSizeStandardE104idV5,
+	api.VMSizeStandardE104idsV5,
+	// Compute optimized
+	api.VMSizeStandardF72sV2,
+	// Memory and storage optimized
+	api.VMSizeStandardGS5,
+	api.VMSizeStandardG5,
+	// Memory optimized
+	api.VMSizeStandardM128ms,
+}
```
facchettos, line 31: Why not use a map[api.VMSize]struct{} or map[api.VMSize]bool instead of doing a lookup with a loop?

facchettos (changes requested):

pkg/api/validate/vm.go
cadenmarchese: Good call, I went with a `map[api.VMSize]bool`

cadenmarchese:
I've tested an install in our dev subscription using this PR, and opened another PR to add the instance types to our support policy when this merges: https://github.com/MicrosoftDocs/azure-docs-pr/pull/196323

pkg/api/validate/vm.go
```diff
@@ -7,66 +7,90 @@ import (
...
-		case api.VMSizeStandardD8sV3,
-			api.VMSizeStandardD16sV3,
-			api.VMSizeStandardD32sV3,
-			api.VMSizeStandardE64iV3,
-			api.VMSizeStandardE64isV3,
-			api.VMSizeStandardF72sV2,
-			api.VMSizeStandardGS5,
-			api.VMSizeStandardG5,
-			api.VMSizeStandardM128ms:
+		if supportedMasterVMSizes[vmSize] {
```
facchettos, line 85: maybe I am missing something but wouldn't `return supportedMasterVMSizes[vmSize] ` be sufficient here ? or do we also need to check for ?
```
	if supportedWorkerVMSizes[vmSize] && !requiredD2sV3Workers ||
		requiredD2sV3Workers && vmSize == api.VMSizeStandardD2sV3 {
		return true
	}
```

pkg/api/validate/vm.go
```diff
@@ -7,66 +7,90 @@ import (
...
-				api.VMSizeStandardL64sV2:
-				return true
-			}
-		}
 	}
+
+	if supportedWorkerVMSizes[vmSize] && !requiredD2sV3Workers ||
+		requiredD2sV3Workers && vmSize == api.VMSizeStandardD2sV3 {
+		return true
+	}
```
facchettos, line 93: Now that I look at it, I'm pretty sure this condition does not do what you expect because it is lacking parenthesis. 
```suggestion
	if (supportedWorkerVMSizes[vmSize] && !requiredD2sV3Workers) ||
		(requiredD2sV3Workers && vmSize == api.VMSizeStandardD2sV3) {
		return true
	}
```

pkg/api/validate/vm.go
cadenmarchese: `return supportedMasterVMSizes[vmSize]` sounds good to me. No need to check for requiredD2sV3workers for masters since we only use that feature on worker nodes in dev mode.

pkg/api/validate/vm.go
cadenmarchese: I'm not sure parentheses are needed, but for readability, we could go with something like this instead:
```
if supportedWorkerVMSizes[vmSize] && !requiredD2sV3Workers {
     return true
}

if requiredD2sV3Workers && vmSize == api.VMSizeStandardD2sV3 {
     return true
}
```
What do you think?

facchettos (approved):

pkg/api/validate/vm_test.go
```diff
@@ -0,0 +1,105 @@
...
+			requireD2sV3Workers: true,
+			isMaster:            true,
+			desiredResult:       true,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			result := VMSizeIsValid(tt.vmSize, tt.requireD2sV3Workers, tt.isMaster)
+
+			if result != tt.desiredResult {
+				t.Errorf("Want %v, got %v", tt.desiredResult, result)
```
ross-bryan, line 101: nit: I think this should begin with lower case, but can be fixed in a refactor/follow up later

pkg/api/validate/vm_test.go
```diff
@@ -0,0 +1,105 @@
...
+			name:          "size is not valid",
+			diskSize:      127,
+			desiredResult: false,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			result := DiskSizeIsValid(tt.diskSize)
+
+			if result != tt.desiredResult {
+				t.Errorf("Want %v, got %v", tt.desiredResult, result)
```
ross-bryan, line 33: nit: I think this should begin with lower case, but can be fixed in a refactor/follow up later

ross-bryan (approved):
lgtm, some nits to address in the future

PR #2088, opened 2022-04-26
Bump fluentbit, mdm, and mdsd images to mitigate P0/P1s

bennerv:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14258164

Need to bump the fluentbit, mdm, and mdsd images for P0/P1 vulnerabilities.  

The fluentbit image in `pkg/util/version/const.go` is not bumped as that doesn't need to be bumped at this time, only the RP one and a follow-up PR in RP-Config will be done to address that one.  

### Test plan for issue:

Run in INT, and confirm we still have logs coming to dgrep / kusto after RP deploy for the AROServiceLogs event/namespace.  

No


jewzaam (approved):

25region (approved):

bennerv:
This is going to fail E2E because I haven't mirrored the MDSD image which is used in the `openshift-azure-logging` namespace.  Mirroring then rerunning e2e.

PR #2089, opened 2022-04-26
Added unit test for the EnsureDeleted function of dynamichelper

xiaoyu74:
To improve unit test coverage for dynamichelper. This ADO task can be separated to multiple PRs for each function in the dynamichelper as suggested. This PR is for the function `EnsureDeleted`

The dynamichelper is used in multiple important parts of our codebase and need better unit test coverage

No, this PR is mainly to meet the requirement in the ADO task [13525478 - Improve unit test coverage for dynamichelper](https://msazure.visualstudio.com/AzureRedHatOpenShift/_boards/board/t/Skippy/Stories/?workitem=13525478)


pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -4,17 +4,92 @@ package dynamichelper
...
 	"github.com/Azure/ARO-RP/pkg/util/cmp"
+	mock_dynamichelper "github.com/Azure/ARO-RP/pkg/util/mocks/dynamichelper"
 )
 
+func TestEsureDeleted(t *testing.T) {
+	ctx := context.Background()
+
+	for _, tt := range []struct {
+		uname     string
+		grpkind   string
```
m1kola, line 28:
```suggestion
		groupKind   string
```

pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -4,17 +4,92 @@ package dynamichelper
...
+		namespace string
+		name      string
```
m1kola, line 30: We can rename this field to `objectName`, for example. This way it will not clash with test case `name`.

pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -4,17 +4,92 @@ package dynamichelper
...
+		wantErr   error
+		grpvr     *schema.GroupVersionResource
```
m1kola, line 32: `gvr` is a common abbreviation for this type. But we will probably not need it if we swtich to a fake REST client.

pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -4,17 +4,92 @@ package dynamichelper
...
+	}{
+		{
+			uname:     "Test EnsureDeleted 1",
```
m1kola, line 35: 1. Need more meaningfull test case names
2. What is the difference between this test case and the next one? They only seem to delete different objects of the same kind.

pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -4,17 +4,92 @@ package dynamichelper
...
+	} {
+		t.Run(tt.uname, func(t *testing.T) {
+			mockController := gomock.NewController(t)
+			defer mockController.Finish()
+
+			mockCore := mock_dynamichelper.NewMockGVRResolver(mockController)
+			mockCore.
+				EXPECT().
+				Resolve("configmap", "").
+				Return(nil, tt.wantErr).
```
m1kola, line 59: `tt.wantErr` - we usually call the field `wantSomething` in test cases for values we expect. In this case - `tt.wantErr` is mock data because in `Return` we specify what mock is going to return.

---

Unrelated to the above. To set up mock functions we usually have `mocks` function as a field on the test case struct. Like this:

https://github.com/Azure/ARO-RP/blob/d673482fb835b02e4b00dfe0e06135be143c3d07/pkg/api/validate/dynamic/diskencryptionset_test.go#L68-L78

And call it like this:

https://github.com/Azure/ARO-RP/blob/d673482fb835b02e4b00dfe0e06135be143c3d07/pkg/api/validate/dynamic/diskencryptionset_test.go#L274-L276

pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -4,17 +4,92 @@ package dynamichelper
...
+				AnyTimes()
```
m1kola, line 60: We have only one `dh.Resolve(groupKind, "")` call. Or am I missing something?

pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -4,17 +4,92 @@ package dynamichelper
...
+		},
+	} {
+		t.Run(tt.uname, func(t *testing.T) {
+			mockController := gomock.NewController(t)
+			defer mockController.Finish()
+
+			mockCore := mock_dynamichelper.NewMockGVRResolver(mockController)
+			mockCore.
+				EXPECT().
+				Resolve("configmap", "").
```
m1kola, line 58: We can use the data from the test case field. This way it will be easier to add test cases for differents kinds.

```suggestion
				Resolve(tt.grpkind, "").
```

pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -4,17 +4,92 @@ package dynamichelper
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/Azure/ARO-RP/pkg/util/cmp"
+	mock_dynamichelper "github.com/Azure/ARO-RP/pkg/util/mocks/dynamichelper"
 )
 
+func TestEsureDeleted(t *testing.T) {
```
m1kola, line 23: We need a test case to check that [not found](https://github.com/Azure/ARO-RP/blob/d673482fb835b02e4b00dfe0e06135be143c3d07/pkg/util/dynamichelper/dynamichelper.go#L75-L77) condition is handled properly.

pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -4,17 +4,92 @@ package dynamichelper
...
+			}
+
+			dh_3 := &dynamicHelper{GVRResolver: mockCore, delete: func(context.Context, *schema.GroupVersionResource, string, string) error {
+				return (tt.wantErr)
+			}}
+
+			err = dh_3.delete(ctx, tt.grpvr, tt.namespace, tt.name)
+			if !reflect.DeepEqual(err, tt.wantErr) {
+				t.Error(err)
+			}
```
m1kola, line 88: I don't understand why we have this. Could you please explain?

m1kola (changes requested):

pkg/util/dynamichelper/dynamichelper.go
```diff
@@ -62,6 +67,8 @@ func New(log *logrus.Entry, restconfig *rest.Config) (Interface, error) {
 		return nil, err
 	}
 
+	dh.delete = dh.deleteUsingRestCli
```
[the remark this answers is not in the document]
xiaoyu74: @m1kola - Just want to confirm that if you want me to replace gomock to use the fake rest client instead?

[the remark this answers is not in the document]
petrkotas: Hi @xiaoyu74, this is exactly what mikalai said. We have the client already in our vendor `vendor/k8s.io/client-go/rest/fake/fake.go` so you can import it as normal go library.

petrkotas (dismissed):
Thank you @xiaoyu74 for your work put into this. 
Everything that improves our unit test coverage is greatly welcome. I have clarified mikalai comment. Would you please look into it? Thank you

pkg/util/dynamichelper/dynamichelper.go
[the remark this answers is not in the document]
xiaoyu74: @m1kola @petrkotas - Sorry for the late response, I have update the unittest using the fake rest client you suggested, please feel free to review. Thanks.

pkg/util/dynamichelper/generate.go
```diff
@@ -4,5 +4,5 @@ package dynamichelper
 // Licensed under the Apache License 2.0.
 
 //go:generate rm -rf ../mocks/$GOPACKAGE
-//go:generate go run ../../../vendor/github.com/golang/mock/mockgen -destination=../mocks/$GOPACKAGE/$GOPACKAGE.go github.com/Azure/ARO-RP/pkg/util/$GOPACKAGE Interface
+//go:generate go run ../../../vendor/github.com/golang/mock/mockgen -destination=../mocks/$GOPACKAGE/$GOPACKAGE.go github.com/Azure/ARO-RP/pkg/util/$GOPACKAGE Interface,GVRResolver
```
m1kola, line 7: I see that we generate a mock for `GVRResolver`, but do not use it in this PR. We should either use `mockGVRResolver` or a mock generated by GoMock.

pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -18,11 +20,89 @@ import (
...
+		grpkind string
+	}{
+		{
+			uname:   "Test with the groupkind configmap",
+			grpkind: "configmap",
+		},
+		{
+			uname:   "Test with the groupkind baremetalhost",
+			grpkind: "baremetalhost.metal3.io",
+		},
```
m1kola, line 54: We have two test cases here with different group kinds, but `mockGVRResolver` only resolves one. So these two cases test the same code paths.

We should:
* Either get rid of test cases and subtests (`t.Run(...)`) and do a simple test with multiple sequential `EnsureDeleted` calls.
* Or have more meaningfull test cases & subtests. This will basically mean transforming sequential `EnsureDeleted` calls into sub tests.

Either option is fine.

m1kola (commented):
Looks good overal. But some cleanup is needed: we generate mocks which we do not use and we have test cases which test the same code paths.

pkg/util/dynamichelper/dynamichelper_test.go
xiaoyu74: @m1kola - I have updated the PR based on your suggestions, please feel free to check and see if I understood you correctly. Thanks a lot.

m1kola (approved):

PR #2090, opened 2022-04-26
Remove ACR Image Override for Geneva

bennerv:

Going forward, we will not override the Geneva images in FF Prod, but instead sync the images between gov cloud and public cloud.  

Removes unnecessary environment variable as we will now sync them.  

### Test plan for issue:

N/A

Yes, docs have been updated
https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki?wikiVersion=GBmaster&_a=edit&pagePath=/ARO%20MAG/How%20To/Mirroring&pageId=174102

### DIFF
--- a/pkg/util/version/const.go
+++ b/pkg/util/version/const.go
@@ -4,8 +4,6 @@ package version
 // Licensed under the Apache License 2.0.
 
 import (
-	"os"
-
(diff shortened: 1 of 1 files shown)


s-amann (dismissed):
lgtm

ross-bryan (approved):

PR #2091, opened 2022-04-27
add minor version

gvanderpotte:

We need a quick and live way to check the list of OCP minor versions along with the resourceID and subscriptionID to easily identify out of support lifecycle clusters. Due to Geneva limitations, we need to provide the minor version along with the full version to be able to generate corresponding report on Geneva side. 

Also pushes the GA version to allow out of lifecycle detection from Geneva

We need a quick and live way to check the list of OCP minor versions along with the resourceID and subscriptionID to easily identify out of support lifecycle clusters. Due to Geneva limitations, we need to provide the minor version along with the full version to be able to generate corresponding report on Geneva side. 

Also pushes the GA version to allow out of lifecycle detection from Geneva

### Test plan for issue:

- Are there unit tests? yes
- Are there integration/e2e tests? no

### DIFF
--- a/pkg/monitor/cluster/clusterversions.go
+++ b/pkg/monitor/cluster/clusterversions.go
@@ -38,14 +38,26 @@ func (mon *Monitor) emitClusterVersions(ctx context.Context) error {
 		availableRP = version.GitCommit
 	}
 
+	actualVersion := actualVersion(cv)
+	actualMinorVersion := ""
+	if len(actualVersion) > 0 {
+		parsedVersion, err := version.ParseVersion(actualVersion)
+		if err != nil {
+			return err
+		}
+		actualMinorVersion = parsedVersion.MinorVersion()
+	}
+
 	mon.emitGauge("cluster.versions", 1, map[string]string{
-		"actualVersion":                        actualVersion(cv),
+		"actualVersion":                        actualVersion,
 		"desiredVersion":                       desiredVersion(cv),
 		"provisionedByResourceProviderVersion": mon.oc.Properties.ProvisionedBy,              // last successful Put or Patch
 		"resourceProviderVersion":              version.GitCommit,                            // RP version currently running
 		"operatorVersion":                      operatorVersion,                              // operator version in the cluster
 		"availableVersion":                     availableVersion(cv, version.UpgradeStreams), // current available version for upgrade from stream
 		"availableRP":                          availableRP,                                  // current RP version available for document update, empty when none
+		"latestGaMinorVersion":                 version.InstallStream.Version.MinorVersion(), // Latest GA in ARO Minor version
+		"actualMinorVersion":                   actualMinorVersion,                           // Minor version, empty if actual version is not in expected form
 	})
 
 	return nil
--- a/pkg/util/version/version.go
+++ b/pkg/util/version/version.go
@@ -77,3 +77,7 @@ func (v *Version) Eq(w *Version) bool {
 	}
 	return true
 }
+
+func (v *Version) MinorVersion() string {
+	return fmt.Sprintf("%d.%d", v.V[0], v.V[1])
+}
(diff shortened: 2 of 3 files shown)


pkg/monitor/cluster/clusterversions.go
```diff
@@ -38,14 +39,21 @@ func (mon *Monitor) emitClusterVersions(ctx context.Context) error {
 		availableRP = version.GitCommit
 	}
 
+	actualVersion := actualVersion(cv)
+	actualMinorVersion := ""
+	if len(actualVersion) > 0 {
+		actualMinorVersion = actualVersion[0:strings.LastIndex(actualVersion, ".")]
+	}
```
m1kola, line 46: I think version can be `4.3.0-0.nightly-2020-04-17-062811` where last `.` is part of suffix.

You might want to use `version.ParseVersion(...)` instead.

ross-bryan (dismissed):
lgtm only concern is another column which our log aggregation seems to have trouble with

gvanderpotte:
> lgtm only concern is another column which our log aggregation seems to have trouble with

I think that log column aggregation and metric column aggregation are different processes, so we should be safe here. That said, this does not prevent us from doing cleaning of unused pre-aggregates

gvanderpotte:
Added push of latest GA version from RP perspective. It is not super optimal from pure monitor perspective, as this could be pushed as an RP metric (as it is the same for all clusters in a given region). That said, this would force us to make a join on Geneva side (between metrics from different accounts), while the goal here is to have an easy way to detect to detect out of support lifecycle clusters

pkg/util/version/version.go
```diff
@@ -77,3 +77,7 @@ func (v *Version) Eq(w *Version) bool {
 	}
 	return true
 }
+
+func (v *Version) GetMinorVersion() string {
```
m1kola, line 81:
```suggestion
func (v *Version) MinorVersion() string {
```

As per https://go.dev/doc/effective_go#Getters

pkg/monitor/cluster/clusterversions.go
```diff
@@ -38,14 +39,26 @@ func (mon *Monitor) emitClusterVersions(ctx context.Context) error {
...
 	}
 
+	actualVersion := actualVersion(cv)
+	actualMinorVersion := ""
+	if len(actualVersion) > 0 {
+		parsedVersion, err := version.ParseVersion(actualVersion)
+		if err != nil {
+			return err
+		}
+		actualMinorVersion = fmt.Sprintf("%d.%d", parsedVersion.V[0], parsedVersion.V[1])
```
m1kola, line 49: Why not `parsedVersion.MinorVersion()`?

pkg/monitor/cluster/clusterversions.go
```diff
@@ -38,14 +39,26 @@ func (mon *Monitor) emitClusterVersions(ctx context.Context) error {
...
-		"availableVersion":                     availableVersion(cv, version.UpgradeStreams), // current available version for upgrade from stream
-		"availableRP":                          availableRP,                                  // current RP version available for document update, empty when none
+		"provisionedByResourceProviderVersion": mon.oc.Properties.ProvisionedBy,                 // last successful Put or Patch
+		"resourceProviderVersion":              version.GitCommit,                               // RP version currently running
+		"operatorVersion":                      operatorVersion,                                 // operator version in the cluster
+		"availableVersion":                     availableVersion(cv, version.UpgradeStreams),    // current available version for upgrade from stream
+		"availableRP":                          availableRP,                                     // current RP version available for document update, empty when none
+		"latestGaMinorVersion":                 version.InstallStream.Version.GetMinorVersion(), // Latest GA in ARO Minor version
```
m1kola, line 60: I'm a bit puzzled by presence of `resourceProviderVersion` and `latestGaMinorVersion` here: this metric is related to cluster. But the above demensions are related RP. Why do we have them here?

m1kola (changes requested):

pkg/monitor/cluster/clusterversions.go
m1kola: @gvanderpotte

> resourcePRoviderVersion is the version of the RP that has provisionned the cluster (last PUCM by).

No, `resourcePRoviderVersion` is a version of the RP which sends the metric. It is static `version.GitCommit` constant.

Last PUCM'ed by is `provisionedByResourceProviderVersion`

m1kola (approved):
Looks good to me, assuming that we cant not solve geneva limitation in some other way.

PR #2093, opened 2022-04-27
Update FIPs e2e test for 2022-04-01 API

tony-schndr:

Fixes 12385409

This PR updates FIPs e2e test to use the 2022-04-01 API.  The e2e test will retrieve the cluster document with the new API and check if FIPs is enabled or disabled.  Then it will assert the expected machine configs are either present or absent.

### Test plan for issue:
- Are there integration/e2e tests? yes

No

### DIFF
--- a/test/e2e/fips.go
+++ b/test/e2e/fips.go
@@ -5,12 +5,13 @@ package e2e
 
 import (
 	"context"
-	"fmt"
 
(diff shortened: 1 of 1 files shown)


bennerv:
Holding on #2035 to be merged and this rebased on that.

ross-bryan (approved):
lgtm

PR #1863, opened 2021-12-01
Cluster install CloudError support

darthhexx:
### Which issue this PR addresses:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/9224646/

### What this PR does / why we need it:

The goal of this PR is to provide more informative error messages about which step in a cluster installation or admin update failed without sending or attempting to sanitise raw error messages. Raw error messages are always logged, but only the "canned" responses are returned to the end-user.

### Test plan for issue:

There are existing unit tests for the condition timeout function, which have been adapted to test against the new timeout error message.

### Is there any documentation that needs to be updated for this PR?

https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/200186/-WIP-Cluster-installation-failures

PR #2094, opened 2022-04-27
Add additional values to CloudError and Cluster Operation Logs

carvalhe:
Fixes (https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14217751)

Cloud error needs more details to show if an issue is client or server error, as well as where the error is coming from. Added to the struct for cloud error to include this information. This doesn't refactor cloud errors made in the code as that will be done in the future.
With cluster health monitoring being adjusted, more detailed information is needed in the logs. Created a result type where there is a result type of 0, 1, 2 being success, user error, internal server error respectively output to the logs. I also included the operation type to the logs.

### Test plan for issue:
Tested manually by using tests that exist in code already and verifying the additional columns for cloud error.

No, just adding more details to errors and logs that already exist.


m1kola (changes requested):
I had a look [at the ticket](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14217751) and it is very light on details. Could you please explain what issue we are trying to solve and provide more context?

carvalhe:
@m1kola Right now in kusto ArmLogs, when a cluster creation/deletion failure occurs, the logs show that the issue is an internal server error. When looked into further it's actually a client error but ARO generates a generic server error instead. As we move towards checking cluster health, sanitizing the signal will be helpful for monitoring the quality of service for clusters. This PR adds more to the framework to provide this information instantly for error type, and refactoring will be done in the future.

pkg/api/error.go
```diff
@@ -39,6 +39,12 @@ type CloudErrorBody struct {
 	// The target of the particular error. For example, the name of the property in error.
 	Target string `json:"target,omitempty"`
 
+	// The category for the error, being either a user or server error
+	Category string
+
+	// The location that depends on this functioning correctly, for example OpenShiftInstaller would be a dependency
```
weinong, line 45:
```suggestion
	// Dependency is the source where this error originates from
```

weinong:
Hi @m1kola , are you good on what we are trying to do now? If not, you can also reach out to Peter and Vikas. Please let us know how to unblock it. Thanks!

pkg/backend/openshiftcluster.go
```diff
@@ -30,13 +30,28 @@ type openShiftClusterBackend struct {
 	newManager func(context.Context, *logrus.Entry, env.Interface, database.OpenShiftClusters, database.Gateway, encryption.AEAD, billing.Manager, *api.OpenShiftClusterDocument, *api.SubscriptionDocument) (cluster.Interface, error)
 }
 
+type clusterResult struct {
```
weinong, line 33: operationResult seems more generic

weinong: please also implement String() function

pkg/backend/openshiftcluster.go
```diff
@@ -235,6 +255,9 @@ func (ocb *openShiftClusterBackend) updateAsyncOperation(ctx context.Context, lo
 				asyncdoc.OpenShiftCluster = &ocCopy
 			}
 
+			clusterResult := newClusterResult(resultType, initialProvisioningState.String(), backendErr)
+			log.Print("Long running result: %v", *clusterResult)
```
weinong, line 259: use %s and add String() to clusterResult

carvalhe: Changed to `log.Printf("Long running result: %s", clusterResult.String())`

pkg/backend/openshiftcluster.go
carvalhe: Changed name to operationResult, and created a String() function to simplify things.

pkg/backend/openshiftcluster.go
```diff
@@ -44,14 +44,18 @@ func newOpenShiftClusterBackend(b *backend) *openShiftClusterBackend {
...
+func newOperationResult(resultType int, operationType string, backendErr error) *operationResult {
+	return &operationResult{
 		resultType:    resultType,
 		operationType: operationType,
 		errorDetails:  backendErr.Error(),
 	}
 }
 
+func (o *operationResult) String() string {
+	return fmt.Sprintf("%d: %s: %s:", o.resultType, o.operationType, o.errorDetails)
```
weinong, line 56:
```suggestion
	return fmt.Sprintf("operationType: %s, resultType: %d, errorDetails: %s:", o.resultType, o.operationType, o.errorDetails)
```

pkg/api/error.go
```diff
@@ -56,7 +62,7 @@ func (b *CloudErrorBody) String() string {
 		}
 	}
 
-	return fmt.Sprintf("%s: %s: %s%s", b.Code, b.Target, b.Message, details)
+	return fmt.Sprintf("%s: %s: %s: %s: %s%s", b.Code, b.Target, b.Message, b.Category, b.Dependency, details)
```
weinong, line 65: for readability, lets put
code: %s, target: %s ...

carvalhe: Sure thing, it would make it more clear for sure

pkg/api/error.go
```diff
@@ -39,6 +39,12 @@ type CloudErrorBody struct {
 	// The target of the particular error. For example, the name of the property in error.
 	Target string `json:"target,omitempty"`
 
+	// The category for the error, being either a user or server error
+	Category string
```
weinong, line 43:
```suggestion
	Category string `json:"-"`
```

pkg/api/error.go
```diff
@@ -39,6 +39,12 @@ type CloudErrorBody struct {
...
+
+	// Dependency is the source where this error originates from
+	Dependency string
```
weinong, line 46:
```suggestion
	Dependency string `json:"-"`
```

pkg/api/error.go
carvalhe: Added to both

pkg/api/error.go
```diff
@@ -39,6 +39,12 @@ type CloudErrorBody struct {
 	// The target of the particular error. For example, the name of the property in error.
 	Target string `json:"target,omitempty"`
 
+	// The category for the error, being either a user or server error
+	Category string `json:"-"`
```
mbarnes, line 43: Suggest defining a `CloudErrorCategory` string type for this, if it can only be one of the two constants defined below.

mbarnes (changes requested):

pkg/api/error.go
carvalhe: Created CloudErrorCategory string type as suggested.

pkg/backend/openshiftcluster.go
```diff
@@ -215,10 +235,14 @@ func (ocb *openShiftClusterBackend) updateAsyncOperation(ctx context.Context, lo
 				// asyncOperations errors. Otherwise - return generic error
 				err, ok := backendErr.(*api.CloudError)
 				if ok {
-					log.Print(backendErr)
+					if err.CloudErrorBody.Category != "" && err.CloudErrorBody.Category == api.AROUserError {
+						resultType = 1
```
petrkotas, line 239: Could you please add the result type as a constant? This would make the code more easily understandable as we have:

```
0 == sucess
1 == user error
2 == internal server error
```

Thank you.

pkg/api/error.go
```diff
@@ -109,6 +123,20 @@ func NewCloudError(statusCode int, code, target, message string, a ...interface{
 	}
 }
 
+// NewCloudError returns a new CloudError with category and dependency
+func NewCloudErrorWithCategory(statusCode int, code, target, message string, category CloudErrorCategory, dependency string, a ...interface{}) *CloudError {
```
petrkotas, line 127: Where is this function used? We have a policy: "No dead code in the repo."
If it is used and will be referenced in the followup PR would you please add it to the followup?
Thank you.

pkg/backend/openshiftcluster.go
```diff
@@ -235,6 +259,9 @@ func (ocb *openShiftClusterBackend) updateAsyncOperation(ctx context.Context, lo
 				asyncdoc.OpenShiftCluster = &ocCopy
 			}
 
+			clusterResult := newOperationResult(resultType, initialProvisioningState.String(), backendErr)
+			log.Printf("Long running result: %s", clusterResult.String())
```
petrkotas, line 263: I am wondering, if this is the only place which uses the result, who will be reading that log?
If it is for us (SREs) I would love to see the text representation returned instead of an `int`.
If it is for analytics in the Kusto, I believe there should be no difference in creating categories based on `string` or `int`.

petrkotas (changes requested):
Hi @carvalhe thank you taking care of the functionality that will make our life easier. I appreciate it.
I tried to undestand the work and proposed few changes.
Thank you for your effort.

m1kola (dismissed):
@carvalhe thanks for the explanation. I still am probably missing some background on what you are doing. 


I'll try to give a bit more info on how we handle and log errors below and suggest alternative approaches.

> Right now in kusto ArmLogs, when a cluster creation/deletion failure occurs, the logs show that the issue is an internal server error. When looked into further it's actually a client error but ARO generates a generic server error instead. 

The idea is that across our codebase:

* We create `api.CloudError` errors if we want to propagate these errors to customers. I think this is what you refer to as "client error".
* We create other error types if we don't want to propagate them to customers. Other types of errors get converted into generic "Internal server error." `api.CloudError` before we return them. However these generic errors will have a special error code which can be used to separate generic errors from client errors.
    * **Important note**: there are some cases when we return a generic error to a customer, but should actually return something meaningful. Some attempts to improve the situation were made [here](https://github.com/Azure/ARO-RP/pull/1863).

We can distinguish client error from generic ones by [error code](https://github.com/Azure/ARO-RP/blob/e070358150e7edc1f5793d2deafadaca66377a61/pkg/api/error.go#L33-L34) in `api.CloudError` body: generic errors will have `api.CloudErrorCodeInternalServerError` error code. Other errors will have some other error code ([full list](https://github.com/Azure/ARO-RP/blob/e070358150e7edc1f5793d2deafadaca66377a61/pkg/api/error.go#L65-L97)). This code is intended to be used programmatically.

The above applies to both frontend (sync operations) and backend (async operations). Here is where we convert errors:

* [On backend](https://github.com/Azure/ARO-RP/blob/e070358150e7edc1f5793d2deafadaca66377a61/pkg/backend/openshiftcluster.go#L216-L226)
* [On frontend](https://github.com/Azure/ARO-RP/blob/e070358150e7edc1f5793d2deafadaca66377a61/pkg/frontend/frontend.go#L378-L389)

There are few other places where we return a generic "Internal server error.". For example when we [recover from a panic](https://github.com/Azure/ARO-RP/blob/e070358150e7edc1f5793d2deafadaca66377a61/pkg/frontend/middleware/panic.go#L18-L24). But all the rest of the places are truly internal errors.

**Question:** can you use error code to distinguish between the errors? `CloudErrorCategory` looks like a more generic duplicate of error code to me.

> Additionally, in the future the plan is to add a new simplified QoS table in kusto. We want to add details to the logs to act similarly as AKS where there is a result type of 0, 1, 2 being success, user error, internal server error respectively.

Is this about async operations specifically? E.g. you want to have more visibility into cluster creation/deletion/update success/failures?

For async operations we have [this code](https://github.com/Azure/ARO-RP/blob/e070358150e7edc1f5793d2deafadaca66377a61/pkg/backend/openshiftcluster.go#L258-L267). This is where we:

* Update async operation record in the DB once the operation is actually complete
* And emit metrics on async operations (duration of the operations and count, see [`emitMetrics`](https://github.com/Azure/ARO-RP/blob/e070358150e7edc1f5793d2deafadaca66377a61/pkg/backend/openshiftcluster.go#L288-L304)). These metrics include old and new provisioning states as dimensions.

**Question**: Can you use metrics for what you are doing?

If we absolutely have to use logs - then we need to change the approach. Instead of logging type of the result and operation name in the message we should look into adding new log fields (e.g `log.WithField("...", "...")`).

pkg/backend/openshiftcluster.go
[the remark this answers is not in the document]
weinong: > Question: can you use error code to distinguish between the errors? CloudErrorCategory looks like a more generic duplicate of error code to me.

Nope. `CloudErrorCodes` lives in the repo and there is no easy way from kusto/metric side to extract that unless hardcoding them which is also no ideal. Yes, a bit of duplicate can get us so much benefit in terms of readability and usability.

> Is this about async operations specifically? E.g. you want to have more visibility into cluster creation/deletion/update success/failures?

yes, we want to track the result of cluster operations such as Create, Delete, Update, etc.

> Question: Can you use metrics for what you are doing?

Metric is only useful at high level view and alerting, but it loses the slice and dice capabilities from kusto to do further analysis

> If we absolutely have to use logs - then we need to change the approach. Instead of logging type of the result and operation name in the message we should look into adding new log fields (e.g log.WithField("...", "...")).

Ideally, we want this event in a different table. Typically, in Azure (CRP, AKS, NRP, you name it), there is a table like `AsyncQosEvents` that records the outcome of long running operations. We would also like to do that for ARO, but getting insights first will provide biggest bang of the bucks.
I don't get why it _has_ to be using `.WithFields()`. We already print the error in `MESSAGE` today, and we really care less of adding more columns to `ServiceLogs` table. Again, the goal is to write them into different table and we can either use Kusto to re-ingest or update mdsd/fluent-bit to do that, whichever is easier.

All we are doing is bringing all the experiences from Azure to ARO that will only help the service :)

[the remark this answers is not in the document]
weinong: To learn more on what we want to achieve, you can also read more in the Resource Provider Level section in https://docs.google.com/document/d/12UG4rpU9uQCziJ0EKsU-rxSqIO1dPgaoo4-2yi1nzr4/edit

pkg/backend/openshiftcluster.go
carvalhe: Created a resultType constant as mentioned, and changed to be a string instead of int.

pkg/api/error.go
carvalhe: Removed and will place in follow-up PR.

pkg/backend/openshiftcluster.go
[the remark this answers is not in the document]
m1kola: > Nope. `CloudErrorCodes` lives in the repo and there is no easy way from kusto/metric side to extract that unless hardcoding them which is also no ideal. Yes, a bit of duplicate can get us so much benefit in terms of readability and usability.

Yes, `CloudErrorCodes` live in the ARO-RP repo, but this is an API: it is very unlikely that we will break support for this. If we do - it will be fairly high priority for us to fix. Besides that - to distinguish between "client errors" and everything else we don't need to know all the error codes. The logic is basically:

* If the error has `api.CloudErrorCodeInternalServerError` error code - it is an RP error
* If the error has the error code which is not `api.CloudErrorCodeInternalServerError` - it is a 'client error'"

I understand that we would like to avoid hardcoding. But I also want to avoid the need to keep error codes in sync with `CloudErrorCategory` across the codebase and having everyone on the team to remember that they need to make sure that their new error is being logged with the right category.

> Ideally, we want this event in a different table. Typically, in Azure (CRP, AKS, NRP, you name it), there is a table like `AsyncQosEvents` that records the outcome of long running operations.

Yep, that is fine. I don't know how to do that, but I'm sure we can find a way.

> I don't get why it _has_ to be using `.WithFields()`. We already print the error in `MESSAGE` today, and we really care less of adding more columns to `ServiceLogs` table.

We can print error in the message. But we should not print category or other additional information as part of the message. The reason is the same why we don't put resource ID, correlation ID and other data into message itself and use structured logs - usability. I don't want to have extra noise in the message and want them to be human readable. I want to be able to easily filter messages in Geneva. And I want to make sure that we don't have to worry about breaking format and breaking downstream consumers which rely on a message being formatted in a specific way.

I'm looking at this mostly from the SRE perspective and manual/semi-manual consumption, but I think having a separate field will also be beneficial for programmatic consumption (e.g. less chances of breaking apps by breaking the format).

> To learn more on what we want to achieve, you can also read more in the Resource Provider Level section in https://docs.google.com/document/d/12UG4rpU9uQCziJ0EKsU-rxSqIO1dPgaoo4-2yi1nzr4/edit

Thanks for sharing this. I'll take a look.

pkg/backend/openshiftcluster.go
```diff
@@ -239,34 +267,39 @@ func (ocb *openShiftClusterBackend) updateAsyncOperation(ctx context.Context, lo
...
+		operationResult, err := ocb.updateAsyncOperation(ctx, log, doc.AsyncOperationID, doc.OpenShiftCluster, provisioningState, failedProvisioningState, backendErr)
 		if err != nil {
 			return err
 		}
-
+		log.WithFields(logrus.Fields{
+			"resultType":    operationResult.resultType,
+			"operationType": initialProvisioningState.String(),
+			"errorDetails":  operationResult.errorDetails,
+		}).Print("Long running operation result")
```
m1kola, line 298: Can we do something like this?

1. Keep signature of `updateAsyncOperation` unchanged. In this iteration of the PR this method returns `operationResult`. And one might think that we always call `updateAsyncOperation` when an async operation is complete (which is not the case). Not returning `operationResult` will help to avoid confusion.
2. Drop `operationResult` struct altogether and incapsulate result logging in a function (see an example below).
    * We might want to keep `resultType` constatns somewhere, but we should probably find it a new home. Maybe somewhere under `github.com/Azure/ARO-RP/pkg/util/log`?

```suggestion
		ocb.asyncOperationResultLog(log, initialProvisioningState, backendErr)
```

P.S.: Better name suggestions for `asyncOperationResultLog` are welcome :)

And `asyncOperationResultLog` will be something like this:

```go
func (ocb *openShiftClusterBackend) asyncOperationResultLog(log *logrus.Entry, initialProvisioningState api.ProvisioningState, backendErr error) {
	log = log.WithFields(logrus.Fields{
		"resultType":    successResultType,
		"operationType": initialProvisioningState.String(),
	})

	if backendErr == nil {
		log.Info("operation successfully completed")
		return
	}

	_, ok := backendErr.(*api.CloudError)
	if ok {
		log = log.WithField("resultType", userErrorResultType)
	} else {
		log = log.WithField("resultType", serverErrorResultType)
	}

	log = log.WithField("errorDetails", backendErr.Error())
	log.Info("operation failed")
	// Or maybe just add an error into `msg` like this,
	// instead of 2 lines of code above?
	log.Error(backendErr)
}

```

m1kola (commented):
I left few suggestions on logging. Also I would like to separate changes for logging from adding dependencies into errors.

pkg/api/error.go
```diff
@@ -39,6 +39,9 @@ type CloudErrorBody struct {
 	// The target of the particular error. For example, the name of the property in error.
 	Target string `json:"target,omitempty"`
 
+	// Dependency is the source where this error originates from
+	Dependency string `json:"-"`
```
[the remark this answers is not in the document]
carvalhe: Yes I agree, meant to remove this before you rereviewed the PR. Removing now to focus on just logging as mentioned.

pkg/backend/openshiftcluster.go
```diff
@@ -285,6 +286,28 @@ func (ocb *openShiftClusterBackend) endLease(ctx context.Context, log *logrus.En
...
+
+	_, ok := backendErr.(*api.CloudError)
+	if ok {
+		log = log.WithField("resultType", utillog.UserErrorResultType)
+	} else {
+		log = log.WithField("resultType", utillog.ServerErrorResultType)
+	}
+
+	log = log.WithField("errorDetails", backendErr.Error())
+	log.Info("Long running operation failed")
```
ross-bryan, line 308: nit: we mostly have all lower case for logging in code base.

ross-bryan (approved):
thanks for this PR.  after reviewing conversation I can see it has gone through some iterations and this one looks clean! I do have concerns about more columns being added but it seems for now we will deal with it. one nit but lgtm

ross-bryan (approved):
lgtm

rogbas (approved):
lgtm

PR #2096, opened 2022-04-29
Add NRMS/Hypernet tags to cluster resource group in CI mode

cadenmarchese:

The RP fails cluster installs if NSGs are attached to subnets. Simply Secure v1 applies NSGs in our subscriptions when these tags aren't present, resulting in a race condition.

Adds the tags needed to stop SSv1 from applying NSGs to subnets: https://eng.ms/docs/security-compliance-identity-and-management-scim/security/azure-security/security-health-analytics/network-isolation/netiso-teamdocs/netiso-program-overview/manage/hypernet-and-nrms-simply-secure-v1-network-security-rules

### Test plan for issue:

Deploy this commit to central us euap (where the issue is faced)

No

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -155,6 +155,13 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 		c.log.Infof("creating resource group")
 		_, err = c.groups.CreateOrUpdate(ctx, vnetResourceGroup, mgmtfeatures.ResourceGroup{
 			Location: to.StringPtr(c.env.Location()),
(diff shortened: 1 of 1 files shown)


s-amann (dismissed):

rogbas (dismissed):
LGTM

cadenmarchese:
`hold` for now as tagging the resource group didn't have the desired effect. We'll need to update these tags with a temporary SSv1 exception number and re-test.

ross-bryan (commented):
We need to get an exception from microsoft, I understand this code was for testing purposes but it should not make it into production.

cadenmarchese:
@ross-bryan we can open a new PR with exception tags later if needed. Closing this one.

PR #2099, opened 2022-05-02
Generate the InfraID without Installer dependency

hawkowl:

Part of cleanup for M5: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14317614

This generates the Infra ID directly, in the same manner as the Installer. This removes a (small) dependency on it, and means we keep it consistent between infra versions. We need the InfraID before the Installer is run to create the Gateway, among other things.

### Test plan for issue:

Unit tests for the code, which can be run before and after the change to verify behaviour.

N/A

### DIFF
--- a/pkg/cluster/deploystorage.go
+++ b/pkg/cluster/deploystorage.go
@@ -8,7 +8,9 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"net/http"
+	"regexp"
 	"strings"
 
 	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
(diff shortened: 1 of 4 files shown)


pkg/cluster/deploystorage.go
```diff
@@ -302,3 +288,29 @@ func (m *manager) setMasterSubnetPolicies(ctx context.Context) error {
...
 	return m.subnet.CreateOrUpdate(ctx, m.doc.OpenShiftCluster.Properties.MasterProfile.SubnetID, s)
 }
+
+// generateInfraID take base and returns a ID that
+// - is of length maxLen
+// - contains randomLen random bytes
+// - only contains `alphanum` or `-`
+// see openshift/installer/pkg/asset/installconfig/clusterid.go for original implementation
+func generateInfraID(base string, maxLen int, randomLen int) string {
+	maxBaseLen := maxLen - (randomLen + 1)
```
cadenmarchese, line 298: Looks good to me, just out of curiosity - why do we set a maximum?

hawkowl: The rationale appears to be that you have a global maximum, and then a maximum of provided characters, so that you leave a few for randomness. Some clouds don't like long IDs, not sure if Azure is really one of them.

s-amann (approved):

m1kola (dismissed):
I had a very quick look. Need to properly review it.

If I understand this PR correctly - we get rid of installer dependency and generate infra id ourselves in the PR. And store it in the DB. However the installer depends on infra ID a lot and we do not pass it into the installer in any way.

Or did I miss something?

hawkowl:
@m1kola it gets "passed" in ensureGraph, where the actual installConfig is finalised by compiling the parents in the graph to produce the final asset.  The graph in here is because the generation happens when the graph is generated, and the actual installConfig that is passed in is completely unused, as far as I can tell.

faves4less (dismissed):

darthhexx (approved):
lgtm

PR #2104, opened 2022-05-03
Split ensureGraph into customisations and saving

hawkowl:

Part of M5 installer cleanup https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14317614

If we move to using Hive for the installer, we will have to save the InstallConfig that is provided to us *after* we call out to Hive to perform the installation. For the current install method, we can get closer to this by making any customisations before the deployment (which would be removed post-Hive) and then saving the graph after (which would remain after Hive).

### Test plan for issue:

E2E should test it.

N/A

### DIFF
--- a/pkg/cluster/deploystorage.go
+++ b/pkg/cluster/deploystorage.go
@@ -170,30 +170,24 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
 	return arm.DeployTemplate(ctx, m.log, m.deployments, resourceGroup, "storage", t, nil)
 }
 
-func (m *manager) ensureGraph(ctx context.Context, installConfig *installconfig.InstallConfig, image *releaseimage.Image) error {
-	resourceGroup := stringutils.LastTokenByte(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
-	clusterStorageAccountName := "cluster" + m.doc.OpenShiftCluster.Properties.StorageSuffix
-	infraID := m.doc.OpenShiftCluster.Properties.InfraID
-
-	exists, err := m.graph.Exists(ctx, resourceGroup, clusterStorageAccountName)
-	if err != nil || exists {
-		return err
-	}
-
+// applyInstallConfigCustomisations modifies the InstallConfig and creates
+// parent assets, then regenerates the InstallConfig for use for Ignition
+// generation, etc.
+func (m *manager) applyInstallConfigCustomisations(ctx context.Context, installConfig *installconfig.InstallConfig, image *releaseimage.Image) (graph.Graph, error) {
 	clusterID := &installconfig.ClusterID{
 		UUID:    m.doc.ID,
-		InfraID: infraID,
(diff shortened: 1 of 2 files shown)


s-amann (dismissed):

pkg/cluster/deploystorage.go
```diff
@@ -169,30 +169,22 @@ func (m *manager) deployStorageTemplate(ctx context.Context, installConfig *inst
...
-	resourceGroup := stringutils.LastTokenByte(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
-	clusterStorageAccountName := "cluster" + m.doc.OpenShiftCluster.Properties.StorageSuffix
-	infraID := m.doc.OpenShiftCluster.Properties.InfraID
-
-	exists, err := m.graph.Exists(ctx, resourceGroup, clusterStorageAccountName)
-	if err != nil || exists {
-		return err
-	}
-
+// applyInstallConfigCustomisations modifies the InstallConfig and creates parent assets to generate the final installconfig that is used to
```
SudoBrendan, line 172: that is used to...?

SudoBrendan (dismissed):
LGTM! One nit if you want to look.

pkg/cluster/deploystorage.go
hawkowl: oops, fixed

pkg/cluster/deploystorage.go
```diff
@@ -170,30 +170,28 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
...
-	}
-
+// applyInstallConfigCustomisations modifies the InstallConfig and creates
+// parent assets, then regenerates the InstallConfig for use for Ignition
+// generation, etc.
+func (m *manager) applyInstallConfigCustomisations(
+	ctx context.Context,
+	installConfig *installconfig.InstallConfig,
+	image *releaseimage.Image,
+) (graph.Graph, error) {
```
darthhexx, line 180: This function definition should be on a single line.

pkg/cluster/deploystorage.go
```diff
@@ -219,18 +217,30 @@ func (m *manager) ensureGraph(ctx context.Context, installConfig *installconfig.
...
 		if err = m.overrideEthernetMTU(g); err != nil {
-			return err
+			return nil, err
 		}
 	}
 
+	return g, nil
+}
+
+func (m *manager) ensureGraph(ctx context.Context, g graph.Graph) error {
```
darthhexx, line 235: Now that this function only saves the graph, IMO it would be clearer to rename this function to `saveGraph` or `persistGraph`.

darthhexx (commented):
LGTM, but I do have 2 nits around style and naming.

pkg/cluster/deploystorage.go
hawkowl: ack

pkg/cluster/deploystorage.go
hawkowl: ack

darthhexx (approved):
lgtm

PR #2105, opened 2022-05-03
Move armDeployTemplate into util

hawkowl:

Part of M5 installer cleanup: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14317614

When the installer code is isolated we will need to use the `armDeployTemplate` code in two locations. This moves it out of `pkg/cluster/` along with the tests, and also removes the need for a wrapper function in `pkg/cluster/`. We could move it into an interface and mock it later, if we wanted.

### Test plan for issue:

Unit tests follow the move, plus E2E should catch any problems.

N/A


s-amann (approved):

bennerv (approved):
LGTM

PR #2108, opened 2022-05-03
Refactor cronjobs on CI VMs and add job to clean up go cache

bennerv:

Fixes [ADO14324822](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14324822)

Go build cache is filling up CI VMs.  Remove old files not accessed within 24 hours

### Test plan for issue:

Build CI VMs, and check the issue is resolved after ~2 days or so.

Nope


bennerv:
Need to update the CI VMs.  Will prob do so tomorrow.

ross-bryan (dismissed):
deleted more than you added, lgtm

pkg/deploy/generator/resources_dev.go
```diff
@@ -380,24 +382,10 @@ fi
 EOF
 chmod +x /usr/local/bin/fix-podman-pause.sh
 
-# HACK - /tmp will fill up causing build failures
-# delete anything not accessed within 2 days
-cat >/usr/local/bin/clean-tmp.sh <<'EOF'
-#!/bin/bash
-
-find /tmp -type f \( ! -user root \) -atime +2 -delete
```
SudoBrendan, line 388: If I'm reading this correctly, it looks like the replaced command `/sbin/tmpwatch 24h /tmp` will delete additional files (e.g. root-owned files). Is this intended/ok?

facchettos (changes requested):
Wouldn't that be a good occasion to get rid of bindata here and use embed instead?

pkg/deploy/generator/resources_dev.go
bennerv: If they're not accessed within 24 hours yes it will delete them.  This is fine as on RHEL8 `/tmp` is cleaned up on reboot anyway so all files are removed.

bennerv:
> Wouldn't that be a good occasion to get rid of bindata here and use embed instead?

+1 @facchettos will do that when i rebase + test.

bennerv:
Closing as we increased the disk size.  If this becomes an issue again we can revisit.

PR #2111, opened 2022-05-04
Better err handling to customer

SudoBrendan:

Fixes

The current error handling will display API errors to the customer, mistaken as input errors by the customer. For example, if the API call has a timeout, we currently display that to the customer and tell them their subnet was wrong. This restricts those messages to only 404 not founds.


bennerv:
e2e has no influence on these changes so merging.

PR #2116, opened 2022-05-05
NSG controller - reconcile nil NSG

Makdaam:

Fixes ADO-10554108

Fixes the issue when there is no NSG assigned to a subnet. Previously it would error out. RIght now it attaches the correct NSG.

### Test plan for issue:

added unit test to cover the case

n/a - bug

### DIFF
--- a/pkg/operator/controllers/subnets/subnet_nsg.go
+++ b/pkg/operator/controllers/subnets/subnet_nsg.go
@@ -21,7 +21,7 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
 	if err != nil {
 		return err
 	}
-	if subnetObject.SubnetPropertiesFormat == nil || subnetObject.SubnetPropertiesFormat.NetworkSecurityGroup == nil {
+	if subnetObject.SubnetPropertiesFormat == nil {
 		return fmt.Errorf("received nil, expected a value in subnetProperties when trying to Get subnet %s", s.ResourceID)
 	}
 
@@ -30,14 +30,18 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
 		return err
 	}
 
-	if !strings.EqualFold(*subnetObject.NetworkSecurityGroup.ID, correctNSGResourceID) {
-		r.log.Infof("Fixing NSG from %s to %s", *subnetObject.NetworkSecurityGroup.ID, correctNSGResourceID)
-
-		subnetObject.NetworkSecurityGroup = &mgmtnetwork.SecurityGroup{ID: &correctNSGResourceID}
-		err = r.subnets.CreateOrUpdate(ctx, s.ResourceID, subnetObject)
-		if err != nil {
-			return err
-		}
+	// if the NSG is assigned && it's the correct NSG - do nothing
+	if subnetObject.SubnetPropertiesFormat.NetworkSecurityGroup != nil && strings.EqualFold(*subnetObject.NetworkSecurityGroup.ID, correctNSGResourceID) {
+		return nil
 	}
-	return nil
+
+	// else the NSG assignment needs to be corrected
+	oldNSG := "nil"
+	if subnetObject.SubnetPropertiesFormat.NetworkSecurityGroup != nil {
+		oldNSG = *subnetObject.NetworkSecurityGroup.ID
+	}
+	r.log.Infof("Fixing NSG from %s to %s", oldNSG, correctNSGResourceID)
+	subnetObject.NetworkSecurityGroup = &mgmtnetwork.SecurityGroup{ID: &correctNSGResourceID}
+	err = r.subnets.CreateOrUpdate(ctx, s.ResourceID, subnetObject)
+	return err
 }
--- a/pkg/operator/controllers/subnets/subnet_controller_test.go
+++ b/pkg/operator/controllers/subnets/subnet_controller_test.go
@@ -311,6 +311,43 @@ func TestReconcileManager(t *testing.T) {
 				instace.Spec.ArchitectureVersion = int(api.ArchitectureVersionV2)
 			},
 		},
+		{
+			name:                        "Architecture V2 - empty NSG",
+			operatorFlagEnabled:         true,
+			operatorFlagNSG:             true,
+			operatorFlagServiceEndpoint: true,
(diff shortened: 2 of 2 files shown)


pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -30,6 +30,13 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
...
 
+	if subnetObject.SubnetPropertiesFormat.NetworkSecurityGroup == nil {
+		r.log.Infof("Fixing NSG from nil to %s", correctNSGResourceID)
+		subnetObject.NetworkSecurityGroup = &mgmtnetwork.SecurityGroup{ID: &correctNSGResourceID}
+		err = r.subnets.CreateOrUpdate(ctx, s.ResourceID, subnetObject)
+		return err
+	}
+
 	if !strings.EqualFold(*subnetObject.NetworkSecurityGroup.ID, correctNSGResourceID) {
 		r.log.Infof("Fixing NSG from %s to %s", *subnetObject.NetworkSecurityGroup.ID, correctNSGResourceID)
```
gvanderpotte, line 41: nit: why not combining the two into one block ? I don't like code duplication :)

gvanderpotte (approved):
one nit

pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -30,6 +30,13 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
 		return err
 	}
 
+	if subnetObject.SubnetPropertiesFormat.NetworkSecurityGroup == nil {
```
m1kola, line 33: Section below is (almost) the same. Cane we combine these?

pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -30,14 +30,16 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
...
+	if subnetObject.SubnetPropertiesFormat.NetworkSecurityGroup != nil && strings.EqualFold(*subnetObject.NetworkSecurityGroup.ID, correctNSGResourceID) {
+		return nil
+	}
 
-		subnetObject.NetworkSecurityGroup = &mgmtnetwork.SecurityGroup{ID: &correctNSGResourceID}
-		err = r.subnets.CreateOrUpdate(ctx, s.ResourceID, subnetObject)
-		if err != nil {
-			return err
-		}
+	oldNSG := "nil"
```
gvanderpotte, line 37: might be good to comment here, that because of the previous "if"s, the only two possibilities are either a nil previous nsg or not the right nsg, and that both require reconcilation

gvanderpotte (dismissed):
LGTM - one small nit, about a comment

pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -30,14 +30,18 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
...
+	}
 
-		subnetObject.NetworkSecurityGroup = &mgmtnetwork.SecurityGroup{ID: &correctNSGResourceID}
-		err = r.subnets.CreateOrUpdate(ctx, s.ResourceID, subnetObject)
-		if err != nil {
-			return err
-		}
+	// else the NSG assignment needs to be corrected
+	oldNSG := "nil"
+	if subnetObject.SubnetPropertiesFormat.NetworkSecurityGroup != nil {
```
cadenmarchese, line 40: Do we need an `oldNSG` var, or does something like this work?
```suggestion
	if subnetObject.SubnetPropertiesFormat.NetworkSecurityGroup != nil {
		r.log.Infof("Fixing NSG from '%s' to '%s'", subnetObject.NetworkSecurityGroup.ID, correctNSGResourceID)
		subnetObject.NetworkSecurityGroup = &mgmtnetwork.SecurityGroup{ID: &correctNSGResourceID}
		
		err = r.subnets.CreateOrUpdate(ctx, s.ResourceID, subnetObject)
		if err != nil {
			return err
		}
	}
```
What you have now makes the log message more clear, but I think an empty string with single quotes would get the point across.

pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -30,14 +30,18 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
...
+		oldNSG = *subnetObject.NetworkSecurityGroup.ID
 	}
-	return nil
+	r.log.Infof("Fixing NSG from %s to %s", oldNSG, correctNSGResourceID)
+	subnetObject.NetworkSecurityGroup = &mgmtnetwork.SecurityGroup{ID: &correctNSGResourceID}
+	err = r.subnets.CreateOrUpdate(ctx, s.ResourceID, subnetObject)
+	return err
```
cadenmarchese, line 46:
```suggestion
	return nil
```

cadenmarchese (commented):
Thank you for adding the functionality! I just have one suggestion, not required.

pkg/operator/controllers/subnets/subnet_nsg.go
SudoBrendan: alternatively use `%q` instead of `'%s'`.

SudoBrendan (approved):
LGTM

pkg/operator/controllers/subnets/subnet_nsg.go
SudoBrendan: you could alternatively use `%q` instead of `'%s'`.

ross-bryan (commented):
haven't tested myself, but lgtm. would you be interested in modifying the NSG e2e test to include the use case of an empty NSG?

s-amann (approved):

ross-bryan (approved):
lgtm, thanks for this fix Leszek!

pkg/operator/controllers/subnets/subnet_nsg.go
cadenmarchese: Thanks, the suggestion to return nil is not clear on its own. I was providing it in context with this suggestion: https://github.com/Azure/ARO-RP/pull/2116#discussion_r869570263

PR #2118, opened 2022-05-06
create lint-go script and call it from Makefile

AldoFusterTurpin:

Fixes vendored linter not working locally

Previous to this PR, when executing `make test-go` or `make lint-go` makefile was invoking vendored golangci-lint and it returned some lint errors as it's an old linter version, thus breaking the local development workflow.  

With the proposed changes, when running any of the above commands:
- if golangci-lint binary is present in dev machine, it will run the local golangci-lint binary
- if golangci-lint binary is NOT present in dev machine, it will just show a warning about skipping linting and the link to download golangci-lint

### Test plan for issue:
Test it locally when golangci-lint exist and when does not exist.

No, simply fix dev workflow.


hack/lint-go.sh
```diff
@@ -0,0 +1,6 @@
+if ! command -v golangci-lint &> /dev/null
+then
+    echo "WARNING golangci-lint could not be found, golang lint skipped.To install it: https://golangci-lint.run/usage/install/#local-installation"
```
facchettos, line 3: I would probably fail the script if we can't find it 
```suggestion
    echo "WARNING golangci-lint could not be found, golang lint skipped.To install it: https://golangci-lint.run/usage/install/#local-installation"
    exit 1
```

s-amann: +1 to failing the script as it more strongly encourages installing the linter

s-amann (dismissed):

hack/lint-go.sh
AldoFusterTurpin: Thank you both for the feedback 🙂
 Done!

s-amann (approved):
LGTM

cadenmarchese (commented):
Thanks, this has been bugging me as well lately.

I think you could get away with having the script inside of the Makefile itself, I am not sure a separate file is needed.

If you'd rather keep it in a separate file, maybe add a [shebang and code comment](https://github.com/Azure/ARO-RP/blob/master/hack/ssh-agent.sh#L1-L5) to align it with our other hack scripts.

Thanks!

AldoFusterTurpin:
@cadenmarchese Great idea. Thanks! 
Done.

petrkotas (approved):
Thanks, good job @AldoFusterTurpin

s-amann (dismissed):

darthhexx (dismissed):
LGTM. Small nit about a \n.

hack/lint-go.sh
```diff
@@ -0,0 +1,18 @@
...
+# In case the binary that is trying to execute does not exist,
+# the script will print a message with instructions to download the binary and will exit with a non-zero exit code.
+
+if ! command -v golangci-lint &> /dev/null
+then
+    echo "ERROR: golangci-lint could not be found, Go linting aborted. To install it visit https://golangci-lint.run/usage/install/#local-installation"
+    exit 1
+else
+    golangci-lint run
+fi
```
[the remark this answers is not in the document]
AldoFusterTurpin: Thank you. Done

petrkotas (approved):
Thanks @AldoFusterTurpin for this improvement!

PR #2120, opened 2022-05-06
improve Python unit testing and refactor _validators.py

AldoFusterTurpin:

Improvement of Python code testing and refactor of _validators.py (application of Guard clauses, give some names to magic numbers, etc).

Before this PR there was no execution of any Python unit tests (even though we have a prompt saying "Python Unit tests", we were just executing the Python linter). 

Now we have a lot more Python code covered by unit tests.

This PR simplifies the code in python/az/aro/azext_aro/_validators.py applying Guard clauses, avoiding having nested ifs and complex boolean conditions in order to improve the understanding of the code at first sight.

This PR also splits the Python test folder in:
- unit tests (no external dependencies) and 
- integration tests (depends on Microsoft testing libraries and always fails -> this is why we did not execute Python tests before this PR). Now that we split tests in unit vs integrations we can safely execute unit tests without depending on MSFT things and without the tests failing all the time.

This PR adds Unit Tests for all the functions in _validators.py (it uses pytest.mark.parametrize for a cleaner data-oriented tests).

Lastly, it includes directives in the Makefile to execute python UNIT TESTS (no integration). -> this could be done directly in the Makefile but I decided to write a simple script because 

1) the Python unit tests are invoked in two recipes of the Makefile: 
    - one to just run the unit test (unit-test-python) 
    - and another one to also run the linter (test-python)

(there are two different recipes in the Makefile because when programming, you just one to make your tests go green without worrying about the linter, and just before committing the code, execute the linter to make sure there are no linting errors. To sum up: it is more convenient to execute N times 
```
make unit-test-python
```
and just once 
```
make test-python
``` 
at the end, all that for a faster feedback loop).

2) IMHO it's better to have a big independent script and a clean Makefile rather than a bloated Makefile. If in the future we change how to execute the unit tests, we will touch a single file with that responsibility, the script, and NO the whole Makefile.

N/A. One could argue that the new Unit tests also serve as executable documentation.


Makefile
```diff
@@ -175,7 +175,12 @@ test-python: pyenv az
 	. pyenv/bin/activate && \
 		azdev linter && \
 		azdev style && \
-		hack/format-yaml/format-yaml.py .pipelines
+		hack/format-yaml/format-yaml.py .pipelines && \
+		cd python/az/aro && python -m unittest discover azext_aro/tests/latest/unit
+
+unit-test-python: pyenv az
+	. pyenv/bin/activate && \
```
s-amann, line 182: does this need to deactivate when done?

s-amann (approved):
nice work! Just one question, otherwise LGTM

Makefile
AldoFusterTurpin: Thanks @s-amann  for your review. 
Still a draft as I need to push some last commits to consider it "done".
Answering your question: it does not need to be deactivated as every Makefile recipe is executed in a [sub-shell](https://www.gnu.org/software/make/manual/html_node/Execution.html). This is why we don't need to worry when doing "cd" in the test-python or unit-test-python of the Makefile as it will not affect the current working directory.

s-amann: Thanks for the answer @AldoFusterTurpin, I didn't notice it was in draft until after I hit submit

petrkotas (dismissed):
Thanks @AldoFusterTurpin for the PR. Great work.

s-amann (changes requested):
overall LGTM, just one typo in a test value

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -0,0 +1,419 @@
...
+
+@pytest.mark.parametrize("test_description, namespace, expected_namespace_vnet_resource_group_name", test_validate_vnet_resource_group_name_data, ids=[i[0] for i in test_validate_vnet_resource_group_name_data])
+def test_validate_vnet_resource_group_name(test_description, namespace, expected_namespace_vnet_resource_group_name):
+    validate_vnet_resource_group_name(namespace)
+    assert(namespace.vnet_resource_group_name == expected_namespace_vnet_resource_group_name)
+
+
+test_validate_worker_count_data = [
+    ("should not raise any Exception because worker count of namespace is None", Mock(worker_count=None), None),
+    ("should not raise any Exception because worker count of namespace is 3", Mock(worker_count=None), None),
```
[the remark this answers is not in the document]
AldoFusterTurpin: Thanks! Fixed now.

s-amann (dismissed):

s-amann (dismissed):

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -0,0 +1,419 @@
...
+from azure.cli.core.azclierror import (
+    InvalidArgumentValueError, RequiredArgumentMissingError, RequiredArgumentMissingError, CLIInternalError
+)
+from azure.core.exceptions import ResourceNotFoundError
+import pytest
+
+
+test_validate_cidr_data = [
+    ("should not raise exception when valid IPv4 address", Mock(key='192.168.0.0/28'), "key", None),
+    ("should raise InvalidArgumentValueError when non valid IPv4 address due to beeing a simple string", Mock(key='this is an invalid network'), "key", InvalidArgumentValueError),
```
SudoBrendan, line 20: nit: spelling on "being"

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -0,0 +1,419 @@
...
+    ("should raise InvalidArgumentValueError when non valid IPv4 address due to invalid network ID", Mock(key='192.168.0.0.0.0/28'), "key", InvalidArgumentValueError),
+    ("should raise InvalidArgumentValueError when non valid IPv4 address due to invalid hostID", Mock(key='192.168.0.0.0.0/2888'), "key", InvalidArgumentValueError),
```
SudoBrendan, line 22: I'm having trouble differentiating between these tests. Both contain invalid IP address CIDRs due to too many IP digits `192.168.0.0.0.0`, but the second also contains an invalid range of `2888`

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -0,0 +1,419 @@
...
+
+
+test_validate_disk_encryption_set_data = [
+    ("should not raise any exception when namespace.disk_encryption_set is None", None, Mock(disk_encryption_set=None), None, None, None, None),
+    ("should raise InvalidArgumentValueError exception when namespace.disk_encryption_set is not None and is_valid_resource_id(namespace.disk_encryption_set) returns False", None, Mock(disk_encryption_set="something different than None"), False, None, InvalidArgumentValueError, None),
+    ("should not raise any exception when compute_client.disk_encryption_sets.get() not raises CludError exception", Mock(cli_ctx=None), Mock(disk_encryption_set="something different than None"), True, Mock(), None, {"resource_group": None, "name": None})
+]
+
+
+@pytest.mark.parametrize("test_description, cmd_mock, namespace, is_valid_resource_id_return_value, compute_client_mock, expected_exception, parse_resource_id_mock_return_value", test_validate_disk_encryption_set_data, ids=[i[0] for i in test_validate_disk_encryption_set_data])
```
SudoBrendan, line 99: By this point we have 7 ordered parameters to our test suite and I'm having difficulty parsing out or remembering what each ordered item is. This makes me think this "unit" is actually doing too much. Can we simplify the code so it is testable?

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -0,0 +1,419 @@
...
+        with pytest.raises(expected_exception):
+            validate_disk_encryption_set(cmd_mock, namespace)
+
+
+test_validate_domain_data = [
+    ("should not raise any exception when namespace.domain is None", Mock(domain=None), None),
+    ("should not raise any exception when namespace.domain has '-'", Mock(domain="my-domain.com"), None),
+    ("should not raise any exception when namespace.domain is google.com.au", Mock(domain="google.com.au"), None),
+    ("should not raise any exception when namespace.domain is some.more.than.expected", Mock(domain="some.more.than.expected"), None),
+    ("should not raise any exception when namespace.domain is azure.microsoft.com", Mock(domain="azure.microsoft.com"), None),
```
SudoBrendan, line 123: These seem like the same (happy-path) case, can we remove the duplicates, or am I missing something?

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -0,0 +1,419 @@
...
+        "should raise RequiredArgumentMissingError exception when subnet is not a valid resource id and namespace.vnet is False",
+        Mock(key='192.168.0.0/28', vnet=False),
+        'key',
+        False,
+        None,
+        None,
+        None,
+        None,
+        RequiredArgumentMissingError
+    ),
```
SudoBrendan, line 186: aha! So we can format like this - can we spread it to every test case above and below? I find this much easier to read - I know where each test begins and ends.

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -0,0 +1,419 @@
...
+        "subscription",
+        Mock(**{"subnets.get.return_value": None}),
+        Mock(cli_ctx=None),
+        None
+    )
+]
+
+
+@ pytest.mark.parametrize(
+    "test_description, namespace, key, is_valid_resource_id_mock_return_value, parse_resource_id_mock_return_value, get_subscription_id_mock_return_value, get_mgmt_service_client_mock_return_value, cmd, expected_exception",
```
SudoBrendan, line 290: 9 parameters - we are likely doing too much.

SudoBrendan (dismissed):
This is really great work! Tests are very much needed, thank you!

I leave some comments about large parameter sets, but those may be outside the scope of this PR - feel free to ignore those and keep as-is... but we should likeley create cards for those items to fix in the future. For now, test are better than no tests!

I would appreciate consistent formatting on the test parameters, some suites are very difficult  to read right now because of so many ordered parameters being squashed onto a single line.

Also, out of curiosity, do the tests still pass when when you revert the changes to _validators.py?

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
AldoFusterTurpin: Thanks, well spotted. Test case on line 22 should fail just due to an invalid range of 288, fixed.

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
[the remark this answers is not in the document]
AldoFusterTurpin: Fixed, thanks.

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
AldoFusterTurpin: Done (removed test case from line 121).

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
AldoFusterTurpin: Done it.

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
AldoFusterTurpin: Yes, I agree with that, but the "problem" is that the function under tests calls different MSFT functions and I want this test to be a unit test and don't rely on external calls. For me here is the boundary of the "unit" so I decided to mock those external dependencies. 
In a different context, I would probably not mock them (e,g: if the unit under test was invoking jus functions of the domain layer), but here it makes sense (IMHO).

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
AldoFusterTurpin: I would say this should be a different PR. For now, having those 74 test cases is a good starting point. 😃

AldoFusterTurpin:
@SudoBrendan Thank you for your review. 
I have applied the proposed changes and fixed a few typos you mentioned. 

Answering your question: yes, I have run all the tests with the code previous to the refactor. 

A side comment, also: even though I was adding unit tests to existing code, I was ALWAYS making my test go Red before going Green to ensure no false positives.

petrkotas (approved):
I see comments are addressed. So lets get this in after it is green.

petrkotas (approved):
THank you @AldoFusterTurpin

petrkotas:
@SudoBrendan I see @AldoFusterTurpin fixed everything requested. I am dismissing your review so i can merge. Thank you for understanding.

PR #2123, opened 2022-05-10
Add an in-cluster alert when mhc remediates frequently

s-amann:

Addresses: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14313673

Will deploy an alert in cluster to notify when the machine health check is remediating continuously. This will later be used in Geneva Monitoring to trigger and ICM for SRE to look into/make note of.

### Test plan for issue:

Tested in dev cluster, existing unit tests cover the deploy of files under staticresources directory.

Yes, as a part of the ADO story documentation in ADO will be updated to reflect the new alert and how it can be used.


s-amann:
This is my first time adding a dependency to this project. Please take a close look at how go.sum/go.mod were changed.

Steps I took to add the dependency:
`go get github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring@v0.48.1`
`go mod tidy`
`go mod vendor`

pkg/operator/controllers/machinehealthcheck/staticresources/mhcremediationalert.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
```
cadenmarchese, line 2: Question about the PrometheusRule - [this section](https://docs.openshift.com/container-platform/4.10/monitoring/managing-alerts.html#creating-alerting-rules-for-user-defined-projects_managing-alerts) of openshift documentation says that we can't create new alerts in `openshift-*` namespaces. Is there something different about the strategy here that makes it work?

s-amann: I was able to create alerting rules with `oc create -f mhcremediationalert.yaml` just fine in the `openshift-machine-api` namespace and the alert fired without issue. I am open to changing the namespace to something more ARO specific to help indicate to the customer where the alert is from when it shows up in their management console.

s-amann: I was able to create alerting rules with `oc create -f mhcremediationalert.yaml` just fine in the `openshift-machine-api` namespace and the alert fired without issue. I am open to changing the namespace to something more ARO specific to help indicate to the customer where the alert is from when it shows up in their management console.

vendor/sigs.k8s.io/controller-runtime/go.mod
```diff
@@ -6,30 +6,24 @@ require (
...
-	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 	k8s.io/api v0.21.2
 	k8s.io/apiextensions-apiserver v0.21.2
 	k8s.io/apimachinery v0.21.2
 	k8s.io/client-go v0.21.2
 	k8s.io/component-base v0.21.2
 	k8s.io/utils v0.0.0-20210527160623-6fdb442a123b
-)
-
-exclude (
```
ross-bryan, line 28: we cant lose these :/

vendor/github.com/spf13/viper/go.mod
```diff
@@ -3,32 +3,38 @@ module github.com/spf13/viper
...
+	github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2 // indirect
+	go.etcd.io/bbolt v1.3.2 // indirect
+	go.uber.org/atomic v1.4.0 // indirect
+	go.uber.org/multierr v1.1.0 // indirect
+	go.uber.org/zap v1.10.0 // indirect
 	gopkg.in/ini.v1 v1.51.0
 	gopkg.in/yaml.v2 v2.2.4
 )
-
-exclude (
```
ross-bryan, line 27: we cant lose these :/

ross-bryan (dismissed):
we need to keep these exclusions... need to figure out a better way to manage these because right now they get stomped on by go mod vendor

vendor/github.com/docker/go-metrics/go.mod
```diff
@@ -3,12 +3,3 @@ module github.com/docker/go-metrics
 go 1.11
 
 require github.com/prometheus/client_golang v1.1.0
-
-exclude (
```
s-amann, line 7: These exclusions should not be removed. They were removed when running `go mod vendor` I'm not sure how to go about keeping them aside from manually adding them back in and making sure any future PR which vendors also adds them back in :(

mbarnes: Partial solution might be append a `go mod edit` command to [https://github.com/Azure/ARO-RP/blob/master/hack/update-go-module-dependencies.sh](https://github.com/Azure/ARO-RP/blob/master/hack/update-go-module-dependencies.sh).

Doesn't fix `go mod vendor` but would fix `make vendor`.

s-amann: I'll play around with this and see if I can get it to work, thank you for the suggestion. I was having issues with getting `make vendor` to run successfully so I'll likely have to figure that out first. Another idea (longer term) could be to add these exclusions upstream.

s-amann: go mod edit is for our go.mod file in the root directory (which already excludes protobuf versions). I manually added the exclusions back into the respective go.mod files being vendored. I think longer term we need to stop relying on locally vendored modules for our production binary and move to using a GOPROXY which we populate and access locally and from CI.

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -31,6 +32,8 @@ type Reconciler struct {
 }
 
 func NewReconciler(arocli aroclient.Interface, dh dynamichelper.Interface) *Reconciler {
+	// registers types such as PrometheusRule
+	monitoringv1.AddToScheme(scheme.Scheme)
```
ross-bryan, line 36: hmm. I'd have to do some more digging but i'm wondering if this should go into `pkg/util/scheme/scheme.go` wdyt?

s-amann: That sounds like a good place for it, I will look into it and make the changes if appropriate.

pkg/operator/controllers/machinehealthcheck/staticresources/mhcremediationalert.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: mhc-remediation-alert
+  namespace: openshift-machine-api
```
SudoBrendan, line 5: Can we also add the best-practice labeling structure? Setting the Prometheus instance is of particular concern - if customers create their own Prometheus, these may fire on thier instance instead of the cluster instance.

ref: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/alerting.md#prometheusrule-labelling

SudoBrendan (dismissed):
LGTM, aside from Labels on the PrometheusRule.

pkg/operator/controllers/machinehealthcheck/staticresources/mhcremediationalert.yaml
SudoBrendan: FYI:

I just happened to have a new cluster spun up, looks like we'd need to set this to `prometheus: k8s` to accommodate the OpenShift defaults:

```sh
$ oc get Prometheus -A
NAMESPACE              NAME   VERSION   REPLICAS   AGE
openshift-monitoring   k8s    2.29.2    2          16m
```

s-amann: Yes, I will add this and update shortly. Thank you for pointing this out!

ross-bryan (approved):
lgtm

PR #2130, opened 2022-05-19
Migrate from AD to MS Graph

### LINKED ISSUE #2129: Migrating to Microsoft Graph for az cli
`az cli` has to migrate from using AD Graph to Microsoft Graph.  AD is being deprecated and there's a push for the migration to happen quite soon.

Azure/azure-cli#22174

nontw:
Also changed the AADManager so that it only returns values
instead of the data structure.  This hides the implementation
details so that in the future if MSAL changes the internal
representation, any required changes will be contained within
the class (vs. right now custom.py has to be changed accordingly).

Fixes #2129 

Migrating from AAD Graph to MS Graph for az cli.  The change also removes all the references of AAD Graph.  It also includes a minor refactoring - AADManager now encapsulates the internal data structures from Microsoft Graph by only returning values and ids instead of maps or classes from the API.

### Test plan for issue:
Testing by trying to create, update and delete a cluster with the newly changed cli.  They were all successful.  More tests will be performed once the changes are submitted to the upstream cli repo.

No, the aro cli will work the same as before.

### DIFF
--- a/python/az/aro/azext_aro/_aad.py
+++ b/python/az/aro/azext_aro/_aad.py
@@ -1,65 +1,41 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the Apache License 2.0.
 
-import datetime
 import time
-import uuid
 
-from azure.cli.core._profile import Profile
-from azure.cli.core.commands.client_factory import configure_common_settings
-from azure.cli.core.azclierror import BadRequestError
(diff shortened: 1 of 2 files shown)


SudoBrendan (approved):
LGTM

petrkotas (approved):
Thank you @nontw this simplifies a whole lot of the code.

PR #2132, opened 2022-05-23
added yaml lint

facchettos:

Fixes


.github/workflows/yamllint.yml
```diff
@@ -0,0 +1,42 @@
...
+# More info at https://megalinter.github.io
+name: mega-linter-yaml
+
+on:
+  # Trigger mega-linter at every push. Action will also be visible from Pull Requests to main
+  push: # Comment this line to trigger action only on pull-requests (not recommended if you don't pay for GH Actions)
+  pull_request:
+
+env: # Comment env block if you do not want to apply fixes
+  APPLY_FIXES: none # When active, APPLY_FIXES must also be defined as environment variable (in github/workflows/mega-linter.yml or other CI tool)
```
SudoBrendan, line 12: This appears to be duplicated in `.mega-linter.yml`

pkg/portal/ssh/proxy.go
```diff
@@ -281,8 +281,8 @@ func (s *ssh) newChannel(ctx context.Context, accessLog *logrus.Entry, nc crypto
 	defer recover.Panic(s.log)
 
 	ch2, rs2, err := downstreamConn.OpenChannel(nc.ChannelType(), nc.ExtraData())
-	if err, ok := err.(*cryptossh.OpenChannelError); ok {
-		return nc.Reject(err.Reason, err.Message)
+	if errAsOpenChannel, ok := err.(*cryptossh.OpenChannelError); ok {
+		return nc.Reject(errAsOpenChannel.Reason, errAsOpenChannel.Message)
```
SudoBrendan, line 285: This seems like a weird addition, was it meant for this PR? I don't hate it, just asking.

SudoBrendan (approved):
Overall, LGTM! A few nits.

pkg/portal/ssh/proxy.go
```diff
@@ -281,8 +281,8 @@ func (s *ssh) newChannel(ctx context.Context, accessLog *logrus.Entry, nc crypto
...
 	} else if err != nil {
```
ross-bryan, line 286: this breaks the control flow in the following else if

ross-bryan (changes requested):
lint changes lgtm, I think some other work snuck into this PR that I'd prefer remove from PR, or at minimum fix the flow

pkg/portal/ssh/proxy.go
facchettos: it does not. if anything, what was there was confusing at best. The `err` in the `else if` is the one from before the whole `if` statement. I originally had that in the PR because I also had a golangci-lint update and the new version was complaining about it (this check specifically https://staticcheck.io/docs/checks#SA9008) but I'll remove it

pkg/portal/ssh/proxy.go
facchettos: yeah it is a leftover. I also had a golangci-lint update and it was complaining about it then.

pkg/portal/ssh/proxy.go
ross-bryan: ope totally see what you mean now. this definitely is a bug fix then as the `else if` was using the closer scoped `err` as type `cryptossh.OpenChannelError` when the `else if` should've been referencing the `err` of type `error`.

ross-bryan: even typing that I confused myself ugh

ross-bryan (approved):
lgtm

pkg/portal/ssh/proxy.go
facchettos: not only that, but also what we would get is `nil` because when we do a type assertion and `ok == false` it means that `err` is the zero value of `*openChannelError` (`nil`), so the `else if` is never true 

I would say that it's generally a bad idea to have variable overrides like this one. It's very rarely useful and quite often very confusing

PR #2135, opened 2022-05-24
Increase the lint timeout

hawkowl:

Ups the lint timeout to 10 minutes, as I have noticed that on some PRs a rebuild can make the cache invalidate enough that it needs to redo it from scratch. This takes about 6 minutes on my local PC, and some ~7 mins on CI. 10 minutes should be enough, and this passed on a PR I tested it on that was failing.

### Test plan for issue:
N/A

N/A

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,5 @@
 run:
-  timeout: 5m
+  timeout: 10m
   skip-dirs:
     - vendor/portal
 


darthhexx (approved):
LGTM. Not sure e2e will go green on the current run, but not _actually_ needed for this change IMO.

facchettos:
I have noticed that too. I think there is something fishy going on with the cache because it used to take 40-ish seconds to complete. I think that the cache has been somehow corrupted

s-amann (approved):

PR #1761, opened 2021-09-22
change shared dev vnet to /9

rolandmkunkel:
The shared environment preparation code used a /23 vnet for all clusters which was too small. This PR changes the shared environment to use a /9 vnet. This was already applied to v4-westeurope and doesn't to be applied to further shared environments to my knowledge. Note: v4-westeurope had a manually configured /9 vnet which was overwritten by running deploy_env_dev without the changes from this PR

PR #2138, opened 2022-05-25
Switch CI to FIPS enabled Go 1.17

darthhexx:
Adds CI vmss deploy changes to support Golang 1.17 in the CI agent pool.

Once the 4.10 PR https://github.com/Azure/ARO-RP/pull/2070 is merged we will need Go 1.17 support in the ARO-CI agent pool for all PRs.

The documentation to deploy the updated ARO-CI pool is here: https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/189005/CI-VM-Setup-and-Permissions-Azure-Pipelines

**Notes:**
* This is tagged as blocked because the 4.10 PR needs to be merged first. Once it is merged this can be used to deploy a new ARO-CI Agent Pool.
* The changes in https://github.com/Azure/ARO-RP/pull/2138/commits/22c11282224fc8475355f0844a721aa3884c5421 will need to be made locally when running `deploy_env_dev_ci` to deploy the new agent CI pool, because the keyvault name changes made for the new dev subscription and https://github.com/Azure/ARO-RP/pull/1761 clash with the existing setup in the CI resource group.

### Test plan for issue:
This branch has been used to deploy the existing CI 1.17 vmss that is running jobs tagged as requiring go 1.17.

No.


pkg/deploy/generator/resources_dev.go
```diff
@@ -328,7 +328,24 @@ enabled=yes
 gpgcheck=yes
 EOF
 
-yum -y install azure-cli podman podman-docker jq gcc gpgme-devel libassuan-devel git make tmpwatch python3-devel go-toolset-1.16.12-1.module+el8.5.0+13637+960c7771
+yum -y install azure-cli podman podman-docker jq gcc gpgme-devel libassuan-devel git make tmpwatch python3-devel htop openssl-devel
+
+# go-toolset-1.17.X is only available for RHEL 8.6 so let's use the Google binaries to build the Microsoft Go BoringSSL code for the time being
```
ross-bryan, line 333: should we switch our CI VMSS to RHEL 8.6 to continue to use the RHEL Go-toolset?

hawkowl: I see that RHEL 8.6 is now deployed in Azure:

```
  {
    "offer": "RHEL",
    "publisher": "RedHat",
    "sku": "8-LVM",
    "urn": "RedHat:RHEL:8-LVM:8.6.2022051813",
    "version": "8.6.2022051813"
  },
```

So if we change our version to `8.6.2022051813` in the Image selection above, we should get RHEL 8.6's go. I suggest we do use this instead of "latest" as then that means we won't accidentally get 8.7 or whatever.

darthhexx: Now that 8.6 has been released we can indeed.

Although, having a dependancy between the upstreams and us leaves us open to not being able to react quickly to version bumps and CVE's, so there is a case to be made for building directly from MSFTs FIPS releases IMO.

deploy/env-development.json
```diff
@@ -348,6 +348,7 @@
                         },
                         "osDisk": {
                             "createOption": "FromImage",
+                            "diskSizeGB": 128,
```
ross-bryan, line 351: i'd consider going as large as 200-300.

ross-bryan (dismissed):
lgtm

bennerv (approved):
lgtm

PR #2141, opened 2022-05-26
Mirror App-SRE images (MUO, Hive) to ACR

hawkowl:

Fixes https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14536813

Mirrors FIPS-compliant images from the MUO and Hive Quay repos to ACR.

### Test plan for issue:

Tested using the instructions for mirroring images in INT. Seems to work fine.

N/A

### DIFF
--- a/cmd/aro/mirror.go
+++ b/cmd/aro/mirror.go
@@ -155,6 +155,21 @@ func mirror(ctx context.Context, log *logrus.Entry) error {
 		}
 	}
 
+	for _, ref := range []string{
+		// Managed Upgrade Operator
+		"quay.io/app-sre/managed-upgrade-operator:v0.1.831-986b967",
+
+		// Hive
+		"quay.io/app-sre/hive:2383a88",
+	} {
+		log.Printf("mirroring %s -> %s", ref, pkgmirror.Dest(dstAcr+acrDomainSuffix, ref))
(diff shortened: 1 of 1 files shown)


darthhexx (dismissed):
lgtm

s-amann (dismissed):

cmd/aro/mirror.go
```diff
@@ -155,6 +155,17 @@ func mirror(ctx context.Context, log *logrus.Entry) error {
 		}
 	}
 
+	for _, ref := range []string{
+		"quay.io/app-sre/managed-upgrade-operator:v0.1.827-8d67cc6",
+	} {
+		log.Printf("mirroring %s -> %s", ref, pkgmirror.Dest(dstAcr+acrDomainSuffix, ref))
```
cadenmarchese, line 161: Since we're reusing lines 161-165 from the `registry.redhat.io` mirror above with the only difference being `srcAuth`, I think we can make a separate function for the `pkgmirror.Copy` stuff. What do you think?

cadenmarchese (dismissed):

cmd/aro/mirror.go
petrkotas: With this particular code I would be fine with copy & paste. Having function will require jumping up and down. Moreover this code will potentially go away and be replaced with `az acr import`. 

Would that work for you Caden?

darthhexx (approved):
LGTM

PR #2143, opened 2022-05-27
Enable MUO M2 by default

hawkowl:

Fixes https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14536879

Enables MUO M2 by default.

### Test plan for issue:

Unit tests

N/A

### DIFF
--- a/pkg/operator/controllers/muo/muo_controller.go
+++ b/pkg/operator/controllers/muo/muo_controller.go
@@ -34,7 +34,7 @@ const (
 	controllerEnabled                = "rh.srep.muo.enabled"
 	controllerManaged                = "rh.srep.muo.managed"
 	controllerPullSpec               = "rh.srep.muo.deploy.pullspec"
-	controllerAllowOCM               = "rh.srep.muo.deploy.allowOCM"
+	controllerForceLocalOnly         = "rh.srep.muo.deploy.forceLocalOnly"
 	controllerOcmBaseURL             = "rh.srep.muo.deploy.ocmBaseUrl"
 	controllerOcmBaseURLDefaultValue = "https://api.openshift.com"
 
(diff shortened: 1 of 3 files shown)


pkg/operator/controllers/muo/muo_controller.go
```diff
@@ -34,7 +34,7 @@ const (
 	controllerEnabled                = "rh.srep.muo.enabled"
 	controllerManaged                = "rh.srep.muo.managed"
 	controllerPullSpec               = "rh.srep.muo.deploy.pullspec"
-	controllerAllowOCM               = "rh.srep.muo.deploy.allowOCM"
+	controllerDisableOCM             = "rh.srep.muo.deploy.disableOCM"
```
s-amann, line 37: Q: for clarity, this flag should not be on clusters yet so changing this won't leave an unused value laying around, correct?

FWIW I liked how it read before, especially once we use it on line 100 where we were avoiding a double negative and now have to read "not disabled" vs "allowed"

s-amann (approved):
LGTM, although not sure on why we prefer "not disabled" vs "allowed"

s-amann (changes requested):
I saw on some clusters we already have: `rh.srep.muo.deploy.allowOCM` deployed. IMO we should keep the flag name the same and reduce the use of double negative variables.

pkg/operator/controllers/muo/muo_controller.go
hawkowl: This flag is for if we need to turn it off, since we still need some control (for local updates), but it is a snowflake case and not the default.

darthhexx: All of our feature flags use lowercase names and `.enabled = true` and we should definitely be consistent for all of them. 

I would suggest using  `rh.srep.muo.deploy.ocm.enabled` or more simply `rh.srep.muo.ocm.enabled`, which could also be extended to the OCM BaseUrl naming; `rh.srep.muo.ocm.baseurl`.

darthhexx (changes requested):
LGTM, but we need to be consistent with the feature flag naming convention.

pkg/operator/controllers/muo/muo_controller.go
hawkowl: This runs into the issue where a missing flag means false, but you want the feature to be on by default. I'll tweak the naming a bit.

darthhexx (approved):
lgtm

PR #2145, opened 2022-05-31
[WIP] Register cluster with hive at install time

UlrichSchlueter:
https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14319615/

Add feature to register cluster with hive at install time

- if the HIVEKUBECONFIGPATH environment variable is set and points to a valid hive cluster kubeapi entrypoint the installation flow will collect all required data about a cluster to register with hive , both from the installer manager (in memory) as well as from the new cluster (via its kubeapi). If that was successful the cluster will be registered with hive and the registration will be verified by checking the Unreachable condition on the Clusterdeployment resource located in the cluster namespace in the hive cluster.
  - if not the behavior should remain as before

A command line version to feed the data into hive is available in the hack/hive folder. That command expects the HIVEKUBECONFIGPATH env to be set and takes a json file with all the required data

Open Items:
- Unit tests, unit tests, unit tests
- Check if the data is picked up correctly
- Check if the installation process is marked successful in all cases even if a step of the registration flow fails
- Documentation for the hack command
- Decide if any related data should be persisted in cosmos DB (for example the namespace name in hive, the registration status etc).
- go.mod changes to be verified

### Test plan for issue:

tbd

yes, tbd


pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,127 @@
...
+		ClientID:       m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.ClientID,
+		ClientSecret:   string(m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.ClientSecret),
+	}
+
+	spBytes, err := json.Marshal(servicePrincipal)
+	if err != nil {
+		return err
+	}
+
+	// TODO: What if a customer adds/modifies a pullsecret after the install ?
```
m1kola, line 40: Note: for adoption of existing clusters we need to grab it from the cluster as customer can update pull secret.

We also need to firgure out whether we need to keep pull secret in sync with what we have in hive. This is probably out of scope for this PR, but we need to think about it.

pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,127 @@
...
+	// TODO: All parameter locations to be verified
+	m.hiveWorkloadCluster = &hive.WorkloadCluster{
```
m1kola, line 42: Are we setting it on manager because we need to share the state between actions? If so, maybe we should follow existing pattern:

https://github.com/Azure/ARO-RP/blob/51fd449e7e6054938388f29f64f412944d31ca15/pkg/cluster/install.go#L128-L131

https://github.com/Azure/ARO-RP/blob/51fd449e7e6054938388f29f64f412944d31ca15/pkg/cluster/install.go#L139-L143

https://github.com/Azure/ARO-RP/blob/51fd449e7e6054938388f29f64f412944d31ca15/pkg/cluster/install.go#L147-L149

pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return err
+	}
+	clusterid := cv.Spec.ClusterID
+
+	servicePrincipal := azure.Credentials{
+		TenantID:       m.env.TenantID(),
+		SubscriptionID: m.env.SubscriptionID(),
+		ClientID:       m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.ClientID,
+		ClientSecret:   string(m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.ClientSecret),
```
m1kola, line 32: Note: for adoption of existing clusters we need to grab it from the cluster as customer can update/rotate service principal.

We also need to firgure out whether we need to keep cluster SP in sync with what we have in hive. This is probably out of scope for this PR, but we need to think about it.

pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	// TODO: What if a customer adds/modifies a pullsecret after the install ?
+	// TODO: All parameter locations to be verified
+	m.hiveWorkloadCluster = &hive.WorkloadCluster{
+		Subscription:                m.subscriptionDoc.ID,
+		Name:                        m.doc.OpenShiftCluster.Name,
+		Location:                    m.doc.OpenShiftCluster.Location,
+		BaseDomain:                  m.env.Domain(),
+		BaseDomainResourceGroupName: m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID,
```
m1kola, line 47: `*.aroapp.io` is going to be a base domain. And in prod we do not put child DNS zones into cluster resource groups. [This](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14008788#16266901) might be useful.

pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,127 @@
...
+
+	m.log.Print("collecting registration data from new cluster")
+
+	cv, err := m.configcli.ConfigV1().ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	clusterid := cv.Spec.ClusterID
+
+	servicePrincipal := azure.Credentials{
```
m1kola, line 28: If possible, we need to avoid having a dependency on the installer. @hawkowl is trying to decouple the instaler from our codebase at the moment. See [this relevant doc](https://docs.google.com/document/d/1iqfAWwA71VS8xVGbeianlr8SLD9YcJgw7LysZB1gLm4/edit).

pkg/hiveregistrator/registrator.go
```diff
@@ -0,0 +1,164 @@
...
+	WorkloadCluster *WorkloadCluster
+}
+
+func NewRegistrator(workloadcluster *WorkloadCluster) Registrator {
+	return &hiveRegistrator{
+		WorkloadCluster: workloadcluster,
+	}
+}
+
+func (hr *hiveRegistrator) InitializeFromFile(ctx context.Context, hiveKubeConfigPath string) error {
```
m1kola, line 49: Sounds like this should be in `New`

pkg/hiveregistrator/registrator.go
```diff
@@ -0,0 +1,164 @@
...
+	ServicePrincipal            string `json:"serviceprincipal,omitempty"`
+}
+
+type hiveRegistrator struct {
+	Hiveclientset   *hiveclient.Clientset
+	Kubernetescli   *kubernetes.Clientset
+	WorkloadCluster *WorkloadCluster
+}
+
+func NewRegistrator(workloadcluster *WorkloadCluster) Registrator {
```
m1kola, line 43: Nit: maybe just `New`? So from outside it is going to be `hiveregistrator.New(...)`

pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,127 @@
...
+	m.log.Print("registering with hive")
+
+	isRegistered, err := m.hiveRegistrator.IsRegistered(ctx)
+	if err != nil {
+		m.log.Print(err)
+		return nil
+	}
+
+	if isRegistered {
+		m.log.Print("cluster is already registered, it seems. Skipping step.")
```
m1kola, line 91: It seems. But we are not 100% sure 🤣

pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (m *manager) verifyRegistration(ctx context.Context) (bool, error) {
+	// Never fail
+	if m.hiveRegistrator == nil {
+		m.log.Print("no  hive registrator, skipping step")
+		return true, nil
+	}
+
+	m.log.Print("Verify cluster registration")
+	time.Sleep(10 * time.Second) // give hive some time to log in to the new cluster
```
m1kola, line 112:
```suggestion
```

Let's remove sleep. This way we will try immidiately (and will probably fail which is ok). Next attempt will be made after waiting for an interval. See relevant code [here](https://github.com/Azure/ARO-RP/blob/51fd449e7e6054938388f29f64f412944d31ca15/pkg/util/steps/condition.go#L55-L63).

pkg/hiveregistrator/util.go
```diff
@@ -0,0 +1,24 @@
...
+	"strings"
+)
+
+func ClusterNamespaceName(w WorkloadCluster) string {
+	// create a namespace name for a given cluster that is unique within a hive cluster (or set if multiple)
+	// region should not be necessary, but who knows
+	// TODO: any length constraints ? 63 chars ? 253 characterss?
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
+
+	return strings.ToLower(fmt.Sprintf("aro-%s-%s-%s", w.Subscription, w.Location, w.Name))
```
m1kola, line 14: At the very minimum we need `subscriptionId`, `resourceGroupName`, and `resourceName`.

In the same subscription in the same region we can have clusters with the same name, but in different resource groups.

pkg/hiveregistrator/util.go
```diff
@@ -0,0 +1,24 @@
...
+
+}
+
+func ClusterDeploymentName(w WorkloadCluster) string {
+	// create a cluster deployment name for a given cluster
+	// TODO: any length constraints ?
+
+	return strings.ToLower(fmt.Sprintf("cd-%s-%s", w.Subscription, w.Name))
```
m1kola, line 22: If we are going to have one cluster per namespace - we can probably just use cluster name? Or even hardcode something.

pkg/hiveregistrator/registrator.go
```diff
@@ -0,0 +1,164 @@
...
+
+		if condition.Type == v1.UnreachableCondition {
+			//checking for false, meaning not unreachable, so is reachable
+			isReachable := condition.Status != corev1.ConditionTrue
+			return isReachable, nil
+		}
+
+	}
+
+	return true, nil
```
m1kola, line 131: If `UnreachableCondition` is not present. What does it mean? Is it reachable?

pkg/hiveregistrator/registrator.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (hr *hiveRegistrator) IsRegistered(ctx context.Context) (bool, error) {
+
+	hiveNS := ClusterNamespaceName(*hr.WorkloadCluster)
+	hiveCD := ClusterDeploymentName(*hr.WorkloadCluster)
+
+	ns, err := hr.Kubernetescli.CoreV1().Namespaces().List(context.Background(), metav1.ListOptions{
```
m1kola, line 139: Why `context.Background()`? Below we also have `context.TODO()`. Any reason for not using a parent context?

pkg/hiveregistrator/registrator.go
```diff
@@ -0,0 +1,164 @@
...
+		//no such namespace
+		return false, nil
+	}
+
+	cds := hr.Hiveclientset.HiveV1().ClusterDeployments(hiveNS)
+	r, err := cds.List(context.TODO(), metav1.ListOptions{
+		FieldSelector: fmt.Sprintf("metadata.name=%s", hiveCD),
+	})
+	if err != nil {
+		fmt.Printf("error getting pods: %v\n", err)
```
m1kola, line 155: Pods?

pkg/hiveregistrator/registrator.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	if len(ns.Items) == 0 {
+		//no such namespace
+		return false, nil
+	}
+
+	cds := hr.Hiveclientset.HiveV1().ClusterDeployments(hiveNS)
+	r, err := cds.List(context.TODO(), metav1.ListOptions{
+		FieldSelector: fmt.Sprintf("metadata.name=%s", hiveCD),
+	})
```
m1kola, line 153: Why do we need to list? Can't we use get here?

pkg/hiveregistrator/resources.go
```diff
@@ -0,0 +1,98 @@
...
+
+func KubeAdminSecret(w WorkloadCluster, secretName string, namespace string) *corev1.Secret {
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: namespace,
+		},
+		Data: map[string][]byte{
+			"kubeconfig":     []byte(w.KubeConfig),
+			"raw-kubeconfig": []byte(w.KubeConfig),
```
m1kola, line 31: Why do we need `raw-kubeconfig`? I don't remember doing anything like this while practicing adoption.

pkg/hiveregistrator/util.go
UlrichSchlueter: Agreed, making both namespace and clusterdeployment name unique is not required

pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,127 @@
...
+		ServicePrincipal:            string(spBytes),
+	}
+
+	return nil
+
+}
+
+func (m *manager) initializeHiveClusterClients(ctx context.Context) error {
+	// don't fail the installation on hive related errors
+	kubeConfigPath := os.Getenv("HIVEKUBECONFIGPATH")
```
m1kola, line 61: We need to get rid of `os.Getenv` call here. Please try to move it as close as possible to the command entrypoint.

pkg/cluster/registerWithHive.go
UlrichSchlueter: Agreed. No action at this time.

pkg/cluster/registerWithHive.go
UlrichSchlueter: removed

pkg/hiveregistrator/util.go
UlrichSchlueter: Concatenating subscriptionId, resourceGroupName, and resourceName easily exceeds the 63 character limit of a namespace name. Either we drop one ,truncate the name or use a hash/UUID as a namespace name

pkg/hiveregistrator/resources.go
UlrichSchlueter: Dropping raw-kubeconfig. Works fine without.

pkg/hiveregistrator/registrator.go
UlrichSchlueter: No good reason anyway. Fixed.

pkg/hiveregistrator/registrator.go
UlrichSchlueter: That approach makes it easier to distinguish between "hive not reachable" and "resources not present"

pkg/cluster/registerWithHive.go
UlrichSchlueter: I'm aware of that, will fix later.. For now it's just more convenient to be able to change the env variable without having to restart the RP

ross-bryan (commented):
a couple of comments/questions/suggestions. overall looks good on the right path to me.

pkg/cluster/registerWithHive.go
UlrichSchlueter: After discussion with the hive team, I will remove the pullsecret completely. The related doc: https://github.com/openshift/hive/blob/master/docs/using-hive.md#example-adoption-clusterdeployment shows a pullsecret in the example cd but not in the text.

pkg/api/openshiftcluster.go
```diff
@@ -150,6 +151,9 @@ type OpenShiftClusterProperties struct {
 	UserAdminKubeconfig SecureBytes `json:"userAdminKubeconfig,omitempty"`
 
 	RegistryProfiles []*RegistryProfile `json:"registryProfiles,omitempty"`
+
+	HiveClusterNamespace      string `json:"hiveClusterNamespace,omitempty"`
```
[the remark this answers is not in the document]
UlrichSchlueter: This is still a bit in flux. I went from one property to three (namespace, clusterdeployment, hive cluster pointer), back to one.  I agree that a HiveProfile type would make complete sense for more than one properties, for one I'm hesitant.

pkg/cluster/install.go
```diff
@@ -197,6 +199,20 @@ func (m *manager) Install(ctx context.Context) error {
 			steps.Action(m.configureIngressCertificate),
 			steps.Condition(m.ingressControllerReady, 30*time.Minute, true),
 			steps.Action(m.configureDefaultStorageClass),
+			steps.Action(func(ctx context.Context) error {
+				workloadCluster, _ = m.collectDataForHive(ctx)
+				return nil //don't fail because of hive
```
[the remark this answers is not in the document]
UlrichSchlueter: next commit will have that.

hack/hive/verify/verify.go
```diff
@@ -0,0 +1,155 @@
...
+
+	clustersByHiveNamespace := make(map[string]*hive.WorkloadCluster)
+
+	hiveclients, err := hive.HiveClientsFromFileEnvVariable()
+	if err != nil {
+		return err
+	}
+
+	for _, doc := range docs.OpenShiftClusterDocuments {
+		fmt.Println(doc.OpenShiftCluster.Properties.HiveClusterNamespace)
```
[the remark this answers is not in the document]
UlrichSchlueter: I'm not sure the two hacks will remain, if yes they need cleanup and documentation/examples.

UlrichSchlueter:
Not sure why this PR was undrafted. Setting it to draft mode again for now.

pkg/hiveregistrator/registrator.go
```diff
@@ -0,0 +1,138 @@
...
+
+	for _, condition := range cd.Status.Conditions {
+		if condition.Type == v1.UnreachableCondition {
+			//checking for false, meaning not unreachable, so is reachable
+			isReachable := condition.Status != corev1.ConditionTrue
+			return isReachable, condition.Message, nil
+		}
+	}
+
+	return true, "", nil
```
m1kola, line 110: Bumping [this question](https://github.com/Azure/ARO-RP/pull/2145#discussion_r885738324):

> If `UnreachableCondition` is not present. What does it mean? Is it reachable?

pkg/cluster/registerWithHive.go
m1kola: Need to check: can we adopt cluster without providing service principals to Hive to avoid sync question?

pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,130 @@
...
+	clusterID := cv.Spec.ClusterID
+
+	namespacename, err := m.hiveClients.UUIDClusterNamespaceName(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: When hive support first party principles we'll need to send both first party and cluster service principles
+	servicePrincipal := azure.Credentials{
+		TenantID:       m.subscriptionDoc.Subscription.Properties.TenantID,
```
m1kola, line 38: Check if we need to do `nil` check here: `Subscription` is a pointer.

pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+func (m *manager) collectDataForHive(ctx context.Context) (*hiveregistrator.WorkloadCluster, error) {
+	m.log.Print("collecting registration data from new cluster")
+
+	cv, err := m.configcli.ConfigV1().ClusterVersions().Get(ctx, "version", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	clusterid := cv.Spec.ClusterID
```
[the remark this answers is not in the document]
m1kola: It is not the same as infa id.

pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,138 @@
...
+	resourcegroupname := stringutils.LastTokenByte(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
+
+	// TODO: What if a customer adds/modifies a pullsecret after the install ?
+	// TODO: All parameter locations to be verified
+	hiveWorkloadCluster := &hiveregistrator.WorkloadCluster{
+		Subscription:                m.subscriptionDoc.ID,
+		Name:                        m.doc.OpenShiftCluster.Name,
+		ResourceGroup:               resourcegroupname,
+		Location:                    m.doc.OpenShiftCluster.Location,
+		BaseDomain:                  m.env.Domain(),
```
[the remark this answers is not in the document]
m1kola: @UlrichSchlueter is trying to do the adoption without DNS related fiels: DNS management stays in RP for now.

pkg/cluster/registerWithHive.go
UlrichSchlueter: Technically speaking the subscription could be nil, but in that case we'd fail much earlier in the installation process and we don't check for nil in the cases either.
Suggest to leave as it.

pkg/hiveregistrator/resources.go
```diff
@@ -0,0 +1,83 @@
...
+			Namespace: namespace,
+		},
+		Data: map[string][]byte{
+			"osServicePrincipal.json": secret,
+		},
+		Type: corev1.SecretTypeOpaque,
+	}
+}
+
+func ClusterDeployment(namespaceName string, w *WorkloadCluster) *hivev1.ClusterDeployment {
```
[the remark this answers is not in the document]
UlrichSchlueter: changed to parameter list instead of workloadCluster

pkg/cluster/registerWithHive.go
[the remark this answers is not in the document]
UlrichSchlueter: DNS related entries are not required if/as long as:
- we use ManagedDNS: false
- we create/delete cluster on the RP side and adopt cluster in hive.

Leaving the BaseDomain and BaseDomainResourceGroup empty as a result

pkg/cluster/registerWithHive.go
bennerv: > Note: for adoption of existing clusters we need to grab it from the cluster as customer can update/rotate service principal.

I believe that the credentials we store in our database should be valid.  A customer should only rotate their service principal using `az aro update --refresh-credentials` which stores their service principal client id + client secret in our database.  I believe we don't have to grab them from the cluster.  

Additionally, our `m.validateResources` step validates that the cluster service principal (client id and client secret) is valid and has the correct permissions, which happens on `Update()` as well as `Install()`.

bennerv: > Note: for adoption of existing clusters we need to grab it from the cluster as customer can update/rotate service principal.

@m1kola
I believe that the credentials we store in our database should be valid.  A customer should only rotate their service principal using `az aro update --refresh-credentials` which stores their service principal client id + client secret in our database.  I believe we don't have to grab them from the cluster.  

Additionally, our `m.validateResources` step validates that the cluster service principal (client id and client secret) is valid and has the correct permissions, which happens on `Update()` as well as `Install()`.

pkg/cluster/install.go
```diff
@@ -183,6 +188,32 @@ func (m *manager) Install(ctx context.Context) error {
...
+				}
+				return nil //don't fail because of hive
+			}),
+			steps.Condition(func(ctx context.Context) (bool, error) {
+				valid, err := m.verifyRegistration(ctx)
+				if err != nil {
+					m.hiveRegistrator = nil
+					m.log.Info(err)
+				}
+				return valid, nil //don't fail because of hive
```
bennerv, line 214: You can return `valid, err` here as the `steps.Condition` function has a `fail` parameter that states it will not fail to progress the installation if a timeout or error occurs.

pkg/api/admin/openshiftcluster.go
```diff
@@ -295,3 +296,7 @@ type SystemData struct {
 	LastModifiedByType CreatedByType `json:"lastModifiedByType,omitempty"`
 	LastModifiedAt     *time.Time    `json:"lastModifiedAt,omitempty"`
 }
+
+type HiveProfile struct {
+	NamespaceName string `json:"namespaceName,omitempty"`
```
bennerv, line 301: nit - can we just call this `namespace` as that's how it's typically referenced in k8s code and whatnot.  

```suggestion
	Namespace string `json:"namespace,omitempty"`

File: pkg/cluster/registerWithHive.go [hunk 28]
```diff
@@ -0,0 +1,110 @@
...
+
+	if !isConnected {
+		m.log.Printf("hive is not able to connect to cluster %s", reason)
+		return true, nil
+	}
+
+	m.log.Info("cluster adopted successfully")
+
+	return true, nil
+}
```
bennerv, line 110: I think this function in it's context needs some tweaking due to it being called in steps.Condition with a timeout of 2 minutes.

The changes below will do the following:
1. return immediately if hive doesn't have a registerator
2. if the hive registrator returns an error, it will log the error, but continue the installation
3. if the hive registrator does not return an error,  but it's not connected it will retry the function
4. else success

```suggestion
func (m *manager) verifyRegistration(ctx context.Context) (bool, error) {
	// Never fail
	if m.hiveRegistrator == nil {
		m.log.Info("no hive registrator, skipping cluster verification")
		return true, nil
	}

	m.log.Info("verifing cluster registration in hive")
	isConnected, reason, err := m.hiveRegistrator.IsConnected(ctx, m.doc.OpenShiftCluster.Properties.HiveProfile.NamespaceName)
	if err != nil {
		return false, err
	}

	if !isConnected {
		m.log.Printf("hive is not able to connect to cluster %s", reason)
		return false, nil
	}

	m.log.Info("cluster adopted successfully")
	return true, nil
}
```


steps.Conditions looks like the below, so if an error happens within the condition function, it will return it but not fail if the fail parameter is `false`.  If the function returns false, it will rerun it.  

```
	err := wait.PollImmediateUntil(pollInterval, func() (bool, error) {
		// We use the outer context, not the timeout context, as we do not want
		// to time out the condition function itself, only stop retrying once
		// timeoutCtx's timeout has fired.
		return c.f(ctx)
	}, timeoutCtx.Done())

	if err != nil && !c.fail {
		log.Warnf("step %s failed but has configured 'fail=%t'. Continuing. Error: %s", c, c.fail, err.Error())
		return nil
	}
	return err
```

pkg/hiveregistrator/registrator.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	kubesecret := KubeAdminSecret(namespaceName, []byte(workloadCluster.KubeConfig))
+
+	_, err = hr.kubernetescli.CoreV1().Secrets(namespaceName).Create(ctx, kubesecret, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	spsecret := ServicePrincipalSecret(namespaceName, []byte(workloadCluster.ServicePrincipal))
```
bennerv, line 72: we need to patch the `Update` workflow to update this secret as well.  see `pkg/cluster/install.go -> Update` function

pkg/hiveregistrator/registrator.go
```diff
@@ -0,0 +1,100 @@
...
+	// the order is not stable, but the condition is expected to be present
+	for _, condition := range cd.Status.Conditions {
+		if condition.Type == hivev1.UnreachableCondition {
+			//checking for false, meaning not unreachable, so is reachable
+			isReachable := condition.Status != corev1.ConditionTrue
+			return isReachable, condition.Message, nil
+		}
+	}
+
+	// we should never arrive here (famous last words)
```
bennerv, line 98:
```
	// we should never arrive here (famous last words)
```
:crossed_fingers:

bennerv (changes requested):
first pass

pkg/cluster/registerWithHive.go
m1kola: @bennerv ok, this solves the first bit (grabbing from the cluster), but we need to somehow keep SP stored in AKS alongside `ClusterDeployment` in sync with what we have in our DB and in the customer cluster.

pkg/hiveregistrator/registrator.go
m1kola: I created a work item for this: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14895480

Let's do it separately from this PR.

Good catch @bennerv 👍

pkg/cluster/install.go
m1kola: Right. And and we are already logging an error here: 

https://github.com/Azure/ARO-RP/blob/4dd364aac52f1604285f42098869576a456812bd/pkg/util/steps/condition.go#L66


And setting `m.hiveRegistrator` to `nil` is not strictly necessary here. So I think we can simplify this down to this:

```go
steps.Condition(m.verifyRegistration, 2*time.Minute, false), // don't fail because of hive
```

pkg/cluster/registerWithHive.go
m1kola: Given [this](https://github.com/Azure/ARO-RP/pull/2145#discussion_r913089133), I think we should only return `true` when cluster addopted. In all other cases we should either return an `false, err` (to stop immediately) or `false, nil` (to continue trying).

P.S.: `wait.PollImmediateUntil` stops immediately when there is an error (even if we return `false`).

pkg/api/admin/openshiftcluster.go
m1kola: Updated

pkg/cluster/install.go
m1kola: Updated

pkg/hive/manager.go
```diff
@@ -4,36 +4,112 @@ package hive
...
 	}
 }
 
-func NewFromConfig(restConfig *rest.Config) (Interface, error) {
-	hiveclientset, err := hiveclient.NewForConfig(restConfig)
+func (hr *clusterManager) Register(ctx context.Context, workloadCluster *WorkloadCluster) (*hivev1.ClusterDeployment, error) {
+	var namespace string
+
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		namespace = "aro-" + uuid.Must(uuid.NewV4()).String()
```
hawkowl, line 71: I think we should reuse the existing UUID that we have for the cluster -- the document ID. There's a few reasons for this:

1. It is common to the cluster, not just our database. In generating the installconfig, we set the cluster UUID as the document ID.
2. It means we can remove the HiveProfile struct. We may need it for other things later, but we at least don't need to list the namespace.
3. At runtime, since it is common to the cluster, it can assist us in working backwards from a cluster in production to the Hive namespace, without having to do a second lookup from the RP.

pkg/hive/resources.go
```diff
@@ -0,0 +1,83 @@
...
+		Spec: hivev1.ClusterDeploymentSpec{
+			BaseDomain:  "",
+			ClusterName: clusterName,
+			Installed:   true,
+			ClusterMetadata: &hivev1.ClusterMetadata{
+				AdminKubeconfigSecretRef: corev1.LocalObjectReference{
+					Name: kubesecretName,
+				},
+				ClusterID: clusterID,
+				InfraID:   infraID},
```
hawkowl, line 69: wonky brace here :)

pkg/cluster/registerWithHive.go
```diff
@@ -0,0 +1,108 @@
...
+	if m.hiveClusterManager == nil {
+		return errors.New("no hive cluster manager, skipping")
+	}
+
+	cd, err := m.hiveClusterManager.Register(ctx, workloadCluster)
+	if err != nil {
+		return err
+	}
+
+	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, func(doc *api.OpenShiftClusterDocument) error {
```
hawkowl, line 73: we can get rid of this by using the doc UUID

pkg/api/admin/openshiftcluster.go
```diff
@@ -295,3 +296,7 @@ type SystemData struct {
 	LastModifiedByType CreatedByType `json:"lastModifiedByType,omitempty"`
 	LastModifiedAt     *time.Time    `json:"lastModifiedAt,omitempty"`
 }
+
+type HiveProfile struct {
```
hawkowl, line 300: this can be removed by using the doc uuid

pkg/api/openshiftcluster.go
```diff
@@ -150,6 +151,8 @@ type OpenShiftClusterProperties struct {
 	UserAdminKubeconfig SecureBytes `json:"userAdminKubeconfig,omitempty"`
 
 	RegistryProfiles []*RegistryProfile `json:"registryProfiles,omitempty"`
+
+	HiveProfile HiveProfile `json:"hiveProfile,omitempty"`
```
hawkowl, line 155: this can be removed by using the doc uuid (etc)

pkg/cluster/cluster.go
```diff
@@ -120,6 +124,15 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 
 	storage := storage.NewManager(_env, r.SubscriptionID, fpAuthorizer)
 
+	// TODO: always set hive once we have it everywhere in prod and dev
+	var hr hive.ClusterManager
```
hawkowl, line 128: I don't super love the idea of landing this -- but I guess it's harmless for now.

hawkowl (changes requested):

pkg/api/admin/openshiftcluster.go
m1kola: Resolving thread in favour of https://github.com/Azure/ARO-RP/pull/2145#discussion_r913319980

pkg/cluster/registerWithHive.go
m1kola: Resolving thread in favour of https://github.com/Azure/ARO-RP/pull/2145#discussion_r913319980

pkg/hive/manager.go
m1kola: @hawkowl this solves (I think?) the issue of uniqueness as we can assume that DB ids are unique and we don't have to retry on conflict here. And, as you said, we will be able to work from cluster to hive.

However very common SRE/tooling use case will be working from IcM to Hive (troubleshooting failed provisioning/deprovisioning/updates). In IcM we will have a cluster ID and a region. With that we can get [`OpenShiftCluster`](https://github.com/Azure/ARO-RP/blob/8f0a38c1919e54a9cf3931a361ca98f26534f858/pkg/api/openshiftcluster.go#L11-L23), but not [`OpenShiftClusterDocument`](https://github.com/Azure/ARO-RP/blob/8f0a38c1919e54a9cf3931a361ca98f26534f858/pkg/api/openshiftclusterdocument.go#L20-L48) which contains an the ID you suggest to use.

So we won't have a way of figuring out namespace on the AKS/Hive cluster via Geneva action for a specific cluster in many cases (failed provisioning or when API server is down).

m1kola:
@hawkowl I'm going to merge this PR to unblock other work related to this. I replied to [your suggestion](https://github.com/Azure/ARO-RP/pull/2145#discussion_r913319980) about uuids and `HiveProfile` Please let me know what you think - we can change it separately despite it being in master (we don't use it just yet) if we find an alternative.

pkg/hive/manager.go
hawkowl: Won't our basic checks automation be able to spit out that ID for us?

m1kola: @hawkowl no. Basic checks get cluster endpoint and we marshall `doc.OpenShiftCluster`, not `doc` so `doc.ID` is not included in the response.

PR #2152, opened 2022-06-03
Panic on AdminUpdate with MaintenanceTaskEverything

LiniSusan:

Fixes: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14361162

Admin put or patch fails with runtime error: index out of range [0] with length 0 when API server is unavailable (due to VMs being powered off)

### Test plan for issue:
Power off the VMs and run admin put or patch: update task starts, it will now skip the process, if there is no VMs with a Warn message

### DIFF
--- a/pkg/cluster/deploystorage_resources.go
+++ b/pkg/cluster/deploystorage_resources.go
@@ -77,6 +77,10 @@ func (m *manager) clusterServicePrincipalRBAC() *arm.Resource {
 // Legacy storage accounts (public) are not encrypted and cannot be retrofitted.
 // The flag controls this behavior in update/create.
 func (m *manager) storageAccount(name, region string, encrypted bool) *arm.Resource {
+        if(len(m.doc.OpenShiftCluster.Properties.WorkerProfiles) == 0){
+              m.log.Warn("Worker Profiles missing.")
+              return nil
+        }
 	virtualNetworkRules := []mgmtstorage.VirtualNetworkRule{
 		{
 			VirtualNetworkResourceID: &m.doc.OpenShiftCluster.Properties.MasterProfile.SubnetID,
--- a/pkg/cluster/deploystorage_resources.go
+++ b/pkg/cluster/deploystorage_resources.go
@@ -77,10 +77,6 @@ func (m *manager) clusterServicePrincipalRBAC() *arm.Resource {
 // Legacy storage accounts (public) are not encrypted and cannot be retrofitted.
 // The flag controls this behavior in update/create.
 func (m *manager) storageAccount(name, region string, encrypted bool) *arm.Resource {
-        if(len(m.doc.OpenShiftCluster.Properties.WorkerProfiles) == 0){
-              m.log.Warn("Worker Profiles missing.")
-              return nil
-        }
 	virtualNetworkRules := []mgmtstorage.VirtualNetworkRule{
 		{
 			VirtualNetworkResourceID: &m.doc.OpenShiftCluster.Properties.MasterProfile.SubnetID,
--- a/pkg/cluster/storageaccounts.go
+++ b/pkg/cluster/storageaccounts.go
@@ -70,16 +70,12 @@ func (m *manager) enableServiceEndpoints(ctx context.Context) error {
 	return nil
 }
 
-func (m *manager) isWorkerProfileAvailable() bool {
-	return len(m.doc.OpenShiftCluster.Properties.WorkerProfiles) == 0
-}
-
 // migrateStorageAccounts redeploys storage accounts with firewall rules preventing external access
 // The encryption flag is set to false/disabled for legacy storage accounts.
 func (m *manager) migrateStorageAccounts(ctx context.Context) error {
 	resourceGroup := stringutils.LastTokenByte(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
-	if m.isWorkerProfileAvailable() {
-		m.log.Error("Unable to create storage resource due to missing worker profiles.")
+	if len(m.doc.OpenShiftCluster.Properties.WorkerProfiles) == 0 {
+		m.log.Error("Skipping migrateStorageAccounts due to missing WorkerProfiles.")
 		return nil
 	}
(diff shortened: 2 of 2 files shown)


pkg/cluster/deploystorage_resources.go
```diff
@@ -77,6 +77,10 @@ func (m *manager) clusterServicePrincipalRBAC() *arm.Resource {
 // Legacy storage accounts (public) are not encrypted and cannot be retrofitted.
 // The flag controls this behavior in update/create.
 func (m *manager) storageAccount(name, region string, encrypted bool) *arm.Resource {
+        if(len(m.doc.OpenShiftCluster.Properties.WorkerProfiles) == 0){
+              m.log.Warn("Worker Profiles missing.")
```
darthhexx, line 81: It would be good to include the fact that we are `Unable to create storage resource due to missing worker profiles.`, so that the log message has some context, and log it as an error not a warning. That said, while it is good practice to defend against calling this function without at least 1 worker profile, none of the callers have any logic handling a `nil` response:
https://github.com/Azure/ARO-RP/blob/ac78ec518b52f305ce
b1d8c2f0166534512cebee/pkg/cluster/deploystorage.go#L121
https://github.com/Azure/ARO-RP/blob/ac78ec518b52f305ceb1d8c2f0166534512cebee/pkg/cluster/deploystorage.go#L124
https://github.com/Azure/ARO-RP/blob/ac78ec518b52f305ceb1d8c2f0166534512cebee/pkg/cluster/storageaccounts.go#L84
https://github.com/Azure/ARO-RP/blob/ac78ec518b52f305ceb1d8c2f0166534512cebee/pkg/cluster/storageaccounts.go#L85
https://github.com/Azure/ARO-RP/blob/ac78ec518b52f305ceb1d8c2f0166534512cebee/pkg/deploy/generator/resources_gateway.go#L829
https://github.com/Azure/ARO-RP/blob/ac78ec518b52f305ceb1d8c2f0166534512cebee/pkg/deploy/generator/resources_rp.go#L1866
https://github.com/Azure/ARO-RP/blob/ac78ec518b52f305ceb1d8c2f0166534512cebee/pkg/deploy/generator/resources_rp.go#L1899

I would say that changing the functionality of all of the functions using this one is outside the scope of this PR. The intention is to skip `migrateStorageAccounts` during an AdminUpdate, so it would be prudent to check for the worker profile length [here in that function](https://github.com/Azure/ARO-RP/blob/ac78ec518b52f305ceb1d8c2f0166534512cebee/pkg/cluster/storageaccounts.go#L76) and gracefully skip running it. Like is being done here: https://github.com/Azure/ARO-RP/blob/ac78ec518b52f305ceb1d8c2f0166534512cebee/pkg/cluster/arooperator.go#L18-L21

darthhexx (changes requested):

pkg/cluster/deploystorage_resources.go
LiniSusan: @darthhexx , made the changes, could you please review it and let me know the comments/suggestions

pkg/cluster/storageaccounts.go
```diff
@@ -70,10 +70,18 @@ func (m *manager) enableServiceEndpoints(ctx context.Context) error {
 	return nil
 }
 
+func (m *manager) isWorkerProfileAvailable() bool {
```
darthhexx, line 73: There is no need to add additional complexity and overhead with a function to check the worker profile length, because we are only doing the length check in one place in this file.

pkg/cluster/storageaccounts.go
```diff
@@ -70,10 +70,18 @@ func (m *manager) enableServiceEndpoints(ctx context.Context) error {
...
+	return len(m.doc.OpenShiftCluster.Properties.WorkerProfiles) == 0
+}
+
 // migrateStorageAccounts redeploys storage accounts with firewall rules preventing external access
 // The encryption flag is set to false/disabled for legacy storage accounts.
 func (m *manager) migrateStorageAccounts(ctx context.Context) error {
 	resourceGroup := stringutils.LastTokenByte(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
+	if m.isWorkerProfileAvailable() {
```
darthhexx, line 81: In this case, just using the `len(m.doc.OpenShiftCluster.Properties.WorkerProfiles) == 0` here would be fine.

pkg/cluster/storageaccounts.go
```diff
@@ -70,10 +70,18 @@ func (m *manager) enableServiceEndpoints(ctx context.Context) error {
...
+		m.log.Error("Unable to create storage resource due to missing worker profiles.")
```
darthhexx, line 82: We should be stating that we are `Skipping migrateStorageAccounts due to missing worker profiles` or similar wording.

pkg/cluster/storageaccounts.go
LiniSusan: @darthhexx : updated as suggested

pkg/cluster/storageaccounts.go
LiniSusan: @darthhexx : done, could you please verify the changes?

darthhexx: Changes look great. We can merge once CI goes green.

pkg/cluster/storageaccounts.go
```diff
@@ -74,6 +74,10 @@ func (m *manager) enableServiceEndpoints(ctx context.Context) error {
 // The encryption flag is set to false/disabled for legacy storage accounts.
 func (m *manager) migrateStorageAccounts(ctx context.Context) error {
 	resourceGroup := stringutils.LastTokenByte(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
+	if len(m.doc.OpenShiftCluster.Properties.WorkerProfiles) == 0 {
+		m.log.Error("Skipping migrateStorageAccounts due to missing WorkerProfiles.")
```
ross-bryan, line 78: nit: we typically log messages starting with lower case character

ross-bryan (approved):
lgtm, 1 nit

PR #2153, opened 2022-06-03
trying out some stuff with github action

facchettos:

Fixes


facchettos:
it already generates the license with you don't use `-validate`.
added `.git` to the list of ignored files

PR #2154, opened 2022-06-03
Adding validations for subnet provisioning state

Shivkumar13:

ADO Card:
Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13937166/

This PR adds the functionality to check the provisioning state for the Subnet.
If the subnet provisioning state is not in Succeeded state, there will be an exception raised.

### Test plan for issue:

Testing can be done by keeping the master/worker subnet state in a `failed` or `updating` state,
 anything other than `Succeeded` and run the installation with `ARM Template`, `Bicept`, `az aro`, `Terraform` etc.
Cluster Installation shouldn't proceed if Subnets are not in the desired state.

### DIFF
--- a/pkg/api/validate/dynamic/dynamic.go
+++ b/pkg/api/validate/dynamic/dynamic.go
@@ -389,10 +389,11 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 			}
 		}
 
-		if ss.SubnetPropertiesFormat.ProvisioningState != "Succeeded" {
+		if ss.SubnetPropertiesFormat == nil ||
+			ss.SubnetPropertiesFormat.ProvisioningState != mgmtnetwork.Succeeded {
 			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidLinkedVNet, s.Path, "The provided subnet '%s' is not in a Succeeded state", s.ID)
 		}
-		 
+
 		_, net, err := net.ParseCIDR(*ss.AddressPrefix)
 		if err != nil {
 			return err
--- a/pkg/api/validate/dynamic/dynamic.go
+++ b/pkg/api/validate/dynamic/dynamic.go
@@ -389,6 +389,10 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 			}
 		}
 
+		if ss.SubnetPropertiesFormat.ProvisioningState != "Succeeded" {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidLinkedVNet, s.Path, "The provided subnet '%s' is not in a Succeeded state", s.ID)
+		}
+		 
 		_, net, err := net.ParseCIDR(*ss.AddressPrefix)
 		if err != nil {
 			return err
--- a/pkg/api/validate/dynamic/dynamic_test.go
+++ b/pkg/api/validate/dynamic/dynamic_test.go
@@ -598,7 +598,7 @@ func TestValidateSubnets(t *testing.T) {
 			},
 		},
 		{
-			name: "fail: subnet doe not exist on vnet",
+			name: "fail: subnet does not exist on vnet",
 			vnetMocks: func(vnetClient *mock_network.MockVirtualNetworksClient, vnet mgmtnetwork.VirtualNetwork) {
 				vnet.Subnets = nil
 				vnetClient.EXPECT().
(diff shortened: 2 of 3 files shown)


bennerv (dismissed):
Is there a reason we're adding this check in the azure-cli command_module instead of the RP validation?  Is it already in the RP validation?  

The reasoning is that not everyone creates an ARO cluster using the azure-cli.  Some people use ARM templates, bicept, terraform, etc.

ross-bryan (changes requested):
+1 to @bennerv, please implement this server side, although client side will be welcome as well.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -389,6 +389,10 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 			}
 		}
 
+		if ss.SubnetPropertiesFormat.ProvisioningState != "Succeeded" {
```
AldoFusterTurpin, line 392: Instead of using the hardcoded string "Succeeded", what about using the [ProvisioningState](https://github.com/Azure/azure-sdk-for-go/blob/main/services/network/mgmt/2020-08-01/network/enums.go#L1896) provided by Azure. 

In fact, we already import the network package on top of the file on line 14:
```go
```
so you just need to change your if with one of the two options:
```go
if ss.SubnetPropertiesFormat.ProvisioningState != mgmtnetwork.ProvisioningState(api.ProvisioningStateSucceeded) {
```
or
```go
if ss.SubnetPropertiesFormat.ProvisioningState != mgmtnetwork.Succeeded {
```
?
I would prefer the second one as does not need any type casting.

AldoFusterTurpin (changes requested):
@Shivkumar13 I have requested changes to not use the hardcoded string in the if condition.

Thanks!

pkg/api/validate/dynamic/dynamic.go
Shivkumar13: Oh cool, fixed it!

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -389,6 +389,10 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 			}
 		}
 
+		if ss.SubnetPropertiesFormat.ProvisioningState != mgmtnetwork.Succeeded {
```
s-amann, line 392: It looks like `ss.SubnetPropertiesFormat` can be nil, can we check for this similar to line 385?

s-amann (changes requested):
Please confirm `ss.SubnetPropertiesFormat` could be nil and add the corresponding guard check.

pkg/api/validate/dynamic/dynamic.go
Shivkumar13: Done!

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -389,6 +389,11 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 			}
 		}
 
+		if ss.SubnetPropertiesFormat == nil ||	
+			    if ss.SubnetPropertiesFormat.ProvisioningState != mgmtnetwork.Succeeded {
```
s-amann, line 393: This won't compile with the second `if` inlined here. Once this is resolved it looks like there will be some tests that need updating in `pkg/api/validate/dynamic_test.go` specifically the test function `TestValidateSubnets` on line 558 with the tests named:
* "pass", 
* "pass (master subnet)", 
* "pass (cluster in creating provisioning status)", 
* "fail: invalid subnet CIDR", "fail: too small subnet CIDR"


While updating these tests to pass, please consider adding a new test for the new condition/error message which can be returned.
```suggestion
			     ss.SubnetPropertiesFormat.ProvisioningState != mgmtnetwork.Succeeded {
```

s-amann (changes requested):
Please see my comment below.

pkg/api/validate/dynamic/dynamic.go
Shivkumar13: Yup rechecked and removed unnecessary if, and added tests too.
Please let me know in case of any changes are needed. Thank you!

AldoFusterTurpin (changes requested):
@Shivkumar13 I think there is a compilation error. Check details below. 

Thank you!

pkg/api/validate/dynamic/dynamic_test.go
```diff
@@ -607,6 +607,16 @@ func TestValidateSubnets(t *testing.T) {
 			},
 			wantErr: "400: InvalidLinkedVNet: : The provided subnet '" + masterSubnet + "' could not be found.",
 		},
+		{
+			name: "fail: subnet provisioning state is not succeeded",
+			vnetMocks: func(vnetClient *mock_network.MockVirtualNetworksClient, vnet mgmtnetwork.VirtualNetwork) {
+				(*vnet.Subnets)[0].ProvisioningState != mgmtnetwork.Succeeded
```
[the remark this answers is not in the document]
Shivkumar13: Should be good now, provided the `mgmtnetwork.succeeded` to the vnet in the `t.Run` block, tests are green now.


(pyenv) Shivkumar's~/go/src/github.com/Azure/ARO-RP>/usr/local/go/bin/go test -timeout 30s -run ^TestValidateSubnets$ github.com/Azure/ARO-RP/pkg/api/validate/dynamic
ok      github.com/Azure/ARO-RP/pkg/api/validate/dynamic        (cached)

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -389,6 +389,11 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 			}
 		}
 
+		if ss.SubnetPropertiesFormat == nil ||
+			ss.SubnetPropertiesFormat.ProvisioningState != mgmtnetwork.Succeeded {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidLinkedVNet, s.Path, "The provided subnet '%s' is not in a Succeeded state", s.ID)
```
s-amann, line 394: nit: for future reference (unless you want to make this small change) `%q` is the same as `'%s'`

s-amann (approved):
LGTM, thank you for addressing all the feedback!

pkg/api/validate/dynamic/dynamic.go
Shivkumar13: Ack!

petrkotas (approved):
I believe this is ok now. Thankyou @Shivkumar13 for the work here.

ross-bryan (approved):

darthhexx (approved):
Great work, LGTM.

ross-bryan (approved):
lgtm

PR #2155, opened 2022-06-03
include openshift-operator-lifecycle-manager in monitoring

s-amann:

Fixes the drop of monitoring 'openshift-operator-lifecycle-manager'. [ADO Ticket 14657169](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14657169/)

We were previously monitoring this namespace which lead to a few clusters with unresolved issues. This PR is for the event we want to continue monitoring it.

### Test plan for issue:
unit test

no.

### DIFF
--- a/pkg/util/namespace/namespace.go
+++ b/pkg/util/namespace/namespace.go
@@ -66,6 +66,7 @@ func IsOpenShiftNamespace(ns string) bool {
 		"openshift-oauth-apiserver":                        {},
 		"openshift-openstack-infra":                        {},
 		"openshift-operators":                              {},
+		"openshift-operator-lifecycle-manager":             {},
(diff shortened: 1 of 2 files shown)


SudoBrendan (approved):

darthhexx (approved):
I know there was a ticket recently and some discussion about whether this namespace should be considered as user workload or not, because it manages user introduced add-ons into clusters. IMO we should monitor it, because its core to enabling a feature we are currently working towards enabling with Hive. LGTM, but please do add a link to the ADO work item in the description @s-amann.

s-amann:
@darthhexx 

Thank you for the reminder, ticket is created and linked!

PR #2157, opened 2022-06-03
Add name length validation on ARO clusters for non-zonal regions

bennerv:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14627987

In OCP 4.10 in non-zonal regions, VMs will be created with availability sets.  The names of these availability sets exceed the maximum length under certain circumstances, so we will prevent these circumstances from occuring by limiting the name length for these regions.  

### Test plan for issue:

Unit tests + shared dev env for australiasoutheast region cluster creation name length at 19 characters.

Peter is working on KCS article.  

Revert this once we have the fix upstream and set our default install version to this.

### DIFF
--- a/pkg/api/validate/openshiftcluster.go
+++ b/pkg/api/validate/openshiftcluster.go
@@ -1,0 +1,53 @@
+package validate
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"strings"
+)
+
+// TODO: remove the VM name validation after https://bugzilla.redhat.com/show_bug.cgi?id=2093044 is resolved
+
+// Max length of a ARO cluster name is determined by the length of the generated availability set
+// in the machine-api provider.  It varies per region, but to remain consistent across non-zonal regions
+// we hardcode this to 19, which is the max for all non-zonal regions.
+//
+// the generated availability set must be <= 80 characters and can be calculated below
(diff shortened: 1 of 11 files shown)


pkg/api/validate/openshiftcluster.go
```diff
@@ -0,0 +1,56 @@
...
+	"switzerlandwest":    true,
+	"uaenorth":           true,
+	"ukwest":             true,
+	"westcentralus":      true,
+	"westus":             true,
+}
+
+// OpenShiftClusterNameLengthValid validates that the name does not exceed the maximumLength
+// which is in place for non-zonal regions due to https://bugzilla.redhat.com/show_bug.cgi?id=2093044
+func OpenShiftClusterNameLength(name, location string) bool {
```
25region, line 50: Did you mean to call it `OpenShiftClusterNameLengthValid`?

bennerv: Fixed.  Calling it: `OpenShiftClusterNameLength` and since it's an exported function, it can be called with:
`validate.OpenShiftClusterNameLength` or `apiValidate.OpenShiftClusterNameLength`

25region: Ok, comment is fixed then. Thanks

s-amann (dismissed):
Thanks for getting this out so quickly, LGTM

ross-bryan (approved):
lgtm

PR #2158, opened 2022-06-03
Truncate cluster names to 19 char in e2e pipelines

tony-schndr:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14628018/

This PR updates all e2e tests where CLUSTER name is generated.  The region under test is checked if it is a non-zonal region, if so the generated name is truncated to 19 characters.  If the region under test is a zonal region, it is generated as it was before.

### Test plan for issue:
Locally tested the bash commands in test script to validate logic / commands are correct.
e2e tests will need to be monitored, visually validating the cluster names for non-zonal regions are truncated to 19 characters and not truncated for all other regions.

N/A

### DIFF
--- a/.pipelines/e2e-with-billing.yml
+++ b/.pipelines/e2e-with-billing.yml
@@ -48,7 +48,16 @@ stages:
     - script: |
         # Pass variables between tasks: https://medium.com/microsoftazure/how-to-pass-variables-in-azure-pipelines-yaml-tasks-5c81c5d31763
         echo "##vso[task.setvariable variable=REGION]$LOCATION"
-        CLUSTER="v4-e2e-V$BUILD_BUILDID-$LOCATION"
+        # TODO: Remove this hack after AvailabilitySet name too long bug is fixed.
+        NONZONAL_REGIONS="australiacentral australiacentral2 australiasoutheast brazilsoutheast canadaeast japanwest northcentralus norwaywest southindia switzerlandwest uaenorth ukwest westcentralus westus"
+        if echo $NONZONAL_REGIONS | grep -wq $LOCATION
+        then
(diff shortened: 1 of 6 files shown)


.pipelines/e2e-with-billing.yml
```diff
@@ -48,7 +48,16 @@ stages:
     - script: |
         # Pass variables between tasks: https://medium.com/microsoftazure/how-to-pass-variables-in-azure-pipelines-yaml-tasks-5c81c5d31763
         echo "##vso[task.setvariable variable=REGION]$LOCATION"
-        CLUSTER="v4-e2e-V$BUILD_BUILDID-$LOCATION"
+        # TODO: Remove this hack after AvailabilitySet name too long bug is fixed.
+        NONZONAL_REGIONS="australiacentral australiacentral2 australiasoutheast brazilsoutheast canadaeast centraluseuap eastus2euap japanwest northcentralus norwaywest southafricanorth southindia switzerlandwest uaenorth ukwest westcentralus westus"
+        if echo $NONZONAL_REGIONS | grep -w $LOCATION > /dev/null;
```
25region, line 53: Could use `grep -q` for no output, but the same thing really :)
```suggestion
        if echo $NONZONAL_REGIONS | grep -wq $LOCATION
```

25region (dismissed):

.pipelines/e2e-with-billing.yml
tony-schndr: Thanks for the feedback!  Cleaner solution :)

ross-bryan (approved):

bennerv (approved):
lgtm

PR #2161, opened 2022-06-06
make monitor Ignore preempted completed pods

SrinivasAtmakuri:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14233929/

### DIFF
--- a/pkg/monitor/cluster/podconditions.go
+++ b/pkg/monitor/cluster/podconditions.go
@@ -9,6 +9,7 @@ import (
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/pkg/kubelet/events"
 
 	"github.com/Azure/ARO-RP/pkg/util/namespace"
 )
@@ -56,6 +57,10 @@ func (mon *Monitor) _emitPodConditions(ps *corev1.PodList) {
 		}
 
 		if p.Status.Phase == corev1.PodSucceeded {
+			continue
+		}
+
+		if p.Status.Reason == events.PreemptContainer {
 			continue
 		}
 
--- a/vendor/k8s.io/kubernetes/pkg/kubelet/events/event.go
+++ b/vendor/k8s.io/kubernetes/pkg/kubelet/events/event.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2014 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


pkg/monitor/cluster/podconditions.go
```diff
@@ -59,6 +59,16 @@ func (mon *Monitor) _emitPodConditions(ps *corev1.PodList) {
...
+		if p.Status.Reason == "Preempting" {
+			for _, cs := range p.Status.ContainerStatuses {
+				if cs.State.Terminated != nil {
+					if cs.State.Terminated.Reason == "Completed" {
+						continue
+					}
+				}
+			}
+		}
+
```
AldoFusterTurpin, line 71: Quick idea, what about:

1. extracting the hardcoded strings into constants (it has the advantatge that if the string changes, the line checking the conditions will not change)

2. simplifying the 2 if's into a single one (advantage of less nested code and a bit cleaner)

Something like:
```go
const Preempting = "Preempting"
const Completed = "Completed"

if p.Status.Reason == Preempting {
    for _, cs := range p.Status.ContainerStatuses {
        if cs.State.Terminated != nil && cs.State.Terminated.Reason == Completed {
            continue
        }
    }
}
```

AldoFusterTurpin (commented):
Just a little suggestion (IMHO), in case you find it useful 🙂

SrinivasAtmakuri:
Thanks, made changes as suggested.

pkg/monitor/cluster/podconditions.go
```diff
@@ -59,6 +62,14 @@ func (mon *Monitor) _emitPodConditions(ps *corev1.PodList) {
 			continue
 		}
 
+		if p.Status.Reason == Preempting {
+			for _, cs := range p.Status.ContainerStatuses {
+				if cs.State.Terminated != nil && cs.State.Terminated.Reason == Completed {
+					continue
+				}
+			}
```
SudoBrendan, line 70: I think this is more a question about the acceptance criteria of the WI rather than the code, but should we just ignore any Pod marked as `Preempting`? We know that, eventually, this Pod will have all it's container Terminated - so do we actually need to check container status (in this case, that any 1 container in this Pod has been terminated) before ignoring, or can we ignore that race condition all-together?

SudoBrendan (commented):
Code looks good, just a question about acceptance criteria.

pkg/monitor/cluster/podconditions.go
SrinivasAtmakuri: Agree with the suggestion, and will make changes to skip all pods which have reason Preempting. Thank you!

pkg/monitor/cluster/podconditions.go
```diff
@@ -22,6 +22,8 @@ var podConditionsExpected = map[corev1.PodConditionType]corev1.ConditionStatus{
 
 var restartCounterThreshold int32 = 10
 
+const Preempting string = "Preempting"
```
hawkowl, line 25: You can access this constant as `k8s.io/kubernetes/pkg/kubelet/events.PreemptContainer` -- you'll just have to run a `go mod vendor` after.

SrinivasAtmakuri: Updated as suggested.

hawkowl (approved):

PR #2162, opened 2022-06-07
Update gateway arm template deployment name and log message

ross-bryan:
Update gateway arm template deployment name and log message

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14657615

### Test plan for issue:

existing unit/e2e, just minor text/string change

I am updating egress lockdown enabling wiki and this fix makes it more clear that the private endpoint was deployed


bennerv (approved):
lgtm;  should have no impact as arm template changing names won't make a difference as long as the resource name (private endpoint) is the same.

PR #2165, opened 2022-06-08
Add OpenShiftVersions to the CosmosDB database

hawkowl:

Part of https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14651828

Adds an OpenShiftVersion database and the boilerplate to use it, as well as an API definition for the admin API. Frontend CRUD work will be in a separate PR.

### Test plan for issue:

Not much to test right now, since it's not yet properly used.

N/A


pkg/api/admin/openshiftversion.go
```diff
@@ -0,0 +1,17 @@
+package admin
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+// OpenShiftVersionList represents a list of OpenShift clusters.
```
s-amann, line 6: nit: this isn't returning a 'cluster' but the versions of openshift that are available, can we reword this a bit?

pkg/api/admin/openshiftversion.go
```diff
@@ -0,0 +1,17 @@
...
+type OpenShiftVersionList struct {
+	// The list of OpenShift clusters.
```
s-amann, line 8: nit: this is redundant 
```suggestion
```

s-amann (dismissed):
LGTM, just minor nits around comments

s-amann (dismissed):

SudoBrendan (dismissed):
LGTM

pkg/api/openshiftversion.go
```diff
@@ -0,0 +1,17 @@
...
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+// OpenShiftVersion represents an OpenShift version that can be installed
+type OpenShiftVersion struct {
+	MissingFields
+
+	ID       string `json:"id,omitempty"`
+	Deleting bool   `json:"deleting,omitempty"` // https://docs.microsoft.com/en-us/azure/cosmos-db/change-feed-design-patterns#deletes
```
petrkotas, line 11: @hawkowl for what use we need to track which version is being deleted?

petrkotas (dismissed):
Beside one question, this is a good one. Thanks

pkg/api/openshiftversion.go
bennerv: > The change feed does not capture deletes. If you delete an item from your container, it is also removed from the change feed. The most common method of handling this is adding a soft marker on the items that are being deleted. You can add a property called "deleted" and set it to "true" at the time of deletion. This document update will show up in the change feed. You can set a TTL on this item so that it can be automatically deleted later.

hawkowl: Yep. This isn't used right now, but I can foresee us perhaps using the changefeed to update/prune/etc images or something, so, might as well keep it.

s-amann (approved):

darthhexx (approved):
LGTM

PR #2166, opened 2022-06-08
To detect and show the unsupported conditions on the console banner

xiaoyu74:
To proactively detect the unsupported conditions of the ARO Cluster and show the notifications on the console banner for the end-user. The detailed description is in the task [ADO-7551905](https://msazure.visualstudio.com/AzureRedHatOpenShift/_boards/board/t/Firefly/Stories/?workitem=7551905)

To implement an ARO operator checker to detect "unsupported conditions" and displaying the notifications on the console banner automatically. Then, make the operator marked as `upgradable=False`.

We haven't considered all the unsupported conditions; this PR will mainly focus on the conditions below: 
- ARO Cluster has `less than 3 worker nodes`
- `Default route` is configured in the Cluster
- `Custom Network Security Group` is configured in the Cluster 

We will constantly update with more conditions based on the opinions/discussions/suggestions from the team

### Test plan for issue:
Testing with an ARO dev cluster to confirm the new checker is working and the notification is shown on the banner as expected.

- Deploy an ARO dev cluster

- In your local ARO-RP repo, run the ARO operator locally to test the new ARO checker:
`oc scale -n openshift-azure-operator deployment/aro-operator-master --replicas=0`
`CLUSTER=cluster-name go run -tags aro,containers_image_openpgp ./cmd/aro operator master`

- Testing the checker with the first unsupported condition which is too less worker nodes
`oc scale --replicas=<number-of-replicas> machineset/<machineset-name> -n openshift-machine-api`
`oc scale --replicas=0 machineset/shawn-dev-mbcp2-worker-australiaeast3 -n openshift-machine-api`

- Then, check if the notification is shown on the cluster console banner as expected. Also, the ARO operator upgradable status is set as "False"

- Scale up the worker node to 3 or more and the notification should be removed from the banner

N/A - This is a new ADO task, and I will document it after the task completed and test passed.


pkg/operator/controllers/checker/unsupportconditionchecker.go
```diff
@@ -0,0 +1,118 @@
...
+		// if there's no errors, object found then update
+		_, err = nc.consolecli.ConsoleV1().ConsoleNotifications().Update(ctx, banner, metav1.UpdateOptions{})
+		return err
+	} else {
+		v1helpers.SetStatusCondition(&co.Status.Conditions, condition[1])
+
+		err = nc.consolecli.ConsoleV1().ConsoleNotifications().Delete(ctx, bannerName, metav1.DeleteOptions{})
+		return err
+	}
+}
```
facchettos, line 95: I know this is still a draft PR, but this function is a little long, if you split it smaller chunks it will be easier to read, and more importantly, easier to test

xiaoyu74: @facchettos - Thanks Jeremy for your time to review and sorry for the late response, will update accordingly.

hawkowl:
@xiaoyu74 I think this is looking good so far.

facchettos:
before you spend more time on it, I think we should have a discussion about it with PMs. I had a talk with Jerome regarding some openshift version out-of-support notification and I am afraid what he told me also applies here. I will start a thread on slack with everyone involved

xiaoyu74:
> before you spend more time on it, I think we should have a discussion about it with PMs. I had a talk with Jerome regarding some openshift version out-of-support notification and I am afraid what he told me also applies here (that we should only do AZComms). I will start a thread on slack with everyone involved

@facchettos - Thanks for your reminder. I checked the discussion in the slack thread. Looks we should do both (console banner and AzComms) based on Jerome's reply. I assume the AzComms implementation will be a separated task and won't affect the current PR for the implementation with console banner.

facchettos:
Correct. Jerome changed his mind regarding console banner in the last few days, so now the banner is cool, just wanted to make sure we were not using engineering time on something the PMs didn't agree

xiaoyu74:
Talked with David for this PR, will do some more refactoring for the code

pkg/operator/controllers/checker/unsupportconditionchecker.go
```diff
@@ -0,0 +1,122 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	clusterOperatorName = "aro"
+	reasonAsExpected    = "AsExpected"
+)
+
+type NodesChecker struct {
```
darthhexx, line 24: This name should change change to `UnsupportedConditionChecker` so we can have functions for each condition we check: `CheckWorkerNodeCount`, `CheckOpenShiftVersion`, `CheckNsg`, `CheckDefaultRoute`, etc.

pkg/operator/controllers/checker/unsupportconditionchecker.go
```diff
@@ -0,0 +1,122 @@
...
+
+		// if there's no errors, object found then update
+		_, err = nc.consolecli.ConsoleV1().ConsoleNotifications().Update(ctx, banner, metav1.UpdateOptions{})
+		return err
+	} else {
+		nc.setCluOperatorStatus(ctx, 1)
+
+		err = nc.consolecli.ConsoleV1().ConsoleNotifications().Delete(ctx, bannerName, metav1.DeleteOptions{})
+		return err
+	}
```
darthhexx, line 77: This `else` is not required because the block above is `return`ing.

pkg/operator/controllers/checker/unsupportconditionchecker.go
```diff
@@ -0,0 +1,122 @@
...
+
+		// if the object doesn't exist Create
+		if err != nil && kerrors.IsNotFound(err) {
+			_, err = nc.consolecli.ConsoleV1().ConsoleNotifications().Create(ctx, banner, metav1.CreateOptions{})
+			return err
+		}
+
+		// if there's no errors, object found then update
+		_, err = nc.consolecli.ConsoleV1().ConsoleNotifications().Update(ctx, banner, metav1.UpdateOptions{})
+		return err
```
darthhexx, line 71: This block should be added to the `newBanner` function, which should be refactored to be an `addConsoleBanner` function that appends the message to any existing messages.

pkg/operator/controllers/checker/unsupportconditionchecker.go
```diff
@@ -0,0 +1,122 @@
...
+	} else {
+		nc.setCluOperatorStatus(ctx, 1)
+
+		err = nc.consolecli.ConsoleV1().ConsoleNotifications().Delete(ctx, bannerName, metav1.DeleteOptions{})
+		return err
+	}
+}
+
+func (nc *NodesChecker) setCluOperatorStatus(ctx context.Context, status_index int) {
```
darthhexx, line 80: Please refactor this function to use condition status' constants and not an int:
```
const (
	OperatorUpgradableFalse = configv1.ConditionFalse
	OperatorUpgradableTrue  = configv1.ConditionTrue
)

// calling it with
ucc.setClusterOperatorStatus(ctx, OperatorUpgradableFalse)

func (ucc *UnsupportedConditionChecker) setClusterOperatorStatus(ctx context.Context, upgradeableStatus configv1.ConditionStatus) {
	co, _ := ucc.getOrCreateClusterOperator(ctx)

	condition := configv1.ClusterOperatorStatusCondition{
		Type:               configv1.OperatorUpgradeable,
		Status:             upgradeableStatus,
		LastTransitionTime: metav1.Now(),
		Reason:             reasonAsExpected,
	}

	v1helpers.SetStatusCondition(&co.Status.Conditions, condition)
}
```

pkg/operator/controllers/checker/unsupportconditionchecker.go
```diff
@@ -0,0 +1,122 @@
...
+			LastTransitionTime: currentTime,
+			Reason:             reasonAsExpected,
+		},
+	}
+	v1helpers.SetStatusCondition(&co.Status.Conditions, condition[status_index])
+}
+
+func (nc *NodesChecker) getOrCreateClusterOperator(ctx context.Context) (*configv1.ClusterOperator, error) {
+	co, err := nc.configcli.ConfigV1().ClusterOperators().Get(ctx, clusterOperatorName, metav1.GetOptions{})
+	if !kerrors.IsNotFound(err) {
```
darthhexx, line 103: This double negative is really hard to read at first and the function is missing the create step. In fact it should only get the cluster operator and return an error if it can't. The creation is l;eft to the operator's reconcile operation:https://github.com/Azure/ARO-RP/blob/9f472ae57106c793ec1af373627eb77eac86d215/pkg/operator/controllers/clusteroperatoraro/clusteroperatoraro_controller.go#L134-L150

pkg/operator/controllers/checker/unsupportconditionchecker.go
```diff
@@ -0,0 +1,122 @@
...
+		configcli:     configcli,
+	}
+}
+
+func (nc *NodesChecker) Name() string {
+	return "WorkerNodesCountChecker"
+}
+
+func (nc *NodesChecker) Check(ctx context.Context) error {
+	bannerName := "aro-sre"
```
darthhexx, line 43: This should be a constant.

darthhexx (changes requested):

facchettos (changes requested):
some little nit pick naming wise. 
`unsupportconditionchecker.go` should probabbly be named `unsupportedconditionchecker.go`

The naming in this package is a little inconsistent. For example we have some files without `checker` at the end (`clusterdns` and `ingresscert`) and some with. Some renaming could be nice. It's not your fault, but I think it should be done and this could be a great addition to that PR. (maybe in a separated commit).  

Also, I would say that the package name should be a plural (since it contains a bunch of checkers), it would be a little more accurate. And since `checker(s)` is in the package name, maybe having it on all file names is a little redundant. 

WDYT?

darthhexx (dismissed):
This LGTM, however the filename _unsupportconditionchecker.go_ does need to be changed to _unsupport**ed**conditionchecker.go_ before we merge.

darthhexx:
@facchettos nice catch on the filename :+1: I agree the naming is a little inconsistent in the package, but not in the files created or modified in this PR and should be vigilant about PR scope-creep.

pkg/operator/controllers/checker/unsupportconditionchecker.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (ucc *UnsupportedConditionChecker) Check(ctx context.Context) error {
+	if ucc.role != operator.RoleMaster {
+		return nil
+	}
+
+	return ucc.CheckWorkerNodeCount(ctx)
+}
+
+func (ucc *UnsupportedConditionChecker) CheckWorkerNodeCount(ctx context.Context) error {
```
facchettos, line 66: This does not need to be exported

facchettos (changes requested):
Now that I have looked a little bit more into the code, I have some more changes to request. 

One of the functions is needlessly exported. That would be great to unexport it. 

Also we should add some tests.

facchettos:
@darthhexx fair enough. That is why I was suggesting to have it in a different commit but I can see your point. As long as a PR is open to fix that I have no problem with it

xiaoyu74:
> some little nit pick naming wise. `unsupportconditionchecker.go` should probabbly be named `unsupportedconditionchecker.go`
> 
> The naming in this package is a little inconsistent. For example we have some files without `checker` at the end (`clusterdns` and `ingresscert`) and some with. Some renaming could be nice. It's not your fault, but I think it should be done and this could be a great addition to that PR. (maybe in a separated commit).
> 
> Also, I would say that the package name should be a plural (since it contains a bunch of checkers), it would be a little more accurate. And since `checker(s)` is in the package name, maybe having it on all file names is a little redundant.
> 
> WDYT?

@facchettos - Thanks a lot for your review feedback. I agree that the naming for the checkers need to be more consistent. I will update the file name to "unsupportedconditionchecker.go" first and have the package name (checkers) and other checker file names as suggested in a separated PR.

pkg/operator/controllers/checker/unsupportedconditionchecker.go
```diff
@@ -0,0 +1,146 @@
...
+		return err
+	}
+
+	if len(nodes.Items) < WORKER_NODE_MINIMUM_COUNT {
+		err = ucc.setClusterOperatorStatus(ctx, OperatorUpgradableFalse)
+		if err != nil {
+			return err
+		}
+
+		return ucc.addConsoleBanner(ctx, fmt.Sprintf("Unsupported Cluster State: There need to be at least %d worker nodes running", WORKER_NODE_MINIMUM_COUNT))
```
SudoBrendan, line 80: need > need**s**

pkg/operator/controllers/checker/unsupportedconditionchecker.go
```diff
@@ -0,0 +1,146 @@
...
+
+	condition := configv1.ClusterOperatorStatusCondition{
+		Type:               configv1.OperatorUpgradeable,
+		Status:             upgradeableStatus,
+		LastTransitionTime: metav1.Now(),
+		Reason:             reasonAsExpected,
+	}
+
+	v1helpers.SetStatusCondition(&co.Status.Conditions, condition)
+	ucc.log.Infof("Set the %q operator upgradeable status to %q", clusterOperatorName, upgradeableStatus)
```
SudoBrendan, line 106: nit: do we need this log message? All logs go through the Gateway, so any traffic we can omit is a good thing :) Will this message enable us to take action on the cluster?

SudoBrendan (dismissed):
This PR has come a long way and is looking great!

Please fix the typo in the customer-facing message.

pkg/operator/controllers/checker/unsupportedconditionchecker.go
xiaoyu74: @SudoBrendan - Thanks for your review.  The reason for having the log message is  that we would like to provide with more details if a customers queries us as to why they received message X on date Y, we are able to look up in the logs and see what message was in the banner and for how long it was there.

facchettos:
I think we need more tests. the coverage is not great here

xiaoyu74:
> 
@facchettos - Thanks Jeremy for your quick review, yes, I will add more test cases, just update the commit to test the CI failing issue.

pkg/operator/controllers/checker/unsupportedconditionchecker.go
```diff
@@ -0,0 +1,153 @@
...
+
+	ucc.log.Infof("Set the %q operator upgradeable status to %q", clusterOperatorName, upgradeableStatus)
+
+	return nil
+}
+
+func (ucc *UnsupportedConditionChecker) addConsoleBanner(ctx context.Context, consoleText string) error {
+	notification, err := ucc.consolecli.ConsoleV1().ConsoleNotifications().Get(ctx, consoleBannerName, metav1.GetOptions{})
+	if err != nil {
+		if !kerrors.IsNotFound(err) {
```
s-amann, line 122: Should we be creating the banner when it is not found and not when it is found (with error)?

Right now this reads as "get the banner, if we got an error and that error is not that we couldn't find the banner, create the banner". 

Instead of "get the banner, if we got an error and that error is that we could not find the banner, create the banner"

s-amann: Alternatively, if there is a way to list the banners it might make it more clear when we need to create vs update.

xiaoyu74: @s-amann - Thanks for your time to review and I apologize for my late response. Let me do some more testing and back to your questions then.

xiaoyu74: > Right now this reads as "get the banner, if we got an error and that error is not that we couldn't find the banner, create the banner".

The main logic of this part was referred from`banner.go` which I thought it's the way of convention to `get and create banner` and I just refactored a bit in my case.

xiaoyu74: > Alternatively, if there is a way to list the banners it might make it more clear when we need to create vs update.

The banner name of this checker is hard-coded as a fixed name, I am not sure why it should be listed ?  Please let me know if anything I misunderstood.

s-amann: Listing the banners is a misunderstanding on my part then, I didn't realize there was only one banner. My initial thinking was if we 'list banners' and the result is `[]` then we know we need to create one, but if it came back `[banner-we-want]` then we would know to update. Please disregard this idea if there is only ever one banner and we can determine its name ahead of time.

pkg/operator/controllers/checker/unsupportedconditionchecker.go
```diff
@@ -0,0 +1,153 @@
...
+			_, err = ucc.consolecli.ConsoleV1().ConsoleNotifications().Create(ctx, newBanner, metav1.CreateOptions{})
+			if err != nil {
+				ucc.log.Errorf("Failed to create the %q banner`: %q", consoleBannerName, err.Error())
+				return err
+			}
+			ucc.log.Infof("Created the %q banner, with text %q", consoleBannerName, consoleText)
+			return nil
+		}
+		return err
+	}
```
facchettos, line 143: the flow would be a little easier to follow if we did it this way
```suggestion
        if kerrors.IsNotFound(err) {
                return err
        }
	if err != nil {
			newBanner := &consolev1.ConsoleNotification{
				ObjectMeta: metav1.ObjectMeta{
					Name: consoleBannerName,
				},
				Spec: consolev1.ConsoleNotificationSpec{
					Text:            consoleText,
					Location:        consolev1.BannerTop,
					Color:           "#000",
					BackgroundColor: "#ff0",
				},
			}
			_, err = ucc.consolecli.ConsoleV1().ConsoleNotifications().Create(ctx, newBanner, metav1.CreateOptions{})
			if err != nil {
				ucc.log.Errorf("Failed to create the %q banner`: %q", consoleBannerName, err.Error())
				return err
			}
			ucc.log.Infof("Created the %q banner, with text %q", consoleBannerName, consoleText)
			return nil
	}
```

facchettos (changes requested):
A little flow simplification and then it's good for me :rocket:

pkg/operator/controllers/checker/unsupportedconditionchecker.go
facchettos: argh github made it look horrible, basically just check if the error is not found first so we can avoid the if nesting and all the negatives

pkg/operator/controllers/checkers/internetchecker/unsupportconditionchecker.go
```diff
@@ -0,0 +1,154 @@
...
+			},
+		}
+		_, err = ucc.consolecli.ConsoleV1().ConsoleNotifications().Create(ctx, newBanner, metav1.CreateOptions{})
+		if err != nil {
+			ucc.log.Errorf("Failed to create the %q banner`: %q", consoleBannerName, err.Error())
+			return err
+		}
+		ucc.log.Infof("Created the %q banner, with text %q", consoleBannerName, consoleText)
+		return nil
+	}
```
facchettos, line 144: With the flow now simplified I think I might have misunderstood what it was doing in my previous review. What are we trying to achieve here? Shouldn't it be
```
If notFound -> create it 
Else if err != nil (actual error) -> return the error 
If err == nil (so object was found) -> update it 
```
?

facchettos (changes requested):
I think I missed something during my last review :( 
I think there is a little error in your creation logic

pkg/operator/controllers/checkers/internetchecker/unsupportconditionchecker.go
xiaoyu74: @facchettos - Sorry for my late response due to variety of leaves in the past 2 weeks.

Yeah, the main logic you mentioned above was what I was going to achieve for handling the banner creation/update. 
```
if err != nil (actual error) && !kerrors.IsNotFound(err)  -> return the error 
If err != nil && kerrors.IsNotFound(err)  -> create it 
If err == nil (so object was found) -> update it 
```

Actually, I referred an existing example in `banner.go` for the banner handling below:
```
case arov1alpha1.BannerContactSupport:
	banner := r.newBanner(fmt.Sprintf(TextContactSupport, instance.Spec.ResourceID))
	_, err := r.consolecli.ConsoleV1().ConsoleNotifications().Get(ctx, BannerName, metav1.GetOptions{})
	if err != nil && !kerrors.IsNotFound(err) {
	    return err
	}

	// if the object doesn't exist Create
	if err != nil && kerrors.IsNotFound(err) {
	     _, err = r.consolecli.ConsoleV1().ConsoleNotifications().Create(ctx, banner, metav1.CreateOptions{})
	    return err
	}

	// if there's no errors, object found then update
	_, err = r.consolecli.ConsoleV1().ConsoleNotifications().Update(ctx, banner, metav1.UpdateOptions{})
	     return err
	}
```

And I think it will be more clear to refactor as below: 
```
notification, err := ucc.consolecli.ConsoleV1().ConsoleNotifications().Get(ctx, consoleBannerName, metav1.GetOptions{})
	if err != nil {
		if kerrors.IsNotFound(err) {
			newBanner := &consolev1.ConsoleNotification{
					Name: consoleBannerName,
				},
				Spec: consolev1.ConsoleNotificationSpec{
					Text:            consoleText,
					Location:        consolev1.BannerTop,
					Color:           "#000",
					BackgroundColor: "#ff0",
				},
			}
			_, err = ucc.consolecli.ConsoleV1().ConsoleNotifications().Create(ctx, newBanner, metav1.CreateOptions{})
			if err != nil {
				ucc.log.Errorf("Failed to create the %q banner`: %q", consoleBannerName, err.Error())
				return err
			}
			ucc.log.Infof("Created the %q banner, with text %q", consoleBannerName, consoleText)
			return nil
		}
		return err
	}

       notification.Spec.Text = consoleText
	_, err = ucc.consolecli.ConsoleV1().ConsoleNotifications().Update(ctx, notification, metav1.UpdateOptions{})
	if err != nil {
		ucc.log.Errorf("Failed to update the %q banner`: %q", consoleBannerName, err.Error())
		return err
	}
	ucc.log.Infof("Updated the %q banner, with combined text %q", consoleBannerName, consoleText)
	return nil
```

facchettos: I think that is fine the way it is now, we just need to do this 
```suggestion
	notification, err := ucc.consolecli.ConsoleV1().ConsoleNotifications().Get(ctx, consoleBannerName, metav1.GetOptions{})
	if kerrors.IsNotFound(err) {
		newBanner := &consolev1.ConsoleNotification{
			ObjectMeta: metav1.ObjectMeta{
				Name: consoleBannerName,
			},
			Spec: consolev1.ConsoleNotificationSpec{
				Text:            consoleText,
				Location:        consolev1.BannerTop,
				Color:           "#000",
				BackgroundColor: "#ff0",
			},
		}
		_, err = ucc.consolecli.ConsoleV1().ConsoleNotifications().Create(ctx, newBanner, metav1.CreateOptions{})
		if err != nil {
			ucc.log.Errorf("Failed to create the %q banner`: %q", consoleBannerName, err.Error())
			return err
		}
		ucc.log.Infof("Created the %q banner, with text %q", consoleBannerName, consoleText)
		return nil
	}

	if err != nil {
		return err
	}
```

facchettos: The reason behind it is that it's easier to follow when you have less nesting

facchettos:
there might be a `go.mod` , or `go.sum`, or `vendor.txt` missing from your commit

xiaoyu74:
> there might be a `go.mod` , or `go.sum`, or `vendor.txt` missing from your commit
@facchettos - Thanks for your feedback, I think I got the same issue discussed here [go vendoring errors](https://redhat-internal.slack.com/archives/C02ULBRS68M/p1675868626299999) and it has been fixed.

facchettos (approved):
:rocket:

pkg/operator/controllers/checkers/internetchecker/unsupportedconditionchecker.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	ucc.log.Infof("Set the %q operator upgradeable status to %q", clusterOperatorName, upgradeableStatus)
+
+	return nil
+}
+
+func (ucc *UnsupportedConditionChecker) addConsoleBanner(ctx context.Context, consoleText string) error {
+	notification, err := ucc.consolecli.ConsoleV1().ConsoleNotifications().Get(ctx, consoleBannerName, metav1.GetOptions{})
+	if kerrors.IsNotFound(err) {
```
s-amann, line 121: this control flow is clean and easy to understand, thank you :rocket:

s-amann (approved):

PR #2167, opened 2022-06-09
Fix azureproviderspec regression

s-amann:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14682300

Fix on a regression for azureproviderspec not utilizing the new way to decode/marshal the contents.  

### Test plan for issue:

Ran PUCM on 4.9 cluster with older version.  

No

### DIFF
--- a/pkg/util/machine/azureproviderspec.go
+++ b/pkg/util/machine/azureproviderspec.go
@@ -1,0 +1,52 @@
+package machine
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	"k8s.io/client-go/kubernetes/scheme"
+
+	_ "github.com/Azure/ARO-RP/pkg/util/scheme"
+)
+
+type MachineType string
+
+const (
+	MachineSet MachineType = "machineset"
+	Machine    MachineType = "machine"
+)
+
+// UnmarshalAzureProviderSpec unmarshals an Azure Provider Spec set on machines or machinesets
+// it contains backward compatibility for the older azureproviderconfig which is no longer in use
+// in OCP 4.10+
+func UnmarshalAzureProviderSpec(name string, mType MachineType, rawProviderSpec []byte) (*machinev1beta1.AzureMachineProviderSpec, error) {
+	var machineProviderSpec *machinev1beta1.AzureMachineProviderSpec
+
+	if strings.Contains(string(rawProviderSpec), "azureproviderconfig.openshift.io") {
+		machineProviderSpec = &machinev1beta1.AzureMachineProviderSpec{}
+		err := json.Unmarshal(rawProviderSpec, machineProviderSpec)
+		if err != nil {
+			return machineProviderSpec, fmt.Errorf("%s %s: failed to unmarshal the 'azureproviderconfig.openshift.io' provider spec: %q", mType, name, err.Error())
+		}
+	} else {
+		o, _, err := scheme.Codecs.UniversalDeserializer().Decode(rawProviderSpec, nil, nil)
+		if err != nil {
+			return machineProviderSpec, err
+		}
+
+		var ok bool
+		machineProviderSpec, ok = o.(*machinev1beta1.AzureMachineProviderSpec)
+		if !ok {
+			// If this happens, the azure machine provider spec type/apiversion may have been updated and
+			// we need to handle it appropriately
+			return machineProviderSpec, fmt.Errorf("%s %s: failed to read provider spec: %T", mType, name, o)
+		}
+	}
+
+	return machineProviderSpec, nil
+}
--- a/pkg/util/machine/azureproviderspec.go
+++ b/pkg/util/machine/azureproviderspec.go
@@ -17,35 +17,34 @@ import (
 type MachineType string
 
 const (
-	MachineSet MachineType = "machineset"
-	Machine    MachineType = "machine"
+	MachineSet        MachineType = "machineset"
+	Machine           MachineType = "machine"
+	azureProviderSpec             = "azureproviderconfig.openshift.io"
 )
 
 // UnmarshalAzureProviderSpec unmarshals an Azure Provider Spec set on machines or machinesets
 // it contains backward compatibility for the older azureproviderconfig which is no longer in use
 // in OCP 4.10+
 func UnmarshalAzureProviderSpec(name string, mType MachineType, rawProviderSpec []byte) (*machinev1beta1.AzureMachineProviderSpec, error) {
-	var machineProviderSpec *machinev1beta1.AzureMachineProviderSpec
-
-	if strings.Contains(string(rawProviderSpec), "azureproviderconfig.openshift.io") {
-		machineProviderSpec = &machinev1beta1.AzureMachineProviderSpec{}
+	if strings.Contains(string(rawProviderSpec), azureProviderSpec) {
+		machineProviderSpec := &machinev1beta1.AzureMachineProviderSpec{}
 		err := json.Unmarshal(rawProviderSpec, machineProviderSpec)
 		if err != nil {
-			return machineProviderSpec, fmt.Errorf("%s %s: failed to unmarshal the 'azureproviderconfig.openshift.io' provider spec: %q", mType, name, err.Error())
+			return nil, fmt.Errorf("%s %s: failed to unmarshal the %q provider spec: %q", mType, name, azureProviderSpec, err.Error())
 		}
-	} else {
-		o, _, err := scheme.Codecs.UniversalDeserializer().Decode(rawProviderSpec, nil, nil)
-		if err != nil {
-			return machineProviderSpec, err
-		}
+		return machineProviderSpec, nil
+	}
 
-		var ok bool
-		machineProviderSpec, ok = o.(*machinev1beta1.AzureMachineProviderSpec)
-		if !ok {
-			// If this happens, the azure machine provider spec type/apiversion may have been updated and
-			// we need to handle it appropriately
-			return machineProviderSpec, fmt.Errorf("%s %s: failed to read provider spec: %T", mType, name, o)
-		}
+	o, _, err := scheme.Codecs.UniversalDeserializer().Decode(rawProviderSpec, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	machineProviderSpec, ok := o.(*machinev1beta1.AzureMachineProviderSpec)
+	if !ok {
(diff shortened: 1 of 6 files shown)


pkg/util/machine/azureproviderspec_test.go
```diff
@@ -0,0 +1,152 @@
...
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			actualProviderSpec, err := UnmarshalAzureProviderSpec(machineName, tt.machineType, tt.providerSpec)
+			if err == nil {
+				if tt.wantErr != "" {
+					t.Error(err)
+				}
+
+				// Don't care about type meta because it should retain it
```
karanmagdani, line 138: Minor. Maybe this should be something like:

// Ignoring check for support for both azureproviderconfig and machine providerSpec Types

pkg/util/machine/azureproviderspec_test.go
```diff
@@ -0,0 +1,152 @@
...
+		t.Fatal(err)
+	}
+
+	return buf.Bytes()
+}
+
+func TestUnmarshalAzureProviderSpec(t *testing.T) {
+	machineName := "myMachine"
+
+	// Register old type soley so we can encode the CR to bytes for use only in test
```
karanmagdani, line 92: minor typo, so we*

pkg/util/machine/azureproviderspec_test.go
karanmagdani, line 138: Minor. Maybe this should be something like:

// Ignoring check for support for both azureproviderconfig and machine providerSpec Types

karanmagdani (changes requested):
LGTM in general, minor typo and comment clarification

pkg/util/machine/azureproviderspec_test.go
s-amann: updated, thank you!

bennerv: fixed

pkg/util/machine/azureproviderspec_test.go
bennerv: ty - fixed

karanmagdani (approved):
LGTM

cadenmarchese (approved):

darthhexx (approved):
lgtm

pkg/util/machine/azureproviderspec.go
```diff
@@ -0,0 +1,52 @@
...
+		var ok bool
+		machineProviderSpec, ok = o.(*machinev1beta1.AzureMachineProviderSpec)
+		if !ok {
+			// If this happens, the azure machine provider spec type/apiversion may have been updated and
+			// we need to handle it appropriately
+			return machineProviderSpec, fmt.Errorf("%s %s: failed to read provider spec: %T", mType, name, o)
+		}
+	}
+
+	return machineProviderSpec, nil
```
facchettos, line 51: Couldn't we get rid of the `var machineProviderSpec` and instantiate it inside the `if` ? That would also get rid of the `var ok bool` since we could just use `:=` on line 43. It would require to put the return inside the branches, but I think that's a little better. 

Acutally, what could make it even clearer would be to have functions for each use case: something like `unmarshalLegacy` and `unmarshalAMPS` which we could just return. That would make it clearer which is the legacy and which is the current

darthhexx: This is a flow starts with a variable that we want to return and ultimately returns that variable with the logic required in between. I can see the use case for 2 distinct functions, but the logic is not verbose and easy to follow so I think there are diminishing returns by refactoring them into individual functions.

facchettos: it is a little confusing though, if you look at line 39, we return `machineProviderSpec` because it has been defined before the if statement, but what we really want to return is `nil`, which is what the variable is 100% of the time. It would not happen if we returned straight from the branches. 

If we had something like this, it would be clearer. 
```go 
if something {
    return resultForSomething
} else {
    return defaultResult
}
``` 

But we should at least have a little comment in the if branches to say which one is the legacy and which one is the current

darthhexx: I do agree that we should be returning an explicit `nil` in those error blocks to be clearer of the return value.

facchettos: btw, I we could even get rid of the else. Unnecessary elses are usually avoided (cf effective go + uber go style guide)
```suggestion
	var machineProviderSpec *machinev1beta1.AzureMachineProviderSpec

	if strings.Contains(string(rawProviderSpec), "azureproviderconfig.openshift.io") {
		machineProviderSpec = &machinev1beta1.AzureMachineProviderSpec{}
		err := json.Unmarshal(rawProviderSpec, machineProviderSpec)
		if err != nil {
			return nil, fmt.Errorf("%s %s: failed to unmarshal the 'azureproviderconfig.openshift.io' provider spec: %q", mType, name, err.Error())
		}
		return  machineProviderSpec, nil
	} 
	
	o, _, err := scheme.Codecs.UniversalDeserializer().Decode(rawProviderSpec, nil, nil)
	if err != nil {
		return nil, err
	}

	var ok bool
	machineProviderSpec, ok = o.(*machinev1beta1.AzureMachineProviderSpec)
	if !ok {
		// If this happens, the azure machine provider spec type/apiversion may have been updated and
		// we need to handle it appropriately
		return nil, fmt.Errorf("%s %s: failed to read provider spec: %T", mType, name, o)
	}
	

	return machineProviderSpec, nil
```

facchettos (dismissed):
See conversation above

pkg/util/machine/azureproviderspec.go
```diff
@@ -0,0 +1,52 @@
...
+	Machine    MachineType = "machine"
+)
+
+// UnmarshalAzureProviderSpec unmarshals an Azure Provider Spec set on machines or machinesets
+// it contains backward compatibility for the older azureproviderconfig which is no longer in use
+// in OCP 4.10+
+func UnmarshalAzureProviderSpec(name string, mType MachineType, rawProviderSpec []byte) (*machinev1beta1.AzureMachineProviderSpec, error) {
+	var machineProviderSpec *machinev1beta1.AzureMachineProviderSpec
+
+	if strings.Contains(string(rawProviderSpec), "azureproviderconfig.openshift.io") {
```
AldoFusterTurpin, line 30: I would extract 'azureproviderconfig.openshift.io' into a constant as it is used in two places (here and in line 34).

AldoFusterTurpin (commented):
Some generals comments. 

I would not reject this PR as I don't see any red flag but I see a yelllow one with the if/else discussed in the review.

gvanderpotte (approved):
LGTM

pkg/util/machine/azureproviderspec_test.go
```diff
@@ -0,0 +1,152 @@
...
+	return buf.Bytes()
+}
+
+func TestUnmarshalAzureProviderSpec(t *testing.T) {
+	machineName := "myMachine"
+
+	// Register old type so we can encode the CR to bytes for use only in test
+	scheme.Scheme.AddKnownTypes(azureproviderv1beta1.SchemeGroupVersion, &azureproviderv1beta1.AzureMachineProviderSpec{})
+
+	machineProviderSpecBytes := marshalAzureMachineProviderSpec(t, machineAzureProviderSpec)
```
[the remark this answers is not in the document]
s-amann: I'm going to leave the instantiation of these objects outside of the method, IMO it is easy to compare them on their own and debloats the function

pkg/util/machine/azureproviderspec_test.go
```diff
@@ -0,0 +1,152 @@
...
+	azProviderSpecBytes := marshalAzureMachineProviderSpec(t, azAzureProviderSpec)
+
+	for _, tt := range []struct {
```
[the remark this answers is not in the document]
s-amann: I agree it is cleaner to have the separate struct, I'll push an update soon.

pkg/util/machine/azureproviderspec.go
s-amann: Thanks for pointing out the simplification, I'll remove the extraneous else and update to explicitly return nil when appropriate.

pkg/util/machine/azureproviderspec_test.go
```diff
@@ -0,0 +1,152 @@
...
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	kjson "k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/client-go/kubernetes/scheme"
+	azureproviderv1beta1 "sigs.k8s.io/cluster-api-provider-azure/pkg/apis/azureprovider/v1beta1"
+)
+
+var machineAzureProviderSpec = &machinev1beta1.AzureMachineProviderSpec{
```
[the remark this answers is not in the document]
s-amann: It is both an expected value and input to the test, I am going to leave the naming as is

pkg/util/machine/azureproviderspec.go
```diff
@@ -0,0 +1,54 @@
...
+	MachineSet        MachineType = "machineset"
+	Machine           MachineType = "machine"
+	azureProviderSpec             = "azureproviderconfig.openshift.io"
+)
+
+// UnmarshalAzureProviderSpec unmarshals an Azure Provider Spec set on machines or machinesets
+// it contains backward compatibility for the older azureproviderconfig which is no longer in use
+// in OCP 4.10+
+func UnmarshalAzureProviderSpec(name string, mType MachineType, rawProviderSpec []byte) (*machinev1beta1.AzureMachineProviderSpec, error) {
+	var machineProviderSpec *machinev1beta1.AzureMachineProviderSpec
```
facchettos, line 29: this is still superfluous

pkg/util/machine/azureproviderspec.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+		return machineProviderSpec, nil
+	}
+
+	o, _, err := scheme.Codecs.UniversalDeserializer().Decode(rawProviderSpec, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var ok bool
```
facchettos, line 45: this too

facchettos (dismissed):
I still think that this is not useful to have the instantiation done this way

pkg/util/machine/azureproviderspec.go
s-amann: fixed, ty

ross-bryan (approved):
this pr got good feedback and lgtm

facchettos (approved):
LGTM

25region (approved):
LGTM

PR #2170, opened 2022-06-10
Create DNS record for the AKS API during provisioning

darthhexx:

AKS generates a random FQDN for the cluster API and, although we can interrogate for it, we should have a common naming convention. 

Provides a known URL pattern to access AKS in the shared dev environments, namely: api.aks.{australiaeast,eastus,westeurope}.osadev.cloud

### Test plan for issue:

The ARM template has already been deployed to the shared environments

No.

### DIFF
--- a/hack/devtools/deploy-shared-env.sh
+++ b/hack/devtools/deploy-shared-env.sh
@@ -76,6 +76,7 @@ deploy_aks_dev() {
         --template-file pkg/deploy/assets/aks-development.json \
         --parameters \
             "adminObjectId=$ADMIN_OBJECT_ID" \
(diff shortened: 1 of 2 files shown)


s-amann (dismissed):
lgtm

SudoBrendan (dismissed):
LGTM

facchettos (changes requested):
It would be great to get rid of bindata here and use embed instead

darthhexx:
@facchettos bindata is gone :smile:

hawkowl (approved):

facchettos (approved):

PR #2172, opened 2022-06-10
Removed temporary timestamp update code

bizz001:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/9128557

Removes temporary code responsable of populating `createdAt` field for clusters created before this field was added.

### Test plan for issue:

Functionality removed not added. Removed corresponding unit test.

No. temporary code cleanup.


bennerv (dismissed):
lgtm - looks like we're all good on clusters according to story notes as well.

facchettos (approved):
LGTM

petrkotas (approved):

PR #2173, opened 2022-06-10
Add pull secret to allow mirroring from pd -> int

bennerv:

Fixes issues running ARO Mirror to INT pipeline that is automagically triggered on building a tagged image

Pipeline doesn't have creds to pull prod image, but now does :) 

### Test plan for issue:

Test it once by hand, then revert latest commit and celebrate (assuming it works)

Nope


.pipelines/mirror-aro-to-int.yml
```diff
@@ -37,8 +37,14 @@ jobs:
       ## set the variable
       echo "##vso[task.setvariable variable=TAG]${TAG}"
   - script: |
+      USERNAME=`echo "$(aro-v4-ci-pd-pull)" | base64 -d | cut -d':' -f1`
+      PASSWORD=`echo "$(aro-v4-ci-pd-pull)" | base64 -d | cut -d':' -f2-`
+
       az acr login --name "${{variables.rpIntImageAcr}}"
       az acr import \
+        --force \
```
bennerv, line 45: Why are you adding a force tag @bennerv ? Well great question.

Our INT e2e runs on master merge. What that does is build the image off of master and deploy to INT. If you tag the repo with a release tag v20220610 prior to it building the image, it will push a versioned image to INT instead of the short commit hash.

What that means is that the INT image may differ just slightly from the production image built during the release. So in order to ensure the images are the exact same, force push over the previous tag. This force push only happens in INT so it cannot affect existing images.

bennerv:
Confirmed fixed now that we're adding in a pull secret to pull the prod image to mirror to INT. 

See _slightly modified_, but successful run here: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=56397488&view=logs&j=85d92508-9606-5d9f-aae1-82b2f0560018&t=85d92508-9606-5d9f-aae1-82b2f0560018

ross-bryan (approved):
lgtm

s-amann (approved):

PR #2175, opened 2022-06-13
Monitor/detect clusters not registered in hive

UlrichSchlueter:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14992346

Monitoring for the hive situation of ARO clusters

Emits hive.clusterdeployment.conditions  metrics for a subset of the clusterdeployment conditions for any monitored cluster
    type: Ready.   (looks as nodes and clusteroperators)
    type: SyncSetFailed.  (shows if all applicable syncsets can be applied successfully)
    type: Unreachable.     (shows if the cluster api is accessible , network and authentication wise)

The metric has a type and reason dimension that can be used to differentiate between the different cases. 

For reference, the full list of hive clusterdeployment conditions for future (M5 and later) extensions
   type: ActiveAPIURLOverride
    type: Hibernating
    type: ControlPlaneCertificateNotFound
    type: Provisioned
    type: Ready
    type: RelocationFailed
    type: SyncSetFailed
    type: Unreachable
    type: AWSPrivateLinkFailed
    type: AWSPrivateLinkReady
    type: AuthenticationFailure
    type: ClusterInstallCompleted
    type: ClusterInstallFailed
    type: ClusterInstallRequirementsMet
    type: ClusterInstallStopped
    type: DNSNotReady
    type: DeprovisionLaunchError
    type: InstallImagesNotResolved
    type: InstallLaunchError
    type: InstallerImageResolutionFailed
    type: ProvisionFailed
    type: ProvisionStopped
    type: RequirementsMet
 
### Test plan for issue:

manually tested with several clusters for proper behaviour with and without hive connectivity, with and without registrations and finally that data arrives is Jarvis and can be processed 
unit tests to follow

no


pkg/monitor/cluster/hiveregistrationstatus.go
```diff
@@ -0,0 +1,90 @@
...
+		})
+	}
+
+	return nil
+}
+
+func (mon *Monitor) hiveRegistrationState(ctx context.Context) (hiveRegistrationState, error) {
+	state := hiveRegistrationState{
+		isRegistered:        false,
+		isUnreachable:       false,
```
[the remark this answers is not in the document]
ross-bryan: I'd also rather default to an empty/bad state, and only change to `isReachable = true` when we've communicated with hive the correct conditions

pkg/monitor/cluster/hiveregistrationstatus.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	state.isRegistered = true
+
+	isConnected, reason, err := hr.IsConnected(ctx)
+	if err != nil {
+
+		return state, err
+	}
+	if !isConnected {
```
ross-bryan, line 85: per my suggestion we should then check for the true value of this bool
```suggestion
        if isConnected {
          state.Reachable = true
        }

Comment (ross-bryan) [reply pkg/monitor/cluster/hiveregistrationstatus.go:85]: and then you'd need to properly set the notReachableReason in the false case

**Review by ross-bryan — CHANGES_REQUESTED** [review]
a couple of comments/suggestions. looks like its on the right path

**Comment by m1kola** [comment]
#2145 was merged. I rebased this PR on top of master and fixed conflicts. Need review now.

File: pkg/monitor/cluster/hiveregistrationstatus.go (hunk 1 above)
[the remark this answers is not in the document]
Comment (UlrichSchlueter) [reply pkg/monitor/cluster/hiveregistrationstatus.go]: renamed accordingly. This double negative logic is confusing .
Also changed the default to bad

File: pkg/monitor/cluster/hiveregistrationstatus.go [hunk 3]
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return err
+	}
+	mon.log.Info(state)
+
+	if !state.isRegistered {
+		mon.emitGauge("hive.registration.error", 1, map[string]string{
+			"reason": state.notRegisteredReason,
+		})
+	}
```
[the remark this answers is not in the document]
UlrichSchlueter: Dropped this metric, but added a case for the namespace value empty or not present. This prevents the monitoring log from getting flooded with errrors for unregistered clusters

pkg/monitor/cluster/hiveregistrationstatus.go
```diff
@@ -0,0 +1,80 @@
...
+
+	if state.isUnreachable {
+		mon.emitGauge("hive.registration.isUnreachable", 1, map[string]string{
+			"reason": state.notReachableReason,
+		})
+	}
```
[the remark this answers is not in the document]
UlrichSchlueter: Not implemented, as discussed

pkg/monitor/cluster/hiveregistrationstatus.go
```diff
@@ -0,0 +1,60 @@
...
+		return err
+	}
+
+	if !state.isReachable {
+		mon.emitGauge("hive.registration.isUnreachable", 1, map[string]string{
+			"reason": state.notReachableReason,
+		})
+	}
+
+	return nil
```
m1kola, line 33: I still think that we should do `ClusterDeployment` conditions instead of this.

I realised that my [previous comments](https://github.com/Azure/ARO-RP/pull/2175#discussion_r915691248) we were missleading: I wanted to refer to conditions, not statuses. And a better example would be [`emitClusterVersionConditions`](https://github.com/Azure/ARO-RP/blob/3ed9676d3672283c4c94d07050da0d805da83634/pkg/monitor/cluster/clusterversionconditions.go) instead of deployment statuses. We have few other exampels such as `emitNodeConditions`, `emitPodConditions`, etc but they include more things which are irrelevant for cluster deployment statuses.

Cluster deployments have a lot of condiditions and we do not care about all of them.

m1kola (changes requested):
I think we need to get rid of [`IsConnected`](https://github.com/Azure/ARO-RP/blob/3ed9676d3672283c4c94d07050da0d805da83634/pkg/hive/manager.go#L29). It is only being used in the monitor at the moment and we will likely have to rely on some other condition for installation in later milestones.

The above is business for a separate PR. But given feedback around cluster deployment conditions, I think we need to modify the monitor in the following way:
* Make it receive `hiveRestConfig` and create `hiveClientset` (we don't need whole hive cluster manager)
* Make it monitors a list of cluster deployment conditions
    * For now we will only need `hivev1.UnreachableCondition`, but we will likely expand the list.

m1kola:
> I think we need to get rid of [`IsConnected`](https://github.com/Azure/ARO-RP/blob/3ed9676d3672283c4c94d07050da0d805da83634/pkg/hive/manager.go#L29). It is only being used in the monitor at the moment and we will likely have to rely on some other condition for installation in later milestones.
> 
> The above is business for a separate PR.

I created https://github.com/Azure/ARO-RP/pull/2261 for this.

m1kola (commented):
Looks good overall. Just a few new comments + [this one](https://github.com/Azure/ARO-RP/pull/2175#discussion_r924339683) is still valid, I think.

pkg/monitor/cluster/hiveregistrationstatus.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (mon *Monitor) emitHiveRegistrationStatus(ctx context.Context) error {
+	if mon.hiveclientset == nil {
+		// TODO(hive): remove this if once we have Hive everywhere
+		mon.log.Info("skipping: no hive cluster manager")
+		return nil
+	}
+
+	if mon.oc.Properties.HiveProfile.Namespace == "" {
+		return fmt.Errorf("cluster %s not adopted. No namespace in the clusterdocument", mon.oc.Name)
```
m1kola, line 31: Not asking to change. Just a tip. It is not necessary to have name in the error message: it will be logged and each log line includes everything we need to identify the cluster:

https://github.com/Azure/ARO-RP/blob/98fc429ecefc51d669cc72dc7cd3963231c18f1c/pkg/util/log/log.go#L156-L161

pkg/monitor/cluster/hiveregistrationstatus.go
```diff
@@ -0,0 +1,50 @@
...
+
+	cd, err := mon.hiveclientset.HiveV1().ClusterDeployments(mon.oc.Properties.HiveProfile.Namespace).Get(ctx, hive.ClusterDeploymentName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, condition := range cd.Status.Conditions {
+		if expectedState, ok := clusterDeploymentConditionsExpected[condition.Type]; ok {
+			if condition.Status != expectedState {
+				mon.emitGauge("hive.conditions", 1, map[string]string{
```
m1kola, line 42: I think that `hive.conditions` is too generic. Maybe something like `hive.clusterdeployment.conditions`?

pkg/monitor/cluster/hiveregistrationstatus.go
```diff
@@ -0,0 +1,50 @@
...
+					"type":   string(condition.Type),
+					"reason": condition.Reason,
```
m1kola, line 44: I'm not sure about `condition.Reason` here: there are probably too many reasons/variations.

We might want to consider replacing it with `condition.Status` (it is either `True` or `False` or `Unknown`) or droping altogether.

m1kola (changes requested):
Looks good. Two more small things:

* I think we should rename `hive.conditions`
* And I'm not sure whether we should add `condition.Reason` into dimensions or not.

pkg/monitor/cluster/hiveregistrationstatus.go
UlrichSchlueter: From what I can see the state alone is does not provide enough information for monitoring logic. For example: at adoption time the Ready condition goes false and then goes through several sub statuses (indicated by the reason), while hive is checking if VMs and COs are up, before it goes to true eventually. The combination reason and state would be required to determine if we had a real issue with the adoption of a cluster or if it's 'just' hive making up its mind. I'd like to keep that data, if possible.

pkg/monitor/cluster/hiveregistrationstatus.go
UlrichSchlueter: Updated accordingly

pkg/monitor/cluster/hiveregistrationstatus.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	if mon.oc.Properties.HiveProfile.Namespace == "" {
+		mon.emitGauge("hive.condition.NoNamespace", 1, map[string]string{
+			"reason": "NoNamespaceInClusterDocument",
+		})
+		return nil
+	}
+
+	cd, err := mon.hiveClusterManager.ClusterDeployment(ctx, mon.oc.Properties.HiveProfile.Namespace)
```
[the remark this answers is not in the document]
UlrichSchlueter: Dropped hiveClusterManager and made exported the clusterdeployment constant instead.

m1kola (approved):
Looks good

@UlrichSchlueter could you please do few things?

1. Create a work item under [14319539](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14319539) to create a dashboard from the metric
2. Create a follow-up PR with unit tests for this. I think this part is solid already: most likely we will not change it significantly.

pkg/monitor/cluster/hiveregistrationstatus.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+var clusterDeploymentConditionsExpected = map[hivev1.ClusterDeploymentConditionType]corev1.ConditionStatus{
+	hivev1.ClusterReadyCondition:  corev1.ConditionTrue,
+	hivev1.UnreachableCondition:   corev1.ConditionFalse,
+	hivev1.SyncSetFailedCondition: corev1.ConditionFalse,
+}
+
+func (mon *Monitor) emitHiveRegistrationStatus(ctx context.Context) error {
+	if interfaces.IsNil(mon.hiveclientset) {
```
m1kola, line 25: Why we do not see the same issue in `pkg/cluster/hive.go`?

pkg/monitor/cluster/cluster.go
```diff
@@ -86,20 +89,31 @@ func NewMonitor(ctx context.Context, log *logrus.Entry, restConfig *rest.Config,
 		return nil, err
 	}
 
+	var hiveclientset *hiveclient.Clientset
```
m1kola, line 92: This should probably be an interface. I also wondering if it is the casue of the issue with `nil` checks.

```suggestion
	var hiveclientset hiveclient.Interface
```

m1kola:
@UlrichSchlueter FYI: I was going trough the tickets and

> 1. Create a work item under [14319539](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14319539) to create a dashboard from the metric

I created [15057927](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15057927) for this.


> 2\. Create a follow-up PR with unit tests for this. I think this part is solid already: most likely we will not change it significantly.

And [15057937](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15057937) for tests.

pkg/monitor/cluster/cluster.go
UlrichSchlueter: Is an interface now.
Could go back to the initial version of the nil check.

m1kola (approved):

PR #2177, opened 2022-06-13
Add e2e tests for ImageConfig operator

dem4gus:

[ADO User Story 11051435](https://msazure.visualstudio.com/AzureRedHatOpenShift/_boards/board/t/Chainsaw/Stories/?workitem=11051435)

This PR adds end-to-end tests for the ImageConfig reconciler. The tests themselves are basically a duplication of the unit tests but in a live cluster. Deployment of images from allowed or blocked registries are not currently tested.

### Test plan for issue:

Run e2e tests.

### DIFF
--- a/test/e2e/operator.go
+++ b/test/e2e/operator.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -19,6 +20,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/ghodss/yaml"
+	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/ugorji/go/codec"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


karanmagdani (changes requested):
I think the missing link here is a wait mechanism for e2e. The reconciler won't instantly make the changes in most cases.

test/e2e/operator.go
```diff
@@ -419,3 +420,131 @@ var _ = Describe("ARO Operator - MHC Deployment", func() {
...
+	It("should have empty allow and block lists by default", func() {
+		instance, err := clients.AROClusters.AroV1alpha1().Clusters().Get(ctx, "cluster", metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		if !instance.Spec.OperatorFlags.GetSimpleBoolean(imageconfigFlag) {
+			Skip("ImageConfig Controller is not enabled, skipping test")
+		}
+
+		imageconfig, err := clients.ConfigClient.ConfigV1().Images().Get(ctx, "cluster", metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
```
s-amann, line 488: This is repeated at the start of each test, is it possible to move this to a `BeforeEach`?

s-amann (dismissed):
lgtm, nice work! I have one suggestion which is not critical.

test/e2e/operator.go
```diff
@@ -419,3 +420,131 @@ var _ = Describe("ARO Operator - MHC Deployment", func() {
...
+
+		imageconfig.Spec.RegistrySources.AllowedRegistries = nil
+		imageconfig.Spec.RegistrySources.BlockedRegistries = nil
+		expectedAllowlist = nil
+		expectedBlocklist = nil
+
+		_, err = clients.ConfigClient.ConfigV1().Images().Update(ctx, imageconfig, metav1.UpdateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		Eventually(verifyLists).WithTimeout(timeout).Should(BeTrue())
```
cadenmarchese, line 476: The testing framework would let you do something like `Eventually(Expect(actualImageConfig).To(BeEquivalentTo(expectedImageConfig))).WithTimeout(timeout)`. If it works, I think it might make more sense than the functions to assert that they match.

test/e2e/operator.go
```diff
@@ -419,3 +420,131 @@ var _ = Describe("ARO Operator - MHC Deployment", func() {
...
+	})
+
+	It("should add optional registries to the block list and skip required registries", func() {
+
+		// Check that Image Config is enabled
+		instance, err := clients.AROClusters.AroV1alpha1().Clusters().Get(ctx, "cluster", metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		if !instance.Spec.OperatorFlags.GetSimpleBoolean(imageconfigFlag) {
```
cadenmarchese, line 530: Could we do this check once at the beginning of the test spec rather than twice?

cadenmarchese (commented):
Nice work! A few initial thoughts:

test/e2e/operator.go
dem4gus: After looking over the [Ginkgo documentation](https://onsi.github.io/ginkgo/#skipping-specs) it does not appear there is any way to skip a container node (`Describe()` and `When()` blocks), only individual test nodes (`It()` and `Specify()`), and the skipping needs to be done inside the test block itself, not in a `Before()`.

dem4gus: After looking over the [Ginkgo documentation](https://onsi.github.io/ginkgo/#skipping-specs) it does not appear there is any way to skip a container node (`Describe()` and `When()` blocks), only individual test nodes (`It()` and `Specify()`), and the skipping needs to be done inside the test block itself, not in a `Before()`. It _could_ be moved to a `BeforeEach()` but that would still result in the flag being checked three times. It would simplify the code though, so it's a good idea.

test/e2e/operator.go
```diff
@@ -419,3 +420,131 @@ var _ = Describe("ARO Operator - MHC Deployment", func() {
...
+
+		for idx, entry := range b {
+			if a[idx] != entry {
+				return false
+			}
+		}
+		return true
+	}
+
+	verifyLists := func() (bool, error) {
```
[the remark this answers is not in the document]
dem4gus: As it turns out `Eventually()` has [another way it can be used](https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode) to make assertions inside the function and I was able to leverage that to pass in the current state.

test/e2e/operator.go
dem4gus: Configuration extracted.

s-amann (dismissed):

cadenmarchese:
The test looks good to me, I would just recommend the following:
- Remove the non-test-specific stuff (such as the return code changes) from this PR and move them into a separate PR
- Squash/re-do your commits so that there's just 1 or 2

Nice work!

karanmagdani (approved):

dem4gus:
Put some constants in a `const()` block to address some offline feedback.

s-amann (approved):

ross-bryan (approved):
lgtm

PR #2178, opened 2022-06-13
e2e to definitively determine a cluster is ARO

karanmagdani:
Signed-off-by: Karan <kmagdani@redhat.com>

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14105220/

This PR adds a new E2E test file. The idea of this file is to add tests that can definitely identify an ARO cluster. A failure here should mean that a cluster is not an ARO cluster. This is needed for clusters created by Hive

### Test plan for issue:

 
Not Applicable 

N/A

### DIFF
--- a/test/e2e/aro.go
+++ b/test/e2e/aro.go
@@ -1,0 +1,79 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+)
+
+const (
+	wSsh = "99-worker-ssh"
+	mSsh = "99-master-ssh"
+)
+
+// The high level idea of these tests is to definitevily say that a cluster is an ARO cluster.
+var _ = Describe("Verify Attributes of ARO Cluster", func() {
+	ctx := context.Background()
+	// acrDomainList should contain acrDomain verifier
+	acrDomainList := []string{"arointsvc.azurecr.io", "arointsvc.azurecr.us", "arosvc.azurecr.io", "arosvc.azurecr.us"}
+	azEnvironmentList := []string{azureclient.PublicCloud.Environment.Name, azureclient.USGovernmentCloud.Environment.Name}
+	It("should be possible to definitely confirm cluster is ARO", func() {
+		// Get cluster object
+		co, err := clients.AROClusters.AroV1alpha1().Clusters().Get(context.Background(), "cluster", metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		mcp, err := clients.MachineConfig.MachineconfigurationV1().MachineConfigPools().List(ctx, metav1.ListOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		// Expect 99-worker-ssh and 99-master-ssh to exist
+		masterSsh, workerSsh := false, false
+		for _, m := range mcp.Items {
+			for _, mc := range m.Spec.Configuration.Source {
+				if mc.Name == wSsh {
+					workerSsh = true
+				}
+				if mc.Name == mSsh {
+					masterSsh = true
+				}
+			}
+		}
+		By("Checking if workerSsh and masterSsh is enabled or disabled")
+		Expect(workerSsh).To(BeTrue())
+		Expect(masterSsh).To(BeTrue())
+
+		// Expect ACR Domain
+		By("Verifying AcrDomain exists and is a value we expect")
+		Expect(co.Spec.ACRDomain).NotTo(BeNil())
+		Expect(acrDomainList).To(ContainElement(co.Spec.ACRDomain))
+
+		// Verify AzEnvironment exists and is equal to cloud
+		By("Verifying AZEnvironment exists and is a value we expect")
+		Expect(co.Spec.AZEnvironment).NotTo(BeNil())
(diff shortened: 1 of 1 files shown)


test/e2e/aro.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/Azure/ARO-RP/pkg/api"
+)
+
+const (
+	wSsh = "99-worker-ssh"
+	mSsh = "99-master-ssh"
+)
+
+// The high level idea of these tests is to definitevily say that a cluster is an ARO cluster.
+var _ = Describe("Get ARO cluster", func() {
```
ross-bryan, line 26: Lets talk about this is a more behavioral approach. Lets use something like `Describe("Verify Attributes of ARO Cluster")

We can use `By` statements such that `By("Getting a list of MachineConfigs")

Followed by `It("should be possible to verify aro master and worker ssh machineconfigs exist")`

ross-bryan: kind of like this https://github.com/Azure/ARO-RP/blob/master/test/e2e/fips.go#L22-L40

karanmagdani: Thanks for the review @ross-bryan ! I'll make the updates you've suggested.

Do you have any feedback regarding the tests that are included? Should we put effort into adding any new tests?

ross-bryan: Looks like its off to a good start. Will cross reference the discussion in the design doc to the e2e code in a future review

test/e2e/aro.go
```diff
@@ -0,0 +1,69 @@
...
+				if mc.Name == mSsh {
+					masterSsh = true
+				}
+			}
+		}
+		Expect(workerSsh).To(BeTrue())
+		Expect(masterSsh).To(BeTrue())
+
+		// Expect ACR Domain
+		Expect(co.Spec.ACRDomain).To(BeEquivalentTo("arointsvc.azurecr.io"))
```
ross-bryan, line 53: This will only pass in CI. During rp releases we create prod clusters and this would have the prod ACR value

karanmagdani: just updated to cover these cases along with the suggestions you made above. Please rereview at your convenience

test/e2e/aro.go
```diff
@@ -0,0 +1,79 @@
...
+	wSsh = "99-worker-ssh"
+	mSsh = "99-master-ssh"
+)
+
+// The high level idea of these tests is to definitevily say that a cluster is an ARO cluster.
+var _ = Describe("Verify Attributes of ARO Cluster", func() {
+	ctx := context.Background()
+	// acrDomainList should contain acrDomain verifier
+	acrDomainList := []string{"arointsvc.azurecr.io", "arointsvc.azurecr.us", "arosvc.azurecr.io", "arosvc.azurecr.us"}
+	azEnvironmentList := []string{azureclient.PublicCloud.Environment.Name, azureclient.USGovernmentCloud.Environment.Name}
```
karanmagdani, line 31: This is added as such because in this e2e specifically, we don't really care if it is PublicCloud / GovCloud. We just want to definitively call it an ARO cluster. 

Also USGovCloud clusters have no plans to be created by hive at the moment, this is just for safeguarding in the future

test/e2e/aro.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+)
+
+const (
+	wSsh = "99-worker-ssh"
+	mSsh = "99-master-ssh"
+)
+
+// The high level idea of these tests is to definitevily say that a cluster is an ARO cluster.
```
s-amann, line 26: nit: typo
```suggestion
// The high level idea of these tests is to definitively say that a cluster is an ARO cluster.
```

test/e2e/aro.go
```diff
@@ -0,0 +1,79 @@
...
+var _ = Describe("Verify Attributes of ARO Cluster", func() {
+	ctx := context.Background()
+	// acrDomainList should contain acrDomain verifier
+	acrDomainList := []string{"arointsvc.azurecr.io", "arointsvc.azurecr.us", "arosvc.azurecr.io", "arosvc.azurecr.us"}
+	azEnvironmentList := []string{azureclient.PublicCloud.Environment.Name, azureclient.USGovernmentCloud.Environment.Name}
+	It("should be possible to definitely confirm cluster is ARO", func() {
```
s-amann, line 32: nit: not sure if you meant definitively here
```suggestion
	It("should be possible to definitively confirm cluster is ARO", func() {
```

s-amann (approved):
lgtm, just some nits

s-amann (approved):
lgtm

test/e2e/aro.go
```diff
@@ -0,0 +1,79 @@
...
+		By("Verifying AZEnvironment exists and is a value we expect")
+		Expect(co.Spec.AZEnvironment).NotTo(BeNil())
+		Expect(azEnvironmentList).To(ContainElement(co.Spec.AZEnvironment))
+
+		// Expect Geneva Logging to be present
+		By("Verifying GenevaLogging Exists")
+		Expect(co.Spec.GenevaLogging).NotTo(BeNil())
+
+		// Expect Clusters operator flags to be equivalent to default operator flags
+		By("Verifying OperatorFlags are set and equivalent to Latest defaults")
```
cadenmarchese, line 71: Should we also check the openshift-azure-operator namespace to ensure that both the master and worker operator deployments exist?

cadenmarchese (commented):
Test is looking good. Nice work!

test/e2e/aro.go
karanmagdani: Thank you for the review! I didn't add that to avoid duplicating existing tests https://github.com/Azure/ARO-RP/blob/master/test/e2e/operator.go#L38-L62

Considering the fact that if it fails, it will fail in that test already.

SudoBrendan (approved):
LGTM

cadenmarchese (approved):

PR #2181, opened 2022-06-14
Remove the flaky machineconfig e2e test

hawkowl:

Fixes some flakiness that's in the way of e2e passing reliably.

We have unit tests for this behaviour and the e2e test is excessively long and flaky for the benefit it provides.

### Test plan for issue:

E2E should pass.

N/A


test/e2e/operator.go
```diff
@@ -279,61 +278,6 @@ var _ = Describe("ARO Operator - Conditions", func() {
 	})
 })
 
-var _ = Describe("ARO Operator - MachineSet Controller", func() {
```
cadenmarchese, line 282: I agree that the test is adding a lot of time and e2e failures to our release process, but maybe it would be better to disable the test with `XDescribe` and create a WI to improve the test / change the test.

One idea would be a quick test that makes a MachineSet change (something non-destructive) and at least confirms that there was a reconcile. What do you think?

test/e2e/operator.go
```diff
@@ -279,61 +278,6 @@ var _ = Describe("ARO Operator - Conditions", func() {
...
-	Specify("operator should maintain at least two worker replicas", func() {
-		ctx := context.Background()
-
-		instance, err := clients.AROClusters.AroV1alpha1().Clusters().Get(ctx, "cluster", metav1.GetOptions{})
-		Expect(err).NotTo(HaveOccurred())
-
-		if !instance.Spec.OperatorFlags.GetSimpleBoolean(machineset.ControllerEnabled) {
```
cadenmarchese, line 289: If we pull this test out, we can remove [this](https://github.com/Azure/ARO-RP/blob/master/test/e2e/scalenodes.go#L70-L72) as well to re-enable the old one.

It's possible that billing e2e will fail if we don't scale at least once: https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/RP-Billing?path=/e2e/testsuites/billingservice/billingservice_test.go&version=GBmaster&line=228&lineEnd=233&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

test/e2e/operator.go
hawkowl: I don't think that testing that a machineset scales up and down is that useful for us as an end-to-end test. We already know that the permissions in the cluster are correct (because it scaled up to 3 nodes in the first place) and we don't really need to test that OpenShift is functioning as it should (that should be OCP's job).

cadenmarchese: Makes sense. In that case, I'd probably pull out the old test as well and adjust billing e2e if needed. This test was meant to test our own reconciler but agree we may be relying too much on Azure's ability to make new VMs quickly (or openshift-machine-api).

hawkowl: Good idea, I've deleted the old tests as well. I kept the check, since it is perhaps still just good for reasonableness checking.

test/e2e/nodes.go
```diff
@@ -0,0 +1,57 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/Azure/ARO-RP/pkg/util/ready"
+)
+
+const (
+	machineSetsNamespace = "openshift-machine-api"
+)
+
+var _ = Describe("Node check", func() {
```
cadenmarchese, line 23: Nice, thanks for separating them out.

cadenmarchese (approved):
LGTM!

bennerv (approved):
LGTM - benefit of removing this test outweighs the benefits of keeping this test to ensure the machine_controller was functioning as expected.

PR #2182, opened 2022-06-14
Pass NIC on CreateOrUpdate call

bennerv:

Fixes failed deletion of failed NIC during cluster deletion

We're not passing in the NIC during cluster deletion, which results in the NIC not being able to be discovered and ARM failing out. 

### Test plan for issue:

This matches the call in reconcilefailednic.go now which we know works.  

no


s-amann (approved):

jhoreman (approved):
Reviewed this change live. LGTM.

ross-bryan (approved):
lgtm

PR #2183, opened 2022-06-14
Use local golangci-lint vendored

bennerv:

Removes the need for developers to download the golangci-lint binary and instead use the one we're already vendoring for use elsewhere in our code. 

### Test plan for issue:

CI should catch any issues.

Updated in PR


bennerv:
Thanks @AldoFusterTurpin for the link.  

https://golangci-lint.run/usage/install/#local-installation
> Note: such go install/go get installation aren't guaranteed to work. We recommend using binary installation.

Recommended approach is to install locally, but still want feedback.  This would unify everyone's dev env, preventing any issues with dependencies between devs.

facchettos:
- As Aldo said , the binary is the recommended way to install. 
- Having in vendored makes it hard to update, I could not update it last time to have it understand proper config files. As a result we were using a version that was more than 2 years old. 
- Vendoring does not guarantee that people have the same version in case of update anyways, and the version in the master branch will always be the source of truth
- If you set up your IDE correctly, golangci-lint lint should show up live on your code, but you need to install it with binary to do it.

And as a more personal note, I think that vendoring should be avoided whenever possible.

facchettos:
I also think that we should consider that https://golangci-lint.run/usage/install/#versioning-policy
As long as we don't change major version, we're fine

PR #2187, opened 2022-06-15
add docs, script, and makefile to log into/use shared cluster

ross-bryan:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14732915

PR describes how the shared cluster was built with az aro commands and in the AME tenant. 

this is docs update


docs/shared-cluster.md
```diff
@@ -0,0 +1,68 @@
...
+
+We have the kubeadmin credentials as well as the kubeadmin kubeconfig file. You can use either to authenticate to the cluster.
+
+* Make secrets:
+
+```
+SECRET_SA_ACCOUNT_NAME=rharosecretsdev make secrets
+```
+
+* Oc login, assuming `env` sources `secrets/env`:
```
s-amann, line 27:
```suggestion
* oc login, assuming `env` sources `secrets/env`:
```

s-amann (approved):
LGTM

s-amann (dismissed):

hack/shared-cluster.sh
```diff
@@ -0,0 +1,68 @@
...
+
+    az network vnet subnet update \
+      --name master-subnet \
+      --resource-group $RESOURCEGROUP \
+      --vnet-name aro-vnet \
+      --disable-private-link-service-network-policies true
+
+    echo "resource group and vnet/subnets created..."
+    echo "creating cluster..."
+    az aro create \
```
SudoBrendan, line 50: Do we need to unset RP_MODE for consistency here, or have a check that it's unset?

SudoBrendan (changes requested):
I think that RP_MODE can play with this functionality, we should have a check or manually unset it.

hack/shared-cluster.sh
ross-bryan: RP_MODE should not be set. the instructions for creating/deleting cluster do not include sourcing `env` file. Instructions are here: https://github.com/Azure/ARO-RP/pull/2187/files#diff-a7eb7df5033da5999c603571053ea9a6811ec9aac786279fb7e5ba61935c182bR55

ross-bryan: this will also be the cloudshell az cli, which will not have development aro extension and will not honor `RP_MODE`

SudoBrendan (dismissed):

karanmagdani (commented):
LGTM! Only question is regarding the kubeadmin credentials being a part of our secrets folder. I think we should instead have a makefile target that retrieves credentials on the fly and logs into the cluster that way instead https://docs.microsoft.com/en-us/azure/openshift/tutorial-connect-cluster#connect-to-the-cluster

ross-bryan:
> LGTM! Only question is regarding the kubeadmin credentials being a part of our secrets folder. I think we should instead have a makefile target that retrieves credentials on the fly and logs into the cluster that way instead https://docs.microsoft.com/en-us/azure/openshift/tutorial-connect-cluster#connect-to-the-cluster. That way we potentially never expose those credentials

If our secrets folder is somehow exposed it will likely have nothing to do with this PR. I don't see any harm or risk in them being variables in `secrets/env` or the admin kubeconfig living under `secrets/shared-cluster.kubeconfig`. 

The cluster is in the AME tenant, so you cannot retrieve credentials "on the fly" unless you are on SAW, with elevated devops JIT, logged into azure portal, and have open a cloud shell session. Only under those conditions can the credentials be retrieved from the RP.

karanmagdani:
> The cluster is in the AME tenant, so you cannot retrieve credentials "on the fly" unless you are on SAW, with elevated devops JIT, logged into azure portal, and have open a cloud shell session. Only under those conditions can the credentials be retrieved from the RP.

Makes Sense, Thanks for the answer

karanmagdani (approved):

docs/shared-cluster.md
```diff
@@ -0,0 +1,68 @@
...
+
+* `oc login`, assuming `env` sources `secrets/env`:
+```
+. ./env
+make shared-cluster-login
+```
+
+* Use kubeconfig
+```
+export KUBECONFIG=$(pwd)/secrets/shared-cluster.kubeconfig
```
AldoFusterTurpin, line 35: What about using 
```bash
$PWD 
```
instead of 
```bash
$(pwd)
``` 
?
It appears more often in our codebase and it doesn't require the execution of a command in a sub-shell.

hack/shared-cluster.sh
```diff
@@ -0,0 +1,68 @@
...
+      --resource-group $RESOURCEGROUP \
+      --cluster-resource-group $CLUSTERRESOURCEGROUP \
+      --name $CLUSTER \
+      --vnet aro-vnet \
+      --master-subnet master-subnet \
+      --worker-subnet worker-subnet || exit 1
```
AldoFusterTurpin, line 56: What about instead of doing
```bash
|| exit 1
```
, we do this on line 57:
```bash
if [[ "$?" != 0 ]]; then 
      exit 1
fi
```
IMHO even though the first form is more compact, the latter is easier to read as the previous command is expanded in multiple lines.

hack/shared-cluster.sh
```diff
@@ -0,0 +1,68 @@
...
+    echo "cluster create completed..."
+    exit 0
+fi
+
+if [[ $1 == "delete" ]]; then
+  az aro delete --name $CLUSTER -g $RESOURCEGROUP --yes
+  az network vnet delete --name aro-vnet -g $RESOURCEGROUP
+  az group delete --resource-group $RESOURCEGROUP --yes
+  exit 0
+fi
```
AldoFusterTurpin, line 66: Just a question: should we check for errors before
```bash
exit 0
```
and do
```bash
exit 1
````
in case any (as it's done in the previous if block)? 

(Maybe I am missing something 🙂)

AldoFusterTurpin (commented):
@ross-bryan Awesome! 
Just a couple of minor suggestions and a question.
Thanks!

docs/shared-cluster.md
ross-bryan: done

hack/shared-cluster.sh
ross-bryan: done

s-amann (dismissed):

hack/shared-cluster.sh
ross-bryan: actually, just removing exit codes...

hack/shared-cluster.sh
ross-bryan: I don't want to add complexity/reduce readability into this part of the script and I don't really care if these commands fail. I would actually want the commands to continue if the previous one fails because we just want to delete all of these resources.

As some compromise, I've changed the conditional flow so that the exit is implicit instead of explicit. I required the exit before because there was a function call to `usage` that default ran, and now we avoid the func call of`usage()` through an else statement.

s-amann (approved):

bennerv (approved):
lgtm

PR #1969, opened 2022-02-15
Filter out unwanted data from azure list geneva action

ArielLima:
### Which issue this PR addresses:

Filters out snapshots from geneva action **_list all azure resources_**  [Ticket](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13151600)

### Test plan for issue:

Run it locally and ensure that snapshot resources are not present.

### Is there any documentation that needs to be updated for this PR?

Unsure, happy to update if it is the case

PR #2188, opened 2022-06-16
Create new ListResourcesScoped geneva action

cadenmarchese:

Our current ListResources / `List Cluster Azure Resouces` Geneva Action consumes a lot of memory and can cause RP VMs to restart. This action is meant to allow resource filters to be passed into the ListResources request so that if needed, filtering is done on the ARM side rather than client or RP side.

As a bonus we could use this to get only virtual machines in a resource group and automate it so that we don't have to run the entirety of `List Cluster Azure Resouces` to determine whether or not a cluster has been deallocated.

### Test plan for issue:

- Can be manually tested like other adminAPI calls. Example:
```
$ curl -X GET -k "https://localhost:8443/admin/subscriptions/tOpenShift/openShiftClusters/$CLUSTER/resourcesscoped?resourceType=Microsoft.Compute/virtualMachines" --header "Content-Type: application/json" -d "{}"
```
- Unit testing
- Do we need a new e2e test for this?

I have added the action to the deploy_development doc.

### DIFF
--- a/pkg/frontend/adminactions/resources_list_scoped.go
+++ b/pkg/frontend/adminactions/resources_list_scoped.go
@@ -1,0 +1,48 @@
+package adminactions
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	mgmtcompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
+
+	"github.com/Azure/ARO-RP/pkg/util/arm"
+	"github.com/Azure/ARO-RP/pkg/util/stringutils"
+)
+
+func (a *azureActions) ResourcesListScoped(ctx context.Context, resourceType string) ([]byte, error) {
+	clusterRGName := stringutils.LastTokenByte(a.oc.Properties.ClusterProfile.ResourceGroupID, '/')
+
+	filter := fmt.Sprintf("resourceType eq '%s'", resourceType)
+	resources, err := a.resources.ListByResourceGroup(ctx, clusterRGName, filter, "", nil)
+	if err != nil {
+		return nil, err
+	}
+
+	armResources := make([]arm.Resource, 0)
+	for _, res := range resources {
+		if *res.Type == "Microsoft.Compute/virtualMachines" {
+			vm, err := a.virtualMachines.Get(ctx, clusterRGName, *res.Name, mgmtcompute.InstanceView)
+			if err != nil {
+				a.log.Warn(err)
+				armResources = append(armResources, arm.Resource{
+					Resource: res,
+				})
+				continue
+			}
+			armResources = append(armResources, arm.Resource{
+				Resource: vm,
+			})
+		}
+
+		armResources = append(armResources, arm.Resource{
+			Resource: res,
+		})
+	}
+
+	return json.Marshal(armResources)
+}
--- a/docs/deploy-development-rp.md
+++ b/docs/deploy-development-rp.md
@@ -171,6 +171,12 @@
   curl -X GET -k "https://localhost:8443/admin/subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$RESOURCEGROUP/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER/resources"
   ```
 
+* List scoped cluster Azure Resources of a dev cluster
+  ```bash
+  RESOURCETYPE="Microsoft.Compute/virtualMachines"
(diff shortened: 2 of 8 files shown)


m1kola (commented):
Instead of introducing a new endpoint and action - should we look into optimising the esisting one? And/or parametrising it?

> Geneva Action consumes a lot of memory and can cause RP VMs to restart.

1. What is the consumption for the existing call? It does indeed hold a lot of data in memory before returning a response, but what are the numbers?
1. In the existing endpoint - can we exclude data we don't need from the output?
1. In the existing endpoint - can we stream data into the response instead of dumping all at once?

m1kola:
@cadenmarchese I think next steps here are:

* Understand what exactly is causing VM/RP restart. Are there any special conditions for VM to go OOM? How do I reproduce it? That is why I was asking about numbers in the first question: typical response is big, but are we sure that it is the cause?
* Once we have better understanding of the root cause - we can discuss potential solutions.

bennerv:
> * Understand what exactly is causing VM/RP restart. Are there any special conditions for VM to go OOM? How do I reproduce it? That is why I was asking about numbers in the first question: typical response is big, but are we sure that it is the cause?
We have directly correlated listing azure resources by ARO SREs to RP instances running OOM in the past.  For this, we added a filter in #1969 on [this user story](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13151600) 

> * Once we have better understanding of the root cause - we can discuss potential solutions.

Geneva actions dictate [best-practices](https://eng.ms/docs/products/geneva/actions/howdoi/bestpracticesauthoringgenevaactions) to "Don't target multiple resources in a single execution of operation"

bennerv:
@m1kola 
> I think that the output of list azure resources is big, but:
> 
>     * It is very rare that we have multiple concurrent "list resources" requests
> 
>     * It is not big enough to consume gigs of memory
> 
> 
> If a single request (or even 5) to an endpoint can crash the instance - then we have an opportunity to optimise our existing code instead of inventing something new which might or might not help us.

With some geneva automation planned to be implemented within IcMs we'll be running actions listing out azure resources to determine if a cluster is unsupported or not.  We'll definitely have an increased usage of geneva actions as a whole.  

> > Geneva actions dictate [best-practices](https://eng.ms/docs/products/geneva/actions/howdoi/bestpracticesauthoringgenevaactions) to "Don't target multiple resources in a single execution of operation"
> 
> We do not target multiple resources here: specific cluster is our target resource. We are getting azure resources of a single cluster (target). IMO, we are fine here.

We're specifying a single ARO resource id to list _all_ resources within the cluster resource group.  The number of resources within this group is arbitrary.  A customer could have 1000 VMs (assuming it was supported at the time), thousands of snapshots, disks, VMs, load balancers, etc.  If we're reading them all into memory at the same time (as we're currently doing, subject to change) it's just a ticking time bomb before it causes issues.  Deletion code also does a similar thing of getting all resources within the cluster resource group.  Could be a problem there if the issue lies with the `ListByResourceGroup` call, rather than the unmarshal or something else.  

@cadenmarchese let's try to determine if the source of the OOMs is the marshalling, or listing out resources as it could affect us in other places down the road.  Then as @m1kola said, we can determine if this solution, streaming, or something else would be a better approach.  I think Jim did a presentation on pprof in the demos section which would be helpful.  Can create a dev cluster, remove the filtering on snapshots, and just create a ton of VM snapshots within the cluster resource group and do a `pprof` analysis on the memory usage during the list clusters action to point to the exact source of OOM.  It may not OOM on your laptop, but it should point you to the functions/objects consuming the most amount of memory.

pkg/frontend/adminactions/resources_list_scoped.go
```diff
@@ -0,0 +1,48 @@
...
+				a.log.Warn(err)
+				armResources = append(armResources, arm.Resource{
+					Resource: res,
+				})
+				continue
+			}
+			armResources = append(armResources, arm.Resource{
+				Resource: vm,
+			})
+		}
```
s-amann, line 40: I'm not sure we need this block since this endpoint would be to give an overview of the resources requested. If we do want to continue to get extra information just for `virtualMachines` we should wrap lines `42-44` in an else block to this condition. Otherwise we would be adding the resource info twice for `virtualMachines`

cadenmarchese:
Pausing work on the code until this comment https://github.com/Azure/ARO-RP/pull/2188#issuecomment-1163385229 is addressed. Reviews in the meantime are welcome!

cadenmarchese:
Closing this PR for now as we have decided to go with a different approach using an optimized version of ListResources rather than a new frontend endpoint for a single resource type. For more info on this, see the following work items:

- https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14717959/
- https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14847686/
- https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14717793/

PR #2189, opened 2022-06-16
operator deployment improvement

facchettos:
Improved the operator deployment process so we can use proper templating instead of custom  (needed for ocm pull secret validation). (needed for https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13933226/)

Fixes

### Test plan for issue:
already covered by existing tests

### DIFF
--- a/pkg/operator/deploy/deploy.go
+++ b/pkg/operator/deploy/deploy.go
@@ -4,15 +4,16 @@ package deploy
 // Licensed under the Apache License 2.0.
 
 import (
+	"bytes"
 	"context"
 	"embed"
 	"errors"
 	"fmt"
 	"strings"
+	"text/template"
 	"time"
 
 	"github.com/sirupsen/logrus"
-	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	extensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
@@ -33,7 +34,6 @@ import (
 	aroclient "github.com/Azure/ARO-RP/pkg/operator/clientset/versioned"
 	"github.com/Azure/ARO-RP/pkg/operator/controllers/genevalogging"
 	"github.com/Azure/ARO-RP/pkg/util/dynamichelper"
-	utilembed "github.com/Azure/ARO-RP/pkg/util/embed"
 	"github.com/Azure/ARO-RP/pkg/util/pullsecret"
 	"github.com/Azure/ARO-RP/pkg/util/ready"
 	"github.com/Azure/ARO-RP/pkg/util/restconfig"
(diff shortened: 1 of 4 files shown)


SudoBrendan (commented):
These changes look OK to me, thank you for the work here! I just have a few questions about the added labels/envs on the deployments - why are those added here? Can you add that to the description of the PR?

facchettos:
> These changes look OK to me, thank you for the work here! I just have a few questions about the added labels/envs on the deployments - why are those added here? Can you add that to the description of the PR?

@SudoBrendan  These used to be added inside the code: 
```go
	// set the image for the deployments
		if d, ok := obj.(*appsv1.Deployment); ok {
			if d.Labels == nil {
				d.Labels = map[string]string{}
			}
			var image string

			if o.oc.Properties.OperatorVersion != "" {
				image = fmt.Sprintf("%s/aro:%s", o.env.ACRDomain(), o.oc.Properties.OperatorVersion)
				d.Labels["version"] = o.oc.Properties.OperatorVersion
			} else {
				image = o.env.AROOperatorImage()
				d.Labels["version"] = version.GitCommit
			}
		for i := range d.Spec.Template.Spec.Containers {
			d.Spec.Template.Spec.Containers[i].Image = image
			if o.env.IsLocalDevelopmentMode() {
					d.Spec.Template.Spec.Containers[i].Env = append(d.Spec.Template.Spec.Containers[i].Env, corev1.EnvVar{
						Name:  "RP_MODE",
						Value: "development",
					})
				}
               }
```

So basically they always were there, it just was not obvious since we had to understand the code to know that they were added to the deployments.

pkg/operator/deploy/deploy.go
```diff
@@ -83,49 +83,74 @@ func New(log *logrus.Entry, env env.Interface, oc *api.OpenShiftCluster, arocli
...
-			var image string
-
-			if o.oc.Properties.OperatorVersion != "" {
-				image = fmt.Sprintf("%s/aro:%s", o.env.ACRDomain(), o.oc.Properties.OperatorVersion)
-				d.Labels["version"] = o.oc.Properties.OperatorVersion
-			} else {
-				image = o.env.AROOperatorImage()
-				d.Labels["version"] = version.GitCommit
-			}
+func templateManifests(data deployData) ([][]byte, error) {
```
AldoFusterTurpin, line 94: IMHO the function can be simplified a bit by using an aux function to execute the templates (I think it reads a bit better and narrows the scope of some variables like "result". I also propose a more specific name rather than "result"). 

Important question🧐: why are you using "buff.Reset()" if buff is declared inside the most nested "for" (so it is instantiated in each iteration)? If you want to reuse "buff" shouldn't you declare it outside that "for" loop? Is that a minor error or I am missing something (this is more probable  🙂)? 

I propose something like this (also a minor change: I have reordered the declaration of templates to be in the same order as they appear in the templatesArray and extract the common string to a constant).

```go
func executeTemplates(templatesArray []*template.Template, data deployData) ([][]byte, error) {
	templateExecutionResult := make([][]byte, 0)

	for _, templates := range templatesArray {
		for _, v := range templates.Templates() {
			buff := &bytes.Buffer{}
			if err := v.Execute(buff, data); err != nil {
				return nil, err
			}
			templateExecutionResult = append(templateExecutionResult, buff.Bytes())
			buff.Reset()
		}
	}

	return templateExecutionResult, nil
}

func templateManifests(data deployData) ([][]byte, error) {
	const ResourcesFolder = "staticresources"

	templatesMaster, err := template.ParseFS(embeddedFiles, ResourcesFolder+"/master/*")
	if err != nil {
		return nil, err
	}

	templatesRoot, err := template.ParseFS(embeddedFiles, ResourcesFolder+"/*.yaml")
	if err != nil {
		return nil, err
	}

	templatesWorker, err := template.ParseFS(embeddedFiles, ResourcesFolder+"/worker/*")
	if err != nil {
		return nil, err
	}

	templatesArray := []*template.Template{templatesMaster, templatesRoot, templatesWorker}

	result, err := executeTemplates(templatesArray, data)
	if err != nil {
		return nil, err
	}

	return result, nil
}
```

pkg/operator/deploy/deploy.go
```diff
@@ -83,49 +83,74 @@ func New(log *logrus.Entry, env env.Interface, oc *api.OpenShiftCluster, arocli
...
+func (o *operator) createObjects() ([]kruntime.Object, error) {
+	image := ""
+	hasVersion := false
+	if o.oc.Properties.OperatorVersion != "" {
+		image = fmt.Sprintf("%s/aro", o.env.ACRDomain())
+		hasVersion = true
+	} else {
+		image = o.env.AROOperatorImage()
+	}
+	data := deployData{IsLocalDevelopment: o.env.IsLocalDevelopmentMode(), Image: image, Version: o.oc.Properties.OperatorVersion, GitCommit: version.GitCommit, HasVersion: hasVersion}
```
AldoFusterTurpin, line 133: Opinionated: probably line too long even is just for a variable instantiation.

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
```diff
@@ -25,8 +30,13 @@ spec:
         args:
         - operator
         - master
-        image: GENERATED
+        image: "{{ .Image }}{{ if .HasVersion}}:{{ if .IsLocalDevelopment }}{{ .GitCommit }}{{ else }}{{ .Version }}{{end}}{{end}}"
```
s-amann, line 33: it looks weird to me to have the colon after the first condition, is this required? The result should look like: `image:gitCommit` or `image:version` so could the colon be moved to after `{{.Image}}`? Same for in the worker file.

pkg/operator/deploy/deploy.go
```diff
@@ -83,49 +83,74 @@ func New(log *logrus.Entry, env env.Interface, oc *api.OpenShiftCluster, arocli
...
-				if o.env.IsLocalDevelopmentMode() {
-					d.Spec.Template.Spec.Containers[i].Env = append(d.Spec.Template.Spec.Containers[i].Env, corev1.EnvVar{
-						Name:  "RP_MODE",
-						Value: "development",
-					})
-				}
+	for _, templates := range templatesArray {
+		for _, v := range templates.Templates() {
+			buff := &bytes.Buffer{}
+			if err := v.Execute(buff, data); err != nil {
```
s-amann, line 114: cool, so `v.Execute` will overlay the `deployData` struct to the template so those in file conditions can be used to determine the values. This is pretty neat!

s-amann (approved):
lgtm, just one question

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
facchettos: It was failing the unit tests if not done this way , because if you don't have a version you end up with `image:` without a tag, and then it's not valid

pkg/operator/deploy/deploy.go
facchettos: exactly ! it makes it very obvious which field comes from where and represents what.

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
s-amann: I see, thank you!

AldoFusterTurpin (commented):
@facchettos Awesome 👏. 

Just some suggestion in case they are useful.

pkg/operator/deploy/deploy.go
facchettos: I think it is fine to keep it this way, there is no complexity in the reading of the templates

pkg/operator/deploy/deploy.go
```diff
@@ -83,49 +83,74 @@ func New(log *logrus.Entry, env env.Interface, oc *api.OpenShiftCluster, arocli
...
 			}
+			result = append(result, buff.Bytes())
+			buff.Reset()
 		}
+	}
+	return result, nil
+}
 
-		results = append(results, obj)
+func (o *operator) createObjects() ([]kruntime.Object, error) {
```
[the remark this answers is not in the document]
facchettos: your first suggestion is only a matter of preference, and I prefer logic is better with my way, since we add when we have the objects, instead of instantiating  zero objects and then assigning value to them.

pkg/operator/deploy/deploy.go
facchettos: for the reset, I just prefer to reset buffers once used. It is just a habit so I don't forget to reset those.

AldoFusterTurpin: Maybe I am wrong but speaking about the "reset", I think it is not just unnecessary but also incorrect as the variable buff is re-instantiated in every iteration of the for loop and the reset is having no effect -> so you are basically resetting an object that will be deleted in the next "line".

For the part of the template, it was just a proposal to apply the [SRP](https://levelup.gitconnected.com/practical-solid-in-golang-single-responsibility-principle-20afb8643483) because if the way the templates are executed changes, you just want to touch the function responsible for that, not a function that it's also responsible to obtain the templates-> mixed responsibilities. 

SRP improves the codebase health in the long run.

pkg/operator/deploy/deploy.go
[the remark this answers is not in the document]
AldoFusterTurpin: Okay  🙂
IMHO SRP still applies for the "k8sObjects" creation, and short functions are always better than long ones.

s-amann (approved):

hawkowl (approved):

PR #2190, opened 2022-06-16
az aro update: startvms, validate api servers, and configure api/ingress certs

ross-bryan:

Allows customers to self-service certificate updates instead of relying on ARO SRE to patch the cluster.

### DIFF
--- a/pkg/cluster/install.go
+++ b/pkg/cluster/install.go
@@ -112,6 +112,10 @@ func (m *manager) Update(ctx context.Context) error {
 		steps.Action(m.clusterSPObjectID),
 		// credentials rotation flow steps
 		steps.Action(m.createOrUpdateClusterServicePrincipalRBAC),
+		steps.Action(m.startVMs),
+		steps.Condition(m.apiServersReady, 30*time.Minute, true),
+		steps.Action(m.configureAPIServerCertificate),
+		steps.Action(m.configureIngressCertificate),
 		steps.Action(m.createOrUpdateDenyAssignment),
 		steps.Action(m.updateOpenShiftSecret),
 		steps.Action(m.updateAROSecret),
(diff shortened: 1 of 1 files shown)


pkg/cluster/install.go
```diff
@@ -112,6 +112,10 @@ func (m *manager) Update(ctx context.Context) error {
 		steps.Action(m.clusterSPObjectID),
 		// credentials rotation flow steps
 		steps.Action(m.createOrUpdateClusterServicePrincipalRBAC),
+		steps.Action(m.startVMs),
```
m1kola, line 115: We still have `startVMs` step in `AdminUpdate`, but as far as I'm aware that is the current policy (decision by BU) is not to start VMs which were powered off.

Just something to check: Are we ok to have it in `Update`?

pkg/cluster/install.go
```diff
@@ -112,6 +112,10 @@ func (m *manager) Update(ctx context.Context) error {
...
+		steps.Condition(m.apiServersReady, 30*time.Minute, true),
+		steps.Action(m.configureAPIServerCertificate),
+		steps.Action(m.configureIngressCertificate),
 		steps.Action(m.createOrUpdateDenyAssignment),
```
bennerv, line 119: Can we move `createOrUpdateDenyAssignment` above the kubernetes things?  it's strictly azure and is more likely to succeed

bennerv (changes requested):

pkg/cluster/install.go
ross-bryan: I think there is a difference between SRE turning them on, and cx running the `az aro update` command. Jerome indicated in a slack thread he was ok with update performing the startVMs check.

rogbas (dismissed):
LGTM

cadenmarchese (approved):
Thanks, looks good to me. 💯

ross-bryan (commented):
move denyassignment step before k8s steps

ross-bryan:
lgtm

bennerv (approved):
lgtm - ship it

PR #2191, opened 2022-06-17
update deployment name/logged arm template deployed

ross-bryan:

Fixes name of deployment logged so it is more clear


bennerv (approved):
lgtm

PR #2194, opened 2022-06-20
Refactor/simplify openshift cluster database

AldoFusterTurpin:

No issue. This PR is a refactor, no behavior was changed.

This PR was created doing pair programming with @petrkotas.

What this PR does?
1. Refactor of pkg/database/openshiftclusters.go to define a custom function type to simplify the code. 
2. Refactor of pkg/cluster/sshkeys.go generateSSHKey() in two functions and files and use them in pkg/cluster/install.go

All that to simplify the codebase in order to be easier to read and understand.

### Test plan for issue:

N/A

N/A

### DIFF
--- a/pkg/cluster/sshkeys.go
+++ b/pkg/cluster/sshkeys.go
@@ -20,7 +20,8 @@ func (m *manager) ensureSSHKey(ctx context.Context) error {
 		return err
 	}
 
-	var f database.OpenShiftDocumentMutator = func(doc *api.OpenShiftClusterDocument) error {
+	var f database.OpenShiftClusterDocumentMutator
+	f = func(doc *api.OpenShiftClusterDocument) error {
 		if doc.OpenShiftCluster.Properties.SSHKey == nil {
 			doc.OpenShiftCluster.Properties.SSHKey = x509.MarshalPKCS1PrivateKey(sshKey)
 		}
--- a/pkg/cluster/sshkeys.go
+++ b/pkg/cluster/sshkeys.go
@@ -13,13 +13,6 @@ import (
 	"github.com/Azure/ARO-RP/pkg/api"
 )
 
-func (m *manager) ensureSSHKey(ctx context.Context) error {
-	var err error
-	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, mutateSSHKey)
-
-	return err
-}
-
 func mutateSSHKey(doc *api.OpenShiftClusterDocument) error {
 	if doc.OpenShiftCluster.Properties.SSHKey != nil {
 		return nil
@@ -33,6 +26,13 @@ func mutateSSHKey(doc *api.OpenShiftClusterDocument) error {
 	doc.OpenShiftCluster.Properties.SSHKey = x509.MarshalPKCS1PrivateKey(sshKey)
 
 	return nil
+}
+
+func (m *manager) ensureSSHKey(ctx context.Context) error {
+	updatedDoc, err := m.db.PatchWithLease(ctx, m.doc.Key, mutateSSHKey)
+	m.doc = updatedDoc
+
+	return err
 }
 
 func randomLowerCaseAlphanumericStringWithNoVowels(n int) (string, error) {
--- a/pkg/cluster/sshkeys.go
+++ b/pkg/cluster/sshkeys.go
@@ -15,12 +15,12 @@ import (
 
 func (m *manager) ensureSSHKey(ctx context.Context) error {
 	var err error
-	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, sshKeyMutatorFn)
+	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, mutateSSHKey)
 
 	return err
 }
 
-func sshKeyMutatorFn(doc *api.OpenShiftClusterDocument) error {
+func mutateSSHKey(doc *api.OpenShiftClusterDocument) error {
 	if doc.OpenShiftCluster.Properties.SSHKey != nil {
 		return nil
 	}
--- a/pkg/cluster/sshkeys.go
+++ b/pkg/cluster/sshkeys.go
@@ -11,17 +11,19 @@ import (
 	"math/big"
 
 	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database"
 )
 
 func (m *manager) generateSSHKey(ctx context.Context) error {
 	var err error
-	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, func(doc *api.OpenShiftClusterDocument) error {
-		if doc.OpenShiftCluster.Properties.SSHKey == nil {
-			sshKey, err := rsa.GenerateKey(rand.Reader, 2048)
-			if err != nil {
-				return err
-			}
 
+	sshKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/cluster/sshkeys.go
```diff
@@ -11,33 +11,24 @@ import (
...
-func (m *manager) generateSSHKey(ctx context.Context) error {
-	var err error
-	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, func(doc *api.OpenShiftClusterDocument) error {
-		if doc.OpenShiftCluster.Properties.SSHKey == nil {
-			sshKey, err := rsa.GenerateKey(rand.Reader, 2048)
-			if err != nil {
-				return err
-			}
+func (m *manager) ensureSSHKey(ctx context.Context) error {
+	sshKey, err := rsa.GenerateKey(rand.Reader, 2048)
```
bennerv, line 18: I think a lot of our other `ensure` or `generate` functions check at the top if the property is nil or empty before performing any work.  Can we do that here as well?

```
if doc.OpenShiftCluster.Properties.SSHKey == nil {
    return nil
}

sshKey, err := rsa.GenerateKey(rand.Reader, 2048)
if err != nil {
    return err
}

var f database.OpenShiftClusterDocumentMutator = func(doc *api.OpenShiftClusterDocument error {
    doc.OpenShiftCluster.Properties.SSHKey = x509.MarshalPKCS1PrivateKey(sshKey)
    return nil
}

m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, f)
return err

File: pkg/cluster/storageSuffix.go [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+	"context"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database"
+)
+
+func (m *manager) ensureStorageSuffix(ctx context.Context) error {
+	var err error
+
+	var f database.OpenShiftClusterDocumentMutator = func(doc *api.OpenShiftClusterDocument) error {
```
bennerv, line 16: Same for this one.  Check if it's empty at the top.

bennerv (changes requested):
Just some nits, with the original function.  Otherwise I'm +1 on the changes.

pkg/cluster/sshkeys.go
```diff
@@ -11,33 +11,24 @@ import (
...
+	if err != nil {
+		return err
+	}
 
+	var f database.OpenShiftClusterDocumentMutator = func(doc *api.OpenShiftClusterDocument) error {
```
facchettos, line 23: I am not sure I like the name of this, something a little bit more expressive would be better imho
I don't really like when we declare functions like that btw, it makes it impossible to test them. Here it is not a big big deal, but it is something to keep in mind

pkg/cluster/storageSuffix.go
facchettos, line 16: Same comment than above for here. Though here I would say that it is really better to declare the function outside in its own func declaration. And then you can test it :)

facchettos (changes requested):
It would be better to have the functions declared as their own func and then used from the other function. This way we can test them. For sshkey it is not absolutely necessary (but I don't like the `f` variable name), but it is for storagesuffix imo

pkg/cluster/sshkeys.go
AldoFusterTurpin: Tks. In this case, we just want to proceed if SSHKey is nil, so your if condition should be reversed. I now have changed that to make it explicit with an early return. See the latest changes.

pkg/cluster/storageSuffix.go
AldoFusterTurpin: Tks. Done. Same comment as before.

pkg/cluster/storageSuffix.go
```diff
@@ -0,0 +1,36 @@
...
+
+	if doc.OpenShiftCluster.Properties.StorageSuffix == "" {
+		var err error
+
+		doc.OpenShiftCluster.Properties.StorageSuffix, err = randomLowerCaseAlphanumericStringWithNoVowels(5)
+		if err != nil {
+			return err
+		}
+
+		doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName = ImageRegistry + doc.OpenShiftCluster.Properties.StorageSuffix
```
s-amann, line 23: nit: we don't need this line as it is duplicated on line 26
```suggestion
```

pkg/cluster/storageSuffix.go
```diff
@@ -0,0 +1,36 @@
...
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+)
+
+func openShiftClusterDocumentMutatorFn(doc *api.OpenShiftClusterDocument) error {
```
s-amann, line 12: nit: can we rename this to `mutateStorageSuffix` so it is inline with the other 'mutate' function in this PR?

s-amann (commented):
overall lgtm, some minor tweaks and this is perfect.

pkg/cluster/sshkeys.go
bennerv: Yeah my bad typo'd it :)

pkg/cluster/storageSuffix.go
```diff
@@ -0,0 +1,36 @@
...
+	const ImageRegistry = "imageregistry"
+
+	if doc.OpenShiftCluster.Properties.StorageSuffix == "" {
+		var err error
+
+		doc.OpenShiftCluster.Properties.StorageSuffix, err = randomLowerCaseAlphanumericStringWithNoVowels(5)
```
facchettos, line 18: Is it necessary to have it this way? wouldn't `:=` be sufficient here ?

pkg/cluster/storageSuffix.go
AldoFusterTurpin: Well spotted. Thanks! Done.

pkg/cluster/storageSuffix.go
AldoFusterTurpin: Good idea. Done. 😄

pkg/cluster/storageSuffix.go
AldoFusterTurpin: In that case no because the result is directly assigned to m.doc and it's not a new variable, so -> Error "expected identifier on left side of :="
But using a temp variable can be done, so done. Thanks for the suggestion 😄

pkg/cluster/sshkeys.go
```diff
@@ -13,32 +13,26 @@ import (
...
-	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, func(doc *api.OpenShiftClusterDocument) error {
-		if doc.OpenShiftCluster.Properties.SSHKey == nil {
-			sshKey, err := rsa.GenerateKey(rand.Reader, 2048)
-			if err != nil {
-				return err
-			}
-
-			doc.OpenShiftCluster.Properties.SSHKey = x509.MarshalPKCS1PrivateKey(sshKey)
-		}
+	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, mutateSSHKey)
```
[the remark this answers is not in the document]
AldoFusterTurpin: Same as before. Used a temp variable to achieve it. Done.

facchettos (approved):
LGTM

s-amann (approved):
lgtm

bennerv (approved):
lgtm

PR #2195, opened 2022-06-21
[M6] Generate database UUIDs in a way that can be deterministic for testing

hawkowl:

splitting out of a followup to #2165

Moves database-related UUID generation into the database layer, where it can then be defined in a consistent manner for testing.

In testing, each database's UUIDs are namespaced where the first 14 bytes are a digit that corresponds with it, and then an incrementing counter.

This PR doesn't touch many tests that could use this functionality, but it's very useful for the followup to #2165 so I decided to make it consistently available.

### Test plan for issue:

mostly tests

N/A


test/util/deterministicuuid/deterministicuuid.go
```diff
@@ -0,0 +1,53 @@
...
+	panic("unimplemented")
+}
+
+// NewV4 generates a UUID that increments each call, using a counter to specify
+// the last two bytes and namespaced by the first byte.
+func (g *gen) NewV4() (uuid.UUID, error) {
+	g.counter++
+	// 16 bits of uuid ought to be enough for any test :)
+	return uuid.FromBytesOrNil([]byte{
+		g.namespace, g.namespace, g.namespace, g.namespace, g.namespace, g.namespace, g.namespace, g.namespace, g.namespace, g.namespace, g.namespace, g.namespace, g.namespace, g.namespace, byte(uint8(g.counter >> 8)), byte(uint8(g.counter)),
```
facchettos, line 51: wouldn't only one `g.namespace` be sufficient here? I don't really understand why we would need to have the namespace on all the other bytes

test/util/deterministicuuid/deterministicuuid.go
```diff
@@ -0,0 +1,53 @@
...
+}
+func (g *gen) NewV5(ns uuid.UUID, name string) uuid.UUID {
+	panic("unimplemented")
+}
+func (g *gen) NewV6() (uuid.UUID, error) {
+	panic("unimplemented")
+}
+func (g *gen) NewV7(uuid.Precision) (uuid.UUID, error) {
+	panic("unimplemented")
+}
```
facchettos, line 43: We are not using any of these in the codebase right ?
Couldn't we have an interface that is called `v4Generator` with a `NewV4` which `gofrs/uuid` would implement instead? That would avoid having these functions that just panic, and thus avoid bad surprises

test/util/deterministicuuid/deterministicuuid.go
hawkowl: The idea is that it makes it a bit more obvious that it's not random, and makes the namespacing more prominent. E.g. `0c0c0c0c-0c0c-0c0c-0c0c-0c0c0c0c0001` just _looks_ like it has something up with it :)

s-amann (approved):

test/util/deterministicuuid/deterministicuuid.go
SudoBrendan: I'm also a little confused by our use of `V4` here, which according to [the UUID docs](https://github.com/gofrs/uuid#uuid) says are generated randomly per [RFC-4122](https://datatracker.ietf.org/doc/html/rfc4122#section-4.4), instead of deterministically. It doesn't sound like we're following any RFC here, so can we call the function something like `NewDeterministic` or something to avoid confusion, or am I missing something?

petrkotas: As per the description, I am suspecting this is used for testing purposes.
@hawkowl would you please rename to more testing friendly name please?

petrkotas (dismissed):
Just a nit, other than that I see this as an iprovement.

facchettos:
@petrkotas We cannot change the naming, because these are needed to implement the `gofr/uuid` interface.

test/util/deterministicuuid/deterministicuuid.go
facchettos: but if we keep it this way, we cannot change the naming since it is needed to implement the gofr interface.

s-amann (dismissed):

facchettos (dismissed):

test/util/deterministicuuid/deterministicuuid.go
```diff
@@ -0,0 +1,57 @@
...
+func (g *gen) NewV5(ns uuid.UUID, name string) uuid.UUID {
+	panic("should not be used in the codebase")
+}
+func (g *gen) NewV6() (uuid.UUID, error) {
+	panic("should not be used in the codebase")
+}
+func (g *gen) NewV7(uuid.Precision) (uuid.UUID, error) {
+	panic("should not be used in the codebase")
+}
+
```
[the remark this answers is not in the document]
hawkowl: Went and implemented this.

facchettos (approved):
Thanks!

pkg/api/v20191231preview/openshiftcluster_validatestatic.go
```diff
@@ -173,8 +173,8 @@ func (sv *openShiftClusterStaticValidator) validateConsoleProfile(path string, c
 }
 
 func (sv *openShiftClusterStaticValidator) validateServicePrincipalProfile(path string, spp *ServicePrincipalProfile) error {
-	_, err := uuid.FromString(spp.ClientID)
-	if err != nil {
+	valid := uuid.IsValid(spp.ClientID)
```
s-amann, line 176: this reads much nicer +1

s-amann (approved):

hack/validate-imports/allowed-import-names.yaml
```diff
@@ -109,3 +109,5 @@ allowedImportNames:
     - hivefake
   github.com/openshift/hive/apis/hive/v1:
     - hivev1
+  github.com/gofrs/uuid:
+    - gofrsuuid
```
s-amann, line 113: am I the only one who reads this as "gophers' uuid"?

s-amann (approved):

hack/validate-imports/allowed-import-names.yaml
hawkowl: I think that 'gofrs' is meant to invoke that, yes :)

darthhexx (approved):
Thanks for making those change, LGTM

s-amann (approved):

PR #2196, opened 2022-06-21
Check if CRG already exists and exit if yes

SrinivasAtmakuri:
WIP: If CRG already exists, prevent the cluster creation,
else proceed.

Fixes: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/10029869/

### DIFF
--- a/pkg/util/azureclient/mgmt/features/resourcegroups.go
+++ b/pkg/util/azureclient/mgmt/features/resourcegroups.go
@@ -18,7 +18,6 @@ type ResourceGroupsClient interface {
 	Get(ctx context.Context, resourceGroupName string) (result mgmtfeatures.ResourceGroup, err error)
 	CreateOrUpdate(ctx context.Context, resourceGroupName string, parameters mgmtfeatures.ResourceGroup) (result mgmtfeatures.ResourceGroup, err error)
 	Delete(ctx context.Context, resourceGroupName string) (result mgmtfeatures.ResourceGroupsDeleteFuture, err error)
-	CheckExistence(ctx context.Context, resourceGroupName string) (result autorest.Response, err error)
 	ResourceGroupsClientAddons
 }
 
--- a/pkg/util/mocks/azureclient/mgmt/features/features.go
+++ b/pkg/util/mocks/azureclient/mgmt/features/features.go
@@ -168,21 +168,6 @@ func (m *MockResourceGroupsClient) EXPECT() *MockResourceGroupsClientMockRecorde
 	return m.recorder
 }
 
-// CheckExistence mocks base method.
-func (m *MockResourceGroupsClient) CheckExistence(arg0 context.Context, arg1 string) (autorest.Response, error) {
-	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "CheckExistence", arg0, arg1)
-	ret0, _ := ret[0].(autorest.Response)
-	ret1, _ := ret[1].(error)
-	return ret0, ret1
-}
-
-// CheckExistence indicates an expected call of CheckExistence.
-func (mr *MockResourceGroupsClientMockRecorder) CheckExistence(arg0, arg1 interface{}) *gomock.Call {
-	mr.mock.ctrl.T.Helper()
-	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "CheckExistence", reflect.TypeOf((*MockResourceGroupsClient)(nil).CheckExistence), arg0, arg1)
-}
-
 // CreateOrUpdate mocks base method.
 func (m *MockResourceGroupsClient) CreateOrUpdate(arg0 context.Context, arg1 string, arg2 features.ResourceGroup) (features.ResourceGroup, error) {
 	m.ctrl.T.Helper()


pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -17,11 +18,14 @@ import (
...
 	"github.com/sirupsen/logrus"
 
+	"regexp"
+
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/api/admin"
 	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
 	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/features"
```
bennerv, line 28: Need to fix the import ordering.  It goes as follows:

1. golang standard library

2. third party imports (i.e. github.com/gorilla/mux) 

3. first party imports (i.e. github.com/Azure/ARO-RP/*)

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -176,6 +180,21 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
...
+
+		fpAuthorizer, err := f.env.FPAuthorizer(f.env.TenantID(), f.env.Environment().ResourceManagerEndpoint)
+		if err != nil {
+			return nil, err
+		}
+
+		resourceGroupsClient := features.NewResourceGroupsClient(f.env.Environment(), a[1], fpAuthorizer)
+		rgexists, err := resourceGroupsClient.CheckExistence(ctx, a[2])
+		if rgexists.StatusCode != 404 {
+			return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeClusterResourceGroupFound, "", "Cluster Resource Group already exists. Delete the Cluster RG.")
```
bennerv, line 195: We might want to run this error message by Jerome.  I would rephrase this and not use acronyms such as `RG` within the message.  Clear and concise if possible.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -176,6 +180,21 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
...
+		resourcegroupIDPattern := regexp.MustCompile(resourcegroupIDPatternText)
+		a := resourcegroupIDPattern.FindStringSubmatch(doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID)
+
+		fpAuthorizer, err := f.env.FPAuthorizer(f.env.TenantID(), f.env.Environment().ResourceManagerEndpoint)
+		if err != nil {
+			return nil, err
+		}
+
+		resourceGroupsClient := features.NewResourceGroupsClient(f.env.Environment(), a[1], fpAuthorizer)
+		rgexists, err := resourceGroupsClient.CheckExistence(ctx, a[2])
```
AldoFusterTurpin, line 193: Minor suggestion (IMHO):
"rgexists" looks like a name for a boolean. What about "resourceGroupExistenceResult" or something similar? 
I prefer long descriptive names rather than short ones (when it makes sense).

AldoFusterTurpin (commented):
@SrinivasAtmakuri Great!
Some suggestions: extract some lines to a function to be testeable and change some variable names. Thanks!

bennerv (dismissed):
I don't think this is something that belongs in the frontend, but rather in the backend during installation.

Our frontend only does static validation, which does not issue any calls to Azure RPs, but rather validates what the user passed in appears to be valid in context with the RP.  We try to minimize the amount of time it takes to create the database record by not running any long-running queries or to Azure at all.  

Adding in a call to check if the resource group exists within the frontend code is an anti-pattern of what the frontend was originally designed for.  Please move this check into it's own function within the backend during the Install portion.  I think that's probably a better suited place as well as you will not have to instantiate the resource client as it will already be initialized and exist within the manager interface.

hawkowl:
> I don't think this is something that belongs in the frontend, but rather in the backend during installation.
> 
> Our frontend only does static validation, which does not issue any calls to Azure RPs, but rather validates what the user passed in appears to be valid in context with the RP. We try to minimize the amount of time it takes to create the database record by not running any long-running queries or to Azure at all.
> 
> Adding in a call to check if the resource group exists within the frontend code is an anti-pattern of what the frontend was originally designed for. Please move this check into it's own function within the backend during the Install portion. I think that's probably a better suited place as well as you will not have to instantiate the resource client as it will already be initialized and exist within the manager interface.

@bennerv I originally suggested putting it inside the frontend since we can then perhaps reduce the amount of create-delete cycles that a user might have to do. The CLI does this role for various other things, but I guess this is a protective instance that would apply more in the case of installs via the Azure Resource Manager?

bennerv:
> @bennerv I originally suggested putting it inside the frontend since we can then perhaps reduce the amount of create-delete cycles that a user might have to do. The CLI does this role for various other things, but I guess this is a protective instance that would apply more in the case of installs via the Azure Resource Manager?

The problem lies with how we have historically architected stuff like this.  If we look at the frontend code, it's all static validation which just checks that values appear sane, but doesn't actually make any requests to other Azure Resource providers.  We do see a call to talk to our cosmosdb instance, but that's it from restful or other service call(s).

If we decide to allow this within the frontend, it's going to raise the question of where future changes for installation validation happen, in the frontend or the backend.

Or we could add a new "intermediary" where we do one-time validation of cluster creation that should fail prior to the cosmos record being created so the customer does not have to delete the record to recreate their cluster.  I'd be curious on @m1kola 's thoughts but I don't know how I feel about this to be honest.  We'd have to ensure the decision was well-documented to give devs an easy understanding of where to put future validation code.  

I think the easy choice, but not necessarily the best customer experience, would be to just add the validation to the backend installation steps based on the current structure of validation and the frontend/backend.

m1kola:
Agree with Ben - this belongs to the dynamic validator/backend.

Yes, this means that in case of existing cluster resource group (wrong input from the customer) the customer will have to delete the cluster object and create it again. This is consistent with rest of validations which we do in the dynamic validator (e.g. vnet validations or permissions).

In addition to the RP side validation we can look into adding AZ CLI validation. In Az CLI we can check - if cluster resource group already exists - don't even create a request to ARO RP. But I'm not convinced that implementing this on AZ CLI side worth the effort.

SrinivasAtmakuri:
Moved the logic to the backend as suggested, since we already have validation in the az-aroclient, mimicked the same error message in the output and tested. Thanks

s-amann (dismissed):

SudoBrendan (dismissed):
LGTM

hawkowl (approved):

hawkowl (dismissed):
Can we get a unit test for the function?

bennerv (changes requested):
@hawkowl @SrinivasAtmakuri I think the implementation of this needs to be addressed again. 

Our installation process is designed to be idempotent.  This change breaks the idempotency design.  At any point, if an RP falls off the face of the earth, another should be able to continue the installation process, but in this scenario it could result in failure even though it's in a good state.  

What we should be doing instead is enhancing our `ensureResourceGroup` function to return a better error message if it fails to `CreateOrUpdate`.  If you take an existing resource group and attempt to set the `ManagedBy` field on it, I believe that field can only be set at creation and will fail.  We should swallow that error and return a better one to the user if the `ManagedBy` field does not equal the ocp resource ID.

bennerv:
Adding on as stated before - you can't modify the `ManagedBy` field on an existing resource group so we can use that in `ensureResourceGroup` instead of the extra function.  

```
Bad Request({"error":{"code":"ResourceGroupManagedByMismatch","message":"The managed by property of the resource group cannot be changed from its current value '/subscriptions/<sub-id>/resourceGroups/tschneid'."}})

File: pkg/cluster/deploystorage.go [hunk 4]
```diff
@@ -41,6 +41,20 @@ func (m *manager) ensureInfraID(ctx context.Context) (err error) {
...
+func (m *manager) checkClusterResourceGroupAlreadyExists(ctx context.Context) error {
+	resourceGroup := stringutils.LastTokenByte(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
+
+	clusterResourceGroupExistenceResult, err := m.resourceGroups.CheckExistence(ctx, resourceGroup)
+	if err != nil {
+		return err
+	}
+
+	if clusterResourceGroupExistenceResult.StatusCode != 404 {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeClusterResourceGroupAlreadyExists, "", "Cluster-resource-group '%s' resource group must not exist.", resourceGroup)
```
hawkowl, line 53: I think this should be `"OpenShiftCluster.Properties.ClusterProfiles.ResourceGroupID: Resource group '%s' must not already exist."`, maybe? I'm not super sure.

SrinivasAtmakuri:
> Adding on as stated before - you can't modify the `ManagedBy` field on an existing resource group so we can use that in `ensureResourceGroup` instead of the extra function.
> 
> ```
> Bad Request({"error":{"code":"ResourceGroupManagedByMismatch","message":"The managed by property of the resource group cannot be changed from its current value '/subscriptions/<sub-id>/resourceGroups/tschneid'."}})
> ```

Added the suggest change, please review.

SrinivasAtmakuri:
Testing results:

$ az group show -g srini-aro-crg-manual
{
  "id": "/subscriptions/xxxx/resourceGroups/srini-aro-crg-manual",
  "location": "yyyy",
  "managedBy": null,
  "name": "srini-aro-crg-manual",
  "properties": {
    "provisioningState": "Succeeded"
  },
  "tags": {
    "createdAt": "xyxyxyxyxyx"
  },
  "type": "Microsoft.Resources/resourceGroups"
}

$ az aro create   --resource-group $RESOURCEGROUP   --name $CLUSTER   --vnet srini-aro-vnet   --master-subnet srini-master-subnet   --worker-subnet srini-worker-subnet   --cluster-resource-group srini-aro-crg-manual
(ClusterResourceGroupAlreadyExists) Resource group /subscriptions/xxxxx/resourcegroups/srini-aro-crg-manual must not already exist.
Code: ClusterResourceGroupAlreadyExists
Message: Resource group /subscriptions/xxxx/resourcegroups/srini-aro-crg-manual must not already exist.

s-amann (changes requested):
lgtm, just one last thing to delete `CheckExistance`

s-amann (approved):
LGTM! nice work!

bennerv (approved):
lgtm - good stuff

PR #2200, opened 2022-06-22
Fix panic bug in centraluseuap

cadenmarchese:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14784728

- Fixes an issue deploying clusters in Central US EUAP due to a panic

### Test plan for issue:

- Confirm that the bug is no longer occurring (the region may still be impacted by the simply secure race condition)

No


weinong (approved):

cadenmarchese:
This is looking healthy on our end: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=57189401&view=logs&j=9a35fcf0-5d3a-58cf-398d-ac55d7edc189&t=dff2a1b5-51a3-5546-8fa7-bd6305e55b78

ross-bryan (approved):
I really dont like this :face_exhaling:

PR #2202, opened 2022-06-23
Returning the error if error is captured instead of nil

LiniSusan:
https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14588871

Fixes
Return the error if error is captured instead of nil

Customer wants to identify the error instead generic error(Internal Server Error)

### DIFF
--- a/pkg/util/aad/aad.go
+++ b/pkg/util/aad/aad.go
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	DefaultTimeout = 5 * time.Minute
+	defaultTimeout = 5 * time.Minute
 )
 
 type TokenClient interface {
@@ -33,14 +33,14 @@ func NewTokenClient() TokenClient {
 }
 
 func (tc *tokenClient) GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, tenantID, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error) {
-	spToken, err := NewServicePrincipalToken(clientID, clientSecret, tenantID, aadEndpoint, resource)
+	spToken, err := newServicePrincipalToken(clientID, clientSecret, tenantID, aadEndpoint, resource)
 	if err != nil {
 		return spToken, err
 	}
 
 	tokenAuthorizer := refreshable.NewAuthorizer(spToken)
 
-	err = AuthenticateServicePrincipalToken(ctx, log, tokenAuthorizer, DefaultTimeout)
+	err = authenticateServicePrincipalToken(ctx, log, tokenAuthorizer, defaultTimeout)
 	return spToken, err
 }
 
@@ -54,8 +54,7 @@ func refreshContext(ctx context.Context, authorizer refreshable.Authorizer, log 
 
 // GetToken authenticates in the customer's tenant as the cluster service
 // principal and returns a token.
-func NewServicePrincipalToken(clientID, clientSecret, tenantID, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error) {
-	//func (tc *tokenClient) GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, tenantID string, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error) {
+func newServicePrincipalToken(clientID, clientSecret, tenantID, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error) {
 	conf := auth.ClientCredentialsConfig{
 		ClientID:     clientID,
 		ClientSecret: clientSecret,
@@ -73,7 +72,7 @@ func NewServicePrincipalToken(clientID, clientSecret, tenantID, aadEndpoint, res
 }
 
 // AuthenticateServicePrincipalToken authenticates in the customer's tenant as the cluster service principal and returns a token.
-func AuthenticateServicePrincipalToken(ctx context.Context, log *logrus.Entry, authorizer refreshable.Authorizer, timeout time.Duration) error {
+func authenticateServicePrincipalToken(ctx context.Context, log *logrus.Entry, authorizer refreshable.Authorizer, timeout time.Duration) error {
 	// during credentials rotation this can take time to propagate
 	// it is overridable so we can have unit tests pass/fail quicker
 	timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
--- a/pkg/util/aad/aad.go
+++ b/pkg/util/aad/aad.go
@@ -18,6 +18,10 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/refreshable"
 )
 
+const (
+	DefaultTimeout = 5 * time.Minute
+)
+
 type TokenClient interface {
 	GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, tenantID string, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error)
 }
@@ -28,9 +32,30 @@ func NewTokenClient() TokenClient {
 	return &tokenClient{}
 }
 
+func (tc *tokenClient) GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, tenantID, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error) {
+	spToken, err := NewServicePrincipalToken(clientID, clientSecret, tenantID, aadEndpoint, resource)
+	if err != nil {
+		return spToken, err
+	}
+
+	tokenAuthorizer := refreshable.NewAuthorizer(spToken)
+
+	err = AuthenticateServicePrincipalToken(ctx, log, tokenAuthorizer, DefaultTimeout)
+	return spToken, err
+}
+
+func refreshContext(ctx context.Context, authorizer refreshable.Authorizer, log *logrus.Entry) (bool, error) {
+	done, err := authorizer.RefreshWithContext(ctx, log)
+	if err != nil {
+		err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalCredentials, "properties.servicePrincipalProfile", "The provided service principal credentials are invalid.")
+	}
+	return done, err
+}
+
 // GetToken authenticates in the customer's tenant as the cluster service
 // principal and returns a token.
-func (tc *tokenClient) GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, tenantID string, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error) {
+func NewServicePrincipalToken(clientID, clientSecret, tenantID, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error) {
+	//func (tc *tokenClient) GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, tenantID string, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error) {
 	conf := auth.ClientCredentialsConfig{
 		ClientID:     clientID,
 		ClientSecret: clientSecret,
@@ -39,26 +64,28 @@ func (tc *tokenClient) GetToken(ctx context.Context, log *logrus.Entry, clientID
 		AADEndpoint:  aadEndpoint,
 	}
 
-	sp, err := conf.ServicePrincipalToken()
+	spToken, err := conf.ServicePrincipalToken()
 	if err != nil {
 		return nil, err
 	}
 
-	authorizer := refreshable.NewAuthorizer(sp)
+	return spToken, nil
+}
 
+// AuthenticateServicePrincipalToken authenticates in the customer's tenant as the cluster service principal and returns a token.
+func AuthenticateServicePrincipalToken(ctx context.Context, log *logrus.Entry, authorizer refreshable.Authorizer, timeout time.Duration) error {
 	// during credentials rotation this can take time to propagate
-	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	// it is overridable so we can have unit tests pass/fail quicker
+	timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
+	var err error
+	done := false
 	// NOTE: Do not override err with the error returned by
 	// wait.PollImmediateUntil. Doing this will not propagate the latest error
 	// to the user in case when wait exceeds the timeout
 	_ = wait.PollImmediateUntil(10*time.Second, func() (bool, error) {
-		var done bool
-		done, err = authorizer.RefreshWithContext(ctx, log)
-		if err != nil {
-			err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalCredentials, "properties.servicePrincipalProfile", "The provided service principal credentials are invalid.")
-		}
+		done, err = refreshContext(ctx, authorizer, log)
 		if !done || err != nil {
 			return false, err
 		}
@@ -84,9 +111,18 @@ func (tc *tokenClient) GetToken(ctx context.Context, log *logrus.Entry, clientID
 
 		return false, nil
 	}, timeoutCtx.Done())
+
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	return sp, nil
+	if !done && authorizer.LastError() != nil {
+		return api.NewCloudError(
+			http.StatusBadRequest,
+			api.CloudErrorCodeInvalidServicePrincipalCredentials,
+			"properties.servicePrincipalProfile",
+			"The provided service principal credentials are invalid.")
+	}
+
+	return nil
 }
--- a/pkg/util/aad/aad_test.go
+++ b/pkg/util/aad/aad_test.go
@@ -1,0 +1,55 @@
+package aad
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/golang/mock/gomock"
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	mock_refreshable "github.com/Azure/ARO-RP/pkg/util/mocks/refreshable"
+)
+
+func TestAuthenticateServicePrincipalToken(t *testing.T) {
+	tests := []struct {
+		name         string
+		wantErr      error
+		refreshError error
+	}{
+		{
+			name:         "Calling ServicePrincipal returns error.",
+			refreshError: errors.New("Some refesh error"),
+			wantErr: api.NewCloudError(
+				http.StatusBadRequest,
+				api.CloudErrorCodeInvalidServicePrincipalCredentials,
+				"properties.servicePrincipalProfile",
+				"The provided service principal credentials are invalid."),
+		},
+	}
+
+	mockController := gomock.NewController(t)
+	ctx := context.Background()
+	log := logrus.NewEntry(logrus.StandardLogger())
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			duration := 1 * time.Second
+			authorizer := mock_refreshable.NewMockAuthorizer(mockController)
+
+			authorizer.EXPECT().RefreshWithContext(gomock.Any(), gomock.Any()).Return(false, tt.refreshError).AnyTimes()
+
+			err := AuthenticateServicePrincipalToken(ctx, log, authorizer, duration)
+			if err != nil && tt.wantErr.Error() != err.Error() {
+				t.Errorf("AuthenticateServicePrincipalToken() error = %v, wantErr = %v", err.Error(), tt.wantErr.Error())
+				return
+			}
+		})
+	}
+}
--- a/pkg/util/aad/aad_test.go
+++ b/pkg/util/aad/aad_test.go
@@ -45,7 +45,7 @@ func TestAuthenticateServicePrincipalToken(t *testing.T) {
 
 			authorizer.EXPECT().RefreshWithContext(gomock.Any(), gomock.Any()).Return(false, tt.refreshError).AnyTimes()
 
-			err := AuthenticateServicePrincipalToken(ctx, log, authorizer, duration)
+			err := authenticateServicePrincipalToken(ctx, log, authorizer, duration)
 			if err != nil && tt.wantErr.Error() != err.Error() {
 				t.Errorf("AuthenticateServicePrincipalToken() error = %v, wantErr = %v", err.Error(), tt.wantErr.Error())
 				return
--- a/pkg/util/refreshable/refreshable.go
+++ b/pkg/util/refreshable/refreshable.go
@@ -16,12 +16,14 @@ import (
 type Authorizer interface {
 	autorest.Authorizer
 	RefreshWithContext(context.Context, *logrus.Entry) (bool, error)
+	LastError() error
 	OAuthToken() string
 }
 
 type authorizer struct {
 	autorest.Authorizer
-	sp *adal.ServicePrincipalToken
+	sp        *adal.ServicePrincipalToken
+	lastError error
 }
 
 // RefreshWithContext attempts to refresh a service principal token.  It should
@@ -54,20 +56,23 @@ type authorizer struct {
 // {"error":"invalid_client","error_description":"AADSTS7000215: Invalid client secret is provided.`
 // Once aad starts propagating credentials you might see occasional success in authentication.
 func (a *authorizer) RefreshWithContext(ctx context.Context, log *logrus.Entry) (bool, error) {
-	err := a.sp.RefreshWithContext(ctx)
-	if err != nil {
-		log.Info(err)
-
-		if !autorest.IsTokenRefreshError(err) ||
-			azureerrors.IsUnauthorizedClientError(err) ||
-			azureerrors.IsInvalidSecretError(err) {
-			return false, nil
-		}
-
-		return false, err
+	a.lastError = a.sp.RefreshWithContext(ctx)
+	if a.lastError == nil {
+		return true, nil
 	}
 
-	return true, nil
+	log.Info(a.lastError)
+
+	if !autorest.IsTokenRefreshError(a.lastError) ||
+		azureerrors.IsUnauthorizedClientError(a.lastError) ||
+		azureerrors.IsInvalidSecretError(a.lastError) {
+		return false, nil
+	}
+	return false, a.lastError
+}
+
+func (a *authorizer) LastError() error {
+	return a.lastError
 }
 
 func (a *authorizer) OAuthToken() string {
--- a/pkg/util/mocks/refreshable/refreshable.go
+++ b/pkg/util/mocks/refreshable/refreshable.go
@@ -36,6 +36,20 @@ func (m *MockAuthorizer) EXPECT() *MockAuthorizerMockRecorder {
 	return m.recorder
 }
 
+// LastError mocks base method.
+func (m *MockAuthorizer) LastError() error {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "LastError")
+	ret0, _ := ret[0].(error)
+	return ret0
+}
+
+// LastError indicates an expected call of LastError.
+func (mr *MockAuthorizerMockRecorder) LastError() *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "LastError", reflect.TypeOf((*MockAuthorizer)(nil).LastError))
+}
+
 // OAuthToken mocks base method.
 func (m *MockAuthorizer) OAuthToken() string {
 	m.ctrl.T.Helper()


kartikaysharma27:
LGTM

pkg/util/refreshable/refreshable.go
```diff
@@ -61,7 +61,7 @@ func (a *authorizer) RefreshWithContext(ctx context.Context, log *logrus.Entry)
 		if !autorest.IsTokenRefreshError(err) ||
 			azureerrors.IsUnauthorizedClientError(err) ||
 			azureerrors.IsInvalidSecretError(err) {
-			return false, nil
+			return false, err
```
ellis-johnson, line 64: I think this was not printed out in the first place to not print out the client secret, It would probably be better to just create a custom error here to inform SRE about the client secret being invalid rather than using the azure error

ellis-johnson (changes requested):
Small change requested

pkg/util/refreshable/refreshable.go
LiniSusan: What ever be the error returned from here, the customer can view only the message and type what we set here in the "NewCloudError" function:
done, err = authorizer.RefreshWithContext(ctx, log)
		if err != nil {
			err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalCredentials, "properties.servicePrincipalProfile", "The provided service principal credentials are invalid.")
		}
		
My question is about what are the possible scenarios that we capture from those function in the if condition

LiniSusan: another way of handling this can be check whether done is false or not then check whether err is nil or not, then apply the custom message for false and nil values, in the parent function.

ellis-johnson: Yep, I fully agree. Since this is only being used to call a generic error further upstream then this is okay.

ellis-johnson (dismissed):

s-amann (approved):
lgtm

pkg/util/refreshable/refreshable.go
darthhexx, line 64: Returning `false, err` is the same as what is being returned after the `if` statement block on line 67, which means the `if` statement isn't actually required to get the same return values from the function. However, the check is here for a reason, as detailed in the comments above the function, we need to retry certain failures to overcome AAD propagation delays.
So, while this change does work and the Invalid Service Principal error is returned it looks like we've lost retries as a result. Is that correct?

darthhexx (changes requested):

pkg/util/refreshable/refreshable.go
LiniSusan: Yes it is being same as the return statement outside the inner if. If we require the retries then we can return false, nil itself here and can make the changes in the file: /pkg/util/aad/aad.go#L48 by adding following code fragment:
if !done  && err == nil {
err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalCredentials, "properties.servicePrincipalProfile", "The provided secret is invalid. or Unauthorized client or Token Error")

// or any new errors 
}

is it fine?

pkg/util/refreshable/refreshable.go
```diff
@@ -59,8 +59,7 @@ func (a *authorizer) RefreshWithContext(ctx context.Context, log *logrus.Entry)
 		log.Info(err)
 
 		if !autorest.IsTokenRefreshError(err) ||
-			azureerrors.IsUnauthorizedClientError(err) ||
-			azureerrors.IsInvalidSecretError(err) {
```
m1kola, line 63: Do we need to update the docstring for this function?

pkg/util/aad/aad.go
```diff
@@ -39,15 +40,21 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
...
 	// NOTE: Do not override err with the error returned by
 	// wait.PollImmediateUntil. Doing this will not propagate the latest error
 	// to the user in case when wait exceeds the timeout
 	_ = wait.PollImmediateUntil(10*time.Second, func() (bool, error) {
 		var done bool
 		done, err = authorizer.RefreshWithContext(ctx, log)
 		if err != nil {
-			err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalCredentials, "properties.servicePrincipalProfile", "The provided service principal credentials are invalid.")
+			deadLine, _ := timeoutCtx.Deadline()
+			if azureerrors.IsInvalidSecretError(err) && !time.Now().After(deadLine.Add(-30*time.Second)) {
```
m1kola, line 51: Why 30 seconds?

pkg/util/aad/aad.go
```diff
@@ -39,15 +40,21 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
...
+				// Ignore the Invalid Secret errors until 30 seconds before the content timeout,
+				// because then we can hopefully assume it isn't caused by an AAD propagation delay
+				err = nil
+			} else {
+				err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalCredentials, "properties.servicePrincipalProfile", "The provided service principal credentials are invalid.")
+			}
```
m1kola, line 57: The ticket says:

> I want receive meaningful error when I provide invalid clientid/clientsecret for CSP instead of Internal Server error

I'm not sure that I understand:

1. Why we were getting internal server error in the first place? Looks like we were returning `CloudError` which should be propogated to the customers.
2. How this change helps to solve the issue?

m1kola (commented):
@LiniSusan This is a high risk change and a regresssion in this code might not be caught by E2Es due to the transient nature of the issue we are trying to solve.

Could you please make sure that цу have as much eyes on this PR as possible?

cc @ross-bryan @bennerv

pkg/util/aad/aad.go
LiniSusan: @m1kola : 
1. we were receiving internal server error because the function RefreshWithContext (https://github.com/Azure/ARO-RP/blob/master/pkg/util/refreshable/refreshable.go) returned nil for 3 types of errors. So it will return sp, nil instead of errors in the parent function GetToken(https://github.com/Azure/ARO-RP/blob/master/pkg/util/aad/aad.go) which leads to a generic error message with wrong sp (servicePrincipalToken)

2. when we return error from RefreshWithContext for secret invalid error also, it will skip the wait and return error immediately, so inorder to add a propagation delay; Ignored the Invalid Secret errors until 30 seconds before the content timeout,  so that we can assume it isn't caused by an AAD propagation delay. This was suggested by @darthhexx , please correct me if i m wrong in the explanation

pkg/util/refreshable/refreshable.go
m1kola: Also - we rely on this function in the following places:

* `pkg/util/aad/aad.go`
* `pkg/util/steps/refreshing.go`

This PR updates the first one. Asuming that the changes in `pkg/util/aad/aad.go` are correct, it looks like we are likely breaking `pkg/util/steps/refreshing.go` and everything which depends on it.

pkg/util/aad/aad.go
m1kola: > we were receiving internal server error because the function RefreshWithContext (https://github.com/Azure/ARO-RP/blob/master/pkg/util/refreshable/refreshable.go) returned nil for 3 types of errors. So it will return sp, nil instead of errors in the parent function GetToken(https://github.com/Azure/ARO-RP/blob/master/pkg/util/aad/aad.go) which leads to a generic error message with wrong sp (servicePrincipalToken)

Thanks for the explanation. I think I now have a better understanding, but the picture is not complete for me.

I'll repeat what you are saying just to make sure that we are on the same page:

1. [`GetToken`](https://github.com/Azure/ARO-RP/blob/93679fb7134afe26cd5daac527188df499551670/pkg/util/aad/aad.go#L23) calls `authorizer.RefreshWithContext(ctx, log)` inside `wait.PollImmediateUntil(...)`
2. `authorizer.RefreshWithContext(ctx, log)` in turn returns `done = false` and `err = nil`. This means that we hit `return` in this condition (it will be `return false, nil`):
    https://github.com/Azure/ARO-RP/blob/93679fb7134afe26cd5daac527188df499551670/pkg/util/aad/aad.go#L52-L54
3. The above return means that we try over and over again inside `wait.PollImmediateUntil` until we timeout. On the timeout, we hit this return:
    https://github.com/Azure/ARO-RP/blob/93679fb7134afe26cd5daac527188df499551670/pkg/util/aad/aad.go#L81
    Since there is no `err` (we deliberetely do not handle an error from `wait.PollImmediateUntil` for propogation reasons).

I think you got the root cause correct, but I'm curious - what happens next? if from `GetToken` we return `err = nil`, then what exactly fails and leads to 500/internal server error?

m1kola: Will something like this be better?

* Keep `RefreshWithContext` as is
* In `GetToken` (do we need to do the same enywhere else?) try to refresh with `wait.PollImmediateUntil`.
* Once we exit `wait.PollImmediateUntil` and if there are no errors - try to refresh one last time. In this case - if it errors - then it is for real and not due to propogation delay.

What do you think?

m1kola (changes requested):
I think we are probably breaking logic for `pkg/util/steps/refreshing.go`. See my comment in `pkg/util/steps/refreshing.go`.

pkg/util/aad/aad.go
```diff
@@ -39,12 +39,11 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
 	// during credentials rotation this can take time to propagate
 	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
 	defer cancel()
-
+	var done bool
```
facchettos, line 42: wouldn't that be a little better with `done := false`

facchettos (changes requested):
left a few small improvements suggestions. 
Also wouldn't it be better to remove the anonymous function and put it into an actual function? That would make it cleaner 

And also, it would be great to have this package tested

pkg/util/refreshable/refreshable.go
```diff
@@ -63,13 +64,17 @@ func (a *authorizer) RefreshWithContext(ctx context.Context, log *logrus.Entry)
...
-
 		return false, err
 	}
 
 	return true, nil
 }
 
+func (a *authorizer) RefreshWithContextRawError(ctx context.Context, log *logrus.Entry) error {
+	err := a.sp.RefreshWithContext(ctx)
+	return err
```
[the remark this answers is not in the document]
darthhexx: +1

darthhexx (dismissed):

pkg/util/refreshable/refreshable.go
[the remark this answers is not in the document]
LiniSusan: improvements done, thank you @facchettos

pkg/util/aad/aad.go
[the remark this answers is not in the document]
LiniSusan: done

pkg/util/aad/aad.go
LiniSusan: @m1kola : suggested changes are made, thank you.

LiniSusan:
@m1kola : New approach : Created a new func in refreshable.go (pkg/util/mocks/refreshable/refreshable.go) and kept the "RefreshWithContext" as it is. And called this new func after the poll wait in aad.go (pkg/util/aad/aad.go)

pkg/util/aad/aad.go
```diff
@@ -39,12 +39,11 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
 	// during credentials rotation this can take time to propagate
 	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
 	defer cancel()
-
+	done := false
 	// NOTE: Do not override err with the error returned by
 	// wait.PollImmediateUntil. Doing this will not propagate the latest error
 	// to the user in case when wait exceeds the timeout
 	_ = wait.PollImmediateUntil(10*time.Second, func() (bool, error) {
```
facchettos, line 46: I still think that this function should not be anonymous

facchettos (changes requested):

pkg/util/aad/aad.go
```diff
@@ -74,6 +73,14 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
 
 		return false, nil
 	}, timeoutCtx.Done())
+	if !done {
+		err = authorizer.RefreshWithContextRawError(ctx, log)
+		if err != nil {
+			err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", "The provided service principal generated an invalid token.")
```
m1kola, line 79: Why `api.CloudErrorCodeInvalidServicePrincipalToken` and invalid token message?

pkg/util/refreshable/refreshable.go
```diff
@@ -63,13 +64,16 @@ func (a *authorizer) RefreshWithContext(ctx context.Context, log *logrus.Entry)
...
 			return false, nil
 		}
-
 		return false, err
 	}
 
 	return true, nil
 }
 
+func (a *authorizer) RefreshWithContextRawError(ctx context.Context, log *logrus.Entry) error {
```
m1kola, line 73: Why do we need `log` here?

pkg/util/refreshable/refreshable.go
```diff
@@ -16,6 +16,7 @@ import (
 type Authorizer interface {
 	autorest.Authorizer
 	RefreshWithContext(context.Context, *logrus.Entry) (bool, error)
+	RefreshWithContextRawError(context.Context, *logrus.Entry) error
```
m1kola, line 19: Ah, right. When I [proposed this originally](https://github.com/Azure/ARO-RP/pull/2202#discussion_r927624201) it looks like I forgot about the fact that `RefreshWithContext` does the error skipping.

Any ideas on how we can make it work without an additional almost idential method?

m1kola (changes requested):
Left few comments.

Also - don't we have the same issue here?

https://github.com/Azure/ARO-RP/blob/e93e0b88d3021494db6992064baf2ed45be82066/pkg/util/steps/refreshing.go#L65-L85

pkg/util/refreshable/refreshable.go
facchettos: The problem here is that there are 2 different `RefreshWithContext` from what I can see . The one from the interface does more complicated things then the one from the `sp *adal.ServicePrincipalToken` (it calls it but does extra stuff with it). 

Possible solution would probably be to separate the call to `a.sp.RefreshWithContext` and add a function which does `error -> bool, error`. This way we don't need the `RefreshWithContextRawError`.

pkg/util/aad/aad.go
LiniSusan: @facchettos : we haven't touched that part for the new changes, this is something already exists.

pkg/util/aad/aad.go
```diff
@@ -74,6 +73,14 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
 
 		return false, nil
 	}, timeoutCtx.Done())
+	if !done {
```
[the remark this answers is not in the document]
LiniSusan: @m1kola : as of my understanding just returning error from GetToken function will not give customer the SP credentials invalid message.

pkg/util/aad/aad.go
LiniSusan: @m1kola : if error exists which means its a invalid/bad SP token

pkg/util/refreshable/refreshable.go
LiniSusan: yes log can be ignored, as we are not logging anything for the new function

pkg/util/aad/aad.go
facchettos: yes you have. you have take the `done` variable out of the function and this is dangerous and makes the code harder to understand.

pkg/util/refreshable/refreshable.go
LiniSusan: @facchettos : the problem here is RefreshWithContext should return nil on certain errors and we are trying to capture one among them to send proper error message to customer, otherwise it will end up in the generic error (Internal Server Error) which is hard to understand the issue.
also didnt get your solution with respect to the current scenario

pkg/util/aad/aad.go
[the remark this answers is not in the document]
m1kola: @LiniSusan yes, but `GetToken` can potentially return some other errors and we don't want to overshadow them. So we need to (in this order):

* Make sure that no error happened inside `wait.PollImmediateUntil`. 
    * If there was an explicit error - return it as is
    * If there was no explicit error, but `done` is `false` (as per [this explanation](https://github.com/Azure/ARO-RP/pull/2202#discussion_r927602737) it means we hit [this condition](https://github.com/Azure/ARO-RP/blob/e93e0b88d3021494db6992064baf2ed45be82066/pkg/util/aad/aad.go#L52-L54) and `return false, nil` from the condition func inside `wait.PollImmediateUntil`) - we need to make sure that it is actually an invalid token error. For that we try to refresh again without `wait.PollImmediateUntil` and additional error handling.

pkg/util/aad/aad.go
m1kola: I thought that we are trying to improve error propogation for the case where customer supplied invalid credentials. Hence `api.CloudErrorCodeInvalidServicePrincipalCredentials` should be used. Am I wrong?

pkg/util/aad/aad.go
petrkotas: The `_ = wait.PollImmediateUntil(10*time.Second, func() (bool, error)` can be changed to
`err = wait.PollImmediateUntil(10*time.Second, func() (bool, error)` and err can be used instead.

If you look how the poller works, it propagetes the error from internal function. It also add some timeout errors or context errors, but is what you want to use when propagating it out.

petrkotas (changes requested):
Hi @LiniSusan thank you for the work in the PR. Would you please take another look and streamline the changes made in the anonymous function to propage error? I am also in favor of Jeremys proposal on the Interface changes.

pkg/util/aad/aad.go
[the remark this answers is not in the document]
LiniSusan: @m1kola : as per the new change we are not changing the previously existing logic in  wait.PollImmediateUntil . In the GetToken function just calling the newly added function to check whether any SP error exists or not after the  wait.PollImmediateUntil , if the function RefreshWithContext is returning false as the first argument.
Also no change in RefreshWithContext function.

Reiterating: no change in existing logic,
changes are:
1. in GetToken: 
just checking the bool value returned from the function RefreshWithContext, if it is false, then it is due to SPToken error (sp *adal.ServicePrincipalToken), so sending that error note 
2. in pkg/util/refreshable/refreshable.go:
added a new function to return error if SPToken is invalid (return a.sp.RefreshWithContext(ctx))

pkg/util/aad/aad.go
LiniSusan: @m1kola : The scenario here is if the client mistakenly given wrong secret value they are getting generic error(Internal Server Error), which is hard to identify what is the mistake. They need a clear message so that they can fix the secret value from their end itself without asking for technical help.
We dont have much to do with error propogation in the existing logic.

pkg/util/aad/aad.go
LiniSusan: @petrkotas : I guess this change is out of scope of the ADO card related to this PR, also check the comments in before this anonymous func:

>   // NOTE: Do not override err with the error returned by
> // wait.PollImmediateUntil. Doing this will not propagate the latest error
> // to the user in case when wait exceeds the timeout 

@facchettos : without taking that done variable out, if you have any suggestion to solve this, please let us know

facchettos:
```go
        done := false 
	_ = wait.PollImmediateUntil(10*time.Second, func() (bool, error) {
		authorizer.RefreshAndParse(&done)
	}, timeoutCtx.Done())



func (authorizer authorizer) RefreshAndParse(done *bool ) (bool, error){

		*done, err = authorizer.RefreshWithContext(ctx, log)
		if err != nil {
			err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalCredentials, "properties.servicePrincipalProfile", "The provided service principal credentials are invalid.")
		}
		if !*done || err != nil {
			return false, err
		}

		p := &jwt.Parser{}
		claims := jwt.MapClaims{}
		_, _, err = p.ParseUnverified(authorizer.OAuthToken(), claims)
		if err != nil {
			err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", "The provided service principal generated an invalid token.")
			return false, err
		}

		for _, claim := range []string{"altsecid", "oid", "puid"} {
			if _, found := claims[claim]; found {
				return true, nil
			}
		}

		// populate err with a user-facing error that will be visible if we're
		// not successful.
		log.Info("token does not contain the required claims")
		err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalClaims, "properties.servicePrincipalProfile", "The provided service principal does not give an access token with at least one of the claims 'altsecid', 'oid' or 'puid'.")

		return false, nil
}
```

now you have something that is testable instead of having anonymous structs that are absolutely unseparable from the code. you could also mock that to test the bigger function.

pkg/util/aad/aad.go
LiniSusan: @m1kola : Yes you are right, it should be CloudErrorCodeInvalidServicePrincipalCredentials. will be updated in the final commit.

pkg/util/aad/aad.go
```diff
@@ -18,6 +18,41 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/refreshable"
 )
 
+func refreshContext(authorizer *refreshable.Authorizer, ctx *context.Context, done *bool, log *logrus.Entry) error {
```
facchettos, line 21: the context usually goes first https://pkg.go.dev/context, also there is no need to use a pointer there
Do you need a pointer to the authorizer ? It seems to be an interface, and the pointer seems unnecessary here

facchettos (changes requested):
It looks better now. It needs some small fixes on the usage of pointers and parameter order.

I also think that there should be tests.

pkg/util/aad/aad.go
```diff
@@ -18,6 +18,41 @@ import (
...
+	if err != nil {
+		err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalCredentials, "properties.servicePrincipalProfile", "The provided service principal credentials are invalid.")
+	}
+	if !*done || err != nil {
+		return err
+	}
+	return nil
+}
+
+func parseAuthToken(authorizer *refreshable.Authorizer, log *logrus.Entry) (bool, error) {
```
facchettos, line 33: here you don't need the pointer either

pkg/util/aad/aad.go
```diff
@@ -39,44 +74,28 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
...
+		if done {
+			done, err = parseAuthToken(&authorizer, log)
 		}
-
-		// populate err with a user-facing error that will be visible if we're
-		// not successful.
-		log.Info("token does not contain the required claims")
-		err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalClaims, "properties.servicePrincipalProfile", "The provided service principal does not give an access token with at least one of the claims 'altsecid', 'oid' or 'puid'.")
-
 		return false, nil
```
facchettos, line 87: actually this should probably be a single function, and it would make it slightly easier to test

pkg/util/aad/aad.go
```diff
@@ -39,44 +74,28 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
...
 	}, timeoutCtx.Done())
```
facchettos, line 88: actually this should probably be a single function, and it would make it slightly easier to test

pkg/util/aad/aad.go
facchettos, line 87: there seems to be something wrong here, before, we had cases were we returned errors, now we never do. I think you missed something

pkg/api/validate/openshiftcluster_validatedynamic.go
```diff
@@ -73,12 +73,20 @@ func (dv *openShiftClusterDynamicValidator) Dynamic(ctx context.Context) error {
...
-	token, err := aad.GetToken(ctx, dv.log, spp.ClientID, string(spp.ClientSecret), dv.subscriptionDoc.Subscription.Properties.TenantID, dv.env.Environment().ActiveDirectoryEndpoint, dv.env.Environment().ResourceManagerEndpoint)
+
+	spToken, err := aad.NewServicePrincipalToken(spp.ClientID, string(spp.ClientSecret), dv.subscriptionDoc.Subscription.Properties.TenantID, dv.env.Environment().ActiveDirectoryEndpoint, dv.env.Environment().ResourceManagerEndpoint)
+	if err != nil {
+		return err
+	}
+
+	tokenAuthorizer := refreshable.NewAuthorizer(spToken)
+
+	err = aad.AuthenticateServicePrincipalToken(ctx, dv.log, tokenAuthorizer, nil)
```
s-amann, line 84: Since we always are creating a new SPtoken and then authenticating with a new Authorizer we could encapsulate this logic to enforce all three operations happen when we create a new token. This would DRY up `pkg/api/validate/openshiftcluster_validatedynamic.go`, pkg/api/validate/dynamic/serviceprincipal.go`, and `pkg/cluster/serviceprincipal.go`

pkg/api/validate/openshiftcluster_validatedynamic.go
```diff
@@ -73,12 +73,20 @@ func (dv *openShiftClusterDynamicValidator) Dynamic(ctx context.Context) error {
...
 	if err != nil {
 		return err
 	}
 
-	spAuthorizer := refreshable.NewAuthorizer(token)
+	spAuthorizer := refreshable.NewAuthorizer(spToken)
```
s-amann, line 89: What is the difference between `spAuthorizer` and `tokenAuthorizer`? can we replace the use of `spAuthorizer` with the new `tokenAuthorizer`?

pkg/util/aad/aad.go
```diff
@@ -74,9 +95,21 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
 
 		return false, nil
```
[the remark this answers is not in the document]
LiniSusan: Hi Spencer, 
No it shouldn't be because we are not posting all the captured errors to customer, only certain errors we are posting to them and for others we send generic error message (Internal Server Error). If we return the error instead of nil, it will directly post the error to customer, which we do not want to do, that is why we are returning nil there instead of err, you can verify the scenario with the existing code in master branch.

pkg/api/validate/openshiftcluster_validatedynamic.go
LiniSusan: Hi @s-amann ,
Didn't change the existing logic in these area, its always creating a new token and authenticating it (you can check with master branch).
Do we really need to change that as part of this requirement and scope?

LiniSusan: Hi @s-amann ,
Did the changes, could you please review and let us know your comments?

facchettos (changes requested):

pkg/util/aad/aad_test.go
```diff
@@ -0,0 +1,134 @@
...
+			},
+			wantErr: nil,
+		},
+		{
+			name: "RefreshWithContext with errors",
+			args: args{
+				authorizer: mock_refreshable.NewMockAuthorizer(mockController),
+				mocksetup: func(refreshauth refreshable.Authorizer) {
+					refreshauth.(*mock_refreshable.MockAuthorizer).EXPECT().RefreshWithContext(gomock.Any(), gomock.Any()).Return(false, errors.New("Azure error")).AnyTimes()
+				},
```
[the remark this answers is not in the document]
LiniSusan: @facchettos : Seems like this is the pattern that we follow, could you please refer below test cases:

- pkg/util/instancemetadata/prod_test.go
- pkg/util/steps/runner_test.go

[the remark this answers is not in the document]
facchettos: I know we are using that pattern some other places. But it isn't a great pattern, so since this is a new file and we are creating the tests, it is better not to use it. Table driven tests are great because we can feed parameters and see in one glance what the test parameters are, this pattern is actually breaking that and this is why we should not use it for new test. 

Here it is bad to use this because it allows for subtest that are widely different from each other, and that should be in different tests. Also, this is duplicating code and now if we are to add new parameters or anything, we need to change all of the functions, while we wouldn't have to do it if that part was directly in the test and we were feeding the parameters to it. Code should only contain the logic, structs are for data. 

Also even though I think you should remove the functions altogether, I am not sure why your functions are using `refreshable.Authorizer` but you only give them `*mock_refreshable.MockAuthorizer` as parameters, and then always typecast them back to the latter.

facchettos:
btw, since this PR has been open for a long time, it would be great to rebase on master before merging because as of today it is 117 commits behind master

pkg/util/refreshable/refreshable.go
```diff
@@ -63,13 +64,16 @@ func (a *authorizer) RefreshWithContext(ctx context.Context, log *logrus.Entry)
...
 		return false, err
 	}
 
 	return true, nil
 }
 
+func (a *authorizer) RefreshWithContextRawError(ctx context.Context) error {
+	return a.sp.RefreshWithContext(ctx)
+}
+
```
facchettos, line 76: I stand by my previous comments, I don't think this should be separated from the other method. we should reduce the scope of the original authorizer.RefreshWithContext and add a function to go from `error` to `bool, error`. 

Tbh I think the whole interface is useless because we only use simple functions from the `*adal.ServicePrincipalToken` and we could probably just use that instead. not necessarily your doing but i'd rather not pile up and add more methods to something like this.

facchettos (changes requested):

pkg/util/refreshable/refreshable.go
```diff
@@ -16,6 +16,7 @@ import (
 type Authorizer interface {
 	autorest.Authorizer
 	RefreshWithContext(context.Context, *logrus.Entry) (bool, error)
+	RefreshWithContextRawError(context.Context) error
 	OAuthToken() string
 }
```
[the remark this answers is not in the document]
darthhexx: This PR are not intended to refactor the AAD authentication module, but to add an error message for the client provided credentials. Your outline above looks like a good path forward at first glance, but I would say it needs to be captured in an ADO card and assigned for someone to pickup in the next sprint.

pkg/util/refreshable/refreshable.go
darthhexx: @facchettos this change was made [at Mikalai's request](https://github.com/Azure/ARO-RP/pull/2202#discussion_r927624201) and is the simplest way to achieve this without breaking the existing pattern. I agree completely that this needs to be refactored, but can be done in a follow up where we address the overall design.

pkg/util/aad/aad.go
```diff
@@ -29,33 +56,39 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
...
+// AuthenticateServicePrincipalToken authenticates in the customer's tenant as the cluster service principal and returns a token.
+func AuthenticateServicePrincipalToken(ctx context.Context, log *logrus.Entry, authorizer refreshable.Authorizer, timeout *time.Duration) error {
 	// during credentials rotation this can take time to propagate
-	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	// it is overridable so we can have unit tests pass/fail quicker
+	ctxTimeout := DEFAULT_TIMEOUT
+	if timeout != nil {
+		ctxTimeout = *timeout
+	}
+	timeoutCtx, cancel := context.WithTimeout(ctx, ctxTimeout)
```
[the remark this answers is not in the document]
LiniSusan: @facchettos :
this cannot be done because our default timeout value is set to 5 minutes, when it comes to the test case the function will wait there till the default time which will cause to test case failure.

pkg/util/aad/aad.go
```diff
@@ -18,9 +18,36 @@ import (
...
+
+func refreshContext(ctx context.Context, authorizer refreshable.Authorizer, log *logrus.Entry) (bool, error) {
+	done, err := authorizer.RefreshWithContext(ctx, log)
+	if err != nil {
+		err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalCredentials, "properties.servicePrincipalProfile", "The provided service principal credentials are invalid.")
+	}
+	if !done || err != nil {
+		return false, err
+	}
+	return done, nil
```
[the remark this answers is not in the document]
darthhexx: Nope, those are not logically the same.

If done := true and err := error, then we return with `false, error` and not `true, error`.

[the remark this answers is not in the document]
LiniSusan: @darthhexx : yes you are correct, logically they are different, but in our case we have only three cases
true nil -- best case
false nil -- special errors
false error -- other errors
so i think we can go with @facchettos 's suggestion, which will not impact the expected result.

pkg/util/refreshable/refreshable.go
facchettos: I stand by my comments. `RefreshWithContext` has the same name than `sp.RefreshWithContext` but has a different logic, while `RefreshWithContextRawError` has a different name but the same logic than `sp.RefreshWithContext`

having  it this way would make more sense:

```go
func (a *authorizer) RefreshWithContext(ctx context.Context, log *logrus.Entry) (error) {
	return a.sp.RefreshWithContext(ctx)
}
```

```go 
func RefreshIsDone(log *logrus.Entry, err error) (bool, error) {
	if err != nil {
		log.Info(err)

		if !autorest.IsTokenRefreshError(err) ||
			azureerrors.IsUnauthorizedClientError(err) ||
			azureerrors.IsInvalidSecretError(err) {
			return false, nil
		}
		return false, err
	}

	return true, nil
}
```

facchettos: then we don't need to have the `RefreshWithContextRawError` method on the interface, which is making it more complicated than it has to be

facchettos: Once this is done, we should be able to do the changes that I outlined below without further modifications

pkg/util/aad/aad.go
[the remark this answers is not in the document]
facchettos: :)

pkg/util/aad/aad.go
```diff
@@ -65,14 +62,10 @@ func NewServicePrincipalToken(clientID, clientSecret, tenantID, aadEndpoint, res
...
-func AuthenticateServicePrincipalToken(ctx context.Context, log *logrus.Entry, authorizer refreshable.Authorizer, timeout *time.Duration) error {
+func AuthenticateServicePrincipalToken(ctx context.Context, log *logrus.Entry, authorizer refreshable.Authorizer, timeout time.Duration) error {
 	// during credentials rotation this can take time to propagate
 	// it is overridable so we can have unit tests pass/fail quicker
-	ctxTimeout := DEFAULT_TIMEOUT
-	if timeout != nil {
-		ctxTimeout = *timeout
-	}
-	timeoutCtx, cancel := context.WithTimeout(ctx, ctxTimeout)
+	timeoutCtx, cancel := context.WithTimeout(ctx, DefaultTimeout)
```
facchettos, line 68: this should be `timeoutCtx, cancel := context.WithTimeout(ctx, timeout)`

facchettos (changes requested):

pkg/util/refreshable/refreshable.go
facchettos:
```go
func RefreshWithContextOldLogic(a Authorizer, ctx context.Context) (book, error) {
    err := a.RefreshWithContext(ctx)
    return RefreshIsDone(err)
}
``` 
to avoid duplication where the current `a.RefreshWithContext` is used

pkg/api/validate/openshiftcluster_validatedynamic.go
LiniSusan: Not required, because that code is no longer changed in the PR with new changes

petrkotas (changes requested):
Thanks @LiniSusan I have one very simple comment. Other than that, I believe we are good.

m1kola (commented):
I didn't manage to spot any serious issues, but the error flow is very complicated. This is very hard to follow and I can't think of a way to simplify this right now.

Is there a way to cover it with E2E or integration tests in addition to unit tests?

pkg/util/aad/aad.go
```diff
@@ -28,9 +32,30 @@ func NewTokenClient() TokenClient {
...
+		err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalCredentials, "properties.servicePrincipalProfile", "The provided service principal credentials are invalid.")
+	}
+	return done, err
+}
+
 // GetToken authenticates in the customer's tenant as the cluster service
 // principal and returns a token.
-func (tc *tokenClient) GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, tenantID string, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error) {
+func NewServicePrincipalToken(clientID, clientSecret, tenantID, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error) {
+	//func (tc *tokenClient) GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, tenantID string, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error) {
```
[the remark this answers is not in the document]
LiniSusan: Done the changes

pkg/util/aad/aad.go
```diff
@@ -18,6 +18,10 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/refreshable"
 )
 
+const (
+	DefaultTimeout = 5 * time.Minute
```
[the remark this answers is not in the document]
LiniSusan: Changed the names to non-exported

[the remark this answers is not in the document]
ellis-johnson: https://issues.redhat.com/browse/ARO-2301 For new task

petrkotas (approved):

pkg/util/refreshable/refreshable.go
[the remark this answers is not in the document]
ellis-johnson: Task covered in https://issues.redhat.com/browse/ARO-2524

PR #2203, opened 2022-06-23
Clear the auth files when the pipeline exits

bennerv:

We used to clear the auth files for docker/podman on exit, but with RHEL8 we switched to podman and it stores it's auth file in a separate location that should be cleared.  

### Test plan for issue:
Run e2e a few times

nah

### DIFF
--- a/.pipelines/templates/template-job-deploy-azure-env-tag.yml
+++ b/.pipelines/templates/template-job-deploy-azure-env-tag.yml
@@ -36,7 +36,7 @@ jobs:
               azureDevOpsJSONSPN: $(aro-v4-ci-devops-spn)
           - script: |
               set -e
-              trap 'set +e; for c in $(docker ps -aq); do docker rm -f $c; done; docker image prune -af ; rm -rf ~/.docker/config.json' EXIT
+              trap 'set +e; for c in $(docker ps -aq); do docker rm -f $c; done; docker image prune -af ; rm -rf ~/.docker/config.json; rm -rf /run/user/$(id -u $USERNAME)/containers/auth.json' EXIT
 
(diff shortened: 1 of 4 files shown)


.pipelines/templates/template-job-deploy-azure-env-tag.yml
```diff
@@ -36,7 +36,7 @@ jobs:
               azureDevOpsJSONSPN: $(aro-v4-ci-devops-spn)
           - script: |
               set -e
-              trap 'set +e; for c in $(docker ps -aq); do docker rm -f $c; done; docker image prune -af ; rm -rf ~/.docker/config.json' EXIT
+              trap 'set +e; for c in $(docker ps -aq); do docker rm -f $c; done; docker image prune -af ; rm -rf ~/.docker/config.json; rm -rf /run/user/1000/containers/auth.json' EXIT
```
25region, line 39: Does $USERNAME work instead of 1000?

bennerv: no, $USERNAME is `cloud-user` which is not the same path.

25region: oh, it's not inside the container. Gotcha.

bennerv: but may want to do a `rm -rf /run/user/$(id $USERNAME)/containers/auth.json`

bennerv: Updated.

ross-bryan (approved):

PR #2205, opened 2022-06-24
[M6] Add an Admin API for OpenShiftVersions

hawkowl:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14651828

Adds an Admin API for OpenShiftVersions that can be used to investigate the database and update it (likely needs GenevaActions for that?)

### Test plan for issue:

Unit tests are added

N/A


m1kola:
Do we need a separate endpoint for that? Can't we use the existing one? Similar to get cluster.

hawkowl:
@m1kola I considered that, but since this endpoint has slightly different semantics (and no subscription, like the public API), a dedicated one here feels a bit better IMO.

pkg/frontend/admin_openshiftversion_put.go
```diff
@@ -0,0 +1,93 @@
...
+			osvd.OpenShiftVersion.Enabled = version.Enabled
+			return nil
+		})
+		if err != nil {
+			api.WriteError(w, http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "Internal server error.")
+			return
+		}
+	} else {
+		isCreate = true
+		returnDoc, err = f.dbOpenShiftVersions.Create(ctx, &api.OpenShiftVersionDocument{
```
darthhexx, line 71: There needs to be some input validate before this line, because as it stands we are able to PUT an empty `{}` and that results in an "OpenShiftVersion" with no `Version` that can never be removed. It also results in a panic in `getAdminOpenShiftVersions`, because the sort is not checking if the Version field is not nil.

We should make all Version, OpenShiftPullspec, and InstallerPullspec required during creation, otherwise we are able to PUT a version that has no image and/or installer pull spec, which is not usable. That would require any calling code to perform the validation in N places, when we should really do it here.

For an update PUT (PATCH?), we could either require all or more concisely allow individual field values to be specified for OpenShiftPullspec, InstallerPullspec, and/or Enabled and only modify the fields that have been specified. As it is now, attempting to enable a version with `PUT {"version":"4.10.20","enabled":true}` results in the OpenShiftPullspec and InstallerPullspec values being removed.

pkg/frontend/admin_openshiftversion_put.go
```diff
@@ -0,0 +1,93 @@
...
+	if docs != nil {
+		for _, doc := range docs.OpenShiftVersionDocuments {
+			if doc.OpenShiftVersion.Version == version.Version {
+				foundDoc = doc
+				break
+			}
+		}
+	}
+
+	isCreate := false
```
darthhexx, line 57: Shouldn't we add a separate PATCH handler to cleanly separate creates and updates?

darthhexx (changes requested):
We also need to set the `DefaultTTL` in the [ARM template generation code](https://github.com/Azure/ARO-RP/blob/cae95089072542d26f24d40dfb61db5ac6c40593/pkg/deploy/generator/resources_rp.go#L1648) to `-1`, otherwise the versions are going to be auto-removed.

pkg/deploy/generator/resources_rp.go
```diff
@@ -1645,7 +1645,6 @@ func (g *generator) database(databaseName string, addDependsOn bool) []*arm.Reso
 							},
 							Kind: mgmtdocumentdb.PartitionKindHash,
 						},
-						DefaultTTL: to.Int32Ptr(7 * 86400), // 7 days
```
darthhexx, line 1648: The OpenShiftVersions database has already been deployed in production with a setting of 7 days expiry therefore, due to the default incremental behaviour of ARM templates, we need to specify `-1` here to [switch to infinite TTL](https://docs.microsoft.com/en-us/azure/cosmos-db/sql/time-to-live#time-to-live-for-containers-and-items). Otherwise they will be left at 7 days afaik.

darthhexx (changes requested):
Thanks for adding the validation and making the fields mandatory. There is one change for the DB entries TTL and then we'd be good to go.

darthhexx (approved):
Thanks for making the changes, lgtm.

PR #2209, opened 2022-06-27
Generate hive clients inside the RP

hawkowl:

Maybe part of M4

Rather than depending on hive/pkg/clientset, which includes dependencies on the whole installer (thus solving none of the goals of M5 for tech debt :) ), we can depend on the API definition and build our own client set. 

### Test plan for issue:

dunno right now

N/A


s-amann (approved):
LGTM (I did not review `vendor/` and only spot checked some of the generated code in `pkg/hive/clientset/versioned/`)

m1kola (changes requested):
I'm not in favour of making hive client generation our job.

> Rather than depending on hive/pkg/clientset, which includes dependencies on the whole installer

Could you please share more details?

I think something is not right with this statement. I'm pretty sure that in this PR you generate the same code as we have in `github.com/openshift/hive/pkg/client/clientset`. So if `github.com/openshift/hive/pkg/client/clientset` has an installer dependency - than this means that our generated clientset also has an installer dependency. Or am I wrong? But I believe that `github.com/openshift/hive/pkg/client/clientset` has no installer dependency: I did grep of `github.com/openshift/hive/pkg/client` and I see no refrences to the installer.

Or are you talking about the fact that `github.com/openshift/hive` as a whole has a dependency on the installer? If so - then I think it is not a problem: all we need from the hive repo is a clientset and its dependencies (`github.com/openshift/hive/apis` mostly). If, as part of M5, we manage to get rid of direct installer dependency - go mod won't even vendor installer depsite the fact that `github.com/openshift/hive` has it in the `go.mod` as there is no code path in `github.com/openshift/hive/pkg/client` which imports the installer as far as I see.

Or am I missing something?

hawkowl:
@m1kola As an aside, Hive isn't going to be keeping pkg/client for much longer: https://issues.redhat.com/browse/HIVE-1950

hawkowl:
@m1kola it's that the package has that dependency, we still need to have a bunch of junk in our go.mod. For example, in a test project with this:

```
package main

import (
	hivev1 "github.com/openshift/hive/pkg/client/clientset/versioned/typed/hive/v1"
)

func main() {
	hivev1.NewForConfig(nil)
}
```

And a go.mod file that is this:

```
module github.com/hawkowl/hivetest

go 1.17

require (
    github.com/openshift/hive v1.1.16
	github.com/openshift/hive/apis v0.0.0
)

replace (
github.com/openshift/hive/apis => github.com/openshift/hive/apis v0.0.0-20220513200946-2a02c5dfd7fe
)
```

Then if I run `go mod tidy` (with or without `-compat=1.17`):

```
~/c/test [1] $ go mod tidy -compat=1.17                                                                                                             12:44:54
go: github.com/openshift/hive@v1.1.16 requires
        sigs.k8s.io/cluster-api-provider-aws@v0.0.0: reading sigs.k8s.io/cluster-api-provider-aws/go.mod at revision v0.0.0: unknown revision v0.0.0
```

If I add the nightmare of `replace` directives from ARO-RP's go.mod, I can then run `go mod tidy -compat=1.17` (the compat tag just means that it does the smarter tree resolution rather than 1.16s, because it will be compatible with 1 version behind), which then fails on:

```
go: github.com/openshift/hive@v1.1.16 requires
        github.com/openshift/installer@v0.9.0-master.0.20210211002944-d237b9dee575 requires
        github.com/terraform-providers/terraform-provider-aws@v1.60.1-0.20211207215258-c9b9d0b9dc46: invalid version: unknown revision c9b9d0b9dc46
```

So, even if it doesn't vendor all of the installer's deps since they're not strictly required, we _do_ have to keep the giant mess of replace directives to keep the installer happy, which is what I'd like to avoid :)

petrkotas (changes requested):
I am honestly not i favor of hard include of dependencies to our codebase. I would much rather see parts being removed rather than added. From the thread I understand we can use the controller-runtime directly https://coreos.slack.com/archives/CE3ETN3J8/p1656434003100069?thread_ts=1656378989.073059&cid=CE3ETN3J8 which is something I would rather see.

I fully understand that this would be the first time ARO does this compared to clientsets used everywhere, but the alternative is to manually maintain hard copy of Hives client files.

m1kola:
@hawkowl thanks for the example. I see what you mean now. I agree with you that having to satisfy indirect dependencies is not fun (especially installer ones).

> As an aside, Hive isn't going to be keeping pkg/client for much longer: https://issues.redhat.com/browse/HIVE-1950

This decision means that every codebase that consumes hive clients will be forced to generate clients themselves. It will be ARO as well as, I think, OSD, OCM and probably more. There are also outside consumers of this package (based on a quick github search).

For ARO specifically, I think it is a slippery slope: if we start generating hive client today - should we also start generating everything which is not in `k8s.io/client-go` and `github.com/openshift/client-go`? For example:
* We import `github.com/openshift/machine-config-operator/pkg/generated/clientset/versioned`. Should we generate it as well?
* We also used to use  `github.com/openshift/machine-api-operator/pkg/generated/clientset/versioned`
* I think it is likely we will be importing more clients which are not in client-go. 

IMO - we need the same approach for these clients. Or these clients should make it easier to consume them (I would prefer way).

hawkowl:
Since it's not entirely clear if we want to use the clientsets long term or use the c-r code, I have made #2220 which just vendors in the existing clientsets so that we can get stuff going, and when we get rid of the rest of the installer dependencies we can look at the c-r rewrite for this code.

Closing in preference to #2220.

PR #2210, opened 2022-06-27
Added cluster install time metrics

ArrisLee:

https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/6470470

Enable SRE to view how long ARO cluster creations are taking across the fleet, together with the time consumption of each install step.

### Test plan for issue:

Unit tested

N/A


pkg/cluster/cluster.go
```diff
@@ -56,6 +57,7 @@ type manager struct {
 	subscriptionDoc   *api.SubscriptionDocument
 	fpAuthorizer      refreshable.Authorizer
 	localFpAuthorizer refreshable.Authorizer
+	me                metrics.Emitter
```
s-amann, line 60: nit: please consider a more descriptive name such as `metricsEmitter`

pkg/cluster/cluster.go
```diff
@@ -97,7 +99,7 @@ type manager struct {
 
 // New returns a cluster manager
 func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database.OpenShiftClusters, dbGateway database.Gateway, aead encryption.AEAD,
-	billing billing.Manager, doc *api.OpenShiftClusterDocument, subscriptionDoc *api.SubscriptionDocument) (Interface, error) {
+	billing billing.Manager, doc *api.OpenShiftClusterDocument, subscriptionDoc *api.SubscriptionDocument, metics metrics.Emitter) (Interface, error) {
```
s-amann, line 102: nit: typo `metics` -> `metrics`

pkg/cluster/install_test.go
```diff
@@ -30,6 +31,23 @@ import (
...
 func failingFunc(context.Context) error { return errors.New("oh no!") }
 
+func normalFunc(context.Context) error {
+	time.Sleep(1 * time.Second)
+	return nil
+}
+
+type fakeMetricsEmitter struct {
+	Topic      string
+	IntallTime int64
```
s-amann, line 41: nit: typo `InstallTime` -> `InstallTime`

s-amann (changes requested):
just a few typos/naming suggestions, otherwise this lgtm

ArrisLee:
PR updated as required

s-amann (approved):
lgtm

ArrisLee:
Got further suggestions from @hawkowl will add time metrics for each step run

pkg/cluster/install_test.go
```diff
@@ -30,6 +31,23 @@ import (
 
 func failingFunc(context.Context) error { return errors.New("oh no!") }
 
+func normalFunc(context.Context) error {
+	time.Sleep(1 * time.Second)
```
hawkowl, line 35: I don't like sleeps in tests :)

Alternatively, perhaps this would be better:

runSteps takes two more arguments -- `startTime int64, currentTime func() int64` -- the former being the time the function was ran, and in the main code just `time.Now`. Then, in the test, you can have startTime be some number, and then currentTime just increment it every time it is called, and then you don't need to have a sleep :)

(alternatively, we could define some fake interface which does the same thing, that we can pass to Cluster, but I'm not sure if that's worth the effort unless we're using it elsewhere.)

hawkowl (dismissed):
see comment wrt tests

pkg/cluster/install_test.go
ArrisLee: @hawkowl Hi Amber, I did it in another way, added a `metricsDryrun` param in step run funcs, if it is set to `true`, the func will set static time value for each step run

pkg/util/steps/runner.go
```diff
@@ -21,19 +21,29 @@ func FriendlyName(f interface{}) string {
...
 		err := step.run(ctx, log)
 
 		if err != nil {
 			log.Errorf("step %s encountered error: %s", step, err.Error())
-			return err
+			return nil, err
 		}
+
+		endTime := time.Now()
+		if metricsDryrun {
```
s-amann, line 43: IMO we should not modify production code in such a way to only be used for testing. The only place I see `true` sent for `metricsDryrun` is in tests. While this fixes the use of sleep from before it now adds complexity throughout the code base as we need to pass `false` in many places where we want to run steps.

s-amann (changes requested):
I don't believe we should be adding a parameter throughout our actions to help a testing case. There are other suggestions about how to improve the tests without having to modify the production code flow.

pkg/util/steps/runner.go
SudoBrendan: I'm thinking dependency injection is probably the best answer... create an interface that `time` can match (and is used in prod), but for the test, you make an implementation that returns reasonable values for `Now` and other functions.

SudoBrendan (dismissed):
Great work so far on this, it's looking very good! I only want to +1 on Spencer's comment.

ArrisLee:
> Great work so far on this, it's looking very good! I only want to +1 on Spencer's comment.

Thanks for the feedback Brendon, your idea is exactly what I was considering at the first place, however the problem here is, some `time` related func in go, e.g., `time.Now()` are pkg based, not struct based, which IMO is not that easy to treated as an interface implementation?

pkg/util/steps/action.go
```diff
@@ -17,17 +17,31 @@ type actionFunction func(context.Context) error
...
+		metricsTopic: metricsTopic,
+	}
 }
 
+// name field is for better naming
+// when processing metrics emitting
 type actionStep struct {
-	f actionFunction
+	f            actionFunction
+	metricsTopic []string
```
s-amann, line 31: I'm not following why we have an array of string when we only consider the first element in it and we only ever pass one string into the `Action` function. Can this be changed to just accept a `metricsTopic string`? The same applies for `Condition` and `AuthorizationRefreshingAction`

pkg/cluster/install.go
```diff
@@ -142,63 +146,63 @@ func (m *manager) callInstaller(ctx context.Context) error {
...
+			steps.AuthorizationRefreshingAction(m.fpAuthorizer, steps.Action(m.validateResources), "validate_resources"),
+			steps.Action(m.ensureACRToken, "ensure_acr_token"),
+			steps.Action(m.ensureInfraID, "ensure_infra_id"),
+			steps.Action(m.ensureSSHKey, "ensure_ssh_key"),
+			steps.Action(m.ensureStorageSuffix, "ensure_storage_suffix"),
+			steps.Action(m.populateMTUSize, "populate_mtu_size"),
+
+			steps.Action(m.createDNS, "create_dns"),
+			steps.Action(m.initializeClusterSPClients, "initialize_cluster_sp_clients"), // must run before clusterSPObjectID
+			steps.Action(m.clusterSPObjectID, "cluster_sp_objectid"),
```
s-amann, line 158: nit: should this be `cluster_sp_object_id`?

pkg/util/steps/action.go
ArrisLee: @s-amann Hi Spencer, thanks for the feedback, the reason why I am using a slice here is to make the param `metricsTopic` optional, so we do not need to pass this param if it is not needed in some other places. If we make it as a `string` field, then we have to pass an empty string when we dun want to set any metrics topic for the certain actions.

s-amann: Ah, I see. Perhaps we should modify the comment above the functions to include `the metricsTopic parameter is optional and only the first element is considered if present` so the comment would eventually look like:
```
// Action returns a Step which will execute the action function `f`.
// The metricsTopic parameter is optional and only the first element is considered if present.
// Errors from `f` are returned directly.
func Action(f actionFunction, metricsTopic ...string) Step {
```
What do you think @ArrisLee ?

s-amann (approved):

pkg/cluster/install.go
```diff
@@ -27,10 +27,14 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
+func currentTime() time.Time {
+	return time.Now()
+}
+
```
bennerv, line 33: Should we follow a similar pattern to `pkg/frontend/frontend.go` -> `now` for consistency? 

```
now func() time.Time

File: pkg/util/steps/action.go [hunk 9]
```diff
@@ -17,17 +17,32 @@ type actionFunction func(context.Context) error
 
 // Action returns a Step which will execute the action function `f`. Errors from
 // `f` are returned directly.
-func Action(f actionFunction) Step {
-	return actionStep{f}
+// The metricsTopic parameter is optional and only the first element is considered if present.
+func Action(f actionFunction, metricsTopic ...string) Step {
```
bennerv, line 21: Why is `metricsTopic` a variadic parameter?  Is it to allow for zero to be specified (i.e. an optional parameter?

I would prefer to take the function and derive a name from said function rather than hardcoding strings above.  Something similar to the `friendlyName` function within steps.  Then we don't have to worry about passing in any data as it's already there within the function.

pkg/util/steps/condition.go
```diff
@@ -25,11 +25,13 @@ type conditionFunction func(context.Context) (bool, error)
 // out with a failure when more time than the provided timeout has elapsed
 // without f returning (true, nil). Errors from `f` are returned directly.
 // If fail is set to false - it will not fail after timeout.
-func Condition(f conditionFunction, timeout time.Duration, fail bool) Step {
+// The metricsTopic parameter is optional and only the first element is considered if present.
+func Condition(f conditionFunction, timeout time.Duration, fail bool, metricsTopic ...string) Step {
```
bennerv, line 29: Same comments as the `Action` function.

pkg/util/steps/refreshing.go
```diff
@@ -22,10 +22,12 @@ var ErrWantRefresh = errors.New("want refresh")
 // `authorizer` if the step returns an Azure AuthenticationError and rerun it.
 // The step will be retried until `retryTimeout` is hit. Any other error will be
 // returned directly.
-func AuthorizationRefreshingAction(authorizer refreshable.Authorizer, step Step) Step {
+// The metricsTopic parameter is optional and only the first element is considered if present.
+func AuthorizationRefreshingAction(authorizer refreshable.Authorizer, step Step, metricsTopic ...string) Step {
```
bennerv, line 26: Same comments as the `Action` function

bennerv (changes requested):
great work so far.  I think this can be a bit cleaner.  Please let me know if anything comments don't make sense and I can clarify.

pkg/util/steps/action.go
ArrisLee: @bennerv Thanks for the comment! Yes, `metricsTopic` is an optional param here, we just need to pass it wherever we need to assign a metrics topic to the action. For the conception of `friendlyName`, that is actually my first solution, but seems it is not working properly in some scenario, I got nil pointer issues when running unit test for the `runSteps` func. The idea of passing `metricsTopic` is based on the suggestions from Amber in slack channel: https://coreos.slack.com/archives/C02ULBRS68M/p1656557874242279?thread_ts=1656556973.718739&cid=C02ULBRS68M

ArrisLee:
> /azp run e2e

@SudoBrendan e2e test passed, are we good to merge it? it's been pending for a very long time :)

ArrisLee:
Updates:
- add a bool param in `steps.Run` function so we can controller if we need to emit the metrics or not, as for now, only steps in cluster installation phases need to be emitted.
- abandon the old metrics name param when init actions, conditions and refreshing actions, instead, create a name converting function based on the `FriendlyName()` functions so the changes will have less breaking impact on the existing code

pkg/util/steps/runner_test.go
```diff
@@ -341,3 +345,33 @@ func TestStepRunner(t *testing.T) {
...
 }
+
+func TestStepMetricsTopicNaming(t *testing.T) {
+	for _, tt := range []struct {
+		desc string
+		step Step
+		want string
+	}{
+		{
+			desc: "test action step naming",
```
hawkowl, line 356: What does the metric topic naming look like with anonymous functions such as `func(ctx context.Context) { return successfulFunc(ctx)} `, like what we use in `pkg/cluster/install.go`?

pkg/util/steps/action.go
```diff
@@ -17,17 +17,27 @@ type actionFunction func(context.Context) error
...
 // `f` are returned directly.
+// The metricsTopic parameter is optional and only the first element is considered if present.
 func Action(f actionFunction) Step {
-	return actionStep{f}
+	return actionStep{
+		f: f,
+	}
 }
 
+// name field is for better naming
```
hawkowl, line 27: no name field anymore

pkg/util/steps/action.go
```diff
@@ -17,17 +17,27 @@ type actionFunction func(context.Context) error
 
 // Action returns a Step which will execute the action function `f`. Errors from
 // `f` are returned directly.
+// The metricsTopic parameter is optional and only the first element is considered if present.
```
hawkowl, line 20: this doesn't seem to be around anymore

pkg/util/steps/refreshing.go
```diff
@@ -84,6 +86,13 @@ func (s authorizationRefreshingActionStep) run(ctx context.Context, log *logrus.
 		return true, err
 	}, timeoutCtx.Done())
 }
+
 func (s authorizationRefreshingActionStep) String() string {
 	return fmt.Sprintf("[AuthorizationRefreshingAction %s]", s.step)
 }
+
+func (s authorizationRefreshingActionStep) MetricsTopic() string {
+	trimedName := strings.ReplaceAll(s.step.String(), "[", "")
```
hawkowl, line 95: `s/trimed/trimmed/` here

pkg/util/steps/runner.go
```diff
@@ -17,23 +18,46 @@ func FriendlyName(f interface{}) string {
...
+		return fields[size-1]
+	}
+	return fullName
+}
+
 // Step is the interface for steps that Runner can execute.
 type Step interface {
 	run(ctx context.Context, log *logrus.Entry) error
 	String() string
+	MetricsTopic() string
```
hawkowl, line 38: does this have to be public?

hawkowl (approved):
Great! I think this will be very useful.

PR #2211, opened 2022-06-28
refactor: use already created friendlyName function

AldoFusterTurpin:
Co-authored-by: Petr Kotas <petrkotas@users.noreply.github.com>
Co-authored-by: Daniel Bizau <dbizau@users.noreply.github.com>

N/A

It adds calls to friendlyName function whenever possible as we were not following DRY. Export friendlyName of steps package as the behavior of this function was hardcoded in different places instead of calling the function. Now we just call this function whenever possible.

### Test plan for issue:

N/A

N/A


pkg/cluster/gatherlogs.go
```diff
@@ -6,9 +6,8 @@ package cluster
 import (
 	"context"
 	"encoding/json"
-	"reflect"
-	"runtime"
 
+	"github.com/Azure/ARO-RP/pkg/util/steps"
```
facchettos, line 10: need a new line here

AldoFusterTurpin: Thanks. Done!

s-amann (approved):
lgtm

PR #2212, opened 2022-06-28
added github action for go ci

facchettos:
https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14705826/

Moves units tests to github actions so we can see the results without having to connect to wvd or a vm.

### DIFF
--- a/.github/workflows/ci-go.yml
+++ b/.github/workflows/ci-go.yml
@@ -6,12 +6,9 @@ on:
       - v*
     branches:
       - master
-      - main
   pull_request:
 permissions:
   contents: read
-  # Optional: allow read access to pull request. Use with `only-new-issues` option.
-  # pull-requests: read
 
 jobs:
   ci-from-docker:
@@ -19,13 +16,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
-      - name: build from docker
-        uses: docker/build-push-action@v3
+      - name: build from buildah
+        uses: redhat-actions/buildah-build@v2
         with:
+          image: test-image
+          tags: ${{ github.sha }}
           context: .
-          push: false
-          file: "./Dockerfile.ci"
+          containerfiles: ./Dockerfile.ci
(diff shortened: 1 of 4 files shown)


facchettos:
this is currently blocked. we need some credentials on the repo settings to be able to fetch ubi-gotoolset

.github/workflows/ci-go.yml
```diff
@@ -0,0 +1,31 @@
+name: ci-go
+
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - main
```
s-amann, line 9: nit: we don't use main should we remove this until/if we do?

s-amann (dismissed):

karanmagdani (commented):
Thank you for the PR! I have a few clarifying questions. Does this mean that the test runs in both environments? 

If yes, What is the source of truth to merge a PR?

For example: In case the build fails in ADO but passes in Github Actions (or vice versa), would the merge action be blocked / allowed?

facchettos:
@karanmagdani yes, tests would run in both environment for now, and the authoritative test would be the one in ADO. It would be an interesting case however, since the code is the same, and the steps are the same.

karanmagdani:
> @karanmagdani yes, tests would run in both environment for now, and the authoritative test would be the one in ADO. It would be an interesting case if one failed and not the other however, since the code is the same, and the steps are the same.

I agree this *should* never happen from the PR side. But we might have other issues with CI-VMSS / ADO which could cause something like this to happen.

s-amann (approved):

Dockerfile.ci
```diff
@@ -0,0 +1,41 @@
...
+COPY .env .env
+COPY .sha256sum .sha256sum
+COPY .config .config
+
+RUN make generate
+RUN hack/git-utils/isClean.sh
+RUN make build-all
+RUN hack/git-utils/isClean.sh
+RUN make unit-test-go
+RUN hack/git-utils/isClean.sh
```
SudoBrendan, line 40: nit: I wonder if putting this into it's own hack script, with proper error messaging, would be helpful. Multiple RUN clauses reduces performance. It might be helpful to have a single script that we can run everywhere, so we know our pipelines are the same.

SudoBrendan (approved):
LGTM, one small nit if you'd like.

SudoBrendan (approved):

s-amann (approved):

petrkotas (approved):
Ok I believe it is time to test this one. Lets approve, add and see how usefultit is.

PR #2213, opened 2022-06-28
Refactor/simplify enable service endpoints

AldoFusterTurpin:
Co-authored-by: Petr Kotas <petrkotas@users.noreply.github.com>
Co-authored-by: Daniel Bizau <dbizau@users.noreply.github.com>

N/A

This PR refactors enableServiceEndpoints (pkg/cluster/storageaccounts.go) in order to simplify the code to be more readable, understandable and testeable.

Before this refactor, enableServiceEndpoints had some problems:
1-  was too large.
2- had too nested ifs and fors (High Cyclomatic Complexity).
3- had multiple responsibilities.
4- was not easy to understand.
5- the tests were not easy to follow and did not cover a lot of branches or errors. 

After this PR, a new Interface and Struct have been created to decouple that responsibility and to mock dependencies in pkg/cluster/enableendpoints.go (new file created replacing old content in storageaccount.go).

Also, the code is now more flexible as the "microsoftsEndpoints" that we want to enable are instantiated in the most outer function (enableServiceEndpoints) and are passed down as a parameter instead of being instantiated at a too low level down the code path (making it impossible to reuse the logic with other endpoints).

The search of an endpoint in a subnet has been improved as we can return true (exit the function) as soon as we encounter the endpoint (done now), instead of iterating all the service endpoints even if we have found it previously (done before this PR).

Last but not least, before this PR the code iterated over all the subnets and modified the endpoints in place and call m.subnet.CreateOrUpdate(), which is not ideal as the function is at the same time querying data and updating objects. After the refactor, we first obtain the subnets that have been updated with the new endpoints and then we call to m.subnet.CreateOrUpdate() for each of those subnets. 

### Test plan for issue:

Added unit tests for new and existing components.

No.


facchettos (changes requested):

pkg/cluster/storageaccounts.go
```diff
@@ -20,47 +20,42 @@ import (
...
-			return fmt.Errorf("WorkerProfile '%s' has no SubnetID; check that the corresponding MachineSet is valid", wp.Name)
+		if len(wp.SubnetID) < 1 {
+			return nil, fmt.Errorf("WorkerProfile '%s' has no SubnetID; check that the corresponding MachineSet is valid", wp.Name)
 		}
+		subnets = append(subnets, wp.SubnetID)
 	}
+	return subnets, nil
+}
 
+func (m *manager) enableServiceEndpointsOf(subnets []string, ctx context.Context) error {
```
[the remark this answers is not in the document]
AldoFusterTurpin: Yes, fixing. Tks.

pkg/cluster/storageaccounts.go
```diff
@@ -20,47 +20,42 @@ import (
 // enableServiceEndpoints should enable service endpoints on
 // subnets for storage account access
 func (m *manager) enableServiceEndpoints(ctx context.Context) error {
+	subnets, err := m.getSubnets()
+	if err != nil {
+		return err
+	}
+
+	err = m.enableServiceEndpointsOf(subnets, ctx)
```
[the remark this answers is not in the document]
AldoFusterTurpin: Renamed.

petrkotas (dismissed):
very nice

facchettos (approved):

ross-bryan (changes requested):
refactor is pretty straight forward, can we get some unit tests for the additional functions?

AldoFusterTurpin:
> refactor is pretty straight forward, can we get some unit tests for the additional functions?

@ross-bryan Thank you for your review!  🙂 You have brought an important point to the table with your question about test!👌 ...

Short answer: tests added!

Long answer:

IMHO it is a good practice to just test directly functions or methods that are exported ("public" in other programming languages) and the private ones should be tested indirectly by being called by the public ones. 

That's why I have now improved the design creating AND testing the new component (subnets.Updater Interface and UpdaterManager struct that implements that Interface). The function "enableServiceEndpoints" was doing too much things without delegating anything to others, so the code had bad cohesion. Now each component does one thing and we have splitted the responsability for better cohesion.

pkg/util/subnet/subnetsupdater.go
```diff
@@ -0,0 +1,74 @@
...
+		subnet.ServiceEndpoints = &[]mgmtnetwork.ServiceEndpointPropertiesFormat{}
+	}
+
+	serviceEndpoint := mgmtnetwork.ServiceEndpointPropertiesFormat{
+		Service:   to.StringPtr(endpoint),
+		Locations: &[]string{"*"},
+	}
+
+	*subnet.ServiceEndpoints = append(*subnet.ServiceEndpoints, serviceEndpoint)
+}
```
facchettos, line 74: why are these pointer functions ? They don't reference the object.

facchettos (changes requested):
some functions have unnecessary receivers. 
Otherwise lgtm

pkg/util/subnet/subnetsupdater.go
AldoFusterTurpin: Thanks for the review! 

I added them (function receivers) because I come from a more strict  OOP background and I wanted to "encapsulate" all that behavior inside the same struct.

But as you mention is not necessary, I have just removed them.  

Tks. 🙂

ross-bryan: They could've remained receiver functions, but they did not need to be pointer receiver functions. They could've been a normal receiver function that copies the struct by value instead of reference.

ross-bryan (approved):
lgtm, a couple comments/questions. nice work

pkg/util/subnet/subnetsupdater.go
facchettos: they could have, but should they have been ? we don't use the receiver fields/functions even once. it seems unnecessary to have a copy of an object we don't use

pkg/util/subnet/subnetsupdater.go
```diff
@@ -0,0 +1,74 @@
...
```
facchettos, line 74: Actually, you know what could be nice here ? having these functions be receiver functions on the subnet. 

for example :

```go
func (subnet mgmtnetwork.Subnet) containsEndpoint(endpoint string) (endpointFound bool, serviceEndpointPtr *mgmtnetwork.ServiceEndpointPropertiesFormat)
func (subnet *mgmtnetwork.Subnet) addEndpoint(endpoint string)
```
then you can even remove the pointer check on subnet for `containsEndpoint` because we are not using a pointer anymore

facchettos (changes requested):
there is some inconsistencies in your code regarding return values, sometimes they have names, sometimes they don't. it would be great to have it consistent. Since we don't really use those elsewhere, I'd rather remove them than add them all over. For example in `pkg/util/subnet/subnetsupdater.go:subnetContainsEndpoint` you have the named return value `endpointFound` which you use in the code (and don't have to instantiate it), but you're actually returning an hard coded value. And then for the other named return value, you don't actually use it. I think that this is dangerous to have things like this in the code. 

if you want more clarity on the return values I'd rather you write some comments like this `//funcsomething does something and returns this and that `

pkg/util/subnet/subnetsupdater.go
```diff
@@ -0,0 +1,74 @@
...
+		if !endpointFound {
+			addEndpointToSubnet(endpoint, subnet)
+			subnetChanged = true
+			continue
+		}
+
+		if serviceEndpointPtr.ProvisioningState != mgmtnetwork.Succeeded {
+			addEndpointToSubnet(endpoint, subnet)
+			subnetChanged = true
+		}
```
facchettos, line 42: I don't really get this part. It does the exact same thing, and the `continue` isn't doing anything. So i don't think it has any value here

facchettos: also, if it makes no sense adding an endpoint twice in the same subnet (given that these are domain names or something I would assume it does not), it would probably be better to have an idempotent `addEndpointToSubnet` function which return true if it was added rather than this one.

pkg/cluster/cluster.go
```diff
@@ -164,9 +167,11 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 
 		dns:     dns.NewManager(_env, localFPAuthorizer),
 		storage: storage,
-		subnet:  subnet.NewManager(_env.Environment(), r.SubscriptionID, fpAuthorizer),
+		subnet:  subnetManager,
```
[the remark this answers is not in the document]
AldoFusterTurpin: I needed to pass subnet manager to a struct I created but now I am not doing it that way and I forgot to revert the change. Changing it now.  Thanks!

pkg/cluster/cluster.go
```diff
@@ -133,6 +134,8 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 		}
 	}
 
+	subnetManager := subnet.NewManager(_env.Environment(), r.SubscriptionID, fpAuthorizer)
```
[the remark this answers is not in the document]
AldoFusterTurpin: I needed to pass subnet manager to a struct I created but now I am not doing it that way and I forgot to revert the change. Changing it now. Thanks!

pkg/util/subnet/subnetsupdater.go
AldoFusterTurpin: I replied too fast.

I agree it was a good idea but we can't do that as  "mgmtnetwork.Subnet" is not a struct we define, it is on "vendor/github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network/models.go" so we can't do it.

AldoFusterTurpin: I thought it was a good idea but I just realized that "mgmtnetwork.Subnet" is from "vendor/github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network/models.go" so I think it is not that easy to change in this case. 

If the struct was defined in this module (ARO-RP) the change would be easy and straightforward but knowing that is from another module I would not touch it for now. 

Right? 🙂

pkg/util/subnet/subnetsupdater.go
AldoFusterTurpin: "it makes no sense adding an endpoint twice in the same subnet"
I have not decided to do it that way 🙂

I agree with you that it is a bit strange, BUT this PR is a refactor and this is how it was done before. I am not sure if changing the behavior now is a good thing because if the code does not work once merged (as this is touching infrastructure), all the work done here will be reverted. That's why I would just ensure the refactor is good and in a future pull request we could change that behavior (because thanks to the refactor will be easy to change).

In the original code, you can check "pkg/cluster/storageaccounts.go:47-48" and you will see that it sets "found to true" when it both finds the serviceEndpoint and the state of the serviceEndpoint is == mgmtnetwork.Succeeded. If the endpoint is present but the state is != "mgmtnetwork.Succeeded", found will still be false and it will add the service endpoint to the subnet even if the service endpoint is already present. And that is what my code does ( on porpuse). 

A part from that, I think that the code block you have marked in this conversation can be simplified to
```go
if !endpointFound ||
			serviceEndpointPtr != nil && serviceEndpointPtr.ProvisioningState != mgmtnetwork.Succeeded {
			addEndpointToSubnet(endpoint, subnet)
			subnetChanged = true
		}

```
so will do it now

pkg/util/subnet/subnetsupdater.go
facchettos: yeah this is what I said in the edit afterwards. there isn't a good way to do it except wrapping it into an other struct. which is probably not worth it.

pkg/util/subnet/subnetsupdater.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	return updatedSubnets
+}
+
+func addEndpointsToSubnet(endpoints []string, subnet *mgmtnetwork.Subnet) (subnetChanged bool) {
+	for _, endpoint := range endpoints {
+		endpointFound, serviceEndpointPtr := subnetContainsEndpoint(subnet, endpoint)
+		if !endpointFound ||
+			serviceEndpointPtr != nil && serviceEndpointPtr.ProvisioningState != mgmtnetwork.Succeeded {
```
facchettos, line 34: you should put some parenthesis on line 34. it may be the same, but we want to keep the exact same statements and not chain them like this

pkg/util/subnet/subnetsupdater.go
AldoFusterTurpin: Either I answered previously to your edit or I missed it. But know we agree on that 🙂

pkg/util/subnet/subnetsupdater.go
AldoFusterTurpin: Parenthesis are not needed  (Unit Tests are covering this and also this is just an "or" following an "and") but I have added an explanatory variable for readability.

facchettos: this is not about readability this is about the logic. if you don't put the parenthesis it means that it is evaluated like this 
`(!endpointFound || serviceEndpointPtr != nil ) && serviceEndpointPtr.ProvisioningState != mgmtnetwork.Succeeded `
So yes the result is the same , but the logic isn't.

AldoFusterTurpin:
> there is some inconsistencies in your code regarding return values, sometimes they have names, sometimes they don't. it would be great to have it consistent. Since we don't really use those elsewhere, I'd rather remove them than add them all over. For example in `pkg/util/subnet/subnetsupdater.go:subnetContainsEndpoint` you have the named return value `endpointFound` which you use in the code (and don't have to instantiate it), but you're actually returning an hard coded value. And then for the other named return value, you don't actually use it. I think that this is dangerous to have things like this in the code.
> 
> if you want more clarity on the return values I'd rather you write some comments like this `//funcsomething does something and returns this and that `

In subnetsupdater.go file that you mention, ALL the methods that return values, use named return parameters, so I don't see the inconsistency there. 

I use them in the code when it makes sense: when they make the code more readable and clear. For me, this happens when we have a function/method that performs an action and at the same time returns one or more things (apart from an error, obviously). E.g, when using named return parameters you know what the function does + returns just by looking at the function header.

According to [Effective Go](https://go.dev/doc/effective_go#named-results) named parameters "are documentation". So I don't think removing named return parameters to add comments in the code is a good idea.

pkg/util/subnet/subnetsupdater.go
facchettos: > "it makes no sense adding an endpoint twice in the same subnet"
> I have not decided to do it that way slightly_smiling_face

I didn't say, "it makes no sense", it was a conditional statement. 

Your function names aren't chosen properly then. Because the current ones imply that in `addEndpointsToSubnet` is just calling `addEndpointToSubnet` for multiple endpoints.

facchettos: Also I missed it when you updated it, but when is the pointer nil and endpointFound==true ? 
Your pointer check is superfluous

AldoFusterTurpin: I agree on that, I will improve it with a Go comment now to clarify it.
 
Also, how should I name the functions?
"AddEndpointsToSubnets" -> "AddMissingAndInvalidEndpointsToSubnets" or "AddMissingAndNotSuccededEndpointsToSubnet" ? 

"addEndpointsToSubnet" -> "addMissingAndInvalidEndpointsToSubnet" or "addMissingAndNotSuccededEndpointsToSubnet"

Any suggestion is welcome.

Tks!

AldoFusterTurpin: > Also I missed it when you updated it, but when is the pointer nil and endpointFound==true ?
Your pointer check is superfluous

What do you mean? I need to do 
```go
serviceEndpointPtr != nil
```
if not, 
```go
serviceEndpointPtr.ProvisioningState
````
on line 33 can panic, no? Maybe I am missing something.

I think I am not checking for "pointer nil and endpointFound==true".

Comment (facchettos) [reply pkg/util/subnet/subnetsupdater.go]: A documentation comment is fine, and that would show up in function hovering if it starts with the function name. 

Basically , the pointer is nil only if `subnetContains... ` returns false. But if it returns false then go will stop at the first condition (which is true because it is `!containsEndpoint`), and will never evaluate the second value of the `or` statement https://go.dev/ref/spec#Logical_operators

Comment (AldoFusterTurpin) [reply pkg/util/subnet/subnetsupdater.go]: > Basically , the pointer is nil only if subnetContains... returns false. But if it returns false then go will stop at the first condition (which is true because it is !containsEndpoint), and will never evaluate the second value of the or statement

> Your pointer check is superfluous

You are wrong about not checking the pointer against nil. 

Facts (check the attached picture): I removed the nil check that you mention and tests executing that part fail exactly at line 33 (with the nil check that already exists the tests don't fail).

Why?
For the simple fact that here (line 33 of pkg/util/subnet/subnetsupdater.go):
```go
endpointSucceded := serviceEndpointPtr != nil && serviceEndpointPtr.ProvisioningState == mgmtnetwork.Succeeded
```
at exactly that point, you are not evaluating the if condition that you mention, that will come later. You just thought about what happens on the execution of the "if" on line 35
but you completely forgot that the nil check is necessary because at line 33 nothing guarantees you that serviceEndpointPtr is not nil. No change is needed.

tests_fail_without_checking_nil

facchettos: that's what you get for using a gazillion useless variables. it was working before when there was still a `continue` statement, so there is no reason it is not working now. 
please stop trying to reinvent the wheel.

it should be like this:
```go
func addEndpointsToSubnet(endpoints []string, subnet *mgmtnetwork.Subnet) (subnetChanged bool) {
	for _, endpoint := range endpoints {
		endpointFound, serviceEndpointPtr := subnetContainsEndpoint(subnet, endpoint)

		if !endpointFound || serviceEndpointPtr.ProvisioningState != mgmtnetwork.Succeeded {
			addEndpointToSubnet(endpoint, subnet)
			subnetChanged = true
		}
	}

	return subnetChanged
}
```

AldoFusterTurpin: > that's what you get for using a gazillion useless variables. 
No, you just proposed something that was not working either because you did not check the current status of the code at that point (your comment proposing a nonworking change was done 5 hours after the last push) or you made some assumptions that break the code as I explained in the previous comment.

"gazillion useless variables"-> this was just one explanatory variable to make the "if" easier to read. I will now remove the variable and just do an "or" as your last suggestion because now it does work and is okay for me.

Thanks! 🙂

pkg/cluster/enableendpoints.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	return subnets, nil
+}
+
+func (m *manager) getAllSubnets(ctx context.Context, subnetIds []string) ([]*mgmtnetwork.Subnet, error) {
+	if len(subnetIds) == 0 {
+		return nil, nil
+	}
+
+	subnets := make([]*mgmtnetwork.Subnet, len(subnetIds))
```
hawkowl, line 59: Can this code be moved into https://github.com/Azure/ARO-RP/blob/master/pkg/util/subnet/subnet.go ?

hawkowl (changes requested):
I don't quite understand why subnetsUpdater is not part of the existing subnets manager, from looking at the code. I've also left some comments.

pkg/cluster/enableendpoints.go
AldoFusterTurpin: Good idea! Tks. Done.

pkg/cluster/cluster.go
```diff
@@ -78,10 +78,11 @@ type manager struct {
...
 	rpPrivateLinkServices network.PrivateLinkServicesClient
 
-	dns     dns.Manager
-	storage storage.Manager
-	subnet  subnet.Manager
-	graph   graph.Manager
+	dns            dns.Manager
+	storage        storage.Manager
+	subnet         subnet.Manager
+	subnetsUpdater subnet.Updater
```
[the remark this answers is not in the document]
AldoFusterTurpin: I 100% agree with that 🙂

I will find one more appropriate and update the code.

pkg/cluster/enableendpoints.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return err
+	}
+
+	subnets, err := m.getAllSubnets(ctx, subnetIds)
+	if err != nil {
+		return err
+	}
+
+	updatedSubnets := m.subnetsUpdater.AddEndpointsToSubnets(api.SubnetsEndpoints, subnets)
```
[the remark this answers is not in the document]
AldoFusterTurpin: @hawkowl  Good question. I asked myself the same when doing the refactor.

The main ideas (all related) behind that are:
1 - ideally we just want to test public methods as this indicates a good design in terms of separation of concerns and testability (components easier to test and with less coupling to the implementation)

2 - It would be great to separate the business logic from communication with infrastructure (related to the layers of [Clean Architecture](https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html)). In other words: we don't want one component (struct in Go, object in a different OO language) be responsible for both:
- calculating which endpoints are new (or without succeded state) in the subnet
- call a component's method from the infrastructure layer to update the subnets in Azure Cloud (or anywhere from outside our domain)

And that is because if we have:
- a list of endpoints (that we want to make sure they are enabled in the subnets)
- a list of subnets that should include the previous endpoints in succeded state

we don't really need to call any infrastructure component to perform that business logic, we can do it by "ourselves" (from the point of view of the domain of our architecture) and return that new business "object": a list of updatedSubnets or subnetsThatShouldBeUpdated. 

This is also related to the principle of not modifying things while iterating over them. 

Instead of doing (before this PR) ❌:
- iterate over variables
- check for a condition in the variable
- depending on the previous condition do a call to update the subnet in Azure with this same component but updated

we are doing (after this PR)✅
- get components (in our case subnets) that should be updated (we don't need the infrastructure to perform the logic to know which ones are missing or aren't in succeded state after we have obtained all the subnets)
- iterate over all the previous components and call to "update subnet" in Azure

I 100% agree with you that "subnetsUpdater" is a confusing name, I will find one more appropriate and update the code. 

Thanks! 🙂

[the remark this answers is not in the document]
facchettos: > 1 - ideally we just want to test public methods as this indicates a good design in terms of separation of concerns and testability (components easier to test and with less coupling to the implementation)

what do you mean by that ?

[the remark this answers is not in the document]
AldoFusterTurpin: [ Disclaimer: This is just my opinion but some people have the same vision.  I would say it is a good practice]

I mean that, ideally, we should just test directly exported (or public in other languages) methods of structs. By directly I mean invoking it explicitly from a test function. 

This is really difficult to do as requires really high cohesive and decoupled components but is a thing we should try to achieve. 

The idea basically says that private methods should be tested indirectly by public methods. This is important because most of the time what changes most are unexported or private methods (the details of components) but the interface of those components (interface in the sense of public methods you can invoke on) are likely to be more stable than the details. 

That's why testing private methods have the implication of ending up with tests that are more coupled to internal details rather than behavior. Or put in different words: only testing directly exported methods you end up with tests less coupled to implementation details and more focused on behavior.

What is not a valid strategy is to just make a method exported because you want to test it. What should be done is to extract this behavior to a new component with high cohesion and test the public methods of the new component.

Maybe I am biased as I am a TDD advocate 👨‍💻

[into](https://twitter.com/unclebobmartin/status/1242655294368419840?lang=en), [info](https://anthonysciamanna.com/2016/02/14/should-private-methods-be-tested.html), [info](https://roadfiresoftware.com/2018/05/can-i-unit-test-private-methods/)

This is [not a strict norm](https://roadfiresoftware.com/2018/05/when-to-unit-test-private-methods/) but instead a good design philosophy.
🙂

pkg/cluster/enableendpoints_test.go
```diff
@@ -0,0 +1,295 @@
...
+				subnetsUpdaterMock.
+					EXPECT().
+					AddEndpointsToSubnets(endpoints, subnets).
+					Return(nil)
+
+				subnetManagerMock.
+					EXPECT().
+					CreateOrUpdate(gomock.Any(), gomock.Any(), gomock.Any()).
+					Times(0)
+			},
```
facchettos, line 84: This doesn't follow DRY. All of these functions do pretty much the same thing but they are just copy pasted and modified a little bit. It would be better to have a single function to which we pass arguments.

facchettos (changes requested):

pkg/cluster/enableendpoints_test.go
AldoFusterTurpin: I do understand what you mean but this is just done this way to prepare the expectations of the mocks. As every test case is different, I don't see how we could create an "enough general" function that can be used to prepare the mocks for all the test cases. 🤔

facchettos: either find the parameters that allow you to have a general function, either split your tests into more specialized one. setup functions like these do not belong in the test parameters

pkg/cluster/enableendpoints_test.go
```diff
@@ -0,0 +1,295 @@
...
+
+	tt := []test{
+		{
+			name: "should not call to subnetManager.CreateOrUpdate() because all endpoints are already in the subnet",
+			oc: &api.OpenShiftCluster{
+				Properties: api.OpenShiftClusterProperties{
+					MasterProfile: api.MasterProfile{SubnetID: subnetIdMaster},
+				},
+			},
+			setMocks: func(subnetManagerMock *mock_subnet.MockManager, subnetsUpdaterMock *mock_subnet.MockEndpointsAdder, tt test) {
```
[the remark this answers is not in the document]
AldoFusterTurpin: Improved the naming. Tks!

pkg/cluster/enableendpoints_test.go
AldoFusterTurpin: I have split the single Table Driven Test into multiple tests. Done.

facchettos (changes requested):
The interface on endpoints adder is not needed, and is actually making the thing needlessly complicated. We don't need to have it as an interface, since it is just a function that has no dependencies on external entities. Once the interface is removed, the testing will be easier to do because we won't have to set up as many mocks. 

As @hawkowl said, you could have it directly in createorupdate, though personally I would maybe have a wrapper around `createorupdate` with `addendpointstosubnets` followed by `createorupdate`.

pkg/util/subnet/endpointsadder.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+//  EndpointsAdder is the interface that wraps the basic logic to add endpoints to subnets
+//
+// AddEndpointsToSubnets adds the endpoints (that either are missing in subnets
+// or aren't in succeded state in subnets) to the subnets and returns those subnets.
+// Implementations of this method should not talk to infrastructure layer (e.g, Azure).
+// In contrast, the result of this function should be passed to a subnet manager to update the subnets
+// in Azure.
+type EndpointsAdder interface {
```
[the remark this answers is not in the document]
AldoFusterTurpin: IMHO even interfaces are used most of the time to avoid talking directly to external dependencies (DB, HTTP, CLI, etc.), they are also used as a design mechanism to decouple components (in this case "enableServiceEndpoints") from details ("AddEndpointsToSubnets"). 

If you don't have it as an interface, how can you ensure when testing pkg/cluster/enableendpoints.go "enableServiceEndpoints" function that "AddEndpointsToSubnets(...)" returns nil (or a specific value), and depending on that value "enableServiceEndpoints" behaves in a specific way? In other words, how can you test the behavior of "enableServiceEndpoints" without depending on the behavior of "EndpointsAdderManager" (implementation of the interface EndpointsAdder)?

Without the interface,  tests of "enableServiceEndpoints" depend on the behavior of "endpointsAdder" as it calls a specific method of a struct (and you are out of control of that).

In order to test "enableServiceEndpoints" independently from "EndpintsAdder" we need to create an interface to expose the behavior of "EndpointsAdder" but not the details. Without this interface, for example, how can we force AddEndpointsToSubnets to return nil?

pkg/util/subnet/endpointsadder.go
```diff
@@ -0,0 +1,84 @@
...
+	AddEndpointsToSubnets(endpoints []string, subnets []*mgmtnetwork.Subnet) (subnetsToBeUpdated []*mgmtnetwork.Subnet)
+}
+
+// EndpointsAdderManager is the basic implementation of EndpointsAdder interface.
+type EndpointsAdderManager struct {
```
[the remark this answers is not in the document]
AldoFusterTurpin: Agree on the improvement of the naming, changing it now. For the PS, check below, please.

pkg/cluster/enableendpoints_test.go
```diff
@@ -0,0 +1,328 @@
...
+	endpoints := []string{"Microsoft.ContainerRegistry", "Microsoft.Storage"}
+	subnets := []*mgmtnetwork.Subnet{subnet}
+
+	controller := gomock.NewController(t)
+	defer controller.Finish()
+
+	subnetManagerMock := mock_subnet.NewMockManager(controller)
+	subnetManagerMock.
+		EXPECT().
+		GetAll(gomock.Any(), subnetIds).
```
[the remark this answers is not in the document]
AldoFusterTurpin: Done.

pkg/cluster/enableendpoints_test.go
```diff
@@ -0,0 +1,328 @@
...
+		Return(subnets, nil)
+
+	subnetManagerMock.
+		EXPECT().
+		CreateOrUpdate(gomock.Any(), gomock.Any(), gomock.Any()).
```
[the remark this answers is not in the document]
AldoFusterTurpin: Done.

pkg/cluster/enableendpoints_test.go
```diff
@@ -0,0 +1,328 @@
...
+	subnets := []*mgmtnetwork.Subnet{subnet, secondSubnet}
+	endpoints := []string{"Microsoft.ContainerRegistry", "Microsoft.Storage"}
+
+	controller := gomock.NewController(t)
+	defer controller.Finish()
+
+	subnetManagerMock := mock_subnet.NewMockManager(controller)
+	subnetManagerMock.
+		EXPECT().
+		GetAll(gomock.Any(), subnetIds).
```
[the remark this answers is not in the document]
AldoFusterTurpin: Done.

pkg/cluster/enableendpoints_test.go
```diff
@@ -0,0 +1,328 @@
...
+		Return(subnets, nil)
+
+	subnetManagerMock.
+		EXPECT().
+		CreateOrUpdate(gomock.Any(), gomock.Any(), gomock.Any()).
```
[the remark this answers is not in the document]
AldoFusterTurpin: Done.

pkg/cluster/enableendpoints_test.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	controller := gomock.NewController(t)
+	defer controller.Finish()
+
+	subnetManagerMock := mock_subnet.NewMockManager(controller)
+
+	subnetManagerMock.
+		EXPECT().
+		Get(gomock.Any(), gomock.Any()).
```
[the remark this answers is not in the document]
AldoFusterTurpin: Done.

pkg/cluster/enableendpoints_test.go
```diff
@@ -0,0 +1,328 @@
...
+
+	subnetManagerMock := mock_subnet.NewMockManager(controller)
+	subnetManagerMock.
+		EXPECT().
+		GetAll(gomock.Any(), subnetIds).
+		Return(initialSubnets, nil)
+
+	subnetManagerMock.
+		EXPECT().
+		CreateOrUpdate(gomock.Any(), *updatedSubnets[0].ID, updatedSubnets[0]).
```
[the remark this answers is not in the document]
AldoFusterTurpin: Done.

pkg/cluster/enableendpoints_test.go
```diff
@@ -0,0 +1,328 @@
...
+		Return(nil)
+
+	subnetManagerMock.
+		EXPECT().
+		CreateOrUpdate(gomock.Any(), *updatedSubnets[1].ID, updatedSubnets[1]).
```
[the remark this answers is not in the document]
AldoFusterTurpin: Done.

pkg/cluster/enableendpoints_test.go
```diff
@@ -0,0 +1,328 @@
...
+		endpointsAdder: endpointsAdderMock,
+		doc: &api.OpenShiftClusterDocument{
+			OpenShiftCluster: oc,
+		},
+	}
+
+	err := m.enableServiceEndpoints(context.Background())
+	expectedError := ""
+
+	if err != nil && err.Error() != expectedError || err == nil && expectedError != "" {
```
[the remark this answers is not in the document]
AldoFusterTurpin: Done.

pkg/util/subnet/endpointsadder.go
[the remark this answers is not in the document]
facchettos: here your design is the problem. 
```go
func (m *manager) enableServiceEndpoints(ctx context.Context) error {
	subnetIds, err := m.getSubnetIds()
	if err != nil {
		return err
	}

	subnets, err := m.subnet.GetAll(ctx, subnetIds)
	if err != nil {
		return err
	}

	subnetsToBeUpdated := m.endpointsAdder.AddEndpointsToSubnets(api.SubnetsEndpoints, subnets)

	return m.updateSubnets(subnetsToBeUpdated, ctx)
}
```

you should not have to call another function from the subnet package. (it's hidden because you have the interface, but still) `m.endpointsAdder.AddEndpointsToSubnets(api.SubnetsEndpoints, subnets)`.  

The step where you add the subnets should be inside `m.updateSubnets()` . whether it is inside createOrUpdate or another function, it's up to you, but this step should be in the `subnet` package, not the `cluster` package.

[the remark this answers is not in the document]
facchettos: I didn't mean that as a mean comment. It is not a very big problem, true, but it still impacts the ability to test easily and to understand the code, so my statement was true nonetheless. 

This line needs to go in subnet `subnetsToBeUpdated:=m.endpointsAdder.AddEndpointsToSubnets(api.SubnetsEndpoints, subnets)`. and please remove the interface, it is not serving any purpose here.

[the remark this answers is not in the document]
AldoFusterTurpin: Done.

pkg/cluster/enableendpoints.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	subnets, err := m.subnet.GetAll(ctx, subnetIds)
+	if err != nil {
+		return err
+	}
+
+	subnetsToBeUpdated := subnet.AddEndpointsToSubnets(api.SubnetsEndpoints, subnets)
+
+	return m.subnet.CreateOrUpdateSubnets(ctx, subnetsToBeUpdated)
```
facchettos, line 29: what is the reason behind this not being a single function in the subnet package ? `CreateOrUpdateSubnetsFromIds` would probably be a good name for such a function.

facchettos (changes requested):

pkg/cluster/enableendpoints.go
AldoFusterTurpin: The main reason is that if we had just one function like "CreateOrUpdateSubnetsFromIds" it would be a function that 
 both will have business logic + will talk with Azure (infrastructure). 

As is now, subnet.CreateOrUpdateSubnets(...) just talks to Azure (infrastructure) and subnet.AddEndpointsToSubnets(...) just have business logic. This makes subnet.AddEndpointsToSubnets() be easier to test as doesn't have side effects and also separates clear responsibilities between the functions.

facchettos: look, you have just shifted the responsibility elsewhere here, and now you need to export a bunch of functions from `subnet`. `enableServiceEndpoint` has both business logic and talks to azure, so you don't have an argument here. create a new function like this in the subnet package. 

```go 
func createFromIds() error {
	subnets, err := m.subnet.GetAll(ctx, subnetIds)
	if err != nil {
		return err
	}

	subnetsToBeUpdated := subnet.AddEndpointsToSubnets(api.SubnetsEndpoints, subnets)

	return m.subnet.CreateOrUpdateSubnets(ctx, subnetsToBeUpdated)
}
```

and you will be able to mock it as bonus.

AldoFusterTurpin: I thought you meant to move all the logic into the same function instead of creating a function that invokes the already created functions, as I see here. Doing it now.

SudoBrendan (dismissed):
LGTM

hawkowl (approved):

PR #2219, opened 2022-06-30
Fix/panic on admin update

AldoFusterTurpin:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14743726

It fixes a panic on admin patch (AdminUpdate) operation on this [line](https://github.com/Azure/ARO-RP/blob/v20220610.00/pkg/cluster/storageaccounts.go#L107) when Config.imageregistry.operator.openshift.io/cluster does not have the structure we expect (when there are missing fields). Now the code returns the appropriate error instead of panic.

Note: other steps depend on having storage account name. We need to carefully look all refrences to ImageRegistryStorageAccountName field.

### Test plan for issue:

Added corresponding unit tests.

N/A

### DIFF
--- a/pkg/api/openshiftclusterdocument.go
+++ b/pkg/api/openshiftclusterdocument.go
@@ -1,5 +1,7 @@
 package api
 
+import "fmt"
+
 // Copyright (c) Microsoft Corporation.
 // Licensed under the Apache License 2.0.
 
@@ -49,4 +51,12 @@ type OpenShiftClusterDocument struct {
 
 func (c *OpenShiftClusterDocument) String() string {
 	return encodeJSON(c)
+}
+
+func (doc *OpenShiftClusterDocument) ImageRegistryStorageAccountName() (string, error) {
+	if doc.OpenShiftCluster == nil {
+		return "", fmt.Errorf("OpenShiftCluster info from OpenShift cluster document is nil")
+	}
+
+	return doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName, nil
 }
--- a/pkg/cluster/deploystorage.go
+++ b/pkg/cluster/deploystorage.go
@@ -109,12 +109,22 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
 	clusterStorageAccountName := "cluster" + m.doc.OpenShiftCluster.Properties.StorageSuffix
 	azureRegion := strings.ToLower(m.doc.OpenShiftCluster.Location) // Used in k8s object names, so must pass DNS-1123 validation
 
+	imageRegistryStorageAccountName, err := m.imageRegistryStorageAccountName()
+	if err != nil {
+		return err
+	}
+
+	err = validateImageRegistryStorageAccountName(imageRegistryStorageAccountName)
+	if err != nil {
+		return err
+	}
+
 	resources := []*arm.Resource{
 		m.storageAccount(clusterStorageAccountName, azureRegion, true),
 		m.storageAccountBlobContainer(clusterStorageAccountName, "ignition"),
 		m.storageAccountBlobContainer(clusterStorageAccountName, "aro"),
-		m.storageAccount(m.doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName, azureRegion, true),
-		m.storageAccountBlobContainer(m.doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName, "image-registry"),
+		m.storageAccount(imageRegistryStorageAccountName, azureRegion, true),
+		m.storageAccountBlobContainer(imageRegistryStorageAccountName, "image-registry"),
 		m.clusterNSG(infraID, azureRegion),
 		m.clusterServicePrincipalRBAC(),
 		m.networkPrivateLinkService(azureRegion),
@@ -146,6 +156,14 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
 	}
 
 	return arm.DeployTemplate(ctx, m.log, m.deployments, resourceGroup, "storage", t, nil)
+}
+
+func (m *manager) imageRegistryStorageAccountName() (string, error) {
+	if m.doc == nil {
+		return "", fmt.Errorf("OpenShift cluster document is nil")
+	}
+
+	return m.doc.ImageRegistryStorageAccountName()
 }
 
 func (m *manager) attachNSGs(ctx context.Context) error {
--- a/pkg/cluster/ipaddresses.go
+++ b/pkg/cluster/ipaddresses.go
@@ -236,12 +236,22 @@ func (m *manager) ensureGatewayCreate(ctx context.Context) error {
 		return errors.New("private endpoint connection not found")
 	}
 
+	imageRegistryStorageAccountName, err := m.imageRegistryStorageAccountName()
+	if err != nil {
+		return err
+	}
+
+	err = validateImageRegistryStorageAccountName(imageRegistryStorageAccountName)
+	if err != nil {
+		return err
+	}
+
 	_, err = m.dbGateway.Create(ctx, &api.GatewayDocument{
 		ID: linkIdentifier,
 		Gateway: &api.Gateway{
 			ID:                              m.doc.OpenShiftCluster.ID,
 			StorageSuffix:                   m.doc.OpenShiftCluster.Properties.StorageSuffix,
-			ImageRegistryStorageAccountName: m.doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName,
+			ImageRegistryStorageAccountName: imageRegistryStorageAccountName,
 		},
 	})
 
@@ -257,7 +267,7 @@ func (m *manager) ensureGatewayCreate(ctx context.Context) error {
 			return err
 		}
 		if !strings.EqualFold(gwyDoc.ID, m.doc.OpenShiftCluster.ID) ||
-			!strings.EqualFold(gwyDoc.Gateway.ImageRegistryStorageAccountName, m.doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName) ||
+			!strings.EqualFold(gwyDoc.Gateway.ImageRegistryStorageAccountName, imageRegistryStorageAccountName) ||
 			!strings.EqualFold(gwyDoc.Gateway.StorageSuffix, m.doc.OpenShiftCluster.Properties.StorageSuffix) {
 			return errors.New("gateway record already exists for a different cluster")
 		}
--- a/pkg/cluster/storageaccounts.go
+++ b/pkg/cluster/storageaccounts.go
@@ -10,9 +10,11 @@ import (
 
 	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
 	"github.com/Azure/go-autorest/autorest/to"
+	imageregistryv1 "github.com/openshift/api/imageregistry/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database"
 	"github.com/Azure/ARO-RP/pkg/util/arm"
 	"github.com/Azure/ARO-RP/pkg/util/stringutils"
 )
@@ -79,14 +81,23 @@ func (m *manager) migrateStorageAccounts(ctx context.Context) error {
 		return nil
 	}
 	clusterStorageAccountName := "cluster" + m.doc.OpenShiftCluster.Properties.StorageSuffix
-	registryStorageAccountName := m.doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName
+
+	imageRegistryStorageAccountName, err := m.imageRegistryStorageAccountName()
+	if err != nil {
+		return err
+	}
+
+	err = validateImageRegistryStorageAccountName(imageRegistryStorageAccountName)
+	if err != nil {
+		return err
+	}
 
 	t := &arm.Template{
 		Schema:         "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
 		ContentVersion: "1.0.0.0",
 		Resources: []*arm.Resource{
 			m.storageAccount(clusterStorageAccountName, m.doc.OpenShiftCluster.Location, false),
-			m.storageAccount(registryStorageAccountName, m.doc.OpenShiftCluster.Location, false),
+			m.storageAccount(imageRegistryStorageAccountName, m.doc.OpenShiftCluster.Location, false),
 		},
 	}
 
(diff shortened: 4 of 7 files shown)


pkg/cluster/storageaccounts.go
```diff
@@ -98,14 +100,63 @@ func (m *manager) populateRegistryStorageAccountName(ctx context.Context) error
...
+func (m *manager) getRegistryConfig(ctx context.Context) (*v1.Config, error) {
+	configName := "cluster"
+
+	registryConfig, err := m.imageregistrycli.ImageregistryV1().Configs().Get(ctx, configName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	return registryConfig, nil
+}
```
facchettos, line 123: This is superfluous imo. If you still want to have it in a function, which i wouldn't do , you can simplify by having 
```go 
func (m *manager) getRegistryConfig(ctx context.Context) (*v1.Config, error) {
	configName := "cluster"

	return m.imageregistrycli.ImageregistryV1().Configs().Get(ctx, configName, metav1.GetOptions{})
}
```
or even 

```go 
func (m *manager) getRegistryConfig(ctx context.Context) (*v1.Config, error) {
    return m.imageregistrycli.ImageregistryV1().Configs().Get(ctx, "cluster", metav1.GetOptions{})
}
```
but this becomes really obvious here that you don't need the function

facchettos (changes requested):

pkg/cluster/storageaccounts.go
AldoFusterTurpin: Interesting point, but I added the function because:
1- I don't want to know how the registry is obtained, I prefer to delegate it to a function so I hide the details when I am reading "populateRegistryStorageAccountName" method. It also favors SRP principle: if the way the registry config is obtained changes, we just have to touch this function and not "populateRegistryStorageAccountName" (same if it is used in more places).
2- it also adds the explanatory variable "configName" because the second parameter to Get() was hardcoded and at first sight was not clear what "cluster" meant. Now it is clear that refers to the config name.
🙂

facchettos: But you turn one line of code, which already follows srp , because it is already its own function , into 6 , and it brings little to no value here.

AldoFusterTurpin: Strictly speaking, if we don't create the aux function to get the config registry, we are breaking the SRP principle:  if the way the registry is obtained changes, we will touch the "populateRegistryStorageAccountName" function when in reality, this function should not be responsible for obtaining the registry config, the same way it should not be responsible for creating the mutator function, "populateRegistryStorageAccountName" just delegates those responsibilities to others. 

I agree that in this case as we just have one line it is not that clear, but if we forgot about the number of lines and just focus on responsibility, I would keep the aux function.

One function should do one thing and do it right. 

A great [conference](https://www.youtube.com/watch?v=QvrNwfuWbCQ) about the topic by Uncle Bob.

🙂

facchettos: but the function is already created and is literally called GET, so SRP is followed. the only thing that you are doing is renaming it and making the code more complex. if we follow your idea we should also have a function for creating the variable and one for calling get with that variable. It makes no sense.

AldoFusterTurpin: > but the function is already created and is literally called GET, so SRP is followed. the only thing that you are doing is renaming it and making the code more complex. if we follow your idea we should also have a function for creating the variable and one for calling get with that variable. It makes no sense.

I don't think "we should also have a function for creating the variable and one for calling get with that variable". Is just about group responsibility together. As I mention, this is not 100% obvious here as it is just one line. 

If the way we get the config file in the future changes, we will need to change it X times where X is the number of times we get the config in our codebase. If you have it on a function, you just need to change it in one place, in the function definition.

I will remove the aux function here (as it is just one line) but I still think IMHO the reasoning makes sense. 🙂

pkg/cluster/deploystorage.go
```diff
@@ -148,6 +157,19 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
 	return arm.DeployTemplate(ctx, m.log, m.deployments, resourceGroup, "storage", t, nil)
 }
 
+func (m *manager) imageRegistryStorageAccountName() (string, error) {
+	if m == nil {
+		return "", fmt.Errorf("cluster manager is nil")
+	}
```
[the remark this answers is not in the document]
AldoFusterTurpin: You can reach that point as you can not ensure that m was initialized before invoking the method.
Here is an example with a nil pointer exception: 
```go
package main

import (
	"log"
)

type manager struct {
}

func (m *manager) Foo() {
	log.Print(m)
	log.Print(*m) // <- this line will panic if pointer receiver (m) was NOT initialised
	log.Print(&m)
	log.Print(&(*m))
}

func main() {
	var m *manager = &manager{}
	m.Foo() // <- no panic as the pointer receiver is initialised

	log.Print("\n\n")

	var m2 *manager
	m2.Foo()
}

```
And here is that example in the [Playground](https://go.dev/play/p/dD3KdnFpIqU)

The line that is giving nil is
```go
log.Print(*m)
```
because you try to dereference a non-nil pointer.

If you add 
```go
if m == nil {
    return
}
```
inside Foo() in the example provided it will not panic.
🙂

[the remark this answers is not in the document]
facchettos: but any references to that function has at least one instance of accessing fields referenced by that pointer. it will never reach that point if it is nil. 
Here nil is not a valid value for the manager, so you should check if it is nil at creation if you do. it would become incredibly cumbersome to check it for every pointer receiver. And by not valid, I mean that we can't do anything of it, not that we can't go out of our way to create a nil pointer.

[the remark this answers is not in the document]
AldoFusterTurpin: "it will never reach that point if it is nil." ->  You can call m.imageRegistryStorageAccountName() when m (of type *manager) have been declared but not initialised, so it is possible to reach that point. 

We can not assume that m will never be nil when calling m.imageRegistryStorageAccountName(), so you must check if m == nil inside the function or outside, but you have to check it at some point. 

The good thing about checking for nil inside the imageRegistryStorageAccountName() is that you would never need to check if m is nil before calling this function. So if in your codebase you have N calls to m.imageRegistryStorageAccountName() you just need to do the nil check inside the function, not N times before calling the function.

I am removing the nil check inside the function because it is something we don't usually do in the codebase but still think it has some interesting usages. 

🙂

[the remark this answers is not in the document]
facchettos: if there is a check it should be at creation time, not inside the function since we are never expecting this to be a nil pointer. Here it makes no sense to call that function on a nil pointer, if it is nil, something failed before and we should have failed beforehand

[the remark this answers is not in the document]
AldoFusterTurpin: "Here it makes no sense to call that function on a nil pointer" ->In fact, it almost never makes sense to call a method on a pointer receiver so it is not specific to that case. As soon as you dereference the pointer inside the function, you should ensure it is not nil.

Obviously, it is better to assert correctness as close to the creation of entities as possible, that's for sure. I completely get your point about checking things before we reach a point where we call to the method on a nil receiver, BUT...

The problem is that even if you do ALL that, nothing prevents the codebase to have in the package cluster something like:

```go
var m *manager
// more code...
m. imageRegistryStorageAccountName()
```
and panic will occur. 

Because m. imageRegistryStorageAccountName() can be called anywhere in the cluster package, you need to check for not nil before calling the method, doing it inside or outside the function is your decision, but is a decision that you should make. 

But again, I removed the check inside the function so we are all happy 🙂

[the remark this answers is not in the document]
AldoFusterTurpin: Sorry, when I said "In fact, it almost never makes sense to call a method on a pointer receiver so it is not specific to that case."
I wanted to say
"In fact, it almost never makes sense to call a method on a NIL pointer receiver so it is not specific to that case." (I forgot the nil), as I wanted to respond to your sentence "Here it makes no sense to call that function on a nil pointer".

Nothing against pointer receivers, just against never checking for nil receivers.

All clear.

pkg/api/openshiftclusterdocument.go
```diff
@@ -50,3 +52,15 @@ type OpenShiftClusterDocument struct {
 func (c *OpenShiftClusterDocument) String() string {
 	return encodeJSON(c)
 }
+
+func (doc *OpenShiftClusterDocument) ImageRegistryStorageAccountName() (string, error) {
+	if doc == nil {
+		return "", fmt.Errorf("OpenShift cluster document is nil")
+	}
```
facchettos, line 59: Here it is the same. no need to check if it is nil, since it checked in `pkg/backend/openshiftcluster.go:48`  already

facchettos (changes requested):

pkg/cluster/deploystorage.go
```diff
@@ -109,12 +109,21 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
 	clusterStorageAccountName := "cluster" + m.doc.OpenShiftCluster.Properties.StorageSuffix
 	azureRegion := strings.ToLower(m.doc.OpenShiftCluster.Location) // Used in k8s object names, so must pass DNS-1123 validation
 
+	accountName, err := m.imageRegistryStorageAccountName()
```
bennerv, line 112: can we rename this to be consistent with `clusterStorageAccountName`? See above.

```suggestion
	imageRegistryStorageAccountName, err := m.imageRegistryStorageAccountName()

File: pkg/api/openshiftclusterdocument.go [hunk 5]
```diff
@@ -50,3 +52,15 @@ type OpenShiftClusterDocument struct {
 func (c *OpenShiftClusterDocument) String() string {
 	return encodeJSON(c)
 }
+
+func (doc *OpenShiftClusterDocument) ImageRegistryStorageAccountName() (string, error) {
```
bennerv, line 56: Do we actually need this function?  Can we just replace it with `doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName`? I don't see any issue with dereferencing the `doc` type or the `doc.OpenShiftCluster` type as they are already populated through cosmos.

pkg/cluster/ipaddresses.go
```diff
@@ -236,12 +236,21 @@ func (m *manager) ensureGatewayCreate(ctx context.Context) error {
...
 	}
 
+	accountName, err := m.imageRegistryStorageAccountName()
+	if err != nil {
+		return err
+	}
+
+	if accountName == "" {
+		return fmt.Errorf("cluster manager's storage account name of image registry config is empty")
+	}
```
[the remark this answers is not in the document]
bennerv: Just a nit on the error message as well to be more concise. 
```suggestion
	if accountName == "" {
		return fmt.Errorf("the cluster's image registry name is empty")
	}
```

I believe the error message should be propagated to our logging system which should enrich the logger with the resource ID of the cluster in question which threw an error.  Can you confirm this, otherwise we should add the clusterId to the error message.

bennerv (changes requested):
Good work so far!  Few questions and nits.

pkg/api/openshiftclusterdocument.go
AldoFusterTurpin: If I remove the condition, I will need to check  m.doc != nil on `pkg/cluster/deploystorage.go:161` before calling this method. For me it makes no sense to traverse the call stack to **see** if at that point m.doc is nil or not, I want to **ensure** it is not nil). 

The good think about doing it inside the method is that we don't ever need to do in the codebase:
```go
if doc != nil {
    err := doc.ImageRegistryStorageAccountName()
}
``` 
we can simply do 
```go
err :=  doc.ImageRegistryStorageAccountName()
```
and forget about the panic case.

If I don't check doc == nil inside the method I will do it just before every call to doc.ImageRegistry(), here is just one call but I am talking about the general idea of the check inside the method. 

I will remove the nil check from inside the method and add the nil check before calling imageRegistryStorageAccountName() because I still can not convince you 😂👌

pkg/cluster/deploystorage.go
AldoFusterTurpin: Done!

pkg/api/openshiftclusterdocument.go
AldoFusterTurpin: Honestly, I did it accessing the nested structs at the beginning but I decided to create this method because:

1- At that point, I felt I was violating the Law of Demeter -> from the perspective of the client talking to OpenShiftClusterDocument, I know it can provide me "something" called "image registry account name" but as I am interacting with that object, I don't want to know how to obtain it (information encapsulation).  What happens when you access the field using all the nested calls doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName you expose the details.

2- In order to access ImageRegistryStorageAccountName, we first need to ensure doc.OpenShiftCluster is not nil, and the benefit of having this function is that you just need to do the check inside the function. If you don't do it inside, you will need to do it N times where N is the number of calls to ImageRegistryStorageAccountName(). In that case, is just 1 but I don't see any drawback of having the function.

I know this type of nested access is common in Go, but for me, point 2 is important here to preserve encapsulation.

Have I convinced you?
If yes, I will directly resolve the conversation.
If not, I will apply your suggestion and resolve the conversation.

Thanks! 🙂

pkg/cluster/ipaddresses.go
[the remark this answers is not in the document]
AldoFusterTurpin: About
> should we bundle this condition within the imageRegistryStorageAccountName() function? We're doing this in two places so would adhere to the DRY principal.

At the beginning, I created a function that did exactly that: get data and validate the data. 

Soon I realized that having a function that both gets data and at the same time validates it was a bad idea because there is a case where we want to get the data and if that data is not empty, just return no error. 

I am talking about pkg/cluster/storageaccounts.go:112
```go
accountName, err := m.imageRegistryStorageAccountName()
if err != nil {
	return err
}

if accountName != "" {
	return nil
}
```

If I have a function that both gets the data and validates it, when calling it in the piece of code I have shared, I will need to get the error, check the error == "cluster manager's storage account name of image registry config is empty" and bypass the error and do business logic depending on the error. In other words, the fact that imageRegistryStorageAccount == "" is an error just depends on the context. 

The best option would be to create a function validateImageRegistryStorageAccountName() and call it from the two places you mention.

About 
> I believe the error message should be propagated to our logging system which should enrich the logger...

I will change the error message as you suggest and check the logging.

Thanks!

pkg/api/openshiftclusterdocument.go
m1kola: I also think it is unnecessary

> I am interacting with that object, I don't want to know how to obtain it (information encapsulation). What happens when you access the field using all the nested calls doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName you expose the details.

`doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName` vs `doc.ImageRegistryStorageAccountName()` - doesn't make much difference.

Also following this logic - we should have a getter for every field we have on the `OpenShiftClusterDocument` level. Doesn't sound reasonable.

> In order to access ImageRegistryStorageAccountName, we first need to ensure doc.OpenShiftCluster is not nil

Also I don't think `doc.OpenShiftCluster` will ever be `nil`. I don't know why it is a pointer, but we set it straight away [on cluster creation](https://github.com/Azure/ARO-RP/blob/98fc429ecefc51d669cc72dc7cd3963231c18f1c/pkg/frontend/openshiftcluster_putorpatch.go#L68-L86) and never set to `nil`, I believe.

>  If you don't do it inside, you will need to do it N times where N is the number of calls to ImageRegistryStorageAccountName(). In that case, is just 1 but I don't see any drawback of having the function.

You still need to do `if err != nil` every time you call `ImageRegistryStorageAccountName`.

pkg/cluster/deploystorage.go
```diff
@@ -109,12 +109,22 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
...
 
+	imageRegistryStorageAccountName, err := m.imageRegistryStorageAccountName()
+	if err != nil {
+		return err
+	}
+
+	err = validateImageRegistryStorageAccountName(imageRegistryStorageAccountName)
+	if err != nil {
+		return err
+	}
```
m1kola, line 120: I think we only run `deployStorageTemplate` at installation time and I believe we will always have storage account name. It is the old clusters which we need to worry about.

pkg/cluster/ipaddresses.go
```diff
@@ -236,12 +236,22 @@ func (m *manager) ensureGatewayCreate(ctx context.Context) error {
...
+		return err
+	}
+
 	_, err = m.dbGateway.Create(ctx, &api.GatewayDocument{
 		ID: linkIdentifier,
 		Gateway: &api.Gateway{
 			ID:                              m.doc.OpenShiftCluster.ID,
 			StorageSuffix:                   m.doc.OpenShiftCluster.Properties.StorageSuffix,
-			ImageRegistryStorageAccountName: m.doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName,
+			ImageRegistryStorageAccountName: imageRegistryStorageAccountName,
```
m1kola, line 254: It is not directly related to the panic, but - does this mean that we probably have some clusters in prod where this field is empty in the  `GatewayDocument`?

I think we run `ensureGatewayUpgrade` & `ensureGatewayCreate` before `populateRegistryStorageAccountName`. So for older clustes I think we create 

cc @bennerv

pkg/cluster/deploystorage.go
m1kola: I had a quick review of all the places which today (in `master` as of 98fc429ecefc51d669cc72dc7cd3963231c18f1c) reference [`ImageRegistryStorageAccountName` field](https://github.com/Azure/ARO-RP/blob/70a3f19716075a23e7324fb42f78c7a30be3c09c/pkg/api/openshiftcluster.go#L135) and it looks like we only have issue in `populateRegistryStorageAccountName`.

Unless I miss something, in all other places we will have this field populates.

EDIT: Potentially `ensureGatewayCreate`, but it is a related, but separate issue.

m1kola (changes requested):
@AldoFusterTurpin could you please make this PR to include only changes required to fix the panic and move refactorings not related to the fix itself to a separate PR?

PR #2221, opened 2022-07-01
Added an API with `installVersion` field in OpenShiftCluster struct

anshulvermapatel:
Work Item: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/10581245

This PR is for the part 1 of the work item i.e adding a new API field to a new API.


bennerv:
I think @ross-bryan mentioned this in email, but we should probably set the date on this to 2022-08-01 and not release it as a preview API (assuming we're including hive stuff within this).

See "Question about Azure API versions for the RP"

PR #2222, opened 2022-07-01
Except NRMS NSGs from subnet validation

cadenmarchese:

NRMS NSGs (Simply Secure v1) policies are applied in INT when building clusters there. This happens after subnets are created and before subnets are validated, failing the installation with "can't have a network security group attached" errors.

Right now, the workaround is to detach NRMS NSGs from worker and master subnets before starting the install.

This is meant as a more permanent fix to allow the NRMS NSGs during subnet validation so that installs don't fail.

We had previously attempted to except our RP Health subscription from NRMS policies using https://eng.ms/docs/security-compliance-identity-and-management-scim/security/azure-security/security-health-analytics/network-isolation/tsgs/simplysecurev1/optout-simple-secure-network-rules, but it didn't work for us at the time.

### Test plan for issue:

- Unit tests and e2e are green
- Deployments in INT no longer fail on subnet validation

No

### DIFF
--- a/pkg/api/validate/dynamic/dynamic.go
+++ b/pkg/api/validate/dynamic/dynamic.go
@@ -436,7 +436,9 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 
 		if oc.Properties.ProvisioningState == api.ProvisioningStateCreating {
 			if ss.SubnetPropertiesFormat != nil &&
-				ss.SubnetPropertiesFormat.NetworkSecurityGroup != nil {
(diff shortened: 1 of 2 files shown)


cadenmarchese:
This change appears to be working on our end: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=57436635&view=logs&j=d85d6967-64be-5836-1dec-1ff674e70d8e

SudoBrendan (dismissed):
lgtm

ross-bryan (commented):
I've added this PR to the next leads sync to discuss impact and risks with this

PR #2224, opened 2022-07-04
Migrate Machine API definitions

yehornaumenko:
Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/12389289

`AzureMachineProviderSpec` type has been deleted from the latest versions of [cluster-api-provider-azure repository](https://github.com/kubernetes-sigs/cluster-api-provider-azure).
This change updates references to `AzureMachineProviderSpec` type with references to the type in [openshift/api repository](https://github.com/openshift/api/blob/master/machine/v1beta1/types_azureprovider.go).

### Test plan for the issue:
Ran PUCM on 4.9 clusters with a version < 4.10

No


pkg/operator/controllers/machine/machine.go
```diff
@@ -37,9 +37,15 @@ func (r *Reconciler) machineValid(ctx context.Context, machine *machinev1beta1.M
 		return []error{fmt.Errorf("machine %s: provider spec missing", machine.Name)}
 	}
 
-	machineProviderSpec, err := utilmachine.UnmarshalAzureProviderSpec(machine.Name, utilmachine.Machine, machine.Spec.ProviderSpec.Value.Raw)
+	o, _, err := scheme.Codecs.UniversalDeserializer().Decode(machine.Spec.ProviderSpec.Value.Raw, nil, nil)
 	if err != nil {
-		errs = append(errs, err)
+		return []error{err}
+	}
+	machineProviderSpec, ok := o.(*machinev1beta1.AzureMachineProviderSpec)
```
facchettos, line 44: same as the other

facchettos (changes requested):

pkg/operator/controllers/machine/machine.go
```diff
@@ -37,9 +37,15 @@ func (r *Reconciler) machineValid(ctx context.Context, machine *machinev1beta1.M
...
+	if !ok {
+		// This should never happen: codecs uses scheme that has only one registered type
+		// and if something is wrong with the provider spec - decoding should fail
+		return []error{fmt.Errorf("machine %s: failed to read provider spec: %T", machine.Name, o)}
```
m1kola, line 48: The thing is - we need to support both types: old one (we run the operator in the older clusters) and the new one. That's why we introduced `utilmachine.UnmarshalAzureProviderSpec`.

Note: it is very likely that there is no better solution than having `utilmachine.UnmarshalAzureProviderSpec`, but I would like you to explore the options.

I was hoping that we could somehow solve it on the scheme level instead of having `utilmachine.UnmarshalAzureProviderSpec` which does relatively unreliable check of `strings.Contains(string(rawProviderSpec), azureProviderSpec)`. Maybe by registring `azureproviderconfig.openshift.io` to be unmarshalled into the new type? Or somehow else?

I'm wondering what other repos in https://github.com/openshift do to deal with this migration and backward compatibility. We can ask around: `#forum-cloud` on slack is probably a good place to start.

m1kola (changes requested):
We need to support both old and rew type (more details below).

pkg/operator/controllers/machine/machine.go
facchettos: do you know in what field is that `azureproviderspec` string supposed to show up ?

pkg/cluster/fixmcsuserdata.go
```diff
@@ -81,10 +82,14 @@ func getUserDataSecretReference(objMeta *metav1.ObjectMeta, spec *machinev1beta1
 		return nil, nil
 	}
 
-	machineProviderSpec, err := utilmachine.UnmarshalAzureProviderSpec(spec.Name, utilmachine.Machine, spec.ProviderSpec.Value.Raw)
+	o, _, err := scheme.Codecs.UniversalDeserializer().Decode(spec.ProviderSpec.Value.Raw, nil, nil)
 	if err != nil {
 		return nil, err
 	}
+	machineProviderSpec, ok := o.(*machinev1beta1.AzureMachineProviderSpec)
```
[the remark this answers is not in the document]
yehornaumenko: Fixed

pkg/operator/controllers/machine/machine.go
yehornaumenko: @m1kola I've updated a scheme. Could you take a look at it, please? Also, I am wondering if we need some tests to be added.

darthhexx (dismissed):
This PR should be using utilmachine.UnmarshalAzureProviderSpec and not removing those files, because as it stands these changes will prevent the ARO operator from decoding machines created by older OCP machinesets. See https://github.com/Azure/ARO-RP/pull/2070#discussion_r867586623

pkg/util/scheme/scheme.go
```diff
@@ -43,4 +44,6 @@ func init() {
 	utilruntime.Must(monitoringv1.AddToScheme(scheme.Scheme))
 	// AzureMachineProviderSpec is not registered by default
 	scheme.Scheme.AddKnownTypes(machinev1beta1.GroupVersion, &machinev1beta1.AzureMachineProviderSpec{})
+	// Need it for backward compatibility
```
m1kola, line 47: I would like to see more verbose comment here, if possible.

pkg/operator/controllers/machine/machine.go
m1kola: @yehornaumenko yes, I think we need to update few tests to make sure that we unmarshall both old and new type correctly. I think we need to update tests for the following code:

* `pkg/cluster/fixmcsuserdata.go`
* `pkg/operator/controllers/machine/machine.go`
* `pkg/util/clusterdata/worker_profiles_task.go`
* `pkg/util/subnet/cluster_subnets.go`
* Something else?

m1kola (commented):
Looks good to me. We need a bit more tests (see https://github.com/Azure/ARO-RP/pull/2224#discussion_r915829827) and probably a bit more verbose comment (see https://github.com/Azure/ARO-RP/pull/2224#discussion_r915828500).

> This PR should be using utilmachine.UnmarshalAzureProviderSpec and not removing those files, because as it stands these changes will prevent the ARO operator from decoding machines created by older OCP machinesets. See https://github.com/Azure/ARO-RP/pull/2070#discussion_r867586623

@darthhexx this is a replacement for `utilmachine.UnmarshalAzureProviderSpec` hack.

This PR is trying to get rid of `utilmachine.UnmarshalAzureProviderSpec` which was introduced as a quick fix for backward compatibility with the older type in #2167. Now we are trying to solve the same issue in a different way: on the `scheme` level. See `pkg/util/scheme/scheme.go`.

hack/validate-imports/allowed-import-names.yaml
```diff
@@ -47,6 +47,8 @@ allowedImportNames:
     - coreosarch
   github.com/googleapis/gnostic/openapiv2:
     - openapi_v2
+  github.com/openshift/api/machine/v1beta1:
+    - machinev1beta1
```
bennerv, line 51: I don't think we need this anymore.  I believe it should pass without it as anything under `openshift/api` -> `api` + `apiVersion` which results in `machinev1beta1`.

Can you please remove the old reference in the file too as that is also not in use.  

`github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1`

bennerv (changes requested):
Looks good - just one nit / confirmation.

pkg/cluster/fixmcsuserdata_test.go
```diff
@@ -184,3 +192,76 @@ func TestFixMCSUserData(t *testing.T) {
 		}
 	}
 }
+
+func TestGetUserDataSecretReference(t *testing.T) {
```
m1kola, line 196: @yehornaumenko you will soon be looking for a new task - here is one for you :)

Could you please look into expanding (in a separate PR) this test to provide better coverage for `getUserDataSecretReference`?

image

pkg/cluster/fixmcsuserdata_test.go
```diff
@@ -184,3 +192,76 @@ func TestFixMCSUserData(t *testing.T) {
...
+	for _, td := range []getUserDataSecretReferenceTestData{
```
m1kola, line 197: We normally just do this in tests (so that you don't have to define `getUserDataSecretReferenceTestData`:

```suggestion
	for _, td := range []struct {
		name        string
		objectMeta  *metav1.ObjectMeta
		machineSpec *machinev1beta1.MachineSpec
		result      *corev1.SecretReference
		shouldFail  bool
	}{
```

m1kola (approved):
Looks good. I'm happy to merge on green e2e.

pkg/util/clusterdata/worker_profiles_task.go
```diff
@@ -74,12 +75,14 @@ func (ef *workerProfilesEnricherTask) FetchData(ctx context.Context, callbacks c
...
-			// we need to handle it appropriately
-			ef.log.Info(err.Error())
+			ef.log.Info(err)
+			continue
+		}
+		machineProviderSpec, ok := obj.(*machinev1beta1.AzureMachineProviderSpec)
+		if !ok {
+			ef.log.Infof("failed to read provider spec from the machine set %q: %T", machineset.Name, obj)
 			continue
 		}
```
facchettos, line 88: We should not be using continue here. The logic is backwards because it is not in a function, and should be. We should be using return. It'd be great to have this refactored instead of adding more things with `continue`

facchettos (changes requested):

pkg/util/clusterdata/worker_profiles_task.go
m1kola: @facchettos 

> We should not be using continue here. The logic is backwards because it is not in a function, and should be. We should be using return.

I don't understand your point. I read it as "If thre is an issue with one `MachineSet` - log and continue to the next one". So logic looks correct to me.

> It'd be great to have this refactored instead of adding more things with `continue`

Anyway we are reverting back to the code we had before we introduced `utilmachine.UnmarshalAzureProviderSpec`. See [this PR](https://github.com/Azure/ARO-RP/pull/2167/files#diff-daed99b70882a6c275d461b2973b19bbf1d09f34100f08ae95c4ce58f935df42L78-L82). We are not adding anything new here.

Refactoring of thing is out of scope of this PR. Feel free to submit a PR to refactor this.

PR #2228, opened 2022-07-06
E2E scripting and documentation improvements

dem4gus:

Changes to e2e testing documentation and scripting. Using `exit` in a dot-sourced script terminates the terminal session immediately and doesn't let you read error messages. Since the script is setting environment variables in the current terminal session it needs to be dot-sourced, using `return` will allow the session to remain open so any errors can be read.

* [StackOverflow answer](https://stackoverflow.com/questions/9640660/any-way-to-exit-bash-script-but-not-quitting-the-terminal/9640718#9640718)
* [Bourne Shell Built-ins](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Bourne-Shell-Builtins) for behavior of `exit` and `return`.

### Test plan for issue:

Updated script has run in CI successfully multiple times as part of #2177. The script is run using ADO variables as fallbacks when they are not set locally so failure states with variables not being set are only reached when running locally.

My particular failure occurred when `LOCAL_E2E` was set but `CLUSTER` was not. In this case the block starting on line 91 would execute, but the message was not readable due to the terminal session closing.

PR includes documentation updates specifying that `CLUSTER` must be set for local e2e.


s-amann (approved):
lgtm

karanmagdani (approved):
LGTM! Thank you

ross-bryan (approved):

PR #2230, opened 2022-07-07
Prevent AdminAction redeploy if VM has an ephemeral OS disk

SrinivasAtmakuri:
Redeploying a VM with an ephemeral OS disk will result in a error message,
stating that the VM has an ephemeral OS disk and cannot be redeployed.

If a VM that has an ephemeral OS disk is redeployed it will never become ready,
due to the fact that the VM's OS is re-provisioned and
the cluster machine approver will not allow the "new" machine
to join the cluster again.

Fixes : https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14751454/

### DIFF
--- a/pkg/frontend/adminactions/azureactions.go
+++ b/pkg/frontend/adminactions/azureactions.go
@@ -74,6 +74,14 @@ func NewAzureActions(log *logrus.Entry, env env.Interface, oc *api.OpenShiftClus
 
 func (a *azureActions) VMRedeployAndWait(ctx context.Context, vmName string) error {
 	clusterRGName := stringutils.LastTokenByte(a.oc.Properties.ClusterProfile.ResourceGroupID, '/')
+	vm, err := a.virtualMachines.Get(ctx, clusterRGName, vmName, mgmtcompute.InstanceView)
+	if err != nil {
+		return err
+	}
+	if vmDisk := vm.StorageProfile.OsDisk; vmDisk != nil && vmDisk.DiffDiskSettings != nil &&
+		vmDisk.Caching == "ReadOnly" && vmDisk.DiffDiskSettings.Option == "Local" && vmDisk.DiffDiskSettings.Placement == "CacheDisk" {
+		return api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden, "", "VM '%s' has an Ephemeral Disk OS and cannot be redeployed.", vmName)
+	}
 	return a.virtualMachines.RedeployAndWait(ctx, clusterRGName, vmName)
 }
 


pkg/frontend/adminactions/azureactions.go
```diff
@@ -74,6 +74,16 @@ func NewAzureActions(log *logrus.Entry, env env.Interface, oc *api.OpenShiftClus
 
 func (a *azureActions) VMRedeployAndWait(ctx context.Context, vmName string) error {
 	clusterRGName := stringutils.LastTokenByte(a.oc.Properties.ClusterProfile.ResourceGroupID, '/')
+	vm, err := a.virtualMachines.Get(ctx, clusterRGName, vmName, mgmtcompute.InstanceView)
+	if err != nil {
+		return err
+	}
+	vmDisk := vm.StorageProfile.OsDisk
+	if vmDisk.Caching == "ReadOnly" {
+		if vmDisk.DiffDiskSettings.Option == "Local" && vmDisk.DiffDiskSettings.Placement == "CacheDisk" {
```
s-amann, line 83: Q: are `DiffDiskSettings` always not nil?

pkg/frontend/adminactions/azureactions.go
```diff
@@ -74,6 +74,16 @@ func NewAzureActions(log *logrus.Entry, env env.Interface, oc *api.OpenShiftClus
 
 func (a *azureActions) VMRedeployAndWait(ctx context.Context, vmName string) error {
 	clusterRGName := stringutils.LastTokenByte(a.oc.Properties.ClusterProfile.ResourceGroupID, '/')
+	vm, err := a.virtualMachines.Get(ctx, clusterRGName, vmName, mgmtcompute.InstanceView)
+	if err != nil {
+		return err
+	}
+	vmDisk := vm.StorageProfile.OsDisk
```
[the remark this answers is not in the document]
SrinivasAtmakuri: In probably most of the scenarios, VM has a storage profile, but as panic are tricky to troubleshoot, I will add an extra validation step. Thank you!

pkg/frontend/adminactions/azureactions.go
SrinivasAtmakuri: I will add the step to check if its `nil`

bennerv (commented):
I don't think everyone understands the context behind this change. Can we get an SOP in the wiki about ephemeral OS disks and what the issues is with them.  

As a summary, an ephemeral OS disk means that when the VM reboots, the OSDisk is wiped.  What that means is when it tries to rejoin the cluster under the same node name, the certificate that it uses to communicate with the kube-apiserver is gone and it is no longer trusted.  the kube-apiserver essentially sees a node which it does not trust and could assume that someone maliciously placed this node here. 

What will then happen is the machine health check operator should come by after 300s and clobber the machine, resulting in a new node being created.

Then a new node will come online.  

The reason we don't want to allow redeploys of this, is that it will never be successful.  in the SOP we should have context on proper steps to perform if a node with an ephemeral disk goes not ready.  Typically the MHC controller will just delete this and all will be fine, but during OCP upgrades where nodes reboot we're also going to see interesting results. 

@darthhexx not sure if there's a story to detect this in basic checks, but we probably want one.  If one doesn't already exist, would you mind creating one?

darthhexx:
The [work item](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14751145/) already exists and there is also an [approved PR](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO-Scripts/pullrequest/6245393) ready to be merged.

darthhexx (changes requested):
Looks good so far, but we only want to prevent the redeploy if we have a positive match, otherwise we may well be changing the default behaviour, i.e. potentially when `vm.StorageProfile.OsDisk == nil`.
The logic could be reduced to a single if statement, looking for the exact case we are checking for:
```bash
if vmDisk := vm.StorageProfile.OsDisk; vmDisk != nil && vmDisk.DiffDiskSettings != nil &&
    vmDisk.Caching == "ReadOnly" && vmDisk.DiffDiskSettings.Option == "Local" && vmDisk.DiffDiskSettings.Placement == "CacheDisk" {
    return api.NewCloudError( .... )
}
```

darthhexx (approved):
LGTM

PR #2231, opened 2022-07-07
Hive: preparation for adoption & reconciliation on update

m1kola:

Preparation for [WI 14895480](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14895480)

`ClusterManager` interface and implementation is now updated to support both creation and update of the hive resources which enables both post-installation adoption and adoption of the existing cluster fleet.

This is roughly how I intent to use it for update: https://github.com/Azure/ARO-RP/pull/2232/commits/a36c0e3d3fcc5fbf2ff4458d7ceeec1880c184f7

### Test plan for issue:

This new code is still disabled at the moment. Green e2e will confirm that we did not break anything.

No, it is refactoring. Code is a no-op at the moment.


bennerv (approved):
lgtm

pkg/hive/manager.go
```diff
@@ -57,44 +60,63 @@ func NewClusterManagerFromConfig(restConfig *rest.Config) (ClusterManager, error
...
-	var namespace string
-
+func (hr *clusterManager) CreateNamespace(ctx context.Context) (*corev1.Namespace, error) {
+	var namespaceName string
+	var namespace *corev1.Namespace
 	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		namespace = "aro-" + uuid.Must(uuid.NewV4()).String()
-		csn := ClusterNamespace(namespace)
-		_, err := hr.kubernetescli.CoreV1().Namespaces().Create(ctx, csn, metav1.CreateOptions{})
+		namespaceName = "aro-" + uuid.Must(uuid.NewV4()).String()
```
hawkowl, line 84: I still don't like this :) I raised a PR so that we can perhaps get the ID we need easier. https://github.com/Azure/ARO-RP/pull/2236

m1kola: I left comments in your PR. Let's continue this discussion there. I'm happy to change this when/if we find a solution for this.

We have time to make this change since, despite being in master already, this code is a no-op for now (no hive in prod at the moment).

ross-bryan (commented):
lgtm

pkg/hive/manager.go
```diff
@@ -57,44 +60,63 @@ func NewClusterManagerFromConfig(restConfig *rest.Config) (ClusterManager, error
...
+
+func (hr *clusterManager) CreateOrUpdate(ctx context.Context, parameters *CreateOrUpdateParameters) error {
+	resources := []kruntime.Object{
+		kubeAdminSecret(parameters.Namespace, []byte(parameters.KubeConfig)),
+		servicePrincipalSecret(parameters.Namespace, []byte(parameters.ServicePrincipal)),
+		clusterDeployment(parameters.Namespace, parameters.ClusterName, parameters.ClusterID, parameters.InfraID, parameters.Location),
+	}
 
-	_, err = hr.kubernetescli.CoreV1().Secrets(namespace).Create(ctx, kubesecret, metav1.CreateOptions{})
+	err := dynamichelper.Prepare(resources)
```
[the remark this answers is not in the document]
hawkowl: During install, it definitely updates the spec once it pulls things like the kubeadmin SSH key out.

PR #2232, opened 2022-07-07
Hive: cluster adoption & resource reconciliation on update

m1kola:

Fixes [WI 14895480](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14895480).

Modifies Update and AdminUpdate

Adds steps to reconcile hive resources which covers two things:
* Initial cluster adoption
* Hive resources reconciliation on updates (admin and customer initiated) to make sure that the hive has the latest configuration and secrets

### Test plan for issue:

This new code is still disabled at the moment. Green e2e will confirm that we did not break anything.

No, it is refactoring. Code is a no-op at the moment.


ross-bryan (approved):
lgtm

PR #2236, opened 2022-07-08
Provide the ClusterID via the Admin API

hawkowl:
Part of the M4 work

Exposes the cluster ID in the admin API, since right now it's not very easily available.

This means we can avoid using a randomly generated namespace in Hive because now the ID we already have is stable.

### Test plan for issue:
Added unit tests for the GET for the admin API, which was missing.

N/A

### DIFF
--- a/pkg/api/admin/openshiftcluster.go
+++ b/pkg/api/admin/openshiftcluster.go
@@ -55,6 +55,7 @@ type OpenShiftClusterProperties struct {
 	ImageRegistryStorageAccountName string                  `json:"imageRegistryStorageAccountName,omitempty"`
 	InfraID                         string                  `json:"infraId,omitempty"`
 	HiveProfile                     HiveProfile             `json:"hiveProfile,omitempty"`
+	ClusterID                       string                  `json:"clusterId,omitempty"`
 }
 
 // ProvisioningState represents a provisioning state.
--- a/pkg/api/admin/openshiftclusterdocument_convert.go
+++ b/pkg/api/admin/openshiftclusterdocument_convert.go
@@ -1,0 +1,35 @@
+package admin
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import "github.com/Azure/ARO-RP/pkg/api"
+
+type openShiftClusterDocumentConverter struct{}
+
+// ToExternalList returns a slice of external representations of the internal
+// objects
+func (c *openShiftClusterDocumentConverter) ToExternalList(docs []*api.OpenShiftClusterDocument, nextLink string) interface{} {
+	l := &OpenShiftClusterList{
+		OpenShiftClusters: make([]*OpenShiftCluster, 0, len(docs)),
+		NextLink:          nextLink,
+	}
+
+	conv := &openShiftClusterConverter{}
+
+	for _, doc := range docs {
+		converted := conv.ToExternal(doc.OpenShiftCluster).(*OpenShiftCluster)
+		converted.Properties.ClusterID = doc.ID
+		l.OpenShiftClusters = append(l.OpenShiftClusters, converted)
+	}
+
+	return l
+}
+
+// ToExternal returns an external representation of the internal object
+func (c *openShiftClusterDocumentConverter) ToExternal(doc *api.OpenShiftClusterDocument) interface{} {
+	conv := &openShiftClusterConverter{}
+	converted := conv.ToExternal(doc.OpenShiftCluster).(*OpenShiftCluster)
+	converted.Properties.ClusterID = doc.ID
+	return converted
+}
--- a/pkg/api/admin/register.go
+++ b/pkg/api/admin/register.go
@@ -15,6 +15,9 @@ func init() {
 		OpenShiftClusterConverter: func() api.OpenShiftClusterConverter {
 			return &openShiftClusterConverter{}
 		},
+		OpenShiftClusterDocumentConverter: func() api.OpenShiftClusterDocumentConverter {
+			return &openShiftClusterDocumentConverter{}
+		},
 		OpenShiftClusterStaticValidator: func(string, string, bool, string) api.OpenShiftClusterStaticValidator {
 			return &openShiftClusterStaticValidator{}
 		},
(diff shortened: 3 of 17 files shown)


pkg/api/admin/openshiftcluster.go
```diff
@@ -55,6 +55,7 @@ type OpenShiftClusterProperties struct {
 	ImageRegistryStorageAccountName string                  `json:"imageRegistryStorageAccountName,omitempty"`
 	InfraID                         string                  `json:"infraId,omitempty"`
 	HiveProfile                     HiveProfile             `json:"hiveProfile,omitempty"`
+	ClusterID                       string                  `json:"clusterId,omitempty"`
```
m1kola, line 58: Background: https://github.com/Azure/ARO-RP/pull/2145#discussion_r913319980

1. `ClusterProfile` looks likea good place for `ClusterID`, but
2. I'm not in faviour of introducing magic/virtual fields here or in any other external representation: each field in the external representation should have a corresponding field in the internal representation (CosmosDB). It is the approach similar to what Kubernetes has with internal (etcd) / external representation (API) of objects.
3. And changes with converters look hacky (very subjective obviously).

IMO if we really, really want to use cluster doc ID we need to look for alternative options. Maybe duplicate copy doc id into `ClusterProfile.ClusterID`? This will require us to run a migration on all the DB records (even failed ones). For we clusters we need to update [this](https://github.com/Azure/ARO-RP/blob/6a5f514b68b8e49f3e6d2268525529cf4dd74fab/pkg/frontend/openshiftcluster_putorpatch.go#L68-L86).

But I'm still not convinced that the amount of changes required for this to happen worth the benefit: I do not see a trouble in having to do 1 extra lookup from RP for a hive namespace when working backwards from a cluster. Maybe I'm missing something and need more explanation why this is an issue?

m1kola (changes requested):

pkg/api/admin/openshiftcluster.go
hawkowl: We already do have magic virtual fields in external representations -- the enricher runs on this API which populates a bunch of virtual fields.

hawkowl: I was also considering the approach of moving _all_ the converters to take a Document, but that would require a few more changes here and there, so this is fairly minimal by comparison. I could do it consistently if not having two sets of converters seems good.

I also did think about making it a non-virtual field, but the database migration seems meh.

My major concern is that a UUID is long, hard to remember, and in a rush is easy to mix up. If we have some systems that have UUIDs (e.g. Telemeter to OCM: https://github.com/openshift/cluster-monitoring-operator/blob/838a238342b2b1ab5c99a18bd271a7b15a1acbd1/pkg/operator/operator.go#L856-L872 ) then we also need to be able to get from a UUID in OCM to a cluster resource ID (which we currently can't easily do).

I'm also extremely wary of operator fatigue with having to correlate two extremely long unique IDs at once while debugging.

bennerv: First off +1 on the approach.  Less ids == better.  

> We already do have magic virtual fields in external representations -- the enricher runs on this API which populates a bunch of virtual fields.

I think what Mikalai means is that we're adding the `ClusterID` to the admin API only, but not to the internal representation. AFAIK the enricher updates non-virtual fields as they are represented within the internal API (clusterversion, master profile, worker profile, ingress profiles, etc. are all within `pkg/api/openshiftcluster.go`). This `ClusterID` field is only being added to `pkg/api/admin/openshiftcluster.go`.  Unless I'm missing a field the enricher updates which isn't in our internal API representation. 

I think the easier approach instead of adding an additional API endpoint + Geneva Action would be to perform a database migration + add the field to the internal representation instead of creating an additional admin endpoint + corresponding GA.  The GA won't provide as much benefit as majority of the fields are not as useful to us (since they're generated).  We already have portions of our code in `AdminUpdate` where we require population of fields, and we should have full access to the internal cluster doc, which gives us the UUID which we can populate the `ClusterId` as a result.

hawkowl: @bennerv to note, this doesn't add an additional API endpoint nor Geneva Action, since that already exists :) The tests are not for testing a brand-new endpoint, but for coverage of the existing GET endpoint w/ the API version being 'admin'. ListClusters is an admin action, but GET is shared with the public API.

bennerv: Thinking about my last comment more:
>I think less ids are better especially since we're dealing with a lot of them.

Why do we care that the document ID matches the Hive namespace?  We never actually read the document ID out at all nor do we natively have access to view it.  How will we confuse it with anything?

m1kola: Re: working from cluster -> hive. I think we can use `InfraID` for that: we have it everywhere: in the cluster, in the DB and in ClusterDeployment.

It, however, is not guaranteed to be unique.

hawkowl: @m1kola Can we make InfraID unique somehow?

m1kola: @hawkowl I can not think of a way which 100% guaratees that there will be no collisions within a Hive cluster. Especially when/if we move infra id generation to the upstream installer.

---

Overall - I think we are spinning wheels on something which is not an issue. From this thread and from [the previous one](https://github.com/Azure/ARO-RP/pull/2145#discussion_r913319980) the thing which concerns me the most is the fact that we use UUID for the namespace name (even if we re-use `InfraID` or cluster doc ID - these are still not human friendly UUIDs). This will be inconvenient for the cases where we need to manually go to the Hive cluster. But I think these cases will be rare as we will be sending logs from Hive and from the installer into Geneva so it should be possible to troubleshoot provisioning/deprovisioning and reconciliation issues without accessing a Hive cluster.

Given the above I propose two options:

1. Leave everything as is (new UUID generated for every cluster specifically for a namespace)
2. Instead of UUID, generate something like `aro-invulnerably-taled-lama` for a namespace. Solves human readability problem.

Both options can guarantee that namespace is unique within the Hive cluster.

hawkowl:
Seems like we're not heading in this direction.

PR #2239, opened 2022-07-11
Update getUserDataSecretReference func and tests

yehornaumenko:
These changes extends tests of getUserDataSecretReference function that have been added in [another PR](https://github.com/Azure/ARO-RP/pull/2224). (Check @m1kola's comments)


bennerv (approved):
lgtm

PR #2240, opened 2022-07-11
Hive cluster manager: do not error deletion on 404

m1kola:

Do not return an error when there is no cluster namespace in the Hive AKS cluster.

### Test plan for issue:

Code path is disabled at the moment. Existing E2E should indicate a regression, if any.

No. Small change in a disabled code.


ross-bryan (approved):
lgtm

PR #2243, opened 2022-07-12
Add AppLens Client and Detector Endpoints

zacharyljones:

[Feature 10651414](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/10651414)

This PR adds API endpoints that will enable AppLens Detectors in the Diagnose & Solve Problems Blade in the Azure Portal for ARO clusters.

### Test plan for issue:

I have tested that I am able to hit the appropriate AppLens endpoints locally and receive a 401: Unauthorized error. There is no way for me to fully test this change in my dev environment because the "public" AppLens endpoints exist within the AME tenant and cross-tenant communication is not allowed. It may be possible to setup a dev RP in the AME tenant, but this is a conversation that I will have with Red Hat/MS offline.

No, the AppLens team has generic [documentation](https://applensdiagnostics.azurewebsites.net/?p=79) for performing this integration with any service (necessary api endpoints, AppLens service endpoints, etc.)


pkg/util/cluster/cluster.go
```diff
@@ -135,36 +135,17 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 	}
 
 	c.log.Infof("creating AAD application")
-	// This is for our CFS scenario where we do not have permissions to allow the RP to create it's own AAD application
```
zacharyljones: Changes to this file were specific to a local development scenario and were accidentally committed. All changes to this file have been reverted.

pkg/util/cluster/cluster.go
```diff
@@ -135,14 +135,32 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 	}
 
 	c.log.Infof("creating AAD application")
-	appID, appSecret, err := c.createApplication(ctx, "aro-"+clusterName)
```
zacharyljones, line 138: Changes to this file were specific to a local development scenario and were accidentally committed. All changes to this file have been reverted.

zacharyljones:
@weinong FYSA

zacharyljones:
The AppLens Client was modeled after the Azure [CosmosDB Client](https://github.com/Azure/azure-sdk-for-go/blob/3f7acd20691214ef2cb1f0132f82115f1df01a8c/sdk/data/azcosmos/cosmos_client.go) in the Azure SDK for Go. Unfortunately this new client cannot be added to the Azure SDK for Go, as the AppLens service is technically private and there is no need for end users to have access to the client.

zacharyljones:
@coreyperkins FYSA

pkg/env/env.go
```diff
@@ -79,6 +80,7 @@ type Interface interface {
 	Domain() string
 	FeatureIsSet(Feature) bool
 	FPAuthorizer(string, string) (refreshable.Authorizer, error)
+	FPNewClientCertificateCredential(string) (*azidentity.ClientCertificateCredential, error)
```
weinong, line 83: we should use `azcore.TokenCredential` interface

pkg/frontend/adminactions/azureactions.go
```diff
@@ -5,6 +5,7 @@ package adminactions
```
weinong, line 5: why does admin actions need applens?

pkg/frontend/adminactions/azureactions.go
```diff
@@ -102,3 +113,20 @@ func (a *azureActions) VMResize(ctx context.Context, vmName string, size string)
 	vm.HardwareProfile.VMSize = mgmtcompute.VirtualMachineSizeTypes(size)
 	return a.virtualMachines.CreateOrUpdateAndWait(ctx, clusterRGName, vmName, vm)
 }
+
+func (a *azureActions) AppLensGetDetector(ctx context.Context, detectorId string) ([]byte, error) {
+	resp, err := a.appLens.GetDetector(ctx, &applens.GetDetectorOptions{ResourceID: a.oc.Properties.ClusterProfile.ResourceGroupID, DetectorID: detectorId})
```
weinong, line 118: `ResourceGroupID`?

pkg/frontend/adminactions/azureactions.go
zacharyljones: I added it to Azure Actions based on nothing more than the fact that this would be connecting to an Azure endpoint. Azure Actions does fall under Admin Actions, but the endpoints I added are not admin endpoints.

A separate actions class could be made if desired/need be.

pkg/util/azureclient/applens/applens.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+type appLensClient struct {
+	*Client
+}
+
+var _ AppLensClient = &appLensClient{}
+
+// NewAppLensClient returns a new AppLensClient
+func NewAppLensClient(env *azureclient.AROEnvironment, clientCertCred *azidentity.ClientCertificateCredential) AppLensClient {
```
weinong, line 28: let's also use `azcore.TokenCredential`

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -0,0 +1,163 @@
...
+		&options.ClientOptions)
+}
+
+func createScopeFromEndpoint(endpoint string) ([]string, error) {
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	return []string{fmt.Sprintf("%s://%s/.default", u.Scheme, u.Hostname())}, nil
```
weinong, line 63: i'm pretty sure the endpoint is not the scope.

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -0,0 +1,163 @@
...
+			for k, v := range *headers {
+				req.Raw().Header.Set(k, v)
+			}
+		}
+	}
+
+	id := uuid.Must(uuid.NewV4()).String()
+	req.Raw().Header.Set(headerXmsClientRequestId, id)
+	req.Raw().Header.Set(headerXmsDate, time.Now().UTC().Format(http.TimeFormat))
+	req.Raw().Header.Set(headerXmsRequestId, id)
```
weinong, line 142: are these headers from CosmosDB? AKS uses different headers
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/toolkit/azureclients/applensclient/applens_client.go&version=GBmaster&line=19&lineEnd=22&lineStartColumn=1&lineEndColumn=32&lineStyle=plain&_a=contents

pkg/util/azureclient/applens/applens_headers.go
```diff
@@ -0,0 +1,12 @@
...
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+const (
+	headerAuthorization      string = "Authorization"
+	headerXmsClientRequestId string = "x-ms-client-request-id"
+	headerXmsDate            string = "x-ms-date"
+	headerXmsPathQuery       string = "x-ms-path-query"
+	headerXmsRequestId       string = "x-ms-request-id"
```
weinong, line 11: these headers are not used by AKS.

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	return azResponse, nil
+}
+
+func (c *Client) sendGetRequest(
+	ctx context.Context,
+	requestOptions appLensRequestOptions,
+	requestEnricher func(*policy.Request)) (*http.Response, error) {
+	req, err := c.createRequest(ctx, http.MethodGet, requestOptions, requestEnricher)
```
weinong, line 112: AppLens api is Post, iirc

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -0,0 +1,163 @@
...
+// NewClient creates a new instance of AppLens client with Azure AD access token authentication. It uses the default pipeline configuration.
+// endpoint - The applens service endpoint to use.
+// cred - The credential used to authenticate with the applens service.
+// options - Optional AppLens client options.  Pass nil to accept default values.
+func NewClient(endpoint string, cred azcore.TokenCredential, o *ClientOptions) (*Client, error) {
+	scope, err := createScopeFromEndpoint(endpoint)
+	if err != nil {
+		return nil, err
+	}
+	return &Client{endpoint: endpoint, pipeline: newPipeline([]policy.Policy{runtime.NewBearerTokenPolicy(cred, scope, nil), &appLensBearerTokenPolicy{}}, o)}, nil
```
weinong, line 41: you will need to update the transport of http client like
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/toolkit/azureclients/applensclient/applens_client.go&version=GBmaster&line=43&lineEnd=49&lineStartColumn=1&lineEndColumn=5&lineStyle=plain&_a=contents

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func newPipeline(authPolicy []policy.Policy, options *ClientOptions) runtime.Pipeline {
```
weinong, line 44: you will need to update the transport of http client like
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/toolkit/azureclients/applensclient/applens_client.go&version=GBmaster&line=43&lineEnd=49&lineStartColumn=1&lineEndColumn=5&lineStyle=plain&_a=contents

weinong:
can you also add user agent please?

pkg/frontend/adminactions/azureactions.go
weinong: i see. i think that's fine by me

pkg/frontend/adminactions/azureactions.go
zacharyljones: Updated to use cluster resource id, and not cluster resource group resource id

pkg/frontend/adminactions/azureactions.go
```diff
@@ -102,3 +113,20 @@ func (a *azureActions) VMResize(ctx context.Context, vmName string, size string)
...
+
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(resp.Body)
+}
+
+func (a *azureActions) AppLensListDetectors(ctx context.Context) ([]byte, error) {
+	resp, err := a.appLens.ListDetectors(ctx, &applens.ListDetectorsOptions{ResourceID: a.oc.Properties.ClusterProfile.ResourceGroupID})
```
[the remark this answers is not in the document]
zacharyljones: Updated to use cluster resource id, and not cluster resource group resource id

pkg/util/azureclient/applens/applens_client.go
zacharyljones: This is a carry over from the Cosmos client - I figured I would resolve any issues related to the scope during testing as I wasn't sure whether it would be needed or not. Do you think the scope should be something else, or removed altogether?

pkg/util/azureclient/applens/applens_client.go
zacharyljones: I think I understand what you are saying - I will look at adding those transport properties to the policies being created in the applens client.

pkg/env/env.go
zacharyljones: I will take a look - the current FPauthorizer uses cert based authentication to AAD, so I thought that made the most sense at the time. After our conversation yesterday, it sounds like we want to use the FP certificates to create an azcore.TokenCredential and then authenticate to AppLens using that credential. Is that correct?

weinong: correct. that being said, `azidentity.ClientCertificateCredential` is also `azcore.TokenCredential`

pkg/util/azureclient/applens/applens.go
weinong: `azidentity.ClientCertificateCredential` is the implementation of `azcore.TokenCredential` interface

pkg/util/azureclient/applens/applens_client.go
weinong: you have to have scope to get AAD token. in ADAL, it's called ResourceID. In MSAL, it's called Scope.
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/resourceprovider/server/microsoft.com/containerservice/templates/implementation.go&version=GBmaster&line=676&lineEnd=677&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

pkg/env/env.go
zacharyljones: I attempted to use the `azcore.TokenCredential` interface, but ran into an issue with code gen. The mock env file that gets generated imports `github.com/Azure/azure-sdk-for-go/sdk/azcore/internal/shared` instead of just `github.com/Azure/azure-sdk-for-go/sdk/azcore`. Similarly, all references to `azcore.TokenCredential` are replaced with `shared.TokenCredential`. This causes go validation to fail, unit tests to fail, etc.

A bit of research turned up this [azure-sdk-for-go issue](https://github.com/Azure/azure-sdk-for-go/issues/17450#issuecomment-1095280305). Apparently this is a bug/issue with how older versions of the SDK 
 export symbols. Some aspects of this issue have been fixed but none of those changes are available in the go 1.17 compatible version of the SDK that I am using.

Sticking with `azidentity.ClientCertificateCredential` for now.

pkg/frontend/adminactions/azureactions.go
zacharyljones: Refer to [previous comment](https://github.com/Azure/ARO-RP/pull/2243#discussion_r920620112)

pkg/util/azureclient/applens/applens.go
zacharyljones: Refer to [previous comment](https://github.com/Azure/ARO-RP/pull/2243#discussion_r920620112)

pkg/util/azureclient/applens/applens_client.go
zacharyljones: Client updated to use Post. Unit tests also updated.

pkg/util/azureclient/applens/applens_client.go
zacharyljones: Added default transporter properties to `applens_client_options.go`

zacharyljones:
> > > can you also add user agent please?
> > 
> > 
> > @weinong Can you elaborate on this? I see that it is added in AKS, but I do not see it in the [AppLens example](https://github.com/ShekharGupta1988/AppServiceDiagnosticsAPISample/blob/master/src/SampleAPIServer/Services/AppServiceDiagnosticsClientService.cs). What does the user agent afford us?
> 
> it helps identify the clients from server side log.

Added default user agent properties (TelemetryOptions) to `applens_client_options.go`

pkg/util/azureclient/applens/applens_client.go
zacharyljones: Understood - with that information I was able to use the appropriate scopes provided by the AppLens team. The code for creating the scope from the target endpoint has been removed.

pkg/util/azureclient/applens/applens_client_options.go
```diff
@@ -0,0 +1,46 @@
...
+	"crypto/tls"
+	"net/http"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+const (
+	userAgent = "AppLens-Client"
```
weinong, line 16:
```suggestion
	userAgent = "ARO-AppLens-Client"
```
```suggestion
	userAgent = "AppLens-Client"
```

pkg/util/azureclient/applens/applens_policy_bearer_token.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func (b *appLensBearerTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
+	currentAuthorization := req.Raw().Header.Get(headerAuthorization)
+	if currentAuthorization == "" {
+		return nil, errors.New("authorization header is missing")
+	}
+
+	token := currentAuthorization[lenBearerTokenPrefix:]
+	req.Raw().Header.Set(headerAuthorization, fmt.Sprintf("type=aad&ver=1.0&sig=%v", token))
```
weinong, line 26: AKS doesn't do this, right?

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -0,0 +1,148 @@
...
+	ctx context.Context,
+	method string,
+	requestOptions appLensRequestOptions,
+	requestEnricher func(*policy.Request)) (*policy.Request, error) {
+	req, err := runtime.NewRequest(ctx, method, c.endpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	if requestOptions != nil {
```
weinong, line 115: i think `policy.WithHttpHeader` is more idiomatic

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (c *Client) createRequest(
+	ctx context.Context,
+	method string,
+	requestOptions appLensRequestOptions,
+	requestEnricher func(*policy.Request)) (*policy.Request, error) {
+	if requestOptions != nil {
+		header := requestOptions.toHeader()
+		ctx = policy.WithHTTPHeader(ctx, header)
```
weinong, line 110: IMO, the headers that applens server required should be added here instead from `requestOptions`. In fact, if we do that, there is no need of `appLensRequestOptions` at all. Although I like your design to allow customization, I think it can be simplified for this client.

weinong: nvm. I see that options is used to pass in the resource ID. fine :)

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -0,0 +1,137 @@
...
+func (c *Client) Endpoint() string {
+	return c.endpoint
+}
+
+// NewClient creates a new instance of AppLens client with Azure AD access token authentication. It uses the default pipeline configuration.
+// endpoint - The applens service endpoint to use.
+// cred - The credential used to authenticate with the applens service.
+// options - Optional AppLens client options.  Pass nil to accept default values.
+func NewClient(endpoint, scope string, cred azcore.TokenCredential, o *ClientOptions) (*Client, error) {
+	return &Client{endpoint: endpoint, pipeline: newPipeline([]policy.Policy{runtime.NewBearerTokenPolicy(cred, []string{scope}, nil), &appLensBearerTokenPolicy{}}, o)}, nil
```
weinong, line 33: I don't think we need our own `appLensBearerTokenPolicy`. The built-in BearerTokenPolicy should do.

zacharyljones:
@weinong I believe all of your comments have been resolved. If there is nothing else, I will remove the change the tags on this PR and start getting feedback from Red Hat.

weinong:
LGTM

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -0,0 +1,138 @@
...
+func (c *Client) Endpoint() string {
+	return c.endpoint
+}
+
+// NewClient creates a new instance of AppLens client with Azure AD access token authentication. It uses the default pipeline configuration.
+// endpoint - The applens service endpoint to use.
+// cred - The credential used to authenticate with the applens service.
+// options - Optional AppLens client options.  Pass nil to accept default values.
+func NewClient(endpoint, scope string, cred azcore.TokenCredential, o *ClientOptions) (*Client, error) {
+	return &Client{endpoint: endpoint, pipeline: newPipeline([]policy.Policy{runtime.NewBearerTokenPolicy(cred, []string{fmt.Sprintf("%s/.default", scope)}, nil)}, o)}, nil
```
zacharyljones, line 34: @weinong Even though I cannot connect to the public AppLens endpoints in my local environment, I wanted to make sure that I could at least see the unauthorized errors in the RP logs. On the first attempt, I received this error:

image


So I updated the client to always suffix "./default" to the end of the scope.

pkg/util/azureclient/applens/applens_headers.go
```diff
@@ -0,0 +1,12 @@
...
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+const (
+	headerXmsClientRequestId string = "x-ms-client-request-id"
+	headerXmsDate            string = "x-ms-date"
+	headerXmsPathQuery       string = "x-ms-path-query"
+	headerXmsRequestId       string = "x-ms-request-id"
+	headerXmsVerb            string = "x-ms-verb"
```
zacharyljones, line 11: > So I updated the client to always suffix "./default" to the end of the scope

After fixing the scope, I received this error:

image

So I added `the x-ms-verb` header and included it in the request the same way all other headers are added.

zacharyljones (commented):
At this point, I have verified that I am attempting to connect to the target AppLens endpoint, but am receiving unauthorized errors, which I believe is as far as I can test in the dev environment.

image

pkg/util/azureclient/environments.go
```diff
@@ -25,6 +29,9 @@ var (
 		ActualCloudName:          "AzureCloud",
 		GenevaMonitoringEndpoint: "https://gcs.prod.monitoring.core.windows.net/",
 		AppSuffix:                "aro.azure.com",
+		AppLensEndpoint:          "https://diag-runtimehost-staging.azurefd.net/api/invoke",
```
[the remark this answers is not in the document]
ross-bryan: There is not a pattern for that today. The RP Config is only used in the `deploy` binary. Without changing anything architecturally, what that could do is plumb the RP Config into environment variables to be pulled into the RP when it inits. Otherwise we'l need to look at adding the RP Config as a parameter to other commands

pkg/frontend/frontend.go
```diff
@@ -217,6 +217,18 @@ func (f *frontend) authenticatedRoutes(r *mux.Router) {
 
 	s.Methods(http.MethodPost).HandlerFunc(f.postOpenShiftClusterKubeConfigCredentials).Name("postOpenShiftClusterKubeConfigCredentials")
 
+	s = r.
```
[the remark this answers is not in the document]
ross-bryan: I see https://github.com/Azure/ARO-RP/pull/2243/files#r919542725 - there are only 2 classes. Customer API or SRE API. If this is SRE API then it is an admin API

ross-bryan (changes requested):
Hi Zach, this looks like its off to a good start. I did a quick look over and this will take me a couple hours to properly go through, but here is initial thoughts/feedback/response

pkg/frontend/frontend.go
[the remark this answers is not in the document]
zacharyljones: I would label them customer facing APIs - Really these endpoints will only be called from the Azure portal in order to enable the "Diagnose and solve problems" blade for ARO clusters. Here is an example from an AKS cluster:

image

pkg/util/azureclient/environments.go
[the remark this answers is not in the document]
zacharyljones: Understood - I don't think there is necessarily a problem with the way these values are being handled right now. Maybe one of the suggestions you mentioned could be implemented for all fields later on then.

pkg/frontend/frontend.go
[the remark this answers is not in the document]
ross-bryan: if this is part of a versioned api release this needs to be planned and coordinated with our releases. we have one coming in september, and then likely a q4 release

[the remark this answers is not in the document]
zacharyljones: Understood - I don't believe this needs to be part of a versioned api release. The AKS RP does take in an api-version parameter for this call (as you can see from the screenshot), but I do not see it passed to AppLens in any way. I also don't see an api version mentioned in the AppLens docs. To me this indicates that the AppLens calls do not need to be associated with an API version.

@weinong am I overlooking anything here?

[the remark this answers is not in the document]
ross-bryan: API is proxied through ARM and it will require ARM manifest with swagger spec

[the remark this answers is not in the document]
weinong: @ross-bryan swagger change is not required as this is not customer facing api. At least I've not seen anyone doing that.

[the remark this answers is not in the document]
weinong: @zacharyljones that being said, we need to add these two apis to https://docs.microsoft.com/en-us/rest/api/openshift/operations/list?tabs=HTTP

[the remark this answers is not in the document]
ross-bryan: Thanks @weinong, TIL!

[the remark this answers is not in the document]
zacharyljones: @weinong I believe I would add these two apis to the [getOperations endpoint](https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/operations_get.go). I'm assuming the `Origin` field for each of these operations would be `system` because they aren't for end users. @ross-bryan is that understanding of the `Origin` field correct?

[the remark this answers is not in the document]
weinong: we use system,user in AKS as this api is called by user identity from portal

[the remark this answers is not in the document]
zacharyljones: @weinong I added a `detectors/read` entry to the `operations` endpoint, similar to AKS.

@ross-bryan Should I be able to hit the `operations` endpoint locally?

image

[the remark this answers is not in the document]
ross-bryan: @zacharyljones   requires api version https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/frontend.go#L306
```
curl -k \
  "https://localhost:8443/providers/Microsoft.RedHatOpenShift/operations?api-version=2020-04-30"
  ```

**Comment by rogbas** [comment]
I'm assuming AppLens relies on an Azure service of same name, correct? Have we checked if it is available in all the Azure regions we support? cc @weinong

**Comment by weinong** [comment]
> I'm assuming AppLens relies on an Azure service of same name, correct? Have we checked if it is available in all the Azure regions we support? cc @weinong

yes, it's available in public and fairfax

File: pkg/util/azureclient/applens/applens_client.go [hunk 23]
```diff
@@ -0,0 +1,138 @@
...
+func (c *Client) GetDetector(
+	ctx context.Context,
+	o *GetDetectorOptions) (*http.Response, error) {
+	if o == nil {
+		o = &GetDetectorOptions{}
+	}
+
+	azResponse, err := c.sendPostRequest(
+		ctx,
+		o,
```
s-amann, line 83: `o` is used for `options` all over. Can we rename this to `options`?

pkg/util/azureclient/applens/mock.go
```diff
@@ -0,0 +1,369 @@
+package applens
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+// Code taken from github.com/Azure/azure-sdk-for-go/internal/mock
```
s-amann, line 5: Should we be publicizing code under `/internal`? Either this directory doesn't exist in `azure-sdk-for-go` or I don't have access to view it.

pkg/util/azureclient/environments.go
```diff
@@ -33,6 +40,9 @@ var (
 		ActualCloudName:          "AzureUSGovernment",
 		GenevaMonitoringEndpoint: "https://gcs.monitoring.core.usgovcloudapi.net/",
 		AppSuffix:                "aro.azure.us",
+		AppLensEndpoint:          "https://diag-runtimehost-prod-bn1-001.azurewebsites.us/api/invoke",
+		AppLensScope:             "https://microsoft.onmicrosoft.com/runtimehost",
```
s-amann, line 44: Why is the scope here a URL?

pkg/util/azureclient/applens/mock.go
cfsmikedrum: "sdk" was left out of the path.
It should be: [github.com/Azure/azure-sdk-for-go/sdk/internal/mock](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/internal/mock)

pkg/util/azureclient/applens/applens_client.go
s-amann: oh I see, thanks for sharing that wiki. I can push aside my want for more descriptive local variables in favor of that recommendation.

pkg/util/azureclient/applens/mock.go
s-amann: ah, that explains it, thank you!

cfsmikedrum:
@s-amann, the rebase is complete

pkg/frontend/adminactions/azureactions.go
```diff
@@ -56,6 +61,11 @@ func NewAzureActions(log *logrus.Entry, env env.Interface, oc *api.OpenShiftClus
 		return nil, err
 	}
 
+	fpClientCertCred, err := env.FPNewClientCertificateCredential(subscriptionDoc.Subscription.Properties.TenantID)
```
weinong, line 64: @cfsmikedrum @zacharyljones I think this is wrong. The AppLens' tenant ID is fixed per cloud:
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/resourceprovider/server/microsoft.com/containerservice/templates/implementation.go&version=GBmaster&line=645&lineEnd=645&lineStartColumn=61&lineEndColumn=76&lineStyle=plain&_a=contents

cfsmikedrum: Nont and I worked together to correct the Tenant ID issue, as well as update the go.mod file.

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -0,0 +1,138 @@
...
+	return req, nil
+}
+
+func (c *Client) executeAndEnsureSuccessResponse(request *policy.Request) (*http.Response, error) {
+	response, err := c.pipeline.Do(request)
+	if err != nil {
+		return nil, err
+	}
+
+	successResponse := (response.StatusCode >= 200 && response.StatusCode < 300) || response.StatusCode == 304
```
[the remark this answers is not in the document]
cadenmarchese: Agreed, I think it would be better to have a list of all success responses from applens and and evaluate based on that list rather than have a huge range.

cadenmarchese (approved):
Some outstanding comments here, but as discussed I'll merge this to unblock AppLens onboarding @s-amann

PR #2244, opened 2022-07-12
Refactor pkg/deploy to use go:embed instead of go-bindata

tony-schndr:

[Work item 14927288](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14927288/)

This PR removes the use of go-bindata within pkg/deploy and uses go:embed instead.

**Please note** that in order to facilitate this change, `deploy/` was moved to `pkg/deploy/staticassets/` as go:embed requires the file to be within the dir/subdir, (ie go:embed cannot embed files in a parent directory).

### Test plan for issue:
Unit test - The test `TestConfigurationFieldParity` in `pkg/deploy/config_test.go` shows the behavior/result of retrieving the asset's binary data is the same between go:embed and go-bindata.  There are a handful of other unit tests that indirectly validate this change, ie the tests show the files can be read and binary data can be processed.

Yes, path updates within `docs/`


tony-schndr:
@facchettos

tony-schndr:
Looks like there is one more `deploy/` path reference in [.gdn/.gdnsuppress](https://github.com/Azure/ARO-RP/blob/master/.gdn/.gdnsuppress#L13).  This looks to be generated by Guardian, I'm not entirely sure how to go about updating it.

pkg/deploy/deploy_rp.go
```diff
@@ -30,7 +31,7 @@ func (d *deployer) DeployRP(ctx context.Context) error {
 
 	deploymentName := "rp-production-" + d.version
 
-	b, err := Asset(generator.FileRPProduction)
+	b, err := staticassets.EmbeddedFiles.ReadFile(generator.FileRPProduction)
```
facchettos, line 34: could you rename the variable at the same time ?

deploy/generate.go
```diff
@@ -1,8 +0,0 @@
-package deploy
-
-// Copyright (c) Microsoft Corporation.
-// Licensed under the Apache License 2.0.
-
-//go:generate go run ../hack/gendeploy
```
facchettos, line 6: what happened to gendeploy here ? I could not find it somewhere else and we might need it to be run during make generate right? I think that passed through the net. 

Could we maybe put it in `pkg/deploy` (that one might require a slight refactor of `hack/gendeploy`) or in `pkg/deploy/assets` (after it has been renamed from `staticassets`)

facchettos (changes requested):
overall LGTM, thanks !  just a few things that I'd probably change. Could we also rename the variables with something meaningful (the ones from when you read the files), `b` is not a good variable name. I know it was named like that before, but since we're modifying the code there, we could also rename it at the same time. 

Since we are creating the directory, could we just use `assets` or `manifests` instead of `staticassets` ? We modify some of those in predeploy so it's not really static. 

There is a line in `deploy/generate.go` that we still need I think, but it has been removed. We'd need to get it back somewhere else so it is run as before

deploy/generate.go
tony-schndr: Good catch!  I'll get it added back.

pkg/deploy/deploy_rp.go
tony-schndr: Yes, I'll get this updated as well.

facchettos (approved):
lgtm now, I don't really know the guardian bit, but once you figure it out, that can be merged (will approve now, since I am off tomorrow)

coreyperkins:
> Looks like there is one more `deploy/` path reference in [.gdn/.gdnsuppress](https://github.com/Azure/ARO-RP/blob/master/.gdn/.gdnsuppress#L13). This looks to be generated by Guardian, I'm not entirely sure how to go about updating it.

@tony-schndr - here is a reference for how the .gdnsuppress file is produced. The short story is, you will not modify it by adjusting any paths. You will submit all your changes, then run your build. That build will produce a new .gdnsuppress file as an artifact with the updated paths. You will then download the new .gdnsuppress file and commit it to your branch and run the build again, it should pass Guardian at that point.

There are other ways to generate this file, but they require downloading windows executables, configuring them locally, etc. At the time we did this work this was the recommended workflow.

https://onebranch.visualstudio.com/OneBranch/_wiki/wikis/OneBranch.wiki/4640/Suppress-SDL-issues?anchor=download-.gdnsuppress-or-.gdnbaselines-file-from-the-failed-build

facchettos (dismissed):

bennerv (dismissed):
+1 - great work

bennerv (approved):
lgtm - ty for rebase.  merge on green

PR #2246, opened 2022-07-13
Fix broken mirror specific version arg

bennerv:

Fixes the mirroring code when you pass in a specific version to be mirrored.  It previously would fail, now it works :1st_place_medal: 

### Test plan for issue:

```bash
[bvesel@fedora ARO-RP]$ go run -tags aro ./cmd/aro mirror 4.10.15
INFO[2022-07-13T15:39:06-04:00]cmd/aro/main.go:54 main.main() starting, git commit unknown                 
INFO[2022-07-13T15:39:06-04:00]pkg/env/core.go:62 env.NewCoreForCI() running in local development mode            
INFO[2022-07-13T15:39:08-04:00]cmd/aro/mirror.go:126 main.mirror() mirroring release 4.10.15                    
INFO[2022-07-13T15:39:08-04:00]pkg/mirror/mirror.go:72 mirror.Mirror() reading imagestream from quay.io/openshift-release-dev/ocp-release@sha256:ddcb70ce04a01ce487c0f4ad769e9e36a10c8c832a34307c1b1eb8e03a5b7ddb 
```

No

### DIFF
--- a/cmd/aro/mirror.go
+++ b/cmd/aro/mirror.go
@@ -99,9 +99,19 @@ func mirror(ctx context.Context, log *logrus.Entry) error {
 					Payload: version.InstallStream.PullSpec,
 				})
 			} else {
+				vers, err := version.ParseVersion(arg)
+				if err != nil {
+					return err
+				}
+
+				node, err := pkgmirror.VersionInfo(vers)
+				if err != nil {
+					return err
+				}
+
 				releases = append(releases, pkgmirror.Node{
-					Version: arg,
-					Payload: arg,
+					Version: node.Version,
+					Payload: node.Payload,
 				})
 			}
 		}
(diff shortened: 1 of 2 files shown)


ArielLima (approved):
LGMT

cmd/aro/mirror.go
```diff
@@ -99,9 +99,19 @@ func mirror(ctx context.Context, log *logrus.Entry) error {
 					Payload: version.InstallStream.PullSpec,
 				})
 			} else {
+				vers, err := version.ParseVersion(arg)
+				if err != nil {
+					return nil
```
bizz001, line 104: Shouldn't we return `err` instead of `nil`?

bennerv: good catch - ty

hawkowl (approved):

PR #2248, opened 2022-07-13
Public IP removal

ArielLima:

Removes public IP from private clusters.

### Test plan for issue:
We deployed a full rp service in DEV with the current changes and created a cluster.

### DIFF
--- a/pkg/api/admin/openshiftcluster.go
+++ b/pkg/api/admin/openshiftcluster.go
@@ -147,7 +147,7 @@ type NetworkProfile struct {
 	PodCIDR      string       `json:"podCidr,omitempty"`
 	ServiceCIDR  string       `json:"serviceCidr,omitempty"`
 	MTUSize      MTUSize      `json:"mtuSize,omitempty"`
-	OutboundType OutboundType `json:"userDefinedRouting,omitempty"`
+	OutboundType OutboundType `json:"outboundType,omitempty" mutable:"true"`
 
 	APIServerPrivateEndpointIP string `json:"privateEndpointIp,omitempty"`
 	GatewayPrivateEndpointIP   string `json:"gatewayPrivateEndpointIp,omitempty"`
--- a/pkg/api/admin/openshiftcluster.go
+++ b/pkg/api/admin/openshiftcluster.go
@@ -130,14 +130,24 @@ const (
 	MTU3900 MTUSize = 3900
 )
 
+// OutboundType represents the type of routing a cluster is using
+type OutboundType string
+
+// OutboundType constants
+const (
+	OutboundTypeUserDefinedRouting OutboundType = "UserDefinedRouting"
+	OutboundTypeLoadbalancer       OutboundType = "Loadbalancer"
+)
+
 // NetworkProfile represents a network profile.
 type NetworkProfile struct {
 	// The software defined network (SDN) to use when installing the cluster.
 	SoftwareDefinedNetwork SoftwareDefinedNetwork `json:"softwareDefinedNetwork,omitempty"`
 
-	PodCIDR     string  `json:"podCidr,omitempty"`
-	ServiceCIDR string  `json:"serviceCidr,omitempty"`
-	MTUSize     MTUSize `json:"mtuSize,omitempty"`
+	PodCIDR      string       `json:"podCidr,omitempty"`
+	ServiceCIDR  string       `json:"serviceCidr,omitempty"`
+	MTUSize      MTUSize      `json:"mtuSize,omitempty"`
+	OutboundType OutboundType `json:"userDefinedRouting,omitempty"`
 
 	APIServerPrivateEndpointIP string `json:"privateEndpointIp,omitempty"`
 	GatewayPrivateEndpointIP   string `json:"gatewayPrivateEndpointIp,omitempty"`
--- a/pkg/api/admin/openshiftcluster_convert.go
+++ b/pkg/api/admin/openshiftcluster_convert.go
@@ -52,6 +52,7 @@ func (c openShiftClusterConverter) ToExternal(oc *api.OpenShiftCluster) interfac
 				PodCIDR:                    oc.Properties.NetworkProfile.PodCIDR,
 				ServiceCIDR:                oc.Properties.NetworkProfile.ServiceCIDR,
 				MTUSize:                    MTUSize(oc.Properties.NetworkProfile.MTUSize),
+				OutboundType:               OutboundType(oc.Properties.NetworkProfile.OutboundType),
 				APIServerPrivateEndpointIP: oc.Properties.NetworkProfile.APIServerPrivateEndpointIP,
 				GatewayPrivateEndpointIP:   oc.Properties.NetworkProfile.GatewayPrivateEndpointIP,
 				GatewayPrivateLinkID:       oc.Properties.NetworkProfile.GatewayPrivateLinkID,
@@ -193,6 +194,7 @@ func (c openShiftClusterConverter) ToInternal(_oc interface{}, out *api.OpenShif
 	out.Properties.NetworkProfile.PodCIDR = oc.Properties.NetworkProfile.PodCIDR
 	out.Properties.NetworkProfile.ServiceCIDR = oc.Properties.NetworkProfile.ServiceCIDR
 	out.Properties.NetworkProfile.MTUSize = api.MTUSize(oc.Properties.NetworkProfile.MTUSize)
+	out.Properties.NetworkProfile.OutboundType = api.OutboundType(oc.Properties.NetworkProfile.OutboundType)
 	out.Properties.NetworkProfile.SoftwareDefinedNetwork = api.SoftwareDefinedNetwork(oc.Properties.NetworkProfile.SoftwareDefinedNetwork)
 	out.Properties.NetworkProfile.APIServerPrivateEndpointIP = oc.Properties.NetworkProfile.APIServerPrivateEndpointIP
 	out.Properties.NetworkProfile.GatewayPrivateEndpointIP = oc.Properties.NetworkProfile.GatewayPrivateEndpointIP
--- a/pkg/api/openshiftcluster.go
+++ b/pkg/api/openshiftcluster.go
@@ -247,6 +247,15 @@ const (
 	MTU3900 MTUSize = 3900
 )
 
+// OutboundType represents the type of routing a cluster is using
+type OutboundType string
+
+// OutboundType constants
+const (
+	OutboundTypeUserDefinedRouting OutboundType = "UserDefinedRouting"
+	OutboundTypeLoadbalancer       OutboundType = "Loadbalancer"
+)
+
 // NetworkProfile represents a network profile
 type NetworkProfile struct {
 	MissingFields
@@ -255,6 +264,7 @@ type NetworkProfile struct {
 	ServiceCIDR            string                 `json:"serviceCidr,omitempty"`
 	SoftwareDefinedNetwork SoftwareDefinedNetwork `json:"softwareDefinedNetwork,omitempty"`
 	MTUSize                MTUSize                `json:"mtuSize,omitempty"`
+	OutboundType           OutboundType           `json:"userDefinedRouting,omitempty"`
 
 	APIServerPrivateEndpointIP string `json:"privateEndpointIp,omitempty"`
 	GatewayPrivateEndpointIP   string `json:"gatewayPrivateEndpointIp,omitempty"`
--- a/pkg/api/openshiftcluster.go
+++ b/pkg/api/openshiftcluster.go
@@ -264,7 +264,7 @@ type NetworkProfile struct {
 	ServiceCIDR            string                 `json:"serviceCidr,omitempty"`
 	SoftwareDefinedNetwork SoftwareDefinedNetwork `json:"softwareDefinedNetwork,omitempty"`
 	MTUSize                MTUSize                `json:"mtuSize,omitempty"`
-	OutboundType           OutboundType           `json:"userDefinedRouting,omitempty"`
+	OutboundType           OutboundType           `json:"OutboundType,omitempty"`
 
 	APIServerPrivateEndpointIP string `json:"privateEndpointIp,omitempty"`
 	GatewayPrivateEndpointIP   string `json:"gatewayPrivateEndpointIp,omitempty"`
--- a/pkg/api/openshiftcluster.go
+++ b/pkg/api/openshiftcluster.go
@@ -264,7 +264,7 @@ type NetworkProfile struct {
 	ServiceCIDR            string                 `json:"serviceCidr,omitempty"`
 	SoftwareDefinedNetwork SoftwareDefinedNetwork `json:"softwareDefinedNetwork,omitempty"`
 	MTUSize                MTUSize                `json:"mtuSize,omitempty"`
-	OutboundType           OutboundType           `json:"OutboundType,omitempty"`
+	OutboundType           OutboundType           `json:"outboundType,omitempty"`
 
 	APIServerPrivateEndpointIP string `json:"privateEndpointIp,omitempty"`
 	GatewayPrivateEndpointIP   string `json:"gatewayPrivateEndpointIp,omitempty"`
--- a/pkg/cluster/deploystorage.go
+++ b/pkg/cluster/deploystorage.go
@@ -130,15 +130,22 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
 		m.clusterNSG(infraID, azureRegion),
 		m.clusterServicePrincipalRBAC(),
 		m.networkPrivateLinkService(azureRegion),
-		m.networkPublicIPAddress(azureRegion, infraID+"-pip-v4"),
 		m.networkInternalLoadBalancer(azureRegion),
-		m.networkPublicLoadBalancer(azureRegion),
 	}
 
-	if m.doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPublic {
+	// Create a public load balancer routing if needed
+	if m.doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType == api.OutboundTypeLoadbalancer {
+		// Normal private clusters still need a public load balancer
 		resources = append(resources,
-			m.networkPublicIPAddress(azureRegion, infraID+"-default-v4"),
+			m.networkPublicIPAddress(azureRegion, infraID+"-pip-v4"),
+			m.networkPublicLoadBalancer(azureRegion),
 		)
+		// If the cluster is public we still want the default public IP address
+		if m.doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPublic {
+			resources = append(resources,
+				m.networkPublicIPAddress(azureRegion, infraID+"-default-v4"),
+			)
+		}
 	}
 
 	if m.doc.OpenShiftCluster.Properties.FeatureProfile.GatewayEnabled {
--- a/pkg/cluster/networkprofile.go
+++ b/pkg/cluster/networkprofile.go
@@ -60,3 +60,20 @@ func patchMTUSize(m *manager, ctx context.Context, mtuSize api.MTUSize) error {
 	})
 	return err
 }
+
+func (m *manager) determineOutboundType(ctx context.Context) error {
+	var err error
+	// Determine if this is a cluster with user defined routing
+	outboundType := api.OutboundTypeLoadbalancer
+	if m.doc.OpenShiftCluster.Properties.APIServerProfile.Visibility == api.VisibilityPrivate &&
+		m.doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPrivate &&
+		feature.IsRegisteredForFeature(m.subscriptionDoc.Subscription.Properties, api.FeatureFlagUserDefinedRouting) {
+		m.doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType = api.OutboundTypeUserDefinedRouting
+		outboundType = api.OutboundTypeUserDefinedRouting
+	}
+	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, func(doc *api.OpenShiftClusterDocument) error {
+		doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType = outboundType
+		return nil
+	})
+	return err
+}
--- a/pkg/cluster/networkprofile.go
+++ b/pkg/cluster/networkprofile.go
@@ -68,7 +68,6 @@ func (m *manager) determineOutboundType(ctx context.Context) error {
 	if m.doc.OpenShiftCluster.Properties.APIServerProfile.Visibility == api.VisibilityPrivate &&
 		m.doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPrivate &&
 		feature.IsRegisteredForFeature(m.subscriptionDoc.Subscription.Properties, api.FeatureFlagUserDefinedRouting) {
-		m.doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType = api.OutboundTypeUserDefinedRouting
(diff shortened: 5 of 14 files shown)


ArielLima:
This is very much a WIP and there is no need of reviews at this time.

ArielLima:
Cards created for this feature:
https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15054503
https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15058440

pkg/installer/custominstallconfig.go
```diff
@@ -54,6 +53,24 @@ func (m *manager) applyInstallConfigCustomisations(ctx context.Context, installC
...
+		}
+		return false
+	}
+	// Determine whether this cluster is a noPublicIP cluster
+	// At creation we can assume a single ingress profile TODO: Maybe `IngressProfiles` should be a map?
+	m.oc.Properties.NetworkProfile.UserDefinedRouting = api.UserDefinedRoutingDisabled
+	if m.oc.Properties.APIServerProfile.Visibility == api.VisibilityPrivate &&
+		m.oc.Properties.IngressProfiles[0].Visibility == api.VisibilityPrivate && _userDefinedRoutingEnabled(m) {
+		m.oc.Properties.NetworkProfile.UserDefinedRouting = api.UserDefinedRoutingEnabled
+	}
```
bennerv, line 72: This shouldn't go here.  This function is specifically for creating custom installation configuration, and we're not doing that with the UserDefinedRouting things.  We should be doing this elsewhere.

pkg/installer/deployresources_resources.go
```diff
@@ -13,25 +13,30 @@ import (
 	"github.com/openshift/installer/pkg/asset/ignition/machine"
 	"github.com/openshift/installer/pkg/asset/installconfig"
 
+	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/util/arm"
 	"github.com/Azure/ARO-RP/pkg/util/azureclient"
 )
 
 func (m *manager) networkBootstrapNIC(installConfig *installconfig.InstallConfig) *arm.Resource {
```
bennerv, line 21: Since we don't require a load balancer anymore, we can just gate on if we're using UserDefinedRouting, then don't add the ip configuration to point to the existing load balancer.

pkg/installer/generateconfig.go
```diff
@@ -110,6 +110,12 @@ func (m *manager) generateInstallConfig(ctx context.Context) (*installconfig.Ins
 		SoftwareDefinedNetwork = string(m.oc.Properties.NetworkProfile.SoftwareDefinedNetwork)
 	}
 
+	// determine outbound type based on cluster visibility
+	OutboundType := azuretypes.LoadbalancerOutboundType
```
bennerv, line 114: nit - don't use upper-case variable names.  
```suggestion
	outboundType := azuretypes.LoadbalancerOutboundType

File: pkg/installer/generateconfig.go [hunk 4]
```diff
@@ -221,7 +227,7 @@ func (m *manager) generateInstallConfig(ctx context.Context) (*installconfig.Ins
 		),
 	}
 
-	if m.oc.Properties.IngressProfiles[0].Visibility == api.VisibilityPrivate {
+	if m.oc.Properties.NetworkProfile.UserDefinedRouting == api.UserDefinedRoutingDisabled {
```
bennerv, line 230: Don't actually need to change this, as the case doesn't differ.

pkg/installer/generateconfig.go
```diff
@@ -181,7 +187,7 @@ func (m *manager) generateInstallConfig(ctx context.Context) (*installconfig.Ins
 					ControlPlaneSubnet:       masterSubnetName,
 					ComputeSubnet:            workerSubnetName,
 					CloudName:                azuretypes.CloudEnvironment(m.env.Environment().Name),
-					OutboundType:             azuretypes.LoadbalancerOutboundType,
+					OutboundType:             OutboundType,
```
bennerv, line 190: what happens if outboundType is empty?

pkg/installer/install.go
```diff
@@ -37,7 +37,7 @@ func (m *manager) Install(ctx context.Context) error {
 		steps.Action(func(ctx context.Context) error {
 			var err error
 			// Applies ARO-specific customisations to the InstallConfig
-			g, err = m.applyInstallConfigCustomisations(ctx, installConfig, image)
+			g, err = m.applyInstallConfigCustomisations(installConfig, image)
```
bennerv, line 40: +1 on removing `ctx` as it's not actually used.

bennerv (changes requested):
good work so far!  Few things I noticed

pkg/installer/custominstallconfig.go
bennerv: Make it it's own step within `pkg/cluster/install.go`.

Call it something like: `populateUserDefinedRouting` and utilize the existing `pkg/util/feature/subscription.go` function called `IsRegisteredForFeature` instead of writing your own.

ArielLima: +1

pkg/installer/deployresources_resources.go
ArielLima: I think we are already doing that

pkg/installer/generateconfig.go
ArielLima: +1

pkg/installer/generateconfig.go
ArielLima: Can't be empty, we initialize it here:
```go
// determine outbound type based on cluster visibility
outboundType := azuretypes.LoadbalancerOutboundType
if m.oc.Properties.NetworkProfile.UserDefinedRouting == api.UserDefinedRoutingDisabled {
	outboundType = azuretypes.UserDefinedRoutingOutboundType
}
```

pkg/api/admin/openshiftcluster_convert.go
```diff
@@ -193,6 +193,7 @@ func (c *openShiftClusterConverter) ToInternal(_oc interface{}, out *api.OpenShi
 	out.Properties.NetworkProfile.PodCIDR = oc.Properties.NetworkProfile.PodCIDR
 	out.Properties.NetworkProfile.ServiceCIDR = oc.Properties.NetworkProfile.ServiceCIDR
 	out.Properties.NetworkProfile.MTUSize = api.MTUSize(oc.Properties.NetworkProfile.MTUSize)
+	out.Properties.NetworkProfile.UserDefinedRouting = api.UserDefinedRouting(oc.Properties.NetworkProfile.UserDefinedRouting)
```
bennerv, line 196: Need to add a conversion in `toExternal` as well so we can view it from an admin API perspective.

pkg/installer/deployresources_resources.go
```diff
@@ -13,25 +13,30 @@ import (
...
+	// Private clusters without Public IPs will not have valid external load balancers
+	LBBackendAddressPool := &[]mgmtnetwork.BackendAddressPool{
```
bennerv, line 23: nit:
lowercase variable names `lbBackendAddressPool`

pkg/installer/install.go
```diff
@@ -37,12 +39,21 @@ func (m *manager) Install(ctx context.Context) error {
...
 		}),
 		steps.Action(func(ctx context.Context) error {
 			return m.persistGraph(ctx, g)
 		}),
+		steps.Action(func(ctx context.Context) error {
+			m.oc.Properties.NetworkProfile.UserDefinedRouting = api.UserDefinedRoutingDisabled
+			if m.oc.Properties.APIServerProfile.Visibility == api.VisibilityPrivate &&
+				m.oc.Properties.IngressProfiles[0].Visibility == api.VisibilityPrivate &&
+				feature.IsRegisteredForFeature(m.sub.Properties, api.FeatureFlagUserDefinedRouting) {
+				m.oc.Properties.NetworkProfile.UserDefinedRouting = api.UserDefinedRoutingEnabled
```
bennerv, line 53: can you pull this into it's own function to make things a bit cleaner?

bennerv (changes requested):
One more thing:

`pkg/api/defaults.go` will need a case if the field is empty to set the `UserDefinedRouting` to disabled.

pkg/installer/install.go
ArielLima: +1

pkg/installer/deployresources_resources.go
ArielLima: +1

pkg/api/admin/openshiftcluster_convert.go
ArielLima: +1

pkg/api/admin/openshiftcluster.go
```diff
@@ -130,14 +130,23 @@ const (
...
 	// The software defined network (SDN) to use when installing the cluster.
 	SoftwareDefinedNetwork SoftwareDefinedNetwork `json:"softwareDefinedNetwork,omitempty"`
 
-	PodCIDR     string  `json:"podCidr,omitempty"`
-	ServiceCIDR string  `json:"serviceCidr,omitempty"`
-	MTUSize     MTUSize `json:"mtuSize,omitempty"`
+	PodCIDR            string             `json:"podCidr,omitempty"`
+	ServiceCIDR        string             `json:"serviceCidr,omitempty"`
+	MTUSize            MTUSize            `json:"mtuSize,omitempty"`
+	UserDefinedRouting UserDefinedRouting `json:"userDefinedRouting,omitempty"`
```
bennerv, line 149: After some discussion with Jerome, we decided we want to do something like this instead.  This will allow us to support more outbound types in the future if desired.  I'm not sure what to call the "default".  I believe within OCP documentation it's `LoadbalancerOutboundType` so maybe that's a good thing to set it as instead of `Default`.

```
OutboundType OutboundType
```

```const (
OutboundTypeUserDefinedRouting OutboundType = "UserDefinedRouting"
OutboundTypeDefault OutboundType = "Default"
```

bennerv (commented):
Couple changes asked by Jerome on the experience.

pkg/api/admin/openshiftcluster.go
ArielLima: +1

pkg/api/admin/openshiftcluster.go
```diff
@@ -130,14 +130,23 @@ const (
 	MTU3900 MTUSize = 3900
 )
 
+// OutboundType represents
```
SudoBrendan, line 133: nit: a slightly more informative comment would be good

pkg/cluster/deploystorage.go
```diff
@@ -118,15 +118,22 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
...
 		m.clusterServicePrincipalRBAC(),
 		m.networkPrivateLinkService(azureRegion),
-		m.networkPublicIPAddress(azureRegion, infraID+"-pip-v4"),
 		m.networkInternalLoadBalancer(azureRegion),
-		m.networkPublicLoadBalancer(azureRegion),
 	}
 
-	if m.doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPublic {
+	// We do not want to create a public IP or LB for clusters with user defined outbound types
+	if m.doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType == api.OutboundTypeDefault {
```
SudoBrendan, line 125: nit: this line is making me think we should have a third constant:

1. User Defined
2. (whatever it is we want to call what we do right now... public?)
3. Default (which may change over time)

This `if` makes a lot more sense if we use a constant like "public" instead of "default"

SudoBrendan (changes requested):
This is looking good so far, nice work! A few nits.

pkg/cluster/deploystorage.go
ArielLima: +1 I don't think it makes sense to introduce a third constant at this time since we either want it user defined or default. I will rename the "default" to `OutboundTypeLoadbalancerRouting` since that will be more clear.

pkg/api/admin/openshiftcluster.go
ArielLima: +1

pkg/api/admin/openshiftcluster.go
```diff
@@ -130,14 +130,24 @@ const (
 	MTU3900 MTUSize = 3900
 )
 
+// OutboundType represents the type of routing a cluster is using
+type OutboundType string
+
+// OutboundType constants
+const (
+	OutboundTypeUserDefinedRouting  OutboundType = "UserDefinedRouting"
+	OutboundTypeLoadbalancerRouting OutboundType = "LoadbalancerRouting"
```
bennerv, line 139: I think we just want to call this `Loadbalancer` not `LoadbalancerRouting`.  

Maintains consistency with https://github.com/openshift/installer/blob/master/pkg/types/azure/platform.go#L18

pkg/util/network/network.go
```diff
@@ -0,0 +1,20 @@
...
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/util/feature"
+)
+
+func DetermineUserDefinedRouting(oc *api.OpenShiftCluster, sub *api.Subscription) error {
```
bennerv, line 11: nit:
```suggestion
func DetermineOutboundType(oc *api.OpenShiftCluster, sub *api.Subscription) error {

File: pkg/util/network/network.go [hunk 15]
```diff
@@ -0,0 +1,20 @@
...
+	// Determine if this is a cluster with user defined routing
+	oc.Properties.NetworkProfile.OutboundType = api.OutboundTypeLoadbalancerRouting
+	if oc.Properties.APIServerProfile.Visibility == api.VisibilityPrivate &&
+		oc.Properties.IngressProfiles[0].Visibility == api.VisibilityPrivate &&
+		feature.IsRegisteredForFeature(sub.Properties, api.FeatureFlagUserDefinedRouting) {
+		oc.Properties.NetworkProfile.OutboundType = api.OutboundTypeUserDefinedRouting
+	}
+	return nil
+}
```
bennerv, line 20: We should move this to the installation package.  This isn't really a utility.  As well, we can define it on the manager struct and call it using `m.determineOutboundType`

pkg/installer/install.go
```diff
@@ -37,12 +39,15 @@ func (m *manager) Install(ctx context.Context) error {
...
-			g, err = m.applyInstallConfigCustomisations(ctx, installConfig, image)
+			g, err = m.applyInstallConfigCustomisations(installConfig, image)
 			return err
 		}),
 		steps.Action(func(ctx context.Context) error {
 			return m.persistGraph(ctx, g)
 		}),
+		steps.Action(func(ctx context.Context) error {
+			return m.determineOutboundType(m.oc, m.sub)
+		}),
```
bennerv, line 50: This has to happen before `m.applyInstallConfigCustomisations` otherwise the field will be nil.

pkg/util/network/network.go
```diff
@@ -0,0 +1,20 @@
+package network
```
bennerv, line 1: I think this file isn't used anywhere now and can be deleted.

bennerv (changes requested):
Almost there - great work!

pkg/installer/install.go
```diff
@@ -33,11 +35,13 @@ func (m *manager) Install(ctx context.Context) error {
 			installConfig, image, err = m.generateInstallConfig(ctx)
 			return err
 		}),
-
+		steps.Action(func(ctx context.Context) error {
+			return m.determineOutboundType(m.oc, m.sub)
+		}),
```
bennerv, line 40: This has to happen before we generate the installation configuration since the `generateInstallConfig` function uses the outbound type to set the installConfig outboundType parameter.  Otherwise it'll nil out and fail? or default to something bad.

pkg/cluster/deploystorage.go
```diff
@@ -120,15 +120,22 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
 		m.clusterNSG(infraID, azureRegion),
 		m.clusterServicePrincipalRBAC(),
 		m.networkPrivateLinkService(azureRegion),
-		m.networkPublicIPAddress(azureRegion, infraID+"-pip-v4"),
 		m.networkInternalLoadBalancer(azureRegion),
-		m.networkPublicLoadBalancer(azureRegion),
 	}
 
-	if m.doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPublic {
+	// Create a public load balancer routing if needed
```
bennerv, line 126: So the ordering on this is interesting.  Can you append to your comment and add something like:
`m.determineOutboundType` will ensure that the outbound type is set correctly so we do not have to evaluate whether the visibility is public or private in this if statement since it's already been confirmed by the function.`

It's very order-specific and it's confusing to review/read in isolation.  We have to look at the installation as a whole to ensure this is doing what we expect.

bennerv (changes requested):
Super close.  Just noticed something on it that won't work.

cblecker:
Would you mind rebasing, @ArielLima? Removing ready-to-review until that's done.

ArielLima:
@cblecker As far as I am aware this is on hold until hive is finished so maybe wait until then to rebase?

pkg/api/admin/openshiftcluster.go
```diff
@@ -130,14 +130,24 @@ const (
...
 	// The software defined network (SDN) to use when installing the cluster.
 	SoftwareDefinedNetwork SoftwareDefinedNetwork `json:"softwareDefinedNetwork,omitempty"`
 
-	PodCIDR     string  `json:"podCidr,omitempty"`
-	ServiceCIDR string  `json:"serviceCidr,omitempty"`
-	MTUSize     MTUSize `json:"mtuSize,omitempty"`
+	PodCIDR      string       `json:"podCidr,omitempty"`
+	ServiceCIDR  string       `json:"serviceCidr,omitempty"`
+	MTUSize      MTUSize      `json:"mtuSize,omitempty"`
+	OutboundType OutboundType `json:"userDefinedRouting,omitempty"`
```
bennerv, line 150: can you make this mutable so we can modify it on PUCM for existing clusters.  Then we can set it for the clusters we performed this action on manually.  


In the following release (after this one is merged), we will remove the field.

bennerv (commented):
Can you open up a PR to https://github.com/hawkowl/ARO-Installer as well to pull in the installer related chagnes (pkg/installer)?  And then just add the mutable flag too then LGTM - ship it.

pkg/api/openshiftcluster.go
```diff
@@ -255,6 +264,7 @@ type NetworkProfile struct {
 	ServiceCIDR            string                 `json:"serviceCidr,omitempty"`
 	SoftwareDefinedNetwork SoftwareDefinedNetwork `json:"softwareDefinedNetwork,omitempty"`
 	MTUSize                MTUSize                `json:"mtuSize,omitempty"`
+	OutboundType           OutboundType           `json:"userDefinedRouting,omitempty"`
```
bennerv, line 267:
```suggestion
	OutboundType           OutboundType           `json:"outboundType,omitempty"`

File: pkg/cluster/networkprofile.go [hunk 22]
```diff
@@ -60,3 +60,20 @@ func patchMTUSize(m *manager, ctx context.Context, mtuSize api.MTUSize) error {
...
 }
+
+func (m *manager) determineOutboundType(ctx context.Context) error {
+	var err error
+	// Determine if this is a cluster with user defined routing
+	outboundType := api.OutboundTypeLoadbalancer
+	if m.doc.OpenShiftCluster.Properties.APIServerProfile.Visibility == api.VisibilityPrivate &&
+		m.doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPrivate &&
+		feature.IsRegisteredForFeature(m.subscriptionDoc.Subscription.Properties, api.FeatureFlagUserDefinedRouting) {
+		m.doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType = api.OutboundTypeUserDefinedRouting
```
bennerv, line 71: this line is redundant since you're already doing this on line 75 below.

bennerv (changes requested):
@ArielLima  can we get these changes and the other comments fixed so we can merge this?

bennerv:
Also, please update the PR description and link to appropriate work items.

pkg/api/openshiftcluster.go
```diff
@@ -255,6 +264,7 @@ type NetworkProfile struct {
 	ServiceCIDR            string                 `json:"serviceCidr,omitempty"`
 	SoftwareDefinedNetwork SoftwareDefinedNetwork `json:"softwareDefinedNetwork,omitempty"`
 	MTUSize                MTUSize                `json:"mtuSize,omitempty"`
+	OutboundType           OutboundType           `json:"OutboundType,omitempty"`
```
bennerv, line 267:
```suggestion
	OutboundType           OutboundType           `json:"outboundType,omitempty"`

**Review by bennerv — COMMENTED** [review]
Just one capitalization issue and I'm good to approve.  Great work here.

**Review by bennerv — APPROVED** [review]
lgtm

File: pkg/cluster/deploystorage.go [hunk 24]
```diff
@@ -130,15 +130,22 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
...
+	if m.doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType == api.OutboundTypeLoadbalancer {
+		// Normal private clusters still need a public load balancer
 		resources = append(resources,
-			m.networkPublicIPAddress(azureRegion, infraID+"-default-v4"),
+			m.networkPublicIPAddress(azureRegion, infraID+"-pip-v4"),
```
s-amann, line 140: why do we name this one with the `-pip-v4` suffix? does it have specific meaning?

s-amann (commented):
just some questions - forgot to submit them before it got merged..

pkg/installer/custominstallconfig.go
```diff
@@ -21,7 +20,7 @@ import (
 // applyInstallConfigCustomisations modifies the InstallConfig and creates
 // parent assets, then regenerates the InstallConfig for use for Ignition
 // generation, etc.
-func (m *manager) applyInstallConfigCustomisations(ctx context.Context, installConfig *installconfig.InstallConfig, image *releaseimage.Image) (graph.Graph, error) {
+func (m *manager) applyInstallConfigCustomisations(installConfig *installconfig.InstallConfig, image *releaseimage.Image) (graph.Graph, error) {
```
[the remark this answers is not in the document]
bennerv: Context isn't used anywhere within the function.  Just asked to remove it.  Could've been a separate commit, but left it in there.

pkg/cluster/deploystorage.go
```diff
@@ -130,15 +130,22 @@ func (m *manager) deployStorageTemplate(ctx context.Context) error {
...
+			m.networkPublicLoadBalancer(azureRegion),
 		)
+		// If the cluster is public we still want the default public IP address
+		if m.doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPublic {
+			resources = append(resources,
+				m.networkPublicIPAddress(azureRegion, infraID+"-default-v4"),
```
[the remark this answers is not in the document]
bennerv: https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/375424/Remove-public-IP-manually#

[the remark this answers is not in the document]
bennerv: >I believe a public cluster would end up with both -pip-v4 and -default-v4.

This is the desired behavior. Existing public ingress & apiserver clusters have both these IP addresses.  We don't want to change the default behavior we've been using for existing clusters.  The default IP address is for ingress, the other is used for outbound explicit rule.

pkg/cluster/deploystorage.go
bennerv: It's how upstream OCP installed named them.  It doesn't have a specific meaning.

PR #2252, opened 2022-07-15
Adds and documents a way to manually test a custom hive image

m1kola:

It is not the best way to test the changes, but for now it is jsut to close a gap.

### Test plan for issue:

Manual test of the docs

### DIFF
--- a/docs/hive.md
+++ b/docs/hive.md
@@ -35,3 +35,36 @@ Installing then simply requires the running of the install script.
 ```bash
 ./hack/hive-dev-install.sh
 ```
+
+
+## Testing own hive build
+
+1. Create a fork of `github.com/openshift/hive` on github
+    1. Make your changes and push into your fork.
+1. Build and push a custom Hive image:
+    ```bash
+    # From a hive repo checkout with your changes
+    export IMG="quay.io/{username}/hive:latest"
+    GOOS=linux GOARCH=amd64 make build image-hive-fedora-dev docker-push
+    ```
+1. Make sure that your image is public so AKS can pull it
+1. Set environment variables like shown in the example below:
+    ```bash
+    export HIVE_REPO=https://github.com/m1kola/hive.git # Point to your fork
+    export HIVE_IMAGE_COMMIT_HASH=c63c9b0               # Commit hash from your fork
+    export HIVE_IMAGE=$IMG                              # Point to your image as previously set in $IMG
+    ```
+1. Run the following commands:
+    ```bash
+    source ./env
+
+    make aks.kubeconfig
+    export KUBECONFIG="$PWD/aks.kubeconfig"
+
+    ./hack/hive-generate-config.sh
+    ./hack/hive-dev-install.sh
+    ```
+1. Restart pods to make sure that hive is running the latest version:
+    ```bash
+    oc delete pods -n hive --all
+    ```
--- a/hack/hive-generate-config.sh
+++ b/hack/hive-generate-config.sh
@@ -1,11 +1,13 @@
 #!/bin/bash
 
+HIVE_REPO="${HIVE_REPO:=https://github.com/openshift/hive.git}"
+
 # This is the commit sha that the image was built from and ensures we use the correct configs for the release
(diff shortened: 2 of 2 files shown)


docs/hive.md
```diff
@@ -35,3 +35,34 @@ Installing then simply requires the running of the install script.
...
+1. Create a fork of `github.com/openshift/hive` on github
+    1. Make your changes and push into your fork.
+1. Build and push a custom Hive image:
+    ```bash
+    # From a hive repo checkout with your changes
+    export IMG="quay.io/{username}/hive:latest"
+    GOOS=linux GOARCH=amd64 make build image-hive-fedora-dev docker-push
+    ```
+1. Make sure that your image is public so AKS can pull it
+1. Update `./hack/hive-generate-config.sh`:
```
dem4gus, line 51: Since this file is committed to source control it may make more sense to pull these values from the local environment rather than changing the script itself, which may lead to accidental updates. See comment on script.

hack/hive-generate-config.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+HIVE_REPO="https://github.com/openshift/hive.git"
```
dem4gus, line 3: This would be a good use of Bash's parameter expansion so these variables can be set on the local execution environment, with the current values as a default if they are not set:
```bash
HIVE_REPO="${HIVE_REPO:=https://github.com/openshift/hive.git}"
```

dem4gus (changes requested):

m1kola:
@dem4gus updated. Please take another look.

hack/hive-generate-config.sh
```diff
@@ -1,11 +1,13 @@
...
+HIVE_REPO="${HIVE_REPO:=https://github.com/openshift/hive.git}"
+
 # This is the commit sha that the image was built from and ensures we use the correct configs for the release
-HIVE_IMAGE_COMMIT_HASH="c63c9b0"
+HIVE_IMAGE_COMMIT_HASH="${HIVE_IMAGE_COMMIT_HASH:=c63c9b0}"
 
 # For now we'll use the quay hive image, but this will change to an ACR once the quay.io -> ACR mirroring is setup
 # Note: semi-scientific way to get the latest image: `podman search --list-tags --limit 10000 quay.io/app-sre/hive | tail -n1`
-HIVE_IMAGE="quay.io/app-sre/hive:${HIVE_IMAGE_COMMIT_HASH}"
+HIVE_IMAGE="${HIVE_IMAGE:=quay.io/app-sre/hive:${HIVE_IMAGE_COMMIT_HASH}}"
```
dem4gus, line 10: The way this variable and `$HIVE_IMAGE_COMMIT_HASH` are interacting is a tad confusing. If this gets set to `$IMG` like in the instructions then `$HIVE_IMAGE_COMMIT_HASH` isn't used at all for the image tag since `$IMG` has `latest`. If this is unset but `$HIVE_IMAGE_COMMIT_HASH` is, then the hash being used is likely one from a local git branch and the tag won't exist in the `quay.io/app-sre/hive` container repository. Other functions in this file use `$HIVE_IMAGE_COMMIT_HASH` to generate configs, and having the image tag default to `latest` could cause some problems when the `hive:latest` and `hive:${HIVE_IMAGE_COMMIT_HASH}` are not the same image.

m1kola: Yes, it is confusing. Any suggestions?


Keep in mind - this is for rare cases when you need to test a custom image. In theory `HIVE_IMAGE_COMMIT_HASH` only needs to be updated when CRDs/manifests change. In most of the cases it will be ok to use `HIVE_IMAGE_COMMIT_HASH` for config generation and using `HIVE_IMAGE` from a different commit.

dem4gus: > In theory HIVE_IMAGE_COMMIT_HASH only needs to be updated when CRDs/manifests change. In most of the cases it will be ok to use HIVE_IMAGE_COMMIT_HASH for config generation and using HIVE_IMAGE from a different commit.

I think leaving this as a note in the docs and/or as a comment in the script would be the best approach then.

PR #2251, opened 2022-07-14
Adding openvpn for civm's to support Simply Secure

swiencki:
### Which issue this PR addresses:

FEATURE [12563373](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/12563373)

Simply Secure is introducing zero trust network policies within nonprod environments that E2E / CI pipelines deploy to.

### What this PR does / why we need it:

Installing openvpn on civm's for VPN and Simply Secure support from civm's to nonprod environments.

### Test plan for issue:

VPN and Simply Secure policies will be tested using E2E. Additional details in FEATURE [12563373](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/12563373).

### Is there any documentation that needs to be updated for this PR?

Not at this point in time.

PR #2258, opened 2022-07-18
Adding e2e openvpn support for Simply Secure

swiencki:

FEATURE [12563373](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/12563373)

Simply Secure is introducing zero trust network policies within nonprod environments that E2E / CI pipelines deploy to.

Adding e2e openvpn support for Simply Secure.

### Test plan for issue:

VPN and Simply Secure policies will be tested using E2E. Additional details in FEATURE [12563373](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/12563373).

Not at this point in time.


.pipelines/e2e.yml
```diff
@@ -41,16 +41,29 @@ jobs:
 
       az account set -s $AZURE_SUBSCRIPTION_ID
 
+      sudo openvpn --config secrets/$VPN --daemon --writepid vpnpid
```
dem4gus, line 44: Is `openvpn` guaranteed to be on the CI agents?

Also, since `kill_vpn` is being put into the hack/e2e script it may also make sense to include this VPN creation in there as well and move the script sourcing to earlier in the pipeline process.

pkg/cluster/nsg.go
```diff
@@ -31,7 +31,7 @@ func (m *manager) clusterNSG(infraID, location string) *arm.Resource {
 					SourceAddressPrefix:      to.StringPtr("*"),
 					DestinationAddressPrefix: to.StringPtr("*"),
 					Access:                   mgmtnetwork.SecurityRuleAccessAllow,
-					Priority:                 to.Int32Ptr(101),
+					Priority:                 to.Int32Ptr(120),
```
dem4gus, line 34: I don't know if there's a reason for this or if the change is just an oversight, pointing it out in case it's the latter.

.pipelines/e2e.yml
swiencki: openvpn is guaranteed on CI agents as per https://github.com/Azure/ARO-RP/pull/2251
Added 'run_vpn' for ./hack/e2e/run-rp-and-e2e.sh 👍

pkg/cluster/nsg.go
swiencki: Simply Secure NSG policies apply with a priority within the 101 range. 120 keeps it out of the range an ensures no overlap and conflict

pkg/util/cluster/cluster.go
```diff
@@ -134,15 +134,29 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 		return fmt.Errorf("fp service principal id is not found")
 	}
 
-	c.log.Infof("creating AAD application")
-	appID, appSecret, err := c.createApplication(ctx, "aro-"+clusterName)
-	if err != nil {
-		return err
+	appID := ""
```
karanmagdani, line 137: The subscription I'm testing this on does not have Microsoft.Graph permissions so instead of creating a new AAD its just reusing RPs. This won't be part of the final PR for merge

swiencki:
PR not required, please see https://github.com/Azure/ARO-RP/pull/2300

PR #2260, opened 2022-07-19
Update a comment to be a todo

m1kola:
Just a comment update so it is easier to grep all these todos.


m1kola:
Closing in favour of #2261

PR #2264, opened 2022-07-19
slight refactor to use of FriendlyName function

ArielLima:

slight refactor to use of FriendlyName function

### DIFF
--- a/pkg/monitor/cluster/cluster.go
+++ b/pkg/monitor/cluster/cluster.go
@@ -118,8 +118,9 @@ func (mon *Monitor) Monitor(ctx context.Context) (errs []error) {
 	statusCode, err := mon.emitAPIServerHealthzCode(ctx)
 	if err != nil {
 		errs = append(errs, err)
-		mon.log.Printf("%s: %s", steps.FriendlyName(mon.emitAPIServerHealthzCode), err)
-		mon.emitGauge("monitor.clustererrors", 1, map[string]string{"monitor": steps.FriendlyName(mon.emitAPIServerHealthzCode)})
+		friendlyFuncName := steps.FriendlyName(mon.emitAPIServerHealthzCode)
+		mon.log.Printf("%s: %s", friendlyFuncName, err)
+		mon.emitGauge("monitor.clustererrors", 1, map[string]string{"monitor": friendlyFuncName})
 	}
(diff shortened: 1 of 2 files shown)


pkg/monitor/cluster/cluster.go
```diff
@@ -120,8 +118,8 @@ func (mon *Monitor) Monitor(ctx context.Context) (errs []error) {
 	statusCode, err := mon.emitAPIServerHealthzCode(ctx)
 	if err != nil {
 		errs = append(errs, err)
-		mon.log.Printf("%s: %s", runtime.FuncForPC(reflect.ValueOf(mon.emitAPIServerHealthzCode).Pointer()).Name(), err)
-		mon.emitGauge("monitor.clustererrors", 1, map[string]string{"monitor": runtime.FuncForPC(reflect.ValueOf(mon.emitAPIServerHealthzCode).Pointer()).Name()})
+		mon.log.Printf("%s: %s", steps.FriendlyName(mon.emitAPIServerHealthzCode), err)
+		mon.emitGauge("monitor.clustererrors", 1, map[string]string{"monitor": steps.FriendlyName(mon.emitAPIServerHealthzCode)})
```
bennerv, line 122: nit:
generate the value once, use it in two different places.  
```suggestion
                friendlyName := steps.FriendlyName(mon.emitAPIServerHealthzCode)
		mon.log.Printf("%s: %s", friendlyName, err)
		mon.emitGauge("monitor.clustererrors", 1, map[string]string{"monitor": friendlyName})

bennerv (dismissed):
lgtm - only one nit

@AldoFusterTurpin curious on your thoughts here as well.

pkg/monitor/cluster/cluster.go
AldoFusterTurpin: Agree 100%. It is the same that I did in this [PR](https://github.com/Azure/ARO-RP/pull/2211/files) that you can see [here](https://github.com/Azure/ARO-RP/blob/master/pkg/monitor/cluster/cluster.go#L152) in this same file.

AldoFusterTurpin (dismissed):
LGTM. Thanks!

bennerv (approved):
lgtm

PR #2265, opened 2022-07-19
Set qualys download link config param

bennerv:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15004784

* Sets the base URL for qualys plugin as it's incorrectly trying to fetch the wrong URL due to a bug in azure-security package.  
* Reverts the previous changes to pinning azure-security to a specific version
* Removes unnecessary parameters we're setting.  They're being unmarshalled into the config struct so we shouldn't have to set them.  

### Test plan for issue:

Full service deploy + deploy to INT.


ross-bryan (approved):
LGTM - will approve as is, but one nit on the name of the var - should we consider a prefix with `azSecPack`? like the TenantID?
```
azureSecPackVSATenantId
```
that way we are grouping the vars together for same process and name provides the context they are related

ross-bryan (dismissed):
lgtm thx

ross-bryan (approved):
lgtm thx

bennerv:
CodeQL is failing after updating it to v2 in #2269 .  Appears to be scanning so many directories that is causing failures in PRs.  There's a new PR #2276 which should fix it, then we can rerun it and merge this in.

PR #2266, opened 2022-07-19
generate and insert public key on deploy

jhoreman:

Since we now manage SSH access through JIT directly, as opposed to using JIT to collect a permanent key, we can now remove the aforementioned private key. In order to pass ARM validation an authentication mechanism is required however so we're choosing to generate this on demand on deploy. 
Fixes

Accompanies the SSH Access JIT changes

### Test plan for issue:

Once deployed we can check the key value in authorised keys 

No docs already exist.

### DIFF
--- a/pkg/deploy/config.go
+++ b/pkg/deploy/config.go
@@ -4,6 +4,9 @@ package deploy
 // Licensed under the Apache License 2.0.
 
 import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
 	"fmt"
 	"io/ioutil"
 	"reflect"
@@ -91,7 +94,7 @@ type Configuration struct {
 	SubscriptionResourceGroupName      *string       `json:"subscriptionResourceGroupName,omitempty" value:"required"`
 	SubscriptionResourceGroupLocation  *string       `json:"subscriptionResourceGroupLocation,omitempty" value:"required"`
 	RPVMSSCapacity                     *int          `json:"rpVmssCapacity,omitempty"`
-	SSHPublicKey                       *string       `json:"sshPublicKey,omitempty" value:"required"`
+	SSHPublicKey                       *string       `json:"sshPublicKey,omitempty"`
 	StorageAccountDomain               *string       `json:"storageAccountDomain,omitempty" value:"required"`
 	VMSize                             *string       `json:"vmSize,omitempty" value:"required"`
 	VMSSCleanupEnabled                 *bool         `json:"vmssCleanupEnabled,omitempty"`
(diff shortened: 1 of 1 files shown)


pkg/deploy/config.go
```diff
@@ -110,6 +113,13 @@ func GetConfig(path, location string) (*RPConfig, error) {
 		return nil, err
 	}
 
+	if config.Configuration.SSHPublicKey == nil {
+		key, _ := rsa.GenerateKey(rand.Reader, 2048)
+		pubKey := x509.MarshalPKCS1PublicKey(&key.PublicKey)
+		publicKey := string(pubKey)
+		config.Configuration.SSHPublicKey = (&publicKey)
+	}
+
```
bennerv, line 122: one more ask - can you put this in the `validate()` function?  otherwise lgtm

bennerv (commented):
one nit

bennerv (dismissed):
LGTM - just need to verify you can SSH to FF Prod with an escort session and get onto the VMSS without using the SSH key.  Once that is done, I'm good to merge.

pkg/deploy/config.go
```diff
@@ -147,6 +150,13 @@ func (conf *RPConfig) validate() error {
 	v := reflect.ValueOf(*configuration)
 	missingFields := []string{}
 
+	if conf.Configuration.SSHPublicKey == nil {
+		key, _ := rsa.GenerateKey(rand.Reader, 2048)
```
cadenmarchese, line 154: Should we handle the error here?

pkg/deploy/config.go
```diff
@@ -147,6 +150,13 @@ func (conf *RPConfig) validate() error {
...
+		pubKey := x509.MarshalPKCS1PublicKey(&key.PublicKey)
+		publicKey := string(pubKey)
+		conf.Configuration.SSHPublicKey = (&publicKey)
+	}
```
dem4gus, line 158: If I'm understanding the description correctly we no longer need a permanently-stored key for SSH access to the servers, which means we don't care that the generated private key is lost after this block?

pkg/deploy/config.go
facchettos: absolutely. if `err!=nil` you have a nil pointer and will probably panic on the next line.

pkg/deploy/config.go
jhoreman: Correct! We actually want to forget about it and have garbage collection kick in.

pkg/deploy/config.go
jhoreman: Thanks, both. Added error handling. Please review.

bennerv (approved):
lgtm - waiting on confirmation vmss JIT works as expected with an escort session then can merge

ross-bryan (approved):
lgtm

cadenmarchese (approved):
Thanks!

PR #2267, opened 2022-07-19
parseUnverified returns cloudError

carvalhe:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15008430

This PR is intended to have errors returned from parseUnverified in aad specifically to return the same cloudError as serviceProviderCredentials. As of now, we see logs where it keeps retrying to get token and the error "Invalid client secret provided." is displayed for 10 minuets only for it to go through and have invalid segments on a token. This change will give a better idea of the issue and split it from an internalServerError to a userError.

### DIFF
--- a/pkg/util/aad/aad.go
+++ b/pkg/util/aad/aad.go
@@ -43,11 +43,12 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
 	// NOTE: Do not override err with the error returned by
 	// wait.PollImmediateUntil. Doing this will not propagate the latest error
 	// to the user in case when wait exceeds the timeout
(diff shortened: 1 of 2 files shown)


SudoBrendan (dismissed):
lgtm

pkg/util/aad/aad.go
```diff
@@ -57,6 +58,7 @@ func GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, te
 		claims := jwt.MapClaims{}
 		_, _, err = p.ParseUnverified(authorizer.OAuthToken(), claims)
 		if err != nil {
+			err = errServicePrincipal
```
weinong, line 61: i am not sure if this level of simplification is needed. I personally prefer to return a new cloud error here with more detail from `err`

ross-bryan (approved):
lgtm

s-amann (approved):

weinong:
LGTM

PR #2268, opened 2022-07-20
Makes ClusterDeployment private endpoint aware

m1kola:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15014584

Updates the way we create/update `ClusterDeployment`s: we need to provide a value for a new field which was introduced in https://github.com/openshift/hive/pull/1817 to make hive private endpoint aware.

### Test plan for issue:

This code path is disabled by default. E2E tests should indicate a regression, if there is any.

No, it is an internal implementation detail.


ross-bryan (approved):
lgtm

s-amann (approved):

bennerv (approved):
+1 - good stuff

PR #2269, opened 2022-07-20
bumped version of codeql to v2

facchettos:
updates codeql which now fails because it was using v1 which is being deprecated
All pipelines are going to start failing without that update

Fixes


petrkotas (approved):

ross-bryan (approved):
lgtm

bennerv:
This doesn't impact e2e at all.  Merging.

PR #2272, opened 2022-07-21
Doc: Add step to Stop/Start/resize for a dev cluster VM

SrinivasAtmakuri:

updates the docs to add CURL calls to start, Stop and
resize VM of a dev cluster using local-rp


ross-bryan (approved):
lgtm

PR #2273, opened 2022-07-21
Verify discovery cache changes in version 4.11.0-rc.4

rhamitarora:
A list of files changed resources between 4.10 and 4.11 when run **make discoverycache**
[Work Item - Verify if DiscoveryCache changes in z-streams will have an effect](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14625899)


ross-bryan (changes requested):
hey what is this for? we are not on 4.11 and wouldn't add in a release candidate as our version we deploy

hawkowl:
@ross-bryan I asked for a draft PR so that I could see the changes -- if we're going to the RP being multiversion, the discoverycache stuff might be a problem, so this is part of a workitem to see what changes.

ross-bryan:
ack, thx

hawkowl:
Since thi

PR #2274, opened 2022-07-21
docs: update aad group name

ross-bryan:

Fixes the group name in our docs

The old group name was under the old tenant, we are now under a new tenant in our dev subscription

### Test plan for issue:

n/a
this is the docs


m1kola (approved):

m1kola:
Merged without wawiting for e2e as it is just a single line doc change.

PR #2275, opened 2022-07-21
DRY implementation to our admin_vm_actions

ArielLima:
Remove a  lot of repeated code in our VM actions.
### Test plan for issue:
Unit tests were already in place.

### DIFF
--- a/pkg/frontend/admin_openshiftcluster_redeployvm.go
+++ b/pkg/frontend/admin_openshiftcluster_redeployvm.go
@@ -12,8 +12,6 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/sirupsen/logrus"
 
-	"github.com/Azure/ARO-RP/pkg/api"
-	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
 	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
 )
 
@@ -21,40 +19,16 @@ func (f *frontend) postAdminOpenShiftClusterRedeployVM(w http.ResponseWriter, r 
 	ctx := r.Context()
 	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
 	r.URL.Path = filepath.Dir(r.URL.Path)
-
-	err := f._postAdminOpenShiftClusterRedeployVM(ctx, r, log)
-
+	err := f._postAdminOpenShiftClusterRedeployVM(log, ctx, r)
 	adminReply(log, w, nil, nil, err)
 }
 
-func (f *frontend) _postAdminOpenShiftClusterRedeployVM(ctx context.Context, r *http.Request, log *logrus.Entry) error {
+func (f *frontend) _postAdminOpenShiftClusterRedeployVM(log *logrus.Entry, ctx context.Context, r *http.Request) error {
 	vars := mux.Vars(r)
-
 	vmName := r.URL.Query().Get("vmName")
-	err := validateAdminVMName(vmName)
+	azActionsWrapper, err := f.newAzureActionsWrapper(log, ctx, vmName, strings.TrimPrefix(r.URL.Path, "/admin"), vars)
 	if err != nil {
 		return err
 	}
-
-	resourceID := strings.TrimPrefix(r.URL.Path, "/admin")
-
-	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
-	switch {
-	case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
-		return api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, "", "The Resource '%s/%s' under resource group '%s' was not found.", vars["resourceType"], vars["resourceName"], vars["resourceGroupName"])
-	case err != nil:
-		return err
-	}
-
-	subscriptionDoc, err := f.getSubscriptionDocument(ctx, doc.Key)
-	if err != nil {
-		return err
-	}
-
-	a, err := f.azureActionsFactory(log, f.env, doc.OpenShiftCluster, subscriptionDoc)
-	if err != nil {
-		return err
-	}
-
-	return a.VMRedeployAndWait(ctx, vmName)
+	return f.adminAction.VMRedeployAndWait(ctx, azActionsWrapper.vmName)
 }
--- a/pkg/frontend/admin_vm_actions.go
+++ b/pkg/frontend/admin_vm_actions.go
@@ -1,0 +1,53 @@
+package frontend
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
+)
+
+type azVmActionsWrapper struct {
+	vmName string
+	doc    *api.OpenShiftClusterDocument
+}
+
+func (f *frontend) newAzureActionsWrapper(log *logrus.Entry, ctx context.Context, vmName, resourceID string, vars map[string]string) (azVmActionsWrapper, error) {
+	err := validateAdminVMName(vmName)
+	if err != nil {
+		return azVmActionsWrapper{}, err
+	}
+	if err != nil {
+		return azVmActionsWrapper{}, err
+	}
+
+	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	switch {
+	case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
+		return azVmActionsWrapper{}, api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, "",
+			"The Resource '%s/%s' under resource group '%s' was not found.",
+			vars["resourceType"], vars["resourceName"], vars["resourceGroupName"])
+	case err != nil:
+		return azVmActionsWrapper{}, err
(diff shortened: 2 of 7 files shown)


pkg/frontend/admin_vm_actions.go
```diff
@@ -0,0 +1,80 @@
...
+	r.URL.Path = filepath.Dir(r.URL.Path)
+
+	return VMContext{
+		context: ctx,
+		logger:  log,
+		r:       r,
+	}
+}
+
+func (f *frontend) BuildAzureActionsFactory(ctx context.Context, r *http.Request, log *logrus.Entry) (AzActionsFactoryWrap, error) {
```
bennerv, line 46: Doesn't need to be exported.  
```suggestion 
func (f *frontend) buildAzureActionsFactory(ctx context.Context, r *http.Request, log *logrus.Entry) (AzActionsFactoryWrap, error) {
```

pkg/frontend/admin_openshiftcluster_redeployvm.go
```diff
@@ -6,55 +6,20 @@ package frontend
...
-
-	"github.com/Azure/ARO-RP/pkg/api"
-	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
-	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
 )
 
 func (f *frontend) postAdminOpenShiftClusterRedeployVM(w http.ResponseWriter, r *http.Request) {
-	ctx := r.Context()
-	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
-	r.URL.Path = filepath.Dir(r.URL.Path)
```
bennerv, line 23: I don't think we need to wrap these in a struct.  They're going to be used in every admin action and the context isn't limited towards just VMs.  

I'm fine with reverting these three lines and living with them in everything for now.

pkg/frontend/admin_vm_actions.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+)
+
+type VMContext struct {
+	context context.Context
+	r       *http.Request
+	logger  *logrus.Entry
+}
+
+type AzActionsFactoryWrap struct {
```
bennerv, line 27: You don't need to export this as it's only used within the package. 

Additionally this contains VM specific parameters and not all our azure actions contain are against VMs.  We have other things that run against NICs instead of strictly Vms.  I think renaming it to something like `azVmActionsWrapper` or `azVmActionsFactoryWrap` would be better.  

Or, wrap all fields that would be needed and replace it in multiple places.   (not just the azure VM actions)

bennerv (changes requested):
Good first pass - we're definitely removing some redundancy here.  

Few nits and requested changes.  @m1kola if you have time would love to see your review on this as you're a lot better at refactoring and clean code than I am :)

ArielLima:
> Good first pass - we're definitely removing some redundancy here.
> 
> Few nits and requested changes. @m1kola if you have time would love to see your review on this as you're a lot better at refactoring and clean code than I am :)

Done

s-amann (dismissed):
lgtm, great work reducing duplicated code!

pkg/frontend/admin_vm_actions.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
+	"github.com/Azure/ARO-RP/pkg/frontend/adminactions"
+)
+
+type azVmActionsWrapper struct {
+	vmName       string
+	adminActions adminactions.AzureActions
```
dem4gus, line 21: Would it be beneficial to define an `AzureVmActions` interface so we could ensure that only VM Actions are being included in this wrapper?

```go
type AzureVmActions interface {
    VMRedeployAndWait(ctx context.Context, vmName string) error
	VMStartAndWait(ctx context.Context, vmName string) error
	VMStopAndWait(ctx context.Context, vmName string) error
	VMSizeList(ctx context.Context) ([]mgmtcompute.ResourceSku, error)
	VMResize(ctx context.Context, vmName string, vmSize string) error
	VMSerialConsole(ctx context.Context, w http.ResponseWriter, log *logrus.Entry, vmName string) error
}
```

We could also use that to compose the `AzureActions` interface

```go
type AzureActions interface {
    ResourcesList(ctx context.Context) ([]byte, error)
	NICReconcileFailedState(ctx context.Context, nicName string) error
    AzureVmActions
}
```

ArielLima: Seems a unnecessary to me but I am happy to get more input. @bennerv @m1kola

ArielLima:
@facchettos Ben would like your input on this. What are your thoughts?

pkg/frontend/admin_vm_actions.go
```diff
@@ -0,0 +1,60 @@
...
+func (f *frontend) buildAzureActionsFactory(ctx context.Context, r *http.Request, log *logrus.Entry) (azVmActionsWrapper, error) {
+	vars := mux.Vars(r)
+
+	vmName := r.URL.Query().Get("vmName")
+	err := validateAdminVMName(vmName)
+	if err != nil {
+		return azVmActionsWrapper{}, err
+	}
+
+	resourceID := strings.TrimPrefix(r.URL.Path, "/admin")
```
facchettos, line 35: same than the others. much easier to forge strings than http.requests

pkg/frontend/admin_vm_actions.go
```diff
@@ -0,0 +1,60 @@
...
+
+	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	switch {
+	case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
+		return azVmActionsWrapper{}, api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, "", "The Resource '%s/%s' under resource group '%s' was not found.", vars["resourceType"], vars["resourceName"], vars["resourceGroupName"])
```
facchettos, line 40: for another PR (but it's also DRY :D), could this newclouderror be made into something like `api.NewCloudErrorResourceNotFound(map[string]string)` or something ? it appears  16 times over the codebase with this exact wording(well maybe 12 after that PR). Maybe I missed some also

facchettos (commented):
Overall LGTM, I just don't like the name you gave to the function because it isn't really descriptive imo. 

I left a few comments which could make testing easier if you decide to test it ;) 
If testing, I would maybe bundle up getting doc and subscriptionDoc into a single function which returns `doc, subDoc, err`, since they are directly related. That could then maybe be mocked in the testing on the main function. 

Also found another potential DRY improvement , but this is for another PR

pkg/frontend/admin_vm_actions.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/Azure/ARO-RP/pkg/frontend/adminactions"
+)
+
+type azVmActionsWrapper struct {
+	vmName       string
+	adminActions adminactions.AzureActions
+	doc          *api.OpenShiftClusterDocument
+}
+
+func (f *frontend) newAzureActionsWrapper(ctx context.Context, vmName string, resourceID string, vars map[string]string, log *logrus.Entry) (azVmActionsWrapper, error) {
```
facchettos, line 23: I just realized that the logrus thing should probably be inside the frontend struct for this function and probably `f.azureActionsFactory` ( maybe others too), so we don't have to pass it around using parameters. Does it belong in this PR or another, I'm not sure, I'll let you be the judge of that.

A little nit while I'm at it, you can write `vmName , resourceID string` instead of writing string twice.

pkg/frontend/admin_vm_actions.go
ArielLima: I started working on a fix for this separately since it is a significant refactor itself.

facchettos: Awesome thanks !

pkg/frontend/admin_vm_actions.go
ArielLima: +1 I think more can be done on the log side but it should be in a separate  PR

facchettos (changes requested):
Actually I think I might have I underestimated the changes required to improve the logging here. that might be easier to revert the commit touching the logs and do it in another PR if we do it. 

With the current thing that you have done you would have some unexpected race conditions.

pkg/frontend/admin_openshiftcluster_redeployvm.go
```diff
@@ -12,49 +12,23 @@ import (
...
 
-	"github.com/Azure/ARO-RP/pkg/api"
-	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
 	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
 )
 
 func (f *frontend) postAdminOpenShiftClusterRedeployVM(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
-	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
+	f.contextLog = ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
```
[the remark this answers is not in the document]
ArielLima: +1

bennerv (dismissed):
lgtm - good work

bennerv:
@ArielLima can you rebase.  There was a python PR which fixed the failing python CI test.

facchettos (approved):
Lgtm thanks ! Was waiting for the ci to be fixed

s-amann (approved):

facchettos:
this introduced a race condition I think... ):

PR #2278, opened 2022-07-22
codeql version bump and fix extraction

facchettos:
updated to v2 and changed the concurrency on the extraction. This slows down the process a little but it avoid the errors. 

This was recommended by someone from codeQL [here](https://github.com/github/codeql/issues/9888) 

update codeql to v2 because it is being deprecated

### Test plan for issue:
https://github.com/Azure/ARO-RP/actions/runs/2733706014 

Ran it multiple times to see if it was working.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -39,7 +39,7 @@ jobs:
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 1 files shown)


facchettos:
This seems to be a lot more stable now with the new env variable.

petrkotas (approved):
Well lets try it out.

PR #2281, opened 2022-07-25
Fixes validate-imports

m1kola:

Fixes the issue is that import aliases (from both regex rules and yaml) are only being enforced if an import under the validation uses alias.

The last commit with the correct behaviour is 00fabbb5a3fdeafdfc9cabc098f026cb4f900139.

### DIFF
--- a/pkg/cluster/hive.go
+++ b/pkg/cluster/hive.go
@@ -7,7 +7,7 @@ import (
 	"context"
 	"encoding/json"
 
-	"github.com/openshift/installer/pkg/asset/installconfig/azure"
+	icazure "github.com/openshift/installer/pkg/asset/installconfig/azure"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/hive"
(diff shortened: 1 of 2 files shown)


pkg/cluster/hive.go
```diff
@@ -7,7 +7,7 @@ import (
 	"context"
 	"encoding/json"
 
-	"github.com/openshift/installer/pkg/asset/installconfig/azure"
+	icazure "github.com/openshift/installer/pkg/asset/installconfig/azure"
```
m1kola, line 10: It is an example of the rule which has not been enforced:

https://github.com/Azure/ARO-RP/blob/ed8ab351eaefb151c5f0f1e6589aab290c5fef68/hack/validate-imports/allowed-import-names.yaml#L49-L50

---

Similar thing can happen with regex rules. For example this one:

https://github.com/Azure/ARO-RP/blob/ed8ab351eaefb151c5f0f1e6589aab290c5fef68/hack/validate-imports/validate-imports.go#L144-L147

If you import like this:

```go
operatorfake1 "github.com/openshift/client-go/operator/clientset/versioned/fake"
```

It will be flagged by `validate-imports`. But if you import without an alias, it will not be flagged:

```go
"github.com/openshift/client-go/operator/clientset/versioned/fake"
```

facchettos (approved):
lgtm, I missed it when I implemented it (or maybe I disagreed?). not sure I agree completely with forcing use of aliases, but it's not a big deal either

m1kola:
> not sure I agree completely with forcing use of aliases

Just to clarify: we are only forcing to use aliases in all cases. By default there is no enforcement. But we do force certain patterns (such as importing fake clients from `client-go` as an alias with suffix `fake`) and some cases where we know we will likely going to have multiple imports with the same package name in the same file (e.g. `github.com/Azure/go-autorest/autorest/azure` and `github.com/openshift/hive/apis/hive/v1/azure`).

bennerv (approved):
lgtm

ross-bryan (approved):

PR #2282, opened 2022-07-25
add logic to choose default disk

dem4gus:
https://msazure.visualstudio.com/AzureRedHatOpenShift/_backlogs/backlog/Chainsaw/Stories/?workitem=14884568

Fixes

The current install config does not pass in a disk type for the OSDisk, which results in it being set to `Premium_LRS` by default. However, not all supported VM Sizes are able to use premium disks, so creation of the nodes would fail. This PR explicitly sets OSDisk type for all VMs, either to Standard for those that need it or Premium for any others.

### Test plan for issue:
Unit tests for disk selection logic created in `generateconfig_test.go`. Further e2e tests to ensure Standard-type VMs are available after creation are in progress.

Updates to customer-facing documentation are part of [this story](https://msazure.visualstudio.com/AzureRedHatOpenShift/_boards/board/t/Chainsaw/Stories/?workitem=14884495)


bennerv (changes requested):
IMO we shouldn't be hardcoding these values.  When azure changes or introduces new instance types which do or do not support premium LRS disk types, we're always going to be changing and updating this file.  

Does the VM SKUs API provide anything that will show if it supports a premium disk type?  

```bash
az vm list-skus --all | jq -r '.[] | .name as $NAME | .capabilities[]? | select(.name == "PremiumIO") | [$NAME, .value]' 
```

dem4gus:
> IMO we shouldn't be hardcoding these values. When azure changes or introduces new instance types which do or do not support premium LRS disk types, we're always going to be changing and updating this file.
> 
> Does the VM SKUs API provide anything that will show if it supports a premium disk type?
> 
> ```shell
> az vm list-skus --all | jq -r '.[] | .name as $NAME | .capabilities[]? | select(.name == "PremiumIO") | [$NAME, .value]' 
> ```

It appears it would be possible to do and is similar to what is being done in [upstream validation](https://github.com/openshift/installer/blob/28a38e3b391068c298bb8162bfafe2c1b7eae8c4/pkg/asset/installconfig/azure/validation.go#L92-L103).

That brings up another question. We already have a list of supported VMSizes in [pkg/api/openshiftcluster.go](https://github.com/Azure/ARO-RP/blob/53ab14babcbb658750a2b450424e8cff771261e4/pkg/api/openshiftcluster.go#L288), should we have something similar for the OSDisk type, or are we not worried about hard-coding these values as strings?

dem4gus:
Refactored code to use existing functions to check for `PremiumIO` capability.

Since we have no defined list of VM capabilities that I could find I double-checked [MSFT documentation](https://docs.microsoft.com/en-us/rest/api/compute/resource-skus/list?tabs=HTTP#lists-all-available-resource-skus) to ensure I had the capability name correct.

pkg/installer/generateconfig.go
```diff
@@ -31,6 +32,20 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
+type capabilityChecker func(*mgmtcompute.ResourceSku, string) bool
```
bennerv, line 35: nit: I don't think we need this.  The only method we're passing to the supportedDefaultDisk is `computeskus.HasCapability` and I don't see that changing.  

IMO remove the argument from the `supportedDefaultDisk function and just call it directly.  Would make things a bit cleaner, unless you can justify a use case of passing it as an argument.

bennerv: I guess it makes testing easier, but imo decreases readability.  In your test you could just pass in a vmskus object with premium const or without it.

bennerv (changes requested):
some nits for readability.

pkg/installer/generateconfig.go
dem4gus: Testing was my primary concern when I originally decided to pass the `HasCapability` function in but I was able to do a little more setup in the test itself to avoid having to mock the function out.

pkg/installer/generateconfig.go
```diff
@@ -31,6 +32,18 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
+func supportedDefaultDisk(vmSku *mgmtcompute.ResourceSku) string {
```
dem4gus, line 35: I'm wondering if it would be better to move this entire function and its tests to the `computeskus` package instead of having it here. This is currently the only place we're using it, but if there's anything else that arises from supported disk types in the future it would be beneficial to have it available there.

cadenmarchese: +1 on moving ResourceSku stuff to `computeskus`.

pkg/installer/generateconfig.go
```diff
@@ -148,6 +163,7 @@ func (m *manager) generateInstallConfig(ctx context.Context) (*installconfig.Ins
 						OSDisk: azuretypes.OSDisk{
 							DiskEncryptionSetID: m.oc.Properties.MasterProfile.DiskEncryptionSetID,
 							DiskSizeGB:          1024,
+							DiskType:            masterDiskType,
```
cadenmarchese, line 166:
```suggestion
							DiskType:            supportedDefaultDisk(masterSKU),
```

pkg/installer/generateconfig.go
```diff
@@ -166,6 +182,7 @@ func (m *manager) generateInstallConfig(ctx context.Context) (*installconfig.Ins
 							OSDisk: azuretypes.OSDisk{
 								DiskEncryptionSetID: m.oc.Properties.WorkerProfiles[0].DiskEncryptionSetID,
 								DiskSizeGB:          int32(m.oc.Properties.WorkerProfiles[0].DiskSizeGB),
+								DiskType:            workerDiskType,
```
cadenmarchese, line 185:
```suggestion
								DiskType:            supportedDefaultDisk(workerSKU),
```

cadenmarchese (commented):
Nice work on this, just a few comments. Thanks!

pkg/util/computeskus/computeskus.go
```diff
@@ -77,3 +77,16 @@ func FilterVMSizes(skus []mgmtcompute.ResourceSku, location string) map[string]*
 
 	return vmskus
 }
+
+// SupportedOSDisk returns the type of OSDisk for the given resource. Most VMs will use Premium disks but some SKUs only support Standard SSDs
+func SupportedOSDisk(vmSku *mgmtcompute.ResourceSku) string {
+	const (
+		standardDisk          = "StandardSSD_LRS"
```
dem4gus, line 84: We do not have any types for disks or capabilities so these are string constants.

pkg/util/computeskus/computeskus.go
```diff
@@ -77,3 +77,16 @@ func FilterVMSizes(skus []mgmtcompute.ResourceSku, location string) map[string]*
 
 	return vmskus
 }
+
+// SupportedOSDisk returns the type of OSDisk for the given resource. Most VMs will use Premium disks but some SKUs only support Standard SSDs
+func SupportedOSDisk(vmSku *mgmtcompute.ResourceSku) string {
+	const (
```
facchettos, line 83: is there a reason why this block is directly in the function ?

dem4gus: Just declaring them in the area they're used, if you have a suggestion on a better place to put them I'd be happy to hear it. From what I could find we don't have these defined as a type anywhere like we do VMSizes.

facchettos: So, I first asked that because I just found it a little curious to have a block like that, so it was more out of curiosity. But I just found that you write them again in the test file, while you would be able to just use them if they were declared outside of the function. I would be a little cleaner this way IMO. (talking about lines 265-275 in the test file)

dem4gus: Your English is fine. Better than my French at the very least ;)

Files updated.

pkg/util/computeskus/computeskus_test.go
```diff
@@ -255,3 +255,42 @@ func TestIsRestricted(t *testing.T) {
...
+			"premium_disk_supported",
+			"True",
+			premiumDisk,
+		},
+		{
+			"Premium disk not supported on VMSize",
+			"premium_disk_not_supported",
+			"False",
+			standardDisk,
+		},
```
facchettos, line 277: one last nit, here we should write the names of the fields.
[here](https://github.com/uber-go/guide/blob/master/style.md#initializing-structs) is the bit from the uber style guide. While we don't follow it 100% in the code, it is a nice resource to look at from time to time

facchettos (approved):
Lgtm

bennerv (approved):
super clean - good stuff; lgtm

darthhexx (approved):
LGTM

s-amann (approved):

ross-bryan (approved):
lgtm thanks

PR #2285, opened 2022-07-28
Add expression to exclude logs from openshift-logging namespace

devamanv:
The PR is to exclude logs from the openshift-logging namespace. 

Fixes [Exclude openshift-logging logs from being ingested into Azure log infrastructure](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13146970/)

This is to exclude logs from openshift-logging and openshift-gitops logs going into the Azure Geneva logging infrastructure due to the fact that incorrectly configured customer log ingestion can result in customer workload logs being ingested into Geneva

**Update**:  The PR has been updated to also exclude logs from the `openshift-user-workload-monitoring` namespace

### DIFF
--- a/pkg/operator/controllers/genevalogging/const.go
+++ b/pkg/operator/controllers/genevalogging/const.go
@@ -73,7 +73,7 @@ const (
 [FILTER]
 	Name grep
 	Match containers
-	Regex NAMESPACE ^(?:default|kube-.*|openshift|(?!openshift-(logging|gitops))(openshift-.*))$
+	Regex NAMESPACE ^(?:default|kube-.*|openshift|(?!openshift-(logging|gitops|user-workload-monitoring))(openshift-.*))$
 
 [FILTER]
 	Name nest
--- a/pkg/operator/controllers/genevalogging/const.go
+++ b/pkg/operator/controllers/genevalogging/const.go
@@ -73,7 +73,7 @@ const (
 [FILTER]
 	Name grep
 	Match containers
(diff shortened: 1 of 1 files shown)


pkg/operator/controllers/genevalogging/const.go
```diff
@@ -73,7 +73,7 @@ const (
 [FILTER]
 	Name grep
 	Match containers
-	Regex NAMESPACE ^(?:default|kube-.*|openshift|openshift-.*)$
+	Regex NAMESPACE ^(?:default|kube-.*|openshift|(?!openshift-(logging|gitops))(openshift-.*))$
```
cadenmarchese, line 76: Should we exclude `openshift-user-workload-monitoring` as well?

https://docs.openshift.com/container-platform/4.10/monitoring/enabling-monitoring-for-user-defined-projects.html#enabling-monitoring-for-user-defined-projects_enabling-monitoring-for-user-defined-projects

hawkowl: Yeah, this is a good idea.

devamanv: Thanks @cadenmarchese for pointing this out. Since the linked work item is to exclude any logs from customer workloads, it seems fitting to exclude the said namespace as well.

darthhexx (approved):
lgtm

dem4gus (approved):
Looks good.

PR #2286, opened 2022-07-29
Hive: `ClusterManager` changes and `dynamichelper` support for `ClusterDeployment`

m1kola:

Credit to @hawkowl: it is taked from PR #2215 to make the PR smaller.

* Adds `dynamichelper` support for `ClusterDeployment`: dynamic helper doesn't update the status, but it logs objects diff. We don't care about the diff in status.
* Renaming constructors to be less verbose (it is clear what we return from signrature/return value


petrkotas (approved):
LGTM

PR #2287, opened 2022-07-29
Hive, M5 prep: `ClusterManager` refactoring

m1kola:
* Drops `CreateOrUpdateParameters` from `pkg/hive`. We now pass sub doc and cluster doc into the cluster manager instead having params struct.
* Minor renamings and extra comments

Refactoring to play more nicely with https://github.com/Azure/ARO-RP/pull/2215

### DIFF
--- a/pkg/hive/resources.go
+++ b/pkg/hive/resources.go
@@ -10,16 +10,18 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Changing values of these constants most likely would require
+// some sort of migration on the Hive cluster for existing clusters.
 const (
-	ClusterDeploymentName      = "cluster"
-	kubesecretName             = "admin-kube-secret"
-	servicePrincipalSecretname = "serviceprincipal-secret"
+	ClusterDeploymentName             = "cluster"
+	aroServiceKubeconfigSecretName    = "aro-service-kubeconfig-secret"
+	clusterServicePrincipalSecretName = "cluster-service-principal-secret"
 )
 
-func kubeAdminSecret(namespace string, kubeConfig []byte) *corev1.Secret {
+func aroServiceKubeconfigSecret(namespace string, kubeConfig []byte) *corev1.Secret {
 	return &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      kubesecretName,
+			Name:      aroServiceKubeconfigSecretName,
 			Namespace: namespace,
 		},
 		Data: map[string][]byte{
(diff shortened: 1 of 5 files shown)


pkg/hive/resources.go
```diff
@@ -10,16 +10,18 @@ import (
...
 const (
-	ClusterDeploymentName      = "cluster"
-	kubesecretName             = "admin-kube-secret"
-	servicePrincipalSecretname = "serviceprincipal-secret"
+	ClusterDeploymentName             = "cluster"
+	aroServiceKubeconfigSecretName    = "aro-service-kubeconfig-secret"
+	clusterServicePrincipalSecretName = "cluster-service-principal-secret"
 )
 
-func kubeAdminSecret(namespace string, kubeConfig []byte) *corev1.Secret {
```
m1kola, line 19: Renamed to `aroServiceKubeconfigSecret` to avoid confusion. While from Hive point of view it is an admin kubeconfig, from ARO service perspective - we should not be using admin kube config (intened to be used by the end customer) and use ARO service kubeconfig.

pkg/hive/resources.go
```diff
@@ -10,16 +10,18 @@ import (
...
 
+// Changing values of these constants most likely would require
+// some sort of migration on the Hive cluster for existing clusters.
 const (
-	ClusterDeploymentName      = "cluster"
-	kubesecretName             = "admin-kube-secret"
-	servicePrincipalSecretname = "serviceprincipal-secret"
+	ClusterDeploymentName             = "cluster"
+	aroServiceKubeconfigSecretName    = "aro-service-kubeconfig-secret"
+	clusterServicePrincipalSecretName = "cluster-service-principal-secret"
```
m1kola, line 18: We should not be changing these values (added a comment about it), but since we are still in development - it is ok.

dem4gus (commented):
I notice that none of these files have tests associated with them. Is there a reason for that or is it just something that hasn't been gotten to?

bennerv (approved):
LGTM - will merge at end of day if noone else has any concerns.

m1kola:
> I notice that none of these files have tests associated with them. Is there a reason for that or is it just something that hasn't been gotten to?

@dem4gus yes, i'm shortcutting a bit here. We are not executing these code paths in prod just yet and we are still changing this a lot (like in this PR). So I'm saving time a bit on tests.

But we have tickets to cover this stuff with tests planned:

* [Monitor bit](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15057937) (probably going to be the most stable one)
* [Cluster manager bit](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14962382)
* [`pkg/cluster` bit](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14962351)

ross-bryan (approved):
lgtm

PR #2289, opened 2022-08-01
Hive: fixes the issue with outdated cluster doc

m1kola:

Fixes the issue raised here: https://github.com/Azure/ARO-RP/pull/2215#discussion_r934123823

If we store a cluster doc or sub doc in `clusterManager` struct we risk it getting outdated during the lifetime of the `clusterManager`.

This happens because in the code calling `clusterManager` we do things like this:

```go
m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, func(doc *api.OpenShiftClusterDocument) error {
	// ...
})
```

So `m.doc` receives a new reference, but `clusterManager`  still holds an old one.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -127,7 +127,7 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 	// TODO(hive): always set hiveClusterManager once we have Hive everywhere in prod and dev
 	var hr hive.ClusterManager
 	if hiveRestConfig != nil {
-		hr, err = hive.NewFromConfig(log, subscriptionDoc, doc, hiveRestConfig)
+		hr, err = hive.NewFromConfig(log, _env, hiveRestConfig)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/cluster/hive.go
+++ b/pkg/cluster/hive.go
@@ -43,7 +43,7 @@ func (m *manager) hiveEnsureResources(ctx context.Context) error {
 		return nil
 	}
 
-	return m.hiveClusterManager.CreateOrUpdate(ctx)
+	return m.hiveClusterManager.CreateOrUpdate(ctx, m.subscriptionDoc, m.doc)
 }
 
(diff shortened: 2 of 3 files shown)


bennerv (approved):
+1 - we also had this problem when we deployed the ARO operator.  As a _hack_ we initialize kubernetes clients much later in the installation process.  Not sure if that needs some love like this as well.

m1kola:
@bennerv it is a bit different kind of problem. With ARO operator we have to do that becuase at the beginning of the installation process we do not have an API server and kubeconfig for the cluster. We can only create clients at the certain point.

> +1 - we also had this problem when we deployed the ARO operator. As a _hack_ we initialize kubernetes clients much later in the installation process. Not sure if that needs some love like this as well.

`clusterManager` will be holding hive related clients and Hive should pre-exist at the moment of creation of `clusterManager` (before we even start the installation process). So we are ok from the client perspective.

It is more about patching cosmos DB and holding an old cluster document inside of the `clusterManager`.

So the flow is:

1. We have a cluster `doc`
2. We create a `clusterManager` with cluster `doc`
3. We run `clusterManager.CreateNamespace(...)` and use return value to update `doc` in the cosmos db like this:
    ```go
    m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, func(doc *api.OpenShiftClusterDocument) error {
    	doc.OpenShiftCluster.Properties.HiveProfile.Namespace = namespace.Name
    	return nil
    })
    ```

So the calling code now has `m.doc` pointing to a new updated document,  but `clusterManager` still holds a reference to the old document we gave it on the step 2.

bennerv:
@m1kola My bad I didn't mean installation, I meant AdminUpdate, specifically #2199 

It's the same premise.  We were passing the operator deployment struct a `m.doc`, but updated the doc after we passed it.  As a result, the `m.doc` that the operator was constructing the `cluster.aro.openshift.io` CRD on was stale and resulted in us having to run an `AdminUpdate` multiple times as the second time the contents of the `m.doc` wouldn't change and it would apply correctly.

PR #2291, opened 2022-08-01
Add hive create namespace unittest

bizz001:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14962351/

Clearing technical debt. Adding unit tests to hive related functions.

### Test plan for issue:

This PR consists of adding unit tests only.

No. Tech debt.


pkg/cluster/hive_test.go
```diff
@@ -0,0 +1,156 @@
...
+	defer mockCtrl.Finish()
+	existingNamespaceName := "existing-namespace"
+
+	m, err := createManagerForTests(t, existingNamespaceName)
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+	}
+
+	mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+	mockClusterManager.EXPECT().CreateNamespace(gomock.Any()).Times(0)
```
m1kola, line 90: You can just drop this line and it will be the same as expecting it to be called 0 times.

```suggestion
```

pkg/cluster/hive_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+//Test that namespace is created if it doesn't exist
+func TestHiveCreateNamespaceShouldCreateNamespace(t *testing.T) {
```
m1kola, line 21: Could you please refactor these tests to be a single test function with subtests (e.g. each test case in `t.Run(...)`).

We have a lot of tests like this in the codebase if you want to use it as an example.

m1kola (changes requested):
Had a very quick look. I noticed that mocks are for the old version of the interface. There were quite a few changes recently: probably needs a rebase and re-generation of mocks.

PR #2298, opened 2022-08-02
kubeconfig from AKS cluster WIP

rolandmkunkel:

Fixes

### DIFF
--- a/pkg/env/prod.go
+++ b/pkg/env/prod.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	mgmtcompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
+	mgmtcontainerservice "github.com/Azure/azure-sdk-for-go/services/containerservice/mgmt/2021-10-01/containerservice"
 	"github.com/Azure/go-autorest/autorest/adal"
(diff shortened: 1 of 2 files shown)


pkg/env/prod.go
```diff
@@ -139,6 +141,20 @@ func newProd(ctx context.Context, log *logrus.Entry) (*prod, error) {
 
 	p.serviceKeyvault = keyvault.NewManager(msiKVAuthorizer, serviceKeyvaultURI)
 
+	client := mgmtclusters.NewManagedClustersClient(p.SubscriptionID())
+	client.Authorizer = msiAuthorizer
```
m1kola, line 145: Will we have permissions for list admin creds?

pkg/env/prod.go
```diff
@@ -139,6 +141,20 @@ func newProd(ctx context.Context, log *logrus.Entry) (*prod, error) {
 
 	p.serviceKeyvault = keyvault.NewManager(msiKVAuthorizer, serviceKeyvaultURI)
 
+	client := mgmtclusters.NewManagedClustersClient(p.SubscriptionID())
```
m1kola, line 144: If we are going this way - we should probably have a wrapper in `pkg/util/azureclient` with an interface so we can replace this with a mock.

hawkowl:
I think we can close this now?

PR #2301, opened 2022-08-03
Makes admin patch skip not ready ClusterDeployment

m1kola:

Some of the clusters in our fleet are not going to be alive enough for Hive to connect to their API server and Hive will not report ready condition. This can happen with clusters where VMs were powered off by customers.

If this happens - we do not want to fail admin patch operation.


gvanderpotte (approved):
LGTM

s-amann (approved):

petrkotas (approved):

PR #2303, opened 2022-08-03
add logKind column and filter FB

carvalhe:
https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14781267

Fixes:
Currently asyncQoS data is in ServiceLogs on kusto. We want to separate this data to be in its own table instead by using fluentBit tags.

By using the rewrite_tag rule for fluentbit, we can search for the key LOGKIND, and search for the new log value of asyncQoS. It will then give it a new tag for that value. In another PR afterwards for the pipeline, mdsd will pick up this new tage and create a new table from it.


pkg/backend/openshiftcluster.go
```diff
@@ -295,6 +295,7 @@ func (ocb *openShiftClusterBackend) endLease(ctx context.Context, log *logrus.En
 
 func (ocb *openShiftClusterBackend) asyncOperationResultLog(log *logrus.Entry, initialProvisioningState api.ProvisioningState, backendErr error) {
 	log = log.WithFields(logrus.Fields{
+		"logKind":       "asyncqos",
```
rogbas, line 298: Could you please link where this tag is going to be used?

carvalhe: This tag is used so that when we get to ARO.Pipeline we can use mdsd to create a new table in kusto based off of the fluentBit tag which in this case, it gets from logKind.

weinong: @rogbas there will be another change in mdsd.xml to route this tag to its own table

hawkowl:
@carvalhe 
> Currently asyncQoS data is in ServiceLogs on kusto. We want to separate this data to be in its own table

Why? The workitem doesn't provide much insight here :)

cc @weinong

carvalhe:
> @carvalhe
> 
> > Currently asyncQoS data is in ServiceLogs on kusto. We want to separate this data to be in its own table
> 
> Why? The workitem doesn't provide much insight here :)
> 
> cc @weinong

Ill update the ticket. Essentially Kusto tables are defined in geneva/namespace-configs/public/prod/arorplogs/main.xml where fluentBit tags are checked as an Mdsd event source. With matching tags, you can route that data to be a new table like we have for ServiceLogs. Since we want to have asyncQoS information in its own Kusto table, which is currently in ServiceLogs. To do this, data must be filtered in the fluentBit log collection on the journalD tag. A rewrite_tag filter will get current logs on a tag and filter it into a new table based on a value and can keep the old tag too. Let me know if you have any questions

carvalhe:
Made changes in the mdsd.xml file specifically just for INT for testing purposes, so now the tags will be routed once this is merged.

rogbas (dismissed):
lgtm

rogbas (approved):

rogbas:
@carvalhe lgtm, can you please squash commits to simplify history before we merge?

carvalhe:
Moving to a different PR here https://github.com/Azure/ARO-RP/pull/2329 because of rebase issues.

PR #2305, opened 2022-08-04
Fix panic on AdminUpdate for clusters with custom storage

AldoFusterTurpin:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14743726

### Test plan for issue:

N/A

N/A


hawkowl:
This prevents a panic (and is therefore a net positive). We do need to do some work in making sure that we can perform Admin Updates without one thing causing the entire process to crash but that is outside of scope of this ticket.

hawkowl (approved):

PR #2307, opened 2022-08-04
Scope the OpenShift install version listing by location

darthhexx:
Implements the API changes to support location specific OpenShift install version listing for the az CLI. [WI](https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14177970).

Each region has a dedicated CosmoDB database for storing the installable OpenShift versions. Adding the location parameter to the request ensures the request is sent to the correct ARO-RP instance.

Another use case for the change to scoping to location is that is comes with a _subscription_ scope and that will provide the RP with the capability to return different supported OpenShift versions for specific subscriptions. For example, an internal CI subscription for testing nightly builds.

### Test plan for issue:
Existing unit and e2e tests.

Once the API is finalised for release.


pkg/frontend/frontend.go
```diff
@@ -218,7 +218,7 @@ func (f *frontend) authenticatedRoutes(r *mux.Router) {
 	s.Methods(http.MethodPost).HandlerFunc(f.postOpenShiftClusterKubeConfigCredentials).Name("postOpenShiftClusterKubeConfigCredentials")
 
 	s = r.
-		Path("/providers/{resourceProviderNamespace}/listinstallversions").
```
bennerv, line 221: How are we going to direct this call to the associated RP?  Can you link some MSFT documentation on how the call goes to the specific resource provider based on this information?  

otherwise, we'd have to provide every version available in all RP cosmosdb databases.

bennerv (commented):
Question on this

pkg/frontend/frontend.go
bennerv: If we look at the `az vm list-skus` call it's making a request to this guy: https://management.azure.com/subscriptions/fe16a035-e540-4ab7-80d9-373fa9a3d6ae/providers/Microsoft.Compute/skus?api-version=2019-04-01

This has to be location agnostic.

bennerv: @darthhexx I think this is what we want to implement. https://eng.ms/docs/products/arm/api_contracts/skuupdatecontract

@hawkowl we might have to revert the openshift versions in cosmosdb.  

Note that I can't view the doc on the page but I requested access.

bennerv: It looks like AKS follows a similar pattern to what you used:
`/subscriptions/subId/providers/Microsoft.ContainerService/locations/eastus/orchestrators`

I'm curious if they have store the configuration in a common place where every resource provider queries, every resource provider has all information about all locations, or something else.  Discussing with Weinong.

bennerv: TIL: 
the `/locations` API endpoint appears to route directly to the RP within that location. This looks good to me actually.  Sorry about the confusion, I didn't know that it's how it would work.  

>The usage can also be based on specific location. And in that case the API would look like below –

> GET https://management.azure.com/subscriptions/{subscriptionId}/providers/{providerNameSpace}/locations/{location}/usages?api-version={version}

hawkowl:
I think this is good, just waiting for master to be unbroke before I land more.

hawkowl (approved):

PR #2308, opened 2022-08-04
added hiveEnsureResources unit tests

bizz001:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14962351/

unit tests for hiveEnsureResources

No

### DIFF
--- a/pkg/cluster/hive_test.go
+++ b/pkg/cluster/hive_test.go
@@ -272,3 +272,51 @@ func createManagerForTests(t *testing.T, existingNamespaceName string) *manager 
 	}
 	return m
 }
+
+func TestHiveEnsureResources(t *testing.T) {
+	ctx := context.Background()
+	for _, tt := range []struct {
+		testName           string
+		clusterManagerMock func(mockCtrl *gomock.Controller, m *manager) *mock_hive.MockClusterManager
+		wantErr            string
+	}{
+		{
+			testName: "returns error if cluster manager returns error",
+			clusterManagerMock: func(mockCtrl *gomock.Controller, m *manager) *mock_hive.MockClusterManager {
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().CreateOrUpdate(ctx, m.subscriptionDoc, m.doc).Return(fmt.Errorf("cluster manager error"))
+				return mockClusterManager
+			},
+			wantErr: "cluster manager error",
+		},
+		{
+			testName: "does not return error if cluster manager is nil",
+		},
+		{
+			testName: "calls cluster manager CreateOrUpdate with correct parameters",
+			clusterManagerMock: func(mockCtrl *gomock.Controller, m *manager) *mock_hive.MockClusterManager {
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().CreateOrUpdate(ctx, m.subscriptionDoc, m.doc).Return(nil)
+				return mockClusterManager
+			},
+		},
+	} {
+		t.Run(tt.testName, func(t *testing.T) {
+			controller := gomock.NewController(t)
+			defer controller.Finish()
+
+			m := createManagerForTests(t, "")
+
+			if tt.clusterManagerMock != nil {
+				m.hiveClusterManager = tt.clusterManagerMock(controller, m)
+			}
+
+			err := m.hiveEnsureResources(ctx)
+			if err != nil && err.Error() != tt.wantErr ||
+				err == nil && tt.wantErr != "" {
+				t.Error(err)
+			}
+		},
+		)
+	}
+}


pkg/cluster/hive_test.go
```diff
@@ -272,3 +272,52 @@ func createManagerForTests(t *testing.T, existingNamespaceName string) *manager
...
+			wantErr: "cluster manager error",
+		},
+		{
+			testName: "does not return error if cluster manager is nil",
+		},
+		{
+			testName: "calls cluster manager CreateOrUpdate with correct parameters",
+			clusterManagerMock: func(mockCtrl *gomock.Controller, m *manager) *mock_hive.MockClusterManager {
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().CreateOrUpdate(ctx, gomock.Any(), gomock.Any()).Return(nil)
```
m1kola, line 299: We are not asserting args here

```suggestion
				mockClusterManager.EXPECT().CreateOrUpdate(ctx, m.subscriptionDoc, m.doc).Return(nil)
```

pkg/cluster/hive_test.go
```diff
@@ -272,3 +272,52 @@ func createManagerForTests(t *testing.T, existingNamespaceName string) *manager
...
+				return mockClusterManager
+			},
+		},
+	} {
+		t.Run(tt.testName, func(t *testing.T) {
+			ctx := context.Background()
```
m1kola, line 305: This one shadows the one on line 277.

One of them must go.

pkg/cluster/hive_test.go
```diff
@@ -272,3 +272,52 @@ func createManagerForTests(t *testing.T, existingNamespaceName string) *manager
...
+	for _, tt := range []struct {
+		testName           string
+		clusterManagerMock func(mockCtrl *gomock.Controller, m *manager) *mock_hive.MockClusterManager
+		wantErr            string
+	}{
+		{
+			testName: "returns error if cluster manager returns error",
+			clusterManagerMock: func(mockCtrl *gomock.Controller, m *manager) *mock_hive.MockClusterManager {
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().CreateOrUpdate(gomock.Any(), m.subscriptionDoc, m.doc).Return(fmt.Errorf("cluster manager error"))
```
m1kola, line 287: Nit: let's either use `gomock.Any()` for `ctx` arg or assert it in both places.

```suggestion
				mockClusterManager.EXPECT().CreateOrUpdate(ctx, m.subscriptionDoc, m.doc).Return(fmt.Errorf("cluster manager error"))
```

m1kola (changes requested):
Looks good overal. Only we need to fix mock assertions.

pkg/cluster/hive_test.go
bizz001: I've accidentally added the type assertion to the wrong unit test. I'll fix it

pkg/cluster/hive_test.go
bizz001: I'm not sure I agree. The purpose of this unit test is not to assert parameters. This should be `gomock.Any()` for all params, which I will fix in a sec.

pkg/cluster/hive_test.go
```diff
@@ -272,3 +272,51 @@ func createManagerForTests(t *testing.T, existingNamespaceName string) *manager
...
+			m := createManagerForTests(t, "")
+
+			if tt.clusterManagerMock != nil {
+				m.hiveClusterManager = tt.clusterManagerMock(controller, m)
+			}
+
+			err := m.hiveEnsureResources(ctx)
+			if err != nil && err.Error() != tt.wantErr ||
+				err == nil && tt.wantErr != "" {
+				t.Error(err)
```
s-amann, line 317: nit: the text printed here wouldn't be too helpful for debugging a test, but since these tests are small and short in number I think it is fine

s-amann (dismissed):
lgtm

pkg/cluster/hive_test.go
```diff
@@ -272,3 +272,51 @@ func createManagerForTests(t *testing.T, existingNamespaceName string) *manager
...
+	for _, tt := range []struct {
+		testName           string
+		clusterManagerMock func(mockCtrl *gomock.Controller, m *manager) *mock_hive.MockClusterManager
+		wantErr            string
+	}{
+		{
+			testName: "returns error if cluster manager returns error",
+			clusterManagerMock: func(mockCtrl *gomock.Controller, m *manager) *mock_hive.MockClusterManager {
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().CreateOrUpdate(gomock.Any(), gomock.Any(), gomock.Any()).Return(fmt.Errorf("cluster manager error"))
```
m1kola, line 287:
```suggestion
				mockClusterManager.EXPECT().CreateOrUpdate(ctx, m.subscriptionDoc, m.doc).Return(fmt.Errorf("cluster manager error"))
```

As in "We expect `CreateOrUpdate` to be called with arguments X, Y and Z, but we want to return an error"

In another case it will be "We expect `CreateOrUpdate` to be called with arguments X, Y and Z, but we want to return no error"

pkg/cluster/hive_test.go
m1kola: @s-amann this is our standard snippet for asserting errors in unit tests and it will actualy be enough: you will get test name, subtest/case name, line where test failed and the actual result. You can figure out an expected result from there.

Example:

```
--- FAIL: TestHiveCreateNamespace (0.00s)
    --- FAIL: TestHiveCreateNamespace/returns_error_if_cluster_manager_returns_error (0.00s)
        /Users/mradchuk/workspace/src/github.com/Azure/ARO-RP/pkg/cluster/hive_test.go:233: cluster manager error
```

m1kola (changes requested):

pkg/cluster/hive_test.go
bizz001: I think each test should only test one functionality. The purpose of this unit test is to determine that the behavior of the function is correct in case of an error, regardless of the parameters used, while the other test takes care of the parameters.

If we were to test parameters in both tests, and we introduce a change that breaks the correctness of the parameters being passed to the Cluster Manager, it would not be so obvious that we broke, as both tests would fail.

SudoBrendan: I agree with Mikalai here. For this error, we want to be as specific as possible. This lets us know the exact intent of the code we are testing, which is more important than coverage/just having the test at all.

SudoBrendan (dismissed):
This is really great work so far, thanks for adding more tests for our code! I just want to +1 on Mikalai's comment.

m1kola (dismissed):

m1kola:
@bizz001 this PR looks good, but python tests are failing. There is a fix for this in #2321. You will probably have to rebase your PRs to include the fix once merged.

SudoBrendan (approved):

m1kola (approved):

PR #2311, opened 2022-08-05
Added hiveDeleteResources unit tests

bizz001:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14962351/

unit tests for hiveDeleteResources func

No

### DIFF
--- a/pkg/cluster/hive_test.go
+++ b/pkg/cluster/hive_test.go
@@ -320,3 +320,61 @@ func TestHiveEnsureResources(t *testing.T) {
 		)
 	}
 }
+
+func TestHiveDeleteResources(t *testing.T) {
+	ctx := context.Background()
+	for _, tt := range []struct {
+		testName           string
+		namespace          string
+		clusterManagerMock func(mockCtrl *gomock.Controller, namespaceName string) *mock_hive.MockClusterManager
+		wantErr            string
+	}{
+		{
+			testName: "doesn't return error if cluster manager is nil",
+		},
+		{
+			testName:  "deletes namespace if it exists",
+			namespace: "existing-namespace",
+			clusterManagerMock: func(mockCtrl *gomock.Controller, namespaceName string) *mock_hive.MockClusterManager {
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().Delete(ctx, namespaceName).Return(nil)
+				return mockClusterManager
+			},
+		},
+		{
+			testName: "doesn't attempt to delete namespace if it doesn't exist",
+			clusterManagerMock: func(mockCtrl *gomock.Controller, namespaceName string) *mock_hive.MockClusterManager {
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().Delete(ctx, namespaceName).Times(0)
+				return mockClusterManager
+			},
+		},
+		{
+			testName:  "returns error if cluster manager returns error",
+			namespace: "existing-namespace",
+			clusterManagerMock: func(mockCtrl *gomock.Controller, namespaceName string) *mock_hive.MockClusterManager {
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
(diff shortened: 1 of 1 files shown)


s-amann (dismissed):
lgtm

pkg/cluster/hive_test.go
```diff
@@ -272,3 +272,61 @@ func createManagerForTests(t *testing.T, existingNamespaceName string) *manager
...
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().Delete(gomock.Any(), namespaceName).Return(nil)
+				return mockClusterManager
+			},
+		},
+		{
+			testName: "doesn't attempt to delete namespace if it doesn't exist",
+			clusterManagerMock: func(mockCtrl *gomock.Controller, namespaceName string) *mock_hive.MockClusterManager {
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().Delete(gomock.Any(), gomock.Any()).Times(0)
```
m1kola, line 299: Same as in [the other PR](https://github.com/Azure/ARO-RP/pull/2308#discussion_r938803574): need to assert args

```suggestion
				mockClusterManager.EXPECT().Delete(gomock.Any(), namespaceName).Times(0)
```

pkg/cluster/hive_test.go
```diff
@@ -272,3 +272,61 @@ func createManagerForTests(t *testing.T, existingNamespaceName string) *manager
...
+				return mockClusterManager
+			},
+		},
+		{
+			testName:  "returns error if cluster manager returns error",
+			namespace: "existing-namespace",
+			clusterManagerMock: func(mockCtrl *gomock.Controller, namespaceName string) *mock_hive.MockClusterManager {
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().Delete(gomock.Any(), gomock.Any()).Return(fmt.Errorf("cluster manager error"))
```
m1kola, line 308: Same as above

m1kola (changes requested):
Looks good overall. Just need to update args assertions.

dem4gus (dismissed):
Always a fan of testing our stuff.

pkg/cluster/hive_test.go
SudoBrendan: I completely agree with the other case where we should be passing the NS - but I think this particular one ("don't expect `Delete` to ever be called") seems reasonable to me.

pkg/cluster/hive_test.go
bizz001: added parameters assertion to all tests

pkg/cluster/hive_test.go
```diff
@@ -272,3 +272,61 @@ func createManagerForTests(t *testing.T, existingNamespaceName string) *manager
...
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().Delete(ctx, namespaceName).Return(nil)
+				return mockClusterManager
+			},
+		},
+		{
+			testName: "doesn't attempt to delete namespace if it doesn't exist",
+			clusterManagerMock: func(mockCtrl *gomock.Controller, namespaceName string) *mock_hive.MockClusterManager {
+				mockClusterManager := mock_hive.NewMockClusterManager(mockCtrl)
+				mockClusterManager.EXPECT().Delete(ctx, namespaceName).Times(0)
```
m1kola, line 300: I think we can just drop this line: it will be the same effect. Like if you do not expect a call of this func then you expect the function to be called 0 times.

It is not a big deal in here as it is just one line, but in functions where we mock multiple different calls it just adds a noise into mock setup.

m1kola (approved):
Looks good, but needs to be rebased on top of https://github.com/Azure/ARO-RP/pull/2321 to fix python tests.

m1kola:
Rebased to fix the conflict with no changes to @bizz001's code.

m1kola (approved):
Looks good. Will merge on green.

PR #2312, opened 2022-08-05
update pipeline to only generate release notes

tony-schndr:

Release notes are not currently being generated with EV2 pipelines.

Removes the build steps from the pipeline so it only generates release notes as release notes are not currently being generated with EV2 pipelines.  We don't need build steps as this is also done in EV2 pipelines.

### Test plan for issue:
Need to test in ADO when able.

Updated `docs/releases.md`

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -23,15 +23,16 @@ you are tagging the right commit with the changes you want to release.
 
 ## Release pipeline
 
-Currently the release is done manually via the [pipeline](https://github.com/Azure/ARO-RP/blob/master/.pipelines/build-and-push-images-tagged.yml).
-The pipeline does not have any parameter, instead the pipeline is started on the `tag` as illustrated on the image.
-
-![Start pipelines with tag](img/pipelines.png "Aro Monitor Architecture")
+Currently the release is done manually via the [EV2 pipelines](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/233405/Performing-Release).
 
 
 ### Release page
 
-Once the release is built, new item is added to the [GitHub release page](https://github.com/Azure/ARO-RP/releases) with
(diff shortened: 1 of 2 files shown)


docs/releases.md
```diff
@@ -23,15 +23,16 @@ you are tagging the right commit with the changes you want to release.
...
 
 
 ### Release page
 
-Once the release is built, new item is added to the [GitHub release page](https://github.com/Azure/ARO-RP/releases) with
+The generate release notes pipeline does not have any parameter, instead the pipeline is started on the `tag` as illustrated on the image.
+
+![Start pipelines with tag](img/pipelines.png "Aro Monitor Architecture")
+
+Once the release notes pipeline is finished, new item is added to the [GitHub release page](https://github.com/Azure/ARO-RP/releases) with
```
dem4gus, line 35: Nit: "once the release notes pipeline is finished, _a_ new item is added ..."

dem4gus (dismissed):
Just a tiny grammar thing but looks good overall.

gvanderpotte:
I have a question: why require a run of this pipeline on top of the EV2 ? why not include the steps to create a release in the EV2 (this is a genuine question, I dunno if it is possible)

SudoBrendan (dismissed):
LGTM

dem4gus (dismissed):

tony-schndr:
> why not include the steps to create a release in the EV2 

@gvanderpotte you raise a valid point here.  @bennerv what do you think?  It sounds easy enough though I don't fully understand the complexities added with EV2.

bennerv:
> what do you think? It sounds easy enough though I don't fully understand the complexities added with EV2.

@gvanderpotte this was an easy fix for the release note problem.  Idea is to move it to OB pipelines that run on binary build, but will require the use of the existing service connection.  This is a temporary work-around to get in place until we can add it there.  I think it's easier said than done though.

bennerv (approved):
lgtm

PR #2313, opened 2022-08-08
[M4/M5/M6] Read Hive kubeconfig from env vars and deploy via `aro deploy`

hawkowl:

Fixes https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14625923 for now

Adds the liveconfig but using the approach in #2298 . Also uses a user config, not an admin config.

### Test plan for issue:

Has unit tests, although not sure if it will function in CI.

probably not


pkg/util/liveconfig/hive.go
```diff
@@ -0,0 +1,52 @@
...
+
+	rpResourceGroup := fmt.Sprintf("rp-%s", p.location)
+	rpResourceName := fmt.Sprintf("aro-aks-cluster-%03d", index)
+
+	res, err := p.managedClusters.ListClusterUserCredentials(ctx, rpResourceGroup, rpResourceName, "")
+	if err != nil {
+		return nil, err
+	}
+
+	p.cachedCredentials[index] = *res.Kubeconfigs
```
rolandmkunkel, line 50: can we cache the parsed kubeconfig, so that we don't have to parse it whenever we read from the cache?

pkg/env/core.go
```diff
@@ -31,6 +34,20 @@ func (c *core) IsLocalDevelopmentMode() bool {
...
+func (c *core) NewLiveConfigManager(ctx context.Context) (liveconfig.Manager, error) {
+	if c.isLocalDevelopmentMode {
+		return liveconfig.NewDev(), nil
+	} else {
+		msiAuthorizer, err := c.NewMSIAuthorizer(MSIContextRP, c.Environment().ResourceManagerEndpoint)
+		if err != nil {
+			return nil, err
+		}
+
+		mcc := containerservice.NewManagedClustersClient(c.Environment(), c.SubscriptionID(), msiAuthorizer)
```
rolandmkunkel, line 46: could we also read the kubeconfig once during deployment, so that the RP/GW MSI needs less runtime permissions?

jharrington22:
/hold

Holding this PR as it implements some of the live config changes that we discussed would be released after our go live date for this project. A new PR will be opened that will either require changes here or for this to be closed. cc/ @UlrichSchlueter @petrkotas

hawkowl:
I think https://github.com/Azure/ARO-RP/pull/2325 is a better solution since vars in this manner are not encrypted.

hawkowl:
Closing in preference for #2325

PR #2316, opened 2022-08-08
marshal resources separately to avoid the huge memory consumption by recursive call in MarshalJSON

yjst2012:
fixes issue https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14847686/
analysis: https://docs.google.com/document/d/1Ie-ZWsm3DD94w-fCIAl9XaIjiHgacFWXKamgVrpLNbw 
thanks @cadenmarchese 

marshal resources separately instead of for the whole array of resources to avoid the exponential memory consumption by recursive call in MarshalJSON

### Test plan for issue:
for detailed test plan and results please refer to [this](https://docs.google.com/document/d/1Ie-ZWsm3DD94w-fCIAl9XaIjiHgacFWXKamgVrpLNbw), thanks Caden for the effort.

No, this is to optimise the memory cost to avoid the OOM when listing resources, no change for outside, etc.

### DIFF
--- a/pkg/frontend/adminactions/resources_list.go
+++ b/pkg/frontend/adminactions/resources_list.go
@@ -6,6 +6,7 @@ package adminactions
 import (
 	"context"
 	"encoding/json"
+	"strings"
 
 	mgmtcompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
 	"github.com/Azure/go-autorest/autorest/azure"
@@ -25,7 +26,7 @@ func (a *azureActions) ResourcesList(ctx context.Context) ([]byte, error) {
 	}
 
 	// +4 because we expect +2 subnet and +1 vnets and optional +1 diskEncryptionSet
-	armResources := make([]arm.Resource, 0, len(resources)+4)
+	armResources := make([]arm.Resource, 0, 4)
 	armResources, err = a.appendAzureNetworkResources(ctx, armResources)
 	if err != nil {
 		a.log.Warnf("error when getting network resources: %s", err)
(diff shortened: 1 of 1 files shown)


cadenmarchese:
@yjst2012 thanks for your work on this. Are you able to correct the Go unit testing so that I can re-test this code?

Looking at results, it appears that `{"Resource":{"properties":{}}}` field is not being marshalled in as expected in the unit tests. I'm not immediately sure what the production impact of this would be, but correcting it may have an impact on memory consumption.

yjst2012:
>  fixed, thanks for pointing out that

yjst2012:
misoperation, reopen it

facchettos:
by using the pointer it is seems to be using the MarshalJSON function too

yjst2012:
> by using the pointer it is seems to be using the MarshalJSON function too
correct

>btw instead of using current it would probably be better to use append and create the array with a length of 0 and a capacity of >len(resources)+1. It would make it easier to read
I tried that initially, but the memory consumption is higher than the current solution

facchettos:
How did you measure ? How much memory does it save ?

yjst2012:
pprof, basically similar to the original version

facchettos:
pprof is not a good tool for that. There is no reason why append would use more memory than this solution, it is however a lot easier to read and understand. If you want to have proper results, you can write a benchmark.

yjst2012:
not sure if it is because of pprof, I agree with you that append is neat and more readable, however with some trade off of dynamic mem allocation / swap, which may need proactive GC to cope with in this situation IMHO. this is why I didn't dig much into pprof details and switched to a conservative way in terms of mem allocation to make sure it works in any (GC) condition.

facchettos:
you are wrong, this is not how slices work. When you allocate them with a size of 0 and a capacity of `len` , the underlying array is going to be of size `len`. When you use append, you just update the size if the capacity is greater than `size` + `nb of elem to add`. It is explained here  https://go.dev/blog/slices-intro

If anything when you create a slice of size `len` you have to create zero values for each element, while it's probably not done with a slice of size 0 and capacity `len`. (here zero values aren't very expensive, but anyways)

PR #1882, opened 2021-12-15
ensure that the ARO operator is running at the desired image

SrinivasAtmakuri:
### Which issue this PR addresses:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/11055603/

### What this PR does / why we need it:

This PR creates an extra check condition for aroDepolymentReady to prevent a race condition while running AdminUpdate between ensureAROOperator and aroDeploymentReady functions. There was a scenario where the aro-operator was not running the desired image but the PUCM succeeds this maybe be because we are checking if aroDepolymentReady even before the deployment was actually applied on the cluster.

This PR helps to prevent such a race condition by checking if the deployment has the right version label.

PR #2318, opened 2022-08-09
Tests for hive.manager CreateNamespace

yehornaumenko:
### Which issue this PR addresses:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14962382/

### What this PR does / why we need it:

Unit tests for `pkg/hive/manager.go`, `CreateNamespace` func.

Is there any documentation that needs to be updated for this PR?

No

PR #2319, opened 2022-08-09
make ensureAROOperatorRunningDesiredVersion a condition

SrinivasAtmakuri:
A race condition is present in the current code, this PR
will fix it my modifying the existing action to a condition
with a timeout of 20 mins.

Earlier PR: https://github.com/Azure/ARO-RP/pull/1882

A race condition between the JSON patch applying to the deployment, the m.aroDeploymentReady function returning true immediately as the older revision is still ready and running, and then m.ensureAROOperatorRunningDesiredVersion fetching the older deployment and seeing the version doesn't match.

Fixes


SrinivasAtmakuri:
Yet to test the changes, will share the results soon.

ross-bryan (commented):
first glance lgtm, not sure why python tests are failing tho :/

SrinivasAtmakuri:
the same exact error message is seen here too: https://github.com/Azure/ARO-RP/pull/2318

pkg/cluster/arooperator.go
```diff
@@ -36,16 +36,15 @@ func (m *manager) aroDeploymentReady(ctx context.Context) (bool, error) {
...
 	if !m.isIngressProfileAvailable() {
 		m.log.Error("skip ensureAROOperatorRunningDesiredVersion")
-		return nil
+		// skip and don't retry
+		return true, nil
 	}
-
-	err := m.aroOperatorDeployer.IsRunningDesiredVersion(ctx)
-	if err != nil {
-		m.log.Errorf("cannot ensureAROOperatorRunningDesiredVersion.IsRunningDesiredVersion: %s", err.Error())
```
SrinivasAtmakuri, line 47: With the current implementation, this error message is not logged anymore but is possibly needed. For arooperatorReady and apiServerReady, we do not log the error in RP.

SrinivasAtmakuri:
I tested the changes., the code doesn't exit directly and waits for a timeout of 20 mins while retrying.

gvanderpotte (dismissed):
LGTM

bennerv (changes requested):

pkg/cluster/arooperator.go
```diff
@@ -36,16 +36,15 @@ func (m *manager) aroDeploymentReady(ctx context.Context) (bool, error) {
 	return m.aroOperatorDeployer.IsReady(ctx)
 }
 
-func (m *manager) ensureAROOperatorRunningDesiredVersion(ctx context.Context) error {
+func (m *manager) ensureAROOperatorRunningDesiredVersion(ctx context.Context) (bool, error) {
 	if !m.isIngressProfileAvailable() {
```
bennerv, line 40: Why are we checking the ingress profile is available?

bennerv: Nevermind - I see that we don't deploy/update the ARO operator if the ingress profile is not available, so all these checks would fail anyway. 

Can you put a comment above this stating that?

pkg/operator/deploy/deploy.go
```diff
@@ -374,53 +374,51 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
...
+	return true, nil
 }
 
-func checkPodImageVersion(ctx context.Context, cli corev1client.PodInterface, namespace string, gitCommit string) error {
+func checkPodImageVersion(ctx context.Context, cli corev1client.PodInterface, namespace string, gitCommit string) (bool, error) {
 	podList, err := cli.List(ctx, metav1.ListOptions{LabelSelector: "app=" + namespace})
 	if err != nil {
-		return err
+		return false, err
 	}
```
bennerv, line 392: Again, the pod may not yet be running.  If it doesn't exist, this will error out.  If it doesn't exist, just return false, nil

```suggestion
	switch {
		case kerrors.IsNotFound(err):
			return false, nil
		case err != nil:
			return false, err
		}

File: pkg/cluster/arooperator.go [hunk 4]
```diff
@@ -29,23 +30,24 @@ func (m *manager) ensureAROOperator(ctx context.Context) error {
...
+		// If the ingress profile is not available, ARO operator update/deploy will fail.
 		m.log.Error("skip ensureAROOperatorRunningDesiredVersion")
-		return nil
+		// skip and don't retry
```
bennerv, line 45: No need for this comment as the log error message says it's going to skip it.  A bit redundant.

pkg/operator/deploy/deploy.go
```diff
@@ -374,53 +374,57 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
...
-		return errors.New(name + " is not running the desired version: " + gitCommit)
+		return false, nil
 	}
-
-	return nil
+	return true, nil
 }
 
-func checkPodImageVersion(ctx context.Context, cli corev1client.PodInterface, namespace string, gitCommit string) error {
+func checkPodImageVersion(ctx context.Context, cli corev1client.PodInterface, namespace string, gitCommit string) (bool, error) {
```
bennerv, line 391: The `namespace` parameter is misleading.  This isn't actually the namespace, but the "role".  Please rename this parameter to `role` as the role of the operator is either a `worker` or `master`.

bennerv (changes requested):
Few more nits.  Can we get some unit tests for these newer functions as well?  

checkOperatorDeploymentVersion
checkPodImageVersion
IsRunningDesiredVersion

SrinivasAtmakuri:
Written unit tests for the functions and made suggested changes. cc @bennerv

pkg/operator/deploy/deploy.go
```diff
@@ -374,53 +374,68 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
...
 
-func (o *operator) IsRunningDesiredVersion(ctx context.Context) error {
-	// check if aro-operator-master is running desired version
-	err := checkOperatorDeploymentVersion(ctx, o.kubernetescli.AppsV1().Deployments(pkgoperator.Namespace), "aro-operator-master", version.GitCommit)
-	if err != nil {
-		return err
+func (o *operator) IsRunningDesiredVersion(ctx context.Context) (bool, error) {
+	// Get the desired Version
+	image := o.env.AROOperatorImage()
+	desiredVersion := version.GitCommit
```
bennerv, line 411: Let's not tie the `desiredVersion` to the `version.GitCommit`.  In local dev mode it will fail always.  Instead do something like this

```
	desiredVersion := "latest"
	if strings.Contains(image, ":") {
		str := strings.Split(image, ":")
		desiredVersion = str[len(str)-1]
	}

File: pkg/operator/deploy/deploy.go [hunk 7]
```diff
@@ -374,53 +374,68 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
...
+		return false, nil
+	case err != nil:
+		return false, err
 	}
 	for _, pod := range podList.Items {
 		imageTag := strings.Split(pod.Spec.Containers[0].Image, ":")
-		if imageTag[len(imageTag)-1] != gitCommit {
-			return errors.New(pod.Name + " pod of namespace " + pod.Namespace + " is not running the desired version: " + gitCommit)
+		if imageTag[len(imageTag)-1] != desiredVersion {
+			return false, nil
```
bennerv, line 402: if `desiredVersion` == `latest` then we it's possible there isn't a tag on it, so this could fail.  We should check if it contains a `:` character, and if not the imageTag is `latest`

bennerv (changes requested):
few things - also rebase and test it locally with the changes of mine that were merged in today.

s-amann (commented):
overall LGTM, I agree with Ben we should default to latest for the desiredVersion

SrinivasAtmakuri:
Ack, made suggested change and testing the PR now, will update results. Thank you.

SrinivasAtmakuri:
Ack, made suggested change and testing the PR now, will update results. Thank you.

SrinivasAtmakuri:
Manual Testing is done for the below scenarios.,

Created a dev cluster via local-RP.

1. ARO_IMAGE is set, but the image doesn't have a tag, PUCM succeeds and looks for desiredVersion == "latest"
2. ARO_IMAGE modified and the image has a tag, PUCM succeeds and checks for desiredVersion == "imageTag"
3. desiredVersion is hard coded to a random string (say unknown), PUCM fails at step ensureAROOperatorRunningDesiredVersion-a-condition after a timeout of 5 mins.

pkg/operator/deploy/deploy.go
```diff
@@ -378,53 +377,68 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
...
+func checkPodImageVersion(ctx context.Context, cli corev1client.PodInterface, role string, desiredVersion string) (bool, error) {
+	podList, err := cli.List(ctx, metav1.ListOptions{LabelSelector: "app=" + role})
+	switch {
+	case kerrors.IsNotFound(err):
+		return false, nil
+	case err != nil:
+		return false, err
 	}
 	for _, pod := range podList.Items {
 		imageTag := strings.Split(pod.Spec.Containers[0].Image, ":")
```
bennerv, line 403: This needs an update.  There might not be a tag on the image.

pkg/cluster/arooperator.go
```diff
@@ -29,23 +30,22 @@ func (m *manager) ensureAROOperator(ctx context.Context) error {
...
-		return nil
+		return true, nil
 	}
-
-	err := m.aroOperatorDeployer.IsRunningDesiredVersion(ctx)
-	if err != nil {
-		m.log.Errorf("cannot ensureAROOperatorRunningDesiredVersion.IsRunningDesiredVersion: %s", err.Error())
+	ok, err := m.aroOperatorDeployer.IsRunningDesiredVersion(ctx)
+	if !ok || err != nil {
+		return false, nil
```
bennerv, line 48:
```suggestion
		return false, err

File: pkg/operator/deploy/deploy.go [hunk 10]
```diff
@@ -378,53 +377,68 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
...
-		if imageTag[len(imageTag)-1] != gitCommit {
-			return errors.New(pod.Name + " pod of namespace " + pod.Namespace + " is not running the desired version: " + gitCommit)
+		if imageTag[len(imageTag)-1] != desiredVersion {
+			return false, nil
```
bennerv, line 405:
```
tag := "latest"
if strings.Contains(pod.Spec.Containers[0].Image, ":") {
    str := strings.Split(pod.Spec.Containers[0].Image, ":"
    tag := str[len(str)  - 1]
}

if tag != desiredVersion {
    return false, nil
}

bennerv (changes requested):
almost there

SrinivasAtmakuri:
Manual Testing is done for the below scenarios.,

I created a dev cluster via local-RP.

1. ARO_IMAGE is set, but the image doesn't have a tag, PUCM succeeds and looks for desiredVersion == "latest"
2. ARO_IMAGE modified and the image has a tag, PUCM succeeds and checks for desiredVersion == "imageTag"
3. desiredVersion is hardcoded to a random string (say unknown), PUCM fails at step ensureAROOperatorRunningDesiredVersion-a-condition after a timeout of 5 mins.
4. OperatorVersion is passed via Payload, PUCM succeeds and looks for desiredVersion == "OperatorVersion">
'{ "Properties": { "OperatorVersion": "v20220729.01" } }'

bennerv (approved):
LGTM

In a follow-up PR (not this one so we can get this change merged in), can we update the code for the following:
`checkOperatorDeploymentVersion` and `checkPodImageVersion` are only called within the operator package.  Let's make them functions specific to the operator, and then we don't have to pass in clients such as `corev1client.PodInterface`, but can use `o.kubernetescli`

ross-bryan (approved):
lgtm

PR #2320, opened 2022-08-09
re-implement cloud aware function for e2e test

dem4gus:
Fixes e2e test failures in USGovCloud environments. [#15119913](https://msazure.visualstudio.com/AzureRedHatOpenShift/_boards/board/t/Chainsaw/Stories/?workitem=15119913)

Fixes a previously hard-coded string that was causing test failures.

### Test plan for issue:
Run test in USGovCloud environment. Test now re-implements cloud aware logic from controller so confidence is high that it will work correctly.

N/A

### DIFF
--- a/test/e2e/operator.go
+++ b/test/e2e/operator.go
@@ -31,6 +31,7 @@ import (
 
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 	"github.com/Azure/ARO-RP/pkg/operator/controllers/monitoring"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
 	"github.com/Azure/ARO-RP/pkg/util/conditions"
 	"github.com/Azure/ARO-RP/pkg/util/ready"
 	"github.com/Azure/ARO-RP/pkg/util/subnet"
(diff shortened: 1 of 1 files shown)


test/e2e/operator.go
```diff
@@ -433,10 +434,20 @@ var _ = Describe("ARO Operator - ImageConfig Reconciler", func() {
...
+		var requiredRegistries []string
+		switch instance.Spec.AZEnvironment {
+		case azureclient.PublicCloud.Environment.Name:
+			requiredRegistries = []string{instance.Spec.ACRDomain, "arosvc." + instance.Spec.Location + ".data." + azure.PublicCloud.ContainerRegistryDNSSuffix}
+
+		case azureclient.USGovernmentCloud.Environment.Name:
+			requiredRegistries = []string{instance.Spec.ACRDomain, "arosvc." + instance.Spec.Location + ".data." + azure.USGovernmentCloud.ContainerRegistryDNSSuffix}
+
+		default:
+			return nil, fmt.Errorf("cloud environment %s is not supported", instance.Spec.AZEnvironment)
```
cadenmarchese, line 448: I agree we should fail e2e if environment is unsupported (or if environment is blank/deleted), but not sure if we should fail it during this test specifically.

cadenmarchese (commented):
Nice fix, one suggestion - I would personally rather skip the test if environment is blank or unsupported rather than fail the test, but it's an edge case.

test/e2e/operator.go
dem4gus: That's a good point but I think I actually like it the way it is now. There's already a check in the `BeforeEach()` that will skip the tests if the feature flag is unset, having an error here will tell us if the flag is enabled on an unsupported cloud environment for whatever reason.

bennerv (approved):
lgtm

ross-bryan (approved):

PR #2323, opened 2022-08-12
Add az list install versions

carlowisse:
ADO Card: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14177970

Extended az command to accept a new command: `az aro list-install-versions --location <LOCATION> -o table`

### Test plan for issue:
Manually tested new command against development RP with and without location string as well as with and without table output.

No, documentation is a part of the command --help flag.


python/az/aro/azext_aro/_help.py
```diff
@@ -31,6 +31,16 @@
     text: az aro list -o table
 """
 
+helps['aro list-install-versions'] = """
```
ross-bryan, line 34: can we change the command to something like `az aro get-versions`?

ross-bryan (changes requested):

python/az/aro/azext_aro/_help.py
bennerv: +1, it'll align [with AKS' command](https://docs.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#azure-portal-and-cli-versions).  We should also run this by Jerome to make sure he's good with it.

petrkotas (approved):
@carlowisse thanks for all the changes. I see you have updated the code according to the comments. And the code looks good.

jboutaud:
+1 for consistency with AKS. The command should be az aro get-versions

python/az/aro/azext_aro/_help.py
carlowisse: I have updated the command to align with AKS'.

ross-bryan (dismissed):
thanks for the update. lgtm.

python/az/aro/azext_aro/tests/latest/integration/test_aro_scenario.py
```diff
@@ -30,6 +30,10 @@ def test_aro(self, resource_group):
             self.check('tags.foo', 'boo')
         ])
 
+        self.cmd('az aro get-versions --location eastus', checks=[
+            self.check('4.10.20')
+        ])
+
```
darthhexx, line 36: While this test will work now, with the hard coded "4.10.20" return value coming from the RP, that is going to very shortly be switched to dynamically loading from the CosmoDB OpenShiftVersions database. When that happens, this test will fail. In short, since we know it will be redundant very shortly, it's better to leave it out.

darthhexx (changes requested):

python/az/aro/azext_aro/tests/latest/integration/test_aro_scenario.py
carlowisse: Thanks David, have updated with a commit to remove the hard coded test.

darthhexx (approved):
lgtm

PR #2324, opened 2022-08-12
10581366-validate the current OCP version

LiniSusan:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/10581366
Fixes

Created a new function: validateVersion in pkg/frontend/validate.go to fetch the list of installed version and compare whether current version is listed in it or not. Invoked this function in pkg/frontend/openshiftcluster_putorpatch.go
Also added new error constant: CloudErrorCodeInvalidVersion in pkg/api/error.go for version validation

### DIFF
--- a/pkg/frontend/openshiftcluster_putorpatch.go
+++ b/pkg/frontend/openshiftcluster_putorpatch.go
@@ -28,19 +28,20 @@ func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Req
 	ctx := r.Context()
 	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
 	vars := mux.Vars(r)
+	installVersion := vars["install-version"]
 
 	var header http.Header
 	var b []byte
 	err := cosmosdb.RetryOnPreconditionFailed(func() error {
 		var err error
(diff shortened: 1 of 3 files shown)


pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -28,19 +28,20 @@ func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Req
 	ctx := r.Context()
 	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
 	vars := mux.Vars(r)
+	installVersion := vars["install-version"]
```
ross-bryan, line 31: install version shouldn't be coming from vars, it should be in the body of the payload

ross-bryan (changes requested):
install version is not a URL parameter, it is in body of payload

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -58,6 +59,10 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 	isCreate := doc == nil
 
 	if isCreate {
+		err := f.validateVersion(installVersion)
```
ross-bryan, line 62: we should also validate during a PATCH request that the clusterProfile version is not modified

darthhexx:
Closing this PR, because the functionality was added into #2374

PR #2325, opened 2022-08-16
[M4/M5/M6] Read Hive kubeconfig from AKS live

hawkowl:

Fixes https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14625923 for now

Adds the liveconfig but using the approach in https://github.com/Azure/ARO-RP/pull/2298 . Also uses a user config, not an admin config.

Test plan for issue:

Has unit tests, although not sure if it will function in CI.

probably not


pkg/util/liveconfig/hive.go
```diff
@@ -0,0 +1,52 @@
...
+		return nil, err
+	}
+
+	return restConfig, nil
+}
+
+func (p *prod) HiveRestConfig(ctx context.Context, index int) (*rest.Config, error) {
+	cached, ext := p.cachedCredentials[index]
+	if ext {
+		return parseKubeconfig(cached)
```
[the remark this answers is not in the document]
hawkowl: not really, fixed :)

pkg/env/env.go
```diff
@@ -51,6 +52,7 @@ const (
 	GatewayKeyvaultSuffix            = "-gwy"
 	PortalKeyvaultSuffix             = "-por"
 	ServiceKeyvaultSuffix            = "-svc"
+	LiveConfigKeyvaultSuffix         = "-cfg"
```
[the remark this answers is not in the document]
hawkowl: yep, good catch, this is from paring back from the keyvault live config PR

pkg/util/liveconfig/hive.go
```diff
@@ -0,0 +1,57 @@
...
+	restConfig, err := clientconfig.ClientConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	return restConfig, nil
+}
+
+func (p *prod) HiveRestConfig(ctx context.Context, index int) (*rest.Config, error) {
+	cached, ext := p.cachedCredentials[index]
```
petrkotas, line 37: Since we know rotation happens every 30 years, this is fine. But in general I would like to see TTL for caching.

petrkotas (commented):
This is a really good work. Thank you @hawkowl. I have one comment, no-op mentioned for the future optimizations, and one question.
I see vendoring changed, but I do not see `go.mod` and `go.sum` changes. Did the changes occur without modifying these two? Just double checking. If yes, than lets get it in.

pkg/env/core.go
```diff
@@ -31,6 +34,20 @@ func (c *core) IsLocalDevelopmentMode() bool {
...
+	} else {
+		msiAuthorizer, err := c.NewMSIAuthorizer(MSIContextRP, c.Environment().ResourceManagerEndpoint)
+		if err != nil {
+			return nil, err
+		}
+
+		mcc := containerservice.NewManagedClustersClient(c.Environment(), c.SubscriptionID(), msiAuthorizer)
+		return liveconfig.NewProd(c.Location(), mcc), nil
+	}
+}
```
facchettos, line 49: The else is unnecessary here

pkg/util/liveconfig/hive.go
facchettos: what does `ext` mean ? Shouldn't it be `exists` or `ok`?

pkg/util/liveconfig/hive.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	parsed, err := parseKubeconfig(*res.Kubeconfigs)
+	if err != nil {
+		return nil, err
+	}
+
+	p.cachedCredentials[index] = parsed
```
facchettos, line 55: Are we sure there isn't going to be concurrent access on that map ? We should have a `RWMutex` to protect the map from concurrent access

facchettos (changes requested):

pkg/util/liveconfig/dev.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"fmt"
+	"os"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const (
+	HIVEENVVARIABLE = "HIVEKUBECONFIGPATH"
```
s-amann, line 16: can we make this a bit more readable?
`HIVE_ENV_VARIABLE = "HIVE_KUBE_CONFIG_PATH"`

pkg/util/liveconfig/hive.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (p *prod) HiveRestConfig(ctx context.Context, index int) (*rest.Config, error) {
+	cached, ext := p.cachedCredentials[index]
+	if ext {
+		return cached, nil
+	}
+
+	rpResourceGroup := fmt.Sprintf("rp-%s", p.location)
+	rpResourceName := fmt.Sprintf("aro-aks-cluster-%03d", index)
```
s-amann, line 43: does each location get its own indexed aks cluster? Is this information held someplace to look up?

s-amann (dismissed):
This seems straightforward, or maybe I'm missing something. Just some nits below.

pkg/util/liveconfig/hive.go
hawkowl: This is to support multiple shards (AKS clusters) per region, since that is how you scale Hive. Right now, we only have one shard.

pkg/env/core.go
hawkowl: fixed

pkg/util/liveconfig/hive.go
hawkowl: yeah, I figured that it would be more complicated than it's worth here. +1 for TTLs for stuff, though.

pkg/util/liveconfig/hive.go
hawkowl: Good call. Added that, isn't too complicated.

pkg/util/liveconfig/hive.go
```diff
@@ -0,0 +1,67 @@
...
+	// for the existing Azure API call to be done.
+	p.hiveCredentialsMutex.Lock()
+	defer p.hiveCredentialsMutex.Unlock()
+
+	rpResourceGroup := fmt.Sprintf("rp-%s", p.location)
+	rpResourceName := fmt.Sprintf("aro-aks-cluster-%03d", index)
+
+	res, err := p.managedClusters.ListClusterUserCredentials(ctx, rpResourceGroup, rpResourceName, "")
+	if err != nil {
+		return nil, err
```
jharrington22, line 57: Do we usually wrap any of these errors to make debugging easier later?

pkg/backend/openshiftcluster.go
```diff
@@ -102,7 +101,7 @@ func (ocb *openShiftClusterBackend) handle(ctx context.Context, log *logrus.Entr
 		return err
 	}
 
-	hiveRestConfig, err := hive.HiveRestConfig()
+	hiveRestConfig, err := ocb.env.LiveConfig().HiveRestConfig(ctx, 1)
 	if err != nil {
 		log.Info(err) // TODO(hive): Update to fail once we have Hive everywhere in prod and dev
```
jharrington22, line 106: Its likely we could get AKS everywhere before this is merged. Either way can we not log an error here now? 

Also if there is an error getting MSI credentials inside `HiveRestConfig` we will we retry this?

pkg/util/liveconfig/hive.go
hawkowl: Usually? No. We don't have a good observability story.

pkg/backend/openshiftcluster.go
hawkowl: I think this is out of scope for this PR. 

As far as retrying the MSI, there's a lot of other places it will crash first.

facchettos (approved):

petrkotas:
The remaining comments are minor changes, which can be done by the follow PR.

s-amann (approved):

petrkotas (approved):
Thanks for all the changes. Lets get it in.

PR #2326, opened 2022-08-16
Tests for hive.manager CreateOrUpdate

yehornaumenko:
https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14962382/

**What this PR does / why we need it:**
Unit tests for pkg/hive/manager.go, CreateOrUpdate func.

No

### DIFF
--- a/pkg/hive/manager_test.go
+++ b/pkg/hive/manager_test.go
@@ -5,9 +5,11 @@ package hive
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"testing"
 
+	"github.com/golang/mock/gomock"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	hivefake "github.com/openshift/hive/pkg/client/clientset/versioned/fake"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/hive/manager_test.go
```diff
@@ -158,3 +162,46 @@ func TestResetCorrelationData(t *testing.T) {
...
+				},
+			},
+		},
+	}
+
+	controller := gomock.NewController(t)
+	dh := mock_dynamichelper.NewMockInterface(controller)
+	cm := clusterManager{dh: dh}
+
+	dh.EXPECT().Ensure(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
```
m1kola, line 198: I think we either need to drop this test as it is not very useful or do some checking on what we receive here. Checking will be tricky.

yehornaumenko: I am not sure what can be checked here. I guess you mean mocking `restcli` inside `dynamicHelper.ensureOne` (and checking for expected API calls), do you?

facchettos: If you were able to mock the call to `func (hr *clusterManager) resources(sub *api.SubscriptionDocument, doc *api.OpenShiftClusterDocument) ([]kruntime.Object, error) {` you'd be able to have something that you can actually check , (and thus replace the 4 `gomock.Any()` with some proper values). Not sure if it is feasible though

petrkotas: I honestly believe covering the `clusterManager` with unit tests does not make much sense. It would be better suited for integration test, when it actually checks that it calls the stuff OK.

PR #2327, opened 2022-08-16
Test hive create namespace

yehornaumenko:
https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14962382/

Unit tests for pkg/hive/manager.go, CreateNamespace func.

No


petrkotas (approved):
@yehornaumenko Thank you for the PR. Good work.

PR #2328, opened 2022-08-16
Persist tags on ensureResourceGroup

bennerv:

Fixes [this](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15272241/)

Currently, if we PUCM a cluster with tags set on the cluster resource group, it will remove them.  We do allow tagging of the clsuter resource group, so we shouldn't be removing these.  

### Test plan for issue:

Unit tests & tested locally. 

no,


ross-bryan (approved):
lgtm

darthhexx:
:1st_place_medal:

PR #2329, opened 2022-08-16
add logKind column and filter FB

carvalhe:
https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14781267

Fixes:
Currently asyncQoS data is in ServiceLogs on kusto. We want to separate this data to be in its own table instead by using fluentBit tags.

By using the rewrite_tag rule for fluentbit, we can search for the key LOGKIND, and search for the new log value of asyncQoS. It will then give it a new tag for that value. In another PR afterwards for the pipeline, mdsd will pick up this new tage and create a new table from it.

This is a remake of a similar PR which was having issues rebasing


rogbas (approved):
lgtm

PR #2330, opened 2022-08-17
Do not fail cluster installation if the default storage account already exists

bennerv:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15298895

If the installation stops for any reason after the default storage account is already created, and then attempts to resume creation, it will fail since the default storage account already was created.  We shouldn't fail on this error.  

It's a very rare scenario, but technically possible and our installation should be idempotent.  

### Test plan for issue:

N/A

N/A


ross-bryan (commented):
lgtm

ross-bryan (approved):

cadenmarchese (approved):
thx!

PR #2331, opened 2022-08-17
Refactor how we version the aro operator deploymennt

bennerv:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15288173

Currently, if `o.oc.properties.OperatorVersion` is not set (i.e. we're not overriding the operator version we deploy into the cluster), then the label on the aro-operator deployments is set to `unknown`, which results in PUCM always failing since we very rarely override the OperatorVersion. 

### Test plan for issue:

This needs some unit tests & manual testing

manual testing:
1.  ARO_IMAGE set & no tag set -> version label should be "latest"
2. ARO_IMAGE set & a tag set -> version label should be tag
3. OperatorVersion set & overridden -> version label should be "operatorVersion"

AdminUpdate with the above scenarios.  They should all work as documented above

no

### DIFF
--- a/pkg/operator/deploy/deploy.go
+++ b/pkg/operator/deploy/deploy.go
@@ -89,9 +89,7 @@ func New(log *logrus.Entry, env env.Interface, oc *api.OpenShiftCluster, arocli 
 type deploymentData struct {
 	Image              string
 	Version            string
-	GitCommit          string
 	IsLocalDevelopment bool
-	HasVersion         bool
 }
 
 func templateManifests(data deploymentData) ([][]byte, error) {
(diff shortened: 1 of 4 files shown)


SrinivasAtmakuri:
LGTM, based on this commit, I should probably add a few additional checks in the ensureAROOperatorRunningDesiredVersion step, ref: https://github.com/Azure/ARO-RP/pull/2319

bennerv:
I want to add unit tests before this gets merged in so put a hold label on it.

pkg/operator/deploy/deploy.go
```diff
@@ -125,18 +123,24 @@ func templateManifests(data deploymentData) ([][]byte, error) {
 
 func (o *operator) createDeploymentData() deploymentData {
 	image := o.env.AROOperatorImage()
-	hasVersion := false
+
+	// HACK: Override for ARO_IMAGE env variable setup in local-dev mode
+	version := "latest"
```
bennerv, line 128: This will override any previous "unknown" value in local dev to "latest".  

So if you're running local dev mode, you're image tag will always be the tag of the image, or latest if there is no tag.  This should allow us to run the admin update tasks Srinivas introduced and still be successful.

petrkotas (approved):
Good one, thanks @benerv

PR #2333, opened 2022-08-17
Switch RP NSG from using rpNsgSourceAddressPrefixes to static GenevaActions service tag

ventifus:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14930561/

Geneva Actions added support for service tags earlier this year. As a result, Microsoft has requested we stop using fixed lists of IPs in our NSGs and instead use the `GenevaActions` service tag. This PR removes support for the rpNsgSourceAddressPrefixes key from the deploy configuration and instead adds an NSG allowing GenevaActions the same  way we currently allow AzureResourceManager. 

There will be two corresponding PRs for the RP-Config repo, one updating the INT and FF INT configuration for testing, and one updating Prod.

### Test plan for issue:

We will deploy these changes to INT/FF INT and confirm that Geneva Actions works there.

There is no documentation for rpNsgSourceAddressPrefixes.


ventifus:
Do not merge. Need review and testing.

ross-bryan (commented):
lgtm, lets deploy to both public and FF INT and confirm geneva actions still work :-)

ventifus:
Test failed.
`2022-08-19 21:47:02.361339: time="2022-08-19T21:47:02Z" level=fatal msg="features.DeploymentsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code=\"InvalidTemplateDeployment\" Message=\"The template deployment 'rp-production-predeploy' is not valid according to the validation procedure. The tracking id is '11097388-19a9-4865-9d7c-c5306dd80860'. See inner errors for details.\" Details=[{\"code\":\"SecurityRuleParameterContainsUnsupportedValue\",\"details\":[],\"message\":\"Security rule parameter SourceAddressPrefix for rule with Id /subscriptions/0cc1cafa-578f-4fa5-8d6b-ddfd8d82e6ea/resourceGroups/aro-int-eastus/providers/Microsoft.Network/networkSecurityGroups/rp-nsg/securityRules/rp_in_geneva cannot specify existing VIRTUALNETWORK, INTERNET, AZURELOADBALANCER, '*' or system tags. Unsupported value used: GenevaActions.\"}]" func="main.main()" file="cmd/aro/main.go:88"
`

ventifus:
Tested successfully in public int; confirmed Geneva Actions can still talk to the RP after updating the NSG.

bennerv (dismissed):
We still need a way to hardcode the deployNSGs to true for the next release.

I think we should add a commit that we plan on reverting with a big `// TODO` in front setting the value to true no matter what, and reverting it after the fact.  So we get the most up-to-date NSG changes we introduce in this PR.

bennerv:
Another thing too:

We probably want to ensure this works in Fairfax, just to make sure the service tag exists.  If you wanted to kick off a deploy to FF INT that would confirm everything works as expected.

ross-bryan (approved):
lgtm, lets just make sure the GA's work in FF still too

PR #2334, opened 2022-08-18
Create update OpenShift versions command

ellis-johnson:

Implements https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/10581314/ 

Creating ARO command to update the minimum ocp versions allowed when creating and updating ARO clusters.

### Test plan for issue:
Unit tests and manual testing

Both in repo documentation and internal ADO Documentation


bennerv:
Few questions here:
1. Why are we not dynamically building the supported versions within the RP itself by parsing Cincinnati stable channel?  
    - I can understand that we haven't tested & vetted the versions, but how do we plan on testing/vetting the versions before updating them anyway? Do we have a plan
    - I guess it would be harder to pull versions in the case of a bad release.  Ideally the release team would do so.
    
2. The current way this is structured it has 4.11 in it which isn't ready (I know it's in draft, but just a heads up) as well as it includes multiple z-stream versions when there is a minSupportedZStream version.  I think this was meant to look differently.

3. What's the current workflow to run / update this?  Is it going to run on every RP release, or run independently of them?  If so, how does this play into SDP?

darthhexx:
@bennerv thanks for the questions:
1. This is what we were implementing, but TL's asked for this to be descoped from the initial implementation and the WI was modified to this effect. I'll be creating a design document to circulate for approval.
2. It was supposed to be just a minimum X.Y.Z version for each supported X.Y version from which the versions from the Cincinnati stable channel would be added. The explanation of this process will be explained in greater detail in the design doc. 
3. The intention is to run this process in a separate pipeline for each sector on a regular, yet TBD, cadence.

cmd/aro/update_ocp_versions.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	dstAcr := os.Getenv("DST_ACR_NAME")
+	acrDomainSuffix := "." + env.Environment().ContainerRegistryDNSSuffix
+
+	dstRepo := dstAcr + acrDomainSuffix
+	var (
+		OpenshiftVersions = []api.OpenShiftVersion{
+			{
+				Version:           "4.10.20",
+				OpenShiftPullspec: "quay.io/openshift-release-dev/ocp-release@sha256:e1fa1f513068082d97d78be643c369398b0e6820afab708d26acda2262940954",
```
hawkowl, line 34: @darthhexx should we take this from version/const.go?

```
Version: version.InstallStream.Version.String(),
OpenShiftPullspec: version.InstallStream.PullSpec,
InstallerPullspec: fmt.Sprintf("%s/aro-installer:release-%s", dstRepo, version.InstallStream.Version.MinorVersion()),
```

darthhexx: Good call, done.

cmd/aro/update_ocp_versions.go
```diff
@@ -0,0 +1,176 @@
...
+	dstAcr := os.Getenv("DST_ACR_NAME")
+	acrDomainSuffix := "." + env.Environment().ContainerRegistryDNSSuffix
+
+	dstRepo := dstAcr + acrDomainSuffix
+	var (
+		OpenshiftVersions = []api.OpenShiftVersion{
+			{
+				Version:           version.InstallStream.Version.String(),
+				OpenShiftPullspec: version.InstallStream.PullSpec,
+				InstallerPullspec: dstRepo + "/aro-installer:release-4.10",
```
jharrington22, line 36: Hard coded value required here?

cmd/aro/update_ocp_versions.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+		}
+	}
+
+	msiAuthorizer, err := _env.NewMSIAuthorizer(env.MSIContextRP, _env.Environment().ResourceManagerEndpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	msiKVAuthorizer, err := _env.NewMSIAuthorizer(env.MSIContextRP, _env.Environment().ResourceIdentifiers.KeyVault)
```
jharrington22, line 74: nit: Could rename the var to be more obvious. Not blocking.

cmd/aro/update_ocp_versions.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return nil, err
```
jharrington22, line 76: Should we consider returning which authorizer failed here for better debugging?

cmd/aro/update_ocp_versions.go
ellis-johnson: Yep, we are temporarily hard coding the releases to get this functionality into the next release. But we will be retrieving the versions from the Cincinnati stable channel at a later stage

cmd/aro/update_ocp_versions.go
ellis-johnson: Done

cmd/aro/update_ocp_versions.go
ellis-johnson: This is duplicated from the other commands, so I think it will need to be fixed everywhere, but this is currently blocking a lot of work that needs to be done so I think it might be best if we forgo this for now

hawkowl (approved):

darthhexx (approved):

PR #2335, opened 2022-08-18
List resource improvements

facchettos:

https://docs.google.com/document/d/1Ie-ZWsm3DD94w-fCIAl9XaIjiHgacFWXKamgVrpLNbw/edit?usp=sharing


pkg/frontend/adminactions/resources_list.go
```diff
@@ -16,24 +17,39 @@ import (
...
 		a.log.Warnf("error when getting DiskEncryptionSet resources: %s", err)
 	}
+	hasWritten := false
+
+	for _, resource := range armResources {
+		if hasWritten {
+			writer.Write(comaAsByteArray)
+		}
+		bytes, _ := resource.MarshalJSON()
+		writer.Write(bytes)
```
cadenmarchese, line 50: Could we leverage your `writeObject` function here as well?

pkg/frontend/adminactions/resources_list.go
```diff
@@ -44,35 +60,48 @@ func (a *azureActions) ResourcesList(ctx context.Context) ([]byte, error) {
...
 		}
 	}
 
-	return json.Marshal(armResources)
+	_, err = writer.Write([]byte("]"))
+
+	return err
+}
+
+func writeObject(writer io.Writer, resource arm.Resource) {
```
cadenmarchese, line 102: We could probably return the error from `MarshallJSON` here just in case a single resource fails to marshall.

Question for my own knowledge - is the individual marshall of each resource required, or could we skip it and pass resources directly to `writer.Write`?

cadenmarchese (commented):
Thanks again for your time on this, I've added you to the associated ADR as an author. I haven't tested anything yet, but I can already see how we avoid writing the resources to memory as a single large object.

@m1kola I would be curious as to your thoughts on this approach.

pkg/frontend/adminactions/resources_list.go
```diff
@@ -16,24 +17,39 @@ import (
...
+var comaAsByteArray = []byte(",")
+
+func (a *azureActions) GroupResourceList(ctx context.Context) ([]mgmtfeatures.GenericResourceExpanded, error) {
 	clusterRGName := stringutils.LastTokenByte(a.oc.Properties.ClusterProfile.ResourceGroupID, '/')
 
-	resources, err := a.resources.ListByResourceGroup(ctx, clusterRGName, "resourceType ne 'Microsoft.Compute/snapshots'", "", nil)
-	if err != nil {
-		return nil, err
-	}
+	return a.resources.ListByResourceGroup(ctx, clusterRGName, "resourceType ne 'Microsoft.Compute/snapshots'", "", nil)
```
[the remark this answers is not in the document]
facchettos: I didn't know what it was for so I preferred not to touch it, will remove it if you think it's better !

pkg/frontend/adminactions/resources_list.go
facchettos: good catch, will fix that tomorrow

pkg/frontend/adminactions/resources_list.go
facchettos: then we would log it but not fail as it was before ? Don't see any problem with that. 

yes, the marshaling is required, writer.Write() only accepts `[]byte` , and also go doesn't make any assumption on how you want to convert it

facchettos: then we would log it but not fail as it was before ? Don't see any problem with that.

yes, the marshaling is required, writer.Write() only accepts []byte , and also go doesn't make any assumption on how you want to convert it

facchettos:
result of the benchmarks: 
```➜  adminactions git:(list-resource-improvements) ✗ go test -bench=. -run=XXX
goos: linux
goarch: amd64
pkg: github.com/Azure/ARO-RP/pkg/frontend/adminactions
cpu: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
BenchmarkResourcesList10-12           	    1699	    695702 ns/op	  348714 B/op	    3071 allocs/op
BenchmarkResourcesList100-12          	     180	   6661380 ns/op	 3387230 B/op	   30017 allocs/op
BenchmarkResourcesList500-12          	      33	  32914196 ns/op	16837106 B/op	  149766 allocs/op
BenchmarkResourcesList1000-12         	      16	  65778108 ns/op	33618915 B/op	  299420 allocs/op
BenchmarkResourcesList10000-12        	       2	 650945102 ns/op	335515996 B/op	 2992800 allocs/op
BenchmarkResourcesList10old-12        	    1672	    657954 ns/op	  346403 B/op	    2890 allocs/op
BenchmarkResourcesList100old-12       	     192	   6384128 ns/op	 3377989 B/op	   28221 allocs/op
BenchmarkResourcesList500old-12       	      39	  30884575 ns/op	16782081 B/op	  140736 allocs/op
BenchmarkResourcesList1000old-12      	      19	  59550317 ns/op	33478402 B/op	  281352 allocs/op
BenchmarkResourcesList10000old-12     	       2	 587785286 ns/op	335090472 B/op	 2812235 allocs/op
BenchmarkResourcesList10Jeff-12       	    1719	    668218 ns/op	  346942 B/op	    2891 allocs/op
BenchmarkResourcesList100Jeff-12      	     189	   6383845 ns/op	 3378893 B/op	   28222 allocs/op
BenchmarkResourcesList500Jeff-12      	      34	  30648379 ns/op	16789058 B/op	  140740 allocs/op
BenchmarkResourcesList1000Jeff-12     	      18	  60727391 ns/op	33483085 B/op	  281393 allocs/op
BenchmarkResourcesList10000Jeff-12    	       2	 586616238 ns/op	332901972 B/op	 2812075 allocs/op```

The memory allocations and speed are similar in all functions. only difference is that in the streaming approach , the bytes can be discarded as soon as they are written to the writer.

**Review by petrkotas — DISMISSED** [review]
LGTM, with nonblocking nit that should be fixed afterwards

File: pkg/frontend/adminactions/resources_list.go [hunk 4]
```diff
@@ -16,24 +17,38 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/subnet"
 )
 
-func (a *azureActions) ResourcesList(ctx context.Context) ([]byte, error) {
+var comaAsByteArray = []byte(",")
```
s-amann, line 20: nit: typo for future PR

pkg/frontend/adminactions/resources_list.go
```diff
@@ -44,35 +59,55 @@ func (a *azureActions) ResourcesList(ctx context.Context) ([]byte, error) {
...
+	return err
+}
+
+func (a *azureActions) writeObject(writer io.Writer, resource arm.Resource) {
+	bytes, err := resource.MarshalJSON()
+	if err != nil {
+		a.log.Warn(err) //very unlikely , only a handful of cases trigger an error
+		//here. and since we get the object from a database , it probably will never happen
+	}
+	_, err = writer.Write(bytes)
```
s-amann, line 107: if the unlikely error happens above during marshalling should we still attempt to write the bytes?

petrkotas: Good catch @s-amann. When I look at the flow again, I would tend to say we keep the flow running. Partial result is better than no result.

s-amann: I agree it would be best not to error out, but the bytes could be empty, nil, or unusable data. My take would be to invert the condition and only write when successful, otherwise log a warning.

cadenmarchese: @s-amann @petrkotas Jeremy gave me push access to his fork and I was able to make the changes. Let me know how it looks.

pkg/frontend/adminactions/resources_list.go
```diff
@@ -44,35 +59,56 @@ func (a *azureActions) ResourcesList(ctx context.Context) ([]byte, error) {
...
+
+	return err
+}
+
+func (a *azureActions) writeObject(writer io.Writer, resource arm.Resource) {
+	bytes, err := resource.MarshalJSON()
+	if err != nil {
+		a.log.Warn(err) //very unlikely , only a handful of cases trigger an error
+		//here. and since we get the object from a database , it probably will never happen
+	} else {
```
s-amann, line 106: imo this is fine to use else, but we could reduce nesting by adding a return after logging on line 104 and then moving the code in the else block outside of the err check condition. 

```go
bytes, err := resource.MarshalJSON()
if err != nil {
  a.log.Warn(err)  // comment
  return
}

_, err = writer.Write(bytes)
if err != nil {
  a.log.Warn(err) //comment
}

s-amann (approved):
lgtm, left one suggestion

cadenmarchese (approved):

s-amann (approved):

PR #2337, opened 2022-08-19
Add ARO ingress controller

ArrisLee:

https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/10116740

If customer scales down IngressController to zero replica, they will lose access to their cluster (probably even to API Server as OAuth will use ingress). 
ARO Operator should have a controller to ensure the replica is set to at least 2 in case it is set by customer to 0 (on purpose or by accident)

### Test plan for issue:

- Unit test
- Manual test via local RP

N/A


facchettos:
doesn't that conflict with guardrails ?

ArrisLee:
> doesn't that conflict with guardrails ?

not quite sure, will check this with Jeff

ArrisLee:
> doesn't that conflict with guardrails ?

@facchettos just checked with Miguel, seems the guardrails project is paused for now. So remove the hold tag.

pkg/operator/controllers/checker/ingressreplica_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+	arofake "github.com/Azure/ARO-RP/pkg/operator/clientset/versioned/fake"
+)
+
+func TestIngressReplicaChecker(t *testing.T) {
+	ctx := context.Background()
+
+	var defaultReplica int32 = 2
+	var zeroReplica int32 = 0
```
cadenmarchese, line 22: These vars probably aren't needed for the test, I think you could get away with just using the int values directly in the tests since you are only using each var once.

pkg/operator/controllers/checker/ingressreplica_test.go
```diff
@@ -0,0 +1,127 @@
...
+			expectedState: operatorv1.OperatorCondition{
+				Type:    arov1alpha1.DefaultIngressReplica,
+				Status:  operatorv1.ConditionTrue,
+				Message: "Default replicas in place",
+				Reason:  "CheckDone",
+			},
+			expectedReplica: 2,
+		},
+		{
+			name: "ingress controller has zero replicas",
```
cadenmarchese, line 56: Some additional test case ideas:
-`ingress controller has greater than default replicas`
-`ingress controller replicas could not be read`
-`ingress controller replicas failed to update`
I didn't check code coverage of the current tests, but these were a few that came to mind.

pkg/operator/controllers/checker/ingressreplica_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+			if *ingress.Spec.Replicas != tt.expectedReplica {
+				t.Errorf("wrong replica count, want: %d, got: %d", tt.expectedReplica, *ingress.Spec.Replicas)
+			}
+
+			cluster, err := arocli.AroV1alpha1().Clusters().Get(ctx, arov1alpha1.SingletonClusterName, metav1.GetOptions{})
+			if err != nil {
+				t.Error(err)
+			}
+
```
dem4gus, line 112: Add debug info for the condition states on all of these:
```suggestion
receivedCondition := cluster.Status.Conditions[0]
if receivedCondition.Message != tt.expectedState.Message {
    t.Errorf("expected message `%v` but got `%v`", receivedCondition.Message, tt.expectedState.Message)
}
```

dem4gus (dismissed):
Looks good so far. Are there plans to add e2e tests for this operator?

pkg/operator/controllers/checker/ingressreplica.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (r *IngressReplicaChecker) Check(ctx context.Context) error {
+	cond := &operatorv1.OperatorCondition{
+		Type:    arov1alpha1.DefaultIngressReplica,
+		Status:  operatorv1.ConditionTrue,
+		Message: "Default replicas in place",
+		Reason:  "CheckDone",
+	}
+
+	ingress, err := r.operatorcli.OperatorV1().IngressControllers("openshift-ingress-operator").Get(ctx, "default", metav1.GetOptions{})
```
bennerv, line 53: can you pull `openshift-ingress-operator` out into a const at the top?

pkg/operator/controllers/checker/ingressreplica.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		cond.Message = err.Error()
+		cond.Reason = "CheckFailed"
+		cond.LastTransitionTime = metav1.NewTime(time.Now().UTC())
+		return conditions.SetCondition(ctx, r.arocli, cond, r.role)
+	}
+
+	if ingress.Spec.Replicas != nil && *ingress.Spec.Replicas < 1 {
+		var minimumReplica int32 = 1
```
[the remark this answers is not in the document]
bennerv: +1 - default in OCP is 2 anyway.  Plus it would make it resilient to machine updates happening.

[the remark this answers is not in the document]
bennerv: Also let's pull the minimum replica count as a const at the top of the file as well.

pkg/operator/controllers/checker/ingressreplica.go
```diff
@@ -0,0 +1,77 @@
...
+		ingress.Spec.Replicas = &minimumReplica
```
bennerv, line 63: nit:
```suggestion
		ingress.Spec.Replicas = to.Int32Ptr(minReplicaCount)

bennerv (dismissed):
One thing: this isn't actually a "checker".  The original design of the checker was to report the status of things, not to make modifications to the cluster.  

Let's pull this out of the checker package and move it into it's own controller.  Since we're making modifications and then we can watch on the default IngressController resource.

facchettos (changes requested):
I am struggling to understand the architecture here. Why are we not preventing them to set the replicas to 0? Here it would seem that a validating webhook is the better choice (either via guardrail or a plain admission hook)

ArrisLee:
updates: refactor the ingress replica checker into aro ingress controller.

hawkowl (approved):
Looks good to me.

hawkowl:
@ArrisLee some E2E tests might be good for this, but I think it would be a bit of a pain to test, so, I'm fine with this.

PR #2300, opened 2022-08-02
CI E2E in SimplySecure

karanmagdani:
Signed-off-by: Karan <kmagdani@redhat.com>

### Which issue this PR addresses:

Fixes https://dev.azure.com/msazure/AzureRedHatOpenShift/_boards/board/t/Chainsaw/Stories/?workitem=15123283

### What this PR does / why we need it:

This PR enables VPN support on our CI-E2E Pipelines. The pipeline connects to a vpn before creating private clusters and  running end to end tests. This will facilitate running pipelines when SimplySecure V2 is enabled on our subscriptions

PR #2338, opened 2022-08-19
Add vpn code run before e2e to connect to the aks

petrkotas:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15325401

Add a vpn connection for e2e runs. The connection is configured with the `$VPN` environment variable. 
The variable can be configured in the environment as a pipeline variable.

The current `$VPN` should be set to `vpn-aks-eastus.ovpn`. 
It will enable direct access to the AKS subnet to allow RP to communicate with AKS API.

IMPORTANT: The variable needs to be set up in the pipelines!!!

### DIFF
--- a/.pipelines/e2e.yml
+++ b/.pipelines/e2e.yml
@@ -43,7 +43,9 @@ jobs:
 
       set -x
       . ./hack/e2e/run-rp-and-e2e.sh
-      trap 'set +e; kill_rp; clean_e2e_db' EXIT
+      trap 'set +e; kill_rp; clean_e2e_db; kill_vpn' EXIT
+
+      run_vpn
 
       deploy_e2e_db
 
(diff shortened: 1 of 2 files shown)


bennerv (approved):
lgtm.

@karanmagdani this will have to be pulled out and replaced with the correct VPN single client once we integrate simply securev2 work.

karanmagdani:
@bennerv @petrkotas I was wondering why we need this. If we get #2300 merged then this is not needed anymore.  I know that SimplySecure Rollout is paused from MSFT side, but our solution in 2300 works in both SimplySecure and non-SimplySecure environment

bennerv:
> @bennerv @petrkotas I was wondering why we need this. If we get #2300 merged then this is not needed anymore. I know that SimplySecure Rollout is paused from MSFT side, but our solution in 2300 works in both SimplySecure and non-SimplySecure environment

This is stop-gap solution until #2300 is merged.  if it's ready and lands before this we can close this.

petrkotas:
@bennerv @karanmagdani thanks, merging.

PR #2340, opened 2022-08-19
make ImageConfig environment aware

dem4gus:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_boards/board/t/Chainsaw/Stories/?workitem=15120407

Makes the ImageConfig controller aware of running in INT/non-INT environments.

### Test plan for issue:
Unit tests created or updated as needed, e2e tests updated.

None

### DIFF
--- a/pkg/operator/controllers/imageconfig/image_controller.go
+++ b/pkg/operator/controllers/imageconfig/image_controller.go
@@ -119,13 +119,13 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // Switch case to ensure the correct registries are added depending on the cloud environment (Gov or Public cloud)
 func getCloudAwareRegistries(instance *arov1alpha1.Cluster) ([]string, error) {
-	const replicationRegistryFormat = "%s.%s.data.%s"
 	var replicationRegistry string
 	var dnsSuffix string
 
-	acrSubdomain := strings.Split(instance.Spec.ACRDomain, ".")[0]
-	if len(acrSubdomain) == 0 {
-		return nil, errors.New("azure container registry domain is not present or malformed")
+	acrDomain := instance.Spec.ACRDomain
+	acrSubdomain := strings.Split(acrDomain, ".")[0]
+	if acrDomain == "" || acrSubdomain == "" {
+		return nil, fmt.Errorf("azure container registry domain is not present or is malformed")
 	}
 
 	switch instance.Spec.AZEnvironment {
@@ -136,8 +136,8 @@ func getCloudAwareRegistries(instance *arov1alpha1.Cluster) ([]string, error) {
 	default:
 		return nil, fmt.Errorf("cloud environment %s is not supported", instance.Spec.AZEnvironment)
 	}
-	replicationRegistry = fmt.Sprintf(replicationRegistryFormat, acrSubdomain, instance.Spec.Location, dnsSuffix)
-	return []string{instance.Spec.ACRDomain, replicationRegistry}, nil
+	replicationRegistry = fmt.Sprintf("%s.%s.data.%s", acrSubdomain, instance.Spec.Location, dnsSuffix)
+	return []string{acrDomain, replicationRegistry}, nil
 }
 
 // Helper function that filters registries to make sure they are added in consistent order
--- a/pkg/operator/controllers/imageconfig/image_controller.go
+++ b/pkg/operator/controllers/imageconfig/image_controller.go
@@ -119,20 +119,25 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // Switch case to ensure the correct registries are added depending on the cloud environment (Gov or Public cloud)
 func getCloudAwareRegistries(instance *arov1alpha1.Cluster) ([]string, error) {
+	const replicationRegistryFormat = "%s.%s.data.%s"
+	var replicationRegistry string
+	var dnsSuffix string
+
 	acrSubdomain := strings.Split(instance.Spec.ACRDomain, ".")[0]
-	var regionalRegistry string
+	if len(acrSubdomain) == 0 {
+		return nil, errors.New("azure container registry domain is not present or malformed")
+	}
 
 	switch instance.Spec.AZEnvironment {
 	case azureclient.PublicCloud.Environment.Name:
-		regionalRegistry = fmt.Sprintf("%s.%s.data.%s", acrSubdomain, instance.Spec.Location, azure.PublicCloud.ContainerRegistryDNSSuffix)
-
+		dnsSuffix = azure.PublicCloud.ContainerRegistryDNSSuffix
 	case azureclient.USGovernmentCloud.Environment.Name:
-		regionalRegistry = fmt.Sprintf("%s.%s.data.%s", acrSubdomain, instance.Spec.Location, azure.USGovernmentCloud.ContainerRegistryDNSSuffix)
-
+		dnsSuffix = azure.USGovernmentCloud.ContainerRegistryDNSSuffix
 	default:
 		return nil, fmt.Errorf("cloud environment %s is not supported", instance.Spec.AZEnvironment)
 	}
-	return []string{instance.Spec.ACRDomain, regionalRegistry}, nil
+	replicationRegistry = fmt.Sprintf(replicationRegistryFormat, acrSubdomain, instance.Spec.Location, dnsSuffix)
+	return []string{instance.Spec.ACRDomain, replicationRegistry}, nil
 }
 
 // Helper function that filters registries to make sure they are added in consistent order
--- a/pkg/operator/controllers/imageconfig/image_controller.go
+++ b/pkg/operator/controllers/imageconfig/image_controller.go
@@ -62,7 +62,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	}
 
 	// Check for cloud type
-	requiredRegistries, err := getCloudAwareRegistries(instance)
+	requiredRegistries, err := GetCloudAwareRegistries(instance)
 	if err != nil {
 		// Not returning error as it will requeue again
 		return reconcile.Result{}, nil
@@ -118,7 +118,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 // Switch case to ensure the correct registries are added depending on the cloud environment (Gov or Public cloud)
-func getCloudAwareRegistries(instance *arov1alpha1.Cluster) ([]string, error) {
+func GetCloudAwareRegistries(instance *arov1alpha1.Cluster) ([]string, error) {
 	var replicationRegistry string
 	var dnsSuffix string
 
--- a/pkg/operator/controllers/imageconfig/image_controller.go
+++ b/pkg/operator/controllers/imageconfig/image_controller.go
@@ -119,18 +119,20 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // Switch case to ensure the correct registries are added depending on the cloud environment (Gov or Public cloud)
 func getCloudAwareRegistries(instance *arov1alpha1.Cluster) ([]string, error) {
-	var requiredRegistries []string
+	acrSubdomain := strings.Split(instance.Spec.ACRDomain, ".")[0]
+	var regionalRegistry string
+
 	switch instance.Spec.AZEnvironment {
 	case azureclient.PublicCloud.Environment.Name:
-		requiredRegistries = []string{instance.Spec.ACRDomain, "arosvc." + instance.Spec.Location + ".data." + azure.PublicCloud.ContainerRegistryDNSSuffix}
+		regionalRegistry = fmt.Sprintf("%s.%s.data.%s", acrSubdomain, instance.Spec.Location, azure.PublicCloud.ContainerRegistryDNSSuffix)
 
 	case azureclient.USGovernmentCloud.Environment.Name:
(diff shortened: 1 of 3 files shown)


test/e2e/operator.go
```diff
@@ -436,18 +436,20 @@ var _ = Describe("ARO Operator - ImageConfig Reconciler", func() {
 
 	// Reimplementation of function from image config controller
 	getCloudAwareRegistries := func(instance *arov1alpha1.Cluster) ([]string, error) {
-		var requiredRegistries []string
+		acrSubdomain := strings.Split(instance.Spec.ACRDomain, ".")[0]
```
petrkotas, line 439: I am not happy about this re-implementation, but this is work for a separate task.

petrkotas (approved):
Thank you @dem4gus good work.

petrkotas (dismissed):
Pardon, I found one think I realized afterwards. Small question.

test/e2e/operator.go
dem4gus: Agreed.

pkg/operator/controllers/imageconfig/image_controller.go
```diff
@@ -119,18 +119,20 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // Switch case to ensure the correct registries are added depending on the cloud environment (Gov or Public cloud)
 func getCloudAwareRegistries(instance *arov1alpha1.Cluster) ([]string, error) {
-	var requiredRegistries []string
+	acrSubdomain := strings.Split(instance.Spec.ACRDomain, ".")[0]
```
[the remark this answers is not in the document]
dem4gus: This is set (for prod) in [pkg/env/prod.go](https://github.com/Azure/ARO-RP/blob/fc656bc7a8feb251227a170186c9ddbb90c5c064/pkg/env/prod.go#L175-L185) by querying the Azure endpoint and building the domain name. This is a quick way of getting `acrResource.ResourceName` without having to re-query the Azure API.

s-amann (dismissed):

pkg/operator/controllers/imageconfig/image_controller.go
```diff
@@ -119,18 +119,20 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	var regionalRegistry string
+
 	switch instance.Spec.AZEnvironment {
 	case azureclient.PublicCloud.Environment.Name:
-		requiredRegistries = []string{instance.Spec.ACRDomain, "arosvc." + instance.Spec.Location + ".data." + azure.PublicCloud.ContainerRegistryDNSSuffix}
+		regionalRegistry = fmt.Sprintf("%s.%s.data.%s", acrSubdomain, instance.Spec.Location, azure.PublicCloud.ContainerRegistryDNSSuffix)
```
bennerv, line 127: technically we don't have ACR replication in all regions and environments, but to make this easy (and it has no additional impact including this) I'm +1 with this.

bennerv (approved):
lgtm

pkg/operator/controllers/imageconfig/image_controller.go
```diff
@@ -119,18 +119,20 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
 	case azureclient.USGovernmentCloud.Environment.Name:
-		requiredRegistries = []string{instance.Spec.ACRDomain, "arosvc." + instance.Spec.Location + ".data." + azure.USGovernmentCloud.ContainerRegistryDNSSuffix}
+		regionalRegistry = fmt.Sprintf("%s.%s.data.%s", acrSubdomain, instance.Spec.Location, azure.USGovernmentCloud.ContainerRegistryDNSSuffix)
 
 	default:
 		return nil, fmt.Errorf("cloud environment %s is not supported", instance.Spec.AZEnvironment)
 	}
-	return requiredRegistries, nil
+	return []string{instance.Spec.ACRDomain, regionalRegistry}, nil
```
cblecker, line 135: Using format strings like this can be fragile, but perhaps we can DRY it out a bit to make it a bit more robust.

e.g.
```golang
var regionalRegistrySuffix string
switch instance.Spec.AZEnvironment {
case azureclient.PublicCloud.Environment.Name:
	regionalRegistrySuffix = azure.PublicCloud.ContainerRegistryDNSSuffix
 [snip -- repeat for USGovernmentCloud and maintain default case)

// only have this format string once, because the format string is the fragile piece
regionalRegistry := fmt.Sprintf("%s.%s.data.%s", acrSubdomain, instance.Spec.Location, regionalRegistrySuffix)

return []string{instance.Spec.ACRDomain, regionalRegistry}, nil
```

pkg/operator/controllers/imageconfig/image_controller.go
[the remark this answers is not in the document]
cblecker: Perhaps a quick error check that the resultant string is of a minimum length? In future, a more robust check would be ensuring the string conforms to a DNS label spec.

test/e2e/operator.go
```diff
@@ -436,18 +436,20 @@ var _ = Describe("ARO Operator - ImageConfig Reconciler", func() {
...
+		var regionalRegistry string
+
 		switch instance.Spec.AZEnvironment {
 		case azureclient.PublicCloud.Environment.Name:
-			requiredRegistries = []string{instance.Spec.ACRDomain, "arosvc." + instance.Spec.Location + ".data." + azure.PublicCloud.ContainerRegistryDNSSuffix}
+			regionalRegistry = fmt.Sprintf("%s.%s.data.%s", acrSubdomain, instance.Spec.Location, azure.PublicCloud.ContainerRegistryDNSSuffix)
 
 		case azureclient.USGovernmentCloud.Environment.Name:
-			requiredRegistries = []string{instance.Spec.ACRDomain, "arosvc." + instance.Spec.Location + ".data." + azure.USGovernmentCloud.ContainerRegistryDNSSuffix}
+			regionalRegistry = fmt.Sprintf("%s.%s.data.%s", acrSubdomain, instance.Spec.Location, azure.USGovernmentCloud.ContainerRegistryDNSSuffix)
```
SudoBrendan, line 447: Just another call for a DRY implementation of this format string (see above comments). At this point, I'm wondering if it would be good to have it as a `const`, so you can import it here for consistency.

SudoBrendan (approved):
Overall LGTM, you may consider pulling the format string out and making the code/tests a little DRYer

pkg/operator/controllers/imageconfig/image_controller.go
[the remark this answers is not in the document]
dem4gus: Addressed in changes. I also included a quick check for `instance.Spec.ACRDomain` that we were already returning as part of the required registries since that's where any problem would originate from.

pkg/operator/controllers/imageconfig/image_controller.go
dem4gus: Do we have any way of querying which environments have replication and which don't? If we do that could be included in the logic of the this function.

pkg/operator/controllers/imageconfig/image_controller.go
dem4gus: I couldn't find any information about fragility of format strings in Go (if you have any good resources I wouldn't mind reading about it), but I did change this to follow the [Uber style guide](https://github.com/uber-go/guide/blob/master/style.md#format-strings-outside-printf) recommendation to use a const as the string template.

pkg/operator/controllers/imageconfig/image_controller.go
bennerv: no.  it's defined within the RP config which isn't present when we deploy the controllers into a cluster.  I think it's fine.

pkg/operator/controllers/imageconfig/image_controller.go
```diff
@@ -119,18 +119,26 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 // Switch case to ensure the correct registries are added depending on the cloud environment (Gov or Public cloud)
 func getCloudAwareRegistries(instance *arov1alpha1.Cluster) ([]string, error) {
-	var requiredRegistries []string
+	const replicationRegistryFormat = "%s.%s.data.%s"
+	var replicationRegistry string
+	var dnsSuffix string
+
+	acrDomain := instance.Spec.ACRDomain
+	acrSubdomain := strings.Split(acrDomain, ".")[0]
+	if len(acrDomain) == 0 || len(acrSubdomain) == 0 {
```
bennerv, line 128: nit: comparing a string to an empty string is a bit more idomatic than checking it's length.  

```suggestion
	if acrDomain == "" || len(acrSubdomain) == 0 {
```

pkg/operator/controllers/imageconfig/image_controller.go
```diff
@@ -119,18 +119,26 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // Switch case to ensure the correct registries are added depending on the cloud environment (Gov or Public cloud)
 func getCloudAwareRegistries(instance *arov1alpha1.Cluster) ([]string, error) {
-	var requiredRegistries []string
+	const replicationRegistryFormat = "%s.%s.data.%s"
```
bennerv, line 122: I'm not really a fan of this approach.  Would like to get other's thoughts but it's a bit more verbose using `fmt.Printf("%s.%s.data.%s", x, y, z)`

It's better though since we're defining it in one place using it in multiple though.  +1 on the DRY

test/e2e/operator.go
bennerv: +1 - dry this out

bennerv (changes requested):
very close - great work so far

pkg/operator/controllers/imageconfig/image_controller.go
dem4gus: It seems we can still have the string inline, Uber's style guide recommendations seem to be more along the lines of "_if_ you're using a separately-declared format string, make it a const so `go vet` can do its thing".

Really I just wish Go had template strings.

test/e2e/operator.go
dem4gus: This entire function is a re-implementation of the function in image_controller.go in order to get the correct registries for the e2e test. I'd like to _not_ have to copy/paste the entire function, but making the original exportable solely for the purposes of the e2e test didn't seem appropriate. Let me know your thoughts.

test/e2e/operator.go
```diff
@@ -436,18 +436,26 @@ var _ = Describe("ARO Operator - ImageConfig Reconciler", func() {
 
 	// Reimplementation of function from image config controller
 	getCloudAwareRegistries := func(instance *arov1alpha1.Cluster) ([]string, error) {
-		var requiredRegistries []string
+		const replicationRegistryFormat = "%s.%s.data.%s"
```
cadenmarchese, line 439: Personally I am good with using this format string in-line with your `replicationRegistry = fmt.Sprintf(...`. Declaring `replicationRegistryFormat` here and then only using it once in the function might be unnecessary.

pkg/operator/controllers/imageconfig/image_controller.go
cadenmarchese: > It's better though since we're defining it in one place using it in multiple

From the changes in this PR it looks like we're defining `const replicationRegistryFormat = "%s.%s.data.%s"` twice (once here, once in the test) and subsequently only using it once in each place, so I don't think having it in-line would be too repetitive.

test/e2e/operator.go
```diff
@@ -436,18 +436,26 @@ var _ = Describe("ARO Operator - ImageConfig Reconciler", func() {
 
 	// Reimplementation of function from image config controller
 	getCloudAwareRegistries := func(instance *arov1alpha1.Cluster) ([]string, error) {
```
cadenmarchese, line 438: Do we need this function reimplemented or is it possible to import the one from `pkg/operator`?

cadenmarchese (changes requested):
I have a few additional suggestions, thanks for your work so far on this.

test/e2e/operator.go
dem4gus: It's a good thought but there are a couple problems with it:

1. The function is in the `imageconfig` controller which would need to be imported here, I'd like to keep the e2e test as separate from the controller as possible to ensure it's truly e2e.

1. The `getCloudAwareRegistries` function isn't exported by the controller and changing that for the purposes of the test didn't seem like a good solution either.

To me the problem is there's no one place to go to get the required ACR endpoints for a cluster. There used to just be "arosvc.azurecr.io" but that's grown to include USGovCloud ("arosvc.azurecr.us"), replication registries ("arosvc.<region>.data.azurecr.io"), and separate INT environments for each ("arosvc" → "arointsvc"). A better solution would be creating a way for the RP to tell the cluster which registries are required on creation, but we don't currently have that and it would require more time and effort to do than just copying and pasting this code for now. But it is something we _should_ get on our backlog now before we get even more possibilities to account for and these functions balloon even more.

cblecker: > The `getCloudAwareRegistries` function isn't exported by the controller and changing that for the purposes of the test didn't seem like a good solution either

Any particular reason why it can't be exported?

dem4gus: Not a technical one. I was hesitant to do export it just for the purposes of the test but given the feedback it seems others don't share my concerns, which is fine.

I still think a better long-term solution would be for a cluster to be configured with all its required registries by the RP at install time. We've been doing that with `arosvc.azurecr.io` when it was the only ACR we were concerned with, but the platform has grown.

s-amann (approved):
lgtm, nice work

cadenmarchese (approved):
Nice, thanks!

cblecker (approved):

ross-bryan (approved):
lgtm

PR #2341, opened 2022-08-22
Deprecation Fix: Replace strings.Title with Caser

SudoBrendan:

One of [100 things on the way to 4.11](https://github.com/Azure/ARO-RP/pull/2339)...

[Golang 1.18 deprecates the `strings.Title` function](https://github.com/golang/go/issues/48367), but a fix is available before we bump golang versions so I'm doing that here in it's own PR.

### Test plan for issue:

`make generate` still returns the same swagger files with identical filename casing (e.g. "openShiftClusters" still gets transformed to "OpenShiftClusters"), this is a no-op refactor.

N/A


bennerv (approved):
lgtm - following recommendation of the golang devs themselves.  +1

SudoBrendan:
> lgtm - following recommendation of the golang devs themselves. +1

The real win, is that now we can FINALLY rename our service to ":rocket:OpenShift" and not worry about those darn unicode capitalization issues! :sunglasses: \s

petrkotas (approved):
@SudoBrendan Perfect, thank you!

PR #2343, opened 2022-08-23
Fix OpenShiftCcluster PUTs causing static validation errors due to ARM metadata

hawkowl:
Fixes a bug in production.

When a PUT is done, copy over the SystemData so that it does not fail static validation. This would cause PUTs to fail on API v2022-04-01 and above.

### Test plan for issue:

Has unit tests.

No, bug fix

### DIFF
--- a/pkg/frontend/openshiftcluster_putorpatch.go
+++ b/pkg/frontend/openshiftcluster_putorpatch.go
@@ -138,6 +138,7 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 					ClientSecret: doc.OpenShiftCluster.Properties.ServicePrincipalProfile.ClientSecret,
 				},
 			},
+			SystemData: doc.OpenShiftCluster.SystemData,
 		})
 
 		// In case of PATCH we take current cluster document, which is enriched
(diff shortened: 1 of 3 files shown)


petrkotas (approved):
Hi @hawkowl thanks for spotting this one. Out of a curiosity, where does it fail? I am not able to find `SystemData` validation in static checks. I believe I am missing something. Thank you for the information in advance. In the meantime LGTM

s-amann (approved):

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -138,6 +138,7 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 					ClientSecret: doc.OpenShiftCluster.Properties.ServicePrincipalProfile.ClientSecret,
 				},
 			},
+			SystemData: doc.OpenShiftCluster.SystemData,
```
jharrington22, line 141: So we weren't sending the SystemData that originates ARM?

jharrington22 (commented):
Left a comment to clarify my understanding but otherwise lgtm

PR #2344, opened 2022-08-23
Refactor pkg/portal to use go:embed instead of go-bindata

xiaoyu74:

[Work Item: 14891926](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14891926)

This PR removes the use of go-bindata within pkg/portal and uses go:embed instead.

### Test plan for issue:
There are a handful of unit tests that indirectly validate this change

We have had updates within `docs/`


pkg/portal/portal.go
```diff
@@ -261,7 +262,8 @@ func (p *portal) unauthenticatedRoutes(r *mux.Router) {
 }
 
 func (p *portal) aadAuthenticatedRoutes(r *mux.Router) {
-	for _, name := range AssetNames() {
+	names, _ := assets.EmbeddedFiles.ReadFile("v1/portalfiles.txt")
```
hawkowl, line 265: Should we use https://pkg.go.dev/embed#FS.ReadDir instead? It might be a hassle to keep portalfiles updated.

xiaoyu74: @hawkowl - Thanks Amber for reviewing and suggestion, I will check to see if `FS.ReadDir` can read the files from the different sub-folders respectively.

cblecker:
Removing ready for review pending rebase and addressing review comments

pkg/portal/portal.go
```diff
@@ -261,21 +265,41 @@ func (p *portal) unauthenticatedRoutes(r *mux.Router) {
...
-			r.NewRoute().Methods(http.MethodGet).Path("/").HandlerFunc(p.index)
-			continue
-		}
+	var names []string
 
-		if name == "v2/build/index.html" {
-			r.NewRoute().Methods(http.MethodGet).Path("/v2").HandlerFunc(p.indexV2)
-			continue
+	err := filepath.Walk(".", func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
```
cblecker, line 271: This error check doesn't fully make sense to me.

`Walk` runs this function on ever file it identifies, right? So nothing in here should have errored yet if I understand correctly.

pkg/portal/portal.go
```diff
@@ -261,21 +265,41 @@ func (p *portal) unauthenticatedRoutes(r *mux.Router) {
...
+			log.Fatal(err)
+			return nil
 		}
 
-		fmtName := strings.TrimPrefix(name, "v1/build/")
-		fmtName = strings.TrimPrefix(fmtName, "v2/build/")
+		names = append(names, path)
+		return nil
+	})
+
```
cblecker, line 279:
```suggestion
```
Makes it clearer that this error check is related to the above Walk

cblecker (dismissed):

pkg/portal/portal.go
xiaoyu74: @cblecker - Thanks for your review feedback. 

`Walk` just walks the file tree and calling `WalkFunc` for each file or directory in the tree. All errors that arise visiting files and directories are filtered by `WalkFunc`. I just use the standard error handling of this function to handle the possible failures accessing a path. 

This is an example [example of using Walk](https://pkg.go.dev/path/filepath#Walk) I referred. Please let me know if anything I misunderstood from you.

facchettos: the error is given by walk to the function. We have to check it and decide how we handle errors and if we should stop the whole processing or just skip the directory etc. Maybe crashing the portal here isn't optimal though? I'd rather keep what they have in the example, which will stop the walking and we can handle the error in the following block.

facchettos: I would also prefer if we took out the function into a real function instead of using an anonymous function here.

pkg/portal/portal.go
facchettos: not sure I understand your suggestion here @cblecker

pkg/portal/portal.go
```diff
@@ -261,21 +265,41 @@ func (p *portal) unauthenticatedRoutes(r *mux.Router) {
...
+				continue
+			}
 
-		r.NewRoute().Methods(http.MethodGet).Path("/" + fmtName).HandlerFunc(p.serve(name))
+			fmtName := strings.TrimPrefix(name, "v1/build/")
+			fmtName = strings.TrimPrefix(fmtName, "v2/build/")
+
+			r.NewRoute().Methods(http.MethodGet).Path("/" + fmtName).HandlerFunc(p.serve(name))
+		}
 	}
```
facchettos, line 303: That is a lot of nesting, it would be nicer to have smaller function to break this big block

pkg/portal/portal.go
```diff
@@ -137,12 +141,12 @@ func (p *portal) setupRouter() error {
...
+	assetv1, err := assets.EmbeddedFiles.ReadFile("v1/build/index.html")
 	if err != nil {
 		return err
 	}
 
-	assetv2, err := Asset("v2/build/index.html")
+	assetv2, err := assets.EmbeddedFiles.ReadFile("v2/build/index.html")
 	if err != nil {
 		return err
 	}
```
facchettos, line 152: I think I'd call that `indexV{1,2}` here.  Not your change but the name didn't really make sense before so we might as well change it

pkg/portal/portal.go
```diff
@@ -261,21 +265,41 @@ func (p *portal) unauthenticatedRoutes(r *mux.Router) {
...
+		names = append(names, path)
+		return nil
+	})
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	for _, name := range names {
+		regexp, _ := regexp.Compile(`v[1,2]/build/.*\..*`)
```
facchettos, line 285: could we use `MustCompile` here if we are not checking the error ?

pkg/portal/portal.go
```diff
@@ -261,21 +265,41 @@ func (p *portal) unauthenticatedRoutes(r *mux.Router) {
...
+				r.NewRoute().Methods(http.MethodGet).Path("/v2").HandlerFunc(p.indexV2)
+				continue
+			}
 
-		r.NewRoute().Methods(http.MethodGet).Path("/" + fmtName).HandlerFunc(p.serve(name))
+			fmtName := strings.TrimPrefix(name, "v1/build/")
+			fmtName = strings.TrimPrefix(fmtName, "v2/build/")
+
+			r.NewRoute().Methods(http.MethodGet).Path("/" + fmtName).HandlerFunc(p.serve(name))
+		}
```
facchettos, line 302:
```go
     switch name {
     case "v1/build/index.html":
         r.NewRoute().Methods(http.MethodGet).Path("/").HandlerFunc(p.index)
     case "v2/build/index.html":                                                               
         r.NewRoute().Methods(http.MethodGet).Path("/").HandlerFunc(p.indexV2)
     case "":
         //the regex exp did not match and the name is empty
         break                                                           
     default:                                                    
         fmtName := strings.TrimPrefix(name, "v1/build/")
         fmtName = strings.TrimPrefix(fmtName, "v2/build/")
                                                                 
         r.NewRoute().Methods(http.MethodGet).Path("/" + fmtName).HandlerFunc(p.serve(name))
     }                                                           
```

maybe this would be a little better

facchettos (changes requested):

pkg/portal/portal.go
xiaoyu74: It's the standard use case of the `Walk` function to have an anonymous `WalkFunc` function type called by Walk. May I know the reason it's really needed to take out into a real function?

pkg/portal/portal.go
```diff
@@ -261,21 +265,40 @@ func (p *portal) unauthenticatedRoutes(r *mux.Router) {
...
+				r.NewRoute().Methods(http.MethodGet).Path("/").HandlerFunc(p.indexV2)
+			case "":
+				break
+			default:
+				fmtName := strings.TrimPrefix(name, "v1/build/")
+				fmtName = strings.TrimPrefix(fmtName, "v2/build/")
+
+				r.NewRoute().Methods(http.MethodGet).Path("/" + fmtName).HandlerFunc(p.serve(name))
+			}
+		}
```
facchettos, line 301: you don't need to check `len(name)>0` because you have the `""` case in the switch, this is why I suggested it

pkg/portal/portal.go
```diff
@@ -261,21 +265,40 @@ func (p *portal) unauthenticatedRoutes(r *mux.Router) {
...
-		}
+	var names []string
 
-		if name == "v2/build/index.html" {
-			r.NewRoute().Methods(http.MethodGet).Path("/v2").HandlerFunc(p.indexV2)
-			continue
+	err := filepath.Walk(".", func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			log.Fatal(err)
+			return nil
```
facchettos, line 273:
```suggestion
			return err
```

This will exit and we will exit in the error handling block. it's better to only have one place where we crash the program ()

facchettos (changes requested):

pkg/portal/portal.go
```diff
@@ -261,21 +265,38 @@ func (p *portal) unauthenticatedRoutes(r *mux.Router) {
...
+		regexp, _ := regexp.Compile(`v[1,2]/build/.*\..*`)
+		name := regexp.FindString(name)
+	nameLoop:
+		switch name {
+		case "v1/build/index.html":
+			r.NewRoute().Methods(http.MethodGet).Path("/").HandlerFunc(p.index)
+		case "v2/build/index.html":
+			r.NewRoute().Methods(http.MethodGet).Path("/").HandlerFunc(p.indexV2)
+		case "":
+			break nameLoop
```
facchettos, line 293: why do you break nameloop here? This didn't use to be this logic. please remove it. The break in the `switch` was to be explicitly exiting the `switch` because there was no action performed inside the `case`.

facchettos (changes requested):

pkg/portal/portal.go
xiaoyu74: @facchettos - Thanks for pointing it out.  Actually, I got  `ineffective break statement. Did you mean to break out of the outer loop?` from go-staticcheck in the IDE if put `break` for the case `""` and the golangci-lint will be failed for that. That's why I put a label `nameLoop` there as a workaround. (yeah, the naming is not that clear, sorry) 

Anyway, I think break statement here has no effect and could be removed.
```
for _, name := range names {
		regexp, _ := regexp.Compile(`v[1,2]/build/.*\..*`)
		name := regexp.FindString(name)
		switch name {
		case "v1/build/index.html":
			r.NewRoute().Methods(http.MethodGet).Path("/").HandlerFunc(p.index)
		case "v2/build/index.html":
			r.NewRoute().Methods(http.MethodGet).Path("/").HandlerFunc(p.indexV2)
		case ""
```

facchettos (dismissed):
Let's squash the commits and have a nice commit message and then we can merge it

facchettos (approved):

ellis-johnson:
This merge has broken admin portal v2 functionality

PR #2346, opened 2022-08-23
clean up naming in liveconfig

s-amann:

Fixes comments from https://github.com/Azure/ARO-RP/pull/2325

Addresses comments made from merged PR.

### Test plan for issue: 
N/A

No.

### DIFF
--- a/pkg/util/liveconfig/dev.go
+++ b/pkg/util/liveconfig/dev.go
@@ -13,19 +13,19 @@ import (
 )
 
 const (
-	HIVEENVVARIABLE = "HIVEKUBECONFIGPATH"
+	HIVE_ENV_VARIABLE = "HIVEKUBECONFIGPATH"
 )
 
 func (d *dev) HiveRestConfig(ctx context.Context, index int) (*rest.Config, error) {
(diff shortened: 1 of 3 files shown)


bennerv (approved):
lgtm - just renames

cblecker (approved):

pkg/util/liveconfig/dev.go
```diff
@@ -13,19 +13,19 @@ import (
 )
 
 const (
-	HIVEENVVARIABLE = "HIVEKUBECONFIGPATH"
```
jharrington22, line 16: @s-amann this is only for local dev right?

jharrington22 (commented):
One comment, thanks for fixing these up!

pkg/util/liveconfig/dev.go
s-amann: @jharrington22 yes, this impl of HiveRestConfig will only be available when `isLocalDevelopmentMode` is set to `true`. Otherwise the HiveRestConfig impl in `pkg/util/liveconfig/hive.go` will be used.

jharrington22 (approved):
lgtm

PR #1310, opened 2021-02-12
Ignore Upgradeable/Unknown condition in pushed metrics

gvanderpotte:
### Which issue this PR addresses:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/9326693

### What this PR does / why we need it:

Currently, kube-storage-version-migrator is constantly reported with status Upgradeable/Unknown. This disrupts the readability of dashboards.

### Test plan for issue:

- Are there unit tests? No - but it is in theory possible to test the filtering of ignored. This PR is meant to gather opinions on that possibility of ignoring this status for now
- Are there integration/e2e tests? No
- If it is not possible to write automated tests, explain why and document how
  to manually test and verify the feature. I am not sure how to perform a manual test on that one.

### Is there any documentation that needs to be updated for this PR?

N/A

PR #2349, opened 2022-08-25
remove kube-storage-version-migrator condition filter (do not Ignore Upgradeable/Unknown condition)

AldoFusterTurpin:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/9338383

In the past: 
"As already [reported](https://github.com/openshift/cluster-kube-storage-version-migrator-operator/issues/25) by another team member, kube-storage-version-migrator is currently being constantly reported as "Upgradeable" condition status "Unknown". This makes the Cluster Operator constantly being reported as unhealthy and creates a lot of noise in our dashboards."

Now:
 kube-storage-version-migrator is NO longer reported as "Upgradeable" condition status "Unknown". So we can now remove the filter that was previously done in https://github.com/Azure/ARO-RP/pull/1310

Explanation:
I have created a new cluster (using local ARO-RP) and after executing 
`oc get clusteroperator -o yaml`
we can see the attached output...

oc_get_clusteroperator_output

With the important block being:
```
status:
    conditions:
    - lastTransitionTime: "2022-08-24T13:09:19Z"
      message: All is well
      reason: AsExpected
      status: "True"
      type: Upgradeable
```

so the operator condition Upgradeable is NO longer "Unkown".

### Test plan for issue:

N/A

N/A 
Explanation added here and in ADO Task.

### DIFF
--- a/pkg/monitor/cluster/clusteroperatorconditions.go
+++ b/pkg/monitor/cluster/clusteroperatorconditions.go
@@ -19,12 +19,11 @@ type clusterOperatorConditionsIgnoreStruct struct {
 // clusterOperatorConditionsIgnore contains list of failures we know we can
 // ignore for now
 var clusterOperatorConditionsIgnore = map[clusterOperatorConditionsIgnoreStruct]struct{}{
-	{"insights", "Disabled", configv1.ConditionFalse}:                                          {},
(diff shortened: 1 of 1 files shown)


s-amann (approved):

bizz001:
lgtm

bennerv (approved):

PR #2353, opened 2022-08-26
Removing yum exclude on WALinuxAgent to resolve dependency errors 

swiencki:
N/A

Removing yum exclude on WALinuxAgent to resolve dependency errors on civmss deployments

### Test plan for issue:

SECRET_SA_ACCOUNT_NAME=e2earosecrets make secrets
. ./secrets/env
. ./hack/devtools/deploy-shared-env.sh
RESOURCEGROUP=e2e-eastus deploy_env_dev_ci

No


swiencki:
E2E failed on

cblecker:
This exclude has been in the code base a long time (earliest instance I can see dates back to https://github.com/Azure/ARO-RP/commit/fe78fbe4401299305fe6512b38123e99837b132d in December 2019).

It's also in resources_gateway.go and resources_rp.go, which were both split out from the _dev file.

Open questions:
- Is there a reason we should be not upgrading the WALinuxAgent?
- If we decide we should upgrade it, should we upgrade it in all three resource files?

swiencki:
> 

I did some digging around on why it was excluded in the first place. The reasoning I got is that it upgrades itself, however we are running into the following error on deploy:

`Error: \n Problem: package WALinuxAgent-2.3.0.2-2.el8.noarch requires WALinuxAgent-udev = 2.3.0.2-2.el8, but none of the providers can be installed\n  - cannot install both WALinuxAgent-udev-2.3.0.2-2.el8_6.2.noarch and WALinuxAgent-udev-2.3.0.2-2.el8.noarch\n  - cannot install the best update candidate for package WALinuxAgent-udev-2.3.0.2-2.el8.noarch\n  - problem with installed package WALinuxAgent-2.3.0.2-2.el8.noarch\n`

Just removing the exclude has been the path of least resistance and ensures the package is latest.

As for the other deployments; resources_gateway.go and resources_rp.go it would be required to test those changes before hand but it's likely they will run into the same issue.

pkg/deploy/generator/resources_dev.go
```diff
@@ -312,7 +312,7 @@ func (g *generator) devCIPool() *arm.Resource {
 sleep 60
 
 for attempt in {1..5}; do
-  yum -y update -x WALinuxAgent && break
+  yum -y update && break
```
cblecker, line 315: Looks like there is one more instance on L41 of this file

bennerv (dismissed):
we shouldn't be seeing other people's changes in this PR, only ones which should diff from our current master.  Can you fix the rebases on this so it only shows changes introduced by this PR?

cblecker (approved):

swiencki:
> we shouldn't be seeing other people's changes in this PR, only ones which should diff from our current master. Can you fix the rebases on this so it only shows changes introduced by this PR?

Rebase corrected :+1:

PR #2355, opened 2022-08-28
[admin-portal]: add a fix for subscription ID filtering

devamanv:

Adds a minor fix for the subscription ID filtering in the admin portal
Ref: [ADO Card](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15132767)

Currently, the filtering by subscription ID feature is broken. This PR provides a fix for that so that filtering works on either the exact match or a substring match of the subscription ID of an ARO cluster.


darthhexx (approved):
lgtm

PR #2358, opened 2022-08-29
Add e2e test for hive kubeconfig

petrkotas:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15372448/ 

Add new e2e test to cover hive kubeconfig manipulation

### DIFF
--- a/test/e2e/aks.go
+++ b/test/e2e/aks.go
@@ -1,0 +1,69 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	"github.com/Azure/ARO-RP/pkg/util/liveconfig"
+)
+
+// Tests the kubeconfig ability to get and manipulate the cluster
+var _ = Describe("AKS cluster present", func() {
+	ctx := context.Background()
+	var liveConfig liveconfig.Manager
+	var kubeConfig *rest.Config
+
+	BeforeEach(func() {
+		var err error
+		liveConfig, err = _env.NewLiveConfigManager(ctx)
+
+		Expect(err).To(BeNil())
+	})
+
+	It("should get kubeconfig", func() {
+		var err error
+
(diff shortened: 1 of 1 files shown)


facchettos (dismissed):
seems fairly straight forward !

test/e2e/aks.go
```diff
@@ -0,0 +1,69 @@
...
+
+		kubeConfig, err = liveConfig.HiveRestConfig(ctx, 0)
+		Expect(err).To(BeNil())
+		Expect(kubeConfig).ToNot(BeNil())
+
+		kubernetescli, err := kubernetes.NewForConfig(kubeConfig)
+		Expect(err).To(BeNil())
+
+		// to avoid name collision by accident
+		testNamespaceName := "e2e-test-namespace" + time.Now().Format("20060102150405")
```
cblecker, line 48: non-blocking nit:
```suggestion
		testNamespaceName := "e2e-test-namespace-" + time.Now().Format("20060102150405")
```

test/e2e/aks.go
```diff
@@ -0,0 +1,69 @@
...
+
+		_, err = kubernetescli.CoreV1().Namespaces().Create(ctx, testNamespace, metav1.CreateOptions{})
+		Expect(err).To(BeNil())
+
+		Eventually(func() error {
+			_, err := kubernetescli.CoreV1().Namespaces().Get(ctx, testNamespaceName, metav1.GetOptions{})
+			return err
+		}).WithTimeout(20 * time.Second).WithPolling(1 * time.Second).Should(Succeed())
+
+		Eventually(func() error {
```
cblecker, line 64: Is this integral to the test, or is this more a `DeferCleanup()` type function?

test/e2e/aks.go
cblecker: We could also use `GinkgoParallelProcess()` to append the process number such that if there happens to be multiple functions using this generic namespace name at the same time, they won't step on each other's toes.

petrkotas: Unfortunately the `GinkgoParallelProcess()` will not help. The AKS/Hive is shared between different pipelines. It can easily happen that multiple pipelines will have all process `1`. This was the motivation for selecting date.

test/e2e/aks.go
petrkotas: It is both. It is cleanup, but integral to this particular function. I was considering `DeferCleanup()` originally, but for this single function it is an overkill. In the background it created `AfterEach()` and it is tied to `BeforeEach()`. Which will come in handy, when there are more tests present. Until that time, I would like to keep it streamlined to single test case.

facchettos (approved):

cblecker (approved):

darthhexx:
This PR doesn't take into consideration the absence of AKS in FairFax and the 6 regions that do not have Hive installed and is going to result in E2E failures in those regions. We are going to need a follow up PR that skips this test in a predefined list of regions.

PR #2359, opened 2022-08-29
CloudError added for clusterDelete

carvalhe:
Usage: Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14409128

Fixes

This PR adds a new CloudErrorCode for the top incorrectly categorized server error case occurring in cluster delete. While producing the newly created ErrorCode, it essentially transforms a DetailedError into a CloudError type. We need this because currently the metrics for cluster delete looks for the CloudError type to understand if the issue is a user or server error, and when not CloudError, it defaults to server error. Here is a related ICM where the loadbalancer can't be deleted with a privateLinkService, but is mitgated as it's a user error, https://portal.microsofticm.com/imp/v3/incidents/details/305928113/home

### Test plan for issue:
When deleting a cluster with a private link service, logs correctly categorize issue as userError.

Correcting categorization/new error type. N/A

### DIFF
--- a/pkg/cluster/delete.go
+++ b/pkg/cluster/delete.go
@@ -231,6 +231,16 @@ func (m *manager) deleteResources(ctx context.Context) error {
 			m.log.Printf("deleting %s", *resource.ID)
 			future, err := m.resources.DeleteByID(ctx, *resource.ID, apiVersion)
 			if err != nil {
+				detailedError, ok := err.(autorest.DetailedError)
+				if ok {
+					if strings.Contains(detailedError.Original.Error(), "CannotDeleteLoadBalancerWithPrivateLinkService") {
+						err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeCannotDeleteLoadBalancerByID,
+							"features.ResourcesClient#DeleteByID", detailedError.Original.Error())
+					} else if strings.Contains(detailedError.Original.Error(), "AuthorizationFailed") {
+						err = api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden,
+							"features.ResourcesClient#DeleteByID", detailedError.Original.Error())
+					}
+				}
 				return err
 			}
 
--- a/pkg/cluster/delete.go
+++ b/pkg/cluster/delete.go
@@ -231,17 +231,7 @@ func (m *manager) deleteResources(ctx context.Context) error {
 			m.log.Printf("deleting %s", *resource.ID)
 			future, err := m.resources.DeleteByID(ctx, *resource.ID, apiVersion)
 			if err != nil {
-				detailedError, ok := err.(autorest.DetailedError)
-				if ok {
-					if strings.Contains(detailedError.Original.Error(), "CannotDeleteLoadBalancerWithPrivateLinkService") {
-						err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeCannotDeleteLoadBalancerByID,
-							"features.ResourcesClient#DeleteByID", detailedError.Original.Error())
-					} else if strings.Contains(detailedError.Original.Error(), "AuthorizationFailed") {
-						err = api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden,
-							"features.ResourcesClient#DeleteByID", detailedError.Original.Error())
-					}
-				}
-				return err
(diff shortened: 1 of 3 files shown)


bennerv (commented):
Question on this.  Ideally we'll provide actionable error messages to the customer if there's not an infrastructure issue.  What sort of action would be expected by the customer to take on this?

carvalhe:
> @bennerv  Question on this. Ideally we'll provide actionable error messages to the customer if there's not an infrastructure issue. What sort of action would be expected by the customer to take on this?

For this issue, the customer would have to go and delete the PrivateLinkService if they were attempting to delete the load balancer. Here is an ICM related to that type of event https://portal.microsofticm.com/imp/v3/incidents/details/305928113/home. The other case would be the user isn't authorized to use the get/delete actions.

pkg/cluster/delete.go
```diff
@@ -231,6 +231,16 @@ func (m *manager) deleteResources(ctx context.Context) error {
...
+				detailedError, ok := err.(autorest.DetailedError)
+				if ok {
+					if strings.Contains(detailedError.Original.Error(), "CannotDeleteLoadBalancerWithPrivateLinkService") {
+						err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeCannotDeleteLoadBalancerByID,
+							"features.ResourcesClient#DeleteByID", detailedError.Original.Error())
+					} else if strings.Contains(detailedError.Original.Error(), "AuthorizationFailed") {
+						err = api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden,
+							"features.ResourcesClient#DeleteByID", detailedError.Original.Error())
+					}
+				}
```
facchettos, line 243: this probably should be in a function, that is a lot of ifs here

facchettos (dismissed):

pkg/cluster/delete.go
carvalhe: Moved to its own function and changed to a switch statement

pkg/cluster/delete.go
```diff
@@ -251,6 +251,22 @@ func (m *manager) deleteResources(ctx context.Context) error {
...
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeCannotDeleteLoadBalancerByID,
+				"features.ResourcesClient#DeleteByID", detailedError.Original.Error())
+
+		case strings.Contains(detailedError.Original.Error(), "AuthorizationFailed"):
+			return api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden,
+				"features.ResourcesClient#DeleteByID", detailedError.Original.Error())
+		}
+	}
+	return err
+}
```
facchettos, line 268: wondering if this wouldn't have a nicer flow... not asking to change, just wondering. WDYT?
```suggestion
func deleteByIdCloudError(err error) error {
	detailedError, ok := err.(autorest.DetailedError)
	if !ok {
		return err
	}
	switch {
	case strings.Contains(detailedError.Original.Error(), "CannotDeleteLoadBalancerWithPrivateLinkService"):
		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeCannotDeleteLoadBalancerByID,
			"features.ResourcesClient#DeleteByID", detailedError.Original.Error())

	case strings.Contains(detailedError.Original.Error(), "AuthorizationFailed"):
		return api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden,
			"features.ResourcesClient#DeleteByID", detailedError.Original.Error())
	default:
		return err	
	}
}
```

carvalhe: I think this is nice, but I prefer the one in place as it matches the style of deleteResourcesAndResourceGroup where it checks if ok. The other reason is that there is only one place where we are returning err, and it acts as a default too.

facchettos: having an explicit flow is better than having one less line of code. Also nesting should be avoided when possible. 
https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

carvalhe: Fair enough, having a good flow is always nice. We can change to the suggested version then.

s-amann (approved):

ross-bryan (approved):
lgtm

PR #2361, opened 2022-08-30
Specify platform when building or running images

cblecker:
None

Specifies to use linux/amd64 when building or running images. This is needed if you are running a non-amd64 architecture (like arm64/v8).

### Test plan for issue:
Ran locally on arm64 based system.

No


facchettos (approved):
seems fairly straight forward

ross-bryan (approved):
lgtm

PR #2332, opened 2022-08-17
[M8] Hive/OCM CRUD Operations

ross-bryan:
### Which issue this PR addresses:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14955492/

~~_This is still a work in progress while I make changes required to pass azure rest spec api validation. However, changes should be minimal and reviews are still welcome._~~ Passing Validation - only feedback is to make list version a proxy resource which will be done in a follow up PR

### What this PR does / why we need it:
Implementation for CRUD operations for hive resources from OCM client. 

This PR receives requests, writes the resources to a new container called `ClusterManagerConfigurations` and stores the resources there. Additional PRs/project milestones will be responsible for taking the items from cosmos and applying to the hive cluster.

### Test plan for issue:

- [x] passing azure-rest-spec-api validation
- [ ] deploy to canary/pilot

Right now just by running RP locally 

### Is there any documentation that needs to be updated for this PR?
TBD

PR #2363, opened 2022-08-30
[DRAFT] [M7] New backend to propagate OCM requests to Hive

mbarnes:
**Note:** This PR builds upon https://github.com/Azure/ARO-RP/pull/2332 and is therefore in draft state ~until the first PR is merged~.
**Update 9/14:** https://github.com/Azure/ARO-RP/pull/2332 is merged but leaving this in **DRAFT** until [SD-ADR-0077](https://docs.google.com/document/d/1CU4Cmjz3nh643bSAlPEfrfpJeAa_xMBVmg4lDQo1n90/) is decided.

Implements feature: [M7 - ARO Hive Day-2 SyncSet Mechanism](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14846838)

Implements a new backend and a new CosmosDB container to store backend state.  One and only one RP instance (per region) processes events from a CosmosDB change feed and propagates the changes to the region's Hive cluster on AKS.

### Test plan for issue:

To meet the required deadline, only manual testing was possible.

No, but some rationale for the backend design is captured in commit messages.


pkg/backend/clustermanagerconfiguration.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	docs, err := cmc.configIterator.Next(ctx, -1)
+	if err != nil {
+		return false, err
+	}
+
+	workWasDone := false
+
+	if docs != nil {
```
petrkotas, line 77: In this loop I see it takes the configuration and passes it directly to the AKS/Hive cluster. What is the plan for linking the configuration with RP clusterDocument? I can easily see AKS going bananas and the need to restore the state. We also need a way to know which configuration is linked to which cluster.

Pardon me, if I missed it, this PR is realy big.

petrkotas (changes requested):
Hi @mbarnes thank you for your work. It is really a lot of effort put into this PR. I have a question related to the document handling bellow.

In general, how is the overal document flow? It is hard to understand the relation between backendDocument and clusterManagerConfiguration.
Would you have a diagram with such info? Thank you a lot and thank you again for the work put into this PR.

pkg/database/backends.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	collClient := cosmosdb.NewCollectionClient(dbClient, dbid)
+	triggerClient := cosmosdb.NewTriggerClient(collClient, collBackends)
+
+	trigger := &cosmosdb.Trigger{
+		ID:               renewLeaseTriggerID,
+		TriggerOperation: cosmosdb.TriggerOperationAll,
+		TriggerType:      cosmosdb.TriggerTypePre,
+		Body: fmt.Sprintf(`function trigger() {
```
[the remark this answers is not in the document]
mbarnes: It is, I stole that from `pkg/database/monitors.go`.

pkg/backend/clustermanagerconfiguration.go
mbarnes: Yeah, it's subtle.  The assumption is the payload from OCM will include the correct Hive namespace for the target cluster.  All we have to do is forward it on.

So if we had to do disaster recovery, I expect we would first recreate the same Hive namespaces for existing clusters based on the RP `clusterdocument`s.  Then we just throw all the resources in the `ClusterManagerConfigurations` container at Hive and they would land in the correct namespaces.

mbarnes:
> It is hard to understand the relation between backendDocument and clusterManagerConfiguration.

The two documents are mostly unrelated.  `backendDocument` is just a place for backends to stash state.  Backend state in this case means 1) a lease that only one RP instance can acquire, and 2) the so-called "continuation value" for the `clusterManagerConfiguration` change feed, which is basically a bookmark.

So if an RP instance holding the `backendDocument` lease dies for whatever reason, the lease will expire after one minute and another RP instance can then acquire the lease and begin reading the `clusterManagerConfiguration` change feed starting from the last recorded "continuation value" (otherwise it would replay the entire change history of the DB container).

Does that help clarify?

mociarain:
> #2332 is merged but leaving this in **DRAFT** until [SD-ADR-0077](https://docs.google.com/document/d/1CU4Cmjz3nh643bSAlPEfrfpJeAa_xMBVmg4lDQo1n90/) is decided.

Looks like that ADR has been accepted. What do we want to do with this guy now?

mbarnes:
Sorry, I didn't realize this was still open.  Closing since all this code is obsolete.

As I understand, at this point OCM integration with ARO Classic is unlikely to ever happen.  The API extensions for Hive resources from OCM (#2332) should never have been published because we're stuck with this unused public API now.

PR #2241, opened 2022-07-11
Feat: Add API support for choosing Software Defined Network (OVN/SDN)

SudoBrendan:
### Which issue this PR addresses:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14940808

### What this PR does / why we need it:

Adds support for selecting the Software Defined Network (SDN) you want when provisioning a new cluster (OVN or SDN). This is a near-revert of https://github.com/Azure/ARO-RP/pull/2210/files (a few tweaks, and in a different api version).

Also includes a `make client`.

### Test plan for issue:

Tests have been modified to reflect expected behavior.

### Is there any documentation that needs to be updated for this PR?

No, swagger has been updated.

PR #2364, opened 2022-08-30
Remove SDN option from the stable API

cblecker:

Fixes

This is a near revert of #2241, which added selecting SDN as an option to the stable API


bennerv (commented):
Why are there file extensions with `.go-e`?

Additionally, we must follow Azure's [deprecation policy](https://eng.ms/docs/products/arm/api_contracts/apideprecationpolicy) on removing our 2021-09-01 preview API which currently supports setting the SDN. 

Since it was a preview API, we're not subject to the 36 month notice we would have had to give customers, just 90 days.

cblecker:
@bennerv
Not looking at removing from the preview API right now, just removing from the stable API.

As for `.go-e`, these were generated by autorest. I just ran a `make client`.

bennerv:
> @bennerv Not looking at removing from the preview API right now, just removing from the stable API.
> 
> As for `.go-e`, these were generated by autorest. I just ran a `make client`.

That's weird.  It should generate `.go` files.  You can't remove it from existing APIs though so any modifications to files non-2022-09-04 are not good.

cblecker:
The new go-e files were the only ones changing for the other API versions. I removed them from the commit.

ross-bryan (approved):
lgtm

bennerv (approved):

PR #2366, opened 2022-08-30
Drain Node E2E Test Fix

zacharyljones:

This PR addresses [Bug 15410721](https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/15410721) which resolves a recurring flake in the drain node E2E tests. For example - https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=59442207&view=logs&j=3965c7ea-79de-5618-9d9a-42b627bfb354&t=4810cc2a-45e4-5107-9a11-ae0073a144ef 

This PR addresses a recurring "flake" in the drain node E2E tests - Tests will fail if there are only daemon set pods running on the target worker node, because the number of pods targeted for deletion will be 0.

### Test plan for issue:

The CI E2E tests that run for this PR will run the modified E2E test.

No - there is no explicit documentation for E2E tests.

### DIFF
--- a/test/e2e/adminapi_drainnode.go
+++ b/test/e2e/adminapi_drainnode.go
@@ -14,7 +14,6 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/kubectl/pkg/drain"
 )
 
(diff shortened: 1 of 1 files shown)


ross-bryan (approved):

ross-bryan:
failed on the drain e2e test :( 

```
  should be able to cordon, drain, and uncordon nodes [It]
  /home/cloud-user/agent/_work/4/go/src/github.com/Azure/ARO-RP/test/e2e/adminapi_drainnode.go:24

  Expected
      <int>: 500
  to equal
      <int>: 200
      ```

cblecker:
Is this test even useful if the result can be any number equal or greater than 0? Do we ever expect this to be a negative number?

zacharyljones:
> Is this test even useful if the result can be any number equal or greater than 0? Do we ever expect this to be a negative number?

Fair point - Originally I had always expected this number to be greater than 0. At the time it seemed unlikely that it would ever be 0, but there have already been several instances of the test failing because it equaled 0. Deleting this check altogether may be more appropriate.

zacharyljones:
@ross-bryan @cblecker - I removed the "number of pods" checks from this test as they seem to be unnecessary and causing nothing but problems.

cblecker (approved):

ross-bryan (approved):

PR #2368, opened 2022-08-30
Forcibly redeploy RP NSG to get new GenevaActions rule

ventifus:

Forces the RP NSG to redeploy in order to get new GenevaActions tag. The commit to do so in #2333 was incorrect and did not cause a redeploy.

### Test plan for issue:

Confirm the NSG gets redeployed properly.

No

### DIFF
--- a/pkg/deploy/assets/rp-production-predeploy-parameters.json
+++ b/pkg/deploy/assets/rp-production-predeploy-parameters.json
@@ -3,7 +3,7 @@
     "contentVersion": "1.0.0.0",
     "parameters": {
         "deployNSGs": {
-            "value": true
+            "value": false
         },
         "extraClusterKeyvaultAccessPolicies": {
             "value": []
--- a/pkg/deploy/assets/rp-production-predeploy.json
+++ b/pkg/deploy/assets/rp-production-predeploy.json
@@ -57,7 +57,7 @@
     "parameters": {
         "deployNSGs": {
             "type": "bool",
-            "defaultValue": true
+            "defaultValue": false
         },
         "extraClusterKeyvaultAccessPolicies": {
             "type": "array",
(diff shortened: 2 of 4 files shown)


bennerv (approved):
lgtm - hack for now but can revert after next deploy

darthhexx:
@bennerv @ross-bryan this kind of toggle is what I was aiming at solving with this [RP-Config PR](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/RP-Config/pullrequest/6413617). This is something we need to replicate exactly what you are doing in this PR for the "[Add the RP VNET creation to the bootstrap ARM template](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines/pullrequest/6337444)" PR, so that the VNETs are redeployed each time. Thoughts on the approach?

bennerv:
> @bennerv @ross-bryan this kind of toggle is what I was aiming at solving with this [RP-Config PR](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/RP-Config/pullrequest/6413617). This is something we need to replicate exactly what you are doing in this PR for the "[Add the RP VNET creation to the bootstrap ARM template](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines/pullrequest/6337444)" PR, so that the VNETs are redeployed each time. Thoughts on the approach?

I would rather be more explicit in what the config parameter is doing.  Change from `fullBootstrapLocations` to `forceDeployNSGLocations` or something like that.

bennerv:
And adding onto this, we need to define the vnets with NSGs right away otherwise we get flagged + NSGs are automatically added to the VNETs causing issues later.  Just for context (or at least this is what it used to be).

darthhexx:
Thanks for the feedback @bennerv, I've made the changes to both the RP-Config and the ARO.Pipelines PR's.

PR #2369, opened 2022-08-31
Fix autorest client generation

cblecker:
relates to ADO-15132425

Minor changes to autorest image and client generation:
- remove superfluous flags when installing autorest
- remove no longer needed sudo commands
- run sed operation inside a container, to eliminate flag differences between BSD/Darwin and GNU/Linux versions of sed

### Test plan for issue:
Tested locally

No, these are just small fixes to scripts


hawkowl (approved):
sure

PR #2375, opened 2022-09-01
Move the disabling of samples and operatorhub sources earlier in the installation process

bennerv:

Fixes [ADO#15431594](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15431594)

There's a issue where the clusterversion CR waits for samples to go healthy.  If we disable samples before we wait for the clusterversion, then it should not error out in installations.  

ICM 331555133 has a similar failure scenario

### Test plan for issue:

I can't replicate this, but customers are experiencing this.  The clusterversion CR shouldn't complain about a disabled operator so it's moved earlier in the install process.  

no


dem4gus (approved):
lgtm

bennerv:
> lgtm. Is it possible to also disable the BareMetal operator the same way or do we just rely on it disabling itself?

It's possible in 4.11 now I guess
https://docs.openshift.com/container-platform/4.11/release_notes/ocp-4-11-release-notes.html#ocp-4-11-capabilities

So soon :)

pkg/cluster/install.go
```diff
@@ -247,11 +247,11 @@ func (m *manager) Install(ctx context.Context) error {
 			steps.Condition(m.operatorConsoleExists, 30*time.Minute, true),
 			steps.Action(m.updateConsoleBranding),
 			steps.Condition(m.operatorConsoleReady, 20*time.Minute, true),
+			steps.Action(m.disableSamples),
+			steps.Action(m.disableOperatorHubSources),
 			steps.Condition(m.clusterVersionReady, 30*time.Minute, true),
 			steps.Condition(m.aroDeploymentReady, 20*time.Minute, true),
 			steps.Action(m.disableUpdates),
```
cblecker, line 254: We should also probably move updates up?

bennerv: disableUpdate just removes the channel from the clusterversion CR.  I was mostly focused on things that could impact the clusterversion status on installation.

cblecker: It will speed it up if it doesn't need to pull cincinnati updates

bennerv: updated - ty for the info.  TIL

cblecker (approved):
lgtm

PR #2376, opened 2022-09-01
Update auth used by rhcos uploads

SudoBrendan:

N/A, ref: https://github.com/Azure/ARO-RP/pull/2339

This updates the authentication mechanism used by the hack script to get access to the (temporary) Storage Account used to store RHCOS images and get SAS links.

### Test plan for issue:

N/A - I did verify that the auth works after sourcing proper envvars and running `az login`.

Yes, I've done that [in our wiki](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/186149/How-To-Publish-New-OCP-Image-for-ARO?anchor=copy-the-image-and-get-os-vhd-link)

### DIFF
--- a/hack/rhcos-sas/rhcos-sas.go
+++ b/hack/rhcos-sas/rhcos-sas.go
@@ -27,17 +27,17 @@ func run(ctx context.Context, log *logrus.Entry) error {
 	resourceGroup := "images"
 	accountName := "openshiftimages"
 
-	authorizer, err := auth.NewAuthorizerFromEnvironment()
+	_env, err := env.NewCore(ctx, log)
 	if err != nil {
 		return err
 	}
 
-	env, err := env.NewCore(ctx, log)
+	authorizer, err := auth.NewAuthorizerFromCLIWithResource(_env.Environment().ResourceManagerEndpoint)
 	if err != nil {
 		return err
 	}
 
-	accounts := storage.NewAccountsClient(env.Environment(), subscriptionID, authorizer)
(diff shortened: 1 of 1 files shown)


cblecker (approved):

hack/rhcos-sas/rhcos-sas.go
```diff
@@ -27,12 +27,12 @@ func run(ctx context.Context, log *logrus.Entry) error {
 	resourceGroup := "images"
 	accountName := "openshiftimages"
 
-	authorizer, err := auth.NewAuthorizerFromEnvironment()
+	env, err := env.NewCore(ctx, log)
```
bennerv, line 30:
```suggestion
_env, err := env.NewCore(ctx, log)
```

if you use a variable name same as the package you're importing, it can lead to some interesting things later on.  

We currently do it in a few places, but should fix it up.

bennerv (changes requested):
one nit

hack/rhcos-sas/rhcos-sas.go
bennerv: Will need to update the usage of the variable below as well.

SudoBrendan: will do, great catch!

bennerv (approved):
lgtm

jharrington22 (commented):
Other than @bennerv suggestion looks good.

jharrington22 (approved):
lgtm

PR #2378, opened 2022-09-02
Refactor "pkg/deploy/generator": extract huge hardcoded strings

AldoFusterTurpin:

Fixes readability issues in the codebase regarding huge hardcoded strings in pkg/deploy/generator/*.go

Having hardcoded strings with 580+ lines in the middle of a big Go function is something we should avoid as we lose focus on what the code was doing (the method becomes hard to follow and understand). 

This PR extracts those hardcoded strings using the embed package in order to not merge infrastructure scripts with ARO-RP logic. 

### Test plan for issue:

No unit tests were added as this is just extracting the huge strings (no behavior to test at "unit level").

I ensured I can access the variables by printing them in the main function (first making the embedded variables exportable and then making them unexportable after checking it).

N/A.


facchettos (changes requested):
I think you should use embed here instead.

AldoFusterTurpin:
> I think you should use embed here instead.

Yes, I completely agree. 
Done. Thanks!

facchettos (dismissed):

facchettos:
if you want to remove these ugly new lines, you can have a new commit (keep both so we have the history and it shows that this is a wanted change) where you remove it and run make generate. Might make it a little nicer (or not). up to you, I think it's fine either way

facchettos (approved):

facchettos (approved):

cblecker (approved):

PR #2380, opened 2022-09-05
Option for Hive Adoption

hawkowl:
Based off #2374 

n/a

Introduces a config flag option for Hive adoption of clusters. Makes accessing Hive fail hard if this option is set (except for deletes), to avoid issues going unnoticed.

### Test plan for issue:

Unit tests are added

N/A


hawkowl:
Nicer diff off #2374: https://github.com/Azure/ARO-RP/pull/2380/files/9b4a6a844f3987745cd6f87f02b29fa624d8278d..514caa083acb0cd44afda036c70df91fbf8d57df

pkg/deploy/generator/templates_rp.go
```diff
@@ -124,6 +125,16 @@ func (g *generator) rpTemplate() *arm.Template {
 				"northcentralus",
 				"uaenorth",
 			}
+
+			// TODO: Replace with Live Service Configuration in KeyVault
+		case "clustersInstallByHive":
```
zacharyljones, line 130: Should be changed to `clustersInstallViaHive` to match value in `params` array

pkg/deploy/generator/resources_rp.go
```diff
@@ -476,6 +476,7 @@ func (g *generator) rpVMSS() *arm.Resource {
 		// TODO: Replace with Live Service Configuration in KeyVault
 		"clustersInstallViaHive",
 		"clusterDefaultInstallerPullspec",
+		"clustersAdoptByHive",
```
zacharyljones, line 479: Having `clustersInstallViaHive` and `clustersAdoptByHive` in this array causes the following ARM template error:

`Unable to process template language expressions for resource 'xxxxxxxxx' at line '1' and column '7182'. 'The template language function 'base64' expects its parameter to be of type 'String'`
 
 To get around this, I moved `clustersInstallViaHive` and `clustersAdoptByHive` to the array on line 489 - the one with `adminApiCaBundle` and `armApiCaBundle` - so `base64` does not get called on these boolean parameters.

zacharyljones: To get around a separate issue with the aro rp service not starting, I reverted this change and tried making these fields strings in the ARM template. This change also allowed the ARM template validation to succeed, but the aro-rp service would not start.

pkg/cluster/install.go
```diff
@@ -186,10 +190,12 @@ func (m *manager) bootstrap() []steps.Step {
 		steps.AuthorizationRefreshingAction(m.fpAuthorizer, steps.Action(m.ensureGatewayCreate)),
 		steps.Action(m.createAPIServerPrivateEndpoint),
 		steps.Action(m.createCertificates),
+	}
 
+	if m.adoptViaHive || m.installViaHive {
 		// We will always need a Hive namespace, whether we are installing
 		// via Hive or adopting
```
darthhexx, line 197: We should remove this comment as the if statement above makes this clear now.

darthhexx (approved):
Overall LGTM, just 2 small nits that aren't blockers and will be addressed in follow up PRs.

pkg/cluster/delete.go
```diff
@@ -456,11 +456,13 @@ func (m *manager) Delete(ctx context.Context) error {
...
-	if err != nil {
-		m.log.Info(err)
+	if m.adoptViaHive || m.installViaHive {
+		// Don't fail the deletion because of hive
+		// This should change when/if we start using hive for cluster deletion
+		err = m.hiveDeleteResources(ctx)
+		if err != nil {
+			m.log.Info(err)
+		}
 	}
```
jharrington22, line 466: Does this mean could lead ClusterDeployments?

pkg/deploy/generator/resources_rp.go
jharrington22: This should have been fixed in the last commit by @hawkowl https://github.com/Azure/ARO-RP/pull/2380/commits/1bbc858a98a1f051a9980fc98cced5ad2abdee63

jharrington22 (approved):

pkg/cluster/delete.go
ross-bryan: It would log it within the RP server logs, but no customer could access these logs.

pkg/backend/openshiftcluster.go
```diff
@@ -102,12 +102,31 @@ func (ocb *openShiftClusterBackend) handle(ctx context.Context, log *logrus.Entr
...
+	}
+
+	var hr hive.ClusterManager
+	if installViaHive || adoptViaHive {
+		hiveShard := 1
+		hiveRestConfig, err := ocb.env.LiveConfig().HiveRestConfig(ctx, hiveShard)
+		if err != nil {
+			return fmt.Errorf("failed getting RESTConfig for Hive shard %d: %w", hiveShard, err)
+		}
+		hr, err = hive.NewFromConfig(log, ocb.env, hiveRestConfig)
```
ross-bryan, line 123: what does `hr` mean in this context? I can see h for hive, but `hr` doesn't mean anything to me

pkg/cluster/install.go
```diff
@@ -131,15 +130,20 @@ func (m *manager) Update(ctx context.Context) error {
...
-		// Hive reconciliation: we mostly need it to make sure that
-		// hive has the latest credentials after rotation.
-		steps.Action(m.hiveCreateNamespace),
-		steps.Action(m.hiveEnsureResources),
-		steps.Condition(m.hiveClusterDeploymentReady, 5*time.Minute, true),
-		steps.Action(m.hiveResetCorrelationData),
 	}
 
-	return m.runSteps(ctx, steps, false)
+	if m.adoptViaHive {
```
ross-bryan, line 135: why are we adding `adoptViaHive` into the `az aro update` customer workflow?

ross-bryan (commented):
LGTM only concern is modifying the `az aro update` path - I don't like introducing Hive into that path at this point.

pkg/cluster/install.go
bennerv: I think it's because hive also requires the cluster service principal to function properly.  So if a customer goes through the Update flow to rotate credentials and their cluster is already adopted by hive, it needs to rotate the credentials within Hive too.

This does have a side-affect of adopting the cluster to hive on an update if the cluster isn't already adopted.

ross-bryan: :+1:

ross-bryan (approved):
lgtm

PR #2385, opened 2022-09-07
Fix Frontend Unit Test Flake in List Install Versions

bennerv:

Fixes a flake in the list install versions unit test

The response from the RP is non-deterministic.  So we must first sort the response before validating it.

Ideally, we could check the presence of values rather than comparing responses, but then we'd have to write a custom function for this specific test and I would rather just sort.  

### Test plan for issue:

```bash
$ while true; do go test -run ^TestListInstallVersions$ -count=1 ./pkg/frontend/; done
PASS
...
```

No

### DIFF
--- a/pkg/frontend/installversions_list_test.go
+++ b/pkg/frontend/installversions_list_test.go
@@ -5,6 +5,7 @@ package frontend
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


s-amann (approved):

jharrington22 (approved):
lgtm

darthhexx (approved):

PR #2386, opened 2022-09-07
Update CI VMSS CSE script to resize LVM partition

tony-schndr:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15481503/

This update adds steps to the CI VMSS CSE script to resize the LVM partition.  This is to mitigate the issue of CI failing due to running out of disk space and ensure the partition resize persists when CI VMSS is redeployed.

### Test plan for issue:

Tested by deploying new CI VMSS, ensure partition is resized correctly, and validate CI jobs succeed.  

NA


tony-schndr:
@darthhexx when you get a chance would you care to review this change?  Your script for this fix was really close.  The only thing that needed to be pulled out was the `N` for the prompt `Do you want to remove the signature?`.  The prompt doesn't occur and defaults to No when running fdisk using a pipe for input.

darthhexx (dismissed):
Thanks for working on this, lgtm

pkg/deploy/generator/resources_dev.go
```diff
@@ -316,6 +316,36 @@ for attempt in {1..5}; do
...
+DEVICE=$(echo $DEVICE_PARTITION | grep -oP '^[a-z]{3}')
+
+# Fix the "GPT PMBR size mismatch (134217727 != 268435455)"
+echo "w" | fdisk /dev/${DEVICE}
+
+# Steps from https://access.redhat.com/solutions/5808001
+# 1. Delete the LVM partition
+# 2. Recreate the partition
+# 3. Accept the default start and end sectors (2 x \n)
+# 4. DO NOT delete the LVM2_member signature
```
SudoBrendan, line 329: either the doc is out of date or something is a little off here. After `n` the solution you linked outlines:

a. selecting the partition type of Primary (default) vs Extended
b. selecting the partition number (default)
c. setting the first sector (default)
d. setting the last sector (default)
e. a `y/n` for removing the signature (no default)

This comment calls for 2 `\n`, but there are 3 in the script, and the doc calls for 4 + a `y/n` response that is commented but not in the script. Can you clarify what's going on here?

SudoBrendan (changes requested):
Comments/Docs/Scripts are not looking aligned. Can you give some clarification?

pkg/deploy/generator/resources_dev.go
SudoBrendan: It might add clarity to put the actual commands as part of these comments:

```
# 1. Delete the LVM partition "d\n2\n"
# 2. Recreate the partition as a primary with default partition number, default start sector (first available), default end sector (last available), and without removing the signature "n\np\n2\n\n\nn\n"
# 3. Set the type of the partition "t\n2\n31\n"
# 4. Write the table "w\n"
```

tony-schndr: I see what you mean.  The solution shows you select partition type but I am never prompted for that locally.  I'm not certain why it is this way.  I will also state that it doesn't look like `@darthhexx` encountered this when running it ad hoc.

As for the comment misaligned with the script, I'll push a update to clarify that.

darthhexx (approved):

darthhexx:
@SudoBrendan please would you take another look after the recent commits.

PR #2387, opened 2022-09-08
Putorpatchfix

facchettos:
some cleanup of putorpatch and some improvements on api registration 

Fixes

### DIFF
--- a/pkg/frontend/openshiftcluster_putorpatch.go
+++ b/pkg/frontend/openshiftcluster_putorpatch.go
@@ -40,7 +40,7 @@ func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Req
 
 	err := cosmosdb.RetryOnPreconditionFailed(func() error {
 		var err error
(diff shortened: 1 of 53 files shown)


ross-bryan (commented):
I have looked over the PR and it makes sense to me. Didn't have the time to go through it too thoroughly beyond scrolling the changes. I don't think there are any issues being introduced from this PR

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -240,7 +240,7 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 	doc.OpenShiftCluster.Properties.ClusterProfile.PullSecret = ""
 	doc.OpenShiftCluster.Properties.ServicePrincipalProfile.ClientSecret = ""
 
-	b, err := json.MarshalIndent(converter.ToExternal(doc.OpenShiftCluster), "", "    ")
+	b, err := json.MarshalIndent(converter.ToExternal(doc.OpenShiftCluster), "", "\t")
```
ventifus, line 243: I doubt anything cares, but why change the indentation to tabs vs spaces?

ventifus (commented):
lgtm

pkg/frontend/openshiftcluster_putorpatch.go
facchettos: it's just a bit more readable imo , can revert it though, don't really care about it

bennerv (approved):
lgtm - makes sense was thinking the same thing when adding operations to the API `init` functions.

PR #2392, opened 2022-09-13
Add Geneva Metrics to E2E Tests

zacharyljones:

[Geneva Metrics E2E Tests](https://dev.azure.com/msazure/AzureRedHatOpenShift/_boards/board/t/Azure%20Red%20Hat%20OpenShift/Stories/?workitem=15060186)

This PR enables the E2E test suite to emit metrics for each E2E test. This is done by configuring a logrus hook and then using information in the log entry to create a metric.

### Test plan for issue:

This change was tested by...
1. Using a slightly modified version of the `deploy_MDM_VM.sh` script to create a VM running the Geneva MDM agent
2. Using a slightly modified version of the `startMDMNetwork.sh` to open a tunnel to the Geneva MDM VM
3. Running the RP locally using existing shared RP infrastructure running the RP locally - This instance of the RP has the necessary certificates to communicate with the new E2E metrics account in the Test environment.
4. Creating a cluster
5. Running E2E tests against the cluster created
6. Verifying that E2E test metrics are created in the new E2E metrics account

The `deploy_MDM_VM.sh` and `startMDMNetwork.sh` scripts along with documentation related to local MDM testing have been updated to use the latest MDM images, get around issues with simply secure, etc. in #2418.

The ARO wiki also needs to be updated to include the new E2E MDM account. This will also be handled separately.

### DIFF
--- a/pkg/metrics/statsd/e2e/metrics.go
+++ b/pkg/metrics/statsd/e2e/metrics.go
@@ -1,0 +1,56 @@
+package e2e
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/metrics"
+)
+
+const (
+	DimensionARMGeoLocation = "armGeoLocation"
+	DimensionARMResourceID  = "armResourceID"
+	DimensionResourceType   = "resourceType"
+	DimensionSucceeded      = "Succeeded"
+
+	LogEntryIsE2EEmittableMetric = "IsE2EEmittableMetric"
+	LogEntryMetricName           = "MetricName"
+	LogEntryMetricStatus         = "MetricStatus"
+)
+
+type E2ELogToMetrics struct {
+	metricsEmitter metrics.Emitter
+}
+
+func NewE2ELogToMetrics(m metrics.Emitter) E2ELogToMetrics {
+	return E2ELogToMetrics{
+		metricsEmitter: m,
+	}
+}
+
+func (e *E2ELogToMetrics) PostMetricsFromLogEntry(entry *logrus.Entry) {
+	if _, ok := entry.Data[LogEntryIsE2EEmittableMetric]; ok {
+		metricName := fmt.Sprint(entry.Data[LogEntryMetricName])
+		metricStatus := fmt.Sprint(entry.Data[LogEntryMetricStatus])
+		metricStatusInt := btoi(strings.EqualFold(metricStatus, "true"))
+		dimensions := map[string]string{
+			DimensionARMResourceID:  fmt.Sprint(entry.Data[DimensionARMResourceID]),
+			DimensionARMGeoLocation: fmt.Sprint(entry.Data[DimensionARMGeoLocation]),
+			DimensionResourceType:   fmt.Sprint(entry.Data[DimensionResourceType]),
+			DimensionSucceeded:      metricStatus,
+		}
+		e.metricsEmitter.EmitGauge(metricName, metricStatusInt, dimensions)
+	}
+}
+
+func btoi(b bool) int64 {
+	if b {
+		return 1
+	}
+	return 0
+}
--- a/pkg/metrics/statsd/e2e/metrics.go
+++ b/pkg/metrics/statsd/e2e/metrics.go
@@ -16,11 +16,11 @@ const (
 	DimensionARMGeoLocation = "armGeoLocation"
 	DimensionARMResourceID  = "armResourceID"
 	DimensionResourceType   = "resourceType"
-	DimensionSucceeded      = "Succeeded"
+	DimensionSucceeded      = "E2ETestSuccessful"
 
+	LogEntryE2ETestName          = "E2ETestName"
 	LogEntryIsE2EEmittableMetric = "IsE2EEmittableMetric"
-	LogEntryMetricName           = "MetricName"
-	LogEntryMetricStatus         = "MetricStatus"
+	LogEntryIsE2ETestSuccessful  = "IsE2ETestSuccessful"
 )
 
 type E2ELogToMetrics struct {
@@ -35,16 +35,16 @@ func NewE2ELogToMetrics(m metrics.Emitter) E2ELogToMetrics {
 
 func (e *E2ELogToMetrics) PostMetricsFromLogEntry(entry *logrus.Entry) {
 	if _, ok := entry.Data[LogEntryIsE2EEmittableMetric]; ok {
-		metricName := fmt.Sprint(entry.Data[LogEntryMetricName])
-		metricStatus := fmt.Sprint(entry.Data[LogEntryMetricStatus])
-		metricStatusInt := btoi(strings.EqualFold(metricStatus, "true"))
+		metricName := fmt.Sprint(entry.Data[LogEntryE2ETestName])
+		metricIsE2ETestSuccessful := fmt.Sprint(entry.Data[LogEntryIsE2ETestSuccessful])
+		metricIsE2ETestSuccessfulInt := btoi(strings.EqualFold(metricIsE2ETestSuccessful, "true"))
 		dimensions := map[string]string{
 			DimensionARMResourceID:  fmt.Sprint(entry.Data[DimensionARMResourceID]),
 			DimensionARMGeoLocation: fmt.Sprint(entry.Data[DimensionARMGeoLocation]),
 			DimensionResourceType:   fmt.Sprint(entry.Data[DimensionResourceType]),
-			DimensionSucceeded:      metricStatus,
+			DimensionSucceeded:      metricIsE2ETestSuccessful,
 		}
-		e.metricsEmitter.EmitGauge(metricName, metricStatusInt, dimensions)
+		e.metricsEmitter.EmitGauge(metricName, metricIsE2ETestSuccessfulInt, dimensions)
 	}
 }
 
--- a/pkg/util/log/e2e/e2e.go
+++ b/pkg/util/log/e2e/e2e.go
@@ -1,0 +1,30 @@
+package e2e
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/metrics"
+	e2estatsd "github.com/Azure/ARO-RP/pkg/metrics/statsd/e2e"
+)
+
+type e2EHook struct {
+	logToMetrics e2estatsd.E2ELogToMetrics
+}
+
+func NewE2EHook(m metrics.Emitter) e2EHook {
+	return e2EHook{
+		logToMetrics: e2estatsd.NewE2ELogToMetrics(m),
+	}
+}
+
+func (e2EHook) Levels() []logrus.Level {
+	return logrus.AllLevels
+}
+
+func (h *e2EHook) Fire(entry *logrus.Entry) error {
+	h.logToMetrics.PostMetricsFromLogEntry(entry)
+	return nil
+}
--- a/test/e2e/setup.go
+++ b/test/e2e/setup.go
@@ -34,6 +34,10 @@ import (
 	"k8s.io/client-go/tools/clientcmd/api/latest"
 
 	"github.com/Azure/ARO-RP/pkg/env"
+	"github.com/Azure/ARO-RP/pkg/metrics"
+	"github.com/Azure/ARO-RP/pkg/metrics/statsd"
+	e2emetrics "github.com/Azure/ARO-RP/pkg/metrics/statsd/e2e"
+	"github.com/Azure/ARO-RP/pkg/metrics/statsd/golang"
 	aroclient "github.com/Azure/ARO-RP/pkg/operator/clientset/versioned"
 	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/compute"
 	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/features"
@@ -42,6 +46,7 @@ import (
 	redhatopenshift20220401 "github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/redhatopenshift/2022-04-01/redhatopenshift"
 	"github.com/Azure/ARO-RP/pkg/util/cluster"
 	utillog "github.com/Azure/ARO-RP/pkg/util/log"
+	e2elogging "github.com/Azure/ARO-RP/pkg/util/log/e2e"
 	"github.com/Azure/ARO-RP/test/util/kubeadminkubeconfig"
 )
 
@@ -75,8 +80,8 @@ var (
 	vnetResourceGroup string
 	clusterName       string
 	clients           *clientSet
-
-	dockerSucceeded bool
+	dockerSucceeded   bool
+	metricsEmitter    metrics.Emitter
 )
 
 func skipIfNotInDevelopmentEnv() {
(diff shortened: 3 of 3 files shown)


test/e2e/setup.go
```diff
@@ -70,6 +76,7 @@ var (
 	vnetResourceGroup string
 	clusterName       string
 	clients           *clientSet
+	m                 metrics.Emitter
```
dem4gus, line 79: This variable needs a descriptive name as well.

pkg/metrics/statsd/e2e/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	DimensionResourceType   = "resourceType"
+	DimensionSucceeded      = "Succeeded"
+
+	LogEntryIsE2EEmittableMetric = "IsE2EEmittableMetric"
+	LogEntryMetricName           = "MetricName"
+	LogEntryMetricStatus         = "MetricStatus"
+)
+
+type E2ELogToMetrics struct {
+	m metrics.Emitter
```
dem4gus, line 27: Give this a descriptive name instead of a single letter.
```suggestion
	metricsEmitter metrics.Emitter
```

pkg/metrics/statsd/e2e/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+			metricName := fmt.Sprint(entry.Data[LogEntryMetricName])
+			metricStatus := fmt.Sprint(entry.Data[LogEntryMetricStatus])
+			metricStatusInt := Btoi(strings.EqualFold(metricStatus, "true"))
+			dimensions := map[string]string{DimensionARMResourceID: fmt.Sprint(entry.Data[DimensionARMResourceID]), DimensionARMGeoLocation: fmt.Sprint(entry.Data[DimensionARMGeoLocation]), DimensionResourceType: fmt.Sprint(entry.Data[DimensionResourceType]), DimensionSucceeded: metricStatus}
+			e.m.EmitGauge(metricName, metricStatusInt, dimensions)
+		}
+	}
+}
+
+func Btoi(b bool) int64 {
```
dem4gus, line 48: Does this need to be exported?

test/e2e/setup.go
```diff
@@ -245,3 +264,10 @@ var _ = AfterSuite(func() {
 		panic(err)
 	}
 })
+
+var _ = JustAfterEach(func() {
+	gtd := CurrentGinkgoTestDescription()
+	metricName := strings.Replace(gtd.ComponentTexts[1], " ", ".", -1)
+
+	log.WithFields(logrus.Fields{e2emetrics.LogEntryIsE2EEmittableMetric: "true", e2emetrics.LogEntryMetricName: metricName, e2emetrics.LogEntryMetricStatus: !gtd.Failed, e2emetrics.DimensionARMResourceID: resourceIDFromEnv(), e2emetrics.DimensionARMGeoLocation: _env.Location(), e2emetrics.DimensionResourceType: "openShiftClusters"}).Info("Emitting Metric: " + metricName)
```
dem4gus, line 272: This line is long (374 columns) and difficult to read. Can we break it up to make it more scannable:
```suggestion
	log.WithFields(logrus.Fields{
		e2emetrics.LogEntryIsE2EEmittableMetric: "true",
		e2emetrics.LogEntryMetricName:           metricName,
		e2emetrics.LogEntryMetricStatus:         !gtd.Failed,
		e2emetrics.DimensionARMResourceID:       resourceIDFromEnv(),
		e2emetrics.DimensionARMGeoLocation:      _env.Location(),
		e2emetrics.DimensionResourceType:        "openShiftClusters"},
	).Info("Emitting Metric: " + metricName)
```

pkg/metrics/statsd/e2e/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+func (e *E2ELogToMetrics) PostMetricsFromLogEntry(entry *logrus.Entry) {
+	if v, ok := entry.Data[LogEntryIsE2EEmittableMetric]; ok {
+		if strings.EqualFold(fmt.Sprint(v), "true") {
+			metricName := fmt.Sprint(entry.Data[LogEntryMetricName])
+			metricStatus := fmt.Sprint(entry.Data[LogEntryMetricStatus])
+			metricStatusInt := Btoi(strings.EqualFold(metricStatus, "true"))
+			dimensions := map[string]string{DimensionARMResourceID: fmt.Sprint(entry.Data[DimensionARMResourceID]), DimensionARMGeoLocation: fmt.Sprint(entry.Data[DimensionARMGeoLocation]), DimensionResourceType: fmt.Sprint(entry.Data[DimensionResourceType]), DimensionSucceeded: metricStatus}
```
dem4gus, line 42: Break up long line.

test/e2e/setup.go
```diff
@@ -184,6 +191,14 @@ func setup(ctx context.Context) error {
 		return err
 	}
 
+	m = statsd.New(ctx, log.WithField("component", "metrics"), _env, os.Getenv("MDM_E2E_ACCOUNT"), os.Getenv("MDM_E2E_NAMESPACE"), os.Getenv("MDM_E2E_STATSD_SOCKET"))
```
dem4gus, line 194: Break up long line.

test/e2e/setup.go
```diff
@@ -235,6 +250,10 @@ var _ = BeforeSuite(func() {
 
 	if err := setup(context.Background()); err != nil {
 		panic(err)
+	} else {
+		m := statsd.New(context.Background(), log.WithField("component", "e2emetrics"), _env, os.Getenv("MDM_E2E_ACCOUNT"), os.Getenv("MDM_E2E_NAMESPACE"), os.Getenv("MDM_E2E_STATSD_SOCKET"))
```
dem4gus, line 254: Break up long line.

pkg/util/log/e2e/e2e.go
```diff
@@ -0,0 +1,30 @@
...
+	}
+}
+
+func (E2EHook) Levels() []logrus.Level {
+	return logrus.AllLevels
+}
+
+func (h *E2EHook) Fire(entry *logrus.Entry) error {
+	h.logToMetrics.PostMetricsFromLogEntry(entry)
+	return nil
```
dem4gus, line 29: What's the purpose of returning an `error` here if we're only going to be returning `nil`? Should there be some sort of checking?

pkg/metrics/statsd/e2e/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+
+func NewE2ELogToMetrics(m metrics.Emitter) E2ELogToMetrics {
+	return E2ELogToMetrics{
+		m: m,
+	}
+}
+
+func (e *E2ELogToMetrics) PostMetricsFromLogEntry(entry *logrus.Entry) {
+	if v, ok := entry.Data[LogEntryIsE2EEmittableMetric]; ok {
+		if strings.EqualFold(fmt.Sprint(v), "true") {
```
dem4gus, line 38: What behavior should we expect to see if either of the checks in these lines fail?

dem4gus (changes requested):
Good work going on. Some formatting cleanup and questions about error states.

test/e2e/setup.go
zacharyljones: line broken up

test/e2e/setup.go
zacharyljones: Line broken up.

pkg/metrics/statsd/e2e/metrics.go
zacharyljones: Line broken up.

pkg/metrics/statsd/e2e/metrics.go
zacharyljones: No - is there a more appropriate place in the RP for reusable functions like this?

test/e2e/setup.go
zacharyljones: Updated name.

pkg/metrics/statsd/e2e/metrics.go
zacharyljones: Updated name.

pkg/util/log/e2e/e2e.go
zacharyljones: The function definition must match the logrus `Hook` interface - https://github.com/sirupsen/logrus/blob/master/hooks.go

pkg/metrics/statsd/e2e/metrics.go
zacharyljones: If either check fails, then the log entry does not correspond to an e2e test result log entry. Technically the `LogEntryIsE2EEmittableMetric` cannot be set to anything other than `true` if it exists, so the second check can be removed.

zacharyljones: Removed the second `if` statement.

pkg/metrics/statsd/e2e/metrics.go
dem4gus: Not that I know of, but it's pretty common to create small functions like this when you need them. Just change it to lowercase so it's not exported.

zacharyljones: No longer exporting `btoi` function

cadenmarchese (approved):
LGTM, thank you for putting this together.

test/e2e/setup.go
```diff
@@ -377,9 +391,34 @@ var _ = BeforeSuite(func() {
...
+		log.Logger.AddHook(&e2eHook)
 	}
 })
 
 var _ = AfterSuite(func() {
 	log.Info("AfterSuite")
 })
+
+var _ = JustAfterEach(func() {
+	gtd := CurrentGinkgoTestDescription()
```
petrkotas, line 413: What is the thinking behind using already deprecated `struct`?

petrkotas (commented):
Hi @zacharyljones if I got the motivation right, this will allow us to create dashboard in geneva showing us which independent tests are failing?

test/e2e/setup.go
zacharyljones: I was unaware of the deprecation - I will look into changing this.

zacharyljones:
> Hi @zacharyljones if I got the motivation right, this will allow us to create dashboard in geneva showing us which independent tests are failing?

Correct! A much more clear indicator of individual test results, than dgrep, ev2 logs, etc.

pkg/util/log/e2e/e2e.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/metrics"
+	"github.com/Azure/ARO-RP/pkg/metrics/statsd/e2e"
+)
+
+type E2EHook struct {
+	logToMetrics e2e.E2ELogToMetrics
+}
```
facchettos, line 15: this doesn't need to be exported does it?

test/e2e/setup.go
```diff
@@ -377,9 +391,35 @@ var _ = BeforeSuite(func() {
 
 	if err := setup(context.Background()); err != nil {
 		panic(err)
+	} else {
```
facchettos, line 394: the else isn't necessary, since we panic in the block right before

test/e2e/setup.go
```diff
@@ -377,9 +391,35 @@ var _ = BeforeSuite(func() {
...
 	log.Info("AfterSuite")
 })
+
+var _ = JustAfterEach(func() {
+	csr := CurrentSpecReport()
+
+	metricName := strings.Replace(csr.ContainerHierarchyTexts[1], " ", ".", -1)
+
+	log.WithFields(logrus.Fields{
+		e2emetrics.LogEntryIsE2EEmittableMetric: "true",
```
facchettos, line 418: I would rather have the bool `true` here instead

pkg/util/log/e2e/e2e.go
```diff
@@ -0,0 +1,30 @@
...
+
+func NewE2EHook(m metrics.Emitter) E2EHook {
+	return E2EHook{
+		logToMetrics: e2e.NewE2ELogToMetrics(m),
```
facchettos, line 19: use a package alias here, it's really confusing because we are in the e2e package here too

test/e2e/setup.go
```diff
@@ -377,9 +391,35 @@ var _ = BeforeSuite(func() {
 
 	if err := setup(context.Background()); err != nil {
 		panic(err)
+	} else {
+		m := statsd.New(
+			context.Background(),
```
facchettos, line 396: don't we want to have the context passed to the setup function here?

facchettos (changes requested):
I left a few change requests. Also, I can't seem to trigger the ci for your PR because of a merge conflict. Lastly, shouldn't you be a collaborator on the repo so you could trigger the ci yourself?

facchettos:
I have added you as a collaborator so you should be able to run ci yourself now

pkg/util/log/e2e/e2e.go
zacharyljones: No, it doesn't - no longer exporting struct.

test/e2e/setup.go
zacharyljones: removed else statement.

test/e2e/setup.go
zacharyljones: Now using bool `true` instead of string.

pkg/util/log/e2e/e2e.go
zacharyljones: Now using package alias `e2estatsd`.

test/e2e/setup.go
zacharyljones: Moved logrus hook setup into `setup` function so that the same context could be used.

zacharyljones:
> I have added you as a collaborator so you should be able to run ci yourself now (you may need to accept it in the email you received)

Thanks! It will be good to have some CF members as collaborators. I will make a note of this so that when CF members roll off the project we can remove these permissions.

zacharyljones:
Added link to #2418 in PR description - contains documentation updates.

test/e2e/setup.go
bizz001: I haven't been able to find any explicit policy saying that, however, when possible it is better to use flags that represent states, instead of binary flags because they can more easily be extended.

test/e2e/setup.go
```diff
@@ -423,3 +438,17 @@ var _ = AfterSuite(func() {
...
 })
+
+var _ = JustAfterEach(func() {
+	gsc := CurrentSpecReport()
+	metricName := strings.Replace(gsc.ContainerHierarchyTexts[0], " ", ".", -1)
+
+	log.WithFields(logrus.Fields{
+		e2emetrics.LogEntryIsE2EEmittableMetric: true,
+		e2emetrics.LogEntryMetricName:           metricName,
+		e2emetrics.LogEntryMetricStatus:         !gsc.Failed(),
```
bizz001, line 449: On the bool vs state flag, for example It could be more useful to store the actual state here and decide when the metric is consumed if it is good or bad.

pkg/metrics/statsd/e2e/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+
+const (
+	DimensionARMGeoLocation = "armGeoLocation"
+	DimensionARMResourceID  = "armResourceID"
+	DimensionResourceType   = "resourceType"
+	DimensionSucceeded      = "Succeeded"
+
+	LogEntryIsE2EEmittableMetric = "IsE2EEmittableMetric"
+	LogEntryMetricName           = "MetricName"
+	LogEntryMetricStatus         = "MetricStatus"
```
facchettos, line 23:
```suggestion
	LogEntryMetricSuccess         = "MetricSuccess"
```
this or `isSuccessful` or something that can be represented by a bool

test/e2e/setup.go
facchettos, line 449: That's a good point. Here the bool doesn't represent a status. (see my other suggestion)

facchettos (changes requested):

test/e2e/setup.go
zacharyljones: @facchettos - I updated the names (variables, log entry fields, metrics dimensions) to be more representative of state/bool. Is this along the lines of what you were looking for?

s-amann (approved):

petrkotas (approved):
Seems good, we will still need the dashboard created.

jaitaiwan (changes requested):
LGTM, I think it's just a good thing to move that one helper function out.

PR #2394, opened 2022-09-15
Add network contributor to NAT gateway

rhamitarora:
[Work Item- Add Network Contributor permissions on az aro update](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14957600) 
Added network contributor to NAT gateway on az aro update.

### DIFF
--- a/python/az/aro/azext_aro/custom.py
+++ b/python/az/aro/azext_aro/custom.py
@@ -278,10 +278,10 @@ def generate_random_id():
     return random_id
 
 
-def get_route_tables_from_subnets(cli_ctx, subnets):
+def get_network_resources_from_subnets(cli_ctx, subnets):
     network_client = get_mgmt_service_client(cli_ctx, ResourceType.MGMT_NETWORK)
 
-    route_tables = set()
+    subnet_resources = set()
     for sn in subnets:
         sid = parse_resource_id(sn)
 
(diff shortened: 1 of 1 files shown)


python/az/aro/azext_aro/custom.py
```diff
@@ -290,6 +290,23 @@ def get_route_tables_from_subnets(cli_ctx, subnets):
     return route_tables
 
 
+def get_nat_gateways_from_subnets(cli_ctx, subnets):
```
bennerv, line 293: The `get_route_tables_from_subnets` function is doing the exact same thing but checking `subnet.route_table` instead of `subnet.nat_gateway`.  Can we dry this up and have a single function which returns both a set of `route_tables` and `nat_gateways`? 

Could name it `get_network_resources_from_subnets` or something like that.

bennerv (changes requested):
Great work so far!  This is a needed change I've seen us run into because of on-call issues.  Just one change.

bennerv:
@hawkowl do we have an accompanying story to update `pkg/api/validate/dynamic/dynamic.go`? the `validateRouteTablePermissions` could use a `validateNatGatewayPermissions`

python/az/aro/azext_aro/custom.py
rhamitarora: Done

SrinivasAtmakuri:
Thanks for the PR. LGTM.

rhamitarora:
> @hawkowl do we have an accompanying story to update `pkg/api/validate/dynamic/dynamic.go`? the `validateRouteTablePermissions` could use a `validateNatGatewayPermissions`

@hawkowl @bennerv No we don't have any story for the same, we will create in future.

s-amann (approved):

bennerv:
> > @hawkowl do we have an accompanying story to update `pkg/api/validate/dynamic/dynamic.go`? the `validateRouteTablePermissions` could use a `validateNatGatewayPermissions`
> 
> @hawkowl @bennerv No we don't have any story for the same, we will create in future.

I'm fine with the changes as they are, but anything we do from a validation/permissions requirement client-side we should also require server-side.  I think before the next release should this get merged we should validate permissions on NAT gateways so these changes are coupled together.

bennerv (approved):
lgtm

PR #2396, opened 2022-09-15
Remove Proxy CI VMs :boom: 

bennerv:
Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15342132/

With the simply securev2 changes coming, we no longer require the proxy VMSS and as a result can remove it.  Additionally, this cleans up s360 vulnerabilities being reported by the VM and container image so two birds one stone :rocket: 

### Test plan for issue:

Test in the ssv2 world on #2300 . 

Nope.  I need to rip out secrets out of our environment files still though.  

/hold on #2300


bennerv:
This isn't going to pass e2e because the current structure of e2e still relies on the proxy VM.  Will have to test this manually or point to simply secure v2 e2e location.

karanmagdani:
@bennerv should we wait for these changes to be merged before redeploying Infra in our development subscription (@redhat.com) ?

bennerv:
> should we wait for these changes to be merged before redeploying Infra in our development subscription (@redhat.com) ?

yes

PR #2399, opened 2022-09-16
Certificate renewal PUCM type to rotate in-cluster MDSD/ingress/API certs

mhsreddy:

Fixes: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15581636/

During regular PUCM few of the clusters are failing and these clusters are unable to fetch the new certificates from RP as these steps never run after the PUCM failure due to customer policies or other issues. With this PR we can just simply run the steps which will fetch new certificates and install them in the cluster without the need to run all the steps in regular PUCM.

### Test plan for issue:

No Test Cases for this issue. Just re-using the existing functions which already have test cases written.

CURL Command to perform Admin Update just for certificates renewal:
curl -X PATCH -k "https://localhost:8443/subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$RESOURCEGROUP/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER?api-version=admin" --header "Content-Type: application/json" -d '{"properties":{"maintenanceTask": "CertificatesRenewal"}}'

### DIFF
--- a/pkg/api/admin/openshiftcluster_validatestatic.go
+++ b/pkg/api/admin/openshiftcluster_validatestatic.go
@@ -29,7 +29,7 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message)
 	}
 
-	if !(oc.Properties.MaintenanceTask == "" || oc.Properties.MaintenanceTask == MaintenanceTaskEverything || oc.Properties.MaintenanceTask == MaintenanceTaskOperator) {
+	if !(oc.Properties.MaintenanceTask == "" || oc.Properties.MaintenanceTask == MaintenanceTaskEverything || oc.Properties.MaintenanceTask == MaintenanceTaskOperator || oc.Properties.MaintenanceTask == MaintenanceTaskRenewCerts) {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "properties.maintenanceTask", "Invalid enum parameter.")
 	}
 
--- a/pkg/cluster/install.go
+++ b/pkg/cluster/install.go
@@ -37,6 +37,7 @@ func (m *manager) adminUpdate() []steps.Step {
 	task := m.doc.OpenShiftCluster.Properties.MaintenanceTask
 	isEverything := task == api.MaintenanceTaskEverything || task == ""
 	isOperator := task == api.MaintenanceTaskOperator
+	isRenewCerts := task == api.MaintenanceTaskRenewCerts
 
 	// Generic fix-up or setup actions that are fairly safe to always take, and
 	// don't require a running cluster
@@ -91,6 +92,20 @@ func (m *manager) adminUpdate() []steps.Step {
 			steps.Condition(m.aroDeploymentReady, 20*time.Minute, true),
 			steps.Condition(m.ensureAROOperatorRunningDesiredVersion, 5*time.Minute, true),
 		)
+	}
+
+	if isRenewCerts {
+		toRun = append(toRun,
+			steps.Action(m.startVMs),
+			steps.Condition(m.apiServersReady, 30*time.Minute, true),
+			steps.Action(m.fixMCSCert),
+			steps.Action(m.configureAPIServerCertificate),
+			steps.Action(m.configureIngressCertificate),
+			steps.Action(m.initializeOperatorDeployer), // depends on kube clients
+			steps.Action(m.renewMDSDCertificate),
+		)
+		// return early, because we only want to run the certificate actions using the existing operator
+		return toRun
 	}
 
 	// Hive cluster adoption and reconciliation
--- a/pkg/cluster/install.go
+++ b/pkg/cluster/install.go
@@ -96,8 +96,6 @@ func (m *manager) adminUpdate() []steps.Step {
 
 	if isRenewCerts {
 		toRun = append(toRun,
-			steps.Action(m.startVMs),
-			steps.Condition(m.apiServersReady, 30*time.Minute, true),
 			steps.Action(m.fixMCSCert),
 			steps.Action(m.configureAPIServerCertificate),
 			steps.Action(m.configureIngressCertificate),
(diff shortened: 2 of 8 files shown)


ventifus:
The MockOperator in pkg/util/mocks/operator/deploy/deploy.go also needs the new RenewMDSDCertificate function. Run `make generate` to auto-generate them.

s-amann (dismissed):

pkg/cluster/install.go
```diff
@@ -93,6 +94,18 @@ func (m *manager) adminUpdate() []steps.Step {
 		)
 	}
 
+	if isRenewCerts {
```
cblecker, line 97: I think it would be preferable if we followed a similar format to the rest of the ordered list, and only have each action listed in one section.

So for https://github.com/Azure/ARO-RP/blob/0f950d78f14cc4391a90520f12958327482c077d/pkg/cluster/install.go#L70-L84, you'd do something like:
```
	// Requires Kubernetes clients
	if isEverything {
		toRun = append(toRun,
			steps.Action(m.fixSREKubeconfig),
			steps.Action(m.fixUserAdminKubeconfig),
			steps.Action(m.createOrUpdateRouterIPFromCluster),
			steps.Action(m.ensureGatewayUpgrade),
			steps.Action(m.populateRegistryStorageAccountName),
			steps.Action(m.ensureMTUSize),
		)
	}

	// Renew all certificate pieces
	if isEverything || isRenewCerts {
		toRun = append(toRun,
			steps.Action(m.fixMCSCert),
			steps.Action(m.fixMCSUserData),
			steps.Action(m.ensureGatewayUpgrade),
			steps.Action(m.configureAPIServerCertificate),
			steps.Action(m.configureIngressCertificate),
			steps.Action(m.renewMDSDCertificate),
		)
	}
```

pkg/cluster/install.go
```diff
@@ -93,6 +94,18 @@ func (m *manager) adminUpdate() []steps.Step {
...
+		toRun = append(toRun,
+			steps.Action(m.fixMCSCert),
+			steps.Action(m.configureAPIServerCertificate),
+			steps.Action(m.configureIngressCertificate),
+			steps.Action(m.initializeOperatorDeployer), // depends on kube clients
```
cblecker, line 102: Do we need this for cert renewals? I don't see anything that actually uses this client, but maybe I'm missing something.

pkg/cluster/install.go
```diff
@@ -93,6 +94,18 @@ func (m *manager) adminUpdate() []steps.Step {
...
+			steps.Action(m.renewMDSDCertificate),
+		)
+		// return early, because we only want to run the certificate actions using the existing operator
+		return toRun
```
cblecker, line 106: If we use the above pattern, we don't really need to return early.

pkg/api/admin/openshiftcluster_validatestatic.go
```diff
@@ -29,7 +29,7 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message)
 	}
 
-	if !(oc.Properties.MaintenanceTask == "" || oc.Properties.MaintenanceTask == MaintenanceTaskEverything || oc.Properties.MaintenanceTask == MaintenanceTaskOperator) {
+	if !(oc.Properties.MaintenanceTask == "" || oc.Properties.MaintenanceTask == MaintenanceTaskEverything || oc.Properties.MaintenanceTask == MaintenanceTaskOperator || oc.Properties.MaintenanceTask == MaintenanceTaskRenewCerts) {
```
cblecker, line 32: If this list is growing, should we format this more like an enum (with consts and then a slice we loop through)? @ross-bryan thoughts?

cblecker (changes requested):
Looks good, but some comments to make the steps clearer

pkg/cluster/install.go
mhsreddy: The function renewMDSDCertificate is dependant on initializeOperatorDeployer which will initialize the Operator.

pkg/api/admin/openshiftcluster_validatestatic.go
ross-bryan: big +1, this conditional is now up to 4 items, and increases cognitive load and reduces readability in its current state. We could add a helper function to return a list of "valid" `MaintenanceTask`'s and verify that our valid list `contains` _this_ `oc.Properties.MaintenanceTask`.

In go1.18 `slices.Contains` ships and we can stop writing our own helper funcs

pkg/cluster/install.go
mhsreddy: Will remove the step `return toRun` as its not needed in this case.

pkg/api/admin/openshiftcluster_validatestatic.go
mhsreddy: I think we can implement the slices.Contains method once we start using 1.18 version.
`if !slices.Contains([]string{"", MaintenanceTaskEverything, MaintenanceTaskOperator, MaintenanceTaskRenewCerts}, oc.Properties.MaintenanceTask) {
	}`

mhsreddy:
```suggestion
	if !(oc.Properties.MaintenanceTask == "" || oc.Properties.MaintenanceTask == MaintenanceTaskEverything || oc.Properties.MaintenanceTask == MaintenanceTaskOperator || oc.Properties.MaintenanceTask == MaintenanceTaskRenewCerts) {
```

mhsreddy: if !slices.Contains([]string{"", MaintenanceTaskEverything, MaintenanceTaskOperator, MaintenanceTaskRenewCerts}, oc.Properties.MaintenanceTask) {
	}

mhsreddy: if !slices.Contains([]string{"", MaintenanceTaskEverything, MaintenanceTaskOperator, MaintenanceTaskRenewCerts}, oc.Properties.MaintenanceTask) {
	}

s-amann (approved):

SrinivasAtmakuri (approved):

darthhexx (approved):
lgtm

cblecker (approved):
Great work!

PR #2401, opened 2022-09-16
Remove the UniqueKeyPolicy from the Subscriptions database

ventifus:
In CosmosDB, adding unique indexes to existing collections with items is not allowed. 
https://docs.microsoft.com/en-gb/azure/cosmos-db/mongodb/mongodb-indexing#limitations-1

This is causing deployments to fail with an error
```
{"code":"Forbidden","message":"Message: {"Errors":["The unique index cannot be modified. To change the unique index, remove the collection and re-create a new one."]}
```
Reverting for now to unblock deployments, need to regroup later to determine how to implement this change in int and prod.

### DIFF
--- a/pkg/deploy/assets/databases-development.json
+++ b/pkg/deploy/assets/databases-development.json
@@ -225,15 +225,6 @@
                             "/id"
                         ],
                         "kind": "Hash"
-                    },
-                    "uniqueKeyPolicy": {
-                        "uniqueKeys": [
-                            {
-                                "paths": [
-                                    "/key"
-                                ]
-                            }
-                        ]
                     }
                 },
                 "options": {}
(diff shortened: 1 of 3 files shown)


bennerv (approved):
lgtm - I missed this change on review, but should resolve our int deploy issues.  Discussed with @ross-bryan as well and this is the correct change.

ross-bryan (commented):
thanks @ventifus @bennerv for the fix. This should've been in the clustermanager's container not the subscription's. Most likely an oversight while editing `resources_rp.go` or a rebase gone wrong.
Potential follow ups
1. Add "e2e' test coverage for our `deploy`. (this should hold true for all of our components to deploy) would've ideally failed to deploy on the PR introducing this change
2. consider refactoring the resources_rp or breaking out functions into smaller more comprehensible chunks as this file is nearly 2,000 lines itself

PR #2411, opened 2022-09-20
Add unit tests for OCM frontend routes

ventifus:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15529572/

Building on #2332, add relevant test cases.

### Test plan for issue:

Make tests pass.

No

### DIFF
--- a/cmd/aro/rp.go
+++ b/cmd/aro/rp.go
@@ -110,9 +110,12 @@ func rp(ctx context.Context, log, audit *logrus.Entry) error {
 		return err
 	}
 
-	dbClusterManagerConfiguration, err := database.NewClusterManagerConfigurations(ctx, _env.IsLocalDevelopmentMode(), dbc)
-	if err != nil {
-		return err
+	var dbClusterManagerConfiguration database.ClusterManagerConfigurations
+	if os.Getenv("ENABLE_OCM") == "1" {
+		dbClusterManagerConfiguration, err = database.NewClusterManagerConfigurations(ctx, _env.IsLocalDevelopmentMode(), dbc)
+		if err != nil {
+			return err
+		}
 	}
 
 	dbBilling, err := database.NewBilling(ctx, _env.IsLocalDevelopmentMode(), dbc)
--- a/cmd/aro/rp.go
+++ b/cmd/aro/rp.go
@@ -110,12 +110,9 @@ func rp(ctx context.Context, log, audit *logrus.Entry) error {
 		return err
 	}
 
-	var dbClusterManagerConfiguration database.ClusterManagerConfigurations
-	if os.Getenv("ENABLE_OCM") == "1" {
-		dbClusterManagerConfiguration, err = database.NewClusterManagerConfigurations(ctx, _env.IsLocalDevelopmentMode(), dbc)
-		if err != nil {
-			return err
-		}
+	dbClusterManagerConfiguration, err := database.NewClusterManagerConfigurations(ctx, _env.IsLocalDevelopmentMode(), dbc)
+	if err != nil {
+		return err
 	}
 
 	dbBilling, err := database.NewBilling(ctx, _env.IsLocalDevelopmentMode(), dbc)
--- a/pkg/frontend/clustermanager_delete_test.go
+++ b/pkg/frontend/clustermanager_delete_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/metrics/noop"
 	testdatabase "github.com/Azure/ARO-RP/test/database"
 )
@@ -97,7 +98,7 @@ func TestDeleteClusterManagerConfiguration(t *testing.T) {
 		},
 	} {
 		t.Run(tt.name, func(t *testing.T) {
-			ti := newTestInfra(t).WithClusterManagerConfigurations().WithSubscriptions()
+			ti := newTestInfraWithFeatures(t, map[env.Feature]bool{env.FeatureRequireD2sV3Workers: false, env.FeatureDisableReadinessDelay: false, env.FeatureEnableOCMEndpoints: true}).WithClusterManagerConfigurations().WithSubscriptions()
 			defer ti.done()
 
 			resourceKey := fmt.Sprintf("/subscriptions/%s/resourcegroups/resourcegroup/providers/microsoft.redhatopenshift/openshiftclusters/resourcename/%s/%s",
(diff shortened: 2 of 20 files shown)


ross-bryan (commented):
a test case to consider (and make code changes for) - is the API version correct? these resources are only supported on 2022-09-04 (or greater), how do we know if an api version supports the resources? you could check api version, but that can be messy, I propose something like this
```
diff --git a/pkg/frontend/clustermanager_putorpatch.go b/pkg/frontend/clustermanager_putorpatch.go
index fc1e6509e..bc8ea873f 100644
--- a/pkg/frontend/clustermanager_putorpatch.go
+++ b/pkg/frontend/clustermanager_putorpatch.go
@@ -45,6 +45,9 @@ func (f *frontend) _putOrPatchClusterManagerConfiguration(ctx context.Context, l
        systemData, _ := r.Context().Value(middleware.ContextKeySystemData).(*api.SystemData) // don't panic
        vars := mux.Vars(r)
 
+       if converter == nil {
+               return []byte(""), api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", "Invalid API Version. Please use 2022-09-04 or newer.")
+       }
```
we would need this for all ocm related routes/http verbs we support

pkg/frontend/clustermanager_get.go
```diff
@@ -34,12 +34,14 @@ func (f *frontend) _getClusterManagerConfiguration(ctx context.Context, log *log
...
 	doc, err := f.dbClusterManagerConfiguration.Get(ctx, r.URL.Path)
-	switch {
-	case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
-		return nil, api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, "", "The Resource '%s/%s/%s/%s' under resource group '%s' was not found.",
-			vars["resourceType"], vars["resourceName"], vars["ocmResourceType"], vars["ocmResourceName"], vars["resourceGroupName"])
-	case err != nil:
-		return nil, err
+	if err != nil {
+		switch {
+		case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
```
cadenmarchese, line 39: Would an http 401 or 403 be possible here as well?

pkg/frontend/clustermanager_get_test.go
```diff
@@ -0,0 +1,153 @@
...
+		ocmResourceType string
+		ocmResourceName string
+		clusterName     string
+		apiVersion      string
+		fixture         func(*testdatabase.Fixture, *test, string)
+		wantStatusCode  int
+		wantResponse    *v20220904.ClusterManagerConfiguration
+		wantError       string
+	}
+	createSingleDocument := func(f *testdatabase.Fixture, tt *test, resourceKey string) {
```
cadenmarchese, line 35: Could we separate these `createSingleDocument` vars out from the individual test functions and re-use them?

cadenmarchese (commented):
Looking good, thanks! Few questions + suggestions on my first pass.

pkg/frontend/clustermanager_get.go
ventifus: In those cases the caller would get a generic error (500?) rather than a `CloudErrorCodeResourceNotFound`.

pkg/frontend/clustermanager_delete_test.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+func TestDeleteClusterManagerConfiguration(t *testing.T) {
+	ctx := context.Background()
+
+	mockSubscriptionId := "00000000-0000-0000-0000-000000000000"
+	resourcePayload := "eyAKICAiYXBpVmVyc2lvbiI6ICJoaXZlLm9wZW5zaGlmdC5pby92MSIsCiAgImtpbmQiOiAiU3luY1NldCIsCiAgIm1ldGFkYXRhIjogewogICAgIm5hbWUiOiAic2FtcGxlIiwKICAgICJuYW1lc3BhY2UiOiAiYXJvLWY2MGFlOGEyLWJjYTEtNDk4Ny05MDU2LWYyZjZhMTgzN2NhYSIKICB9LAogICJzcGVjIjogewogICAgImNsdXN0ZXJEZXBsb3ltZW50UmVmcyI6IFtdLAogICAgInJlc291cmNlcyI6IFsKICAgICAgewogICAgICAgICJhcGlWZXJzaW9uIjogInYxIiwKICAgICAgICAia2luZCI6ICJDb25maWdNYXAiLAogICAgICAgICJtZXRhZGF0YSI6IHsKICAgICAgICAgICJuYW1lIjogIm15Y29uZmlnbWFwIgogICAgICAgIH0KICAgICAgfQogICAgXQogIH0KfQo="
```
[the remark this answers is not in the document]
ventifus: done

pkg/frontend/clustermanager_get.go
ross-bryan: our web server wouldn't return a 401/403 at this point in the code, ARM handles the authentication and authorization, and we have middleware that performs some authorization as well, see https://github.com/Azure/ARO-RP/blob/8fc2fddea4eee96c8ff05730368cd05292e128eb/pkg/frontend/middleware/authenticated.go#L30-L32

pkg/frontend/clustermanager_get_test.go
ross-bryan: it seems like a DRY principal we could leverage here :+1:

ross-bryan (dismissed):
this lgtm, thanks @ventifus! I see Caden had some comments regarding DRY-ing the code, but I am willing to approve as is. The refactor can be done in this PR or in a follow up :)

cmd/aro/rp.go
```diff
@@ -110,9 +110,12 @@ func rp(ctx context.Context, log, audit *logrus.Entry) error {
 		return err
 	}
 
-	dbClusterManagerConfiguration, err := database.NewClusterManagerConfigurations(ctx, _env.IsLocalDevelopmentMode(), dbc)
-	if err != nil {
-		return err
+	var dbClusterManagerConfiguration database.ClusterManagerConfigurations
+	if os.Getenv("ENABLE_OCM") == "1" {
```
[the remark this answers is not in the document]
ventifus: How about a _real_ feature flag :smile: . We can now enable this on a per-region per-rp basis by configuring it in (int|prod)-config.yml
```
rpFeatures:
    - EnableOCMEndpoints
```

ross-bryan (approved):
lgtm, thanks for implementing a real feature toggle :+1: :1st_place_medal:

cadenmarchese (approved):

PR #2413, opened 2022-09-21
Expose aro operator version

ventifus:
Currently `oc get co` doesn't list a version for the aro operator. This change shows the version as `1.0-<commithash>`. Might save new people some confusion over why one operator doesn't have a version number.

### DIFF
--- a/pkg/operator/controllers/clusteroperatoraro/clusteroperatoraro_controller.go
+++ b/pkg/operator/controllers/clusteroperatoraro/clusteroperatoraro_controller.go
@@ -5,6 +5,7 @@ package clusteroperatoraro
 
 import (
 	"context"
+	"fmt"
 
 	configv1 "github.com/openshift/api/config/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
@@ -23,6 +24,7 @@ import (
 
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 	aroclient "github.com/Azure/ARO-RP/pkg/operator/clientset/versioned"
+	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


SrinivasAtmakuri:
Thanks for the PR, LGTM.

Please note, with PR: https://github.com/Azure/ARO-RP/pull/2331/ the ARO Operator deployment version is detached from using version.GitCommit and we use the below conditions to set the version tag, Maybe we need something similar? (So that the deployment version and operator versions don't differ and cause confusion)

ARO_IMAGE set & no tag set -> version label should be "latest"
ARO_IMAGE set & a tag set -> version label should be tag
OperatorVersion set & overridden -> version label should be "operatorVersion"

ventifus:
@srinivasAtmakuri Thanks for the pointer, I will look into it.

facchettos:
When we deploy the operator, we have the version available, and it would be easy to set it to env variable in `pkg/operator/deploy/` , then we won't have to care about the special cases because the deployment code already took care of it for us

pkg/operator/controllers/clusteroperatoraro/clusteroperatoraro_controller.go
```diff
@@ -156,6 +158,12 @@ func (r *Reconciler) defaultOperator() *configv1.ClusterOperator {
 			Name: clusterOperatorName,
 		},
 		Status: configv1.ClusterOperatorStatus{
+			Versions: []configv1.OperandVersion{
+				{
+					Name:    "operator",
+					Version: fmt.Sprintf("1.0-%s", version.GitCommit),
```
cadenmarchese, line 164: Is there a reason we use a prefix, or could we get away with just the commit here?
```suggestion
					Version: version.GitCommit,
```

ventifus:
Revisiting this.
Even in the case where we're overriding the OperatorVersion in at deployment time, when the aro operator controller actually executes in the cluster it is the desired version. This means the operator controller can use it's own `version.GitCommit`, there's no need to expose OperatorVersion separately.

ventifus:
For your test, did you follow [operator/README.md](https://github.com/Azure/ARO-RP/blob/master/pkg/operator/README.md)? Here's what I get:

```shell
$ oc get co
NAME                                       VERSION   AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
aro                                        70b48f0   True        False         False      2m43s   
authentication                             4.10.20   True        False         False      3m32s
```

```shell
$ oc get po -n openshift-azure-operator -o jsonpath="{.items[0].spec.containers[0].image}"
quay.io/rh_ee_adenton/aro:70b48f0
```

You can use my image for `$ARO_IMAGE` if you'd like.

cadenmarchese (approved):

PR #2420, opened 2022-09-26
Add SecretVersion to OCP Cluster Document

SrinivasAtmakuri:
We currently require rotation of encryption keys every 90 days. With this PR, we have a way to distinguish which secret version is used to encrypt each cluster document, so that we can determine if/when encryption secrets can be removed.

WorkItem: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/12715836

Fixes: 12715836

### Test plan for issue:

1. Create a cluster
2. AdminUpdate an existing cluster without the feature
3. AdminUpdate a cluster with the feature already existing.

Tries to re-implement changes from here:
https://github.com/Azure/ARO-RP/compare/master...bennerv:encryption-autorotation-first-attempt?expand=1

### DIFF
--- a/pkg/database/test/openshiftclusters_test.go
+++ b/pkg/database/test/openshiftclusters_test.go
@@ -1,0 +1,128 @@
+package test
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database"
+	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
+	testdatabase "github.com/Azure/ARO-RP/test/database"
+)
+
+func TestSecretVersion(t *testing.T) {
+	ctx := context.Background()
+	mockSubID := "00000000-0000-0000-0000-000000000000"
+
+	aead := testdatabase.NewFakeAEAD("newversion")
+
+	testCases := []struct {
+		desc        string
+		doc         *api.OpenShiftClusterDocument
+		expectedDoc *api.OpenShiftClusterDocument
+	}{
+		{
+			desc: "secretversion is set",
+			doc: &api.OpenShiftClusterDocument{
+				ID:           "00000000-2222-0000-0000-000000000000",
+				PartitionKey: mockSubID,
+				Key:          strings.ToLower(testdatabase.GetResourcePath(mockSubID, "resourceName")),
+				OpenShiftCluster: &api.OpenShiftCluster{
+					ID:   testdatabase.GetResourcePath(mockSubID, "resourceName"),
+					Name: "resourceName",
+					Type: "Microsoft.RedHatOpenShift/openshiftClusters",
+					Properties: api.OpenShiftClusterProperties{
+						KubeadminPassword: api.SecureString("FAKE123"),
+					},
+				},
+			},
+			expectedDoc: &api.OpenShiftClusterDocument{
+				ID:            "00000000-2222-0000-0000-000000000000",
+				Key:           strings.ToLower(testdatabase.GetResourcePath(mockSubID, "resourceName")),
+				PartitionKey:  mockSubID,
+				SecretVersion: "newversion",
+				OpenShiftCluster: &api.OpenShiftCluster{
+					ID:   testdatabase.GetResourcePath(mockSubID, "resourceName"),
+					Name: "resourceName",
+					Type: "Microsoft.RedHatOpenShift/openshiftClusters",
+					Properties: api.OpenShiftClusterProperties{
+						KubeadminPassword: api.SecureString("FAKE123"),
+						ProvisionedBy:     "new",
+					},
+				},
+			},
+		},
+		{
+			desc: "existing secret version is updated",
+			doc: &api.OpenShiftClusterDocument{
+				ID:            "00000000-2222-0000-0000-000000000000",
+				PartitionKey:  mockSubID,
+				Key:           strings.ToLower(testdatabase.GetResourcePath(mockSubID, "resourceName")),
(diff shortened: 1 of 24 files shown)


SrinivasAtmakuri:
Testing:

1. Create a local dev cluster., and verify if the cluster object has the SecertVersion set, [DONE]
2. A cluster which doesn't have secretVersion previously set, and on AdminUpdating, the cluster document should get secretVerion, [DONE]

```
curl -X GET -k "https://localhost:8443/subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$RESOURCEGROUP/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER?api-version=admin" --header "Content-Type: application/json" -d "{}" | grep secretVersion
        "secretVersion": "8d19c3001e1041579be0738680ee01e5"
```

SrinivasAtmakuri:
@bennerv, using your changes, I have made this draft PR and also added details on the testing and result, may you please review.

pkg/util/keyvault/keyvault.go
```diff
@@ -196,7 +196,7 @@ func (m *manager) GetCertificateSecret(ctx context.Context, secretName string) (
 	return key, certs, nil
 }
 
-func (m *manager) GetSecret(ctx context.Context, secretName string) (azkeyvault.SecretBundle, error) {
+func (m *manager) GetSecret(ctx context.Context, secretName string, version string) (azkeyvault.SecretBundle, error) {
```
s-amann, line 199: Will `version` be used at a later time? why add this now?

pkg/util/encryption/xchacha20poly1305.go
```diff
@@ -53,3 +53,7 @@ func (c *xChaCha20Poly1305) Seal(input []byte) ([]byte, error) {
 
 	return append(nonce, c.aead.Seal(nil, nonce, input, nil)...), nil
 }
+
+func (c *xChaCha20Poly1305) SealSecretVersion() string {
+	return ""
```
s-amann, line 58: This is my first glance at the encryption package, why does this return `""`?

hawkowl: @SrinivasAtmakuri This probably needs a comment. I'm guessing that it's blank because it will never have a secret version?

pkg/util/keyvault/keyvault.go
hawkowl: @SrinivasAtmakuri it seems this isn't used?

pkg/database/test/openshiftclusters_test.go
```diff
@@ -0,0 +1,131 @@
...
+					Properties: api.OpenShiftClusterProperties{
+						KubeadminPassword: api.SecureString("FAKE123"),
+						ProvisionedBy:     "new",
+					},
+				},
+			},
+		},
+	}
+	for _, tC := range testCases {
+		t.Run(tC.desc, func(t *testing.T) {
```
jaitaiwan, line 93: Love this!

pkg/database/test/openshiftclusters_test.go
```diff
@@ -0,0 +1,131 @@
...
+			jsonHandle, err := database.NewJSONHandle(aead)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			for _, isPatch := range []bool{false, true} {
+				db, client := testdatabase.NewFakeOpenShiftClustersWithProvidedJSONHandle(jsonHandle, aead)
```
jaitaiwan, line 100: Could this also be inside another `t.Run` and be run `t.Parallel`?

pkg/util/keyvault/keyvault.go
s-amann: should line 200 be `m.kv.GetSecret(ctx, m.keyvaultURI, secretName, version)`?

petrkotas:
Hi @SrinivasAtmakuri I see there are unadressed comments. Would you please take a look?

pkg/util/keyvault/keyvault.go
SrinivasAtmakuri: Made the change.

pkg/util/keyvault/keyvault.go
```diff
@@ -196,7 +196,7 @@ func (m *manager) GetCertificateSecret(ctx context.Context, secretName string) (
...
 	return m.kv.GetSecret(ctx, m.keyvaultURI, secretName, "")
```
petrkotas, line 200:
```suggestion
return m.kv.GetSecret(ctx, m.keyvaultURI, secretName, version)
```

pkg/util/encryption/xchacha20poly1305.go
petrkotas: So this puzzles me too.

pkg/util/encryption/multi.go
```diff
@@ -28,7 +36,8 @@ func NewMulti(ctx context.Context, serviceKeyvault keyvault.Manager, secretName,
 	}
 
 	m := &multi{
-		sealer: aead,
+		sealer:      aead,
+		sealVersion: filepath.Base(*secret.ID),
```
petrkotas, line 40: To be honest I am lost with how this PR should work. 
Srinivas would you please provide a detailed explanation of the logical flow and how this is supposed to work?

petrkotas (changes requested):
I have a questions about this PR. I see it adding secret version, but it is never used. Morover I see `sealversion` created based on the secret.id.
I believe the whole PR deserves detailed explanation what problem it solves and how it is solving it.

SrinivasAtmakuri:

PR #2422, opened 2022-09-27
[wip] Keep this e2e

ventifus:
keep the e2e test cluster.

### DIFF
--- a/.pipelines/e2e.yml
+++ b/.pipelines/e2e.yml
@@ -55,5 +55,6 @@ jobs:
       register_sub
 
       export CI=true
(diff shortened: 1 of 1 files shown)


.pipelines/e2e.yml
```diff
@@ -55,5 +55,6 @@ jobs:
       register_sub
 
       export CI=true
+      export E2E_DELETE_CLUSTER=false
```
ross-bryan, line 58: wondering if we should document the use of this env var outside of it just existing within `test/e2e/setup.go` and having to determine where to plumb it in every time :thinking:

PR #2423, opened 2022-09-27
Adjust comment formatting to appease newer golangci-lint versions

ventifus:

Fixes https://coreos.slack.com/archives/C02ULBRS68M/p1664313661925289

Current golangci-lint (I tested 1.49.0) enforces some go1.19 rules on comment formatting. I have used go1.19.1's `gofmt -s` to fix those comments that golangci-lint complains about. Here's an example of the kind of lint errors I see, although this process had to be repeated on other files that golangci-lint complains about once these first three are gofmt'ed.

```
$ make lint-go
hack/lint-go.sh
WARN [runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused. 
pkg/util/pullsecret/pullsecret.go:23: File is not `gofmt`-ed with `-s` (gofmt)
//   {"auths": {"secret key": {"auth": "secret value"}, "secret key": {"auth": "secret value"}}}
pkg/util/status/clusterversionoperator_status.go:12: File is not `gofmt`-ed with `-s` (gofmt)
//TODO: this is duplicate from clusterversioncondition.go
pkg/operator/controllers/subnets/subnets_controller.go:71: File is not `gofmt`-ed with `-s` (gofmt)
//Reconcile fixes the Network Security Groups
make: *** [Makefile:184: lint-go] Error 1
```

### Test plan for issue:

Comment-only change

### DIFF
--- a/pkg/monitor/cluster/machineconfigpool.go
+++ b/pkg/monitor/cluster/machineconfigpool.go
@@ -41,11 +41,9 @@ func (mon *Monitor) getNodeCounts(ctx context.Context) (int64, error) {
 	return int64(len(ns.Items)), nil
 }
 
-/***
-Count the number of nodes available
-Total the nodes under machineconfigpool control
-Alert if different
-*/
+// Count the number of nodes available
+// Total the nodes under machineconfigpool control
+// Alert if different
 func (mon *Monitor) emitMachineConfigPoolUnmanagedNodeCounts(ctx context.Context) error {
 	mcpcount, err := mon.getMachineConfigPoolNodeCounts(ctx)
 	if err != nil {
--- a/hack/licenses/main.go
+++ b/hack/licenses/main.go
@@ -146,7 +146,7 @@ func applyPythonLicense() error {
 	})
 }
 
-//returns the lists of files that don't have a license but should
+// returns the lists of files that don't have a license but should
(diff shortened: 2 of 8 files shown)


facchettos (dismissed):
We should probably also bump the version of golangci-lint.

pkg/monitor/cluster/machineconfigpool.go
```diff
@@ -41,7 +41,8 @@ func (mon *Monitor) getNodeCounts(ctx context.Context) (int64, error) {
 	return int64(len(ns.Items)), nil
 }
 
-/***
+/*
+**
```
SudoBrendan, line 45: nit: just remove `**`?

ross-bryan (approved):
LGTM!

SudoBrendan (dismissed):
LGTM one nit.

pkg/monitor/cluster/machineconfigpool.go
ventifus: Or switch to line comments since this is the only place that uses block comments (besides a few `doc.go` files).

facchettos (approved):

cblecker (approved):

PR #2424, opened 2022-09-28
az CLI changes required for the latest OpenShiftVersion API version and multi-version functionality

darthhexx:

Fixes the ability to install multi-version OCP clusters and the `az aro get-versions` output.

### Test plan for issue:

Tested multi-version cluster creation using the az CLI.

Nope.


m1kola (approved):
Looks good to me, but I'm not super familiar with multi-version support. Might be a good idea to get some more eyes on it.

PR #2425, opened 2022-09-28
Fix InternalServerError for decoding outdated securecookie

darthhexx:

Fixes: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15664001

Instead of responding with an HTTP 500 when a cookie decode error is encountered, we should be clearing the cookie and returning an HTTP 403. This PR changes the behaviour, so that when we are loading the SRE portal in a secret rotated region that we were previously logged into we are presented with a login dialog instead of an internal server error.

### Test plan for issue:

Local testing and existing, slightly modified, unit tests.

Nope.

### DIFF
--- a/pkg/portal/middleware/aad.go
+++ b/pkg/portal/middleware/aad.go
@@ -142,7 +142,8 @@ func (a *aad) AAD(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		session, err := a.store.Get(r, SessionName)
 		if err != nil {
-			cookieError := err.(securecookie.Error)
+                        cookieError, ok := err.(securecookie.Error)
+                        if ok && cookieError != nil && cookieError.IsDecode() {
 			if cookieError != nil && cookieError.IsDecode() {
 				cookie := &http.Cookie{
 					Name:    SessionName,
--- a/pkg/portal/middleware/aad.go
+++ b/pkg/portal/middleware/aad.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/gorilla/mux"
+	"github.com/gorilla/securecookie"
 	"github.com/gorilla/sessions"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/oauth2"
@@ -141,7 +142,18 @@ func (a *aad) AAD(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		session, err := a.store.Get(r, SessionName)
 		if err != nil {
-			a.internalServerError(w, err)
+			cookieError := err.(securecookie.Error)
+			if cookieError != nil && cookieError.IsDecode() {
+				cookie := &http.Cookie{
+					Name:    SessionName,
+					Path:    "/",
+					Expires: time.Now().Add(-1),
+				}
+				http.SetCookie(w, cookie)
+				http.Error(w, http.StatusText(http.StatusForbidden), http.StatusForbidden)
+			} else {
+				a.internalServerError(w, err)
+			}
 			return
 		}
 
--- a/pkg/portal/middleware/aad.go
+++ b/pkg/portal/middleware/aad.go
@@ -147,7 +147,7 @@ func (a *aad) AAD(h http.Handler) http.Handler {
 				cookie := &http.Cookie{
 					Name:    SessionName,
 					Path:    "/",
-					Expires: time.Now().Add(-1),
+					Expires: time.Unix(0, 0),
 				}
 				http.SetCookie(w, cookie)
 				http.Error(w, http.StatusText(http.StatusForbidden), http.StatusForbidden)
--- a/pkg/portal/middleware/aad.go
+++ b/pkg/portal/middleware/aad.go
@@ -142,9 +142,8 @@ func (a *aad) AAD(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		session, err := a.store.Get(r, SessionName)
 		if err != nil {
-                        cookieError, ok := err.(securecookie.Error)
-                        if ok && cookieError != nil && cookieError.IsDecode() {
-			if cookieError != nil && cookieError.IsDecode() {
+			cookieError, ok := err.(securecookie.Error)
+			if ok && cookieError != nil && cookieError.IsDecode() {
 				cookie := &http.Cookie{
 					Name:    SessionName,
 					Path:    "/",
--- a/pkg/portal/middleware/aad_test.go
+++ b/pkg/portal/middleware/aad_test.go
@@ -137,7 +137,7 @@ func TestAAD(t *testing.T) {
 					},
 				}, nil
 			},
-			wantStatusCode: http.StatusInternalServerError,
+			wantStatusCode: http.StatusForbidden,
 		},
 	} {
 		t.Run(tt.name, func(t *testing.T) {


pkg/portal/middleware/aad.go
```diff
@@ -141,7 +142,18 @@ func (a *aad) AAD(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		session, err := a.store.Get(r, SessionName)
 		if err != nil {
-			a.internalServerError(w, err)
+			cookieError := err.(securecookie.Error)
+			if cookieError != nil && cookieError.IsDecode() {
+				cookie := &http.Cookie{
+					Name:    SessionName,
+					Path:    "/",
+					Expires: time.Now().Add(-1),
```
facchettos, line 150: It would be a better to use time.Unix(0,0) here. If there is time drift between the client and the server this may not work.

facchettos (changes requested):

pkg/portal/middleware/aad.go
darthhexx: Good call, done.

pkg/portal/middleware/aad.go
```diff
@@ -141,7 +142,18 @@ func (a *aad) AAD(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		session, err := a.store.Get(r, SessionName)
 		if err != nil {
-			a.internalServerError(w, err)
+			cookieError := err.(securecookie.Error)
```
ross-bryan, line 145: should we handle the type assertion? otherwise the portal could panic here?
something like this perhaps: 
```suggestion
                        cookieError, ok := err.(securecookie.Error)
                        if ok && cookieError != nil && cookieError.IsDecode() {
```

ross-bryan (changes requested):
lgtm but wondering if we should more cautiously handle the type assertion

pkg/portal/middleware/aad.go
facchettos: wouldn't 
```suggestion
			cookieError, ok  := err.(securecookie.Error)
			if !ok {
			    a.internalServerError(w, err)
			    return
			}
```
be more explicit ?

pkg/portal/middleware/aad.go
```diff
@@ -141,7 +142,18 @@ func (a *aad) AAD(h http.Handler) http.Handler {
...
+				cookie := &http.Cookie{
+					Name:    SessionName,
+					Path:    "/",
+					Expires: time.Unix(0, 0),
+				}
+				http.SetCookie(w, cookie)
+				http.Error(w, http.StatusText(http.StatusForbidden), http.StatusForbidden)
+			} else {
+				a.internalServerError(w, err)
+			}
```
facchettos, line 156: maybe we could also get rid of the else with a `return` after line 153. The else is not necessary here

darthhexx: I agree on the general practice of not introducing unnecessary else's, but I am not sure how we can get rid of the `else` in this instance. We need the `a.internalServerError()` to catch all errors that are not a `cookieError.IsDecode()` error, with the return preventing any of the rest of the function from running.

facchettos:
```

			cookieError, ok := err.(securecookie.Error)
			if ok && cookieError != nil && cookieError.IsDecode() {
				cookie := &http.Cookie{
					Name:    SessionName,
					Path:    "/",
					Expires: time.Unix(0, 0),
				}
				http.SetCookie(w, cookie)
				http.Error(w, http.StatusText(http.StatusForbidden), http.StatusForbidden)
				return
			} 
			a.internalServerError(w, err)
			return
```

facchettos: this is not what duplication is. Also , it may have 2 returns but is has less nesting which reduces complexity.

pkg/portal/middleware/aad.go
```diff
@@ -141,7 +142,18 @@ func (a *aad) AAD(h http.Handler) http.Handler {
...
```
AldoFusterTurpin, line 156: Some interesting discussion here 🙂

I agree with @facchettos about the deletion of the "else" block whenever possible. I know this can seem like a "own" school or type of doing things but in Go this is a common thing and almost a standard. Even in [Effective Go](https://go.dev/doc/effective_go#control-structures) there is a section where they say: 

"In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted." ... " the resulting code needs no else statements." 

And they provide an example:
```go
f, err := os.Open(name)
if err != nil {
    return err
}
d, err := f.Stat()
if err != nil {
    f.Close()
    return err
}
codeUsing(f, d)
```

So, in the code in this PR, I agree with the already proposed alternatives, but we can simplify it a bit. Note that we already have a function a.internalServerError(w, err) with http.ResponseWriter and error as parameters, in that order.

This is my suggestion: 

- create a new method (handleErr) to encapsulate the handling of the error to simplify AAD public method.
- invert the if condition (in handleErr) to first check if there is an error so the code inside the if block is just two lines. 

Something like that:

```go
// AAD is the early stage handler which adds a username to the context if it
// can.  It lets the request through regardless (this is so that failures can be
// logged).
func (a *aad) AAD(h http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		session, err := a.store.Get(r, SessionName)
		if err != nil {
			a.handleErr(w, err)
			return
		}

		expires, ok := session.Values[SessionKeyExpires].(time.Time)
		if !ok || expires.Before(a.now()) {
			h.ServeHTTP(w, r)
			return
		}

		ctx := r.Context()
		ctx = context.WithValue(ctx, ContextKeyUsername, session.Values[SessionKeyUsername])
		ctx = context.WithValue(ctx, ContextKeyGroups, session.Values[SessionKeyGroups])
		r = r.WithContext(ctx)

		h.ServeHTTP(w, r)
	})
}

func (a *aad) handleErr(w http.ResponseWriter, err error) {
	cookieError, ok := err.(securecookie.Error)

	isCookieError := ok && cookieError != nil && cookieError.IsDecode()
	if !isCookieError {
		a.internalServerError(w, err)
		return
	}

	cookie := &http.Cookie{
		Name:    SessionName,
		Path:    "/",
		Expires: time.Unix(0, 0),
	}
	http.SetCookie(w, cookie)
	http.Error(w, http.StatusText(http.StatusForbidden), http.StatusForbidden)
}
```

AldoFusterTurpin (changes requested):
@darthhexx Great. 👌
But having a public method returning an anonymous function that contains an "if-else" inside an "if" is probably too much nesting in Go.

I propose to delete the else and follow the Go convention of returning as soon as possible. Take a look at my proposed suggestion in the case is useful.

Thanks 🙂

pkg/portal/middleware/aad.go
AldoFusterTurpin: I 100% agree with the removal of the "else", take a look at my review if you wish.

pkg/portal/middleware/aad.go
s-amann: A smaller change to remove the else:

```go
                session, err := a.store.Get(r, SessionName)
		if err != nil {
			cookieError, ok := err.(securecookie.Error)
			if ok && cookieError != nil && cookieError.IsDecode() {
				cookie := &http.Cookie{
					Name:    SessionName,
					Path:    "/",
					Expires: time.Unix(0, 0),
				}
				http.SetCookie(w, cookie)
				http.Error(w, http.StatusText(http.StatusForbidden), http.StatusForbidden)
				return
			} 
			a.internalServerError(w, err)
			return
		}

**Review by s-amann — APPROVED** [review]
lgtm!

File: pkg/portal/middleware/aad.go (hunk 3 above)
Comment (darthhexx) [reply pkg/portal/middleware/aad.go]: I prefer the first option in cases when the condition follows logically to the else and the number of lines is low, because the logic this simple to parse.

The example you are using is contrived though, because its not a conditional on the error output. It is on whether to _log an error_ **or** _perform an action_. Obviously that reads better in the second example.

The code in this PR is conceptually doing this and it reads easily this way:
```go
if error-is-a-particular-type {
    handleIt()
} else {
    handleItDifferently()
}
return // to prevent the rest of the code running
```
This is a well practised approach to handling this kind of logic flow and does not benefit from the early-return approach to reduce nesting.

pkg/portal/middleware/aad.go
darthhexx: This is starting to feel a lot like an interrogation by the Else-Police :D

Thanks for the feedback @AldoFusterTurpin. I do think that adding a new function adds complexity that isn't necessary for this simple change. Also note that you quoted "when an if statement doesn't flow into the next statement" as a predicate and motivation for changing the code in this PR. However, as I mentioned above, the code does flow logically to the else because it is conditioning on the **type** of error we are handling.

In the Effective Go documentation they are describing, "The code reads well if the successful flow of control runs down the page, eliminating error cases as they arise. Since error cases tend to end in return statements".

"eliminating error cases as they arise" - in the code we are handling the **same error** conditionally.
"Since error cases tend to end in return statements" - this is why there is a return statement directly after we have conditionally handled the **same error**.

AldoFusterTurpin: @darthhexx 🚓🚨👮‍♂️ "Why aren't you using guard clauses?"

Jokes apart 🙂, there is nothing wrong with discussing different ways to achieve the same thing as we all can learn from each other. 

I like your argument provided in [mentioned above](https://github.com/Azure/ARO-RP/pull/2425#discussion_r984139453) but I don't see the benefits of it over the early return (that's usually used in Go). 

Effective Go which I have mentioned talks about a particular case, but there is nothing wrong to apply the same principle here. Even at [The Zen of Go](https://the-zen-of-go.netlify.app/) we can see "Return early rather than nesting deeply"-> "Rather than nesting deeply, keep the success path to the left using guard clauses." as general advice.

I will try to provide a few "objective" arguments to point out why I think the version with an auxiliary function is an improvement that we can apply:

Introducing an auxiliary function doesn't add complexity here. A function should do just one thing and do it well, and creating the auxiliary function improves the cohesion of that block because you are encapsulating the logic to handle the cookie error in a separate block of code (and thanks to that it lets you have just 2 lines in the body in the if).

Benefits: 
- We are reducing the length of AAD.
- We group together everything needed to decide if we have a cookie error or an internal server error in the same place.
- When reading AAD method, I can understand what it does without going into the details about how the error is handled as I don't care at that point (that is the point of composition at the end of the day).
- If new errors need to be handled or cookie struct changes, I don't need to touch AAD method.

I will give you a more practical example. Imagine that now you need to add more fields to the cookie struct, you will see that with the auxiliary function, the piece of code you are touching is narrower than the original, and this is a big benefit as the mental overhead to understand the context is less than if you had a bigger function.

Now imagine that apart from having to check against a "err.(securecookie.Error)" error, you need to check a couple more errors that can occur. Where do you want to handle those errors? 
- In a function that has the only purpose of handling errors, or 
- in a function that is responsible for doing a lot of things + handling errors? 
I honestly will prefer to have a short, simple function to handle those errors to not mix AAD logic with error handling and error construction logic. 

When we talk about creating an aux function and making the code return as soon as possible is all about readability and maintainability.

I just wanted to provide my 5 cents about the implications/benefits of the suggestion. If you disagree I would approve anyway as there is nothing wrong. 🙂

darthhexx: With regards to the Zen of Go, the premise is: _"Return early rather than nesting deeply"_. Th logic in this PR does not introduce any deep nesting with it's single else. It isn't even nested by the definition of nesting.

The next premise for your example states, "Imagine that now you need to add more fields to the cookie struct". The struct is in an external Gorilla module, so that would never happen and is purely an academic example. However, the next statement does have some merit to it, in that if we _"need to check a couple more errors that can occur"_ then I would agree that implementing a function would better encapsulate the logic of checking and branching on the different error scenarios.

That said, I don't agree that we are at that level of complexity with the current narrow scope of the changes in this PR. There are 2 logic branches to handle and that suites, and could even be the definition of, an if-else approach.

AldoFusterTurpin: Perfect. Thanks for the response. 🙂👌

AldoFusterTurpin (approved):
LGTM!

bennerv (approved):

pkg/portal/middleware/aad.go
devamanv: @darthhexx  I had a chance to take a look at the comments. Some good discussion there. From my experience, it's always a good practice to eliminate unnecessary `else` s. Moreover, it's also considered a linting issue. One question though. I immediately noticed a return statement, which was also pointed by @facchettos  as well. Why do we even need that return statement at the end, couldn't we just return from the if block to stop further execution and wouldn't the inside function return any way when it has executed a.internalServerError(w, err) . Unless, the last return statement stops the outside function from executing further, which it doesn't, I think we should be able to get rid of the last return statement, move it inside if and remove the else.

PR #2427, opened 2022-09-28
Implement baseline functionality of validating OCM APIs 

s-amann:

[15529556 Build Authentication with allowed ClientIDs for OCM Routes](https://msazure.visualstudio.com/AzureRedHatOpenShift/_boards/board/t/Azure%20Red%20Hat%20OpenShift/Stories/?workitem=15529556)

We will eventually have a list of IDs in RP-Config / LiveConfig to be pulled in via the liveConfig manager. This implementation will simply reject any ocm request until their ID is populated. Additional validation is done to ensure the api is called only for specific resourceTypes.
### Test plan for issue:

Unit tests, can deploy to INT given that the rp-config is setup to inject the env var `OCM_VALID_CLIENT_IDS` into the RP during deploy (via ev2).

Unsure if anything related to OCM APIs has been documented, will need to ensure the api feature as a whole includes how we are authorizing the requests.


pkg/util/ocm/validator.go
```diff
@@ -0,0 +1,45 @@
...
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+)
+
+var ValidOCMResourceTypes = []string{"syncsets", "secrets", "machinepools", "syncidentityproviders"}
+
+func ValidateOCMResourceType(resourceType string) bool {
```
cadenmarchese, line 17: I think https://github.com/Azure/ARO-RP/pull/2446 will give us good coverage for this case. Do you think it's still needed?

cadenmarchese (commented):
Nice, thanks for your work on this.

pkg/util/ocm/validator.go
s-amann: Yes that validation is great. I can update this PR to remove this extra step.

s-amann: @cadenmarchese I removed this extra validation of the `ocmResourceType`

petrkotas (commented):
Thank you for the work on this PR. I have a clarification question.

I see the authentication verifies the usrID in the header. So I am a bit puzzled, where does the rest of authentication and authorization works? Is this left for the ARM which will use the AAD and we count on the request comming from ARM to be verified?

s-amann:
> 

The added code will only 'authorize' the clientID in the header against a known set of valid IDs AFTER the main authorization occurs as we only validate the clientID on a subset of API calls. The authorization of the request as a whole (as it happens for most of the other API calls into the RP) happens when the appropriate `clientAuthorizer` has its `isAuthorized()` method called. This can be seen on [line 30 of authenticated.go](https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/middleware/authenticated.go#L30)

pkg/util/liveconfig/hive.go
```diff
@@ -116,3 +117,17 @@ func (p *prod) AdoptByHive(ctx context.Context) (bool, error) {
 	}
 	return false, nil
 }
+
+func (p *prod) OCMValidClientIDs() ([]string, error) {
+	// TODO: Replace with RP Live Service Config (KeyVault)
+	ocmValidClientIDString := os.Getenv(ocmValidClientIDs)
```
cadenmarchese, line 123: Consider a comment here on where we get/set ocmValidClientIDs and how to add new ones

pkg/util/liveconfig/dev.go
```diff
@@ -64,10 +65,23 @@ func (d *dev) DefaultInstallerPullSpecOverride(ctx context.Context) string {
...
-func (p *dev) AdoptByHive(ctx context.Context) (bool, error) {
+func (d *dev) AdoptByHive(ctx context.Context) (bool, error) {
 	adopt := os.Getenv(hiveAdoptEnableEnvVar)
 	if adopt != "" {
 		return true, nil
 	}
 	return false, nil
 }
+
+func (d *dev) OCMValidClientIDs() ([]string, error) {
```
cadenmarchese, line 76: This looks like the same function as in prod. It might be worth defining this function separately from the methods and then calling the same function in both dev and prod.

cadenmarchese (commented):
Some small comments but overall this LGTM

pkg/frontend/middleware/authenticated.go
```diff
@@ -32,6 +32,13 @@ func Authenticated(env env.Interface) func(http.Handler) http.Handler {
 				return
 			}
 
+			if _, isOCMAPI := vars["ocmResourceType"]; isOCMAPI {
```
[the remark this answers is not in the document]
s-amann: Agreed, I'll make the change

pkg/util/liveconfig/hive.go
s-amann: I can add a comment on the format the IDs will need to be in in RP-Config. However, I don't know how we will be obtaining these IDs from OCM.

pkg/util/liveconfig/dev.go
s-amann: It is only the same function for now. Once we introduce proper live config Prod will be different from Dev in that Dev will continue to pull from the ENV and Prod will pull from a KeyVault. I'll extract it for now to reduce the duplicated code.

pkg/util/liveconfig/hive.go
s-amann: Going through this we will have to make some additional changes to this branch so that we can properly pipe in configuration values in RP-Config into the environment of the RP

pkg/util/liveconfig/hive.go
```diff
@@ -116,3 +116,10 @@ func (p *prod) AdoptByHive(ctx context.Context) (bool, error) {
 	}
 	return false, nil
 }
+
+func (p *prod) OCMValidClientIDs() ([]string, error) {
+	// TODO: Replace with RP Live Service Config (KeyVault)
+	// until then, IDs can be added to RP-Config `configuration.validOCMClientIDs: '["id1", "id2"]'
```
s-amann, line 122: I'd like some input here. We ultimately want to be able to support multiple valid OCM clientIDs. The way this is currently implemented we would have one ENV variable set to a bash array `["id1", "id2"]` but I'm not sure if this will work having to convert the RP-Config from a yaml to json and then into the Configuration struct. 

Do we have a better way to accomplish this goal today?

s-amann:
added hold label as I would like to test the env piping code in INT.

pkg/frontend/middleware/authenticated_test.go
```diff
@@ -0,0 +1,108 @@
...
+			name:             "non ocm api called, system data header is not validated",
+			method:           "GET",
+			ocmResourceType:  "",
+			systemDataHeader: `{"systemData":{"lastModifiedBy":"unused"}}`,
+			wantStatus:       200,
+			isValid:          true,
+			mocks: func(_env *mock_env.MockInterface, req http.Request, isValid bool) {
+				_env.EXPECT().ArmClientAuthorizer().Return(clientauthorizer.NewAll())
+				_env.EXPECT().ValidateOCMClientID(req.Header.Get(ArmSystemDataHeaderKey)).Times(0)
+			},
```
facchettos, line 52: these funcs as variables aren't great. We could maybe just get away with having a `expectCalled` `expectArg` and `return` args so we would not have duplication of `_env.EXPECT().ArmClientAuthorizer().Return(clientauthorizer.NewAll())` and no func vars

facchettos (dismissed):

pkg/frontend/middleware/authenticated_test.go
s-amann: yeah I can refactor the test to remove the function variables

pkg/deploy/devconfig.go
```diff
@@ -197,6 +197,7 @@ func DevConfig(_env env.Core) (*Config, error) {
 			InstallViaHive:           to.StringPtr(os.Getenv("ARO_INSTALL_VIA_HIVE")),
 			DefaultInstallerPullspec: to.StringPtr(os.Getenv("ARO_HIVE_DEFAULT_INSTALLER_PULLSPEC")),
 			AdoptByHive:              to.StringPtr(os.Getenv("ARO_ADOPT_BY_HIVE")),
+			ValidOCMClientIDs:        to.StringPtr(os.Getenv("ARO_VALID_OCM_CLIENT_IDS")),
```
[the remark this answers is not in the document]
s-amann: I like that idea, might be good to get it in before liveconfig is fully implemented as that should remove the need for the env variables here.

pkg/frontend/middleware/authenticated.go
```diff
@@ -28,6 +30,13 @@ func Authenticated(env env.Interface) func(http.Handler) http.Handler {
 				api.WriteError(w, http.StatusForbidden, api.CloudErrorCodeForbidden, "", "Forbidden.")
 				return
 			}
+			vars := mux.Vars(r)
+			if _, isOcmResourceType := vars["ocmResourceType"]; isOcmResourceType {
+				if systemDataHeader := r.Header.Get(ArmSystemDataHeaderKey); !env.ValidateOCMClientID(systemDataHeader) {
+					api.WriteError(w, http.StatusForbidden, api.CloudErrorCodeForbidden, "", "Forbidden.")
+					return
+				}
+			}
```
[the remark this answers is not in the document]
facchettos: you are correct, I haven't modified authenticated in my pr ... I do too many things at once during oncall...

but yeah I think that would make sense to have it in a different file (and use the pattern from 2679 to)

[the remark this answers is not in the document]
s-amann: I'll get that changed in a new commit then, I've pushed the other changes.

pkg/frontend/middleware/validateocmclient.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/gorilla/mux"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/env"
+)
+
+type OCMValidator struct {
+	Env env.Interface
+}
```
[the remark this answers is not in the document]
s-amann: Having the env passed in at this point did make it easier to inject the validator into the creation of a frontend. With this change to only pass in the IDs the liveConfig needs to be retrieved at instantiation time of the frontend which ends up blowing up every test with `NewFrontend` :(.

I don't think that means we should inject the env, but it does have some benefits with how the codebase is currently structured.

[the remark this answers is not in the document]
facchettos: ... could we not have it as a parameter but just extract it from the env passed in the frontend creation? (genuine question, need to leave early tonight, will take a deeper look tomorrow)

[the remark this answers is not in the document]
s-amann: I was able to work it out, commit pushed. If you think there is an easier/better way I'm all :ear:s

PR #2428, opened 2022-09-29
add color back into e2es, easier to see pass/fails in ADO

ross-bryan:

Fixes ADO e2e tests not displaying color, it had bad output in ev2 which is why it was removed, but e2e now uses its own parameters see https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines?path=/ev2/Deployment/ServiceGroupRoot/bin/rpe2e.sh

Makes viewing e2e failures in ADO much easier. 

### Test plan for issue:
:fire: 

no


bennerv (approved):
lgtm - no more black and white

EV2 e2e will still not have color :rocket: https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines?path=/ev2/Deployment/ServiceGroupRoot/bin/rpe2e.sh

PR #2429, opened 2022-09-30
Validate perms req create cluster

carlowisse:
ADO Card: [14618792](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14618792)

Adds: `az` command to validate permissions: `az aro validate-permissions --resource-group MyResourceGroup --vnet MyVnet --master-subnet MyMasterSubnet --worker-subnet MyWorkerSubnet`
Allows users to validate their resource permissions before creating a cluster.

### Test plan for issue:
- Integration tests added
- General CI

Help documentation has been added as part of the command help flag: `az aro validate-permissions -h`


bennerv (commented):
Couple questions:

1. Based on the user story description, we should be validating permissions only, not assigning them.  Based on the current implementation, this calls `ensure_resource_permissions` which will create role assignments.  Is this the intention?
1. If a service principal isn't passed in during validation, it will create one.  If we're just doing validation, it's most likely going to create the service principal + permissions only to be thrown away as that info is not provided to the user.

I think we need to break this up a bit:
1. Validate the FP SP has Network Contributor over the vnet, route tables (if applicable), and NAT gateways (coming soon in another PR if applicable)
1. Break this up into two further use cases for the cluster service principal
    * Cluster already exists:
        * If cluster already exists, query the service principal and check proper permissions on it
    * Cluster does not yet exist:
        * If a cluster service principal is passed in, then validate it has the proper permissions 
        * If the cluster service principal is __not__ passed in, then do nothing

I think we should decouple the validation code path from the create code path because customers could be running this post-creation (such as before an az aro update) or pre-creation.  

Curious on others thoughts. 

**Final Thing**
Technically, the person running the command may not have the permissions to create the service principal, query the resources, and/or create role assignments.  That could be another validation someone might care about, but that doesn't sound like the scope of the story.

PR #2430, opened 2022-09-30
Splits operator E2E tests into steps

m1kola:

Part of [WI 13667437](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13667437).

Splits ARO operator E2Es into steps. This is helpful from several perspectives:
* Documenting what we do in our E2E tests
* Makes it bit easier to narrow down the area of failure

### Test plan for issue:

CI runs E2E

No, just E2E improvements

### DIFF
--- a/test/e2e/operator.go
+++ b/test/e2e/operator.go
@@ -79,7 +79,7 @@ func dumpEvents(ctx context.Context, namespace string) error {
 var _ = Describe("ARO Operator - Internet checking", func() {
 	var originalURLs []string
 	BeforeEach(func() {
-		// save the originalURLs
+		By("saving the original URLs")
 		co, err := clients.AROClusters.AroV1alpha1().Clusters().Get(context.Background(), "cluster", metav1.GetOptions{})
 		if kerrors.IsNotFound(err) {
 			Skip("skipping tests as aro-operator is not deployed")
@@ -89,7 +89,7 @@ var _ = Describe("ARO Operator - Internet checking", func() {
 		originalURLs = co.Spec.InternetChecker.URLs
 	})
 	AfterEach(func() {
-		// set the URLs back again
+		By("restoring the original URLs")
 		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 			co, err := clients.AROClusters.AroV1alpha1().Clusters().Get(context.Background(), "cluster", metav1.GetOptions{})
 			if err != nil {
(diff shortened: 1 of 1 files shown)


test/e2e/operator.go
```diff
@@ -226,22 +230,22 @@ var _ = Describe("ARO Operator - Cluster Monitoring ConfigMap", func() {
...
+		By("deleting for the ConfigMap")
 		err = clients.Kubernetes.CoreV1().ConfigMaps("openshift-monitoring").Delete(context.Background(), "cluster-monitoring-config", metav1.DeleteOptions{})
 		Expect(err).NotTo(HaveOccurred())
 
+		By("waiting for the ConfigMap to make sure it was restored")
 		err = wait.PollImmediate(30*time.Second, 15*time.Minute, configMapExists)
 		Expect(err).NotTo(HaveOccurred())
-
-		_, err = clients.Kubernetes.CoreV1().ConfigMaps("openshift-monitoring").Get(context.Background(), "cluster-monitoring-config", metav1.GetOptions{})
-		Expect(err).NotTo(HaveOccurred())
```
m1kola, line 239: This get doesn't make much sense since we are doing exactly the same thing in `PollImmediate` just above. `wait.PollImmediate` will error on timeout. Also we should probably look into replacing `wait.PollImmediate` with ginkgo's `Eventually`.

test/e2e/operator.go
```diff
@@ -250,22 +254,22 @@ var _ = Describe("ARO Operator - RBAC", func() {
...
+		By("deleting for the ClusterRole")
 		err = clients.Kubernetes.RbacV1().ClusterRoles().Delete(context.Background(), "system:aro-sre", metav1.DeleteOptions{})
 		Expect(err).NotTo(HaveOccurred())
 
+		By("waiting for the ClusterRole to make sure it was restored")
 		err = wait.PollImmediate(30*time.Second, 15*time.Minute, clusterRoleExists)
 		Expect(err).NotTo(HaveOccurred())
-
-		_, err = clients.Kubernetes.RbacV1().ClusterRoles().Get(context.Background(), "system:aro-sre", metav1.GetOptions{})
-		Expect(err).NotTo(HaveOccurred())
```
m1kola, line 263: This get doesn't make much sense since we are doing exactly the same thing in `PollImmediate` just above. `wait.PollImmediate` will error on timeout. Also we should probably look into replacing `wait.PollImmediate` with ginkgo's `Eventually`.

test/e2e/operator.go
```diff
@@ -402,13 +412,15 @@ var _ = Describe("ARO Operator - MUO Deployment", func() {
 })
 
 var _ = Describe("ARO Operator - MHC Deployment", func() {
-	Specify("MHC should be enabled and managed by default", func() {
+	It("must be deployed and managed by default", func() {
```
m1kola, line 415: I think this test (and probably some other "by default" tests) are a bit odd. I think it should be an unit test, not an e2e.

I decided not to touch it for now: just add splitting into steps.

s-amann (approved):
lgtm!

test/e2e/operator.go
cadenmarchese: +1 on `Eventually` everywhere.

cadenmarchese (approved):
Nice, looking at [the e2e run]( https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=61171769&view=logs&j=10f66467-59e5-53bc-0c8a-42cc89770de4&t=3bbe7e73-d181-5aca-4b77-91acc0ef0f4a) I like how the tests now provide more information and are more logically written within the testing framework.

test/e2e/operator.go
bennerv: I agree.  Unless we're testing the actual reconciliation of a controller (creating, deleting, updating resources) by changing the flags, we shouldn't have these.  

If you want to remove them in another PR, be my guest :rocket:

bennerv (approved):
lgtm - nice work

PR #2432, opened 2022-09-30
Merge existing operator flags with defaults

bennerv:

Fixes TBD

There has been in the past a situation where a cluster was missing operatorflags for one of our controllers because.  This can happen under one of the following circumstances:

1. We release a new operator flag
1. We don't PUCM an existing cluster with the new operator flag
1. Time passes

1) Before operatorflags, there was nothing
2) We PUCM'd an existing cluster without operatorflags with an operatorflag payload
3) That cluster now only has the operatorflags specified within the above payload an no defaults

This should prevent any above circumstance from happening.  

Ideally during a release with a new operator flag this is what we'll do:
1. Create a new operator flag, but don't set it in `DefaultOperatorFlags`
2. PUCM a bunch of clusters to confirm it works
3. Next release, set the default
4. PUCM down the road
5. Any clusters between steps 2 and 3 that were created will now get the added `DefaultOperatorFlag` on next PUCM

In the future with LiveConfig (there's a few variations, but I think the below explains the gist of it)
1. The RP will have a set of defaultOperatorFlags we can modify for cluster creation at any time
2. We will release the RP without the default set
3. We will test across the fleet to ensure it works as expected
4. We will modify the defaultOperatorFlags liveconfig to set it
5. We can PUCM without setting the flag and not worry. 

### Test plan for issue:

Unit tests

No

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -36,9 +36,18 @@ func SetDefaults(doc *OpenShiftClusterDocument) {
 				doc.OpenShiftCluster.Properties.MaintenanceTask = MaintenanceTaskEverything
 			}
 		}
+
 		// If there's no operator flags, set the default ones
 		if doc.OpenShiftCluster.Properties.OperatorFlags == nil {
 			doc.OpenShiftCluster.Properties.OperatorFlags = DefaultOperatorFlags()
+		} else {
+			// Merge operator flags with existing ones
+			defaultOperatorFlags := DefaultOperatorFlags()
+			for k, v := range defaultOperatorFlags {
+				if _, ok := doc.OpenShiftCluster.Properties.OperatorFlags[k]; !ok {
+					doc.OpenShiftCluster.Properties.OperatorFlags[k] = v
+				}
+			}
 		}
 	}
 }
--- a/pkg/api/defaults_test.go
+++ b/pkg/api/defaults_test.go
@@ -31,9 +31,9 @@ func validOpenShiftClusterDocument() *OpenShiftClusterDocument {
 			},
 		},
 	}
-
 	return &doc
 }
+
 func TestSetDefaults(t *testing.T) {
 	for _, tt := range []struct {
 		name  string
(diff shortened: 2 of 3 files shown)


bennerv:
@hawkowl I'd like your input on this change.  I think it's a net-positive and will help us not missing anything and reconcile potentially missing operator flags.

hawkowl:
@bennerv without reading the changes since I'm on hols -- not merging with defaults is on purpose since then it can inadvertently turn on changes on existing clusters for something that we want to turn on by default for new ones (e.g. guardrails may have different options initially for fresh clusters, so the flag being missing meaning off = desired)

There's a Q4 toil item for Skippy to implement the ability to merge flags though that's written in the operator flags design document, which ideally does what I think you're doing here.

hawkowl:
@bennerv that is, merging a new set of values on top of defaults to maintain snowflakes -- for which a straight up reset wouldn't contain a body

bennerv:
@hawkowl I think we need some clarification on how we work with operator flags and releases.

Ideally we'll have a consistent fleet with all operatorFlags that are set in `DefaultOperatorFlags` set within all cluster documents, even if values are toggled on or off.   They would be toggled the opposite of default under the idea that an SRE did this on purpose.

IMO, when releasing a new controller, we shouldn't set the values in DefaultOperatorFlags because right now (without liveconfig to update the DefaultOperatorFlags dynamically for a region), clusters may get these controllers enabled out of the box, it could be broken, and we need a hotfix rollout.  We've had this issue multiple times in the past.

So under the above assumption, when we roll out a new controller, we'll start with the first SDP region, PUCM all clusters with the flag explicitly, and test it on existing clusters.  Once we're confident with the flag being enabled across regions, we'll update the `DefaultOperatorFlags` for the next release and set the default values.  

Any clusters created between the PUCM finishing and before the next release will be missing the flags, and we need to come back and enable them.  This will be accomplished by merging the `DefaultOperatorFlags` with the existing `OperatorFlags` on a cluster document during the next round of PUCM.  This will ensure we don't miss any opeartor flag enablement (which again we have in the past) for some clusters. 

If this isn't the way we're currently doing things, it would be great to ensure we're all on the same page.

pkg/api/defaults.go
```diff
@@ -36,9 +36,18 @@ func SetDefaults(doc *OpenShiftClusterDocument) {
 				doc.OpenShiftCluster.Properties.MaintenanceTask = MaintenanceTaskEverything
 			}
 		}
+
 		// If there's no operator flags, set the default ones
 		if doc.OpenShiftCluster.Properties.OperatorFlags == nil {
 			doc.OpenShiftCluster.Properties.OperatorFlags = DefaultOperatorFlags()
+		} else {
```
s-amann, line 43: instead of adding an else can we return early within the `if`?

pkg/api/defaults.go
```diff
@@ -36,9 +36,18 @@ func SetDefaults(doc *OpenShiftClusterDocument) {
...
+			// Merge operator flags with existing ones
+			defaultOperatorFlags := DefaultOperatorFlags()
```
cadenmarchese, line 45: Do we need this var?

pkg/api/defaults_test.go
```diff
@@ -116,14 +116,18 @@ func TestSetDefaults(t *testing.T) {
 			},
 		},
 		{
-			name: "preserve flags",
+			// TODO: Once the operatorflags are changed from strings to consts, we can correctly
+			// modify a single flag, until then it's hardcoded
+			name: "merge flags",
```
cadenmarchese, line 121: Do we want to test the other case as well (`doc.OpenShiftCluster.Properties.OperatorFlags == nil`, set them)?

bennerv:
Sounds like we're going a different route.  Closing

PR #2433, opened 2022-09-30
Test hive CreateNamespace in e2e

yehornaumenko:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15583619/

E2E tests for hive clusterManager in pkg/hive/manager.go

No

### Tests 
Tested manually. Checked that namespaces have been created and deleted.

### DIFF
--- a/test/e2e/hive_manager.go
+++ b/test/e2e/hive_manager.go
@@ -1,0 +1,42 @@
+package e2e
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Hive manager creates a namespace", func() {
+
+	var ns *corev1.Namespace
+
+	AfterEach(func() {
+		if ns != nil {
+			Eventually(func() error {
+				return clients.HiveAKS.CoreV1().Namespaces().Delete(context.Background(), ns.Name, metav1.DeleteOptions{})
+			}).WithTimeout(20 * time.Second).WithPolling(1 * time.Second).Should(Succeed())
+		}
+	})
+
+	It("Should be created successfully", func() {
+		var err error
+		ns, err = clients.HiveClusterManager.CreateNamespace(context.Background())
+		Expect(err).NotTo(HaveOccurred())
+		Expect(ns).NotTo(BeNil())
+
+		res, err := clients.HiveAKS.CoreV1().Namespaces().Get(context.Background(), ns.Name, metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(res).NotTo(BeNil())
+
+		Expect(res.String()).To(Equal(ns.String()))
+	})
+
+})
--- a/test/e2e/aks.go
+++ b/test/e2e/aks.go
@@ -16,24 +16,11 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
-
-	"github.com/Azure/ARO-RP/pkg/util/liveconfig"
 )
 
 // Tests the kubeconfig ability to get and manipulate the cluster
 var _ = Describe("AKS cluster present", Pending, func() {
 	ctx := context.Background()
-	var liveConfig liveconfig.Manager
-	var kubeConfig *rest.Config
-
-	BeforeEach(func() {
-		var err error
-		liveConfig, err = _env.NewLiveConfigManager(ctx)
-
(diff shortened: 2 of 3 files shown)


test/e2e/hive_manager.go
```diff
@@ -0,0 +1,42 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Hive manager", func() {
+	Describe("CreateNamespace", func() {
+		var ns *corev1.Namespace
```
facchettos, line 19: maybe instead of having two describes you could just have something like this 
```suggestion
var _ = Describe("Hive manager creates a namespace", func() {
		var ns *corev1.Namespace
```

facchettos (changes requested):

test/e2e/hive_manager.go
```diff
@@ -0,0 +1,42 @@
...
+		It("Should be created successfully", func() {
+			var err error
+			ns, err = clients.HiveClusterManager.CreateNamespace(context.Background())
+			Expect(err).NotTo(HaveOccurred())
+			Expect(ns).NotTo(BeNil())
+
+			res, err := clients.HiveAKS.CoreV1().Namespaces().Get(context.Background(), ns.Name, metav1.GetOptions{})
```
facchettos, line 26: would it make sense to have a timeout on that context so we can be sure that the test will finish in a specific amount of time ?

yehornaumenko: Yeah, it makes sense to have time limits for tests. As I understand, there are some ways to set timeout via ginkgo – `SpecTimeout` and `NodeTimeout`. But these things are available only in the latest versions of ginkgo.

Now I am wondering why we don't have timeouts for k8s calls in all other places.

Yeah, we can set a timeout locally, using `context.WithTimeout`, but actually it is not the only option we have.
Also, we can set it once while initiating the k8s client in the e2e `setup()` function.

@m1kola could you share your thoughts, please?

m1kola: We limit e2e total test execution time to 180 minutes [here](https://github.com/Azure/ARO-RP/blob/4240ea46193a8a72d7275ecddcfb66406640e866/Makefile#L5). Ginkgo will fail the job if exeed 180 min.

> But these things are available only in the latest versions of ginkgo.

We already have [version of Ginkgo which supports this](https://github.com/Azure/ARO-RP/pull/2503), but I don't think that setting a custom timeout on a spec or node will be beneficial: global timeout seems to be enough for me. Or am I missing something?

But we must get rid of creating own context for each call and use context provided by a node/spec. I updated all of our existing tests to do that in [this PR](https://github.com/Azure/ARO-RP/pull/2505).

test/e2e/hive_manager.go
```diff
@@ -0,0 +1,42 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(res).NotTo(BeNil())
+
+			Expect(res.String()).To(Equal(ns.String()))
+		})
+
+		AfterEach(func() {
```
m1kola, line 33: Could you please put `BeforeEach` and `AfterEach` before `It` nodes? `AfterEach` is like a `defer` call: should be close to resource creation.

test/e2e/hive_manager.go
```diff
@@ -0,0 +1,42 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Hive manager", func() {
+	Describe("CreateNamespace", func() {
+		var ns *corev1.Namespace
+		It("Should be created successfully", func() {
+			var err error
+			ns, err = clients.HiveClusterManager.CreateNamespace(context.Background())
```
m1kola, line 22: I do not see a benefit of this test: we already have an unit test for `CreateNamespace` which is sufficent for ensuring that this method works as expected.

We should not be testing metohds of the hive manager separately in E2Es: we need to test whether the whole flow of cluster creation/adoption works end to end. From customer's request to successfull cluster creation/adoption via Hive.

m1kola (changes requested):
We should not be testing metohds of the hive manager separately in E2Es: we need to test whether the whole flow of cluster creation/adoption works end to end. From customer's request to successfull cluster creation/adoption via Hive.

test/e2e/hive_manager.go
yehornaumenko: @m1kola I entirely agree that this test is almost useless in terms of coverage of Hive-related processes.
But I consider this as something that proves that our e2e infra is ready for Hive testing to some extent.

test/e2e/hive_manager.go
yehornaumenko: @m1kola but if some test is stuck for some reason, we have to wait for ~180min to see it failed, right? does it make sense? It is actually the only case I wanna avoid by setting a timeout on node/spec.

facchettos (dismissed):

petrkotas (approved):
Thank you @yehornaumenko for the PR. It lays down the base for additional Hive tests perfectly.

facchettos (approved):

PR #2434, opened 2022-09-30
workaround - FPSP Role Assignment Doesn't Have Get

bennerv:
Don't "GET" a resource group unless it already exists.  

Fixes TODO

https://github.com/Azure/ARO-RP/pull/2328/files#diff-fc2ecaea9c3fb03887bc89a5290dd452dc919a651168e73a3b3a83a25d8ff525R48

### Test plan for issue:

Test in prod because that's the only place to accurately test this.  We also need to PUCM an existing cluster to ensure we can still "GET" an existing resource group.

### DIFF
--- a/pkg/cluster/deploystorage.go
+++ b/pkg/cluster/deploystorage.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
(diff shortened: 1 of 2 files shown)


s-amann (approved):
lgtm!

ross-bryan (approved):
ick. this is not ideal, but it should work until we can fix the READ permissions

rogbas (approved):

PR #2435, opened 2022-09-30
Stylistic: makes e2e flags consistent

m1kola:
Changes nothing: only adds missing `-` (apperently ginkgo works with single - in flags as well) and re-orders flags so that `go test` flags are grouped together and `ginkgo` flags are together.


bennerv (approved):
:shipit:

m1kola:
For historic reasons: it is not just stylistic: it actually fixes ginkgo timeout: previously it was using go test timeouts, but now has own timeout flag.

PR #2436, opened 2022-10-03
Removes e2e which duplicates unit test coverage

m1kola:

Part of [WI 13667437](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13667437).

Follow up from https://github.com/Azure/ARO-RP/pull/2430#discussion_r984763943: I think we already have unit test coverage for this in [`pkg/api/defaults_test.go`](https://github.com/Azure/ARO-RP/blob/f7895a143c348cd16e83043d14429504c29f1a13/pkg/api/defaults_test.go).

### Test plan for issue:

Run E2E

No, just a E2E clean up

### DIFF
--- a/test/e2e/operator.go
+++ b/test/e2e/operator.go
@@ -5,11 +5,9 @@ package e2e
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"regexp"
 	"sort"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cblecker (commented):
Yeah it looks like we have a unit test for ensuring default flags are set in the cluster document: https://github.com/Azure/ARO-RP/blob/f7895a143c348cd16e83043d14429504c29f1a13/pkg/api/defaults_test.go#L110-L117

I'm wondering if it's useful to have some generic test coverage somewhere to ensure that the flags from the cluster doc actually make their way into the actual CR on the cluster.

bennerv (approved):
LGTM

This should check the existence of the resources the MHC controller deploys rather than just checking if it's enabled by default right?   That would be an actual e2e test to confirm the resources were successfully created and there wasn't a conflict.

m1kola:
> I'm wondering if it's useful to have some generic test coverage somewhere to ensure that the flags from the cluster doc actually make their way into the actual CR on the cluster.

We update flags in CR from the RP [here](https://github.com/Azure/ARO-RP/blob/f7895a143c348cd16e83043d14429504c29f1a13/pkg/operator/deploy/deploy.go#L248-L249): the test will end up being a copy paste of all flags. I think it won't be very useful.

> This should check the existence of the resources the MHC controller deploys rather than just checking if it's enabled by default right? That would be an actual e2e test to confirm the resources were successfully created and there wasn't a conflict.

Something along these lines, yes. I created [this work item](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15689695) to make sure that HMC resources get re-created after deletion. I think this should be enough.

m1kola:
> I'm wondering if it's useful to have some generic test coverage somewhere to ensure that the flags from the cluster doc actually make their way into the actual CR on the cluster.

@cblecker we update flags in CR from the RP [here](https://github.com/Azure/ARO-RP/blob/f7895a143c348cd16e83043d14429504c29f1a13/pkg/operator/deploy/deploy.go#L248-L249): the test will end up being a copy paste of all flags. I think it won't be very useful.

> This should check the existence of the resources the MHC controller deploys rather than just checking if it's enabled by default right? That would be an actual e2e test to confirm the resources were successfully created and there wasn't a conflict.

@bennerv something along these lines, yes. I created [this work item](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15689695) to make sure that HMC resources get re-created after deletion. I think this should be enough.

PR #2438, opened 2022-10-04
E2E clients cleanup

m1kola:

Part of the [WI 13667437](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13667437).

* Removes unused clients
* Switches some tests to the newer API version

### Test plan for issue:

Run E2E

No, just test refactoring.

### DIFF
--- a/test/e2e/disk_encryption.go
+++ b/test/e2e/disk_encryption.go
@@ -20,7 +20,7 @@ var _ = Describe("Encryption at host should be enabled", func() {
 		ctx := context.Background()
 
 		By("getting the resource group where the VM instances live in")
-		oc, err := clients.OpenshiftClustersv20200430.Get(ctx, vnetResourceGroup, clusterName)
+		oc, err := clients.OpenshiftClustersv20220401.Get(ctx, vnetResourceGroup, clusterName)
 		Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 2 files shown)


test/e2e/disk_encryption.go
```diff
@@ -20,7 +20,7 @@ var _ = Describe("Encryption at host should be enabled", func() {
 		ctx := context.Background()
 
 		By("getting the resource group where the VM instances live in")
-		oc, err := clients.OpenshiftClustersv20200430.Get(ctx, vnetResourceGroup, clusterName)
+		oc, err := clients.OpenshiftClustersv20220904.Get(ctx, vnetResourceGroup, clusterName)
```
m1kola, line 23: I'll be adding additional checks to these two tests: want to check that cluster resource has appropriate fields set

cblecker (approved):

bennerv:
@m1kola should we /hold this PR until we perform an ARM manifest rollout with the new API version?  

Otherwise if e2e runs against production, or we have to cut a new release with this PR it will fail since the API is not released and the client GETs using the newest API (which isn't yet released) will not work.

bennerv (dismissed):

m1kola:
@bennerv that makes sense, but it will block my next PR which relies on some new fields. Let's switch to 2022-04-01 for now and then we can switch again to the newer one later.

I updated this PR.

bennerv (approved):
lgtm

PR #2439, opened 2022-10-05
Region dropdown admin portal

ellis-johnson:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/9777050 

Added a nav menu to access other prod regions on admin portal

### Test plan for issue:
E2E Tests to be added for admin portal

No


portal/v2/.eslintrc.yml
```diff
@@ -22,4 +22,6 @@ settings:
     fragment: "Fragment"
     version: "detect"
     flowVersion: "0.53"
-rules: {}
+rules: {
+  no-undef: "off"
```
dem4gus, line 26: `no-undef` presumably being disabled because of errors with JSX typing global variables. Eslint [recommendation](https://github.com/typescript-eslint/typescript-eslint/blob/main/docs/linting/TROUBLESHOOTING.md#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors) is to disable the `no-undef` rule for TypeScript projects because TypeScript's type checker is much more robust.

NIT: That said, this rules section no longer needs the `{}` in the yaml.

pkg/portal/regions.go
```diff
@@ -0,0 +1,98 @@
...
+	Regions []Region `json:"regions"`
+}
+
+func (p *portal) regions(w http.ResponseWriter, r *http.Request) {
+	resp := &RegionInfo{
+		Regions: make([]Region, 0, len(PROD_REGIONS)),
+	}
+
+	if value, found := os.LookupEnv("AZURE_ENVIRONMENT"); found {
+		var err error
```
dem4gus, line 75: This may be some Go idiom I'm unfamiliar with but won't `err` be initialized as `nil` always, making the check in the next line useless?

dem4gus (changes requested):
Looks good, just a couple nits and questions about things unfamiliar to me.

pkg/portal/regions.go
```diff
@@ -0,0 +1,98 @@
+package portal
```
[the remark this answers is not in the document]
ellis-johnson: Test currently feels a bit redundant since we are just returning a static list. But am happy to add one if that is the consensus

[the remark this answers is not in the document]
ellis-johnson: Have made a test anyway since the object that comes out from the static list is slightly different

portal/v2/.eslintrc.yml
ellis-johnson: Yep it was firing this check on some pretty unnecessary stuff so I added this, and fixed!

pkg/portal/regions.go
ellis-johnson: Nope, just a silly add on my side. Removed

test/e2e/admin_portal.go
```diff
@@ -197,4 +197,45 @@ var _ = Describe("Admin Portal E2E Testing", func() {
 			SaveScreenshotAndExit(wd, err)
 		}
 	})
+
+	It("Should be able to navigate to other regions", func() {
+		NUMBER_OF_REGIONS := 40
```
dem4gus, line 202: Will this assumption for region number be valid for environments and clouds, like INT and Fairfax? From my own experiences with the ImageConfig controller (see #2340) e2e tests will still run in those environments but variables such as `AZURE_ENVIRONMENT` will be different, which affects the list we expect to get back.

pkg/portal/regions.go
```diff
@@ -0,0 +1,93 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+)
+
+var (
+	PROD_REGIONS = []string{
```
dem4gus, line 16: I looked through the repo and couldn't find anywhere else we're defining or using a list like this, but if the need arises anywhere else it would probably be a good idea to pull this out to a central location so we only have to change one place when we add a new region. No change needed for now, just wanted to make a note of it.

pkg/portal/regions_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/go-test/deep"
+	"github.com/gorilla/mux"
+
+	testdatabase "github.com/Azure/ARO-RP/test/database"
+)
+
+func TestRegionList(t *testing.T) {
```
dem4gus, line 18: Need a test for behavior when `AZURE_ENVIRONMENT` is not the same as `azureclient.PublicCloud.Environment.Name`. Returning an empty list is fine as long as the frontend component knows how to deal with it.

dem4gus (changes requested):
Clarifying reasoning for my earlier suggestions. Need some more information about the expected behavior in environments that are not public prod.

test/e2e/admin_portal.go
ellis-johnson: I believe these tests still only run in development as there are variables like this which currently make it difficult to test properly in production, int and FF

pkg/portal/regions_test.go
ellis-johnson: Have added two clear tests for when the environment is set to public and when it is set to Fairfax

pkg/portal/regions.go
ellis-johnson: A ticket has been created to create an api for our cloud regions which this will eventually be pointed at

petrkotas (approved):
Seems good. Thanks.

dem4gus (approved):
Looks good.

pkg/portal/regions.go
s-amann: Until then can we add a note in our region onboarding doc to make sure we update this list? [region rollout](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/59387/New-region-rollout)

portal/v2/src/RegionList.tsx
```diff
@@ -0,0 +1,53 @@
...
+    visibility: 'hidden'
+  }
+};
+
+interface RegionComponentProps {
+  item: IRegionList
+}
+
+interface IRegionComponentState {
+  item: IRegion // why both state and props?
```
s-amann, line 15: nit: why indeed? :stuck_out_tongue:

s-amann (dismissed):

pkg/portal/regions.go
ellis-johnson: Absolutely, I will do this prior to merging

portal/v2/src/RegionList.tsx
ellis-johnson: I was under the impression props were for passing in values into a it class and state was for maintaining the values of the class during runtime. If this is the incorrect way to this I am happy to change it

darthhexx (approved):

portal/v2/src/RegionList.tsx
s-amann: I don't know that it is incorrect, I was just curious the answer to `why`. Thank you!

PR #2444, opened 2022-10-05
Add .dockerignore file :rocket: 

bennerv:

Adds a dockerignore file so when you build with podman or docker, it won't copy unnecessary files into the container image.  

`pyenv` is like 2Gi, so this should save us a bunch of time.

```bash
$ du -h ./pyenv -d0
1.9G	./pyenv
```

### Test plan for issue:

Run the multistage build and check the build artifact to see if the files are excluded.  They are

Nope


ross-bryan (approved):
python is only used in the client upstream so we dont need it in our builds

bennerv:
> python is only used in the client upstream so we dont need it in our builds

Our `Dockerfile.ci` slurps in our python code and runs python linting + unit tests against it so we can't exclude it.

PR #2238, opened 2022-07-08
Move Dev Environment VPN to Enable Connection to RP and Cluster Private IPs Simultaneously

zacharyljones:
### Which issue this PR addresses:

[USER STORY 14932248](https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/14932248/)

### What this PR does / why we need it:

The current dev RP infrastructure has one VPN connected to the dev VNET. This can be used to access the private ip addresses associated with dev clusters, but cannot be used to access private IP addresses associated with the dev RP. Certain actions, like some E2E tests, require access to both simultaneously. Setting the VPN up in a way where the RP and dev clusters can be accessed simultaneously will allow other SREs to test changes more effectively.

### Test plan for issue:

I used the new ARM templates to create a shared RP and then used that RP to...
1. create a private cluster
2. pull the kubeconfig for that cluster and issue kubectl commands against it
3. curl the admin api endpoints to perform various cluster related actions
4. ran e2e tests using the cluster

### Is there any documentation that needs to be updated for this PR?

No - the process for creating this shared RP is exactly the same as before

PR #2445, opened 2022-10-05
Remove the proxy VM and custom dial code

karanmagdani:

Fixes


pkg/deploy/assets/rp-development-predeploy.json
```diff
@@ -113,13 +113,13 @@
             "properties": {
                 "addressSpace": {
                     "addressPrefixes": [
-                        "10.0.4.0/22"
+                        "10.0.3.0/22"
```
karanmagdani, line 116: use /24

karanmagdani:
I have a created an RP (RG= km-eastus) in our RH tenant with these changes and the cluster is creating. I believe this should work but there are some big concerns:

dev-vnet was changed to `10.0.0.0/16`  here https://github.com/Azure/ARO-RP/pull/2238. While this does not affect our CI environment. Development Environment has two issues:

1. We can created only limited cluster in development environment due to smaller dev-vnet (~128)
2. Generate subnets function needs a [hacky condition](https://github.com/Azure/ARO-RP/pull/2445/files#diff-c7dba7bf3908043db4c458af9edd806998129f1ef7f58ee8fa54a312ce93f23cR352) to check if RP_Mode=development & CI=False. 

Before merging these changes and redeploying RP, i think we need to discuss whether the right solution is keep it as it is in this PR, redesign the address spaces in a better/more efficient way or have separate address  spaces for dev environment and CI e2e environment

karanmagdani:
Should be noted that any cluster creations/operations after this change will need need to happen over VPN

PR #2446, opened 2022-10-05
Remove ClusterManagerConfiguration from v20220904, add static validation

cadenmarchese:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15533307/, and also removes ClusterManagerConfiguration from API.

- [x] Refactor the v20220904 API to no longer use the `ClusterManagerConfiguration` type
- [x] Fix the unit testing
- [x] Add static validation to frontend


pkg/frontend/clustermanager_putorpatch.go
```diff
@@ -22,24 +22,55 @@ func (f *frontend) putOrPatchClusterManagerConfiguration(w http.ResponseWriter,
...
+			b, err = f._putOrPatchSecret(ctx, log, r, &header, f.apis[vars["api-version"]].SecretConverter)
+		}
 		return err
 	})
 
 	reply(log, w, header, b, err)
 }
 
-func (f *frontend) _putOrPatchClusterManagerConfiguration(ctx context.Context, log *logrus.Entry, r *http.Request, header *http.Header, converter api.ClusterManagerConfigurationConverter) ([]byte, error) {
+func (f *frontend) _putOrPatchSyncSet(ctx context.Context, log *logrus.Entry, r *http.Request, header *http.Header, converter api.SyncSetConverter) ([]byte, error) {
```
cadenmarchese, line 73: Currently I have separate putorpatch functions for each resource type (`_putOrPatchSyncSet`, `_putOrPatchMachinePool`, `_putOrPatchSyncIdentityProvider`, `_putOrPatchSecret`). This seems required since we now have separate converters for each resource type (`machinepool_convert.go`, `syncset_convert.go`, etc..).

Should I have kept the `clustermanager_convert.go` coverter and used it instead of having the separate putorpatch functions?

pkg/frontend/clustermanager_get.go
```diff
@@ -20,17 +21,48 @@ func (f *frontend) getClusterManagerConfiguration(w http.ResponseWriter, r *http
...
+
+		b, err = f._getSyncIdentityProviderConfiguration(ctx, log, r, f.apis[vars["api-version"]].SyncIdentityProviderConverter)
+	case "secret":
+		if f.apis[vars["api-version"]].SecretConverter == nil {
+			api.WriteError(w, http.StatusBadRequest, api.CloudErrorCodeInvalidResourceType, "", badRequestError)
+			return
+		}
+
+		b, err = f._getSecretConfiguration(ctx, log, r, f.apis[vars["api-version"]].SecretConverter)
 	}
```
ventifus, line 61: how about a `default:` clause here that raises an error for unsupported resource type?

pkg/frontend/clustermanager_putorpatch.go
ventifus: I like having separate functions per resource type, even though it's a lot of similar code it feels cleaner.

pkg/frontend/clustermanager_get.go
cadenmarchese: Done, thanks!

pkg/api/v20220904/clustermanager_validatestatic.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+type clusterManagerStaticValidator struct{}
+
+func (c clusterManagerStaticValidator) Static(body string, vars map[string]string) error {
+	ocmResourceType := vars["ocmResourceType"]
+	var resource map[string]interface{}
+
+	if IsBase64(body) {
+		decodedBody, err := base64.StdEncoding.DecodeString(body)
```
ventifus, line 20: How about this, to avoid decoding twice?
```go
if decodedBody, err := base64.StdEncoding.DecodeString(body); decodedBody != nil {
```

pkg/api/v20220904/clustermanager_validatestatic.go
```diff
@@ -0,0 +1,48 @@
...
+		if err != nil {
```
ventifus, line 21: How about this to avoid decoding twice?
```go
	if decodedBody, err := base64.StdEncoding.DecodeString(body); err == nil {
```

pkg/frontend/clustermanager_get.go
```diff
@@ -20,19 +21,100 @@ func (f *frontend) getClusterManagerConfiguration(w http.ResponseWriter, r *http
...
 
-	if f.apis[vars["api-version"]].ClusterManagerConfigurationConverter == nil {
-		api.WriteError(w, http.StatusBadRequest, api.CloudErrorCodeInvalidResourceType, "", "The resource type '%s' could not be found in the namespace '%s' for api version '%s'.", vars["resourceType"], vars["resourceProviderNamespace"], vars["api-version"])
+	var (
+		b   []byte
+		err error
+	)
+
+	switch vars["ocmResourceType"] {
+	case "syncset":
```
s-amann, line 33: can we encapsulate this logic of valid resources / non nil converts into a shared function to reduce the duplication between our REST operations? It might also help down the line if we want to support new resource types

s-amann (changes requested):
Please consider DRYing up our check on the `ocmResourceType` in the GET/DETLETE/PUTPATCH operations.

pkg/api/v20220904/secret_convert.go
```diff
@@ -0,0 +1,39 @@
...
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"github.com/Azure/ARO-RP/pkg/api"
+)
+
+type secretConverter struct{}
+
+func (c secretConverter) ToExternal(s *api.Secret) interface{} {
```
ventifus, line 12: It looks like Microsoft doesn't want secrets to be retrievable. I'm getting an error from their linter: Secret property "secretResources" cannot be sent in the response

https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/Semantic-and-Model-Violations-Reference.md#SECRET_PROPERTY

cadenmarchese: I've removed `SecretResources` from the `ToExternal` secret converter.

pkg/frontend/clustermanager_type.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+)
+
+func (f *frontend) readOcmResourceType(vars map[string]string) error {
+	badRequestError := fmt.Errorf("the resource type '%s' could not be found in the namespace '%s' for api version '%s'", vars["ocmResourceType"], vars["resourceProviderNamespace"], vars["api-version"])
+
+	switch vars["ocmResourceType"] {
+	case "syncset":
+		if f.apis[vars["api-version"]].SyncSetConverter == nil {
+			return fmt.Errorf("the resource type '%s' could not be found in the namespace '%s' for api version '%s'", vars["ocmResourceType"], vars["resourceProviderNamespace"], vars["api-version"])
```
ventifus, line 16: return badRequestError here

pkg/frontend/clustermanager_type.go
```diff
@@ -0,0 +1,35 @@
...
+	case "syncidentityprovider":
+		if f.apis[vars["api-version"]].SyncIdentityProviderConverter == nil {
+			return badRequestError
+		}
+	case "secret":
+		if f.apis[vars["api-version"]].SecretConverter == nil {
+			return badRequestError
+		}
+	default:
+		return fmt.Errorf("the resource type '%s' is not supported for api version '%s'", vars["ocmResourceType"], vars["api-version"])
```
ventifus, line 31: The difference in error message is subtle, might as well use badRequestError here too.

pkg/frontend/clustermanager_type.go
```diff
@@ -0,0 +1,35 @@
+package frontend
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"fmt"
+)
+
+func (f *frontend) readOcmResourceType(vars map[string]string) error {
```
ventifus, line 10: This doesn't really "read" anything does it? How about `validateOcmResourceType`?

ventifus (dismissed):

pkg/frontend/clustermanager_type.go
s-amann: only difference is the interpretation of the error, the previous errors indicate the resourceType could be supported in a different api-version, where as this error states it is not supported at all.

pkg/frontend/clustermanager_putorpatch.go
```diff
@@ -102,52 +120,369 @@ func (f *frontend) _putOrPatchClusterManagerConfiguration(ctx context.Context, l
...
+		return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidRequestContent, "", "The request content was invalid and could not be deserialized: %q.", err)
+	}
+
+	err = staticValidator.Static(resources, vars)
+	if err != nil {
+		return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidRequestContent, "", "The 'Kind' in the request payload does not match 'Kind' in the request path: %q.", err)
+	}
+
+	isCreate := ocmdoc == nil
+	uuid := f.dbClusterManagerConfiguration.NewUUID()
```
s-amann, line 182: I believe this chunk is duplicated 4 times in this file. Simply refactoring this chunk out leaves much to be desired as there are way too many return values. I played around a little and was able to extract lines 145-163 into a function returning an error and the 'originalPath'. 

```go
func validateOriginalPath(ctx context.Context, r *http.Request, vars map[string]string) (string, error) {
	_, err := f.validateSubscriptionState(ctx, r.URL.Path, api.SubscriptionStateRegistered)
	if err != nil {
		return "", err
	}

	originalPath := r.Context().Value(middleware.ContextKeyOriginalPath).(string)
	armResource, err := arm.ParseArmResourceId(originalPath)
	if err != nil {
		return "", err
	}

	ocp, err := f.dbOpenShiftClusters.Get(ctx, armResource.ParentResource())
	if err != nil && !cosmosdb.IsErrorStatusCode(err, http.StatusNotFound) {
		return "", err
	}

	if ocp == nil || cosmosdb.IsErrorStatusCode(err, http.StatusNotFound) {
		return "", api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, "", "The Resource '%s/%s' under resource group '%s' was not found.", vars["resourceType"], vars["resourceName"], vars["resourceGroupName"])
	}
	return originalPath, err
}
```

s-amann (dismissed):
lgtm other than the duplication in `clustermanager_putorpatch.go`. If we are pressed for time I'd be ok with leaving this as a future work item. Let me know if you want to work together to knock this out

pkg/frontend/clustermanager_type.go
```diff
@@ -0,0 +1,35 @@
...
+	badRequestError := fmt.Errorf("the resource type '%s' could not be found in the namespace '%s' for api version '%s'", vars["ocmResourceType"], vars["resourceProviderNamespace"], vars["api-version"])
```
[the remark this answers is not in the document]
cadenmarchese: Would it be better to remove it from the error or add a check for it?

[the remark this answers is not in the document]
s-amann: If it is something we need to validate then I would add the validation, I'm just not sure why we say it can't be found in the namespace when we have an explicit list of resourceTypes we support.

pkg/frontend/clustermanager_putorpatch.go
cadenmarchese: Thanks, agreed that this is a good chunk to DRY. I've worked in your changes.

ventifus:
lgtm

s-amann (dismissed):

PR #2450, opened 2022-10-06
Refactor conditions to return error

SrinivasAtmakuri:
Currently, all conditions steps return InternalServerError and only looking at the RP logs will give detailed understanding on the failure, this PR attempts to change this by adding a wrapper function to exiting poll function to return readable error.

Fixes: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/9224646

### Test plan for issue:

Add Unit tests for AdminUpdate and Test_Runner

This PR is based on implementation done by David here: https://github.com/Azure/ARO-RP/pull/1863

### DIFF
--- a/pkg/util/steps/condition.go
+++ b/pkg/util/steps/condition.go
@@ -5,12 +5,34 @@ package steps
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
+	"strings"
 	"time"
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/Azure/ARO-RP/pkg/api"
 )
+
+// Functions that run as condition-steps return Error
+// instead of InternalServerError
+// Efforts are being made  to not have generic Hive errors but specific, actionable failure cases.
+// Instead of providing Hive-specific error messages to customers, the below will send a timeout error message.
+var timeoutConditionErrors = map[string]string{
+	"apiServersReady":                        "Kube API has not initialised successfully and is unavailable.",
+	"minimumWorkerNodesReady":                "Minimum number of worker nodes have not been successfully created.",
+	"operatorConsoleExists":                  "Console Cluster Operator has failed to initialize successfully.",
+	"operatorConsoleReady":                   "Console Cluster Operator has not started successfully.",
+	"clusterVersionReady":                    "Cluster Verion is not reporting status as ready.",
+	"ingressControllerReady":                 "Ingress Cluster Operator has not started successfully.",
+	"aroDeploymentReady":                     "ARO Cluster Operator has failed to initialize successfully.",
+	"ensureAROOperatorRunningDesiredVersion": "ARO Cluster Operator is not running desired version.",
+	"hiveClusterDeploymentReady":             "Timed out waiting for a condition, cluster Installation is unsuccessful.",
+	"hiveClusterInstallationComplete":        "Timed out waiting for a condition, cluster Installation is unsuccessful.",
+}
 
 // conditionFunction is a function that takes a context and returns whether the
 // condition has been met and an error.
@@ -66,7 +88,27 @@ func (c conditionStep) run(ctx context.Context, log *logrus.Entry) error {
 		log.Warnf("step %s failed but has configured 'fail=%t'. Continuing. Error: %s", c, c.fail, err.Error())
 		return nil
 	}
+	if errors.Is(err, wait.ErrWaitTimeout) {
+		return enrichConditionTimeoutError(c.f)
+	}
 	return err
+}
+
+// Instead of giving Generic, timed out waiting for a condition, error
+// returns enriched error messages mentioned in timeoutConditionErrors
+func enrichConditionTimeoutError(f conditionFunction) error {
+	funcNameParts := strings.Split(FriendlyName(f), ".")
+	funcName := strings.TrimSuffix(funcNameParts[len(funcNameParts)-1], "-fm")
+
+	message, exists := timeoutConditionErrors[funcName]
+	if !exists {
+		return errors.New("timed out waiting for the condition")
+	}
+	return api.NewCloudError(
+		http.StatusInternalServerError,
+		api.CloudErrorCodeDeploymentFailed,
+		"", message+"Please retry, if issue persists: raise azure support ticket",
+	)
 }
 
 func (c conditionStep) String() string {
--- a/pkg/util/steps/condition_test.go
+++ b/pkg/util/steps/condition_test.go
@@ -1,0 +1,95 @@
+package steps
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"testing"
+)
+
+// functionnames that will be used in the conditionFunction below
+// All the keys of map timeoutConditionErrors
+func apiServersReady(context.Context) (bool, error)                        { return false, nil }
+func minimumWorkerNodesReady(context.Context) (bool, error)                { return false, nil }
+func operatorConsoleExists(context.Context) (bool, error)                  { return false, nil }
+func operatorConsoleReady(context.Context) (bool, error)                   { return false, nil }
+func clusterVersionReady(context.Context) (bool, error)                    { return false, nil }
+func ingressControllerReady(context.Context) (bool, error)                 { return false, nil }
+func aroDeploymentReady(context.Context) (bool, error)                     { return false, nil }
+func ensureAROOperatorRunningDesiredVersion(context.Context) (bool, error) { return false, nil }
+func hiveClusterDeploymentReady(context.Context) (bool, error)             { return false, nil }
+func hiveClusterInstallationComplete(context.Context) (bool, error)        { return false, nil }
+
+func TestEnrichConditionTimeoutError(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/util/steps/condition.go
```diff
@@ -66,9 +70,48 @@ func (c conditionStep) run(ctx context.Context, log *logrus.Entry) error {
...
+		return c.enrichConditionTimeoutError()
+	}
 	return err
 }
 
+func (c conditionStep) enrichConditionTimeoutError() error {
+	funcNameParts := strings.Split(FriendlyName(c.f), ".")
+	funcName := strings.Replace(funcNameParts[len(funcNameParts)-1], "-fm", "", 1)
+
+	message := "timed out polling for a prerequisite cluster condition to resolve"
```
SudoBrendan, line 83: You set this string twice - once here, once in `default`. Likely best to only be once.

pkg/util/steps/condition.go
```diff
@@ -66,9 +70,48 @@ func (c conditionStep) run(ctx context.Context, log *logrus.Entry) error {
...
+		message = "ARO Cluster Operator is not running desired version"
+	case "hiveClusterDeploymentReady":
+		message = "Cluster Hive deployment is unsuccessful and is unavailable"
+	case "hiveClusterInstallationComplete":
+		message = "Cluster Hive Installation is unsuccessful"
+	default:
+		message = "timed out polling for a prerequisite cluster condition to resolve"
+	}
+	return api.NewCloudError(
+		http.StatusFailedDependency,
```
SudoBrendan, line 109: Hmm... IDK about code 424 here - this indicates a client-side error, that being, that the customer previously submitted a request to our API that caused a subsequent one (this one) to fail - this is likely not the case. IMO we should be returning a 500 Internal Server Error for these, but open to discussion from the team.

SudoBrendan (commented):
Thanks for the PR! This is looking OK so far - the only topic I want to raise is the HTTP code we are returning. These don't feel like client-side errors.

I'd like to see some tests added before merging.

pkg/util/steps/condition.go
```diff
@@ -66,9 +71,48 @@ func (c conditionStep) run(ctx context.Context, log *logrus.Entry) error {
...
+	case "ingressControllerReady":
+		message = "Ingress Cluster Operator has not started successfully"
+	case "aroDeploymentReady":
+		message = "ARO Cluster Operator has failed to initialize successfully"
+	case "ensureAROOperatorRunningDesiredVersion":
+		message = "ARO Cluster Operator is not running desired version"
+	case "hiveClusterDeploymentReady":
+		message = "Cluster Hive deployment is unsuccessful and is unavailable"
+	case "hiveClusterInstallationComplete":
+		message = "Cluster Hive Installation is unsuccessful"
```
jaitaiwan, line 105: I'd like to see this as a map rather than a switch. e.g.

```
const TimeoutConditionErrors = map[string]string{
   "apiServersReady": "Kube API has not initialised successfully and is unavailable",
  "minimumWorkerNodesReady": ""minimum number of worker nodes have not been successfully created"
  // ...
}

message, exists := TimeoutConditionErrors[funcName]
if exists == false {
  return errors.New("timed out waiting for the condition")
}
```

jaitaiwan (changes requested):

pkg/util/steps/condition.go
SrinivasAtmakuri: Made the change.

pkg/util/steps/condition.go
```diff
@@ -5,13 +5,33 @@ package steps
...
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/Azure/ARO-RP/pkg/api"
 )
 
+// Functions that run as condition-steps return Error
+// instead of InternalServerError
+var TimeoutConditionErrors = map[string]string{
```
facchettos, line 22: does this need to be exported?

facchettos (changes requested):

jaitaiwan (dismissed):
LGTM!

pkg/util/steps/condition.go
```diff
@@ -66,9 +86,27 @@ func (c conditionStep) run(ctx context.Context, log *logrus.Entry) error {
 		log.Warnf("step %s failed but has configured 'fail=%t'. Continuing. Error: %s", c, c.fail, err.Error())
 		return nil
 	}
+	if err == wait.ErrWaitTimeout {
+		return c.enrichConditionTimeoutError()
+	}
 	return err
 }
 
+func (c conditionStep) enrichConditionTimeoutError() error {
```
s-amann, line 95: this utility function is prime for a test. Please add some coverage.

s-amann (dismissed):
please add unit tests for the new conditions added in the PR.

pkg/util/steps/condition.go
SrinivasAtmakuri: I have added test coverage for enrichConditionTimeoutError. Thanks

pkg/util/steps/condition.go
```diff
@@ -5,13 +5,33 @@ package steps
...
 	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/Azure/ARO-RP/pkg/api"
 )
 
+// Functions that run as condition-steps return Error
+// instead of InternalServerError
+var timeoutConditionErrors = map[string]string{
+	"apiServersReady":                        "Kube API has not initialised successfully and is unavailable",
+	"minimumWorkerNodesReady":                "minimum number of worker nodes have not been successfully created",
```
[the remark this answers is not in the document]
SrinivasAtmakuri: Made the change.

pkg/util/steps/condition.go
```diff
@@ -5,13 +5,33 @@ package steps
...
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/Azure/ARO-RP/pkg/api"
 )
 
+// Functions that run as condition-steps return Error
+// instead of InternalServerError
+var timeoutConditionErrors = map[string]string{
```
SrinivasAtmakuri, line 22: Looking for inputs on the error messages here, as these will customer visible when install/update/delete fails.
Also, I am cross-minded if we wanna share with customers saying `Hive deployment failed`, 

I want to make these errors not very generic, but more precise and actionable, the aim is to prevent InternalServerError with an actionable error or at least append `Please contact Support if retry fails`

Any thoughts?

pkg/util/steps/condition.go
```diff
@@ -66,9 +86,29 @@ func (c conditionStep) run(ctx context.Context, log *logrus.Entry) error {
 		log.Warnf("step %s failed but has configured 'fail=%t'. Continuing. Error: %s", c, c.fail, err.Error())
 		return nil
 	}
+	if err == wait.ErrWaitTimeout {
+		return c.enrichConditionTimeoutError()
```
cadenmarchese, line 90: This utility function `c.enrichConditionTimeoutError()` might be easier to test if placed outside of `c`. You could set up something like the below, which passes `conditionFunction` as a parameter to utility function `enrichConditionTimeoutError()` that then provides the expected error. 
```suggestion
		return enrichConditionTimeoutError(c.f)
```
Functionally it would be the same but the unit test would be simpler.

pkg/util/steps/condition.go
SrinivasAtmakuri: cc @bennerv @s-amann @cadenmarchese

pkg/util/steps/condition.go
SrinivasAtmakuri: Done.

pkg/util/steps/condition.go
```diff
@@ -66,9 +86,29 @@ func (c conditionStep) run(ctx context.Context, log *logrus.Entry) error {
 		log.Warnf("step %s failed but has configured 'fail=%t'. Continuing. Error: %s", c, c.fail, err.Error())
 		return nil
 	}
+	if err == wait.ErrWaitTimeout {
```
tsatam, line 89: I don't think it necessarily needs to be handled in this PR, but I have noticed that we see `ErrWaitTimeout` errors returned in the internal implementation of some condition steps (e.g. a k8s client list call takes too long to resolve). 

I've opened a PR to wrap "internal" ErrWaitTimeout errors within the condition runner [here](https://github.com/Azure/ARO-RP/pull/3005), which will change the shape of the log message we see but preserve the error type as `ErrWaitTimeout`. 

If we want this error enrichment to handle both "the condition timed out" and "something within the condition timed out" cases, this if statement should be changed to `errors.Is(err, wait.ErrWaitTimeout)` instead of the equality comparison, in order to match the wrapped error.

SrinivasAtmakuri: Yes, agreed, made the change.

pkg/util/steps/condition.go
SrinivasAtmakuri: I have modified the code to prevent HiveRelated errors and only pass a generic timeout message, instead, please review.

facchettos (approved):
good job! I'll let someone else check if the messages are appropriate but I don't see any problem with them I think

petrkotas (approved):
@SrinivasAtmakuri awesome work. This will make the UX much more pleasant.

PR #2451, opened 2022-10-06
added difference checker

facchettos:
Improves the diff checking so we don't rely on `git status`
Fixes the problem introduced by @bennerv 's commit. :tada: 

Fixes


ross-bryan (approved):

yehornaumenko:
@facchettos could you explain the difference between the native `git diff` command and hack/ci-utils/differenceChecker/main.go, please? I tried to look over all options (flags) `git diff` provides and compare it to your method but didn't manage to find a difference. Probably I missed something :/

petrkotas (approved):
LGTM, thanks

PR #2460, opened 2022-10-13
toil: metrics for nonstandard cluster support status

ArrisLee:
https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/10614778

As an SRE, I want to be able to use metrics dashboards and alerts to determine if feature flags are set such that they disable parts of the ARO Operator, and also to know if a "contact support" banner is being displayed in their console.
By doing this, we can more easily monitor the fleet and keep track of snowflake or unsupported clusters

Test plan for issue:
Unit test

N/A


pkg/monitor/cluster/cache.go
```diff
@@ -59,3 +60,13 @@ func (mon *Monitor) listARODeployments(ctx context.Context) (*appsv1.DeploymentL
 	mon.cache.arodl, err = mon.cli.AppsV1().Deployments(pkgoperator.Namespace).List(ctx, metav1.ListOptions{})
 	return mon.cache.arodl, err
 }
+
+func (mon *Monitor) listAROClusters(ctx context.Context) (*arov1alpha1.ClusterList, error) {
```
ross-bryan, line 64: I am not sure this is a good idea. Do you see in this file how every list call has a comment saying it is an anti-pattern and should be avoided? Since we are not under strict business pressures to implement this I would ask we re-think this logic and do this correctly with pagination.

pkg/monitor/cluster/clusterflagsandbanner.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, cluster := range cs.Items {
+		if cluster.Spec.OperatorFlags != nil {
+			defualtFlags := api.DefaultOperatorFlags()
+			nonStandardOperatorFlags := make(map[string]string)
+			//check if the current set flags matches the default ones
+			for flag, status := range cluster.Spec.OperatorFlags {
```
ross-bryan, line 29: do you think we should loop over the defaultFlags rather than the current cluster flags? If the cluster flags are missing some then it won't be included in the loop or emitted via the monitor

pkg/monitor/cluster/cache.go
ross-bryan: Perhaps a limit parameter should be exposed in this function call here

pkg/monitor/cluster/clusterflagsandbanner.go
```diff
@@ -0,0 +1,46 @@
...
+	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+)
+
+const (
+	operatorFlagMetricsTopic  = "cluster.nonstandard.operator.featureflags"
+	supportBannerMetricsTopic = "cluster.nonstandard.contact.support.banner"
+)
+
+func (mon *Monitor) emitOperatorFlagsAndSupportBanner(ctx context.Context) error {
+	cs, err := mon.listAROClusters(ctx)
```
ross-bryan, line 19: see comment above regarding pagination, this now has 2 copies of the entire list of clusters in memory. one in the cache and now a second in this function. this is where we could paginate and perform logic as needed.

pkg/monitor/cluster/clusterflagsandbanner.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, cluster := range cs.Items {
+		if cluster.Spec.OperatorFlags != nil {
```
ross-bryan, line 25: what if operatorFlags are nil? should we do anything about that? do we care?

pkg/monitor/cluster/clusterflagsandbanner.go
```diff
@@ -0,0 +1,46 @@
...
+			defualtFlags := api.DefaultOperatorFlags()
+			nonStandardOperatorFlags := make(map[string]string)
```
ross-bryan, line 27: nit: you can set the initialized length of this map to the length of the default flags

pkg/monitor/cluster/clusterflagsandbanner.go
```diff
@@ -0,0 +1,46 @@
...
+
+func (mon *Monitor) emitOperatorFlagsAndSupportBanner(ctx context.Context) error {
+	cs, err := mon.listAROClusters(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, cluster := range cs.Items {
+		if cluster.Spec.OperatorFlags != nil {
+			defualtFlags := api.DefaultOperatorFlags()
```
ross-bryan, line 26: nit: typo 
```suggestion
                        defaultFlags := api.DefaultOperatorFlags()
```

pkg/monitor/cluster/clusterflagsandbanner.go
```diff
@@ -0,0 +1,46 @@
...
+					nonStandardOperatorFlags[flag] = status
+				}
+			}
+			if len(nonStandardOperatorFlags) > 0 {
+				mon.emitGauge(operatorFlagMetricsTopic, 1, nonStandardOperatorFlags)
+			}
+		}
+
+		//check if the contact support banner is activated
+		if cluster.Spec.Banner.Content == arov1alpha1.BannerContactSupport {
```
ross-bryan, line 40: Can the banner be nil?

ross-bryan (changes requested):
Hi Arris, thanks for this PR. A couple of comments that I think we should consider and review how this is implemented.

pkg/monitor/cluster/clusterflagsandbanner.go
ArrisLee: actually I am not 100% sure about this since I believe some of the old-version clusters do not have any valid operator flags, should we regard them as unsupported ones or not?

pkg/monitor/cluster/clusterflagsandbanner.go
ArrisLee: the `Banner` obj here is a struct value not a pointer, so no need to worry about the nil pointer exception

pkg/monitor/cluster/clusterflagsandbanner.go
s-amann: I agree this should be flipped and add a check for when the cluster flags are missing a default flag key.

s-amann (dismissed):
Ross has made some valid points and I'd like to call out specifically looping over the defaultFlags rather than the clusters flags in the event the cluster is missing a flag.

pkg/monitor/cluster/clusterflagsandbanner.go
```diff
@@ -0,0 +1,67 @@
...
+					}
+				}
+
+				for flag, status := range cluster.Spec.OperatorFlags {
+					if defaultFlags[flag] != status {
+						nonStandardOperatorFlags[flag] = status
+					}
+				}
+				if len(nonStandardOperatorFlags) > 0 {
+					mon.emitGauge(operatorFlagMetricsTopic, 1, nonStandardOperatorFlags)
```
SudoBrendan, line 51: nit: this code was a little hard to read because `nonStandardOperatorFlags` is ultimately used as a Dimension `string[string]` map. Maybe naming it something like `nonStandardOperatorFlagDims` would make this more clear - we are building up a list of dimensions.

pkg/monitor/cluster/clusterflagsandbanner.go
```diff
@@ -0,0 +1,67 @@
...
+
+				//check if the current set flags matches the default ones
+				for name, value := range defaultFlags {
+					flag, ok := cluster.Spec.OperatorFlags[name]
+					if !ok {
+						// if flag is not there, put "not exist" in the metrics
+						nonStandardOperatorFlags[name] = "not exist"
+					} else if flag != value {
+						// if flag value does not match the defualt one, record the current set value in the metrics
+						nonStandardOperatorFlags[name] = flag
```
SudoBrendan, line 41: nit: my brain is also struggling with the name `flag` - this is actually the feature flag _value_ in the cluster, not the _name_ of the flag, right? Might be worth renaming stuff to `currentFlagValue` and `defaultFlagValue` or something - it seems like we are setting the dimension's (feature flag name's) value to the actual value in the cluster if it is not what we expect - so something similar to `aro.imageconfig.enabled=false` and/or `aro.imageconfig.enabled="not exist"`?

pkg/monitor/cluster/clusterflagsandbanner.go
```diff
@@ -0,0 +1,67 @@
...
+			if cluster.Spec.OperatorFlags != nil {
+				defaultFlags := api.DefaultOperatorFlags()
+				nonStandardOperatorFlags := make(map[string]string, len(defaultFlags))
+
+				//check if the current set flags matches the default ones
+				for name, value := range defaultFlags {
+					flag, ok := cluster.Spec.OperatorFlags[name]
+					if !ok {
+						// if flag is not there, put "not exist" in the metrics
+						nonStandardOperatorFlags[name] = "not exist"
```
SudoBrendan, line 38: nit: maybe "does not exist" (or "DNE" for short - a common mathematical shorthand for this... Though it is less clear, it saves on bytes. If using DNE, adding a comment would help.)

SudoBrendan (approved):
All things told, LGTM. I'm not seeing caching, and pagination is there, excellent. Callout on this testing, excellent format and very clear! Nice work.

I left a few nits on names of things if you want to have a look.

ArrisLee:
> All things told, LGTM. I'm not seeing caching, and pagination is there, excellent. Callout on this testing, excellent format and very clear! Nice work.
> 
> I left a few nits on names of things if you want to have a look.

@SudoBrendan Hi Brendan, thanks a lot for the review!

hawkowl (approved):

PR #2461, opened 2022-10-13
[WIP] Az aro CLI update to add cluster creation without public ip

ArielLima:
Create clusters without public ip through the cli

Customers can create their own private clusters

### Test plan for issue:
Will be tested by using the cli to create the cluster

We could add the functionality to the cluster creation doc somewhere here when it is live https://learn.microsoft.com/en-us/azure/openshift/tutorial-create-cluster

### DIFF
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -2,6 +2,7 @@
 # Licensed under the Apache License 2.0.
 
 from azext_aro._validators import validate_cidr
+from azext_aro._validators import validate_outbound_type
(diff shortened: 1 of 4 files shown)


bennerv:
This should not be merged until the next stable API release.  **Not** the 20220904 release.

darthhexx:
Due to the fact that this code requires a new API release, shouldn't we only be either be bundling it in with the first iteration of the API or waiting until the first iteration has been merged and then creating this PR?

tony-schndr:
closing this, replaced with #2622

PR #2462, opened 2022-10-14
removed useless parameter in clustermonitor

facchettos:
There seems to be a useless parameter in the newmonitor function

Fixes


rolandmkunkel (approved):

petrkotas (approved):
Yes it seems like it.

PR #2463, opened 2022-10-14
Update Proxy VMs to pull image every week in case of updates

bennerv:

Performs a weekly pull of the latest proxy image

Will help resolve vulnerabilities.  If we have a container vulnerability, we can just push up a new proxy image, wait until sunday, and then it will fix itself.  

### Test plan for issue:

Deploy proxy vm

Yes - need to doc the above


petrkotas (approved):
Thanks @bennerv. I see the value in this one.

karanmagdani (approved):
LGTM

PR #2464, opened 2022-10-14
[Admin Portal]: Add resource ID param in the URL

devamanv:
[ADO Card: SRE Portal v2 resourceid as a parameter](https://dev.azure.com/msazure/AzureRedHatOpenShift/_boards/board/t/Firefly/Stories/?workitem=15132772)

Fixes

The PR contains changes  to the SRE portal so that it accepts Resource ID as a URI parameter which opens the detailed of the desired cluster directly. 

### Test plan for issue:

E2E tests have been added as part of the changes


portal/v2/src/ClusterList.tsx
```diff
@@ -417,8 +491,11 @@ export function ClusterList(props: {
         styles={controlStyles}
       />
       <Separator styles={separatorStyle} />
-
+
```
ellis-johnson, line 494: Need to get rid of whitespace

ellis-johnson (commented):
LGTM! Just one nitpick

portal/v2/src/ClusterList.tsx
devamanv: Done

facchettos (dismissed):
would be good to get rid of bindata here and use embed instead. edit: Nevermind, taken care of by shawn

portal/v2/src/ClusterList.tsx
```diff
@@ -71,6 +76,59 @@ const separatorStyle = {
...
+                  window.history.replaceState({}, "", "/v2")
+                  props.hidePopup()
+                }}>
+                  Close
+                </DefaultButton>
+              </div>
+            </FocusTrapZone>
+          </Popup>
+        </Layer>
+    </>
```
jaitaiwan, line 128: There's only 1 top-level element `<Layer>` so the fragment `<>` is unnecessary.

portal/v2/src/index.tsx
```diff
@@ -14,4 +14,19 @@ mergeStyles({
...
+  let paramStringFromURL = searchParams.split('?')[1];
+  let paramsArr = paramStringFromURL.split('&');
+
+  for (let i = 0; i < paramsArr.length; i++) {
+      let pair = paramsArr[i].split('=');
+      urlParams = {[pair[0]]: pair[1]}
+  }
+
+  return urlParams
+}
```
jaitaiwan, line 30: This is a helper function that should be extracted into some sort of libs file.

test/e2e/admin_portal.go
```diff
@@ -244,4 +244,36 @@ var _ = Describe("Admin Portal E2E Testing", func() {
 			Expect(link.GetAttribute("href")).To(MatchRegexp(`https://([a-z]|[0-9])+\.admin\.aro\.azure\.com`))
 		}
 	})
+
+	It("Should open an error modal for an invalid resource ID parameter in the URL", func() {
+		wd.Get(host + "/v2" + "?resourceid=" + "invalidResourceId")
+		time.Sleep(time.Second * 2) // this is done to give enough time to fetch the data
+
+		wd.Wait(conditions.ElementIsLocated(selenium.ByCSSSelector, "div[role='document']"))
```
jaitaiwan, line 252: We shouldn't have both time.Sleep and wd.Wait. We should be waiting with specific enough selectors that it executes the test when the dom is ready with the data fetched.

test/e2e/admin_portal.go
```diff
@@ -244,4 +244,36 @@ var _ = Describe("Admin Portal E2E Testing", func() {
...
+			SaveScreenshotAndExit(wd, err)
+		}
+
+		Expect(errorModal.IsDisplayed()).To(BeTrue())
+	})
+
+	It("Should display the correct cluster detail view for the resource ID parameter in the URL", func() {
+		wd.Get(host + "/v2" + "?resourceid=" + resourceIDFromEnv())
+		time.Sleep(time.Second * 2)
+		wd.Wait(conditions.ElementIsLocated(selenium.ByID, "ClusterDetailPanel"))
```
jaitaiwan, line 264: Same as previous comment

jaitaiwan (changes requested):
Provided some more feedback on the PR. A few nicks a few best practice things.

mociarain:
We have this now (unless I'm misunderstanding ResourceId in this context??)

Happy to close this?

mociarain:
This has been implemented here AFAIK: https://github.com/Azure/ARO-RP/pull/2640.

Closing for now but do re-open if I've misunderstood

PR #2466, opened 2022-10-17
Fix race conditions in Test functions

AldoFusterTurpin:

 
[Fix race conditions in test functions](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/16201104)

Help requested 👀 

I have used the [data race detector](https://go.dev/doc/articles/race_detector) and have detected some race conditions.

If you run 
```sh
go test -race -tags=aro,containers_image_openpgp ./… 
```
with a fresh pull of master, you will see that there are some race conditions/data races ([not the same](https://cronokirby.github.io/posts/data-races-vs-race-conditions) BTW).

Those race conditions are just affecting some unit tests and not production code (AFAIK):

- there is a data race in the test/util/listener/listener.go Close method because we try to read “l.closed” which can be accessed by more than one go routine at a time, causing a data race. The first commit in this PR solves it.

- there is another race condition that happens in the pkg/portal/security_test.go TestSecurity function. There is a comment on line 355 of that file talking a bit more about it.

- there is one last race condition happening in the pkg/portal/prometheus/proxy_test.go:121 fakeServer function. It happens in:
  - pkg/portal/prometheus/proxy_test.go:69
  - pkg/portal/prometheus/proxy_test.go:97
  - pkg/portal/prometheus/proxy_test.go:117
  - pkg/portal/prometheus/proxy_test.go:121

A note about this last one, the error on line 121, occurs in the call to portforward.NewStreamConn(nil, serverconn, dataStream, errorStream) and NOT in podl.Enqueue(...)

Split line 121 like this:
```go
streamConn := portforward.NewStreamConn(nil, serverconn, dataStream, errorStream)
podl.Enqueue(streamConn)
```
run again, and you will see it more clearly.

Just wondering if anyone has any idea or a direct solution to get rid of these two last race conditions. I have tried a few things but wasn’t able to solve the last two. I would really appreciate any help.

My suggestion if you work on that: comment all the tests that have race conditions except the one you want to fix first.
Feel free to open a new PR with the fix, if you want.

Thank you a lot.

### Test plan for issue:

There are already Unit Tests.

N/A

### DIFF
--- a/test/util/listener/listener.go
+++ b/test/util/listener/listener.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"sync"
 
 	"github.com/Azure/ARO-RP/test/util/bufferedpipe"
 )
@@ -17,8 +18,8 @@ func (addr) Network() string { return "testlistener" }
 func (addr) String() string  { return "testlistener" }
 
 type Listener struct {
-	c      chan net.Conn
-	closed bool
+	c    chan net.Conn
+	once sync.Once
 }
 
 func NewListener() *Listener {
@@ -36,10 +37,10 @@ func (l *Listener) Accept() (net.Conn, error) {
 }
 
 func (l *Listener) Close() error {
-	if !l.closed {
+	l.once.Do(func() {
 		close(l.c)
-		l.closed = true
-	}
+	})
+
 	return nil
 }
 


test/util/listener/listener.go
```diff
@@ -36,10 +38,18 @@ func (l *Listener) Accept() (net.Conn, error) {
...
+	l.RUnlock()
+
+	if !isListenerClosed {
 		close(l.c)
+
+		l.Lock()
 		l.closed = true
+		l.Unlock()
 	}
+
```
facchettos, line 52: The name should probably not be a question but a statement, like `listenerIsClosed`, it reads better in the if.

facchettos, line 52: There can still be a race condition here and you can have a double close with that code. You probably want to have a simple mutex which protects the whole function instead

facchettos (dismissed):

test/util/listener/listener.go
AldoFusterTurpin: Agree. Removed the variable as I don't need it when locking the whole method. Thanks!

AldoFusterTurpin: You're right. Wanted to optimize it with a RWMutex after the "basic" use of a Lock, but that is wrong here. Thanks!

test/util/listener/listener.go
```diff
@@ -19,6 +20,7 @@ func (addr) String() string  { return "testlistener" }
 type Listener struct {
 	c      chan net.Conn
 	closed bool
+	sync.Mutex
```
yehornaumenko, line 23: I would introduce sync.Once here, e.g.
```suggestion
	closeOnce sync.Once
```

test/util/listener/listener.go
```diff
@@ -36,10 +38,14 @@ func (l *Listener) Accept() (net.Conn, error) {
...
 func (l *Listener) Close() error {
+	l.Lock()
+	defer l.Unlock()
+
 	if !l.closed {
 		close(l.c)
 		l.closed = true
 	}
+
 	return nil
```
yehornaumenko, line 49: So we can do this way:
```suggestion
	l.closeOnce.Do(func() {
	    close(l.c)
	    l.closed = true
	})
	return nil
```

test/util/listener/listener.go
AldoFusterTurpin: Thanks for the review @yehornaumenko. 

I am not sure if that would work. Looking at the sync.Once [documentation](https://pkg.go.dev/sync#Once), it says that the action will be performed just once (in total, not "concurrently"), so the behavior of the code will be different as we want it to be executed by a maximum of one goroutine at an exact given time, not in total. 

If we also take a look at the .Do [documentation](https://pkg.go.dev/sync#Once.Do), we see that it says 
_Do calls the function f if and only if Do is being called for the first time for this instance of Once_, 
so we can confirm this is different than what we want.

It is also interesting to look at the Go [source](https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/sync/once.go;l=69;drc=03f6d81fc7d52ec53deb94cff69b63d04e689c24;bpv=0;bpt=1) code to understand how the behavior is different. 

Rejecting the suggestion but an interesting comparison.
😃

AldoFusterTurpin:
Thanks for the review @yehornaumenko.

I am not sure if that would work. Looking at the sync.Once [documentation](https://pkg.go.dev/sync#Once), it says that the action will be performed just once (in total, not "concurrently"), so the behavior of the code will be different as we want it to be executed by a maximum of one goroutine at an exact given time, not in total.

If we also take a look at the .Do [documentation](https://pkg.go.dev/sync#Once.Do), we see that it says
Do calls the function f if and only if Do is being called for the first time for this instance of Once,
so we can confirm this is different than what we want.

It is also interesting to look at the Go [source](https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/sync/once.go;l=69;drc=03f6d81fc7d52ec53deb94cff69b63d04e689c24;bpv=0;bpt=1) code to understand how the behavior is different.

Rejecting the suggestion but an interesting comparison.
😃

test/util/listener/listener.go
AldoFusterTurpin: Good idea. Tks!

petrkotas (approved):
LGTM

s-amann (approved):

PR #2469, opened 2022-10-18
Draft -Move hive config retrieval to separate goroutine

UlrichSchlueter:
Improving error handling for ICM341446812 

Suggestion to separate concerns and move the retrieval of the hive configuration to a separate go routine. That should make error handling and makes the code easier to understand (as opposed to letting x go routines dealing with amongst each other. 
With this approach comes with a rudimentary TTL mechanism

### Test plan for issue:

Suggestion only

n/a

### DIFF
--- a/pkg/util/liveconfig/manager.go
+++ b/pkg/util/liveconfig/manager.go
@@ -9,10 +9,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/sirupsen/logrus"
 	"k8s.io/client-go/rest"
 
 	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/containerservice"
-	"github.com/sirupsen/logrus"
 )
 
 type Manager interface {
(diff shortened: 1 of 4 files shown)


pkg/util/liveconfig/manager.go
```diff
@@ -35,11 +38,58 @@ type prod struct {
...
-func NewProd(location string, managedClustersClient containerservice.ManagedClustersClient) Manager {
-	return &prod{
+func NewProd(location string, managedClustersClient containerservice.ManagedClustersClient, log *logrus.Entry) Manager {
+	p := &prod{
 		location:              location,
 		managedClustersClient: managedClustersClient,
 		cachedCredentials:     make(map[int]*rest.Config),
 		hiveCredentialsMutex:  &sync.RWMutex{},
 	}
+	go p.hiveConfigRetrieveLoop(1, log)
```
s-amann, line 48: Having the index be passed in at this point may make it tricky when we move to using a dynamic index vs hardcoding to 1

mociarain:
Is this still something we can act on?

mociarain:
No longer needed after speaking with @UlrichSchlueter. We took a different approach

PR #2470, opened 2022-10-18
Liveconfig: Global AKS shard admin credentials

darthhexx:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15877607

Switches to using AKS admin dynamic credential loading based on the configured location and shard number. We only have a single cluster shard deployed per location, so its currently hard coded to 1.

Implements a global map of Hive (AKS) shard kubeConfigs, instead of calling the `liveConfig.HiveRestConfig` in the cluster.monitor code path which causes contention and slows down a critical code path, e.g. API health monitoring.

For local development, it is still possible to override the dynamic loading of the AKS config by setting the `HIVE_KUBE_CONFIG_PATH_1` environment variable to a valid kubeconfig path. The kubeconfig can be generated locally by running `make aks.kubeconfig`.

Please note when testing Hive cluster creations you have to be connected to the AKS APN by running: `sudo openvpn secrets/vpn-aks-$LOCATION.ovpn`.

### Test plan for issue:

I have created many clusters and modfied an existing unit test.

No.


s-amann (approved):
lgtm, I like listing out the clusters in search for ours versus hardcoding the RG pattern. Thank you for working on this!

pkg/util/liveconfig/hive.go
```diff
@@ -21,14 +23,52 @@ const (
...
 	if err != nil {
 		return nil, err
 	}
 
-	clientconfig, err := clientcmd.NewClientConfigFromBytes(res)
+	return parseKubeconfig(*res.Kubeconfigs)
+}
+
+func parseKubeconfig(credentials []mgmtcontainerservice.CredentialResult) (*rest.Config, error) {
+	clientconfig, err := clientcmd.NewClientConfigFromBytes(*credentials[0].Value)
```
s-amann, line 71: According to this doc, the `value` is base64 encoded, does `NewCLientConfigFromBytes` account for this?
https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/list-cluster-user-credentials?tabs=Go#credentialresult

pkg/util/liveconfig/hive.go
```diff
@@ -41,37 +81,32 @@ func parseKubeconfig(credentials []mgmtcontainerservice.CredentialResult) (*rest
...
-	res, err := p.managedClustersClient.ListClusterUserCredentials(ctx, rpResourceGroup, rpResourceName, "")
+	kubeConfig, err := getAksKubeconfig(ctx, p.managedClustersClient, p.location, index, credentialType)
 	if err != nil {
 		return nil, err
 	}
 
-	parsed, err := parseKubeconfig(*res.Kubeconfigs)
-	if err != nil {
-		return nil, err
+	p.hiveCredentialsMutex.Lock()
```
UlrichSchlueter, line 103: Consideration: Moving the lock here reduces the time we block the individual goroutines but will likely see multiple calls to the Azure API. We add random delays up to a minute at monitor goroutines creation time, so the number of calls are a function of the number of clusters a monitor instance is handling (currently ~100 max) and the runtime of the Azure API calls, which should be below 500 ms => if the randomisation is good enough we should be fine and not trigger denial of service attacks to the Azure API.

UlrichSchlueter: Consideration 2: If all calls fail (current situation) and the cache remains empty every monitor goroutine will trigger an API Call every minute, so we'll create 1 to 2 call/sec, which still should be fine.

darthhexx:
When debugging the monitor workers closely, after the switch to Warning logs, I discovered a log line that I had missed before:
```sh
no Auth Provider found for name "azure"
```
After taking a look, I noticed that the azure plugin was missing from our vendored `k8s.io/client-go` module so I referenced it and re-vendored. Problem solved....not quite.
```sh
To sign in, use a web browser to open the page https://microsoft.com/devicelogin and enter the code EXD2HBR7Y to authenticate.
```
Turns out that that the plugin requires [manual intervention before it will work](https://pkg.go.dev/k8s.io/client-go/plugin/pkg/client/auth/azure#section-readme). 🤦 Utterly useless to us.

As a result, I am going to change this PR to use admin credentials for all AKS access, because it does not suffer from the same issues and retrieves a bearer token without fail.

pkg/util/liveconfig/hive.go
darthhexx: I went back to the original, to avoid any authentication "stampedes". Since we already have the cluster admin role assigned to the RP MSI for all of the AKS clusters in production, we aren't going to see the level of credential requests we have been and we should be fine from that aspect.

The only consideration is that the context is [only set to 50 seconds](https://github.com/Azure/ARO-RP/blob/aacda3a665eca2bf7faa9f9b768c63d20951f95a/pkg/monitor/worker.go#L192) so in cases where Azure does respond slowly we may drop some metrics for a collection cycle, but it should only be once.

s-amann (approved):

pkg/util/liveconfig/dev.go
```diff
@@ -12,24 +12,45 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
-func (d *dev) HiveRestConfig(ctx context.Context, index int) (*rest.Config, error) {
+func (d *dev) HiveRestConfig(ctx context.Context, shard int) (*rest.Config, error) {
 	// Indexes above 0 have _index appended to them
```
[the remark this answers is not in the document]
darthhexx: Thanks, fixed.

pkg/util/liveconfig/hive.go
darthhexx: I have refactored the method of fetching the kubeConfigs, taking to a global level, because we only need to do it once for all monitors and yet it was in the critical monitoring path of all of them.

pkg/util/liveconfig/hive.go
```diff
@@ -21,14 +23,46 @@ const (
...
+			if *cluster.Name == aksClusterName && *cluster.Location == location {
+				aksCluster = &cluster
+				break outerLoop
+			}
+		}
+		err = aksClusters.NextWithContext(ctx)
+		if err != nil {
+			return nil, err
+		}
+	}
```
facchettos, line 48: it would be nicer to have that in a function so we don't have to use loop annotations

pkg/monitor/worker.go
```diff
@@ -80,6 +97,21 @@ func (mon *monitor) changefeed(ctx context.Context, baseLog *logrus.Entry, stop
...
+					_, exists := mon.hiveShardConfigs[shard]
+					mon.shardMutex.RUnlock()
+					if !exists {
+						// set this to `nil` so cluster monitors will ignore it until its populated with config
+						mon.hiveShardConfigs[shard] = nil
+						go mon.populateHiveShardRestConfig(ctx, shard)
+					}
+
 					// TODO: improve memory usage by storing a subset of doc in mon.docs
 					mon.upsertDoc(doc)
```
facchettos, line 116: could we have that in a separate function? the function is already long enough as it is

pkg/monitor/worker.go
```diff
@@ -80,6 +97,21 @@ func (mon *monitor) changefeed(ctx context.Context, baseLog *logrus.Entry, stop
 							fps == api.ProvisioningStateDeleting):
 					mon.deleteDoc(doc)
 				default:
+					// check if we have a Hive shard config and if not start the recursive auth call
+					// in the future we will have the shard index set on the api.OpenShiftClusterDocument
+					// but for now we simply select Hive (AKS) shard 1
+					// e.g. shard := mon.hiveShardConfigs[doc.shardIndex]
+					shard := 1
```
facchettos, line 104: seems to be somewhat duplicated with line 233

pkg/monitor/worker.go
```diff
@@ -198,10 +230,15 @@ func (mon *monitor) workOne(ctx context.Context, log *logrus.Entry, doc *api.Ope
...
 	}
 
-	hiveRestConfig, err := mon.liveConfig.HiveRestConfig(ctx, 1)
-	if err != nil {
-		// TODO(hive): Update to fail once we have Hive everywhere in prod and dev
-		log.Info(err)
+	// in the future we will have the shard index set on the api.OpenShiftClusterDocument but for
+	// now we simply select Hive (AKS) shard 1 because we only have one and sharding is yet to come
+	// e.g. shard := mon.hiveShardConfigs[doc.shardIndex]
+	shard := 1
```
facchettos, line 236: seems to be somewhat duplicated with line 100

pkg/monitor/worker.go
```diff
@@ -80,6 +97,21 @@ func (mon *monitor) changefeed(ctx context.Context, baseLog *logrus.Entry, stop
...
 				default:
+					// check if we have a Hive shard config and if not start the recursive auth call
+					// in the future we will have the shard index set on the api.OpenShiftClusterDocument
+					// but for now we simply select Hive (AKS) shard 1
+					// e.g. shard := mon.hiveShardConfigs[doc.shardIndex]
+					shard := 1
+
+					mon.shardMutex.RLock()
+					_, exists := mon.hiveShardConfigs[shard]
+					mon.shardMutex.RUnlock()
```
facchettos, line 108: we should do an accessor for that field if we have to handle mutexes

pkg/monitor/worker.go
```diff
@@ -198,10 +230,15 @@ func (mon *monitor) workOne(ctx context.Context, log *logrus.Entry, doc *api.Ope
...
-	if err != nil {
-		// TODO(hive): Update to fail once we have Hive everywhere in prod and dev
-		log.Info(err)
+	// in the future we will have the shard index set on the api.OpenShiftClusterDocument but for
+	// now we simply select Hive (AKS) shard 1 because we only have one and sharding is yet to come
+	// e.g. shard := mon.hiveShardConfigs[doc.shardIndex]
+	shard := 1
+	mon.shardMutex.RLock()
+	hiveRestConfig, exists := mon.hiveShardConfigs[shard]
+	mon.shardMutex.RUnlock()
```
facchettos, line 239: we should do an accessor for that field if we have to handle mutexes

pkg/util/liveconfig/dev.go
```diff
@@ -12,24 +12,45 @@ import (
...
+			return nil, err
+		}
+		return restConfig, nil
 	}
 
-	restConfig, err := clientcmd.BuildConfigFromFlags("", kubeConfigPath)
+	// Hive shards are planned but not deployed or implemented yet
+	d.hiveCredentialsMutex.RLock()
+	cached, exists := d.cachedCredentials[shard]
+	d.hiveCredentialsMutex.RUnlock()
```
facchettos, line 35: we should do an accessor for that field if we have to handle mutexes

pkg/util/liveconfig/hive.go
```diff
@@ -41,37 +75,29 @@ func parseKubeconfig(credentials []mgmtcontainerservice.CredentialResult) (*rest
...
-func (p *prod) HiveRestConfig(ctx context.Context, index int) (*rest.Config, error) {
-	// NOTE: This RWMutex locks on a fetch for any index for simplicity, rather
-	// than a more granular per-index lock. As of the time of writing, multiple
-	// Hive shards are planned but unimplemented elsewhere.
+func (p *prod) HiveRestConfig(ctx context.Context, shard int) (*rest.Config, error) {
+	// Hive shards are planned but not implemented yet
 	p.hiveCredentialsMutex.RLock()
-	cached, ext := p.cachedCredentials[index]
+	cached, exists := p.cachedCredentials[shard]
 	p.hiveCredentialsMutex.RUnlock()
```
facchettos, line 82: we should do an accessor for that field if we have to handle mutexes

pkg/util/liveconfig/hive.go
```diff
@@ -41,37 +75,29 @@ func parseKubeconfig(credentials []mgmtcontainerservice.CredentialResult) (*rest
...
-	parsed, err := parseKubeconfig(*res.Kubeconfigs)
-	if err != nil {
-		return nil, err
-	}
+	p.cachedCredentials[shard] = kubeConfig
+	p.hiveCredentialsMutex.Unlock()
 
-	p.cachedCredentials[index] = parsed
-	return rest.CopyConfig(parsed), nil
+	return rest.CopyConfig(kubeConfig), nil
```
facchettos, line 100: defer would be nice here

pkg/util/liveconfig/manager.go
```diff
@@ -21,10 +21,20 @@ type Manager interface {
 	DefaultInstallerPullSpecOverride(context.Context) string
 }
 
-type dev struct{}
+type dev struct {
+	location              string
+	managedClustersClient containerservice.ManagedClustersClient
+
+	hiveCredentialsMutex *sync.RWMutex
```
facchettos, line 28: the pointer isn't need here, (will need to probably change prod too), because we are only using the pointers to that struct and using it as an interface. that allows to not instantiate the mutex

pkg/util/liveconfig/manager.go
```diff
@@ -21,10 +21,20 @@ type Manager interface {
 	DefaultInstallerPullSpecOverride(context.Context) string
 }
 
-type dev struct{}
+type dev struct {
```
facchettos, line 24: shouldn't that struct definition/creation be in dev.go instead? same comment for prod (but in prod.go)

facchettos (changes requested):

pkg/util/liveconfig/dev.go
s-amann: I agree, if we need to always lock before reading and unlock after we can ensure this is done by wrapping them together in an accessor.

pkg/monitor/worker.go
```diff
@@ -18,6 +18,23 @@ import (
...
+// This function will recurse until such time as it has a config to add to the global Hive shard map
+// Note that because the mon.hiveShardConfigs[shard] is set to `nil` when its created the cluster
+// monitors will simply ignore Hive stats until this function populates the config
+func (mon *monitor) populateHiveShardRestConfig(ctx context.Context, shard int) {
+	hiveRestConfig, err := mon.liveConfig.HiveRestConfig(ctx, shard)
+	if err != nil {
+		mon.baseLog.Printf("error fetching Hive kubeconfig for shard %d: %s", shard, err.Error())
+		mon.baseLog.Printf("pausing for a minute before retrying...")
+		time.Sleep(time.Duration(60 * time.Second))
+		mon.populateHiveShardRestConfig(ctx, shard)
```
UlrichSchlueter, line 30: For permanent errors like our current 403 we might run out of memory over time with one recursive call per minute.  Would it make sense to change this to a loop ?

facchettos: Agree with @UlrichSchlueter , that or having a different context than the one from the changefeed so there is a way out.

pkg/util/azureclient/mgmt/containerservice/managedcluster_addons.go
```diff
@@ -11,9 +11,14 @@ import (
 
 // RegistriesAddons contains addons for RegistriesClient
 type ManagedClustersAddons interface {
```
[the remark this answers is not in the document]
darthhexx: I never put it there, but it sure looks like copy-pasta. Will remove it.

pkg/util/liveconfig/dev.go
```diff
@@ -12,24 +12,45 @@ import (
...
 	if err != nil {
+		d.hiveCredentialsMutex.Unlock()
 		return nil, err
 	}
 
-	return restConfig, nil
+	d.cachedCredentials[shard] = kubeConfig
+	d.hiveCredentialsMutex.Unlock()
+
+	return rest.CopyConfig(kubeConfig), nil
```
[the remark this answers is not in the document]
darthhexx: In this time sensitive case it makes more sense to call the unlock deterministically at the exact moment it can be unlocked instead of relying on Golang to do it after the function exits.

pkg/monitor/worker.go
darthhexx: Done.

pkg/util/liveconfig/hive.go
darthhexx: Good call.

pkg/monitor/worker.go
darthhexx: Intentionally duplicated so the reader is aware of why the shard is being set to 1 here and also marks the place where we can come and change it to  `mon.hiveShardConfigs[doc.shardIndex]`. Do you really think it needs to be removed or just reworded to not be a copy?

pkg/monitor/worker.go
darthhexx: Intentionally duplicated. See the comment below.

pkg/util/liveconfig/manager.go
darthhexx: Changed.

pkg/monitor/worker.go
darthhexx: I refactored this not to use recursion.

pkg/util/liveconfig/manager.go
darthhexx: Agreed. I never put them there in the first place, but I've refactored both of them.

pkg/util/liveconfig/dev.go
darthhexx: Good call, done.

pkg/monitor/worker.go
darthhexx: I modified the comment after adding the accessor functions. Have a look in the next round and let me know if it still looks duplicated.

pkg/util/liveconfig/manager.go
darthhexx: Well now I know why its like that. The mocking is quite tricky with the Azure SDK. In the interest of reducing complexity of this PR I am going to leave it for now and we can do a follow up PR to work on it when time isn't of the essence.

pkg/util/liveconfig/dev.go
darthhexx: When looking at implementing this, the mutex is only locked/unlocked in this one function. There seems to be a lot of them because there is one for dev and one for prod. It made sense to add accessors in the monitors, but it doesn't really here.

pkg/util/liveconfig/dev.go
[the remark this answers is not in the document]
facchettos: how much time are we saving by not using defer? it's a few nanoseconds at best, if this is the determining factor, we have much bigger problems. Unless there is data that shows that this is the problem in the monitor issue, I think we should use defer here since it's reducing code complexity.

pkg/monitor/worker.go
```diff
@@ -18,6 +20,30 @@ import (
...
+			return
+		}
+
+		mon.baseLog.Warn(fmt.Sprintf("error fetching Hive kubeconfig for shard %d", shard))
+		if err != nil {
+			mon.baseLog.Error(err.Error())
+		}
+
+		mon.baseLog.Info("pausing for a minute before retrying...")
+		time.Sleep(time.Duration(60 * time.Second))
```
petrkotas, line 43: `time.Sleep(60 * time.Second)` should be enough.
Moreover, I am tempted to prolong the interval to give AKS some rest. `2-5` minutes?

pkg/monitor/worker.go
facchettos: we could probably have shard as a package const so we don't need to look for every instance of that pattern when we remove it, just delete the const and fix all the errors instead

petrkotas (changes requested):
Hi @darthhexx thank you for the astonishing work. I have few minor questions and considerations. Besides I would be happy to get this one in.

pkg/util/liveconfig/manager.go
facchettos: has it ?

pkg/monitor/worker.go
```diff
@@ -18,6 +20,30 @@ import (
...
+	var err error
+
+	for {
+		hiveRestConfig, err = mon.liveConfig.HiveRestConfig(ctx, shard)
+		if hiveRestConfig != nil {
+			mon.setHiveShardConfig(shard, hiveRestConfig)
+			return
+		}
+
+		mon.baseLog.Warn(fmt.Sprintf("error fetching Hive kubeconfig for shard %d", shard))
```
s-amann, line 37: This will be a nice log to end up in a dashboard, I propose we combine the two logs (this warn and the below error). By having static text along with the the error returned it will make identifying what the error was when fetching the kubeconfig easier.

pkg/monitor/worker.go
```diff
@@ -198,10 +236,12 @@ func (mon *monitor) workOne(ctx context.Context, log *logrus.Entry, doc *api.Ope
...
-	hiveRestConfig, err := mon.liveConfig.HiveRestConfig(ctx, 1)
-	if err != nil {
-		// TODO(hive): Update to fail once we have Hive everywhere in prod and dev
-		log.Info(err)
+	// once sharding is implemented, we will have the shard set on the api.OpenShiftClusterDocument
+	// e.g. shard := mon.hiveShardConfigs[doc.shardIndex]
+	shard := 1
+	hiveRestConfig, exists := mon.getHiveShardConfig(shard)
+	if !exists {
+		log.Errorf("no hiveShardConfigs set for shard %d", shard)
```
s-amann, line 244: IMO this should be a warning since we 'plan' on not having a config and already have an error log when we can't fetch it.

pkg/util/liveconfig/hive.go
darthhexx: Done.

pkg/util/liveconfig/manager.go
darthhexx: Well, funny story... it was, but when I switched back to not moving hive -> prod, etc. (because of mock-hell) I reverted it by mistake. Fixed in the latest push.

pkg/util/liveconfig/dev.go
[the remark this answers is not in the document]
darthhexx: I missed those release notes, thanks for the background info. I've switched back to defer.

pkg/hive/install.go
```diff
@@ -112,17 +115,24 @@ func servicePrincipalSecretForInstall(oc *api.OpenShiftCluster, sub *api.Subscri
 	if isDevelopment {
 		// In development mode, load in the proxy certificates so that clusters
 		// can be accessed from a local (not in Azure) Hive
-		proxyCert, err := os.ReadFile("secrets/proxy.crt")
+		// This assumes we are running from an ARO-RP checkout in development
```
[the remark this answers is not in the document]
darthhexx: It is not directly relevant, but it prevents anybody from successfully testing this PR in the development environment, so it's necessary by association.

pkg/util/liveconfig/dev.go
```diff
@@ -12,24 +12,45 @@ import (
...
+	cached, exists := d.cachedCredentials[shard]
+	d.hiveCredentialsMutex.RUnlock()
+	if exists {
+		return rest.CopyConfig(cached), nil
+	}
+
+	// Lock the RWMutex as we're starting to fetch so that new readers will wait
+	// for the existing Azure API call to be done.
+	d.hiveCredentialsMutex.Lock()
+
```
s-amann, line 43: nit: `hive.go` uses defer for unlocking. Can we update `dev.go` to match?

s-amann (dismissed):

pkg/env/core.go
```diff
@@ -35,16 +35,17 @@ func (c *core) IsLocalDevelopmentMode() bool {
...
-		return liveconfig.NewDev(), nil
-	}
-
 	msiAuthorizer, err := c.NewMSIAuthorizer(MSIContextRP, c.Environment().ResourceManagerEndpoint)
 	if err != nil {
 		return nil, err
 	}
 
 	mcc := containerservice.NewManagedClustersClient(c.Environment(), c.SubscriptionID(), msiAuthorizer)
+
```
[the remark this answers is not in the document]
darthhexx: It works and is necessary for us to be able to [dynamically fetch the AKS kubeconfig here](https://github.com/Azure/ARO-RP/pull/2470/files#diff-bf806692328781c7ebc9962bcdfe5d97530ed6f4ff2699090035ceeae1ed7f6cR44) as we do in production.

pkg/monitor/worker.go
darthhexx: So the reason for keeping them separate is for generating a Sev 1 alert when we get N of these consecutively, which is why I intentionally kept the low level alert out of it. Isn't the clean separation better?

pkg/monitor/worker.go
darthhexx: I spent quite a bit of time deciding on the intervals actually and I think 60 seconds if a good compromise for not wanting to be "aggressive" and also needing to get the config quickly. We currently [poll the AAD endpoint every 10 seconds](https://github.com/Azure/ARO-RP/blob/f3d5f276bb2f07ec05d428aec1aaf575479d98fa/pkg/util/aad/aad.go#L46), when we are trying to get valid credentials, so 60 doesn't seem to high in this instance.

pkg/util/liveconfig/dev.go
darthhexx: Missed that, thanks.

pkg/monitor/worker.go
darthhexx: The only place we plan on not having config is in FairFax. I've changed it, but we should change it back once we go live there.

facchettos (approved):

petrkotas (approved):
I see all concerns are addressed.

pkg/monitor/worker.go
s-amann: I'm ok with that reasoning, thank you for the clarification.

PR #2474, opened 2022-10-19
Set OCP pullspec back to quay.io after the Hive installation completes

darthhexx:

Fixes: https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/15728770

Due to the ACR Domain constraints in production, we  set the domain to the ACR Domain during the cluster installation. Once the installation has succeeded we need to put the quay.io domain back, because that is the expected customer experience.

### Test plan for issue:

Tested by creating numerous clusters.

No.

### DIFF
--- a/pkg/cluster/disableupdates.go
+++ b/pkg/cluster/disableupdates.go
@@ -5,7 +5,9 @@ package cluster
 
 import (
 	"context"
+	"strings"
 
+	configv1 "github.com/openshift/api/config/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
 )
@@ -19,6 +21,18 @@ func (m *manager) disableUpdates(ctx context.Context) error {
 
 		cv.Spec.Upstream = ""
 		cv.Spec.Channel = ""
+
+		// when installing via Hive we end up with an ACR image pullspec and we should leave it for the customer as quay.io
+		if m.installViaHive {
+			version, err := m.openShiftVersionFromVersion(ctx)
+			if err != nil {
+				return err
+			}
+			cv.Spec.DesiredUpdate = &configv1.Update{
+				Version: version.Properties.Version,
+				Image:   strings.Replace(version.Properties.OpenShiftPullspec, m.env.ACRDomain(), "quay.io", 1),
+			}
+		}
 
 		_, err = m.configcli.ConfigV1().ClusterVersions().Update(ctx, cv, metav1.UpdateOptions{})
 		return err
--- a/pkg/cluster/disableupdates.go
+++ b/pkg/cluster/disableupdates.go
@@ -22,7 +22,7 @@ func (m *manager) disableUpdates(ctx context.Context) error {
 		cv.Spec.Upstream = ""
 		cv.Spec.Channel = ""
 
-		// when installing via Hive we end up with an ACR image pullspec and we should leave it for the customer as quay.io
+		// when installing via Hive we replace the quay.io domain with the ACR domain, so now we set it back to the expected domain
 		if m.installViaHive {
 			version, err := m.openShiftVersionFromVersion(ctx)
 			if err != nil {


pkg/cluster/disableupdates.go
```diff
@@ -20,6 +22,18 @@ func (m *manager) disableUpdates(ctx context.Context) error {
...
+		// when installing via Hive we replace the quay.io domain with the ACR domain, so now we set it back to the expected domain
+		if m.installViaHive {
+			version, err := m.openShiftVersionFromVersion(ctx)
+			if err != nil {
+				return err
+			}
+			cv.Spec.DesiredUpdate = &configv1.Update{
+				Version: version.Properties.Version,
+				Image:   strings.Replace(version.Properties.OpenShiftPullspec, m.env.ACRDomain(), "quay.io", 1),
+			}
```
facchettos, line 34: is there a version change when we change the image?

m1kola:
@darthhexx could you please provide a bit more context for this?

darthhexx:
@m1kola we will be using the ACR Domain for the OpenShift and ARO-Installer images for multi-version, because we can't use quay.io in production. The installation process currently creates the default version and [replaces the version.InstallStream domain with the ACR Domain](https://github.com/Azure/ARO-RP/blob/4e36c1fe0498cb8e2463b5e967a5bd2d890e3d0b/pkg/cluster/install_version.go#L47).

Once the cluster has completed its installation, the URL in cluster version is obviously the ACR DOmain, but that is not what it is expected to be. This PR switches it back to quay.io to ensure that the customer experience is what is expected.

pkg/cluster/disableupdates.go
darthhexx: Potentially, yes. I did not test that though.

m1kola:
@darthhexx do we need to update [this doc](https://learn.microsoft.com/en-us/azure/openshift/howto-restrict-egress)? quay.io is currently in the optional list. Sounds like it is a requirement now.

pkg/cluster/disableupdates.go
facchettos: Would be nice if you could test it

petrkotas: From the doc, it seems it is relevant only when doing upgrade. However we are doing install. So we indeed might just be able to skip this field.

petrkotas (commented):
Minor non-blocking comment.

pkg/cluster/disableupdates.go
darthhexx: @facchettos I certainly didn't have time to do so last week, but I'll look after my shift today.

darthhexx:
@m1kola we don't have to update the docs, because quay.io is still automatically sent to the ACR domain using the content policy in the cluster.

The only reason we need to set it to the ACR before the install via Hive, and back to quay.io afterwards, is because we don't have an easy way to get content policies to work inside AKS. [This method was attempted](https://github.com/containerd/containerd/blob/0def98e462706286e6eaeff4a90be22fda75e761/docs/hosts.md#host-fields-in-the-toml-table-format) but it does not support token authentication and does not work for us.

pkg/cluster/disableupdates.go
darthhexx: As expected its functional, but not a great user experience, because the `Version` in `DesiredUpdate` is an empty string as a result.
```yaml
$ oc get clusterversion -oyaml
apiVersion: v1
items:
- apiVersion: config.openshift.io/v1
  kind: ClusterVersion
...snip...
  spec:
    clusterID: 59568119-41e1-4d57-a513-a988d527ac6a
    desiredUpdate:
      force: false
      image: quay.io/openshift-release-dev/ocp-release@sha256:58a68bea6f4d5c23a10596953f50529417b33c331195deb9041d9fecfa95b400
      version: ""
  status:
    availableUpdates: null
...snip...
```
This should be left as it is, so that the expected version is listed.

darthhexx:
> Minor non-blocking comment.

@petrkotas as [mentioned in the other thread above](https://github.com/Azure/ARO-RP/pull/2474#discussion_r1002920621), we should be setting the so that the `DesiredUpdate.Version` field is set accordingly and not listed as an empty string.

pkg/cluster/disableupdates.go
facchettos: indeed

facchettos (approved):

petrkotas (approved):
Lets get it in.

PR #2476, opened 2022-10-19
Clean up dnsmasq machineconfig when machineconfigpool is deleted

tsatam:

Fixes: https://issues.redhat.com/browse/ARO-1490

Deletes the `99-%s-aro-dns` MachineConfig made by the operator for each MachineConfigPool when that MachineConfigPool is deleted. This will prevent alerts firing for these MachineConfigs when they do not apply to any MachineConfigPool. 

### Test plan for issue:

Unit tests and an E2E test were added for the overall MachineConfig creation as well as the delete case.

No.

### DIFF
--- a/pkg/operator/controllers/dnsmasq/cluster_controller.go
+++ b/pkg/operator/controllers/dnsmasq/cluster_controller.go
@@ -65,14 +65,7 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 		return reconcile.Result{}, err
 	}
 
-	roles := make([]string, 0, len(mcps.Items))
-	for _, mcp := range mcps.Items {
-		if mcp.GetDeletionTimestamp() == nil {
-			roles = append(roles, mcp.Name)
-		}
-	}
-
-	err = reconcileMachineConfigs(ctx, instance, r.dh, roles...)
+	err = reconcileMachineConfigs(ctx, instance, r.dh, mcps.Items...)
 	if err != nil {
 		r.log.Error(err)
 		return reconcile.Result{}, err
@@ -81,10 +74,15 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 	return reconcile.Result{}, nil
 }
 
-func reconcileMachineConfigs(ctx context.Context, instance *arov1alpha1.Cluster, dh dynamichelper.Interface, roles ...string) error {
+func reconcileMachineConfigs(ctx context.Context, instance *arov1alpha1.Cluster, dh dynamichelper.Interface, roles ...mcv1.MachineConfigPool) error {
 	var resources []kruntime.Object
 	for _, role := range roles {
-		resource, err := dnsmasq.MachineConfig(instance.Spec.Domain, instance.Spec.APIIntIP, instance.Spec.IngressIP, role, instance.Spec.GatewayDomains, instance.Spec.GatewayPrivateEndpointIP)
+		resource, err := dnsmasq.MachineConfig(instance.Spec.Domain, instance.Spec.APIIntIP, instance.Spec.IngressIP, role.Name, instance.Spec.GatewayDomains, instance.Spec.GatewayPrivateEndpointIP)
+		if err != nil {
+			return err
+		}
+
+		err = dynamichelper.SetControllerReferences([]kruntime.Object{resource}, &role)
 		if err != nil {
 			return err
 		}
@@ -92,12 +90,7 @@ func reconcileMachineConfigs(ctx context.Context, instance *arov1alpha1.Cluster,
 		resources = append(resources, resource)
 	}
 
-	err := dynamichelper.SetControllerReferences(resources, instance)
-	if err != nil {
-		return err
-	}
-
-	err = dynamichelper.Prepare(resources)
+	err := dynamichelper.Prepare(resources)
 	if err != nil {
 		return err
 	}
--- a/pkg/operator/controllers/dnsmasq/cluster_controller.go
+++ b/pkg/operator/controllers/dnsmasq/cluster_controller.go
@@ -86,15 +86,15 @@ func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
-func reconcileMachineConfigs(ctx context.Context, instance *arov1alpha1.Cluster, dh dynamichelper.Interface, roles ...mcv1.MachineConfigPool) error {
+func reconcileMachineConfigs(ctx context.Context, instance *arov1alpha1.Cluster, dh dynamichelper.Interface, mcps ...mcv1.MachineConfigPool) error {
 	var resources []kruntime.Object
-	for _, role := range roles {
-		resource, err := dnsmasq.MachineConfig(instance.Spec.Domain, instance.Spec.APIIntIP, instance.Spec.IngressIP, role.Name, instance.Spec.GatewayDomains, instance.Spec.GatewayPrivateEndpointIP)
+	for _, mcp := range mcps {
+		resource, err := dnsmasq.MachineConfig(instance.Spec.Domain, instance.Spec.APIIntIP, instance.Spec.IngressIP, mcp.Name, instance.Spec.GatewayDomains, instance.Spec.GatewayPrivateEndpointIP)
 		if err != nil {
 			return err
 		}
 
-		err = dynamichelper.SetControllerReferences([]kruntime.Object{resource}, &role)
+		err = dynamichelper.SetControllerReferences([]kruntime.Object{resource}, &mcp)
 		if err != nil {
 			return err
 		}
--- a/pkg/operator/controllers/dnsmasq/cluster_controller.go
+++ b/pkg/operator/controllers/dnsmasq/cluster_controller.go
@@ -74,6 +74,18 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 	return reconcile.Result{}, nil
 }
 
+// SetupWithManager setup our mananger
+func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	aroClusterPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.GetName() == arov1alpha1.SingletonClusterName
+	})
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&arov1alpha1.Cluster{}, builder.WithPredicates(aroClusterPredicate)).
+		Named(ClusterControllerName).
+		Complete(r)
+}
+
 func reconcileMachineConfigs(ctx context.Context, instance *arov1alpha1.Cluster, dh dynamichelper.Interface, roles ...mcv1.MachineConfigPool) error {
 	var resources []kruntime.Object
 	for _, role := range roles {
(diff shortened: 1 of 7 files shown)


tsatam:
@microsoft-github-policy-service agree [company="Red Hat"]

tsatam:

pkg/operator/controllers/dnsmasq/machineconfigpool_controller_test.go
```diff
@@ -0,0 +1,254 @@
...
+			request: ctrl.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: "",
+					Name:      "custom",
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "MachineConfigPool marked for deletion with no finalizers does nothing",
```
rolandmkunkel, line 94: why do we test this case? This can only happen if you delete the MCP before it was reconciled once, right?

pkg/operator/controllers/dnsmasq/machineconfigpool_controller_test.go
```diff
@@ -0,0 +1,254 @@
...
+			request: ctrl.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: "",
+					Name:      "custom",
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "MachineConfigPool with no finalizer adds finalizer",
```
rolandmkunkel, line 196: maybe we can also test that the finalizer is not readded after the mcp has been marked for deletion and the finalizer was run

rolandmkunkel (commented):
Since we are just deleting the 99-%s-aro-dns machineConfig when the machineConfigPool is deleted, could we also just set an owner reference on it and let it be deleted together with the machineConfigPool?

pkg/operator/controllers/dnsmasq/machineconfigpool_controller_test.go
tsatam: My understanding was that this scenario can also happen after the finalizer is handled and removed, but before the MCP itself gets deleted. If that is not the case, this test case and the corresponding guard check (machineconfigpool_controller.go#L69-71) could be removed.

pkg/operator/controllers/dnsmasq/machineconfigpool_controller_test.go
tsatam: The `MachineConfigPool marked for deletion with no finalizers does nothing` case validates this by ensuring the MCP is not modified.

tsatam:
> Since we are just deleting the 99-%s-aro-dns machineConfig when the machineConfigPool is deleted, could we also just set an owner reference on it and let it be deleted together with the machineConfigPool?

I experimented with this by replacing the existing owner of the MachineConfig set by the operator (the singleton Cluster) with the MachineConfigPool, however the MachineConfig would not get deleted when deleting the MachineConfigPool. I could be missing something, though.

tsatam:
> > > Since we are just deleting the 99-%s-aro-dns machineConfig when the machineConfigPool is deleted, could we also just set an owner reference on it and let it be deleted together with the machineConfigPool?
> > 
> > 
> > I experimented with this by replacing the existing owner of the MachineConfig set by the operator (the singleton Cluster) with the MachineConfigPool, however the MachineConfig would not get deleted when deleting the MachineConfigPool. I could be missing something, though.
> 
> Could you push the code?

Sorry for the delay, returning to this after a while.

I have an experimental branch pushed up with an owner reference implementation: https://github.com/tsatam/ARO-RP/commit/affd8391532a75a5ce7c459c6d0919f36b141a8f

I was able to get it to perform the cleanup as desired, but only by removing the existing OwnerReference that the generated MachineConfigs had to the Cluster resource. I am not sure if this is a safe and backwards compatible change, but if it is, I will merge it into this PR's branch.

test/e2e/operator.go
```diff
@@ -508,3 +509,70 @@ var _ = Describe("ARO Operator - ImageConfig Reconciler", func() {
...
+
+var _ = Describe("ARO Operator - dnsmasq", func() {
+	const (
+		timeout = 1 * time.Minute
+		polling = 10 * time.Second
+	)
+	mcpName := "test-aro-custom-mcp"
+	mcName := fmt.Sprintf("99-%s-aro-dns", mcpName)
+
+	ctx := context.Background()
```
m1kola, line 521: Please use Ginkgo context. See https://github.com/Azure/ARO-RP/pull/2505

test/e2e/operator.go
```diff
@@ -508,3 +509,70 @@ var _ = Describe("ARO Operator - ImageConfig Reconciler", func() {
...
+			Kind:       "MachineConfigPool",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: mcpName,
+		},
+		Spec: mcv1.MachineConfigPoolSpec{},
+	}
+
+	getMachineConfigNames := func(g Gomega) []string {
+		machineConfigs, err := clients.MachineConfig.MachineconfigurationV1().MachineConfigs().List(ctx, metav1.ListOptions{})
```
m1kola, line 535: If we define `getMachineConfigNames` like this - we capture `ctx` from the `Describe` node. We should be using context from `Eventually`.

m1kola (changes requested):
I only looked into e2es briefly. Did not pay a lot of attention to rest of the PR.

test/e2e/operator.go
tsatam: I've updated these tests to use the Ginkgo context.

pkg/operator/controllers/dnsmasq/machineconfigpool_controller_test.go
```diff
@@ -0,0 +1,166 @@
...
+					Name:       "custom",
+					Finalizers: []string{MachineConfigPoolControllerName},
+				},
+				Status: mcv1.MachineConfigPoolStatus{},
+				Spec:   mcv1.MachineConfigPoolSpec{},
+			},
+		)
+
+		dh := fakeDh(controller)
+		dh.EXPECT().Ensure(gomock.Any(), gomock.AssignableToTypeOf(&mcv1.MachineConfig{})).Times(1)
```
SudoBrendan, line 144: More of the same - I'd rather see a lookup/verification rather than relying on `Ensure`.

SudoBrendan (commented):
This is looking really close, great work! I like the format of these tests, and always great to see more added. Before a merge, I'd like to see a stricter end result test (behavior check) rather than a function call check in a few places.

pkg/operator/controllers/dnsmasq/cluster_controller.go
```diff
@@ -79,38 +74,38 @@ func (r *ClusterReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-func reconcileMachineConfigs(ctx context.Context, instance *arov1alpha1.Cluster, dh dynamichelper.Interface, roles ...string) error {
+// SetupWithManager setup our mananger
+func (r *ClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	aroClusterPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.GetName() == arov1alpha1.SingletonClusterName
+	})
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&arov1alpha1.Cluster{}, builder.WithPredicates(aroClusterPredicate)).
+		Named(ClusterControllerName).
```
cadenmarchese, line 85: Consider a `Watches` for machine config pool resources (if needed)

pkg/operator/controllers/dnsmasq/cluster_controller_test.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		_, err := r.Reconcile(context.Background(), ctrl.Request{})
+
+		if err != nil {
+			t.Errorf("wanted no error, got error: %v", err)
+		}
+	})
+
+	t.Run("when valid MachineConfigPool is present, creates ARO DNS MachineConfig", func(t *testing.T) {
```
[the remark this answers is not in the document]
cadenmarchese: Would this suggestion be better for an integration / e2e test though? It sounds like this behavior would be better tested by testing `Ensure` specifically, not as much by building out the test here.

cadenmarchese (commented):
We covered this PR briefly in code review yesterday - here's a quick summary of discussion we had on first pass. Nice work :)

pkg/operator/controllers/dnsmasq/machineconfig_controller.go
```diff
@@ -71,8 +71,11 @@ func (r *MachineConfigReconciler) Reconcile(ctx context.Context, request ctrl.Re
 		r.log.Error(err)
 		return reconcile.Result{}, err
 	}
+	if mcp.GetDeletionTimestamp() != nil {
```
tsatam, line 74: TODO: remove this as it is no longer necessary (remnant of old implementation via finalizers).

pkg/operator/controllers/dnsmasq/machineconfig_controller_test.go
```diff
@@ -0,0 +1,145 @@
...
+				cluster(true),
+				&mcv1.MachineConfigPool{
+					ObjectMeta: metav1.ObjectMeta{Name: "custom"},
+					Status:     mcv1.MachineConfigPoolStatus{},
+					Spec:       mcv1.MachineConfigPoolSpec{},
+				},
+			).
+			Build()
+		dh := fakeDh(controller)
+		dh.EXPECT().Ensure(gomock.Any(), gomock.Any()).Times(1)
```
bennerv, line 124: Let's leverage similar logic as here: https://github.com/Azure/ARO-RP/pull/2476/files?diff=unified&w=1#diff-71aa82188a09b613595179318fb771896687ffd693acc5187521d9586175b05eR115

```suggestion
		dh.EXPECT().Ensure(gomock.Any(), gomock.AssignableToTypeOf(&mcv1.MachineConfig{})).Times(1)

File: test/e2e/operator.go [hunk 10]
```diff
@@ -463,3 +464,72 @@ var _ = Describe("ARO Operator - ImageConfig Reconciler", func() {
...
+		return names
+	}
+
+	BeforeEach(func(ctx context.Context) {
+		By("Create custom MachineConfigPool")
+		_, err := clients.MachineConfig.MachineconfigurationV1().MachineConfigPools().Create(ctx, &customMcp, metav1.CreateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterEach(func(ctx context.Context) {
```
bennerv, line 503: The e2e test cleans this up already to ensure the owner reference on the MCP is garbage collecting the MC that is created.  i don't think we actually need the `AfterEach` section with anything in it.  

The clusters that run e2e in our PRs and production pipelines are deleted after e2e completes.  This would only be applicable if someone was running multiple e2e tests against a single cluster they created.  Not sure if this is necessary.

pkg/operator/controllers/dnsmasq/cluster_controller.go
bennerv: We don't need this as the `cluster_controller` is only watching the ARO Cluster CR, the `machine_config_controller` is watching the machine configs.

pkg/operator/controllers/dnsmasq/cluster_controller_test.go
```diff
@@ -0,0 +1,129 @@
...
+			Status:     arov1alpha1.ClusterStatus{},
+			Spec: arov1alpha1.ClusterSpec{
+				OperatorFlags: arov1alpha1.OperatorFlags{
+					controllerEnabled: strconv.FormatBool(enabled),
+				},
+			},
+		}
+	}
+
+	t.Run("when no cluster resource is present, returns error", func(t *testing.T) {
```
bennerv, line 39: I wish we could use tabular tests here to remain consistent with the rest of our unit tests.

bennerv (changes requested):
Looks really good - just a few nits.

pkg/operator/controllers/dnsmasq/cluster_controller_test.go
tsatam: I have brought back the tabular tests for these controllers.

test/e2e/operator.go
tsatam: Removed the AfterEach definition.

pkg/operator/controllers/dnsmasq/machineconfig_controller_test.go
tsatam: Done

SudoBrendan (dismissed):
LGTM

s-amann (approved):

SudoBrendan (approved):
LGTM

PR #2477, opened 2022-10-20
WIP - POC for MS/1ES hosted agent pool

cfsmikedrum:
MS requires usage of 1ES hosted agent pools for all pipelines. The ARO-CI agent pool is not compliant, so a new agent pool must be created to replace it. 

ADO user story: https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/15764753

As a POC, the ci.yml file was modified from using the "ARO-CI" agent pool to the "aro-ci-1es" agent pool. 
The pipeline using the ci.yml file will be executed in ADO to test and changes made there. 
Once the pipeline can run on the new agent pool, other pipeline yml files will be updated to use this pool and tested in the same way.

### Test plan for issue:
Run pipelines associated with the modified yml files

tech debt cleanup, N/A


ross-bryan (commented):
Hi Mike, Are you able to answer these question?

- How is this hosted pool provisioned? 

- Does it have FIPS validated go available on it? 

- How are the operating system packages configured and container runtime configured? 

- Whats size disk are we working with? What if we need more? What if it fills up?

PR #2482, opened 2022-10-24
add MUO reconciliation E2E test

AldoFusterTurpin:

Fixes [Implement an E2E test confirming that MUO deployment gets recreated by ARO operator in case of deletion](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15689363)

ARO operator reconciles MUO deployment on customer clusters and must restore MUO resources in case of deletion.

We already have a test that proves that MUO is deployed by default on the customer clusters, but no test that proves that the deployment is being reconciled. So the task is to create that last missing test.

### Test plan for issue:

Created the new E2E test (the task is itself to write an E2E test).

N/A

### DIFF
--- a/test/e2e/operator.go
+++ b/test/e2e/operator.go
@@ -311,9 +311,14 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
 })
 
 var _ = Describe("ARO Operator - MUO Deployment", func() {
+	const (
+		managedUpgradeOperatorNamespace  = "openshift-managed-upgrade-operator"
+		managedUpgradeOperatorDeployment = "managed-upgrade-operator"
+	)
+
 	It("must be deployed by default with FIPS crypto mandated", func(ctx context.Context) {
 		By("getting MUO pods")
-		pods, err := clients.Kubernetes.CoreV1().Pods("openshift-managed-upgrade-operator").List(ctx, metav1.ListOptions{
+		pods, err := clients.Kubernetes.CoreV1().Pods(managedUpgradeOperatorNamespace).List(ctx, metav1.ListOptions{
 			LabelSelector: "name=managed-upgrade-operator",
 		})
 		Expect(err).NotTo(HaveOccurred())
@@ -321,11 +326,38 @@ var _ = Describe("ARO Operator - MUO Deployment", func() {
 
 		By("verifying that MUO has FIPS crypto mandated by reading logs")
 		Eventually(func(g Gomega, ctx context.Context) {
-			b, err := clients.Kubernetes.CoreV1().Pods("openshift-managed-upgrade-operator").GetLogs(pods.Items[0].Name, &corev1.PodLogOptions{}).DoRaw(ctx)
+			b, err := clients.Kubernetes.CoreV1().Pods(managedUpgradeOperatorNamespace).GetLogs(pods.Items[0].Name, &corev1.PodLogOptions{}).DoRaw(ctx)
 			g.Expect(err).NotTo(HaveOccurred())
 
 			g.Expect(string(b)).To(ContainSubstring(`msg="FIPS crypto mandated: true"`))
 		}).WithContext(ctx).Should(Succeed())
+	})
+
+	It("must be restored if deleted", func(ctx context.Context) {
+		deleteMUODeployment := func(ctx context.Context) error {
+			return clients.Kubernetes.
+				AppsV1().
+				Deployments(managedUpgradeOperatorNamespace).
+				Delete(ctx, managedUpgradeOperatorDeployment, metav1.DeleteOptions{})
+		}
+
+		muoDeploymentExists := func(g Gomega, ctx context.Context) {
+			_, err := clients.Kubernetes.
+				AppsV1().
+				Deployments(managedUpgradeOperatorNamespace).
+				Get(ctx, managedUpgradeOperatorDeployment, metav1.GetOptions{})
+
+			g.Expect(err).ToNot(HaveOccurred())
+		}
+
+		By("waiting for the MUO deployment to be ready")
+		Eventually(muoDeploymentExists).WithContext(ctx).Should(Succeed())
+
+		By("deleting the MUO deployment")
+		Expect(deleteMUODeployment(ctx)).Should(Succeed())
+
+		By("waiting for the MUO deployment to be reconciled")
+		Eventually(muoDeploymentExists).WithContext(ctx).Should(Succeed())
 	})
 })
 


facchettos:
moving to draft

facchettos:
draft pr have e2e no ?

test/e2e/operator.go
```diff
@@ -379,33 +379,79 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
+		if err != nil {
+			return false, err
+		}
+
+		By("verifying that MUO has FIPS crypto mandated by reading logs")
+		return strings.Contains(string(b), `msg="FIPS crypto mandated: true"`), nil
+	}
+
+	BeforeEach(func() {
+		ctx = context.Background()
```
rolandmkunkel, line 415: is there any benefit to just setting it to `context.Background()` once?

test/e2e/operator.go
```diff
@@ -379,33 +379,79 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
-	ctx := context.Background()
+	const (
+		managedUpgradeOperatorNamespace = "openshift-managed-upgrade-operator"
+		interval                        = 30 * time.Second
+		duration                        = 10 * time.Minute
+	)
+
+	var ctx context.Context
+
+	muoIsDeployed := func() (bool, error) {
```
rolandmkunkel, line 390: I believe this should be part of beforeEach, and the checks for `mandated: true` can be refactored out into the first test below. Right now you are testing the same case twice in both tests (the second one checking more things afterwards) and also listing the pod twice. This can be done once as a before step. It is actually good practice to even have assertions for correct setups in your beforeEach step.

test/e2e/operator.go
```diff
@@ -379,33 +379,79 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
+			listOptions := metav1.ListOptions{
 				LabelSelector: "name=managed-upgrade-operator",
-			})
-			if err != nil {
-				return false, err
-			}
-			if len(pods.Items) != 1 {
-				return false, fmt.Errorf("%d managed-upgrade-operator pods found", len(pods.Items))
 			}
+			podList, err := clients.Kubernetes.CoreV1().Pods(managedUpgradeOperatorNamespace).List(ctx, listOptions)
```
rolandmkunkel, line 433: should be part of a beforeEach

rolandmkunkel (changes requested):

test/e2e/operator.go
rolandmkunkel: by "also listing the pod twice" I mean duplicating the code

rolandmkunkel:
It should also be checked if deployment itself is recreated. It also uses many more resources related to prometheus and other things, but I am not sure if these are supposed to be reconciled as well. Right now the tests only checks if pods are recreated

test/e2e/operator.go
AldoFusterTurpin: Done.

test/e2e/operator.go
AldoFusterTurpin: Refactored to avoid duplication. Thanks!

test/e2e/operator.go
```diff
@@ -378,34 +385,127 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
-			b, err := clients.Kubernetes.CoreV1().Pods("openshift-managed-upgrade-operator").GetLogs(pods.Items[0].Name, &corev1.PodLogOptions{}).DoRaw(ctx)
-			if err != nil {
-				return false, err
-			}
+func getMUODeployment(ctx context.Context) *appsv1.Deployment {
+	deploymentList, err := clients.Kubernetes.
+		AppsV1().
+		Deployments(managedUpgradeOperatorNamespace).
+		List(ctx, metav1.ListOptions{})
+
```
facchettos, line 408: you should use get here, you know the name of the deployment you want so you might as well just try to retrieve it directly instead of listing and checking the length. This test would break if we are to deploy extra deployments in that namespace.

facchettos (changes requested):

test/e2e/operator.go
AldoFusterTurpin: Yes, you're right. Done. Thanks! 😃

test/e2e/operator.go
```diff
@@ -379,36 +387,125 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
+		By("verifying that MUO has FIPS crypto mandated by reading logs")
+		logs := string(rawLogs)
+		return hasFIPSCryptoMandated(logs), nil
+	}
+
+	return hasMUOFIPSCryptoMandatedFn
+}
+
+// doesMUODeploymentExistHigherOrderFn returns a function which does not return any error as it
```
rolandmkunkel, line 496: couple typos in there: mechanisnm, does not exists, comma missing before the first and

test/e2e/operator.go
```diff
@@ -379,36 +387,125 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
-			return strings.Contains(string(b), `msg="FIPS crypto mandated: true"`), nil
-		}
+			By("waiting for the MUO deployment to be ready")
+			err := wait.PollImmediate(interval, duration, doesMUODeploymentExistFn)
+			Expect(err).NotTo(HaveOccurred())
 
-		err := wait.PollImmediate(30*time.Second, 10*time.Minute, muoIsDeployed)
-		Expect(err).NotTo(HaveOccurred())
+			By("getting the MUO deployment to delete")
+			deploymentToDelete := getMUODeployment(ctx)
```
rolandmkunkel, line 409: do we need to get the deployment here or can we just call delete on the deployment name? We know it already because it uses a conventional name

test/e2e/operator.go
```diff
@@ -379,36 +387,125 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
 
-			By("getting logs from MUO")
-			b, err := clients.Kubernetes.CoreV1().Pods("openshift-managed-upgrade-operator").GetLogs(pods.Items[0].Name, &corev1.PodLogOptions{}).DoRaw(ctx)
-			if err != nil {
-				return false, err
-			}
+	When("MUO is deleted on the ARO cluster", func() {
+		It("should reconcile the MUO deployment", func() {
+			ctx := context.Background()
+			doesMUODeploymentExistFn := doesMUODeploymentExistHigherOrderFn(ctx)
```
rolandmkunkel, line 402: Could we just inline the function below? Not sure if it is easier to read with that name

rolandmkunkel (changes requested):
Almost done I think :)

test/e2e/operator.go
AldoFusterTurpin: Done. Tks!

test/e2e/operator.go
AldoFusterTurpin: You're right, thanks! Removed that.

test/e2e/operator.go
AldoFusterTurpin: Declaring a function inside another function is something I don't really like, neither in production nor in tests, as it pollutes the test's assertions with business logic you are not really interested to see at first sight. It gets even worse when the inlined function is big (> 10 lines).

When I read a test, I want to see something like [arrange, act, assert] in a straightforward way without worrying about how the act part is performed (if I want to see the details I jump to the function to see those details). 

That's why I created the High Order Functions to simplify the content of the tests. Ideally, I would directly define the functions outside the tests, but I need the High Order Functions (as a wrapper) because I need to access an outer variable (ctx) that is not passed in the wait.ConditionFunc type definition but that I need in the function's body.

After your comment, I have simplified the code and know it reads a bit better. 
😃

test/e2e/operator.go
```diff
@@ -378,34 +385,108 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
+		})
+	})
 
-			By("verifying that MUO has FIPS crypto mandated by reading logs")
-			return strings.Contains(string(b), `msg="FIPS crypto mandated: true"`), nil
-		}
+	When("MUO is deleted on the ARO cluster", func() {
+		It("should reconcile the MUO deployment", func() {
+			ctx := context.Background()
+			fn := doesMUODeploymentExistFn(ctx)
```
rolandmkunkel, line 400: NIT: it's a little bit confusing that fn now means something different depending on were you look

rolandmkunkel (dismissed):

test/e2e/operator.go
AldoFusterTurpin: @rolandmkunkel 
I chose _fn_ because the scope of the variable is really narrow (like using _b_ for bytes) and fn just means the **f**u**n**ction that we want to "apply". In a short function like this, I think it is clear. 

But I can revert back to something like  doesMUODeploymentExistFn := doesMUODeploymentExistFn(ctx) if you want.

😃

cadenmarchese: Agreed with Roland, I think we should rename this var.

test/e2e/operator.go
```diff
@@ -378,34 +385,108 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
-			pods, err := clients.Kubernetes.CoreV1().Pods("openshift-managed-upgrade-operator").List(ctx, metav1.ListOptions{
-				LabelSelector: "name=managed-upgrade-operator",
-			})
-			if err != nil {
-				return false, err
-			}
-			if len(pods.Items) != 1 {
-				return false, fmt.Errorf("%d managed-upgrade-operator pods found", len(pods.Items))
-			}
+	When("we have a fresh ARO cluster", func() {
```
cadenmarchese, line 388: Is this `When` needed? It doesn't check that the ARO cluster is fresh, and the same assumption is made in other operator e2e tests. I think you could follow the convention of the other operator e2e tests, and just use the `It`.

test/e2e/operator.go
```diff
@@ -35,6 +35,13 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/subnet"
 )
 
+const (
+	managedUpgradeOperatorNamespace  = "openshift-managed-upgrade-operator"
+	managedUpgradeOperatorDeployment = "managed-upgrade-operator"
+	interval                         = 30 * time.Second
+	duration                         = 10 * time.Minute
+)
```
cadenmarchese, line 43: Can we define these inside of the test?

test/e2e/operator.go
```diff
@@ -378,34 +385,108 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			By("waiting for the MUO deployment to be reconciled")
+			err = wait.PollImmediate(interval, duration, fn)
+			Expect(err).NotTo(HaveOccurred())
+		})
 	})
 })
 
+func getMUOPods(ctx context.Context) (*corev1.PodList, error) {
```
cadenmarchese, line 417: Similar comment to above - the convention in this file is to define necessary functions such as this inside of the test spec rather than outside of it (`ARO Operator - Azure Subnet Reconciler` is a good example of one that has multiple) so I think we should keep that convention here.

cadenmarchese (dismissed):
Looking close, a few suggestions.

test/e2e/operator.go
AldoFusterTurpin: Done.

test/e2e/operator.go
AldoFusterTurpin: Done.

test/e2e/operator.go
AldoFusterTurpin: Done.

test/e2e/operator.go
```diff
@@ -311,22 +311,54 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
 		Eventually(func(g Gomega, ctx context.Context) {
-			b, err := clients.Kubernetes.CoreV1().Pods("openshift-managed-upgrade-operator").GetLogs(pods.Items[0].Name, &corev1.PodLogOptions{}).DoRaw(ctx)
+			b, err := clients.Kubernetes.CoreV1().Pods(managedUpgradeOperatorNamespace).GetLogs(pods.Items[0].Name, &corev1.PodLogOptions{}).DoRaw(ctx)
 			g.Expect(err).NotTo(HaveOccurred())
 
 			g.Expect(string(b)).To(ContainSubstring(`msg="FIPS crypto mandated: true"`))
 		}).WithContext(ctx).Should(Succeed())
 	})
+
+	It("should reconcile the MUO deployment when it is delated", func(ctx context.Context) {
```
m1kola, line 352: You should be able to read it as a sentence: "ARO Operator - MUO Deployment must be restored if delated" vs "ARO Operator - MUO Deployment should reconcile the MUO deployment when it is delated".

Another nit: use "must". "Should" sounds indecisive in this context :) 

```suggestion
	It("must be restored if delated", func(ctx context.Context) {
```

m1kola (commented):
Looks good to me apart few minor things.

The bigest concern is about `Expect(deleteMUODeployment(ctx)).Should(Succeed())` - I'm not sure that this works as expected.

test/e2e/operator.go
```diff
@@ -311,22 +311,54 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
+		By("waiting for the MUO deployment to be ready")
+		Eventually(muoDeploymentExists).WithContext(ctx).Should(Succeed())
```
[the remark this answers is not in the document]
AldoFusterTurpin: Done. Also moving _deleteMUODeployment_ for the same reason. Thanks!

test/e2e/operator.go
AldoFusterTurpin: Done. Thanks!

test/e2e/operator.go
```diff
@@ -311,22 +311,54 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
+
+		By("deleting the MUO deployment")
+		Expect(deleteMUODeployment(ctx)).Should(Succeed())
```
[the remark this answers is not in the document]
AldoFusterTurpin: I think it is correct. The official Gomega docs have a [section](https://onsi.github.io/gomega/#handling-errors) explaining that syntax. 
It is also mentioned in the Ginkgo [docs](https://onsi.github.io/ginkgo/#suite-setup-and-cleanup-beforesuite-and-aftersuite).


image

Thanks!

petrkotas (approved):

PR #2487, opened 2022-10-25
Revert the svc KV access for the AKS agentpool MSI

darthhexx:

Reverts the Key-vault access policy added in #2440 because it is not being dynamically evaluated in production due to [the way that the access policies are allocated](https://github.com/Azure/ARO-RP/blob/8c0842f6fadc18317f6cc61ddf62efc1ee19d1b2/pkg/deploy/assets/rp-production-predeploy.json#L329). We will be reviting this when we don't have a pending RP release on the cards, because we can't allow the RP to remove the permissions everytime it is deployed.

### Test plan for issue:

Tested the inclusion, which failed, so by logical deduction removing it should result in success.

No.

### DIFF
--- a/pkg/deploy/assets/rp-development-predeploy.json
+++ b/pkg/deploy/assets/rp-development-predeploy.json
@@ -292,18 +292,6 @@
                     },
                     {
                         "tenantId": "[subscription().tenantId]",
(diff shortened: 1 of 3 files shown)


UlrichSchlueter:
lgtm

facchettos (approved):

petrkotas (approved):

Issue #1593, opened 2021-07-08
Add unit test coverage for service principal validation (dynamic validator and operator)


m1kola:
We can solve the following todo items:

https://github.com/Azure/ARO-RP/blob/57e4bf000d38df650b29247319f5796e0280ed32/pkg/api/validate/dynamic/serviceprincipal.go#L18

https://github.com/Azure/ARO-RP/blob/57e4bf000d38df650b29247319f5796e0280ed32/pkg/operator/controllers/checker/serviceprincipalchecker_test.go#L19

by mocking `aad.GetToken` in a similar we mock `time.Now()` from go standard library.

Example code:

https://github.com/Azure/ARO-RP/blob/57e4bf000d38df650b29247319f5796e0280ed32/pkg/util/clientauthorizer/arm_test.go#L124

Example test:

https://github.com/Azure/ARO-RP/blob/57e4bf000d38df650b29247319f5796e0280ed32/pkg/util/clientauthorizer/arm.go#L55

kitarp29:
Hi @m1kola 
I have experience with Golang, but not much in writing Unit tests.
This issue might be a good starting point for me.
I would like to take it up.
Could you assist me with it?

m1kola:
@kitarp29 unfortunately I'm not sure whether I can provide enough assistance in learning unit testing to external contributors at the moment.

But I'll try to give you some pointers. I recommend checking out [this tutorial](https://go.dev/doc/tutorial/add-a-test) on unit testing from Go website (it is a part of a bigger tutorial). Also take a look at [documentation for `testing` package](https://pkg.go.dev/testing) from standard library. Apart from very good overview of the `testing` package it also has some [examples](https://pkg.go.dev/testing#hdr-Examples). Another commonly used package for unit testing is [gomock](https://github.com/golang/mock). We use it in our codebase for generating mocks from interfaces and a lot of other go projects use this package.

Apart from go specific things it might be a good idea to look something up on different types of tests. And find an answer to the question "what is a good unit tests?". Unit tests are often unintentionally turned into tests which test multiple units. Think about the reasons why it is not good.

Once you are more comfortable with the basics - take a look at the real world examples (in our codebase, in upstream Kubernetes or in some other go project) and try to write more complex tests yourself.

AldoFusterTurpin:
[...]

PR #2488, opened 2022-10-25
Add unit tests for dynamic/serviceprincipal, azureclient/environments, clusterauthorizer

s-fairchild:
https://github.com/Azure/ARO-RP/issues/1593

Fixes
Cleanup and separate test cases in `serviceprincipalchecker_test.go`

ServicePrincipalChecker now accepts a function with the signature of aad.GetToken, allowing the use of the mocked GetToken function.
Added tests to azureclient/environments and clusterauthorizer that were previously in serviceprincipalchecker.
Add mock of dynamic service principal validator to prevent testing code outside of the service principal checker package.
serviceprincipal is tested by manually creating and modifying the service principal tokens.

### Test plan for issue:

New unit tests are added to clusterauthorizer and serviceprincipalchecker packages.
run `make test-go`

No, only unit tests are added.


hawkowl:
e2e is failing with a null pointer dereference, can you check that? @s-fairchild

s-fairchild:
@hawkowl Yes I'm looking into the e2e tests

pkg/api/validate/dynamic/serviceprincipal_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/Azure/go-autorest/autorest/adal"
+	"github.com/golang/mock/gomock"
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+	mock_aad "github.com/Azure/ARO-RP/pkg/util/mocks/aad"
+)
+
+func TestValidateServicePrincipal(t *testing.T) {
```
cadenmarchese, line 23: It might also be possible to test for a failure to get the token, and a failure to parse the token (lines 21 and 28 of serviceprincipal.go)

pkg/operator/controllers/checker/serviceprincipalchecker_test.go
```diff
@@ -61,18 +105,45 @@ func TestServicePrincipalValid(t *testing.T) {
...
-			kubernetescli: kubernetescli,
+			log:                      log,
+			arocli:                   arocli,
+			kubernetescli:            kubernetescli,
+			getToken:                 aad.GetToken,
+			validateServicePrincipal: dynamic,
 		}
 
 		t.Run(tt.name, func(t *testing.T) {
 			err := sp.Check(ctx)
```
cadenmarchese, line 142: Could we also add either direct or indirect coverage for `updateFailedCondition()` in the service principal checker?

cadenmarchese (commented):
This is looking good, I ran some coverage reports and added some additional ideas. In case it helps, this is what I'm using for coverage:
```
cover () {
        t="/tmp/go-cover.$$.tmp"
        go test -coverprofile=$t $@ && go tool cover -html=$t && unlink $t
}
```
Thanks for your work on this!

pkg/util/clusterauthorizer/authorizer_test.go
```diff
@@ -0,0 +1,154 @@
...
+	log := logrus.NewEntry(logrus.StandardLogger())
+
+	for _, tt := range []struct {
+		name       string
+		secret     *corev1.Secret
+		azCloudEnv *azureclient.AROEnvironment
+		wantErr    string
+	}{
+		{
+			name:       "fail: Invalid principal credentials",
```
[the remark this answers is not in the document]
s-fairchild: Certainly

pkg/api/validate/dynamic/serviceprincipal_test.go
s-fairchild: That should be possible by creating multiple expects and returning those errors. I didn't create those initially because it's a mocked return, so it should hit the `if err != nil { return err }` line.
I can add some tests to cause the parser to fail as well.

s-fairchild:
@cadenmarchese Let me know what you think of my last commit, I was able to increase the coverage for the areas you pointed out. Thanks for sharing that cover function it's really helpful!

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -69,6 +72,7 @@ type dynamic struct {
 	resourceSkusClient compute.ResourceSkusClient
 	spComputeUsage     compute.UsageClient
 	spNetworkUsage     network.UsageClient
+	getToken           func(ctx context.Context, log *logrus.Entry, clientID, clientSecret, tenantID string, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error)
 }
 
 type AuthorizerType string
```
facchettos, line 78: What is the purpose of this ? It should be an interface (getToken) not a function.

facchettos (changes requested):

pkg/api/validate/dynamic/dynamic.go
s-fairchild: Thanks for pointing this out. I've update the packages to use interfaces rather than a function.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -43,7 +42,7 @@ type Subnet struct {
 
 type ServicePrincipalValidator interface {
 	ValidateServicePrincipal(ctx context.Context, clientID, clientSecret, tenantID string) error
-	SetGetToken(gt func(ctx context.Context, log *logrus.Entry, clientID, clientSecret, tenantID string, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error))
+	SetGetToken(gt aad.TokenClient)
```
facchettos, line 45: Why do we need this? Can't we just create the objects with that field?

pkg/api/validate/dynamic/serviceprincipal_test.go
```diff
@@ -0,0 +1,193 @@
...
+		{
+			name:      "pass: Successful Validation",
+			tokenReqs: newTokenRequirements(),
+			aadMock: func(aad *mock_aad.MockTokenClient, tr *tokenRequirements) {
+				token, err := createToken(tr)
+				if err != nil {
+					t.Errorf("failed to create testing service principal token: %v\n", err)
+				}
+				aad.EXPECT().GetToken(ctx, log, tr.clientID, tr.clientSecret, tr.tenantID, tr.aadEndpoint, tr.graphEndpoint).MaxTimes(1).Return(token, nil)
+			},
```
facchettos, line 54: There is lots of code duplication here, I think it would be better to have that code in the test section and only feed it the arguments

facchettos (changes requested):

pkg/util/clusterauthorizer/authorizer_test.go
```diff
@@ -0,0 +1,344 @@
...
+			azCloudEnv: &azureclient.PublicCloud,
+			secret:     newV1CoreSecret(azureSecretName, nameSpace),
+			tokenReqs:  newTokenRequirements(),
+			aadMock: func(aad *mock_aad.MockTokenClient, tr *tokenRequirements) {
+				aad.EXPECT().GetToken(ctx, log, tr.clientID, tr.clientSecret, tr.tenantID, tr.aadEndpoint, tr.resource).MaxTimes(1).Return(
+					nil, fmt.Errorf("400: InvalidServicePrincipalCredentials: properties.servicePrincipalProfile: The provided service principal credentials are invalid."),
+				)
+			},
```
facchettos, line 102: same comment than above

pkg/api/validate/dynamic/dynamic.go
s-fairchild: The dynamic struct and fields are private, so when `serviceprincipal_test.go` runs on line 125 this is used to replace the `GetToken` with the mock.
The fields aren't modifiable from the `psDynamic` object.

Is there a way to directly assign the mock functions without this?

facchettos: Add an argument to `NewValidator` and `NewServicePrincipalValidator`. Then you'll be able to create them directly with your mock tokenClient in it

pkg/api/validate/dynamic/serviceprincipal_test.go
s-fairchild: Thanks Jeremy. I originally did that to change the error returned. Now it seems obvious to create an another error field in the test struct. I moved the mock code in my last commit.

pkg/api/validate/dynamic/dynamic.go
s-fairchild: I've updated them to accept the token client as an argument.

facchettos:
looks good :+1: can you squash your commits or do a fixup to have only a single one and then let's get this in

pkg/operator/controllers/checker/serviceprincipalchecker.go
```diff
@@ -69,29 +72,35 @@ func (r *ServicePrincipalChecker) Check(ctx context.Context) error {
...
 
-func updateFailedCondition(cond *operatorv1.OperatorCondition, err error) {
+func updateFailedCondition(cond *operatorv1.OperatorCondition, err error) *operatorv1.OperatorCondition {
 	cond.Status = operatorv1.ConditionFalse
 	if tErr, ok := err.(*api.CloudError); ok {
 		cond.Message = tErr.Message
 	} else {
 		cond.Message = err.Error()
 	}
+	return cond
```
facchettos, line 105: Actually this is redundant. we are already working with a pointer so the change will be reflected onto the struct. 
```suggestion
```

pkg/operator/controllers/checker/serviceprincipalchecker_test.go
```diff
@@ -61,18 +108,109 @@ func TestServicePrincipalValid(t *testing.T) {
...
+			messageErr: fmt.Errorf("service principal is invalid"),
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			var cond *operatorv1.OperatorCondition
+			if tt.cloudErr != nil {
+				cond = updateFailedCondition(tt.cond, tt.cloudErr)
+			} else if tt.messageErr != nil {
+				cond = updateFailedCondition(tt.cond, tt.messageErr)
+			}
```
facchettos, line 208: the assignment of cond here isn't necessary.

facchettos (changes requested):
I was a little quick on the trigger! I have one extra change that I want before we get it in and then :rocket:

pkg/operator/controllers/checker/serviceprincipalchecker.go
s-fairchild: Thanks for pointing that out. I've removed this and squashed my commits.

pkg/util/aad/aad.go
```diff
@@ -18,9 +18,21 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/refreshable"
 )
 
+type TokenClient interface {
+	GetToken(ctx context.Context, log *logrus.Entry, clientID, clientSecret, tenantID string, aadEndpoint, resource string) (*adal.ServicePrincipalToken, error)
+}
+
+type tokenClient struct {
+	TokenClient
+}
```
facchettos, line 27: This isn't needed, as go uses duck typing for interfaces. (line 26)

facchettos (changes requested):
aaaand, I missed another small thing... :cry:

pkg/util/aad/aad.go
s-fairchild: Yes that was my intent, I've removed that field.

facchettos (approved):

cadenmarchese (approved):
LGTM - thanks @facchettos for your review

PR #2490, opened 2022-10-26
Revert "Temporarily force the rp nsg to redeploy"

ventifus:
Now that we have deployed the NSG update, we no longer need to force create it. Revert to previous behavior.

### DIFF
--- a/pkg/deploy/predeploy.go
+++ b/pkg/deploy/predeploy.go
@@ -126,11 +126,7 @@ func (d *deployer) PreDeploy(ctx context.Context) error {
 	// key the decision to deploy NSGs on the existence of the gateway
 	// predeploy.  We do this in order to refresh the RP NSGs when the gateway
 	// is deployed for the first time.
-
-	// TODO: *** REVERT AFTER RELEASE ***
-	// *** Forcing this to true here to ensure the rp_in_geneva NSG gets updated with GenevaActions service tag.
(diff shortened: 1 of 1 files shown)


pkg/deploy/predeploy.go
```diff
@@ -126,11 +126,7 @@ func (d *deployer) PreDeploy(ctx context.Context) error {
 	// key the decision to deploy NSGs on the existence of the gateway
 	// predeploy.  We do this in order to refresh the RP NSGs when the gateway
 	// is deployed for the first time.
-
-	// TODO: *** REVERT AFTER RELEASE ***
-	// *** Forcing this to true here to ensure the rp_in_geneva NSG gets updated with GenevaActions service tag.
-	// *** Once the NSGs get updated, revert this change (set the isCreate declaration back to `var isCreate bool`) so that the NSGs aren't getting redeployed every release.
-	var isCreate = true
+	var isCreate bool
```
facchettos, line 129:
```suggestion
	isCreate := false
```

facchettos (changes requested):
reads better this way

pkg/deploy/predeploy.go
ventifus: Sure that works

facchettos (approved):

cadenmarchese (approved):

PR #2493, opened 2022-10-27
validate permissions on NAT gateways

rhamitarora:

We should update our azure-cli command module to include adding the network contributor on the NAT gateway (if one exists). Right now we're only adding the vnet and route tables.

This PR is related to server side changes for Nat Gateway validation.
Client Side Implementation: [https://github.com/Azure/ARO-RP/pull/2394](https://github.com/Azure/ARO-RP/pull/2394)

### Test plan for issue:
Below two test cases added:
_TestValidateNatGatewaysPermissions_
_TestGetNatGatewayID_

### DIFF
--- a/pkg/api/validate/dynamic/dynamic.go
+++ b/pkg/api/validate/dynamic/dynamic.go
@@ -150,6 +150,12 @@ func (dv *dynamic) ValidateVnet(ctx context.Context, location string, subnets []
 		}
 	}
 
+	for _, s := range subnets {
+		err := dv.validateNatGatewayPermissions(ctx, s)
+		if err != nil {
+			return err
+		}
+	}
 	return dv.validateCIDRRanges(ctx, subnets, additionalCIDRs...)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/api/validate/dynamic/dynamic.go
```diff
@@ -229,6 +237,55 @@ func (dv *dynamic) validateRouteTablePermissions(ctx context.Context, s Subnet)
...
+	if dv.authorizerType == AuthorizerClusterServicePrincipal {
+		errCode = api.CloudErrorCodeInvalidServicePrincipalPermissions
+	}
+
+	err = dv.validateActions(ctx, &ngr, []string{
+		"Microsoft.Network/natGateways/join/action",
+		"Microsoft.Network/natGateways/read",
+		"Microsoft.Network/natGateways/write",
+	})
+	if err == wait.ErrWaitTimeout {
```
s-amann, line 279: Can `wait.ErrWaitTimeout` happen for other issues other than lack of permissions? Usually a 403 is returned when the permissions don't line up and a timeout means we could not connect or the serving application did not respond in time

s-amann (dismissed):
overall lgtm, just one clean up and one question

s-amann (dismissed):

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -229,6 +235,55 @@ func (dv *dynamic) validateRouteTablePermissions(ctx context.Context, s Subnet)
...
+
+	vnet, err := dv.virtualNetworks.Get(ctx, vnetr.ResourceGroup, vnetr.ResourceName, "")
+	if err != nil {
+		return err
+	}
+
+	ngID, err := getNatGatewayID(&vnet, s.ID)
+	if err != nil || ngID == "" { // error or no nat gateway
+		return err
+	}
```
facchettos, line 260: what does that mean if `ngID == ""` ?

rhamitarora: The data type of ngID is string. In this statement, we need to check if there is no Nat Gateway associated to the Subnet.

facchettos: but does that mean that this is a valid case? 
I would rather have a separate `if` if that is the case. 
 ```suggestion
	ngID, err := getNatGatewayID(&vnet, s.ID)
	if err != nil {
		return err
	}
	if ngID == "" { // no nat gateway so valid
		return nil
	}
```

rhamitarora: Suggested changes implemented.

pkg/api/validate/dynamic/dynamic.go
hawkowl: This is a userspace wait timeout, which means that the loop inside validateActions (which reruns it because sometimes it takes a few goes if it's just been applied) didn't pass with success -- if the other end times out it'll be a HTTP error below.

hawkowl (approved):

PR #2494, opened 2022-10-27
Selenium fix

facchettos:
followup to #2492 
This should probably be reverted once we figured out a solution to docker in int

Fixes


UlrichSchlueter:
LGTM

petrkotas (approved):

PR #2495, opened 2022-10-27
Selenium fix

facchettos:
follow up of #2494 

fixing code smells and disabling tests if docker can't be ran

Fixes

### DIFF
--- a/test/e2e/setup.go
+++ b/test/e2e/setup.go
@@ -75,11 +75,21 @@ var (
 	vnetResourceGroup string
 	clusterName       string
 	clients           *clientSet
+
+	dockerSucceeded bool
 )
 
 func skipIfNotInDevelopmentEnv() {
 	if !_env.IsLocalDevelopmentMode() {
-		Skip("skipping tests in non-development environment")
+		Skip("skipping portal tests in non-development environment")
+	}
+}
+
+func skipIfDockerNotWorking() {
+	// docker cmds will fail in INT until we figure out a solution since
+	// it is running from docker already
+	if !dockerSucceeded {
+		Skip("skipping portal tests as docker is not available")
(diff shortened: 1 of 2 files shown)


SudoBrendan (changes requested):
So very close, would like to see separate error messages for localDevMode vs FailedDocker

test/e2e/setup.go
```diff
@@ -75,15 +75,19 @@ var (
...
 	clients           *clientSet
+
+	dockerSucceeded bool
 )
 
 func skipIfNotInDevelopmentEnv() {
-	if !_env.IsLocalDevelopmentMode() {
+	// docker cmds will fail in INT until we figure out a solution since
+	// it is running from docker already
+	if !_env.IsLocalDevelopmentMode() || !dockerSucceeded {
```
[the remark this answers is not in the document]
facchettos: I added a new function to skip only the portal tests when docker does not work

test/e2e/setup.go
```diff
@@ -75,15 +75,25 @@ var (
 	vnetResourceGroup string
 	clusterName       string
 	clients           *clientSet
+
+	dockerSucceeded bool
 )
 
 func skipIfNotInDevelopmentEnv() {
+	// docker cmds will fail in INT until we figure out a solution since
```
cadenmarchese, line 83: We can move this comment into the separate function.

cadenmarchese (changes requested):

test/e2e/setup.go
facchettos: yes I messed up the refactoring haha, should be fine now

karanmagdani (approved):

petrkotas (approved):

PR #2496, opened 2022-10-28
Split the image mirroring up so that it can run piecemeal

hawkowl:

Part of https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/16033731

Breaks apart the image mirroring a little, so that each part can be run alone, and the OpenShift portion can be broken up to individual tasks for each minor version.

### Test plan for issue:

A bit of minor local testing, although this shouldn't break the existing pipelines

Not yet

### DIFF
--- a/cmd/aro/main.go
+++ b/cmd/aro/main.go
@@ -24,7 +24,20 @@ func usage() {
 	fmt.Fprintf(flag.CommandLine.Output(), "  %s dbtoken\n", os.Args[0])
 	fmt.Fprintf(flag.CommandLine.Output(), "  %s deploy config.yaml location\n", os.Args[0])
 	fmt.Fprintf(flag.CommandLine.Output(), "  %s gateway\n", os.Args[0])
-	fmt.Fprintf(flag.CommandLine.Output(), "  %s mirror [release_image...]\n", os.Args[0])
+	fmt.Fprintf(flag.CommandLine.Output(), "  %s mirror\n", os.Args[0])
+
+	// indent the mirror flags... probably a better way to do this? :)
+	s := &strings.Builder{}
+	mirror := mirrorFlags()
+	mirror.SetOutput(s)
+	mirror.PrintDefaults()
+	for _, l := range strings.Split(s.String(), "\n") {
+		if l == "" {
+			continue
+		}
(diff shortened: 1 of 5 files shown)


cmd/aro/mirror.go
```diff
@@ -84,15 +118,104 @@ func mirror(ctx context.Context, log *logrus.Entry) error {
...
+	doBaseMirror := func(ref string) error {
+		log.Printf("mirroring %s -> %s", ref, pkgmirror.Dest(dstAcr+acrDomainSuffix, ref))
+
+		srcAuth := srcAuthRedhat
+		if strings.Index(ref, "quay.io") == 0 {
+			srcAuth = srcAuthQuay
+		}
+
+		return pkgmirror.Copy(ctx, pkgmirror.Dest(dstAcr+acrDomainSuffix, ref), ref, dstAuth, srcAuth)
+	}
```
facchettos, line 165: it probably would be better to have a proper function definition for that

facchettos (changes requested):
some suggestions + a question

cmd/aro/main.go
```diff
@@ -24,7 +24,20 @@ func usage() {
...
+	s := &strings.Builder{}
+	mirror := mirrorFlags()
+	mirror.SetOutput(s)
+	mirror.PrintDefaults()
+	for _, l := range strings.Split(s.String(), "\n") {
+		if l == "" {
+			continue
+		}
+		fmt.Fprintf(flag.CommandLine.Output(), "  %s\n", l)
+	}
```
facchettos, line 39: maybe we can look into tabwriter ? https://pkg.go.dev/text/tabwriter

pkg/util/version/version.go
```diff
@@ -12,6 +12,7 @@ import (
 )
 
 var rxVersion = regexp.MustCompile(`^(\d+)\.(\d+)\.(\d+)(.*)`)
+var rxVersionMinorOnly = regexp.MustCompile(`^(\d+)\.(\d+)(.*)`)
 
 type Version struct {
 	V      [3]uint32
```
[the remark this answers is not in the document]
hawkowl: No clue :)

hlipsig:
Closing old PR's. Please open again if needed.

PR #2497, opened 2022-10-28
Fix permissions error for image mirroring

hawkowl:
fixes image mirroring failing

RHEL8 hardcodes a root-writable place for lookaside signature storage, which our rootless image copying can't use and fails with. This makes it look at a non-existent directory so that it falls back to the user-dir-based default.

### Test plan for issue:

Testing the pipeline, probably?

N/A


hawkowl:
Might not be needed if the EV2 mirroring succeeds :)

hawkowl:
Likely not needed because EV2

PR #2501, opened 2022-10-30
Enable copy.Image's optimisation for cases where the destination image already exists

hawkowl:

Minor improvement I saw while trying to fix the mirroring in EV2.

Since our mirroring will be replicating images that already exist nearly all of the time, it appears like this optimisation will speed up the usual case where the image already exists.

### Test plan for issue:

Don't have a good one, but we could merge it and run it :)

N/A


darthhexx (approved):
Works very well.

PR #2504, opened 2022-10-31
Move sshd_config step first in boot script, add basic logging

cadenmarchese:

When the scripts fail to execute fully on prod deploy, we can't get in to debug them since the SSH auth config step is at the end of the script. Additionally, if any individual step is failing, we don't have visibility on which one.

- Moves ssh step of script to the very beginning
- Adds some basic logging so we know what step (if any) is being executed during a failure of one step


cblecker (approved):

pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -1,12 +1,19 @@
+# We need to manually set PasswordAuthentication to true in order for the VMSS Access JIT to work
+echo "setting ssh password authentication"
+sed -i 's/PasswordAuthentication no/PasswordAuthentication yes/g' /etc/ssh/sshd_config
+
```
darthhexx, line 4: We need to reload the SSH config or this won't take effect until the reboot.

pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -1,12 +1,19 @@
+# We need to manually set PasswordAuthentication to true in order for the VMSS Access JIT to work
+echo "setting ssh password authentication"
+sed -i 's/PasswordAuthentication no/PasswordAuthentication yes/g' /etc/ssh/sshd_config
+
```
darthhexx, line 4: Here as well.

darthhexx (changes requested):

pkg/deploy/generator/scripts/gatewayVMSS.sh
cadenmarchese: Indeed, thanks!

darthhexx (dismissed):
lgtm

pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -1,12 +1,20 @@
+# We need to manually set PasswordAuthentication to true in order for the VMSS Access JIT to work
+echo "setting ssh password authentication"
+sed -i 's/PasswordAuthentication no/PasswordAuthentication yes/g' /etc/ssh/sshd_config
+systemctl restart sshd.service
```
s-amann, line 4: Do we need to restart and could we get away with `reload` instead?

s-amann (approved):

pkg/deploy/generator/scripts/gatewayVMSS.sh
cadenmarchese: We can move to `reload` if it's preferred, for our case either should work (I just had more confidence in the restart)

s-amann (dismissed):

s-amann (approved):

darthhexx (approved):

PR #2505, opened 2022-10-31
Improves context usage in E2Es

m1kola:

Unifies context usage across e2e specs: makes our tests use Ginkgo provided context. This should help with troubleshooting as [we supposed get better picture](https://onsi.github.io/ginkgo/#spec-timeouts-and-interruptible-nodes) in case of failures (such as timeouts).

### Test plan for issue:

Run E2Es

No, just E2E refactoring.


test/e2e/cluster.go
```diff
@@ -29,32 +29,30 @@ const (
...
+	var _ = AfterEach(func(ctx context.Context) {
 		By("deleting a test namespace")
-		err := p.Delete(context.Background())
+		err := p.Delete(ctx)
 		Expect(err).NotTo(HaveOccurred(), "Failed to delete test namespace")
 
 		By("verifying the namespace is deleted")
-		Eventually(func() error {
-			return p.VerifyProjectIsDeleted(context.Background())
-		}, 5*time.Minute, 10*time.Second).Should(BeNil())
```
m1kola, line 52: Here and in other places polling interval and timeout match default values we set [here](https://github.com/Azure/ARO-RP/blob/6191c11a035462cba7c4bc53b0f9fe18e8868a59/test/e2e/setup.go#L411-L412) so I removed them. Probably makes sense to set only custom values.

PR #2507, opened 2022-10-31
Flake hunting - Improves E2E clean up for internet checker

m1kola:

Previously in `AfterEach` we were restoring original URLs for the internet checker, but we were not waiting for original conditions to be set before proceeding to the next test.

This led to a flaky `It` containers where we were asserting conditions.

### Test plan for issue:

Run E2E locally with repeats:
```bash
ginkgo run --tags e2e,codec.safe --timeout 180m -v --until-it-fails --focus "ARO Operator - Internet checking"  ./test/e2e/...
```

Note `--until-it-fails` flag.

No, just E2E improvements.

### DIFF
--- a/test/e2e/operator.go
+++ b/test/e2e/operator.go
@@ -98,17 +98,15 @@ var _ = Describe("ARO Operator - Internet checking", func() {
 			return err
 		})
 		Expect(err).NotTo(HaveOccurred())
-	})
-	It("sets InternetReachableFromMaster to true when the default URL is reachable from master nodes", func(ctx context.Context) {
-		co, err := clients.AROClusters.AroV1alpha1().Clusters().Get(ctx, "cluster", metav1.GetOptions{})
-		Expect(err).NotTo(HaveOccurred())
-		Expect(conditions.IsTrue(co.Status.Conditions, arov1alpha1.InternetReachableFromMaster)).To(BeTrue())
-	})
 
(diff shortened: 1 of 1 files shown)


s-amann (approved):

test/e2e/operator.go
```diff
@@ -98,17 +98,15 @@ var _ = Describe("ARO Operator - Internet checking", func() {
 			return err
 		})
 		Expect(err).NotTo(HaveOccurred())
-	})
-	It("sets InternetReachableFromMaster to true when the default URL is reachable from master nodes", func(ctx context.Context) {
```
cblecker, line 102: Is it intentional that we are removing these tests?

s-amann: The two tests are replaced by these two lines
```g.Expect(conditions.IsTrue(co.Status.Conditions, arov1alpha1.InternetReachableFromMaster)).To(BeTrue())
			g.Expect(conditions.IsTrue(co.Status.Conditions, arov1alpha1.InternetReachableFromWorker)).To(BeTrue())

cadenmarchese (approved):
LGTM

test/e2e/operator.go
cblecker: Okay so as opposed to a separate test, we are going to wait for these conditions to be true after each test. I understand.

cblecker (approved):

PR #2508, opened 2022-10-31
add param names to Emitter's interface methods and unify naming

AldoFusterTurpin:

Improve naming: add parameter names to Emitter's interface methods and unify the naming of metrics and monitor.

In the codebase, we refer to the same concepts regarding metrics with different names in different places ("metric", "topic", etc.). This PR unifies the usage of those names following the [statsd Protocol Reference of MSFT](https://eng.ms/docs/products/geneva/collect/metrics/statsdref). It also adds names to the parameters of the methods of metrics.Emitter interface as those methods has two unnamed parameters of the same type, which makes it impossible to know which param is what without looking at the interface implementation.

### Test plan for issue:

Already covered by unit tests/E2E test. Most of the refactoring was done using the IDE renaming feature.

N/A

### DIFF
--- a/pkg/metrics/statsd/metrics.go
+++ b/pkg/metrics/statsd/metrics.go
@@ -12,11 +12,11 @@ import (
 
 // metric represents generic metric structure
 type metric struct {
-	metric    string
-	account   string
-	namespace string
-	dims      map[string]string
-	ts        time.Time
+	name       string
+	account    string
+	namespace  string
+	dimensions map[string]string
+	timestamp  time.Time
 
 	valueGauge *int64
 	valueFloat *float64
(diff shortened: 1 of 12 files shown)


pkg/metrics/statsd/metrics.go
```diff
@@ -12,11 +12,11 @@ import (
 
 // metric represents generic metric structure
 type metric struct {
-	metric    string
```
AldoFusterTurpin, line 15: IMHO having an attribute with the same name as the struct to which it belongs is a bit weird.

petrkotas (approved):

pkg/util/steps/runner_test.go
```diff
@@ -379,8 +379,8 @@ func TestStepMetricsTopicNaming(t *testing.T) {
 		},
 	} {
 		t.Run(tt.desc, func(t *testing.T) {
-			if got := tt.step.metricsTopic(); got != tt.want {
-				t.Errorf("incorrect step metrics topic, want: %s, got: %s", tt.want, got)
+			if got := tt.step.metricsName(); got != tt.want {
+				t.Errorf("incorrect step metrics name, want: %s, got: %s", tt.want, got)
```
s-amann, line 383: can we rename this test from `TestStepMetricsTopicNaming` -> `TestStepMetricsNameFormatting` or similar?

s-amann (changes requested):
minor nit but leaves us without further changes. Thank you!

pkg/util/steps/runner_test.go
AldoFusterTurpin: Yes! Aligned with the refactoring. Done! 
Thanks!

s-amann (approved):

petrkotas (approved):
I like how it simplified. LGTM

facchettos (approved):

PR #2512, opened 2022-11-02
Exclude WALinuxAgent from vm update scripts

s-amann:

[WI](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/16113483)

Fixes hung up WALinuxAgent due to being updated

Updating the WALinuxAgent will stop the agent causing the script to not complete.

### Test plan for issue:

N/A

N/A


pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -1,4 +1,4 @@
-yum -y update
+yum -y -x WALinuxAgent update
```
cadenmarchese, line 1: There's a yum update later in the script as well (after repositories are added), should we exclude there as well?

cadenmarchese (commented):
Looking good, one question:

pkg/deploy/generator/scripts/rpVMSS.sh
s-amann: I only see two yum installs in the remainder of the files

jharrington22: I don't see the other update there is only an install of a specific rpm [here](https://github.com/Azure/ARO-RP/pull/2512/files#diff-a7a118c3717408a3425f9209b28c891edad1c68d9b9a67cb41b18545f8ce37e3R16) and then a list of specific packages on [here](https://github.com/Azure/ARO-RP/pull/2512/files#diff-a7a118c3717408a3425f9209b28c891edad1c68d9b9a67cb41b18545f8ce37e3R74) am I missing it?

cadenmarchese: You're right, they are not updates. LGTM!

jharrington22 (approved):
/lgtm

s-amann:
> 

I believe the idea behind us excluding the WALinuxAgent from our update is Azure should be taking care of which version is to be used by a new VMSS. I have reached out to MSFT to get clarification on this.

jhoreman:
We should consider addressing the other potential race condition of the reboot command as well. Proposing we increase the sleep count on the last line to 120 (up from 30)

Line #625: (sleep 120; reboot) &

s-amann:
> We should consider addressing the other potential race condition of the reboot command as well. Proposing we increase the sleep count on the last line to 120 (up from 30)
> 
> Line #625: (sleep 120; reboot) &

done

jhoreman (approved):
LGTM thanks

jhoreman (approved):
lgtm

cadenmarchese (approved):

PR #2516, opened 2022-11-03
Ensure SSH access and LB backend pool for Master Node replacement

mhsreddy:

Fixes: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14539058/
Fixes: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/16195191/

This PR will fix SSH to the new node did not work after the master node replacement  and a PUCM was performed to reconcile the NIC and restore ssh access (as PUCM will call fixssh to do that reconciliation). Also this PR  fixes the issue whether the new NIC  is not being added to Internal LoadBalancer API Backend Address Pool after the master is replaced even after running the Admin Update or PUCM.

### Test plan for issue:

Added unit tests. Also tested manually by replacing the master nodes in test clusters.

<!--

Added new Go Mock Testcases.

https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/253699/Replace-a-Master-Node-VM

### DIFF
--- a/pkg/cluster/fixssh.go
+++ b/pkg/cluster/fixssh.go
@@ -79,7 +79,7 @@ func (m *manager) checkandUpdateNIC(ctx context.Context, resourceGroup string, i
 
 		nic, err = m.interfaces.Get(ctx, resourceGroup, nicName, "")
 		if err != nil {
-			//iErr = err // preserve original error
+			m.log.Warnf("Fetching details for NIC %s has failed with err %s", nicName, err)
 			fallbackNIC = true
 		} else if nic.InterfacePropertiesFormat != nil && nic.InterfacePropertiesFormat.VirtualMachine == nil {
 			err = m.removeBackendPoolsFromNIC(ctx, resourceGroup, nicName, &nic)
@@ -101,7 +101,7 @@ func (m *manager) checkandUpdateNIC(ctx context.Context, resourceGroup string, i
 			}
 		}
 
-		changed := updateNIC(&nic, &lb, i)
+		changed := updateNIC(&nic, &lb, i, infraID)
 
 		if changed {
 			m.log.Printf("updating %s", nicName)
@@ -124,7 +124,7 @@ func (m *manager) removeBackendPoolsFromNIC(ctx context.Context, resourceGroup, 
 	return nil
 }
 
-func updateNIC(nic *mgmtnetwork.Interface, lb *mgmtnetwork.LoadBalancer, i int) bool {
+func updateNIC(nic *mgmtnetwork.Interface, lb *mgmtnetwork.LoadBalancer, i int, infraID string) bool {
 	id := fmt.Sprintf("%s/backendAddressPools/ssh-%d", *lb.ID, i)
 	ipc := (*nic.InterfacePropertiesFormat.IPConfigurations)[0]
 	if ipc.LoadBalancerBackendAddressPools == nil {
--- a/pkg/cluster/fixssh.go
+++ b/pkg/cluster/fixssh.go
@@ -48,7 +48,7 @@ func (m *manager) fixSSH(ctx context.Context) error {
 		}
 	}
 
-	for i := 0; i < 3; i++ {
+	for i := 2; i < 3; i++ {
 		// NIC names might be different if customer re-created master nodes
 		// see https://bugzilla.redhat.com/show_bug.cgi?id=1882490 for more details
 		// installer naming  - <foo>-master{0,1,2}-nic
@@ -70,6 +70,16 @@ func (m *manager) fixSSH(ctx context.Context) error {
 				m.log.Warnf("fallback failed with err %s", err)
 				return iErr
 			}
+
+		} else if nic.InterfacePropertiesFormat != nil && nic.InterfacePropertiesFormat.VirtualMachine == nil {
+			nicName = nicNameMachineAPI
+			m.log.Warnf("fallback to check MachineAPI NIC name format for %s", nicName)
+			nic, err = m.interfaces.Get(ctx, resourceGroup, nicName, "")
+			if err != nil {
+				m.log.Warnf("fallback failed with err %s", err)
+				return err
+			}
+
 		}
 
 		changed = updateNIC(&nic, &lb, i)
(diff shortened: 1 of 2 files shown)


pkg/cluster/fixssh.go
```diff
@@ -33,9 +31,24 @@ func (m *manager) fixSSH(ctx context.Context) error {
...
+	}
+
+	err = m.checkandUpdateNIC(ctx, resourceGroup, infraID, lb)
+	if err != nil {
+		m.log.Warnf("Failed checking and Updating Network Interface with err %s", err)
+	}
+	return nil
+}
+
+func (m *manager) checkAndUpdateLB(ctx context.Context, resourceGroup string, lbName string) (lb mgmtnetwork.LoadBalancer, err error) {
```
s-amann, line 48: now that we have some more methods defined can you please add tests around the specific functionality they provide? Currently the tests in `fixssh_test.go` do not cover any scenarios where `fixssh()` would return an error.

s-amann (changes requested):
Thank you for the work on this! I just have two concerns regarding the error paths in the changes.

pkg/cluster/fixssh.go
mhsreddy: Thanks Spencer for your feedback. Now I added couple of tests where the fixssh() would return an error.

pkg/cluster/fixssh.go
```diff
@@ -82,22 +112,34 @@ func (m *manager) fixSSH(ctx context.Context) error {
 			}
 		}
 	}
+	return nil
+}
 
+func (m *manager) removeBackendPoolsFromNIC(ctx context.Context, resourceGroup, nicName string, nic *mgmtnetwork.Interface) error {
+	ipc := (*nic.InterfacePropertiesFormat.IPConfigurations)[0]
```
darthhexx, line 119: Please can we defend against `*nic.InterfacePropertiesFormat.IPConfigurations` being invalid here with a check for:
```go
if nic.InterfacePropertiesFormat.IPConfigurations == nil || len(*nic.InterfacePropertiesFormat.IPConfigurations) == 0 {
	return fmt.Errorf("unable to remove backendpools from NIC as there are no IP configurations for %s in resource group %s", nicName, resourceGroup)
}
```

pkg/cluster/fixssh.go
```diff
@@ -82,22 +112,34 @@ func (m *manager) fixSSH(ctx context.Context) error {
...
+		*(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools = []mgmtnetwork.BackendAddressPool{}
+		return m.interfaces.CreateOrUpdateAndWait(ctx, resourceGroup, nicName, *nic)
+	}
 	return nil
 }
 
-func updateNIC(nic *mgmtnetwork.Interface, lb *mgmtnetwork.LoadBalancer, i int) bool {
+func updateNIC(nic *mgmtnetwork.Interface, lb *mgmtnetwork.LoadBalancer, i int, infraID string) bool {
 	id := fmt.Sprintf("%s/backendAddressPools/ssh-%d", *lb.ID, i)
+	ipc := (*nic.InterfacePropertiesFormat.IPConfigurations)[0]
```
darthhexx, line 130: We also need a check for the IPConfiguration's validity at the beginning of this function.

pkg/cluster/fixssh.go
```diff
@@ -82,22 +112,34 @@ func (m *manager) fixSSH(ctx context.Context) error {
...
+	if ipc.LoadBalancerBackendAddressPools != nil {
+		m.log.Printf("Removing Loadbalancer Backend Address Pools from NIC %s with no VMs attached", nicName)
+		*(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools = []mgmtnetwork.BackendAddressPool{}
+		return m.interfaces.CreateOrUpdateAndWait(ctx, resourceGroup, nicName, *nic)
+	}
 	return nil
 }
 
-func updateNIC(nic *mgmtnetwork.Interface, lb *mgmtnetwork.LoadBalancer, i int) bool {
+func updateNIC(nic *mgmtnetwork.Interface, lb *mgmtnetwork.LoadBalancer, i int, infraID string) bool {
```
darthhexx, line 128: Is there a reason that `updateNIC` and `updateLB` functions are defined as part of the `*manager`, where as `removeBackendPoolsFromNIC` is?

darthhexx (changes requested):

pkg/cluster/fixssh.go
mhsreddy: Updated check for NIC IP configurations validation.

pkg/cluster/fixssh.go
mhsreddy: Updated check for NIC IPConfiguration's validity

pkg/cluster/fixssh.go
mhsreddy: `updateNIC `and `updateLB` functions are now part of *manager and also we can now use manger to log messages and warnings directly.

pkg/cluster/fixssh.go
```diff
@@ -56,52 +74,99 @@ func (m *manager) fixSSH(ctx context.Context) error {
...
+			ID: &ilbID,
+		})
+	}
 
-	return true
+	return updateSSHPool || updateILBPool
 }
 
-func updateLB(lb *mgmtnetwork.LoadBalancer) (changed bool) {
+func (m *manager) updateLB(ctx context.Context, lb *mgmtnetwork.LoadBalancer) (lbChanged bool) {
```
darthhexx, line 169: All the changes to this function aren't required. If you were going to use the logger, then is would make sense to make it part of the manager interface, but since we aren't we shouldn't make any of these changes.

pkg/cluster/fixssh.go
```diff
@@ -33,21 +31,41 @@ func (m *manager) fixSSH(ctx context.Context) error {
...
+func (m *manager) checkAndUpdateLB(ctx context.Context, resourceGroup string, lbName string) (lb mgmtnetwork.LoadBalancer, err error) {
+	lb, err = m.loadBalancers.Get(ctx, resourceGroup, lbName, "")
+	if err != nil {
+		return lb, err
+	}
+
+	lbChanged := m.updateLB(ctx, &lb)
 
-	if changed {
+	if lbChanged {
```
darthhexx, line 58: It would read better if we dropped the `lbChanged` variable altogether:
```go
if m.updateLB(ctx, &lb) {
```

pkg/cluster/fixssh.go
```diff
@@ -56,52 +74,99 @@ func (m *manager) fixSSH(ctx context.Context) error {
...
+				m.log.Warnf("Fallback failed with err %s", err)
+				return err
 			}
 		}
 
-		changed = updateNIC(&nic, &lb, i)
+		nicChanged := m.updateNIC(ctx, &nic, nicName, &lb, i, infraID)
 
-		if changed {
+		if nicChanged {
```
darthhexx, line 107: It would also read better if we dropped the `changed` variable:
```go
if updateNIC(&nic, &lb, i) {
```

darthhexx (changes requested):

pkg/cluster/fixssh.go
mhsreddy: Dropped `lbchanged` variable altogether as you suggested.

pkg/cluster/fixssh.go
mhsreddy: Dropped `changed` variable in `updateNIC` function

pkg/cluster/fixssh_test.go
```diff
@@ -149,16 +152,47 @@ func TestFixSSH(t *testing.T) {
...
+				},
+			},
+		}
+	}
+
+	ifNoVmAfter := func(nic *mgmtnetwork.Interface) *mgmtnetwork.Interface {
+		emptyAddressPool := make([]mgmtnetwork.BackendAddressPool, 0)
+		(*nic.InterfacePropertiesFormat.IPConfigurations)[0].InterfaceIPConfigurationPropertiesFormat.LoadBalancerBackendAddressPools = &emptyAddressPool
+		return nic
+	}
```
facchettos, line 178: I know you are following the existing pattern, but this is a bad pattern. could we avoid having function variables in the test structs? (and in a follow up, fix the whole test file)

facchettos (changes requested):

pkg/cluster/fixssh_test.go
darthhexx: I agree on the functions vs variables change. Let's make that change @mhsreddy and we can merge after that.

As an FYI, there is a large PUCM planned for Hive, so we'll need to be extra vigilant on merging this change.

darthhexx (approved):
We must create a ticket to refactor these tests in a follow up, but the functionality LGTM.

pkg/cluster/fixssh_test.go
mhsreddy: @facchettos I converted the variables to functions as you suggested. Please review the changes. I created a new Jira Story to refactor the entire test file.

facchettos (approved):
I'll let @darthhexx merge it since he has more context on the thing/timing required 
Good job sudharsan!

s-amann (approved):

PR #2517, opened 2022-11-04
Az aro validate cmd

ellis-johnson:
Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/14618792 

Adding new az aro validate extension to validate permissions needed to create an ARO Cluster

Test plan for issue:
Unit Testing

Yes


python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -0,0 +1,279 @@
...
+
+            parts = parse_resource_id(subnet)
+
+            network_client = get_mgmt_service_client(
+                cmd.cli_ctx, ResourceType.MGMT_NETWORK)
+
+            auth_client = get_mgmt_service_client(
+                cmd.cli_ctx, ResourceType.MGMT_AUTHORIZATION, api_version="2015-07-01")
+
+            try:
```
s-amann, line 140: Is it possible to also check at this time if the subnet has an NSG already attached? And if so, should we?

I know we currently run into a race if a newly created VNET/subnet is used when creating a cluster the RP can lose the race to azure and azure applies its own NSG before the RP can.

python/az/aro/azext_aro/tests/latest/unit/test_dynamic_validators.py
```diff
@@ -0,0 +1,371 @@
...
+            "child_type_1": "subnets",
+            "resource_group": None,
+            "name": None,
+            "child_name_1": None
+        },
+        Mock(),
+        None
+    ),
+    (
+        "should return missing permission when actions are not permitted",
```
s-amann, line 232:
```suggestion
        "should return disabled permission when actions are not permitted",
```

python/az/aro/azext_aro/_dynamic_validators.py
ellis-johnson: I'm actually not too sure on this. Would this be a check that is needed in order to successfully create the cluster? @darthhexx Thoughts?

darthhexx: I haven't come across this in my tests before, but if we didn't have the ACLs to replace the Azure added NSG with the RP NSG, then this could well be an issue. Let's sync on this @ellis-johnson and get some test cases together to see if we can deterministically test for this edge case.

ellis-johnson: I discussed with @darthhexx and decided to add a check in to see if the subnets are already assigned to an NSG

s-amann (approved):
lgtm, thank you!

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -0,0 +1,286 @@
...
+                          "is already assigned to this subnet. Ensure there a no Network "\
+                          "Security Groups assigned to cluster subnets before cluster creation"
+                error = [f"{key}", parts['child_name_1'], message]
+                errors.append(error)
+
+        return errors
+    return _validate_subnet
+
+
+def dyn_validate_cidr_ranges():
```
tony-schndr, line 173: Would it make sense to also check if the pod cidr is large enough to provision master nodes like we do in basic checks?

tony-schndr (commented):
Hey Ellis, just a few comments/questions.  Looks great overall!

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -0,0 +1,279 @@
...
+from azure.cli.core.azclierror import CLIInternalError, InvalidArgumentValueError, \
+    RequiredArgumentMissingError
+from azure.core.exceptions import ResourceNotFoundError
+from knack.log import get_logger
+from msrestazure.tools import is_valid_resource_id
+from msrestazure.tools import parse_resource_id
+from msrestazure.azure_exceptions import CloudError
+from azext_aro._validators import validate_vnet, validate_cidr
+from azext_aro._rbac import has_role_assignment_on_resource
+from humanfriendly.terminal.spinners import AutomaticSpinner
```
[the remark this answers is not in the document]
ellis-johnson: We are only able to import safe imports currently in python so these were already accessible in az

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -0,0 +1,286 @@
...
+
+            validate_vnet(cmd, namespace)
+
+            parts = parse_resource_id(vnet)
+
+            network_client = get_mgmt_service_client(
+                cmd.cli_ctx, ResourceType.MGMT_NETWORK)
+
+            auth_client = get_mgmt_service_client(
+                cmd.cli_ctx, ResourceType.MGMT_AUTHORIZATION, api_version="2015-07-01")
```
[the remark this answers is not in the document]
ellis-johnson: So the validation of the resourse_id at the top is slightly different for the subnet vs the vnet. The vnet will try and validate itself and fail if its not valid. Whereas the subnet will try and validate itself and if it fails it will try and validate the vnet and retrieve the subnet from the vnet so it wouldn't be a very nice refactor. However, I have refactored the retrieving of the parts, network_client and auth_client into a new function as this is identical between the functions

python/az/aro/azext_aro/_dynamic_validators.py
ellis-johnson: Added this in!

python/az/aro/azext_aro/custom.py
```diff
@@ -155,6 +158,93 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
...
+                 client_secret,  # pylint: disable=unused-argument
+                 vnet_resource_group_name=None,  # pylint: disable=unused-argument
+                 pod_cidr=None,  # pylint: disable=unused-argument
+                 service_cidr=None,  # pylint: disable=unused-argument
+                 master_vm_size=None,  # pylint: disable=unused-argument
+                 disk_encryption_set=None,  # pylint: disable=unused-argument
+                 worker_vm_size=None,  # pylint: disable=unused-argument
+                 worker_vm_disk_size_gb=None,  # pylint: disable=unused-argument
+                 worker_count=None,  # pylint: disable=unused-argument
+                 ):
```
darthhexx, line 178: This function call is missing the new features, apiserver_visibility, ingress_visibility, and version. For validating version you will need to [fetch the list from the RP](https://github.com/Azure/ARO-RP/blob/cab5146c98df281050a99c065bbb2a46af6ff8ce/python/az/aro/azext_aro/custom.py#L318) and validate dynamically.

darthhexx (changes requested):

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -0,0 +1,291 @@
...
+
+            if subnet_obj.network_security_group is not None:
+                message = f"A Network Security Group \"{subnet_obj.network_security_group.id}\" "\
+                          "is already assigned to this subnet. Ensure there a no Network "\
+                          "Security Groups assigned to cluster subnets before cluster creation"
+                error = [f"{key}", parts['child_name_1'], message]
+                errors.append(error)
+
+        return errors
+    return _validate_subnet
```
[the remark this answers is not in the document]
ellis-johnson: Fixed

python/az/aro/azext_aro/custom.py
```diff
@@ -71,9 +75,19 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
...
+        worker_vm_size = worker_vm_size or 'Standard_D2s_v3'
+    else:
+        worker_vm_size = worker_vm_size or 'Standard_D4s_v3'
+
+    if apiserver_visibility is not None:
+        apiserver_visibility = apiserver_visibility.capitalize()
+
+    if ingress_visibility is not None:
+        ingress_visibility = ingress_visibility.capitalize()
+
```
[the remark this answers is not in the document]
ellis-johnson: Fixed

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -0,0 +1,291 @@
...
+from azure.cli.core.azclierror import CLIInternalError, InvalidArgumentValueError, \
+    RequiredArgumentMissingError
+from azure.core.exceptions import ResourceNotFoundError
+from knack.log import get_logger
+from msrestazure.tools import is_valid_resource_id
+from msrestazure.tools import parse_resource_id
+from msrestazure.azure_exceptions import CloudError
+from azext_aro._validators import validate_vnet, validate_cidr
+from azext_aro._rbac import has_role_assignment_on_resource
+from humanfriendly.terminal.spinners import AutomaticSpinner
```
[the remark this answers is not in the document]
ellis-johnson: Replaced with azure.core version of spinner

python/az/aro/azext_aro/custom.py
ellis-johnson: Added

python/az/aro/azext_aro/custom.py
```diff
@@ -155,6 +160,116 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
...
+        if len(errors) > 0:
+            logger.error("Issues found blocking cluster creation.\n")
+            headers = ["Type", "Name", "Error"]
+            table = tabulate(errors, headers=headers, tablefmt="grid")
+            print(f"\n{table}")
+        else:
+            print("\nNo Issues on subscription blocking cluster creation\n")
+        if tempAppCreated:
+            aad.delete_application(client_id)
+    except Exception as e:
```
kimorris27, line 267: Wherever there is a `raise` statement within the `try` clause, the code already checks `if tempAppCreated` and calls `aad.delete_application()` accordingly, so the call to `aad.delete_application()` within this `except` clause will fail in those cases - the app will already have been deleted and will no longer exist.

There are different ways to fix this, but I would do it by removing all instances of:
```
if tempAppCreated:
    aad.delete_application(client_id)
```

from the `try` clause and replacing the `except` clause with a `finally` clause. That way if an exception is raised from within the `try` clause, the code will always check whether there is a temporary app that needs to be deleted, delete it if necessary, and then raise the exception.

kimorris27 (changes requested):
I still need to review the actual validation logic, but I found some problems in the case where the code creates and uses a temporary AAD app/SP. I put comments/recommendations on the relevant places in the code.

kimorris27:
I've finished looking over the whole PR, and on the whole it looks good to me! I have a few questions though:

- I noticed the code still isn't creating a temporary SP to go with the temporary app. I ended up testing the command with a temporary SP by adding in the code to create it. Did you leave it out intentionally?
- If I understand correctly, the subnets don't have to have route tables to be valid and ready for cluster creation. Why does the command error out if one of the subnets doesn't have a route table?
- If I use a temporary SP and pass a valid vnet, valid subnets, etc., the table printed at the end will include a row telling me that the temporary SP is missing the network contributor role (and it looks like the same thing would happen for the reader role if you pass a disk encryption set). This seems like it could be misleading for users. Would it make sense to either 1) avoid outputting these rows when the command is using a temporary SP or 2) print some sort of notification that these rows don't require any action from the user since they are referring to a temporary SP?

ellis-johnson:
> I've finished looking over the whole PR, and on the whole it looks good to me! I have a few questions though:
> 
> * I noticed the code still isn't creating a temporary SP to go with the temporary app. I ended up testing the command with a temporary SP by adding in the code to create it. Did you leave it out intentionally?
> * If I understand correctly, the subnets don't have to have route tables to be valid and ready for cluster creation. Why does the command error out if one of the subnets doesn't have a route table?
> * If I use a temporary SP and pass a valid vnet, valid subnets, etc., the table printed at the end will include a row telling me that the temporary SP is missing the network contributor role (and it looks like the same thing would happen for the reader role if you pass a disk encryption set). This seems like it could be misleading for users. Would it make sense to either 1) avoid outputting these rows when the command is using a temporary SP or 2) print some sort of notification that these rows don't require any action from the user since they are referring to a temporary SP?

I have created a temporary SP for the temporary app and make sure both are cleaned up at the end of validation. I have removed a necessity for a route table and just check permissions if there is a route table. And I have not included checks against the app service principal if it is created temporarily! Thanks for the thorough review!

python/az/aro/azext_aro/custom.py
```diff
@@ -191,22 +191,28 @@ def __init__(self, disk_encryption_id, master_subnet_id, worker_subnet_id):
...
 
     tempAppCreated = False
+    tempSPCreated = False
+    client_sp_id = None
 
     try:
         if client_id is None or client_secret is None:
-            random_id = generate_random_id()
+            # random_id = generate_random_id()
+            random_id = "ellis-test"
```
kimorris27, line 200: I assume you want to revert this change 😉

python/az/aro/azext_aro/custom.py
```diff
@@ -252,11 +258,14 @@ def __init__(self, disk_encryption_id, master_subnet_id, worker_subnet_id):
         if len(errors) > 0:
             logger.error("Issues found blocking cluster creation.\n")
```
kimorris27, line 259: Given that `az aro create` will assign the needed roles to both the RP SP and the client SP if they are missing them, should information about missing role assignments be printed separately from the other validation failures? That way the information is still there in case customers want to create the role assignments themselves, but it's clear that cluster creation won't be blocked, which is what the output currently indicates.

kimorris27 (changes requested):
The changes in response to my other comments LGTM! I found two last things.

kimorris27 (approved):

python/az/aro/azext_aro/_aad.py
```diff
@@ -51,6 +54,19 @@ def create_service_principal(self, app_id):
                 logger.warning("%s; retry %d of %d", ex, retries, max_retries)
                 time.sleep(10)
 
+    def delete_service_principal(self, app_id):
+        max_retries = 3
```
jaitaiwan, line 58: Because we keep track of max_retries for a couple of functions, it would make more sense to make this a class-wide property.

python/az/aro/azext_aro/_aad.py
```diff
@@ -51,6 +54,19 @@ def create_service_principal(self, app_id):
...
+        retries = 0
+        while True:
+            try:
+                return self.client.service_principal_delete(app_id)
+            except GraphError as ex:
+                if retries >= max_retries:
+                    raise
```
jaitaiwan, line 65: Stylistic preference but as per PEP8 "explicit is better than implicit" I'd prefer to see `raise ex` not just `raise`.

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -0,0 +1,335 @@
...
+    return errors
+
+
+def get_subnet(client, subnet, subnet_parts):
+    try:
+        subnet_obj = client.subnets.get(subnet_parts['resource_group'],
+                                        subnet_parts['name'],
+                                        subnet_parts['child_name_1'])
+    except Exception as err:
+        if isinstance(err, ResourceNotFoundError):
```
jaitaiwan, line 66: I believe you can simply use `except ResourceNotFoundError as err` and then another `except` underneath without needing `isinstance`

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -0,0 +1,335 @@
...
+
+    return _validate_cidr_ranges
+
+
+def dyn_validate_resource_permissions(service_principle_ids, resources):
+    def _validate_resource_permissions(cmd,
+                                       namespace):  # pylint: disable=unused-argument
+        errors = []
+
+        hook = cmd.cli_ctx.get_progress_controller()
```
jaitaiwan, line 259: I see this hook functionality being repeated a bit. It's looks like it would be a great candidate for a decorator function like 
```
@shows_progress('Validating resource permissions')
def _validate_resource_permissions(cmd, namespace):
```

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -0,0 +1,335 @@
...
+            errors.append(["OpenShift Version",
+                           namespace.version,
+                           f"{namespace.version} is not a valid version, valid versions are {versions}"])
+
+        hook.end()
+        return errors
+    return _validate_version
+
+
+def validate_cluster_create(cmd,  # pylint: disable=unused-argument
```
jaitaiwan, line 314: Unused arguments should be able to be turned off by prefixing them with an underscore.

python/az/aro/azext_aro/custom.py
```diff
@@ -155,6 +160,116 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
                        parameters=oc)
 
 
+def aro_validate(cmd,  # pylint: disable=too-many-locals,too-many-statements
```
jaitaiwan, line 163: Same comment again, could use underscores instead of disabling pylint rules.

jaitaiwan (changes requested):
I'd like to see the validations have comments for each function as well so it's clear the what & how of the validation function without having to read the function body.

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -83,13 +84,29 @@ def get_clients(key, cmd):
     return parts, network_client, auth_client
 
 
+def progress_tracking_factory(msg):
```
jaitaiwan, line 87: Idiomatic python means we probably wouldn't name it "factory" as it's a decorator function. Using more imperative language like `display_progress` or `track_progress` is more idiomatic in my opinion. I'd also like to see it documented.

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -172,18 +191,21 @@ def _validate_subnet(cmd, namespace):
...
 
+# Validating that the cidr ranges between the master_subnet, worker_subnet,
+# service_cidr and pod_cidr do not overlap at all
 def dyn_validate_cidr_ranges():
+    prog = progress_tracking_factory("Validating no overlapping CIDR Ranges on subnets")
+
+    @prog
     def _validate_cidr_ranges(cmd, namespace):
-        addresses = []
+        MIN_CIDR_PREFIX = 23
```
jaitaiwan, line 206: Love this!

jaitaiwan (dismissed):
The PR is looking great now IMO. I have a couple of nitpicky points if you wanna address them. Otherwise looking schmick!

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -0,0 +1,345 @@
...
+
+        try:
+            network_client.virtual_networks.get(parts['resource_group'], parts['name'])
+        except ResourceNotFoundError as err:
+            raise InvalidArgumentValueError(
+                f"Invalid --{key.replace('_', '-')}, error when getting '{vnet}': {str(err)}") from err
+
+        except Exception as err:
+            raise CLIInternalError(
+                f"Unexpected error when getting subnet '{vnet}': {str(err)}") from err
```
s-amann, line 128: nit
```suggestion
                f"Unexpected error when getting vnet '{vnet}': {str(err)}") from err
```

python/az/aro/azext_aro/custom.py
```diff
@@ -153,6 +158,90 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
...
+    aad = AADManager(cmd.cli_ctx)
+
+    rp_client_sp_id = aad.get_service_principal_id(resolve_rp_client_id())
+    if not rp_client_sp_id:
+        raise ResourceNotFoundError("RP service principal not found.")
+
+    sp_obj_ids = [rp_client_sp_id]
+
+    if client_id is not None:
+        client_sp_id = aad.get_service_principal_id(client_id)
```
hawkowl, line 198: You can probably collapse this into one line

python/az/aro/azext_aro/_aad.py
```diff
@@ -39,16 +44,27 @@ def get_application_object_id_by_client_id(self, client_id):
...
+        retries = 0
+        while True:
+            try:
+                return self.client.service_principal_delete(app_id)
+            except GraphError as ex:
+                if retries >= MAX_RETRIES:
+                    raise ex
                 retries += 1
-                logger.warning("%s; retry %d of %d", ex, retries, max_retries)
+                logger.warning("%s; retry %d of %d", ex, retries, MAX_RETRIES)
```
hawkowl, line 67: I don't think you need this anymore either

hawkowl (changes requested):
Some minor cleanups

hawkowl (approved):
thumbs up once tests pass

PR #2520, opened 2022-11-07
Add a Fluentbit image build using Mariner containers

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-1384

This builds fluentbit from source in a Mariner build container and then places it in a Mariner runtime container.

The build is minimised to include as little functionality as possible, removing all available features except ones required by ARO for use in conjunction with MDSD.

Build portion at https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines/pullrequest/7089561 (merged)

### Test plan for issue:

Needs to be manually tested in a cluster.

N/A

### DIFF
--- a/Dockerfile.fluentbit
+++ b/Dockerfile.fluentbit
@@ -1,9 +1,133 @@
-ARG REGISTRY
-FROM ${REGISTRY}/ubi8/ubi-minimal
+ARG MARINER_VERSION
+FROM mcr.microsoft.com/cbl-mariner/base/core:${MARINER_VERSION} as builder
 ARG VERSION
-RUN echo -e '[td-agent-bit]\nname=td-agent-bit\nbaseurl=https://packages.fluentbit.io/centos/$releasever/$basearch' >/etc/yum.repos.d/td-agent-bit.repo && \
-    rpm --import https://packages.fluentbit.io/fluentbit.key && \
-    microdnf update && \
-    microdnf install td-agent-bit-$VERSION && \
-    microdnf clean all
+RUN tdnf repolist --refresh && \
+    tdnf update -y && \
+    tdnf install -y ca-certificates flex build-essential openssl-devel systemd-devel cmake && \
+    tdnf clean all
+
+RUN mkdir /build && \
+    cd /build && \
+    wget -q https://github.com/fluent/fluent-bit/archive/refs/tags/v$VERSION.tar.gz && \
+    tar -xf v$VERSION.tar.gz && \
+    cd fluent-bit-$VERSION/build && \
+    cmake -DCMAKE_INSTALL_PREFIX=/opt/td-agent-bit/ \
+        -DCMAKE_INSTALL_SYSCONFDIR=/etc/ \
+        -DFLB_AWS="Off" \
+        -DFLB_CONFIG_YAML="Off" \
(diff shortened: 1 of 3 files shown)


hawkowl:
I tested the images (as well as with Mariner-based MDSD for comparison) on a test cluster:

cpu-mariner
memory-mariner

The overall CPU usage of the fluentbit container was reduced by 30% (so ~0.5% of a core across the cluster), but more interestingly, the memory usage was reduced by 60% (60MB across the base cluster!)

The logging rate of the test cluster was consistent at 8200 logs/hr.

SudoBrendan (commented):
Seems like a quick win for a huge benefit - nice! Worth mentioning that the description of this PR mentions not changing the default installed version - but we see changes to /pkg/util/version/const.go - which will change what customer clusters use.

Have these changes also been tested on the RP side? I'm seeing it using a different version than shown here in INT atm.

pkg/util/version/const.go
```diff
@@ -64,7 +64,7 @@ var (
 
 // FluentbitImage contains the location of the Fluentbit container image
 func FluentbitImage(acrDomain string) string {
-	return acrDomain + "/fluentbit:1.7.8-1"
+	return acrDomain + "/fluentbit:1.9.9-1"
```
jhoreman, line 67: Should we try to add some way to populate this fluentbit version dynamically? It updates pretty often and the older versions aren't typically available anymore as fluentbit releases. Otherwise this change looks great and the impact is measurable.

s-amann (approved):

pkg/util/version/const.go
hawkowl: This compiles them, so the requirement for an upstream release is not required.

Dockerfile.fluentbit
```diff
@@ -1,9 +1,131 @@
...
+    -DFLB_OUT_PROMETHEUS_REMOTE_WRITE="Off" \
+    -DFLB_OUT_S3="Off" \
+    -DFLB_OUT_SKYWALKING="Off" \
+    -DFLB_OUT_SLACK="Off" \
+    -DFLB_OUT_SPLUNK="Off" \
+    -DFLB_OUT_STACKDRIVER="Off" \
+    -DFLB_OUT_STDOUT="Off" \
+    -DFLB_OUT_SYSLOG="Off" \
+    -DFLB_OUT_TD="Off" \
+    -DFLB_OUT_WEBSOCKET="Off"
```
s-amann, line 112: are all these options required, or is this *just* to be more explicit with how we build the image?

pkg/util/version/const.go
s-amann, line 67: should this be `1.9.10` like we have in the makefile

s-amann (commented):
lgtm, just some questions below

Dockerfile.fluentbit
hawkowl: In v2 (not v1, but, uhhh, v2 would be nice later) those are all "On" by default.

pkg/util/version/const.go
hawkowl: No, since this should build it and such but not activate it in production yet.

PR #2521, opened 2022-11-07
Liveconfig: Use case insensitive matching for the AKS cluster lookup

darthhexx:

Fixes AKS cluster location matching in regions that return mixed case instance meta locations.

Currently the AKS cluster credentials are not matching the correct cluster in Canary due to EastUS2EUAP != eastus2euap and CentralUSEUAP != centraluseuap. This is an assumption, but we have logs with EastUS2EUAP and CentralUSEUAP and when running an `az aks list` the results show the locations are lowercase.

### Test plan for issue:

Tested against a Dev AKS cluster.

No.


s-amann (approved):

UlrichSchlueter:
lgtm

jharrington22:
LGTM

Issue #54, opened 2020-01-09
Tests for the database package


m1kola:
An issue for tracking progress of adding tests for the `database` package

PR #2529, opened 2022-11-09
Add unit tests for database package

s-fairchild:
Billing and Subscriptions tests are placed in `test/database` to reuse code in that package.

httptest is used to send http responses.
Add Mock for databaseaccounts.
Add DatabaseClient, MasterKeyClient interfaces to allow for mock testing.
Add unit tests for subscriptions.
Add billing unit tests.
Add default http.Client to database.go, update references to use DatabaseClient interface. This will keep from importing "net/http" in every package calling it and allow easy changing within database if needed."
Add unit tests for clustermanager in database package
Add asyncoperations unit tests
Add Monitor tests
Add gateway unit tests

Tests for billing and subscriptions are placed in test/database to reuse
fake testing code due without import cycling.

https://github.com/Azure/ARO-RP/issues/54

Fixes

Increase test coverage for database package.

### Test plan for issue:
Ran `make test-go` without issues, run tests in CI pipelines.
Ran `go test -coverprofile=file -coverpkg ../../pkg/database/`

No, test coverage only


pkg/database/billing_test.go
```diff
@@ -0,0 +1,209 @@
...
+			err := billing.Delete(ctx, doc)
+			if err != nil && err.Error() != tt.wantErr ||
+				err == nil && tt.wantErr != "" {
+				t.Errorf("\n%v\n !=\n%v", err, tt.wantErr)
+			}
+		})
+	}
+}
+
+// Cannot import test/database due to import cycling
```
hawkowl, line 187: Why not put these tests in test/database?

pkg/database/subscriptions_test.go
```diff
@@ -0,0 +1,449 @@
...
+			if err != nil && err.Error() != tt.wantErr ||
+				err == nil && tt.wantErr != "" {
+				t.Errorf("\n%v\n !=\n%v", err, tt.wantErr)
+			}
+		})
+	}
+}
+
+// Cannot import "test/database" due to import cycling
+func NewFakeSubscriptions() (db Subscriptions, client *cosmosdb.FakeSubscriptionDocumentClient) {
```
hawkowl, line 407: Can this also move to test/database?

hawkowl (changes requested):
I think we should try and minimise the duplication as much as possible and see if we can move the tests to somewhere that doesn't cause import loops, or otherwise restructure it.

One of the main motivators behind the database fakes was that we had duplication and mocks of database portions scattered about which often meant that we had wholly insufficient or invalid tests.

s-fairchild:
@hawkowl The tests can be moved to `test/database`. I hadn't really thought about that, I was mostly keeping them in `database` because that's the package being tested. Moving them should work as everything being tested is public.
I'll work on moving the tests.

cadenmarchese (approved):
LGTM, thank you for improving coverage.

PR #2532, opened 2022-11-10
fix/delete usage of flags at package-level and include Go test direct…

AldoFusterTurpin:

Fixes [Remove usage of Go flags at package level to allow usage of -coverpkg=./... flag in go test](https://msazure.visualstudio.com/AzureRedHatOpenShift/_backlogs/backlog/Red%20Hat%20EMEA/Stories/?workitem=16201266)

This task is NOT JUST to ensure that we can execute go test with -coverpkg=./..., it is also to get rid of the bad practice of having variables defined at package level.

When we execute Go tests with the -coverpkg=./... option, we encounter 165 errors about the redefinition of the same flag (flag package) with the same name. 

The errors look like:
=== FAIL: test/util/log  (0.00s)
/var/folders/3w/x8khwvt57lj7t8fvqhy5sb8m0000gn/T/go-build4150742334/b5045/log.test flag **redefined: keyFile**
panic: /var/folders/3w/x8khwvt57lj7t8fvqhy5sb8m0000gn/T/go-build4150742334/b5045/log.test **flag redefined: keyFile**
...

You can reproduce the errors going to master and doing (the same command we use to test but using the -coverpkg=./... option):
_go run ./vendor/gotest.tools/gotestsum/main.go --format pkgname --junitfile report.xml -- -tags=aro,containers_image_openpgp -coverpkg=./... -coverprofile=cover_coverpkg.out ./..._

This happens because we try to have multiple flag variables at package level with the same name at the same time. That is not just bad practice, it also panics, even if our intention is NOT to use those different flags with the same name at the same time, the code will panic when initializing the packages indirectly when running the above go test command).

The new go test command included in the Makefile in this PR lets you see (not in the CLI output but in the coverage file using the web browser) the production code that is covered by test from different packages rather than just the same packages where the production code leaves. i.e. it includes indirect references coverage. (More info [here](https://youtu.be/JTLB7j8M85A?t=1687) and [here](https://pkg.go.dev/cmd/go#hdr-Testing_flags))

### Test plan for issue:

There are already unit tests

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ gomock_reflect_*
 /mdm_statsd.socket
 /uts.txt
 /cover.out
+/cover_coverpkg.out
 /coverage.*
 /report.xml
 /deploy/config.yaml
--- a/Makefile
+++ b/Makefile
@@ -180,6 +180,9 @@ validate-fips:
 unit-test-go:
 	go run ./vendor/gotest.tools/gotestsum/main.go --format pkgname --junitfile report.xml -- -tags=aro,containers_image_openpgp -coverprofile=cover.out ./...
 
+unit-test-go-coverpkg:
+	go run ./vendor/gotest.tools/gotestsum/main.go --format pkgname --junitfile report.xml -- -tags=aro,containers_image_openpgp -coverpkg=./... -coverprofile=cover_coverpkg.out ./...
+
 lint-go:
 	hack/lint-go.sh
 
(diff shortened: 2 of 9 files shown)


hawkowl (approved):

s-amann (approved):

SudoBrendan (approved):

petrkotas (approved):
Thank you @aldofusterturpin

PR #2534, opened 2022-11-10
Moves ARO operator E2E tests to use `Eventually`

m1kola:

This helps with troubleshooting of E2Es as it makes easier to understand what assertion within the wait code failed and why.

### Test plan for issue:

Run E2E

No, just E2E refactoring

### DIFF
--- a/test/e2e/operator.go
+++ b/test/e2e/operator.go
@@ -19,12 +19,10 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/ghodss/yaml"
 	configv1 "github.com/openshift/api/config/v1"
-	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/ugorji/go/codec"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


test/e2e/operator.go
```diff
@@ -123,39 +109,27 @@ var _ = Describe("ARO Operator - Internet checking", func() {
...
-				log.Warn(err)
-			}
-			return done, nil // swallow error
+
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(done).To(BeTrue())
 		}
 
 		By("checking that mdsd DaemonSet is ready before the test")
-		err := wait.PollImmediate(30*time.Second, 15*time.Minute, mdsdReady)
```
m1kola, line 152: `Eventually` will be using default polling and wait intervals set here:

https://github.com/Azure/ARO-RP/blob/00f1604fd229bf5b4e92f96b83040741b5a99494/test/e2e/setup.go#L411-L412

They are lower than what we had, but we can increase themon per-`Eventually` basis if we need it.

But I don't think that any of these waits need to be 15 minutes: I believe these situations either recover quickly or do not recover at all.

SudoBrendan (approved):
LGTM

facchettos (approved):

PR #2535, opened 2022-11-10
removes explicit rules from loadbalancer and enables outbound snat

tony-schndr:

Removes explicit outbound rule and enables outbound snat.

Remove the outbound rule to remove SNAT limitation and scale cluster past 62 nodes.

### Test plan for issue:

Made changes and created cluster.

NA for now


jharrington22 (changes requested):
@tony-schndr this `disableOutboundSnat` parameter. Reading the Azure [documentation](https://learn.microsoft.com/en-us/azure/load-balancer/outbound-rules) it seems setting this to `false` will enable Azure to automatically configure SNAT configuration based on the table referenced [here](https://learn.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#preallocatedports). I don't believe it will solve our issue, unless i'm missing something.

tony-schndr:
@jharrington22 Reading those docs I believe your right, it doesn't necessarily remove the SNAT limitation but just allows Azure to manage it for you.  200 nodes would still be 320 ports per node...

I will note though that this is very close to how the vanilla ocp installer is configuring the lb by comparing a cluster created from this PR and a vanilla ocp cluster using load balancer outbound type.

PR #2537, opened 2022-11-11
While create if cluster exists check provisioning state

SrinivasAtmakuri:
During cluster creation via hack script/e2e tests, if the cluster already exists we return nil, instead we should also check for the provisioning state of the cluster and return nil only if its not Failed.

Background:
For a failed e2e test case, due to resource quota limitation, on re-run e2e was still using the failed cluster and panicking. 
Instead, give a better error message.

Fixes

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -140,8 +140,11 @@ func New(log *logrus.Entry, environment env.Core, ci bool) (*Cluster, error) {
 }
 
 func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName string) error {
-	_, err := c.openshiftclustersv20200430.Get(ctx, vnetResourceGroup, clusterName)
+	clusterGet, err := c.openshiftclustersv20220904.Get(ctx, vnetResourceGroup, clusterName)
 	if err == nil {
+		if clusterGet.ProvisioningState == mgmtredhatopenshift20220904.Failed {
+			return fmt.Errorf("cluster exists and is in failed provisioning state, please delete and retry")
+		}
 		c.log.Print("cluster already exists, skipping create")
 		return nil
 	}


SrinivasAtmakuri:
Would be nice to trigger the delete code automatically instead of returning an error and then trying to delete? Would this be wise? Looking for opinions. Thanks!

carlowisse (approved):
LGTM.

pkg/util/cluster/cluster.go
```diff
@@ -140,8 +141,11 @@ func New(log *logrus.Entry, environment env.Core, ci bool) (*Cluster, error) {
 }
 
 func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName string) error {
-	_, err := c.openshiftclustersv20200430.Get(ctx, vnetResourceGroup, clusterName)
+	clusterGet, err := c.openshiftclustersv20200430.Get(ctx, vnetResourceGroup, clusterName)
 	if err == nil {
+		if clusterGet.ProvisioningState == mgmtredhatopenshift20200430.ProvisioningState(api.ProvisioningStateFailed) {
+			return fmt.Errorf("cluster is in failed provisioning state, please delete and retry")
```
anshulvermapatel, line 147: I think the return message can also be - 
`return fmt.Errorf("cluster exists and is in failed provisioning state, please delete and retry")`

anshulvermapatel:
> Would be nice to trigger the delete code automatically instead of returning an error and then trying to delete? Would this be wise? Looking for opinions. Thanks!

I agree

jhoreman:
> 

While I agree that executing the delete code automatically would be a cleaner process, it may mask issues as to why the cluster was not successfully created in the first place and may make for a frustrating customer experience. If we are confident that we capture enough logging / information around this then this is a good idea. Just want to make sure we don't lose data points upon failed creation that could help us troubleshoot.

s-amann (dismissed):

hawkowl (approved):

PR #2542, opened 2022-11-15
Migrates rest of the E2Es to use `Eventually` instead of `wait.Poll*`

m1kola:

This helps with troubleshooting of E2Es as it makes easier to understand what assertion within the wait code failed and why.

This is basically the same as https://github.com/Azure/ARO-RP/pull/2534, but for rest of E2Es.

### Test plan for issue:

Run E2E

No, just E2E refactoring

### DIFF
--- a/test/e2e/adminapi_cluster_getlogs.go
+++ b/test/e2e/adminapi_cluster_getlogs.go
@@ -5,19 +5,16 @@ package e2e
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 var _ = Describe("[Admin API] Kubernetes get pod logs action", func() {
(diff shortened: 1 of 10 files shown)


petrkotas (approved):
LGTM, thanks Mikalai

test/e2e/adminapi_cluster_getlogs.go
```diff
@@ -59,24 +56,12 @@ func testGetPodLogsOK(ctx context.Context, containerName, podName, namespace str
...
 		pod, err = clients.Kubernetes.CoreV1().Pods(namespace).Get(ctx, pod.Name, metav1.GetOptions{})
-		Expect(err).NotTo(HaveOccurred())
+		g.Expect(err).NotTo(HaveOccurred())
 
-		switch pod.Status.Phase {
-		case corev1.PodSucceeded:
-			return true, nil
-		case corev1.PodPending:
-			if pod.CreationTimestamp.Time.Add(5*time.Minute).Unix() < time.Now().Unix() {
-				return false, errors.New("pod was pending for more than 5min")
```
cblecker, line 71: Do we not care about the "takes more than 5 mins to schedule pod" case?

m1kola: It is covered on line 63:

```go
g.Expect(pod.Status.Phase).To(Equal(corev1.PodSucceeded))))
```

`Eventually` has a timeout (5 minutes, [default](https://github.com/Azure/ARO-RP/blob/180eb992fce1e2a132e7643a31edc4c3c04d3bd4/test/e2e/setup.go#L436)). So if we exeed the timeout due to not satisfying this matcher - we will see it in the failure message produced by line 63. It will say that we expected phase `Succeeded`, but got `Pending` (or something else if there was another reason).

cblecker: Makes sense. Thanks!

cblecker (approved):

PR #2547, opened 2022-11-17
Fixed input on Admin Portal V1. Updated loader-utils to 2.0.4 to mitigate vulnerability.

carlowisse:
[ADO Card](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/16298560)

Fixes - Input search selector for admin portal v1
Fixes - [loader-utils vulnerability](https://github.com/webpack/loader-utils/pull/225/files) (upgrade to 2.0.4)

After a previous merge that upgraded Bootstrap 4 to 5, JQuery dependency from Bootstrap was removed. We were using a sister Bootstrap package called `bootstrap-select` that no longer worked. In changing the element to `<select>`, the input to search the select element no longer worked.

There is no native way to have a searchable select element in Bootstrap5/HTML. The first implementation I used `<datalist>` but the width didn't suit ResourceIds length. And by design you cannot style a datalist. datalists filter search is also not reliable.

Opted to create and implement a fully custom searchable select element.

Vulnerability in loader-utils for admin portal v2 has been patched and packaged needed to be updated.

### Test plan for issue:
Manual
E2E

No.


portal/v1/src/select/select.js
```diff
@@ -0,0 +1,46 @@
...
+function selectResourceId() {
+    select_input.val(select_element.find(":selected").text());
+    select_element.css("visibility", "hidden");
+}
+
+select_input.on("keyup change", function () {
+    if (select_input.val().length >= 2) {
+        select_element.children().each(function (i) {
+            if (i != 0) {
+                if (!$(this).text().toLowerCase().match(select_input.val().toLowerCase())) {
```
s-amann, line 15: nit: can we invert this condition? imo it is better to avoid using leading `!` as they can be sneaky. 

e.g. `if (textIsEqual) { show } else { hide }`

s-amann (dismissed):
lgtm, just one minor nitpick

darthhexx (changes requested):
As discussed, we need to be able to differentiate between customer clusters that have the same subscription ID and resource group with a different cluster name. This can be accomplished with a horizontal scrollbar, plus a smaller font is also preferable.

There is also a small glitch when trying to re-select the currently selected cluster that needs to be addressed.

darthhexx (approved):

s-amann (approved):

PR #2551, opened 2022-11-21
Hive: exit if we enter ProvisionFailed status

darthhexx:

Fixes https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/15728844

Currently the Hive manager is only checking for `spec.Installed`, but this does not check for any failure conditions and results in all failures taking 60 minutes to return. This PR adds a single failure condition, Provision Failed, but scaffolding has been put in place making it trivial to add any additional error conditions in the future.

### Test plan for issue:

Tested cluster installation failure propagation in a dedicated RP instance.

No.

### DIFF
--- a/pkg/hive/manager.go
+++ b/pkg/hive/manager.go
@@ -5,6 +5,7 @@ package hive
 
 import (
 	"context"
+	"fmt"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	hiveclient "github.com/openshift/hive/pkg/client/clientset/versioned"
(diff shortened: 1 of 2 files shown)


pkg/hive/manager.go
```diff
@@ -160,10 +161,26 @@ func (hr *clusterManager) IsClusterDeploymentReady(ctx context.Context, doc *api
...
-	return false, err
+
+	for _, cond := range cd.Status.Conditions {
+		conditionStatus, found := checkFailureConditions[cond.Type]
+		if found && conditionStatus == cond.Status {
+			return false, fmt.Errorf("clusterdeployment has failed: %s == %s", cond.Type, cond.Status)
+		}
+	}
+
+	return false, nil
```
s-amann, line 183: wdyt about adding manager tests? I don't think we have any, and the logic here is growing

s-amann (approved):
lgtm, we should have a ticket for enhancing the tests if we are in a rush to get this out

pkg/hive/manager.go
darthhexx: Good point. I was racing to get this across the line, but there actually is time. I've added some tests.

s-amann (approved):
lgtm, thank you for adding the tests

cadenmarchese (approved):

cblecker (approved):

PR #2554, opened 2022-11-23
Adding VM's allocation status column to the machines table

anshulvermapatel:
Fixes
https://msazure.visualstudio.com/AzureRedHatOpenShift/_boards/board/t/Firefly/Stories/?workitem=9777520

Adds a column to the Machines table which displays the allocation status of the corresponding VM.


pkg/portal/cluster/machines.go
```diff
@@ -59,6 +68,59 @@ func (c *client) Machines(ctx context.Context) (*MachineListInformation, error)
...
+	computeResources, err := NewResourceClientFunction(env.Environment(), subscriptionDoc.ID, fpAuth).ListByResourceGroup(ctx, clusterRGName, "resourceType eq 'Microsoft.Compute/virtualMachines'", "", nil)
+	if err != nil {
+		return nil, err
+	}
+
+	vmAllocationStatus := make(VMAllocationStatus)
+	virtualMachineClient := NewVirtualMachineClientFunction(env.Environment(), subscriptionDoc.ID, fpAuth)
+	for _, res := range computeResources {
+		var vmName, allocationStatus string
+		if *res.Type != "Microsoft.Compute/virtualMachines" {
```
cadenmarchese, line 97: Should we leave out this `if` statement since we're already using the `resourceType eq 'Microsoft.Compute/virtualMachines` filter in the call to `ListByResourceGroup` above?

pkg/portal/cluster/fetcher.go
```diff
@@ -41,13 +44,28 @@ type client struct {
...
+
+func newAzureSideFetcher(resourceGroupName string, subscriptionDoc *api.SubscriptionDocument, env env.Interface) azureSideFetcher {
+	return azureSideFetcher{
+		resourceGroupName: resourceGroupName,
+		subscriptionDoc:   subscriptionDoc,
+		env:               env,
+	}
+}
+
+func newRealFetcher(log *logrus.Entry, dialer proxy.Dialer, doc *api.OpenShiftClusterDocument, azureazureSideFetcher azureSideFetcher) (*realFetcher, error) {
```
cadenmarchese, line 68: Maybe:
```suggestion
func newRealFetcher(log *logrus.Entry, dialer proxy.Dialer, doc *api.OpenShiftClusterDocument, fetcher azureSideFetcher) (*realFetcher, error) {
```

pkg/portal/cluster/machines_test.go
```diff
@@ -97,3 +111,148 @@ func TestMachines(t *testing.T) {
...
+		subscriptionID string,
+		authorizer autorest.Authorizer) features.ResourcesClient {
+		return mockResourcesClient
+	}
+	NewVirtualMachineClientFunction = func(environment *azureclient.AROEnvironment, subscriptionID string, authorizer autorest.Authorizer) compute.VirtualMachinesClient {
+		return mockVirtualMachinesClient
+	}
+
+	defer func() {
+		NewResourceClientFunction = features.NewResourcesClient
```
devamanv, line 130:
```suggestion
		NewResourceClientFunction = features.NewResourcesClient
		NewVirtualMachineClientFunction = compute.NewVirtualMachinesClient
```
Both assignments should be reversed.

pkg/portal/cluster/machines_test.go
```diff
@@ -97,3 +111,148 @@ func TestMachines(t *testing.T) {
...
+	}()
+
+	type test struct {
+		name    string
+		mocks   func(*test, *mock_env.MockInterface, *mock_refreshable.MockAuthorizer, *mock_features.MockResourcesClient, *mock_compute.MockVirtualMachinesClient)
+		wantErr string
+	}
+	for _, tt := range []*test{
+		{
+			name: "Everything runs fine",
```
devamanv, line 140: The test name could be made a little more descriptive and should clearly state the functionality you're trying to test.

pkg/portal/cluster/fetcher.go
```diff
@@ -12,14 +12,17 @@ import (
...
 	"github.com/Azure/ARO-RP/pkg/proxy"
 	"github.com/Azure/ARO-RP/pkg/util/restconfig"
+	"github.com/Azure/ARO-RP/pkg/util/stringutils"
 )
 
 // FetchClient is the interface that the Admin Portal Frontend uses to gather
 // information about clusters. It returns frontend-suitable data structures.
 type FetchClient interface {
 	Nodes(context.Context) (*NodeListInformation, error)
+	VMAllocationStatus(context.Context) (VMAllocationStatus, error)
```
devamanv, line 25: Since we have made changes to the interface, can we check to see if it requires generating and committing new  mocks for the interface?

pkg/portal/cluster/machines.go
```diff
@@ -5,11 +5,20 @@ package cluster
...
 
+	mgmtcompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/compute"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/features"
 )
 
+// Creating local vars for these functions in order to make them testable. The function definition will is altered while in the tests written.
```
devamanv, line 18: The intent here is to not make them local(which they aren't anyway, as they are exported), but to be able to swap the function definitions or make them swappable during testing. So, consider changing the comment.

pkg/portal/cluster/machines.go
anshulvermapatel: I think yes .. I am not able to recall .. why I put that condition there .. looking at the code now, it looks like the `if` condition is not required, hence removing it.

cmd/aro/portal.go
```diff
@@ -26,7 +26,9 @@ import (
 )
 
 func portal(ctx context.Context, log *logrus.Entry, audit *logrus.Entry) error {
-	_env, err := env.NewCore(ctx, log)
+	//_env, err := env.NewCore(ctx, log)
```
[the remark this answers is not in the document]
anshulvermapatel: done

pkg/portal/cluster/fetcher.go
anshulvermapatel: So there is another interface called fetcher .. so I don't want someone to confuse that fetcher with this one. This struct is specifically to fetch something from Azure side .. anything in future we want to fetch from Azure, this struct can be used or can be enhanced to fetch more stuff may be.

pkg/portal/cluster/machines_test.go
anshulvermapatel: Ohh yes .. my bad .. made the changes.

pkg/portal/cluster/machines_test.go
anshulvermapatel: Yup .. changed the title, do check and let me know.

pkg/portal/cluster/fetcher.go
anshulvermapatel: We do not have mocks for fetcher right now. Maybe we can generate them in a separate work item.

pkg/portal/cluster/machines.go
anshulvermapatel: So here .. in the comment, I have talked about creating local variables not making the functions local. Yes they are exported .. which is not required so I will make them local and also will change the comment to make it more descriptive to give an idea of why that is done.

cmd/aro/portal.go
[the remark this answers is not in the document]
facchettos: no :D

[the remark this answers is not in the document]
anshulvermapatel: pushed again with removed comments. This time for sure.

pkg/portal/cluster.go
```diff
@@ -166,6 +166,31 @@ func (p *portal) machines(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	b, err := json.MarshalIndent(machineVMAllocationStatus, "", "    ")
+	if err != nil {
+		p.internalServerError(w, err)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	_, _ = w.Write(b)
```
facchettos, line 191: can we check the error here ?

pkg/portal/portal.go
```diff
@@ -305,6 +311,7 @@ func (p *portal) aadAuthenticatedRoutes(r *mux.Router) {
 	r.NewRoute().PathPrefix("/api/{subscription}/{resourceGroup}/{clusterName}/clusteroperators").HandlerFunc(p.clusterOperators)
 	r.NewRoute().Methods(http.MethodGet).Path("/api/{subscription}/{resourceGroup}/{clusterName}").HandlerFunc(p.clusterInfo)
 	r.NewRoute().PathPrefix("/api/{subscription}/{resourceGroup}/{clusterName}/nodes").HandlerFunc(p.nodes)
+	r.NewRoute().PathPrefix("/api/{subscription}/{resourceGroup}/{clusterName}/vmallocationstatus").HandlerFunc(p.VMAllocationStatus) // Mine, change the route name and also the function name.
```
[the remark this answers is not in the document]
facchettos: also, why do we use path prefix here? shouldn't it be just `Path`?

pkg/portal/security_test.go
```diff
@@ -49,7 +49,8 @@ func TestSecurity(t *testing.T) {
 	controller := gomock.NewController(t)
 	defer controller.Finish()
 
-	_env := mock_env.NewMockCore(controller)
+	//_env := mock_env.NewMockCore(controller)
```
facchettos, line 52: comment

pkg/portal/info_test.go
```diff
@@ -23,7 +23,8 @@ func TestInfo(t *testing.T) {
 	controller := gomock.NewController(t)
 	defer controller.Finish()
 
-	_env := mock_env.NewMockCore(controller)
+	//_env := mock_env.NewMockCore(controller)
```
facchettos, line 26: comment

pkg/portal/cluster/machines.go
```diff
@@ -59,6 +68,55 @@ func (c *client) Machines(ctx context.Context) (*MachineListInformation, error)
...
+			// armResources = append(armResources, arm.Resource{
+			// 	Resource: res,
+			// })
+			continue
+		}
+
+		vmName = *vm.Name
+		instanceViewStatuses := vm.InstanceView.Statuses
+		for _, status := range *instanceViewStatuses {
+			if strings.Contains(*status.Code, "PowerState") {
```
facchettos, line 109: maybe doesn't matter too much, but isn't the real condition `string.HasPrefix(*status.Code, "PowerState/")` ?

pkg/portal/cluster/fetcher.go
```diff
@@ -41,13 +44,28 @@ type client struct {
...
+	machineClient    machineclient.Interface
+	azureSideFetcher azureSideFetcher
 }
 
-func newRealFetcher(log *logrus.Entry, dialer proxy.Dialer, doc *api.OpenShiftClusterDocument) (*realFetcher, error) {
+type azureSideFetcher struct {
+	resourceGroupName string
+	subscriptionDoc   *api.SubscriptionDocument
+	env               env.Interface
+}
```
facchettos, line 58: I am not sure I understand the point of this, it is not really a fetcher, and its fields could just be in the `realFetcher`

facchettos (changes requested):

pkg/portal/cluster/machines.go
```diff
@@ -5,11 +5,20 @@ package cluster
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/compute"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/features"
 )
 
+// Creating local vars for these functions in order to make them swappable to make them testable. The function definition will is altered in the tests written.
+var newResourceClientFunction = features.NewResourcesClient
+var newVirtualMachineClientFunction = compute.NewVirtualMachinesClient
+
```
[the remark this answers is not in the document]
anshulvermapatel: Since these are the directly exported functions from a package, there were not getting mocked with a function with desired return value hence I need to resort to this.

pkg/portal/cluster.go
anshulvermapatel: Yes, .. I have updated the code and will push it with other changes.

pkg/portal/cluster/machines.go
```diff
@@ -59,6 +68,55 @@ func (c *client) Machines(ctx context.Context) (*MachineListInformation, error)
 	return c.fetcher.Machines(ctx)
 }
 
+// VM Allocation Status Stuff
+type VMAllocationStatus map[string]string
```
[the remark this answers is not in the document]
anshulvermapatel: I thought of making it a type to make sure whoever reads my code knows what this map is for.

pkg/portal/cluster/machines.go
```diff
@@ -59,6 +68,55 @@ func (c *client) Machines(ctx context.Context) (*MachineListInformation, error)
...
+
+		vmName = *vm.Name
+		instanceViewStatuses := vm.InstanceView.Statuses
+		for _, status := range *instanceViewStatuses {
+			if strings.Contains(*status.Code, "PowerState") {
+				allocationStatus = *status.Code
+			}
+		}
+
+		vmAllocationStatus[vmName] = allocationStatus
```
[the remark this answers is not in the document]
anshulvermapatel: Whats wrong with continue here? Is this for performance optimization or readability optimization?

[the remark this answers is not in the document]
anshulvermapatel: PowerState is at least some state as it comes in the for of "PowerState/{somestate}" for example "PowerState/running" and I take only the status code which has "PowerState" in it. 
Since it is looping through the VMs .. there is no chance of a VMs powerstate getting overridden.

pkg/portal/security_test.go
anshulvermapatel: done

pkg/portal/portal.go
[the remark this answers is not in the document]
anshulvermapatel: No, the comment is not necessary.
I just used path prefix because it was used before. I am changing this for all the other routes as well

pkg/portal/info_test.go
anshulvermapatel: done

pkg/portal/cluster/machines.go
anshulvermapatel: this condition can also be used, however, I have seen the statuses, they have powerstate only for actual power status of the VM.

pkg/portal/cluster/fetcher.go
anshulvermapatel: So the realFetcher is something which has stuff used for fetching details about a cluster which is in cosmosDB. I wanted to open doors to fetch from Azure directly like VM status VMID etc and things required for that can land in this struct. If any other stuff in future requires to fetch anything from Azure which might require any other information for that, it can be added here.

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
facchettos: it breaks the flow, so it's more readability related. and also most `continue` are actually just a sign that you should split your code into smaller functions

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
anshulvermapatel: Okay .. but wherever I pass the interface as parameter to the `VMAllocationStatus` method from, there I will have to call these functions to get the interface object and I will have to anyways test that as well .. which will put me in the same situation ..

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
anshulvermapatel: Yeah okay .. I too now think that .. I will change it ..

pkg/portal/cluster/machines.go
anshulvermapatel: if the convention changes .. we will have to change the code anyway ..

pkg/portal/cluster/fetcher.go
anshulvermapatel: but then we will do it in a segregated manner which I think will help a new engineer to understand the code easily.

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
anshulvermapatel: Are you saying to put it like this -
```
func putAllocationStatusToMap(ctx context.Context, clusterRGName string, vmAllocationStatus map[string]string, res mgmtfeatures.GenericResourceExpanded, virtualMachineClient compute.VirtualMachinesClient, log *logrus.Entry) {
	var vmName, allocationStatus string
	vm, err := virtualMachineClient.Get(ctx, clusterRGName, *res.Name, mgmtcompute.InstanceView)
	if err != nil {
		log.Warn(err) // can happen when the ARM cache is lagging
		return
	}
	vmName = *vm.Name
	instanceViewStatuses := vm.InstanceView.Statuses
	for _, status := range *instanceViewStatuses {
		if strings.Contains(*status.Code, "PowerState") {
			allocationStatus = *status.Code
		}
	}
	vmAllocationStatus[vmName] = allocationStatus
}
```

pkg/portal/cluster/fetcher.go
facchettos: this does not help right now, it's not even a fetcher so it's confusing. I stand by my comment

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
facchettos: Having the interfaces in the struct is a standard pattern (passing it as a parameter is also possible but imo having these as a field is better here), and with what you have done it means that we can modify the functions at runtime and this is dangerous (package wide function variable is not a good idea)

pkg/portal/cluster/machines.go
facchettos: in any case, one is more correct than the other so I think we should go for the more precise one

anshulvermapatel: Okay .. I can do that ..

pkg/portal/cluster/fetcher.go
anshulvermapatel: so what design do you suggest here?

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
anshulvermapatel: I have added the above function and removed the continue line.

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
anshulvermapatel: in the tests .. I am resetting these variable to the original functions in the defer statement so that any other tests which might use those variables have the original function from the `features` package.

pkg/portal/cluster/machines.go
anshulvermapatel: done, resolving this conversation

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
anshulvermapatel: done, resolving

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
anshulvermapatel: Changed, resolving conversation.

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
facchettos: use an interface.

pkg/portal/cluster/fetcher.go
facchettos: I already suggested a way to do it

anshulvermapatel: to put the fields in the realfetcher itself .. I really think a segregation must be here to the required things to fetch from Azure(not from kube-apiserver).. maybe we can call it something else as you mentioned it is not a fetcher .. we can call it something else but I think the segregation must be there.

facchettos: actually it should be a completely separated client. we don't need any of the existing realfetcher underlying clients do we? (only log ) 
the real fetcher fetches stuff from kubernetes and it should keep doing just that. If we need to fetch from azure, we should create a new and separate client

facchettos (changes requested):
I think the azurefetcher should be separated from the realfetcher as it actually is something else entirely and we don't need to extend the existing realfetcher which only deals with the kube api with azure capabilities

pkg/portal/cluster/machines_test.go
```diff
@@ -97,3 +111,154 @@ func TestMachines(t *testing.T) {
...
+										s := new(string)
+										*s = "PowerState/running"
+										return s
+									}(),
+								},
+							},
+						},
+					},
+				}, nil)
+			},
```
facchettos, line 182: setting up the mock is better inside the test itself, not in the test data. if it proves difficult to generalize your test, that probably means that it should be in different test functions

pkg/portal/cluster/machines.go
```diff
@@ -59,7 +65,52 @@ func (c *client) Machines(ctx context.Context) (*MachineListInformation, error)
 	return c.fetcher.Machines(ctx)
 }
 
+func (c *client) VMAllocationStatus(ctx context.Context) (map[string]string, error) {
+	return c.fetcher.VMAllocationStatus(ctx)
+}
+
+func (f *realFetcher) VMAllocationStatus(ctx context.Context) (map[string]string, error) {
```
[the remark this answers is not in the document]
anshulvermapatel: we use it elsewhere, we use it in the `portal` package.

pkg/portal/cluster/fetcher.go
```diff
@@ -6,20 +6,32 @@ package cluster
...
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/features"
 	"github.com/Azure/ARO-RP/pkg/util/restconfig"
+	"github.com/Azure/ARO-RP/pkg/util/stringutils"
 )
 
+type ResourceFactory interface {
+	NewResourcesClient(environment *azureclient.AROEnvironment, subscriptionID string, authorizer autorest.Authorizer) features.ResourcesClient
+	NewVirtualMachinesClient(environment *azureclient.AROEnvironment, subscriptionID string, authorizer autorest.Authorizer) compute.VirtualMachinesClient
+}
+
```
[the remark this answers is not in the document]
anshulvermapatel: Okay, I have made the changes, will pushed it after making more changes as per the review.

pkg/portal/cluster/machines_test.go
anshulvermapatel: I should depend on the test case right .. the data we want and test our function according to that data. That should come in the test data only.

pkg/portal/cluster/machines.go
```diff
@@ -59,7 +65,52 @@ func (c *client) Machines(ctx context.Context) (*MachineListInformation, error)
...
+
+	vmName = *vm.Name
+	instanceViewStatuses := vm.InstanceView.Statuses
+	for _, status := range *instanceViewStatuses {
+		if strings.HasPrefix(*status.Code, "PowerState/") {
+			allocationStatus = *status.Code
+		}
+	}
+
+	vmAllocationStatus[vmName] = allocationStatus
```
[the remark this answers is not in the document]
facchettos: then remove that line from my suggestion and keep the rest of it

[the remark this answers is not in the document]
anshulvermapatel: ohh yaa, sorry, if we do not get any status from then it is going to be empty string, I did this to keep the UI clean as this is going to get seen in the UI as is.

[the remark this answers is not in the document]
anshulvermapatel: I just want to have all the VMNames as keys in the map.

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
facchettos: well one of these seems to be exported for no reason. Because it portal we apparently use the fetchclient interface, not the realfetcher itself.

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
facchettos: then change `"unknown"` to `""`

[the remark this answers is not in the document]
anshulvermapatel: No, all the values of the map will have empty string, I have declared the variable `allocationStatus` at the starting of the function, if there is value for the vm allocationStatus code, the var will have that value else it would be empty string only. So I think it is good as is.

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
anshulvermapatel: Yes, it is fetchClient, I think I made it capital because VM is a short form and wrote it in caps I think. I can make it in small letter.

pkg/portal/cluster/machines_test.go
anshulvermapatel: Sorry, I might have misunderstood you. 
So what I think is, it should depend on the test, support if the test is about a situation where the function will return an error in the first few lines itself, so we would not expect the later function/methods to get executed and we won't write <mockInterface>.EXPECT().<function>.whatever. 

If you meant something else than what I understood, please do elaborate a little.

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
facchettos: take a look again at my suggestion, it does exactly that. the `return` statement break the flow

pkg/portal/cluster/machines_test.go
facchettos: but you should do that setup in the test itself, if there is no expected call you should know it in the data, and set it up accordingly.

pkg/portal/cluster/machines.go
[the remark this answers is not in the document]
anshulvermapatel: Yes yes .. you are right .. I am changing it to empty String.

pkg/portal/cluster/machines_test.go
anshulvermapatel: I have seen this as standard structure in our codebase only .. check this - https://github.com/Azure/ARO-RP/blob/782894071cf91e8c1ca6e182417b851925f823cc/pkg/api/validate/dynamic/dynamic_test.go#L43-L110

facchettos: I know, jim and mj loved that , but it creates loads of code duplication, unreadable and overly complicated tests and we should stop using it

ellis-johnson (commented):
Changes look awesome! Just two things

Firstly it seems like the machines state isn't getting updated when you switch between different machines. It just stays on the first one you selected regardless of which one you select next. I tried this for the other InfoLists we use for Nodes and MachineSets and it seems to work fine so I think this is an isolated issue

Secondly it would be awesome if we can get an e2e test for the Machines tab from this

pkg/portal/cluster/machines_test.go
anshulvermapatel: I have updated this in the latest push, please do check and let me know.

ellis-johnson (changes requested):
Looks good! Just need to rebuild the portal and commit the files and it will be all good to go

pkg/portal/cluster/machines_test.go
```diff
@@ -97,3 +111,159 @@ func TestMachines(t *testing.T) {
...
+				if tt.mocks["env.Environment"] > 0 {
+					switch tt.name {
+					case "Successfully fetching VMs allocation status. Calling all the required methods.", "No VM resource found", "Empty FP Authorizer":
+						mockEnv.EXPECT().Environment().Return(&azureclient.AROEnvironment{
+							Environment: azure.Environment{
+								ResourceManagerEndpoint: "temp",
+							},
+						})
+					}
+				}
```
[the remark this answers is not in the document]
anshulvermapatel: Done please do check

pkg/portal/cluster/machines_test.go
```diff
@@ -97,3 +113,133 @@ func TestMachines(t *testing.T) {
...
+				env:                          mockEnv,
+				resourceClientFactory:        mcf,
+				virtualMachinesClientFactory: mcf,
+			}
+			azureClient := &azureClient{fetcher: azureSideFetcher, log: log}
+			vmAllocationStatuses, err := azureClient.VMAllocationStatus(ctx)
+			if !reflect.DeepEqual(vmAllocationStatuses, tt.wantOutput) {
+				t.Error("Expected output", tt.wantOutput, "Got", vmAllocationStatuses)
+			}
+			if err != nil && err /*.Error()*/ != tt.wantErr || err == nil && tt.wantErr != nil /*""*/ {
```
facchettos, line 240: left over comments

facchettos (dismissed):
something to fix in a follow up but otherwise LGTM (conditional to what ellis requested being implemented)

pkg/portal/cluster/machines_test.go
anshulvermapatel: removed these leftover comments.
And yes the bug which Ellis pointed out, I fixed it in this PR.

ellis-johnson (approved):
LGTM!

facchettos (approved):
please make sure to squash before merging (whoever merges)

PR #1609, opened 2021-07-13
Add limits to ARO operator

cadenmarchese:
### Which issue this PR addresses:

Adds limits to the operator deployment so that the kubelet can invoke OOMKiller in the event that it goes rogue.

Fixes https://github.com/Azure/ARO-RP/issues/1325 and ADO card #10247139

### Test plan for issue:

Followed the steps in https://github.com/Azure/ARO-RP/tree/master/pkg/operator#how-to-run-a-custom-operator-image to patch in a new operator deployment, and confirmed that the following resource block was added to the pod yaml:

    resources:
      limits:
        cpu: "1"
        memory: 1000Mi
      requests:
        cpu: 10m
        memory: 100Mi

### Is there any documentation that needs to be updated for this PR?

No

PR #2556, opened 2022-11-24
Add limits to ARO operator

rhamitarora:
The PR is created to add requests / limits to aro-master and aro-worker.
ADO Card : #[13399871](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/13399871) 

Picked a few random clusters in a couple different regions and looked at CPU/Mem usage of the openshift-azure-operator pods. During analysis I found the values are missing, hence added the required values.
### Test plan for issue:
oc project openshift-azure-operator
oc describe pod aro-operator-master-XXXXXXXXXXXXX
oc describe pod aro-operator-worker-XXXXXXXXXXXXX
oc -n openshift-azure-operator get all
oc describe deployment.apps/aro-operator-master
oc describe deployment.apps/aro-operator-worker
```
    resources:
          limits:
            cpu: 50m
            memory: 500Mi
          requests:
            cpu: 50m
            memory: 500Mi
```

No


pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
```diff
@@ -36,6 +36,13 @@ spec:
         ports:
         - containerPort: 8080
           name: http
+        resources:
+          limits:
```
cadenmarchese, line 40: The master operator deployment runs a lot of controllers (and it's expected that customer workloads are not typically running on master nodes), so I am of the opinion that we probably shouldn't impose any limits on the master deployment. Do we know how/if other OpenShift operators do requests and limits?

cadenmarchese (commented):
I'm not sure on limiting/throttling the operator CPU. Memory limits probably make sense especially on the worker deployment because we don't want to crowd out a customer workload.

I see it's been mentioned in the ADO card, but for visibility, we attempted this in https://github.com/Azure/ARO-RP/pull/1609 and it was reverted in https://github.com/Azure/ARO-RP/pull/1646. It was a release issue so we didn't make time to investigate, however we saw that limits were causing operators to restart / take a long time to start, not updating statuses, and failing e2e in certain regions.

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
facchettos: It should probably still have limits even if it is running on the master. just in case of a memory leak it avoids crashing the whole master node. having it that low might not be super optimal though.

cadenmarchese: I checked over the shared cluster today (specifically at openshift-machine-api operators/pods that run on masters) and see that there are no limits set. There are requests, though, which I agree we should have.

This is just my 0.02 though from past experience. If we need limits, we should just set them a lot higher.

jaitaiwan:
@rhamitarora I was looking at the ticket history on this. I'm wondering if we've run the CPU/Memory tests more recently. What I'm particularly interested in is what the startup cpu/memory is because in the past I've seen containers take up 2-3x the resources during startup as during the running period.

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
```diff
@@ -36,6 +36,13 @@ spec:
...
+            cpu: 50m
```
SudoBrendan, line 41: CPU limits are generally considered bad practice, since CPU is a compressible resource. Limits tend to harm way more than they help. Here's one of many articles explaining this further: https://medium.com/directeam/kubernetes-resources-under-the-hood-part-3-6ee7d6015965

pkg/operator/deploy/staticresources/worker/deployment.yaml.tmpl
```diff
@@ -33,6 +33,13 @@ spec:
         - name: "RP_MODE"
           value: "development"
         {{ end }}
+        resources:
+          limits:
+            cpu: 10m
```
SudoBrendan, line 38: Same as above. We shouldn't set CPU limits.

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
SudoBrendan: I'm +1 with Caden here. Requests are generally considered very healthy, and allow the underlying OS to appropriately trigger things like eviction when resources are high on a node. Limits cause strange behavior, like restarting containers and crashlooping when memory is high, and unnecessary throttling on CPU resources (which can lead to outages, like if your context will timeout 100% of the time if you don't get enough cycles).

The only real benefit we get to setting Limits is getting put in the QoS tier of Guaranteed, which I don't think is helpful when other (honestly, more critical) operators are NOT in this class as Caden has discovered (meaning, we'd end up evicting other, more critical components before our operator if we achieve Guaranteed...): https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed

SudoBrendan (changes requested):

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
facchettos: requests have nothing to do with the os, it is for the scheduler

PR #2558, opened 2022-11-24
Skips AKS tests for defined regions without AKS

petrkotas:

Disables AKS in a region without AKS. The list of regions is hard coded without an option to
configure. 

The reason for not configurable variant is to make the PR as small as possible and not introduce
changes in the pipeline and in the rp-config.

Fixes

### DIFF
--- a/test/e2e/aks.go
+++ b/test/e2e/aks.go
@@ -34,7 +34,28 @@ var _ = Describe("AKS cluster present", func() {
 		Expect(err).To(BeNil())
 	})
 
+	// TODO: remove this when all regions have the AKS
+	//       since this is going to happen in a weeks,
+	//       no need for external configuration option
+	regionsWithoutAKS := []string{
+		"australiacentral",
+		"australiacentral2",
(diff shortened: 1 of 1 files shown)


cblecker (approved):

test/e2e/aks.go
```diff
@@ -34,7 +34,28 @@ var _ = Describe("AKS cluster present", func() {
...
+		"switzerlandwest",
+		"uaecentral",
+		"usgovvirginia",
+	}
+
 	It("should get kubeconfig", func() {
+		By("region = " + _env.Location())
+		for _, region := range regionsWithoutAKS {
+			// uses the region information stored in core environment, which reads it from instance metadata.
+			if _env.Location() == region {
```
darthhexx, line 54: We need to use a `strings.EqualFold` check here instead of a direct comparison. Otherwise we will end up with [the same issue that was fixed in this PR](https://github.com/Azure/ARO-RP/pull/2521).

PR #2362, opened 2022-08-30
Add HIVEKUBECONFIGPATH variable to e2e

petrkotas:
### Which issue this PR addresses:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15372448/

### What this PR does / why we need it:

Adds environment variable to the E2E pipeline to read the kubeconfig from the `secrets` folder.
E2E does not have MSI deployed, which means this is the only way to read kubeconfig.

Signed-off-by: Petr Kotas <pkotas@redhat.com>

### Test plan for issue:
This is for the tests to be able to run.

PR #2559, opened 2022-11-24
Update docs for shared rp-dev-env installation

yehornaumenko:

None

We have started using AKS kubeconfig in some e2e tests to test AKS connectivity/Hive/etc. In some of our bash scripts, Makefiles, and docs files we use aks.kubeconfig as the name of AKS kubeconfig file.
This change actually updates two things:

- [Prepare shared environment docs](https://github.com/Azure/ARO-RP/blob/master/docs/prepare-a-shared-rp-development-environment.md) (aka E2E infra installation docs) with steps: update a kubeconfig on the SA, and install Hive on the new AKS
- The kubeconfig filename on E2E. For some reason, the e2e AKS kubeconfig is stored in the file called e2e-aks-kubeconfig. I suggest renaming this file to aks.kubeconfig to reach some naming consistency

### Test plan for issue:

Green E2E is enough

### DIFF
--- a/.pipelines/e2e.yml
+++ b/.pipelines/e2e.yml
@@ -42,7 +42,7 @@ jobs:
       set -o pipefail
 
       . secrets/env
-      export HIVE_KUBE_CONFIG_PATH_1="secrets/e2e-aks-kubeconfig_1"
+      export HIVE_KUBE_CONFIG_PATH_1="secrets/aks.kubeconfig"
 
       az account set -s $AZURE_SUBSCRIPTION_ID
 
--- a/docs/prepare-a-shared-rp-development-environment.md
+++ b/docs/prepare-a-shared-rp-development-environment.md
@@ -27,7 +27,7 @@ locations.
    Set SECRET_SA_ACCOUNT_NAME to the name of the storage account:
 
    ```bash
-   SECRET_SA_ACCOUNT_NAME=rharosecretsdev
+   SECRET_SA_ACCOUNT_NAME=e2earosecrets
    ```
 
 1. You will need an AAD object (this could be your AAD user, or an AAD group of
@@ -368,7 +368,7 @@ az ad app credential reset \
 
 5. The RP makes API calls to kubernetes cluster via a proxy VMSS agent. For the agent to get the updated certificates, this vm needs to be redeployed. Proxy VM is currently deployed by the `deploy_env_dev` function in `deploy-shared-env.sh`. It makes use of `env-development.json`
 
-6. Run `[rharosecretsdev|aroe2esecrets] make secrets-update` to upload it to your
+6. Run `[rharosecretsdev|e2earosecrets] make secrets-update` to upload it to your
 storage account so other people on your team can access it via `make secrets`
 
 # Environment file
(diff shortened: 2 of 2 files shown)


.pipelines/e2e.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       set -o pipefail
 
       . secrets/env
-      export HIVE_KUBE_CONFIG_PATH="secrets/e2e-aks-kubeconfig"
+      export HIVE_KUBE_CONFIG_PATH="secrets/aks.kubeconfig"
```
yehornaumenko, line 45: I guess we use `aks.kubeconfig` filename everywhere, but not in the storage account (e2e-aks-kubeconfig is used in the SA).
So I suggest changing the filename in the SA. Hence we should update the e2e config.

cblecker: Please provide more context, as this isn't just a docs update if it's editing our e2e pipeline

yehornaumenko: @cblecker thanks, I have updated the PR description

cblecker: @petrkotas Is this correct? This was introduced here from what I can tell: https://github.com/Azure/ARO-RP/pull/2362

and without updating the secrets bundle, this would break things, correct?

petrkotas: yes and at the same time @yehornaumenko is right, we definitely need to chose one name. I would be in favor of having it `aks.kubeconfig` because it is visible from the context of account logged in to which I talk to.

docs/prepare-a-shared-rp-development-environment.md
```diff
@@ -472,6 +472,16 @@ each of the bash functions below.
    when running the `deploy_env_dev_override` command, delete the `-pip` resource
    and re-run.
 
+1. Get the AKS kubeconfig and upload it to the storage account:
```
petrkotas, line 475: @yehornaumenko can you also update the documentation to state that these changes should be done with `SECRET_SA_ACCOUNT_NAME=e2earosecrets`? I cannot find such information present.

petrkotas (changes requested):
Thank @yehornaumenko for keeping the documentation in sync with the real world state. I have two comment about this docs. Otherwise it seems good to me.

docs/prepare-a-shared-rp-development-environment.md
yehornaumenko: Done here https://github.com/Azure/ARO-RP/pull/2559/files#diff-9b3780a23507353cbb9a0ef8435545f8c4c6d483c105190658d2b7e60a7d7a7eR30

.pipelines/e2e.yml
yehornaumenko: I have just updated the secrets bundle. So at this moment we have 2 kubeconfig files there – an old one (e2e-aks-kubeconfig) and a new one (aks.kubeconfig).

petrkotas: as per https://github.com/Azure/ARO-RP/pull/2568 this variable needs to end with `_1` otherwise it will fail in e2e, because we need to select shard `1`

petrkotas (changes requested):
one nit and I believe we are done

.pipelines/e2e.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       set -o pipefail
 
       . secrets/env
-      export HIVE_KUBE_CONFIG_PATH_1="secrets/e2e-aks-kubeconfig_1"
+      export HIVE_KUBE_CONFIG_PATH_1="secrets/aks.kubeconfig"
```
cadenmarchese, line 45: Why `HIVE_KUBE_CONFIG_PATH_1` instead of just `HIVE_KUBE_CONFIG_PATH`?

yehornaumenko: @cadenmarchese This changed was made by @petrkotas, you can find full explanation [here](https://github.com/Azure/ARO-RP/pull/2568#discussion_r1037488551) but in short we need it because of hive shards mechanism

cadenmarchese (approved):

PR #2561, opened 2022-11-25
fixed race condition in frontend

facchettos:
fixes a race condition introduced by https://github.com/Azure/ARO-RP/pull/2275 . This race condition can lead to adminActions overriding each others' parameters and lead to wrong VMs in wrong subscription to be redeployed for example

This is because the admin action is shared and unique (accessed by the frontend pointer). In case of concurrent use, we have no guarantee of which admin action interface we are using and thus trigger geneva actions with the wrong parameters. Catastrophic bug isn't really likely because vms will have a somewhat unique names across subs but the geneva action will fail when that rc happens. 

This just removes the adminaction field from the frontend struct so we are forced to use local variables instead and can't have that race condition anymore

Fixes

### DIFF
--- a/pkg/frontend/admin_vm_actions.go
+++ b/pkg/frontend/admin_vm_actions.go
@@ -11,43 +11,34 @@ import (
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
+	"github.com/Azure/ARO-RP/pkg/frontend/adminactions"
 )
 
-type azVmActionsWrapper struct {
-	vmName string
-	doc    *api.OpenShiftClusterDocument
-}
-
-func (f *frontend) newAzureActionsWrapper(log *logrus.Entry, ctx context.Context, vmName, resourceID string, vars map[string]string) (azVmActionsWrapper, error) {
-	err := validateAdminVMName(vmName)
+func (f *frontend) prepareAdminActions(log *logrus.Entry, ctx context.Context, vmName, resourceID string, vars map[string]string) (azureActions adminactions.AzureActions, doc *api.OpenShiftClusterDocument, err error) {
+	err = validateAdminVMName(vmName)
 	if err != nil {
-		return azVmActionsWrapper{}, err
-	}
-	if err != nil {
-		return azVmActionsWrapper{}, err
+		return nil, nil, err
 	}
 
-	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	doc, err = f.dbOpenShiftClusters.Get(ctx, resourceID)
 	switch {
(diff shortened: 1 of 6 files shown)


darthhexx:
@facchettos could you please detail the race condition scenario that you are fixing in this PR?

facchettos:
@darthhexx imagine you have 2 concurrent calls to one of the functions that I modified (let say _postAdminOpenShiftClusterStopVM). What value of the `frontend.adminaction` field would we end up using?

pkg/frontend/admin_vm_actions.go
```diff
@@ -11,43 +11,36 @@ import (
...
-		return azVmActionsWrapper{}, err
+		return
 	}
 
-	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	doc, err = f.dbOpenShiftClusters.Get(ctx, resourceID)
 	switch {
 	case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
-		return azVmActionsWrapper{}, api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, "",
+		return nil, doc, api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, "",
```
s-amann, line 29: nit: if the doc retrieval errors out with 'not found' why return the doc here? Can this be `return nil, nil, api.NewCloudError(...)`?

s-amann (approved):

pkg/frontend/admin_vm_actions.go
facchettos: you're right, I actually thought it could be useful to have the doc even if it was not found, but it's `nil` anyways so it's better to be explicit

s-amann (dismissed):

pkg/frontend/admin_vm_actions.go
```diff
@@ -11,43 +11,33 @@ import (
...
-	vmName string
-	doc    *api.OpenShiftClusterDocument
-}
-
-func (f *frontend) newAzureActionsWrapper(log *logrus.Entry, ctx context.Context, vmName, resourceID string, vars map[string]string) (azVmActionsWrapper, error) {
-	err := validateAdminVMName(vmName)
-	if err != nil {
-		return azVmActionsWrapper{}, err
-	}
+func (f *frontend) prepareAdminActions(log *logrus.Entry, ctx context.Context, vmName, resourceID string, vars map[string]string) (azureActions adminactions.AzureActions, doc *api.OpenShiftClusterDocument, err error) {
```
cblecker, line 17: I'd prefer we didn't mix return types here.. Can we please swap this back to not use naked returns?

cblecker (dismissed):

s-amann (approved):

cblecker (approved):

PR #2562, opened 2022-11-28
Mirror all app-sre tags since a given time, rather than specific tags

hawkowl:

Fixes https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/16407158

Mirrors app-sre images since a given time automatically. This ensures that we have the versions in ACR to be tested or to provide bugfixes to customers without having to roll an RP release.

### Test plan for issue:
Needs to be run manually, 

N/A

### DIFF
--- a/cmd/aro/mirror.go
+++ b/cmd/aro/mirror.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/containers/image/v5/types"
 	"github.com/sirupsen/logrus"
@@ -82,6 +83,19 @@ func mirror(ctx context.Context, log *logrus.Entry) error {
 		if err != nil {
 			return err
 		}
+	}
+
+	var mirrorSince time.Time
+	mirrorSinceArg := os.Getenv("MIRROR_SINCE")
+
+	if mirrorSinceArg != "" {
+		mirrorSince, err = time.Parse("2006-01-02", mirrorSinceArg)
+		if err != nil {
+			log.Errorf("unable to parse MIRROR_SINCE, falling back: %s\n", err)
+			mirrorSince = time.UnixMilli(1664586000000)
+		}
+	} else {
+		mirrorSince = time.UnixMilli(1664586000000)
 	}
 
 	var releases []pkgmirror.Node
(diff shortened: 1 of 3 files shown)


cmd/aro/mirror.go
```diff
@@ -162,9 +176,6 @@ func mirror(ctx context.Context, log *logrus.Entry) error {
 		"registry.access.redhat.com/ubi7/go-toolset:1.16.12",
 		"registry.access.redhat.com/ubi8/go-toolset:1.17.7",
 		"mcr.microsoft.com/azure-cli:latest",
-
-		"quay.io/app-sre/managed-upgrade-operator:v0.1.856-eebbe07",
-		"quay.io/app-sre/hive:fec14dc",
```
s-amann, line 167: we use the hive commit as the tag in our [deployment of hive pipeline](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines?path=/ev2/generator/templates/hive-deployment/hive-deployment.tpl&version=GBmaster&line=277&lineEnd=278&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents). Will we need to update how we are deploying hive since we are changing how we are mirroring the image? From what I understand the answer should be 'no' as this new mirroring method should grab all tags from a date, and the tag should still remain as first 7 of the commit.

s-amann (commented):
lgtm, just a question and minor nitpick

darthhexx:
I have some concerns with this approach:
1. We have to update the image tag we are deploying in the pipelines and [the hive config for the dev environment](https://github.com/Azure/ARO-RP/blob/f1f1a9b19565c9fec8bc662d4d9468dec1e5bb96/hack/hive-generate-config.sh#L4), so I don't see why we can just update it in the mirror code, as we have been, in the same PR? I am not sure about MUO though.
2. We are trying to speed up the mirror pipeline and merging this functionality would mean that many more unused tags are going to be mirrored in the process, which is going to slow it down, even though we still have to update the tag anyway.
3. We also don't have to roll a new RP release to deploy the new image to the ACRs, we only have to merge to master and the next pipeline run will include the new image(s).

cblecker:
Due to the concerns raised, removing `next-release`

PR #2564, opened 2022-11-29
Quota validation in frontend

facchettos:
as discussed in the ado ticket, this moves the quota validation to the frontend and thus avoids having failed clusters when the quotas prevent cluster creation. 
https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15811405/

Fixes

### Test plan for issue:
added unit tests


pkg/frontend/quota_validation.go
```diff
@@ -0,0 +1,178 @@
...
+
+type QuotaValidator interface {
+	ValidateQuota(ctx context.Context, azEnv *azureclient.AROEnvironment, environment env.Interface, subscriptionID, tenantID string, oc *api.OpenShiftCluster) error
+	//ValidateNewCluster(ctx context.Context, envValue env.Interface, subscription *api.SubscriptionDocument, cluster *api.OpenShiftCluster, staticValidator api.OpenShiftClusterStaticValidator, ext interface{}, path string) error
+}
+
+type quotaValidator struct{}
+
+func addRequiredResources(requiredResources map[string]int, vmSize api.VMSize, count int) error {
+	vmTypesMap := map[api.VMSize]struct {
```
cadenmarchese, line 42: We keep a similar list in pkg/api/validate/vm.go, would it be possible to combine them so we only need to update one list when we add sizes?

pkg/api/error.go
```diff
@@ -99,6 +99,7 @@ const (
 	CloudErrorCodeQuotaExceeded                      = "QuotaExceeded"
 	CloudErrorCodeResourceProviderNotRegistered      = "ResourceProviderNotRegistered"
 	CloudErrorCodeCannotDeleteLoadBalancerByID       = "CannotDeleteLoadBalancerWithPrivateLinkService"
+	CloudErrorUnsupportedSKU                         = "InvalidVMSKU"
```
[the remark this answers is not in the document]
facchettos: yeah I missed it. will remove it

pkg/frontend/quota_validation.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// ValidateQuota checks usage quotas vs. resources required by cluster before cluster
+// creation
+// It is a method on struct so we can make use of interfaces.
+func (q quotaValidator) ValidateQuota(ctx context.Context, azEnv *azureclient.AROEnvironment, environment env.Interface, subscriptionID, tenantID string, oc *api.OpenShiftCluster) error {
+	fpAuthorizer, err := environment.FPAuthorizer(tenantID, environment.Environment().ResourceManagerEndpoint)
+	if err != nil {
+		return err
+	}
```
m1kola, line 45: I think this is the place where we might hit AAD propagation delay.

https://github.com/Azure/ARO-RP/blob/07a08c6b59c47694b80d5168226977d4bced1acd/pkg/util/refreshable/refreshable.go#L27-L55

cadenmarchese (approved):

PR #2568, opened 2022-12-01
Fix AKS shard number and add equalfold

petrkotas:
AKS shards start with 1
EqualFold is required because not all regions use lowercase

Fixes

### DIFF
--- a/.pipelines/e2e.yml
+++ b/.pipelines/e2e.yml
@@ -42,7 +42,7 @@ jobs:
       set -o pipefail
 
       . secrets/env
-      export HIVE_KUBE_CONFIG_PATH="secrets/e2e-aks-kubeconfig"
+      export HIVE_KUBE_CONFIG_PATH_1="secrets/e2e-aks-kubeconfig_1"
 
       az account set -s $AZURE_SUBSCRIPTION_ID
 
--- a/test/e2e/aks.go
+++ b/test/e2e/aks.go
@@ -8,6 +8,7 @@ package e2e
 
 import (
 	"context"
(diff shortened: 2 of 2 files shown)


.pipelines/e2e.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       set -o pipefail
 
       . secrets/env
-      export HIVE_KUBE_CONFIG_PATH="secrets/e2e-aks-kubeconfig"
+      export HIVE_KUBE_CONFIG_PATH_1="secrets/e2e-aks-kubeconfig_1"
```
s-amann, line 45: why was this change mage? the `liveconfig/manager.go` is expecting the env variable `HIVE_KUBE_CONFIG_PATH`

petrkotas: it is not, the issue is it expects `%s_%d` in case of shards according to https://github.com/Azure/ARO-RP/blob/master/pkg/util/liveconfig/dev.go#L19

Issue is that all shards are starting from `001` meaning the ode need to use shard N=1, this works for INT as it asks AKS. In e2e it does not work, because we are using kubeconfig. The link I shared is expecting different variable names for multiple shards.

This is an example of solving problems too early when problems are not known yet. 

Therefore this variable name change.

s-amann: ah, thank you for the detailed explanation and catching my logical error in understanding e2e, vs int, vs prod.

s-amann (approved):

cblecker (approved):

PR #2570, opened 2022-12-02
Clouderror for parseunverified

carvalhe:
Fixes
(https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/16455663)

QoS for cluster creation is showing the top internal server error as "invalid number of segments." This was fixed in a previous PR a few months ago, but there are some more areas this change is needed. This PR adds the CloudError to areas ParseUnverified is called so when the error occurs, it will mark it as UserError instead of InternalServerError for QoS purposes. 
Similar PR in the past: https://github.com/Azure/ARO-RP/pull/2267

### Test plan for issue:
error/log based changes

### DIFF
--- a/pkg/api/validate/dynamic/serviceprincipal.go
+++ b/pkg/api/validate/dynamic/serviceprincipal.go
@@ -25,6 +25,7 @@ func (dv *dynamic) ValidateServicePrincipal(ctx context.Context, clientID, clien
 	c := &azureclaim.AzureClaim{}
 	_, _, err = p.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
+		dv.log.Error(err)
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", "The provided service principal generated an invalid token.")
 	}
 
--- a/pkg/api/validate/dynamic/serviceprincipal.go
+++ b/pkg/api/validate/dynamic/serviceprincipal.go
@@ -25,7 +25,7 @@ func (dv *dynamic) ValidateServicePrincipal(ctx context.Context, clientID, clien
 	c := &azureclaim.AzureClaim{}
 	_, _, err = p.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
-		return err
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", "The provided service principal generated an invalid token.")
 	}
 
 	for _, role := range c.Roles {
--- a/pkg/util/clusterauthorizer/authorizer.go
+++ b/pkg/util/clusterauthorizer/authorizer.go
@@ -72,6 +72,7 @@ func (a *azRefreshableAuthorizer) NewRefreshableAuthorizerToken(ctx context.Cont
 	c := &azureclaim.AzureClaim{}
 	_, _, err = p.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
+		a.log.Error(err)
 		return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", "The provided service principal generated an invalid token.")
 	}
 
--- a/pkg/util/clusterauthorizer/authorizer.go
+++ b/pkg/util/clusterauthorizer/authorizer.go
@@ -6,12 +6,14 @@ package clusterauthorizer
 import (
 	"context"
 	"fmt"
+	"net/http"
 
 	"github.com/form3tech-oss/jwt-go"
 	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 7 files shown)


pkg/api/validate/dynamic/serviceprincipal.go
```diff
@@ -25,7 +25,7 @@ func (dv *dynamic) ValidateServicePrincipal(ctx context.Context, clientID, clien
 	c := &azureclaim.AzureClaim{}
 	_, _, err = p.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
-		return err
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", "The provided service principal generated an invalid token.")
```
SudoBrendan, line 28: Unfortunately, the docs for ParseUnverified don't specify what errors are returned. Do we want to at least log them before masking/returning them like this?

SudoBrendan (commented):
I think we might benefit from at least logging the original error in all these cases so we don't mask failures we are unaware of or might actually be our fault.

pkg/util/instancemetadata/prod.go
```diff
@@ -71,7 +72,7 @@ func (p *prod) populateTenantIDFromMSI(ctx context.Context) error {
 	c := &azureclaim.AzureClaim{}
 	_, _, err = parser.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
-		return err
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", "The provided service principal generated an invalid token.")
```
darthhexx, line 75: It doesn't make sense to switch this error into an CloudError because the code path here is during RP instancemeta initialisation and there is no API request involved.

darthhexx (changes requested):

pkg/api/validate/dynamic/serviceprincipal.go
carvalhe: Agreed, I changed this and the similar files to now produce a log with the error before returning the cloudError type

pkg/util/instancemetadata/prod.go
carvalhe: Reverted the changes, thank you.

pkg/api/validate/dynamic/serviceprincipal.go
darthhexx: `dynamic.ValidateServicePrincipal` is not only called from the [API SP validation](https://github.com/Azure/ARO-RP/blob/7447d1b01df6974a8069b265f85fa59d7b15a0c5/pkg/api/validate/openshiftcluster_validatedynamic.go#L90) but also from the [ARO operator when performing SP validation](https://github.com/Azure/ARO-RP/blob/7447d1b01df6974a8069b265f85fa59d7b15a0c5/pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go#L60), so this change would mean that [the error message in OpenShift would be set to the CloudError message](https://github.com/Azure/ARO-RP/blob/7447d1b01df6974a8069b265f85fa59d7b15a0c5/pkg/operator/controllers/checkers/serviceprincipalchecker/controller.go#L104) instead of the actual root cause error.

I think this needs to be tested, so that we can see the difference between the two and decide whether we need to add conditional logic in the ARO Operator to parse the CloudError. Potentially only displaying the message segment after the last `:`, i.e. "The provided service principal generated an invalid token."

pkg/util/clusterauthorizer/authorizer.go
```diff
@@ -70,7 +72,8 @@ func (a *azRefreshableAuthorizer) NewRefreshableAuthorizerToken(ctx context.Cont
 	c := &azureclaim.AzureClaim{}
 	_, _, err = p.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
-		return nil, err
+		a.log.Error(err)
+		return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", "The provided service principal generated an invalid token.")
```
darthhexx, line 76: `azRefreshableAuthorizer.NewRefreshableAuthorizerToken`  is currently only used in the ARO Operator controllers:

[Preview Feature Controller](https://github.com/Azure/ARO-RP/blob/7447d1b01df6974a8069b265f85fa59d7b15a0c5/pkg/operator/controllers/previewfeature/previewfeature_controller.go#L88)
[Storage Account Controller](https://github.com/Azure/ARO-RP/blob/7447d1b01df6974a8069b265f85fa59d7b15a0c5/pkg/operator/controllers/storageaccounts/storageaccount_controller.go#L104)
[Subnets Controller](https://github.com/Azure/ARO-RP/blob/7447d1b01df6974a8069b265f85fa59d7b15a0c5/pkg/operator/controllers/subnets/subnets_controller.go#L108)

I don't think these will be called from customer facing API requests, so we shouldn't change this to be a CloudError, but we could potentially just return a cleaner "The provided service principal generated an invalid token" message?

darthhexx (changes requested):
Did further digging into the ramifications of each change and have some recommendations.

PR #2573, opened 2022-12-02
Disable AKS test as it is not working every where

petrkotas:

Fixes

Makes AKS tests as WIP with use of `Pending` because we cannot run them properly in INT.
We still need to fix the INT and than enable AKS test again.


cadenmarchese (approved):
I think in the future we should continue to use `XDescribe`, `XIt`, etc.. to disable tests to keep things consistent, but looks like this method will work for the release

PR #2574, opened 2022-12-02
Require "official" builds have a valid tag.

ventifus:
Git has two different kinds of tags: annotated and light-weight. Our release process only supports annotated tags (`git tag -a`). If the tag is a light-weight tag then the build process will ignore it and use the git revision instead of the tag.

This change enforces the use of an annotated tag for "official" builds. A new pipeline variable `is_official_release` is passed down to `Dockerfile.aro-multistage` which, if true, runs a new `make aro-release` target. This Makefile target bails out if the tag is not specified.

Similar logic is added to the `Build_ARO` onebranch pipeline which doesn't use the Makefile.

Official release with valid tag:
```bash
$ make aro-release
go generate ./...
bindata.go
bindata.go
bindata.go
bindata.go
go build -tags aro,containers_image_openpgp,codec.safe -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=v20221114.01" ./cmd/aro
```

Official release without valid tag:
```bash
$ make aro-release
go generate ./...
bindata.go
bindata.go
bindata.go
fatal: no tag exactly matches '9daa6b845741b0f2c59300ace65ce0ac8d915cc6'
Ensure there is an annotated tag (git tag -a) for git commit 9daa6b8
make: *** [Makefile:43: aro-release] Error 1
```

### DIFF
--- a/.pipelines/onebranch/templates/template-buildrp-buildaro.yml
+++ b/.pipelines/onebranch/templates/template-buildrp-buildaro.yml
@@ -9,6 +9,12 @@ steps:
       export COMMIT=$(git rev-parse --short=7 HEAD)$([[ $(git status --porcelain) = "" ]] || echo -dirty)
       if [ -z "$TAG" ];
       then
+        if [ -z "$is_official_release"]
+        then
+          git describe --exact-match
+          echo "Ensure there is an annotated tag (git tag -a) for git commit $(COMMIT)"
+          exit 1
+        fi
         export VERSION=${COMMIT}
       else
         export VERSION=${TAG}
--- a/Dockerfile.aro-multistage
+++ b/Dockerfile.aro-multistage
@@ -12,7 +12,7 @@ WORKDIR ${GOPATH}/src/github.com/Azure/ARO-RP
 USER root
 RUN yum update -y
 COPY . ${GOPATH}/src/github.com/Azure/ARO-RP/
-RUN if [ "${IS_OFFICIAL_RELEASE}" == "true" ]; then make aro-release; else make aro; fi && make e2e.test
+RUN make aro RELEASE=${IS_OFFICIAL_RELEASE} && make e2e.test
 
 FROM ${REGISTRY}/ubi8/ubi-minimal
 RUN microdnf update && microdnf clean all
--- a/Dockerfile.aro-multistage
+++ b/Dockerfile.aro-multistage
@@ -12,7 +12,7 @@ WORKDIR ${GOPATH}/src/github.com/Azure/ARO-RP
 USER root
 RUN yum update -y
 COPY . ${GOPATH}/src/github.com/Azure/ARO-RP/
-RUN make aro && make e2e.test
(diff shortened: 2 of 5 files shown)


Dockerfile.aro-multistage
```diff
@@ -12,7 +12,7 @@ WORKDIR ${GOPATH}/src/github.com/Azure/ARO-RP
 USER root
 RUN yum update -y
 COPY . ${GOPATH}/src/github.com/Azure/ARO-RP/
-RUN make aro && make e2e.test
+RUN if [ "${IS_OFFICIAL_RELEASE}" == "true" ]; then make aro-release; else make aro; fi && make e2e.test
```
cadenmarchese, line 15:
```suggestion
RUN make aro RELEASE=${IS_OFFICIAL_RELEASE}; fi && make e2e.test
```

Makefile
```diff
@@ -39,6 +39,10 @@ build-all:
 aro: generate
 	go build -tags aro,containers_image_openpgp,codec.safe -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=$(VERSION)" ./cmd/aro
```
cadenmarchese, line 40: Then, here we could have an `if` statement in `make aro` that runs the tag check if `RELEASE`/IS_OFFICIAL_RELEASE is true. Otherwise it uses `VERSION`

cadenmarchese (commented):
Thanks for the fix! This issue got us in the last release and the one before and should be addressed.

I think it might be worth trying to give the existing `make aro` makefile target this functionality rather than create a new makefile target, if possible. I left a potential idea.

Makefile
ventifus: That makes more sense, I like it. I'll give it a shot and update this PR.

cadenmarchese (approved):

.pipelines/onebranch/templates/template-buildrp-buildaro.yml
```diff
@@ -9,6 +9,12 @@ steps:
       export COMMIT=$(git rev-parse --short=7 HEAD)$([[ $(git status --porcelain) = "" ]] || echo -dirty)
       if [ -z "$TAG" ];
       then
+        if [ -z "$is_official_release"]
+        then
+          git describe --exact-match
+          echo "Ensure there is an annotated tag (git tag -a) for git commit $(COMMIT)"
+          exit 1
```
cblecker, line 16: I'm not sure if I'm missing something, but doesn't this make if `is_official_release == true` then it will always `exit 1`?

cblecker (changes requested):

.pipelines/onebranch/templates/template-buildrp-buildaro.yml
ventifus: It's a bit of action-at-a-distance: you can only get here (past the test on line 10) if $TAG didn't get set to anything up on line 2, because git failed to resolve an (annotated) tag for the current revision.

PR #2577, opened 2022-12-05
Update deploy full rp service in dev documentation

s-fairchild:
Update documentation to inform readers that `SSH_PUBLIC_KEY` must be set if `~/.ssh/id_rsa.pub` doesn't exist. Update `gendevconfig.go` to log a warning if `SSH_PUBLIC_KEY` is unset

No existing issue

Fixes

If you are using `ed25519` encryption keys the deploy er fails without `SSH_PUBLIC_KEY` being set.
This is to update the deployment guide and log a warning if it's unset.

Document updated in PR

### DIFF
--- a/docs/deploy-full-rp-service-in-dev.md
+++ b/docs/deploy-full-rp-service-in-dev.md
@@ -17,7 +17,7 @@
     ```
 
 1. Create a full environment file, which overrides some default `./env` options when sourced
-    * if your ssh keypair is using `ed25519` or your key isn't named `~/.ssh/id_rsa.pub`, add `export SSH_PUBLIC_KEY=~/.ssh/id_ed25519.pub`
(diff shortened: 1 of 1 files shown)


docs/deploy-full-rp-service-in-dev.md
```diff
@@ -17,6 +17,7 @@
     ```
 
 1. Create a full environment file, which overrides some default `./env` options when sourced
+    * if your ssh keypair is using `ed25519` or your key isn't named `~/.ssh/id_rsa.pub`, add `export SSH_PUBLIC_KEY=~/.ssh/id_ed25519.pub`
```
cadenmarchese, line 20: Might make more sense:
```suggestion
    * if using a public key separate from `~/.ssh/id_rsa.pub`, source it with `export SSH_PUBLIC_KEY=~/.ssh/id_separate.pub`
```
Also, not sure if we need the docs change if we add your !found function. Would like others to weigh in.

SudoBrendan (approved):
LGTM

cadenmarchese (approved):

PR #2578, opened 2022-12-05
Improves internet checker test coverage

m1kola:

Improves internet checker test coverage.

I'm looking into splitting internet checker from other checks and noticed that test coverage can be improved: we were not checking what cnoditions controller was actually setting after the checks. Also the test now tets both master and worker roles.

### Test plan for issue:

Run tests.

No, just test updates.

### DIFF
--- a/pkg/operator/controllers/checker/internetchecker_test.go
+++ b/pkg/operator/controllers/checker/internetchecker_test.go
@@ -14,6 +14,12 @@ import (
 	"testing"
 	"time"
 
+	operatorv1 "github.com/openshift/api/operator/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/Azure/ARO-RP/pkg/operator"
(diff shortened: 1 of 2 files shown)


pkg/operator/controllers/checker/internetchecker_test.go
```diff
@@ -68,38 +68,95 @@ var (
...
+					var condition operatorv1.OperatorCondition
+					for _, condition = range instance.Status.Conditions {
+						if condition.Type == roleToConditionTypeMap[testRole] {
+							break
+						}
+					}
+
+					if condition.Status != test.wantCondition {
+						t.Errorf(string(condition.Status))
+					}
```
m1kola, line 158: I would like to separate the code which makes actual http calls to the list of URLs from updating status of the operator.

But in this PR I'm just adding coverage for the exiting code with minimal changes (just to make it testable).

s-amann: This sounds like a good maintenance task when we get the opportunity next year

s-amann (approved):

petrkotas (approved):
Good one.

Issue #2110, opened 2022-05-04
CLI --pull-secret option and error does not specify input format


cadenmarchese:
Attempting to install with `az aro create --pull-secret ~/path/to/pull-secret.txt` fails with the following:
```
Invalid --pull-secret.
```
The CLI only accepts the JSON string and not the file format that you get when you download a pull secret from Red Hat. We should either specify the requirement in the error message, or configure the CLI to accept/parse the .txt.

This is also mentioned in https://github.com/MicrosoftDocs/azure-docs/issues/53604.

cadenmarchese:
@ross-bryan told me that `az aro create --pull-secret @~/path/to/pull-secret.txt` would have worked. IMO, we can still clarify this better (in the man page or in the error message) for those who wish to pass in a pull secret.

PR #2581, opened 2022-12-06
Update az aro extension to support specifying --pull-secret file without @ prefix

s-fairchild:
Update az aro extension to handle reading the pull-secret file directly without having to prefix the path with @.
The pull-secret file can still be specified with or without @ as  `--pull-secret @pull-secret.txt`, or `--pull-secret pull-secret.txt`.

https://github.com/Azure/ARO-RP/issues/2110

Fixes

The pull secret file has to be prefixed with the @ symbol rather than the file path directly. Also brought up here: https://github.com/MicrosoftDocs/azure-docs/issues/53604

### Test plan for issue:
Tested locally with `az aro create -g $RESOURCEGROUP --name steven-cluster-2 --vnet aro-vnet --master-subnet master-subnet --worker-subnet worker-subnet --pull-secret ~/Downloads/pull-secret`

Unit tests pass, filesystem mocking will need to be done to test reading a file in addition to json strings.

Azure docs may want to be updated, although it will work with the existing documentation.

### DIFF
--- a/python/az/aro/azext_aro/_validators.py
+++ b/python/az/aro/azext_aro/_validators.py
@@ -5,6 +5,7 @@ import ipaddress
 import json
 import re
 import uuid
+from os.path import exists
 
 from azure.cli.core.commands.client_factory import get_mgmt_service_client
 from azure.cli.core.commands.client_factory import get_subscription_id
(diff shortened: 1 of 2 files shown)


cadenmarchese (approved):

s-amann (approved):
lgtm, nice work for a quick win

SudoBrendan (approved):
This LGTM, but if we are changing how customers interact with the CLI without backwards compatibility (unclear to me if this is happening - do both now work - with or without `@`?), we might need to talk to Jerome about getting a customer notification sent out before merging this upstream.

s-fairchild:
@SudoBrendan JSON can be passed as a string or file with this change, so using `--pull-secret @pull_secret.txt` works the same as `--pull-secret pull_secret.txt`.
We may want to ask them to update their documentation to show that `@` isn't required when referencing a file.

PR #2582, opened 2022-12-06
Add a new rp-nsg NSG rule portal_in_proxy, but only in fairfax

ventifus:
Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/16263929/
Issue uncovered in https://coreos.slack.com/archives/CCV9YF9PD/p1668715362789389?thread_ts=1668018543.974949&cid=CCV9YF9PD

This PR allows for a new RP-Config variable rpNsgPortalSourceAddressPrefixes. If present, an additional portal_in NSG rule will be added, allowing those source IPs to connect to the proxy.

In fairfax, we will need to be able to access the admin portal via escort sessions which will egress via a set of proxy IPs. In normal prod environments, SAWs can connect to the portal via rules added by SimplySecure. Those don't exist in fairfax so we must allow in whichever IPs we are connecting from. Proxy IPs are documented here: https://msazure.visualstudio.com/AzureWiki/_wiki/wikis/AzureWiki.wiki/82109/MS-Proxy

Once this PR looks good, I will submit another PR to add the proxy IPs to RP-Config for fairfax.


pkg/deploy/generator/resources_rp.go
```diff
@@ -37,98 +37,132 @@ func (g *generator) rpManagedIdentity() *arm.Resource {
...
+					Priority:                 to.Int32Ptr(130),
 					Direction:                mgmtnetwork.SecurityRuleDirectionInbound,
 				},
-				Name: to.StringPtr("dbtoken_in_gateway_8445"),
+				Name: to.StringPtr("rp_in_geneva"),
 			},
-			mgmtnetwork.SecurityRule{
+		}
+
+		if !g.production {
```
cadenmarchese, line 85: Where is this set?

cadenmarchese (commented):
The change and your rework make sense to me for the most part - just a few thoughts / questions initially. Thanks!

pkg/deploy/assets/rp-production-predeploy.json
```diff
@@ -157,13 +161,102 @@
...
+                    },
+                    {
+                        "properties": {
+                            "protocol": "Tcp",
+                            "sourcePortRange": "*",
+                            "destinationPortRange": "444",
+                            "sourceAddressPrefixes": "[parameters('rpNsgPortalSourceAddressPrefixes')]",
+                            "destinationAddressPrefix": "*",
+                            "access": "Allow",
+                            "priority": 130,
```
darthhexx, line 173: This priority will clash with the GenevaActions rule. It needs to be lower than 145 (the DENY rule) and not clash with another. 142 would be a good choice.

pkg/deploy/generator/resources_rp.go
```diff
@@ -37,98 +37,132 @@ func (g *generator) rpManagedIdentity() *arm.Resource {
...
-				SourceAddressPrefix:      to.StringPtr("GenevaActions"),
-				DestinationAddressPrefix: to.StringPtr("*"),
-				Access:                   mgmtnetwork.SecurityRuleAccessAllow,
-				Priority:                 to.Int32Ptr(130),
-				Direction:                mgmtnetwork.SecurityRuleDirectionInbound,
-			},
-			Name: to.StringPtr("rp_in_geneva"),
-		},
-	}
+func (g *generator) rpSecurityGroup() []*arm.Resource {
```
darthhexx, line 40: This function has been changed to be too complicated to achieve the conditional adding of a single security rule. Duplicate rp-nsg blocks in the ARM template is also not optimal.

Please can you rather add a new function `(func (g *generator) rpSecurityGroupFairFaxAdminPortal() *arm.Resource` that effectively adds this ARM resource:
```
{
    "type": "Microsoft.Network/networkSecurityGroups/securityRules",
    "apiVersion": "2022-07-01",
    "name": "rp-nsg/portal_in",
    "condition": "[not(empty(parameters('rpNsgPortalSourceAddressPrefixes')))]",
    "properties": {
        "protocol": "Tcp",
        "sourcePortRange": "*",
        "destinationPortRange": "444",
        "sourceAddressPrefixes": "[parameters('rpNsgPortalSourceAddressPrefixes')]",
        "destinationAddressPrefix": "*",
        "access": "Allow",
        "priority": 142,
        "direction": "Inbound"
    }
}
```` 
That way we are cleanly adding the security rule when the RP-Config array is set with no changes to the existing NSG deployment, i.e. in FairFax.

File: pkg/deploy/generator/templates_rp.go [hunk 4]
```diff
@@ -297,14 +298,17 @@ func (g *generator) rpPredeployTemplate() *arm.Template {
...
+			p.DefaultValue = []string{}
 		case "keyvaultPrefix":
 			p.MaxLength = 24 - max(len(env.ClusterKeyvaultSuffix), len(env.ServiceKeyvaultSuffix), len(env.PortalKeyvaultSuffix))
 		}
 		t.Parameters[param] = p
 	}
 
+	t.Resources = append(t.Resources, g.rpSecurityGroup()...)
 	t.Resources = append(t.Resources,
-		g.rpSecurityGroup(),
```
darthhexx, line 307: This removal would be reverted and a new one would be added for the new function `g.rpSecurityGroupFairFaxAdminPortal()`.

darthhexx (changes requested):
I recommend a change of approach with the ARM template changes. In conjunction with this PR we have to change the bootstrap ARM template to match the changes here.

pkg/deploy/generator/resources_rp.go
ventifus: pkg/deploy/assets/generate.go:
```go
//go:generate go run ../../../hack/gendeploy
```

gendeploy runs it both ways, true and false:
```go
func run() error {
	// dev artifacts
	err := generator.New(false).Artifacts()
	if err != nil {
		return err
	}

	// prod artifacts
	return generator.New(true).Artifacts()
}
```

pkg/deploy/generator/resources_rp.go
ventifus: I like this approach much better, I'll give it a shot this way.

darthhexx (approved):
LGTM

cadenmarchese (approved):
I like this approach, thank you @darthhexx for your review

PR #2557, opened 2022-11-24
[Testing-Ci] Make CI jobs run in container

SrinivasAtmakuri:
Please ignore, this is just for testing.

### Which issue this PR addresses:

Fixes

PR #2583, opened 2022-12-07
Containerize CI pipeline

SrinivasAtmakuri:
With this commit, the CI jobs running in shell will run inside containers.

Fixes: https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/16018371/

### Test plan for issue:

Testing done: https://github.com/Azure/ARO-RP/pull/2557

### DIFF
--- a/Dockerfile.python
+++ b/Dockerfile.python
@@ -5,31 +5,7 @@ RUN mkdir -p /app/ARO-RP
 WORKDIR /app/ARO-RP
 
 #we have multiple steps for copy so we can make use of caching
-COPY vendor/ vendor
-
-COPY docs/ docs
-COPY hack/ hack
-
-COPY swagger/ swagger
-COPY test/ test
-COPY python/ python
-COPY portal/ portal
-
-COPY cmd/ cmd
-COPY pkg/ pkg
-
-#COPY all files with an extension (directories not copied)
-COPY ?*.* /app/ARO-RP/
-COPY Makefile LICENSE /app/ARO-RP/
-
-COPY .git .git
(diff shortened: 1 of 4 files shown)


darthhexx (changes requested):
I have raised this before, but I am not seeing any mention in this PR of the fact that this is change going to prevent all Admin Portal E2E tests from executing.

That is going to open us up to potential issues, like we recently had when all Admin Portal tests were disabled by mistake, instead of just tests in production, that resulted in a broken Admin Portal being deployed to production.

We need a plan to work around this issue and that work needs to be ready to commit, before this PR is merged.

facchettos (changes requested):
already talked with srini about it on slack, but for the record, this is breaking the github action with the change from run to entryproint.

Dockerfile.python
```diff
@@ -0,0 +1,12 @@
...
+USER root
+RUN mkdir -p /app/ARO-RP
+WORKDIR /app/ARO-RP
+
+#we have multiple steps for copy so we can make use of caching
+COPY . /app/ARO-RP
+
+RUN microdnf update && \
+    microdnf install make python39 git && \
+    microdnf clean all
```
facchettos, line 12: does that mean that we will update all the packages before every run ? couldn't we build an image with those built in so we don't have to redownload and apply those everytime?

SrinivasAtmakuri:
A better approach is being tested here: https://github.com/Azure/ARO-RP/pull/2624. With the current approach of this PR, Golang runs are taking longer to run inside the custom build image, the 2624 approach doesn't seem to have this problem.

PR #2584, opened 2022-12-07
Operator checker split - move internet checker into a separate controller

m1kola:

Trying to enable the operator to run checks independently from each other.

### Test plan for issue:

Operator behaviour before the change should be the same as the behaviour with this change.

No, refactoring.


cmd/aro/operator.go
```diff
@@ -237,12 +238,17 @@ func operator(ctx context.Context, log *logrus.Entry) error {
 			arocli, operatorcli)).SetupWithManager(mgr); err != nil {
 			return fmt.Errorf("unable to create controller %s: %v", ingress.ControllerName, err)
 		}
+		if err = (checker.NewReconciler(
+			log.WithField("controller", checker.ControllerName),
+			arocli, kubernetescli, maocli, operatorcli, configcli, role)).SetupWithManager(mgr); err != nil {
+			return fmt.Errorf("unable to create controller %s: %v", checker.ControllerName, err)
+		}
```
m1kola, line 245: Only internet checker needs to be run on workers. Everything else - runs on master. So moving the whole thing into master.

pkg/operator/controllers/checkers/internetchecker/controller.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (r *Reconciler) conditionType() string {
+	switch r.role {
+	case "master":
+		return arov1alpha1.InternetReachableFromMaster
+	case "worker":
+		return arov1alpha1.InternetReachableFromWorker
+	default:
+		r.log.Warnf("unknown role %s, assuming worker role", r.role)
+		return arov1alpha1.InternetReachableFromWorker
```
s-amann, line 105: do we just want to return `"UnknownRole"` here? I think defaulting to `worker` is fine as this should be a weird corner case to start with.

pkg/operator/controllers/checkers/internetchecker/controller_test.go
```diff
@@ -0,0 +1,131 @@
...
+			controllerDisabled: true,
+			wantResult:         reconcile.Result{},
+		},
+	}
+
+	roleToConditionTypeMap := map[string]string{
+		"master": arov1alpha1.InternetReachableFromMaster,
+		"worker": arov1alpha1.InternetReachableFromWorker,
+	}
+	for _, testRole := range []string{operator.RoleMaster, operator.RoleWorker} {
```
s-amann, line 67: since our `conditionType()` switches over 'three values' should we add a nonworker/nonmaster role in this slice? - might be overkill though.

s-amann (approved):
lgtm, just some minor notes to consider

pkg/operator/controllers/checkers/internetchecker/controller.go
m1kola: We should not hit this as long as we have this in place:

https://github.com/Azure/ARO-RP/blob/b45b3a1d6d2c21acddb866920dd327dc224fde52/cmd/aro/operator.go#L56-L59

However I'm not a big fan of implicit things and I would love to `NewReconciler` to return an error if internet checker controller is being constructed with a role other than master or worker. For rest of the controllers we will only support master role so their `NewReconciler` should return an error for anything other than master.

I will try to do that once I finish the split (assuming I'll have a bit of time before I need to pick up my next card).

But I agree with your other comment: since we have this behaviour - it should be tested. I updated the test to do that (and thanks to you found an issue with the test).

Also your comment reminded me that we should probably set `Unknown` condition status in case if the controller is disabled.

s-amann: Awesome, I agree with the controller enforcing which roles it deals with too. If you run out of time LMK, I'd be happy to pick it up.

s-amann (approved):
lgtm, thank you!

cadenmarchese (approved):

hawkowl (approved):

PR #2585, opened 2022-12-08
Hive install OpenShift 4.10 and 4.11

darthhexx:

Addresses a few compatibility issues when installing using the ARO-Installer for OCP 4.10 and the [newly minted PR for 4.11](https://github.com/openshift/ARO-Installer/pull/4).

There are 2 compatibility issues that come up when installing both 4.10 and 4.11 clusters from this code base. Most notably, the installConfig has new optional parameters in the 4.11 installer that provide UltraSSD and Accelerated Networking support. There is an additional OSImage parameter and a change to the diskEncryptionSet (using a DiskEncryptionSet instead of an ID)  and the ability to manage to core cluster capabilities.

If we do not have these defined, the [updateClusterData function](https://github.com/Azure/ARO-RP/blob/a928f0ea2c2b13b7b2c2ef5dc82deb31fd8f4c63/pkg/cluster/ipaddresses.go#L29-L39) fails to parse the 4.11 installConfig object and fails to complete the cluster creation successfully.

The second part this PR addresses is supporting encrypting the default storage class with the BYOK provided by a customer.

### Test plan for issue:

Tested creating 4.10 and 4.11 clusters with both default platform encryption and user provided encryption keys.

Not directly, but for 4.11 in general there may be updates.

### DIFF
--- a/pkg/cluster/storageclass.go
+++ b/pkg/cluster/storageclass.go
@@ -12,11 +12,17 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
+
+	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
 const (
 	defaultStorageClassName          = "managed-premium"
 	defaultEncryptedStorageClassName = "managed-premium-encrypted-cmk"
+	defaultProvisioner               = "kubernetes.io/azure-disk"
+	ocp411StorageClassName           = "managed-csi"
+	ocp411EncryptedStorageClassName  = "managed-csi-encrypted-cmk"
+	ocp411Provisioner                = "disk.csi.azure.com"
 )
 
 // configureDefaultStorageClass replaces default storage class provided by OCP with
@@ -26,8 +32,23 @@ func (m *manager) configureDefaultStorageClass(ctx context.Context) error {
 		return nil
 	}
 
+	installVersion, err := version.ParseVersion(m.doc.OpenShiftCluster.Properties.ClusterProfile.Version)
+	if err != nil {
+		return err
+	}
+
+	storageClassName := defaultStorageClassName
+	encryptedStorageClassName := defaultEncryptedStorageClassName
+	provisioner := defaultProvisioner
+
+	if installVersion.V[0] == 4 && installVersion.V[1] == 11 {
+		storageClassName = ocp411StorageClassName
+		encryptedStorageClassName = ocp411EncryptedStorageClassName
+		provisioner = ocp411Provisioner
+	}
+
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		oldSC, err := m.kubernetescli.StorageV1().StorageClasses().Get(ctx, defaultStorageClassName, metav1.GetOptions{})
+		oldSC, err := m.kubernetescli.StorageV1().StorageClasses().Get(ctx, storageClassName, metav1.GetOptions{})
 		if err != nil {
 			return err
 		}
@@ -41,7 +62,7 @@ func (m *manager) configureDefaultStorageClass(ctx context.Context) error {
 			return err
 		}
 
-		encryptedSC := newEncryptedStorageClass(m.doc.OpenShiftCluster.Properties.WorkerProfiles[0].DiskEncryptionSetID)
+		encryptedSC := newEncryptedStorageClass(m.doc.OpenShiftCluster.Properties.WorkerProfiles[0].DiskEncryptionSetID, encryptedStorageClassName, provisioner)
 		_, err = m.kubernetescli.StorageV1().StorageClasses().Create(ctx, encryptedSC, metav1.CreateOptions{})
 		if err != nil && !kerrors.IsAlreadyExists(err) {
 			return err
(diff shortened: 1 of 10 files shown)


darthhexx:
Vendor checking is obviously failing on this PR, because I've manually modified installer files. The decision is going to have to be taken on how we progress on this. Should we re-vendor for the 4.11 installer or create a custom interface to deal with the multi-version compatibility?

pkg/cluster/storageclass.go
```diff
@@ -12,11 +12,17 @@ import (
...
+	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
 const (
 	defaultStorageClassName          = "managed-premium"
 	defaultEncryptedStorageClassName = "managed-premium-encrypted-cmk"
+	defaultProvisioner               = "kubernetes.io/azure-disk"
+	ocp411StorageClassName           = "managed-csi"
+	ocp411EncryptedStorageClassName  = "managed-csi-encrypted-cmk"
+	ocp411Provisioner                = "disk.csi.azure.com"
```
lranjbar, line 25: Can we make this more generic? Will we have to add more constants for 4.12 support? 4.13 support?

One of the patterns we used in Assisted Installer is to keep JSON files for the different versions to handle defaults see: https://github.com/openshift/assisted-service/blob/master/data/default_os_images.json

In Assisted Installer these are loaded at runtime instead of compile time. We could use this pattern and modify it for compile time if ARO needs it. Just thinking about maintainability of multi-version in the future. WDYT?

lranjbar:
> Vendor checking is obviously failing on this PR, because I've manually modified installer files. The decision is going to have to be taken on how we progress on this. Should we re-vendor for the 4.11 installer or create a custom interface to deal with the multi-version compatibility?

My gut tells me to vendor for 4.11 and then expand on the custom interface idea. It sounds interesting.

darthhexx:
> 

I found a cleaner way, by conditionally allowing incoming unknown fields when the graph in loaded.

pkg/cluster/storageclass.go
lranjbar: I like the changes to make it more generic. :)

> They may change, but we are going to have to adapt the RP to handle any changes as we GA new versions, which means making version specific tweaks to our code in order to maintain multi-version compatibility. If 4.12 keeps the same storage class as 4.11, then we could remove the special handling when we have 4.11 and 4.12 as install options.

That's the pattern I was seeing if we have multiple versions of OCP in ARO for install we may end up with very complicated logic to deal with defaults that could be hard to back out. Different versions of OCP will tend to have different installation defaults. For now I think this change is good! 

It has got me thinking about a more proactive approach (external definitions or not) rather than reactive, to try to control the different settings for versions of OCP in ARO.

SudoBrendan (commented):
This PR makes me nervous for multi-version. If we have 1 codebase with 1 branch that must support N many versions (through conditionals, like the ones in this PR), I wonder how long it will take before we get into scaling/permutation issues. At the very least, we should be E2E'ing all versions we expect to support if this is the path we're choosing.

None of the code changes here look inherently wrong, but I don't know how this pattern of conditionals per version will scale (e.g. how will we deprecate the "4.10 logic" when we deprecate that version? How would we identify it? How would we tear it out without affecting other versions?). This is an idea that goes beyond the scope of this PR, but something we need to think very hard about moving forward.

darthhexx:
> 
As noted above, we have a single conditional introduced to support both 4.10 and 4.11. One which we already know will be defunct when 4.12 goes GA and 4.10 falls out of support. I really don't see this as an issue for the support of 4.10 and 4.11 multi-version.

Moreover, the fact that we are going to remove the installer module and carve a new path forward in Q1, I'd say we should gather all of our ideas of what that should look like and be ready to plan out what that system looks like.

cblecker (approved):

cblecker (commented):
This is fine for storage, but we'll also need to set a conditional to use OVN.

Should we do that in this PR or a follow up? cc @SudoBrendan

pkg/installer/deployresources.go
```diff
@@ -27,7 +27,7 @@ func (m *manager) deployResourceTemplate(ctx context.Context) error {
 
 	var installConfig *installconfig.InstallConfig
 	var machineMaster *machine.Master
-	err = pg.Get(&installConfig, &machineMaster)
+	err = pg.Get(true, &installConfig, &machineMaster)
```
s-amann, line 30: imo since it becomes critical what the value of `disallowUnknownFields` is passed in for the various cases of getting the graph we should have at least mocking expectations of the desired value. Looking through the existing tests for some of these files that would be a huge ask for this PR. WDYT about a work item being created to enhance the tests in this regard?

darthhexx: Sounds like a good idea to me. More coverage is always a good thing.

petrkotas: @s-amann  have you created that workitem please?

petrkotas (approved):
Thank you @darthhexx good one. I have only two questions, other than that LGTM.
1) I see that get has now `bool` to set ignoring missing fields, we set it in the functions, we know the missing field occur, and do not care. Such update. Did I get that right?
2) The storage class and even the first change done here, are part of the ARO embedded installer. How is this related to the external 4.11 installer planned to be released?

darthhexx:
> * I see that get has now `bool` to set ignoring missing fields, we set it in the functions, we know the missing field occur, and do not care. Such update. Did I get that right?

It s actually making the [DisallowUnknownFields togglable](https://github.com/Azure/ARO-RP/blob/bb76a8449a7e658ca0c5f5920e51776b34aa2518/pkg/cluster/graph/persistedgraph.go#L24-L26) so we can turn it off for the cluster install data coming back from Hive.

This is because the 4.11 installer has a few new fields that are not present in our vendored 4.10 installer struct, but we only care about the common ones. So we turn **off** **Disallow**UnknownFields (double negative, so it translates to Allow Unknown Fields) to let the common fields be populated, which we [use here](https://github.com/Azure/ARO-RP/blob/bb76a8449a7e658ca0c5f5920e51776b34aa2518/pkg/cluster/ipaddresses.go#L42-L44).

All other calls to the function remain strict in the parsing and do not allow unknown fields.

> * The storage class and even the first change done here, are part of the ARO embedded installer. How is this related to the external 4.11 installer planned to be released?

These changes allow both the 4.10 and 4.11 ARO-Installer to work together with the RP that currently has the 4.10 installer vendored into it.

PR #2587, opened 2022-12-09
Add operator info admin portal

ellis-johnson:

Fixes 9777208

Added Cluster Operator Information to Admin Portal so it is easy to see what operators are not available. This will be beneficial to on call.

### Test plan for issue:

Added unit tests

No


portal/v2/src/ClusterDetailListComponents/ClusterOperatorsWrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { ICluster } from "../App"
+import { IMessageBarStyles, MessageBar, MessageBarType, Stack } from '@fluentui/react';
+import { clusterOperatorsKey } from "../ClusterDetail";
+import { ClusterOperatorListComponent } from "./ClusterOperatorList";
+
+export interface ICondition {
+  status: string,
+  reason: string,
+  lastUpdated: string,
+  message: string
```
s-amann, line 13: should we also add `type: string` here as well?

ellis-johnson: So we don't need to actually add it here as the type is what is used for the header of the InfoList
image

If we added type to this interface as well the heading would double up with an entry to the InfoList

s-amann (approved):

darthhexx (changes requested):
This PR still requires a `make build-portal`. Also could we list the operator status' in a single column, like we would read them from the `oc get co` output, instead of them being in rows and adaptive columns based on screen width?

portal/v2/src/ClusterDetailListComponents/ClusterOperatorList.tsx
```diff
@@ -0,0 +1,188 @@
...
+    }
+  ])
+
+  const [clusterOperatorList, setClusterOperatorList] = useState<IClusterOperatorList[]>([])
+  const [clusterOperatorDetailsVisible, setClusterOperatorDetailsVisible] = useState<boolean>(false)
+  const [currentClusterOperator, setCurrentclusterOperator] = useState<string>("")
+  const [shimmerVisibility, SetShimmerVisibility] = useState<boolean>(true)
+
+  useEffect(() => {
+    setClusterOperatorList(createClusterOperatorList(props.clusterOperators))
```
jaitaiwan, line 84: You might not need the `useEffect` here. This code will cause a double-render (this will show when developing with React strict mode on): https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops

portal/v2/src/ClusterDetailListComponents/ClusterOperatorList.tsx
```diff
@@ -0,0 +1,188 @@
...
+  }, [props.clusterOperators] );
+
+  useEffect(() => {
+    const newColumns: IColumn[] = columns.slice();
```
jaitaiwan, line 88: This could likely be done using an inline function rather than a useEffect. For example `getNewColumns` which calculates the slice etc within the view.

portal/v2/src/ClusterDetailListComponents/ClusterOperatorList.tsx
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+  }, [clusterOperatorList] )
+
+  function _onClusterOperatorInfoLinkClick(operator: string) {
+    setClusterOperatorDetailsVisible(!clusterOperatorDetailsVisible)
+    setCurrentclusterOperator(operator)
+  }
+
+  function _copyAndSort<T>(items: T[], columnKey: string, isSortedDescending?: boolean): T[] {
```
jaitaiwan, line 105: Should this be here? It looks like a utility function that should be in some sort of utilities/helpers file.

portal/v2/src/ClusterDetailListComponents/ClusterOperatorsWrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+  available: string,
+  conditions?: ICondition[],
+}
+
+export function ClusterOperatorsWrapper(props: {
+  currentCluster: ICluster
+  detailPanelSelected: string
+  loaded: boolean
+}) {
+  const [data, setData] = useState<any>([])
```
jaitaiwan, line 27: Two things here, I think we should use a more descriptive name than `data` and I don't think we should use `any` here as we control what data we set.

portal/v2/src/ClusterDetailListComponents/ClusterOperatorsWrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+          name: element.name,
+          available: element.available
+        }
+        clusterOp.conditions = []
+        element.conditions.forEach((condition: ICondition) => {
+          clusterOp.conditions!.push(condition)
+        });
+        clusterOperatorList.push(clusterOp)
+      });
+      state.current.setState({ clusterOperators: clusterOperatorList })
```
jaitaiwan, line 69: Why are we using a `useRef` for this?

portal/v2/src/ClusterDetailListComponents/ClusterOperatorsWrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+      setFetching("FETCHING")
+      FetchClusterOperators(props.currentCluster).then(onData)
+    }
+  }, [data, props.loaded, props.detailPanelSelected])
+  
+  return (
+    <Stack>
+      <Stack.Item grow>{error && errorBar()}</Stack.Item>
+      <Stack>
+        <ClusterOperatorListComponent clusterOperators={data!} ref={state} clusterName={props.currentCluster != null ? props.currentCluster.name : ""} />
```
jaitaiwan, line 96: Can we actually assume that we have `data` here at this point? If we have an `error` will that mean that `data` may be undefined?

portal/v2/src/Request.tsx
```diff
@@ -75,6 +75,17 @@ export const FetchMachineSets = async (cluster: ICluster): Promise<AxiosResponse
   }
 }
 
+export const FetchClusterOperators = async (cluster: ICluster): Promise<AxiosResponse | null> => {
+  try {
+    const result = await axios(
+      "/api/" + cluster.subscription + "/" + cluster.resourceGroup + "/" + cluster.name + "/clusteroperators")
```
jaitaiwan, line 81: I'd optionally suggest using `.join` here. E.g. `[cluster.subscription, cluster.resourceGroup, cluster.name, "clusteroperators"].join("/")`

pkg/portal/cluster/clusteroperators.go
```diff
@@ -10,9 +10,18 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type OperatorCondition struct {
```
jaitaiwan, line 13: It would be good if we could document the type, and what it achieves.

pkg/portal/cluster/clusteroperators.go
```diff
@@ -10,9 +10,18 @@ import (
...
+	Type        configv1.ClusterStatusConditionType `json:"type"`
+	LastUpdated string                              `json:"lastUpdated"`
+	Status      configv1.ConditionStatus            `json:"status"`
+	Reason      string                              `json:"reason"`
+	Message     string                              `json:"message"`
+}
+
 type OperatorInformation struct {
```
jaitaiwan, line 21: Also documenting this struct as well

pkg/portal/cluster/clusteroperators_test.go
```diff
@@ -55,33 +56,173 @@ func TestClusterOperators(t *testing.T) {
...
+				},
 			},
 		},
 	}
 
 	sort.SliceStable(info.Operators, func(i, j int) bool { return info.Operators[i].Name < info.Operators[j].Name })
 	sort.SliceStable(expected.Operators, func(i, j int) bool { return expected.Operators[i].Name < expected.Operators[j].Name })
 
+	layout := "2006-01-02 15:04:05 -0700 MST"
+	for i, r := range info.Operators {
```
jaitaiwan, line 215: You could possibly make use of parallel tests here: https://clavinjune.dev/en/blogs/speed-up-golang-test-execution-time/

portal/v2/src/ClusterDetailList.tsx
```diff
@@ -4,6 +4,7 @@ import { NodesWrapper } from './ClusterDetailListComponents/NodesWrapper';
 import { MachinesWrapper } from "./ClusterDetailListComponents/MachinesWrapper";
 import { MachineSetsWrapper } from "./ClusterDetailListComponents/MachineSetsWrapper";
 import { ICluster } from "./App"
+import { ClusterOperatorsWrapper } from "./ClusterDetailListComponents/ClusterOperatorsWrapper";
 
 interface ClusterDetailComponentProps {
   item: any
```
jaitaiwan, line 10: If possible could we replace this `any`?

portal/v2/src/ClusterDetailList.tsx
```diff
@@ -70,6 +71,12 @@ export class ClusterDetailComponent extends Component<ClusterDetailComponentProp
             <MachineSetsWrapper currentCluster={this.props.cluster!} detailPanelSelected={this.props.detailPanelVisible} loaded={this.props.isDataLoaded}/>
           );
         }
+        case "clusteroperators":
+        {
+          return (
+            <ClusterOperatorsWrapper currentCluster={this.props.cluster!} detailPanelSelected={this.props.detailPanelVisible} loaded={this.props.isDataLoaded}/>
```
jaitaiwan, line 77: Can we truly expect cluster to be defined? if so I think it would better to make a change to the type or add a conditional to do type narrowing (https://www.typescriptlang.org/docs/handbook/2/narrowing.html).

portal/v2/src/ClusterDetailListComponents/ClusterOperator.tsx
```diff
@@ -0,0 +1,85 @@
+import { Component } from "react"
+import { Stack, Text, PivotItem, IStackItemStyles } from '@fluentui/react';
+import { ICondition, IClusterOperator } from "./ClusterOperatorsWrapper";
+import { contentStackStylesNormal } from "../App";
+import { MultiInfoList } from "./InfoList"
+
+export interface ClusterOperatorComponentProps {
+    clusterOperators: any
```
jaitaiwan, line 8: Are we able to define a concrete type here? If not `unknown` would be a better option.

portal/v2/src/ClusterDetailListComponents/ClusterOperatorList.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { IClusterOperator } from './ClusterOperatorsWrapper';
+import { ClusterOperatorsComponent } from './ClusterOperator';
+
+export declare interface IClusterOperatorList {
+  name: string;
+  available: string;
+}
+
+interface ClusterOperatorComponentProps {
+  clusterOperators: any
```
jaitaiwan, line 16: Another place it would be better to specify a concrete type or `unknown` if possible.

jaitaiwan (changes requested):
Hi Ellis, I've added a bunch of comments which could be thought of as semantic and possibly ignorable. I'm only just getting aboard Red Hat culture so please let me know! Thanks for the opportunity to review :)

pkg/portal/cluster/clusteroperators_test.go
```diff
@@ -55,33 +56,173 @@ func TestClusterOperators(t *testing.T) {
...
+		for j, c := range r.Conditions {
+			_, err := time.Parse(layout, c.LastUpdated)
+			if err != nil {
+				t.Error("Given date was not in the correct format " + layout + err.Error())
+			} else {
```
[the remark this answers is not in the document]
ellis-johnson: Why would we want to continue after the error? I've removed the else as that is unnecessary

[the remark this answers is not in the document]
jaitaiwan: I completely forgot that t.Error() in go stops execution of the test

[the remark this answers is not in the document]
ellis-johnson: Done!

portal/v2/src/ClusterDetailListComponents/ClusterOperator.tsx
```diff
@@ -0,0 +1,85 @@
...
+    props.clusterOperators.forEach((operator: IClusterOperator) => {
+        if (operator.name === props.clusterOperatorName) {
+            currentClusterOperator = operator
+            return
+        }
+    })
+    
+    return (
+            <>
+                {renderClusterOperators(currentClusterOperator!)}
```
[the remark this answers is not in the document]
ellis-johnson: Done!

portal/v2/src/ClusterDetailListComponents/ClusterOperator.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    <Text variant="large" styles={contentStackStylesNormal}>Conditions</Text>
+                </Stack>
+                <Stack wrap styles={stackItemStyles}horizontal grow>
+                    <MultiInfoList headers={ConditionDetails} items={operator.conditions} title="Conditions" subProp="type" titleSize="medium"/>
+                </Stack>
+            </PivotItem>;
+};
+
+function PivotOverflowMenuExample(props: {
+    clusterOperators: any,
```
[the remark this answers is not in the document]
ellis-johnson: Used IClusterOperator[]

portal/v2/src/ClusterDetailListComponents/ClusterOperatorList.tsx
```diff
@@ -0,0 +1,188 @@
...
+  
+  public render() {
+    return (
+        <ClusterOperatorListHelperComponent clusterOperators={this.state.clusterOperators} clusterName={this.state.clusterName}/>
+      )
+  }
+}
+
+export function ClusterOperatorListHelperComponent(props: {
+     clusterOperators: any,
```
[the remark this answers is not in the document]
ellis-johnson: used IClusterOperator[]

portal/v2/src/ClusterDetailList.tsx
ellis-johnson: Done!

portal/v2/src/ClusterDetailListComponents/ClusterOperatorList.tsx
ellis-johnson: Fixed

portal/v2/src/ClusterDetailListComponents/ClusterOperatorList.tsx
ellis-johnson: Fixed

portal/v2/src/ClusterDetailListComponents/ClusterOperatorsWrapper.tsx
ellis-johnson: Fixed

portal/v2/src/ClusterDetailListComponents/ClusterOperatorsWrapper.tsx
ellis-johnson: I believe so that it can be passed down to the child component and updated

portal/v2/src/ClusterDetailListComponents/ClusterOperatorsWrapper.tsx
ellis-johnson: Was able to just remove this since we use an array now it can be empty

portal/v2/src/Request.tsx
ellis-johnson: Done

portal/v2/src/ClusterDetailList.tsx
ellis-johnson: Have added a check for cluster defined prior to this component

pkg/portal/cluster/clusteroperators_test.go
ellis-johnson: This is a singular test, what do you mean by this here?

pkg/portal/cluster/clusteroperators.go
```diff
@@ -25,17 +36,38 @@ func clusterOperatorsInformationFromOperatorList(operators *configv1.ClusterOper
...
+		var conditions []OperatorCondition
 		for _, cnd := range co.Status.Conditions {
 			if cnd.Type == "Available" {
-				Available = cnd.Status
+				available = cnd.Status
+			} else if cnd.Type == "Progressing" {
+				progressing = cnd.Status
+			} else if cnd.Type == "Degraded" {
+				degraded = cnd.Status
+			}
```
facchettos, line 51: a switch case would be better here, also, please use the constants provided in configv1

facchettos (dismissed):

portal/v2/src/ClusterDetailListComponents/ClusterOperatorList.tsx
```diff
@@ -0,0 +1,188 @@
...
+  clusterOperators: any
+  clusterName: string
+}
+
+export interface IClusterOperatorListState {
+  clusterOperators: IClusterOperator[]
+  clusterName: string
+}
+
+export class ClusterOperatorListComponent extends React.Component<ClusterOperatorComponentProps, IClusterOperatorListState> {
```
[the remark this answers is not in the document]
ellis-johnson: I think this is something we should keep in mind as currently we use classes in a bunch of places. We might need a large refactoring effort in order to resolve this

portal/v2/src/ClusterDetailListComponents/ClusterOperatorList.tsx
ellis-johnson: I am not entirely sure what you mean here, where would this be called if we don't use a useeffect?

pkg/portal/cluster/clusteroperators.go
ellis-johnson: Done

pkg/portal/cluster/clusteroperators.go
ellis-johnson: Done

pkg/portal/cluster/clusteroperators.go
ellis-johnson: Fixed

pkg/portal/cluster/clusteroperators_test.go
jaitaiwan: And a small warning too: https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721

portal/v2/src/ClusterDetailList.tsx
jaitaiwan, line 1: At a quick glance of the switch statement all of the nodes you're using here have the same prop structure except overview.

So what I'd expect here would that you would use a variable to store the component and then use the variable to render it e.g.:

```
const detailComponents = {
  'nodes': NodesWrapper,
  'machines': MachinesWrapper,
  ... 
}

const DetailView = detailComponents[this.props.detailpanel.Visible.toLowerCase()]
return <DetailView ...thepropsgohere />
```

Then all that's left is 2 things:

1. An if statement to handle when cluster is null
2. An if statement to render the overview

This should really simplify this component down a ton.

As a side note, I'm also not a fan of the prop `detailPanelVisible` as it is named almost like it's a boolean rather than a which view is selected. If you're feeling like it, seeing it named as `selectedPanel` or something like that would be very nice.

portal/v2/src/ClusterDetailListComponents/ClusterOperator.tsx
```diff
@@ -0,0 +1,87 @@
...
+        }
+    })
+    
+    return (
+
+            <>
+                    {(currentClusterOperator != undefined) ? 
+                        renderClusterOperators(currentClusterOperator!) : null}
+            </>
+    );
```
jaitaiwan, line 64: The `forEach` here seems like extra work to avoid just doing a [for..of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of). If you used a for..of you could immediately return the renderedClusterOperator if found, and then return null as default for the function.

portal/v2/src/ClusterDetailListComponents/ClusterOperatorList.tsx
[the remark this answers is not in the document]
jaitaiwan: I would advise taking the bite out of the elephant approach. One-at-time just change each class component. Obviously don't have to now but that would be my preference.

portal/v2/src/ClusterDetailListComponents/ClusterOperatorList.tsx
jaitaiwan: I won't worry about this for now, I might setup a phone call to go through with those who do react stuff, common useEffect overuses.

portal/v2/src/ClusterDetailListComponents/ClusterOperatorsWrapper.tsx
jaitaiwan: I might add this to things to talk about on a react call.

jaitaiwan (changes requested):
A few more comments but lookin' great!

portal/v2/src/ClusterDetailListComponents/ClusterOperator.tsx
ellis-johnson: `props.clusterOperators.forEach((operator: IClusterOperator) => {
        if (operator.name === props.clusterOperatorName) {
            return (

                <>
                        {renderClusterOperators(operator)}
                </>
        );
        }
    })
    return <></>`
    
  I've managed to do this suggestion while still using foreach, is there any reason that for..of is better?

portal/v2/src/ClusterDetailList.tsx
ellis-johnson: Is this a refactor that makes sense for this PR or should we branch it out into a seperate task?

pkg/portal/cluster/clusteroperators_test.go
ellis-johnson: This is undeniabally an optimisation on time, but it feels like an unneccessary change given that the entire portal suite takes 100ms to run and the entire ARO-RP unit test suite takes 60 seconds.

I'd like to hear opinions from others in regards if we want to start thinking about time optimisation for unit testing?

portal/v2/src/ClusterDetailList.tsx
jaitaiwan: It's a 5 minute change. Here's what I was thinking, you might need to just make sure types are working:

```
import { Component } from "react"
import { OverviewWrapper } from './ClusterDetailListComponents/OverviewWrapper';
import { NodesWrapper } from './ClusterDetailListComponents/NodesWrapper';
import { MachinesWrapper } from "./ClusterDetailListComponents/MachinesWrapper";
import { MachineSetsWrapper } from "./ClusterDetailListComponents/MachineSetsWrapper";
import { ICluster } from "./App"
import { ClusterOperatorsWrapper } from "./ClusterDetailListComponents/ClusterOperatorsWrapper";

interface ClusterDetailComponentProps {
  item: IClusterDetails
  cluster: ICluster | null
  isDataLoaded: boolean
  detailPanelVisible: string
}

export interface IClusterDetails {
  apiServerVisibility: string
  apiServerURL: string
  architectureVersion: string
  consoleLink: string
  createdAt: string
  createdBy: string
  failedProvisioningState: string
  infraId: string
  lastAdminUpdateError: string
  lastModifiedAt: string
  lastModifiedBy: string
  lastProvisioningState: string
  location: string
  name: string
  resourceId: string
  provisioningState: string
  version: string
  installStatus: string
}

interface IClusterDetailComponentState {
  item: IClusterDetails // why both state and props?
  detailPanelSelected: string
}

const panelComponents = {
  'nodes': NodesWrapper,
  'machines': MachinesWrapper,
  'machinesets': MachineSetsWrapper,
  'clusteroperators': ClusterOperatorsWrapper
}

export class ClusterDetailComponent extends Component<ClusterDetailComponentProps, IClusterDetailComponentState> {

  constructor(props: ClusterDetailComponentProps | Readonly<ClusterDetailComponentProps>) {
    super(props);
  }

  public render() {
    if (this.props.cluster == null) {
      return null
    }

    const panelType = this.props.detailPanelVisible.toLowerCase()
    if (panelType === "overview") {
      return <OverviewWrapper clusterName= {this.props.item.name} currentCluster={this.props.cluster!} detailPanelSelected={this.props.detailPanelVisible} loaded={this.props.isDataLoaded}/>
    }

    const Panel = panelComponents[panelType]
    if (!Panel) {
      return null
    }
    
    return <Panel currentCluster={this.props.cluster!} detailPanelSelected={this.props.detailPanelVisible} loaded={this.props.isDataLoaded}/>
  }
}
```

portal/v2/src/ClusterDetailListComponents/ClusterOperator.tsx
jaitaiwan: This will just return `<></>` every time. If you do it like you've just posted, all that's happening is that you're creating an array something like `[undefined, undefined, undefined, <component>, undefined]` whose value isn't used anywhere.

pkg/portal/cluster/clusteroperators_test.go
jaitaiwan: I don't think it's that big a change to warrant caring one way or the other... but here's a code snippet for it to encourage having the change:
```
	for i, r := range info.Operators {
		// Ensure vars are accessible to goroutine in t.Run
		r := r
		i := i
		t.Run(r.Name, func(t *testing.T) {
			t.Parallel()
			for j, c := range r.Conditions {
				// Ensure vars are accessible to goroutine in t.Run
				c := c
				j := j
				t.Run(c.Type, func(t *testing.T) {
					t.Parallel()
					_, err := time.Parse(layout, c.LastUpdated)
					if err != nil {
						t.Error("Given date was not in the correct format " + layout + err.Error())
						continue
					}
					info.Operators[i].Conditions[j].LastUpdated = ""
				})
				
			}
		})
	}
```

portal/v2/src/ClusterDetailListComponents/ClusterOperator.tsx
ellis-johnson: Fixed

portal/v2/src/ClusterDetailList.tsx
ellis-johnson: Done!

pkg/portal/cluster/clusteroperators_test.go
ellis-johnson: Done!

portal/v2/src/ClusterDetailList.tsx
```diff
@@ -38,38 +39,31 @@ interface IClusterDetailComponentState {
   detailPanelSelected: string
 }
 
+const detailComponents: Map<string, any> = new Map<string, any>([
```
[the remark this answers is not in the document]
jaitaiwan: Alternatively use something like ReactComponent<interface>

jaitaiwan (changes requested):

pkg/portal/cluster/clusteroperators_test.go
```diff
@@ -53,32 +53,186 @@ func TestClusterOperators(t *testing.T) {
...
 		},
 	}
 
+	for i, r := range info.Operators {
+		for j := range r.Conditions {
+			info.Operators[i].Conditions[j].LastUpdated = ""
+		}
+	}
+
+	for _, r := range info.Operators {
```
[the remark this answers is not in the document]
ellis-johnson: Fixed!

jaitaiwan (dismissed):
LGTM!

s-amann (dismissed):

s-amann (dismissed):

s-amann (approved):

PR #2588, opened 2022-12-09
Operator checker split - move service principal checker into a separate controller

m1kola:

Apart from the split I did the following changes:

* Now we write the full error into the condition in case of a `CloudError`: previously it was only message from the error (which does not include status code, error code, etc).
* Now we watch for the changes in the service principal secret. This should give us feedback faster (no need to wait for 1 hour in the worst case scenario)
* Also fixes unit test coverage for this checker: currently in master we have a separate code branch just for unit tests and production code was in a different branch.

No, refactoring/optimisation.


pkg/operator/controllers/checkers/serviceprincipalchecker/controller.go
```diff
@@ -0,0 +1,121 @@
...
+
+// Reconcile will keep checking that the has a valid cluster service principal.
+func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	instance, err := r.arocli.AroV1alpha1().Clusters().Get(ctx, arov1alpha1.SingletonClusterName, metav1.GetOptions{})
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if !instance.Spec.OperatorFlags.GetSimpleBoolean(checkercommon.ControllerEnabled) {
+		r.log.Info("controller is disabled")
```
m1kola, line 67:
```suggestion
		r.log.Debug("controller is disabled")
```

pkg/operator/controllers/checkers/internetchecker/controller.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Reconcile will keep checking that the cluster can connect to essential services.
+func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	instance, err := r.arocli.AroV1alpha1().Clusters().Get(ctx, arov1alpha1.SingletonClusterName, metav1.GetOptions{})
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if !instance.Spec.OperatorFlags.GetSimpleBoolean(checkercommon.ControllerEnabled) {
+		r.log.Info("controller is disabled")
```
m1kola, line 62:
```suggestion
		r.log.Debug("controller is disabled")
```

pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go
```diff
@@ -0,0 +1,61 @@
...
+
+type servicePrincipalChecker interface {
+	Check(ctx context.Context, AZEnvironment string) error
+}
+
+type checker struct {
+	log *logrus.Entry
+
+	credentialsGetter      func(ctx context.Context) (*clusterauthorizer.Credentials, error)
+	spValidatorConstructor func(azEnv *azureclient.AROEnvironment) (dynamic.ServicePrincipalValidator, error)
```
facchettos, line 26: are these required to be functions or could we use interfaces here instead? 
also, we probably could just have `spValidator` and delegate the construction to the implementer

pkg/operator/controllers/checkers/serviceprincipalchecker/controller.go
```diff
@@ -0,0 +1,133 @@
...
+	return &Reconciler{
+		log:  log,
+		role: role,
+
+		arocli:  arocli,
+		checker: newServicePrincipalChecker(log, kubernetescli),
+	}
+}
+
+// Reconcile will keep checking that the has a valid cluster service principal.
```
facchettos, line 59: I think you forgot a word here

pkg/operator/controllers/checkers/serviceprincipalchecker/controller_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+	"github.com/Azure/ARO-RP/pkg/util/cmp"
+	utillog "github.com/Azure/ARO-RP/pkg/util/log"
+)
+
+type fakeChecker func(ctx context.Context, AZEnvironment string) error
+
+func (fc fakeChecker) Check(ctx context.Context, AZEnvironment string) error {
+	return fc(ctx, AZEnvironment)
+}
```
facchettos, line 30: this pattern seems a little weird tbh. what does it achieve, is it a function on a function? not 100% sure I understand it (maybe just missing something)
shouldn't the checker be an interface if we need to mock it to test the reconciler?

pkg/operator/controllers/checkers/serviceprincipalchecker/controller_test.go
```diff
@@ -0,0 +1,132 @@
...
+			r := &Reconciler{
+				log:  utillog.GetLogger(),
+				role: "master",
+				checker: fakeChecker(func(ctx context.Context, AZEnvironment string) error {
+					if !reflect.DeepEqual(AZEnvironment, azureclient.PublicCloud.Environment.Name) {
+						t.Error(cmp.Diff(AZEnvironment, azureclient.PublicCloud.Environment.Name))
+					}
+
+					return tt.checkerReturnErr
+				}),
```
facchettos, line 94: I am having a hard time following that part, do we need to type cast it as a `fakechecker` ? wouldn't the function satisfy the type directly? (I think it should be an interface if we have to mock it though, as per previous comment)

pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func newServicePrincipalChecker(log *logrus.Entry, kubernetescli kubernetes.Interface) servicePrincipalChecker {
```
facchettos, line 29: we probably don't need to return the interface type here. We are losing access to the struct fields because of that and will need to typecast it. Returning the actual type will work just as well

pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go
facchettos: the `dynamic` package does weird things actually, we might want to look into it to fix it because the current implementation is unnecessarily complicated. e.g. this (which really seems to be an unneeded constructor, which returns the interface type instead of the actual type): 
```go 
func NewServicePrincipalValidator(log *logrus.Entry, azEnv *azureclient.AROEnvironment, authorizerType AuthorizerType, tokenClient aad.TokenClient) (ServicePrincipalValidator, error) {
	return &dynamic{
		log:            log,
		authorizerType: authorizerType,
		azEnv:          azEnv,
		tokenClient:    tokenClient,
	}, nil
}
```

but this is for another PR

facchettos (changes requested):

pkg/operator/controllers/checkers/serviceprincipalchecker/controller.go
m1kola: Yep, I know. I didn't want to rebase and wait for e2e to pass, so I fixed it in [another PR](https://github.com/Azure/ARO-RP/pull/2589/files#diff-09378a77e4f4272513cbac218616bb64044eace567cf6cd181a07eff526f7fcfR59).

pkg/operator/controllers/checkers/serviceprincipalchecker/controller_test.go
m1kola: @facchettos I'll answer two related comments here.

> wouldn't the function satisfy the type directly?

No - `checker` field in the `Reconciler` expects an instance which implements `servicePrincipalChecker` interface. The anonymous function on these lines won't statisfy the interface.

Hence we need type casting to `fakeChecker` which does statisfy it.

> (I think it should be an interface if we have to mock it though, as per previous comment)

It already is an interface and we mock it with `fakeChecker`. We do not use `mockgen` from `go-mock` becaue it is unexported interface and only is required once. So I see no point of generating a mock for it.

No about [the above comment](https://github.com/Azure/ARO-RP/pull/2588/files#r1048210367):

> this pattern seems a little weird tbh. what does it achieve, is it a function on a function? not 100% sure I understand it (maybe just missing something)
shouldn't the checker be an interface if we need to mock it to test the reconciler?

This is the same pattern as with `http.HandlerFunc` from the standart library ([source](https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/net/http/server.go;l=2101-2110)).

We just make a function to statisfy an interface with one method.

pkg/operator/controllers/checkers/serviceprincipalchecker/controller_test.go
m1kola: Answered in [the thread below](https://github.com/Azure/ARO-RP/pull/2588/files#r1048214246).

pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go
m1kola: But we don't need access to the fields. We only need access to the method defined in the interface.

* When testing the `checker` struct itself - we have access to the fields.
* When testing `Reconciler` - we do not have access to the feilds and we replace the whole thing by a mock which statisfies `servicePrincipalChecker` interface.

pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go
m1kola: > are these required to be functions or could we use interfaces here instead?

These are needed so that we can swap constructors (`clusterauthorizer.AzCredentials(...)` and `dynamic.NewServicePrincipalValidator(...)`) with constructors which return mocks.

We can define an interface, but defining an interface for a constructure seems like an overkill to me.

> the `dynamic` package does weird things actually, we might want to look into it to fix it because the current implementation is unnecessarily complicated.

I agere with this part

> e.g. this (which really seems to be an unneeded constructor, which returns the interface type instead of the actual type)

I think nothing wrong with it: we return an interface becuase we don't want to expose fields of `dynamic`. If you need a validator somewhere - you use interface and mock the whole thing if you want to test your unit.

facchettos: but since it is for a another PR, that is why I was recommending to delegate it to an interface so we didn't have to/would be easier change it later when we fix `dynamic`

pkg/operator/controllers/checkers/serviceprincipalchecker/controller_test.go
facchettos: is it really simplifying if we compare it to that though?
```go
type mockChecker struct {
  returnError error
}
func (fc mockChecker) Check(ctx context.Context, AZEnvironment string) error {
	if !reflect.DeepEqual(AZEnvironment, azureclient.PublicCloud.Environment.Name) {
		t.Error(cmp.Diff(AZEnvironment, azureclient.PublicCloud.Environment.Name))
	}

	return returnError
}

//and in the test
fakeChecker := mockChecker{tt.checkReturnError}
r := &Reconciler{
				log:  utillog.GetLogger(),
				role: "master",
				checker: fakeChecker,
}
```

pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go
m1kola: > and this is why we end up having to swap out useless constructors.

No, it is not the reason. We have to do that because of two reasons:
* For the validator - info about the environment (e.g. public cloud, gov cloud) comes from an instance which we know only on the reconcile stage. If we can address this - we can drop a swappable constructor for the validator. For rest of the codebase - I think it can already be stored in the field. E.g. [here](https://github.com/Azure/ARO-RP/blob/d5485932f428cf2c61d51820a825b1c04fe97d20/pkg/api/validate/openshiftcluster_validatedynamic.go#L59-L63) we should be able to store it in a field of `openShiftClusterDynamicValidator`.
* For `credentialsGetter` - we can not store credentials between the reconcile calls becasue we want to ensure that whatever we have in the cluster is valid hence we need to call `clusterauthorizer.AzCredentials(...)` on every reconciliation. `credentialsGetter` simply makes it swappable so that we can mock the whole thing (`clusterauthorizer.AzCredentials` itself is tested separately).

Can we do it differently? Yes. (e.g. some sort of `Setup` method as you suggested). Does it make sense? I don't know... Probably not.

> Here if the servicePrincipal had a `Setup` function instead of a constructor (because the fields are private), we could just use the interface and avoid having this extra function field. and then we could just use a common mock for all the places where we use it to test it

Do you mean that you would prefer having validator as a field in the `checker` and do a `Setup` call on every reconcile? Yes, that would work, but you don't have to expose a sepcific type for this: `Setup` can be part of your interface.

However I personally wouldn't support this appraoch since this introduces unnecessary state management inside of the validaotr.

In my opinion swappable constructor which returns a disposable validator instance is a better appraoch than having a validator hanging around for the lifetime of the controller and having to deal with credentials replacement. I think it is harder to modify code which deals with the state like this: easier to introduce a regression.

> also, you chose a naming that is for interfaces, functions names should be `getCredentials` and `constructSpValidator` or `newSpValidator`

I agree that `newSpValidator` is probably better. `credentialsGetter` should probably just become `credentials` (so that we follow [this advice](https://go.dev/doc/effective_go#Getters)).

I would like to update it in a follow up PR so we don't have to wait for E2E here just because of naming, if that's ok with you?

---

Off topic:

> why don't we want to expose those types? As long as the fields in it are private, there isn't really any need to hide those fields.

But why do you want to do that? You have a contract between packages (an interface). Why should you care about a specific type/implementation details?

>  I think this is just a bad design on the dynamic part

I don't think so. Maybe I miss something, but I don't believe that returning an interface is a bad design. Your package provides an interface and consumers (even withing the same codebase) shouldn't care about the implementation details. You might want to swap the implementation one day or add another one which meets the interface and you don't want to update consuming code because of that.

pkg/operator/controllers/checkers/serviceprincipalchecker/controller_test.go
m1kola: @facchettos I think with this approach `mockChecker` would have to be defined within nested `t.Run` somehow. Or you would have to store `t *testing.T` inside of `mockChecker` struct. Otherwise `t.Error` will be giving you wrong context in case of failure.

So instead of defining the whole type inside `t.Run` I just define an anonymous function and use `http.HandlerFunc` pattern.

I think what we have right now in this PR is marginally better than struct approach.

pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go
facchettos: https://github.com/golang/go/wiki/CodeReviewComments#interfaces this says to return concrete types instead of interfaces 
```The implementing package should return concrete (usually pointer or struct) types: that way, new methods can be added to implementations without requiring extensive refactoring.```

facchettos: when I was saying that we should have a `Setup` function on the struct, I was thinking of adding it to the `ServicePrincipalValidator` interface. But that would require to not have the `dynamic` struct implementing all the interfaces at once. (to be completely honest, I don't understand why it is done this way, it makes it very cumbersome)

pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go
m1kola: > but this is just a bad pattern though

I don't agree. Ideally no one should be concerned about underlying implementation inside of the package hence we should keep stuff non-exported untill there is a really good reason for this.

> just the other day I had to fix a bunch of frontend typecasting because NewFrontend didn't return a frontend struct but an interface type.

I would like to look into this with you. You are probably talking about our front-end unit tests which are more like integration tests. Yes, we have to cast into a specific type there, but it is not the issue with the pattern. It is an issue with the tests. Hapyp to discuss this.

facchettos (approved):
to avoid rebasing hell, approving for now. we are working on fixing the concerns tomorrow

PR #2589, opened 2022-12-09
Operator checker split - move DNS checker into a separate controller

m1kola:

Some changes:
* When `default` DNS object is missing we now set condition to `False` as we expect it to exist.
* Now we watch for the changes in the DNS object. This should give us feedback faster (no need to wait for 1 hour in the worst case scenario)

No, refactoring/optimisation.


pkg/operator/controllers/checkers/clusterdnschecker/controller.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Reconcile will keep checking that the has a valid cluster service principal.
+func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	instance, err := r.arocli.AroV1alpha1().Clusters().Get(ctx, arov1alpha1.SingletonClusterName, metav1.GetOptions{})
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if !instance.Spec.OperatorFlags.GetSimpleBoolean(checkercommon.ControllerEnabled) {
+		r.log.Info("controller is disabled")
```
m1kola, line 65:
```suggestion
		r.log.Debug("controller is disabled")
```

Same in other places

facchettos (approved):

PR #2593, opened 2022-12-12
Update redeployvm e2e test to use unique name for each uptime pod

tsatam:

[ADO task](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15682807)

Adds a random, unique suffix to the end of the names of pods deployed during the adminapi_redeployvm E2E test in order to check uptime. These pods are not deleted immediately, causing them to conflict if the test deploys a second uptime pod before the first is finished deleting.

### Test plan for issue:

Run the E2E test under question and ensure it still passes/fails as expected 

No (E2E flake reduction)

### Note

I have discovered additional flakiness in this test that I do not believe to be related. When running these tests against the same cluster (e.g. in a local dev environment), subsequent runs after the first passing one will fail, triggering Ginkgo's automatic retry which passes and marks the test as flaky. Further investigation/fixes for this other issue will be the subject of a future change, not this PR.

### DIFF
--- a/test/e2e/adminapi_redeployvm.go
+++ b/test/e2e/adminapi_redeployvm.go
@@ -6,6 +6,7 @@ package e2e
 import (
 	"bufio"
 	"context"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
@@ -17,6 +18,7 @@ import (
 	mgmtcompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilrand "k8s.io/apimachinery/pkg/util/rand"
 
 	"github.com/Azure/ARO-RP/pkg/util/ready"
 	"github.com/Azure/ARO-RP/pkg/util/stringutils"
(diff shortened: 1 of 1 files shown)


SudoBrendan (approved):
LGTM!

cadenmarchese (approved):

test/e2e/adminapi_redeployvm.go
```diff
@@ -80,7 +82,7 @@ var _ = Describe("[Admin API] VM redeploy action", func() {
 func getNodeUptime(g Gomega, ctx context.Context, node string) (time.Time, error) {
 	// container kernel = node kernel = `uptime` in a Pod reflects the Node as well
 	namespace := "default"
-	name := node
+	name := fmt.Sprintf("%s-uptime-%s", node, utilrand.String(5))
```
cblecker, line 85: Instead of using a rand function, can we leverage `GinkgoParallelProcess()` to make it unique but deterministic? https://onsi.github.io/ginkgo/#patterns-for-testing-against-singletons

cblecker (changes requested):

test/e2e/adminapi_redeployvm.go
tsatam: Updated to use `GinkgoParallelProcess()`

cblecker (approved):

PR #2594, opened 2022-12-12
Added frontendResultLog for enrichment

carvalhe:
Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/16179270

Currently frontend logs do not give out much information for QoS. The returned frontend response says "sent response" with not much detail if there was an error, and if there is, it's just an error code. This PR creates a separate result log that will look similar to the one created for backend. It sends a message stating "Succeeded" or "Failed", and marks it as a QoS log. It also sends the error information as well as the operation type being done from the http request method. For now, this is only being used on openshiftcluster put/patch, get, and delete. Potentially other places in the future. With these changes, we can easily measure the health by checking what's succeeding and what's failing. 

### Test plan for issue:
error/log based changes


pkg/frontend/frontend.go
```diff
@@ -503,3 +504,37 @@ func reply(log *logrus.Entry, w http.ResponseWriter, header http.Header, b []byt
...
+				log = log.WithField("resultType", utillog.UserErrorResultType)
+			} else {
+				log = log.WithField("resultType", utillog.ServerErrorResultType)
+			}
+		default:
+			log = log.WithField("resultType", utillog.ServerErrorResultType)
+		}
+	}
+
+	log = log.WithField("errorDetails", err.Error())
```
s-amann, line 538: I'm not too familiar with this log api, is this appending the error in addition to the `operationType` etc set on line 509?

carvalhe: Yes, so the errorDetails will be blank until it gets to this point and will append it to the log. Similarly done [here ](https://github.com/Azure/ARO-RP/blob/2ca94a547ffa2522da45967abbd400b35a6c3ce8/pkg/backend/openshiftcluster.go#L340) for the back end.

s-amann (dismissed):

pkg/frontend/frontend.go
s-amann: very cool, thanks for the additional example!

pkg/frontend/frontend.go
```diff
@@ -503,3 +504,37 @@ func reply(log *logrus.Entry, w http.ResponseWriter, header http.Header, b []byt
...
+		"resultType":    utillog.SuccessResultType,
+		"operationType": method,
+	})
+
+	if err == nil {
+		log.Info("front end operation succeeded")
+		return
+	}
+
+	if err != nil {
```
darthhexx, line 520: We don't need this check, because of the check above, we know it's not `nil`.

pkg/frontend/frontend.go
```diff
@@ -503,3 +504,37 @@ func reply(log *logrus.Entry, w http.ResponseWriter, header http.Header, b []byt
...
+		switch err := err.(type) {
+		case *api.CloudError:
+			log = log.WithField("resultType", utillog.UserErrorResultType)
+		case statusCodeError:
+			if int(err) < 300 {
```
darthhexx, line 525: I've not tested a timeout (status code 0), or whether we'd ever encounter one here, but it would be treated as success the way I read the code. Just something to consider.

darthhexx (changes requested):
Small nit and a logic check on timeouts.

pkg/frontend/frontend.go
carvalhe: Removed unneeded if statement

pkg/frontend/frontend.go
carvalhe: Changed it to 200 <= X < 300 so reduce possible errors that might come up as you mentioned

darthhexx (approved):

darthhexx:
@carvalhe please rebase this PR so I can retest

PR #2597, opened 2022-12-14
Ensure monitor captures the deletion

yehornaumenko:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/12371799/

[Google doc](https://docs.google.com/document/d/1x6VPMWaD_hADHB3UOsksSwc7dJAuzbd5LhekFXgRXkc/edit?usp=sharing)

`docs/monitoring.md`

### DIFF
--- a/docs/monitoring.md
+++ b/docs/monitoring.md
@@ -31,6 +31,12 @@ monitoring from inside the cluster as well as a complementary near-term goal.
   locally (like k8s list/watch).  At startup, the cosmos DB change feed returns
   the current state of all of the OpenShiftClusterDocuments; subsequently as
   OpenShiftClusterDocuments it returns the updated documents.
+* At the moment of writing, the change feed does not log record deletions. It logs
+  only changes. Deallocated clusters are deleted from the monitoring list only if
+  they were seen in the `DeletingProvisioningState` by the monitor.
+  The monitor reads the change feed every 10 seconds, so we should except exclude
+  cases when `OpenShiftClusterDocuments` have the `DeletingProvisioningState` less
+  than 10 seconds.
 * Each monitor aims to check each cluster it "owns" every 5 minutes; it walks
   the local database map and distributes checking over lots of local goroutine
   workers.
--- a/pkg/backend/openshiftcluster.go
+++ b/pkg/backend/openshiftcluster.go
@@ -172,6 +172,7 @@ func (ocb *openShiftClusterBackend) handle(ctx context.Context, log *logrus.Entr
 
 	case api.ProvisioningStateDeleting:
 		log.Print("deleting")
+		t := time.Now()
 
 		err = m.Delete(ctx)
 		if err != nil {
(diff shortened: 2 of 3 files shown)


jhoreman:
Hey Yehor, thanks for this. Can we include the work item associated with this PR? Would like to know more about the reasoning and goals behind this change.

yehornaumenko:
Hi @jhoreman. I have updated the description of this PR, could you check it, please?

jhoreman:
> Hi @jhoreman. I have updated the description of this PR, could you check it, please?

Perfect! Thanks for adding all this detail.

pkg/backend/openshiftcluster.go
```diff
@@ -185,6 +186,11 @@ func (ocb *openShiftClusterBackend) handle(ctx context.Context, log *logrus.Entr
 
 		stop()
 
+		// This Sleep ensures that the monitor has enough time
+		// to capture the deletion (by reading from the changefeed)
+		// and stop monitoring the cluster.
+		// TODO: Provide better communication between RP and Monitor
+		time.Sleep(time.Until(t.Add(time.Second * 20)))
```
darthhexx, line 193: Adding 20 seconds to _every_ cluster delete is not a great customer experience to introduce for this edge case. Ideally we would follow the [recommended delete approach of using a TTLs](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/change-feed-design-patterns#deletes), 10 seconds in our case, but due to use-cases where users are creating and tearing down clusters using the same name/RG/VNET, etc, we may run into issues when the TTL has not expired yet.

To limit the effect of this change, wouldn't it make sense to add a `deleteStartedAt := time.Now()` variable before `err = m.Delete(ctx)`, and then only sleep here if the delete has taken less than 10 seconds? Or 15 to be absolutely certain.

darthhexx (changes requested):

pkg/backend/openshiftcluster.go
darthhexx: Ignore my first comment. I see that you have added a conditional sleep on the deletion start time.

darthhexx (approved):

pkg/backend/openshiftcluster.go
darthhexx: Ignore my previous comment, I see you have added a conditional sleep using the deletion start time.

petrkotas (approved):
@yehornaumenko perfect, this is exactly the good enough for now solution, we agreed upon. It will allow us to resolve some sev2.

In the future, we can improve the deletion mechanism.

docs/monitoring.md
```diff
@@ -31,6 +31,12 @@ monitoring from inside the cluster as well as a complementary near-term goal.
   locally (like k8s list/watch).  At startup, the cosmos DB change feed returns
   the current state of all of the OpenShiftClusterDocuments; subsequently as
   OpenShiftClusterDocuments it returns the updated documents.
+* At the moment of writing, the change feed does not log record deletions. It logs
+  only changes. Deallocated clusters are deleted from the monitoring list only if
+  they were seen in the `DeletingProvisioningState` by the monitor.
+  The monitor reads the change feed every 10 seconds, so we should except exclude
```
facchettos, line 37:
```suggestion
  The monitor reads the change feed every 10 seconds, so we should expect exclude
```

facchettos (approved):
lgtm, little typos but that can be fixed later

PR #2599, opened 2022-12-14
Operator checker split - move ingress certificate checker into a separate controller

m1kola:

TBD

No, refactoring/optimisation.


pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
```diff
@@ -0,0 +1,64 @@
...
+
+	ingress, err := r.operatorcli.OperatorV1().IngressControllers("openshift-ingress-operator").Get(ctx, "default", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	if ingress.Spec.DefaultCertificate == nil {
+		// TODO: fix constant reconciliation.
+		// This causes constant reconciliation in dev where we do not have default cert set.
+		// In the current implementation we seem to always set condition to False in dev,
```
m1kola, line 53: I think we also set it to false when there is a custom domain in use because of this bit:

https://github.com/Azure/ARO-RP/blob/340ebb161f590eaf4ee069d519e6cfdd9d5c5f01/pkg/cluster/tls.go#L29-L36

facchettos (approved):

PR #2602, opened 2022-12-15
override the api url used by hive with the internal api

s-amann:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/16592928

### Test plan for issue:

manual testing on hive

no, we need to create some detailed hive documentation.

### DIFF
--- a/pkg/hive/install.go
+++ b/pkg/hive/install.go
@@ -10,6 +10,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	hivev1azure "github.com/openshift/hive/apis/hive/v1/azure"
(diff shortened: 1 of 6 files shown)


pkg/hive/install.go
```diff
@@ -195,6 +214,7 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 			},
 			ControlPlaneConfig: hivev1.ControlPlaneConfigSpec{
 				APIServerIPOverride: doc.OpenShiftCluster.Properties.NetworkProfile.APIServerPrivateEndpointIP,
+				APIURLOverride:      "api-int." + installConfig.Config.ObjectMeta.Name + "." + installConfig.Config.BaseDomain + ":6443",
```
hawkowl, line 217: Can't we do this:
```
	domain := m.doc.OpenShiftCluster.Properties.ClusterProfile.Domain
	if !strings.ContainsRune(domain, '.') {
		domain += "." + m.env.Domain()
	}
```

I would like to not read the graph if we can absolutely help it.

pkg/hive/install.go
```diff
@@ -166,6 +171,20 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
...
+	resourceGroup := stringutils.LastTokenByte(doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
+	account := "cluster" + doc.OpenShiftCluster.Properties.StorageSuffix
+
+	pg, err := c.graph.LoadPersisted(context.TODO(), resourceGroup, account)
+	if err != nil {
+		return nil, err
+	}
+
+	var installConfig *installconfig.InstallConfig
+	var kubeadminPassword *password.KubeadminPassword
```
hawkowl, line 183: Is this used?

PR #2603, opened 2022-12-16
Hive: Set the APIURLOverride with the internal API URL

darthhexx:

Fixes certificate validation errors when Hive is communicating with the Cluster API server.

Hive can't communicate with the Cluster API server if we don't.

### Test plan for issue:

Created a test cluster.

None yet, but we need Hive debug documentation.

### DIFF
--- a/pkg/hive/install.go
+++ b/pkg/hive/install.go
@@ -6,10 +6,12 @@ package hive
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"path"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	hivev1azure "github.com/openshift/hive/apis/hive/v1/azure"
@@ -167,6 +169,11 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 		}
 	}
 
+	clusterDomain := doc.OpenShiftCluster.Properties.ClusterProfile.Domain
+	if !strings.ContainsRune(clusterDomain, '.') {
+		clusterDomain += "." + os.Getenv("DOMAIN_NAME")
+	}
+
 	// Do not set InfraID here, as Hive wants to do that
 	return &hivev1.ClusterDeployment{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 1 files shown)


pkg/hive/install.go
```diff
@@ -167,6 +169,11 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 		}
 	}
 
+	clusterDomain := doc.OpenShiftCluster.Properties.ClusterProfile.Domain
+	if !strings.ContainsRune(clusterDomain, '.') {
+		clusterDomain += "." + os.Getenv("DOMAIN_NAME")
+	}
+
```
cblecker, line 176: This bit of code is in our code base so many times. It would be great if we could extract it into a function and apply some testing to the result (like a regex match to ensure it's a properly formed FQDN).

pkg/hive/install.go
```diff
@@ -167,6 +169,11 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 		}
 	}
 
+	clusterDomain := doc.OpenShiftCluster.Properties.ClusterProfile.Domain
+	if !strings.ContainsRune(clusterDomain, '.') {
+		clusterDomain += "." + os.Getenv("DOMAIN_NAME")
```
cblecker, line 174:
```suggestion
		if envDomainName, ok := os.LookupEnv("DOMAIN_NAME"); ok {
			clusterDomain += "." + envDomainName
		} else {
			// return some sort of error if the env var is unset
			// this ensures we don't return this with an empty string
		}
```

I think I'd still feel more comfortable if we were using `m.env.Domain()` so that we had a standardized call and that one already has error checking too, but I'm okay with this.

pkg/hive/install.go
```diff
@@ -195,6 +202,7 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 			},
 			ControlPlaneConfig: hivev1.ControlPlaneConfigSpec{
 				APIServerIPOverride: doc.OpenShiftCluster.Properties.NetworkProfile.APIServerPrivateEndpointIP,
+				APIURLOverride:      fmt.Sprintf("api-int.%s:6443", clusterDomain),
```
cblecker, line 205: might be overkill, but
```suggestion
				APIURLOverride:      net.JoinHostPort("api-int." + clusterDomain, "6443"),
```

cblecker (approved):
non-blocking suggestions. I am fine with this as is, but also, I'd love to see this made more robust in the future. Maybe calls for a `//TODO`? Or something we can clean up in the tech debt pay down?

pkg/hive/install.go
darthhexx: Definitely makes sense, but we should do so in a follow up PR.

pkg/hive/install.go
darthhexx: The Hive managed doesn't have access to `Domain()` so we'll need to refactor it a bit. Since we want to get this deployed to test before the deploy freeze, I'd say that we should do that in a follow up PR as well.

pkg/hive/install.go
darthhexx: [Created a WI](https://dev.azure.com/msazure/AzureRedHatOpenShift/_workitems/edit/16616511) for follow-up work.

PR #2605, opened 2022-12-19
Reduces the number of non-cached API server calls

m1kola:

This PR is an attempt to optimise in-cluster operator. I need to properly mesure it, but as far as I can see from the code we currently send a burst of 26 requests to the API just to get `cluster/cluster` object.

I think we can get rid of most of them (if not all) by using a split client (which reads from a shared cache and writes into the API server).

Ideally I would like to replace other versioned clients in the operator with this client, but the first step is to get rid of calls like this:

```go
instance, err := r.arocli.AroV1alpha1().Clusters().Get(ctx, arov1alpha1.SingletonClusterName, metav1.GetOptions{})
if err != nil {
}
``` 

**Next steps (not in this PR):**

* Get rid of versioned ARO client dependecy [here](https://github.com/Azure/ARO-RP/blob/d6b3c3c87b1b2ed88f1eee61aa13b771a200e49f/pkg/util/conditions/condition.go#L24-L48). This will enable to remove `arocli` argument/dependency in rest of the controllers (all checker controllers and machine controller)
* Replace rest of the versioned clients in the operator with a split client.

### Test plan for issue:

E2E should cover most of it, but I will also be manually triggering reconciliation in some cases.

### DIFF
--- a/pkg/operator/controllers/alertwebhook/alertwebhook_controller.go
+++ b/pkg/operator/controllers/alertwebhook/alertwebhook_controller.go
@@ -20,7 +20,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
-	aroclient "github.com/Azure/ARO-RP/pkg/operator/clientset/versioned"
 )
 
 const (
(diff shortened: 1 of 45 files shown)


pkg/operator/controllers/alertwebhook/alertwebhook_controller.go
```diff
@@ -35,21 +34,22 @@ var alertManagerName = types.NamespacedName{Name: "alertmanager-main", Namespace
...
-		arocli:        arocli,
 		kubernetescli: kubernetescli,
 	}
 }
 
 // Reconcile makes sure that the Alertmanager default webhook is set.
 func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
-	instance, err := r.arocli.AroV1alpha1().Clusters().Get(ctx, arov1alpha1.SingletonClusterName, metav1.GetOptions{})
+	instance := &arov1alpha1.Cluster{}
+	err := r.client.Get(ctx, types.NamespacedName{Name: arov1alpha1.SingletonClusterName}, instance)
```
s-amann, line 52: Since we can't guarantee at the time of construction in of any of these controllers its ability to get the cluster, should we be introducing a guard statement around if `r.client` was populated before attempting to utilize it?

m1kola: `InjectClient` will be called by the builder (down the `builder.Named(ControllerName).Complete(r)` call stack). So at the moment of reconciliation there will always be a client here: 

And this is how the docs suggests to set the client in this [example](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/builder#example-Builder).

However thi is very implicit and I prefer being explicit. I think there might be a way to explicitly pass the client into the constructor. I was thinking - `ctrl.Manager` should already have an instance of a client and we create it before calling any of the constructors, so we should be able `mgr.GetClient()` and pass into the constructor. But I didn't have time to dig into this and I don't fully understand implications so I decided to follow an example from the docs for now.

s-amann: Thank you for sharing this, I agree to start we should follow the example and enhance as we see fit down the line.

s-amann (approved):

petrkotas (commented):
One small question, otherwise looks good to me.

PR #2606, opened 2022-12-21
systemData is initially NIL on ClusterManagerConfigurationDocument

s-amann:

Fixes https://issues.redhat.com/browse/ARO-1634

When the clustermanager put or patch API is called and the ClusterManagerConfigurationDocument does not already have the appropriate SystemData initialized the code will panic.

### Test plan for issue:

unit tests

no, documentation on the OCM API usage has yet to be created AFIAK.

### DIFF
--- a/pkg/frontend/clustermanager_putorpatch_test.go
+++ b/pkg/frontend/clustermanager_putorpatch_test.go
@@ -244,7 +244,9 @@ func TestEnrichSyncSetSystemData(t *testing.T) {
 
 	enrichSyncSetSystemData(&doc, &populatedSystemData)
 
-	compareSystemData(t, *doc.SyncSet.SystemData, populatedSystemData)
+	if err := compareSystemData(*doc.SyncSet.SystemData, populatedSystemData); err != nil {
+		t.Fatal(err)
+	}
 }
 
 func TestEnrichSyncIdentityProviderSystemData(t *testing.T) {
@@ -254,7 +256,9 @@ func TestEnrichSyncIdentityProviderSystemData(t *testing.T) {
 
 	enrichSyncIdentityProviderSystemData(&doc, &populatedSystemData)
 
-	compareSystemData(t, *doc.SyncIdentityProvider.SystemData, populatedSystemData)
+	if err := compareSystemData(*doc.SyncIdentityProvider.SystemData, populatedSystemData); err != nil {
+		t.Fatal(err)
+	}
 }
 
 func TestEnrichMachinePoolSystemData(t *testing.T) {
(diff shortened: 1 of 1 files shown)


jhoreman (dismissed):
Thanks for this, Spencer. Straightforward and looks good to me!

pkg/frontend/clustermanager_putorpatch_test.go
```diff
@@ -226,3 +227,73 @@ func TestPutOrPatchClusterManagerConfiguration(t *testing.T) {
...
+	}
+
+	enrichSecretSystemData(&doc, &populatedSystemData)
+
+	compareSystemData(t, *doc.Secret.SystemData, populatedSystemData)
+}
+
+func compareSystemData(t *testing.T, docSystemData, expectedSystemData api.SystemData) {
+	if docSystemData.CreatedAt == nil || docSystemData.CreatedAt != expectedSystemData.CreatedAt {
+		t.Fatalf("CreatedAt was %q expected %q", docSystemData.CreatedAt, expectedSystemData.CreatedAt)
```
cadenmarchese, line 282: Suggestion for readability, but totally optional - might make more sense to fail the test inside of the `Test` functions themselves rather than in this external one. We would still probably want to return these specific errors rather than one single nonspecific error, and then fail the tests if `err != nil`.

cadenmarchese (dismissed):

pkg/frontend/clustermanager_putorpatch_test.go
s-amann: I'll make this change, thanks for the suggestion

pkg/frontend/clustermanager_putorpatch_test.go
```diff
@@ -226,3 +227,82 @@ func TestPutOrPatchClusterManagerConfiguration(t *testing.T) {
...
+	}
+
+	enrichSecretSystemData(&doc, &populatedSystemData)
+
+	if err := compareSystemData(t, *doc.Secret.SystemData, populatedSystemData); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func compareSystemData(t *testing.T, docSystemData, expectedSystemData api.SystemData) error {
```
cadenmarchese, line 288: Do we still need `t`?
```suggestion
func compareSystemData(docSystemData, expectedSystemData api.SystemData) error {
```

s-amann: doh! I'll remove this, thanks.

cadenmarchese (approved):

PR #2607, opened 2022-12-22
feat: add gatekeeper policies for pull-secret delete deny

ArrisLee:
https://issues.redhat.com/browse/ARO-1393?workflowName=OJA-WF-S&stepId=5

Currently the customer can run oc -n openshift-azure-operator delete secret cluster and break ARO operator: it will not be able to reconcile a deleted cluster pull secret: `oc -n openshift-config delete secret pull-secret`.
 
The secret is comming from RP and we can not reconcile it inside of the cluster, but the cluster is still recoverable using admin update (requires an SRE action).
 
We can try and make it harder to delete it. We need to invesitage and implement (if possible) a way to do this.

Test plan for issue:
Unit test with Gator
Acceptance test with local RP and dev cluster

N/A


pkg/operator/controllers/guardrails/gktemplates/aro-deny-delete.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: templates.gatekeeper.sh/v1beta1
```
yjst2012, line 1: version has been upgraded to templates.gatekeeper.sh/v1 for gatekeeper 3.10.x

pkg/operator/controllers/guardrails/gktemplates/aro-deny-delete.yaml
```diff
@@ -0,0 +1,31 @@
...
+kind: ConstraintTemplate
+metadata:
+  name: aro-deny-delete
```
yjst2012, line 4: the name here should be all lower case of spec.crd.spec.names.kind, ie, arodenydelete, as required by gatekeeper

pkg/operator/controllers/guardrails/gktemplates/aro-deny-delete.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - target: admission.k8s.gatekeeper.sh
+    rego: |
+      package arodenydelete
+
+      deny[msg] {
+        input.review.operation == "DELETE"
+        deny_resource_name := input.parameters.name
+        deny_resource_name == input.review.object.metadata.name
+        msg = "Deleting resource - " + name + " is not allowed"
+      }
```
yjst2012, line 31: got some error messages like:
invalid ConstraintTemplate: invalid rego: invalid module: missing required rules: [violation] 

try modified some places and finally get it work:
```
    rego: |
      package arodenydelete
      violation[{"msg": msg}] {
        input.review.operation == "DELETE"
        deny_resource_name := input.parameters.name
        deny_resource_name == input.review.object.metadata.name
        msg := sprintf("Deleting resource - %v is not allowed", [deny_resource_name])
      }
```
`$ oc delete secrets pull-secret  -n openshift-config
Error from server (Forbidden): admission webhook "validation.gatekeeper.sh" denied the request: [aro-pull-secret-deny-delete] Deleting resource - pull-secret is not allowed
`

ArrisLee: thanks for the correction 👍🏼

pkg/operator/controllers/guardrails/gktemplates/aro-deny-delete.yaml
```diff
@@ -0,0 +1,31 @@
...
+  targets:
+  - target: admission.k8s.gatekeeper.sh
+    rego: |
+      package arodenydelete
+
+      violation[{"msg": msg}] {
+        input.review.operation == "DELETE"
+        deny_resource_name := input.parameters.name
+        deny_resource_name == input.review.object.metadata.name
+        msg := "Deleting resource - " + name + " is not allowed"
```
yjst2012, line 30: this does not work:
```
template:6: rego_type_error: plus: invalid argument(s)
	have: (number, string, ???)
	want: (x: number, y: number, z: number)
template:6: rego_type_error: plus: invalid argument(s)
	have: (string, any, ???)
	want: (x: number, y: number, z: number) 
```
and has to be something like:
`        msg := sprintf("Deleting resource - %v is not allowed", [deny_resource_name])
`

PR #2608, opened 2022-12-22
operator feature: check ingress certificate and cluster domain

AldoFusterTurpin:

Fixes [Improve detection of cases for ingress certificate](https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/15990932)

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go contains code that currently detects if the default cert is specified in the ingress controller. But this check does not distinguish between the managed domain case and custom domain case.

This PR takes into account if we have a manage domain or not and report the error accordingly.
The [design doc](https://docs.google.com/document/d/13KgQOvUv3rV2vImzop74ouefhza7Owwk5K1bXoQLVR8/edit?usp=sharing) with the logic.

### Test plan for issue:

Unit Tests added.

N/A


pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
```diff
@@ -59,13 +62,41 @@ func (r *checker) Check(ctx context.Context) error {
...
+
+func certificateNameIsValid(certificateName string, clusterID configv1.ClusterID) bool {
+	ingressNameSuffix := "-ingress"
+	expectedCertName := string(clusterID) + ingressNameSuffix
+	return certificateName == expectedCertName
+}
+
+func clusterHasManagedDomain(clusterDomain string) bool {
+	managedDomainSuffix := "aroapp.io"
+	return strings.HasSuffix(clusterDomain, managedDomainSuffix)
```
facchettos, line 101: the variable isn't necessary here, at least not at this scope. maybe we should have a const or just `return strings.HasSuffix(clusterDomain, "aroapp.io")`

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
```diff
@@ -59,13 +62,41 @@ func (r *checker) Check(ctx context.Context) error {
...
+	cluster, err := r.arocli.AroV1alpha1().Clusters().Get(ctx, arov1alpha1.SingletonClusterName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	return validateCertificate(cv.Spec.ClusterID, ingress.Spec.DefaultCertificate, cluster.Spec.Domain)
+}
+
+func validateCertificate(clusterId configv1.ClusterID, certificate *corev1.LocalObjectReference, clusterDomain string) error {
+	clusterHasManagedDomain := clusterHasManagedDomain(clusterDomain)
```
gvanderpotte, line 74: Is not this redefining the logic in dns.ManagedDomain ?

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
```diff
@@ -59,13 +62,41 @@ func (r *checker) Check(ctx context.Context) error {
...
 		return errNoDefaultCertificate
 	}
 
-	if ingress.Spec.DefaultCertificate.Name != string(cv.Spec.ClusterID)+"-ingress" {
-		return fmt.Errorf("custom ingress certificate in use: %q", ingress.Spec.DefaultCertificate.Name)
+	certificateName := certificate.Name
+	certificateNameIsValid := certificateNameIsValid(certificateName, clusterId)
+
+	if clusterHasManagedDomain && !certificateNameIsValid {
+		return fmt.Errorf("custom ingress certificate in use: %q", certificateName)
```
gvanderpotte, line 87: Not sure if we should log an error message with the certificate name. That's not specifically secret, but the name of the cert does not really matter, as long as it is not the ARO provisionned one

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
```diff
@@ -59,13 +62,41 @@ func (r *checker) Check(ctx context.Context) error {
...
 	}
 
 	return nil
 }
+
+func certificateNameIsValid(certificateName string, clusterID configv1.ClusterID) bool {
+	ingressNameSuffix := "-ingress"
+	expectedCertName := string(clusterID) + ingressNameSuffix
```
facchettos, line 95: that's a little too much here `string(clusterID) + "-ingress"` would be sufficient

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
```diff
@@ -59,13 +62,41 @@ func (r *checker) Check(ctx context.Context) error {
...
+
+	return validateCertificate(cv.Spec.ClusterID, ingress.Spec.DefaultCertificate, cluster.Spec.Domain)
+}
+
+func validateCertificate(clusterId configv1.ClusterID, certificate *corev1.LocalObjectReference, clusterDomain string) error {
+	clusterHasManagedDomain := clusterHasManagedDomain(clusterDomain)
+
+	if certificate == nil {
+		if clusterHasManagedDomain {
+			return fmt.Errorf("ARO has not provisioned the default ingress certificate because it is empty and cluster has a managed domain")
```
facchettos, line 78: does it ever happen? wouldn't we have errors beforehand? like, errors when we create the cluster?

facchettos (dismissed):

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
gvanderpotte: I think it is necessary, because either:
- if ARO could fail at provisionning the certifcate during the 			steps.Action(m.configureIngressCertificate) step in installation.  Cluster installation will fail (provisionning failed), but the cluster would still be there and mostly operational (this step is one of the last steps). This is an ARO problem in that case.
- customer could have deleted the cert. You are right, that one is also a possibility. Maybe we should mention the two possibilities

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
AldoFusterTurpin: Good point.
[dns.ManagedDomain()](https://github.com/Azure/ARO-RP/blob/master/pkg/util/dns/dns.go#L193) expects as a parameter the whole env.Interface which is overkill, so after your comment, I thought it would be better to move the function which checks if a clusterDomain is managed to the dns package and make it public and directly unit-tested.

Tks!

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
AldoFusterTurpin: Good point. It was already like this before this PR. Changing it now. 
Tks!

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
AldoFusterTurpin: I don't like magic hardcoded strings and as the string is just used once, declaring the variable inside the short function was enough for me.
Going for the const option now as you suggest  😄

Note that I have also moved it to the dns package and tested the public function so it is reusable.
Tks!

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
AldoFusterTurpin: Declaring ingressNameSuffix as a package constant as I don't like magic hardcoded strings.
Tks!

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
AldoFusterTurpin: As far as I understood when talking with @petrkotas, we have to differentiate those errors into different types because one case is the customer's fault and the other is ARO's fault. Even if the procedure is the same, we will have a better understanding of the situation). 

Agree on the rewording of the error message.
Thank you both!

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
```diff
@@ -59,13 +62,35 @@ func (r *checker) Check(ctx context.Context) error {
...
+
+	return validateCertificate(cv.Spec.ClusterID, ingress.Spec.DefaultCertificate, cluster.Spec.Domain)
+}
+
+func validateCertificate(clusterId configv1.ClusterID, ingressCertificate *corev1.LocalObjectReference, clusterDomain string) error {
+	isManagedDomain := dns.IsManagedDomain(clusterDomain)
+
+	if ingressCertificate == nil {
+		if isManagedDomain {
+			return errors.New("missing ingress certificate for cluster with managed domain")
```
facchettos, line 78: would be slightly better to extract it into a var at package level so you can reuse it in the test

pkg/operator/controllers/checkers/ingresscertificatechecker/controller_test.go
```diff
@@ -58,7 +58,7 @@ func TestReconcile(t *testing.T) {
 		{
 			name:                 "no default cert set",
 			wantConditionStatus:  operatorv1.ConditionFalse,
-			wantConditionMessage: "ingress has no default certificate set",
+			wantConditionMessage: "missing ingress certificate for cluster with custom domain",
```
facchettos, line 61: maybe have a extra test case here too with your new error

pkg/operator/controllers/checkers/ingresscertificatechecker/checker_test.go
```diff
@@ -31,37 +34,101 @@ func TestCheck(t *testing.T) {
...
+			ingressController: fakeIngressController(nil),
+			cluster:           nil,
+			wantErr:           "clusters.aro.openshift.io \"cluster\" not found",
+		},
+		{
+			name:              "should return certificate error (ARO's fault) when no certificate is set and we have a managed domain",
+			clusterVersion:    clusterVersion,
+			ingressController: fakeIngressController(nil),
+			cluster:           fakeCluster("whatever.aroapp.io"),
+			wantErr:           "missing ingress certificate for cluster with managed domain",
```
facchettos, line 62: would it make sense to use `errVar.Error()` instead of retyping it? would prevent us to have to change the tests if we want to change the error messages

pkg/operator/controllers/checkers/ingresscertificatechecker/checker_test.go
```diff
@@ -31,37 +34,101 @@ func TestCheck(t *testing.T) {
...
+			wantErr:           "missing ingress certificate for cluster with custom domain",
+		},
+		{
+			name:           "should return error when we have a managed domain and we have an invalid certificate name",
+			clusterVersion: clusterVersion,
+			ingressController: fakeIngressController(&corev1.LocalObjectReference{
+				Name: "fake-custom-name-ingress",
+			}),
+			cluster: fakeCluster("whatever.aroapp.io"),
+			wantErr: `invalid ingress certificate name for cluster with managed domain`,
```
facchettos, line 78: why are your string not always with `"` ?

AldoFusterTurpin: ` was already there, I just changed the content of the string, so I forgot to change it to ", doing it now.
Tks!

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
AldoFusterTurpin: In fact, this is something I asked Mikalai some months ago: "why do we compare the error string in test cases instead of declaring the errors at package level and check that the error is the same". We could use == or errors.Is in test functions instead of hardcoded strings.

So I agree about putting the error at the package level. 
Doing it now.

pkg/operator/controllers/checkers/ingresscertificatechecker/checker_test.go
AldoFusterTurpin: Yes! I 100% agree. I prefer sentinel errors. I was just following the tendency in the repo for consistency 😃 
Tks!

pkg/operator/controllers/checkers/ingresscertificatechecker/controller_test.go
AldoFusterTurpin: Interesting! I already asked this question myself before. 

The key thing is that in this test (controller_test.go) we don't really want to test all error cases because the way the controller handles the error is the same for all the errors that the checker returns. I think that we should test just that the error returned by the checker is well used by the controller, no matter which error it is. (so the use case to test in the controller would be: "when the checker returns any error, the controller can handle it properly").

Having in controller_test.go a test case for each of the errors that the checker returns is redundant with the tests of the checker and a change in the prod code will imply touching 2 different unit tests.
 
I believe that all those test cases should be in the checker as it is the one that creates them and there you want to make sure each of them is well created/returned.

What I mentioned about the controller: in the [function](https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/checkers/ingresscertificatechecker/controller.go#L104), it just checks if the error is not nil, so no specific logic with the error. -> we cover all code paths with just one test cases for the error handling.

I changed the name of the test to better reflect that we just want to test that the controller behaves as expected when the checker returns an error, no matter which error. 

Thanks for pointing it out. 
😃

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
```diff
@@ -59,13 +62,35 @@ func (r *checker) Check(ctx context.Context) error {
...
+		}
+		return errNoCertificateAndCustomDomain
 	}
 
-	if ingress.Spec.DefaultCertificate.Name != string(cv.Spec.ClusterID)+"-ingress" {
-		return fmt.Errorf("custom ingress certificate in use: %q", ingress.Spec.DefaultCertificate.Name)
+	certificateName := ingressCertificate.Name
+	isCertificateNameValid := isCertificateNameValid(certificateName, clusterId)
+
+	if !isCertificateNameValid && isManagedDomain {
```
cadenmarchese, line 86:
```suggestion
	if !isCertificateNameValid(certificateName, clusterId) && isManagedDomain {
```

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
```diff
@@ -59,13 +62,35 @@ func (r *checker) Check(ctx context.Context) error {
...
+		if isManagedDomain {
+			return errNoCertificateAndManagedDomain
+		}
+		return errNoCertificateAndCustomDomain
 	}
 
-	if ingress.Spec.DefaultCertificate.Name != string(cv.Spec.ClusterID)+"-ingress" {
-		return fmt.Errorf("custom ingress certificate in use: %q", ingress.Spec.DefaultCertificate.Name)
+	certificateName := ingressCertificate.Name
+	isCertificateNameValid := isCertificateNameValid(certificateName, clusterId)
```
cadenmarchese, line 84:
```suggestion
```
I think you could leave this variable out, it's clear from the name of the function what it's doing here, and it's only used once in the code below.

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
facchettos: adding up on this, renaming the function name to a statement instead of a question would probably be good too. `certificateIsValid(...)` is easier to read as a boolean

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
petrkotas: I agree on this one. It will simplify the logic a bit.

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
petrkotas: I leave this one on Aldo. We use both in our codebase. I personally do not have strong opinions on the naming.

AldoFusterTurpin: Done. Tks!

pkg/operator/controllers/checkers/ingresscertificatechecker/checker.go
AldoFusterTurpin: Done. Tks!

pkg/util/dns/dns.go
```diff
@@ -18,7 +18,10 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/dns"
 )
 
-const resourceID = "resourceId"
+const (
+	resourceID          = "resourceId"
+	managedDomainSuffix = ".aroapp.io"
+)
```
bennerv, line 24: This doesn't take into account the govcloud domains.  Can we update this to switch on the ARO cluster CR based on which cloud we are in?

For reference, in gov cloud the `managedDomainSuffix` would be: `aroapp.azure.us`

bennerv (dismissed):

pkg/util/dns/dns.go
AldoFusterTurpin: Thanks @bennerv! Good catch. Doing it now.

petrkotas (approved):
Looks good to me. Thanks @AldoFusterTurpin for the thorough work. We will need to wait for e2e to pass, as it fails currently globally.

gvanderpotte:
LGTM - test coverage is good and does what we want

petrkotas (approved):
Thank you @AldoFusterTurpin LGTM and I see all the comments has been addressed.

pkg/operator/controllers/checkers/ingresscertificatechecker/checker_test.go
```diff
@@ -11,81 +11,167 @@ import (
...
-			clusterVersion: clusterVersion,
+			name:           "Check returns error when cluster has a managed domain and certificate name is empty",
+			clusterVersion: fakeClusterVersion(),
 			ingressController: fakeIngressController(&corev1.LocalObjectReference{
-				Name: "fake-custom-name-ingress",
+				Name: "",
 			}),
-			wantErr: `custom ingress certificate in use: "fake-custom-name-ingress"`,
+			cluster: fakeCluster(managedDomain),
+			wantErr: "invalid ingress certificate name for cluster with managed domain",
```
tiguelu, line 93: For consistency with the other tests, can we use the custom error here instead of a string?

```suggestion
			wantErr: errInvalidCertificateAndManagedDomain.Error(),
```

tiguelu (changes requested):

pkg/operator/controllers/checkers/ingresscertificatechecker/checker_test.go
AldoFusterTurpin: Yes, I forgot to change it. 
Thanks!
Done.

tiguelu (approved):
LGTM, thank you Aldo!

petrkotas (approved):
@aldofusterturpin thank you for the PR. An excelent work. Thank you addressing all the comments.

facchettos (approved):
Can you move your helper function to `util/dns` instead of `util/net` in a follow up PR?

AldoFusterTurpin:
> Can you move your helper function to `util/dns` instead of `util/net` in a follow up PR?

[Done](https://github.com/Azure/ARO-RP/pull/2783) 😄 

Thanks!

PR #2609, opened 2023-01-02
added token refresh heartbeat. removed lastrefresh

bizz001:

Fixes https://issues.redhat.com/browse/ARO-1714

Removes existing unused Gateway db token timestamp metric. Adds new heartbeat type metric for token refresh monitoring. This will allow us to create a monitor that detects lapses in token refresh.

### Test plan for issue:

Check that the new metric is correctly emitted.

No

### DIFF
--- a/pkg/dbtoken/refresher.go
+++ b/pkg/dbtoken/refresher.go
@@ -16,6 +16,7 @@ import (
 	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
 	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/metrics"
+	"github.com/Azure/ARO-RP/pkg/util/heartbeat"
 	utilrecover "github.com/Azure/ARO-RP/pkg/util/recover"
 )
 
@@ -33,8 +34,9 @@ type refresher struct {
 
 	lastRefresh atomic.Value //time.Time
 
-	m            metrics.Emitter
-	metricPrefix string
+	m              metrics.Emitter
+	metricPrefix   string
+	tokenRefreshed bool
 }
 
 func NewRefresher(log *logrus.Entry, env env.Core, authorizer autorest.Authorizer, insecureSkipVerify bool, dbc cosmosdb.DatabaseClient, permission string, m metrics.Emitter, metricPrefix string) (Refresher, error) {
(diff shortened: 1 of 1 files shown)


pkg/dbtoken/refresher.go
```diff
@@ -35,6 +36,7 @@ type refresher struct {
 
 	m            metrics.Emitter
 	metricPrefix string
+	refreshCount int
```
s-amann, line 39: should this be managed atomically?

bizz001: We are not interested in the exact value so it might as well be a bool. Also any race conditions between reading the value and setting to 0 would actually make the heartbeat shut off one cycle sooner in case of an issue that starts right when heartbeat triggers, so I think it's fine like this.

s-amann (dismissed):

pkg/dbtoken/refresher.go
s-amann: I like the refactor to a bool, thank you

s-amann (approved):

jhoreman (commented):
Hey Daniel, thanks for this. We use the LastRefresh metric in the Gateway dashboard [1] to determine whether or not the token was indeed refreshed. Is this something that we could potentially implement there as well? We should create the dashboard to monitor this metric for dbtoken as well once this is merged. 

[1] https://portal.microsoftgeneva.com/dashboard/AzureRedHatOpenShiftRP/Gateway/Gateway?overrides=[{%22query%22:%22//*[id%3D%27location%27]%22,%22key%22:%22value%22,%22replacement%22:%22%22}]%20

bizz001:
> Hey Daniel, thanks for this. We use the LastRefresh metric in the Gateway dashboard [1] to determine whether or not the token was indeed refreshed. Is this something that we could potentially implement there as well? We should create the dashboard to monitor this metric for dbtoken as well once this is merged.
> 
> [1] [https://portal.microsoftgeneva.com/dashboard/AzureRedHatOpenShiftRP/Gateway/Gateway?overrides=[{%22query%22:%22//*[id%3D%27location%27]%22,%22key%22:%22value%22,%22replacement%22:%22%22}]%20](https://portal.microsoftgeneva.com/dashboard/AzureRedHatOpenShiftRP/Gateway/Gateway?overrides=%5B%7B%22query%22:%22//*%5Bid%3D%27location%27%5D%22,%22key%22:%22value%22,%22replacement%22:%22%22%7D%5D%20)

Yes, I'll add this to the list of things to update. Thanks for pointing it out.

petrkotas (approved):
LGTM

PR #2610, opened 2023-01-04
removing useless typecasting because of bad constructor

facchettos:
removing useless typecasting because of bad constructor

Fixes


petrkotas (approved):
lgtm

PR #2611, opened 2023-01-04
aad session expiration simplification

facchettos:
simplifies the session timeout process. Instead of storing a whole struct into the session cookies, now only store an integer which avoids to use `gob` and sending whole struct in cookies

### DIFF
--- a/pkg/portal/middleware/aad.go
+++ b/pkg/portal/middleware/aad.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"crypto/rsa"
 	"crypto/x509"
-	"encoding/gob"
 	"errors"
 	"io"
 	"net/http"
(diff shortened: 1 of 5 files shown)


m1kola (approved):
I think it looks good.

Just to double check: are we going to be ok if someone will send an old cookie value?

pkg/portal/middleware/aad.go
```diff
@@ -157,8 +153,8 @@ func (a *aad) AAD(h http.Handler) http.Handler {
 			return
 		}
 
-		expires, ok := session.Values[SessionKeyExpires].(time.Time)
-		if !ok || expires.Before(a.now()) {
+		expires, ok := session.Values[SessionKeyExpires].(int64)
+		if !ok || time.Unix(expires, 0).Before(a.now()) {
 			h.ServeHTTP(w, r)
 			return
 		}
```
facchettos, line 160: @m1kola if someone sends an old value, we are treating it as not logged in and we will redirect them to /api/login from what I can see

PR #2614, opened 2023-01-05
create test functions for error comparison

AldoFusterTurpin:

Removes duplication of lines used to test errors simplifying how the errors are compared by using an auxiliary function.

This PR simplifies how we assert that the errors we obtain are the expected ones in test functions. I think it is a better idea to have a function to do that instead of copy-pasting the same 4 lines every time we want to compare the errors.
### Test plan for issue:

There are already Unit Tests.

N/A, just cleaned the code.

### DIFF
--- a/pkg/api/v20220904/clustermanager_validatestatic_test.go
+++ b/pkg/api/v20220904/clustermanager_validatestatic_test.go
@@ -4,8 +4,9 @@ package v20220904
 // Licensed under the Apache License 2.0.
 
 import (
-	"fmt"
 	"testing"
+
+	"github.com/Azure/ARO-RP/test/util/matcher"
 )
 
 var ocmResource = string(`
@@ -72,11 +73,7 @@ func TestStatic(t *testing.T) {
 			c := &clusterManagerStaticValidator{}
 
 			err := c.Static(tt.ocmResource, tt.ocmResourceType)
-			if err != nil && tt.wantErr {
-				if fmt.Sprint(err) != tt.err {
-					t.Errorf("wanted '%v', got '%v'", tt.err, err)
-				}
-			}
+			matcher.AssertErrHasWantMsg(t, err, tt.err)
 		})
 	}
 }
--- a/pkg/api/v20220904/clustermanager_validatestatic_test.go
+++ b/pkg/api/v20220904/clustermanager_validatestatic_test.go
@@ -6,7 +6,7 @@ package v20220904
 import (
 	"testing"
 
-	"github.com/Azure/ARO-RP/test/util/matcher"
+	utilerror "github.com/Azure/ARO-RP/test/util/error"
 )
 
 var ocmResource = string(`
@@ -73,7 +73,7 @@ func TestStatic(t *testing.T) {
 			c := &clusterManagerStaticValidator{}
 
 			err := c.Static(tt.ocmResource, tt.ocmResourceType)
-			matcher.AssertErrHasWantMsg(t, err, tt.err)
+			utilerror.AssertErrorMessage(t, err, tt.err)
 		})
 	}
 }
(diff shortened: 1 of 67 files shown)


pkg/api/v20220904/clustermanager_validatestatic_test.go
```diff
@@ -38,7 +39,6 @@ func TestStatic(t *testing.T) {
 		name        string
 		ocmResource string
 		vars        map[string]string
-		wantErr     bool
```
AldoFusterTurpin, line 41: We don't really need this boolean. Removed after the refactoring.

pkg/operator/deploy/deploy_test.go
```diff
@@ -41,8 +41,8 @@ func TestCheckIngressIP(t *testing.T) {
 					},
 				}
 			},
-			want:    "1.2.3.4",
-			wantErr: nil,
+			want:       "1.2.3.4",
+			wantErrMsg: "",
```
s-amann, line 45: nit: we can omit setting `wantErrMsg` to `""`

s-amann: same for `want` down below

pkg/operator/deploy/deploy_test.go
```diff
@@ -373,7 +370,7 @@ func TestCheckPodImageVersion(t *testing.T) {
 			},
 			desiredVersion: "abcde",
 			want:           true,
-			wantErr:        nil,
+			wantErrMsg:     "",
```
s-amann, line 373: same thing we can omit this

pkg/util/conditions/condition_test.go
```diff
@@ -35,7 +36,7 @@ func TestSetCondition(t *testing.T) {
 		input   *operatorv1.OperatorCondition
 
 		expected arov1alpha1.ClusterStatus
-		wantErr  error
+		wantErr  string
```
s-amann, line 39: I was wondering how the type could change and none of its usage, we can delete `wantErr` from the struct since it is not used in a useful way.

s-amann (changes requested):
overall this is awesome and will help greatly in reducing test complexity, thank you! I just have a few minor nits below.

pkg/operator/deploy/deploy_test.go
AldoFusterTurpin: Done. Tks!

pkg/operator/deploy/deploy_test.go
AldoFusterTurpin: Done.Tks!

pkg/util/conditions/condition_test.go
AldoFusterTurpin: Right. Done. Tks!

s-amann (approved):

petrkotas (dismissed):
This one took me some time. Good cleanup Aldo!

petrkotas (dismissed):

test/util/matcher/matcher.go
```diff
@@ -60,3 +61,14 @@ func (m *AsyncOperationDocument) Matches(x interface{}) bool {
...
+// AssertErrHasWantMsg asserts that err.Error() is equal to wantMsg.
+func AssertErrHasWantMsg(t *testing.T, err error, wantMsg string) {
+	if err == nil && wantMsg != "" {
+		t.Errorf("did not get an error, but wanted error '%v'", wantMsg)
+	}
+
+	if err != nil && err.Error() != wantMsg {
+		t.Errorf("got error '%v', but wanted error '%v'", err, wantMsg)
+	}
+}
```
facchettos, line 74: this function doesn't really belong in here. You should probably put it in `test/util/error` (create the new package)

also the name is reading a little funny. Wouldn't be `AssertErrorMessage` be a bit nicer?

facchettos (changes requested):

test/util/matcher/matcher.go
AldoFusterTurpin: Done.

facchettos: awesome thanks !

facchettos (approved):

test/util/matcher/matcher.go
AldoFusterTurpin: Np. Thanks!

PR #2538, opened 2022-11-11
Revert "collect must-gather on failed CI e2e runs"

m1kola:
### What this PR does / why we need it:

This reverts commit 05e8d390a38b780de31668357e2c39921f31fc3d.

We need to rever this because we need this piece of code back to ensure proper E2E teardown in EV2:

https://github.com/Azure/ARO-RP/blob/d6d376477b6bb75aac86da6f0700a36c024ad336/test/e2e/setup.go#L203-L218

Once we fix EV2 pipeline we need to look into:

* Whether we actually need must gather in E2E pipeline (have anyone ever actually looked into it? Is it useful?)
* If we need must gather - restore this functionality, but keep EV2 pipelines in mind: we need proper clean up

---

I had to fix conflicts in the following files:

* `test/e2e/setup.go`
* `.pipelines/e2e.yml`

### Test plan for issue:

Run E2E in ADO and run E2E in EV2 for canary

PR #1955, opened 2022-02-09
Collect must-gathers on failed e2e runs

cadenmarchese:
### Which issue this PR addresses:

Fixes https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/11004136 and https://msazure.visualstudio.com/AzureRedHatOpenShift/_workitems/edit/12519725/

For reviewers - here is a recent pipeline run with these changes where e2e failed, and a must-gather was collected and appended as an artifact: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=59130662&view=logs&j=0eff6622-f27f-5719-c5b3-a82191700046&t=0eff6622-f27f-5719-c5b3-a82191700046

### What this PR does / why we need it:

- Adds a step to CI runs that appends a must-gather
- Executes the step only when the E2E job fails

### Test plan for issue:

Confirm in CI that behavior is as expected

### Is there any documentation that needs to be updated for this PR?

Yes, we will need to update documentation on debugging e2e: https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/166757/Debug-E2E

PR #2618, opened 2023-01-05
Reintroduce CI must-gathers, retain existing cluster deletion

cadenmarchese:

Follow-up PR for an issue with the must-gather feature where CI e2e clusters were getting deleted successfully, but Ev2 e2e clusters were getting left behind.

- Reverts  https://github.com/Azure/ARO-RP/pull/2538 with fixes
- Reintroduces https://github.com/Azure/ARO-RP/pull/1955
- Fixes https://issues.redhat.com/browse/ARO-1502

When must-gather CI was first introduced, it removed cluster deletion from the e2e binary and moved it to a pipeline step. This PR is almost identical, except for that rather than making any changes to the e2e binary, it passes in `E2E_DELETE_CLUSTER=false` so that only CI clusters are retained for must-gather collection, then deleted later using the pipeline.

### Test plan for issue:

Confirm functionality in CI (we should be OK to stop there, since we're no longer making any changes to the e2e binary)

### DIFF
--- a/.pipelines/e2e.yml
+++ b/.pipelines/e2e.yml
@@ -38,7 +38,7 @@ jobs:
 
       run_portal
       validate_portal_running
-      
+
       run_rp
       validate_rp_running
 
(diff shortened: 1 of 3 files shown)


petrkotas (approved):
It seems good. We can try this and see if anything breaks.

PR #2620, opened 2023-01-05
Add etcd-fix operator to detect and correct master node IP change breaking etcd - ARO-1534

s-fairchild:
This operator is needed to resolve a crash loop failure of etcd pod due to hardcoded etcd endpoints in pod spec. oc observe is used to detect this condition and start a bash script.
Information from the Updated event is passed to the script that performs recover steps detailed here: https://docs.openshift.com/container-platform/4.4/backup_and_restore/replacing-unhealthy-etcd-member.html#restore-replace-crashlooping-etcd-member_replacing-unhealthy-etcd-member

Fixes
https://issues.redhat.com/browse/ARO-1534

When the master node IP changes etcd pod spec has a hard coded ETCD_ENDPOINTS variable. If a member changes IP addresses the endpoints var isn't updated, causing a crashlooping etcd container on that node.
This PR provides an automated way to detect and correct this condition on the problem master node.

The script only performs actions when the problem node is the same as the host node. This prevents peers from taking actions they shouldn't.

### Test plan for issue:

Adding unit tests.
Tested against development cluster.

1. Execute script below to reproduce the IP change.
  * `./change_vm_nic.sh steven-cluster-hrkhk-master-2`

```bash
#!/bin/bash

main() {
    if [[ -z $1 ]]; then
        abort "node name must be provided"
    fi

    node="$1"
    nic2="${node}-nic-$((1 + $RANDOM % 10))"

    subnetID="$(az network vnet subnet show -n $CLUSTER-master --vnet-name dev-vnet --query id)" || abort "failed to get subnet ID"
    aro="aro-${CLUSTER}"
    az network nic create --name "$nic2" --subnet "$( tr -d \" <<< "$subnetID")" -g "$aro" || abort "failed to create nic $nic2"

    az vm deallocate -g "$aro" -n "$node" || abort "failed to deallocate $node"

    az vm nic add --nics "$nic2" --vm-name "$node" --primary-nic "$nic2" -g "$aro" || abort "failed to add $nic2 to $node"

    az vm start -n "$node" -g "$aro" || abort "failed to start $node"
}

abort() {
    echo "$1"
    echo "Exiting"
    exit 1
}

main "$1"
```

3. Create operator
  * `go run -tags aro ./cmd/aro/ operator master`
4. Monitored operator logs
  * `oc logs -n openshift-etcd pod/etcd-fix-nrsd2`
5. Example logs output
```bash

openshift-etcd etcd-steven-cluster-hrkhk-master-2 ''\''{"name":"ETCDCTL_ENDPOINTS","value":"https://10.125.80.6:2379,https://10.125.80.8:2379,https://10.125.80.9:2379"} {"name":"NODE_steven_cluster_hrkhk_master_0_IP","value":"10.125.80.6"} {"name":"NODE_steven_cluster_hrkhk_master_1_IP","value":"10.125.80.9"} {"name":"NODE_steven_cluster_hrkhk_master_2_IP","value":"10.125.80.7"} False'\'''
Event is: Updated
Container status: False
10.125.80.7 not found in https://10.125.80.6:2379,https://10.125.80.8:2379,https://10.125.80.9:2379
bad node name is: steven-cluster-hrkhk-master-2
bad pod name is: etcd-steven-cluster-hrkhk-master-2
All good node IPs: 10.125.80.6 10.125.80.9
Creating /host/var/lib/etcd-backup
Moving /host/etc/kubernetes/manifests/etcd-pod.yaml to /host/var/lib/etcd-backup
Moving /host/var/lib/etcd/ to /host/tmp
found peer pod: etcd-steven-cluster-hrkhk-master-0
found peer pod: etcd-steven-cluster-hrkhk-master-1
rshing into pod/etcd-steven-cluster-hrkhk-master-0 now to remove member id 9e10a4626851f482
Member 9e10a4626851f482 removed from cluster bab5f47cf9607d11
id not found in etcd member list for pod etcd-steven-cluster-hrkhk-master-1
secret "etcd-peer-steven-cluster-hrkhk-master-2" deleted
secret "etcd-serving-steven-cluster-hrkhk-master-2" deleted
secret "etcd-serving-metrics-steven-cluster-hrkhk-master-2" deleted
etcd.operator.openshift.io/cluster patched
```

Toil work, I'm unsure if this needs to be documented?

### DIFF
--- a/pkg/operator/controllers/etcdfix/etcd-fix-operator.sh
+++ b/pkg/operator/controllers/etcdfix/etcd-fix-operator.sh
@@ -1,0 +1,170 @@
+#!/bin/bash
+#
+# Perform actions to fix etcd members if the node's IP address doesn't match etcd endpoints and container status is not ready
+# See for more information: https://docs.openshift.com/container-platform/4.4/backup_and_restore/replacing-unhealthy-etcd-member.html#restore-replace-crashlooping-etcd-member_replacing-unhealthy-etcd-member
+
+main() {
+    create_sym_links
+    PATH+="${PATH}:/host/usr/bin"
+
+    # ns="$1"
+
+    host_pod="$2"
+    local -i i=0
+    local -A ips
+    # Generate associative array of IPs
+    for a in $@; do
+        l="$(remove_single_quotes $a)"
+        if jq -e . >/dev/null 2>&1 <<< "$l" && [[ $i -gt 1 ]] ; then
+            d="$(jq -r '.[]' <<< "$l" | tr '\n' ' ')"
+            IFS=' ' read -a arr <<< "$d"
+            ips[${arr[0]}]="${arr[1]}"
+        elif [[ $l =~ ^True|False$ ]] && [[ $i -gt 1 ]]; then
+            container_status="$l"
+            echo "Container status: $container_status"
+        fi
+        ((i++))
+    done
+
+    endpoints="${ips[ETCDCTL_ENDPOINTS]}"
+    unset "ips[ETCDCTL_ENDPOINTS]"
+
+    for k in "${!ips[@]}"; do
+        if ! test_ips "${ips[$k]}" "$endpoints" ; then
+            bad_node="$(get_name "node" $k)"
+            bad_pod="$(get_name "pod" $k)"
+            echo "bad node name is: $bad_node"
+            echo "bad pod name is: $bad_pod"
+            unset "ips[$k]"
+        fi
+    done
+    echo "All good node IPs: ${ips[*]}"
+
+    if [[ $container_status == "False" ]] && [[ "$bad_node" == "$K8S_NODE" ]]; then
+        fix_host_etcd
+        fix_peer_etcd
+    fi
+}
+
+get_name() {
+    s="$(tr '_' '-' <<< $2)"
+    if [[ $1 == "pod" ]]; then
+        s="${s//NODE/etcd}"
+        echo "${s:0: -3}"
+    elif [[ $1 == "node" ]]; then
+        s="${s//NODE-/}"
+        echo "${s//-IP/}"
+    fi
+}
+
+remove_single_quotes() {
+    local line
+    line="$1"
+    if [ "${line:0:1}" == \' ] || [ "${line: -1}" == \' ]; then
+        line="$(echo $line | tr -d \')"
+    fi
+    echo "$line"
+}
+
+create_sym_links() {
+    jq_lib1="/usr/lib64/libjq.so.1"
+    jq_lib2="/usr/lib64/libonig.so.5"
+
+    if [[ ! -f $jq_lib1 ]]; then
+        ln -s "/host${jq_lib1}" "$jq_lib1"
+    fi
+    if [[ ! -f $jq_lib2 ]]; then
+        ln -s "/host${jq_lib2}" "$jq_lib2"
+    fi
+}
+
+fix_host_etcd() {
+    local bdir etcd_yaml
+    bdir=/host/var/lib/etcd-backup
+    etcd_yaml=/host/etc/kubernetes/manifests/etcd-pod.yaml
+    etcd_dir=/host/var/lib/etcd/
+    if [[ ! -d $bdir ]] && [[ -f $etcd_yaml ]]; then
+        echo "Creating $bdir"
+        mkdir -p "$bdir" || abort "failed to make backup directory"
+        echo "Moving $etcd_yaml to $bdir"
+        mv "$etcd_yaml" "$bdir/" || abort "failed to move $etcd_yaml to $bdir"
+        echo "Moving $etcd_dir to /host/tmp"
+        mv "$etcd_dir" /host/tmp/ || abort "failed to move $etcd_dir to /host/tmp"
+    else
+        echo "$bdir already exists or $etcd_yaml has already been moved"
+        echo "Not taking host etcd backup"
+    fi
+}
+
+########################################
+# fix_peer_etcd removes the unhealthy member from healthy peer members lists
+# and deletes unhealthy pod's secrets.
+# Afterwards etcd is redeployed
+########################################
+fix_peer_etcd() {
+    for p in ${!ips[@]}; do
+        peer_pods+=("$(get_name "pod" "$p")")
+    done
+
+    # If at least one member was deleted from a peer's member list, delete secrets and patch
+    cont="false"
+    remove_peer_members
+    if [[ $cont == "true" ]]; then
+        remove_etcd_secrets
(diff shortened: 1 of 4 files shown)


gvanderpotte:
Hi @s-fairchild that's good that we have this being taken. 
I think that the operator aproach and more generally automatic detect/recovery is probably a better aproach than waiting for a PUCM to repair the problem. 
to @darthhexx 's point, I think that indeed, etcd reconciliation, ssh fixing and NIC reconciliation could be part of a greater feature. But rather than putting that in a PUCM, I would rather go for the aproach taken with the operator. 
(without looking at the code itself, I have not reviewed it, just talking about the approach here)

s-fairchild:
@darthhexx No there isn't a design document for this feature that I'm aware of.
The goal of this PR was to automate a fix for [this issue](https://issues.redhat.com/browse/ARO-1534) about the [problem described here](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/64663/Current-known-and-outstanding-issues?anchor=apiserver-errors-due-to-broken-etcd).

Will the automated master node replacement monitor the cluster from an operator live, or be triggered another way? I'm curious mostly to see if this PR can fit (or modified to fit) into the larger feature design.

Also Would this particular issue be better handled by an automated full node replacement, rather than just etcd?

darthhexx:
> Will the automated master node replacement monitor the cluster from an operator live, or be triggered another way?

It would be triggered as a new SRE triggered RP process that runs as a set of predefined actions and condition steps, just like we have [here for a cluster update](https://github.com/Azure/ARO-RP/blob/d6b3c3c87b1b2ed88f1eee61aa13b771a200e49f/pkg/cluster/install.go#L154-L182). This gives us specific control over the actions and remediation processes, much like your steps are doing in the bash script, but in the RP itself.

> Also Would this particular issue be better handled by an automated full node replacement, rather than just etcd?

Personally, I am not comfortable having ectd replacement code running on all clusters all of the time waiting to perform an etcd pod replacement as soon as it detects an ectd member not available. From that I can see from a quick glance, there are also no locks to prevent multiple replacements and adding that alone will raise the detection and remediation complexity substantially.

Having the process automated is a must, but I do think it is better suited as an SRE initiated process on a confirmed failed etcd pod, due to the ramifications of "getting it wrong", rather than a permanent process running on all clusters.

With regards to using this process in the automated master node replacement, it would be ideal if the process was converted to Golang functions in a steps-based approach, which could be tested, instead of a daemonset and a bash script. However, the removing of the etcd member would still need to be performed in a pod, but that could be an Action to deploy followed by a Condition to verify and a subsequent Action to remove the pod.

Of course there are always many ways to design a system, so my suggestions are just my viewpoint and I'm happy to hash this out further in a meeting, design doc, or this PR.

pkg/operator/controllers/etcdfix/etcd-fix-operator.sh
```diff
@@ -0,0 +1,211 @@
...
+    line="$1"
+    if [ "${line:0:1}" == \' ] || [ "${line: -1}" == \' ]; then
+        line="$(echo $line | tr -d \')"
+    fi
+    echo "$line"
+}
+
+########################################
+# create_sym_links creates symbolic links for shared libraries required by jq
+########################################
```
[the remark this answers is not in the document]
s-fairchild: Thanks for the tip!

pkg/operator/controllers/etcdfix/etcdfix_controller.go
```diff
@@ -0,0 +1,109 @@
...
+		return reconcile.Result{}, err
+	}
+
+	r.deploy(ctx, instance)
+
+	r.log.Debug("running")
+	return ctrl.Result{}, nil
+}
+
+func (r *Reconciler) deploy(ctx context.Context, instance *arov1alpha1.Cluster) (ctrl.Result, error) {
```
cadenmarchese, line 61: We'll also want unit testing for `Reconcile` and any functions/methods that it relies on.

pkg/operator/controllers/etcdfix/etcdfix_controller.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// SetupWithManager creates the controller
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	aroClusterPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return o.GetName() == arov1alpha1.SingletonClusterName
+	})
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&arov1alpha1.Cluster{}, builder.WithPredicates(aroClusterPredicate)).
```
cadenmarchese, line 103: Another idea for this would be to watch the master machine objects for changes, and execute remediation or other operator actions (write a status) if the right conditions are detected.

cadenmarchese (commented):
Thanks for your PR, I've given it a first pass. Overall, I agree with comments that we will most likely want to take a Go/reconciler approach to the remediation rather than have a controller that enforces a DaemonSet and relies on a bash script for remediation. Some potential paths forward from here:

- Use this operator to detect a (breaking) master node IP change, and use it to write an operator status, and potentially alert SRE so that remediation steps can be taken
- Use this operator for remediation, but only trigger the remediation when a specific flag is set (maybe a feature flag in the cluster document / cluster CRD that we set only if remediation is needed)

Some open questions I have:

- How automated do we want this process to be? Is this issue something we just want visibility on when it happens, or do we want to go full auto with regards to detect / remediate?
- What extent of this work is covered by the planned master node replacement feature, and can we use this operator to trigger the master node replacement feature if/ when it is part of production?

I think an arch meeting topic on this with a quick proposal would help in answering the above and getting agreement on how to best proceed. So far, this is great work.

s-fairchild:
@darthhexx Thanks for your input. I agree a tested Go step based remediation process, SRE triggered sounds like the safest approach. I'd like to meet to discuss this design and incorporating it into the larger master node replacement feature when you have time. I'll reach out to set something up.

> Use this operator to detect a (breaking) master node IP change, and use it to write an operator status, and potentially alert SRE so that remediation steps can be taken

@cadenmarchese Are we using operators to write statuses for other events?
Maybe I can update this PR shifting this operator from taking action to alerting, then an SRE triggering an automated fix from there.

cadenmarchese (commented):
@s-fairchild we have a few controllers that write statuses to the cluster custom resource (`oc get cluster cluster -o yaml` in an ARO cluster) - the checker controllers are a good example of this.

I agree that as a first step we can detect / write the status if a master node IP has changed, and then decide from there if we want to remediate either using the operator reconciler or something else.

s-fairchild:
Closing this and opening https://github.com/Azure/ARO-RP/pull/2725

PR #2622, opened 2023-01-06
add outbound-type param to az cli extension

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-1478

Bumps api version used and adds parameter `--outbound-type` to az aro cli extension for public IP removal.  OutboundType can be Loadbalancer in all cases of apiserver/ingress visibility but OutboundType of UserDefinedRouting is only valid if both apiserver and ingress are private.  This should probably not not be merged until the next stable API is released.

### Test plan for issue:

Added unit tests and created clusters with OutboundType of UserDefinedRouting and Loadbalancer.

Yes, the [Azure Doc](https://learn.microsoft.com/en-us/azure/openshift/howto-create-private-cluster-4x#create-a-private-cluster-without-a-public-ip-address-preview) will need to be updated though that will be a separate effort.

### DIFF
--- a/python/az/aro/azext_aro/_validators.py
+++ b/python/az/aro/azext_aro/_validators.py
@@ -118,6 +118,13 @@ def validate_pull_secret(namespace):
         raise InvalidArgumentValueError("Invalid --pull-secret.") from e
 
 
+def validate_outbound_type(namespace):
+    outbound_type = getattr(namespace, 'outbound_type')
+    if outbound_type in {'UserDefinedRouting', 'Loadbalancer', None}:
+        return
+    raise InvalidArgumentValueError('Outbound type must be "UserDefinedRouting" or "Loadbalancer"')
+
+
 def validate_subnet(key):
     def _validate_subnet(cmd, namespace):
         subnet = getattr(namespace, key)
--- a/python/az/aro/azext_aro/_validators.py
+++ b/python/az/aro/azext_aro/_validators.py
@@ -120,9 +120,21 @@ def validate_pull_secret(namespace):
 
 def validate_outbound_type(namespace):
     outbound_type = getattr(namespace, 'outbound_type')
-    if outbound_type in {'UserDefinedRouting', 'Loadbalancer', None}:
-        return
-    raise InvalidArgumentValueError('Outbound type must be "UserDefinedRouting" or "Loadbalancer"')
+    if outbound_type not in {'UserDefinedRouting', 'Loadbalancer', None}:
+        raise InvalidArgumentValueError('Invalid --outbound-type: must be "UserDefinedRouting" or "Loadbalancer"')
+
+    ingress_visibility = getattr(namespace, 'ingress_visibility')
+    apiserver_visibility = getattr(namespace, 'apiserver_visibility')
+
+    if (outbound_type == 'UserDefinedRouting' and
+            (is_visibility_public(ingress_visibility) or is_visibility_public(apiserver_visibility))):
+        raise InvalidArgumentValueError('Invalid --outbound-type: cannot use UserDefinedRouting when ' +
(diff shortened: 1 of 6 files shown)


SudoBrendan (changes requested):
@tony-schndr This PR does add the flag to the python code, including validation, but doesn't leverage it in any capacity when actually creating clusters, so I don't think this is ready-for-review. In this PR, we'll need to bump the API version to the next stable (once it's available and has the outboundtype as an exposed parameter), and then leverage that API's new functionality for selecting an outbound type when creating clusters.

SudoBrendan (dismissed):
(removed)

tony-schndr:
Depends on https://github.com/Azure/ARO-RP/pull/2760

python/az/aro/azext_aro/_validators.py
```diff
@@ -118,6 +118,26 @@ def validate_pull_secret(namespace):
...
+    outbound_type = getattr(namespace, 'outbound_type')
+    if outbound_type not in {'UserDefinedRouting', 'Loadbalancer', None}:
+        raise InvalidArgumentValueError('Invalid --outbound-type: must be "UserDefinedRouting" or "Loadbalancer"')
+
+    ingress_visibility = getattr(namespace, 'ingress_visibility')
+    apiserver_visibility = getattr(namespace, 'apiserver_visibility')
+
+    if (outbound_type == 'UserDefinedRouting' and
+            is_visibility_public(ingress_visibility) or
+            is_visibility_public(apiserver_visibility)):
```
SudoBrendan, line 131: I think this is bugged, needs some `()` here to wrap the `or` condition (that is, I'm not sure why tests are passing? Maybe `getattr` is returning `""` instead of `None`? I'd recommend a debug session to confirm...). This is a result of Python order of operations on `and` vs `or`.

For example, this can result in a:

```
if False and True or True
```

which is equivalent to
```
if (False and True) or True
```

Basically, any time they set public API, they'll pass this `if`. So, the user doesn't set UserDefinedRouting, then we tell them, "you can't use that!"

SudoBrendan (changes requested):

python/az/aro/azext_aro/_validators.py
tony-schndr: Python evaluates these expressions lazily, if the first expression is not UserDefinedRouting then the rest of the expression will not be evaluated.  If the first expression is True, then the next expression is evaluated. If that next expression is False then the next is evaulated.  So either way here we are going to encounter the exception if a user passes in UserDefinedRouting and if any visibility is public.

SudoBrendan (dismissed):
Overall LGTM, added a nit for another test case, but this can merge.

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -775,3 +775,62 @@ def test_validate_refresh_cluster_credentials(test_description, namespace, expec
     else:
         with pytest.raises(expected_exception):
             validate_refresh_cluster_credentials(namespace)
+
+
+test_validate_outbound_type_data = [
+    (
+        "Should not raise exception when key is Loadbalancer.",
+        Mock(outbound_type='Loadbalancer'),
+        None
```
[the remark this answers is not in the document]
tony-schndr: Added additional test cases.

s-amann (approved):

petrkotas (approved):
Hi @tony-schndr LGTM, I also see dependent PR is already merged. Do we need to wait for publishing the new API version too?

tony-schndr:
@petrkotas From my understanding we can merge this since we generate the python client/sdk in the ARO-RP for the cli extension.  We will need to wait for the API to be merged upstream and SDKs generated before we can upstream this to azure-cli.

PR #2567, opened 2022-12-01
E2E Containerization

rhamitarora:
### Which issue this PR addresses:
E2E Containerization
WIP

Fixes

PR #2625, opened 2023-01-09
Create a utility container for troubleshooting network issues

ventifus:

Part of TSG https://issues.redhat.com/browse/ARO-1494

Provides a tool-box of advanced network troubleshooting utilities, intended to be run on customer clusters as a one-shot debug pod.

### DIFF
--- a/Dockerfile.net-utils
+++ b/Dockerfile.net-utils
@@ -1,0 +1,23 @@
+FROM registry.fedoraproject.org/fedora:latest
+
+LABEL MAINTAINER="aos-azure"
+
+# Autorest prerequisites
+RUN INSTALL_PKGS="\
+    bind-utils \
+    dnstop \
+    ethtool \
+    iftop \
+    iproute \
+    iputils \
+    ldns-utils \
+    mtr \
+    nmap \
+    nmap-ncat \
+    procps-ng \
+    traceroute \
+    wireshark-cli \
+    " && \
+    yum -y install $INSTALL_PKGS && rpm -V --nosize --nofiledigest --nomtime --nomode $INSTALL_PKGS && yum clean all && rm -rf /var/cache/*
+
+ENTRYPOINT ["/bin/bash"]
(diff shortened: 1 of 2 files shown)


hawkowl (changes requested):
1. This should be built in CI instead of as a thing that's manually pushed -- see stuff like the Fluentbit containers in ADO.
2. We have to pull base images from the ACR for running in prod, see the fluentbit dockerfile for that. We would need to mirror the Fedora image into the ACR if we wanted to build off it.
3. We already have several debugging containers, such as support-tools and ose-toolbox. Can we instead use one of those (which we already mirror) as a base? I think that's probably better than Fedora, since I don't entirely know if Fedora has all the FIPS/etc that we need for places such as MAG.

The two containers we already mirror:

https://catalog.redhat.com/software/containers/rhel8/support-tools/5ba3eaf9bed8bd6ee819b78b?container-tabs=dockerfile
https://catalog.redhat.com/software/containers/openshift4/ose-tools-rhel8/5f748d3399cc5b9e7c1a8747?container-tabs=dockerfile

Dockerfile.net-utils
```diff
@@ -0,0 +1,23 @@
+FROM registry.fedoraproject.org/fedora:latest
```
cadenmarchese, line 1: Since we have a requirement to pull the base image from ACR, we could potentially use one of our existing images here, and instead add a fedora repository to it in your dockerfile to install the pagkages needed.

This is currently being done in https://github.com/Azure/ARO-RP/pull/2567 to install openvpn since it was not available in the base image.

ventifus: In that PR we're installing openvpn from EPEL rather than Fedora proper. One approach would be to maintain EPEL packages for the tools we want to have available.

cadenmarchese: The only compliance requirement that would hit us on this change that I am aware of is the FROM directive pointing to something other than our ACRs - adding repositories or pulling down packages later in the dockerfile should be fair game. Do you mean that we could host the packages somewhere on our end?

ventifus:
My current thoughts on this is we need to start from ose-tools or support-tools base image, and then maintain EPEL versions of whichever tools we'd like to see available. Since this would be an ongoing maintenance burden, we'd want to be careful with which tools we adopt.

PR #2627, opened 2023-01-11
Removes dependency on a versioned ARO client from the operator

m1kola:

Operator is not longer dependant on a version client for ARO objects and is now using a split client.

This continues the rafactoring started in #2605

### Test plan for issue:

No, just refactoring.


petrkotas (approved):
Good one, LGTM

PR #2631, opened 2023-01-12
e2e test for Hive-managed ARO clusters

kimorris27:

https://issues.redhat.com/browse/ARO-1602

This PR adds new e2e tests that run only when the test cluster is managed by Hive. The tests attempt to validate that the installation/adoption of the test cluster went as expected by comparing some data values stored in the test cluster's corresponding `ClusterDeployment` object (retrieved from the AKS/Hive cluster) to the expected values (the values that the RP would have set at the time that it told Hive to install/adopt the cluster).

### Test plan for issue:

I ran the new e2e test locally using my dev RP instance. When using a test cluster created by Hive, the new tests all pass. When using a test cluster *not* created by Hive, the tests are skipped (the desired behavior).

Not to my knowledge, but please let me know if there is any documentation I'll need to update.

### Note

I wasn't sure which values from the `ClusterDeployment` are worth checking in this context, so I just picked a subset that seemed important. Let me know if there are others that we should be checking and I'll add them in.

### DIFF
--- a/test/e2e/hive.go
+++ b/test/e2e/hive.go
@@ -1,0 +1,55 @@
+package e2e
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/Azure/ARO-RP/pkg/api/admin"
+	"github.com/Azure/ARO-RP/pkg/hive"
+)
+
+var (
+	clusterPlatformLabelKey string = "hive.openshift.io/cluster-platform"
+	clusterRegionLabelKey   string = "hive.openshift.io/cluster-region"
+	clusterPlatformValue    string = "azure"
+)
+
+var _ = Describe("Hive-managed ARO cluster", func() {
+	var oc *admin.OpenShiftCluster
+	var resourceID string
+
+	BeforeEach(func(ctx context.Context) {
+		resourceID = resourceIDFromEnv()
+		oc = adminGetCluster(Default, ctx, resourceID)
+
+		if oc.Properties.HiveProfile == (admin.HiveProfile{}) {
+			Skip("skipping tests because this ARO cluster has not been created/adopted by Hive")
+		}
+	})
+
+	It("has been properly created/adopted by Hive", func(ctx context.Context) {
+		By("verifying that a corresponding ClusterDeployment object exists in the expected namespace in the Hive cluster")
+		cd, err := clients.Hive.HiveV1().ClusterDeployments(oc.Properties.HiveProfile.Namespace).Get(ctx, hive.ClusterDeploymentName, metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		By("verifying that the ClusterDeployment object has the expected name and labels")
+		Expect(cd.ObjectMeta).NotTo(BeNil())
+		Expect(cd.ObjectMeta.Name).To(Equal(hive.ClusterDeploymentName))
+		Expect(cd.ObjectMeta.Labels).Should(HaveKey(clusterPlatformLabelKey))
+		Expect(cd.ObjectMeta.Labels[clusterPlatformLabelKey]).To(Equal(clusterPlatformValue))
+		Expect(cd.ObjectMeta.Labels).Should(HaveKey(clusterRegionLabelKey))
+		Expect(cd.ObjectMeta.Labels[clusterRegionLabelKey]).To(Equal(oc.Location))
+
+		By("verifying that the ClusterDeployment object spec correctly includes the ARO cluster's Azure region and RG name")
+		Expect(cd.Spec).NotTo(BeNil())
+		Expect(cd.Spec.ClusterName).To(Equal(oc.Name))
+		Expect(cd.Spec.Platform).NotTo(BeNil())
+		Expect(cd.Spec.Platform.Azure).NotTo(BeNil())
+		Expect(cd.Spec.Platform.Azure.BaseDomainResourceGroupName).To(Equal(oc.Properties.ClusterProfile.ResourceGroupID))
+		Expect(cd.Spec.Platform.Azure.Region).To(Equal(oc.Location))
+	})
+})
--- a/test/e2e/hive.go
+++ b/test/e2e/hive.go
@@ -20,7 +20,6 @@ import (
 var (
 	clusterPlatformLabelKey string = "hive.openshift.io/cluster-platform"
 	clusterRegionLabelKey   string = "hive.openshift.io/cluster-region"
-	clusterPlatformValue    string = "azure"
 
 	controlPlaneAPIURLOverride = func(clusterDomain string, clusterLocation string) string {
 		if !strings.ContainsRune(clusterDomain, '.') {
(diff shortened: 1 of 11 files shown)


s-amann (approved):

test/e2e/hive.go
```diff
@@ -0,0 +1,58 @@
...
+
+		By("verifying that the ClusterDeployment object has the expected name and labels")
+		Expect(cd.ObjectMeta).NotTo(BeNil())
+		Expect(cd.ObjectMeta.Name).To(Equal(hive.ClusterDeploymentName))
+		Expect(cd.ObjectMeta.Labels).Should(HaveKey(clusterPlatformLabelKey))
+		Expect(cd.ObjectMeta.Labels[clusterPlatformLabelKey]).To(Equal(clusterPlatformValue))
+		Expect(cd.ObjectMeta.Labels).Should(HaveKey(clusterRegionLabelKey))
+		Expect(cd.ObjectMeta.Labels[clusterRegionLabelKey]).To(Equal(oc.Location))
+
+		By("verifying that the ClusterDeployment object spec correctly includes the ARO cluster's Azure region and RG name")
```
s-amann, line 50: @kimorris27 , can we add to check for these two properties? They are critical for a successful install

```go
  APIURLOverride:      fmt.Sprintf("api-int.%s:6443", clusterDomain),
},
```

kimorris27: Yep, I've added those in! Will re-request review.

s-amann (dismissed):

test/e2e/hive.go
```diff
@@ -0,0 +1,74 @@
...
+
+var _ = Describe("Hive-managed ARO cluster", func() {
+	var oc *admin.OpenShiftCluster
+	var resourceID string
+
+	BeforeEach(func(ctx context.Context) {
+		resourceID = resourceIDFromEnv()
+		oc = adminGetCluster(Default, ctx, resourceID)
+
+		if oc.Properties.HiveProfile == (admin.HiveProfile{}) {
```
cadenmarchese, line 43: We may want to make a function for this in `test/e2e/setup.go` like what we have for `skipIfNotInDevelopmentEnv()` and `skipIfDockerNotWorking()` that we can call inside of your test. This will be helpful if we add more tests that depend on the HiveProfile in the future.

test/e2e/hive.go
```diff
@@ -0,0 +1,74 @@
...
+			clusterDomain += "." + clusterLocation + "." + os.Getenv("PARENT_DOMAIN_NAME")
+		}
+
+		return fmt.Sprintf("api-int.%s:6443", clusterDomain)
+	}
+)
+
+var _ = Describe("Hive-managed ARO cluster", func() {
+	var oc *admin.OpenShiftCluster
+	var resourceID string
```
cadenmarchese, line 37: These vars can probably be defined at package level since they are used a lot in the e2e suite. Feel free to make the changes to other files in the package if necessary.

test/e2e/hive.go
```diff
@@ -0,0 +1,74 @@
...
+		Expect(cd.Spec.ClusterName).To(Equal(oc.Name))
+		Expect(cd.Spec.Platform).NotTo(BeNil())
+		Expect(cd.Spec.Platform.Azure).NotTo(BeNil())
+		Expect(cd.Spec.Platform.Azure.BaseDomainResourceGroupName).To(Equal(oc.Properties.ClusterProfile.ResourceGroupID))
+		Expect(cd.Spec.Platform.Azure.Region).To(Equal(oc.Location))
+
+		By("verifying that the ClusterDeployment object spec includes the expected ControlPlaneConfig overrides")
+		Expect(cd.Spec.ControlPlaneConfig).NotTo(BeNil())
+		Expect(cd.Spec.ControlPlaneConfig.APIServerIPOverride).To(Equal(oc.Properties.NetworkProfile.APIServerPrivateEndpointIP))
+		Expect(cd.Spec.ControlPlaneConfig.APIURLOverride).To(Equal(controlPlaneAPIURLOverride(oc.Properties.ClusterProfile.Domain, oc.Location)))
```
cadenmarchese, line 72: I would move these `Expect` assertions up a line and remove the whitespace so it's clear to the reader that these assertions are part of the "verifying that the ClusterDeployment object spec includes the expected ControlPlaneConfig overrides" test.

test/e2e/hive.go
```diff
@@ -0,0 +1,74 @@
...
+		Expect(cd.ObjectMeta.Labels).Should(HaveKey(clusterPlatformLabelKey))
+		Expect(cd.ObjectMeta.Labels[clusterPlatformLabelKey]).To(Equal(clusterPlatformValue))
+		Expect(cd.ObjectMeta.Labels).Should(HaveKey(clusterRegionLabelKey))
+		Expect(cd.ObjectMeta.Labels[clusterRegionLabelKey]).To(Equal(oc.Location))
+
+		By("verifying that the ClusterDeployment object spec correctly includes the ARO cluster's Azure region and RG name")
+		Expect(cd.Spec).NotTo(BeNil())
+		Expect(cd.Spec.ClusterName).To(Equal(oc.Name))
+		Expect(cd.Spec.Platform).NotTo(BeNil())
+		Expect(cd.Spec.Platform.Azure).NotTo(BeNil())
```
cadenmarchese, line 65: IMO, if possible we should test exactly what we want to see rather than test that it's not empty.

test/e2e/hive.go
```diff
@@ -0,0 +1,74 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/Azure/ARO-RP/pkg/api/admin"
+	"github.com/Azure/ARO-RP/pkg/hive"
+)
+
+var (
+	clusterPlatformLabelKey string = "hive.openshift.io/cluster-platform"
+	clusterRegionLabelKey   string = "hive.openshift.io/cluster-region"
+	clusterPlatformValue    string = "azure"
```
cadenmarchese, line 24: It looks like this var is only used once, so you probably don't need it here.

cadenmarchese (changes requested):
This is great work overall, I have a few suggestions - thanks!

test/e2e/hive.go
kimorris27: If I understand correctly, the two lines below this (66 and 67 at the time of writing of this comment) are already doing what you're recommending:

```
Expect(cd.Spec.Platform.Azure.BaseDomainResourceGroupName).To(Equal(oc.Properties.ClusterProfile.ResourceGroupID))
Expect(cd.Spec.Platform.Azure.Region).To(Equal(oc.Location))
```

I didn't check all of the attributes though - I just picked two of the three that the RP sets.

test/e2e/hive.go
kimorris27: Are you referring specifically to `clusterPlatformValue` here or to the whole set of variables?

test/e2e/hive.go
kimorris27: I'm not seeing any whitespace between the `By` call on line 69 and the `Except` calls on lines 71 and 72. Did you miss the `Except` call on line 70?

test/e2e/hive.go
kimorris27: If I understand correctly, lines 66-67 (they check `cd.Spec.Platform.Azure.BaseDomainResourceGroupName` and `cd.Spec.Platform.Azure.Region`) are already checking what you're recommending. I didn't check every attribute of `Platform` though - I just picked a subset of the attributes that the RP sets. Do you see others you think we should check in addition to the two I picked?

test/e2e/hive.go
kimorris27: Made this change 👍🏻

test/e2e/hive.go
kimorris27: I made both of these package level, but then I realized that there's an e2e test for `adminGetCluster()` in `adminapi_cluster_get.go`. If we initialize a package level `*admin.OpenShiftCluster` during test setup, it defeats the point of having that test. What do you think of keeping the cluster resource ID as a package level variable and changing the admin cluster struct back to being a local variable?

test/e2e/hive.go
cadenmarchese: Just the `clusterPlatformValue` var.

test/e2e/hive.go
cadenmarchese: LGTM now, I think my diff must have been off.

cadenmarchese (approved):
LGTM

PR #2633, opened 2023-01-12
Onebranch template-buildrp-buildaro: fix logic

ventifus:

Fixes logic causing the pullrequest pipeline to fail


cadenmarchese (approved):
Makes sense to me, assuming that the PullRequest pipeline isn't setting `is_official_release`:
```
$ echo $is_official_release
<blank>
$ if [ -z "$is_official_release"]; then echo "check for annotated tag"; fi
check for annotated tag
```
And using this PR, it looks like the pipeline is happy again: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=67118765&view=results

PR #2634, opened 2023-01-13
logs added for frontend enrichment

carvalhe:
Fixes https://issues.redhat.com/browse/ARO-1899
Added frontendResultLog for enrichment

Currently frontend logs do not give out much information for QoS. The returned frontend response says "sent response" with not much detail if there was an error, and if there is, it's just an error code. This PR creates a separate result log that will look similar to the one created for backend. It sends a message stating "Succeeded" or "Failed", and marks it as a QoS log. It also sends the error information as well as the operation type being done from the http request method. For now, this is only being used on openshiftcluster put/patch, get, and delete. Potentially other places in the future. With these changes, we can easily measure the health by checking what's succeeding and what's failing.


darthhexx (approved):
Thanks for addressing the changes in #2594 and rebasing for this PR.

PR #2637, opened 2023-01-13
Feature/aro-1608 return hive deployment cr info

s-amann:

Fixes https://issues.redhat.com/browse/ARO-1608

Will add an admin endpoint to be wired for a Geneva Action to pull the Cluster Deployment of a hive managed cluster.

### Test plan for issue:

Manually tested using local RP as well as unit test and a development mode e2e test

not yet.

### DIFF
--- a/pkg/frontend/admin_hive_clusterdeployment_get.go
+++ b/pkg/frontend/admin_hive_clusterdeployment_get.go
@@ -5,7 +5,6 @@ package frontend
 
 import (
 	"context"
-	"errors"
 	"net/http"
 	"path/filepath"
 	"strings"
@@ -14,53 +13,47 @@ import (
 	"github.com/ugorji/go/codec"
 
 	"github.com/Azure/ARO-RP/pkg/api"
-	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
 	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
 )
 
 func (f *frontend) getAdminHiveClusterDeployment(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
+	resourceID := strings.TrimPrefix(filepath.Dir(r.URL.Path), "/admin")
+	b, err := f._getAdminHiveClusterDeployment(ctx, resourceID)
 
-	b, err := f._getAdminHiveClusterDeployment(ctx, r)
-
-	switch {
-	case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
-		api.WriteError(w, http.StatusNotFound, api.CloudErrorCodeNotFound, "", "Cluster not found.")
-		return
-	case err != nil:
-		api.WriteError(w, http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
+	if cloudErr, ok := err.(*api.CloudError); ok {
+		api.WriteCloudError(w, cloudErr)
 		return
 	}
 
 	adminReply(log, w, nil, b, err)
 }
 
-func (f *frontend) _getAdminHiveClusterDeployment(ctx context.Context, r *http.Request) ([]byte, error) {
+func (f *frontend) _getAdminHiveClusterDeployment(ctx context.Context, resourceID string) ([]byte, error) {
 	// we have to check if the frontend has a valid clustermanager since hive is not everywhere.
 	if f.hiveClusterManager == nil {
-		return nil, errors.New("hive is not enabled")
+		return nil, api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "hive is not enabled")
 	}
-	url := filepath.Dir(r.URL.Path)
-	resourceID := strings.TrimPrefix(url, "/admin")
+
 	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
 	if err != nil {
-		return nil, err
+		return nil, api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeNotFound, "", "cluster not found")
 	}
 
 	if doc.OpenShiftCluster.Properties.HiveProfile.Namespace == "" {
-		return nil, errors.New("cluster is not managed by hive")
+		return nil, api.NewCloudError(http.StatusNoContent, api.CloudErrorCodeResourceNotFound, "", "cluster is not managed by hive")
 	}
 
 	cd, err := f.hiveClusterManager.GetClusterDeployment(ctx, doc)
 	if err != nil {
-		return nil, err
+		return nil, api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeNotFound, "", "cluster deployment not found")
 	}
 
 	var b []byte
 	err = codec.NewEncoderBytes(&b, &codec.JsonHandle{}).Encode(cd)
 	if err != nil {
-		return nil, err
+		return nil, api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "unable to marshal response")
 	}
 
 	return b, nil
--- a/pkg/frontend/admin_hive_clusterdeployment_get.go
+++ b/pkg/frontend/admin_hive_clusterdeployment_get.go
@@ -1,0 +1,68 @@
+package frontend
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"github.com/ugorji/go/codec"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
+	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+)
+
+func (f *frontend) getAdminHiveClusterDeployment(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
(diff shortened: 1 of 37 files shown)


pkg/frontend/admin_hive_clusterdeployment_get.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/ugorji/go/codec"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
+	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+)
+
+func (f *frontend) getAdminHiveClusterDeployment(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
```
facchettos, line 23: should we check if the typecast is successful?

pkg/frontend/admin_hive_clusterdeployment_get.go
```diff
@@ -0,0 +1,69 @@
...
+	case err != nil:
+		api.WriteError(w, http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
+		return
+	}
+
+	adminReply(log, w, nil, b, err)
+}
+
+func (f *frontend) _getAdminHiveClusterDeployment(ctx context.Context, r *http.Request, log *logrus.Entry) ([]byte, error) {
+	r.URL.Path = filepath.Dir(r.URL.Path)
```
facchettos, line 40: it's probably equivalent in the end but it seems a little weird to modify the request path instead of storing this in a different variable

pkg/frontend/admin_hive_clusterdeployment_get_test.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	for _, tt := range []*test{
+		{
+			name:       "cluster has hive profile with namespace",
+			resourceID: fmt.Sprintf("/subscriptions/%s/resourcegroups/resourceGroup/providers/Microsoft.RedHatOpenShift/openShiftClusters/hive", fakeUUID),
+			properties: api.OpenShiftClusterProperties{HiveProfile: api.HiveProfile{Namespace: fmt.Sprintf("aro-%s", fakeUUID)}},
+			mocks: func(t *test, clusterManager *mock_hive.MockClusterManager) {
+				clusterManager.EXPECT().GetClusterDeployment(gomock.Any(), gomock.Any()).Return(&clusterDeployment, nil).Times(1)
+			},
```
facchettos, line 51: setting up the mocks should be done in the test, not here. it was a bad pattern and I'd rather not add new tests with that pattern

pkg/frontend/frontend.go
```diff
@@ -41,6 +42,8 @@ func (err statusCodeError) Error() string {
 	return fmt.Sprintf("%d", err)
 }
 
+type hiveActionsFactory func(*logrus.Entry, env.Interface) (hive.ClusterManager, error)
```
facchettos, line 45: do we really need a factory here? these are not great , the others needed the factory because they were created with the openshift cluster as a parameter (and it's was a bad design), here I don't think we need it and it'd be great to not add more of those

facchettos (dismissed):

pkg/frontend/frontend.go
s-amann: I did this because I think eventually this will be updated to include the OpenshiftCluster so we can know which shard to access. Otherwise we could just create the hiveclientset and add it as a part of the frontend without a factory.

pkg/frontend/admin_hive_clusterdeployment_get_test.go
s-amann: I've removed the mocks from the test struct.

pkg/frontend/admin_hive_clusterdeployment_get.go
s-amann: this was the existing pattern, but I can create a new variable to store the modified url string.

pkg/frontend/admin_hive_clusterdeployment_get.go
s-amann: I don't see the harm in checking, however most if not all the other api handlers do this type assertion without checking if it was ok. The check is kind of clunky, do you think we need to add it?

```golang
log, ok := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
if !ok {
	api.WriteError(w, http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "Internal Server Error")
	return
}

File: pkg/frontend/frontend.go (hunk 4 above)
Comment (facchettos) [reply pkg/frontend/frontend.go:45]: still, I don't think the factory is adding any value here, because 
1- the env is never changing so reconstructing it every time isn't really useful
2- the log is changing but we are only using it once so we can just add it as a parameter on the one function
3- we can add the openshiftcluster as an argument for the functions that need it

overall removing it is just going to simplify the code imho

Comment (s-amann) [reply pkg/frontend/frontend.go:45]: I agree with those points, I will directly inject the `hive.ClusterManager` to the frontend and get it pushed up.

File: pkg/frontend/admin_hive_clusterdeployment_get.go [hunk 5]
```diff
@@ -0,0 +1,63 @@
...
+func (f *frontend) _getAdminHiveClusterDeployment(ctx context.Context, r *http.Request) ([]byte, error) {
+	url := filepath.Dir(r.URL.Path)
+	resourceID := strings.TrimPrefix(url, "/admin")
+	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	if err != nil {
+		return nil, err
+	}
+
+	if doc.OpenShiftCluster.Properties.HiveProfile.Namespace == "" {
+		return nil, errors.New("cluster is not managed by hive")
```
cadenmarchese, line 48: Should we return a 204, `http.StatusNoContent` or similar here?

pkg/frontend/admin_hive_clusterdeployment_get.go
```diff
@@ -0,0 +1,63 @@
...
+	b, err := f._getAdminHiveClusterDeployment(ctx, r)
+
+	switch {
+	case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
+		api.WriteError(w, http.StatusNotFound, api.CloudErrorCodeNotFound, "", "Cluster not found.")
+		return
+	case err != nil:
+		api.WriteError(w, http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
+		return
+	}
```
cadenmarchese, line 34: It might be worth moving this logic into `_getAdminHiveClusterDeployment()` at the moment that you Get the cluster document (line 42)

cadenmarchese (commented):
This is straightforward to me, thanks! A few questions/suggestions looking over the action.

pkg/frontend/admin_hive_clusterdeployment_get.go
s-amann: I like that idea, I'll refactor so the error handling is in one spot and slot this return code in.

pkg/frontend/admin_hive_clusterdeployment_get.go
s-amann: I agree, the error handling can be simplified. I'll get this refactoring in.

s-amann: changes done!

cadenmarchese (approved):

pkg/frontend/admin_hive_clusterdeployment_get.go
petrkotas: We can decide for a different approach later on for all handlers. Lets keep this one concise.

petrkotas (approved):
I am fine with the PR, however I do not see `NewFromEnv` used anywhere. Where is the plan to use it?

pkg/frontend/frontend.go
```diff
@@ -401,6 +405,12 @@ func (f *frontend) authenticatedRoutes(r *mux.Router) {
 
 	s.Methods(http.MethodPost).HandlerFunc(f.postAdminOpenShiftClusterDrainNode).Name("postAdminOpenShiftClusterDrainNode")
 
+	s = r.
```
bennerv, line 408: If we ever decide to expose this via the cluster manager, then we should remove the `/admin` route and use `api-version=admin` logic to route accordingly.

pkg/hive/manager.go
```diff
@@ -51,6 +52,31 @@ type clusterManager struct {
...
+func NewFromEnv(ctx context.Context, log *logrus.Entry, env env.Interface) (ClusterManager, error) {
+	adoptByHive, err := env.LiveConfig().AdoptByHive(ctx)
+	if err != nil {
+		return nil, err
+	}
+	installViaHive, err := env.LiveConfig().InstallViaHive(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if !adoptByHive && !installViaHive {
```
bennerv, line 68: Should this be `!adoptByHive` && `!installViaHive`?  Shouldn't it be one or the other?

pkg/hive/manager.go
```diff
@@ -51,6 +52,31 @@ type clusterManager struct {
...
+		log.Infof("hive is disabled, skipping creation of ClusterManager")
+		return nil, nil
+	}
+	hiveShard := 1
```
bennerv, line 72: should this get a `todo` for when we scale our hive?

bennerv (commented):
Couple comments, otherwise LGTM

pkg/hive/manager.go
bennerv: Nevermind this logic is good.  It's hard to read sometimes :)

bennerv (approved):

PR #2638, opened 2023-01-16
Migrate E2E to 1ES pool

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-1390

Ports E2E to use 1ES.

### Test plan for issue:

It is tests :) It might not work perfectly (e.g. when it crashes), but we can make incremental updates as needed.

N/A

### DIFF
--- a/.pipelines/e2e.yml
+++ b/.pipelines/e2e.yml
@@ -10,8 +10,6 @@ resources:
   - container: container
     image: registry.access.redhat.com/ubi8/toolbox:8.7
     options: --user=0 --cap-add=NET_ADMIN --device /dev/net/tun --name vpn
-    ports:
-      - 8444:8444
   - container: selenium
     image: docker.io/selenium/standalone-edge:latest
     options: --shm-size=2g
@@ -27,8 +25,8 @@ jobs:
   timeoutInMinutes: 180
   pool:
     name: 1es-aro-ci-pool
-  services:
-    selenium: selenium
+  #services:
+  #  selenium: selenium
   steps:
   - template: ./templates/template-checkout.yml
 
@@ -65,8 +63,8 @@ jobs:
       . ./hack/e2e/run-rp-and-e2e.sh
 
       run_vpn
-      run_portal
-      validate_portal_running
+      # run_portal
+      # validate_portal_running
 
       run_rp
       validate_rp_running
--- a/.pipelines/e2e.yml
+++ b/.pipelines/e2e.yml
@@ -10,6 +10,10 @@ resources:
   - container: container
     image: registry.access.redhat.com/ubi8/toolbox:8.7
     options: --user=0 --cap-add=NET_ADMIN --device /dev/net/tun --name vpn
+    ports:
+      - 8443:8443
+      - 8444:8444
+
 
 # Azure DevOps Pipeline running e2e tests
 variables:
@@ -68,6 +72,13 @@ jobs:
       validate_rp_running
 
       register_sub
+    displayName: Setup Tests
+    target: container
+
+  - script: |
+      export CI=true
+      . secrets/env
+      . ./hack/e2e/run-rp-and-e2e.sh
 
       make test-e2e -o e2e.test
     displayName: Execute Tests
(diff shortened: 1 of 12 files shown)


.pipelines/e2e.yml
```diff
@@ -16,26 +20,43 @@ jobs:
...
-    name: ARO-CI
-    demands: go-1.17
+    name: 1es-aro-ci-pool
   steps:
   - template: ./templates/template-checkout.yml
 
+  - script: |
+      set -xe
+      sudo rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+      sudo dnf install -y openvpn make
```
facchettos, line 30: do we need sudo here? (since we are running as user 0)

.pipelines/e2e.yml
```diff
@@ -16,26 +20,43 @@ jobs:
...
-      run_vpn
       deploy_e2e_db
+    displayName: Setup (Azure)
+
+  - script: |
+      export CI=true
+      . secrets/env
+      . ./hack/e2e/run-rp-and-e2e.sh
 
+      run_vpn
```
facchettos, line 59: I am probably missing something, but isn't that function removed?

hack/e2e/run-rp-and-e2e.sh
```diff
@@ -139,16 +137,11 @@ clean_e2e_db() {
...
-    go run ./hack/cluster delete
-}
-
-run_vpn() {
-    sudo openvpn --config secrets/$VPN --daemon --writepid vpnpid
-    sleep 10
-}
-
-kill_vpn() {
-    while read pid; do sudo kill $pid; done < vpnpid
```
facchettos, line 151: removed here

Dockerfile.aro-e2e
```diff
@@ -0,0 +1,17 @@
+# Uses a multi-stage container build to build the RP & E2E components.
+#
+ARG REGISTRY
+FROM ${REGISTRY}/ubi8/go-toolset:1.17.7 AS builder
+ENV GOOS=linux \
+    GOPATH=/go/
+WORKDIR ${GOPATH}/src/github.com/Azure/ARO-RP
+USER root
+COPY . ${GOPATH}/src/github.com/Azure/ARO-RP/
+RUN make aro RELEASE=${IS_OFFICIAL_RELEASE} -o generate && make e2e.test e2etools
```
facchettos, line 10: how do we pass that variable (IS_OFFICIAL_RELEASE) into the container? would it make sense to have an docker arg for it?

Dockerfile.aro-multistage
```diff
@@ -1,18 +1,13 @@
...
 ARG REGISTRY
 FROM ${REGISTRY}/ubi8/go-toolset:1.17.7 AS builder
 ENV GOOS=linux \
     GOPATH=/go/
 WORKDIR ${GOPATH}/src/github.com/Azure/ARO-RP
 USER root
-RUN yum update -y
 COPY . ${GOPATH}/src/github.com/Azure/ARO-RP/
-RUN make aro RELEASE=${IS_OFFICIAL_RELEASE} && make e2e.test
+RUN make aro RELEASE=${IS_OFFICIAL_RELEASE} -o generate && make e2e.test
```
facchettos, line 10: If I got this correctly, that prevents from rerunning make generate right?

facchettos (commented):
looks good! left a few questions to better understand it :)

.pipelines/e2e.yml
hawkowl: there was a duplicate in the file I removed :D

Dockerfile.aro-multistage
hawkowl: yep (since generated stuff should be committed)

Dockerfile.aro-e2e
hawkowl: not a clue, I copied it from the other makefiles :thinking:

.pipelines/e2e.yml
hawkowl: the Azure Pipelines stuff switches our user to the one that it creates after it sets up the container. The --user=0 is so that the initial setup doesn't run as user 1001 as set in the dockerfile.

hack/e2e/run-rp-and-e2e.sh
hawkowl: this was duplicate

facchettos:
I think that portal tests will be disabled with this, because they require docker to be available to them to run selenium. We have to find a solution to run them even if the tests are running inside docker, or run them separately from this. 

Or did I miss something and you found a way to run the portal tests?

facchettos (approved):

PR #2639, opened 2023-01-16
removed gorilla from the gateway

facchettos:
removes gorilla from the gateway and replaces it with http.mux since we are not using any of the advanced features of gorilla and also it is deprecated

Fixes


pkg/deploy/generator/resources_gateway.go
```diff
@@ -138,7 +138,7 @@ func (g *generator) gatewayLB() *arm.Resource {
 					{
 						ProbePropertiesFormat: &mgmtnetwork.ProbePropertiesFormat{
 							Protocol:       mgmtnetwork.ProbeProtocolHTTP,
-							Port:           to.Int32Ptr(80),
+							Port:           to.Int32Ptr(8081),
```
m1kola, line 141: Can we make the same server to serve healthz?

SudoBrendan: I think I'm inclined to agree with @m1kola on this, and while I agree it does add more code to the root handler, it is more idiomatic go, I think. I've seen plenty of `switch` statements on handlers based on method, then routing the response to a secondary handler or other (testable) method:

```
// ServeHTTP implements http.Handler
func (h *SomeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	var err error
	switch r.Method {
    case http.MethodConnect:
		err = h.Proxy(r, w)
    case http.MethodGet:
		err = h.Healthz(r, w)
	default:
		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
		return
	}
	if err != nil {
        // not prod ready, but you get the idea...
		http.Error(w, err.Error(), http.StatusServiceUnavailable)
		return
	}
}
```

This seems like a less invasive/bloated implementation than adding an additional server process... and since we added a check in this PR to validate the HTTP method anyways on the proxy handler, it's computationally equivalent.

facchettos: I'd be curious to see where it is said that this way is more idiomatic. Plus to me it still doesn't make much sense to have the health check for 2 different services on one of the services (why not on the other, why not have it on both? ). If we only reported health for the http proxy I'd be fine having it being served on port 80, but here IMO it makes more sense to have a separate port for health check because we also report the health of the https proxy

cmd/aro/gateway.go
```diff
@@ -90,14 +90,19 @@ func gateway(ctx context.Context, log *logrus.Entry) error {
 		return err
 	}
 
+	healthl, err := utilnet.Listen("tcp", ":8081", pkggateway.SocketSize)
```
bennerv, line 93: Should we have an `EXPOSE` statement for this port for our docker image?

facchettos: I don't think it is necessary, since we have the `-p` directive on the command line. `EXPOSE` if for when you want to have your port available inside a docker network, it is redundant when you want to expose the port to the outside world

pkg/deploy/generator/resources_gateway.go
petrkotas: I have only once concern and that is who is consuming this health endpoint. Where alse we need to update the port?

facchettos: It's for the load balancer to direct traffic to the vm

petrkotas (approved):
This was fun one, but I learned how the stuff is proxied. Good work Jeremy

PR #2642, opened 2023-01-17
Removes dependency on a versioned operator client

m1kola:

Operator is not longer dependant on a version config client and is now using a split client.

### Test plan for issue:

No, just refactoring.

### DIFF
--- a/pkg/operator/controllers/checkers/ingresscertificatechecker/checker_test.go
+++ b/pkg/operator/controllers/checkers/ingresscertificatechecker/checker_test.go
@@ -9,8 +9,6 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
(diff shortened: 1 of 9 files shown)


pkg/operator/controllers/checkers/ingresscertificatechecker/checker_test.go
```diff
@@ -66,23 +64,17 @@ func TestCheck(t *testing.T) {
...
 	} {
 		t.Run(tt.name, func(t *testing.T) {
-			operatorcliFake := operatorfake.NewSimpleClientset()
-			if tt.ingressController != nil {
-				operatorcliFake.Tracker().Add(tt.ingressController)
-			}
-			configcliFake := configfake.NewSimpleClientset()
-			if tt.clusterVersion != nil {
-				configcliFake.Tracker().Add(tt.clusterVersion)
-			}
```
m1kola, line 76: Forgot to remove this bit in https://github.com/Azure/ARO-RP/pull/2636

facchettos (approved):

PR #2644, opened 2023-01-18
Fix/do not adopt hive created clusters during pucm

AldoFusterTurpin:

Fixes [Don't adopt Hive-created clusters during PUCM runs](https://issues.redhat.com/browse/ARO-1447).

We don't want a PUCM run to update/change the Hive clusterdeployment resource for a cluster if that cluster was created by Hive. Doing adoption again would mean deleting existing clusterdeployment already created in Hive and would destroy the exiting state, we don't want that.

This PRs:
- modify the *adminUpdate()* behaviour to adopt clusters that were not created by Hive.
- add new info at cluster installation time (in cosmosDB and in a label in Cluster Deployment) to know if the cluster was created by hive or not. This is new info is used in the new adminUpdate() behaviour. 
- adds some unit tests.

### Test plan for issue:

Added Unit Tests. I created a separate commit to refactor *(m *manager) openShiftVersionFromVersion()* and extracted the funcitonality to a separate interface and struct in order to being able to mock that function in the newly created test *TestRunHiveInstallerSetsCreatedByHiveFieldToTrueInClusterDoc()* (needed to ensure the *createdByHive* field is properly set when a cluster is created by Hive). I also added a new test case for the test TestAdminUpdateSteps to skip adoption of clusters created by hive.

After the refactoring, a new struct field is added to the cluster manager with the functionality that previously was in *(m *manager) openShiftVersionFromVersion* in *pkg/cluster/install_version.go*. It has been moved to its own file, struct and interface. I have done that to be able to mock a function but this is a win by itself as the cluster manager is TOO big and complicated and we are using it as a GOD class. 

@UlrichSchlueter and I have tested the functionality manually and end-to-end with both a regular ARO cluster and a cluster created by Hive. We have ensured that the adoption is just performed in the cluster that was not created by hive (and a few more things). So all automated tests are green and the manual tests we performed were also "green" for the new functionality. 🟢 

N/A.

### DIFF
--- a/pkg/cluster/install.go
+++ b/pkg/cluster/install.go
@@ -152,7 +152,6 @@ func (m *manager) adminUpdate() []steps.Step {
 }
 
 func (m *manager) clusterWasCreatedByHive() bool {
-	// Check if the below "if" is still true!
 	if m.doc.OpenShiftCluster.Properties.HiveProfile.Namespace == "" {
 		return false
 	}
@@ -204,7 +203,7 @@ func (m *manager) runIntegratedInstaller(ctx context.Context) error {
 
 func (m *manager) runHiveInstaller(ctx context.Context) error {
 	var err error
-	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, createdByHiveMutator())
+	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, setFieldCreatedByHive(true))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


pkg/cluster/install.go
```diff
@@ -150,6 +151,14 @@ func (m *manager) adminUpdate() []steps.Step {
 	return toRun
 }
 
+func (m *manager) clusterWasCreatedByHive() bool {
```
cadenmarchese, line 154: Would something simpler work, or are there cases where `Namespace` is not empty but`CreatedByHive` is false?

```suggestion
func (m *manager) clusterWasCreatedByHive() bool {
	return m.doc.OpenShiftCluster.Properties.HiveProfile.Namespace != "" 
}
```

AldoFusterTurpin: Good question. 
Short answer: no, we can't.
Long answer:

As this discussed in the slack [thread](https://redhat-internal.slack.com/archives/CE3ETN3J8/p1673621733329919), checking just for the namespace is not enough. We need to differentiate if it was created by hive vs adopted by hive.

As hive doesn't have this information, wee need to keep track of that. That's why we (@UlrichSchlueter and me) decided to place the information in the cluster document in CosmosDB and in a label in the cluster deployment during install time (when the cluster is created via Hive).

cadenmarchese (approved):

petrkotas (approved):
Thank you @AldoFusterTurpin for this PR. Good work!

PR #2646, opened 2023-01-19
Added fix for diff issue in MacOS

sankur-codes:

The documentation for testing has been updated with the fix to the issue

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -8,7 +8,7 @@ To run RP unit tests:
 make test-go
 ```
 
-In case of MacOS, the go-diff module creates [issue](https://github.com/golangci/golangci-lint/issues/3087) making the test fail. Till a new release of the module with the [fix](https://github.com/sourcegraph/go-diff/pull/65) is available, an easy workaround to mitigate the issue is to install diffutils using `brew install diffutils`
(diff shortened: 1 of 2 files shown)


SudoBrendan (changes requested):
Thanks for documenting this fix. If an additional tool is required for local dev, we should also add it to our docs here: https://github.com/Azure/ARO-RP/blob/master/docs/prepare-your-dev-environment.md#macos-packages

sankur-codes:
> Thanks for documenting this fix. If an additional tool is required for local dev, we should also add it to our docs here: https://github.com/Azure/ARO-RP/blob/master/docs/prepare-your-dev-environment.md#macos-packages

Thanks for pointing out. Done with the addition.

petrkotas (approved):
LGTM

docs/testing.md
```diff
@@ -8,6 +8,8 @@ To run RP unit tests:
 make test-go
 ```
 
+In case of MacOS, the go-diff module creates [issue](https://github.com/golangci/golangci-lint/issues/3087) making the test fail. Till a new release of the module with the [fix](https://github.com/sourcegraph/go-diff/pull/65) is available, an easy workaround to mitigate the issue is to install diffutils using `brew install diffutils`
```
bennerv, line 11: Super nitpicky here:
```suggestion
In case of MacOS, the go-diff module creates [issue](https://github.com/golangci/golangci-lint/issues/3087) making the test fail. Until a new release of the module with the [fix](https://github.com/sourcegraph/go-diff/pull/65) is available, an easy workaround to mitigate the issue is to install diffutils using `brew install diffutils`

bennerv (changes requested):
one nitpick, otherwise lgtm

bennerv (approved):

PR #2647, opened 2023-01-19
Increase nodeStartupTimeout for the Machine Health Check

LiniSusan:
https://issues.redhat.com/browse/ARO-1814

Fixes

Added a new field to the file: `pkg/operator/controllers/machinehealthcheck/staticresources/machinehealthcheck.yaml `- 
`nodeStartupTimeout: "20m"` - to specify the timeout duration that a machine health check must wait for a node to join the cluster before a machine is determined to be unhealthy.
The current issue is that the alert is received, then it gets auto-mitigated at some point and SREs don't really take any action on it. So trying to reduce the number of auto-mitigating alerts

### Test plan for issue:
Will monitor how frequently the auto-mitigating alerts are being generated after the change and whether it is affecting other MHC alert scenarios

Yes, past three months MHC alert data is uploaded in the Jira ticket mentioned above.

### DIFF
--- a/pkg/operator/controllers/machinehealthcheck/staticresources/machinehealthcheck.yaml
+++ b/pkg/operator/controllers/machinehealthcheck/staticresources/machinehealthcheck.yaml
@@ -21,3 +21,4 @@ spec:
     timeout: "300s"
     status: "Unknown"
   maxUnhealthy: "1"
+  nodeStartupTimeout: "20m"


cadenmarchese (approved):
+1, I have seen a few incidents where 10m isn't enough time to start a node from scratch, especially if there are compute issues in the region or if an alternate vm size is used

darthhexx (approved):

PR #2648, opened 2023-01-19
Removes dependency on a versioned machine client

m1kola:

Operator is not longer dependant on a versioned machine client and is now using a split client which uses cache for read operations.

### Test plan for issue:

No, just refactoring.


bennerv (approved):
lgtm

PR #2649, opened 2023-01-19
Update subnet generation for local dev

karanmagdani:

Since the introduction of aks-vnet and dev-vpn-vnet. Dev-vnet has /16 address space here https://github.com/Azure/ARO-RP/pull/2238.

Development clusters use address space of dev-vnet to create subnets, changing the generatesubnets function to allow clusters to be created within the /16 address space. This limits the number of clusters we can create to 128

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -339,12 +339,15 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 
 func (c *Cluster) generateSubnets() (vnetPrefix string, masterSubnet string, workerSubnet string) {
 	// pick a random 23 in range [10.3.0.0, 10.127.255.0]
-	// 10.0.0.0 is used by dev-vnet to host CI
-	// 10.1.0.0 is used by rp-vnet to host Proxy VM
-	// 10.2.0.0 is used by dev-vpn-vnet to host VirtualNetworkGateway
+	// 10.0.0.0/16 is used by dev-vnet to host CI
+	// 10.1.0.0/24 is used by rp-vnet to host Proxy VM
+	// 10.2.0.0/24 is used by dev-vpn-vnet to host VirtualNetworkGateway
 	var x, y int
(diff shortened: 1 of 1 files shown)


karanmagdani:
This will also facilitate removal of proxy here: https://github.com/Azure/ARO-RP/pull/2396

karanmagdani:
CI is failing with 
```
##[warning]Failed to download task '1ESHostedPoolValidation'. Error Unable to read data from the transport connection: Connection reset by peer.
##[warning]Inner Exception: Connection reset by peer
##[warning]Back off 22.994 seconds before retry.
```

jhoreman (approved):
As discussed over the past few days. These changes look good and will help keep our dev subscription deployment sane.

pkg/util/cluster/cluster.go
```diff
@@ -339,12 +339,15 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
...
+	// 10.2.0.0/24 is used by dev-vpn-vnet to host VirtualNetworkGateway
 	var x, y int
 	rand.Seed(time.Now().UnixNano())
-	for x == 0 && y == 0 {
+	// Local Dev clusters are limited to /16 dev-vnet
+	if !c.ci {
+		x, y = 0, 2*rand.Intn(128)
+	} else {
 		x, y = rand.Intn((124))+3, 2*rand.Intn(128)
 	}
```
facchettos, line 352: what is the reason behind only using even numbers ?

karanmagdani: i just changed the `x` to have 10.0.y.0 so it falls under the /16 instead of /9 address space. 

But I assume its only even numbers because the worker subnet takes the odd number right after 
```workerSubnet = fmt.Sprintf("10.%d.%d.0/24", x, y+1)```

facchettos: thanks! then lgtm

facchettos (approved):

PR #2651, opened 2023-01-23
Bump image mirror tags

hawkowl:

Need the Mariner container for #2520 

Updates image mirroring, adds Mariner

### Test plan for issue:

Shouldn't need any

N/A

### DIFF
--- a/cmd/aro/mirror.go
+++ b/cmd/aro/mirror.go
@@ -152,18 +152,20 @@ func mirror(ctx context.Context, log *logrus.Entry) error {
 	}
 
 	for _, ref := range []string{
-		"registry.redhat.io/rhel7/support-tools:latest",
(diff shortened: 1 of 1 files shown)


SudoBrendan (approved):
These changes look fine, and while I agree tests won't change, it would be good if we could document the process for safely deprecating images from this list, I remember hearing that we keep images for a reason, but I can't remember exactly why... maybe building old images in emergencies?

hawkowl:
@SudoBrendan The images won't "go away" if they're not synced, and since we have a global registry, they'll be available in other regions. We're just not interested in updating the given tag anymore. Maybe we do need to have a method of cleaning out old images, however.

bennerv (approved):
lgtm - thanks for adding relevant links

PR #2653, opened 2023-01-23
Update old dbtoken-service documentation

bennerv:

Fixes old dbtoken-service docs that aren't relevant.

### DIFF
--- a/docs/dbtoken-service.md
+++ b/docs/dbtoken-service.md
@@ -47,10 +47,26 @@ in depth in the face of an attack on the gateway component.
 
 ## Setup
 
-* At rollout time, create an AAD application whose *Application ID URI*
-  (`identifierUris` in the application manifest) is
-  `https://dbtoken.aro.azure.com`.  It is not necessary for the application to
-  have any permissions, credentials, etc.
+* Create the application and set `requestedAccessTokenVersion`
+
+   ```bash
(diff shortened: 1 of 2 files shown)


mbarnes (approved):
Thanks for this!  LGTM

docs/dbtoken-service.md
```diff
@@ -47,10 +47,26 @@ in depth in the face of an attack on the gateway component.
...
+
+   ```bash
+   AZURE_DBTOKEN_CLIENT_ID="$(az ad app create --display-name dbtoken \
+      --oauth2-allow-implicit-flow false \
+      --query appId \
+      -o tsv)"
+
+   OBJ_ID="$(az ad app show --id $AZURE_DBTOKEN_CLIENT_ID --query id)"
+
+   > __NOTE:__: the graph API requires this to be done from a managed machine
```
SudoBrendan, line 60: super-nit, but this doesn't format correctly in github, should we just make it a `#` instead of `>__NOTE:__`?

SudoBrendan (approved):
one nit, lgtm

docs/dbtoken-service.md
cadenmarchese: I'm seeing that too - it looks like the note is meant to be outside of the markdown and instead it's getting included in the markdown.

bennerv: Done - good catch.  Thank you :)

PR #2624, opened 2023-01-09
Switch CI to running on 1ES

hawkowl:
### Which issue this PR addresses:

Fixes https://issues.redhat.com/browse/ARO-1401 maybe

PR #2654, opened 2023-01-24
Run CI on the 1es-aro-ci-pool pool

hawkowl:
Amending #2624  to use the final pool

### Test plan for issue:
Should pass because it is tests :)


darthhexx (approved):
Passed the tests of the pools that were changed.

PR #2655, opened 2023-01-24
Change az aro create install-version argument to version

darthhexx:

Changes the `az aro create` CLI argument for multi-version from `install-version` to `version`.

This change was requested from product management to ensure a consistent experience for end-users, because we list the versions as simply `Version` and also the ARM template cluster creation accepts `version`.

### Test plan for issue:

Tested cluster creation and version validation manually.

Yes. The documentation for multi-version should specify `version` and not `install-version`.


SudoBrendan (approved):
lgtm

cadenmarchese (approved):

PR #2656, opened 2023-01-24
Track debug pods

yehornaumenko:

[Detect customer debug pod activity](https://issues.redhat.com/browse/ARO-1697)

We want to have this metric to track debug pod activity

### Test plan for issue:

Unit tests

TBD

### DIFF
--- a/pkg/monitor/cluster/debugpods.go
+++ b/pkg/monitor/cluster/debugpods.go
@@ -1,0 +1,89 @@
+package cluster
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	kuval "k8s.io/apimachinery/pkg/util/validation"
+
+	"github.com/Azure/ARO-RP/pkg/util/stringutils"
+)
+
+// Each cluster is being checked every 1 minute. emitDebugPodsCount
+// lists all events (age < 120 seconds) in the default namespace
+// in order to detect recent events of debug-pod creation.
+func (mon *Monitor) emitDebugPodsCount(ctx context.Context) error {
+	m := map[string]string{
+		"reason":              "Started",
+		"involvedObject.kind": "Pod",
+		// oc debug node creates a pod with a "container-00"
+		// container name.
+		"involvedObject.fieldPath": "spec.containers{container-00}",
+	}
+	lo := metav1.ListOptions{
+		FieldSelector: fields.SelectorFromSet(m).String(),
+	}
+	events, err := mon.cli.CoreV1().Events("default").List(ctx, lo)
+	if err != nil {
+		return err
+	}
+
+	nodes, err := mon.cli.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	podNames := getDebugPodNames(nodes.Items)
+
+	// User can spawn a debug pod, make some activity and delete it
+	// in the short timespan when the monitoring go routine is sleeping.
+	// That's why we check the events of debug pod creation, not pods.
+	count := countDebugPods(podNames, events.Items)
+	mon.emitGauge("debugpods.count", int64(count), nil)
+
+	return nil
+}
+
+func countDebugPods(debugPods []string, events []corev1.Event) int {
+	count := 0
+	for _, e := range events {
+		if eventIsNew(e) && stringutils.Contains(debugPods, e.InvolvedObject.Name) {
+			count++
+		}
+	}
+	return count
+}
+
+func eventIsNew(event corev1.Event) bool {
+	return time.Since(event.LastTimestamp.Time) < time.Second*120
+}
+
+func getDebugPodNames(nodes []corev1.Node) []string {
+	names := make([]string, len(nodes))
+	for i, n := range nodes {
+		names[i] = fmt.Sprintf("%s-debug", MakeSimpleName(n.Name))
+	}
+	return names
+}
+
+// MakeSimpleName is a copy of the function that is used by oc CLI tool
+// to generate names for the debug pods.
+func MakeSimpleName(name string) string {
+	invalidServiceChars := regexp.MustCompile("[^-a-z0-9]")
+
+	name = strings.ToLower(name)
+	name = invalidServiceChars.ReplaceAllString(name, "")
+	name = strings.TrimFunc(name, func(r rune) bool { return r == '-' })
+	if len(name) > kuval.DNS1035LabelMaxLength {
+		name = name[:kuval.DNS1035LabelMaxLength]
+	}
+	return name
+}
--- a/pkg/monitor/cluster/debugpods_test.go
+++ b/pkg/monitor/cluster/debugpods_test.go
@@ -1,0 +1,182 @@
+package cluster
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/golang/mock/gomock"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+	ktesting "k8s.io/client-go/testing"
+
+	mock_metrics "github.com/Azure/ARO-RP/pkg/util/mocks/metrics"
+)
+
+func TestEmitDebugPodsCount(t *testing.T) {
+	ctx := context.Background()
+
+	cli := fake.NewSimpleClientset(
+		&corev1.Node{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "master-1",
+			},
+		},
+		&corev1.Node{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "master-2",
+			},
+		},
+	)
+
+	cli.PrependReactor("list", "events", reactorFn)
+
+	controller := gomock.NewController(t)
+	defer controller.Finish()
+
+	m := mock_metrics.NewMockEmitter(controller)
+	mon := &Monitor{
+		cli: cli,
+		m:   m,
+	}
+
+	m.EXPECT().EmitGauge("debugpods.count", int64(1), map[string]string{})
+	err := mon.emitDebugPodsCount(ctx)
+	if err != nil {
+		t.Fatalf("got unexpected error: %v", err)
+	}
+}
+
+func reactorFn(_ ktesting.Action) (handled bool, ret kruntime.Object, err error) {
+	now := metav1.Now()
+	longAgo := metav1.Date(1991, time.August, 24, 0, 0, 0, 0, time.UTC)
+	return true, &corev1.EventList{Items: []corev1.Event{
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "master-1-started",
+				Namespace: "default",
+			},
+			Reason: "Started",
+			InvolvedObject: corev1.ObjectReference{
+				Kind: "Pod",
+				Name: "master-1-debug",
+			},
+			LastTimestamp: longAgo,
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "master-2-started",
+				Namespace: "default",
+			},
+			Reason: "Started",
+			InvolvedObject: corev1.ObjectReference{
+				Kind: "Pod",
+				Name: "master-2-debug",
+			},
+			LastTimestamp: now,
+		},
+	}}, nil
+}
+
+func TestEventIsNew(t *testing.T) {
+	for _, tt := range []struct {
+		event corev1.Event
+		isNew bool
+	}{
+		{
+			event: corev1.Event{
+				LastTimestamp: metav1.Now(),
+			},
+			isNew: true,
+		},
+		{
+			event: corev1.Event{
+				LastTimestamp: metav1.NewTime(metav1.Now().Add(-1 * time.Minute)),
+			},
+			isNew: true,
+		},
+		{
+			event: corev1.Event{
+				LastTimestamp: metav1.Date(2020, 02, 18, 0, 0, 0, 0, time.UTC),
+			},
+			isNew: false,
+		},
+	} {
+		if tt.isNew != eventIsNew(tt.event) {
+			t.Fatalf("test failed for the event: %v", tt.event)
+		}
+	}
+}
+
+func TestGetDebugPodNames(t *testing.T) {
+	for _, tt := range []struct {
+		node                 corev1.Node
+		expectedDebugPodName string
+	}{
+		{
+			node: corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "master-0",
+				},
+			},
+			expectedDebugPodName: "master-0-debug",
+		},
+		{
+			node: corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "-master-",
+				},
+			},
+			expectedDebugPodName: "master-debug",
+		},
+	} {
+		if tt.expectedDebugPodName != getDebugPodNames([]corev1.Node{tt.node})[0] {
+			t.Fatalf("test failed for the node: %v", tt.node.Name)
+		}
+	}
+}
+
+func TestCountDebugPods(t *testing.T) {
+	now := metav1.Now()
+	longAgo := metav1.Date(2020, 02, 18, 0, 0, 0, 0, time.UTC)
+
+	for _, tt := range []struct {
+		debugPodNames []string
+		events        []corev1.Event
+		expectedCount int
+	}{
+		{
+			debugPodNames: []string{"m-0-debug", "m-1-debug", "m-2-debug", "m-3-debug"},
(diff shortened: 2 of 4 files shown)


pkg/monitor/cluster/debugpods.go
```diff
@@ -0,0 +1,69 @@
...
+			count++
+		}
+	}
+	return count
+}
+
+func eventIsNew(event corev1.Event) bool {
+	// corev1.Event.LastTimeStamp is deprecated in newer k8s releases.
+	// Update this check when ARO's k8s version is upgraded
+	return time.Since(event.LastTimestamp.Time) < time.Second*70
```
yehornaumenko, line 60: The monitoring check delay [is set to 1 minute](https://github.com/Azure/ARO-RP/blob/master/pkg/monitor/cache.go#L68). It means that [all monitoring emit methods](https://github.com/Azure/ARO-RP/blob/master/pkg/monitor/cluster/cluster.go#L144) are called every X minutes, where X = 1 minute + the execution time of all other emit methods.
I think it makes sense to check events for the last 5-10 minutes, but I am not sure what is the best value.

pkg/monitor/cluster/debugpods.go
```diff
@@ -0,0 +1,69 @@
...
+func (mon *Monitor) emitDebugPodsCount(ctx context.Context) error {
+	events, err := mon.cli.CoreV1().Events("default").List(ctx, metav1.ListOptions{
+		FieldSelector: fields.SelectorFromSet(map[string]string{
+			"reason":              "Started",
+			"involvedObject.kind": "Pod",
+			// oc debug node creates a pod with a "container-00"
+			// container name.
+			"involvedObject.fieldPath": "spec.containers{container-00}",
+		}).String(),
+	})
```
AldoFusterTurpin, line 30: The nested fields + .String() call make it a little bit difficult to read.

What about something like:
```go
m := map[string]string{
	"reason":                   "Started",
	"involvedObject.kind":      "Pod",
	"involvedObject.fieldPath": "spec.containers{container-00}",
}

lo := metav1.ListOptions{
    FieldSelector: fields.SelectorFromSet(m).String(),
}

events, err := mon.cli.CoreV1().Events("default").List(ctx, lo)
```

pkg/monitor/cluster/debugpods.go
```diff
@@ -0,0 +1,69 @@
...
+
+func countDebugPods(debugPods []string, events []corev1.Event) int {
+	count := 0
+	for _, e := range events {
+		if eventIsNew(e) && stringutils.Contains(debugPods, e.InvolvedObject.Name) {
+			count++
+		}
+	}
+	return count
+}
```
AldoFusterTurpin, line 55: I really like the split in narrow functions. 😄 

If we know that corev1.Event.LastTimeStamp is deprecated in newer k8s releases and we will need to update this check when ARO's k8s version is upgraded, I think we can modify the code a bit to be easier to maintain.

What about something like:
```go
func countDebugPods(debugPods []string, events []corev1.Event) int {
	count := 0
	for _, e := range events {
		// corev1.Event.LastTimeStamp is deprecated in newer k8s releases.
		// Update this check when ARO's k8s version is upgraded
		t := e.LastTimestamp.Time
		if eventIsNew(t) && stringutils.Contains(debugPods, e.InvolvedObject.Name) {
			count++
		}
	}
	return count
}
```

pkg/monitor/cluster/debugpods.go
AldoFusterTurpin, line 60: Following with the last comment, the function can now be updated to not depend on any change in k8s as we pass as a parameter a standard time.Time

```go
func eventIsNew(t time.Time) bool {
	return time.Since(t) < AgeTargetEvent
}
```

And we can also extract the AgeTargetEvent into a constant in top of the file (for easier modification in the future):
```go
const (
	AgeTargetEvent = time.Second * 70
)
```
or we can pass to eventIsNew a second argument instead of extracting it into a constant, we have different options that could also work here.

pkg/monitor/cluster/debugpods_test.go
```diff
@@ -0,0 +1,82 @@
...
+				},
+				Reason: "Started",
+				InvolvedObject: corev1.ObjectReference{
+					Kind: "Pod",
+					Name: "master-2-debug",
+				},
+				LastTimestamp: metav1.NewTime(now),
+			},
+		}}, nil
+	})
```
AldoFusterTurpin, line 66: Anonymous functions can become really ugly in the long run. I propose a simple change to extract it to a variable.
```go
prependReactorFn := func(action ktesting.Action) (handled bool, ret kruntime.Object, err error) {
        return true, &corev1.EventList{Items: []corev1.Event{
	        {
			        Name:      "master-1-started",
			        Namespace: "default",
		        },
		        Reason: "Started",
		        InvolvedObject: corev1.ObjectReference{
			        Kind: "Pod",
			        Name: "master-1-debug",
		        },
		        LastTimestamp: metav1.NewTime(longAgo),
	        },
	        {
			        Name:      "master-2-started",
			        Namespace: "default",
		        },
		        Reason: "Started",
		        InvolvedObject: corev1.ObjectReference{
			        Kind: "Pod",
			        Name: "master-2-debug",
		        },
		        LastTimestamp: metav1.NewTime(now),
	        },
        }}, nil
}
cli.PrependReactor("list", "events", prependReactorFn)
```

AldoFusterTurpin (approved):
LGTM! 🚀 
I really like the split of things in short functions! 😃 

I just left a few suggestions that thing that can make the code a bit easier to maintain in the long run in case you think any of them make sense.

Thanks!

pkg/monitor/cluster/debugpods.go
AldoFusterTurpin: Note: the reason I put e.LastTimestamp.Time in a different line instead of inlining it in the function call below it is to have a narrower scope of change in the future and not affect lines that shouldn't be affected by that type of change.

facchettos: is it being deprecated? do you have a link to it?

AldoFusterTurpin: I just trusted the comment on [pkg/monitor/cluster/debugpods.go](https://github.com/Azure/ARO-RP/pull/2656/files#diff-e6aaf68ce82f0311e70154300b03c0a6a7f8c3cb614795f64289a782bf9db4edR58) 😄

pkg/monitor/cluster/debugpods.go
yehornaumenko: good idea!

pkg/monitor/cluster/debugpods.go
rolandmkunkel: If we refactor it like this, the function name `eventIsNew` doesn't make sense anymore. I would say it is actually better to only change the lower level function `eventIsNew` if `LastTimestamp` is removed.

pkg/monitor/cluster/debugpods.go
```diff
@@ -0,0 +1,69 @@
...
+			// oc debug node creates a pod with a "container-00"
+			// container name.
+			"involvedObject.fieldPath": "spec.containers{container-00}",
+		}).String(),
+	})
+	if err != nil {
+		return err
+	}
+
+	nodes, err := mon.cli.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
```
rolandmkunkel, line 35: this monitor and `emitNodeConditions` both (and possibly more) list nodes on the cluster, duplicating these calls. I think that should be refactored so that we do these calls only once. However, probably not in the scope of your PR

pkg/monitor/cluster/debugpods.go
```diff
@@ -0,0 +1,69 @@
...
+func eventIsNew(event corev1.Event) bool {
+	// corev1.Event.LastTimeStamp is deprecated in newer k8s releases.
+	// Update this check when ARO's k8s version is upgraded
+	return time.Since(event.LastTimestamp.Time) < time.Second*70
+}
+
+func getDebugPodNames(nodes []corev1.Node) []string {
+	names := make([]string, len(nodes))
+	for i, n := range nodes {
+		names[i] = fmt.Sprintf("%s-debug", n.Name)
```
rolandmkunkel, line 66: I believe this has a low chance to not work in case the node names are weird. If you check how the debug pod name is created in source code for `oc debug`, it does this:
`fmt.Sprintf("%s-debug", generateapp.MakeSimpleName(infos[0].Name))` (see
https://github.com/openshift/oc/blob/master/pkg/cli/debug/debug.go#L446) with `MakeSimpleName` stripping some characters. Could we use the same helperfunction? https://github.com/openshift/oc/blob/f94afb52dc8a3185b3b9eacaf92ec34d80f8708d/pkg/helpers/newapp/app/pipeline.go#L327

pkg/monitor/cluster/debugpods.go
```diff
@@ -0,0 +1,69 @@
...
+	"k8s.io/apimachinery/pkg/fields"
+
+	"github.com/Azure/ARO-RP/pkg/util/stringutils"
+)
+
+// Each cluster is being checked every 1 minute. emitDebugPodsCount
+// lists all events (age < 70 seconds) in the default namespace
+// in order to detect recent events of debug-pod creation.
+func (mon *Monitor) emitDebugPodsCount(ctx context.Context) error {
+	events, err := mon.cli.CoreV1().Events("default").List(ctx, metav1.ListOptions{
```
rolandmkunkel, line 22: Could we run this query after generating the pod names for debug pods and query only for events related to the pods? Then we would save looping over all events another time to check the name of the involved object?

pkg/monitor/cluster/debugpods.go
rolandmkunkel: maybe the chance isn't even that low, because it also lowercases the name

pkg/monitor/cluster/debugpods.go
yehornaumenko: @AldoFusterTurpin the deprecation issue I referred to was closed several years ago, so these fields will not be deprecated. Sorry for this :)

pkg/monitor/cluster/debugpods_test.go
yehornaumenko: agreed 😎

pkg/monitor/cluster/debugpods.go
yehornaumenko: @rolandmkunkel Yeah, I have thought about this kind of duplicate API calls. I guess we should balance the code readability and the performance here. Could you please provide an example of how you see the refactoring here?

pkg/monitor/cluster/debugpods.go
yehornaumenko: @rolandmkunkel I have been thinking about this, but I haven't found an appropriate FieldSelector for filtering the events this way

pkg/monitor/cluster/debugpods.go
petrkotas: The thing is called informer https://aly.arriqaaq.com/kubernetes-informers/ and I believe it should be able to watch events. 

More resources:
- https://www.cncf.io/blog/2019/10/15/extend-kubernetes-via-a-shared-informer/
- https://pkg.go.dev/k8s.io/client-go/informers

pkg/monitor/cluster/debugpods.go
```diff
@@ -0,0 +1,72 @@
...
+func eventIsNew(event corev1.Event) bool {
+	// corev1.Event.LastTimeStamp is deprecated in newer k8s releases.
+	// Update this check when ARO's k8s version is upgraded
+	return time.Since(event.LastTimestamp.Time) < time.Second*120
+}
+
+func getDebugPodNames(nodes []corev1.Node) []string {
+	names := make([]string, len(nodes))
+	for i, n := range nodes {
+		names[i] = fmt.Sprintf("%s-debug", oc.MakeSimpleName(n.Name))
```
petrkotas, line 69: Do I understand correctly, we use `oc` only for SimpleName?
If yes, how would you solve the problem without `oc`?
I would rather not vendor in so many dependencies just for this tiny function. Thank you.

petrkotas (changes requested):
Few nits.

yehornaumenko:
@petrkotas thanks for the review. The Informers feature is really cool, but it looks like it is impossible to use it in the combination with the current monitoring approach we have. Informers are supposed to run in parallel as a goroutine by design, but we have a `time.Sleep` thing in the each of goroutines we run for the monitoring of each cluster. Theoretically, I can add a new goroutine that will do this, but I think informers should be added as a part of the monitoring redesign PR. (It can be done with a fix of the bug @rolandmkunkel found 2-3 weeks ago)
So I suggest adding informers in another PR. Does it make sense to you?

pkg/monitor/cluster/debugpods.go
yehornaumenko: Yeah, we need it only for this function.
This problem can be solved by looping through all pod names and checking the last 6 symbols of their names (`-debug`). It is not a super reliable solution, but it will work unless users try to confuse us consciously.
I agree that importing this library is overkill. Are you ok with the idea I proposed?

petrkotas (approved):
thank you @yehornaumenko good work, lets get it in.

petrkotas:
```
[BeforeSuite]! [PANICKED] [55.183 seconds]
[BeforeSuite] 
/go/src/github.com/Azure/ARO-RP/test/e2e/setup.go:501

  Test Panicked
  In [BeforeSuite] at: /go/src/github.com/Azure/ARO-RP/test/e2e/setup.go:508

  redhatopenshift.OpenShiftClustersClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="ResourceQuotaExceeded" Message="Resource quota of standardDSv3Family exceeded. Maximum allowed: 600, Current in use: 586, Additional requested: 30."

  Full Stack Trace
    github.com/Azure/ARO-RP/test/e2e.glob..func33()
    	/go/src/github.com/Azure/ARO-RP/test/e2e/setup.go:508 +0xbf
------------------------------
[AfterSuite] 
/go/src/github.com/Azure/ARO-RP/test/e2e/setup.go:512
```

we are hitting the corecpu limit

pkg/monitor/cluster/debugpods.go
```diff
@@ -0,0 +1,91 @@
...
+			count++
+		}
+	}
+	return count
+}
+
+func eventIsNew(event corev1.Event) bool {
+	// corev1.Event.LastTimeStamp is deprecated in newer k8s releases.
+	// Update this check when ARO's k8s version is upgraded
+	return time.Since(event.LastTimestamp.Time) < time.Second*120
```
s-amann, line 68: Can we increase the code coverage for these helper methods?

s-amann (changes requested):
nice work, the code is easy to read. Please add some more unit tests around the helper functions since they are nicely broken out.

pkg/monitor/cluster/debugpods.go
rolandmkunkel, line 68: why is this 120 seconds now? the comment on the function still says 70 seconds

pkg/monitor/cluster/debugpods_test.go
```diff
@@ -0,0 +1,83 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "master-2",
+			},
+		},
+	)
+
+	now := time.Now()
+	longAgo := time.Date(1991, time.August, 24, 0, 0, 0, 0, time.UTC)
+	reactorFn := func(action ktesting.Action) (handled bool, ret kruntime.Object, err error) {
+		return true, &corev1.EventList{Items: []corev1.Event{
```
rolandmkunkel, line 40: I think there should also be a test that tests that pod started events for non debug pods are not counted

rolandmkunkel (changes requested):

pkg/monitor/cluster/debugpods_test.go
yehornaumenko: It is not yet possible to test k8s API call settings, including field-selector
https://github.com/kubernetes/client-go/issues/326

pkg/monitor/cluster/debugpods.go
yehornaumenko: I have updated the comment. The timeout for doing all emit monitor funcs is 50 seconds. There is a chance that we count the same event twice, but it is better than skipping it.

yehornaumenko: @s-amann  I have added some tests, could you check it, please?

pkg/util/stringutils/stringutils.go
```diff
@@ -9,3 +9,12 @@ import "strings"
 func LastTokenByte(s string, sep byte) string {
 	return s[strings.LastIndexByte(s, sep)+1:]
 }
+
+func Contains(list []string, value string) bool {
```
rolandmkunkel, line 13: one day this can be https://pkg.go.dev/golang.org/x/exp/slices#Contains

rolandmkunkel (dismissed):

petrkotas (dismissed):
Good one. LGTM

facchettos (approved):

pkg/monitor/cluster/debugpods_test.go
```diff
@@ -0,0 +1,182 @@
...
+			t.Fatalf("test failed for the node: %v", tt.node.Name)
+		}
+	}
+}
+
+func TestCountDebugPods(t *testing.T) {
+	now := metav1.Now()
+	longAgo := metav1.Date(2020, 02, 18, 0, 0, 0, 0, time.UTC)
+
+	for _, tt := range []struct {
```
s-amann, line 148: We need a loop with only one test in it, I think this is fine for now but lets consider either removing the loop or adding additional tests

s-amann (approved):
lgtm

petrkotas:
We can merge this one, as soon as the E2E passes.

petrkotas (approved):
re approve becuse of rebase

facchettos (approved):

PR #2658, opened 2023-01-26
updated docs to reflect new AAD application after migration

karanmagdani:
Signed-off-by: Karan <kmagdani@redhat.com>

Updates AAD applications post-migration

doc update


bennerv (commented):
Should we publish these APP IDs?  I guess they're already out there, but are they necessary to specify the exact APP ID? 

We could instead point to: This appid is contained within x environment variable in our secrets for a given environment.  

The data isn't secret, but could be sensitive.

karanmagdani:
> Should we publish these APP IDs? I guess they're already out there, but are they necessary to specify the exact APP ID?
> 
> We could instead point to: This appid is contained within x environment variable in our secrets for a given environment.
> 
> The data isn't secret, but could be sensitive.

I can update the doc to remove the app ID and instead provide the location for where to find them in secrets

bennerv (approved):

PR #2661, opened 2023-01-26
Add validation and unrestricted endpoint for kubernetes objects

cblecker:

Fixes ARO-2148

Updates `getAdminKubernetesObjects` to only work for system level objects (specific clusterwide objects, and openshift system namespaces). Adds a `getAdminKubernetesObjectsUnrestricted` method and route to allow us to bypass this in specific circumstances.

This also moves us to use standard client-go restmapper libraries instead of custom ones to resolve cluster objects.

### DIFF
--- a/pkg/frontend/admin_openshiftcluster_kubernetesobjects_test.go
+++ b/pkg/frontend/admin_openshiftcluster_kubernetesobjects_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/golang/mock/gomock"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/env"
@@ -52,6 +53,7 @@ func TestAdminKubernetesObjectsGetAndDelete(t *testing.T) {
 				k.EXPECT().
 					KubeGet(gomock.Any(), tt.objKind, tt.objNamespace, tt.objName).
 					Return([]byte(`{"Kind": "test"}`), nil)
+				k.EXPECT().ResolveGVR(tt.objKind).Return(&schema.GroupVersionResource{Resource: "configmaps"}, nil)
 			},
 			wantStatusCode: http.StatusOK,
 			wantResponse:   []byte(`{"Kind": "test"}` + "\n"),
@@ -66,6 +68,7 @@ func TestAdminKubernetesObjectsGetAndDelete(t *testing.T) {
 				k.EXPECT().
 					KubeList(gomock.Any(), tt.objKind, tt.objNamespace).
 					Return([]byte(`{"Kind": "test"}`), nil)
+				k.EXPECT().ResolveGVR(tt.objKind).Return(&schema.GroupVersionResource{Resource: "configmaps"}, nil)
 			},
 			wantStatusCode: http.StatusOK,
 			wantResponse:   []byte(`{"Kind": "test"}` + "\n"),
@@ -77,9 +80,10 @@ func TestAdminKubernetesObjectsGetAndDelete(t *testing.T) {
 			objNamespace: "openshift",
 			objName:      "config",
 			mocks: func(tt *test, k *mock_adminactions.MockKubeActions) {
+				k.EXPECT().ResolveGVR(tt.objKind)
 			},
 			wantStatusCode: http.StatusBadRequest,
-			wantError:      "400: InvalidParameter: : The provided groupKind '' is invalid.",
+			wantError:      "400: InvalidParameter: : The provided resource is invalid.",
 		},
 		{
 			method:       http.MethodGet,
@@ -89,6 +93,7 @@ func TestAdminKubernetesObjectsGetAndDelete(t *testing.T) {
 			objNamespace: "openshift",
 			objName:      "config",
 			mocks: func(tt *test, k *mock_adminactions.MockKubeActions) {
+				k.EXPECT().ResolveGVR(tt.objKind).Return(&schema.GroupVersionResource{Resource: "secrets"}, nil)
 			},
 			wantStatusCode: http.StatusForbidden,
 			wantError:      "403: Forbidden: : Access to secrets is forbidden.",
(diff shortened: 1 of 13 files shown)


pkg/frontend/validate.go
```diff
@@ -78,7 +78,33 @@ func (f *frontend) validateOpenShiftUniqueKey(ctx context.Context, doc *api.Open
 // prevent mischief
 var rxKubernetesString = regexp.MustCompile(`(?i)^[-a-z0-9.]{0,255}$`)
 
+func validatePermittedClusterwideObjects(groupKind string) bool {
```
bennerv, line 81: This now requires SREs to suffix the resource/kind with the corresponding API group correct?  

This will make things a bit harder for us as it'll require some remembering / lookups.

pkg/frontend/admin_openshiftcluster_kubernetesobjects_test.go
```diff
@@ -229,124 +229,6 @@ func TestAdminKubernetesObjectsGetAndDelete(t *testing.T) {
 	}
 }
 
-func TestValidateAdminKubernetesObjectsNonCustomer(t *testing.T) {
```
cblecker, line 232: I moved this test to validate_test.go, in alignment with the function it's testing

pkg/frontend/frontend.go
```diff
@@ -289,6 +289,12 @@ func (f *frontend) authenticatedRoutes(r *mux.Router) {
 	s.Methods(http.MethodGet).HandlerFunc(f.getAppLensDetector).Name("getAppLensDetector")
 
 	// Admin actions
+	s = r.
+		Path("/admin/subscriptions/{subscriptionId}/resourcegroups/{resourceGroupName}/providers/{resourceProviderNamespace}/{resourceType}/{resourceName}/kubernetesobjects/force").
```
cblecker, line 293: open to input on naming/differentiation on the path

pkg/frontend/validate.go
```diff
@@ -78,7 +78,33 @@ func (f *frontend) validateOpenShiftUniqueKey(ctx context.Context, doc *api.Open
...
+		"aro.openshift.io",
+		"config.openshift.io",
+		"console.openshift.io",
+		"machine.openshift.io",
+		"machineconfiguration.openshift.io",
+		"imageregistry.operator.openshift.io",
+		"certificatesigningrequest.certificates.k8s.io",
+		"apirequestcounts.apirequestcount.openshift.io",
+		"rbac.authorization.k8s.io",
+		"authorization.openshift.io",
```
bennerv, line 93: Any chance we could put `operator.openshift.io` in here?  

`dnses, imagecontentsourcepolicies, kubeapiserver, schedulers, openshiftapiservers` are all content we've used in the past. 

`security.openshift.io` -> securitycontextconstraints are also useful, but might expose customer SCCs.  But they are just names so they might be fine.

pkg/frontend/frontend.go
bennerv: /force implies something else IMO.  I would rather have a /restricted endpoint for the restricted one.

cblecker: I wanted to leave the current endpoint to ensure that everything that goes through it gets the restriction. Open to better names for the unrestricted endpoint.

cblecker:
This isn't fully working yet, but I think this is the right direction.

pkg/frontend/validate.go
facchettos: if this was a `map[string]bool` that'd reduce the complexity a little bit

pkg/frontend/validate.go
```diff
@@ -78,26 +79,48 @@ func (f *frontend) validateOpenShiftUniqueKey(ctx context.Context, doc *api.Open
...
+	}
+	for _, apiGroup := range validApiGroups {
+		if gvr.Group == apiGroup {
+			return true
+		}
+	}
+	return false
```
facchettos, line 101: if this was a `map[string]bool` that'd reduce the complexity a little bit because we'd be able to do `return isValid[gvr.Group]`

pkg/frontend/validate.go
hawkowl: +1 on `operator.openshift.io`, since that includes things such as network/DNS/ingress which are essential for us for determining a cluster state.

pkg/frontend/validate.go
```diff
@@ -78,26 +79,69 @@ func (f *frontend) validateOpenShiftUniqueKey(ctx context.Context, doc *api.Open
 // prevent mischief
 var rxKubernetesString = regexp.MustCompile(`(?i)^[-a-z0-9.]{0,255}$`)
 
-func validateAdminKubernetesObjectsNonCustomer(method, groupKind, namespace, name string) error {
+func validatePermittedClusterwideObjects(gvr *schema.GroupVersionResource) bool {
```
hawkowl, line 82: Can we put this somewhere common so that it can generate the counterpart in https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/rbac/staticresources/clusterrole.yaml ?

pkg/frontend/validate.go
cblecker: the contents of an SCC can contain usernames (unfortunately)

but I'm cool with operator.openshift.io

pkg/frontend/validate.go
cblecker: I have no objections to it being moved into a common/generic location, but I would rather not blow the scope of this PR any further than it already is. If you'd like to follow up with that, I'd be happy to review.

pkg/frontend/validate.go
cblecker: I had to update this function to check for two different conditions.

rolandmkunkel:
Is there a ticket that tracks all the work we need for this?

bennerv (commented):
It makes sense to me.  I think we hit the first iteration of apigroups.  Hoping this is all we'll need from cluster-scoped level.  

I think [Jeremy's comments](https://github.com/Azure/ARO-RP/pull/2661#discussion_r1088941647) still stand but lgtm otherwise

pkg/frontend/validate.go
cblecker: @bennerv @facchettos Is this change something that is blocking, or can it be done in a follow up?

facchettos: I'd rather have it directly in it tbh, but I'm not the boss of aro so :D

cblecker: @facchettos Why? Is there a functional difference?

petrkotas (approved):
Thank you @cblecker for the changes to enable us in MAG. And most importantly thank you for reducing the complixity of a dynamicHelper. 
This PR is good and we can get it in. 

I do agree with @facchettos on the change, however we can do it as a followup work, the same as @hawkowl proposed change.

rogbas:
This PR LGTM, I suggest we address @hawkowl and @facchettos suggestions separately.

cblecker:
Thank you all for the reviews. Merging..

PR #2662, opened 2023-01-26
Update namespace list

cblecker:

Fixes

Updates the namespace list with two additional openshift system namespaces


bennerv (approved):
lgtm

PR #2663, opened 2023-01-27
replaced deprecated gorilla with standard library

facchettos:
this is part of the effort to get rid of the deprecated gorilla library
https://issues.redhat.com/browse/ARO-1731

Fixes

### Test plan for issue:
It should pass existing tests (bar the one that I modified because there is no segregation on methods in servemux)

### DIFF
--- a/pkg/dbtoken/server.go
+++ b/pkg/dbtoken/server.go
@@ -16,7 +16,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/gorilla/mux"
 	"github.com/sirupsen/logrus"
 
 	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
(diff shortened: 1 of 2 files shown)


pkg/dbtoken/server_test.go
```diff
@@ -64,7 +64,7 @@ func TestServer(t *testing.T) {
 					Path:   "/token",
 				},
 			},
-			wantStatusCode: http.StatusMethodNotAllowed,
```
bennerv, line 67: This is changing because previously we set a route only for POST on the `/token` endpoint, and now using the stdlib mux it handles all HTTP method requests for the endpoint correct?  So it's hitting the auth endpoint which is why we're returning Forbidden now instead of MethodNotAllowed.  

If so, can we add a unit test case for an authenticated non-GET request to cover the new introduction?
```go
	if r.Method != http.MethodPost {
		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
	}
```

pkg/dbtoken/server.go
```diff
@@ -122,14 +121,6 @@ func (s *server) Run(ctx context.Context) error {
 	return srv.Serve(s.l)
 }
 
-func (s *server) unauthenticatedRoutes(r *mux.Router) {
-	r.NewRoute().Methods(http.MethodGet).Path("/healthz/ready").HandlerFunc(func(http.ResponseWriter, *http.Request) {})
```
bennerv, line 126: Previously we only allowed a `GET` request on this path, and no we accept anything.  I know it technically doesn't matter since the endpoint is just returning a 200 status code, but to maintain consistency across the removal should we have a filter on the health probe endpoint to only accept HTTP GET requests in case we make future changes to the health endpoint?

bennerv (commented):
Couple questions, otherwise makes sense.

pkg/dbtoken/server_test.go
facchettos: done

pkg/dbtoken/server.go
facchettos: I don't see any problems with it, although it is unlikely imho. I implemented it

bennerv (approved):
lgtm

PR #2664, opened 2023-01-27
update docs: configure local ARO-RP to create clusters using a Hive d…

AldoFusterTurpin:

Simple PR to update the docs explaining how to configure the ARO-RP running locally to create clusters using an AKS Hive dev cluster (instead of the standard way of creating clusters that does not use Hive).

This process is not documented and it will be nice having a simple explanation in the docs.

### Test plan for issue:

N/A

Yes, updated the *docs/deploy-development-rp.md* file.

### DIFF
--- a/docs/deploy-development-rp.md
+++ b/docs/deploy-development-rp.md
@@ -137,6 +137,18 @@
      journalctl _COMM=aro -o json --since "15 min ago" -f | jq -r 'select (.COMPONENT != null and (.COMPONENT | contains("access"))|not) | .MESSAGE'
      ```
 
+## Connect ARO-RP with a Hive development cluster
+The env variables names defined in pkg/util/liveconfig/manager.go control the communication of the ARO-RP with Hive.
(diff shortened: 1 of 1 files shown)


petrkotas (dismissed):
LGTM thank you @AldoFusterTurpin

docs/deploy-development-rp.md
```diff
@@ -137,6 +137,17 @@
      journalctl _COMM=aro -o json --since "15 min ago" -f | jq -r 'select (.COMPONENT != null and (.COMPONENT | contains("access"))|not) | .MESSAGE'
      ```
 
+## Create clusters using local ARO-RP + Hive
+If you want to create clusters using the local ARO-RP + Hive instead of using the standard cluster creation process, you can do so by setting the environment variables defined in pkg/util/liveconfig/manager.go:
+- Set *ARO_ADOPT_BY_HIVE* and *ARO_INSTALL_VIA_HIVE* to *true*.
```
bizz001, line 142: Why do we need to set ADOPT variable to true when we are creating cluster via hive?

AldoFusterTurpin: I have updated the PR to better reflect in the docs that *ARO_ADOPT_BY_HIVE* is not needed for the cluster creation (with a more curated explanation). Tks!

petrkotas (approved):

bennerv (approved):

PR #2621, opened 2023-01-06
Update dedicated RP deployment docs to include AKS, VPN, and Hive

darthhexx:
### Which issue this PR addresses:

Updates dedicated RP deployment docs to include the AKS and Hive installation steps.

### What this PR does / why we need it:

In order for engineers to be able to create dedicated RP infrastructure for Installer and infrastructure development.

### Test plan for issue:

Manually tested the process.

### Is there any documentation that needs to be updated for this PR?

Nope.

PR #2667, opened 2023-01-27
Migrate RP from adal to azidentity package

mbarnes:

[[ARO-1917] AAD: Migrate ADAL -> MSAL](https://issues.redhat.com/browse/ARO-1917)

Microsoft has [announced](https://techcommunity.microsoft.com/t5/microsoft-entra-azure-ad-blog/azure-ad-change-management-simplified/ba-p/2967456) the impending end-of-support for Azure Active Directory Authentication Library (ADAL):

> **Update December 15th, 2022**: ADAL end of support is now extended to June 30th, 2023. We will retire AAD Graph API any time after June 30th, 2023. Through the next six months (January 2023 – June 2023) we will continue informing customers about the upcoming end of support along with providing guidance on migration.

We need to move the entire RP code to Microsoft Authentication Library (MSAL) which for us means using the [Azure Identity Client Module for Go or "azidentity"](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity) in place of [Azure Active Directory authentication for Go or "adal"](https://pkg.go.dev/github.com/Azure/go-autorest/autorest/adal).

(Note: "azidentity", despite its slightly outdated [README](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/README.md), [does now use MSAL](https://github.com/Azure/azure-sdk-for-go/issues/14920).)

### Test plan for issue:

Primary method of testing was [deploying a full RP development service](https://github.com/Azure/ARO-RP/blob/master/docs/deploy-full-rp-service-in-dev.md) (with help from https://github.com/Azure/ARO-RP/pull/2621), and verifying all ARO services on the RP and Gateway VMs start up and function correctly.

Also, unit tests were updated where needed and verified to pass, along with running end-to-end tests.

None that I'm aware of.  This is an internal RP change and should not alter its behavior.

### Additional notes:

- This does bump the Go requirement to 1.18 since [azidentity requires Go 1.18 as of v1.0](https://github.com/Azure/azure-sdk-for-go/commit/9514e5f7779d3406fa8320777da06867b79dfcc1#diff-5f1a0df6b303a95cf03dc9fb99c9bef2ddaddbcb963ffe6a95900ae263169791).

- This is a large change, so I've tried to break most of the commits up by source directory for less overwhelming review.

- azidentity [handles token refreshing automatically](https://github.com/Azure/azure-sdk-for-go/issues/10567#issuecomment-921376485), so we should no longer have to worry about this in the RP.  Therefore, this Pull Request eliminates [AuthorizationRefreshingAction](https://github.com/Azure/ARO-RP/blob/master/pkg/util/steps/refreshing.go) and [refreshable.Authorizer](https://github.com/Azure/ARO-RP/blob/master/pkg/util/refreshable/refreshable.go).

- There is one remaining import of [adal](https://pkg.go.dev/github.com/Azure/go-autorest/autorest/adal) in [pkg/portal/middleware/aad.go](https://github.com/Azure/ARO-RP/blob/master/pkg/portal/middleware/aad.go), however the RP code is just using offline utility functions from [adal](https://pkg.go.dev/github.com/Azure/go-autorest/autorest/adal) that currently have no public equivalent in [Microsoft Authentication Library (MSAL) for Go](https://github.com/AzureAD/microsoft-authentication-library-for-go) (as far as I can see).  So I left that code alone for now.

- **Followup work:** Per the [azidentity Migration Guide](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/MIGRATION.md#use-azidentity-credentials-with-older-packages), this PR vendors in an unofficial module [github.com/jongio/azidext](https://github.com/jongio/azidext) to serve as a temporary shim between azidentity's TokenCredential APIs and the older Azure SDK client packages under [github.com/Azure/go-autorest](https://github.com/Azure/go-autorest).<BR>
 We should endeavor to further migrate the RP code to use the newer Azure SDK client packages under [github.com/Azure/azure-sdk-for-go/sdk/resourcemanager](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/resourcemanager), and remove the need for [github.com/jongio/azidext](https://github.com/jongio/azidext).


mbarnes:
CI checks will be failing until we have Go 1.18 support.

pkg/util/clusterauthorizer/authorizer.go
```diff
@@ -44,36 +48,46 @@ func NewAzRefreshableAuthorizer(log *logrus.Entry, azEnv *azureclient.AROEnviron
...
-	// create service principal token from azure-credentials
-	token, err := a.tokenClient.GetToken(ctx,
-		a.log,
-		string(credentials.ClientID),
-		string(credentials.ClientSecret),
-		string(credentials.TenantID),
-		a.azureEnvironment.ActiveDirectoryEndpoint,
-		a.azureEnvironment.ResourceManagerEndpoint)
+
+	tokenCredential := a.tokenCredential // for unit testing only
```
SudoBrendan, line 61: nit: what if, instead of having a field that was only useful during unit tests, we used the field for dependency injection? We'd define an interface, say `TokencredentialGetter` with a `GetTokenCredential` function (returns tokenCredential and error), which in prod used the logic below, but in our tests, was some other implementation? This feels a little cleaner, and doesn't require all the "only useful for tests" stuff, I'm not sure, just a thought.

pkg/util/clusterauthorizer/authorizer.go
```diff
@@ -44,36 +48,46 @@ func NewAzRefreshableAuthorizer(log *logrus.Entry, azEnv *azureclient.AROEnviron
...
+
+	scopes := []string{a.azureEnvironment.ResourceManagerEndpoint + "/.default"}
+
+	tokenRequestOptions := policy.TokenRequestOptions{Scopes: scopes}
+	token, err := tokenCredential.GetToken(ctx, tokenRequestOptions)
 	if err != nil {
 		return nil, err
 	}
 
+	// XXX Is this validation still necessary?
```
SudoBrendan, line 82: todo?

SudoBrendan (commented):
Taking this in pieces, it's quite large, but the commit structure is EXCELLENT and makes this much easier to review! Thanks! I've made it past the first 6 commits... Looking good so far, a few questions/nits.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -301,7 +296,7 @@ func (dv *dynamic) validateActions(ctx context.Context, r *azure.Resource, actio
 
 		if detailedErr, ok := err.(autorest.DetailedError); ok &&
 			detailedErr.StatusCode == http.StatusForbidden {
-			return false, steps.ErrWantRefresh
```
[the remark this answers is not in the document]
SudoBrendan: Going further, can we just eliminate the `ErrWantRefresh` object all together? I haven't seen that in the commits yet, maybe it's coming :p

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -108,7 +108,7 @@ func (c *Client) createRequest(
 	requestEnricher func(*policy.Request)) (*policy.Request, error) {
 	if requestOptions != nil {
 		header := requestOptions.toHeader()
-		ctx = policy.WithHTTPHeader(ctx, header)
```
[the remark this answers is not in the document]
mbarnes: Breaking change in [azcore v0.22.0](https://github.com/Azure/azure-sdk-for-go/blob/71b4e49ab14f7dee8fa19db3f0dec6edab44e430/sdk/azcore/CHANGELOG.md#0220-2022-03-03): function moved from `policy` to `runtime`.

pkg/util/refreshable/refreshable.go
```diff
@@ -1,82 +0,0 @@
-package refreshable
```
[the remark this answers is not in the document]
mbarnes: I verified it only insofar as finding where it's implemented (where you linked) but haven't tested it specifically in an ARO RP context.  I imagine the maintainers have a test cases for it, which I can try to hunt down.

pkg/util/clusterauthorizer/authorizer.go
mbarnes: Yeah, this is pretty fugly.  I was trying to quickly fix the conflicts introduced when these unit tests were merged to `master`, and azidentity works significantly differently.  Your interface idea sounds good, I'll see if I can make this cleaner.

pkg/api/validate/dynamic/dynamic.go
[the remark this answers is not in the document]
mbarnes: Yeah, code archaeology on ARO is hard.  Git blame on that line traces it to one of Jim Minter's commits whose [commit message is helpfully titled "WIP"](https://github.com/Azure/ARO-RP/commit/690c767cd62ed485f8227e0e8c4cb01763f96751) with no further explanation.  An explanation may be buried in some old team demo video, but best as I can tell from the code alone it's a special case for needing manual token refreshing, which azidentity should handle now.

The commit that deletes `pkg/util/steps/refreshing.go` takes `ErrWantRefresh` with it, but I'll take another look at the error handling code here.

pkg/util/clusterauthorizer/authorizer.go
mbarnes: Was originally an on-the-fly test to make sure the provided ADAL token was valid, or at least parsable.  Unclear why it was needed, [commit that introduced](https://github.com/Azure/ARO-RP/commit/144606bb004b64623d4eec969c7098253f7ba56a#diff-9b05764fc49f59be0df3e53cc7ff53930ad08391ac5a4e3740f70960db70a73d) it doesn't explain.  Hence my comment.

Wrote that early on when I was still figuring things out.  I'd lean now toward deleting this part.  I trust no `GetToken` implementation in azidentity is going to return an invalid token.

pkg/util/clusterauthorizer/authorizer.go
mbarnes: Implemented something close to your suggestion in a "fixup" commit below.  You're right, it feels cleaner.
There's another instance of this pattern in a subsequent commit, which I'll also rewrite.

pkg/api/validate/dynamic/dynamic.go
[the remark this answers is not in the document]
mbarnes: Addressed this in a "fixup" commit below.

pkg/util/refreshable/refreshable.go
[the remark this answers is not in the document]
mbarnes: Located unit tests: [resource_test.go](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/internal/temporal/resource_test.go)

The call chain between [azidext.NewTokenCredentialAdapter](https://github.com/jongio/azidext/blob/ecb3be2645af39c19b2fcc8409732170b6fba5f2/go/azidext/azure_identity_credential_adapter.go#L17-L25) and the [refresh logic](https://github.com/Azure/azure-sdk-for-go/blob/5a72021a515303a89f838f231ec6b64f84f25ef2/sdk/internal/temporal/resource.go#L45-L114) is pretty straight-forward.  I can elaborate if requested.

Also, FWIW, while I did say I haven't specifically verified the refresh functionality in an ARO RP context, I can say in all my cluster creation/deletion testing with azidentity (before multi version) I never observed any failure due to expired tokens.

mbarnes:
https://github.com/Azure/ARO-RP/pull/2691 (Go 1.18) merged so I rebased this branch and squashed the "fixup" commits.

mbarnes:
Tweaked some commits and force-pushed again to get more checks passing.

pkg/util/refreshable/refreshable.go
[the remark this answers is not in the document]
SudoBrendan: Do we know what the default token TTL is?

pkg/util/clusterauthorizer/authorizer.go
SudoBrendan: lgtm!

pkg/api/validate/dynamic/serviceprincipal_test.go
```diff
@@ -5,101 +5,79 @@ package dynamic
```
SudoBrendan, line 5: I just have to comment that this file's changes is one of the most satisfying things I've seen in a while, it reads like a sigh of relief. Excellent.

pkg/api/validate/openshiftcluster_validatedynamic.go
```diff
@@ -39,7 +45,49 @@ type openShiftClusterDynamicValidator struct {
 
 	oc              *api.OpenShiftCluster
 	subscriptionDoc *api.SubscriptionDocument
-	fpAuthorizer    refreshable.Authorizer
+	fpAuthorizer    autorest.Authorizer
+}
+
+func ensureAccessTokenClaims(ctx context.Context, tokenCredential *azidentity.ClientSecretCredential, scopes []string) error {
```
SudoBrendan, line 51: Q: Why is this necessary now when it wasn't before? It makes sense to validate the FPAuthorizer generates tokens with proper scope, but curious how this impacts downstream code.

pkg/api/validate/openshiftcluster_validatedynamic.go
```diff
@@ -39,7 +45,49 @@ type openShiftClusterDynamicValidator struct {
...
+	var err error
+
+	timeoutCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
SudoBrendan, line 54: Q: 5 min on this seems a bit excessive for some reason, but I'm most certainly wrong. What are we expecting to take 4+ minutes that would still result in success? GetToken?

pkg/env/dev.go
```diff
@@ -84,30 +84,26 @@ func (d *dev) Listen() (net.Listener, error) {
...
-		return nil, err
-	}
-
+func (d *dev) FPAuthorizer(tenantID string, scopes ...string) (autorest.Authorizer, error) {
 	fpPrivateKey, fpCertificates := d.fpCertificateRefresher.GetCertificates()
 
-	sp, err := adal.NewServicePrincipalTokenFromCertificate(*oauthConfig, d.fpClientID, fpCertificates[0], fpPrivateKey, resource)
+	options := d.Environment().ClientCertificateCredentialOptions()
+	tokenCredential, err := azidentity.NewClientCertificateCredential(
+		tenantID, d.fpClientID, fpCertificates, fpPrivateKey, options)
```
SudoBrendan, line 92: It looks like old functionality only passed the first certificate instead of the whole array. Do we know this only contains the 1 cert we want on this, or are we potentially expanding the privilege here?

pkg/env/dev.go
```diff
@@ -84,30 +84,26 @@ func (d *dev) Listen() (net.Listener, error) {
...
 
 func (d *dev) FPNewClientCertificateCredential(tenantID string) (*azidentity.ClientCertificateCredential, error) {
 	fpPrivateKey, fpCertificates := d.fpCertificateRefresher.GetCertificates()
 
-	credential, err := azidentity.NewClientCertificateCredential(tenantID, d.fpClientID, fpCertificates, fpPrivateKey, &azidentity.ClientCertificateCredentialOptions{
-		AuthorityHost:        d.Environment().AuthorityHost,
-		SendCertificateChain: true,
-	})
+	options := d.Environment().ClientCertificateCredentialOptions()
+	options.SendCertificateChain = true
```
SudoBrendan, line 104: Do we also need AuthorityHost as an option here, or was that removed?

pkg/env/msiauthorizer.go
```diff
@@ -18,28 +20,34 @@ const (
...
+			os.Getenv("AZURE_"+string(msiContext)+"_CLIENT_ID"),
+			os.Getenv("AZURE_"+string(msiContext)+"_CLIENT_SECRET"),
+			options)
+	}
+	if err != nil {
+		return nil, err
 	}
 
-	return config.Authorizer()
+	return azidext.NewTokenCredentialAdapter(tokenCredential, scopes), nil
```
SudoBrendan, line 52: good catch returning `nil`, I'm honestly surprised this even compiled.

pkg/env/prod.go
```diff
@@ -320,20 +320,17 @@ func (p *prod) FeatureIsSet(f Feature) bool {
...
-		return nil, err
-	}
-
+func (p *prod) FPAuthorizer(tenantID string, scopes ...string) (autorest.Authorizer, error) {
 	fpPrivateKey, fpCertificates := p.fpCertificateRefresher.GetCertificates()
 
-	sp, err := adal.NewServicePrincipalTokenFromCertificate(*oauthConfig, p.fpClientID, fpCertificates[0], fpPrivateKey, resource)
+	options := p.Environment().ClientCertificateCredentialOptions()
+	tokenCredential, err := azidentity.NewClientCertificateCredential(
+		tenantID, p.fpClientID, fpCertificates, fpPrivateKey, options)
```
SudoBrendan, line 328: Same as above, we are shipping all certs instead of only the first.

pkg/env/prod.go
```diff
@@ -375,11 +372,10 @@ func (p *prod) LiveConfig() liveconfig.Manager {
 func (p *prod) FPNewClientCertificateCredential(tenantID string) (*azidentity.ClientCertificateCredential, error) {
 	fpPrivateKey, fpCertificates := p.fpCertificateRefresher.GetCertificates()
 
-	credential, err := azidentity.NewClientCertificateCredential(tenantID, p.fpClientID, fpCertificates, fpPrivateKey, &azidentity.ClientCertificateCredentialOptions{
-		AuthorityHost:        p.Environment().AuthorityHost,
-		SendCertificateChain: true,
-	})
+	options := p.Environment().ClientCertificateCredentialOptions()
+	options.SendCertificateChain = true
```
SudoBrendan, line 376: Same as above - no AuthorityHost?

SudoBrendan (commented):
OK, round 2, but my mind is melting, so, for my future self, I'm at commit `Migrate RP from adal to azidentity package`, and made it through /pkg/env...

Still looking great so far, a few more Q's. Thanks!

pkg/util/refreshable/refreshable.go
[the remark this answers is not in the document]
mbarnes: > The default lifetime of an access token is variable. When issued, the default lifetime of an access token is assigned a random value ranging between 60-90 minutes (75 minutes on average). The variation improves service resilience by spreading access token demand over a time, which prevents hourly spikes in traffic to Azure AD.

([source](https://learn.microsoft.com/en-us/azure/active-directory/develop/access-tokens#access-token-lifetime))

pkg/env/dev.go
mbarnes: **AuthorityHost** was removed in favor of a **Cloud** option in the embedded [ClientOptions](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.3.1/policy#ClientOptions) struct.

Essentially does the same thing: specifies which Azure cloud to use.  `d.Environment().ClientCertificateCredentialOptions()` sets it appropriately.

pkg/api/validate/openshiftcluster_validatedynamic.go
mbarnes: I probably should have added a comment there.

I copied that logic from the old [GetToken function in pkg/util/aad/aad.go](https://github.com/Azure/ARO-RP/blob/f6e624fcd928468b8208e825bde841a89bcb8303/pkg/util/aad/aad.go#L49-L86) before I deleted the file.  This was the one place that seemed like the token polling logic might still be needed, based on the comment
```
// during credentials rotation this can take time to propagate
```

pkg/api/validate/openshiftcluster_validatedynamic.go
mbarnes: See above answer.  Old logic pasted into a new function.

pkg/env/dev.go
mbarnes: Don't think I considered that.  It was just "function wants an array, I have an array, here ya go!"

I suspect there's only one certificate present but I'll try to track this down to make sure.

mbarnes:
E2E tests pass for me when run locally but are consistently failing here on Azure Pipelines.

Going to add some temporary debug commits to help figure out where the failure's coming from.  If you're reviewing code, disregard these.

pkg/env/dev.go
mbarnes: So I kinda side-stepped your question in a "fixup" commit below by rewriting `FPAuthorizer` to get its token credential from the recently-added `FPNewClientCertificateCredential` method (the two methods were _almost_ identical).

But `FPNewClientCertificateCredential` is also passing the full certificate array returned by the certificate refresher.

Did some debugging of E2E failures and in the course of that discovered in the `e2e-eastus-svc` keyvault, the `rp-firstparty` certificate bundle that we're fetching here consists of 3 certificates.  I don't have permission to examine them in the Azure portal, but I assume the other two are just certificate authorities.

I can look into that further if you want.  I guess I could just as easily change it to `fpCertificates[:1]` but I'm not yet convinced there's a risk.

pkg/env/prod.go
mbarnes: See above reply.

pkg/env/prod.go
mbarnes: See above reply.

SudoBrendan (commented):
OK, I've finally made it past the last commit, again stuff is generally looking good and E2E is passing! :tada:

The last question I'll post is: this PR introduces a lot of `+ "/.default"` to endpoints, is there ever a case in the code now where we _don't_ use that `/.default` uri? Can we safely migrate that into the Environment instead, in one place?

mbarnes:
> The last question I'll post is: this PR introduces a lot of + "/.default" to endpoints, is there ever a case in the code now where we don't use that /.default uri? Can we safely migrate that into the Environment instead, in one place?

The whole "/.default" scope thing is ugly and I wish azidentity just handled it transparently.

There is only one place where we don't use "/.default", which is the MSI authorizer in `cmd/aro/gateway.go` for DB token.

I considered adding getter methods in `Environment` for the few scopes we use, but decided against it since this is temporary ugliness.  I believe the newer Azure clients I suggested migrating to as followup work hide all this scope stuff.  I've only looked briefly at the APIs but I think it sets the appropriate scope(s) automatically for the type of Azure resource.

cadenmarchese (dismissed):
This looks good to me without any testing done on my end. I've left a few questions that may help me with https://issues.redhat.com/browse/ARO-2383. Thanks @mbarnes !

cmd/aro/dbtoken.go
```diff
@@ -47,12 +47,12 @@ func dbtoken(ctx context.Context, log *logrus.Entry) error {
 		}
 	}
 
-	msiAuthorizer, err := _env.NewMSIAuthorizer(env.MSIContextRP, _env.Environment().ResourceManagerEndpoint)
+	msiAuthorizer, err := _env.NewMSIAuthorizer(env.MSIContextRP, _env.Environment().ResourceManagerEndpoint+"/.default")
```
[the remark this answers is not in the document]
mbarnes: Yeah, these are "scopes" in the lingo of Microsoft's Identity Platform, which this change moves us to.

The [".default" scope](https://learn.microsoft.com/EN-US/azure/active-directory/develop/scopes-oidc#the-default-scope) is a special case that daemon applications like ours would typically use.

From the docs:
> Another use of .default is to request app roles (also known as application permissions) in a non-interactive application like a daemon app that uses the [client credentials](https://learn.microsoft.com/EN-US/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow) grant flow to call a web API.

I commented earlier in the PR that these scope strings will go away once we migrate to the [newer Azure SDK](https://issues.redhat.com/browse/ARO-2527) whose client packages already know the appropriate scope(s) to use.  So it's only temporary ugliness until we fully transition.

mbarnes:
Thanks for the review @cadenmarchese!

mbarnes:
Looks like I lost @cadenmarchese's approval after I force-pushed today to resolve merge conflicts.

cadenmarchese (approved):
From my perspective, this PR looks good. I might reach out with some more questions in the coming weeks. Thanks!

mbarnes:
In order to not lose approvals, I'll resolve any further merge conflicts once I'm told we're ready to merge this.

pkg/api/validate/openshiftcluster_validatedynamic.go
```diff
@@ -39,7 +45,49 @@ type openShiftClusterDynamicValidator struct {
...
+	defer cancel()
+
+	_ = wait.PollImmediateUntil(10*time.Second, func() (bool, error) {
```
jaitaiwan, line 57: I don't think err needs to be set here, the condition can return errors directly inside the function (like it's already doing) that error should be the error result from wait.PollImmediateUtil instead.

e.g.
```
- var err error
+ err := wait.PollImmediateUntil(...
return err
```

pkg/api/validate/openshiftcluster_validatedynamic.go
```diff
@@ -39,7 +45,49 @@ type openShiftClusterDynamicValidator struct {
...
+		if err != nil {
+			err = api.NewCloudError(
+				http.StatusBadRequest,
+				api.CloudErrorCodeInvalidServicePrincipalToken,
+				"properties.servicePrincipalProfile",
+				"The provided service principal generated an invalid token.")
+			return false, err
+		}
+
+		for _, claim := range []string{"altsecid", "oid", "puid"} {
```
jaitaiwan, line 76: Is there any documentation that can be linked to here about the significance of these 3 strings?

pkg/cluster/cluster.go
```diff
@@ -117,17 +117,17 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
...
 	}
 
-	localFPAuthorizer, err := _env.FPAuthorizer(_env.TenantID(), _env.Environment().ResourceManagerEndpoint)
+	localFPAuthorizer, err := _env.FPAuthorizer(_env.TenantID(), _env.Environment().ResourceManagerEndpoint+"/.default")
 	if err != nil {
 		return nil, err
 	}
 
-	fpAuthorizer, err := _env.FPAuthorizer(subscriptionDoc.Subscription.Properties.TenantID, _env.Environment().ResourceManagerEndpoint)
+	fpAuthorizer, err := _env.FPAuthorizer(subscriptionDoc.Subscription.Properties.TenantID, _env.Environment().ResourceManagerEndpoint+"/.default")
```
jaitaiwan, line 125: I personally would like to see this string put into a const somewhere and used everywhere so that the significance of the string is documented.

jaitaiwan (changes requested):
LGTM just a couple nit picks

pkg/api/validate/openshiftcluster_validatedynamic.go
mbarnes: The original function I copied this from had a [note warning not to do that](https://github.com/Azure/ARO-RP/blob/f8cfc0338a481cc7ce6f007f18c4348d5217bea2/pkg/util/aad/aad.go#L83-L86).  I don't know who's right, but I'll reproduce that note here and leave the logic in its original form.  I don't want to introduce a potential regression.

pkg/cluster/cluster.go
mbarnes: I was originally pushing back on doing this since these scope strings are temporary and will go away once we fully migrate to the newer Azure SDK that uses the [TokenCredential interface](https://github.com/Azure/azure-sdk-for-go/blob/sdk/azcore/v1.4.0/sdk/azcore/internal/exported/exported.go#L80) instead of [ADAL-based Authorizers](https://pkg.go.dev/github.com/Azure/go-autorest/autorest#Authorizer).  But since you're the 3rd person to ask, I'll go ahead and add some scope fields to the `AROEnvironment` struct.

pkg/api/validate/openshiftcluster_validatedynamic.go
mbarnes: Added a new permanent commit for this.

s-amann (approved):
lgtm

mbarnes:
Rebased on https://github.com/Azure/ARO-RP/pull/2818, so the 1st commit for dependencies has changed substantially.
No other code changes.

s-amann (approved):
lgtm, and with e2e passing I think we are good to get this in!

bennerv (approved):

PR #2670, opened 2023-01-31
Hide hive config behind localdev flag

petrkotas:

Fixes

Hides the hive config behind localdev flag. The managed identity is not yet supported in the prod e2e, which renders these tests unusable. 


UlrichSchlueter (approved):
LGTM

darthhexx:
To be clear, these tests would not work because the E2E container does not have network access to the private AKS cluster. Apologies for not reviewing this test initially :(

PR #2673, opened 2023-01-31
Always use lowercase "true" for the Official Release flag

ventifus:
Fixes a case mismatch in PR #2574 that causes the tag checking to never happen.


petrkotas (approved):
LGTM

PR #2674, opened 2023-02-01
Created Cluster Health Dashboard in Admin Portal

anshulvermapatel:
https://issues.redhat.com/browse/ARO-1767

It has health dashboard for the cluster to get health related metrics which eventually reduce toil for the SRE engineers.


pkg/portal/cluster/statistics.go
```diff
@@ -0,0 +1,115 @@
...
+		Step:  time.Minute * 2,
+	})
+	if len(warning) > 0 {
+		f.log.Warn(warning)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	valueMatrix := value.(model.Matrix)
```
facchettos, line 60: are we sure this is always this type? Otherwise we might want to check the type casting result

pkg/portal/cluster/statistics.go
```diff
@@ -0,0 +1,115 @@
...
+		return endTime.Add(-7 * 24 * time.Hour), nil
+	case "2w":
+		return endTime.Add(-14 * 24 * time.Hour), nil
+	case "4w":
+		return endTime.Add(-28 * 24 * time.Hour), nil
+	case "8w":
+		return endTime.Add(-56 * 24 * time.Hour), nil
+	}
+	return time.Time{}, errors.New("invalid duration")
+}
```
facchettos, line 115: I think `time.ParseDuration` would help you quite a bit here

pkg/portal/portal.go
```diff
@@ -307,7 +307,7 @@ func (p *portal) aadAuthenticatedRoutes(r *mux.Router) {
 	r.NewRoute().PathPrefix("/api/{subscription}/{resourceGroup}/{clusterName}/nodes").HandlerFunc(p.nodes)
 	r.NewRoute().PathPrefix("/api/{subscription}/{resourceGroup}/{clusterName}/machines").HandlerFunc(p.machines)
 	r.NewRoute().PathPrefix("/api/{subscription}/{resourceGroup}/{clusterName}/machine-sets").HandlerFunc(p.machineSets)
-	r.NewRoute().PathPrefix("/api/{subscription}/{resourceGroup}/{clusterName}").HandlerFunc(p.clusterInfo)
+	r.NewRoute().PathPrefix("/api/{subscription}/{resourceGroup}/{clusterName}/statistics/{statisticsType}").HandlerFunc(p.statistics)
```
facchettos, line 310: I'm not sure pathPrefix is needed here

pkg/portal/cluster/statistics.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	valueMatrix := value.(model.Matrix)
+	return convertToStringMap(valueMatrix), nil
+}
+
+func convertToStringMap(v model.Matrix) []Metrics {
```
facchettos, line 64: I'm confused, this doesn't convert it to a stringmap does it?

pkg/portal/cluster.go
```diff
@@ -190,3 +195,88 @@ func (p *portal) machineSets(w http.ResponseWriter, r *http.Request) {
...
+	case "dnsalltraffic":
+		promQuery = "histogram_quantile(0.95, sum(rate(coredns_dns_request_duration_seconds_bucket[5m])) by (le))"
+	//Ingress
+	case "ingresscontrollercondition":
+		promQuery = "sum(ingress_controller_conditions) by (condition)"
+
+	default:
+		p.internalServerError(w, errors.New("invalid statistic type '"+statisticsType+"'"))
+		return
+	}
```
facchettos, line 249: maybe we could convert that into a map and have it in a different function

pkg/portal/cluster.go
```diff
@@ -190,3 +195,88 @@ func (p *portal) machineSets(w http.ResponseWriter, r *http.Request) {
...
+	//DNS
+	case "dnsresponsecodes":
+		promQuery = "sum(rate(coredns_dns_responses_total[5m])) by (rcode)"
+	case "dnserrorrate":
+		promQuery = "sum(rate(coredns_dns_responses_total{rcode=~\"SERVFAIL|NXDOMAIN\"}[5m])) by (pod) / sum(rate(coredns_dns_responses_total{rcode=~\"NOERROR\"}[5m])) by (pod)"
+
+	case "dnshealthcheck":
+		promQuery = "histogram_quantile(0.99, sum(rate(coredns_health_request_duration_seconds_bucket[5m])) by (le))"
+	case "dnsforwardedtraffic":
+		promQuery = ""
```
facchettos, line 239: is it empty on purpose?

pkg/portal/cluster/statistics.go
```diff
@@ -0,0 +1,115 @@
...
+	Value []MetricValue `json:"metricvalue"`
+}
+
+func (c *client) Statistics(ctx context.Context, httpClient *http.Client, promQuery string, duration string, endTime time.Time) ([]Metrics, error) {
+	return c.fetcher.Statistics(ctx, httpClient, promQuery, duration, endTime)
+}
+
+func (f *realFetcher) Statistics(ctx context.Context, httpClient *http.Client, promQuery string, duration string, endTime time.Time) ([]Metrics, error) {
+	promConfig := prometheusAPI.Config{
+		Address: "http://prometheus-k8s-0:9090",
```
facchettos, line 33: couldn't we reuse the domain name from the director function here?

pkg/portal/cluster/statistics.go
```diff
@@ -0,0 +1,115 @@
...
+type Metrics struct {
+	Name  string        `json:"metricname"`
+	Value []MetricValue `json:"metricvalue"`
+}
+
+func (c *client) Statistics(ctx context.Context, httpClient *http.Client, promQuery string, duration string, endTime time.Time) ([]Metrics, error) {
+	return c.fetcher.Statistics(ctx, httpClient, promQuery, duration, endTime)
+}
+
+func (f *realFetcher) Statistics(ctx context.Context, httpClient *http.Client, promQuery string, duration string, endTime time.Time) ([]Metrics, error) {
```
facchettos, line 31: why is this a method on the real fetcher? you don't use any of its fields

facchettos (changes requested):
haven't looked at the typescript, but there are things to change in the go part at least

pkg/portal/cluster/statistics.go
anshulvermapatel: `time.ParseDuration` does not take `1d`, `1w` into account .. I can make the frontend to send the string in hours like for 1d send request for 24h but any place either frontend or backend .. I will have to do this switch case thing with ParseDuration, a little less but  will have to do that.

pkg/portal/cluster/statistics.go
anshulvermapatel: hmmm, Actually I wanted to make this `client's` method and since the `client` has `realfetcher` in it so in order to call it from the portal package I did this, but yeah .. I learned from the azure-side-information PR and will try correcting this and let you know.

pkg/portal/cluster.go
anshulvermapatel: ohh no .. I need to put in a query there.

pkg/portal/cluster/statistics.go
anshulvermapatel: no .. yaa that name is consufing, I will change that name.

pkg/portal/cluster.go
anshulvermapatel: I think creating a map would still be the same thing since the number of elements are less, it could have been good at performance if the elements would have been more(if I am not wrong), but yeah, this can be in a different function.

pkg/portal/cluster/statistics.go
anshulvermapatel: This data is actually coming from prometheus only from the cluster, so it being in the realfetcher makes sense since that is also for fetching any details from the cluster.

pkg/portal/cluster/statistics.go
anshulvermapatel: where is this domain in the director function? :thinking:

pkg/portal/cluster/statistics.go
facchettos: I do not understand what you're trying to say

facchettos: we should probably use a standard duration format that is parsable by golang instead of having things like that

pkg/portal/cluster/statistics.go
facchettos:
```go
	r.RequestURI = ""
	r.URL.Scheme = "http"
	r.URL.Host = "prometheus-k8s-0:9090"
	r.URL.Path = "/" + strings.Join(strings.Split(r.URL.Path, "/")[10:], "/")
	r.Header.Del("Cookie")
	r.Header.Del("Referer")
	r.Host = r.URL.Host
```

facchettos: line 58-64

pkg/portal/cluster/statistics.go
anshulvermapatel: name changed resolving this conversation.

pkg/portal/cluster.go
anshulvermapatel: I did put that query in the latest push, resolving.

pkg/portal/cluster/statistics.go
anshulvermapatel: okay done. I am converting date into hours at the frontend itself now and then sending to the backend so that ParseDuration() fucntion can identify the time.

pkg/portal/cluster/statistics.go
anshulvermapatel: yes, this is how we use it.

pkg/portal/cluster/statistics.go
anshulvermapatel: Okay, I get it why you want me to have it from there, to keep this URL at the single place and we don't have to explicitly change it everywhere when the URL will get changed. 

However, I will have to change the director function itself, in order to not to panic as the resourceID grepping logic is making the program panic, and then will have to extract the URL from the request.
So, why can't we create a new Function which actually gives this domain(to the Director functions as well)??

facchettos: that's a good idea :+1:

anshulvermapatel: Done.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
```diff
@@ -0,0 +1,228 @@
...
+  export const convertTimeToHours = (duration: string): string => {
+    switch(duration) {
+      case "5m":
+      case "10m":
+      case "30m":
+      case "1h":
+      case "2h":
+      case "6h":
+      case "12h":
+        return duration
```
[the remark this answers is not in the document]
anshulvermapatel: Oh yes .. I will remove those.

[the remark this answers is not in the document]
anshulvermapatel: Done

ellis-johnson (commented):
This is so awesome, great work!

One thing, would you be able to add an e2e test for these changes?

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsComponent.tsx
```diff
@@ -0,0 +1,165 @@
...
+    DefaultPalette.themeTertiary
+  ]
+
+  const _onLegendClickHandler = (selectedLegend: string | null | string[]): void => {
+    if (selectedLegend !== null) {
+      console.log(`Selected legend - ${selectedLegend}`);
+    }
+  };
+
+  function _styledExample(): JSX.Element {
```
ellis-johnson, line 74: Should probably rename this

ellis-johnson (changes requested):
Small change requested. Also I think you might need to delete your package-lock.json and local npm_modules and run a npm install and portal build again and commit the changes

ellis-johnson (approved):
LGTM!

pkg/portal/cluster.go
```diff
@@ -190,3 +195,56 @@ func (p *portal) machineSets(w http.ResponseWriter, r *http.Request) {
...
+		fmt.Sprintf(
+			"/subscriptions/%s/resourceGroups/%s/providers/Microsoft.RedHatOpenShift/openShiftClusters/%s",
+			subscription, resourceGroup, clusterName))
+
+	promQuery, err := cluster.GetPromQuery(statisticsType)
+	if err != nil {
+		p.internalServerError(w, err)
+		return
+	}
+	prom := prometheus.New(p.log, p.dbOpenShiftClusters, p.dialer) //, p.authenticatedRouter)
```
facchettos, line 223: isn't this comment a leftover of some kind?

facchettos (changes requested):
minor changes and it should be good to go afterwards

pkg/portal/cluster.go
```diff
@@ -190,3 +195,56 @@ func (p *portal) machineSets(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	b, err := json.MarshalIndent(APIStatistics, "", "    ")
+	if err != nil {
+		p.internalServerError(w, err)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	_, _ = w.Write(b)
```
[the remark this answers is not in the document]
anshulvermapatel: okay, changed it.

pkg/portal/cluster.go
anshulvermapatel: yes, my bad, removed it and pushed the changes.

facchettos (dismissed):
LGTM on the go code, will let others review the TS. Please squash when merging

ellis-johnson (approved):

pkg/portal/cluster.go
```diff
@@ -190,3 +195,59 @@ func (p *portal) machineSets(w http.ResponseWriter, r *http.Request) {
...
+		p.internalServerError(w, err)
+		return
+	}
+	promHost, promScheme := prom.GetPrometheusHostAndScheme()
+	prometheusURL := promScheme + "://" + promHost
+	APIStatistics, err := fetcher.Statistics(ctx, httpClient, promQuery, duration, endTime, prometheusURL)
+	if err != nil {
+		p.internalServerError(w, err)
+		return
+	}
```
jaitaiwan, line 241: Query parameters aren't being validated until this function. This means that if a client provided invalid input data it would receive a `500 Internal Server Error` which would really explain what has actually happened. I'd expect to see input validation inside this handler so that you can return to the client a `400 Bad Request` in the case the client is sending something we don't expect.

pkg/portal/cluster/statistics.go
```diff
@@ -0,0 +1,114 @@
...
+	Timestamp time.Time `json:"timestamp"`
+	Value     float64   `json:"value"`
+}
+
+type Metrics struct {
+	Name  string        `json:"metricname"`
+	Value []MetricValue `json:"metricvalue"`
+}
+
+func (c *client) Statistics(ctx context.Context, httpClient *http.Client, promQuery string, duration string, endTime time.Time, prometheusURL string) ([]Metrics, error) {
```
jaitaiwan, line 27: It would be good to ensure that all exported types and functions are documented.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
```diff
@@ -0,0 +1,220 @@
+import { TimePicker, IComboBox, DatePicker, IDatePickerStyles, Stack, Text, TooltipHost, IStackStyles, IconButton, mergeStyleSets } from "@fluentui/react"
+import { iconButtonStyles } from "./Statistics"
+
+export function GraphOptionsComponent(props: {duration: string, setDuration: React.Dispatch<React.SetStateAction<string>> , endDate: Date, setEndDate: React.Dispatch<React.SetStateAction<Date>>}): any {    
+    const calculatorAddition =  { iconName: "CalculatorAddition" }
+    const calculatorSubtract =  { iconName: "CalculatorSubtract" }
+    const _increaseDuration = () => {
+      switch(props.duration) {
```
jaitaiwan, line 8: Change to a JS object

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsComponent.tsx
```diff
@@ -0,0 +1,165 @@
...
+    DefaultPalette.themeLighter,
+    DefaultPalette.themeLighterAlt,
+    DefaultPalette.themePrimary,
+    DefaultPalette.themeSecondary,
+    DefaultPalette.themeTertiary
+  ]
+
+  const _onLegendClickHandler = (selectedLegend: string | null | string[]): void => {
+    if (selectedLegend !== null) {
+      console.log(`Selected legend - ${selectedLegend}`);
```
jaitaiwan, line 70: We shouldn't be committing console.logs

portal/v2/src/Request.tsx
```diff
@@ -114,3 +115,15 @@ export const RequestKubeconfig = async (
     return OnError(err)
   }
 }
+
+export const FetchStatistics = async (cluster: ICluster, statisticsName: string, duration: string, endDate: Date): Promise<AxiosResponse | null> => {
+  duration = convertTimeToHours(duration)
+  try {
+    const result = await axios(
+      "/api/" + cluster.subscription + "/" + cluster.resourceGroup + "/" + cluster.name + "/statistics/" + statisticsName + "?duration=" + duration + "&endtime=" + endDate.toJSON())
```
jaitaiwan, line 123: This is a great place for using a javascript template literal to reduce the line size.

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsWrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+      }
+      metrics.push(metric)
+    });
+    setMetrics(metrics)   
+  }
+
+  useEffect(() => {    
+    const onData = (result: AxiosResponse | null) => {
+      if (result?.status === 200) {        
+        updateData(result.data)
```
jaitaiwan, line 73: The error doesn't appear to get cleared when data is returned.

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsComponent.tsx
```diff
@@ -0,0 +1,165 @@
...
+            <LineChart
+              data={data}
+              strokeWidth={2}
+              tickFormat={timeFormat}
+              height={height}
+              width={width}
+              legendProps={{ canSelectMultipleLegends: true, allowFocusOnLegends: true }}
+            />
+          </div>
+        }
```
jaitaiwan, line 155: I would split these into individual rendering consts or functions. It's not clear at a glance that the two dom trees are displayed dependent on `spinnerVisible`

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsComponent.tsx
```diff
@@ -0,0 +1,165 @@
...
+            <StackItem> 
+              <Spinner size={SpinnerSize.large} />
+            </StackItem>
+            <div style={rootStyle}>
+              <LineChart
+                data={data}
+                strokeWidth={2}
+                tickFormat={timeFormat}
+                height={height}
+                legendProps={{ canSelectMultipleLegends: true, allowFocusOnLegends: true }}
```
jaitaiwan, line 140: There's barely any difference in props between this invocation and the invocation on line 146. Consider creating a map of variables and remove/add props to it: `<LineChart {{...lineChartProps}} />`

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsComponent.tsx
```diff
@@ -0,0 +1,165 @@
...
+      }
+    },[props.metrics])
+
+    useEffect(() => {   
+      setData({
+        chartTitle: 'Line Chart',
+        lineChartData: points,
+      });
+      (points.length > 0) ? setSpinnerVisible(false) : setSpinnerVisible(true)
+    }, [points])
```
jaitaiwan, line 107: These use effects don't appear to me to be needed. If the props are updated the component will be re-rendered.

portal/v2/src/ClusterDetailListComponents/Statistics/Statistics.tsx
```diff
@@ -0,0 +1,229 @@
...
+                <Stack.Item grow>
+                  <Text>{"Global Graph Options"}</Text>
+                </Stack.Item>
+                <Stack.Item>
+                  <GraphOptionsComponent duration={globalDuration} setDuration={setGlobalDuration} endDate={globalEndDate} setEndDate={setGlobalEndDate}/>
+                </Stack.Item>
+              </Stack>
+            </ThemeProvider>
+          </Stack>
+        </>
```
jaitaiwan, line 70: The fragment `<> </>` around this is not needed as it has a single root element.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
```diff
@@ -0,0 +1,220 @@
...
+        return "168h"
+      case "2w":
+        return "336h"
+      case "4w":
+        return "672h"
+      case "8w":
+        return "1344h"
+    }
+    return duration
+  }
```
jaitaiwan, line 220: This entire function should just be a js object constant.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+  function timeToString(date: Date): string {
+    var str
+    let hourString = date.getHours().toString()
+    if (hourString.length === 1) {
+      str = "0" + hourString + ":"
+    } else {
+      str = hourString + ":"
+    }
```
jaitaiwan, line 184: No need for the else. Just set `str = ...` above the `if`.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
```diff
@@ -0,0 +1,220 @@
...
+    let minuteString = date.getMinutes().toString()
+    if (minuteString.length === 1) {
+      str = str + "0" + minuteString
+    } else {
+      str += minuteString
+    }
```
jaitaiwan, line 190: No need for this else. Just remove `minuteString` from 187 and then keep line 189 unconditional after the `if`.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
```diff
@@ -0,0 +1,220 @@
...
+            styles={timePickerStyles}
+            allowFreeform
+            placeholder={timeToString(convertToUTC(props.endDate))}
+            autoComplete="on"
+            onChange={onTimeChange}
+            defaultValue={convertToUTC(props.endDate)}
+            useComboBoxAsMenuWidth
+          />
+        </Stack>
+      </Stack>
```
jaitaiwan, line 172: This does not need a fragment `<></>` because there is a single root element.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  export const convertToUTC = (date: Date): Date => {
+    let localDate = new Date()
+    localDate.setFullYear(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate())
+    localDate.setHours(date.getUTCHours())
+    localDate.setMinutes(date.getUTCMinutes())
+    
+    return localDate
+  }
```
jaitaiwan, line 202: [GraphOptionsComponent.tsx](https://github.com/Azure/ARO-RP/pull/2674/files#diff-98976a43957c74e6d37bd36960c9565126b2f9abbbeadae63f19212660d27892) Contains similar or identical code. Consider turning it into a util function and importing it in both places.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
```diff
@@ -0,0 +1,220 @@
...
+      localDate.setUTCMinutes(date.getMinutes());
+      props.setEndDate(localDate)        
+     };
+    const onDateChange = (date: Date | null | undefined): void => { 
+      let localDate: Date = new Date();
+      localDate.setUTCFullYear(date!.getFullYear(), date!.getMonth(), date!.getDate())
+      localDate.setHours(props.endDate.getHours());
+      localDate.setMinutes(props.endDate.getMinutes());
+      props.setEndDate(localDate)  
+     };
```
jaitaiwan, line 129: These two functions appear to be almost duplicates and could be combined.

portal/v2/src/ClusterDetailList.tsx
```diff
@@ -70,6 +73,36 @@ export class ClusterDetailComponent extends Component<ClusterDetailComponentProp
...
+            <Statistics currentCluster={this.props.cluster!} detailPanelSelected={this.props.detailPanelVisible} loaded={this.props.isDataLoaded} statisticsType={"dns"}/>
+            </>
+          );
+        }
+        case "ingressstatistics":
+        {
+          return (
+            <>
+            <Statistics currentCluster={this.props.cluster!} detailPanelSelected={this.props.detailPanelVisible} loaded={this.props.isDataLoaded} statisticsType={"ingress"}/>     
+            </>
```
jaitaiwan, line 101: No need for fragments `<></>` around these two Components.

jaitaiwan (changes requested):
Really great PR and TONS going on with it. I gave some feedback on some react and general JS idioms as well as some notes on providing a nicer HTTP API. One thing I'd really like to see is just a big more in-code documentation especially around exported functions in go, and some simple tests for them too wouldn't hurt.

Great job!

pkg/portal/cluster.go
```diff
@@ -190,3 +195,59 @@ func (p *portal) machineSets(w http.ResponseWriter, r *http.Request) {
...
+			"/subscriptions/%s/resourceGroups/%s/providers/Microsoft.RedHatOpenShift/openShiftClusters/%s",
+			subscription, resourceGroup, clusterName))
+
+	promQuery, err := cluster.GetPromQuery(statisticsType)
+	if err != nil {
+		p.internalServerError(w, err)
+		return
+	}
+	prom := prometheus.New(p.log, p.dbOpenShiftClusters, p.dialer)
+	httpClient, err := prom.Cli(ctx, resourceID)
```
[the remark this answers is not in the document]
anshulvermapatel: yup .. agree but yeah I agree with Jeremy as well, not in the scope of this PR, though we can work on that by creating another JIRA ticket for this. Thanks for the suggestion.

pkg/portal/cluster.go
anshulvermapatel: Yup you are right ... I have made the changes in the code and will pushed them shortly.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
anshulvermapatel: Okay yup .. made the changes.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
anshulvermapatel: yup right .. made the changes.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
anshulvermapatel: No actually there is a difference, here functions are given as props to different FluentUI component firstly and secondly, `onDateChange` only the date gets changed and on `onTimeChange` on the time gets changed the date being same so they are different.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
anshulvermapatel: Done, will push it in a while.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
anshulvermapatel: yup done.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
anshulvermapatel: I would like to keep it as a function only .. but I will replace the switch case with a js object.

portal/v2/src/ClusterDetailListComponents/Statistics/Statistics.tsx
anshulvermapatel: done

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsComponent.tsx
anshulvermapatel: Okay, yaa anyways we are not doing anything right now at a Legend click so I removed the whole function itself.

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
jaitaiwan: I don't think the change I  had in mind is worth pushing for here.

portal/v2/src/ClusterDetailList.tsx
anshulvermapatel: done

portal/v2/src/ClusterDetailListComponents/Statistics/GraphOptionsComponent.tsx
anshulvermapatel: In that code the time/date is actually getting changed, here the time is only getting converted to UTC without any change.

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsComponent.tsx
anshulvermapatel: It is required, because, the useEffect just above this one updates the `points` but in react the state does not get updated immediately and hence if the put the logic of the useEffect in question to the above useEffect then, there would be no data for the graph to show, it would be spinner only, hence I have to set the data to be given to graph when the points get updated and then, there would be data for the graph.

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsComponent.tsx
anshulvermapatel: Okay, I get you, changing the same.

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsComponent.tsx
anshulvermapatel: I created a function and used that there. Do check. I will be pushing the changes shortly.

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsWrapper.tsx
anshulvermapatel: Oh yaa, I too noticed that while I was working on your review. setting the error null at receiving status 200.

portal/v2/src/Request.tsx
anshulvermapatel: Ohh yes, thanks for that. I have changed it.

pkg/portal/cluster/statistics.go
anshulvermapatel: Done.

pkg/portal/cluster.go
```diff
@@ -190,3 +195,59 @@ func (p *portal) machineSets(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	apiVars := mux.Vars(r)
+	subscription := apiVars["subscription"]
+	resourceGroup := apiVars["resourceGroup"]
+	clusterName := apiVars["clusterName"]
+	statisticsType := apiVars["statisticsType"]
+	resourceID := strings.ToLower(
+		fmt.Sprintf(
+			"/subscriptions/%s/resourceGroups/%s/providers/Microsoft.RedHatOpenShift/openShiftClusters/%s",
```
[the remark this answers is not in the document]
anshulvermapatel: I think it would be a good idea to create a function which will take these values and return the resourceID. What do you think?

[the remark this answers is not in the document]
anshulvermapatel: Haven't made this change for now, waiting for your reply. I am pushing rest of the changes.

portal/v2/src/ClusterDetailListComponents/Statistics/Statistics.tsx
```diff
@@ -0,0 +1,226 @@
...
+            </Stack.Item>  
+          </Stack>
+        </ThemeProvider>
+        <StatisticsWrapper currentCluster={props.currentCluster} detailPanelSelected={props.detailPanelSelected} loaded={props.loaded} statisticsName={lprops.statisticsName} duration={duration} endDate={endDate} graphHeight={200} graphWidth={740}/>
+      </Stack>
+      </>
+      )
+    }
+    
+    if (props.statisticsType === "api") {
```
ellis-johnson, line 143: Is there anyway we can create a another function to create these objects? Seems like we a reusing code a bit here

pkg/portal/portal.go
```diff
@@ -455,7 +436,23 @@ func (p *portal) getSubscriptionDocument(ctx context.Context, key string) (*api.
...
+func (p *portal) getResourceID(r *http.Request) string {
+	apiVars := mux.Vars(r)
+	subscriptionID := apiVars["subscription"]
+	resourceGroup := apiVars["resourceGroup"]
+	clusterName := apiVars["clusterName"]
+	return strings.ToLower(
+		fmt.Sprintf(
+			"/subscriptions/%s/resourceGroups/%s/providers/Microsoft.RedHatOpenShift/openShiftClusters/%s",
+			subscriptionID, resourceGroup, clusterName))
+}
```
jaitaiwan, line 448: I think this function should be changed to receive subscriptionId, resourceGroup and clusterName directly and not worry about getitng out the variables from the request, because the place you're calling `getResourceID` you're already extracting those variables out form apiVars.

pkg/portal/portal.go
```diff
@@ -455,7 +436,23 @@ func (p *portal) getSubscriptionDocument(ctx context.Context, key string) (*api.
...
+
 func (p *portal) internalServerError(w http.ResponseWriter, err error) {
 	p.log.Warn(err)
 	http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 }
+
+func (p *portal) badRequest(w http.ResponseWriter, err error) {
+	p.log.Warn(err)
```
jaitaiwan, line 456: I don't think we need to use `Warn` here. I expect `Debug` would be more appropriate because we don't really care about bad requests from clients.

portal/v2/src/ClusterDetailList.tsx
```diff
@@ -70,6 +73,32 @@ export class ClusterDetailComponent extends Component<ClusterDetailComponentProp
...
+          return (
+            <Statistics currentCluster={this.props.cluster!} detailPanelSelected={this.props.detailPanelVisible} loaded={this.props.isDataLoaded} statisticsType={"dns"}/>
+          );
+        }
+        case "ingressstatistics":
+        {
+          return (
+            <Statistics currentCluster={this.props.cluster!} detailPanelSelected={this.props.detailPanelVisible} loaded={this.props.isDataLoaded} statisticsType={"ingress"}/>     
+          );
+        }
```
jaitaiwan, line 99: I'd like to see this refactored into a map. See my comment on this PR here: https://github.com/Azure/ARO-RP/pull/2587#discussion_r1144217368

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsComponent.tsx
```diff
@@ -0,0 +1,167 @@
...
+          var lineChartPoint: ILineChartPoints = {
+            legend: metric.Name,
+            data: dataPoints,
+            color: colors[i]
+          }
+          newPoints.push(lineChartPoint)
+        })
+        setPoints(newPoints);
+      }
+      (props.fetchStatus === "error") ? setSpinnerVisible(false) : setSpinnerVisible(true)
```
jaitaiwan, line 98: Consider refactoring this so that you remove 1 if statement:
```
if (props.fetchStatus === "error") {
  ... do things
  return
}

// success actions here
```

portal/v2/src/Request.tsx
```diff
@@ -125,3 +126,16 @@ export const RequestKubeconfig = async (
     return OnError(err)
   }
 }
+
+export const FetchStatistics = async (cluster: ICluster, statisticsName: string, duration: string, endDate: Date): Promise<AxiosResponse | null> => {
```
jaitaiwan, line 130: Variable name shouldn't start with an upper-case letter as it's not a component.

jaitaiwan (changes requested):

portal/v2/src/Request.tsx
anshulvermapatel: it is actually the fetch function and it was as is from the starting for I continued with the same style.

portal/v2/src/ClusterDetailListComponents/Statistics/StatisticsComponent.tsx
anshulvermapatel: Yup, got it ... will refactor it.

portal/v2/src/ClusterDetailList.tsx
anshulvermapatel: Yes, you are right, but since this PR did not initiated the whole switch design, I think it would be better if we correct the whole thing and create another work item to refactor those and have whole another PR just for that for not only this piece of code but any other code as well in admin-portal.

pkg/portal/portal.go
anshulvermapatel: I actually wrote it to receive subscriptionId, resourceGroup, clusterName only previously but since I saw that most of the places where it is being used, these values are being extracted, so to make once piece of code reusable, I then changed it to take a *http.Request to extract these variables inside the function itself.

portal/v2/src/ClusterDetailList.tsx
jaitaiwan: If you're introducing new code, it should be following better standards, otherwise we keep creating PR debt. I'd understand if it wasn't your own code.

portal/v2/src/Request.tsx
jaitaiwan: I understand, however ideally the code should be following known React conventions, so while we're here touching this code, it would be better to make it right rather than leaving it for the next person to have to clean it up.

pkg/portal/portal.go
jaitaiwan: Hmmm, cause now we're extracting these variables twice, and if others use this code, they'd be forced to validate these same variables outside of this function. My suggestion then is either:
1. Move the apiVars validation code into this function, returning errors or something like that so that the calling function can send the appropriate http error response OR
2. Change it back to accepting only those values without extracting them.

portal/v2/src/ClusterDetailList.tsx
anshulvermapatel: Okay I will change the switch case to a map

portal/v2/src/ClusterDetailListComponents/Statistics/Statistics.tsx
anshulvermapatel: yup .. I will change it.

pkg/portal/portal.go
anshulvermapatel: I will change it back then.

portal/v2/src/Request.tsx
anshulvermapatel: I have changed all the functions in that file.

portal/v2/src/ClusterDetailList.tsx
```diff
@@ -38,38 +41,85 @@ interface IClusterDetailComponentState {
...
-          return (
-            <MachineSetsWrapper currentCluster={this.props.cluster!} detailPanelSelected={this.props.detailPanelVisible} loaded={this.props.isDataLoaded}/>
-          );
-        }
-      }
+    interface Map {
+      [key: string]: JSX.Element
+    }
+
+    const menus: Map = {
```
ellis-johnson, line 57: I think Daniel meant more like this https://github.com/Azure/ARO-RP/blob/9cb41357eb32f70a1f872be37435c3baeeb084fc/portal/v2/src/ClusterDetailList.tsx 

If you revert this change we can just handle it in the merge conflict

ellis-johnson (changes requested):

ellis-johnson (dismissed):
LGTM!

pkg/portal/portal.go
```diff
@@ -406,7 +401,20 @@ func (p *portal) serve(path string) func(w http.ResponseWriter, r *http.Request)
...
+	p.log.Debug(err)
```
ellis-johnson, line 418: Last little nitpick but do we need both warn and debug here?

ellis-johnson (changes requested):
One little comment

pkg/portal/portal.go
anshulvermapatel: ohh sorry, it should be only Debug

ellis-johnson (approved):

PR #2679, opened 2023-02-02
middleware improvements in frontend

facchettos:
this is part of https://issues.redhat.com/browse/ARO-1724?filter=-1

This removes the `function that returns a functions that returns a function` that is really hard to understand, and diverges from the standard middleware pattern in go. This allows us to have a simpler design and will reduce the difficulty for gorilla removal and the reviews
the diff is not looking good because whole files lose a level of nesting. This overall simplify the middleware stack 

Fixes


pkg/frontend/middleware/validate.go
```diff
@@ -23,100 +22,103 @@ const (
...
+		}
 
-			if err != nil || hasVariableAPIVersion {
-				if apiVersion != "" {
-					if _, found := apis[apiVersion]; !found {
-						api.WriteError(w, http.StatusBadRequest, api.CloudErrorCodeInvalidResourceType, "", "The resource type '%s' could not be found in the namespace '%s' for api version '%s'.", vars["resourceType"], vars["resourceProviderNamespace"], apiVersion)
-						return
-					}
+		if err != nil || hasVariableAPIVersion {
+			if apiVersion != "" {
```
s-amann, line 114: nit: can we reduce this nesting at least one level by combining 113 and 114
`(err != nil || hasVariableAPIVersion) && apiVersion != ""`

s-amann (changes requested):
lgtm, but we can get some further cleanup with a few minor changes, thanks!

pkg/frontend/middleware/validate.go
```diff
@@ -23,100 +22,103 @@ const (
...
-			if _, found := vars["resourceGroupName"]; found {
-				if !rxResourceGroupName.MatchString(vars["resourceGroupName"]) {
-					api.WriteError(w, http.StatusBadRequest, api.CloudErrorCodeResourceGroupNotFound, "", "Resource group '%s' could not be found.", vars["resourceGroupName"])
-					return
-				}
+			api.WriteError(w, http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "Internal server error.")
+			return
+		}
+
+		if _, found := vars["subscriptionId"]; found {
```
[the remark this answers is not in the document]
facchettos: Vars is going to go away (next ish step in my plan to remove gorilla), so I am not 100% sure it is worth doing it now since I will do it next anyways

pkg/frontend/middleware/validate.go
facchettos: wasn't the main purpose of that pr but yeah that's a simple enough change that doesn't blow up the scope. will do

pkg/frontend/middleware/validate.go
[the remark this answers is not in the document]
s-amann: ack, fine to leave it for now

pkg/frontend/middleware/validate.go
s-amann: ah, I thought part of the goal was to reduce nesting, thanks for including this change.

facchettos: so the point was to remove `a function that returns a function that returns a function` and go back to the standard middleware pattern

facchettos: and also my bad for not having a nice description

pkg/frontend/frontend.go
```diff
@@ -407,11 +421,11 @@ func (f *frontend) authenticatedRoutes(r *mux.Router) {
 
 func (f *frontend) setupRouter() *mux.Router {
 	r := mux.NewRouter()
-	r.Use(middleware.Log(f.env, f.auditLog, f.baseLog.WithField("component", "access")))
```
s-amann, line 410: I don't see this log field being added in the new changes, is its remove purposeful and why don't we want it anymore then?

facchettos: fixed

s-amann (approved):

m1kola:
> the diff is not looking good because whole files lose a level of nesting. This overall simplify the middleware stack

I recommend reviewing this with "Hide whitespace" option on (`git diff -w`).

m1kola (approved):

s-amann (approved):

PR #2681, opened 2023-02-03
Added new instance type

kartikaysharma27:

https://issues.redhat.com/browse/ARO-1295

Fixes

Added new instance types.
refer: https://learn.microsoft.com/en-us/azure/openshift/support-policies-v4


hawkowl:
@kartikaysharma27 I ran `az vm list-usage -l eastus | grep NCS -C 5` to check the family name -- it looks like there's just NCSv3, and the rs is just a part of that family. Since there's other NCS versions (probably in the future, v2 is retired), I also changed it to be "NCSv3".

cblecker (approved):

PR #2682, opened 2023-02-03
changing bootstrap vmsize from d4s to d8s to match default master size

jhoreman:

ARO-1485
Fixes

Sets the default deploy size of the boostrap VM to be equal to our default Master deploy size. This will ensure consistency with OCP deployment models as well as future proofing. 

### Test plan for issue:

Local development RP loaded with the change stood up and cluster deployment issued through it. Bootstrap VM was created with the correct size 

Size
Standard D8s v3
vCPUs
8
RAM
32 GiB

No, done for consistency purposes with other openshift installers.

### DIFF
--- a/pkg/installer/deployresources_resources.go
+++ b/pkg/installer/deployresources_resources.go
@@ -106,7 +106,7 @@ func (m *manager) computeBootstrapVM(installConfig *installconfig.InstallConfig)
 	vm := &mgmtcompute.VirtualMachine{
 		VirtualMachineProperties: &mgmtcompute.VirtualMachineProperties{
 			HardwareProfile: &mgmtcompute.HardwareProfile{
-				VMSize: mgmtcompute.VirtualMachineSizeTypesStandardD4sV3,
+				VMSize: mgmtcompute.VirtualMachineSizeTypesStandardD8sV3,
 			},
 			StorageProfile: &mgmtcompute.StorageProfile{
 				ImageReference: &mgmtcompute.ImageReference{


pkg/installer/deployresources_resources.go
```diff
@@ -106,7 +106,7 @@ func (m *manager) computeBootstrapVM(installConfig *installconfig.InstallConfig)
 	vm := &mgmtcompute.VirtualMachine{
 		VirtualMachineProperties: &mgmtcompute.VirtualMachineProperties{
 			HardwareProfile: &mgmtcompute.HardwareProfile{
-				VMSize: mgmtcompute.VirtualMachineSizeTypesStandardD4sV3,
+				VMSize: mgmtcompute.VirtualMachineSizeTypesStandardD8sV3,
```
SudoBrendan, line 109: I think instead of a hard-coded value, we'll want to use whatever the Master size is, which is variable based on customer selection. Can we update this to use whatever was passed in as the master size?

SudoBrendan (changes requested):

pkg/installer/deployresources_resources.go
SudoBrendan: In addition, this change shouldn't go in the "old" installer code, let's put it into the 4.11 branch on the ARO-Installer repo: https://github.com/openshift/ARO-Installer/blob/main/pkg/installer/deployresources_resources.go

bennerv (commented):
Our current implementation of the quota checker doesn't include the bootstrap VM.  I believe the bootstrap VM is torn down after the control plane VMs are created, so it should probably be included in the quota validation as well.

SudoBrendan:
@jhoreman - are we good to close this one out? It sounds like you've moved to a new PR in the OpenShift/ARO-Installer repo for this change.

PR #2684, opened 2023-02-07
Detect expected NSG's during install retries

darthhexx:

Fixes: Retried cluster installs that have already attached our expected NSG to the subnets.

If an installation process gets past the attachNSG step and then restarts, either due to an RP restart from an upgrade, panic, or CosmoDB lease expiry, the install fails due to a false positive check for any NSG being attached to the master or worker subnets.

This PR adds the same [check that already takes place during install](https://github.com/Azure/ARO-RP/blob/a3c9e536a3b509c99f5961b35168a1fdd3467b26/pkg/cluster/deploystorage.go#L194-L209) "forward" into the dynamic validation.

### Test plan for issue:

Create test clusters using a shared environment.

Nope.

### DIFF
--- a/pkg/api/validate/dynamic/dynamic.go
+++ b/pkg/api/validate/dynamic/dynamic.go
@@ -436,7 +436,14 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 		if oc.Properties.ProvisioningState == api.ProvisioningStateCreating {
 			if ss.SubnetPropertiesFormat != nil &&
 				ss.SubnetPropertiesFormat.NetworkSecurityGroup != nil {
-				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidLinkedVNet, s.Path, "The provided subnet '%s' is invalid: must not have a network security group attached.", s.ID)
+				expectedNsgID, err := subnet.NetworkSecurityGroupID(oc, s.ID)
+				if err != nil {
+					return err
+				}
+
+				if !strings.EqualFold(*ss.SubnetPropertiesFormat.NetworkSecurityGroup.ID, expectedNsgID) {
+					return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidLinkedVNet, s.Path, "The provided subnet '%s' is invalid: must not have a network security group attached.", s.ID)
+				}
 			}
 		} else {
 			nsgID, err := subnet.NetworkSecurityGroupID(oc, *ss.ID)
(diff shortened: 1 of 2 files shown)


facchettos (approved):

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -436,7 +436,14 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 		if oc.Properties.ProvisioningState == api.ProvisioningStateCreating {
 			if ss.SubnetPropertiesFormat != nil &&
 				ss.SubnetPropertiesFormat.NetworkSecurityGroup != nil {
-				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidLinkedVNet, s.Path, "The provided subnet '%s' is invalid: must not have a network security group attached.", s.ID)
+				expectedNsgID, err := subnet.NetworkSecurityGroupID(oc, s.ID)
```
m1kola, line 439: I think we are adressing symptom, not the root cause: we need to understand why in some cases (like [what we discussed](https://redhat-internal.slack.com/archives/C02ULBRS68M/p1675371441574689) in Slack recently) validation fails.

This validation should happen before any provisioning of NSG.

m1kola (changes requested):

pkg/api/validate/dynamic/dynamic.go
darthhexx: The issue arises if the cluster document is picked up again and the install process rerun. This can happen if the RP panics, RP is upgraded, AKS is upgraded, or if another instance picks up a lease for an install operation.

You can emulate the failure by stopping the install process manually after [the attachNSG step](https://github.com/Azure/ARO-RP/blob/78acd1eed34b9c63af50ed8a2485a61836429cfa/pkg/cluster/install.go#L246) has been run.

The RP should be able to rerun the install process if it fails due to the reasons listed above, but currently it cannot.

bennerv: Yeah the cluster install should always be idempotent, but it's not in this specific instance.  (unless this PR is merged)

cadenmarchese (approved):

pkg/api/validate/dynamic/dynamic.go
darthhexx: @m1kola I don't think there is anything else to add to this PR, since its intention is to unblock re-queued installs, would you mind giving it another look-over?

m1kola: Thanks for providing more context. Makes sense.

m1kola (approved):

PR #2686, opened 2023-02-07
Remove systemdata from admin API

bennerv:

Removes SystemData from being returned in the admin API "GET" response.  

It is no longer needed and to comply with MAG requirements.

### Test plan for issue:

testing in INT tomorrow

nope


cblecker (approved):
lgtm

PR #2688, opened 2023-02-07
Remove unused clusterimageset

bennerv:
According to @darthhexx we don't use this CR, so I'm removing it so we don't confuse the team :rocket:


darthhexx (approved):
lgtm

s-amann (approved):

PR #2693, opened 2023-02-08
Refactor/env vars access

AldoFusterTurpin:

This PR moves the access to environment variable to the main package and injects those values in parameters to not couple internal packages with infrastructure details. (this can be consideres some sort of dependency inversion as some packages don't depend now in the runtime environment and just depend on standard Go variables (strings, ints, etc.) injected as parameters. It also has the effect of reducing the number of times we access env variables as we just read them once and inject them several times (eg: in cmd/aro/monitor.go, cmd/aro/portal.go and others).

In fact, this PR removes more lines of code than it adds as I created env.ValidateVars() to stick to DRY to stop copy-pasting every time we want to validate a list of environtment variables.

(
This PR does not represent the direct application of Clean Architecture as it would be very difficult due to the nature and complexity of the project.

I performed a [presentation](https://drive.google.com/file/d/1O0l-cUDu5agokl8B-WFNRnK3yvBz1zKY/view?usp=sharing) talking about key topics of these PR changes and about some principles that can help the Design of the project in the long term to have a shape more similar to Clean Architecture improving what we have today.

This -> [PR](https://github.com/Azure/ARO-RP/pull/2694) is also part of the same task. Using both PRs I  presented key topics that move the direction of the design towards Clean Architecture.

This PR contains > 1 commit so it is easier to navigate and understand. The idea is to squash them when merging but right now I want it that way to use it  as a reference in the presentation I conducted.
)

### Test plan for issue:

Unit tests exist. I performed small steps in each commit to not mess up things.

N/A.

---

### Problems of using env vars everywhere
- We couple our code to infrastructure details (runtime environment).
- We can not provide different values without modifying the environment (as they are hardcoded).
- Changing the name of an env variable affects multiple parts of the codebase (as the code is coupled to the name we use to identify that env var).
- We can not write unit tests for those functions as one Unit Test should not touch the runtime environment (I/O, HTTP, file system, etc.) to work.
- We loose control of the environment we set as we read/set/validate env variables in multiple places instead of centralising that in a single place.

The image below shows some ideas of the refactoring.
example2

### Code evolution walkthrough (what was explained in the presentation, optional)
We first try to extract the common parts to follow DRY principle.

After the commit  *refactor: use env.ValidateVars() when possible* with ID e202, we see in:
- pkg/util/instancemetadata/dev.go
- pkg/util/instancemetadata/prodfromenv.go
- pkg/util/keyvault/uri.go
that we can't use the env package because we will have the following compile errors due to import cycle errors:
```
go build -tags aro,containers_image_openpgp ./...
package github.com/Azure/ARO-RP/cmd/aro
        imports github.com/Azure/ARO-RP/pkg/api/v20191231preview
        imports github.com/Azure/ARO-RP/pkg/api/validate
        imports github.com/Azure/ARO-RP/pkg/api/validate/dynamic
        imports github.com/Azure/ARO-RP/pkg/env
        imports github.com/Azure/ARO-RP/pkg/util/keyvault
        imports github.com/Azure/ARO-RP/pkg/env: import cycle not allowed
package github.com/Azure/ARO-RP/cmd/aro
        imports github.com/Azure/ARO-RP/pkg/api/v20191231preview
        imports github.com/Azure/ARO-RP/pkg/api/validate
        imports github.com/Azure/ARO-RP/pkg/api/validate/dynamic
        imports github.com/Azure/ARO-RP/pkg/env
        imports github.com/Azure/ARO-RP/pkg/util/keyvault
        imports github.com/Azure/ARO-RP/pkg/env: import cycle not allowed
make: *** [build-all] Error 1
```

Even though we have improved so far (we now apply the DRY principle), we can do better if instead of reading the env variables inside those packages, we directly inject the values from the outer level (*main* package, where the app is started). 

In other words: whenever possible, just touch (read/write) env variables in the *main* package and inject those values down (in the methods call) to the corresponding packages (Dependencies Injection principle). By doing that we will decouple our functions/packages from the *os* runtime dependency. This is what we do in the next commits.

After the commit *refactor: inject dbAccountName into NewMasterKeyAuthorizer* with ID 8eab we see in cmd/aro/monitor.go:85 we can get just *once* the DatabaseAccountName and use it twice. Before that commit, we were checking and accessed that environment variable twice. Now that has been fixed thanks to the injection in the parameter of the *NewDatabaseClient* and *NewMasterKeyAuthorizer* functions. We also extracted the hardcoded string of the env variable into a const.go file for better maintainability (where we will define all the env var names).

In the next commit we will try to get rid of the pkg/database/database.go:95 *Name* function because it has some drawbacks:
- validates env variables
- gets the value of env variables
- it is used in 11 different places (propagating the coupling with the runtime environment and reading a lot of times from the runtime when it is not necessary. It is bad both from the design perspective and execution time)

We will inject the database name instead of retrieving the value from inside the *Name* function.

After the commit *refactor: move function Name from database package to main package and inject dbName as parameter* with ID ba8d, the database package does not depend on the *os* package and we inject dbName everywhere where it's possible instead of accessing env vars from everywhere were Name function was called. We also reduce the number of times that we access the runtime environment as we see in cmd/aro/rp.go where we read once the value *dbName* and inject it multiple places.

Following this idea, if we search for *os.Getenv* in the codebase, we see 82 matches in the *pkg/* folder. We should get the env variables from outside and inject them to the different components (maybe the *pkg/env* package is an exception but still can be improved).

After the commit *refactor: inject url into NewClient and NewRefresher, move PROXY_HOSTNAME validation outside newDev function, replace env.ValidateVars with ValidateVars and create constants*, with ID:
- we are injecting more values to more functions and we can simplify some of them as they do not return errors (because they no longer validate env vars). 
- We have extracted some hardcoded strings into constants.
- We replace env.ValidateVars with ValidateVars when possible. 

After the commit *refactor: use env.ValidateVars() and remove the others ValidateVars() function declarations* with commit ID 11d6, we just define once ValidateVars() in the env package and call it from multiple files (most of the time from *main* package). This is not a problem now because if there is a package apart from *main* that can read/write env variables this is the *env* package. (I would prefer just handle env vars in the *main* package but at the moment it is not possible as there are 23 occurrences of *os.Getenv* in the env package). At least at this point we have moved a lot of env vars validation and reading outside the internal packages and we inject the values from the *main* package. We could keep iterating and fixing that but the *IsLocalDevelopmentMode* flag present in a lot of places is not helping as we would need to move all that logic to outer layers.

### DIFF
--- a/cmd/aro/const.go
+++ b/cmd/aro/const.go
@@ -4,6 +4,8 @@ package main
 // Licensed under the Apache License 2.0.
 
 const (
+	DatabaseName        = "DATABASE_NAME"
 	DatabaseAccountName = "DATABASE_ACCOUNT_NAME"
 	KeyVaultPrefix      = "KEYVAULT_PREFIX"
+	DBTokenUrl          = "DBTOKEN_URL"
 )
--- a/cmd/aro/const.go
+++ b/cmd/aro/const.go
@@ -5,4 +5,5 @@ package main
 
 const (
 	DatabaseAccountName = "DATABASE_ACCOUNT_NAME"
+	KeyVaultPrefix      = "KEYVAULT_PREFIX"
 )
--- a/cmd/aro/const.go
+++ b/cmd/aro/const.go
@@ -1,0 +1,8 @@
+package main
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+const (
+	DatabaseAccountName = "DATABASE_ACCOUNT_NAME"
+)
--- a/cmd/aro/dbtoken.go
+++ b/cmd/aro/dbtoken.go
@@ -55,7 +55,7 @@ func dbtoken(ctx context.Context, log *logrus.Entry) error {
 
 	go g.Run()
 
-	if err := env.ValidateVars(DatabaseAccountName); err != nil {
+	if err := ValidateVars(DatabaseAccountName); err != nil {
 		return err
 	}
 
--- a/cmd/aro/dbtoken.go
+++ b/cmd/aro/dbtoken.go
@@ -83,11 +83,11 @@ func dbtoken(ctx context.Context, log *logrus.Entry) error {
 		return err
 	}
 
-	dbtokenKeyvaultURI, err := keyvault.URI(_env, env.DBTokenKeyvaultSuffix)
-	if err != nil {
+	if err := ValidateVars(KeyVaultPrefix); err != nil {
 		return err
 	}
-
+	keyVaultPrefix := os.Getenv(KeyVaultPrefix)
+	dbtokenKeyvaultURI := keyvault.URI(_env, env.DBTokenKeyvaultSuffix, keyVaultPrefix)
 	dbtokenKeyvault := keyvault.NewManager(msiKVAuthorizer, dbtokenKeyvaultURI)
 
 	servingKey, servingCerts, err := dbtokenKeyvault.GetCertificateSecret(ctx, env.DBTokenServerSecretName)
--- a/cmd/aro/dbtoken.go
+++ b/cmd/aro/dbtoken.go
@@ -5,7 +5,6 @@ package main
 
 import (
 	"context"
-	"fmt"
 	"net"
 	"os"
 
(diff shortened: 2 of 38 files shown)


facchettos (approved):

tsatam (approved):
LGTM, thanks for this refactor!

This shouldn't block this PR at all, but is there planned work to handle the TODOs touched by this? (specifically the usage of the service keyvault in places marked with TODOs to remove it)

AldoFusterTurpin:
> LGTM, thanks for this refactor!
> 
> This shouldn't block this PR at all, but is there planned work to handle the TODOs touched by this? (specifically the usage of the service keyvault in places marked with TODOs to remove it)

Good question. I am not aware of any plans to handle those TODOs. As I did not touch in any way the behaviour of it I just left the comments.

kimorris27 (approved):
This is good work! Thank you for the level of detail you put in the PR description.

ellis-johnson (approved):
Fantastic refactor, great work

petrkotas (approved):
@AldoFusterTurpin Thank you for the PR, it is extremely useful and a really good work! LGTM

PR #1373, opened 2021-03-09
DNS update fixes

jim-minter:
don't remove dns private zones if not all nodes are under MCPs
don't remove dns private zones on 4.3 clusters
remove private dns zone config setting on upgrade
also update dns config on clusters which are already upgraded

PR #2694, opened 2023-02-08
Refactor/remove private dns zone

AldoFusterTurpin:

This PR does not represent the direct application of Clean Architecture as it would be very difficult due to the nature and complexity of the project.

I will perform a presentation talking about key topics of these PR changes and about some principles that can help the Design of the project in the long term to have a shape more similar to Clean Architecture improving what we have today.

The [PR](https://github.com/Azure/ARO-RP/pull/2693) is also part of the same task. Using both PRs I will be able to present key topics that move the direction of the design towards Clean Architecture.

This PR contains > 1 commit so it is easier to navigate and understand. The idea is to squash them when merging but right now I want it that way in case I need to refer to a particular step in the presentation I will conduct.

### Test plan for issue:

Unit tests exist. I performed small steps in each commit to not mess up things.

N/A. I will conduct a presentation with additional material + explanations.


facchettos (approved):

petrkotas (approved):
Good work @AldoFusterTurpin I believe this is simplifying our code.

pkg/cluster/removeprivatednszone.go
```diff
@@ -27,125 +28,85 @@ func (m *manager) removePrivateDNSZone(ctx context.Context) error {
...
-		r, err := azure.ParseResourceID(*zone.ID)
-		if err != nil {
-			m.log.Print(err)
-			return nil
+	return true, nil
+}
+
+// totalMachinesInTheMCPs returns the total number of machines in the machineConfigPools
+// and an error, if any.
+func totalMachinesInTheMCPs(machineConfigPools []mcv1.MachineConfigPool) (int, error) {
```
s-amann, line 98: I'd like to see unit tests for new functions, especially ones with clearly defined scope.

pkg/cluster/removeprivatednszone.go
```diff
@@ -27,125 +28,85 @@ func (m *manager) removePrivateDNSZone(ctx context.Context) error {
...
-			return err
-		}
-
-		if dns.Spec.PrivateZone == nil ||
-			!strings.HasPrefix(strings.ToLower(dns.Spec.PrivateZone.ID), strings.ToLower(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID)) {
-			return nil
-		}
+// sameNumberOfNodesAndMachines returns true if the cluster has the same number of nodes and total machines,
+// and an error if any. Otherwise it returns false and nil.
+func (m *manager) sameNumberOfNodesAndMachines(ctx context.Context) (bool, error) {
```
s-amann, line 72: same idea, this method has a clear scope and now can have clear unit tests around it.

pkg/cluster/removeprivatednszone.go
```diff
@@ -27,125 +28,85 @@ func (m *manager) removePrivateDNSZone(ctx context.Context) error {
...
 
-	if v.Lt(version.NewVersion(4, 4)) {
-		// 4.3 uses SRV records for etcd
-		m.log.Printf("cluster version < 4.4, not removing private DNS zone")
+	if err := utilnet.RemoveZones(ctx, m.virtualNetworkLinks, m.privateZones, zones, resourceGroup); err != nil {
+		m.log.Print(err)
 		return nil
 	}
+	return nil
+}
```
s-amann, line 68: the tests around this function in `pkg/cluster/removeprivatednszone_test.go` had to do its mocking with `mock_privatedns` I'd expect with this type of refactoring the tests to be changed and added to `pkg/util/net.go` and the mocks here to be updated to capture the new use of `utilnet`.

s-amann (commented):
Great work on the refactor and explanation on how you've applied coding principles to the ARO-RP codebase. While it is great the refactor does not break existing tests, those tests are now crossing the boundary of separation you've created to abstract out the DNS related code. 

I don't think this warrants blocking this from merge, but I would like to hear your thoughts on the refactoring of tests, and if you wouldn't mind having a new WI to implement those changes.

pkg/cluster/removeprivatednszone.go
AldoFusterTurpin: Done 😄

AldoFusterTurpin:
@s-amann Thanks for the review. I didn't update the Unit Tests in the first place as the new functions/types are not used elsewhere but I have updated the Unit Tests now as it is an improvement.
Thanks! 😄

pkg/util/net/net.go
```diff
@@ -72,3 +89,163 @@ func setBuffers(rc syscall.RawConn, sz int) error {
...
+
+	if clusterVersionIsLessThan4_4 {
+		log.Printf("cluster version < 4.4, not removing private DNS zone")
+		return nil
+	}
+
+	if err = UpdateDNSs(ctx, configcli.ConfigV1().DNSes(), resourceGroupID); err != nil {
+		log.Print(err)
+		return nil
+	}
```
jaitaiwan, line 159: We have UpdateDNSs run twice but there's not really any obvious reason why the way this function works, works. Can we possibly document this with some comments?

jaitaiwan (commented):
Didn't have time to review more than 1 of the main files.

pkg/util/net/net.go
AldoFusterTurpin: Good question, I am not sure about the why as this is a refactor.

In fact, UpdateDNSs is never run twice due to the return on line pkg/util/net/net.go:122 so it is kind of an early return when  len(zones) == 0. The original chunk of code is coming from https://github.com/Azure/ARO-RP/pull/1373.

pkg/util/net/net.go
```diff
@@ -72,3 +89,163 @@ func setBuffers(rc syscall.RawConn, sz int) error {
...
+
+	if clusterVersionIsLessThan4_4 {
+		log.Printf("cluster version < 4.4, not removing private DNS zone")
+		return nil
+	}
+
+	if err = UpdateDNSs(ctx, configcli.ConfigV1().DNSes(), resourceGroupID); err != nil {
```
dem4gus, line 156: nit: plural on `UpdateDNSs` does not match plural on `configcli.ConfigV1().DNSes()`

pkg/util/net/net.go
```diff
@@ -72,3 +89,163 @@ func setBuffers(rc syscall.RawConn, sz int) error {
...
+	mcocli mcoclient.Interface,
+	kubernetescli kubernetes.Interface,
+	vNetLinksClient privatedns.VirtualNetworkLinksClient,
+	resourceGroupID string) error {
+	if privateZonesClient == nil {
+		return errors.New("privateZonesClient is nil")
+	}
+
+	resourceGroup := stringutils.LastTokenByte(resourceGroupID, '/')
+	zones, err := privateZonesClient.ListByResourceGroup(ctx, resourceGroup, nil)
```
dem4gus, line 107: Elsewhere you're verbose enough to specify that these are private zones, like in RemoveZones. Can we do that here too and make this variable `privateZones`?

dem4gus (commented):
Overall a great refactor that should make things must more testable. I just have a couple comments/questions.

pkg/util/ready/ready.go
```diff
@@ -193,3 +194,75 @@ func CheckMachineConfigPoolIsReady(ctx context.Context, cli mcoclientv1.MachineC
...
+		if !MachineConfigPoolIsReady(&mcp) {
+			return 0, fmt.Errorf("MCP %s not ready", mcp.Name)
+		}
+
+		totalMachines += int(mcp.Status.MachineCount)
+	}
+	return totalMachines, nil
+}
+
+func McpContainsARODNSConfig(mcp mcv1.MachineConfigPool) bool {
```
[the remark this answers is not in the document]
AldoFusterTurpin: Done.

pkg/util/net/net.go
AldoFusterTurpin: Done.

pkg/util/ready/ready.go
```diff
@@ -193,3 +194,75 @@ func CheckMachineConfigPoolIsReady(ctx context.Context, cli mcoclientv1.MachineC
...
+	List(ctx context.Context, opts metav1.ListOptions) (*mcv1.MachineConfigPoolList, error)
+}
+
+type NodeLister interface {
+	List(ctx context.Context, opts metav1.ListOptions) (*corev1.NodeList, error)
+}
+
+// SameNumberOfNodesAndMachines returns true if the cluster has the same number of nodes and total machines,
+// and an error if any. Otherwise it returns false and nil.
+func SameNumberOfNodesAndMachines(ctx context.Context, mcpLister MCPLister, nodeLister NodeLister) (bool, error) {
```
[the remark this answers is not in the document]
AldoFusterTurpin: I am not sure about that (as this is just a refactor I just focused about keeping the same behaviour)
What is the alternative you have in mind? 🙂

pkg/util/net/net_test.go
```diff
@@ -0,0 +1,351 @@
...
+
+var (
+	resourceGroupName = "testGroup"
+	subscriptionID    = "0000000-0000-0000-0000-000000000000"
+	resourceGroupID   = "/subscriptions/" + subscriptionID + "/resourceGroups/" + resourceGroupName
+	vnetName          = "testVnet"
+	resourceID        = resourceGroupID + "/providers/Microsoft.Network/virtualNetworks/" + vnetName
+)
+
+const id = "/subscriptions/0000000-0000-0000-0000-000000000000/resourceGroups/testGroup/providers/Microsoft.Network/privateDnsZones/zone1"
```
mociarain, line 40: Do we have any concept of common test fixtures or helpers?

pkg/cluster/removeprivatednszone.go
```diff
@@ -5,147 +5,11 @@ package cluster
...
-		err = m.privateZones.DeleteAndWait(ctx, resourceGroup, r.ResourceName, "")
-		if err != nil {
-			m.log.Print(err)
-			return nil
-		}
-	}
-
-	return nil
+	resourceGroupID := m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID
+	return utilnet.RemovePrivateDNSZone(ctx, m.log, m.privateZones, m.configcli, m.mcocli, m.kubernetescli, m.virtualNetworkLinks, resourceGroupID)
```
mociarain, line 14: This is a lot of arguments to pass IMO. Why not collapse it to some other structure?

mociarain (commented):
This is great. It's much clearer and easier for me (a newbie to Go and the repo as a whole) to follow and understand.

Let me know if I can help push this over the line.

pkg/cluster/removeprivatednszone.go
AldoFusterTurpin: Good idea

pkg/util/net/net_test.go
AldoFusterTurpin: Yes, we use them mainly in the e2e (like BeforeEach), but in Unit Tests we simply define redundant things as a const if it is used by all (or many) test cases.

pkg/util/net/net.go
```diff
@@ -72,3 +89,163 @@ func setBuffers(rc syscall.RawConn, sz int) error {
...
+		return errors.New("configcli is nil")
+	}
+
+	if len(privateZones) == 0 {
+		// fix up any clusters that we already upgraded
+		if err := UpdateDNSes(ctx, configcli.ConfigV1().DNSes(), resourceGroupID); err != nil {
+			log.Print(err)
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
AldoFusterTurpin: Good question. I don't know 🙂 

The main idea is to not modify the behavior so I just keep it is for now (if that is okay).

pkg/util/net/net.go
```diff
@@ -72,3 +89,163 @@ func setBuffers(rc syscall.RawConn, sz int) error {
...
+func RemovePrivateDNSZone(
+	ctx context.Context,
+	log *logrus.Entry,
+	privateZonesClient privatedns.PrivateZonesClient,
+	configcli configclient.Interface,
+	mcocli mcoclient.Interface,
+	kubernetescli kubernetes.Interface,
+	vNetLinksClient privatedns.VirtualNetworkLinksClient,
+	resourceGroupID string) error {
+	if privateZonesClient == nil {
```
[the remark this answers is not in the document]
AldoFusterTurpin: Yeah, checking an interface against nil is a pretty common thing. _Can we push back on the use of interfaces?_ -> Passing an interface vs passing a struct have different use cases. We usually pass an interface in the tests when we want to mock that interface because it is a dependency that some function or constructor requires and want to control what it returns. As you mention, in some cases changing an interface by a struct is not a win as with the struct you still need to compare against an empty struct of that type (sometimes people use a pointer to the struct to check if it is empty or not). 

In our case is a bit different because the struct that has the same name as the interface is just a wrapper of our Azure clients that exists for mocking purposes (if I am not mistaken) +  it contains some extra functionality on top of the standard mgmtprivatedns.PrivateZonesClient (pkg/util/azureclient/mgmt/privatedns/privatezones.go). It is a common pattern you will see in our clients.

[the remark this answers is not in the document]
mociarain: Thanks for the explanation. I can't say I love it but I guess I'll learn to live with it ;)

mociarain (approved):

pkg/cluster/delete.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/azureclient"
 	"github.com/Azure/ARO-RP/pkg/util/azureerrors"
 	"github.com/Azure/ARO-RP/pkg/util/dns"
+	utilnet "github.com/Azure/ARO-RP/pkg/util/net"
```
dkeohane, line 28: minor preference: naming to follow stringutils, so netutils instead of utilnet

pkg/util/net/net.go
```diff
@@ -72,3 +89,165 @@ func setBuffers(rc syscall.RawConn, sz int) error {
 
 	return err
 }
+
+type PrivateZoneDeletionConfig struct {
```
dkeohane, line 93: Do we make any distinction between Delete and Remove in this context? Should this be renamed to PrivateZoneRemovalConfig or something? 

The reason I ask is because we've a mix of remove/delete in this file and wanted to ensure I didn't miss something

pkg/util/net/net.go
```diff
@@ -72,3 +89,165 @@ func setBuffers(rc syscall.RawConn, sz int) error {
...
+	if config.Kubernetescli == nil {
+		return errors.New("kubernetescli is nil")
+	}
+
+	mcpLister := config.Mcocli.MachineconfigurationV1().MachineConfigPools()
+	nodeLister := config.Kubernetescli.CoreV1().Nodes()
+	sameNumberOfNodesAndMachines, err := ready.SameNumberOfNodesAndMachines(ctx, mcpLister, nodeLister)
+	if err != nil {
+		config.Log.Print(err)
+		return nil
```
dkeohane, line 140: Should we be returning the err here? Same applies to the rest of the return nils under this block

pkg/cluster/delete.go
AldoFusterTurpin: Good point. I have chosen utilnet as we already use it in a few places:
- ./pkg/gateway/gateway_test.go
- ./pkg/gateway/https.go
- ./pkg/proxy/proxy.go

Maybe this should change in the future, not sure. 🙂

pkg/util/net/net.go
AldoFusterTurpin: Right. Changed now. Thanks!

pkg/util/net/net.go
AldoFusterTurpin: Good question. The thing is that we weren't doing that in the original code and here I just want to keep the same behavior. Even if returning nil when there is an error is counterintuitive (it is for sure!), I should keep the logic the same way it was before the PR.

dkeohane: Sure thing, I didn't really check to see what changes you made from the original, just something I noticed :) Should we cut a card to ensure this doesn't get lost?

PR #2696, opened 2023-02-08
Move all e2e tests to use latest API version

cblecker:
Fixes

Updates the e2e tests to all use the latest API client version

### Test plan for issue:
The e2e test it self will show any issues

### DIFF
--- a/test/e2e/disk_encryption.go
+++ b/test/e2e/disk_encryption.go
@@ -33,7 +33,7 @@ import (
 var _ = Describe("Encryption at host", func() {
 	It("must be enabled on the test cluster and each VM must have encryption at host enabled", func(ctx context.Context) {
 		By("getting the test cluster resource")
-		oc, err := clients.OpenshiftClustersv20220401.Get(ctx, vnetResourceGroup, clusterName)
+		oc, err := clients.OpenshiftClusters.Get(ctx, vnetResourceGroup, clusterName)
 		Expect(err).NotTo(HaveOccurred())
 
 		By("checking that encryption at host is enabled for masters")
@@ -69,7 +69,7 @@ var _ = Describe("Encryption at host", func() {
 var _ = Describe("Disk encryption at rest", func() {
 	It("must be enabled with customer managed key for the cluster and each disk must have it enabled", func(ctx context.Context) {
 		By("getting the test cluster resource")
-		oc, err := clients.OpenshiftClustersv20220401.Get(ctx, vnetResourceGroup, clusterName)
+		oc, err := clients.OpenshiftClusters.Get(ctx, vnetResourceGroup, clusterName)
 		Expect(err).NotTo(HaveOccurred())
 
 		By("checking that disk encryption at rest is enabled for masters")
--- a/test/e2e/disk_encryption.go
+++ b/test/e2e/disk_encryption.go
@@ -12,7 +12,6 @@ import (
 	mgmtcompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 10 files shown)


test/e2e/setup.go
```diff
@@ -50,9 +49,8 @@ import (
 const seleniumContainerName = "selenium-edge-standalone"
 
 type clientSet struct {
-	OpenshiftClustersv20200430 redhatopenshift20200430.OpenShiftClustersClient
-	Operationsv20200430        redhatopenshift20200430.OperationsClient
-	OpenshiftClustersv20220401 redhatopenshift20220401.OpenShiftClustersClient
```
bennerv, line 55: We would reintroduce different versions back in to ensure specific APIs had specific fields in the future correct?

bennerv (approved):
lgtm - just one question

test/e2e/setup.go
cblecker: Maybe? I guess if we want to have e2e tests for future/preview APIs. I could name this as "OpenshiftClustersDefault" or something if we wanted to be able to add more clients later.

I think @m1kola might also have some input here as I believe he's been working on our tests in this area.

m1kola (approved):
This specific change is ok, but we need to do better at testing permutations in general.

test/e2e/disk_encryption.go
```diff
@@ -33,20 +32,20 @@ import (
...
 		By("getting the test cluster resource")
-		oc, err := clients.OpenshiftClustersv20220401.Get(ctx, vnetResourceGroup, clusterName)
+		oc, err := clients.OpenshiftClusters.Get(ctx, vnetResourceGroup, clusterName)
 		Expect(err).NotTo(HaveOccurred())
 
 		By("checking that encryption at host is enabled for masters")
 		Expect(oc.OpenShiftClusterProperties).To(Not(BeNil()))
 		Expect(oc.OpenShiftClusterProperties.MasterProfile).To(Not(BeNil()))
-		Expect((*oc.OpenShiftClusterProperties.MasterProfile).EncryptionAtHost).To(Equal(mgmtredhatopenshift20220401.Enabled))
+		Expect((*oc.OpenShiftClusterProperties.MasterProfile).EncryptionAtHost).To(BeEquivalentTo("Enabled"))
```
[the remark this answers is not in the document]
cblecker: This was actually an issue when I bumped the API version.. I got the following error:
```
Expected
    <redhatopenshift.EncryptionAtHost>: Enabled
to equal
    <redhatopenshift.EncryptionAtHost>: Enabled
```

Normally I'd agree on using a constant and not a hard coded string.. My concern though is that this is the only test that is using a versioned constant and would need to be bumped each time the API is or else it will fail (which is what `BeEquivalentTo` is meant to test). If the value of that constant changed in the future for some reason, this test would still catch that and would need to be updated. This makes the test generic where it can test multiple API versions and still ensure that the actual value that's stored in the cluster doc remains the same. Thoughts?

test/e2e/setup.go
cblecker: Agreed @m1kola, those would both be great enhancements to the e2e tests.. both testing all supported API versions, as well as some way to flag certain tests as only applying to specific versions.

Are you able to take a look for that historical card? Happy to get it pulled over to JIRA if you can find it. Otherwise we can get a new fresh one in JIRA cut.

test/e2e/disk_encryption.go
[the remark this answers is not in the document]
bennerv: I think it's fair to use a hard coded string in this case to eliminate "test toil".  If every time we bump the API being used here and forget to update the e2e test with the correct client's constant for Enabled, the test is going to fail, we're going to context switch into looking at why the test failed, realize it's because we didn't update the client, and then wait again for it to complete.

This will still catch if a change in the string is detected.  Should eliminate some churn.

cblecker:
Going to merge this as is, but TLs are tracking the follow ups.

PR #2697, opened 2023-02-08
Add e2e test to confirm systemData

cblecker:
Fixes

Adds a test step to validate that systemData is returned from the user API, and use `DisallowUnknownFields()` on the admin API to verify that we are *not* returning unknown fields, namely systemData shouldn't be returned.


cblecker:
This won't pass tests until #2695 and #2696 are merged.

bennerv (approved):

PR #2699, opened 2023-02-09
Update the default fluentbit image to 1.9.9

hawkowl:

Part of https://issues.redhat.com/browse/ARO-1384 so we're not jumping too far

Updates to the fluentbit 1.9.9 that we're using on the RP/

### Test plan for issue:

This was tested on a prod cluster and appears to work.

N/A


bennerv (approved):
lgtm

PR #2702, opened 2023-02-09
Move to latest github.com/go-logr/logr

mbarnes:

Turned out to be a prerequisite for [[ARO-1919] AAD: Migrate ARO-RP Azure AD Graph -> Microsoft Graph](https://issues.redhat.com/browse/ARO-1919)

ARO-RP is currently stuck on an old version of [github.com/go-logr/logr](https://pkg.go.dev/github.com/go-logr/logr) (v0.4.0).

Version 1.0.0 introduced breaking API changes, but thereafter promises to be stable.  Both Kubernetes and Microsoft Graph modules use this logging API, so we can't stay on the old version forever or we'll eventually encounter irreconcilable dependency conflicts.

This pull request brings us up to the latest version (currently v1.2.3), and adapts the `logrWrapper` interface in [pkg/util/log/logr.go](https://github.com/Azure/ARO-RP/blob/master/pkg/util/log/logr.go).  The `logrWrapper` interface passes log messages through to a different logging module, [github.com/sirupsen/logrus](https://pkg.go.dev/github.com/sirupsen/logrus), and is used only for the [ARO operators](https://github.com/Azure/ARO-RP/blob/1f86566b927624c12953e18b033116337e8f4635/cmd/aro/operator.go#L60).

### Test plan for issue:

Ran existing unit tests.

No, this is an internal change only.


ellis-johnson (changes requested):
I pulled down this branch and ran a go mod tidy and there were a few extra additions to the go.sum if you want to quickly just add them

mbarnes:
Thanks for taking a look.  Ran `go mod tidy` and added the `go.sum` changes to the 1st commit.

ellis-johnson (dismissed):

mbarnes:
I'm not sure why **ci-go / vendor-check** is failing now.  Can't reproduce it locally.

SudoBrendan (approved):
lgtm

petrkotas (approved):
To sum up, we need this wrapper to enable using logrus in an operator. Operator requires it, because it is using the controller-runtime, which uses logr to delegate choice of logging library to the user of the library. Meaning this is an important part of the code. 

Perfect, thank you!

cadenmarchese (approved):

PR #2703, opened 2023-02-09
Spike - Cluster Operator Errors

tsatam:

Fixes https://issues.redhat.com/browse/ARO-1628

This PR changes the ARO Cluster Operator Status controller to derive the Cluster Operator's Available/Progressing/Degraded conditions from the state of all Controller status conditions present on the ARO cluster resource. 

It also implements controller status conditions on the node operator controller in order to demonstrate this. 

### Test plan for issue:

- Unit tests were added for the utility functions used to update controller conditions. The existing unit tests on the node controller were updated to validate expected conditions. 
- E2E tests were added to ensure the existing behavior of the cluster operator's conditions is preserved. There were no existing E2E tests for the node controller. 

TBD

### Notes

Vendored dependency upgrades were done in isolated commits for easier review.


pkg/util/conditions/controllerconditions.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	return nil
+}
+
+func getCluster(ctx context.Context, c client.Client) (*arov1alpha1.Cluster, error) {
+	cluster := &arov1alpha1.Cluster{}
+	err := c.Get(ctx, types.NamespacedName{Name: arov1alpha1.SingletonClusterName}, cluster)
+	if err != nil {
+		return nil, err
+	} else {
```
jaitaiwan, line 85: Unnecessary else

jaitaiwan (commented):
A couple super minor comments, but I reviewed and LGTM!

SudoBrendan (commented):
This is looking pretty good! A few nit-picks, and a question. I'm also wondering if we can hijack an E2E on node cordoning/draining to test the E2E on percolating of controller status up to the operator level, is that possible to add a quick check for?

pkg/operator/controllers/clusteroperatoraro/clusteroperatoraro_controller.go
```diff
@@ -74,44 +77,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	})
 
-	for _, c := range conditions {
-		v1helpers.SetStatusCondition(&co.Status.Conditions, c)
+	if equality.Semantic.DeepEqual(clusterOperatorObj.Status.Conditions, originalClusterOperatorObj.Status.Conditions) {
+		return nil
 	}
 
-	return r.client.Status().Update(ctx, co)
+	return r.client.Status().Update(ctx, clusterOperatorObj)
```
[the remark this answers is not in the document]
SudoBrendan: Fair enough - we save a call to the k8s api. LGTM.

pkg/operator/controllers/node/node_controller_test.go
```diff
@@ -54,6 +79,15 @@ func TestReconciler(t *testing.T) {
 			},
 			featureFlag: true,
 			wantErr:     `nodes "nonexistent-node" not found`,
+			wantConditions: []operatorv1.OperatorCondition{defaultAvailable, defaultProgressing,
+				{
+					Type: ControllerName + "Controller" + operatorv1.OperatorStatusTypeDegraded,
+					// Status:             operatorv1.ConditionTrue,
```
s-amann, line 85: should this be uncommented, along with all the `cnds.Degraded.Status = operatorv1.ConditionTrue` in the error flows in `node_controller.go`?

s-amann (changes requested):
please verify if we should uncomment the degraded status code, or remove the commented out code.

pkg/operator/controllers/node/node_controller_test.go
tsatam: Good catch. I think, as a part of this spike, we have determined that we should _not_ be propagating Degraded statuses to the ARO Cluster Operator, as that would impact cluster upgradeability. 

It is already the case that any controller Degraded=True statuses will _not_ propagate to the CO in this PR change (https://github.com/Azure/ARO-RP/blob/d5453e724abe3ddd8e5c3597fa8c2eb4ffd924dc/pkg/operator/controllers/clusteroperatoraro/clusteroperatoraro_controller.go#L89). 

As a result, we have two options:
- Do not set any controller Degraded=True statuses at all, and delete all of the error flows in `node_controller.go` that do so. 
- Do set these statuses (uncommenting the code), leaving those statuses and errors available to view on the ARO Cluster resource.

s-amann: ok, if our goal is to not prevent upgrades when our own operator is degraded, then I agree with removing the commented out code.

test/e2e/operator.go
```diff
@@ -228,6 +233,24 @@ var _ = Describe("ARO Operator - Conditions", func() {
...
+			g.Expect(err).NotTo(HaveOccurred())
+
+			g.Expect(cov1Helpers.IsStatusConditionTrue(co.Status.Conditions, configv1.OperatorAvailable))
+			g.Expect(cov1Helpers.IsStatusConditionFalse(co.Status.Conditions, configv1.OperatorProgressing))
+			g.Expect(cov1Helpers.IsStatusConditionFalse(co.Status.Conditions, configv1.OperatorDegraded))
+		}).WithContext(ctx).WithTimeout(timeout).Should(Succeed())
+
+		By("restoring the original cluster status")
+		clients.AROClusters.AroV1alpha1().Clusters().UpdateStatus(ctx, originalCluster, metav1.UpdateOptions{})
+	})
```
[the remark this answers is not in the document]
tsatam: This was scaffolding put in place to handle E2E tests that directly mutated the cluster resource's status and expected statuses to propagate to the operator status conditions. I ultimately ended up not going that route, so I've removed these scaffolds and retained the existing test to check the "normal" conditions on the CO are present.

pkg/operator/controllers/node/node_controller_test.go
tsatam: Discussing this with @SudoBrendan, we decided that we will set Controller degraded statuses and ensure they do not propagate to the cluster operator. I've enabled the Degraded=false handling on errors in the Node controller now.

pkg/operator/controllers/clusteroperatoraro/clusteroperatoraro_controller.go
```diff
@@ -74,44 +77,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	v1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, configv1.ClusterOperatorStatusCondition{
+		Type:               configv1.OperatorDegraded,
+		Status:             configv1.ConditionFalse,
+		LastTransitionTime: metav1.Now(),
+		Reason:             reasonAsExpected,
+	})
 
-	for _, c := range conditions {
-		v1helpers.SetStatusCondition(&co.Status.Conditions, c)
+	if equality.Semantic.DeepEqual(clusterOperatorObj.Status.Conditions, originalClusterOperatorObj.Status.Conditions) {
```
dem4gus, line 97: Won't this comparison always return false due to the LastTransitionTime being modified on line 93?

dem4gus (approved):
Just a small question but looks good overall

pkg/operator/controllers/clusteroperatoraro/clusteroperatoraro_controller.go
tsatam: The `v1helpers.SetStatusCondition` function from openshift/library-go will do its own diff on the "existing" and "new" status conditions, and will only update the tiemstamp to the "new" condition's timestamp timestamp if the actual status has changed:

https://github.com/openshift/library-go/blob/f3277c7720118c0b91bda69112f46a4928bf22b8/pkg/config/clusteroperator/v1helpers/status.go#L29

SudoBrendan (approved):
LGTM

jhoreman (approved):

s-amann (approved):

dem4gus (approved):
Changes look good. Is there a overview of this new pattern and how it's supposed to help anywhere?

SudoBrendan:
> Changes look good. Is there a overview of this new pattern and how it's supposed to help anywhere?

The Jira Epic explains a bit on why we're doing this, basically, so that we don't have to parse logs of every controller to see overall status of things. We'll be grooming additional work to add more controllers to fit this pattern.

PR #2704, opened 2023-02-09
Update user admin kubeconfig generation to happen every 3 months minimum

bennerv:

Creates consistency between when managed ingress and api certificates are rotated vs. the admin kubeconfig rotation which happens.  

The ingress & api certificates are configured to regenerate new certs with a 1 year validity period 275 days prior to expiration.  This does the same for the admin kubeconfig, so if we go >3 months without AdminUpdating/PUCMing clusters, customers should still have access to an admin kubeconfig that's valid.  

### Test plan for issue:

I'm fairly certain there isn't anything special here, but going to test a rotation and make sure it doesn't reboot nodes.

### DIFF
--- a/pkg/cluster/kubeconfig.go
+++ b/pkg/cluster/kubeconfig.go
@@ -36,7 +36,7 @@ func (m *manager) generateAROSREKubeconfig(pg graph.PersistedGraph) (*kubeconfig
 }
 
 // checkUserAdminKubeconfigUpdated checks if the user kubeconfig is
-// present, has >90days until expiry, has the right settings
+// present, has >275days until expiry, has the right settings
 func (m *manager) checkUserAdminKubeconfigUpdated() bool {
 	if len(m.doc.OpenShiftCluster.Properties.UserAdminKubeconfig) == 0 {
 		// field empty, not updated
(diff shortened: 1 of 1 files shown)


gvanderpotte:
LGTM, thank you

facchettos (approved):

facchettos:
as a side question, why is that done in pucm? couldn't we do it straight from the rp/monitor or something like that? It would seem that reducing the scope of pucms could be beneficial

bennerv:
>as a side question, why is that done in pucm? couldn't we do it straight from the rp/monitor or something like that? It would seem that reducing the scope of pucms could be beneficial

The monitor's sole purpose is to emit metrics & alert on the state of the cluster, as well as keep the current state of the cluster document master, worker, and ingress profiles consistent with the cluster.  Hence it is not an appropriate place to push config or update to the cluster. 

The RP doesn't have a mechanism currently to perform these tasks on a schedule, so we do them through PUCM.  It might make sense to add these tasks to regular update so if this happens again a customer can run `az aro update` and it will fix the issue, but I would rather push for automated PUCMs as a long-term solution.

bennerv:
I removed hold flag as I got around to testing this.  Looks good, generated a new cert + private key for kubeconfig data. I had to change the timeframe to 365 days to rotate the cert every PUCM, but it doesn't affect the cluster at all.  
Old Cert
```
$ openssl x509 -in cert.old -enddate 
notAfter=Feb 13 20:05:08 2024 GMT
```

New Cert
```
$ openssl x509 -in cert.new -enddate 
notAfter=Feb 13 20:32:07 2024 GMT
```

PR #2707, opened 2023-02-10
Removes dependency on a versioned kube client

m1kola:

Operator is not longer dependant on a versioned kube client and is now using a split client which uses cache for read operations.

### Test plan for issue:

No, just refactoring.

### DIFF
--- a/pkg/operator/controllers/monitoring/monitoring_controller.go
+++ b/pkg/operator/controllers/monitoring/monitoring_controller.go
@@ -13,8 +13,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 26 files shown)


pkg/operator/controllers/monitoring/monitoring_controller.go
```diff
@@ -99,22 +96,27 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	return reconcile.Result{}, nil
 }
 
-func (r *Reconciler) reconcilePVC(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
-	pvcList, err := r.kubernetescli.CoreV1().PersistentVolumeClaims(monitoringName.Namespace).List(ctx, metav1.ListOptions{LabelSelector: prometheusLabels})
+func (r *Reconciler) reconcilePVC(ctx context.Context) (ctrl.Result, error) {
+	pvcList := &corev1.PersistentVolumeClaimList{}
+	selector, _ := labels.Parse(prometheusLabels)
```
s-amann, line 101: I'm ok with ignoring this error since the labels are hardcoded and the current formatting should not result in an error.

pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
```diff
@@ -12,38 +12,37 @@ import (
...
-	newFakecli := func(s *corev1.Secret, c *corev1.Secret) *fake.Clientset {
-		c.ObjectMeta = metav1.ObjectMeta{
-			Name:      operator.SecretName,
-			Namespace: operator.Namespace,
-		}
-		c.Type = corev1.SecretTypeOpaque
-		if s == nil {
-			return fake.NewSimpleClientset(c)
-		}
+	// newFakecli := func(s *corev1.Secret, c *corev1.Secret) *fake.Clientset {
```
s-amann, line 27: can this commented out code be deleted?

s-amann (changes requested):
overall looks good, please delete the commented out code mentioned below.

pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
m1kola: Good catch! Updated

s-amann (approved):
lgtm, thanks!

PR #2709, opened 2023-02-10
Add checkaccess Client

nwnt:
Fixes https://issues.redhat.com/browse/ARO-1800

Migrating from ARM getAccess to CheckAccessV2

### Test plan for issue:

Client tested locally.  Changes to the RP will be sent in a separate PR.

https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/409184/CheckAccess-API-migration


pkg/util/azureclient/authz/remotepdp/client.go
```diff
@@ -0,0 +1,88 @@
...
+
+// remotePDPClient implements RemotePDPClient
+type remotePDPClient struct {
+	endpoint string
+	pipeline runtime.Pipeline
+}
+
+// NewRemotePDPClient returns an implementation of RemotePDPClient
+// endpoint - the fqdn of the regional specific endpoint of PDP
+// scope - the oauth scope required by the PDP serer
```
niontive, line 26: Nit: server

pkg/util/azureclient/authz/remotepdp/client.go
```diff
@@ -0,0 +1,88 @@
...
+// cred - the credential of the client to call the PDP server
+func NewRemotePDPClient(endpoint, scope string, cred azcore.TokenCredential) RemotePDPClient {
```
niontive, line 28: What are the different possible values for scope?

niontive: I see - differs between clouds

pkg/util/azureclient/environments.go
```diff
@@ -34,6 +42,10 @@ var (
 		AppLensScope:             "b9a1efcd-32ee-4330-834c-c04eb00f4b33",
 		AppLensTenantID:          "72f988bf-86f1-41af-91ab-2d7cd011db47",
 		AuthorityHost:            azidentity.AzurePublicCloud,
+		AzureRbacPDPEnvironment: AzureRbacPDPEnvironment{
```
weinong, line 45: has it been gofmt'ed?

pkg/util/azureclient/authz/remotepdp/client.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil, err
+	}
+	if res.StatusCode != http.StatusOK {
+		return nil, newCheckAccessError(res)
+	}
+
+	var accessDecision AuthorizationDecisionResponse
+	err = runtime.UnmarshalAsJSON(res, &accessDecision)
+
+	if err != nil {
```
weinong, line 68: nit:

if err := runtime.UnmarshalAsJSON(res, &accessDecision); err != nil {

weinong (commented):
LGTM, can you fix the link so that we can merge?

pkg/util/azureclient/environments.go
nwnt: yes it's been fmted.

pkg/util/azureclient/authz/remotepdp/client.go
nwnt: fixed :)

weinong (dismissed):
LGTM. Let's merge since it's been tested locally.

pkg/util/azureclient/authz/remotepdp/const.go
```diff
@@ -0,0 +1,17 @@
+package remotepdp
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+const (
+	modulename = "aro-pdpclient"
+	// version is the semantic version of this module
+	version = "0.0.1" //nolint
```
facchettos, line 9: why do we have nolint here?

pkg/util/azureclient/authz/remotepdp/client.go
```diff
@@ -0,0 +1,90 @@
...
+type remotePDPClient struct {
+	endpoint string
+	pipeline runtime.Pipeline
+}
+
+// NewRemotePDPClient returns an implementation of RemotePDPClient
+// endpoint - the fqdn of the regional specific endpoint of PDP
+// scope - the oauth scope required by the PDP server
+// cred - the credential of the client to call the PDP server
+func NewRemotePDPClient(endpoint, scope string, cred azcore.TokenCredential) RemotePDPClient {
```
facchettos, line 31: is there a specific reason why we are returning the interface type there? returning the actual type would make more sense otherwise and you'd lose no functionality, plus you'd be able to test that function without needing type casting. 

Plus having some testing on public functions is always nice

pkg/util/azureclient/authz/remotepdp/client.go
```diff
@@ -0,0 +1,90 @@
...
+		return err
+	}
+	var checkAccessError CheckAccessErrorResponse
+	err = json.Unmarshal(payload, &checkAccessError)
+	if err != nil {
+		return err
+	}
+	resErr.ErrorCode = checkAccessError.Message
+	return &resErr
+}
```
facchettos, line 90: this would be a better flow
```suggestion
func newCheckAccessError(r *http.Response) error {
        payload, err := runtime.Payload(r)
	if err != nil {
		return err
	}
	var checkAccessError CheckAccessErrorResponse
	err = json.Unmarshal(payload, &checkAccessError)
	if err != nil {
		return err
	}
	return &azcore.ResponseError{
		StatusCode:  r.StatusCode,
		RawResponse: r,
		ErrorCode: checkAccessError.Message
	}
}
```

facchettos (changes requested):

pkg/util/azureclient/environments.go
```diff
@@ -46,6 +58,10 @@ var (
 		AppLensScope:             "https://microsoft.onmicrosoft.com/runtimehost",
 		AppLensTenantID:          "cab8a31a-1906-4287-a0d8-4eef66b95f6e",
 		AuthorityHost:            azidentity.AzureGovernment,
+		AzureRbacPDPEnvironment: AzureRbacPDPEnvironment{
+			Endpoint:   "https://%s.authorization.azure.us/providers/Microsoft.Authorization/checkAccess?api-version=2021-06-01-preview",
```
s-amann, line 62: are we missing the population of `%s` in these two endpoints?

s-amann (commented):
some questions below

pkg/util/azureclient/authz/remotepdp/client.go
nwnt: *misclicked and accidentally deleted my previous comment...

I was trying to be consistent with how the same thing is done in the ARO-RP code base.  Maybe I've mistaken it for a pattern, but you can look at these functions as they return an interface type:
- refreshable.NewAuthorizer
- validate.NewOpenshiftClusterDynamicValidator
- instancemetadata.New
- dynamic.NewValidator

I don't mind making it return the concrete type though, just let me know what the right thing to do.

For testing, I couldn't find a good example of how to do this.  Do you have any suggestions?  I was thinking about using mocks but I'm not sure what tests should actually be performed on this data compositional logic.

pkg/util/azureclient/authz/remotepdp/const.go
nwnt: I got this from the applens code.  I can take it out and see what happens.

pkg/util/azureclient/environments.go
nwnt: nope, this is intentional. checkaccessv2 only has regional endpoint urls; there's no single global url for us to call.  So how I currently do it is using fmt.Sprintf(url, region) to get the full regional url when creating the client like below:
```go
		pdpClient = remotepdp.NewRemotePDPClient(
			fmt.Sprintf(aroEnv.AzureRbacPDPEnvironment.Endpoint, env.Location()),
			aroEnv.AzureRbacPDPEnvironment.OAuthScope,
			fpClientCred,
```

pkg/util/azureclient/authz/remotepdp/types.go
```diff
@@ -0,0 +1,101 @@
...
+type SubjectInfo struct {
+	Attributes SubjectAttributes `json:"Attributes"`
+}
+
+// SubjectAttributes contains the possible attributes to describe the subject
+// of query (i.e. if IT has the access). The ObjectId field is the UUID value of
+// the subject and is required.
+type SubjectAttributes struct {
+	ObjectId         string   `json:"ObjectId"`
+	Groups           []string `json:"Groups"`
```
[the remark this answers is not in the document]
nwnt: Good catch.  `omitempty` added to `Groups`

pkg/util/azureclient/authz/remotepdp/const.go
nwnt: Removed `nolint` in the latest commit.

pkg/util/azureclient/authz/remotepdp/client.go
facchettos: yeah I know we are using that pattern, but it never was a good pattern. the problem with this pattern is that you lose the type, and so you cannot access the fields anymore unless you do some type casting but this is very cumbersome and error prone. If you return the concrete type, it will be much easier to test. 

Are any of the fields relying on distant services or is everything local ? I'd just test that we are creating the objects with the proper fields

pkg/util/azureclient/environments.go
s-amann: ok gotcha, thank you. The missing piece was how this was going to be used.

pkg/util/azureclient/authz/remotepdp/client.go
nwnt: Everything is local until the client is used.  I have a created a test to call this function with the expected value.  Let me know if this is ok.

pkg/util/azureclient/authz/remotepdp/client_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+func TestClientCreate(t *testing.T) {
+	endpoint := "https://westus.authorization.azure.net/providers/Microsoft.Authorization/checkAccess?api-version=2021-06-01-preview"
+	scope := "https://authorization.azure.net/.default"
+	cred, err := azidentity.NewClientSecretCredential("888988bf-86f1-31ea-91cd-2d7cd011db48", "clientID", "clientSecret", nil)
+	if err != nil {
+		t.Error("Unable to create a new PDP client")
+	}
+	NewRemotePDPClient(endpoint, scope, cred)
+}
```
facchettos, line 26: there doesn't seem to be any check in here, we should test that the resulting object has the proper fields. here there is not assertion on it

pkg/util/azureclient/authz/remotepdp/types.go
```diff
@@ -0,0 +1,101 @@
...
+	ActionId       string `json:"actionId,omitempty"`
+	AccessDecision `json:"accessDecision,omitempty"`
+	IsDataAction   bool `json:"isDataAction,omitempty"`
+	RoleAssignment `json:"roleAssignment,omitempty"`
+	DenyAssignment RoleDefinition `json:"denyAssignment,omitempty"`
+	TimeToLiveInMs int            `json:"timeToLiveInMs,omitempty"`
+}
+
+// AccessDecision can be: Allowed, NotAllowed, Denied.
+type AccessDecision string
```
facchettos, line 75: this should probably be in the file where all the consts are defined. it's a little confusing that the consts aren't close to the type definition

pkg/util/azureclient/authz/remotepdp/client.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+)
+
+var _ RemotePDPClient = &remotePDPClient{}
+
+// RemotePDPClient represents the Microsoft Remote PDP API Spec
+type RemotePDPClient interface {
+	CheckAccess(context.Context, AuthorizationRequest) (*AuthorizationDecisionResponse, error)
+}
```
facchettos, line 21: can you add a little comment explaining why we have that line ? (16) (it provides compile time assertion that our type is implementing the interface, but it's not necessarily obvious to every one)

facchettos (changes requested):
some minor changes and then I think we should be good to go

pkg/util/azureclient/authz/remotepdp/client_test.go
nwnt: added one just now.

pkg/util/azureclient/authz/remotepdp/types.go
nwnt: Moved accordingly.

pkg/util/azureclient/authz/remotepdp/client.go
nwnt: Added as suggested.

facchettos (approved):

pkg/util/azureclient/authz/remotepdp/client_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+
+	testhttp "github.com/Azure/ARO-RP/test/util/http"
+)
+
+func TestClientCreate(t *testing.T) {
+	endpoint := "https://westus.authorization.azure.net/providers/Microsoft.Authorization/checkAccess?api-version=2021-06-01-preview"
+	scope := "https://authorization.azure.net/.default"
+	cred, err := azidentity.NewClientSecretCredential("888988bf-86f1-31ea-91cd-2d7cd011db48", "clientID", "clientSecret", nil)
```
cblecker, line 22: Is this dummy data?

nwnt: Yep, dummy data.

cblecker (approved):

PR #2713, opened 2023-02-13
Add newer gen instance types, remove deprecated and non-s types

cadenmarchese:

Fixes https://issues.redhat.com/browse/ARO-1344

Adds new generation instance types, also adds some customer-requested types from CEE. Adds some that were previously only supported on worker to be supported on master also. Cleans up some formatting and adds recently added SKUs to admin API.

Also, we want to remove support for certain deprecated sizes that don't support premium disk.

### Test plan for issue:
- [x] Create a test cluster with each new family according to https://github.com/Azure/ARO-RP/blob/master/docs/adding-new-instance-types.md
  - [x] Standard_D8as_v5
  - [x] Standard_D8s_v4
  - [x] Standard_D8s_v5
  - [x] Standard_E8s_v4
  - [x] Standard_E32s_v5
  - [x] Standard_E8as_v4
  - [x] Standard_E8as_v5
- [x] Confirm billing for each new instance type

Yes, we will need to submit a docs change to `azure-docs-pr` once we have released this.

### DIFF
--- a/pkg/api/admin/openshiftcluster.go
+++ b/pkg/api/admin/openshiftcluster.go
@@ -176,29 +176,87 @@ type VMSize string
 
 // VMSize constants.
 const (
-	VMSizeStandardD2sV3 VMSize = "Standard_D2s_v3"
-
-	VMSizeStandardD4asV4  VMSize = "Standard_D4as_v4"
-	VMSizeStandardD8asV4  VMSize = "Standard_D8as_v4"
-	VMSizeStandardD16asV4 VMSize = "Standard_D16as_v4"
-	VMSizeStandardD32asV4 VMSize = "Standard_D32as_v4"
-
+	VMSizeStandardD2sV3  VMSize = "Standard_D2s_v3"
 	VMSizeStandardD4sV3  VMSize = "Standard_D4s_v3"
 	VMSizeStandardD8sV3  VMSize = "Standard_D8s_v3"
 	VMSizeStandardD16sV3 VMSize = "Standard_D16s_v3"
 	VMSizeStandardD32sV3 VMSize = "Standard_D32s_v3"
 
-	VMSizeStandardE4sV3     VMSize = "Standard_E4s_v3"
-	VMSizeStandardE8sV3     VMSize = "Standard_E8s_v3"
-	VMSizeStandardE16sV3    VMSize = "Standard_E16s_v3"
-	VMSizeStandardE32sV3    VMSize = "Standard_E32s_v3"
+	VMSizeStandardD4sV4  VMSize = "Standard_D4s_v4"
+	VMSizeStandardD8sV4  VMSize = "Standard_D8s_v4"
+	VMSizeStandardD16sV4 VMSize = "Standard_D16s_v4"
+	VMSizeStandardD32sV4 VMSize = "Standard_D32s_v4"
+	VMSizeStandardD64sV4 VMSize = "Standard_D64s_v4"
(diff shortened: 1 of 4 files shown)


cadenmarchese:
This is tested and ready for review, but we should hold on merging/releasing until we confirm billing is set up for new instance types (tracked in https://issues.redhat.com/browse/ARO-1344 on the Red Hat side, work being completed by Microsoft).

pkg/api/admin/openshiftcluster.go
```diff
@@ -179,23 +179,87 @@ const (
...
 	VMSizeStandardD16sV3 VMSize = "Standard_D16s_v3"
+	VMSizeStandardD16sV4 VMSize = "Standard_D16s_v4"
+	VMSizeStandardD16sV5 VMSize = "Standard_D16s_v5"
 	VMSizeStandardD32sV3 VMSize = "Standard_D32s_v3"
+	VMSizeStandardD32sV4 VMSize = "Standard_D32s_v4"
+	VMSizeStandardD32sV5 VMSize = "Standard_D32s_v5"
+	VMSizeStandardD64sV5 VMSize = "Standard_D64s_v5"
+	VMSizeStandardD96sV5 VMSize = "Standard_D96s_v5"
+
+	VMSizeStandardE8V4  VMSize = "Standard_E8_v4"
```
dem4gus, line 205: The VM SKUs from here to line 219 do not support Premium disk types [[ref](https://learn.microsoft.com/en-us/azure/virtual-machines/vm-naming-conventions#naming-convention-explanation)]. Have we verified that these create with standard disks correctly? The [function I added](https://github.com/Azure/ARO-RP/blob/902ae3f3b79ab39dfc60b96c31346c6c8bddd892/pkg/util/computeskus/computeskus.go#L87-L93) *should* be using the capabilities reported by the SKU itself but it would be nice to get confirmation.

dem4gus (approved):
Small question about VM details, but the changes themselves look good.

pkg/api/admin/openshiftcluster.go
cadenmarchese: @dem4gus I need to test a few more VM families still (including this one) - I'll update the PR when I have confirmation that things are working well.

cadenmarchese: @dem4gus as of now, instances that don't support PremiumIO are not installing correctly if selected for master due to this code: https://github.com/Azure/ARO-RP/blob/master/pkg/installer/deployresources_resources.go#L123-L125 && https://github.com/Azure/ARO-RP/blob/master/pkg/installer/deployresources_resources.go#L194-L196 - so even though we were setting this correctly in install config, the master and bootstrap machines were still hardcoded to use Premium.

OCP doesn't technically support non-Premium instance types anyway, so I have removed them in this PR.

ellis-johnson (commented):
How do we know that these machines all function on ARO. Have they all been tested?

SudoBrendan (commented):
Nothing looks out of place here. As others have mentioned, getting confirmation for each type is the key to merging this. Maybe adding a checklist in this PR for all types tested would help clarify for reviewers?

cadenmarchese:
@ellis-johnson @SudoBrendan I have added the testing checklist to PR description as requested. The idea is to test one instance from each family added and confirm basic functionality. I have the general process documented here: https://github.com/Azure/ARO-RP/blob/master/docs/adding-new-instance-types.md

s-amann (approved):

PR #2715, opened 2023-02-14
Fix Admin Portal region select data population and add Qatar region

ellis-johnson:

Fixes https://issues.redhat.com/browse/ARO-2279

Currently the AZURE_ENVIRONMENT validation that is performed to return the regions is incorrect, so it needs to be fixed. Also Qatar needs to be added as a region

### Test plan for issue:

Unit Testing

Yes, ARO Wiki


pkg/portal/regions.go
```diff
@@ -72,7 +74,8 @@ func (p *portal) regions(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if value, found := os.LookupEnv("AZURE_ENVIRONMENT"); found {
-		if value == azureclient.PublicCloud.Environment.Name {
+		// AZURE_ENVIRONMENT variable can either be AZUREPUBLICCLOUD or AZUREUSGOVERNMENTCLOUD
+		if value == strings.ToUpper(azureclient.PublicCloud.Environment.Name) {
```
darthhexx, line 78: We should use `strings.EqualFold()` here, instead of relying on upping the case.

pkg/portal/regions_test.go
```diff
@@ -32,7 +33,7 @@ func TestRegionListPublic(t *testing.T) {
 		dbOpenShiftClusters: dbOpenShiftClusters,
 	}
 
-	os.Setenv("AZURE_ENVIRONMENT", azureclient.PublicCloud.Environment.Name)
+	os.Setenv("AZURE_ENVIRONMENT", strings.ToUpper(azureclient.PublicCloud.Environment.Name))
```
darthhexx, line 36: You can remove this, if you use `strings.EqualFold()` in the comparison above.

pkg/portal/regions_test.go
```diff
@@ -239,7 +244,7 @@ func TestRegionListFF(t *testing.T) {
 		dbOpenShiftClusters: dbOpenShiftClusters,
 	}
 
-	os.Setenv("AZURE_ENVIRONMENT", azureclient.USGovernmentCloud.Environment.Name)
+	os.Setenv("AZURE_ENVIRONMENT", strings.ToUpper(azureclient.USGovernmentCloud.Environment.Name))
```
darthhexx, line 247: And this one.

darthhexx (changes requested):

pkg/portal/regions.go
ellis-johnson: Fixed!

pkg/portal/regions_test.go
ellis-johnson: Removed

pkg/portal/regions_test.go
ellis-johnson: Removed

s-amann (approved):

darthhexx (approved):

PR #2716, opened 2023-02-14
Add operatorhub images to mirror

ellis-johnson:

Fixes https://issues.redhat.com/browse/ARO-2164 

We need to include the Red Hat operatorhub image indexes in our acr so that clusters that do not have outbound internet access can still use them

### Test plan for issue:

Unit Testing

N/A


bennerv:
We don't have imagecontentsourcepolicies on these images to pull from our ACR.  Is there a reason we need these images to be able to be pulled?

bennerv:
>We don't have imagecontentsourcepolicies on these images to pull from our ACR. Is there a reason we need these images to be able to be pulled?

Also something to consider would be samples images pulling successfully.  I'm looking at epic and will reach out to Jerome.

bennerv:
@ellis-johnson I added some comments to the epic, but we're working with Jerome and Oren to refine the scope a bit on the epic and it will go through some doc review next week.  For now we can pause on this.

PR #2717, opened 2023-02-14
Removes RetryOnConflict from operator controllers

m1kola:

Conflict indicates that we acted on a stale date. Abort reconciliation and requeue.


s-amann (approved):

cadenmarchese (approved):
Thanks!

PR #2718, opened 2023-02-14
systemreserved: Fixes `Remove`, eleminates dependency on dynamichelper & adds more test coverage

m1kola:

Changes:
* Removes dynamichelper. Using dynamichelper to create just one object looks like overkill here. The same can be done with the client.
* Updates `Remove`: we previously were removing the label from `MachineConfigPool`, but keeping `KubeletConfig` handing around. Now we remove both
* Extra unit test coverage

### Test plan for issue:

Existing and new unit tests.

No, refactoring

### DIFF
--- a/pkg/operator/controllers/workaround/systemreserved_test.go
+++ b/pkg/operator/controllers/workaround/systemreserved_test.go
@@ -10,37 +10,99 @@ import (
 
 	"github.com/golang/mock/gomock"
 	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	ctrlfake "sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	"github.com/Azure/ARO-RP/pkg/util/cmp"
 	utillog "github.com/Azure/ARO-RP/pkg/util/log"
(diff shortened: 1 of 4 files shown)


s-amann (approved):
lgtm, good to see the additional tests, thank you!

pkg/operator/controllers/workaround/systemreserved_test.go
```diff
@@ -10,37 +10,99 @@ import (
...
 
 	"github.com/Azure/ARO-RP/pkg/util/cmp"
 	utillog "github.com/Azure/ARO-RP/pkg/util/log"
-	mock_dynamichelper "github.com/Azure/ARO-RP/pkg/util/mocks/dynamichelper"
+	_ "github.com/Azure/ARO-RP/pkg/util/scheme"
 )
 
 func TestSystemreservedEnsure(t *testing.T) {
+	kubeletConfig := func(resourceVersion string) *mcv1.KubeletConfig {
+		return &mcv1.KubeletConfig{
```
cblecker, line 26: Does this output match what this object currently looks like on clusters exactly? Would want to make sure the resultant object doesn't change as that would cause a node reboot to apply it

m1kola: @cblecker there previously was `TestKubeletConfig` which was checking the output of `kubeletConfig()` method. This definition is copy-pasted from `TestKubeletConfig`. So despite the fact that we are now constructing the `KubeletConfig`diferently in the controller - it is still has the same contents.

However there was a change which I mentioned in the PR description:

> Updates `Remove`: we previously were removing the label from `MachineConfigPool`, but keeping `KubeletConfig` handing around. Now we remove both

There is a chance that we have clusters where `aro.autosizednodes.enabled` is set to `"true"`[1] which makes this controller to call `Remove` method during reconciliation. For these clusters new version of the controller will remove leftover `KubeletConfig`.


[1] Default is `"true"`

petrkotas (approved):
Good work mikalai!

PR #2719, opened 2023-02-14
Operator standardise dynamichelper

m1kola:

Previously `dynamichelper` was sometimes passed into other constructors, sometimes created inside the controllers. This PR standardises on passing it into the constructors.

### DIFF
--- a/pkg/operator/controllers/genevalogging/genevalogging_controller.go
+++ b/pkg/operator/controllers/genevalogging/genevalogging_controller.go
@@ -11,7 +11,6 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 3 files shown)


pkg/operator/controllers/genevalogging/genevalogging_controller.go
```diff
@@ -69,21 +63,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 	r.log.Debug("running")
-	mysec, err := r.kubernetescli.CoreV1().Secrets(operator.Namespace).Get(ctx, operator.SecretName, metav1.GetOptions{})
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-
-	// TODO: dh should be a field in r, but the fact that it is initialised here
-	// each time currently saves us in the case that the controller runs before
-	// the SCC API is registered.
-	dh, err := dynamichelper.New(r.log, r.restConfig)
```
m1kola, line 80: I think this should not be a problem - if there is an error due to API not being registered or something - we just return an error and requeue with backoff.

s-amann (approved):

PR #2721, opened 2023-02-16
Conditionally add the AKS VM cluster subnet to the CosmoDB firewall allow list in GovCloud

darthhexx:

Fixes: 

The OpenShift Versions Pipeline requires PUT access to the CosmoDB to sync the OpenShift versions per region. Now that we have added a VM deploy ACI fallback option, we need to conditionally add the firewall rule to CosmoDB in GovCloud.

Once we have ACI support, we will be able to remove this rule from the code and the next RP release would remove it from CosmoDB.

### Test plan for issue:

Tested by deploying to INT and testing the [OpenShift Versions pipeline deploys successfully](https://msazure.visualstudio.com/AzureRedHatOpenShift/_releaseProgress?_a=release-pipeline-progress&releaseId=2949).

No.


UlrichSchlueter:
LGTM

UlrichSchlueter (approved):
Approved

PR #2722, opened 2023-02-16
Run ensureResourceGroup for ARO delete

SrinivasAtmakuri:
Slackthread: https://redhat-internal.slack.com/archives/C02ULBRS68M/p1676340311290279

Current Scenario:
azcli delete only ensures if ARO SP has NETWORK_CONTRIBUTOR_ROLE on the VNET and adds it if missing, if OWNER Permission is missing on ARO SP for CRG, the deletion fails and only AdminUpdate can fix such missing permission which will be tricky if the cluster is in ProvisioningFailed state.

With this PR:
We fix any missing permission on ARO SP so that deletion works without failure.

### Which issue this PR address:

Fixes

### DIFF
--- a/pkg/cluster/delete.go
+++ b/pkg/cluster/delete.go
@@ -398,8 +398,14 @@ func (m *manager) deleteResourcesAndResourceGroup(ctx context.Context) error {
 }
 
 func (m *manager) Delete(ctx context.Context) error {
+	m.log.Printf("running ensureResourceGroup")
+	err := m.ensureResourceGroup(ctx) // re-create RP RBAC if needed/missing on best-effort basics
+	if err != nil {
+		m.log.Error(err)
+	}
+
 	m.log.Printf("deleting dns")
-	err := m.dns.Delete(ctx, m.doc.OpenShiftCluster)
+	err = m.dns.Delete(ctx, m.doc.OpenShiftCluster)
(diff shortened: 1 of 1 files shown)


s-amann (approved):

pkg/cluster/delete.go
```diff
@@ -398,8 +398,14 @@ func (m *manager) deleteResourcesAndResourceGroup(ctx context.Context) error {
 }
 
 func (m *manager) Delete(ctx context.Context) error {
+	m.log.Printf("running ensureResourceGroup")
+	err := m.ensureResourceGroup(ctx) // re-create RP RBAC if needed/missing
+	if err != nil {
+		return err
+	}
```
bennerv, line 405: As @gvanderpotte said, I think we should just log out this error.  It's best-effort and could have consequences in environments where the resource group already exists.

SrinivasAtmakuri: @bennerv @gvanderpotte changed to log the error message from returning it.

gvanderpotte (approved):
LGTM

s-amann (approved):

petrkotas (approved):

rolandmkunkel (approved):

bennerv (approved):

PR #2723, opened 2023-02-16
policy: deny pod with master toleration taints

ArrisLee:
https://issues.redhat.com/browse/ARO-2222

Prevents non-privileged namespaces from having pods that have masters toleration taints.
non-privileged namespaces can match (openshift-logging|openshift-operators)

Test plan for issue:
Rego unit test + Gator test

May need to update the design doc for guardrail


yjst2012 (approved):
LGTM

PR #2726, opened 2023-02-16
Validate master resize target is a supported master vmSize

cadenmarchese:

Fixes

Resizing masters to unsupported/untested VM sizes could have unintended negative consequences for a cluster. This adds validating logic to the master resize action so that SRE doesn't have to go and check the list manually.

### Test plan for issue:

Unit tests

No

### DIFF
--- a/pkg/api/validate/vm.go
+++ b/pkg/api/validate/vm.go
@@ -7,7 +7,7 @@ import (
 	"github.com/Azure/ARO-RP/pkg/api"
 )
 
-var supportedMasterVmSizes = map[api.VMSize]api.VMSizeStruct{
+var SupportedMasterVmSizes = map[api.VMSize]api.VMSizeStruct{
 	// General purpose
 	api.VMSizeStandardD8sV3:  api.VMSizeStandardD8sV3Struct,
 	api.VMSizeStandardD16sV3: api.VMSizeStandardD16sV3Struct,
@@ -98,7 +98,7 @@ func DiskSizeIsValid(sizeGB int) bool {
 
 func VMSizeIsValid(vmSize api.VMSize, requiredD2sV3Workers, isMaster bool) bool {
 	if isMaster {
-		_, supportedAsMaster := supportedMasterVmSizes[vmSize]
+		_, supportedAsMaster := SupportedMasterVmSizes[vmSize]
 		return supportedAsMaster
 	}
 
@@ -124,7 +124,7 @@ func VMSizeFromName(vmSize api.VMSize) (api.VMSizeStruct, bool) {
 		return size, true
 	}
 
-	if size, ok := supportedMasterVmSizes[vmSize]; ok {
+	if size, ok := SupportedMasterVmSizes[vmSize]; ok {
 		return size, true
 	}
 	return api.VMSizeStruct{}, false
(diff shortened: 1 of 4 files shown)


petrkotas (approved):
LGTM

pkg/frontend/validate.go
```diff
@@ -189,11 +189,19 @@ func validateNetworkInterfaceName(nicName string) error {
...
-func validateAdminVMSize(vmSize string) error {
+func validateAdminMasterVMSize(vmSize string) error {
 	if vmSize == "" {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", "The provided vmSize '%s' is invalid.", vmSize)
 	}
-	return nil
+
+	// check to ensure that the size is supported as a master size
+	for k := range validate.SupportedMasterVmSizes {
+		if string(k) == vmSize {
```
s-amann, line 199: since `vmSize` is coming off the URL I believe it will be [lowercased](https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/middleware/lowercase.go), should this be `strings.EqualFold()` instead?

pkg/api/validate/vm.go
```diff
@@ -98,7 +98,7 @@ func DiskSizeIsValid(sizeGB int) bool {
 
 func VMSizeIsValid(vmSize api.VMSize, requiredD2sV3Workers, isMaster bool) bool {
```
tony-schndr, line 99: I'm curious, why did you choose to expose `SupportedMasterVmSizes` instead of just using this function?

cadenmarchese: I'm going back and forth on this. `VMSizeIsValid` is case sensitive and checks some other things we don't care about here, so it might be better to just look directly at the list rather than make VMSizeIsValid work in this context. But I also think it would be nice to just use the single function for both, we'd have to tweak it though. Thoughts @petrkotas @s-amann?

pkg/frontend/validate.go
cadenmarchese: Good catch, I've changed the comparison and added tests

pkg/api/validate/vm.go
tony-schndr: This seems to work without breaking unit tests, which is equivalent to what you have in the frontend
```
func VMSizeIsValid(vmSize api.VMSize, requiredD2sV3Workers, isMaster bool) bool {
	if isMaster {
		for k := range SupportedMasterVmSizes {
			if strings.EqualFold(string(k), string(vmSize)) {
				return true
			}
		}
	}
```

I do agree there is a tradeoff here, as this function is used in each API Version.

pkg/frontend/validate.go
```diff
@@ -189,11 +189,19 @@ func validateNetworkInterfaceName(nicName string) error {
 	return nil
 }
 
-func validateAdminVMSize(vmSize string) error {
+func validateAdminMasterVMSize(vmSize string) error {
 	if vmSize == "" {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", "The provided vmSize '%s' is invalid.", vmSize)
```
s-amann, line 194: nit: we can delete this since it will fail in the condition on line 199

s-amann (approved):
lgtm, a nit for the future would be consolidating the errors returned by `validateAdminMasterVMSize`

s-amann (approved):

PR #2728, opened 2023-02-17
Use VirtualNetworkRule structs and not strings for the CosmoDB VNET rules

darthhexx:

Fixes: The root cause behind the deploy failures introduced in https://github.com/Azure/ARO-RP/pull/2721 (for realz this time).

Deploying the RP is borked without it.

### Test plan for issue:

Manually tested the changes in a dedicated RP, although it's not directly possible because a dedicated RP instance is a sortof dev not pure production.

No.


UlrichSchlueter (approved):
lgtm

PR #2729, opened 2023-02-17
Remove unnecessary checking of SDN for install config generation

bennerv:

in `pkg/frontend/openshiftcluster_putorpatch.go`, we set defaults `api.SetDefaults` which always ensures that the SDN is set on cluster creation or cluster update.  This if statement is no longer necessary as the SDN will always be set on a cluster during installconfig generation.

### DIFF
--- a/pkg/installer/generateconfig.go
+++ b/pkg/installer/generateconfig.go
@@ -104,10 +104,7 @@ func (m *manager) generateInstallConfig(ctx context.Context) (*installconfig.Ins
 		masterZones = []string{""}
 	}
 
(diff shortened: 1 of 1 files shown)


bennerv:
SetDefaults happens in frontend -> https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/openshiftcluster_putorpatch.go#L208-L209

Defaults sets the SDN if it isn't already set -> https://github.com/Azure/ARO-RP/blob/master/pkg/api/defaults.go#L13-L15

This code is called much later: https://github.com/Azure/ARO-RP/blob/master/pkg/cluster/install.go#L201

SudoBrendan (approved):
lgtm

PR #2730, opened 2023-02-17
Support OVNKubernetes in OpenShift 4.11

bennerv:

Fixes ARO-1476

The aro-installer will now set SDN if it's not set.  It will change based on which aro-installer is invoked.  4.11 will be OVNK8s and 4.10 will be OpenShiftSDN.  Relevant PR in aro-installer: https://github.com/openshift/ARO-Installer/pull/16/files

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -9,11 +9,6 @@ package api
 // when moving between old and new versions
 func SetDefaults(doc *OpenShiftClusterDocument) {
 	if doc.OpenShiftCluster != nil {
-		// SoftwareDefinedNetwork was introduced in 2021-09-01-preview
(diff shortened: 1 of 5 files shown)


bennerv:
Couple things this PR does:
1. Revert a previous PR because the assumption that SDN is set is no longer true.
2. Remove SDN out of defaults.  This will be set by the ARO-Installer or the integrated installer if not using hive
3. Pull the SDN out of the install graph after installation
4. Add a comment to update SDN if we upgrade the integrated installer to 4.11
5.

bennerv:
All clusters that are in a valid state (not failed creating or failed deleting) have SDN set so I removed /hold.

s-amann (approved):

PR #2732, opened 2023-02-20
Remove SDN in Python as field no longer used in v2022-09-04 or later 

bennerv:

Removes an unused field that doesn't take any effect in cluster creation using `az aro` command module.  

### Test plan for issue:

@tony-schndr tested an `az aro create --version 4.11.x` and even though this field was present, it still defaulted to OVNKubernetes as it should.  

Nope

### DIFF
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -7,7 +7,6 @@ from azext_aro._validators import validate_cluster_resource_group
 from azext_aro._validators import validate_disk_encryption_set
 from azext_aro._validators import validate_domain
 from azext_aro._validators import validate_pull_secret
-from azext_aro._validators import validate_sdn
 from azext_aro._validators import validate_subnet
 from azext_aro._validators import validate_client_secret
 from azext_aro._validators import validate_visibility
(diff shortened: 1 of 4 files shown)


python/az/aro/azext_aro/_params.py
```diff
@@ -64,10 +63,6 @@ def load_arguments(self, _):
         c.argument('service_cidr',
                    help='CIDR of service network. Must be a minimum of /18 or larger.',
                    validator=validate_cidr('service_cidr'))
-        c.argument('software_defined_network', arg_type=get_enum_type(['OVNKubernetes', 'OpenShiftSDN']),
```
s-amann, line 67: is there a deprecation process for this type of change? or are we free to rip it out as we want?

s-amann (approved):

python/az/aro/azext_aro/_params.py
bennerv: it's not in the API that the aro command module uses anymore so I think we're just good to rip it out.  It's not even honored if you pass it anyway right now so we should remove it to remove confusion as well.

cadenmarchese (approved):
LGTM

PR #2735, opened 2023-02-21
Fixing External LoadBalancer Backend Address Pool after Master node replacement

mhsreddy:
https://issues.redhat.com/browse/ARO-2321

This PR fixes the issue when the master node in the cluster is replaced, the newly replaced master's network interface is not automatically being added the External Load Balancer Address Pool in the old architecture version(V1) clusters. Also in the clusters with newer architecture(V2) when the network interface is removed manually from the External Load Balancer for any reason then its not automatically being  added to the External  Load Balancer Backend Address Pool again. This issue can be fixed when we run PUCM/Admin Update after the master node is replaced in the cluster.

### Test plan for issue:
Test manually by replacing the master nodes in test clusters in both new and old version architectures. Also add unit test cases.

### DIFF
--- a/pkg/cluster/fixssh.go
+++ b/pkg/cluster/fixssh.go
@@ -44,6 +44,7 @@ func (m *manager) fixSSH(ctx context.Context) error {
 		m.log.Warnf("Failed checking and Updating Network Interface with error: %s", err)
 		return err
 	}
+
 	return nil
 }
 
@@ -100,7 +101,28 @@ func (m *manager) checkandUpdateNIC(ctx context.Context, resourceGroup string, i
 			}
 		}
 
-		if m.updateNIC(ctx, &nic, nicName, &lb, i, infraID) {
+		if m.updateILBAddressPool(ctx, &nic, nicName, &lb, i, infraID) {
+			m.log.Printf("updating Network Interface %s", nicName)
+			err = m.interfaces.CreateOrUpdateAndWait(ctx, resourceGroup, nicName, nic)
+			if err != nil {
+				return err
+			}
+		}
+
+		var elbName string
+		switch m.doc.OpenShiftCluster.Properties.ArchitectureVersion {
+		case api.ArchitectureVersionV1:
+			elbName = infraID + "-public-lb"
+		case api.ArchitectureVersionV2:
+			elbName = infraID
+		}
+
+		elb, err := m.loadBalancers.Get(ctx, resourceGroup, elbName, "")
+		if err != nil {
+			return err
+		}
+
+		if m.updateELBAddressPool(ctx, &nic, nicName, &elb, i, infraID) {
 			m.log.Printf("updating Network Interface %s", nicName)
 			err = m.interfaces.CreateOrUpdateAndWait(ctx, resourceGroup, nicName, nic)
 			if err != nil {
@@ -108,6 +130,7 @@ func (m *manager) checkandUpdateNIC(ctx context.Context, resourceGroup string, i
 			}
 		}
 	}
+
 	return nil
 }
 
@@ -118,19 +141,20 @@ func (m *manager) removeBackendPoolsFromNIC(ctx context.Context, resourceGroup, 
 	ipc := (*nic.InterfacePropertiesFormat.IPConfigurations)[0]
 	if ipc.LoadBalancerBackendAddressPools != nil {
 		m.log.Printf("Removing Load balancer Backend Address Pools from NIC %s with no VMs attached", nicName)
-		*(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools = []mgmtnetwork.BackendAddressPool{}
+		*ipc.LoadBalancerBackendAddressPools = []mgmtnetwork.BackendAddressPool{}
 		return m.interfaces.CreateOrUpdateAndWait(ctx, resourceGroup, nicName, *nic)
 	}
 	return nil
 }
 
-func (m *manager) updateNIC(ctx context.Context, nic *mgmtnetwork.Interface, nicName string, lb *mgmtnetwork.LoadBalancer, i int, infraID string) bool {
+func (m *manager) updateILBAddressPool(ctx context.Context, nic *mgmtnetwork.Interface, nicName string, lb *mgmtnetwork.LoadBalancer, i int, infraID string) bool {
 	if nic.InterfacePropertiesFormat.IPConfigurations == nil || len(*nic.InterfacePropertiesFormat.IPConfigurations) == 0 {
 		m.log.Warnf("unable to update NIC as there are no IP configurations for %s", nicName)
 		return false
 	}
 
 	var ilbBackendPool string
+
 	switch m.doc.OpenShiftCluster.Properties.ArchitectureVersion {
 	case api.ArchitectureVersionV1:
 		ilbBackendPool = infraID + "-internal-controlplane-v4"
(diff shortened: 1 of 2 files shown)


pkg/cluster/fixssh.go
```diff
@@ -100,14 +101,50 @@ func (m *manager) checkandUpdateNIC(ctx context.Context, resourceGroup string, i
...
+
+		var elbName string
+		switch m.doc.OpenShiftCluster.Properties.ArchitectureVersion {
+		case api.ArchitectureVersionV1:
+			elbName = infraID + "-public-lb"
+		case api.ArchitectureVersionV2:
+			elbName = infraID
+		}
+
+		if m.doc.OpenShiftCluster.Properties.ArchitectureVersion == api.ArchitectureVersionV1 {
```
ellis-johnson, line 120: Could this if statement be moved inside the archv1 switch statement?

ellis-johnson (dismissed):
One small change but LGTM!

pkg/cluster/fixssh.go
mhsreddy: As discussed with you the **if** the statement moved inside the switch statement it makes the code less readable and adds little confusion even though the functionality does not does not change.

s-amann: an alternative would be to encapsulate the logic into a helper function, which it looks like you partially did. if the retrieval of the load balancers on line 121 were moved into the `updateV1ELBAddressPool` function then it would be fine to move this call into the switch. 

Also worth noting, is if `updateV1ELBAddressPool` returns an error instead of a `bool` the code that is actually doing the update can be moved into it making the switch on line 115 more in line with what @ellis-johnson suggested

pkg/cluster/fixssh.go
```diff
@@ -100,14 +101,50 @@ func (m *manager) checkandUpdateNIC(ctx context.Context, resourceGroup string, i
...
+				}
+			}
+		}
+
+		elb, err := m.loadBalancers.Get(ctx, resourceGroup, elbName, "")
+		if err != nil {
+			return err
+		}
+
+		if m.updateELBAddressPool(ctx, &nic, nicName, &elb, i, infraID) {
```
s-amann, line 139: should this be executed for v1 and v2 architecture?

s-amann: should this be renamed to updateV2ElBAddressPool? or does it indeed need to be executed for both v1 and v2?

pkg/cluster/fixssh.go
```diff
@@ -124,52 +161,117 @@ func (m *manager) removeBackendPoolsFromNIC(ctx context.Context, resourceGroup,
...
+		if strings.EqualFold(*p.ID, elbBackendPoolID) {
+			updateV1ELBPool = true
+			index = i
+		}
+	}
+
+	if updateV1ELBPool {
+		m.log.Printf("Removing NIC %s from Public Load Balancer API Address Pool %s", nicName, elbBackendPoolID)
+		lbpool = append(lbpool[:index], lbpool[index+1:]...)
+		(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools = &lbpool
```
s-amann, line 237: lets refactor this to avoid side effects. we can move
```
m.log.Printf("Updating Network Interface %s", nicName)
		err = m.interfaces.CreateOrUpdateAndWait(ctx, resourceGroup, nicName, *nic)
		if err != nil {
			return err
		}
```
into this condition and then we won't be relying on side effects to call update.

pkg/cluster/fixssh.go
```diff
@@ -124,52 +161,117 @@ func (m *manager) removeBackendPoolsFromNIC(ctx context.Context, resourceGroup,
...
 		*(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools = append(*(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools, mgmtnetwork.BackendAddressPool{
-			ID: &ilbID,
+			ID: &ilbBackendPoolID,
 		})
 	}
 
 	return updateSSHPool || updateILBPool
 }
 
+func (m *manager) updateV1ELBAddressPool(ctx context.Context, nic *mgmtnetwork.Interface, nicName string, lb *mgmtnetwork.LoadBalancer, i int, infraID string) bool {
```
s-amann, line 217: `i` is not used here, `lb` does not need to be passed in, if we change `updateV1ELBAddressPool` to accept a `resourceGroup string` and to return a `error` instead of `bool` the first lines of the method can retrieve the `lb` and we can clean up the switch to just call `updateV1ELBAddressPool` and return its possible error.

s-amann (changes requested):
this is shaping up but I still think there is room to reduce the complexity of the switch + condition, as well as reduce some side effects.

pkg/cluster/fixssh.go
mhsreddy: `updateELBAddressPool` function needs to be executed for both V1 and V2.

pkg/cluster/fixssh.go
mhsreddy: Refactored the code as per your suggestion

pkg/cluster/fixssh.go
```diff
@@ -124,52 +153,122 @@ func (m *manager) removeBackendPoolsFromNIC(ctx context.Context, resourceGroup,
...
+	lbpool := *(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools
+	var index int
+	for i, p := range lbpool {
+		if strings.EqualFold(*p.ID, elbBackendPoolID) {
+			updateV1ELBPool = true
+			index = i
+		}
+	}
+
+	if updateV1ELBPool {
```
s-amann, line 226: IMO we should not use 'tracking booleans' and do the work when we need to / the soonest we know which nic needs to be removed. I think we can also avoid having to track the index and avoid possible index out of range errors if we create a new slice and only insert the NICs we want to keep.

I think something like this would work, what do you think?

```golang
	if nic.InterfacePropertiesFormat.IPConfigurations == nil || len(*nic.InterfacePropertiesFormat.IPConfigurations) == 0 {
		return fmt.Errorf("unable to update NIC as there are no IP configurations for %s", nicName)
	}

	lb, err := m.loadBalancers.Get(ctx, resourceGroup, infraID, "")
	if err != nil {
		return err
	}
	elbBackendPoolID := fmt.Sprintf("%s/backendAddressPools/%s", *lb.ID, infraID)
	currentPool := *(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools
	newPool := make([]mgmtnetwork.BackendAddressPool, len(currentPool))
	for _, pool := range currentPool {
		if strings.EqualFold(*pool.ID, elbBackendPoolID) {
			m.log.Printf("Removing NIC %s from Public Load Balancer API Address Pool %s", nicName, elbBackendPoolID)
		} else {
			newPool = append(newPool, pool)
		}
	}

	(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools = &newPool
	m.log.Printf("Updating Network Interface %s", nicName)
	return m.interfaces.CreateOrUpdateAndWait(ctx, resourceGroup, nicName, *nic)
```

pkg/cluster/fixssh.go
```diff
@@ -124,52 +153,122 @@ func (m *manager) removeBackendPoolsFromNIC(ctx context.Context, resourceGroup,
...
+	elbBackendPoolID := fmt.Sprintf("%s/backendAddressPools/%s", *lb.ID, elbBackendPool)
+
+	updateELBPool := true
+	for _, p := range *(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools {
+		if strings.EqualFold(*p.ID, elbBackendPoolID) {
+			updateELBPool = false
+		}
+	}
+
+	if updateELBPool {
```
s-amann, line 262: can we make this similar to the v1 version where we do the work once we determine we need to instead of using 'tracking booleans' and pushing the logic to multiple places?

s-amann (changes requested):
I have a few more suggestions to clean up this code, keep up the good work!

pkg/cluster/fixssh.go
```diff
@@ -124,52 +153,122 @@ func (m *manager) removeBackendPoolsFromNIC(ctx context.Context, resourceGroup,
...
+		return false
+	}
+
+	var elbBackendPool string
+	switch m.doc.OpenShiftCluster.Properties.ArchitectureVersion {
+	case api.ArchitectureVersionV1:
+		elbBackendPool = infraID + "-public-lb-control-plane-v4"
+	case api.ArchitectureVersionV2:
+		elbBackendPool = infraID
+	}
```
jaitaiwan, line 251: Because the switch case has only two options at this point and one of the options is just to use the infraID verbatim, you could simplify it into a single if statement to see if its version 1 and add the extra info.

jaitaiwan (changes requested):
LGTM bar a really minor nit pick.

pkg/cluster/fixssh.go
jaitaiwan:
```
elbBackendPool := infraID
if m.doc.OpenShiftCluster.Properties.ArchitectureVersion == api.ArchitectureVersion {
  elbBackendPool = elbBackendPool + "-public-lb-control-plane-v4"
}
```

pkg/cluster/fixssh.go
```diff
@@ -124,52 +153,122 @@ func (m *manager) removeBackendPoolsFromNIC(ctx context.Context, resourceGroup,
...
 	return updateSSHPool || updateILBPool
 }
 
+func (m *manager) updateV1ELBAddressPool(ctx context.Context, nic *mgmtnetwork.Interface, nicName string, resourceGroup string, infraID string) error {
+	if nic.InterfacePropertiesFormat.IPConfigurations == nil || len(*nic.InterfacePropertiesFormat.IPConfigurations) == 0 {
+		return fmt.Errorf("unable to update NIC as there are no IP configurations for %s", nicName)
+	}
+
+	var updateV1ELBPool bool
+	lb, err := m.loadBalancers.Get(ctx, resourceGroup, infraID, "")
```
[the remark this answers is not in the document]
mhsreddy: Updated code to check for error and return error when it occurs.

pkg/cluster/fixssh.go
mhsreddy: Made changes as per the suggestions.

pkg/cluster/fixssh.go
mhsreddy:
```	elbBackendPool := infraID
	if m.doc.OpenShiftCluster.Properties.ArchitectureVersion == api.ArchitectureVersionV1 {
		elbBackendPool = elbBackendPool + "-public-lb-control-plane-v4"
	}

pkg/cluster/fixssh.go
mhsreddy: Change the code similar to V1 version

jaitaiwan (approved):

jaitaiwan:
LGTM!

s-amann (approved):
lgtm, thank you for the changes.

PR #2736, opened 2023-02-21
Add 4.11 latest version to update-version

petrkotas:

Fixes

Adds 4.11 version to the aro update-versions.

### DIFF
--- a/pkg/util/version/const.go
+++ b/pkg/util/version/const.go
@@ -31,6 +31,13 @@ var InstallStream = &Stream{
 	PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:b9fad814fb4442e7e852b0614d9bb4e2ebc5e1a2fa51623aa838b4ee0e4a5369",
 }
 
+var AdditionalStreams = map[string]*Stream{
+	"4.11.26": {
+		Version:  NewVersion(4, 11, 26),
+		PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:1c3913a65b0a10b4a0650f54e545fe928360a94767acea64c0bd10faa52c945a",
+	},
+}
+
 // UpgradeStreams describes list of streams we support for upgrades
 var (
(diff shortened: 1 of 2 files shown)


pkg/util/version/const.go
```diff
@@ -31,6 +31,13 @@ var InstallStream = &Stream{
 	PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:b9fad814fb4442e7e852b0614d9bb4e2ebc5e1a2fa51623aa838b4ee0e4a5369",
 }
 
+var AdditionalStreams = map[string]*Stream{
+	"4.11.26": {
+		Version:  NewVersion(4, 11, 26),
+		PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:1c3913a65b0a10b4a0650f54e545fe928360a94767acea64c0bd10faa52c945a"
```
s-amann, line 37: please add a comma 
```suggestion
		PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:1c3913a65b0a10b4a0650f54e545fe928360a94767acea64c0bd10faa52c945a",
```

s-amann (changes requested):
there's a need for a comma below

pkg/util/version/const.go
petrkotas: AH I missed it. Thank you!

s-amann (approved):

pkg/util/version/const.go
```diff
@@ -31,6 +31,13 @@ var InstallStream = &Stream{
 	PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:b9fad814fb4442e7e852b0614d9bb4e2ebc5e1a2fa51623aa838b4ee0e4a5369",
 }
 
+var AdditionalStreams = map[string]*Stream{
```
bennerv, line 34: Why use a map?  Why not a slice?

PR #2738, opened 2023-02-21
Create cluster keyvault

dem4gus:

[ARO-1513](https://issues.redhat.com/browse/ARO-1513)

Creates an Azure Keyvault in the Cluster Resource Group to store ACR tokens (and possibly other secrets, but that is outside the scope of this story. See also [MSFT recommended best practices](https://learn.microsoft.com/en-us/azure/key-vault/general/best-practices#use-separate-key-vaults)). The keyvault is created during the Bootstrap phase of cluster installation and uses the AAD tenant of the cluster's service principal.

### Test plan for issue:

Tests will be added before PR leaves draft status.

Preliminary documentation will be added before PR leaves draft status. Changes to the ARO support policy advising against altering the keyvault's access policies may be necessary.

### DIFF
--- a/pkg/cluster/keyvaults.go
+++ b/pkg/cluster/keyvaults.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/Azure/ARO-RP/pkg/util/stringutils"
+	"github.com/Azure/ARO-RP/pkg/util/uuid"
 	"github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2019-09-01/keyvault"
 	"github.com/Azure/go-autorest/autorest/to"
 )
@@ -13,7 +14,6 @@ import (
 // Licensed under the Apache License 2.0.
 
 func (m *manager) createKeyvault(ctx context.Context) error {
-
 	// We want random characters from end of infraID to help ensure KV name uniqueness
 	infraID := m.doc.OpenShiftCluster.Properties.InfraID
 	suffixMaxLength := 20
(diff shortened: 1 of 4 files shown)


pkg/cluster/keyvaults.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+
+	"github.com/Azure/ARO-RP/pkg/util/stringutils"
+	"github.com/Azure/ARO-RP/pkg/util/uuid"
+	"github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2019-09-01/keyvault"
+	"github.com/Azure/go-autorest/autorest/to"
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
```
facchettos, line 14: it would look a bit nicer right after the package name

pkg/cluster/keyvaults.go
```diff
@@ -0,0 +1,77 @@
...
+	infraID := m.doc.OpenShiftCluster.Properties.InfraID
+	suffixMaxLength := 20
+	var suffix string
+
+	// This runs after ensureInfraID bootstrap step, but just in case
+	if infraID == "" {
+		suffix = m.doc.OpenShiftCluster.Name
+	} else {
+		suffix = infraID
+	}
```
facchettos, line 27:
```suggestion
        suffix := infraID
        
	// This runs after ensureInfraID bootstrap step, but just in case
	if infraID == "" {
		suffix = m.doc.OpenShiftCluster.Name
	} 
```

pkg/cluster/keyvaults.go
```diff
@@ -0,0 +1,77 @@
...
+
+	suffixLength := len(suffix)
+	if suffixLength > suffixMaxLength {
+		suffix = suffix[suffixLength-suffixMaxLength : suffixLength]
```
facchettos, line 31: what are we trying to achieve here? couldn't we just truncate the suffix like this? 
```go
suffix = suffix[:suffixMaxLength]
```

(also you can write `suffix = suffix[suffixLength-suffixMaxLength:]`  which would be equivalent to what you have right now, it simplifies it)

facchettos (changes requested):
a few suggestions

pkg/cluster/keyvaults.go
dem4gus: The comment from line 17 still applies here, we want the _end_ of the `InfraID` that has the randomly-generated characters because KV names need to be globally unique. I could just grab the 5 characters at the end that we randomly generate, there are 36^5 > 60 million possible combinations, so the likelihood of collision is small, but including as much of the infraID with the cluster name as possible reduces that likelihood even further.

PR #2739, opened 2023-02-21
fix e2e tests when running against 4.11 clusters

tony-schndr:

Fixes e2e tests when running against 4.11 clusters

default storage classes changed in 4.11, from managed-premium -> managed-csi.  This change checks the cluster version from the cluster document and sets the default storage class.

### Test plan for issue:

Ran e2e tests against 4.11 cluster.

No

### DIFF
--- a/test/e2e/cluster.go
+++ b/test/e2e/cluster.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/Azure/ARO-RP/pkg/util/ready"
+	"github.com/Azure/ARO-RP/pkg/util/version"
 	"github.com/Azure/ARO-RP/test/util/project"
 )
 
@@ -51,6 +52,7 @@ var _ = Describe("Cluster", func() {
 	})
 
 	It("can run a stateful set which is using Azure Disk storage", func(ctx context.Context) {
+
 		By("creating stateful set")
 		err := createStatefulSet(ctx, clients.Kubernetes)
 		Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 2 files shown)


test/e2e/cluster.go
```diff
@@ -161,7 +170,7 @@ func createStatefulSet(ctx context.Context, cli kubernetes.Interface) error {
 						AccessModes: []corev1.PersistentVolumeAccessMode{
 							corev1.ReadWriteOnce,
 						},
-						StorageClassName: to.StringPtr("managed-premium"),
+						StorageClassName: to.StringPtr(defaultStorageClass),
```
SudoBrendan, line 173: how about we just remove this entirely. Is it critically important that we hard-code the default Storage Class name, or can we just let k8s to k8s things and use the default SC if none is specified? I feel like this E2E is more a test of "can the cluster create default storage?" vs "can the cluster create storage of a particular named type?" - I feel like in a multi-version world, saying "use the default" is the easiest path that tests what we are actually after.

tony-schndr: I was meaning to test this, thanks for the reminder.

test/e2e/cluster.go
```diff
@@ -114,6 +116,13 @@ var _ = Describe("Cluster", func() {
 })
 
 func createStatefulSet(ctx context.Context, cli kubernetes.Interface) error {
+	oc, _ := clients.OpenshiftClusters.Get(ctx, vnetResourceGroup, clusterName)
+	installVersion, _ := version.ParseVersion(*oc.ClusterProfile.Version)
+
+	defaultStorageClass := "managed-premium"
+	if installVersion.V[0] == 4 && installVersion.V[1] == 11 {
```
tony-schndr, line 123: I wasn't sure, but should I open up this conditional to match >= 4.11?  Not that it matters at this time, but when 4.12 comes to ARO we would have to update this.  I just haven't had the chance to check what the default storage classes are for 4.12 to confirm.

UlrichSchlueter (approved):

PR #2740, opened 2023-02-22
clean up of portal

facchettos:
some rework of portal before removing gorilla

this centralizes the route definitions in the portal.go file instead of having some extra route definitions added to the routers from other packages. 

Fixes

### DIFF
--- a/pkg/portal/portal.go
+++ b/pkg/portal/portal.go
@@ -44,16 +44,13 @@ type Runnable interface {
 }
 
 type portal struct {
-	env                 env.Core
-	audit               *logrus.Entry
-	log                 *logrus.Entry
-	baseAccessLog       *logrus.Entry
-	l                   net.Listener
-	sshl                net.Listener
-	verifier            oidc.Verifier
-	baseRouter          *mux.Router
-	authenticatedRouter *mux.Router
-	publicRouter        *mux.Router
+	env           env.Core
+	audit         *logrus.Entry
+	log           *logrus.Entry
+	baseAccessLog *logrus.Entry
+	l             net.Listener
+	sshl          net.Listener
+	verifier      oidc.Verifier
 
 	hostname     string
 	servingKey   *rsa.PrivateKey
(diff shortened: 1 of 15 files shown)


dem4gus (dismissed):
Looks good, I'm a big fan of consolidating the route registrations in one place.

pkg/portal/portal.go
```diff
@@ -256,13 +244,23 @@ func (p *portal) Run(ctx context.Context) error {
 	return s.Serve(tls.NewListener(p.l, config))
 }
 
+func (p *portal) bearerRoutes(r *mux.Router, k *kubeconfig.Kubeconfig) {
```
yehornaumenko, line 247: it seems that there is no need to have bearerRoutes func as a method of `*portal` as it does not take any variables from `p *portal` and does not mutate it

pkg/portal/portal.go
```diff
@@ -132,43 +129,40 @@ func NewPortal(env env.Core,
...
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	p.templateV2, err = template.New("index.html").Parse(string(assetv2))
 	if err != nil {
-		return err
+		return nil, err
 	}
```
yehornaumenko, line 154: Have you considered moving this logic to another small function?

facchettos: yes, but I want to stay as close as possible to the structure that we had before, we can use smaller functions later.
here I want the scope of the PRs to be as small as possible because it is a very complex piece of code and it instrumental to our operations

yehornaumenko: oh, got you, really makes sense!

jaitaiwan:
I have a couple of suggestions if I may?
 - It would be good if the functions that are now being exported as a result of this PR be documented so we're able to see at a glance what functions they perform in their contexts; I'm not sure if this is going against your mandate of making this a small PR with as little changes as possible tho.
 - I would like to see slightly more specific commit messages if at all possible e.g. "Centralise route definitions for portal" would be more meaningful to me than "clean up of portal".
 
 Apart from these couple things, looks great!

facchettos:
@jaitaiwan I think that's fair. 
Also, you may always give suggestions or request changes on the PRs, worst case scenario is that your wrong and you learn something new :D , and people should be able to justify what they are doing

yehornaumenko (approved):
Great job!

s-amann (approved):

PR #2742, opened 2023-02-22
Refactor hive install streams

bennerv:

Refactors hive install streams used in `update_ocp_versions`.  We're currently using a map but there is no need.  Additionally, we shouldn't be hardcoding minor version strings for installer pull spec in case default changes for some reason.  

### Test plan for issue:

Run the pipeline against INT.


s-amann (approved):
lgtm, `DefaultInstallStream` is now clear what it is for and where we rely on defaults.

PR #2743, opened 2023-02-22
Reconcile differences between frontend & backend default version behavior

bennerv:

Fixes logical differences between frontend & backend handling the default version

We need to support the default version no matter what.  This is because if a customer does not specify a version, we set the default version.  So we cannot disable it as it would provide a poor experience.  

### Test plan for issue:

Create a default version cluster when a record for it does not exist in cosmosdb.  

Probably, but don't think we have existing documentation on this.

### DIFF
--- a/pkg/cluster/version.go
+++ b/pkg/cluster/version.go
@@ -28,6 +28,30 @@ type openShiftClusterDocumentVersionerService struct{}
 func (service *openShiftClusterDocumentVersionerService) Get(ctx context.Context, doc *api.OpenShiftClusterDocument, dbOpenShiftVersions database.OpenShiftVersions, env env.Interface, installViaHive bool) (*api.OpenShiftVersion, error) {
 	requestedInstallVersion := doc.OpenShiftCluster.Properties.ClusterProfile.Version
 
+	// Honor any installer pull spec override
+	installerPullSpec := env.LiveConfig().DefaultInstallerPullSpecOverride(ctx)
+	if installerPullSpec == "" {
+		installerPullSpec = fmt.Sprintf("%s/aro-installer:release-%s", env.ACRDomain(), version.DefaultInstallStream.Version.MinorVersion())
+	}
+
+	// add the default OCP version as we require it as an install target
+	// if this is removed, we need to also update the logic in
+	// pkg/frontend/frontend.go, pkg/frontend/validate.go
+	defaultVersion := &api.OpenShiftVersion{
+		Properties: api.OpenShiftVersionProperties{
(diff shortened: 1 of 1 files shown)


s-amann (dismissed):

pkg/cluster/version.go
```diff
@@ -40,34 +58,6 @@ func (service *openShiftClusterDocumentVersionerService) Get(ctx context.Context
...
 
-	errUnsupportedVersion := api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "properties.clusterProfile.version", "The requested OpenShift version '%s' is not supported.", requestedInstallVersion)
-
-	// when we have no OpenShiftVersion entries in CosmoDB, default to building one using the InstallStream
-	if len(activeOpenShiftVersions) == 0 {
-		if requestedInstallVersion != version.InstallStream.Version.String() {
-			return nil, errUnsupportedVersion
-		}
-
-		installerPullSpec := env.LiveConfig().DefaultInstallerPullSpecOverride(ctx)
```
darthhexx, line 51: Noting that in this change we're losing the "LiveConfig" capability to override the default installer pullspec. We aren't using this feature currently, so its not breaking anything, but worth noting that we may want to reintroduce this when LiveConfig is put live.

pkg/cluster/version.go
```diff
@@ -28,6 +28,24 @@ type openShiftClusterDocumentVersionerService struct{}
...
 	requestedInstallVersion := doc.OpenShiftCluster.Properties.ClusterProfile.Version
 
+	// add the default OCP version as we require it as an install target
+	// if this is removed, we need to also update the logic in
+	// pkg/frontend/frontend.go, pkg/frontend/validate.go
+	defaultVersion := &api.OpenShiftVersion{
+		Properties: api.OpenShiftVersionProperties{
+			Version:           version.InstallStream.Version.String(),
+			OpenShiftPullspec: version.InstallStream.PullSpec,
+			InstallerPullspec: fmt.Sprintf("%s/aro-installer:release-%s", env.ACRDomain(), version.InstallStream.Version.MinorVersion()),
```
darthhexx, line 38: Also wanted to highlight that by using this hard-coded InstallerPullspec, together with dropping the LiveConfig override, we are losing the ability to set a different value for the default.

bennerv: Fixed by adding the installer spec override at the top of the function.

petrkotas (approved):
Works for me. Thanks 

One thing that caught my eye is, that if we set the override wrong it is not validated. However, it is us who sets the override, so it is not a big issue.

PR #2744, opened 2023-02-23
Add new clouderror for inUseSubnet

carvalhe:

Fixes https://issues.redhat.com/browse/ARO-2298

Client was having an issue deleting a cluster and was getting internal server error. When digging into the logs, the issue was shown to be InUseSubnet, which is marked as an internalServerError with error code 400. This change will change the error to a userError and the user will be able to see more details surrounding the failure and be able to correct it.


cadenmarchese (approved):
Makes sense to me 👍

kimorris27 (approved):

bennerv:
this _shouldn't_ be necessary because what this error message means is that a customer created a vnet within the managed cluster resource group, but we have seen this multiple times, so I'm approving & merging as it's a better customer experience to output a message than to keep returning Internal Server Error

bennerv (approved):

PR #2746, opened 2023-02-23
Portal aad cleanup

facchettos:
depends on #2740 , blocked by our integration setup 

moves the route definition to portal where it is supposed to be. Also fixes the AAD middleware which uses its own way of storing the user info instead of getting it from the jwt we are getting from AAD

Fixes


pkg/portal/middleware/aad.go
```diff
@@ -113,149 +121,71 @@
...
+	http.SetCookie(w, &http.Cookie{
+		Name:  stateCookie,
+		Value: state,
 	})
-}
 
-// Login will redirect the user to a login page.
-func (a *aad) Login(w http.ResponseWriter, r *http.Request) {
-	a.redirect(w, r)
+	http.Redirect(w, r, a.oauther.AuthCodeURL(state), http.StatusTemporaryRedirect)
```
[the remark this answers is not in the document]
facchettos: state is generated with a uuid generator so it's not constant

pkg/portal/security_test.go
```diff
@@ -1,458 +0,0 @@
-package portal
```
facchettos, line 1: This test's only value was to ensure we are running 2 middlewares (Log and AAD). with the redesign and various fixes it is hard to mock it in the same way. Since we are moving to chi in the next days and that chi allows us to list the middlewares applied on routes, I removed it to avoid unnecessary work and will try to merge the PRs at the same time so we don't lose those tests

pkg/portal/middleware/aad.go
```diff
@@ -286,35 +221,63 @@ func (a *aad) callback(w http.ResponseWriter, r *http.Request) {
...
-		http.Error(w, http.StatusText(http.StatusForbidden), http.StatusForbidden)
+	err = token.Claims(&claims)
+	if err != nil {
+		return "", nil, err
 	}
 
-	session.Values[SessionKeyUsername] = claims.PreferredUsername
-	session.Values[SessionKeyGroups] = groupsIntersect
-	session.Values[SessionKeyExpires] = a.now().Add(a.sessionTimeout).Unix()
+	return claims.PreferredUsername, claims.Groups, nil
```
yehornaumenko, line 259: why not return the entire claim object?

yehornaumenko (changes requested):
Now it looks much better, good job

pkg/portal/middleware/aad.go
```diff
@@ -286,35 +221,63 @@ func (a *aad) callback(w http.ResponseWriter, r *http.Request) {
...
+		// token has expired
+		return false, "", nil, idToken.Expiry, err
+	}
+	username, groups, err = extractInfoFromToken(idToken)
+	if err != nil {
+		return false, username, groups, idToken.Expiry, err
+	}
+	groupsIntersect := GroupsIntersect(d.allowedGroups, groups)
+	if len(groupsIntersect) == 0 {
+		return true, username, groups, idToken.Expiry, nil
```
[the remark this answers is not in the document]
facchettos: if it's not allowed, we might want to log things, so we have a trace (e.g. in case of malicious attempts of pen testing). for the other you are right, I changed it

pkg/portal/middleware/aad.go
```diff
@@ -286,35 +221,63 @@ func (a *aad) callback(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
-		a.internalServerError(w, err)
-		return
+		return false, "", nil, time.Time{}, err
 	}
 
-	http.Redirect(w, r, "/", http.StatusTemporaryRedirect)
+	if idToken.Expiry.Before(now) {
+		// token has expired
+		return false, "", nil, idToken.Expiry, err
```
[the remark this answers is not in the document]
facchettos: I ended up copy-pasting a little bit too much, good catch :D

pkg/portal/middleware/aad.go
facchettos: it just avoids having to deal with a claim later on, because after that we really only need about the username and the groups, not the claim.

yehornaumenko (approved):

petrkotas (approved):
I think this is a good upgrade. We can merge this, when we are doing next release, because this is going to be fully tested in canary as we do not have AAD tenant setup and prepared in CI and INT.

petrkotas (approved):
Thanks @facchettos great work. Lets get it in and test in canary

pkg/portal/portal.go
```diff
@@ -160,22 +161,36 @@ func (p *portal) setupRouter(kconfig *kubeconfig.Kubeconfig, prom *prometheus.Pr
 	allGroups := append([]string{}, p.groupIDs...)
 	allGroups = append(allGroups, p.elevatedGroupIDs...)
 
-	p.aad, err = middleware.NewAAD(p.log, p.audit, p.env, p.baseAccessLog, p.hostname, p.sessionKey, p.clientID, p.clientKey, p.clientCerts, allGroups, unauthenticatedRouter, p.verifier)
+	// we get the env var from within the function because it is not meant to be configurable
+	disableOauthOption := os.Getenv("DISABLE_OAUTH")
+	if disableOauthOption == "true" || env.IsCI() || env.IsLocalDevelopmentMode() {
```
hawkowl, line 166: Since we will only ever test this in INT (since we don't/can't run the portal e2e in PROD), can we make it so that we need to give the flag AND be in local development mode/CI for this to work? So, I guess, `disableOauthOption == "true" && (env.IsCI() || env.IsLocalDevelopmentMode())`?

pkg/portal/portal.go
```diff
@@ -160,22 +161,36 @@ func (p *portal) setupRouter(kconfig *kubeconfig.Kubeconfig, prom *prometheus.Pr
...
+		p.baseAccessLog.Errorf("running in int, disableoauth=%s", disableOauthOption)
+		p.aad, err = middleware.NewIntAAD(p.elevatedGroupIDs, p.audit)
+	} else {
+		p.log.Error("not running in int")
```
hawkowl, line 170: I don't think we need to log this when we're starting in regular mode, especially as error. I would just remove it.

pkg/portal/ssh/ssh.go
```diff
@@ -132,7 +133,11 @@ func (s *SSH) New(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	elevated := len(middleware.GroupsIntersect(s.elevatedGroupIDs, ctx.Value(middleware.ContextKeyGroups).([]string))) > 0
+	groups, ok := ctx.Value(middleware.ContextKeyGroups).([]string)
+	if !ok {
+		s.internalServerError(w, errors.New("could not find any groups"))
```
hawkowl, line 138: We need a return here, I think?

test/e2e/setup.go
```diff
@@ -154,8 +152,9 @@ func SaveScreenshotAndExit(wd selenium.WebDriver, e error) {
 
 func adminPortalSessionSetup() (string, *selenium.WebDriver) {
 	const (
-		hubPort  = 4444
-		hostPort = 8444
+		hubPort        = 4444
+		hostPort       = 8444
+		KeyVaultPrefix = "KEYVAULT_PREFIX"
```
hawkowl, line 157: Is this const used?

pkg/deploy/devconfig.go
```diff
@@ -101,6 +101,7 @@ func DevConfig(_env env.Core) (*Config, error) {
 					KeyvaultDNSSuffix:           &_env.Environment().KeyVaultDNSSuffix,
 					KeyvaultPrefix:              &keyvaultPrefix,
 					StorageAccountDomain:        to.StringPtr(os.Getenv("USER") + "aro" + _env.Location() + ".blob." + _env.Environment().StorageEndpointSuffix),
+					DisableOauth:                to.StringPtr("true"),
```
hawkowl, line 104: Do we need to disable oauth for dev use? I think we have the AAD groups set up for it already, it's just CI that's a problem.

hawkowl (changes requested):
Minor comments, overall looks good. The only thing I'd want to confirm before merging is the disable flag checking logic and whether that missing return is actually needed, the rest are optional.

pkg/portal/portal.go
facchettos: pr e2e is actually running with local development mode, I had that condition before but it wasn't working. Something like `(disableOauthOption == "true" && env.IsCI()) || env.IsLocalDevelopmentMode()` should work though

pkg/portal/portal.go
facchettos: this is some leftover from debugging

test/e2e/setup.go
facchettos: yeah no, also some leftover from another try

pkg/deploy/devconfig.go
facchettos: ok will try that

PR #2749, opened 2023-02-27
Fixes an issue with RHUI certificate errors during deploy

darthhexx:

Fixes: RP and gateway deploys.

There is a known issue in the RHEL 8 VM image that requires this command to be run `yum update -y --disablerepo='*' --enablerepo='rhui-microsoft-azure*'` as highlighted in [the IcM Ticket](https://portal.microsofticm.com/imp/v3/incidents/details/370581805/home). This is currently causing deploy failures in eastus2euap, Public INT, and FairFax INT for RP deploys as well as transient-VM-deploys used fro ACI Fallback. The fix for the [transient VM deploys is in this related PR](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines/pullrequest/7684622).

### Test plan for issue:

Tested the fix using the VM deploy ACI fallback mechanism, which was failing for the same reasons, and [it fixed the issue](https://msazure.visualstudio.com/AzureRedHatOpenShift/_releaseProgress?releaseId=3029&_a=release-pipeline-progress).

Successfully tested a deploy to a dedicate RP instance as well.

No.


UlrichSchlueter (approved):
LGTM, let's see what the e2e test says

pkg/deploy/generator/scripts/devProxyVMSS.sh
```diff
@@ -1,3 +1,6 @@
+echo "running RHUI fix"
+yum update -y --disablerepo='*' --enablerepo='rhui-microsoft-azure*'
+
 yum -y -x WALinuxAgent -x WALinuxAgent-udev update
 yum -y install docker
```
facchettos, line 5: why do we install plain docker here instead of podman docker like the others?

facchettos (commented):
seems like there is a lot of code duplication in here, could we look into fixing that? (in a subsequent PR, if this merge is really urgent)

pkg/deploy/generator/scripts/devProxyVMSS.sh
petrkotas: Historically we found out, that podman is not 1to1 replacement and our deployment did not worked.

petrkotas (approved):
LGTM

facchettos:
but we should have the same stuff all over the place, this seems to be just a problem with code duplication, this is a recipe for weird bugs to appear here

pkg/deploy/generator/scripts/devProxyVMSS.sh
darthhexx: I only moved the RP and gateway over to using podman when doing the [RHEL 8 migration](https://github.com/Azure/ARO-RP/pull/2518), but I don't think we actually use this devProxy anymore either. I simply changed it to be consistent across all the scripts.

PR #2750, opened 2023-02-27
Deploy dynamic Service Keyvault ACLs to GovCloud

darthhexx:

Fixes: AKS monitoring stack access to the Service Keyvault.

Historically, there haven't been AKS instances in GovCloud and we've been skipping deploying Keyvault permissions that rely on an AKS instance for the kubelet identity. However, we now do have AKS deployed in FairFax INT, with FairFax Production to be deployed this week.

This PR removed the conditional deployment, so that the ACLs required for the monitoring stack work as expected.

### Test plan for issue:

Not required since this simply removes conditional so that the ARM deployment takes place in all clouds and regions.

Nope.

### DIFF
--- a/pkg/deploy/assets/rp-development-predeploy.json
+++ b/pkg/deploy/assets/rp-development-predeploy.json
@@ -319,7 +319,6 @@
             "dependsOn": [
                 "[concat(parameters('keyvaultPrefix'), '-svc')]"
             ],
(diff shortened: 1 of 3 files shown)


UlrichSchlueter (approved):

facchettos (approved):

petrkotas (approved):
LGTM

PR #2752, opened 2023-03-01
fix MHC failure on node count reaching supported maximum

rajdeepc2792:
Fixes https://issues.redhat.com/browse/ARO-1931

PR provides a fix by implementing a check on nodes count by setting up a watch on Node API in ARO's Machine Health Check Controller.
Existing MHC CRs are deleted if the node count reaches the support maximum value, i.e. 62. Otherwise recreated.

### Test plan for issue:

- [ ] Create a test cluster
- [ ] Change the `maxSupportedNodes` value from 62 to 6(current count of nodes) in [MHC Controller](https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go)
- [ ] Run the local operator
- [ ] Check MHC CRs existence

Unit Test Cases added for MHC Deletion Ensure if nodes count is equivalent to supported maximum.

N/A

### DIFF
--- a/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
@@ -35,10 +35,11 @@ var machinehealthcheckYaml []byte
 var mhcremediationalertYaml []byte
 
 const (
-	ControllerName    string = "MachineHealthCheck"
-	managed           string = "aro.machinehealthcheck.managed"
-	enabled           string = "aro.machinehealthcheck.enabled"
-	maxSupportedNodes int    = 62
+	ControllerName string = "MachineHealthCheck"
+	managed        string = "aro.machinehealthcheck.managed"
+	enabled        string = "aro.machinehealthcheck.enabled"
+	// Note: We're going to review this functionality or change this const once ARO supports more than 62 nodes out of the box.
+	maxSupportedNodes int = 62
 )
 
 type Reconciler struct {
--- a/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
@@ -12,14 +12,17 @@ import (
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
 	kruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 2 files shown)


rajdeepc2792:
@microsoft-github-policy-service agree [company="redhat"]

rajdeepc2792:
@microsoft-github-policy-service agree company="redhat"

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -66,7 +70,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	}
 
 	r.log.Debug("running")
-	if !instance.Spec.OperatorFlags.GetSimpleBoolean(managed) {
+	nodeList := &corev1.NodeList{}
+	err = r.client.List(ctx, nodeList)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	removeMHCCR := len(nodeList.Items) == maxSupportedNodes
```
cadenmarchese, line 78: This looks good to me, I think we could add one more unit test in `machinehealthcheck_controller_test.go` that confirms MHC resource is deleted at the desired node count.

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -32,9 +35,10 @@ var machinehealthcheckYaml []byte
 var mhcremediationalertYaml []byte
 
 const (
-	ControllerName string = "MachineHealthCheck"
-	managed        string = "aro.machinehealthcheck.managed"
-	enabled        string = "aro.machinehealthcheck.enabled"
+	ControllerName    string = "MachineHealthCheck"
+	managed           string = "aro.machinehealthcheck.managed"
+	enabled           string = "aro.machinehealthcheck.enabled"
+	maxSupportedNodes int    = 62
```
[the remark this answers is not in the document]
rajdeepc2792: Done. Comment Added.

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
rajdeepc2792: Done. Added unit test cases for the desired node count.

cblecker (changes requested):
Left comment on ARO-1931 as I don't believe this is the right fix

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
facchettos: could we have cases where we actually have more than 62 nodes? what I'm getting at is that to be completely sure we should maybe use `>=` instead of `==`

s-amann: +1 to use `>= 60` we could have 63 total, in the JIRA it notes 60 workers + 3 masters. In addition we should only list out worker nodes on line 75 since the aro-mhc will not take action on masters anyway.

rajdeepc2792: As per @cadenmarchese , the MHC failure is an issue only if the total nodes count reaches 62. If the node count is above that, it means some manual support is provided to handle more than 62 nodes and it's okay to keep the MHC CRs present in the cluster for that scenario.

For `>=60 and listing only worker nodes`, I have a concern that if any case a new initiated node is a non-worker and that increments the node count to 62, in that case too the MHC CRs should be deleted, am I right?

PR #2753, opened 2023-03-01
API OperatorFlagsMergeStrategy

SrinivasAtmakuri:
JIRA: https://issues.redhat.com/browse/ARO-1885

Fixes

### DIFF
--- a/pkg/api/admin/operatorflags.go
+++ b/pkg/api/admin/operatorflags.go
@@ -1,0 +1,68 @@
+package admin
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/operator"
+)
+
+type operatorFlagsMergeStrategyStruct struct {
+	OperatorFlagsMergeStrategy string
+	Cluster                    *api.OpenShiftCluster
+}
+
+const (
+	operatorFlagsMergeStrategyDefault string = "merge"
+	operatorFlagsMergeStrategyMerge   string = "merge"
+	operatorFlagsMergeStrategyReset   string = "reset"
+)
+
+// When a cluster is edited via the PATCH Cluster Geneva Action (aka an Admin Update)
+// the flags given are treated according to the provided Update Strategy,
+// provided in operatorFlagsMergeStrategy
+
+// merge (default): The provided cluster flags are laid on top of the cluster’s existing flags.
+// reset: The provided cluster flags are laid on top of the default cluster flags,
+// essentially ‘resetting’ the flags if no new flags are provided.
+func OperatorFlagsMergeStrategy(oc *api.OpenShiftCluster, body []byte) error {
+	payload := operatorFlagsMergeStrategyStruct{
+		OperatorFlagsMergeStrategy: operatorFlagsMergeStrategyDefault,
+		Cluster:                    &api.OpenShiftCluster{},
+	}
+
+	err := json.Unmarshal(body, &payload)
+	if err != nil {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidRequestContent, "", "The request content was invalid and could not be deserialized: %q.", err)
+	}
+
+	// return error if OperatorFlagsMergeStrategy is not merge or reset, default is merge
+	if payload.OperatorFlagsMergeStrategy != operatorFlagsMergeStrategyMerge &&
+		payload.OperatorFlagsMergeStrategy != operatorFlagsMergeStrategyReset {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", "invalid operatorFlagsMergeStrategy '%s', can only be 'merge' or 'reset'", payload.OperatorFlagsMergeStrategy)
+	}
+	// return if payload is empty
+	if payload.Cluster == nil {
+		return nil
+	}
+	properties := &payload.Cluster.Properties
+	if properties == nil || properties.OperatorFlags == nil {
+		return nil
+	}
+	// return nil, if OperatorFlagsMergeStrategy is merge and payload has not operatorFlags
+	// return operatorFlags of payload, if OperatorFlagsMergeStrategy is merge and payload has operatorFlags
+	// return defaultOperatorFlags, if OperatorFlagsMergeStrategy is reset and payload has not operatorFlags
+	// return defaultOperatorFlags + operatorFlags of payload, if OperatorFlagsMergeStrategy is reset and payload has operatorFlags
+	if payload.OperatorFlagsMergeStrategy == operatorFlagsMergeStrategyReset {
+		oc.Properties.OperatorFlags = operator.DefaultOperatorFlags()
+		for operatorflag, value := range payload.Cluster.Properties.OperatorFlags {
+			oc.Properties.OperatorFlags[operatorflag] = value
+		}
+	}
+
+	return nil
+}
--- a/pkg/api/admin/operatorflags_test.go
+++ b/pkg/api/admin/operatorflags_test.go
@@ -1,0 +1,60 @@
+package admin
+
+import (
+	"testing"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	utilerror "github.com/Azure/ARO-RP/test/util/error"
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+func TestOperatorFlagsMergeStrategy(t *testing.T) {
+	tests := []struct {
+		name    string
+		oc      *api.OpenShiftCluster
+		wantOc  *api.OpenShiftCluster
+		body    []byte
+		wantErr string
+	}{
+		{
+			name:    "invalid_json",
+			oc:      nil,
+			body:    []byte(`{{}`),
+			wantErr: `400: InvalidRequestContent: : The request content was invalid and could not be deserialized: "invalid character '{' looking for beginning of object key string".`,
+		},
+		{
+			name:    "OperatorFlagsMergeStrategy_is_not_merge_or_reset",
+			oc:      nil,
+			body:    []byte(`{"operatorFlagsMergeStrategy": "xyz"}`),
+			wantErr: `400: InvalidParameter: : invalid operatorFlagsMergeStrategy 'xyz', can only be 'merge' or 'reset'`,
+		},
+		{
+			name: "OperatorFlagsMergeStrategy_payload_is_empty",
+			oc: &api.OpenShiftCluster{
+				Properties: api.OpenShiftClusterProperties{
+					OperatorFlags: api.OperatorFlags{"aro.feature1.enabled": "false"},
+				},
+			},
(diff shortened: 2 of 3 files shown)


bennerv:
Is there a reason we need a flag to merge these?  IMO it should be default behavior to take whatever is present within the cluster document + whatever our defaults are, and merge them together.  

I don't really like adding additional series of maintenance tasks because they're slowly getting out of control.  

@hawkowl what are your thoughts?

pkg/api/admin/operatorflags.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/Azure/ARO-RP/pkg/api"
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+type operatorFlagsMergeStrategyStruct struct {
+	OperatorFlagsMergeStrategy string
+	Oc                         api.OpenShiftCluster
```
xiaoyu74, line 15: The variable oc is a bit vague. It might be more clear to name it `cluster` or `OpenShiftCluster` ?

pkg/api/admin/operatorflags.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// return error if OperatorFlagsMergeStrategy is not merge or reset, default is merge
+	if payload.OperatorFlagsMergeStrategy != "merge" && payload.OperatorFlagsMergeStrategy != "reset" {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", "invalid operatorFlagsMergeStrategy '%s', can only be 'merge' or 'reset'", payload.OperatorFlagsMergeStrategy)
+	}
+	// return nil, if OperatorFlagsMergeStrategy is merge and payload has not operatorFlags
+	// return operatorFlags of payload, if OperatorFlagsMergeStrategy is merge and payload has operatorFlags
+	// return defaultOperatorFlags, if OperatorFlagsMergeStrategy is reset and payload has not operatorFlags
+	// return defaultOperatorFlags + operatorFlags of payload, if OperatorFlagsMergeStrategy is reset and payload has operatorFlags
+	if payload.OperatorFlagsMergeStrategy == "reset" {
```
xiaoyu74, line 46: Before accessing payload.Oc.Properties.OperatorFlags, it may be a good idea to ensure that payload.Oc.Properties is not nil to prevent a potential panic ? 
```
if payload.OpenShiftCluster.Properties == nil {
    return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidRequestContent, "", "The request content was invalid: Properties is nil.")
}
```

pkg/api/admin/operatorflags.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	// return error if OperatorFlagsMergeStrategy is not merge or reset, default is merge
+	if payload.OperatorFlagsMergeStrategy != "merge" && payload.OperatorFlagsMergeStrategy != "reset" {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", "invalid operatorFlagsMergeStrategy '%s', can only be 'merge' or 'reset'", payload.OperatorFlagsMergeStrategy)
+	}
+	// return nil, if OperatorFlagsMergeStrategy is merge and payload has not operatorFlags
+	// return operatorFlags of payload, if OperatorFlagsMergeStrategy is merge and payload has operatorFlags
+	// return defaultOperatorFlags, if OperatorFlagsMergeStrategy is reset and payload has not operatorFlags
+	// return defaultOperatorFlags + operatorFlags of payload, if OperatorFlagsMergeStrategy is reset and payload has operatorFlags
```
xiaoyu74, line 45: Before accessing payload.Oc.Properties.OperatorFlags, it may be a good idea to ensure that payload.Oc.Properties is not nil to prevent a potential panic ?
```
if payload.OpenShiftCluster.Properties == nil {
    return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidRequestContent, "", "The request content was invalid: Properties is nil.")
}
```

pkg/api/admin/operatorflags.go
```diff
@@ -0,0 +1,54 @@
...
+	"encoding/json"
+	"net/http"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+type operatorFlagsMergeStrategyStruct struct {
```
xiaoyu74, line 13: nit:  follow the CamelCase to rename  `operatorFlagsMergeStrategyStruct` to `OperatorFlagsMergeStrategyStruct` for better clarity ?

pkg/api/admin/operatorflags.go
```diff
@@ -0,0 +1,54 @@
...
+	OperatorFlagsMergeStrategy string
+	Oc                         api.OpenShiftCluster
+}
+
+const operatorFlagsMergeStrategyDefault string = "merge"
```
xiaoyu74, line 18: Is that better to replace the string literals "merge" and "reset" with constants, especially if they are used in multiple places in your codebase to prevent bugs due to typos?
```
const (
	OperatorFlagsMergeStrategyDefault = "merge"
	OperatorFlagsMergeStrategyMerge   = "merge"
	OperatorFlagsMergeStrategyReset   = "reset"
)
```

pkg/api/admin/operatorflags.go
SrinivasAtmakuri: We should skip on an empty payload, I added an extra check for this, thank you,

pkg/api/admin/operatorflags.go
SrinivasAtmakuri: Made the change.

pkg/api/admin/operatorflags.go
SrinivasAtmakuri: Agreed, thats better, change made.

pkg/api/admin/operatorflags.go
```diff
@@ -0,0 +1,63 @@
...
+	// return error if OperatorFlagsMergeStrategy is not merge or reset, default is merge
+	if payload.OperatorFlagsMergeStrategy != operatorFlagsMergeStrategyMerge &&
+		payload.OperatorFlagsMergeStrategy != operatorFlagsMergeStrategyReset {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", "invalid operatorFlagsMergeStrategy '%s', can only be 'merge' or 'reset'", payload.OperatorFlagsMergeStrategy)
+	}
+	// return nil, if OperatorFlagsMergeStrategy is merge and payload has not operatorFlags
+	// return operatorFlags of payload, if OperatorFlagsMergeStrategy is merge and payload has operatorFlags
+	// return defaultOperatorFlags, if OperatorFlagsMergeStrategy is reset and payload has not operatorFlags
+	// return defaultOperatorFlags + operatorFlags of payload, if OperatorFlagsMergeStrategy is reset and payload has operatorFlags
+	if payload.OperatorFlagsMergeStrategy == operatorFlagsMergeStrategyReset {
```
xiaoyu74, line 55: I read the logic of the function a bit more and IIUC, looks the code only handles the 'reset' strategy. The 'merge' strategy is mentioned in the comments and constants, but there's no code that specifically handles this case. Do you need to include the logic for the "merge" strategy as well ?
```
if payload.OperatorFlagsMergeStrategy == operatorFlagsMergeStrategyMerge {
    for operatorflag, value := range payload.Cluster.Properties.OperatorFlags {
        oc.Properties.OperatorFlags[operatorflag] = value
    }
}
```

SrinivasAtmakuri: I have added comments, maybe thats not super clear, I will try to edit it, so when the payload.OperatorFlagsMergeStrategy == operatorFlagsMergeStrategyMerge , we just return nil, I have not explicitly added this as the existing code already does this and I decided to use comments instead.

pkg/api/admin/operatorflags.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	// return error if OperatorFlagsMergeStrategy is not merge or reset, default is merge
+	if payload.OperatorFlagsMergeStrategy != operatorFlagsMergeStrategyMerge &&
+		payload.OperatorFlagsMergeStrategy != operatorFlagsMergeStrategyReset {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", "invalid operatorFlagsMergeStrategy '%s', can only be 'merge' or 'reset'", payload.OperatorFlagsMergeStrategy)
+	}
+	// return nil, if OperatorFlagsMergeStrategy is merge and payload has not operatorFlags
+	// return operatorFlags of payload, if OperatorFlagsMergeStrategy is merge and payload has operatorFlags
+	// return defaultOperatorFlags, if OperatorFlagsMergeStrategy is reset and payload has not operatorFlags
+	// return defaultOperatorFlags + operatorFlags of payload, if OperatorFlagsMergeStrategy is reset and payload has operatorFlags
```
petrkotas, line 54: These three lines are a bit confusing. The code should be returning 'error' but these lines say it will return operatorFlags or defaultOperatorFlags. I believe these comments should be changed to mention the code is modifying the  document passed in.

I am also looking for the exact place, where the flags are merged. I cannot find it.

pkg/api/admin/operatorflags.go
```diff
@@ -0,0 +1,63 @@
...
+	if payload.OperatorFlagsMergeStrategy == operatorFlagsMergeStrategyReset {
+		oc.Properties.OperatorFlags = api.DefaultOperatorFlags()
+		for operatorflag, value := range payload.Cluster.Properties.OperatorFlags {
+			oc.Properties.OperatorFlags[operatorflag] = value
+		}
+	}
```
petrkotas, line 60: I am convinced that these lines do not reset. They rather merge. What am I missing to have the full context?

petrkotas (changes requested):
Hi @SrinivasAtmakuri I have two questions and one humble request. Would you please cover this function with unit test to verify it is doing what is expected? Thank you

pkg/api/admin/operatorflags.go
hawkowl: They reset to the default flags, and then merge the flags in the request (so that you don't have a reset with potentially incompatible-with-the-cluster flags)

hawkowl: I don't think we actually need to do this here, though, since the payload's flags will be merged in at https://github.com/Azure/ARO-RP/blob/614efbb55f6c7b18788cc6773a953fe3a918701f/pkg/frontend/openshiftcluster_putorpatch.go#L158

pkg/api/admin/operatorflags.go
hawkowl: It's a bit silly, but it's done in https://github.com/Azure/ARO-RP/blob/614efbb55f6c7b18788cc6773a953fe3a918701f/pkg/frontend/openshiftcluster_putorpatch.go#L158 , because the json decoding automatically merges maps.

pkg/api/admin/operatorflags.go
petrkotas: Lets keep it here, for the sake of making it explicit.

SrinivasAtmakuri:
Adding more test cases and retesting the code. Will update once done. Thanks.

SrinivasAtmakuri:

PR #2754, opened 2023-03-02
CosmoDB firewall rules required for the transient-VM-deploy in all regions

darthhexx:

Fixes: https://issues.redhat.com/browse/ARO-2393

We were only using transient-VM deploys on GovCloud, due to the lack of ACI VNET user-assigned MSI support, but we are changing all ACI fallback regions to use the same method.

This means we only have a regular ACI deploy or a VM deploy option and no 3rd option to use ACI in a fallback region. This simplifies the infrastructure required to use a fallback, however we do need to have the CosmoDB firewall setup to allow traffic from the AKS subnet that the transient-VM is deployed into.

This PR makes the change to deploy the required firewall rules to all CosmoDB instances, instead of having conditional deploys based on an RP-Config setting. This method has been selected because it reduces the flakes in production. Basically all CosmoDB instances in production will have the same subnet firewall rules.

Note: This PR needs to be merged at the same time as these 2:
* [RP-Config changeover to using vmFallbackDeployLocations](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/RP-Config/pullrequest/7700254).
* [ARO.Pipelines removal of ACI fallback region support and the enablement of transient-VM-deploys for configured regions](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines/pullrequest/7700263).

### Test plan for issue:

Tested on a dedicated RP instance.

No.


petrkotas (approved):
ok LGTM

s-amann (approved):

cadenmarchese (approved):

UlrichSchlueter:
The PR looks good, but I'd prefer to hold this until we are finished with the 4.11 multi-version rollout

PR #2755, opened 2023-03-02
Feature/guardrails

yjst2012:

Fixes 
https://issues.redhat.com/browse/ARO-1400 
https://issues.redhat.com/browse/ARO-1410

It integrates gatekeeper and helpful policies to provide cluster guardrails for ARO end users.

ADR:  https://docs.google.com/document/d/1e4fyaKYKQAZ9Ske0qqCU7vD7KT82080FvQXDfNebmzk/edit#heading=h.sarh46666fmj

Deployment guide: https://docs.google.com/document/d/1zKacAg8JCRP86UBiib01CAPcXVcpuk8NPVwuGHUMF4o/edit#heading=h.4n3ih7t5vz82

### Test plan for issue:

[A test plan](https://docs.google.com/document/d/1lfJrpP6NvDsl6wtIuw3gIFiNXoyQCiyQgeFczpWbdKY/edit#heading=h.epoi5jsn8v1a) is in progress

Gatekeeper Policy development and testing instruction: https://github.com/Azure/ARO-RP/tree/feature/guardrails/pkg/operator/controllers/guardrails/policies#gatekeeper-policy-development-and-testing

Gator test instruction: https://github.com/Azure/ARO-RP/tree/feature/guardrails/pkg/operator/controllers/guardrails/policies#gator-test

not as I'm aware of

### some gatekeeper docs that may help your understanding ###

[Gatekeeper Doc](https://open-policy-agent.github.io/gatekeeper/website/docs/)
[Gatekeeper design doc](https://github.com/open-policy-agent/gatekeeper/tree/master/docs/design)
[Policy language reference (Rego)](https://www.openpolicyagent.org/docs/latest/policy-language/)
[Policy unit testing (ie. opa test)](https://www.openpolicyagent.org/docs/latest/policy-testing/)
[Policy integration testing (gator verify)](https://open-policy-agent.github.io/gatekeeper/website/docs/gator/)


pkg/operator/controllers/guardrails/guardrails_config.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+//go:embed staticresources
+var staticFiles embed.FS
+
+//go:embed policies/gktemplates
+var gkPolicyTemplates embed.FS
+
+//go:embed policies/gkconstraints
+var gkPolicyConraints embed.FS
```
bennerv, line 72:
```suggestion
var gkPolicyConstraints embed.FS

Comment (yjst2012) [reply pkg/operator/controllers/guardrails/guardrails_config.go:72]: good catch

File: pkg/util/dynamichelper/dynamichelper.go [hunk 2]
```diff
@@ -289,3 +422,132 @@ func makeURLSegments(gvr *schema.GroupVersionResource, namespace, name string) (
...
+		hasChanged = handleMutatingWebhook(new, expected)
+	}
+	var diff string
+	if _, ok := expected.(*corev1.Secret); !ok { // Don't show a diff if kind is Secret
+		diff = cmp.Diff(old, expected)
+	}
+	return expected, hasChanged, diff, nil
+}
+
+func handleDeployment(new, expected kruntime.Object) bool {
```
jaitaiwan, line 464: This function should be named hasDeploymentChanged as it returns a bool

pkg/util/dynamichelper/dynamichelper.go
```diff
@@ -66,12 +77,34 @@ func New(log *logrus.Entry, restconfig *rest.Config) (Interface, error) {
 	return dh, nil
 }
 
+func (dh *dynamicHelper) resolve(groupKind, optionalVersion string) (*schema.GroupVersionResource, error) {
```
jaitaiwan, line 80: I think it's dangerous to have two functions with effectively the same name `resolve` and `Resolve`

jaitaiwan (changes requested):
I've made some initial comments, I'll take a look at the rest tomorrow

pkg/util/dynamichelper/dynamichelper.go
yjst2012: it does not just check if it changed, it also copy stuff around, the **"has"** will be confusing then

pkg/util/dynamichelper/dynamichelper.go
yjst2012: Resolve() is a method of GVRResolver, while resolve() here is an local revised version of Resolve() with a call to Refresh() with resolve() fails, logically they are doing the same thing, this is why having the "same" name.

yjst2012: as discussed, moved the logic back to GVRResolver.

pkg/operator/controllers/guardrails/guardrails_policy.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (r *Reconciler) getPolicyConfig(ctx context.Context, na string) (string, string, error) {
+	parts := strings.Split(na, ".")
+	if len(parts) < 1 {
+		return "", "", errors.New("unrecognised name: " + na)
+	}
+	name := parts[0]
+
+	instance := &arov1alpha1.Cluster{}
+	err := r.client.Get(ctx, types.NamespacedName{Name: arov1alpha1.SingletonClusterName}, instance)
```
hawkowl, line 35: instead of fetching the policy config in this function (which is inside a loop), could we move it outside and pass it to the function?

pkg/util/deployer/deployer.go
```diff
@@ -127,6 +137,47 @@ func (depl *deployer) Remove(ctx context.Context, data interface{}) error {
...
+	if err != nil {
+		return "", err
+	}
+	nsValue, err := getField(obj, "Namespace")
+	if err != nil {
+		return "", err
+	}
+	name := nameValue.String()
+	ns := nsValue.String()
+	if reflect.TypeOf(obj).String() == "*v1.Namespace" {
```
hawkowl, line 152: Can we instead compare against `obj.GetObjectKind().GroupVersionKind().GroupKind()` here?

hawkowl (dismissed):
Fix up the go.mod merge conflict and we're all good!

pkg/operator/controllers/guardrails/guardrails_config.go
```diff
@@ -0,0 +1,120 @@
...
+	controllerMutatingWebhookTimeout         = "aro.guardrails.mutatingwebhook.timeoutSeconds"
+
+	controllerReconciliationMinutes     = "aro.guardrails.reconciliationMinutes"
+	controllerPolicyManagedTemplate     = "aro.guardrails.policies.%s.managed"
+	controllerPolicyEnforcementTemplate = "aro.guardrails.policies.%s.enforcement"
+
+	defaultNamespace = "openshift-azure-guardrails"
+
+	defaultManagerRequestsCPU = "100m"
+	defaultManagerLimitCPU    = "1000m"
```
SudoBrendan, line 43: Generally, limits on CPU (a compressible resource) are considered bad practice in k8s, in fact, you'll notice that most operators in OCP do not ship with CPU limits, and only workloads that are capable of failing 100% without disruption to the cluster set limits (e.g. cleanup tasks). ref: `oc get po -o custom-columns="Name:metadata.name,CPU-limit:spec.containers[*].resources.limits.cpu" -A` on the shared cluster.

For this workload specifically, I could eventually see this manifest as a large portion of requests timing out due to not having enough CPU cycles to handle a large number of API requests, with the Node this is being hosted on being at 25% CPU utilization on a 4-core machine.

So, to get to the point: When we've finally matured Gatekeeper with enforcement, what would a large portion of requests to gatekeeper timing out look like to a customer? My guess is (unwarranted) 401s, but it could be something else?

pkg/operator/controllers/guardrails/policies/README.md
```diff
@@ -0,0 +1,305 @@
...
+package aroprivilegednamespace
+
+test_input_allowed_ns {
+  input := { "review": input_ns(input_allowed_ns) }
+  results := violation with input as input
+  count(results) == 0
+}
+
+test_input_disallowed_ns1 {
+  input := { "review": input_review(input_disallowed_ns1) }
```
SudoBrendan, line 90: I've never looked at rego before, so pardon if I'm way off base, but should this be a call to `input_ns` vs `input_review`, which I don't see defined?

pkg/operator/controllers/guardrails/policies/README.md
```diff
@@ -0,0 +1,305 @@
...
+        kinds: ["Job", "CronJob"]
+      - apiGroups: ["rbac.authorization.k8s.io"]
+        kinds: ["Role", "RoleBinding"]
+      - apiGroups: ["policy"]
+        kinds: ["PodDisruptionBudget"]
+```
+
+## Test the rego
+
+* install opa cli, refer https://github.com/open-policy-agent/opa/releases/
```
SudoBrendan, line 142: This (and other lines here) look like we need a change to our onboarding docs to set up dev environments with all required tooling.

pkg/operator/controllers/guardrails/policies/README.md
```diff
@@ -0,0 +1,305 @@
...
+
+* after _test.go is done, test it out, and fix the problem
+  ```sh
+  opa test ../library/common.rego *.rego [-v] #-v for verbose
```
SudoBrendan, line 146: I think this would be better as a `make` target, like `make test-rego` or something

pkg/operator/controllers/guardrails/policies/README.md
```diff
@@ -0,0 +1,305 @@
...
+  ```
+
+## Generate the Constraint Templates
+
+* install gomplate which is used by generate.sh, see https://docs.gomplate.ca/installing/
+
+* execute generate.sh under policies, which will generate the acutal Constraint Templates under gktemplates folder, example:
```
SudoBrendan, line 153: also better as a `make` target. Can we add this to `make generate`?

pkg/operator/controllers/guardrails/policies/README.md
```diff
@@ -0,0 +1,305 @@
...
+          cpu: "100m"
+          memory: "30Mi"
+```
+the assertions section is the expected result
+
+gator test is done via cmd:
+
+test.sh executes both opa test and gator verify
+```sh
+ARO-RP/pkg/operator/controllers/guardrails/policies$ ./scripts/test.sh
```
SudoBrendan, line 221: more `make` opportunities, I'm going to stop here - I'd argue every shell command in this README section should somehow be part of a single `make` flow.

SudoBrendan (dismissed):
Alright - I got a good start (through the README), but there's a lot to look through here so I'll come back later.

So far, the biggest concern I have is missing docs on new required dev tooling, and not enough `make` targets to make those commands easy for day-1 developers to use instead of needing deep knowledge of our tooling. This will also make it easier to facilitate validation of our stuff during CI, which I think we should be doing and is missing from this PR - what good are tests if we don't execute them?

Other than that, looking ahead, I'm confused by a few files that have landed in `/pkg` - in particular, `*.md` and `*.sh` already have a well-defined place on our repo. I guess I can get behind `*.rego`, and `*.yaml` appearing here, since several other controllers already define `staticresources`, though it is worth pointing out that this is the first time we're putting non-golang files in `/pkg` that are not an artifact of `go generate`... To conform to our current conventions, I'd rather see:

1. `*.md` in /docs
2. `*.sh` in /hack (and have any relevant `make` targets created/amended in `Makefile`)
3. Any generated files in `/pkg` be ultimately tied into `go generate` (and, by extension, `make generate`)

Another thing I'm generally seeing is a verbiage disconnect, we seem to be using `Guardrails` and `Gatekeeper` (`gk`), I'd rather pick one and use it everywhere to prevent multiple names from creating confusion, or is there a distinction I'm missing?

hawkowl:
@SudoBrendan A few things from your comments:

- The plan is to move all the rego/etc into its own repo so that it can be shared with others (e.g. ROSA). It's here purely as a stop-gap.
- @yjst2012 Could you make a followup PR that removes the CPU limit on the gatekeeper pod by default? Just the request will do fine, as per Brendan's rationale.
- I agree that things should go in a makefile -- we're planning on having the rego tests and the like a part of CI, but that will be a follow-up.

With that said, I am going to merge this as-is so that we can follow up with fixes instead of holding this up much longer.

hawkowl (approved):

yjst2012:
@SudoBrendan @hawkowl 
the cpu limit (and request) came from [the upstream gatekeeper yaml](https://raw.githubusercontent.com/open-policy-agent/gatekeeper/master/deploy/gatekeeper.yaml), what I'm doing here is to make it configurable, so it can be modified via feature flag during runtime.
`        resources:
          limits:
            memory: 512Mi

to remove it will generate a deviation with upstream and extra work for every upgrade of gatekeeper version, do you still want me to remove it?

PR #2756, opened 2023-03-02
Fix failure to parse token errors as Cloud Errors - ARO-1577

s-fairchild:

Fixes https://issues.redhat.com/browse/ARO-1577

This PR updates all cases when `ParseUnverified` is called and returns a cloud error, which will then be converted to a `UserError`. This prevents the generic `InternalServerError` status code from being the only error.

### Test plan for issue:
Updated existing unit tests.

I don't believe so.

### DIFF
--- a/pkg/api/validate/dynamic/serviceprincipal.go
+++ b/pkg/api/validate/dynamic/serviceprincipal.go
@@ -25,7 +25,7 @@ func (dv *dynamic) ValidateServicePrincipal(ctx context.Context, clientID, clien
 	c := &azureclaim.AzureClaim{}
 	_, _, err = p.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
-		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", err.Error())
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalCredentials, "properties.servicePrincipalProfile", err.Error())
 	}
 
 	for _, role := range c.Roles {
--- a/pkg/api/validate/dynamic/serviceprincipal.go
+++ b/pkg/api/validate/dynamic/serviceprincipal.go
@@ -25,8 +25,7 @@ func (dv *dynamic) ValidateServicePrincipal(ctx context.Context, clientID, clien
 	c := &azureclaim.AzureClaim{}
 	_, _, err = p.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
-		err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", err.Error())
-		return err
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", err.Error())
 	}
 
 	for _, role := range c.Roles {
--- a/pkg/api/validate/dynamic/serviceprincipal.go
+++ b/pkg/api/validate/dynamic/serviceprincipal.go
@@ -25,6 +25,7 @@ func (dv *dynamic) ValidateServicePrincipal(ctx context.Context, clientID, clien
 	c := &azureclaim.AzureClaim{}
 	_, _, err = p.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
+		err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", err.Error())
 		return err
 	}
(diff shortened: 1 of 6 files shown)


pkg/util/instancemetadata/prod.go
```diff
@@ -71,6 +72,7 @@ func (p *prod) populateTenantIDFromMSI(ctx context.Context) error {
 	c := &azureclaim.AzureClaim{}
 	_, _, err = parser.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
+		err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", err.Error())
 		return err
```
cadenmarchese, line 76:
```suggestion
		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", err.Error())
```

pkg/util/clusterauthorizer/authorizer.go
```diff
@@ -71,6 +73,7 @@ func (a *azRefreshableAuthorizer) NewRefreshableAuthorizerToken(ctx context.Cont
 	c := &azureclaim.AzureClaim{}
 	_, _, err = p.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
+		err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", err.Error())
```
cadenmarchese, line 76: How did you confirm that all `token contains` errors come from `p.ParseUnverified(token.OAuthToken(), c)`? Mostly just curious.

pkg/api/validate/dynamic/serviceprincipal.go
```diff
@@ -25,6 +25,7 @@ func (dv *dynamic) ValidateServicePrincipal(ctx context.Context, clientID, clien
 	c := &azureclaim.AzureClaim{}
 	_, _, err = p.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
+		err = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", err.Error())
 		return err
```
cadenmarchese, line 29:
```suggestion
		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", err.Error())
```

pkg/util/clusterauthorizer/authorizer.go
s-fairchild: Searching for that string in our repo (excluding vendor/*) results in one result from a test. If you search for this string '"The provided service principal generated an invalid token."' you find one result in `aad.go`. The real failure to parse token error comes from [parser.go](https://github.com/Azure/ARO-RP/blob/master/vendor/github.com/form3tech-oss/jwt-go/parser.go#L99).

That makes me believe it's the source of all those errors.

pkg/util/instancemetadata/prod_test.go
```diff
@@ -205,7 +205,7 @@ func TestPopulateTenantIDFromMSI(t *testing.T) {
 					OAuthToken().
 					Return("invalid")
 			},
-			wantErr: "token contains an invalid number of segments",
+			wantErr: "400: InvalidServicePrincipalToken: properties.servicePrincipalProfile: token contains an invalid number of segments",
```
rogbas, line 208: Can you please replace in the error message `token contains an invalid number of segments` with something that's more meaningful to end customers in this context, like `The provided service principal is invalid` ?

pkg/util/instancemetadata/prod.go
```diff
@@ -71,7 +72,7 @@ func (p *prod) populateTenantIDFromMSI(ctx context.Context) error {
 	c := &azureclaim.AzureClaim{}
 	_, _, err = parser.ParseUnverified(token.OAuthToken(), c)
 	if err != nil {
-		return err
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalToken, "properties.servicePrincipalProfile", err.Error())
```
darthhexx, line 75: This function is called by the RP / Gateway /Monitor, etc. when they are first starting up and performing MSI authentication so there is no concept of an API or end user. This should stay as a regular Internal Server Error.

pkg/util/instancemetadata/prod_test.go
s-fairchild: @rogbas Yes of course, I'll update the PR.

pkg/util/instancemetadata/prod.go
s-fairchild: I changed it back to returning a regular error.

cadenmarchese (approved):
LGTM

PR #2759, opened 2023-03-03
Posting all cluster conditions irrespective of status check

LiniSusan:
… run

https://issues.redhat.com/browse/ARO-1702

Fixes
Removed the 'status' check condition before sending the metric to Geneva in 'emitAroOperatorConditions' function.

If the check for 'False' condition is removed, it will send all the details hourly, from which we can easily set the alerts for ServicePrincipal Invalid. 
Also in future if we require any cluster condition alerts or checks in geneva then we can easily set it because all the metrics will be readily available.

### DIFF
--- a/pkg/monitor/cluster/clusterobjectconditions.go
+++ b/pkg/monitor/cluster/clusterobjectconditions.go
@@ -13,30 +13,45 @@ import (
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 )
 
-var aroOperatorConditionsExpected = map[string]operatorv1.ConditionStatus{
+var clusterObjectConditionsExpected = map[string]operatorv1.ConditionStatus{
+	// ARO Operator Conditions
 	arov1alpha1.InternetReachableFromMaster: operatorv1.ConditionTrue,
 	arov1alpha1.InternetReachableFromWorker: operatorv1.ConditionTrue,
+	// Service Principal Condition
+	arov1alpha1.ServicePrincipalValid: operatorv1.ConditionTrue,
 }
 
-func (mon *Monitor) emitAroOperatorConditions(ctx context.Context) error {
+var metricname = map[string]string{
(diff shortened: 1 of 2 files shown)


SrinivasAtmakuri:
@LiniSusan  I added the changes as we discussed.

pkg/monitor/cluster/clusterobjectconditions.go
```diff
@@ -13,30 +13,45 @@ import (
...
+var clusterObjectConditionsExpected = map[string]operatorv1.ConditionStatus{
+	// ARO Operator Conditions
 	arov1alpha1.InternetReachableFromMaster: operatorv1.ConditionTrue,
 	arov1alpha1.InternetReachableFromWorker: operatorv1.ConditionTrue,
+	// Service Principal Condition
+	arov1alpha1.ServicePrincipalValid: operatorv1.ConditionTrue,
 }
 
-func (mon *Monitor) emitAroOperatorConditions(ctx context.Context) error {
+var metricname = map[string]string{
```
s-amann, line 24: instead of having two separate maps to add the same key to, can we have a struct to contain the ExpectedStatus and the MetricName for each Condition?

```golang
type conditionMetric struct {
	expectedStatus operatorv1.ConditionStatus
	metricName     string
}

var clusterObjectConditionMetrics = map[string]conditionMetric{
	arov1alpha1.InternetReachableFromMaster: {
		expectedStatus: operatorv1.ConditionTrue,
		metricName: "arooperator.conditions",
	},
}
```

s-amann (changes requested):
I think we can simplify this and make it easier to add new conditions we want to report metrics on.

pkg/monitor/cluster/clusterobjectconditions.go
SrinivasAtmakuri: Made the suggested changes.

SrinivasAtmakuri:
The existing code already meets the needed requirement, hence closing the changes. Thank you.

PR #2760, opened 2023-03-03
Public IP removal - add outboundType to api

tony-schndr:

Fixes https://issues.redhat.com/projects/ARO/issues/ARO-1862

Adds outboundType to the latest API while preserving the feature flag functionality.  The outboundType is honored if passed in via the api, if outboundType is empty the feature flag will be evaluated.

### Test plan for issue:
Unit tests and manually tested with full service RP in DEV.

Yes, need to update https://learn.microsoft.com/en-us/azure/openshift/howto-create-private-cluster-4x#create-a-private-cluster-without-a-public-ip-address-preview


SudoBrendan (commented):
Overall, this is LGTM - waiting to see the results of the testing. Great work!

kimorris27:
LGTM! Will approve once E2E tests have run successfully.

kimorris27 (approved):

s-amann (dismissed):
lgtm!

SudoBrendan (commented):
LGTM - some of these changes need to land in OpenShift/ARO-Installer as well, including `pkg/api/featureflags.go`. We need to test those changes in multiversion-land as well. I'm not entirely sure what is the source of truth anymore on featureflags...

tony-schndr:
@SudoBrendan I opened https://github.com/openshift/ARO-Installer/pull/19 to remove the feature flag from the installer.  From what I can tell, once the cluster doc gets to the ARO installer, this field will be set and the existing logic in the ARO-Installer will pick it up.  I don't see anything else that needs to change, worth a second look IMO.

tony-schndr:
Testing looks good for multiversion, successfully deployed a 4.10 / 4.11 cluster with UserDefinedRouting, as well as a 4.10 cluster with Loadbalancer.

pkg/api/v20230401/openshiftcluster.go
```diff
@@ -126,13 +126,23 @@ type ServicePrincipalProfile struct {
...
 // NetworkProfile represents a network profile.
 type NetworkProfile struct {
 	// The CIDR used for OpenShift/Kubernetes Pods.
 	PodCIDR string `json:"podCidr,omitempty"`
 
 	// The CIDR used for OpenShift/Kubernetes Services.
 	ServiceCIDR string `json:"serviceCidr,omitempty"`
+
+	// The OutboundType use for egress traffic.
+	OutboundType OutboundType `json:"outboundType,omitempty"`
```
bennerv, line 145: We still have `mutable:true` in the admin API so you can manually patch the clusters which we did this to from a support exception.  We don't need the `mutable:true` flag on this, just calling it out as something we need to get back to eventually.  

I'll put in a PR with a TODO as a follow-up item when we GA day 2 changing of outbound type.

bennerv (changes requested):
Couple things missing, otherwise really close:

1. Static validation of the outbound type.  We need to validate the outbound type is expected. `pkg/api/v20230401/openshiftcluster_validatestatic.go`


Regarding our conversation earlier today, we actually don't have to update the defaults, as in `pkg/api/defaults.go` we already set a default for outbound type.  This should have been happening on new cluster creation as well as PUCM so we will be good there.  

Otherwise I think it's good.

tony-schndr:
> * There is one way we can mitigate this, which I'd like you to follow-up with Jerome on.  We could take the value passed in by the user, and if empty, take the status of the feature flag for a single release, until the newest API is rolled out completely.

This has been addressed by honoring the API Value first, if its not specified we will use the FeatureFlag.  The check `determineOutboundType` was moved to the frontend, this is because when the backend reads the cluster doc for work the defaults have been set on the document (by the frontend).  Therefore we determine the outbound type just prior to setting defaults.

pkg/frontend/outbound_type.go
```diff
@@ -0,0 +1,28 @@
...
+
+func determineOutboundType(ctx context.Context, doc *api.OpenShiftClusterDocument, subscription *api.SubscriptionDocument) error {
+	var err error
+	// Honor the value of OutboundType if it was passed in via the API
+	if doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType != "" {
+		return err
+	}
+
+	// Determine if this is a cluster with user defined routing
+	doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType = api.OutboundTypeLoadbalancer
```
cadenmarchese, line 21: I think this line should execute in an `else` or default for the `if` immediately after it, rather than always executing and then checking if it's actually correct later.

cadenmarchese (commented):
Validation code looks good / readable to me. I have one comment regarding `determineOutboundRule` which will run when we go to PUCM the fleet.

pkg/frontend/outbound_type.go
tony-schndr: Agreed and updated.

pkg/frontend/outbound_type.go
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/util/feature"
+)
+
+func determineOutboundType(ctx context.Context, doc *api.OpenShiftClusterDocument, subscription *api.SubscriptionDocument) error {
```
bennerv, line 13: it always returns nil.  No point in having an error return type.

pkg/frontend/outbound_type.go
```diff
@@ -0,0 +1,29 @@
...
+
+	// Determine if this is a cluster with user defined routing
+	if doc.OpenShiftCluster.Properties.APIServerProfile.Visibility == api.VisibilityPrivate &&
+		doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPrivate &&
+		feature.IsRegisteredForFeature(subscription.Subscription.Properties, api.FeatureFlagUserDefinedRouting) {
+		doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType = api.OutboundTypeUserDefinedRouting
+	} else {
+		doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType = api.OutboundTypeLoadbalancer
+	}
+	return err
```
bennerv, line 28:
```suggestion
        doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType = api.OutboundTypeLoadbalancer

	// Determine if this is a cluster with user defined routing
	if doc.OpenShiftCluster.Properties.APIServerProfile.Visibility == api.VisibilityPrivate &&
		doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPrivate &&
		feature.IsRegisteredForFeature(subscription.Subscription.Properties, api.FeatureFlagUserDefinedRouting) {
		doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType = api.OutboundTypeUserDefinedRouting
	}

Comment (bennerv) [reply pkg/frontend/outbound_type.go:28]: nit: no if-else

File: pkg/frontend/openshiftcluster_putorpatch.go [hunk 5]
```diff
@@ -208,6 +208,9 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 		doc.Dequeues = 0
 	}
 
+	// TODO: Remove this once 23-04-01 API release is complete.
+	determineOutboundType(ctx, doc, subscription)
```
bennerv, line 212: should only be in `if isCreate` 

Otherwise clusters which are already created an run an update might accidentally get the UDR outbound type.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -208,6 +208,9 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
...
+
 	// SetDefaults will set defaults on cluster document
 	api.SetDefaults(doc)
```
bennerv, line 215: does api.SetDefaults still have the default setting for UDR?  If so, we're good as it will cover the existing cluster PUCMs from a frontend perspective.

pkg/frontend/outbound_type.go
bennerv: also we're not even checking it anyway so drop it.

bennerv (changes requested):

pkg/frontend/openshiftcluster_putorpatch.go
tony-schndr: Yes, api.SetDefaults will set it to Loadbalancer

bennerv (approved):
lgtm

PR #2768, opened 2023-03-09
Cluster RG tagging

kimorris27:

https://issues.redhat.com/browse/ARO-2168

This PR adds tagging for the cluster resource group, including Azure CLI updates that allow customers to pass a set of tags through both `az aro create` and `az aro update`. It introduces a new API field called ~~`ClusterResourceGroupTags`~~ `ResourceTags` for storing the tags. This same set of tags will also be applied to the cluster resources in a later PR.

#### Description of how the tags are applied:

- The set of tags can be applied at cluster creation time or as a day 2 operation, and the set can also be updated throughout the cluster lifecycle.
- Every time a PUT/PATCH request is made to a cluster, whether the request modifies the ~~`ClusterResourceGroupTags`~~ `ResourceTags` API field or not, the tags on the resource group will be updated in an idempotent manner to line up with the contents of the API field (as part of the `ensureResourceGroup()` function).
- The RP doesn't monitor the actual tags on the resource group and guarantee that they line up with the contents of the API field at all times. The RP just reconciles the tags upon every PUT/PATCH to the cluster.
- The RP only adds/updates tags that are added to the API field. It doesn't delete tags when they are removed from the API field. If a customer wants to delete old/unwanted tags, they can do it through the Azure portal.
- Up to 10 tags are allowed. ~~We could increase this limit if we wanted to, assuming that this set of tags will only be used for the cluster RG and *not* for the cluster resources.~~ This is because this set of tags will also apply to the cluster resources.

### Test plan for issue:

Unit tests and E2E tests; included in this PR

I'm not sure. Please let me know of any documentation I should update.


pkg/api/v20230401/openshiftcluster_validatestatic.go
```diff
@@ -346,6 +365,24 @@ func (sv openShiftClusterStaticValidator) validateIngressProfile(path string, p
...
 }
 
+func (sv openShiftClusterStaticValidator) validateClusterResourceGroupTags(path string, t *Tags) error {
+	if len(*t) > maxTags {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path, fmt.Sprintf("The provided set of cluster resource group tags is too large; it can contain at most %d tags.", maxTags))
+	}
+
+	for k, v := range *t {
+		if !sv.clusterResourceGroupTagIsValid(k, v) {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path, "One or more of the provided cluster resource group tags are invalid.")
```
s-amann, line 375: Are the rules for RG naming exposed in some msft docs we can provide along with the error message?

pkg/api/v20230401/openshiftcluster_validatestatic.go
```diff
@@ -346,6 +365,24 @@ func (sv openShiftClusterStaticValidator) validateIngressProfile(path string, p
...
+		}
+	}
+
+	return nil
+}
+
+func (sv openShiftClusterStaticValidator) clusterResourceGroupTagIsValid(key string, value string) bool {
```
s-amann, line 382: if this returned the invalid tag name in addition or in replace of the bool we could then let the customer know which tag they provided was invalid.

pkg/api/v20230401/openshiftcluster_validatestatic.go
kimorris27: This part is validating tags, not the RG name. But to answer the question with respect to tags, the rules are _not_ exposed in MSFT docs (see my response to your other question about the regex for validating tags).

I could definitely still add the rules into the error message though.

cadenmarchese (commented):
Great work, here's a first pass review. Mostly nits and questions, I do think we will need customer-facing documentation later down the road when this is released.

pkg/api/v20230401/openshiftcluster_validatestatic_test.go
```diff
@@ -771,6 +771,64 @@ func TestOpenShiftClusterStaticValidateIngressProfile(t *testing.T) {
...
+				oc.Properties.ClusterResourceGroupTags = map[string]string{
+					"key": "value",
+				}
+			},
+		},
+		{
+			name: "too many tags",
+			modify: func(oc *OpenShiftCluster) {
+				oc.Properties.ClusterResourceGroupTags = map[string]string{}
+				for i := 0; i < 11; i++ {
```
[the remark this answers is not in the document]
kimorris27: Updated this test case 👍🏻

pkg/api/v20230401/openshiftcluster_validatestatic.go
```diff
@@ -20,6 +21,21 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/uuid"
 )
 
+// The maximum allowed number of cluster resource group tags.
+//
+// See https://github.com/openshift/enhancements/blob/master/enhancements/api-review/azure_user_defined_tags.md
+// to understand how the value was decided.
+var maxTags = 10
```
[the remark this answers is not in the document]
kimorris27: I think we're all set to use 10 based on the discussion in Slack yesterday.

pkg/util/azureclient/mgmt/redhatopenshift/2023-04-01/redhatopenshift/openshiftclusters_addons.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (c *openShiftClustersClient) CreateOrUpdateAndWait(ctx context.Context, resourceGroupName string, resourceName string, parameters mgmtredhatopenshift20230401.OpenShiftCluster) error {
+	future, err := c.CreateOrUpdate(ctx, resourceGroupName, resourceName, parameters)
+	if err != nil {
+		return err
+	}
+
+	return future.WaitForCompletionRef(ctx, c.Client)
+}
```
facchettos, line 28: is this used anywhere? I have not found the references to that function apart from its definition and the interface definition. (although I have only done a search on the pr so I could have missed it)

The others seem to be also not used in files in this pr?

facchettos (changes requested):
some suggestions, some questions. 
But this is a great PR good job! :)

pkg/util/azureclient/mgmt/redhatopenshift/2023-04-01/redhatopenshift/openshiftclusters_addons.go
kimorris27: This client is used in E2E tests, specifically in `test/e2e/tags.go`. I copied most of the code for this package over from the corresponding one for the `2022-09-04` API version, though I added the `UpdateAndWait()` function.

pkg/cluster/deploystorage_test.go
```diff
@@ -43,6 +43,13 @@ func TestEnsureResourceGroup(t *testing.T) {
 		"yeet": to.StringPtr("yote"),
 	}
 
+	modifyManagerClusterResourceGroupTags := func(m *manager) {
+		m.doc.OpenShiftCluster.Properties.ClusterResourceGroupTags = map[string]string{
+			"foo": "bar",
+			"bar": "baz",
+		}
+	}
```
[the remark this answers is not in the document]
kimorris27: I'm cool with changing this, but what difference does it make in this case?

pkg/api/v20230401/openshiftcluster_validatestatic.go
kimorris27: I've updated this error message now.

pkg/cluster/deploystorage_test.go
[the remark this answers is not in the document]
facchettos: Variable functions are bad practice when not specifically needed. They often end up getting unneeded values from the context they're in and that makes them untestable. Here it doesn't but if we can avoid using those I'd rather we do

pkg/api/v20230401/openshiftcluster_validatestatic.go
```diff
@@ -346,6 +365,45 @@ func (sv openShiftClusterStaticValidator) validateIngressProfile(path string, p
...
+Tag names must:
+- Start with a letter
+- End with a letter, number, or underscore
+- Contain only letters, numbers, underscores, periods, and hyphens
+- Have length <= 128
+
+Tag values have no character restrictions, but must have length <= 256.`
+
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path, errorMsg)
+	}
```
jaitaiwan, line 398: You're using two loops when you really only need one I think. Process the error message at the same time as you validate the tag exists.

pkg/api/v20230401/openshiftcluster_validatestatic.go
```diff
@@ -346,6 +365,45 @@ func (sv openShiftClusterStaticValidator) validateIngressProfile(path string, p
...
+
+	return nil
+}
+
+func (sv openShiftClusterStaticValidator) resourceTagIsValid(key string, value string) bool {
+	return tagNameRegexp.MatchString(key) && len(value) <= 256
```
jaitaiwan, line 404: I think the max length of a tag should also be set as a top-level constant

jaitaiwan (changes requested):
Couple minor requests but LGTM

pkg/api/v20230401/openshiftcluster_validatestatic.go
facchettos: adding on this, your `errorMsg` string starts with a `\n` is that on purpose? 

usually it's better to use a `strings.Builder` than using `+=` for strings because `+=` does a lot more allocations than a builder. 

Here the best would probably be using `strings.Join` to create the string containing all the tags actually.

kimorris27: Thanks for the suggestions! I reduced it one loop and used `strings.Join()` instead of concatenation.

pkg/api/v20230401/openshiftcluster_validatestatic.go
kimorris27: Made this change👍🏻

pkg/api/v20230401/openshiftcluster_validatestatic.go
```diff
@@ -346,6 +367,45 @@ func (sv openShiftClusterStaticValidator) validateIngressProfile(path string, p
...
+- Contain only letters, numbers, underscores, periods, and hyphens
+- Have length <= 128
+
+Tag values have no character restrictions, but must have length <= 256.`)
+
+	if len(invalidTags) > 0 {
+		errorMsg := strings.Join(errorMsgComponents, "")
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path, errorMsg)
+	}
+
```
facchettos, line 401: how about something like this ? 

```suggestion
	for k, v := range t {
		if !sv.resourceTagIsValid(k, v) {
			invalidTags = append(invalidTags, k)
			errorMsgComponents = append(errorMsgComponents, fmt.Sprintf("- %s\n", k))
		}
	}

	if len(invalidTags) > 0 {
		errorTemplate := `The following cluster resource tags either have an invalid name or an invalid value:
	- %s
Tag names must:
	- Start with a letter
	- End with a letter, number, or underscore
	- Contain only letters, numbers, underscores, periods, and hyphens
	- Have length <= 128
Tag values have no character restrictions, but must have length <= 256.
	`
		errorMsg := fmt.Sprintf(errorTemplate, strings.Join(invalidTags, "\n\t- "))
		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path, errorMsg)
	}

```

facchettos (changes requested):

pkg/api/v20230401/openshiftcluster_validatestatic.go
kimorris27: That makes sense. The tabs help with readability too. I've changed it to be like this.

jaitaiwan (approved):
LGTM

kimorris27:
Closing following change of approach; will open a separate new PR when ready.

PR #2771, opened 2023-03-14
setting bootstrap size to master size through installconfig

jhoreman:

Fixes

Sets bootstrap VM deploy size to selected master size


SudoBrendan (approved):
LGTM

bennerv (commented):
Can we update `pkg/frontend/quota.go` to increase the number of "masters" to 4?  We create the bootstrap and master VMs at the same time so technically the quota requires 4 VMs instead of three VMs of the master size.

hawkowl:
Since we use ARO-Installer everywhere now, we probably shouldn't land this here -- but we should do the quota change?

jhoreman:
> Since we use ARO-Installer everywhere now, we probably shouldn't land this here -- but we should do the quota change?

Hey Amber, so this was a question I had as well. Long story short is that I believe we're still referencing this here on cluster create... I have an identical change in the ARO-Installer repo (ref: https://github.com/openshift/ARO-Installer/pull/20) just in case, but when I empirically tested this on my own full service RP changing the value in the ARO-Installer repo had no effect, but making the same change in ARO-RP ultimately did. What are your thoughts on this?

jhoreman:
> Can we update `pkg/frontend/quota.go` to increase the number of "masters" to 4? We create the bootstrap and master VMs at the same time so technically the quota requires 4 VMs instead of three VMs of the master size.

@bennerv incremented requiredresourced for masters by 1 (now 4)

tony-schndr:
> > Since we use ARO-Installer everywhere now, we probably shouldn't land this here -- but we should do the quota change?
> 
> @hawkowl Hey Amber, so this was a question I had as well. Long story short is that I believe we're still referencing this here on cluster create... I have an identical change in the ARO-Installer repo (ref: [openshift/ARO-Installer#20](https://github.com/openshift/ARO-Installer/pull/20)) just in case, but when I empirically tested this on my own full service RP changing the value in the ARO-Installer repo had no effect, but making the same change in ARO-RP ultimately did. What are your thoughts on this?

I do think @hawkowl is correct here, any chance that you did not enable installation/adoption via hive in `dev-config.yaml`?  Those values default to "" and aren't mentioned in the full service rp documentation unfortunately.  If you missed that then the integrated installer was used.
```
  clustersAdoptByHive: "true"
  clustersInstallViaHive: "true
```

jhoreman:
> > > Since we use ARO-Installer everywhere now, we probably shouldn't land this here -- but we should do the quota change?
> > 
> > 
> > @hawkowl Hey Amber, so this was a question I had as well. Long story short is that I believe we're still referencing this here on cluster create... I have an identical change in the ARO-Installer repo (ref: [openshift/ARO-Installer#20](https://github.com/openshift/ARO-Installer/pull/20)) just in case, but when I empirically tested this on my own full service RP changing the value in the ARO-Installer repo had no effect, but making the same change in ARO-RP ultimately did. What are your thoughts on this?
> 
> I do think @hawkowl is correct here, any chance that you did not enable installation/adoption via hive in `dev-config.yaml`? Those values used to default to "" and aren't mentioned in the full service rp documentation unfortunately. If you missed that then the integrated installer was used.
> 
> ```
>   clustersAdoptByHive: "true"
>   clustersInstallViaHive: "true
> ```

Thanks for this insight @tony-schndr this is exactly what is happening. I have an identical change to this one in the aro-installer repo as well which can be found here: https://github.com/openshift/ARO-Installer/pull/20. For parity I think it's okay to leave the change here as well, but let me know if you want me to revert the vmsize change but keep the quota checking here. @bennerv @hawkowl

bennerv:
Relevant Installer change: https://github.com/openshift/installer/pull/5515

bennerv:
@jhoreman we should keep the quota check here, but backport this to the installer for 4.11.

openshift-installer included a similar change in 4.11.  Any thoughts on backporting it to 4.10 as well?  I'm +1 since we would only have to update the quota here and not have a conditional based on target install version.

pkg/installer/deployresources_resources.go
```diff
@@ -106,7 +106,7 @@ func (m *manager) computeBootstrapVM(installConfig *installconfig.InstallConfig)
 	vm := &mgmtcompute.VirtualMachine{
 		VirtualMachineProperties: &mgmtcompute.VirtualMachineProperties{
 			HardwareProfile: &mgmtcompute.HardwareProfile{
-				VMSize: mgmtcompute.VirtualMachineSizeTypesStandardD4sV3,
+				VMSize: mgmtcompute.VirtualMachineSizeTypes(installConfig.Config.ControlPlane.Platform.Azure.InstanceType),
```
SudoBrendan, line 109: Need to remove from here, will be included in openshift/installer-aro instead.

SudoBrendan (commented):
This has come a long way, thanks Jory! I think at this point we need to remove the /pkg/installer/* code, since that's going away very soon anyways...

SudoBrendan (approved):
LGTM

s-amann (approved):

PR #2777, opened 2023-03-15
Add providers validation to frontend

tsatam:

Fixes [ARO-1891](https://issues.redhat.com/browse/ARO-1891)

This PR moves the validation for whether users have registered the necessary resource providers for cluster creation to the frontend from the API. This allows us to provide a friendly error to the user indicating that the necessary resource providers must be registered, rather than giving them a generic "cluster creation failed" error.  

### Test plan for issue:

- All existing tests for the actual validation logic have been moved alongside the functionality
- New tests have been added to the frontend cluster creation to validate that the correct error is propagated when specific resource providers are not registered.

No (TBD)


pkg/frontend/providers_validation_test.go
```diff
@@ -0,0 +1,148 @@
...
+							Namespace:         to.StringPtr("otherRegisteredProvider"),
+							RegistrationState: to.StringPtr("Registered"),
+						},
+						{
+							Namespace:         to.StringPtr("otherNotRegisteredProvider"),
+							RegistrationState: to.StringPtr("NotRegistered"),
+						},
+					}, nil)
+			},
+		},
```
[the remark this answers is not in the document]
tsatam: I have the removal of the validation from the API in a separate commit that I haven't pushed up yet, I just put this draft PR up for CI checks. 

I agree with that test pattern change, I'll push that up in another commit.

SudoBrendan (approved):
LGTM

pkg/frontend/providers_validation.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return err
+	}
+
+	providersClient := features.NewProvidersClient(azEnv, subscriptionID, fpAuthorizer)
+
+	return validateProviders(ctx, providersClient)
+}
+
+func validateProviders(ctx context.Context, providersClient features.ProvidersClient) error {
```
cadenmarchese, line 35: This duplicate name might be confusing. I am tempted to say that we don't need this extra function since we are just 1) authorizing and then 2) checking against a list, we could do it all in ValidateProviders() above

pkg/frontend/providers_validation.go
```diff
@@ -0,0 +1,60 @@
...
+
+	for _, provider := range providers {
+		providerMap[*provider.Namespace] = provider
+	}
+
+	for _, provider := range []string{
+		"Microsoft.Authorization",
+		"Microsoft.Compute",
+		"Microsoft.Network",
+		"Microsoft.Storage",
```
cadenmarchese, line 51: Should we define the required resource providers somewhere outside of this function? Something like:
```
var requiredResourceProviders = []string{
	"Microsoft.Authorization",
	"Microsoft.Compute",
	"Microsoft.Network",
	"Microsoft.Storage",
}
```
Then we can range over that here.

pkg/frontend/providers_validation_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	for _, tt := range []struct {
+		name            string
+		mockProviders   []mgmtfeatures.Provider
+		mockProviderErr error
+		wantErr         string
+	}{
+		{
+			name: "pass",
+			mockProviders: []mgmtfeatures.Provider{
```
cadenmarchese, line 29: For cleanliness, we don't really need to pass the entire list on each spec. We could reduce the duplication by having a single list that we optionally modify in each test. I think this is what Jeremy is suggesting

cadenmarchese (dismissed):
This looks good to me, I have a few optional suggestions but otherwise approve. Thanks!

pkg/frontend/providers_validation_test.go
facchettos: not really what I was suggesting no, but I agree with your comment :)
what I was suggesting (and he did it), was to remove function from test variables, because it's kind of an equivalent to GOTO or something and makes reading the test code much harder

pkg/frontend/providers_validation.go
facchettos: to provide some context, this is how I did it in quotavalidation and this is (I think) why he did it this way. The reason why I did it this way is that it's not necessarily meaningful to test it with all of these parameters, and makes it harder to set up in the tests (in the tests we only test the private method). Now, if you want to change it, I don't really mind if you found a better way (but then we might want to apply the change to quotavalidator and invalidparameters)

tsatam: I did copy your pattern for QuotaValidation basically wholesale for this PR, so that is why I have the same setup. It also allowed me to migrate the validation unit tests as-is, since the private function shared the same signature as the old validation function.  

I agree that long-term we should inline these private functions, but should find a pattern to do it for all of the new frontend validators.

pkg/frontend/providers_validation.go
```diff
@@ -0,0 +1,62 @@
...
+
+	mgmtfeatures "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-07-01/features"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/env"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/features"
+)
+
+type ProvidersValidator interface {
```
jaitaiwan, line 18: I'd like to see documentation around what the ProviderValidator is/does.

PR #2778, opened 2023-03-15
Add the ability to enable dnsmasq retries via operator flags

ventifus:

Fixes https://issues.redhat.com/browse/ARO-2427. Introduces an operator flag `aro.dnsmasq.retry.enabled` that turns on retries in dnsmasq.

Depends on 
- release-4.10-azure: https://github.com/jewzaam/installer-aro/pull/14
- release-4.10-azure: https://github.com/jewzaam/installer-aro/pull/15

If a query is forwarded by dnsmasq is lost, then dnsmasq will hold on to that query for 40 seconds waiting for a response. If additional queries for that same name (and type) come in during that time, they will block on waiting for the first forwarded query.

### Test plan for issue:

Deploy to a test cluster. Let it run for a while (1 day or more) and check CoreDNS logs for "i/o timeout".

Yes, internal documentation needs to be created detailing how to enable the `aro.dnsmasq.retry.enabled` operator flag, and specify exactly when to enable it.
Once we have this deployed in the field solving real customer issues, we will consider making this default behavior and removing the flag.

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -83,5 +83,6 @@ func DefaultOperatorFlags() OperatorFlags {
 		"aro.autosizednodes.enabled":               flagFalse,
 		"rh.srep.muo.enabled":                      flagTrue,
 		"rh.srep.muo.managed":                      flagTrue,
+		"aro.dnsmasq.retry.enabled":                flagFalse,
 	}
 }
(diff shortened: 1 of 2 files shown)


ventifus:
Abandoning this approach because no version of RHCOS has a dnsmasq new enough to support this feature

PR #2779, opened 2023-03-16
Use CheckAccess with AFEC to test in INT and Canary regions

nwnt:
Fixes ARO-1800

This is for migrating from the current use of list permissions API to check access v2 API.  The list permissions API is to be deprecated by May..

### Test plan for issue:

Tested with local development environment for both with and without subscription flag scenarios.  All passed.

To be tested further in INT and Canary.

No, it changes from using ListPermissions API to Azure RBAC CheckAccess API v2.

The CheckAccess v2 API is not GAed yet, so we're going to test the preview endpoint in INT and Canary with a subscription feature flag.  Subscriptions set with the feature flag will direct the RP to use the new API while the rest will continue using the existing ListPermissions.  We expect to see the results from the logs whether the RP uses the new API during dynamic validations.


pkg/api/validate/dynamic/dynamic.go
```diff
@@ -292,34 +297,95 @@ func (dv *dynamic) validateNatGatewayPermissions(ctx context.Context, s Subnet)
...
-				// TODO(jminter): I don't understand if there are genuinely
-				// cases where CanDoAction can return false then true shortly
-				// after. I'm a little skeptical; if it can't happen we can
-				// simplify this code.  We should add a metric on this.
-				return false, err
-			}
+// usingCheckAccessV2 uses the new RBAC checkAccessV2 API
+func (c closure) usingCheckAccessV2() (bool, error) {
+	c.dv.log.Debug("retry validationActions with CheckAccessV2")
+	oid, err := aad.GetObjectId(c.dv.authorizer.OAuthToken())
```
niontive, line 351: ARO-RP has several authorizers. Could be useful to specify what authorizer is being used here

pkg/api/validate/openshiftcluster_validatedynamic.go
```diff
@@ -72,21 +117,14 @@ func (dv *openShiftClusterDynamicValidator) Dynamic(ctx context.Context) error {
...
-	tokenClient := aad.NewTokenClient()
-	token, err := tokenClient.GetToken(ctx, dv.log, spp.ClientID, string(spp.ClientSecret), dv.subscriptionDoc.Subscription.Properties.TenantID, dv.env.Environment().ActiveDirectoryEndpoint, dv.env.Environment().ResourceManagerEndpoint)
-	if err != nil {
-		return err
-	}
-
-	spAuthorizer := refreshable.NewAuthorizer(token)
-
-	spDynamic, err := dynamic.NewValidator(dv.log, dv.env, dv.env.Environment(), dv.subscriptionDoc.ID, spAuthorizer, dynamic.AuthorizerClusterServicePrincipal, aad.NewTokenClient())
+	//TODO create the cluster SP and add to NewValidator
```
niontive, line 120: Nit: `// TODO`

pkg/api/validate/dynamic/dynamic.go
niontive: Looks like this is specified in `[openshiftcluster_validatedynamic.go]`

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -292,34 +297,95 @@ func (dv *dynamic) validateNatGatewayPermissions(ctx context.Context, s Subnet)
...
+		return false, err
+	}
+
+	authReq := createAuthorizationRequest(oid, c.resource.String(), c.actions...)
+	results, err := c.dv.pdpClient.CheckAccess(c.ctx, authReq)
+	if err != nil {
+		return false, err
+	}
+
+	for _, result := range results.Value {
```
niontive, line 362: The return value for `CheckAccess` is a pointer to a struct, so this may lead to nil pointer dereference issues. Consider checking if `results` is nil before accessing it.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -291,34 +300,107 @@ func (dv *dynamic) validateNatGatewayPermissions(ctx context.Context, s Subnet)
...
+
+	authReq := createAuthorizationRequest(oid, c.resource.String(), c.actions...)
+	results, err := c.dv.pdpClient.CheckAccess(c.ctx, authReq)
+	if err != nil {
+		return false, err
+	}
+
+	if results == nil {
+		c.dv.log.Info("nil response returned from CheckAccessV2")
+		return false, nil
```
weinong, line 374: should this return an error?

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -291,34 +300,107 @@ func (dv *dynamic) validateNatGatewayPermissions(ctx context.Context, s Subnet)
...
+	}
+
+	for _, result := range results.Value {
+		if result.AccessDecision != remotepdp.Allowed {
+			c.dv.log.Debug(fmt.Sprintf("%s has no access to %s", oid, result.ActionId))
```
weinong, line 379: does debug log get logged in geneva?

pkg/util/aad/aad.go
```diff
@@ -125,3 +125,19 @@ func authenticateServicePrincipalToken(ctx context.Context, log *logrus.Entry, a
 
 	return nil
 }
+
+// GetObjectId extracts the "oid" claim from a given access jwtToken
+func GetObjectId(jwtToken string) (string, error) {
```
weinong, line 130: add unit test?

pkg/api/validate/openshiftcluster_validatedynamic.go
```diff
@@ -22,24 +26,66 @@ type OpenShiftClusterDynamicValidator interface {
...
+		aroEnv := env.Environment()
+		pdpClient = remotepdp.NewRemotePDPClient(
+			fmt.Sprintf(aroEnv.AzureRbacPDPEnvironment.Endpoint, env.Location()),
+			aroEnv.AzureRbacPDPEnvironment.OAuthScope,
+			fpClientCred,
+		)
+	}
+
+	tokenClient := aad.NewTokenClient()
+	token, err := tokenClient.GetToken(ctx, log, spp.ClientID, string(spp.ClientSecret), subscriptionDoc.Subscription.Properties.TenantID, env.Environment().ActiveDirectoryEndpoint, env.Environment().ResourceManagerEndpoint)
```
weinong, line 58: do we need to get Token in this function? can this be done lazily later?

pkg/util/aad/aad.go
```diff
@@ -125,3 +125,19 @@ func authenticateServicePrincipalToken(ctx context.Context, log *logrus.Entry, a
...
+	p := jwt.NewParser(jwt.WithoutClaimsValidation())
+	c := &custom{}
+	_, _, err := p.ParseUnverified(jwtToken, c)
```
niontive, line 133: The JWT Go package [recommends](https://pkg.go.dev/github.com/golang-jwt/jwt/v4#Parser.ParseUnverified) using `Parse` because `ParseUnverified` doesn't validate signatures. Is there a reason we are using `ParseUnverified` instead of `Parse`?

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -291,34 +300,107 @@ func (dv *dynamic) validateNatGatewayPermissions(ctx context.Context, s Subnet)
...
-			if !ok || err != nil {
-				// TODO(jminter): I don't understand if there are genuinely
-				// cases where CanDoAction can return false then true shortly
-				// after. I'm a little skeptical; if it can't happen we can
-				// simplify this code.  We should add a metric on this.
-				return false, err
-			}
+// usingCheckAccessV2 uses the new RBAC checkAccessV2 API
+func (c closure) usingCheckAccessV2() (bool, error) {
+	c.dv.log.Info("retry validationActions with CheckAccessV2")
```
niontive, line 353: Nit: `"retry validateActions ..."

pkg/util/aad/aad.go
nwnt: I thought about this and my conclusion was this is for getting the oid claim rather than doing an authentication (there's just not an easy way to get it from aad).  It also does this immediately after calling AAD.  The sender and the receiver are the same party (in this case the RP), unlike in other scenarios where the sender and receiver can't be trusted. 

ParseUnverified is also used in this codebase to get the tenant ID and roles as well.

niontive: Makes sense. Maybe we when deprecate go-autorest, we can revisit this and see if it makes sense to verify the token.

pkg/api/validate/openshiftcluster_validatedynamic.go
nwnt: This is how the existing code works.  I just moved these client creations into the New method rather than them being in the validate method.

weinong: i'm not arguing the functional or behavioral correctness, but i don't think a New function should make such RPC call.

pkg/util/aad/aad.go
s-amann, line 1: heads up: we plan on deleting the `aad` package in this PR. https://github.com/Azure/ARO-RP/pull/2667

cc @mbarnes

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -287,34 +296,107 @@ func (dv *dynamic) validateNatGatewayPermissions(ctx context.Context, s Subnet)
...
-			if !ok || err != nil {
-				// TODO(jminter): I don't understand if there are genuinely
-				// cases where CanDoAction can return false then true shortly
-				// after. I'm a little skeptical; if it can't happen we can
-				// simplify this code.  We should add a metric on this.
-				return false, err
-			}
+// usingCheckAccessV2 uses the new RBAC checkAccessV2 API
+func (c closure) usingCheckAccessV2() (bool, error) {
+	c.dv.log.Info("retry validateActions with CheckAccessV2")
```
weinong, line 349: in the old function, it is using `Debug()`. perhaps add a todo to clean it up after fully migrated

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -55,7 +58,9 @@ type Dynamic interface {
 
 type dynamic struct {
 	log            *logrus.Entry
+	authorizer     refreshable.Authorizer
 	authorizerType AuthorizerType
+	cred           azcore.TokenCredential
```
weinong, line 63: is it fp cred or cluster sp cred? please use an easily identifiable name

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -287,34 +296,107 @@ func (dv *dynamic) validateNatGatewayPermissions(ctx context.Context, s Subnet)
...
+func (c closure) usingCheckAccessV2() (bool, error) {
+	c.dv.log.Info("retry validateActions with CheckAccessV2")
+
+	scope := c.dv.azEnv.Environment.ResourceManagerEndpoint + "/.default"
+	t, err := c.dv.cred.GetToken(c.ctx, policy.TokenRequestOptions{Scopes: []string{scope}})
+	if err != nil {
+		return false, err
+	}
+
+	oid, err := aad.GetObjectId(t.Token)
```
weinong, line 357: during retry, can we skip this if oid is already found?

pkg/api/validate/dynamic/dynamic.go
nwnt: Changed to Infof now.

pkg/util/aad/aad.go
nwnt: Added

pkg/api/validate/dynamic/dynamic.go
nwnt: It could be both like the above `authorizer`.  The caller (dynamic() in `pkg/api/validate/openshiftcluster_validatedynamic.go` specifies which one to use.  It's not fixed to fp or sp.

weinong: perhaps call it `checkAccessSubjectCred` and add some comments that it may be either cluster SP or FP and is used in check access subject info

nwnt: that makes sense.  done.

pkg/api/validate/dynamic/dynamic.go
nwnt: Added accordingly.

pkg/api/validate/dynamic/dynamic.go
nwnt: done. now it makes a call only when oid is not found.

pkg/api/validate/openshiftcluster_validatedynamic.go
nwnt: done.  I moved all of them back to `dynamic()`.

pkg/util/aad/aad.go
nwnt: Thanks for the heads-up!  I moved this and its test cases to `pkg/util/token`

pkg/util/token/inspect.go
```diff
@@ -0,0 +1,22 @@
...
+	p := jwt.NewParser(jwt.WithoutClaimsValidation())
+	c := &custom{}
+	_, _, err := p.ParseUnverified(jwtToken, c)
+	if err != nil {
+		return "", err
+	}
+	return c.ObjectId, nil
+}
+
+type custom struct {
```
weinong, line 19: nit: can you move types before funcs?

pkg/util/aad/aad.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
-	"github.com/form3tech-oss/jwt-go"
+	"github.com/golang-jwt/jwt/v4"
```
weinong, line 13: why bother to change when this file is going away?

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -55,9 +59,12 @@ type Dynamic interface {
 
 type dynamic struct {
 	log            *logrus.Entry
+	authorizer     refreshable.Authorizer
```
s-amann, line 62: @mbarnes, will this need to be updated once #2667 is merged?

pkg/util/aad/aad.go
nwnt: it was part of a rebase yesterday.  this will go away with the new rebase.

pkg/api/validate/dynamic/dynamic.go
nwnt: I think this can be removed.  It's probably from before @mbarnes's PR.  I can see that there's no need to store this into dynamic.

nwnt: I removed it in my latest squashed push.  I believe this one is no longer needed so I removed it.  But apart from here, I changed all of refreshable to autorest accordingly.

pkg/api/validate/openshiftcluster_validatedynamic.go
```diff
@@ -22,7 +26,14 @@ type OpenShiftClusterDynamicValidator interface {
 }
 
 // NewOpenShiftClusterDynamicValidator creates a new OpenShiftClusterDynamicValidator
-func NewOpenShiftClusterDynamicValidator(log *logrus.Entry, env env.Interface, oc *api.OpenShiftCluster, subscriptionDoc *api.SubscriptionDocument, fpAuthorizer refreshable.Authorizer) OpenShiftClusterDynamicValidator {
+func NewOpenShiftClusterDynamicValidator(
+	ctx context.Context,
```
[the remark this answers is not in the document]
nwnt: Removed

s-amann (approved):

weinong (approved):
LGTM

cblecker (approved):

PR #2780, opened 2023-03-17
Update client retry delay when calling out to applens endpoint

cloudfit-nick:
This updates the retry delay from 10 seconds to 3 seconds to ensure that all their calls if needed happen with in 60 seconds.

ARO-2567

### DIFF
--- a/pkg/util/azureclient/applens/applens_client_options.go
+++ b/pkg/util/azureclient/applens/applens_client_options.go
@@ -26,6 +26,12 @@ func NewClientOptions() *ClientOptions {
 		azcore.ClientOptions{
 			Retry: policy.RetryOptions{
 				MaxRetries: 3,
+				// ARO-2567
+				// If the retry logic takes longer than 60 seconds,
+				// the correct error message will not be captured.
+				// With a setting of 3 seconds it was erroring out
+				// in ~30 seconds (3 seconds + 12 seconds + round
+				// trip / response time from all 3 calls).
 				RetryDelay: time.Second * 3,
 			},
 			Telemetry: policy.TelemetryOptions{


weinong (approved):
LGTM. However, we should probably create a common retry policy to be reused. cc: @nwnt

SudoBrendan (commented):
This seems fine, is there a work item or bug linked to this? What's the significance of 60 seconds?

weinong:
> This seems fine, is there a work item or bug linked to this? What's the significance of 60 seconds?

All RP frontend apis have 60s budget before ARM times out

pkg/util/azureclient/applens/applens_client_options.go
```diff
@@ -26,7 +26,7 @@ func NewClientOptions() *ClientOptions {
 		azcore.ClientOptions{
 			Retry: policy.RetryOptions{
 				MaxRetries: 3,
-				RetryDelay: time.Second * 10,
+				RetryDelay: time.Second * 3,
```
cblecker, line 29: 3 retries times 10 seconds.. that's 30 seconds. What's the timeout per try?

cloudfit-nick: It has an exponential back off, with a setting of 3 seconds it was erroring out in ~30 seconds (3 seconds + 12 seconds + round trip / response time from all 3 calls).  With a setting of 4 seconds, it was erroring out around 50 seconds.

To be safe I went with a setting of 3 seconds.

cloudfit-nick:
> This seems fine, is there a work item or bug linked to this? What's the significance of 60 seconds?

https://issues.redhat.com/browse/ARO-2567

pkg/util/azureclient/applens/applens_client_options.go
cblecker: That makes sense, but isn't immediately obvious. Can we add some code comments here as breadcrumbs for the next person that comes across this particular setting?

cloudfit-nick: Done

cloudfit-nick: Note on the place where it defined the setting it already calls out that its exponential.

petrkotas (approved):
LGTM, thanks

PR #2782, opened 2023-03-17
Capture subnet reconciliations

yehornaumenko:

https://issues.redhat.com/browse/ARO-2113

We wanna have a metric that shows if the subnet nsg controller is in the infinite loop. You can find more details in the JIRA ticket description.

### Test plan for issue:

E2Es (are updated in this PR)

### DIFF
--- a/pkg/monitor/cluster/nsgreconciliations.go
+++ b/pkg/monitor/cluster/nsgreconciliations.go
@@ -1,0 +1,50 @@
+package cluster
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+	subnetscontroller "github.com/Azure/ARO-RP/pkg/operator/controllers/subnets"
+)
+
+func (mon *Monitor) emitNSGReconciliation(ctx context.Context) error {
+	co, err := mon.arocli.AroV1alpha1().Clusters().Get(ctx, arov1alpha1.SingletonClusterName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	updated, err := annotationUpdated(co.Annotations)
+	if err != nil {
+		return err
+	}
+
+	if updated {
+		mon.emitGauge("nsg.reconciliations", int64(1), nil)
+	}
+
+	return nil
+}
+
+func annotationUpdated(annotations map[string]string) (bool, error) {
+	if annotations == nil {
+		return false, nil
+	}
+
+	t := annotations[subnetscontroller.AnnotationTimestamp]
+	timestamp, err := time.Parse(time.RFC1123, t)
+	if err != nil {
+		return false, err
+	}
+
+	if time.Since(timestamp) < time.Second*70 {
+		return true, nil
+	}
+
+	return false, nil
+}
--- a/pkg/operator/controllers/subnets/subnet_nsg.go
+++ b/pkg/operator/controllers/subnets/subnet_nsg.go
@@ -7,11 +7,16 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/util/subnet"
+)
+
+const (
+	AnnotationTimestamp = "aro.openshift.io/lastSubnetReconcileTimestamp"
 )
 
 func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet) error {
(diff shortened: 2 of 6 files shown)


pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -43,5 +48,17 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
...
 	err = r.subnets.CreateOrUpdate(ctx, s.ResourceID, subnetObject)
-	return err
+	if err != nil {
+		return err
+	}
+
+	return r.updateReconcileSubnetAnnotation(ctx)
+}
+
+func (r *reconcileManager) updateReconcileSubnetAnnotation(ctx context.Context) error {
```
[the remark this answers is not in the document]
yehornaumenko: Hi Petr.
We have decided with Gilou to change the design of this mechanism a bit. I have just updated the last passage of the JIRA ticket description. Could you take a look at it, please? 
Sorry for confusing you

jaitaiwan (dismissed):

s-amann (approved):

test/e2e/operator.go
```diff
@@ -342,6 +359,12 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
 				s, err := clients.Subnet.Get(ctx, resourceGroup, vnetName, subnet, "")
 				g.Expect(err).NotTo(HaveOccurred())
 				g.Expect(*s.NetworkSecurityGroup.ID).To(Equal(*correctNSG))
+
+				co, err := clients.AROClusters.AroV1alpha1().Clusters().Get(ctx, "cluster", metav1.GetOptions{})
+				log.Printf("Time now is: %v", time.Now().Format(time.RFC1123))
+				log.Printf("The cluster annotation is: %v", co.Annotations[subnetController.AnnotationTimestamp])
```
yehornaumenko, line 365: Debug logs to be deleted

pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -43,5 +48,17 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
 	r.log.Infof("Fixing NSG from %s to %s", oldNSG, correctNSGResourceID)
 	subnetObject.NetworkSecurityGroup = &mgmtnetwork.SecurityGroup{ID: &correctNSGResourceID}
 	err = r.subnets.CreateOrUpdate(ctx, s.ResourceID, subnetObject)
-	return err
+	if err != nil {
+		return err
+	}
+
+	return r.updateReconcileSubnetAnnotation(ctx)
```
rolandmkunkel, line 55: Could we do the update together with the update above, so that we can save one operation?

pkg/monitor/cluster/nsgreconciliations.go
```diff
@@ -0,0 +1,53 @@
...
+		return false, nil
+	}
+
+	t := annotations[subnetscontroller.AnnotationTimestamp]
+	timestamp, err := time.Parse(time.RFC1123, t)
+	if err != nil {
+		return false, err
+	}
+
+	if time.Since(timestamp) < time.Second*120 {
```
rolandmkunkel, line 48: I think metrics are emitted not in a fixed interval, but as quickly as it takes to run the stack once. Meaning with a fixed lookback time of 120 seconds, we would read the same event multiple times, and even worse, that  behavior would change if we add more metrics, or if the stack takes different times to run, depending on some cluster metric. Could we maybe cut off seconds and only take full minutes?

pkg/operator/controllers/subnets/subnet_nsg.go
yehornaumenko: No actually, because the first request is sent to Azure, and the second is to k8s/OpenShift API

pkg/monitor/cluster/nsgreconciliations.go
yehornaumenko: We have a ticker inside the `func (mon *monitor) worker(stop <-chan struct{}, delay time.Duration, id string)` function, so we actually emit metrics every 60 seconds

petrkotas (dismissed):
Alright. Good work Yehor.

pkg/monitor/cluster/nsgreconciliations.go
```diff
@@ -0,0 +1,53 @@
...
+		return err
+	}
+
+	if updated {
+		reconciled = 1
+	}
+
+	mon.emitGauge("nsg.reconciliations", int64(reconciled), nil)
+
+	return nil
```
facchettos, line 34: this could be simplified I think 
```suggestion
	updated, err := annotationUpdated(co.Annotations)
	if err != nil {
		return err
	}

	if updated {
            mon.emitGauge("nsg.reconciliations", int64(1), nil)
	}


	return nil
```

`emitGauge(...,0)` is equivalent to not calling it I think, since it is kind of a counter

facchettos (changes requested):

pkg/monitor/cluster/nsgreconciliations.go
yehornaumenko: There are different sampling strategies provided by Jarvis/Geneva, such as Sum, Count, and Avg. If we use only the Sum sampling method, we can neglect zeros emitted by this metric, but we will not be able to use other sampling methods. I am not sure if we actually need the other methods, but I am also not sure if it is good practice to simplify the code at the expense of sacrificing the ability to use different sampling methods.

@gvanderpotte, what are your thoughts?

gvanderpotte: I think that in that precise case, we can put "fill with 0s" when missing values, so i think it's equivalent in this precise case. As JEremy said, maybe let's not overflow Geneva if the value is normal (that's what we do for other metrics, we send only the value if it's abnormal state)

facchettos (approved):

petrkotas (approved):
LGTM. thanks!

PR #2784, opened 2023-03-20
now recreates the gateway vmss if there is a change in the probe

facchettos:
When updating the gateway to have a separate probe for the probe, it blocks the update of the deployment because the vmss is still using the probe. This deletes the vmss first in case it happens so we can update the probe and then recreate the vmss

Fixes

### DIFF
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -158,16 +158,26 @@ func (d *deployer) deploy(ctx context.Context, rgName, deploymentName, vmssName 
 	for i := 0; i < 3; i++ {
 		d.log.Printf("deploying %s", deploymentName)
 		err = d.deployments.CreateOrUpdateAndWait(ctx, rgName, deploymentName, deployment)
-		if serviceErr, ok := err.(*azure.ServiceError); ok &&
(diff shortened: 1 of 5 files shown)


petrkotas (approved):
So this is because we update the VMSS, which triggers rolling reboot, instead of first deleting the whole set and creating new ones. This is a reasonable workaround.

bennerv (dismissed):
All OCP clusters which have egress lockdown enabled require the gateway VMSS to be online.  During a deployment, we create the new Gateway Instances, confirm they're healthy, and then delete the older instances.  

This change will delete all gateway vmss instances, which will result in any clusters within that region which have egress lockdown in place to lose connectivity to the services the gateway acts as a proxy for.  

We need to find a way to either update the probe in real time or continue using the older probe port.

petrkotas (approved):
So when there are probes on VMSS, first remove them, still preserving traffic to the gateway. When removed, redeploy causing rolling redeploy.  New probes are setup during the deployment. 

This should work.

facchettos:
@petrkotas yes, but only if we face that specific error, otherwise it would be business as usual

PR #2786, opened 2023-03-20
update pucm to patch clusters that had their public IPs removed manually

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-2024

This PR adds a new maintenance task to PUCM so that the OutboundType for clusters that have had their public IP manually removed can be updated to UserDefinedRouting.  We have a list of all clusters that have had their public IP removed and will manually patch each one.

### Test plan for issue:

Added unit tests and tested the patch on a cluster.

No

### DIFF
--- a/pkg/api/admin/openshiftcluster_validatestatic.go
+++ b/pkg/api/admin/openshiftcluster_validatestatic.go
@@ -29,7 +29,7 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message)
 	}
 
-	if !(oc.Properties.MaintenanceTask == "" || oc.Properties.MaintenanceTask == MaintenanceTaskEverything || oc.Properties.MaintenanceTask == MaintenanceTaskOperator || oc.Properties.MaintenanceTask == MaintenanceTaskRenewCerts) {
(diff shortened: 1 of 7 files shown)


pkg/api/admin/openshiftcluster_validatestatic.go
```diff
@@ -29,7 +29,7 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message)
 	}
 
-	if !(oc.Properties.MaintenanceTask == "" || oc.Properties.MaintenanceTask == MaintenanceTaskEverything || oc.Properties.MaintenanceTask == MaintenanceTaskOperator || oc.Properties.MaintenanceTask == MaintenanceTaskRenewCerts) {
+	if !(oc.Properties.MaintenanceTask == "" || oc.Properties.MaintenanceTask == MaintenanceTaskEverything || oc.Properties.MaintenanceTask == MaintenanceTaskOperator || oc.Properties.MaintenanceTask == MaintenanceTaskRenewCerts || oc.Properties.MaintenanceTask == MaintenanceTaskPatchUserDefinedRouting) {
```
jaitaiwan, line 32: I think this would look better as a switch statement:
```
switch oc.Properties.MaintenanceTask {
  case "",
  MaintenanceTaskEverything,
  MaintenanceTaskOperator
  ...:
}
```

jaitaiwan (changes requested):
This PR looks good to me! I just had one suggestion for some code aesthetics :)

PR #2787, opened 2023-03-21
Pin the Azure SDK version so it doesn't try and get the version without preview SDKs

hawkowl:

fixes local dev (running go mod tidy)

pins it to the version so go mod will *actually* get that version

### Test plan for issue:
it no longer fails locally

n/a


hawkowl:
Don't need this now

PR #2788, opened 2023-03-21
fixes wrong return value for retry in vmss cleaner in deployer

facchettos:
fixes the wrong return value in the cleaner which prevented the deployer to redeploy the vmss

Fixes

### DIFF
--- a/pkg/deploy/vmsscleaner/clean.go
+++ b/pkg/deploy/vmsscleaner/clean.go
@@ -85,9 +85,8 @@ func (c *cleaner) UpdateVMSSProbes(ctx context.Context, rgName string) (retry bo
 		err = c.vmss.CreateOrUpdateAndWait(ctx, rgName, name, vmss)
 		if err != nil {
 			c.log.Warn(err)
-			return false // If update failed, gateway vmss still exists. Don't retry.
+			return false // If update failed, gateway vmss still has a reference to the probe
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/deploy/vmsscleaner/clean.go
```diff
@@ -85,9 +85,8 @@ func (c *cleaner) UpdateVMSSProbes(ctx context.Context, rgName string) (retry bo
 		err = c.vmss.CreateOrUpdateAndWait(ctx, rgName, name, vmss)
 		if err != nil {
 			c.log.Warn(err)
-			return false // If update failed, gateway vmss still exists. Don't retry.
+			return false // If update failed, gateway vmss still has a reference to the probe
 		}
 	}
-	// no scaleset matched, so we should not retry and return the error
-	return false
+	return true
```
s-amann, line 91: should we have at least a unit test for this?

s-amann (dismissed):

pkg/deploy/vmsscleaner/clean.go
facchettos: it's a temporary fix so I figured we didn't need them, but yes, we should definitely have had unit tests on that, will add them tomorrow

s-amann (dismissed):
lgtm, bonus points if somehow a test for `if !strings.HasPrefix(name, "gateway-vmss-")` gets added in the future

s-amann (approved):

PR #2789, opened 2023-03-22
Move CI/dev tool usage to use go run pkg@version to reduce the number of vendored dependencies

hawkowl:

Fixes my local dev environment because go mod tidy breaks

This:
- Removes some dev/tool only dependencies and replaces them with `go run pkg@version` invocations. This removes a large amount of dependencies that clutter up our vendor folder.
- Fixes one or two little imports that include more than they need
-  Updates one or two dependencies that don't have valid downloads anymore for unknown reasons

### Test plan for issue:

CI

N/A


facchettos (approved):
I love the idea. Question: does that mean that we are now dependent on github being up and running to run the CI? Wouldn't be compiling these deps into a container possible (because running on azure infra maybe not?)? If not I'm all in for your solution

facchettos:
for things like go-mock and similar things I'd rather have it in a dockerfile instead of compiling it everytime (since we could run those on github actions and it would be less restrictive)

hawkowl:
@facchettos It uses the go module cache, so it should ideally be not too flaky :)

PR #2793, opened 2023-03-22
temporarily serving the health probe on 80 to allow us to update lb

facchettos:
when moving the probe to a different port, we might be facing a short service outage on the gateway if not for this PR. The move to the new probe should be done in 2 steps, step 1 deploying the vmss which serves the 2 health probes and then change the probe to the new port. 

Fixes


facchettos:
superseeded by https://github.com/Azure/ARO-RP/pull/2801

PR #2794, opened 2023-03-22
Add crt to ARO Docker container for AME Root

cloudfit-nick:
Add AME Root the ARO RP Docker Container

The AppLens endpoint is using an AME Root cert so the CA needs to be added to the container so that Go will not reject the endpoint.

ARO-2612


Dockerfile.aro
```diff
@@ -1,6 +1,7 @@
 ARG REGISTRY
 FROM ${REGISTRY}/ubi8/ubi-minimal
 RUN microdnf update && microdnf clean all
+RUN curl -o /etc/pki/ca-trust/source/anchors/AMEROOT_ameroot.crt http://crl.microsoft.com/pkiinfra/certs/AMEROOT_ameroot.crt && update-ca-trust
```
petrkotas, line 4: How do we know this is the correct certificate? I see it is over http and it is public. Shoud it be?

cloudfit-nick: That is a good question, I also was surprised it was not over https.  I will look into it more.

weinong: the thumbprint is verified in Readme.md

weinong (approved):
LGTM

bennerv (commented):
Is there a well-known endpoint for this certificate?  Can we not just issue a request to the certificate on run, and then modify our tls config to trust the additional CA?  

We do something similar with client certificates for the ARM client authorizer.  `pkg/util/clientauthorizer/arm.go`

Then when the CA expires, assuming they update the well-known endpoint we will not have to make any code changes.

weinong:
@bennerv indeed. https://aka.ms/getissuers can be leveraged to get this CA, I think

cblecker:
Closing in favour of #2838

PR #2795, opened 2023-03-22
Increase cosmosdb default throughput to 1000

tsatam:

Fixes [ARO-2615](https://issues.redhat.com/browse/ARO-2615)

Increases the default throughput for the RP cosmosdb from 500 to 1000. 

### Test plan for issue:

None (config change)

No

### DIFF
--- a/pkg/deploy/assets/databases-development.json
+++ b/pkg/deploy/assets/databases-development.json
@@ -16,7 +16,7 @@
                     "id": "[parameters('databaseName')]"
                 },
                 "options": {
-                    "throughput": 500
+                    "throughput": 1000
                 }
             },
             "name": "[concat(parameters('databaseAccountName'), '/', parameters('databaseName'))]",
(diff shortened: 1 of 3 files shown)


s-amann (approved):

cadenmarchese:
Should we make updates to these as well (gateway and portal throughput)?

https://github.com/Azure/ARO-RP/blob/master/pkg/deploy/assets/rp-production.json#L889
https://github.com/Azure/ARO-RP/blob/master/pkg/deploy/assets/rp-production.json#L978

SrinivasAtmakuri (approved):
LGTM!

Shivkumar13 (approved):

Shivkumar13:
LGTM!

cadenmarchese:
> If this is the case, I think that, pending further investigation, we should leave these alone since neither the gateway nor portal container were involved in the throughput issue we are trying to mitigate here, and defer any potential changes to those two until [ARO-2617](https://issues.redhat.com/browse/ARO-2617).

Works for me, let's get it in for upcoming release.

PR #2796, opened 2023-03-23
Update Makefile to not allow unset variables

s-fairchild:
Adding SHELLFLAGS and switch to double dollar signs to prevent make from expanding bash variables

No current issues.

Currently unset variables are allowed in Makefile. I don't believe there should be a case where we need/want unset variables, I believe this a safer option.  

`.SHELLFLAGS := -o nounset -c` sets `nounset` and variables are expanded by bash by escaping with double dollar signs.

### Test plan for issue:
Existing tests that utilize make

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,16 @@
 SHELL = /bin/bash
+.SHELLFLAGS := -o nounset -c
 TAG ?= $(shell git describe --exact-match 2>/dev/null)
 COMMIT = $(shell git rev-parse --short=7 HEAD)$(shell [[ $$(git status --porcelain) = "" ]] || echo -dirty)
-ARO_IMAGE_BASE = ${RP_IMAGE_ACR}.azurecr.io/aro
+ARO_IMAGE_BASE = $${RP_IMAGE_ACR}.azurecr.io/aro
 E2E_FLAGS ?= -test.v --ginkgo.v --ginkgo.timeout 180m --ginkgo.flake-attempts=2 --ginkgo.junit-report=e2e-report.xml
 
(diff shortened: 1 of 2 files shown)


facchettos (dismissed):
:+1:

SrinivasAtmakuri (dismissed):
LGTM!

dem4gus (dismissed):
Good change 👍

Makefile
```diff
@@ -132,7 +133,7 @@ publish-image-e2e: image-e2e
 	docker push $(ARO_IMAGE)
 
 extract-aro-docker:
-	hack/ci-utils/extractaro.sh ${ARO_IMAGE}
+	hack/ci-utils/extractaro.sh $${ARO_IMAGE:-}
```
cadenmarchese, line 136: I think this is where your e2e failure is coming from:
```suggestion
	hack/ci-utils/extractaro.sh $${ARO_IMAGE}
```
~~~
+ DOCKERID=
make: *** [Makefile:136: extract-aro-docker] Error 1
~~~

s-fairchild: I was able to fix the e2e pipeline by hardcoding `ARO_IMAGE` here: https://github.com/Azure/ARO-RP/blob/ce4e22ae71e5d2886db1cdba5857ef19e067a682/.pipelines/e2e.yml#L47

cadenmarchese:
I also ran build and push pipeline to ensure that functionality isn't impacted:

https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=71135590&view=results

PR #2797, opened 2023-03-23
remove --enable-soft-delete option

mbukatov:
This option was removed in azure-cli 2.41.0, for more context see:

https://learn.microsoft.com/en-us/azure/key-vault/general/soft-delete-change


dem4gus (approved):
:+1:

cadenmarchese (approved):

PR #2800, opened 2023-03-24
Guardrail policy: Prevent regular user from managing resources in the certain API group

xiaoyu74:

Guardrail policy: Regular User Validation: https://issues.redhat.com/browse/ARO-2140

Prevent non-authorized user from managing resources in the certain APIgroup

### Test plan for issue:
Gator test and unit test

N/A


pkg/operator/controllers/guardrails/policies/gktemplates-src/aro-regularuser-validation/src_test.rego
```diff
@@ -0,0 +1,91 @@
+package aroregularuservalidation
+
+import data.aroregularuservalidation
+
+test_denies_regular_user_access {
+    input := {
```
ArrisLee, line 6: I think you can create a function for generating inputs instead of creating input obj in each cases

ArrisLee (commented):
left some comments, others LGTM

pkg/operator/controllers/guardrails/policies/gkconstraints/aro-regularuser-apigroup-access.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: ARODenyRegularUserApigroups
```
jaitaiwan, line 2: Probably a silly sticking point, but I think that it should be APIGroups not `Apigroups`

pkg/operator/controllers/guardrails/policies/gktemplates-src/aro-deny-regularuser-apigroups/aro-deny-regularuser-apigroups.tmpl
```diff
@@ -0,0 +1,24 @@
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: arodenyregularuserapigroups
```
jaitaiwan, line 4: For consistency this name should probably be aro-deny- etc.

pkg/operator/controllers/guardrails/policies/gktemplates/aro-deny-regularuser-apigroups.yaml
jaitaiwan, line 1: Maybe it would be wise to put rego and libs arguments into an external file so we have validation in editors?

jaitaiwan (commented):
I have some minor nit-picks. LGTM otherwise :D

pkg/operator/controllers/guardrails/policies/gktemplates/aro-deny-regularuser-apigroups.yaml
yjst2012: not sure if you meant the ConstraintTemplate yaml file is too long and difficult to read, if yes, this is maybe not easy, as yaml has to have all the contents and gatekeeper has [a long template file in their lib](https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/containerlimits/template.yaml) as well.
btw, this is not a problem, because it is auto-generated from the tmpl + rego files, which is the place to write our logic and review.

yjst2012 (changes requested):

pkg/operator/controllers/guardrails/policies/gkconstraints/aro-regularuser-apigroup-access.yaml
xiaoyu74: It's a good point. thanks

pkg/operator/controllers/guardrails/policies/gktemplates-src/aro-deny-regularuser-apigroups/src.rego
```diff
@@ -0,0 +1,12 @@
+package arodenyregularuserapigroups
+
+import future.keywords.in
+import data.lib.common.get_user
+
+# Prevent regular users from managing certain resources in restricted API group
+violation[{"msg": msg}] {
+    input.review.operation in ["CREATE", "UPDATE", "DELETE"]
+    user := get_user(input.review)
+    not regex.match("^(system|kube):.+", user)
```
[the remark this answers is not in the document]
xiaoyu74: Thanks @ArrisLee , it's a good point. 

As we discussed, this initial implementation is based on the [SPIKE doc](https://docs.google.com/document/d/1lbxVaGZ5T__foLRQ_npoa9QCg2mqOi4GBiSXHwLjPT8/edit#heading=h.dzeyoh17ibsy) (Not accurate enough though). The key point is how we define a regular user and considering both user groups and leveraging a service account are worth to check further. cc: #yjst2012

pkg/operator/controllers/guardrails/policies/gktemplates-src/aro-deny-regularuser-apigroups/aro-deny-regularuser-apigroups.tmpl
xiaoyu74: I think all the Constraint Template follow the same name patter without the `-` in between.

pkg/operator/controllers/guardrails/policies/gktemplates-src/aro-deny-regularuser-apigroups/src.rego
```diff
@@ -0,0 +1,12 @@
+package arodenyregularuserapigroups
+
+import future.keywords.in
+import data.lib.common.get_user
+
+# Prevent regular users from managing certain resources in restricted API group
+violation[{"msg": msg}] {
+    input.review.operation in ["CREATE", "UPDATE", "DELETE"]
+    user := get_user(input.review)
```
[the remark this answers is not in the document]
xiaoyu74: Thanks Jeff. I will use the common lib `is_exempted_account(input.review)` to check the authorized user.
Also need to check how the groups related to the authorized user identity.

yjst2012 (commented):
LGTM, please fix the yaml lint issues
btw, have you tested it in a real cluster

yjst2012 (approved):

xiaoyu74:
> LGTM, please fix the yaml lint issues btw, have you tested it in a real cluster

Checked with the [gatekeeper upstream](https://openpolicyagent.slack.com/archives/CDTN970AX/p1685725958551539?thread_ts=1685666994.371049&cid=CDTN970AX) looks our previous assumption that the restriction to access the resources ValidatingWebhookConfiguration, and MutatingWebhookConfiguration in the APIGroup doesn't take effect is an expected behavior but it is partial right, it just isn't specific to Gatekeeper, it's built into Kubernetes ([link](https://github.com/kubernetes/api/blob/63505c4e796f6267d5c811a45f020427447b6ebc/admissionregistration/v1beta1/types.go#L171))

Discussed in our Azure Arch meeting today and we probably cannot restrict the access from the gatekeeper policy level but from guardrail controller instead. So, we probably need to create a new SPIKE ticket to do some investigation about the new implementation.  WDYT @yjst2012

PR #2801, opened 2023-03-24
replaces the standard library mux with chi in the gateway

facchettos:
replaces the standard library with go-chi since we apparently lost the access logs with it. Standard library returned 404 for requests which should have been found. 

also put the probe back to the port 80 where it used to be so we don't run into the problem of `ProbeInUSeByVMSS` 

Fixes

### DIFF
--- a/pkg/deploy/assets/gateway-production.json
+++ b/pkg/deploy/assets/gateway-production.json
@@ -178,7 +178,7 @@
                     {
                         "properties": {
                             "protocol": "Http",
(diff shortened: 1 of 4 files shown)


jaitaiwan:
What's the "Why" of this PR?

cadenmarchese (approved):
Approved with similar comment as https://github.com/Azure/ARO-RP/pull/2802. Thanks for the fix!

PR #2802, opened 2023-03-24
replaced standard library with chi in dbtoken

facchettos:
When replacing `gorilla/mux` with the standard library we lost the access logging. Using `chi` instead of the standard library will add the access logging back

Fixes

### DIFF
--- a/pkg/dbtoken/server_test.go
+++ b/pkg/dbtoken/server_test.go
@@ -55,7 +55,7 @@ func TestServer(t *testing.T) {
 			wantStatusCode: http.StatusOK,
 		},
 		{
-			name: "GET unauthorized /token returns 403",
+			name: "GET unauthorized /token returns 405",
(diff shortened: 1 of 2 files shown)


pkg/dbtoken/server_test.go
```diff
@@ -55,7 +55,7 @@ func TestServer(t *testing.T) {
 			wantStatusCode: http.StatusOK,
 		},
 		{
-			name: "GET unauthorized /token returns 403",
+			name: "GET unauthorized /token returns 405",
```
facchettos, line 58: this is reverting the change and the behavior is back to exactly what it was with gorilla/mux

cadenmarchese (approved):
Without any testing on my end, this looks 👍 to me. I see why we don't want to go back to Gorilla, and if there are any issues with this we should pick it up very quickly during release

PR #2804, opened 2023-03-24
fix e2e tests when running against 4.12 clusters

mbukatov:

none

The default storage class on 4.12 cluster is the same as on 4.11, but  the correct default storage class is used on 4.11 clusters only,  which will make  2 test cases fail on 4.12 cluster.

When the current condition was created, we [were not sure what the default for 4.12 will be](https://github.com/Azure/ARO-RP/pull/2739/files#r1114609783), but now we [have this assumption in the code already](https://github.com/Azure/ARO-RP/pull/2585/files#diff-288bcd4df3d10098fe8b909ead3bf744c5e0e9fd5849623157d3046d11025c87R45).

For this reason, we need to update the version check so that the new storage classes are used with >= 4.11.

### Test plan for issue:

I run the e2e tests on 4.12 cluster during upgrade testing.

No documentation needs update.

### DIFF
--- a/test/e2e/cluster.go
+++ b/test/e2e/cluster.go
@@ -119,9 +119,9 @@ func createStatefulSet(ctx context.Context, cli kubernetes.Interface) error {
 	oc, _ := clients.OpenshiftClusters.Get(ctx, vnetResourceGroup, clusterName)
 	installVersion, _ := version.ParseVersion(*oc.ClusterProfile.Version)
 
-	defaultStorageClass := "managed-premium"
-	if installVersion.V[0] == 4 && installVersion.V[1] >= 11 {
-		defaultStorageClass = "managed-csi"
+	defaultStorageClass := "managed-csi"
+	if installVersion.Lt(version.NewVersion(4, 11)) {
+		defaultStorageClass = "managed-premium"
 	}
 	pvcStorage, err := resource.ParseQuantity("2Gi")
 	if err != nil {
--- a/test/e2e/cluster.go
+++ b/test/e2e/cluster.go
@@ -120,7 +120,7 @@ func createStatefulSet(ctx context.Context, cli kubernetes.Interface) error {
 	installVersion, _ := version.ParseVersion(*oc.ClusterProfile.Version)
 
 	defaultStorageClass := "managed-premium"
-	if installVersion.V[0] == 4 && installVersion.V[1] == 11 {
+	if installVersion.V[0] == 4 && installVersion.V[1] >= 11 {
 		defaultStorageClass = "managed-csi"
 	}
 	pvcStorage, err := resource.ParseQuantity("2Gi")
--- a/test/e2e/disk_encryption.go
+++ b/test/e2e/disk_encryption.go
@@ -74,9 +74,9 @@ var _ = Describe("Disk encryption at rest", func() {
 		Expect(err).NotTo(HaveOccurred())
 
 		installVersion, _ := version.ParseVersion(*oc.ClusterProfile.Version)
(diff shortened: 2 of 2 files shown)


tiguelu (approved):
LGTM

facchettos (approved):
LGTM, let's get it in once e2e is green

test/e2e/cluster.go
```diff
@@ -120,7 +120,7 @@ func createStatefulSet(ctx context.Context, cli kubernetes.Interface) error {
 	installVersion, _ := version.ParseVersion(*oc.ClusterProfile.Version)
 
 	defaultStorageClass := "managed-premium"
-	if installVersion.V[0] == 4 && installVersion.V[1] == 11 {
+	if installVersion.V[0] == 4 && installVersion.V[1] >= 11 {
```
bennerv, line 123: Any reason we aren't using the `version` package `Lt` functions? (or create a `Gt` function and use that)

Something like

```golang
	if v.Lt(version.NewVersion(4, 11)) {

**Review by jaitaiwan — DISMISSED** [review]
LGTM

File: test/e2e/cluster.go (hunk 1 above)
Comment (mbukatov) [reply test/e2e/cluster.go]: PR updated

**Review by jaitaiwan — DISMISSED** [review]

**Review by bennerv — DISMISSED** [review]

**Review by petrkotas — APPROVED** [review]

**Review by petrkotas — APPROVED** [review]

File: test/e2e/cluster.go [hunk 2]
```diff
@@ -119,9 +119,9 @@ func createStatefulSet(ctx context.Context, cli kubernetes.Interface) error {
 	oc, _ := clients.OpenshiftClusters.Get(ctx, vnetResourceGroup, clusterName)
 	installVersion, _ := version.ParseVersion(*oc.ClusterProfile.Version)
 
-	defaultStorageClass := "managed-premium"
-	if installVersion.V[0] == 4 && installVersion.V[1] == 11 {
-		defaultStorageClass = "managed-csi"
+	defaultStorageClass := "managed-csi"
+	if installVersion.Lt(version.NewVersion(4, 11)) {
+		defaultStorageClass = "managed-premium"
```
AldoFusterTurpin, line 124: Maybe this can be extracted to a function: _getDefaultStorageClass()_.

test/e2e/disk_encryption.go
```diff
@@ -74,9 +74,9 @@ var _ = Describe("Disk encryption at rest", func() {
 		Expect(err).NotTo(HaveOccurred())
 
 		installVersion, _ := version.ParseVersion(*oc.ClusterProfile.Version)
-		defaultStorageClass := "managed-premium-encrypted-cmk"
-		if installVersion.V[0] == 4 && installVersion.V[1] == 11 {
-			defaultStorageClass = "managed-csi-encrypted-cmk"
+		defaultStorageClass := "managed-csi-encrypted-cmk"
+		if installVersion.Lt(version.NewVersion(4, 11)) {
+			defaultStorageClass = "managed-premium-encrypted-cmk"
```
AldoFusterTurpin, line 79: Maybe this can be extracted to a function: _getDefaultStorageClass()_.

AldoFusterTurpin (approved):
LGTM with a minor optional suggestion to better locate future changes depending on the version.

test/e2e/cluster.go
mbukatov: This is a good idea. That said, I would rather do this later during improvement of storage class testing in general (eg.  listing all storage classes for given configuration along with info which one is the default). Moreover I have to admit I'm not sure where would I place this function.

PR #2806, opened 2023-03-27
Remove reliance on hiveclient set which in-turn relies on openshift installer

jaitaiwan:
## Note
This PR will be rebased upon the outcome of https://github.com/Azure/ARO-RP/pull/2900

Fixes ARO-1419

We have many issues with the ARO-RP repository relying on the openshift installer, often pinning us to certain versions of the installer and preventing upgrades. These changes will be one in many steps towards removing the reliance on the installer in the repo.

Important note: You will not see any references to the openshift installer in this PR. This is because it's actually the underlying package for the hiveclientsets that has the dependency there.

### Test plan for issue:
Existing unittests are updated as this is a modification of existing code; a local dev install of hive and running this branch against it is the other way in which this has been tested. Open to further suggestions to include further testing.

No, this is tech debt clean-up.

### Note for reviewers
There's a lot of vendor updates which you can ignore.


jaitaiwan:
Need to add more tests for the dynamic helper

pkg/hive/manager_test.go
```diff
@@ -8,23 +8,38 @@ import (
...
+
+/**
+ * NOTE: I have issue with this...
+ * I really shouldn't be manually stating the type of the group or
+ * specifying its version etc. This is an antipractice imho as I
+ * should be getting this somewhere from the hivev1 library. Not
+ * having it listed there means this is hardcoded and will have to
+ * be manually updated with each version of the library that changes
+ * this thing.
+ */
```
SudoBrendan, line 40: Agreed that it's a little frustrating, but I'm not sure of an alternative. We might consider a quick refactor to add a single function that returns the entire `metav1.TypeMeta` object (so everything becomes `TypeMeta: GetClusterDeploymentTypeMeta()`) so we just have 2-3 loc change each time we'd need to update this, but considering it's a `v1` I'm guessing it'll remain stable for a long while.

pkg/util/dynamichelper/dynamichelper.go
SudoBrendan, line 1: Everything here is really helpful and I think goes in a good direction, though I'm wondering if it could/should be split into a separate PR for separate review since DynamicHelper is used so much through our code. We could then have the Hive-specific stuff come later?

pkg/util/dynamichelper/gvrresolver_test.go
```diff
@@ -195,17 +199,35 @@ func TestFindGVR(t *testing.T) {
...
-	for _, tt := range tests {
+	for index, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			r := &gvrResolver{apiresources: tt.resources}
 
 			got, err := r.Resolve(tt.kind, "")
-			if !reflect.DeepEqual(err, tt.wantErr) {
-				t.Errorf("got: %#v, expected: %#v", err, tt.wantErr)
+			if err != nil {
+				for _, i := range deep.Equal(err.Error(), tt.wantErr.Error()) {
```
SudoBrendan, line 208: I'm noticing this was code that existed before, but `deep.Equal` or `reflect.DeepEqual`, etc on type Error is not best practice, linters typically fail on stuff like this: https://github.com/golang/lint/issues/232

SudoBrendan (dismissed):
Great work on this so far. I think splitting out the DynamicHelper changes to a new PR would facilitate better reviews of those changes since they are more widespread.

pkg/util/dynamichelper/gvrresolver_test.go
jaitaiwan: Thanks for the feedback! What's the recommended method for truly fixing this?

pkg/hive/manager_test.go
jaitaiwan: Sounds like a good point, that should make the tests more concise.

pkg/util/dynamichelper/dynamichelper.go
jaitaiwan: The hive changes can't be done without the dynamic helper having the new `GetResource` function. From there the rest of the changes are all just documenting what's there which I did as a step towards understanding the code, refactoring the mock as a part of testing GetResource in a standardised way and the rest is history haha. If you think I should still pull out the dynamichelper / GVRResolver changes as it's own PR and base the hive changes off of that one, happy to do that.

jaitaiwan:
I'll fix linting errors once I have confirmation on whether or not I need to split this PR out into dynamichelper and hive pkg changes.

pkg/util/dynamichelper/dynamichelper.go
jaitaiwan: I've pulled the changes out: https://github.com/Azure/ARO-RP/pull/2900

pkg/util/dynamichelper/gvrresolver_test.go
SudoBrendan: That's discussed in the PR I linked :)

pkg/hive/manager.go
```diff
@@ -22,6 +21,7 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/dynamichelper"
 	utillog "github.com/Azure/ARO-RP/pkg/util/log"
 	"github.com/Azure/ARO-RP/pkg/util/uuid"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
bennerv, line 24: our imports in files are structured as follows: 
* system library imports
* third party imports 
* first party imports (github.com/Azure/ARO-RP)

So the bottom line here needs to move up above to the third party import line.

pkg/hive/manager.go
```diff
@@ -225,7 +234,7 @@ func (hr *clusterManager) ResetCorrelationData(ctx context.Context, doc *api.Ope
 			return err
 		}
 
-		_, err = hr.hiveClientset.HiveV1().ClusterDeployments(doc.OpenShiftCluster.Properties.HiveProfile.Namespace).Update(ctx, cd, metav1.UpdateOptions{})
+		err = hr.hiveClientset.Update(ctx, cd)
 		return err
```
bennerv, line 238: nit
```suggestion
		return hr.hiveClientset.Update(ctx, cd)

bennerv (changes requested):
good first pass - much simpler than the gvr stuff :)

pkg/hive/manager.go
jaitaiwan: Thank you, the error messaging around this, was a lot less helpful.

bennerv (approved):
lgtm - good work

bennerv:
@jaitaiwan are we removing all occurrances of hive clientset?  Is that the point of this PR, or just in this one location?

bennerv (approved):
-7.6k :shipit:  lgtm

hawkowl (approved):

PR #2811, opened 2023-03-29
Allow dnsmasq caching config to be modified

ventifus:

Fixes https://issues.redhat.com/browse/ARO-2427. Introduces a new field to ClusterSpec `dnsmasq` that allows configuring of dnsmasq cache-size, min-cache-ttl, max-cache-ttl and no-negcache.

Depends on 
- release-4.10-azure: https://github.com/jewzaam/installer-aro/pull/20
- release-4.11-azure: https://github.com/jewzaam/installer-aro/pull/19

If a query is forwarded by dnsmasq is lost, then dnsmasq will hold on to that query for 40 seconds waiting for a response. If additional queries for that same name (and type) come in during that time, they will block on waiting for the first forwarded query. To reduce the number of upstream requests needed, this PR allows enabling and tuning caching.

### Test plan for issue:

Deploy to a test cluster. Let it run for a while (1 day or more) and check CoreDNS logs for "i/o timeout". **PENDING**

Install the new aro operator and make sure it doesn't update any machineconfig :heavy_check_mark: **PASS**

Test that each setting updates the machineconfig appropriately. Also verify the resulting dnsmasq.conf config file is formatted coorectly and dnsmasq loads.
  - cacheSize  :heavy_check_mark: **PASS**
  - maxCacheTTL  :heavy_check_mark: **PASS**
  - minCacheTTL :heavy_check_mark: **PASS**
  - noNegCache  :heavy_check_mark: **PASS**

Internal documentation needs to be created detailing how to configure dnsmasq caching, when to enable it, and how to tune it for workload.
Once we have this deployed in the field solving real customer issues, we will consider enabling cache by default. Data will need to be collected to determine how big a cache to configure; the dnsmasq upstream default is 150 entries.


dem4gus:
Were the changes in the vendor folder made upstream and are being incorporated here or are we just changing the vendor files directly?

ventifus:
@dem4gus I'm changing the vendor files directly only to make it easier to test and review. I'll split out those changes to aro-installer once I'm sure this approach will work and I'm ready to take this out of draft.

go.mod
```diff
@@ -456,7 +456,7 @@ replace (
 	github.com/openshift/cluster-api-provider-libvirt => github.com/openshift/cluster-api-provider-libvirt v0.2.1-0.20191219173431-2336783d4603
 	github.com/openshift/cluster-api-provider-ovirt => github.com/openshift/cluster-api-provider-ovirt v0.1.1-0.20211215231458-35ce9aafee1f
 	github.com/openshift/console-operator => github.com/openshift/console-operator v0.0.0-20220318130441-e44516b9c315
-	github.com/openshift/installer => github.com/jewzaam/installer-aro v0.9.0-master.0.20220524230743-7e2aa7a0cc1a
+	github.com/openshift/installer => github.com/ventifus/installer-aro v0.0.0-20230404230942-a99152bf4d63
```
s-amann, line 459: should this get reverted? or are you the new owner of `installer-aro`?

pkg/operator/deploy/deploy.go
```diff
@@ -219,14 +219,17 @@ func (o *operator) resources() ([]kruntime.Object, error) {
...
-			Domain:                 domain,
-			ACRDomain:              o.env.ACRDomain(),
-			AZEnvironment:          o.env.Environment().Name,
-			Location:               o.env.Location(),
-			InfraID:                o.oc.Properties.InfraID,
-			ArchitectureVersion:    int(o.oc.Properties.ArchitectureVersion),
-			VnetID:                 vnetID,
-			StorageSuffix:          o.oc.Properties.StorageSuffix,
+			DnsmasqConfig: arov1alpha1.DnsmasqConfigSpec{
+				CacheSize: 0,
```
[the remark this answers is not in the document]
ventifus: Added

go.mod
ventifus: The intent here was to make CI and such pass. This won't be needed and I'll revert once those changes are merged into installer-aro.

pkg/operator/deploy/deploy.go
```diff
@@ -219,14 +219,20 @@ func (o *operator) resources() ([]kruntime.Object, error) {
...
-			Domain:                 domain,
-			ACRDomain:              o.env.ACRDomain(),
-			AZEnvironment:          o.env.Environment().Name,
-			Location:               o.env.Location(),
-			InfraID:                o.oc.Properties.InfraID,
-			ArchitectureVersion:    int(o.oc.Properties.ArchitectureVersion),
-			VnetID:                 vnetID,
-			StorageSuffix:          o.oc.Properties.StorageSuffix,
+			DnsmasqConfig: arov1alpha1.DnsmasqConfigSpec{
+				CacheSize:   0,
```
ellis-johnson, line 223: Are these just being set as 0 as a default value?

SudoBrendan: +1 here, I feel like these should be interpreted from some kind of higher-level config, that we can modify without code changes, right? (envvars, RP-Config, etc)

SudoBrendan (changes requested):
A great start! I still have open questions about caching in general and if that fits into our ARO strategy moving into the upstream installer, I'll let Ben comment, since I'm a little removed from the intricacies. If that aligns, the only change I'd like to see is some kind of configuration of DNSMasq that we can modify/rollout/revert with a config release rather than a code change.

bennerv:
> I'd be curious to know if @bennerv sees any issues with us adding caching to dns as it relates to installer changes? I know that in a recent openshift/installer sync, we had mentioned that DNSMasq is a non-caching DNS server (ref: https://docs.google.com/presentation/d/1axmIrS3dyzAyGlLcJOly5vPFv1uymvu-jOy2uATYi_E/edit#slide=id.g136f68161c2_0_118), would this implementation change our ability to use upstream in the future?

Didn't get a chance to address this yet, but it's been in the back of my mind.  

Here's my issue with the caching; CoreDNS already leverages caching.  So by adding caching at the top-level (dnsmasq), we're introducing a chain of caches which can result in unexpected behavior from a DNS side.  If a DNS record is updated, the TTL on the DNS record isn't honored from Corefile, and then there's also caching at dnsmasq level.  

The reason dnsmasq was never meant to cache things is that we wanted the normal behavior to happen with the addition of dnsmasq.  Unfortunately it has had some unexpected consequences.  I would rather use the fast retries option if possible, but it seems that's not the case quite yet, and won't be unless we ask the RHEL team to backport.  I don't see this as an option short-term.  

I'm not sure if it's possible to force dnsmasq to use TCP to make it more reliable when querying upstream DNS servers.  That might be one possibility, otherwise we're stuck with this without using a different resolver or waiting for fast retries.




```
$ oc -n openshift-dns get cm dns-default  -o yaml
apiVersion: v1
data:
  Corefile: |
    .:5353 {
        bufsize 512
        cache 900 {
            denial 9984 30
        }
```

vendor/modules.txt
```diff
@@ -878,7 +878,7 @@ github.com/openshift/hive/apis/hive/v1/ovirt
 github.com/openshift/hive/apis/hive/v1/vsphere
 github.com/openshift/hive/apis/hiveinternal/v1alpha1
 github.com/openshift/hive/apis/scheme
-# github.com/openshift/installer v0.16.1 => github.com/jewzaam/installer-aro v0.9.0-master.0.20220524230743-7e2aa7a0cc1a
+# github.com/openshift/installer v0.16.1 => github.com/ventifus/installer-aro v0.0.0-20230404230942-a99152bf4d63
```
jhoreman, line 881: Assuming these should also be reverted as part of the above comment chain.

PR #2812, opened 2023-03-30
Add the registry arg for the fluentbit dockerfile

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-1384

Docker is a bit more restrictive than Podman in terms of args. For building, we'll need this.

### Test plan for issue:
Will test in CI/building
N/A


ellis-johnson (approved):
LGTM!

PR #2813, opened 2023-03-30
Bump the 4.10 version to latest stable

bennerv:

This version should include the fix for: https://access.redhat.com/solutions/6844091.  Additionally, bumps the default install version to latest stable 4.10 version.

Note: we have to maintain the 4.10.40 version as a valid install target until the RP is rolled out.  We have to do that since we need to run the `update_ocp_version` pipeline first.  We have to run that first because there's a discrepency in the way the backend and frontend handle the default version.

* The frontend will always ensure the default version is available for consumption, even if it's not present in the database
* The backend will allow the default version as a valid target if and only if there are no versions in the database, therefore since there are versions in the database already, we have to maintain "backwards" compatibility until it's fixed.  

### Test plan for issue:

e2e tests

Nope


cadenmarchese (approved):

bennerv:
We don't actually have to run update_ocp_versions first, but still probably should if https://github.com/Azure/ARO-RP/pull/2743 is merged in.

petrkotas (approved):
LGTM

PR #2814, opened 2023-03-30
Change mariner to pull from MCR

hawkowl:

Part of https://issues.redhat.com/browse/ARO-1384

We don't need to clone from MCR for Mariner images/etc that are needed in EV2, since they are fine to pull directly. We still need to clone images that might be used in clusters or are on registry.redhat (e.g. go-toolset, or azure-cli from MCR), but for base images from MSFT the first party repo is supposed to be used.

### Test plan for issue:

Will be testing by building

N/A


bennerv (approved):

ellis-johnson (approved):
LGTM

PR #2815, opened 2023-03-30
Cluster resource tagging

kimorris27:
**Important**: As of the time of opening, this PR includes the contents of https://github.com/Azure/ARO-RP/pull/2768 (cluster RG tagging implementation). A lot of the code in this PR depends on that one, but that one hasn't been merged yet.

---

https://issues.redhat.com/browse/ARO-2169

This PR implements tagging for the cluster resources and allows both install time tagging and day 2 updates. It uses a combination of policy-based tagging and "manual" tagging (e.g. iterating over the resources in the cluster RG and adding tags to each of them).

High-level summary of changes made in this PR:
- Updated `pkg/cluster` to create our tagging policy *definition* in the customer subscription at cluster creation time and "ensure" it on each cluster update. There is one policy definition per cluster if the customer has multiple clusters in their subscription.
- Updated `pkg/cluster` to create and "ensure" the policy *assignment* at the scope of the cluster RG. The assignment fills in the policy parameters with the contents of the `ResourceTags` API field such that those tags will be placed on any resources that are newly created within the cluster RG by the policy's [modify effect](https://learn.microsoft.com/en-us/azure/governance/policy/concepts/effects#modify).
- Adding on to the last bullet point: updating the policy assignment parameters post-creation does not automatically remediate resource tags. So I've also added the `remediateTags()` function, which runs on each cluster update and reconciles the tags. I was originally planning to use Azure Policy [remediation tasks](https://learn.microsoft.com/en-us/azure/governance/policy/how-to/remediate-resources?tabs=azure-portal) instead of iterating over the cluster resources like this, but I found that remediation tasks are extremely slow (>= 20 minutes) compared to doing it this way.
- The policy assignment takes some time to come into effect; it's around 2 minutes from what I found while I was working on this PR. In the case where the customer has a policy on their subscription that denies resource creation if missing tags, I found that cluster installation fails at the `deployStorageTemplate()` step because the policy assignment isn't actively working yet. So I updated the ARM template deployments in `pkg/cluster` and `pkg/installer` to explicitly add the `ResourceTags` that the customer has passed. This "hack" allows us to avoid adding an intentional delay to the cluster installation process.
- The policy assignment includes a managed identity with the `TagContributor` built-in Azure role. This is so that customers can manually trigger tag remediation tasks if they'd like to. They can also remediate their tags our way with an `az aro update` though.

There is a lot going on here, and I've probably missed a detail (or a few), so please ask questions.

### Test plan for issue:

I've included unit tests and E2E tests that validate both the "manual" tagging and the policy-based tagging added in this PR, but I did quite a bit of ad-hoc testing as well. Since the policy assignments are somewhat laggy, the goal of my ad-hoc testing was to find out what amount of time is needed for them to be fully active and make sure that this lag doesn't cause weird/unexplained tag behavior or cluster installation failures (for customers with policies that deny resource creation if missing tags). To this end, I did all of my ad-hoc tests with such a creation denial policy in place.

#### Ad-hoc testing summary by installer

**Integrated installer**: Since the storage deployment ARM template and integrated ARM template both explicitly tag their resources, there is about 3-5 minutes between when the policy assignment is created and the time when the policy assignment *must* be working correctly (the time when the installer creates the worker VMs). So there's naturally enough of a delay present in the cluster installation process. I only tested this case once since the integrated installer is no longer used in production.

**Hive installation**: There is less of a natural delay here than in the case of the integrated installer, but it seems to be enough. I found that there is about 2-3 minutes between when the policy assignment is created and the installer starts deploying resources, which aren't explicitly tagged and thus depend on the policy assignment to be fully active. I was concerned that 2 minutes wasn't enough of a delay, but I did > 10 cluster installations and didn't have any failures. We could consider adding an intentional delay of 1-3 minutes if we wanted to reduce risk.

#### Note about E2E tests

The tagging E2E tests take about 15 minutes because they have to account for the policy assignment lag... if I removed the intentional delays, I wouldn't be able to effectively test the policy. Maybe we need a flag to turn these tests on and off so they don't have to run every time?

Yes, I'll need to update customer-facing documentation but haven't done it yet.


jaitaiwan:
I intend to review this once the parent PR is merged.

kimorris27:
Closing following change of approach; will open a separate new PR when ready.

PR #2817, opened 2023-03-30
Run local development and PR E2E in podman

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-2532

Runs local development and PR E2E in podman, instead of directly. Removes the remaining Installer dependencies.

### Test plan for issue:

Local testing, E2E.

Some local dev documentation likely needs revising.


cmd/aro/main.go
```diff
@@ -44,11 +44,6 @@ func main() {
 	audit := utillog.GetAuditEntry()
 	log := utillog.GetLogger()
 
-	// TODO: Use `azuretypes.Platform.IsARO()` from github.com/openshift/installer/pkg/types/azure
-	if !platformIsAro {
```
cadenmarchese, line 48: I don't have a good understanding of why we're removing this tag. It looks like it was originally used to disable managed identity. Do we disable managed identity somewhere else now?

cadenmarchese (commented):
I have some questions on it, but things are making sense to me!

.pipelines/e2e.yml
```diff
@@ -23,13 +22,19 @@ variables:
...
+  variables:
+      ARO_PODMAN_SOCKET: "tcp://localhost:8888"
   pool:
     name: 1es-aro-ci-pool
   #services:
   #  selenium: selenium
   steps:
   - template: ./templates/template-checkout.yml
 
+  - script: |
```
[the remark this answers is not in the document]
hawkowl: this is the old approach :) you might need to refresh

cmd/aro/main.go
hawkowl: This tag controls code in the vendored Installer. Since that code no longer exists, we can remove this. It remains in openshift/ARO-Installer.

pkg/containerinstall/install.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	_, err := containers.Remove(m.conn, containerName, &containers.RemoveOptions{Force: to.BoolPtr(true), Ignore: to.BoolPtr(true)})
+	if err != nil {
+		m.log.Errorf("unable to remove container: %v", err)
+	}
+
+	for _, secretName := range []string{"99_aro.json", "99_sub.json", "proxy.crt", "proxy-client.crt", "proxy-client.key"} {
+		err = secrets.Remove(m.conn, m.doc.ID+"-"+secretName)
+		if err != nil {
```
facchettos, line 250: would it make sense to not log anything if the error is `isNotFound` or something?

pkg/containerinstall/install.go
```diff
@@ -0,0 +1,255 @@
...
+			return true, fmt.Errorf("container exited with %d", inspectData.State.ExitCode)
+		} else {
+			m.success = true
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (m *manager) writeFiles(ctx context.Context) error {
```
facchettos, line 175: is the naming right here? half the time we won't be writing any files anywhere

pkg/containerinstall/install.go
```diff
@@ -0,0 +1,255 @@
...
+		err = m.secretFile(filepath.Join(basepath, "secrets/proxy-client.key"), "proxy-client.key")
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (m *manager) secretFile(from, name string) error {
```
facchettos, line 225:
```suggestion
func (m *manager) secretFromFile(from, name string) error {
```

pkg/containerinstall/podman.go
```diff
@@ -0,0 +1,76 @@
...
+
+	go func() {
+		for {
+			v, ok := <-stderr
+			if !ok {
+				return
+			}
+			log.Errorf("stderr: %s", v)
+		}
+	}()
```
facchettos, line 53: these would be better with a range over the channel instead of checking the ok value

pkg/containerinstall/podman.go
```diff
@@ -0,0 +1,76 @@
...
+	err := containers.Logs(ctx, containerName, nil, stdout, stderr)
+	return err
+}
+
+func pullContainer(ctx context.Context, pullspec string, options *images.PullOptions) error {
+	_, err := images.Pull(ctx, pullspec, options)
+	return err
+}
```
facchettos, line 61: is that really useful? I'm not really sure I see an added value for that function

pkg/containerinstall/podman.go
```diff
@@ -0,0 +1,76 @@
...
+
+func runContainer(ctx context.Context, log *logrus.Entry, s *specgen.SpecGenerator) (string, error) {
+	createResponse, err := containers.CreateWithSpec(ctx, s, nil)
```
facchettos, line 64: not really sure I understand the naming here. couldn't it be called `container` or something? it doesn't create a response or even is a response? Or have I missed something?

pkg/containerinstall/install.go
```diff
@@ -0,0 +1,255 @@
...
+		"AZURE_RP_CLIENT_SECRET",
+		"AZURE_SUBSCRIPTION_ID",
+		"AZURE_TENANT_ID",
+		"DOMAIN_NAME",
+		"KEYVAULT_PREFIX",
+		"LOCATION",
+		"PROXY_HOSTNAME",
+		"PULL_SECRET",
+		"RESOURCEGROUP",
+	}
```
facchettos, line 43: could we make this a map instead? That'd allow us to give names to the env variables instead of just appending numbers at the end of the env variables

pkg/containerinstall/manager.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/env"
+)
+
+type ContainerInstaller interface {
+	Install(ctx context.Context, sub *api.SubscriptionDocument, doc *api.OpenShiftClusterDocument, version *api.OpenShiftVersion) error
+}
```
facchettos, line 17: Could we maybe add a setup function or something? Because getting the env variables directly inside the install function makes it tightly bound with the environment

facchettos (dismissed):
I left some suggestions but overall that's awesome

hack/update-go-module-dependencies.sh
```diff
@@ -27,46 +27,35 @@
...
 	esac
 done
 
-for x in aws azure openstack; do
-	go mod edit -replace sigs.k8s.io/cluster-api-provider-$x=$(go list -mod=mod -m github.com/openshift/cluster-api-provider-$x@release-4.10 | sed -e 's/ /@/')
+for x in vendor/k8s.io/*; do
+	case $x in
+	# Review the list of special cases on each release.
+
+	# This needs to be something different
```
petrkotas, line 45: What it needs to be than?

petrkotas (dismissed):
Please split the PR and separate the context of work. I see 

- update of e2e and CI
- removing of vendored installer
- unrelated whitespace changes

By splitting into multiple PRs, this work can be merged faster

pkg/containerinstall/install.go
hawkowl: I'm actually appending the env variable after "ARO_", no numbers in any of it :)

pkg/containerinstall/install.go
hawkowl: I set it to debug for now

pkg/containerinstall/podman.go
hawkowl: I copied what podman does directly, I am not super up on channels and such

pkg/containerinstall/install.go
```diff
@@ -0,0 +1,255 @@
...
+
+	if inspectData.State.Status == "exited" || inspectData.State.Status == "stopped" {
+		if inspectData.State.ExitCode != 0 {
+			getContainerLogs(m.conn, m.log, containerName)
+			return true, fmt.Errorf("container exited with %d", inspectData.State.ExitCode)
+		} else {
+			m.success = true
+			return true, nil
+		}
```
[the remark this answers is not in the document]
hawkowl: done

pkg/containerinstall/podman.go
hawkowl: yeah it used to do more and I did some refactoring, I've elided it completely now

pkg/containerinstall/podman.go
hawkowl: done,

pkg/containerinstall/install.go
facchettos: ah right, always confuse key and value stuff ><

pkg/containerinstall/podman.go
facchettos: this is weird, because this is essentially reimplementing the range functionality 
```suggestion
	go func() {
		for v := range stdout {
			log.Infof("stdout: %s", v)
		}
	}()

	go func() {
		for v := range stderr {
			log.Errorf("stderr: %s", v)
		}
	}()
```

pkg/containerinstall/manager.go
hawkowl: I tweaked this a bit more, but since it all comes from the dev environment it's a bit of a hassle to do super nicely unless we were going to use it in not-dev

pkg/containerinstall/manager.go
```diff
@@ -0,0 +1,55 @@
...
+
+type manager struct {
+	conn context.Context
+	log  *logrus.Entry
+	env  env.Interface
+
+	clusterUUID string
+	pullSecret  [2]string
```
facchettos, line 26: why do we use an array here? it's seems like a struct or having the username and password as fields inside the manager would make the code easier to understand.

facchettos (changes requested):

pkg/containerinstall/manager.go
hawkowl: cleaned this up

SudoBrendan (approved):
This is all making sense to me, can't wait!!

pkg/containerinstall/install.go
```diff
@@ -0,0 +1,216 @@
...
+				Username: to.StringPtr(m.pullSecret.Username),
+				Password: to.StringPtr(m.pullSecret.Password),
+			}
+
+			_, err := images.Pull(m.conn, version.Properties.InstallerPullspec, options)
+			return err
+		}),
+		steps.Action(func(context.Context) error { return m.createSecrets(ctx, doc, sub) }),
+		steps.Action(func(context.Context) error { return m.startContainer(ctx, version) }),
+		steps.Condition(m.containerFinished, 60*time.Minute, false),
```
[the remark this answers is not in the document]
hawkowl: except in eastus2 :')

.golangci.yml
```diff
@@ -5,6 +5,12 @@ run:
     - vendor
   skip-dirs-use-default: true
   modules-download-mode: vendor
+  build-tags:
+    - "aro"
+    - "containers_image_openpgp"
+    - "exclude_graphdriver_devicemapper"
+    - "exclude_graphdriver_btrfs"
```
facchettos, line 12: does this have a reason to be in this PR? or is it just some extra change bundled in it?

Makefile
facchettos, line 1: wondering if these changes belong in this pr too

pkg/util/pullsecret/extract_test.go
facchettos, line 1: why do we use ginkgo and gomega here? these are standard unit tests

pkg/containerinstall/install_test.go
facchettos, line 1: can you explain why ginkgo and gomega are needed here? a comment would be nice in the file, since we are only using those in the e2e and not in `pkg`

facchettos (changes requested):

.golangci.yml
hawkowl: Podman's bindings vendors in more of the image lib that causes build dep side effects.

pkg/util/pullsecret/extract_test.go
hawkowl: We use gomega in some other places, and I've got mildly sick of dealing with stdlib golang testing, so I just wrote it in ginkgo instead.

Makefile
hawkowl: As above, these flags need to be changed. I decided to clean it up a smidge, since it will make the diff when the aro build tag is removed a little bit smaller.

pkg/util/pullsecret/extract_test.go
facchettos: what about ginkgo? We only use it in e2e, we should have some consistency, we should either use it everywhere in `pkg` or not at all

PR #2819, opened 2023-04-03
Improve error check on failed RP deployment

kimorris27:

Recent deployment issues we saw in INT when trying out some changes to deploy.go.

- ~~Adds `UpdateVMSSProbes()` back in with corrections to the code that was checking for the `CannotModifyProbeUsedByVMSS` error.~~
- Makes the check for the already familiar `ResourceNotFound` error with the load balancer more robust.
- Adds nil checks before checking VMSS health statuses

These changes are needed to make the deployments more robust to failures. They should also make it easier to diagnose the root cause when we run into other types of failures.

### Test plan for issue:

I wrote a script to duplicate the known LB `ResourceNotFound` error so that I could look at the structure of the Azure `ServiceError` and know how to get to the errors nested inside. I updated the unit tests accordingly.

No

### DIFF
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -198,7 +198,7 @@ func (d *deployer) deploy(ctx context.Context, rgName, deploymentName, vmssName 
 }
 
 // checkForKnownError is a helper function that checks the errors nested within an Azure ServiceError
-// for a known error and returns the corresponding KnownDeploymentError if applicable.
+// for a known error and returns the corresponding KnownDeploymentErrorType if applicable.
 func (d *deployer) checkForKnownError(serviceErr *azure.ServiceError, deployAttempt int) (KnownDeploymentErrorType, error) {
 	if serviceErr.Code == "DeploymentFailed" && len(serviceErr.Details) > 0 {
 		outerErr := azure.ServiceError{}
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -158,28 +158,26 @@ func (d *deployer) deploy(ctx context.Context, rgName, deploymentName, vmssName 
 		err = d.deployments.CreateOrUpdateAndWait(ctx, rgName, deploymentName, deployment)
 		serviceErr, isServiceError := err.(*azure.ServiceError)
 
-		// Check for two known errors that we know how to handle.
+		// Check for a known error that we know how to handle.
 		if isServiceError && serviceErr.Code == "DeploymentFailed" && len(serviceErr.Details) > 0 {
-			outerCode, outerCodeOk := serviceErr.Details[0]["code"].(string)
-			message, messageOk := serviceErr.Details[0]["message"].(string)
+			outerErr := azure.ServiceError{}
+			jsonEncoded, _ := json.Marshal(serviceErr.Details[0])
+			json.Unmarshal(jsonEncoded, &outerErr)
 
-			if outerCodeOk && messageOk && outerCode == "BadRequest" {
-				innerErr := map[string]interface{}{}
+			innerErr := azure.ServiceError{}
+			json.Unmarshal([]byte(outerErr.Message), &innerErr)
 
-				if jsonErr := json.Unmarshal([]byte(message), &innerErr); jsonErr == nil {
-					innerCode, innerCodeOk := innerErr["code"].(string)
-					innerMessage, innerMessageOk := innerErr["message"].(string)
+			isKnownLBError := i < 1 && innerErr.Code == "ResourceNotFound" && strings.Contains(innerErr.Message, "Microsoft.Network/loadBalancers/rp-lb")
 
-					if i < 1 && innerCodeOk && innerMessageOk && innerCode == "ResourceNotFound" && strings.Contains(innerMessage, "Microsoft.Network/loadBalancers/rp-lb") {
-						// on new RP deployments, we get a spurious DeploymentFailed error
-						// from the Microsoft.Insights/metricAlerts resources indicating
-						// that rp-lb can't be found, even though it exists and the
-						// resources correctly have a dependsOn stanza referring to it.
-						// Retry once.
-						d.log.Print(err)
-						continue
-					}
-				}
+			// On new RP deployments, we get a spurious DeploymentFailed error
+			// from the Microsoft.Insights/metricAlerts resources indicating
+			// that rp-lb can't be found, even though it exists and the
+			// resources correctly have a dependsOn stanza referring to it.
+			// Retry once, and only if this error is encountered on the first
+			// deployment attempt.
+			if isKnownLBError {
+				d.log.Print(err)
+				continue
 			}
 		}
 
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -200,33 +200,35 @@ func (d *deployer) deploy(ctx context.Context, rgName, deploymentName, vmssName 
 // checkForKnownError is a helper function that checks the errors nested within an Azure ServiceError
 // for a known error and returns the corresponding KnownDeploymentErrorType if applicable.
 func (d *deployer) checkForKnownError(serviceErr *azure.ServiceError, deployAttempt int) (KnownDeploymentErrorType, error) {
-	if serviceErr.Code == "DeploymentFailed" && len(serviceErr.Details) > 0 {
-		outerErr := azure.ServiceError{}
-		jsonEncoded, err := json.Marshal(serviceErr.Details[0])
+	if serviceErr.Code != "DeploymentFailed" || len(serviceErr.Details) == 0 {
+		return "", nil
+	}
 
-		if err != nil {
-			return "", err
-		}
+	outerErr := azure.ServiceError{}
+	jsonEncoded, err := json.Marshal(serviceErr.Details[0])
 
-		err = json.Unmarshal(jsonEncoded, &outerErr)
+	if err != nil {
+		return "", err
+	}
 
-		if err != nil {
-			return "", err
-		}
+	err = json.Unmarshal(jsonEncoded, &outerErr)
 
-		innerErr := azure.ServiceError{}
-		err = json.Unmarshal([]byte(outerErr.Message), &innerErr)
+	if err != nil {
+		return "", err
+	}
 
-		if err != nil {
-			return "", err
-		}
+	innerErr := azure.ServiceError{}
+	err = json.Unmarshal([]byte(outerErr.Message), &innerErr)
 
-		isFirstAttempt := deployAttempt < 1
-		isRPLBNotFound := innerErr.Code == "ResourceNotFound" && strings.Contains(innerErr.Message, "Microsoft.Network/loadBalancers/rp-lb")
+	if err != nil {
+		return "", err
+	}
 
-		if isFirstAttempt && isRPLBNotFound {
-			return KnownDeploymentErrorTypeRPLBNotFound, nil
-		}
+	isFirstAttempt := deployAttempt < 1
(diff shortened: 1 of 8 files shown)


kimorris27:
@facchettos , would appreciate your review on this since you were originally working on addressing the issue with the health probes. 🙂

pkg/deploy/upgrade_gateway.go
```diff
@@ -41,7 +41,8 @@ func (d *deployer) gatewayWaitForReadiness(ctx context.Context, vmssName string)
 	return wait.PollImmediateUntil(10*time.Second, func() (bool, error) {
 		for _, vm := range scalesetVMs {
 			r, err := d.vmssvms.GetInstanceView(ctx, d.config.GatewayResourceGroupName, vmssName, *vm.InstanceID)
-			if err != nil || *r.VMHealth.Status.Code != "HealthState/healthy" {
+			canGetVMStatus := r.VMHealth != nil && r.VMHealth.Status != nil && r.VMHealth.Status.Code != nil
+			if err != nil || canGetVMStatus && *r.VMHealth.Status.Code != "HealthState/healthy" {
```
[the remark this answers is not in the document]
cadenmarchese: Same suggestion would be applied to lines 44-45 if you agree.

cadenmarchese (dismissed):
This looks good to me and should ensure that we're checking exactly what we're looking for during deployment. Thanks! 👍

pkg/deploy/upgrade_gateway.go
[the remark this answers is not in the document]
kimorris27: That makes sense to me. I went ahead and changed it.

cadenmarchese (dismissed):
LGTM, would like others to weigh in as well.

pkg/deploy/deploy.go
```diff
@@ -158,20 +159,42 @@ func (d *deployer) deploy(ctx context.Context, rgName, deploymentName, vmssName
...
+					}
+				}
+			}
 		}
+
+		// For errors we don't know how to handle, delete the failed VMSS and retry the deployment.
 		if err != nil && *d.config.Configuration.VMSSCleanupEnabled {
 			if retry := d.vmssCleaner.RemoveFailedNewScaleset(ctx, rgName, vmssName); retry {
-				continue // Retry deployment after deleting failed VMSS.
+				continue
```
facchettos, line 197: why is the comment removed here, is it not applicable anymore?

pkg/deploy/deploy.go
```diff
@@ -158,20 +159,42 @@ func (d *deployer) deploy(ctx context.Context, rgName, deploymentName, vmssName
...
+						continue
+					} else if innerCodeOk && innerCode == "CannotModifyProbeUsedByVMSS" {
+						// removes the probe reference so we can update the lb rules
+						if retry := d.vmssCleaner.UpdateVMSSProbes(ctx, rgName); retry {
+							continue
+						}
+					}
+				}
+			}
 		}
```
facchettos, line 192: that is a lot of nesting in here. it'd be great to flatten that out

pkg/deploy/deploy.go
```diff
@@ -158,20 +159,42 @@ func (d *deployer) deploy(ctx context.Context, rgName, deploymentName, vmssName
...
+
+		// Check for two known errors that we know how to handle.
+		if isServiceError && serviceErr.Code == "DeploymentFailed" && len(serviceErr.Details) > 0 {
+			outerCode, outerCodeOk := serviceErr.Details[0]["code"].(string)
+			message, messageOk := serviceErr.Details[0]["message"].(string)
+
+			if outerCodeOk && messageOk && outerCode == "BadRequest" {
+				innerErr := map[string]interface{}{}
+
+				if jsonErr := json.Unmarshal([]byte(message), &innerErr); jsonErr == nil {
```
facchettos, line 172: i'm not sure this helps with readability here, although this is purely cosmetic

pkg/deploy/deploy.go
```diff
@@ -158,20 +159,42 @@ func (d *deployer) deploy(ctx context.Context, rgName, deploymentName, vmssName
...
+						// resources correctly have a dependsOn stanza referring to it.
+						// Retry once.
+						d.log.Print(err)
+						continue
+					} else if innerCodeOk && innerCode == "CannotModifyProbeUsedByVMSS" {
+						// removes the probe reference so we can update the lb rules
+						if retry := d.vmssCleaner.UpdateVMSSProbes(ctx, rgName); retry {
+							continue
+						}
+					}
```
facchettos, line 189: why is it needed to add that check again? We are not modifying the probes (at least not yet), and so that effectively isn't ever going to be used until we decide to move the probe to a different port or modify the path (which isn't super likely), until then I am not sure that we want to have unused code (this is why I removed it completely and didn't let that piece of code in the PR that I reverted)

pkg/deploy/vmsscleaner/clean.go
```diff
@@ -62,3 +64,30 @@ func (c *cleaner) RemoveFailedNewScaleset(ctx context.Context, rgName, vmssToDel
...
+		vmss.VirtualMachineProfile.NetworkProfile.HealthProbe = nil
+		c.log.Printf("removing the probe reference from the vmss %s", name)
+		err = c.vmss.CreateOrUpdateAndWait(ctx, rgName, name, vmss)
+		if err != nil {
+			c.log.Warn(err)
+			return false // If update failed, gateway vmss still exists. Don't retry.
+		}
+	}
+	// no scaleset matched, so we should not retry and return the error
+	return false
```
facchettos, line 92: here we are supposed to retry actually, this was a mistake when I first wrote that function (I think there was another commit that fixed it). Because if no scaleset is matched, we would want to recreate one, and if the update succeeded we would also want to retry

facchettos (changes requested):

facchettos:
before we move forward, I think this needs a deeper discussion because we need a proper strategy if we ever decide to allow the probe to be changed. The problem is that the gateway has to be up all the time to allow for egress lockdown to work, and when we remove the probe from the vmss and loadbalancer, there will be some downtime which potentially will bring some down time because some of the gateways will be unresponsive for a while when updated and since they would have no probe, the load balancer probably won't know which ones are ready.

facchettos:
before we move forward, we should probably have a deeper discussion if we indeed plan on modifying the probes, since an update to the probe can lead to some downtime in the gateway which is required to be up for egress lockdown to work. Having the deployment failing if a probe is modified might be a safeguard in the end.

pkg/deploy/deploy.go
kimorris27: I reworded it for a bit more clarity and moved it to be above the outer if block 2 lines up from here.

pkg/deploy/deploy.go
kimorris27: Any recommendations on how you'd flatten it out from here? I think the way I've done it lines up with the logical steps you have to go through to get to the innermost error (check if outermost is `DeploymentFailed` -> check if next is `BadRequest` -> unmarshal the JSON-encoded innermost one -> check the contents), so I'm not sure how to flatten it further.

Maybe I could just add a comment above each `if` statement making it a bit faster to read over the levels and understand what each one means?

pkg/deploy/deploy.go
```diff
@@ -158,20 +159,42 @@ func (d *deployer) deploy(ctx context.Context, rgName, deploymentName, vmssName
...
-			continue
+		serviceErr, isServiceError := err.(*azure.ServiceError)
+
+		// Check for two known errors that we know how to handle.
+		if isServiceError && serviceErr.Code == "DeploymentFailed" && len(serviceErr.Details) > 0 {
+			outerCode, outerCodeOk := serviceErr.Details[0]["code"].(string)
+			message, messageOk := serviceErr.Details[0]["message"].(string)
+
+			if outerCodeOk && messageOk && outerCode == "BadRequest" {
+				innerErr := map[string]interface{}{}
```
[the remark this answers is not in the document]
kimorris27: When you say to use a struct do you mean to unmarshal the inner error into an `azure.ServiceError` instead of a `map`?

If that's what you meant, I was trying to do that at first and it didn't work for some reason, but I'll go back and take one more look. It would definitely be nicer than having to do the type casting.

pkg/deploy/deploy.go
kimorris27: Hmm, I might have been missing some of the context on why you added this in originally.

I added it back here because my understanding was that it was an improvement to make the deployment process more robust to errors and didn't have any downsides. But if it's not needed yet or might cause problems, I'll go ahead and take it out and we can reconsider it in another PR.

pkg/deploy/deploy.go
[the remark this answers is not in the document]
facchettos: you can have a custom struct that you unmarshal into. here for example you can get your innerErr unmarshalled into something like this:
```go
type innerError struct{
    code string
    message string
}
```
and then you'd avoid the type casting on line 173/174

pkg/deploy/deploy.go
kimorris27: Gotcha, I'll go ahead and remove it from this PR in an upcoming commit then.

pkg/deploy/deploy.go
kimorris27: No longer a problem. The structure simplified itself naturally when I cast the errors to `ServiceError` structs.

pkg/deploy/deploy.go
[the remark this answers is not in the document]
kimorris27: I'm not sure what I was doing wrong when I first tried, but casting the nested errors to the `ServiceError` type ended up working fine (and I've committed the change).

pkg/deploy/deploy.go
```diff
@@ -158,20 +159,35 @@ func (d *deployer) deploy(ctx context.Context, rgName, deploymentName, vmssName
...
-		if serviceErr, ok := err.(*azure.ServiceError); ok &&
-			serviceErr.Code == "DeploymentFailed" &&
-			i < 1 {
-			// on new RP deployments, we get a spurious DeploymentFailed error
+		serviceErr, isServiceError := err.(*azure.ServiceError)
+
+		// Check for a known error that we know how to handle.
+		if isServiceError && serviceErr.Code == "DeploymentFailed" && len(serviceErr.Details) > 0 {
+			outerErr := azure.ServiceError{}
+			jsonEncoded, _ := json.Marshal(serviceErr.Details[0])
```
facchettos, line 167: can we check the error here?

facchettos (changes requested):

pkg/deploy/deploy.go
kimorris27: I intentionally left it out to keep things simple because if we get to the conditional where we check the error contents and there have been problems with the JSON marshaling/unmarshaling, `outerErr` and `innerErr`'s fields will either be empty or not contain what we're looking for. Do you think it's still worth checking it?

s-amann: IMO this json marshalling/unmarshalling should be in a helper function where we can have unit tests to confirm which error handling strategy we want to go with.

kimorris27: I've moved that functionality to a helper function. I'm not 100% confident in the way I handled the errors it returns though, so let me know what you think about it.

s-amann: I think you did a fine job, there is still a bit of nesting that could be removed, but that's more of a nitpick

s-amann (approved):

pkg/deploy/deploy.go
kimorris27: Yeah, I see what you mean. I found one spot where I removed some nesting.

s-amann (approved):

facchettos (approved):

PR #2823, opened 2023-04-04
Draft PR for Preflight validation

carvalhe:

Fixes https://issues.redhat.com/browse/ARO-1945

Implementing this API will benefit ARM template deployment and Portal deployment. Having this separate Api call, we can validate before the deploy phase and speed up the process on error catching. Currently a work in progress

### Test plan for issue:

Will have local set up to run Api call


pkg/frontend/frontend.go
```diff
@@ -238,36 +238,43 @@ func (f *frontend) chiAuthenticatedRoutes(router chi.Router) {
...
-
-					r.Delete("/", f.deleteOpenShiftCluster)
-					r.Get("/", f.getOpenShiftCluster)
-					r.Patch("/", f.putOrPatchOpenShiftCluster)
-					r.Put("/", f.putOrPatchOpenShiftCluster)
-
-					r.Post("/listcredentials", f.postOpenShiftClusterCredentials)
-
-					r.Post("/listadmincredentials", f.postOpenShiftClusterKubeConfigCredentials)
+		r.Route("/resourcegroups/{resourceGroupName}/providers/{resourceProviderNamespace}", func(r chi.Router) {
```
weinong, line 241: let's not mix the resource level route configuration with preflight api. instead, add one after this resourceType route configuration

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,120 @@
...
+		var err error
+		b, err = f._preflightValidation(ctx, log, body, correlationData, systemData, r.URL.Path, originalPath, r.Method, referer, &header, f.apis[apiVersion].OpenShiftClusterConverter, f.apis[apiVersion].OpenShiftClusterStaticValidator, subId, resourceProviderNamespace, apiVersion)
+		return err
+	})
+
+	frontendOperationResultLog(log, r.Method, err)
+	reply(log, w, header, b, err)
+}
+
+func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry, body []byte, correlationData *api.CorrelationData, systemData *api.SystemData, path, originalPath, method, referer string, header *http.Header, converter api.OpenShiftClusterConverter, staticValidator api.OpenShiftClusterStaticValidator, subId, resourceProviderNamespace string, apiVersion string) ([]byte, error) {
```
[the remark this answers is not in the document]
weinong: this is existing pattern that we can refactor later

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/go-chi/chi/v5"
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
+	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+)
+
+// NOTE: Make sure to change and always return 200, with any errors inside of the response message
+func (f *frontend) preflightValidation(w http.ResponseWriter, r *http.Request) {
```
niontive, line 19: Would be helpful to add a comment describing what a `preflightValidation` request URL looks. [Example in ARO-RP frontend.](https://github.com/Azure/ARO-RP/blob/30370deecd647c9e70662f2c302299407eebfce1/pkg/frontend/admin_openshiftcluster_vmsizelist.go#L22)

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,109 @@
...
+		return err
+	})
+
+	frontendOperationResultLog(log, r.Method, err)
+	reply(log, w, header, b, err)
+}
+
+func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry, body []byte, correlationData *api.CorrelationData, systemData *api.SystemData, path, originalPath, method, referer string, header *http.Header, converter api.OpenShiftClusterConverter, staticValidator api.OpenShiftClusterStaticValidator, subId, resourceProviderNamespace string, apiVersion string) ([]byte, error) {
+	// multiple errors should be able to return, just add them to the message
+	/*subscription, err := f.validateSubscriptionState(ctx, path, api.SubscriptionStateRegistered)
```
niontive, line 48: Nit: `/* subscription`

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,109 @@
...
+	correlationData := r.Context().Value(middleware.ContextKeyCorrelationData).(*api.CorrelationData)
+	systemData, _ := r.Context().Value(middleware.ContextKeySystemData).(*api.SystemData) // don't panic
+	originalPath := r.Context().Value(middleware.ContextKeyOriginalPath).(string)
+	referer := r.Header.Get("Referer")
+
+	subId := chi.URLParam(r, "subscriptionId")
+	resourceProviderNamespace := chi.URLParam(r, "resourceProviderNamespace")
+
+	apiVersion := r.URL.Query().Get(api.APIVersionKey)
+	err := cosmosdb.RetryOnPreconditionFailed(func() error {
```
niontive, line 36: Why is `_preflightValidation` used in `cosmosdb.RetryonPreconditionFailed`? Is this for the following scenario:
1) customer creates a cluster
2) customer wants to update cluster config via ARM template
3) we run preflight validation on the new ARM template to ensure it's an OK update, and that validation may require looking at the cluster config in CosmosDB

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if !newProvisionState {
+		err = validateTerminalProvisioningState(doc.OpenShiftCluster.Properties.ProvisioningState)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// validateTerminalProvisionState which if creating, wouldn't have. maybe save that check for deploying
+
+		// preflight doc mentions SKU and quota validation are good candidates to do preflight validation on
```
niontive, line 98: More things this code could check:
- Ensure VNet has no NSGs attached. I've gotten this error before.
- Ensure service principal client ID and secret aren't empty. Customers have done this before.
- Verify subnet CIDR ranges. Since we attach a load balancer, we need to have enough IP addresses.

nwnt: Let me know if you're adding the NSG check as I'm working on BYO NSG.  That would allow pre-attached NSGs on the subnet.

pkg/frontend/frontend.go
carvalhe: separated preflight route, and added after resourceType

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	doc, err := f.dbOpenShiftClusters.Get(ctx, path)
+	if err != nil && !cosmosdb.IsErrorStatusCode(err, http.StatusNotFound) {
+		return nil, err
+	}
+
+	newProvisionState := doc == nil
+
+	// dont create this as it would require creating openshiftclusterdoc
```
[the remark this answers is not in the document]
carvalhe: This is temporary but I did remove a few of the comments like these

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: Added url as mentioned

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: removed retryonPreconditionFailed

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry, body []byte, correlationData *api.CorrelationData, systemData *api.SystemData, path, originalPath, method, referer string, header *http.Header, converter api.OpenShiftClusterConverter, staticValidator api.OpenShiftClusterStaticValidator, subId, resourceProviderNamespace string, apiVersion string) ([]byte, error) {
+	// multiple errors should be able to return, just add them to the message
+	subscription, err := f.validateSubscriptionState(ctx, path, api.SubscriptionStateRegistered)
+	if err != nil {
+		return nil, err
+	}
+
+	doc, err := f.dbOpenShiftClusters.Get(ctx, path)
```
weinong, line 53: the `path` is passed in from the `originalPath` in L40 which will be something like `/subscriptions/{subscriptionId}/resourcegroups/{resourceGroupName}/providers/{resourceProviderNamespace}/deployments/{deploymentName}/preflight?api-version={api-version}`. So you actually need to open the body and loop for resources to call this function

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,109 @@
...
+
+	apiVersion := r.URL.Query().Get(api.APIVersionKey)
+	b, err := f._preflightValidation(ctx, log, body, correlationData, systemData, r.URL.Path, originalPath, r.Method, referer, &header, f.apis[apiVersion].OpenShiftClusterConverter, f.apis[apiVersion].OpenShiftClusterStaticValidator, subId, resourceProviderNamespace, apiVersion)
+
+	frontendOperationResultLog(log, r.Method, err)
+	reply(log, w, header, b, err)
+}
+
+func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry, body []byte, correlationData *api.CorrelationData, systemData *api.SystemData, path, originalPath, method, referer string, header *http.Header, converter api.OpenShiftClusterConverter, staticValidator api.OpenShiftClusterStaticValidator, subId, resourceProviderNamespace string, apiVersion string) ([]byte, error) {
+	// multiple errors should be able to return, just add them to the message
```
weinong, line 47: what does this comment mean? the code below will return one error as soon as possible

carvalhe: Thats just a current note to myself since this is a WIP PR and I focused on other areas, currently it does only return one error as you mentioned but will change.

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,136 @@
...
+	systemData, _ := r.Context().Value(middleware.ContextKeySystemData).(*api.SystemData)
+	originalPath := r.Context().Value(middleware.ContextKeyOriginalPath).(string)
+	referer := r.Header.Get("Referer")
+
+	subId := chi.URLParam(r, "subscriptionId")
+	resourceGroupName := chi.URLParam(r, "resourceGroupName")
+
+	resources, err := unmarshalRequest(body)
+	if err != nil {
+		log.Warningf("bad Request. The request content was invalid and could not be deserialized: %s.", err)
```
niontive, line 40: Why do we log bad requests instead of replying with the error?

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,136 @@
...
+				} else {
+					b = append(b, resultByte...)
+				}
+			}
+		}
+	}
+	reply(log, w, header, b, nil)
+}
+
+func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry, raw json.RawMessage, correlationData *api.CorrelationData, systemData *api.SystemData, path, originalPath, method, referer string, header *http.Header, converter api.OpenShiftClusterConverter, staticValidator api.OpenShiftClusterStaticValidator, subId, apiVersion string, resourceGroupName string) api.ValidationResult {
```
niontive, line 76: In general, it's a bad practice to have functions with tons of arguments. This makes these functions (1) tightly coupled and (2) harder to test. Would it be possible to organize some of these arguments into a struct so we can just past the struct?

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+			if strings.EqualFold(typeMeta.Type, "Microsoft.RedHatOpenShift/openShiftClusters") {
+				path := PreflightResourceId(subId, resourceGroupName, typeMeta.Name)
+				ch <- f._preflightValidation(ctx, log, raw, correlationData, systemData, path, originalPath, r.Method, referer, &header, f.apis[typeMeta.APIVersion].OpenShiftClusterConverter, f.apis[typeMeta.APIVersion].OpenShiftClusterStaticValidator, subId, typeMeta.APIVersion, resourceGroupName)
+			}
+		}
+		close(ch)
+
+		for res := range ch {
+			// go through and serialize into an array and make that the return body
```
niontive, line 61: Make this a separate function?

pkg/frontend/openshiftcluster_preflightvalidation.go
weinong: it's more like an existing (or established) coding style. i'd suggest deferring this till later

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: Since this is only used one time, we decided it would be more clear to leave as is unless something is getting called multiple times.

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: I've made changes so that we reply with the error, specifically we are looking to reply with a 200 as that is what the api wants even if there is a validation failure. We return 400 on only a couple instances.

pkg/api/preflight.go
```diff
@@ -0,0 +1,67 @@
...
+	Code *string `json:"code,omitempty"`
+	// Message - The error message returned from the server.
+	Message *string `json:"message,omitempty"`
+	// Target - The target of the error.
+	Target *string `json:"target,omitempty"`
+	// Details - Validation error.
+	Details *[]ManagementErrorWithDetails `json:"details,omitempty"`
+}
+
+func (m *ManagementErrorWithDetails) String() string {
```
weinong, line 32: can you add a simple unit test?

pkg/api/preflight.go
```diff
@@ -0,0 +1,67 @@
...
+
+// TypeMeta describes an individual API model object
+type TypeMeta struct {
+	// APIVersion is on every object
+	APIVersion string `json:"apiVersion"`
+}
+
+// ResourceTypeMeta is the Typemeta inside request body of preflight
+type ResourceTypeMeta struct {
+	Name       string `json:"name"`
```
weinong, line 51: I think you should include ID here

pkg/api/preflight.go
carvalhe: I could do that, but I also might just remove it since I'm not currently using it. Will check aks again

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,152 @@
...
+			// get typeMeta from the raw data
+			typeMeta := &api.ResourceTypeMeta{}
+			err = json.Unmarshal(raw, &typeMeta)
+			if err != nil {
+				// failing to parse the preflight body is not considered a validation failure. continue
+				log.Warningf("bad request. Failed to unmarshal ResourceTypeMeta: %s", err)
+				continue
+			}
+			if strings.EqualFold(typeMeta.Type, "Microsoft.RedHatOpenShift/openShiftClusters") {
+				path := PreflightResourceId(subId, resourceGroupName, typeMeta.Name)
```
weinong, line 51: you should get the reousrce ID from ResourceTypeMeta. See my comment on L51. The resource group here may not be the actual resource group of the resource.

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,152 @@
...
+
+	resources, err := unmarshalRequest(body)
+	if err == nil {
+		resourceCount := len(resources.Resources)
+		ch := make(chan api.ValidationResult, resourceCount)
+		for _, raw := range resources.Resources {
+			// get typeMeta from the raw data
+			typeMeta := &api.ResourceTypeMeta{}
+			err = json.Unmarshal(raw, &typeMeta)
+			if err != nil {
```
weinong, line 45: if all the resources result in unmarshal err, you will end up sending that err to reply() in L69, which would result in InternalServerError. We should always return StatusOk

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,152 @@
...
+			}
+			b, err = marshalValidationResult(&res)
+		}
+		log.Info("preflight validation succeeded")
+	}
+	reply(log, w, header, b, err)
+}
+
+func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry, raw json.RawMessage, path, originalPath, method, referer string, header *http.Header, converter api.OpenShiftClusterConverter, staticValidator api.OpenShiftClusterStaticValidator) api.ValidationResult {
+	_, err := f.validateSubscriptionState(ctx, path, api.SubscriptionStateRegistered)
```
weinong, line 73: nit:
```
if _, err := ...; er != nil {
```

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// unmarshal raw to OpenShiftCluster type
+	doc := &api.OpenShiftCluster{}
+	if !f.env.IsLocalDevelopmentMode() /* not local dev or CI */ {
+		doc.Properties.FeatureProfile.GatewayEnabled = true
+	}
+
+	var ext = converter.ToExternal(doc)
+	err = json.Unmarshal(raw, &ext)
+	if err != nil {
```
weinong, line 91: nit, L91 and L90 can be combined

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,152 @@
...
+		return api.ValidationResult{
+			Status: api.ValidationStatusFailed,
+			Error: &api.ManagementErrorWithDetails{
+				Message: to.StringPtr(err.Error()),
+			},
+		}
+	}
+
+	// For Put operation
```
weinong, line 100: what does this comment mean?

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,152 @@
...
+				return
+			}
+			b, err = marshalValidationResult(&res)
+		}
+		log.Info("preflight validation succeeded")
+	}
+	reply(log, w, header, b, err)
+}
+
+func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry, raw json.RawMessage, path, originalPath, method, referer string, header *http.Header, converter api.OpenShiftClusterConverter, staticValidator api.OpenShiftClusterStaticValidator) api.ValidationResult {
```
weinong, line 72: i think you should remove unused parameters such as `referer` and `header`

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: Was a old note since we are only looking at cluster create at the moment and should validate other requests later. Removed

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: Removed unused parameters

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,152 @@
...
+			typeMeta := &api.ResourceTypeMeta{}
+			err = json.Unmarshal(raw, &typeMeta)
+			if err != nil {
+				// failing to parse the preflight body is not considered a validation failure. continue
+				log.Warningf("bad request. Failed to unmarshal ResourceTypeMeta: %s", err)
+				continue
+			}
+			if strings.EqualFold(typeMeta.Type, "Microsoft.RedHatOpenShift/openShiftClusters") {
+				path := PreflightResourceId(subId, resourceGroupName, typeMeta.Name)
+				ch <- f._preflightValidation(ctx, log, raw, path, originalPath, r.Method, referer, &header, f.apis[typeMeta.APIVersion].OpenShiftClusterConverter, f.apis[typeMeta.APIVersion].OpenShiftClusterStaticValidator)
```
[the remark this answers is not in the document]
carvalhe: Passed just the version and create within _preflightValidation

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: Grabs ID now and uses that for the path instead

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: Always returns StatusOK

pkg/frontend/frontend.go
```diff
@@ -274,6 +274,11 @@ func (f *frontend) chiAuthenticatedRoutes(router chi.Router) {
 			})
 		})
 
+		r.Route("/resourcegroups/{resourceGroupName}/providers/{resourceProviderNamespace}/deployments/{deploymentName}/preflight", func(r chi.Router) {
+			r.Use(f.apiVersionMiddleware.ValidatePreflightAPIVersion) // TODO: Make new validation
```
weinong, line 278: what does this comment mean?

pkg/frontend/middleware/validate_api.go
```diff
@@ -29,3 +29,17 @@ func (a ApiVersionValidator) ValidateAPIVersion(h http.Handler) http.Handler {
 		h.ServeHTTP(w, r)
 	})
 }
+
+func (a ApiVersionValidator) ValidatePreflightAPIVersion(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		apiVersion := r.URL.Query().Get(api.APIVersionKey)
+		resourceProviderNamespace := chi.URLParam(r, "resourceProviderNamespace")
+		_, apiVersionExists := a.APIs[apiVersion]
+		if !apiVersionExists {
```
weinong, line 38:
```suggestion
		if _, apiVersionExists := a.APIs[apiVersion]; !apiVersionExists {
```

pkg/frontend/middleware/validate_api.go
```diff
@@ -29,3 +29,17 @@ func (a ApiVersionValidator) ValidateAPIVersion(h http.Handler) http.Handler {
...
+			api.WriteError(w, http.StatusBadRequest, api.CloudErrorCodeInvalidResourceType, "", "Could not be found in the namespace '%s' for api version '%s'.", resourceProviderNamespace, apiVersion)
```
weinong, line 39:
```suggestion
			api.WriteError(w, http.StatusBadRequest, api.CloudErrorCodeInvalidResourceType, "", "Api version %q is invalid for preflight validation in resource provider %q.", apiVersion, resourceProviderNamespace)
```

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,137 @@
...
+	ctx := r.Context()
+	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
+
+	var header http.Header
+	var b []byte
+
+	body := r.Context().Value(middleware.ContextKeyBody).([]byte)
+
+	resources, err := unmarshalRequest(body)
+	if err == nil {
```
weinong, line 31: what happens when err is not nil?

pkg/api/preflight.go
```diff
@@ -0,0 +1,57 @@
...
+	// Message - The error message returned from the server.
+	Message *string `json:"message,omitempty"`
+	// Target - The target of the error.
+	Target *string `json:"target,omitempty"`
+	// Details - Validation error.
+	Details *[]ManagementErrorWithDetails `json:"details,omitempty"`
+}
+
+// TypeMeta describes an individual API model object
+type TypeMeta struct {
```
weinong, line 33: this struct isn't really used by itself outside the `ResourceTypeMeta` below, right? I think you can just include `APIVersion` in `ResourceTypeMeta` and get rid of this one

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,137 @@
...
+		resourceCount := len(resources.Resources)
+		ch := make(chan api.ValidationResult, resourceCount)
+		for _, raw := range resources.Resources {
+			// get typeMeta from the raw data
+			typeMeta := &api.ResourceTypeMeta{}
+			err = json.Unmarshal(raw, &typeMeta)
+			if err != nil {
```
weinong, line 38:
```suggestion
			if err := json.Unmarshal(raw, &typeMeta); err != nil {
```

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,137 @@
...
+
+	body := r.Context().Value(middleware.ContextKeyBody).([]byte)
+
+	resources, err := unmarshalRequest(body)
+	if err == nil {
+		resourceCount := len(resources.Resources)
+		ch := make(chan api.ValidationResult, resourceCount)
+		for _, raw := range resources.Resources {
+			// get typeMeta from the raw data
+			typeMeta := &api.ResourceTypeMeta{}
```
weinong, line 36:
```suggestion
			typeMeta := api.ResourceTypeMeta{}
```

since you use pointer in the unmarshal below, this line should be a struct

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidRequestContent, "", "The request content was invalid and could not be deserialized: %q.", err)
+	}
+	return preflightRequest, nil
+}
+
+var validationSuccess = api.ValidationResult{
+	Status: api.ValidationStatusSucceeded,
+}
+
+func marshalValidationResult(results *api.ValidationResult) []byte {
```
weinong, line 130: you don't need to pass pointer for marshalling

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,137 @@
...
+	doc := &api.OpenShiftCluster{}
+	doc.Properties.ProvisioningState = api.ProvisioningStateSucceeded
+
+	if !f.env.IsLocalDevelopmentMode() /* not local dev or CI */ {
+		doc.Properties.FeatureProfile.GatewayEnabled = true
+	}
+
+	converter := f.apis[apiVersion].OpenShiftClusterConverter
+	staticValidator := f.apis[apiVersion].OpenShiftClusterStaticValidator
+	var ext = converter.ToExternal(doc)
```
weinong, line 85:
```suggestion
	ext := converter.ToExternal(doc)
```

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,125 @@
...
+				// failing to parse the preflight body is not considered a validation failure. continue
+				log.Warningf("bad request. Failed to unmarshal ResourceTypeMeta: %s", err)
+				continue
+			}
+			if strings.EqualFold(typeMeta.Type, "Microsoft.RedHatOpenShift/openShiftClusters") {
+				ch <- f._preflightValidation(ctx, log, raw, typeMeta.APIVersion)
+			}
+		}
+		close(ch)
+		// define the input, and what expected output will be. Better off to write my own, instead of using existing UT
```
weinong, line 47: please clean up this comment

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,125 @@
...
+	var header http.Header
+	var b []byte
+
+	body := r.Context().Value(middleware.ContextKeyBody).([]byte)
+
+	resources, err := unmarshalRequest(body)
+	if err == nil {
+		resourceCount := len(resources.Resources)
+		ch := make(chan api.ValidationResult, resourceCount)
+		for _, raw := range resources.Resources {
```
weinong, line 34: looking at Fleet's implementation, I don't believe a channel is needed here. can you simply this code so that reviewers don't get confused

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,125 @@
...
+
+func unmarshalRequest(body []byte) (*api.PreflightRequest, error) {
+	preflightRequest := &api.PreflightRequest{}
+	if err := json.Unmarshal(body, preflightRequest); err != nil {
+		return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidRequestContent, "", "The request content was invalid and could not be deserialized: %q.", err)
+	}
+	return preflightRequest, nil
+}
+
+var validationSuccess = api.ValidationResult{
```
weinong, line 114: can you move this var before the functions, which is a more general go convention

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,125 @@
...
+		// define the input, and what expected output will be. Better off to write my own, instead of using existing UT
+		for res := range ch {
+			// go through and return the first error
+			if res.Status == api.ValidationStatusFailed {
+				log.Warningf("preflight validation failed")
+				b = marshalValidationResult(res)
+				reply(log, w, header, b, statusCodeError(http.StatusOK))
+				return
+			}
+			b = marshalValidationResult(res)
```
weinong, line 56: this line can move to before the reply

pkg/frontend/openshiftcluster_preflightvalidation_test.go
```diff
@@ -0,0 +1,140 @@
...
+						`),
+					},
+				}
+			},
+			wantStatusCode: http.StatusOK,
+			wantResponse: &api.ValidationResult{
+				Status: api.ValidationStatusSucceeded,
+			},
+		},
+	} {
```
weinong, line 104: can you add a test case with template expression?

carvalhe: As discussed, will do this after we test on int so that we can see the preflight payload.

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidRequestContent, "", "The request content was invalid and could not be deserialized: %q.", err)
+	}
+	return preflightRequest, nil
+}
+
+func marshalValidationResult(results api.ValidationResult) []byte {
+	body, err := json.Marshal(results)
+	if err != nil {
+		return nil
+	} else {
```
weinong, line 115: nit: this else clause is redundant :)

weinong (approved):
LGTM

bennerv (commented):
Just a few comments first pass.  Not necessarily needed to be done in this PR, but could be in a follow-up.

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,118 @@
...
+	if err := json.Unmarshal(body, preflightRequest); err != nil {
+		return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidRequestContent, "", "The request content was invalid and could not be deserialized: %q.", err)
+	}
+	return preflightRequest, nil
+}
+
+func marshalValidationResult(results api.ValidationResult) []byte {
+	body, err := json.Marshal(results)
+	if err != nil {
+		return nil
```
cadenmarchese, line 114: Why return nil if there's an error with the marshaling?

pkg/frontend/openshiftcluster_preflightvalidation.go
cadenmarchese: This is just personal preference, but it might be nicer to go
```
if err != nil {
		b = marshalValidationResult(api.ValidationResult{
			Status: api.ValidationStatusFailed,
			Error: &api.ManagementErrorWithDetails{
				Message: to.StringPtr(err.Error()),
			},
		})
		reply(log, w, header, b, statusCodeError(http.StatusOK))
}
```
And then do your `for` loop after, separately. It would help flatten the code out a little

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -0,0 +1,118 @@
...
+			Status: api.ValidationStatusFailed,
+			Error: &api.ManagementErrorWithDetails{
+				Message: to.StringPtr(err.Error()),
+			},
+		})
+		reply(log, w, header, b, statusCodeError(http.StatusOK))
+	}
+}
+
+func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry, raw json.RawMessage, apiVersion string) api.ValidationResult {
```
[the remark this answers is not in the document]
carvalhe: The goal is to run preflight as on as much validation as we can. Preflight won't be able to be used on static and dynamic as some of the information from arm might not be available, but after testing on INT we will have a better idea of what can be validated.

pkg/frontend/middleware/validate_api.go
```diff
@@ -20,12 +20,24 @@ func (a ApiVersionValidator) ValidateAPIVersion(h http.Handler) http.Handler {
...
+		if _, apiVersionExists := a.APIs[apiVersion]; !apiVersionExists {
 			api.WriteError(w, http.StatusBadRequest, api.CloudErrorCodeInvalidResourceType, "", "The resource type '%s' could not be found in the namespace '%s' for api version '%s'.", resourceType, resourceProviderNamespace, apiVersion)
 			return
 		}
 
 		h.ServeHTTP(w, r)
 	})
 }
+
+func (a ApiVersionValidator) ValidatePreflightAPIVersion(h http.Handler) http.Handler {
```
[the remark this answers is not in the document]
carvalhe: ValidateAPIVersion() takes in the resourceType from the url which preflight doesn't have available to it.

bennerv (approved):
lgtm - please address feedback in a follow-up PR.  

Approving & merging so we can start release.

PR #2826, opened 2023-04-05
Allow aro operator to tolerate noschedule and noexecute

bennerv:

Fixes an issue where the customer marked all nodes as unschedulable and the ARO operator wouldn't run 

Above

### Test plan for issue:

Manually added it on the shared cluster and it ran successfully when all nodes were tainted with unschedulable.  

Nope

### DIFF
--- a/pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
+++ b/pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
@@ -46,6 +46,8 @@ spec:
       serviceAccount: aro-operator-master
       priorityClassName: system-cluster-critical
       tolerations:
-      - key: node-role.kubernetes.io/master
+      - effect: NoExecute
         operator: Exists
-        effect: NoSchedule
+      - effect: NoSchedule
+        operator: Exists
+
--- a/pkg/operator/deploy/staticresources/worker/deployment.yaml.tmpl
+++ b/pkg/operator/deploy/staticresources/worker/deployment.yaml.tmpl
@@ -42,3 +42,9 @@ spec:
       serviceAccountName: aro-operator-worker
       serviceAccount: aro-operator-worker
       priorityClassName: system-cluster-critical
+      tolerations:
+      - effect: NoExecute
+        operator: Exists
+      - effect: NoSchedule
+        operator: Exists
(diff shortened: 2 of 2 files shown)


pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
```diff
@@ -49,3 +49,8 @@ spec:
       - key: node-role.kubernetes.io/master
```
s-amann, line 49: do we need this now that we will tolerate NoSchedule on its own (matching any key)?

s-amann (approved):

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
bennerv: We don't actually need the one below.  Good catch.  We still need the key one so we tell it explicitly to run on masters.  This keeps the worker pod on workers and the master pod on masters.

bennerv: We need them both.  One ensures the master pod gets scheduled on the master nodes, the other ensures that if the node is cordoned, it can still schedule on the node.  

Ran a few tests to confirm: 
1. Place anti-affinity rules on worker deployment to not run on the same node
2. Scaled up the number of replicas, it did not schedule on the master nodes

1. Cordoned all nodes `oc adm cordon <node> <...>`
2. Deleted the master pod
3. Master pod wouldn't schedule
4. Added these tolerations to master and worker nodes
5. Master and worker pods scheduled

cblecker: @s-amann is right, we likely don't need this. The nodeSelector will enforce the scheduling, but if we are tolerating NoSchedule for any key, we don't need to also specifically tolerate it for the master taint. That said, it shouldn't have any negative effects to leave all three tolerations.

bennerv: Fixed.

s-amann (approved):

cadenmarchese (approved):
Good to merge on green e2e 👍

PR #2832, opened 2023-04-06
Fix `clean_e2e_db` not running in CI when e2e fails

cadenmarchese:

`clean_e2e_db` is responsible for cleaning up the cluster database from `e2e-eastus`, but because it was [only being executed in cases of succeeded e2e](https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=71317109&view=logs&j=10f66467-59e5-53bc-0c8a-42cc89770de4&t=f00aaa95-0643-5e60-2886-48bbeb9c10f0), it was causing failed e2e runs to leak databases. This eventually resulted in a full CosmosDB quota which was fixed by manually deleting cluster databases using `az`.

Ensures that `clean_e2e_db` always runs.

### Test plan for issue:

e2e passes, cluster database is deleted. e2e fails, cluster database is also deleted.

### DIFF
--- a/.pipelines/e2e.yml
+++ b/.pipelines/e2e.yml
@@ -110,6 +110,7 @@ jobs:
       . ./hack/e2e/run-rp-and-e2e.sh
       clean_e2e_db
     displayName: Cleanup (Azure)
+    condition: always()
   - template: ./templates/template-az-cli-logout.yml
 
   - task: PublishTestResults@2


cadenmarchese:
Marking `ready-for-review` since this is working as intended, and the `Cleanup (Azure)` pipeline step is executed even when e2e fails: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=71332267&view=logs&j=3965c7ea-79de-5618-9d9a-42b627bfb354&t=d9d16192-1f8f-5a67-50e7-e33d65609d41

The e2e itself is still having issues - will open up a separate PR for debugging further.

bennerv:
What about timeouts?  If the e2e takes too long, the pipeline times out.  Will the `always()` step still run?

cadenmarchese:
@bennerv yes, because when the e2e suite itself times out, it's a timeout of that particular pipeline step, not a timeout of the whole pipeline.

s-amann (approved):
lgtm

PR #2833, opened 2023-04-06
Fetch AME root certs and add them to the RP Docker containers

cloudfit-nick:
Add AME Root the ARO RP Docker Container

The AppLens endpoint is using an AME Root cert so the CA needs to be added to the container so that Go will not reject the endpoint.

[ARO-2612](https://issues.redhat.com/browse/ARO-2612)

### DIFF
--- a/Dockerfile.aro
+++ b/Dockerfile.aro
@@ -2,6 +2,7 @@ ARG REGISTRY
 FROM ${REGISTRY}/ubi8/ubi-minimal
 RUN microdnf update && microdnf clean all
 COPY aro e2e.test /usr/local/bin/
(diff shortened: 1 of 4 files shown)


bennerv:
@cloudfit-nick I was thinking more along the lines of: https://stackoverflow.com/questions/38582937/how-to-create-a-tls-client-with-ca-certificates-in-go

Make an HTTP request to the endpoint, and then set the TLS configuration in the http client to include the CA at runtime.  

It's similar here, but not quite the same: https://github.com/Azure/ARO-RP/blob/master/pkg/util/clientauthorizer/arm.go#L107-L152

@weinong  were you thinking along the same lines as myself?

weinong:
yes, I was telling Nick the same idea

PR #2835, opened 2023-04-07
Add changes for BYO NSG

nwnt:

Fixes ARO-1802

This is a private preview implementation for allowing customers to attach their own NSGs to the master and worker subnets.

### Test plan for issue:

- Currently tested successfully in my local development environment.

Feature private preview document is to be updated.


pkg/api/validate/dynamic/dynamic.go
```diff
@@ -554,8 +557,12 @@ func (dv *dynamic) validateVnetLocation(ctx context.Context, vnetr azure.Resourc
 	return nil
 }
 
+var (
```
weinong, line 560: it's more conventional to declare vars before any functions in the beginning of the file

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -61,6 +61,7 @@ type dynamic struct {
 	checkAccessSubjectInfoCred azcore.TokenCredential
 	env                        env.Interface
 	azEnv                      *azureclient.AROEnvironment
+	byoNSG                     bool
```
weinong, line 64: i think `allowBYONSG` is better as `byoNSG` doesn't state clearly whether this aro has BYO NSG or this subscription has BYO NSG Feature Flag enabled.

weinong:
LGTM

weinong (dismissed):

pkg/api/validate/dynamic/dynamic.go
nwnt: updated

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -590,34 +598,76 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
-
-				if !strings.EqualFold(*ss.SubnetPropertiesFormat.NetworkSecurityGroup.ID, expectedNsgID) {
-					return api.NewCloudError(
-						http.StatusBadRequest,
-						api.CloudErrorCodeInvalidLinkedVNet,
-						s.Path,
-						"The provided subnet '%s' is invalid: must not have a network security group attached.",
-						s.ID,
-					)
+				if !isTheSameNSG(*ss.SubnetPropertiesFormat.NetworkSecurityGroup.ID, expectedNsgID) {
```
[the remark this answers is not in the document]
nwnt: Nope, this is not the same logic as those lines.  It's actually the current logic that rejects the request when it sees if an NSG is attached to either of the subnets.  I changed these lines... kinda for my understanding of how the previous code works (not sure if I'd call it a refactor because it's not like it makes things better).

The only real change that I made is on the line 652 to skip this check when the subscription feature flag is on.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -590,34 +598,76 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+	allSubnetsAreAttached := len(subnetByID)
+
+	var attached int
+	for _, subnet := range subnetByID {
+		if subnet.NetworkSecurityGroup != nil && subnet.NetworkSecurityGroup.ID != nil {
+			attached++
+		}
+	}
+	if attached == allSubnetsAreAttached {
+		dv.log.Info("all subnets are attached, BYO NSG")
```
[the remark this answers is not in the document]
nwnt: Do you mean those on lines 34 - 36?  They are ... part of the "refactor" I mentioned on your comment about line 658.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -590,34 +598,76 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+		return true, nil // correct setup by customer
+	}
+	if attached == noNSGAttached {
+		dv.log.Info("no subnets are attached, no longer BYO NSG")
+		return false, nil
+	}
+
+	dv.log.Info("BYO NSG: not all subnets are attached")
+	return dv.byoNSG, fmt.Errorf(errMsgNSGNotProperlyAttached)
```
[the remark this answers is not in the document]
nwnt: done.

pkg/api/validate/openshiftcluster_validatedynamic.go
```diff
@@ -37,14 +37,15 @@ func NewOpenShiftClusterDynamicValidator(
 	oc *api.OpenShiftCluster,
 	subscriptionDoc *api.SubscriptionDocument,
 	fpAuthorizer autorest.Authorizer,
-) OpenShiftClusterDynamicValidator {
+) *openShiftClusterDynamicValidator {
```
weinong, line 40: why is this change needed?

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -87,13 +94,15 @@ func NewValidator(
 	authorizerType AuthorizerType,
 	cred azcore.TokenCredential,
 	pdpClient remotepdp.RemotePDPClient,
-) (Dynamic, error) {
+	byoNSG bool,
+) (*dynamic, error) {
```
weinong, line 98: why is this change needed?

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -590,34 +598,75 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+		subnetByID[s.ID] = ss
+	}
+	return subnetByID, nil
+}
+
+// checkByoNSG checks whether all the subnets have or don't have NSG attached.
+// when the BYONsg feature flag is on and only some of the subnets are attached with an NSG,
+// it returns an error.  If none of the subnets is attached, it's no longer BYO NSG and the
+// cluster installation process should fall back to using the managed nsg.
+func (dv *dynamic) checkByoNSG(ctx context.Context, subnetByID map[string]*mgmtnetwork.Subnet) (bool, error) {
```
weinong, line 610: is ctx being used?

weinong:
generally LGTM besides a few minor questions

pkg/api/validate/openshiftcluster_validatedynamic.go
nwnt: The validation has to propagate the flag back to the call, so I added a func to the object to return it.  I think the flag is implementation specific and shouldn't be on the interface, that's why I made it this way.

The other approach I thought of was to pass in a pointer and have the func mutate it, but if possible I'd like all validation funcs to be immutable.

pkg/api/validate/dynamic/dynamic.go
nwnt: Same as above.  It's another layer that I need to propagate the result flag from the validation.

pkg/api/validate/dynamic/dynamic.go
nwnt: Nope it's not.  I initially thought I would need a call to ARM, but turned out I didn't.  Should I just remove it?

weinong: i think we should since there is no rpc call

pkg/api/validate/openshiftcluster_validatedynamic.go
weinong: then `OpenShiftClusterDynamicValidator` is no longer used, right?

pkg/api/validate/dynamic/dynamic.go
nwnt: Removed accordingly.

pkg/api/validate/openshiftcluster_validatedynamic.go
nwnt: I think so.  I was also thinking the current way might be just temporary and when the API change is needed I can put the flag on the OpenshiftDoc instead and the changes here can be removed.  Would that be better or should I proceed to removing the `OpenshiftClusterDynamicValidator` for now?

rajdeepc2792:
@nwnt, Subscription feature flag does add the complexities to the code and thus the usage of `byonsg` in manager.
But IMO the dynamic validation logic should be used to validate/confirm certain roles, permissions and configurations, and not to populate a cluster document value via figuring out `byonsg` value from dynamic validation.
I am not sure if it I am thinking it right, but I think if vnet validation impacts any cluster document value, the validation code and thus populating of byonsg value could be managed at `/pkg/cluster` level.

pkg/util/cluster/cluster.go
```diff
@@ -520,7 +520,7 @@ func (c *Cluster) registerSubscription(ctx context.Context) error {
 			TenantID: c.env.TenantID(),
 			RegisteredFeatures: []api.RegisteredFeatureProfile{
 				{
-					Name:  "Microsoft.RedHatOpenShift/RedHatEngineering",
+					Name:  "Microsoft.RedHatOpenShift/BYONsg",
```
s-amann, line 523: why is this being changed?

nwnt: It shouldn't be there... got spilled from my testing. sorry about that!

nwnt: Removed now.

nwnt:
> @nwnt, Subscription feature flag does add the complexities to the code and thus the usage of `byonsg` in manager. But IMO the dynamic validation logic should be used to validate/confirm certain roles, permissions and configurations, and not to populate a cluster document value via figuring out `byonsg` value from dynamic validation. I am not sure if it I am thinking it right, but I think if vnet validation impacts any cluster document value, the validation code and thus populating of byonsg value could be managed at `/pkg/cluster` level.

@rajdeepc2792 That's what the current behavior is right now.  BYO NSG feature has a new validation that errors out when the subnets are semi-attached.  If none of the subnets is attached, it falls back to the default behavior (no longer BYO NSG).  Subsequent validations depend on this and that's why I'm propagating it back.  Currently no subscription document is modified.

pkg/api/validate/openshiftcluster_validatedynamic.go
nwnt: Removed in the latest commit.

pkg/api/validate/dynamic/dynamic.go
nwnt: This has also been removed in the latest commit.

weinong (dismissed):
LGTM

pkg/api/featureflags.go
```diff
@@ -22,4 +22,8 @@ const (
 	// FeatureFlagCheckAccessTestToggle is used for safely testing the new check access
 	// API in production. The toggle will be removed once the testing has been completed.
 	FeatureFlagCheckAccessTestToggle = "Microsoft.RedHatOpenShift/CheckAccessTestToggle"
+
+	// FeatureFlagBYONsg is used for indicating whether a customer subscription
+	// is registered for BYO NSG feature.
+	FeatureFlagBYONsg = "Microsoft.RedHatOpenShift/BYONsg"
```
bennerv, line 28: Since this will be an API option in future releases, we should track this in `pkg/api/openshiftcluster.go` and display it in our admin API

`pkg/api/admin/openshiftcluster.go`.  We will need conversions for both to internal and to external

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -32,6 +32,12 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/token"
 )
 
+var (
+	errMsgNSGAttached            = "The provided subnet '%s' is invalid: must not have a network security group attached."
+	errMsgNSGNotAttached         = "The provided subnet '%s' is invalid: must have network security group '%s' attached."
+	errMsgNSGNotProperlyAttached = "BYO NSG mode requires that all subnets are attached with an NSG."
```
bennerv, line 38: Assuming Puneet is good with these error messages.  if not please check with him :)

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -45,22 +51,14 @@ type ServicePrincipalValidator interface {
...
 
 // Dynamic validate in the operator context.
-type Dynamic interface {
-	ServicePrincipalValidator
-
-	ValidateVnet(ctx context.Context, location string, subnets []Subnet, additionalCIDRs ...string) error
-	ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster, subnets []Subnet) error
-	ValidateDiskEncryptionSets(ctx context.Context, oc *api.OpenShiftCluster) error
-	ValidateEncryptionAtHost(ctx context.Context, oc *api.OpenShiftCluster) error
-}
```
bennerv, line 55: would prefer not to get rid of our interfaces because we can mock them if they're public rather than structs.

pkg/api/featureflags.go
bennerv: We'll need to update our powershell scripts and geneva automation as well.

pkg/cluster/cluster.go
```diff
@@ -184,5 +187,6 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 		hiveClusterManager:                hiveClusterManager,
 		now:                               func() time.Time { return time.Now() },
 		openShiftClusterDocumentVersioner: new(openShiftClusterDocumentVersionerService),
+		byoNSG:                            feature.IsRegisteredForFeature(subscriptionDoc.Subscription.Properties, api.FeatureFlagBYONsg),
```
bennerv, line 190: if we set this in the cluster doc, we don't have to add the parameter to the manager struct.

pkg/cluster/defaults.go
```diff
@@ -7,19 +7,37 @@ import (
...
+}
+
+// ensureBYONsg updates the cluster docs to disable operators that might interfere
+// with BYO NSG feature
+func (m *manager) ensureBYONsg(ctx context.Context) error {
+	if m.byoNSG {
+		return patchClusterDoc(ctx, m, func(doc *api.OpenShiftClusterDocument) error {
+			flags := doc.OpenShiftCluster.Properties.OperatorFlags
+			flags["aro.azuresubnets.nsg.managed"] = "false"
+			flags["aro.azuresubnets.enabled"] = "false"
```
bennerv, line 29: is there a reason we're setting subnets to disabled as well?  I think subnets should still be setting service endpoints.

pkg/api/validate/openshiftcluster_validatedynamic.go
```diff
@@ -37,24 +37,26 @@ func NewOpenShiftClusterDynamicValidator(
...
-	return &openShiftClusterDynamicValidator{
+) *OpenShiftClusterDynamicValidator {
+	return &OpenShiftClusterDynamicValidator{
 		log: log,
 		env: env,
 
 		oc:              oc,
 		subscriptionDoc: subscriptionDoc,
 		fpAuthorizer:    fpAuthorizer,
+		byoNSG:          feature.IsRegisteredForFeature(subscriptionDoc.Subscription.Properties, api.FeatureFlagBYONsg),
```
bennerv, line 48: moving the feature flag to the oc document and resolving it early will clean this up.

bennerv (changes requested):
A fundamental change to move the field within the openshiftclusterdoc under the networkprofile will clean this up a lot.  We can leverage existing oc structs to perform checks rather than modifying the code.  

Plus it will give us some leverage when we GA this in an API.

pkg/api/openshiftcluster.go
```diff
@@ -274,6 +274,7 @@ type NetworkProfile struct {
 	APIServerPrivateEndpointIP string `json:"privateEndpointIp,omitempty"`
 	GatewayPrivateEndpointIP   string `json:"gatewayPrivateEndpointIp,omitempty"`
 	GatewayPrivateLinkID       string `json:"gatewayPrivateLinkId,omitempty"`
+	PreconfiguredNSG           bool   `json:"preconfigureNSG,omitempty"`
```
bennerv, line 277: same here - no bool

pkg/api/featureflags.go
bennerv: 2 stories - one for powershell script updates in ARO-Scripts, the other for geneva automation updates in ARO.Pipelines (or ARO-Pipelines, if it's migrated already)

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -607,34 +612,80 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+			&api.CloudError{
+				StatusCode: http.StatusBadRequest,
+				CloudErrorBody: &api.CloudErrorBody{
+					Code:    api.CloudErrorCodeInvalidLinkedVNet,
+					Message: errMsgNSGNotProperlyAttached,
+				},
+			}
+	}
+	if attached == allSubnetsAreAttached {
+		dv.log.Info("all subnets are attached, BYO NSG")
```
bennerv, line 646: This has  "fallback" logic in it.  I would prefer not to switch on this.  Let's talk with Puneet as if a customer specifies an API option or feature flag today but doesn't have everything in place we error out with a relevant error message.  

How will this work when we have an API option?  I would prefer the simpler approach of erroring out.

bennerv: IMO - Since we're toggling controllers internally it would be good to error on this so the customer knows why their feature isn't working.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -607,34 +612,80 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+// when the BYONsg feature flag is on and only some of the subnets are attached with an NSG,
+// it returns an error.  If none of the subnets is attached, it's no longer BYO NSG and the
+// cluster installation process should fall back to using the managed nsg.
+func (dv *dynamic) checkByoNSG(subnetByID map[string]*mgmtnetwork.Subnet) (bool, error) {
+	noNSGAttached := 0
+	allSubnetsAreAttached := len(subnetByID)
+
+	var attached int
+	for _, subnet := range subnetByID {
+		if subnet.NetworkSecurityGroup != nil && subnet.NetworkSecurityGroup.ID != nil {
```
bennerv, line 630: you wrote a helper function for this logic below.  Use that.

pkg/cluster/validate.go
```diff
@@ -6,12 +6,27 @@ package cluster
...
 		m.log, m.env, m.doc.OpenShiftCluster, m.subscriptionDoc, m.fpAuthorizer,
 	)
-	return ocDynamicValidator.Dynamic(ctx)
+	err := ocDynamicValidator.Dynamic(ctx)
+	if err != nil {
+		return err
+	}
+	// If the validation found that it's no longer BYONSG, update the doc
+	// TODO this very like not needed when the API change for BYONSG is introduced
+	if byoNSG != m.doc.OpenShiftCluster.Properties.NetworkProfile.PreconfiguredNSG {
```
bennerv, line 24: As mentioned before, let's ask Puneet if we can remove this.  I think we should be consistent with how we have done this in the past on other features.

bennerv (changes requested):
`pkg/api/defaults.go` should have a default value of Disabled/false as well.  So when we PUCM clusters it will pick up the flag.

pkg/cluster/install.go
```diff
@@ -46,6 +47,7 @@ func (m *manager) adminUpdate() []steps.Step {
 		steps.Action(m.initializeKubernetesClients), // must be first
 		steps.Action(m.ensureBillingRecord),         // belt and braces
 		steps.Action(m.ensureDefaults),
+		steps.Action(m.ensureBYONsg),
```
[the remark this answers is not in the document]
bennerv: > Clusters that were not created with the newly introduced option will stay the same (an NSG is to be created by ARO-RP and attached to the subnets).  The clusters however will not be able to switch to using BYOD NSG afterward.

pkg/api/featureflags.go
nwnt: ARO-3051 and ARO-3052 have been created for those stories.

pkg/api/validate/dynamic/dynamic.go
nwnt: That one was not the same although the name sounds like it (it's for a different type).  I have added a new one to keep things consistent anyway.

pkg/cluster/defaults.go
```diff
@@ -7,19 +7,36 @@ import (
...
 	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database"
 )
 
 // ensureDefaults will ensure cluster documents has all default values
 // for new api versions
 func (m *manager) ensureDefaults(ctx context.Context) error {
-	var err error
-	m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, func(doc *api.OpenShiftClusterDocument) error {
+	return patchClusterDoc(ctx, m, func(doc *api.OpenShiftClusterDocument) error {
```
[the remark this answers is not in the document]
nwnt: Removed the refactoring to keep things consistent.

pkg/cluster/install.go
```diff
@@ -373,6 +376,13 @@ func (m *manager) startInstallation(ctx context.Context) error {
 			doc.OpenShiftCluster.Properties.Install = &api.Install{
 				Now: time.Now().UTC(),
 			}
+
+			// TODO remove this when introducing the BYONSG CLI option
+			byoNSG := feature.IsRegisteredForFeature(m.subscriptionDoc.Subscription.Properties, api.FeatureFlagBYONsg)
+			if byoNSG {
+				m.log.Logger.Info("BYO NSG feature flag is on")
+				doc.OpenShiftCluster.Properties.NetworkProfile.PreconfiguredNSG = true
+			}
```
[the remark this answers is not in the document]
nwnt: Done

pkg/cluster/validate.go
nwnt: From the meeting last Wed, we'll keep this for now.

pkg/cluster/install.go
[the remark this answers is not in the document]
nwnt: Removed.

pkg/api/openshiftcluster.go
nwnt: Done

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -669,6 +720,26 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+	return strings.EqualFold(found, inDB)
+}
+
+func subnetHasNSGAttached(subnet *mgmtnetwork.Subnet) bool {
+	return subnet.NetworkSecurityGroup != nil && subnet.NetworkSecurityGroup.ID != nil
+}
+
+func nsgExists(subnetProp *mgmtnetwork.SubnetPropertiesFormat) bool {
+	return subnetProp != nil && subnetProp.NetworkSecurityGroup != nil
+}
```
bennerv, line 737: these functions are "almost" the same.  Let's consolidate it into one.

pkg/cluster/install.go
```diff
@@ -373,6 +375,12 @@ func (m *manager) startInstallation(ctx context.Context) error {
 			doc.OpenShiftCluster.Properties.Install = &api.Install{
 				Now: time.Now().UTC(),
 			}
+
+			// TODO remove this when introducing the BYONSG CLI option
+			if feature.IsRegisteredForFeature(m.subscriptionDoc.Subscription.Properties, api.FeatureFlagBYONsg) {
+				m.log.Logger.Info("BYO NSG feature flag is on")
+				doc.OpenShiftCluster.Properties.NetworkProfile.PreconfiguredNSG = api.PreconfiguredNSGEnabled
+			}
```
bennerv, line 383: move this into pkg/frontend/openshiftcluster_putorpatch.go` around the `determineOutboundType` function.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -607,34 +612,80 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+	}
+
+	if oc.Properties.NetworkProfile.PreconfiguredNSG == api.PreconfiguredNSGEnabled {
+		oc.Properties.NetworkProfile.PreconfiguredNSG, err = dv.checkByoNSG(subnetByID)
+		if err != nil {
+			return err
+		}
+	}
+
+	for _, s := range subnets {
```
bennerv, line 666: Can we add a comment at the top of this stating something along the lines of:
```
// we're parsing through the subnets slice, not the map because we will return consistent error messages on creation.
```

this is because map iteration is not consistent.  This gives the user a more consistent experience by going through an "ordered" list

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -607,34 +612,80 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+	if err != nil {
+		return err
+	}
+
+	if oc.Properties.NetworkProfile.PreconfiguredNSG == api.PreconfiguredNSGEnabled {
+		oc.Properties.NetworkProfile.PreconfiguredNSG, err = dv.checkByoNSG(subnetByID)
+		if err != nil {
+			return err
+		}
+	}
```
bennerv, line 664: This needs to be surrounded in an if statement because we only want this to change if the cluster is in creating.

bennerv: no need to change the logic on the `pkg/cluster/validate.go` as it will only change on creating.

bennerv (changes requested):
I think the NSG validation could be moved to the frontend for a faster failure / response to the customer.  

Maybe follow-up PR / story to do so.  Might clean up some of the logic in `pkg/cluster/validate.go` & `dynamic.go` from ensuring it's creating.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -607,34 +612,80 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+					Message: errMsgNSGNotProperlyAttached,
+				},
+			}
+	}
+	if attached == allSubnetsAreAttached {
+		dv.log.Info("all subnets are attached, BYO NSG")
+		return api.PreconfiguredNSGEnabled, nil // correct setup by customer
+	}
+	dv.log.Info("no subnets are attached, no longer BYO NSG. Fall back to using cluster NSG.")
+	return api.PreconfiguredNSGDisabled, nil
```
bennerv, line 650: Let's clean this up just a bit
```suggestion
	// all subnets have an attached NSG
	if attached == len(subnetByID) {
		dv.log.Info("all subnets are attached, BYO NSG")
		return api.PreconfiguredNSGEnabled, nil // correct setup by customer
	}

	// no subnets have attached NSG, fallback
	if attached == 0 {
		dv.log.Info("no subnets are attached, no longer BYO NSG. Fall back to using cluster NSG.")
		return api.PreconfiguredNSGDisabled, nil
	}

	// some subnets have NSGs attached, error out
	return api.PreconfiguredNSGDisabled,
		&api.CloudError{
			StatusCode: http.StatusBadRequest,
			CloudErrorBody: &api.CloudErrorBody{
				Code:    api.CloudErrorCodeInvalidLinkedVNet,
				Message: errMsgNSGNotProperlyAttached,
			},
		}

File: pkg/cluster/validate.go [hunk 28]
```diff
@@ -6,12 +6,27 @@ package cluster
...
+		m.log.Infof("No longer BYONSG, updating the doc's flag from %s to %s", byoNSG, m.doc.OpenShiftCluster.Properties.NetworkProfile.PreconfiguredNSG)
+		m.doc, err = m.db.PatchWithLease(ctx, m.doc.Key, func(doc *api.OpenShiftClusterDocument) error {
+			doc.OpenShiftCluster.Properties.NetworkProfile.PreconfiguredNSG = api.PreconfiguredNSGDisabled
+			return nil
+		})
+	}
+	return err
```
bennerv, line 31: After your change in `ValidateSubnets`, this will only happen on creation.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -617,34 +622,85 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
 				return err
 			}
 
-			if ss.SubnetPropertiesFormat == nil ||
-				ss.SubnetPropertiesFormat.NetworkSecurityGroup == nil ||
-				!strings.EqualFold(*ss.SubnetPropertiesFormat.NetworkSecurityGroup.ID, nsgID) {
-				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidLinkedVNet, s.Path, "The provided subnet '%s' is invalid: must have network security group '%s' attached.", s.ID, nsgID)
+			if !subnetHasNSGAttached(ss) ||
+				!isTheSameNSG(*ss.SubnetPropertiesFormat.NetworkSecurityGroup.ID, nsgID) {
+				return newSubnetValidationError(s.Path, errMsgNSGNotAttached, s.ID, nsgID)
```
bennerv, line 703: Needs an if statement around if BYO NSG is enabled.

bennerv:
```suggestion
if !subnetHasNSGAttached(ss) ||
				( !isTheSameNSG(*ss.SubnetPropertiesFormat.NetworkSecurityGroup.ID, nsgID) &&
				 	oc.Properties.NetworkProfile.PreconfiguredNSG != api.PreconfiguredNSGEnabled ){
				return newSubnetValidationError(s.Path, errMsgNSGNotAttached, s.ID, nsgID)
			}

File: pkg/api/validate/dynamic/dynamic_test.go [hunk 30]
```diff
@@ -1423,3 +1423,61 @@ func TestValidateNatGatewaysPermissionsWithCheckAccess(t *testing.T) {
...
+				log: logrus.NewEntry(logrus.StandardLogger()),
+			}
+			byoNSG, err := dv.checkByoNSG(tt.subnetByID)
+			utilerror.AssertErrorMessage(t, err, tt.wantErr)
+			if byoNSG != tt.byoNSG {
+				t.Errorf("byoNSG got %s, want %s", byoNSG, tt.byoNSG)
+			}
+		})
+	}
+}
```
bennerv, line 1483: a couple test cases around validateSubnet for both Creating & !Creating provisioning states would be good here.

pkg/cluster/validate.go
bennerv: wrapped around creation so will not change otherwise.  lgtm

bennerv (changes requested):
Super close - great work.  Just one small fix to make + few test cases.

bennerv:
One more question:

Our subnet service endpoint controller modifies the service endpoints on a subnet.  If we have a customer's NSG which we don't have network contributor over, will this controller run successfully?  

Can test this by bringing your own NSG, ensuring the cluster service principal only has network contributor over the VNET, (not the nsg), and check the subnet endpoint reconciliation.  

Note that we do have a feature to remove the subnet endpoint reconciliation, but that's only in egress lockdown enabled clusters.  

if so, we will need to make changes from an azure-cli perspective for BYO NSG + update permissions to ensure we have them in validate.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -33,6 +33,13 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/token"
 )
 
+var (
+	errMsgNSGAttached            = "The provided subnet '%s' is invalid: must not have a network security group attached."
```
cadenmarchese, line 37: IMO, it seems odd to create new vars for errors related to BYO NSG, but not for other error messages in this file (from other validators). We should either keep the existing convention, or do this for other validators as well.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -617,34 +623,92 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+		subnetByID[s.ID] = ss
+	}
+	return subnetByID, nil
+}
+
+// checkByoNSG checks whether all the subnets have or don't have NSG attached.
+// when the BYONsg feature flag is on and only some of the subnets are attached with an NSG,
+// it returns an error.  If none of the subnets is attached, it's no longer BYO NSG and the
+// cluster installation process should fall back to using the managed nsg.
+func (dv *dynamic) checkByoNSG(subnetByID map[string]*mgmtnetwork.Subnet) (api.PreconfiguredNSG, error) {
```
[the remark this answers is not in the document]
cadenmarchese: Looking back, it's not clear to me why we need `checkByoNSG()` at all (can we just have this logic in ValidateSubnets)?

The validation logic in the PR is split between `checkByoNSG()` and `ValidateSubnets()`, where the first one determines if subnets have / don't have NSGs attached, adjusting the doc if needed; and the second one takes it a step further and makes sure it's the same NSG.

Do we want them in separate functions because one is making changes to the cluster document, and the other is just validating? Or can we combine logic a little?

cadenmarchese (commented):
overall LGTM, but some optional suggestions / things I noticed on second review.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -581,34 +588,33 @@ func (dv *dynamic) validateVnetLocation(ctx context.Context, vnetr azure.Resourc
 	return nil
 }
 
-func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster, subnets []Subnet) error {
-	dv.log.Printf("validateSubnet")
```
SudoBrendan, line 585: I think we rely on this message for debugging cluster operations. This log message has been dropped from the ValidateSubnets flow, we should likely add it back in the `ValidateSubnets` function below.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -617,34 +623,92 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+			}
+		}
+	}
+
+	// we're parsing through the subnets slice, not the map because we'll return consistent error messages on creation
+	for _, s := range subnets {
+		ss := subnetByID[s.ID]
+
+		switch oc.Properties.ProvisioningState {
+		case api.ProvisioningStateCreating:
```
SudoBrendan, line 686: nit: it feels strange to define a `switch` block for a single `case`. `if/else` reads a bit more like natural english here.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -617,34 +623,92 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
 			if err != nil {
 				return err
 			}
 
-			if ss.SubnetPropertiesFormat == nil ||
-				ss.SubnetPropertiesFormat.NetworkSecurityGroup == nil ||
-				!strings.EqualFold(*ss.SubnetPropertiesFormat.NetworkSecurityGroup.ID, nsgID) {
-				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidLinkedVNet, s.Path, "The provided subnet '%s' is invalid: must have network security group '%s' attached.", s.ID, nsgID)
+			switch oc.Properties.NetworkProfile.PreconfiguredNSG {
+			case api.PreconfiguredNSGDisabled:
```
SudoBrendan, line 703: nit: another `switch` with (basically) a single case that could be an `if/else`, and has potential to have no path executed, even though it "shouldn't ever happen" :p

pkg/api/validate/dynamic/dynamic.go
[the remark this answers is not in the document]
SudoBrendan: I'm +1 on the `byo` vs `preconfigured` language - I'd rather pick one and go with it. If the BU-approved AFEC is `BYO`, let's use that everywhere so we have a 1:1 correlation of names to features.

SudoBrendan (approved):
I don't see anything that would stop me from merging, so LGTM. Left some nits you can action or dismiss. Thanks for your hard work on this!

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -679,6 +743,22 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 	return nil
 }
 
+func getRecordedNsgID(oc *api.OpenShiftCluster, subnetID string) (string, error) {
```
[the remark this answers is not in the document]
nwnt: Removed

pkg/api/validate/dynamic/dynamic.go
nwnt: Done - converted other validators as well.

pkg/api/validate/dynamic/dynamic.go
nwnt: Good catch.  I somehow accidentally deleted this.  Putting it back now.

pkg/api/validate/dynamic/dynamic.go
nwnt: Done.

pkg/api/validate/dynamic/dynamic.go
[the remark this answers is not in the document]
nwnt: I have changed all to `preconfigured` instead of `byo`.  

About the validation logic, `checkPreconfiguredNSG` (renamed) has to happen first before the existing validation in `ValidateSubnets()`.  This is because depending on the result of `checkPreconfiguredNSG`, it could turn the feature off and the validation logic then must change accordingly.  Having it in a separate function makes it easier to write unit tests for its test scenarios.  Also, `checkPreconfiguredNSG` most likely will be removed once the feature is GAed.  Once the real API change is introduced, the validation will be different (e.g. also rejecting when no NSG is attached at all).  So I think we should not combine them, at least for now.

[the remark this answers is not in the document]
cadenmarchese: Ok - thanks for the explanation.

cadenmarchese (approved):

SudoBrendan (approved):
There is still some `byo` vs `preconfigured` language in this PR, but it still ultimately LGTM!

PR #2839, opened 2023-04-10
Update dev proxy instance type

bennerv:

Uses a smaller dev proxy instance type for local development.  Cost savings yay

### Test plan for issue:
Deployed the instance type to dev.


s-amann (approved):
lgtm, but missing a JIRA

PR #2841, opened 2023-04-10
Geneva MDM Domain Change for INT in hack+devconfig

SrinivasAtmakuri:
Updates the -FrontEndUrl agrument to use
https://global.ppe.microsoftmetrics.com/ for INT.

Fixes


bennerv (approved):
lgtm - good catch

PR #2842, opened 2023-04-11
Remove initial installer dependencies

hawkowl:

Part of https://issues.redhat.com/browse/ARO-1418

Removes places where we use the Installer inside ARO-RP (excluding things that will be removed outright, later, like pkg/installer)

### Test plan for issue:

E2E should verify that this works.

N/A


pkg/util/installer/kubeadmin.go
```diff
@@ -0,0 +1,31 @@
...
+	Password string
+}
+
+type AdminKubeConfigSignerCertKey struct {
+	SelfSignedCertKey
+}
+
+// AdminKubeConfigClientCertKey is the asset that generates the key/cert pair for admin client to apiserver.
+type AdminKubeConfigClientCertKey struct {
+	SignedCertKey
```
facchettos, line 21: what are those for? if this is just to get the nested fields in json I think we should be explicit with json tags and maybe a comment. Embedding can be a little confusing when it's not used to extend objects/interfaces

pkg/cluster/kubeconfig.go
```diff
@@ -115,47 +117,58 @@ func (m *manager) generateKubeconfigs(ctx context.Context) error {
...
 }
 
-func generateKubeconfig(pg graph.PersistedGraph, commonName string, organization []string, validity time.Duration, internal bool) (*kubeconfig.AdminInternalClient, error) {
-	var ca *tls.AdminKubeConfigSignerCertKey
-	var adminInternalClient *kubeconfig.AdminInternalClient
-	err := pg.Get(true, &ca, &adminInternalClient)
+func generateKubeconfig(pg graph.PersistedGraph, commonName string, organization []string, validity time.Duration, internal bool) ([]byte, error) {
+	var ca *installer.AdminKubeConfigSignerCertKey
+	var adminInternalClient *installer.AdminInternalClient
+	err := pg.GetByName(false, "*tls.AdminKubeConfigSignerCertKey", &ca)
```
facchettos, line 132: are we using a pointer of a pointer here? It seems a little weird...

pkg/util/installer/certs.go
```diff
@@ -0,0 +1,76 @@
...
+	Cert() []byte
+}
+
+// CertKeyInterface contains a private key and the associated cert.
+// See openshift/installer/pkg/asset/tls/tls.go
+type CertKeyInterface interface {
+	CertInterface
+	// Key returns the private key.
+	Key() []byte
+}
```
facchettos, line 26: do these need to be interfaces? or is it just so we can use the type from installer directly?

pkg/cluster/kubeconfig.go
hawkowl: the type name as-saved in the map is "*tls.AdminKubeConfigSignerCertKey" and then we pull it out to a struct via a pointer, yeah

pkg/util/installer/kubeadmin.go
hawkowl: part of this is to maintain parity with upstream, even if it's not super nice

ellis-johnson (approved):
LGTM!

PR #2844, opened 2023-04-12
Force Resize option in Master Resize GA

SrinivasAtmakuri:
With this PR, a new option to Force resize a master node GA is added, this parameter prevents nodeName check, this will be helpful if the API server is unavailable or the control plane is under extreme load.

jira: https://issues.redhat.com/browse/ARO-2797

### DIFF
--- a/pkg/frontend/admin_openshiftcluster_vmresize.go
+++ b/pkg/frontend/admin_openshiftcluster_vmresize.go
@@ -62,7 +62,7 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
 	return action.VMResize(ctx, vmName, vmSize)
 }
 
-// A bland check, to validate if the node is master by checking the vmName has substring 'master'
+// A bland check, to validate if the node is master by checking vmName ends with pattern "-master-[0-9]"
 // return false, when the node is either not a master node or not adhering to the standard naming convention.
 // return true, if regexp satisfies
 func nodeIsMaster(vmName string) bool {
--- a/pkg/frontend/admin_openshiftcluster_vmresize.go
+++ b/pkg/frontend/admin_openshiftcluster_vmresize.go
@@ -5,16 +5,13 @@ package frontend
 
 import (
 	"context"
-	"encoding/json"
 	"net/http"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/sirupsen/logrus"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	kruntime "k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
@@ -34,7 +31,7 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
 	resourceType := chi.URLParam(r, "resourceType")
 	resourceGroupName := chi.URLParam(r, "resourceGroupName")
 
-	action, doc, err := f.prepareAdminActions(log, ctx, vmName, strings.TrimPrefix(r.URL.Path, "/admin"), resourceType, resourceName, resourceGroupName)
+	action, _, err := f.prepareAdminActions(log, ctx, vmName, strings.TrimPrefix(r.URL.Path, "/admin"), resourceType, resourceName, resourceGroupName)
 	if err != nil {
 		return err
 	}
@@ -45,44 +42,30 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
 		return err
 	}
 
-	k, err := f.kubeActionsFactory(log, f.env, doc.OpenShiftCluster)
+	// checks if the Virtual machines exists in the Cluster RG
+	exists, err := action.ResourceGroupHasVM(ctx, vmName)
 	if err != nil {
 		return err
 	}
-
-	nodeList, err := k.KubeList(ctx, "node", "")
-	if err != nil {
-		return err
-	}
-
-	var u unstructured.Unstructured
-	var nodes corev1.NodeList
-	if err = json.Unmarshal(nodeList, &u); err != nil {
-		return err
-	}
-
-	err = kruntime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &nodes)
-	if err != nil {
-		return err
-	}
-
-	nodeExists := false
-	for _, node := range nodes.Items {
-		if _, ok := node.ObjectMeta.Labels["node-role.kubernetes.io/master"]; !ok {
-			continue
-		}
-
-		if strings.EqualFold(vmName, node.ObjectMeta.Name) {
-			nodeExists = true
-			break
-		}
-	}
-
-	if !nodeExists {
+	if !exists {
 		return api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeNotFound, "",
-			`"The master node '%s' under resource group '%s' was not found."`,
+			`"The VirtualMachine '%s' under resource group '%s' was not found."`,
 			vmName, resourceGroupName)
 	}
 
+	if !nodeIsMaster(vmName) {
+		return api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden, "",
+			`"The vmName '%s' provided cannot be resized. It is either not a master node or not adhering to the standard naming convention."`,
+			vmName)
+	}
+
 	return action.VMResize(ctx, vmName, vmSize)
+}
+
+// A bland check, to validate if the node is master by checking the vmName has substring 'master'
+// return false, when the node is either not a master node or not adhering to the standard naming convention.
+// return true, if regexp satisfies
+func nodeIsMaster(vmName string) bool {
+	r := regexp.MustCompile(`.*-master-[0-9]{1,}$`)
+	return r.MatchString(vmName)
 }
--- a/pkg/frontend/admin_openshiftcluster_vmresize_test.go
+++ b/pkg/frontend/admin_openshiftcluster_vmresize_test.go
@@ -5,7 +5,6 @@ package frontend
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
@@ -13,8 +12,6 @@ import (
 
 	"github.com/golang/mock/gomock"
 	"github.com/sirupsen/logrus"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/env"
@@ -35,7 +32,7 @@ func TestAdminVMResize(t *testing.T) {
 		vmName         string
 		vmSize         string
 		fixture        func(f *testdatabase.Fixture)
-		mocks          func(*test, *mock_adminactions.MockAzureActions, *mock_adminactions.MockKubeActions)
+		mocks          func(*test, *mock_adminactions.MockAzureActions)
 		wantStatusCode int
 		wantResponse   []byte
 		wantError      string
@@ -44,7 +41,7 @@ func TestAdminVMResize(t *testing.T) {
 	for _, tt := range []*test{
 		{
 			name:       "basic coverage",
-			vmName:     "aro-fake-node-0",
+			vmName:     "aro-fake-node-master-0",
 			vmSize:     "Standard_D8s_v3",
 			resourceID: testdatabase.GetResourcePath(mockSubID, "resourceName"),
 			fixture: func(f *testdatabase.Fixture) {
@@ -69,34 +66,15 @@ func TestAdminVMResize(t *testing.T) {
 					},
 				})
 			},
-			mocks: func(tt *test, a *mock_adminactions.MockAzureActions, k *mock_adminactions.MockKubeActions) {
-				node := corev1.Node{
-					TypeMeta: metav1.TypeMeta{
-						Kind: "node",
-					},
-					ObjectMeta: metav1.ObjectMeta{
-						Name: "aro-fake-node-0",
-
-						Labels: map[string]string{
-							"node-role.kubernetes.io/master": "true",
-						},
-					},
-				}
-				nodeList := corev1.NodeList{
-					TypeMeta: metav1.TypeMeta{
-						Kind: "List",
-					},
-					Items: []corev1.Node{node},
-				}
-				marsh, _ := json.Marshal(nodeList)
-				k.EXPECT().KubeList(gomock.Any(), "node", "").Return(marsh, nil)
+			mocks: func(tt *test, a *mock_adminactions.MockAzureActions) {
+				a.EXPECT().ResourceGroupHasVM(gomock.Any(), tt.vmName).Return(true, nil)
 				a.EXPECT().VMResize(gomock.Any(), tt.vmName, tt.vmSize).Return(nil)
 			},
 			wantStatusCode: http.StatusOK,
 		},
 		{
 			name:       "cluster not found",
-			vmName:     "aro-fake-node-0",
+			vmName:     "aro-fake-node-master-0",
 			vmSize:     "Standard_D8s_v3",
 			resourceID: testdatabase.GetResourcePath(mockSubID, "resourceName"),
 			fixture: func(f *testdatabase.Fixture) {
@@ -110,13 +88,13 @@ func TestAdminVMResize(t *testing.T) {
 					},
 				})
 			},
-			mocks:          func(tt *test, a *mock_adminactions.MockAzureActions, k *mock_adminactions.MockKubeActions) {},
+			mocks:          func(tt *test, a *mock_adminactions.MockAzureActions) {},
 			wantStatusCode: http.StatusNotFound,
 			wantError:      `404: ResourceNotFound: : The Resource 'openshiftclusters/resourcename' under resource group 'resourcegroup' was not found.`,
 		},
 		{
 			name:       "subscription doc not found",
-			vmName:     "aro-fake-node-0",
+			vmName:     "aro-fake-node-master-0",
 			vmSize:     "Standard_D8s_v3",
 			resourceID: testdatabase.GetResourcePath(mockSubID, "resourceName"),
 			fixture: func(f *testdatabase.Fixture) {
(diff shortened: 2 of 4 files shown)


bennerv:
I made a comment in the story, but I would rather this take the approach of not validating the node name within ARO/OpenShift and use the vm client in azure-sdk-for-go to fetch the names of VMs and target it that way.  

There's some relations between VM name in Azure and node name in OpenShift, but it's not technically a guarantee.  I would prefer some validation, but let's remove the apiserver validation out of the way if that's the problem.  Use Azure and the state of the cluster resource group (what VMs are present) to do node/vm name validation instead of the kube-apiserver.

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -45,6 +45,21 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
 		return err
 	}
 
+	force := strings.EqualFold(r.URL.Query().Get("force"), "true")
+	if !force {
```
s-amann, line 49: please update the unit tests to include when we want to force the operation.

s-amann (changes requested):
please add a unit test to cover when we want to use `force`

jhoreman:
I am fine with having a force option, or even bypassing node validation entirely. This Geneva Action already has procedural checks and balances and the impact of resizing a worker node incorrectly (for example) is virtually nil and easily reverted. 

My only worry with introducing a force option as opposed to downing the workers and allowing for an overloaded control plane to calm down is that I am not sure what would happen to the control plane if, on an already overloaded control plane, we take one more node out of rotation. 

We'd potentially be looking at unexpected impact with etcd or other components. I would prefer the expected impact of downing workers over unexpected impact if we force the resize on an overloaded control plane. If we can somehow test this in a scenario where a force would be required and note the impact that would help alleviate these concerns.

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -45,6 +45,28 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
 		return err
 	}
 
+	force := strings.EqualFold(r.URL.Query().Get("force"), "true")
```
cadenmarchese, line 48: Could we get away with an exact comparison?

cadenmarchese (commented):
Thanks for the PR! I have a question on where the force itself happens, and a few more comments.

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -45,6 +45,28 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
...
+		// Skip if node exists check on force
+		// This parameter will come in handy if the controlplane is under extreme load
+		// Acts as a workaround if API server is unavailable
+		err = f.validateMasterNodeExists(log, ctx, vmName, resourceGroupName, doc)
+		if err != nil {
+			return err
+		}
+	} else {
+		// Bland check to prevent forceful Resize of a non-master node.
+		if !strings.Contains(strings.ToLower(vmName), "master") {
```
SudoBrendan, line 59: Per discussion in the Jira, and looking at this code, it seems sufficient to me to just rewrite the original "non forced" code to use the updated azure clients (vm level) rather than openshift (node level), if that's sufficient to determine a master node's status (which appears to be the only change here - it's a validation change, not a change to how we actually resize).

As I commented in the Jira as well, I think we can easily determine control plane status by looking at two non-openshift resources: vms and their subnets. Relying on strings like "master" is a little hacky, even though I'm sure it'll work in nearly all (if not all) cases.

SudoBrendan (changes requested):
If all we change is validation, I don't see a reason to name this anything like "force" or have branching strategies of validation. Let's just use the Azure clients for 100% of validation, since that has fewer dependencies than the OpenShift (node) level validation.

pkg/frontend/admin_openshiftcluster_vmresize.go
SrinivasAtmakuri: Based on the Slack discussion, we are letting this blank check live by going with the team consensus. Thanks.

pkg/frontend/admin_openshiftcluster_vmresize_test.go
AldoFusterTurpin, line 1: The mock would be affected after changing the interface.

AldoFusterTurpin (changes requested):
This looks nice @SrinivasAtmakuri. 😃 
I have left some suggestions that in my humble opinion can simplify it a bit. 

I think the change from action.ListVMsOfResourceGroup to action.ResourceGroupHasVM(vmName) is important as it sticks to the _tell don't ask principle_ (do not list and then manually filter, simply ask for what we want) and it improve the design a bit.

Thank you!

pkg/frontend/admin_openshiftcluster_vmresize_test.go
```diff
@@ -163,37 +147,62 @@ func TestAdminVMResize(t *testing.T) {
...
-			wantError:      `404: NotFound: : "The master node 'aro-fake-node-0' under resource group 'resourcegroup' was not found."`,
+			mocks: func(tt *test, a *mock_adminactions.MockAzureActions) {
+				a.EXPECT().ListVMsOfResourceGroup(gomock.Any()).Return([]mgmtcompute.VirtualMachine{
+					{
+						Name: to.StringPtr("aro-fake-node-0"),
+					},
+				}, nil)
+			},
+			wantStatusCode: http.StatusForbidden,
+			wantError:      `403: Forbidden: : "The provide vmName 'aro-fake-node-0' cannot be force resized as it is not a master node."`,
```
edisonLcardenas, line 198: Just typo "provide" -> "provide**d**"

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -45,44 +43,37 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
...
+			vmName, resourceGroupName)
 	}
 
-	err = kruntime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &nodes)
-	if err != nil {
-		return err
+	// bland check, to validate if the node is master by checking the vmName has substring 'master'
+	if !strings.Contains(strings.ToLower(vmName), "master") {
+		return api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden, "",
+			`"The provide vmName '%s' cannot be force resized as it is not a master node."`,
```
edisonLcardenas, line 63: Just typo "provide" -> "provide**d**"

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -45,44 +43,37 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
...
+		return api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeNotFound, "",
+			`"The VirtualMachine '%s' under resource group '%s' was not found."`,
+			vmName, resourceGroupName)
 	}
 
-	err = kruntime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &nodes)
-	if err != nil {
-		return err
+	// bland check, to validate if the node is master by checking the vmName has substring 'master'
+	if !strings.Contains(strings.ToLower(vmName), "master") {
```
yjst2012, line 61: nit: would it be a little better to use regex matching like .*-master-[0-9]{1,}, so for corner cases that the cluster name contains "master", the worker nodes wont be allowed.

pkg/frontend/admin_openshiftcluster_vmresize_test.go
```diff
@@ -44,7 +43,7 @@ func TestAdminVMResize(t *testing.T) {
 	for _, tt := range []*test{
 		{
 			name:       "basic coverage",
-			vmName:     "aro-fake-node-0",
+			vmName:     "aro-fake-master-node-0",
```
s-amann, line 46: can we update the names of the test nodes to reflect how they are named in production?

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -45,44 +43,37 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
...
+		return api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden, "",
+			`"The provide vmName '%s' cannot be force resized as it is not a master node."`,
+			vmName)
 	}
```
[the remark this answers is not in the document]
SrinivasAtmakuri: +1 on the suggestion, making the changes.

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -45,44 +43,37 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
...
-	var u unstructured.Unstructured
-	var nodes corev1.NodeList
-	if err = json.Unmarshal(nodeList, &u); err != nil {
-		return err
+	// checks if the provided VMname is under the RG
+	if !stringutils.Contains(filteredVmList, vmName) {
+		return api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeNotFound, "",
+			`"The VirtualMachine '%s' under resource group '%s' was not found."`,
+			vmName, resourceGroupName)
 	}
```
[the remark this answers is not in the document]
SrinivasAtmakuri: Thanks for the detailed review, I agree this makes the code a little hard to traverse and read but the reason I chose the current approach instead of taking the suggested one was to leverage having a func ListVMsOfResourceGroup in the azureActions, so it can come in handy at a later date, as we would need a function which would return all the VMs metadata of an RG instead of a name.

pkg/frontend/admin_openshiftcluster_vmresize.go
SrinivasAtmakuri: Done.

pkg/frontend/admin_openshiftcluster_vmresize_test.go
```diff
@@ -163,37 +147,62 @@ func TestAdminVMResize(t *testing.T) {
...
+								ResourceGroupID: fmt.Sprintf("/subscriptions/%s/resourceGroups/test-cluster", mockSubID),
+							},
 						},
 					},
-				}
-				nodeList := corev1.NodeList{
-					TypeMeta: metav1.TypeMeta{
-						Kind: "List",
+				})
+				f.AddSubscriptionDocuments(&api.SubscriptionDocument{
```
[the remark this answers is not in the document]
SrinivasAtmakuri: Done.

pkg/frontend/admin_openshiftcluster_vmresize_test.go
SrinivasAtmakuri: Done.

pkg/frontend/admin_openshiftcluster_vmresize.go
SrinivasAtmakuri: Done.

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -45,44 +43,37 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
...
+}
 
-		if strings.EqualFold(vmName, node.ObjectMeta.Name) {
-			nodeExists = true
-			break
+func (f *frontend) filterVMList(vmList []mgmtcompute.VirtualMachine) []string {
+	filteredVmList := []string{}
+	for _, vm := range vmList {
+		if vm.Name != nil {
+			filteredVmList = append(filteredVmList, *vm.Name)
```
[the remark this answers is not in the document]
SrinivasAtmakuri: I have outline the benefit in the current approach, will make changes if needed once a consensus is reached.

pkg/frontend/adminactions/azureactions.go
```diff
@@ -34,6 +34,7 @@ type AzureActions interface {
 	VMStopAndWait(ctx context.Context, vmName string, deallocateVM bool) error
 	VMSizeList(ctx context.Context) ([]mgmtcompute.ResourceSku, error)
 	VMResize(ctx context.Context, vmName string, vmSize string) error
+	ListVMsOfResourceGroup(ctx context.Context) ([]mgmtcompute.VirtualMachine, error)
```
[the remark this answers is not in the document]
SrinivasAtmakuri: I have outlined the benefit of the current approach, and will make changes if needed once a consensus is reached.

pkg/frontend/admin_openshiftcluster_vmresize.go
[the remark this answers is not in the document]
AldoFusterTurpin: I think that even if we keep the code as in the original version (something I am against of as previously mentioned), this function should not be in the frontend struct as it has already almost 50 dependencies. -> pkg/frontend/frontend.go:50

 I think this should be a regular function, not a method on the struct.

pkg/frontend/admin_openshiftcluster_vmresize.go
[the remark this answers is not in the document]
AldoFusterTurpin: > so it can come in handy at a later date
I suspected that this was the reason 😄 

Nevertheless, I'm a fervent believer in the YAGNI principle, so keeping the code more complex than necessary for potential additional use cases is something I'm almost always against. IMHO even if you want to keep a function that returns a list for future usage, it is done in a wrong way right now: you should put the function in *azureActions like:
```go
func (a *azureActions) ResourceGroupHasVM(ctx context.Context, vmName string) (bool, error) {
    // etc like in my previous suggestion
```
and you will be reusing  a.virtualMachines.List(ctx, clusterRGName). So the idea is that you don't want the clients of this struct to do the filter by hand (as it happens right now). We should expose the behaviour as a method for them.  I think the suggestion have the best of both worlds as it sticks to the tell don't ask principle but at the same time it gives you the possibility to wrap the internal function (the one that produces the whole list) and expose it to clients in the future by just creating an exported method that internally also calls a.virtualMachines.List(ctx, clusterRGName).

pkg/frontend/admin_openshiftcluster_vmresize.go
SrinivasAtmakuri: Yes correct, we are not forcing, I have corrected the comments and error messages. Thanks.

pkg/frontend/admin_openshiftcluster_vmresize.go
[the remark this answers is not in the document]
SrinivasAtmakuri: Made all the suggested changes. Thanks

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -45,44 +44,42 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
...
-			`"The master node '%s' under resource group '%s' was not found."`,
-			vmName, resourceGroupName)
-	}
+	return filteredVmList
+}
 
-	return action.VMResize(ctx, vmName, vmSize)
+// bland check, to validate if the node is master by checking the vmName has substring 'master'
+func nodeIsMaster(vmName string) bool {
+	r := regexp.MustCompile(`.*-master-[0-9]{1,}`)
```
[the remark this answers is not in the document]
SrinivasAtmakuri: Done! changed

pkg/frontend/admin_openshiftcluster_vmresize_test.go
```diff
@@ -110,13 +88,13 @@ func TestAdminVMResize(t *testing.T) {
 					},
 				})
 			},
-			mocks:          func(tt *test, a *mock_adminactions.MockAzureActions, k *mock_adminactions.MockKubeActions) {},
+			mocks:          func(tt *test, a *mock_adminactions.MockAzureActions) {},
```
facchettos, line 91: did you know that you can use `nil` here instead? (not really asking to change because the we should not use functions as test variables, but this is for another pr imho) (btw could you open it if you have 1 or 2 spare cycles please :relaxed: {at least for the `mock` functions})

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -45,44 +42,30 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
...
+	if !nodeIsMaster(vmName) {
+		return api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden, "",
+			`"The vmName '%s' provided cannot be resized. It is either not a master node or not adhering to the standard naming convention."`,
+			vmName)
+	}
+
 	return action.VMResize(ctx, vmName, vmSize)
 }
+
+// A bland check, to validate if the node is master by checking the vmName has substring 'master'
```
facchettos, line 65: that could use an update I would assume, since it is not correct anymore after your change

facchettos (approved):
I left a very small nit on a comment and a small refactor request(for another pr) but otherwise it's looking very good!

s-amann (dismissed):

pkg/frontend/admin_openshiftcluster_vmresize_test.go
SrinivasAtmakuri: Sure, will create a new PR for this. thanks.

pkg/frontend/admin_openshiftcluster_vmresize.go
SrinivasAtmakuri: Made the change!

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -45,44 +42,30 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
...
+	}
+
 	return action.VMResize(ctx, vmName, vmSize)
 }
+
+// A bland check, to validate if the node is master by checking vmName ends with pattern "-master-[0-9]"
+// return false, when the node is either not a master node or not adhering to the standard naming convention.
+// return true, if regexp satisfies
+func nodeIsMaster(vmName string) bool {
+	r := regexp.MustCompile(`.*-master-[0-9]{1,}$`)
```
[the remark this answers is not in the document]
facchettos: this isn't really required tbh. It's used only once right there

facchettos:
if they are only cosmetic/comments issues for now, I propose that we address them in a following PR so that it can go and have proper testing before next release

s-amann (approved):

pkg/frontend/admin_openshiftcluster_vmresize.go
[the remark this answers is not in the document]
AldoFusterTurpin: Yes. I agree it is not really required (that's why it was a LGTM + comment) but I think it helps in the long term because if the regex changes in the future, the function (_nodeIsMaster_) will not be touched, just the const on top of the file. 😃

bennerv (approved):

PR #2849, opened 2023-04-17
Update mirror image syncs to the latest versions

hawkowl:

just some chores

Bumps image versions

### Test plan for issue:

None for this, images to be tested once mirrored


cmd/aro/mirror.go
```diff
@@ -130,12 +131,18 @@ func mirror(ctx context.Context, log *logrus.Entry) error {
 	if env.Environment().Environment == azure.PublicCloud {
 		srcAcrGeneva := "linuxgeneva-microsoft" + acrDomainSuffix
 
-		mirrorImages := []string{
-			version.MdsdImage(srcAcrGeneva),
-			version.MdmImage(srcAcrGeneva),
+		// Mirror the versions that we have defined, as well as future versions
+		// for testing
+		mirrorImages := map[string]bool{
```
s-amann, line 136: if every value is true, can we keep this as a slice?

s-amann: we then wouldn't need the `maps` package either

ellis-johnson (approved):
LGTM!

s-amann (approved):

PR #2850, opened 2023-04-17
VM Allocation Status

anshulvermapatel:
This adds a column to the machine list to show allocation status of the machine VMs.

This is the updated PR of this PR - https://github.com/Azure/ARO-RP/pull/2554
The above one got merged but it broke the Portal when released due to the use of fpAuthorizer. This PR does the same thing but uses spAuthorizer which does not need AZURE_FP_CLIENT_ID env. Also the fpAuthorizer is mainly for RP not for portal.
This is for [ARO-1766](https://issues.redhat.com/browse/ARO-1766)


ellis-johnson (approved):
LGTM!

UlrichSchlueter:
Can you please provide more details on the background of this PR (along the lines of usual template) . I'm particularly interested in the testing that was performed (if any)

anshulvermapatel:
@bennerv , Hi Ben, Thanks for the explanation.
To get any information from azure side for the admin portal we need some kind of authentication mechanism for that. I do get your point about the discussion which is going on, Can you please make me a part of that discussion?

Meanwhile for sure we can hold this PR.(btw there are other work items which are block because of this however of course we want a permanent solution for this.)

bennerv:
Had a conversation with @anshulvermapatel on this today.  We might be able to leverage [ASI](https://aka.ms/asi) for some of the portal tooling rather than implementing it ourselves.  Will reach out to MSFT for a demo and the onboarding process, and we might be able to not maintain or carry any SRE portal work for the same functionality.

PR #2851, opened 2023-04-17
Reconcile upstream changes / remove Network SDK

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-1953

Reconciles azure-cli upstream changes which also removes the network sdk.

### Test plan for issue:

Existing unit tests and manual validation

NA


dem4gus:
I believe the upstream copyright headers have changed as well, can we get those in this PR?

```
# --------------------------------------------------------------------------------------------
# Copyright (c) Microsoft Corporation. All rights reserved.
# Licensed under the MIT License. See License.txt in the project root for license information.
# --------------------------------------------------------------------------------------------
```

dem4gus (dismissed):
The upstream copyright headers in all the files have been changed as well, can we get that into this PR?

```
# --------------------------------------------------------------------------------------------
# Copyright (c) Microsoft Corporation. All rights reserved.
# Licensed under the MIT License. See License.txt in the project root for license information.
# --------------------------------------------------------------------------------------------
```

tony-schndr:
> The upstream copyright headers in all the files have been changed as well, can we get that into this PR?
> 
> ```
> # --------------------------------------------------------------------------------------------
> # Copyright (c) Microsoft Corporation. All rights reserved.
> # Licensed under the MIT License. See License.txt in the project root for license information.
> # --------------------------------------------------------------------------------------------
> ```



> The upstream copyright headers in all the files have been changed as well, can we get that into this PR?
> 
> ```
> # --------------------------------------------------------------------------------------------
> # Copyright (c) Microsoft Corporation. All rights reserved.
> # Licensed under the MIT License. See License.txt in the project root for license information.
> # --------------------------------------------------------------------------------------------
> ```

@dem4gus I question this, mainly because validate-go CI fails at this step  `go run ./hack/licenses -validate -ignored-go vendor,pkg/client,.git -ignored-python python/client,vendor,.git`.  It complains about not finding licenses in the files added to `python/az/aro/azext_aro/aaz`, though there is a license it is just not the license that the code expects.  These files are directly copied from upstream azure-cli.  I'm not really sure if I should update the licenses in `python/az/aro/azext_aro/aaz` to the apache license or update validate-go to include MIT.

tony-schndr:
@ellis-johnson Upstream azure-cli removed the Network SDK which forced me to have to refactor quite a bit in the dynamic validation code.  Please review this PR when you have a chance.

python/az/aro/azext_aro/_dynamic_validators.py
```diff
@@ -57,11 +59,13 @@ def validate_resource(client, key, resource, actions):
...
-                                        subnet_parts['name'],
-                                        subnet_parts['child_name_1'])
+        subnet_obj = subnet_show(cli_ctx=cli_ctx)(command_args={
+            'resource_group': subnet_parts['resource_group'],
+            'name': subnet_parts['child_name_1'],
+            'vnet_name': subnet_parts['name']
+        })
     except ResourceNotFoundError as err:
         raise InvalidArgumentValueError(
             f"Invalid -- subnet, error when getting '{subnet}': {str(err)}") from err
```
tony-schndr, line 71: @ellis-johnson when refactoring this code I noticed this error message looked like it was missing `--{key.replace('_', '-')}`, is that right?  If so I could work that into this PR but I think it will need some additional changes to work with `dyn_validate_cidr_ranges`, or just pass None as key value since this exception shouldn't be raised in `dyn_validate_cidr_ranges` as it would be caught when validating subnets.

ellis-johnson: Woops I didn't realise that was the intention of `--` in these error messages so that was just a bad copy on my behalf. I'm happy to just pass None for the error in `dyn_validate_cidr_ranges`. Or even to remove the `--` in the message itself.

ellis-johnson (commented):
Great refactor. Just one change around the get_subnet error message that needs action

python/az/aro/azext_aro/_dynamic_validators.py
tony-schndr: Done!  Passed key as None so this aligns with the rest of the code.  Thanks for checking this!

ellis-johnson (approved):
LGTM!

python/az/aro/azext_aro/_validators.py
```diff
@@ -156,11 +156,14 @@ def _validate_subnet(cmd, namespace):
         if parts['child_type_1'].lower() != 'subnets':
             raise InvalidArgumentValueError(f"--{key.replace('_', '-')} child type '{subnet}' must equal subnets.")
 
-        client = get_mgmt_service_client(
-            cmd.cli_ctx, ResourceType.MGMT_NETWORK)
+        from .aaz.latest.network.vnet.subnet import Show
```
SudoBrendan, line 159: If we put this here, every time we call this function, we have the overhead of sourcing this library function. Best to have all `import` statements at the top of the file so we only see that cost once.

python/az/aro/azext_aro/custom.py
```diff
@@ -355,8 +355,7 @@ def generate_random_id():
 
 
 def get_network_resources_from_subnets(cli_ctx, subnets, fail):
-    network_client = get_mgmt_service_client(cli_ctx, ResourceType.MGMT_NETWORK)
-
+    from .aaz.latest.network.vnet.subnet import Show
```
SudoBrendan, line 358: Same as above. `import`s are best at the top of the file.

python/az/aro/azext_aro/__init__.py
```diff
@@ -20,6 +20,17 @@ def __init__(self, cli_ctx=None):
                          custom_command_type=aro_custom)
 
     def load_command_table(self, args):
+        from azure.cli.core.aaz import load_aaz_command_table
```
SudoBrendan, line 23: same as other comments. The import below in the `try` stays, I think.

SudoBrendan (changes requested):
It's a little difficult to tell what is a refactor vs a direct pull from upstream so take these comments with a grain of salt...

One thing I noticed is a few inline `import` statements, which:

1. Have (minor? major?) computational overhead each time we call the functions in question
2. More importantly, hide dependencies inside the code instead of grouping all dependencies at the header for each file

However, maybe a special case talked about [here](https://stackoverflow.com/questions/128478/should-import-statements-always-be-at-the-top-of-a-module) or elsewhere is being considered?

python/az/aro/azext_aro/_validators.py
tony-schndr: I agree 100%, the reason the import is not at the top is this is how upstream had it.  I'd be happy to change it to align more with what I did in _dynamic_validators.py for the imports, I wasn't sure how much we wanted to deviate from the upstream implementation. We could fix it in upstream since this is a bad practice to do with python.

SudoBrendan: We can upstream whatever we'd like to our extension as long as they accept the PR, so I'd say if we're all on board here, let's try it

python/az/aro/azext_aro/__init__.py
tony-schndr: I put the `try` at the top as well, tested the ext and it works as it it.

python/az/aro/azext_aro/custom.py
tony-schndr: fixed

python/az/aro/azext_aro/_validators.py
tony-schndr: Done, all imports are now at the top.

SudoBrendan (approved):
LGTM

jhoreman (approved):
Thanks for this, LGTM

PR #2854, opened 2023-04-19
Rewrite some of the cert/key encoding and decoding

hawkowl:
Splitting out some of #2842 that is unrelated

Rewrites some of the PEM encoding/decoding which will be used when the installer dependencies are removed

### Test plan for issue:
Unit tests, should mostly be a no-op

N/A


pkg/util/pem/pem_test.go
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the Apache License 2.0.
+
+import (
+	"bytes"
+	"testing"
+
+	utiltls "github.com/Azure/ARO-RP/pkg/util/tls"
+)
+
+func TestEncode(t *testing.T) {
```
SudoBrendan, line 13: It looks like this one function has multiple given/when/then test cases centered around `Encode` calls. Can we break out each of these into named `t.Run` clauses to give some context for each, which would have added benefit to adding context to each `t.Fatal` (which, as it stands now, would be hard to know at which point something went wrong)

SudoBrendan (changes requested):
Overall, looks really good! I think adding context for each `t.Fatal` in some way would be helpful.

jaitaiwan (commented):
Agree w/ Brendan's comments, apart from that LGTM

pkg/util/pem/pem_test.go
jaitaiwan, line 1: Since we're doing more verbal-like testing, I'd personally expect something like:
```
describe "pem cert package"
  describe "certificate encoding keys"
    it "successfully encodes a private key"
```

etc etc

jaitaiwan: So more "wordy" or "explanatory" in the descriptions.

jaitaiwan (changes requested):

pkg/env/certificateRefresher_test.go
```diff
@@ -117,33 +117,6 @@ PeihXgcD511I3eMq7A8r7+xlNFEFYaleYwmoYSOt0cBGtoZttXDYacR/xvEkVaxO
 1LQ=
 -----END CERTIFICATE-----`
 
-// equalPrivKey is added manually, it has been added to go1.15, whilst ARO uses go1.14
-// TODO: Remove once ARO is on go1.15
```
s-amann, line 121: whoot its about time!

s-amann (approved):

ellis-johnson (approved):
LGTM!

PR #2855, opened 2023-04-20
Update to use new fluentbit, mdm, and mdsd

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-1384 and https://issues.redhat.com/browse/ARO-1386

Updates the fluentbit, MDM, and MDSD images to use latest Mariner 2 versions.

### Test plan for issue:

E2E to verify that the images work.

N/A


ellis-johnson:
```
2023/04/20 02:12:37 pkg/deploy/assets/rp-production.json has been modified after make generate
2023/04/20 02:12:37 pkg/deploy/assets/gateway-production.json has been modified after make generate
exit status 1
```

Looks like you just need to commit these 2 files and you should be good to go

hawkowl:
> ```
> 2023/04/20 02:12:37 pkg/deploy/assets/rp-production.json has been modified after make generate
> 2023/04/20 02:12:37 pkg/deploy/assets/gateway-production.json has been modified after make generate
> exit status 1
> ```
> 
> Looks like you just need to commit these 2 files and you should be good to go

oops, fixed

pkg/util/version/const.go
```diff
@@ -76,19 +76,19 @@ var (
...
 func MdmImage(acrDomain string) string {
-	return acrDomain + "/distroless/genevamdm:mariner_20221026.2"
+	return acrDomain + "/genevamdm:2.2023.331.1521-399d45-20230331t1638"
 }
 
 // MdsdImage contains the location of the MDSD container image
 // see https://eng.ms/docs/products/geneva/collect/references/linuxcontainers
 func MdsdImage(acrDomain string) string {
-	return acrDomain + "/genevamdsd:master_20221018.2"
+	return acrDomain + "/genevamdsd:mariner_20230413.1"
```
cblecker, line 91: Shouldn't this be `mariner_20230418.1` (according to the sourced doc)?

hawkowl: That updated since I did the mirror :D

Dockerfile.muo
```diff
@@ -1,30 +0,0 @@
-ARG REGISTRY
```
bennerv, line 1: This image is FIPs compliant now which is why we don't have to manually build it, correct?

bennerv: Did we remove any associated pipeline references in ADO if so?  Or were there not any?

cmd/aro/mirror.go
```diff
@@ -130,19 +129,10 @@ func mirror(ctx context.Context, log *logrus.Entry) error {
 	// soverign clouds a separate mirror process mirrors from the public cloud
 	if env.Environment().Environment == azure.PublicCloud {
 		srcAcrGeneva := "linuxgeneva-microsoft" + acrDomainSuffix
-
-		// Mirror the versions that we have defined, as well as future versions
-		// for testing
 		mirrorImages := []string{
-			version.MdsdImage(srcAcrGeneva),
-			version.MdmImage(srcAcrGeneva),
```
bennerv, line 138: Why are we removing mirroring for these versions?  

I understand why we would add more, but why would we remove the ones that get set in cluster?

bennerv (commented):
Couple questions.   Additionally, if we're rolling these out can you update them in RP-Config in ADO as well?

cmd/aro/mirror.go
hawkowl: The `distroless` gets dropped so we can't reuse the same pull spec :(

bennerv: ack - so on mirror, `/distroless` gets dropped when it's mirrored into our ACR.  Would it make sense to have a comment somewhere stating that?

bennerv (approved):
LGTM - would like to see a comment about how mirroring drops the `distroless` so when the next person goes to update it, we're not confused when it doesn't work.

PR #2860, opened 2023-04-20
Validate portal build assets are up-to-date in PRs

tsatam:

None

Currently, we store the portal's build artifacts in the repository. If these are invalid, either out-of-sync with itself or with the corresponding source, this can cause issues when deploying the portal. 

This PR adds a step to the `node-lint` GitHub action to validate that the build artifacts are up-to-date with source, by simply running the build and checking if the files have changed. 

### Test plan for issue:

See the `npm_build_check` check on this very PR. We can play with pushing temp changes to the portal on this branch and view failing checks. 

N/A

### DIFF
--- a/.github/workflows/npm-audit.yml
+++ b/.github/workflows/npm-audit.yml
@@ -36,6 +36,6 @@ jobs:
       - name: Run NPM Build
         working-directory: ./portal/v2
         run: |
-          npm install
+          npm ci
           npm run build
           ../../hack/ci-utils/isClean.sh
--- a/.github/workflows/npm-audit.yml
+++ b/.github/workflows/npm-audit.yml
@@ -23,3 +23,19 @@ jobs:
       - name: Run NPM Audit
         run: bash ${GITHUB_WORKSPACE}/hack/github-actions/npm_audit.sh
         shell: bash
+  npm-build-check:
+    name: npm-build-check
+    runs-on: ubuntu-latest
+    steps:
+      - name: setup Node.JS
+        uses: actions/setup-node@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-audit.yml
```diff
@@ -23,3 +23,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: setup Node.JS
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.16.0
+      - name: Checkout Code
+        uses: actions/checkout@v3
+      - name: Run NPM Build
+        working-directory: ./portal/v2
```
bennerv, line 37: Do we need this for both v1 and v2?

tsatam: We could add this for both v1 and v2. I just followed in the footsteps of the existing audit step that only runs against v2.

dem4gus (changes requested):
Looks good, just a small change on installing.

jhoreman (approved):
Looks good to me pending changes from other reviewers. Thanks for this!

.github/workflows/npm-audit.yml
SudoBrendan: I think I'm fine with just v2 since that's the direction we're heading in, yea? I imagine we will eventually deprecate v1 for the sake of maintainability?

SudoBrendan (approved):
LGTM

s-amann (approved):

cadenmarchese (approved):

PR #2861, opened 2023-04-20
Func to list SupportedVMSizesByRole

SrinivasAtmakuri:
This PR creates a new function SupportedVMSizesByRole to provide supportedMasterVmSizes and supportedWorkerVmSizes thus preventing accidental changes to the variables by changing the scope of these variables from public.

Jira: https://issues.redhat.com/browse/ARO-2882

### DIFF
--- a/pkg/api/validate/vm.go
+++ b/pkg/api/validate/vm.go
@@ -13,7 +13,23 @@ import (
 // To add new instance types, needs Project Managment's involment and instructions are below.,
 // https://github.com/Azure/ARO-RP/blob/master/docs/adding-new-instance-types.md
 
-var SupportedMasterVmSizes = map[api.VMSize]api.VMSizeStruct{
+const VMRoleMaster string = "master"
+const VMRoleWorker string = "worker"
+
+var supportedVMSizesByRoleMap = map[string]map[api.VMSize]api.VMSizeStruct{
+	VMRoleMaster: supportedMasterVmSizes,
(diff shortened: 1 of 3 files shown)


facchettos (approved):
Lgtm !

facchettos:
Let's rebase on master when python is fixed, then ping me again :)

pkg/api/validate/vm.go
```diff
@@ -13,7 +13,23 @@ import (
 // To add new instance types, needs Project Managment's involment and instructions are below.,
 // https://github.com/Azure/ARO-RP/blob/master/docs/adding-new-instance-types.md
 
-var SupportedMasterVmSizes = map[api.VMSize]api.VMSizeStruct{
+const VMRoleMaster string = "master"
+const VMRoleWorker string = "worker"
+
+var supportedVMSizesByRoleMap = map[string]map[api.VMSize]api.VMSizeStruct{
+	VMRoleMaster: supportedMasterVmSizes,
```
jhoreman, line 20: Hey Srini. How are we validating that these sizes are *actually* supported for the master? I see we have an explicit list of them in vm.go, but whether or not masters can actually support this size is based on the subscription level allotment, regional availability (which may be dynamic) etc. Are we doing any specific checking inside these parameters?

facchettos: There's the sku validator that does this kind of this iirc. It's in the frontend

PR #2863, opened 2023-04-20
Test cases for GA SupportedVMSizes

SrinivasAtmakuri:

Jira: https://issues.redhat.com/browse/ARO-2883

### DIFF
--- a/pkg/frontend/admin_supportedvmsizes_list_test.go
+++ b/pkg/frontend/admin_supportedvmsizes_list_test.go
@@ -1,0 +1,72 @@
+package frontend
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/go-test/deep"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/api/validate"
+)
+
+func TestSupportedvmsizes(t *testing.T) {
+	mastervmsizes := validate.SupportedVMSizesByRole(validate.VMRoleMaster)
+	workervmsizes := validate.SupportedVMSizesByRole(validate.VMRoleWorker)
+
+	type test struct {
+		name         string
+		vmRole       string
+		wantResponse map[api.VMSize]api.VMSizeStruct
+		wantError    string
+	}
+
+	for _, tt := range []*test{
+		{
+			name:         "vmRole is invalid",
+			vmRole:       "invalidVMRole",
+			wantError:    `400: InvalidParameter: : The provided vmRole 'invalidVMRole' is invalid. vmRole can only be master or worker`,
+			wantResponse: nil,
+		},
+		{
+			name:         "vmRole is empty",
+			vmRole:       "",
+			wantError:    `400: InvalidParameter: : The provided vmRole '' is invalid. vmRole can only be master or worker`,
+			wantResponse: nil,
+		},
+		{
+			name:         "master as vmRole",
+			vmRole:       "master",
+			wantError:    "",
+			wantResponse: mastervmsizes,
+		},
+		{
+			name:         "worker as vmRole",
+			vmRole:       "worker",
+			wantError:    "",
+			wantResponse: workervmsizes,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			f := &frontend{}
+			gotResponse, err := f.supportedVMSizesForRole(tt.vmRole)
+			if err != nil && err.Error() != tt.wantError || err == nil && tt.wantError != "" {
+				t.Error(err)
+			}
+			// if gotResponse and wantResponse are nil, skip
+			if gotResponse != nil || tt.wantResponse != nil {
+				v := map[api.VMSize]api.VMSizeStruct{}
+				err = json.Unmarshal(gotResponse, &v)
+				if err != nil {
+					t.Error(err)
+				}
+				if diff := deep.Equal(v, tt.wantResponse); diff != nil {
+					t.Errorf("unexpected response %s, wanted to match %#v (%s)", string(gotResponse), tt.wantResponse, diff)
+				}
+			}
+		})
+	}
+}
--- a/pkg/frontend/admin_supportedvmsizes_list_test.go
+++ b/pkg/frontend/admin_supportedvmsizes_list_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/api/validate"
+	utilerror "github.com/Azure/ARO-RP/test/util/error"
 )
 
 func TestSupportedvmsizes(t *testing.T) {
@@ -53,15 +54,7 @@ func TestSupportedvmsizes(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			f := &frontend{}
 			gotResponse, err := f.supportedVMSizesForRole(tt.vmRole)
-			// err is not nil, but not the expected error
-			if err != nil && err.Error() != tt.wantError {
-				t.Error(err)
-			}
-			// err is nil but we expected an error
-			if err == nil && tt.wantError != "" {
-				t.Errorf("unexpected error %v , wanted error %s", err, tt.wantError)
-			}
-			// if gotResponse and wantResponse are nil, skip
+			utilerror.AssertErrorMessage(t, err, tt.wantError)
 			if gotResponse != nil || tt.wantResponse != nil {
 				v := map[api.VMSize]api.VMSizeStruct{}
 				err = json.Unmarshal(gotResponse, &v)
--- a/pkg/frontend/admin_supportedvmsizes_list_test.go
+++ b/pkg/frontend/admin_supportedvmsizes_list_test.go
@@ -53,8 +53,13 @@ func TestSupportedvmsizes(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			f := &frontend{}
 			gotResponse, err := f.supportedVMSizesForRole(tt.vmRole)
-			if err != nil && err.Error() != tt.wantError || err == nil && tt.wantError != "" {
+			// err is not nil, but not the expected error
+			if err != nil && err.Error() != tt.wantError {
 				t.Error(err)
+			}
+			// err is nil but we expected an error
+			if err == nil && tt.wantError != "" {
+				t.Errorf("unexpected error %v , wanted error %s", err, tt.wantError)
 			}
 			// if gotResponse and wantResponse are nil, skip
 			if gotResponse != nil || tt.wantResponse != nil {


SrinivasAtmakuri:
Test cases for https://github.com/Azure/ARO-RP/pull/2798

cc @cadenmarchese

pkg/frontend/admin_supportedvmsizes_list_test.go
```diff
@@ -0,0 +1,76 @@
...
+		{
+			name:           "master as vmRole",
+			vmRole:         "master",
+			wantStatusCode: http.StatusOK,
+			wantError:      "",
+			wantResponse:   mastervmsizes,
+		},
+		{
+			name:           "worker as vmRole",
+			vmRole:         "master",
```
cadenmarchese, line 48:
```suggestion
			vmRole:         "worker",
```

pkg/frontend/admin_supportedvmsizes_list_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/api/validate"
+	"github.com/Azure/ARO-RP/pkg/metrics/noop"
+)
+
+func TestSupportedvmsizes(t *testing.T) {
+	ctx := context.Background()
+	mastervmsizes, _ := json.MarshalIndent(validate.SupportedMasterVmSizes, "", "    ")
+	workervmsizes, _ := json.MarshalIndent(validate.SupportedMasterVmSizes, "", "    ")
```
cadenmarchese, line 21:
```suggestion
	workervmsizes, _ := json.MarshalIndent(validate.SupportedWorkerVmSizes, "", "    ")
```
Right now we are testing the master path, but not the worker path.

cadenmarchese: Also, why no error checking here?

cadenmarchese (commented):
This approach looks good to me, just some quick fixes. We are duplicating the marshaling code, but I think that's fine in a unit test context.

pkg/frontend/admin_supportedvmsizes_list_test.go
SrinivasAtmakuri: Added the error check, and corrected the worker path

pkg/frontend/admin_supportedvmsizes_list_test.go
SrinivasAtmakuri: Made the change.

jhoreman (dismissed):
Approach looks good to me.

cadenmarchese (dismissed):

bitoku:
I don't think comparing the json bytes is not a good idea, because `validate.Supported**VmSizes` is a map and a map is not ordered.
I think we should avoid comparing in the way depending on the order of the map.

https://go.dev/ref/spec#Map_types
> A map is an unordered group of elements of one type, called the element type, indexed by a set of unique keys of another type, called the key type.

If you use `*map[api.VMSize]api.VMSizeStruct` for `wantResponse`, you can compare `wantResponse` and its response directly with `validateResponse`, which uses `deep.Equal` and not dependent on the order.
```
	type test struct {
		name           string
		query          string
		wantStatusCode int
		wantResponse   *map[api.VMSize]api.VMSizeStruct
		wantError      string
	}
```

SrinivasAtmakuri:
> I don't think comparing the json bytes is not a good idea, because `validate.Supported**VmSizes` is a map and a map is not ordered. I think we should avoid comparing in the way depending on the order of the map.
> 
> https://go.dev/ref/spec#Map_types
> 
> > A map is an unordered group of elements of one type, called the element type, indexed by a set of unique keys of another type, called the key type.
> 
> If you use `*map[api.VMSize]api.VMSizeStruct` for `wantResponse`, you can compare `wantResponse` and its response directly with `validateResponse`, which uses `deep.Equal` and not dependent on the order.
> 
> ```
> 	type test struct {
> 		name           string
> 		query          string
> 		wantStatusCode int
> 		wantResponse   *map[api.VMSize]api.VMSizeStruct
> 		wantError      string
> 	}
> ```

I wish to use the map[api.VMSize]api.VMSizeStruct` for `wantResponse` so it's a lot easier to compare and I don't need to json.MarshallIndent, but the API returns a []byte response as you can see here., https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/admin_supportvmsizes_list.go#L34

bitoku:
Yes, the API returns []byte response in json format.
but `validateResponse` unmarshals the response if `wantResponse` can't be casted to `[]byte`.

https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/shared_test.go#L262

so when you use `*map[api.VMSize]api.VMSizeStruct` as a `wantResponse`'s type, `validateResponse` unmarshals the response and use `deep.Equal` to compare.

I looked over some APIs that return `[]byte` and their tests.
They just use the pointer of struct as a wantResponse's type and don't marshal it.

https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/admin_openshiftversion_list_test.go
https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/asyncoperationsstatus_get_test.go

SrinivasAtmakuri:
Now I see what you meant, made the changes!

pkg/frontend/admin_supportedvmsizes_list_test.go
```diff
@@ -0,0 +1,86 @@
...
+			if err != nil {
+				t.Error(err)
+			}
+		})
+	}
+}
+
+func Fatal(err error) {
+	panic("unimplemented")
+}
```
facchettos, line 86: Where is this used? I only see `t.Fatal` and not direct calls to that function

facchettos, line 86: Where is this used? I only see `t.Fatal` and not direct calls to that function

pkg/frontend/admin_supportedvmsizes_list_test.go
```diff
@@ -0,0 +1,86 @@
...
+			defer ti.done()
+			f, err := NewFrontend(ctx, ti.audit, ti.log, ti.env, nil, nil, nil, nil, nil, api.APIs, &noop.Noop{}, nil, nil, nil, nil, nil)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			go f.Run(ctx, nil, nil)
+			resp, b, err := ti.request(http.MethodGet,
+				fmt.Sprintf("https://server/admin/supportedvmsizes?vmRole=%s", tt.vmRole),
+				nil, nil)
```
facchettos, line 71: you don't need the testinfra, you can just create an empty `frontend` struct and just test the function directly. You can just call `frontend.GetVMSize` or whatever that function is called instead of creating a whole http request.

facchettos (changes requested):

pkg/frontend/admin_supportedvmsizes_list_test.go
SrinivasAtmakuri: cleaned this.

pkg/frontend/admin_supportedvmsizes_list_test.go
SrinivasAtmakuri: Made the changes.

cadenmarchese (approved):

pkg/frontend/admin_supportedvmsizes_list_test.go
```diff
@@ -0,0 +1,74 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			f := &frontend{}
+			gotResponse, err := f.supportedVMSizesForRole(tt.vmRole)
+			if err != nil && err.Error() != tt.wantError || err == nil && tt.wantError != "" {
+				t.Error(err)
+			}
+			// if gotResponse and wantResponse are nil, skip
+			if gotResponse == nil && tt.wantResponse == nil {
+				t.Skip()
+			}
```
SudoBrendan, line 63: I don't understand this bit - why `Skip`? I'd rather see logic that results in a PASS test result.

pkg/frontend/admin_supportedvmsizes_list_test.go
```diff
@@ -0,0 +1,74 @@
...
+			v := reflect.New(reflect.TypeOf(tt.wantResponse).Elem()).Interface()
```
SudoBrendan, line 64: `reflect` seems like overkill here when we already know this type is going to be `*map[api.VMSize]api.VMSizeStruct`... Is there a strong reason we need to use `reflect` here, or can we just unmarshal the json into a strong type we expect? I'd rather avoid `reflect`'s use at all costs, since it loses us all the benefits of using a strongly typed language.

SudoBrendan (changes requested):
Looking good so far, and a good use of loop testing. I think there's a few concerning bits of test logic, however.

SudoBrendan (approved):
Ultimately, LGTM. One nit if you want to address it, otherwise, close it out :) Nice work.

pkg/frontend/admin_supportedvmsizes_list_test.go
```diff
@@ -0,0 +1,72 @@
...
+			wantError:    "",
+			wantResponse: workervmsizes,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			f := &frontend{}
+			gotResponse, err := f.supportedVMSizesForRole(tt.vmRole)
+			if err != nil && err.Error() != tt.wantError || err == nil && tt.wantError != "" {
+				t.Error(err)
+			}
```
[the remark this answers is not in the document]
SrinivasAtmakuri: Nice catch, agreed, changes made.

facchettos (dismissed):

pkg/frontend/admin_supportedvmsizes_list_test.go
```diff
@@ -0,0 +1,77 @@
...
+			gotResponse, err := f.supportedVMSizesForRole(tt.vmRole)
+			// err is not nil, but not the expected error
+			if err != nil && err.Error() != tt.wantError {
+				t.Error(err)
+			}
+			// err is nil but we expected an error
+			if err == nil && tt.wantError != "" {
+				t.Errorf("unexpected error %v , wanted error %s", err, tt.wantError)
+			}
+			// if gotResponse and wantResponse are nil, skip
```
AldoFusterTurpin, line 64: Can we remove the comment ? I think it just repeats what the code already says.

AldoFusterTurpin (changes requested):
Just two minor changes and we are good to Go! 😄
Thanks!

pkg/frontend/admin_supportedvmsizes_list_test.go
SrinivasAtmakuri: Yes, done

pkg/frontend/admin_supportedvmsizes_list_test.go
```diff
@@ -0,0 +1,77 @@
...
+			f := &frontend{}
+			gotResponse, err := f.supportedVMSizesForRole(tt.vmRole)
+			// err is not nil, but not the expected error
+			if err != nil && err.Error() != tt.wantError {
+				t.Error(err)
+			}
+			// err is nil but we expected an error
+			if err == nil && tt.wantError != "" {
+				t.Errorf("unexpected error %v , wanted error %s", err, tt.wantError)
+			}
```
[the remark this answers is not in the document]
SrinivasAtmakuri: Good to know, change made.

facchettos (approved):

AldoFusterTurpin (approved):
LGTM.

PR #2869, opened 2023-04-25
Adding version in log

sankur-codes:

Fixes [ARO-2786](https://issues.redhat.com/browse/ARO-2786)

This PR includes a new field **_`version`_** in RP logs which has the value of Openshift cluster version. This is later intended to be able to grab it out of the logs and update our dashboards based on cluster installation or failure on it

### Test plan for issue:

Ran RP in development mode. version parameter was populated when cluster was available. 

No


pkg/portal/assets/v2/build/asset-manifest.json
```diff
@@ -1,8 +1,10 @@
 {
   "files": {
-    "main.js": "/static/js/main.61fbb3a0.js",
+    "main.js": "/static/js/main.0b4eab3b.js",
```
ellis-johnson, line 3: Why are all these portal assets included? Was this an accidental change?

ellis-johnson (changes requested):
LGTM! Have one comment about left over portal changes and still needs tests

sankur-codes:
> LGTM! Have one comment about left over portal changes

hawkowl (dismissed):
We already collect this information in Geneva for dashboards -- is there a usage gap that I am not aware of?

ellis-johnson:
> We already collect this information in Geneva for dashboards -- is there a usage gap that I am not aware of?

Where abouts is it stored? Or where do we use it in geneva? The premise behind this change is to use this information to check install success between different versions

SudoBrendan (approved):
This LGTM if it's needed, pending discussions already here.

bennerv:
@hawkowl
> We already collect this information in Geneva for dashboards -- is there a usage gap that I am not aware of?

Our cluster failure dashboards key off of cluster logs to determine why a cluster failed to install, and report it up in a dashboard.  So the goal is to aggregate dashboards + minor versions + failure reasons.  It's not as simple as using existing metrics as a result :/

PR #2875, opened 2023-04-27
Fix reading the AppLens resp to return Json string

cloudfit-nick:
Fix a problem where the json.Marshal(resp.Body) was returning an {} instead of the expected Json string


weinong (approved):
LGTM

niontive (approved):

s-amann (approved):

PR #2876, opened 2023-04-27
Wrap errors into CloudError

nwnt:

Fixes ARO-2934

When cluster sp has no permissions on the target vnet, ARM returns HTTP 403.  The current logic doesn't account for such scenario and would return it without wrapping into a CloudError.  This ultimately makes the RP return "Internal Server Error" which obscures what's actually happening.  

This PR addresses such issue by having a dedicated case for 403 responses and also wrapping all other errors into a CloudError.

### Test plan for issue:
Added unit tests and also tested with local dev environment by reproducing the issue and validating the fix.

No.  This should be the expected behavior of the RP.


pkg/api/validate/dynamic/dynamic.go
```diff
@@ -197,27 +197,41 @@ func (dv *dynamic) validateVnetPermissions(ctx context.Context, vnet azure.Resou
...
+		case http.StatusNotFound:
+			return api.NewCloudError(
+				http.StatusBadRequest,
+				api.CloudErrorCodeInvalidLinkedVNet,
+				"",
+				"The vnet '%s' could not be found.",
+				vnet.String(),
+			)
+		case http.StatusForbidden:
+			return noPermissionsErr
```
weinong, line 223: i think we should also include the detailed error that will provide more clarity

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -197,27 +197,41 @@ func (dv *dynamic) validateVnetPermissions(ctx context.Context, vnet azure.Resou
...
+		}
 	}
-	return err
+	if err != nil { // avoiding Internal Server Error
```
weinong, line 226: i'm not sure about this catch all clause that turns everything to user error. what if they are not?

s-amann: +1 we should only categorize errors as a user error when we are sure. The general case should remain service error / internal server error

s-amann (changes requested):

pkg/api/validate/dynamic/dynamic.go
nwnt: done.

pkg/api/validate/dynamic/dynamic.go
nwnt: This now includes the original error message.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -197,27 +197,46 @@ func (dv *dynamic) validateVnetPermissions(ctx context.Context, vnet azure.Resou
 		"Microsoft.Network/virtualNetworks/subnets/write",
 	})
 
+	noPermissionsErr := api.NewCloudError(
+		http.StatusBadRequest,
+		errCode,
+		"",
+		"The %s service principal does not have Network Contributor permission on vnet '%s'.",
```
weinong, line 204:
```suggestion
		"The %s service principal does not have Network Contributor role on vnet '%s'.",
```

weinong: is it possible to include the application ID? say, if `authorizerType` is `AuthorizerFirstParty`, the user still wouldn't know what app ID to give role to, right?

pkg/api/validate/dynamic/dynamic.go
weinong: with checkAccessV2, will it return Forbidden at all?

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -197,27 +197,46 @@ func (dv *dynamic) validateVnetPermissions(ctx context.Context, vnet azure.Resou
...
+				"%s Original error message: %s",
+				noPermissionsErr.Message,
+				detailedErr.Message,
+			)
+			return noPermissionsErr
+		}
 	}
-	return err
+	if err != nil { // avoiding blank Internal Server Error
+		return &api.CloudError{
```
weinong, line 232: you don't need to return CloudError, which is only used when it's user error. returning `err` as is will result in internal server error like it has been doing.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -197,27 +197,46 @@ func (dv *dynamic) validateVnetPermissions(ctx context.Context, vnet azure.Resou
...
-		detailedErr.StatusCode == http.StatusNotFound {
-		return api.NewCloudError(
-			http.StatusBadRequest,
-			api.CloudErrorCodeInvalidLinkedVNet,
-			"",
-			"The vnet '%s' could not be found.",
-			vnet.String(),
-		)
+	if detailedErr, ok := err.(autorest.DetailedError); ok {
+		switch detailedErr.StatusCode {
```
weinong, line 213: can we always log detailedErr? that will help if we miss catching anything in the future

SudoBrendan (commented):
Thanks for the PR, great work so far. I've got nothing to add that hasn't already been commented on - in particular, adding the ID of the identity that failed the permissions check.

pkg/api/validate/dynamic/dynamic.go
s-amann: +1 we do not need to duplicate the logic we have to wrap uncaught errors as internal server errors. The logic can be viewed here: https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/middleware/panic.go#L23

s-amann (changes requested):
please unwrap the default error case as the logic is duplicated, and will currently interfere with how we report errors internally.

pkg/api/validate/dynamic/dynamic.go
nwnt: Yes, it can return forbidden as well but we're only going to use the FP to call it.  That's going to be a more of a fatal case.

pkg/api/validate/dynamic/dynamic.go
nwnt: The linked code above and [this](https://github.com/Azure/ARO-RP/blob/12aa588dbc7c95c07817bda86bb0b57774259eb7/pkg/backend/openshiftcluster.go#L256) are what I'm trying to address (see the original issue from a Hashicorp engineer [here](https://github.com/Azure/azure-rest-api-specs/issues/23355#issuecomment-1522415332).  When it is not a cloud error, the original message is lost and only "Internal Server Error" is sent back to the caller.

What I had in mind was most unexpected/uncaught errors during the validation were interactions with ARM, so it might be ok to package the original error message and return to the caller so that they have an idea about what's happening.  Alternatively, we could make the locations above include the original message, but I'm not so sure if that's a good idea.. if we're not careful, it could send irrelevant or even leak some sensitive info.

What do you think about this?  I'm open to suggestions here.

pkg/api/validate/dynamic/dynamic.go
nwnt: done.  changed from permission to role and included the appID field which is printed in the error message.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -197,27 +197,46 @@ func (dv *dynamic) validateVnetPermissions(ctx context.Context, vnet azure.Resou
...
+			return api.NewCloudError(
+				http.StatusBadRequest,
+				api.CloudErrorCodeInvalidLinkedVNet,
+				"",
+				"The vnet '%s' could not be found.",
+				vnet.String(),
+			)
+		case http.StatusForbidden:
+			noPermissionsErr.Message = fmt.Sprintf(
+				"%s Original error message: %s",
```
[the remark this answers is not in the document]
nwnt: done

pkg/api/validate/dynamic/dynamic.go
nwnt: done

pkg/api/validate/dynamic/dynamic.go
nwnt: Unwrapped now.

s-amann (approved):

PR #2877, opened 2023-04-27
Update the mega linter because of GitHub Action deprecations

hawkowl:
Chore from warnings in github

Fixes up deprecation warnings

### Test plan for issue:

It is tests :)

N/A

### DIFF
--- a/.mega-linter.yml
+++ b/.mega-linter.yml
@@ -1,14 +1,15 @@
 # Configuration file for MegaLinter
-# See all available variables at https://megalinter.github.io/configuration/ and in linters documentation
+# See all available variables at https://oxsecurity.github.io/megalinter/configuration/ and in linters documentation
 
 APPLY_FIXES: none
-# ENABLE:  # If you use ENABLE variable, all other languages/formats/tooling formats will be disabled by default
 ENABLE_LINTERS:
-  - YAML_YAMLLINT # If you use ENABLE_LINTERS variable, all other linters will be disabled by default
+  - YAML_YAMLLINT
 EXCLUDED_DIRECTORIES:
-  - vendor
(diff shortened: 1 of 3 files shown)


.mega-linter.yml
```diff
@@ -1,14 +1,15 @@
...
 APPLY_FIXES: none
-# ENABLE:  # If you use ENABLE variable, all other languages/formats/tooling formats will be disabled by default
 ENABLE_LINTERS:
-  - YAML_YAMLLINT # If you use ENABLE_LINTERS variable, all other linters will be disabled by default
+  - YAML_YAMLLINT
 EXCLUDED_DIRECTORIES:
-  - vendor
+  - "vendor"
+  - "node_modules"
 YAML_YAMLLINT_CONFIG_FILE: .yaml-lint.yml
```
SudoBrendan, line 10: just noticing that the file edited in this PR and this reference do not align. Am I missing something or is this a typo from a previous commit? The new linter seems to use a different default, one that would work if we actually removed this "custom" one: https://megalinter.io/latest/descriptors/yaml_yamllint/

SudoBrendan (commented):
Changes look good, I wonder if we need to modify the configured ref to the yamllint config file path while using the new linter?

.mega-linter.yml
hawkowl: the path is still different, dunno why we have that name in particular. I'm not too keen to change it if it works :)

hawkowl: OH, right. This yaml-lint is for the yaml-lint config in the proj root dir, the one changed in the PR is for the yaml linting GH action.

ellis-johnson (approved):
LGTM!

PR #2886, opened 2023-05-02
log correlation data in outbound requests

AldoFusterTurpin:

Fixes [Log outbound requests correlationIds and/or clientRequestIds](https://issues.redhat.com/browse/ARO-3406) 

We encourage you to check the [SPIKE](https://docs.google.com/document/d/1RbnKKPNjw7kJZeR-2me4euM5CClGqF62FkaHVlWGxtM/edit?usp=sharing) we created to understand what this is about as it is not a trivial PR.

During investigations on RP issues, we faced difficulties to find out why some messages fail to be sent by the ARO-RP to ARM. Logs of messages (correlation data) sent by ARO RP to ARM are missing to help know what actually happened to the requests made by the RP, and in particular if they were actually sent to ARM.

The purpose of this is to have at least correlationID of outbound requests done by the ARO RP logged in ARO RP logs so that troubleshooting is possible for such complex cases.

The PR includes changes about the config of the log of outgoing requests to its own Geneva/Kusto table using a different marker like [this](https://github.com/Azure/ARO-RP/blob/a03c6ebd5f187b0028eb777098d89ee07c8cb06f/pkg/backend/openshiftcluster.go#L325), similar to HttpIncoming and Outgoing request tables in ARM log (modifcation of the fluentbit config in pkg/deploy/generator/scripts/rpVMSS.sh), also explained in the Spike document mentioned before.

### Test plan for issue:

Unit Tests, e2e tests and tested this PR in INT with the corresponding Geneva config changes.

The [spike document](https://docs.google.com/document/d/1RbnKKPNjw7kJZeR-2me4euM5CClGqF62FkaHVlWGxtM/edit?usp=sharing) contains detailed information that can be used as reference for specific topics as it was written in general terms


pkg/util/azureclient/roundtripper.go
```diff
@@ -0,0 +1,84 @@
...
+	if correlationData == nil {
+		correlationData = api.CreateCorrelationDataFromReq(req)
+	}
+
+	l := updateCorrelationDataAndEnrichLog(correlationData, utillog.GetLogger(), startTime)
+	l.Info("HttpRequestStart")
+
+	res, err := crt.next.RoundTrip(req)
+
+	l = updateCorrelationDataAndEnrichLogWithResponse(correlationData, l, res, startTime)
```
mociarain, line 44: I think I'm missing something here related to the RequestTime/StartTime.

1. Naming. Should we unify it to StartTime or RequestTime?
2. Why not just set the RequestTime when creating CorrelationData?

pkg/api/correlation.go
```diff
@@ -1,12 +1,22 @@
...
+	"github.com/Azure/ARO-RP/pkg/util/uuid"
 )
 
 // Copyright (c) Microsoft Corporation.
 // Licensed under the Apache License 2.0.
 
+type contextKey int
+
+const (
+	contextKeyCorrelationData contextKey = iota
```
mociarain, line 17: What is the purpose of using `iota` here?

pkg/util/azureclient/decorators.go
```diff
@@ -0,0 +1,38 @@
...
+
+func DecorateSenderWithLogging(sender autorest.Sender) autorest.Sender {
+	loggingDecorator := LoggingDecorator()
+	return autorest.DecorateSender(sender, loggingDecorator, autorest.DoCloseIfError())
+}
+
+// LoggingDecorator returns a function which is used to wrap and modify
+// the behaviour of an autorest.Sender.
+// Azure Clients will have the sender wrapped by that function
+// in order to intercept http calls using our custom RoundTripper (through the adapter).
```
mociarain, line 20: This comment (which I've likely misunderstood) seems to be explaining what the following code does and (weak challenge) seems unneeded. What do you think?

pkg/util/azureclient/decorators.go
```diff
@@ -0,0 +1,38 @@
...
+
+// roundTripperAdapter converts from autorest.Sender (internal field)
+// to http.RoundTripper (external method).
+type roundTripperAdapter struct {
+	Sender autorest.Sender
+}
+
+func (rta *roundTripperAdapter) RoundTrip(req *http.Request) (*http.Response, error) {
+	return rta.Sender.Do(req)
+}
```
mociarain, line 38: I don't fully understand this part and if you have a little time I'd appreciate going over this.

pkg/util/azureclient/roundtripper.go
```diff
@@ -0,0 +1,84 @@
...
+	l.Info("HttpRequestEnd")
+
+	return res, err
+}
+
+// updateCorrelationDataAndEnrichLog receives a non nil correlationData and updates the start time. It also returns a new logrus entry
+// updated with the new field for the LOGKIND field, corresponding to a specific custom DGrep table for outgoing requests.
+func updateCorrelationDataAndEnrichLog(correlationData *api.CorrelationData, l *logrus.Entry, startTime time.Time) *logrus.Entry {
```
mociarain, line 52: Is this type of docstring comment a convention we follow?

pkg/util/azureclient/roundtripper.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	l = utillog.EnrichWithCorrelationData(l, correlationData)
+
+	fields := logrus.Fields{
+		"response_status_code":   responseStatusCode,
+		"contentLength":          contentLength,
+		"durationInMilliseconds": time.Since(startTime).Milliseconds(),
+	}
+	return l.WithFields(fields)
```
mociarain, line 83: If I understand this correctly we lose the `CorrelationID` from the logs if the response is `nil`... that seems odd to me. Why don't we want to be apply to correlate such a case?

mociarain (commented):
Sorry for all the "new guy" questions but hopefully some of this is helpful to you as well as me :wink:

pkg/api/correlation.go
```diff
@@ -1,12 +1,22 @@
...
+)
+
 // CorrelationData represents any data, used for metrics or tracing.
 // More on these values:
 // https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/common-api-details.md
```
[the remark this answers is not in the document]
AldoFusterTurpin: True. I have replaced it with: https://eng.ms/docs/products/arm/rpaas/contract/requestheaders

You can check the Spike doc for a explanation about the correlation data 🙂 

Thanks!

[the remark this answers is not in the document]
mociarain: So that link is only accessible to users with a B-account (AFAIK). If that's true I'm not exactly sure what my point is but it's another thing that makes having this repo public a bit... strange. :man_shrugging:

pkg/api/correlation.go
```diff
@@ -26,3 +39,31 @@ type CorrelationData struct {
...
+	return correlationData
+}
+
+func CreateCorrelationDataFromReq(req *http.Request) *CorrelationData {
+	if req == nil {
+		return nil
+	}
+
+	h := req.Header
+	dg := uuid.DefaultGenerator
```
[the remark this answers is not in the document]
AldoFusterTurpin: Before learning Go I tended to use long descriptive names, but in Go there is the well known convention to use single letter or short names for variables that have a narrow scope. They are pretty common and are present in the Go source code and in many projects (req for request, r for Reader, w for Writer, buf for buffer, etc.).

More info in the Google [styleguide](https://google.github.io/styleguide/go/decisions#variable-names)

I have inlined the variables as it is simpler. Thanks!

pkg/api/correlation.go
```diff
@@ -26,3 +39,31 @@ type CorrelationData struct {
 	// RequestTime is the time that the request was received
 	RequestTime time.Time `json:"requestTime,omitempty"`
 }
+
+func ContextWithCorrelationData(ctx context.Context, correlationData *CorrelationData) context.Context {
+	return context.WithValue(ctx, contextKeyCorrelationData, correlationData)
+}
+
+func GetCorrelationDataFromCtx(ctx context.Context) *CorrelationData {
```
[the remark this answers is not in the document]
AldoFusterTurpin: You're right. Changed ContextWithCorrelationData to CtxWithCorrelationData. Thanks.

pkg/api/correlation.go
AldoFusterTurpin: We could use an integer, but using iota we have the benefit that if we add more const types below _contextKeyCorrelationData_ it will automatically have an int assigned. It is just about making it more extensible. Without the iota we need to manually assign a number for this constant key and the new ones in the future. It could be the literal 0, though. I don't see any drawback about using the iota here (I could be wrong). 🙂 

The important thing is to provide a function at package level that allows you to create a context with the given value (without exposing the key used as it is just an implementation detail that could change in the future).

pkg/util/azureclient/decorators.go
AldoFusterTurpin: In general terms I do not like to put comments that explain what the code does (but instead why-> explaining what code can not explain).

In that case is different. I decided to explain the what as the usage of this wrapper function is new and can be a bit misleading to people not used to the azure sender thingy. I just added it tho clarify where the function is intended to be used. I am happy to delete the comment if we find it trivial.

pkg/util/azureclient/decorators.go
AldoFusterTurpin: It is explained in the spike document as it is needed to modify the deprecated Azure clients. We can talk about that later in a call to clarify it if you want 🙂

pkg/util/azureclient/roundtripper.go
AldoFusterTurpin: Comments in Go have a special convention followed by the community: "Go Doc Comments". We don't always use them in the project but I find them useful sometimes to explain functions that are not trivial. Ideally you want to document the public API of your packages, but we don't do that strictly->this is not the case, just a general comment -never better said : ) 

In that case I think the comment makes sense because:
- it explicitly says that correlation data should not be nil (explicit precondition)
- it adds context about what is the LOGKIND field as it is a trivial convention that we use to configure our logging to write to a different table (this is also explained in the Spike as it is not trivial)

Here is the explanation of the standard convention: https://tip.golang.org/doc/comment

WDYT ?

pkg/util/azureclient/roundtripper.go
AldoFusterTurpin: I was testing it (changed that a couple of times) to see the behavior. Updated that to just modify the correlation data when res is not nil. Thanks!

pkg/api/correlation.go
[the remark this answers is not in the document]
AldoFusterTurpin: AFAIK everything under eng.ms requires b- account.

AldoFusterTurpin:
> Sorry for all the "new guy" questions but hopefully some of this is helpful to you as well as me 😉

Every review is welcome. No "sorry" 🙂

pkg/util/azureclient/roundtripper.go
mociarain: That's perfect. Ty for the context

AldoFusterTurpin: No problem. Sometimes the line between useful and redundant in that kind of comments is not 100% clear.

mociarain (approved):

bennerv (commented):
echoing one of my questions in the review session.

The logging of this data relies on updating the HTTP client/senders for the associated azure-sdk-for-go client.  

is there a way we can have a linter rule or something in place to ensure we don't miss this in future clients?  

I wouldn't worry about putting it in this PR, but just brainstorming and jotting a card down to prevent drift / inaccuracies in the "tracing"/"mapping" of outbound requests + inbound correlation data.

AldoFusterTurpin:
> echoing one of my questions in the review session.
> 
> The logging of this data relies on updating the HTTP client/senders for the associated azure-sdk-for-go client.
> 
> is there a way we can have a linter rule or something in place to ensure we don't miss this in future clients?
> 
> I wouldn't worry about putting it in this PR, but just brainstorming and jotting a card down to prevent drift / inaccuracies in the "tracing"/"mapping" of outbound requests + inbound correlation data.

Great question!
(IMHO) I think it would be possible to write a validation Go code similar to what we have for "validate imports" but for the azure clients, applying a rule for everything under pkg/util/azureclient/*. I see that as a better option than unit testing (as the test per see would need to be able to traverse all the target files).

AldoFusterTurpin:
As we no longer support PRs from Forks, I am closing this PR. The new one in upstream is [here](https://github.com/Azure/ARO-RP/pull/3503).

PR #2887, opened 2023-05-02
Fix failure in az aro update when missing azure-credentials object

dem4gus:

Fixes [ARO-2898](https://issues.redhat.com/browse/ARO-2898)

Fixes

### Test plan for issue:

Unit tests updated for new functionality.

Manual tests performed on locally running RP.
1. Create a new cluster and get admin kubeconfig
1. Save existing secret `oc get secret -n kube-system azure-credentials -o yaml > azure-credentials-old.yaml`
1. Delete secret from cluster `oc delete secret -n kube-system azure-credentials`
1. Run `az aro update` on cluster
1. Get newly created secret `oc get secret -n kube-system azure-credentials -o yaml > azure-credentials-new.yaml`
1. Compare secret contents `diff --color azure-credentials-old.yaml azure-credentials-new.yaml`

None

### DIFF
--- a/pkg/cluster/clusterserviceprincipal.go
+++ b/pkg/cluster/clusterserviceprincipal.go
@@ -185,27 +185,7 @@ func (m *manager) updateOpenShiftSecret(ctx context.Context) error {
 		// azure_tenant_id: tenant_id
 		secret, err := m.kubernetescli.CoreV1().Secrets(clusterauthorizer.AzureCredentialSecretNameSpace).Get(ctx, clusterauthorizer.AzureCredentialSecretName, metav1.GetOptions{})
 		if kerrors.IsNotFound(err) {
-			// rebuild secret
-			resourceGroupID := m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID
-			secret = &corev1.Secret{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      clusterauthorizer.AzureCredentialSecretName,
-					Namespace: clusterauthorizer.AzureCredentialSecretNameSpace,
-				},
-				Type: corev1.SecretTypeOpaque,
-				Data: map[string][]byte{
-
-					"azure_subscription_id": []byte(m.subscriptionDoc.ID),
-					"azure_resource_prefix": []byte(m.doc.OpenShiftCluster.Properties.InfraID),
-					"azure_resourcegroup":   []byte(resourceGroupID[strings.LastIndex(resourceGroupID, "/")+1:]),
-					"azure_region":          []byte(m.doc.OpenShiftCluster.Location),
-					// values set below
-					"azure_client_id":     []byte(""),
-					"azure_client_secret": []byte(""),
-					"azure_tenant_id":     []byte(""),
-				},
-			}
-			recreate = true
+			secret = m.newAzureCredentialSecret()
 		} else if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


s-amann (dismissed):

SudoBrendan (dismissed):
LGTM

pkg/cluster/clusterserviceprincipal.go
```diff
@@ -180,10 +182,31 @@ func (m *manager) updateOpenShiftSecret(ctx context.Context) error {
 		// azure_client_id: secret_id
 		// azure_client_secret: secret_value
 		// azure_tenant_id: tenant_id
-		secret, err := m.kubernetescli.CoreV1().Secrets("kube-system").Get(ctx, "azure-credentials", metav1.GetOptions{})
-		if err != nil {
+		secret, err := m.kubernetescli.CoreV1().Secrets(clusterauthorizer.AzureCredentialSecretNameSpace).Get(ctx, clusterauthorizer.AzureCredentialSecretName, metav1.GetOptions{})
+		if err != nil && !kerrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
dem4gus: Done

mbarnes (dismissed):

s-amann (dismissed):

pkg/cluster/clusterserviceprincipal.go
```diff
@@ -173,15 +175,36 @@ func (m *manager) updateAROSecret(ctx context.Context) error {
...
 	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		//data:
 		// azure_client_id: secret_id
 		// azure_client_secret: secret_value
 		// azure_tenant_id: tenant_id
-		secret, err := m.kubernetescli.CoreV1().Secrets("kube-system").Get(ctx, "azure-credentials", metav1.GetOptions{})
-		if err != nil {
+		secret, err := m.kubernetescli.CoreV1().Secrets(clusterauthorizer.AzureCredentialSecretNameSpace).Get(ctx, clusterauthorizer.AzureCredentialSecretName, metav1.GetOptions{})
+		if kerrors.IsNotFound(err) {
+			// rebuild secret
```
cadenmarchese, line 188: Do we have to construct the secret here, or can we just set `recreate = true` here and move the secret construction somewhere else for cleanliness?

pkg/cluster/clusterserviceprincipal.go
```diff
@@ -173,15 +175,36 @@ func (m *manager) updateAROSecret(ctx context.Context) error {
...
+			secret = &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      clusterauthorizer.AzureCredentialSecretName,
+					Namespace: clusterauthorizer.AzureCredentialSecretNameSpace,
+				},
+				Type: corev1.SecretTypeOpaque,
+				Data: make(map[string][]byte),
```
cadenmarchese, line 195: Why not use the composite literal?
```suggestion
				Data: map[string][]byte{
				     "azure_subscription_id": []byte(m.subscriptionDoc.ID),
				     etc...
				},
```

cadenmarchese (commented):
Thanks for the fix! I have a few suggestions before merge.

SudoBrendan (changes requested):
I feel like this whole PR effectively replicates existing functionality in CoreV1().Secrets.Apply() function - can we just construct the secret we want, and then `Apply()` it unconditionally? It removes all the logic we currently have, and reduces api calls to exactly 1 instead of `GET` followed by error checking and `CREATE` or `UPDATE`. I'd prefer that we leverage the (heavily tested and proven) desired state machine of K8s to do all the work for us.

pkg/cluster/clusterserviceprincipal.go
dem4gus: Moved

pkg/cluster/clusterserviceprincipal.go
dem4gus: Why not indeed.

dem4gus:
@SudoBrendan It took a lot more doing than I thought it would, but I managed to get it to use `Apply()` rather than `Create()`/`Update()`. There's an outstanding bug with `fake.Clientset` (kubernetes/client-go#1184) that makes `Apply` not work properly but I was able to use the same Reactor that's in the bug's discussion.

cadenmarchese (approved):
lgtm

kimorris27 (approved):

PR #2889, opened 2023-05-03
Update operator version when version.GitCommit changes

ventifus:

Fixes operator version not updating

Currently, the operator sets it's version when it's first installed, but then never updates it even if the operator is updated.

### DIFF
--- a/pkg/operator/controllers/clusteroperatoraro/clusteroperatoraro_controller.go
+++ b/pkg/operator/controllers/clusteroperatoraro/clusteroperatoraro_controller.go
@@ -94,7 +94,15 @@ func (r *Reconciler) setClusterOperatorStatus(ctx context.Context, originalClust
 		Reason:             reasonAsExpected,
 	})
 
-	if equality.Semantic.DeepEqual(clusterOperatorObj.Status.Conditions, originalClusterOperatorObj.Status.Conditions) {
+	clusterOperatorObj.Status.Versions = []configv1.OperandVersion{
+		{
+			Name:    "operator",
+			Version: version.GitCommit,
+		},
+	}
+
+	if equality.Semantic.DeepEqual(clusterOperatorObj.Status.Conditions, originalClusterOperatorObj.Status.Conditions) &&
+		equality.Semantic.DeepEqual(clusterOperatorObj.Status.Versions, originalClusterOperatorObj.Status.Versions) {
 		return nil
 	}
 


tsatam (approved):
LGTM, thanks for this! 

I wonder if it might make sense to perform this version diff+change in a separate controller and not have it run on every change to the aro cluster resource as is necessary for conditions updates, but I think it's simpler this way and fine as is.

pkg/operator/controllers/clusteroperatoraro/clusteroperatoraro_controller.go
```diff
@@ -94,7 +94,15 @@ func (r *Reconciler) setClusterOperatorStatus(ctx context.Context, originalClust
...
-	if equality.Semantic.DeepEqual(clusterOperatorObj.Status.Conditions, originalClusterOperatorObj.Status.Conditions) {
+	clusterOperatorObj.Status.Versions = []configv1.OperandVersion{
+		{
+			Name:    "operator",
+			Version: version.GitCommit,
+		},
+	}
+
+	if equality.Semantic.DeepEqual(clusterOperatorObj.Status.Conditions, originalClusterOperatorObj.Status.Conditions) &&
+		equality.Semantic.DeepEqual(clusterOperatorObj.Status.Versions, originalClusterOperatorObj.Status.Versions) {
```
facchettos, line 105: what's the purpose of the `equality` deepequal compared to the `reflect` one? (it's probably valid but i'm still curious)

tsatam: I originally added the first check when handling operator conditions. I lifted the comparison from [openshift/library-go](https://github.com/openshift/library-go/blob/d7e4a8ca058837a972a4cdef06334be4f7becdde/pkg/operator/status/status_controller.go#L208)'s equivalent status controller, whose helper methods I used directly but we did not use their controller itself.

tsatam: Looking at [the implementation](https://github.com/kubernetes/apimachinery/blob/v0.27.1/pkg/api/equality/semantic.go), it appears to just be a wrapper of (a fork of) golang reflect.DeepEqual with some thoughtful k8s-specific equality functions provided.

SudoBrendan (approved):

cadenmarchese (approved):
lgtm

PR #2893, opened 2023-05-04
Ad-hoc Hive Image with Red Hat Package Updates at Old Commit

bennerv:

There's an issue in hive in production where there appears to be memory leaks.  We also have to take into account older versions of hive have vulnerabilities.  

This hive image is a hotfix image rebuilt at the commit we were using which was good, but should have red hat update packages so no P0s :+1: 

### Test plan for issue:

Test in INT + roll to prod.  

Nope


s-amann (approved):

cadenmarchese (approved):

bennerv:
Status are showing as running, but they're green in ADO.  Going to force merge

PR #2895, opened 2023-05-05
Default to v5, but use v4 or v3 if unavailable

cadenmarchese:

We want to default to the v5 or v4 instances when no instances are specified because they have better performance, but the v5 and v4 aren't available in all regions and subscriptions, so we need to fall back to v3 in those cases.

It's a draft PR / proposal, feel free to make suggestions.

### Test plan for issue:

- [x] unit test coverage
- [x] manual testing with `go run ./hack/cluster create`
- [x] manual testing wtih `az aro create`

### DIFF
--- a/pkg/frontend/sku_validation.go
+++ b/pkg/frontend/sku_validation.go
@@ -45,20 +45,46 @@ func validateVMSku(ctx context.Context, oc *api.OpenShiftCluster, resourceSkusCl
 	}
 
 	filteredSkus := computeskus.FilterVMSizes(skus, location)
-	masterProfileSku := string(oc.Properties.MasterProfile.VMSize)
 
-	err = checkSKUAvailability(filteredSkus, location, "properties.masterProfile.VMSize", masterProfileSku)
-	if err != nil {
+	// If the default machine size (Standard_D8s_v5) isn't available, use Standard_D8s_v4, or Standard_D8s_v3
+	err = checkSKUAvailability(filteredSkus, location, "properties.masterProfile.VMSize", string(oc.Properties.MasterProfile.VMSize))
+	if err != nil && oc.Properties.MasterProfile.VMSize == api.VMSizeStandardD8sV5 {
+		oc.Properties.MasterProfile.VMSize = api.VMSizeStandardD8sV4
+
+		err = checkSKUAvailability(filteredSkus, location, "properties.masterProfile.VMSize", string(oc.Properties.MasterProfile.VMSize))
+		if err != nil && oc.Properties.MasterProfile.VMSize == api.VMSizeStandardD8sV4 {
+			oc.Properties.MasterProfile.VMSize = api.VMSizeStandardD8sV3
+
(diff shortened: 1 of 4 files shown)


bennerv:
Does ProdMgmt know about this coming change?  They might want to communicate?

pkg/frontend/sku_validation.go
```diff
@@ -45,20 +45,46 @@ func validateVMSku(ctx context.Context, oc *api.OpenShiftCluster, resourceSkusCl
...
+				workerprofile.VMSize = api.VMSizeStandardD4sV3
+
+				err = checkSKUAvailability(filteredSkus, location, fmt.Sprintf("properties.workerProfiles[%d].VMSize", i), string(workerprofile.VMSize))
+				if err != nil {
+					return err
+				}
+			} else if err != nil {
+				return err
+			}
+		} else if err != nil {
```
SudoBrendan, line 87: This looks like the same logic twice, once for masters, once for workers. Is there a way to extract this out into a single function so we can maintain the logic in one place?

cadenmarchese: I gave that a try locally, but even though we're calling the same function in the same sequence for both workers and masters, only `filteredSkus` and `location` are the same between master and workers. Everything else - `path`, `profile`, and default size for workers vs masters, which changes on each attempt - is different, so it would all need to be parameterized and passed in to this new function. I think that having a separate function wouldn't simplify things much, especially since we already have `checkSKUAvailability()`, but I am definitely open to further ideas or pairing on it.

pkg/frontend/sku_validation.go
```diff
@@ -45,20 +45,46 @@ func validateVMSku(ctx context.Context, oc *api.OpenShiftCluster, resourceSkusCl
...
 
-	err = checkSKUAvailability(filteredSkus, location, "properties.masterProfile.VMSize", masterProfileSku)
-	if err != nil {
+	// If the default machine size (Standard_D8s_v5) isn't available, use Standard_D8s_v4, or Standard_D8s_v3
+	err = checkSKUAvailability(filteredSkus, location, "properties.masterProfile.VMSize", string(oc.Properties.MasterProfile.VMSize))
+	if err != nil && oc.Properties.MasterProfile.VMSize == api.VMSizeStandardD8sV5 {
+		oc.Properties.MasterProfile.VMSize = api.VMSizeStandardD8sV4
+
+		err = checkSKUAvailability(filteredSkus, location, "properties.masterProfile.VMSize", string(oc.Properties.MasterProfile.VMSize))
+		if err != nil && oc.Properties.MasterProfile.VMSize == api.VMSizeStandardD8sV4 {
```
s-amann, line 55: I don't yet have the full idea, but with hard coded conditionals like these we will have more work when we have new versions / removing older versions. I was thinking instead, we could have a list of versions to iterate through for masters/workers, and have the list be ordered such we try v4 before v5. If we can do that then maybe we can also reduce the seemingly duplicated blocks of conditions.

cadenmarchese:
Going to close this one for now - will re-open in case this is something we want to consider.

PR #2896, opened 2023-05-08
Refactor the dynamic validator into FPSP/CSP so that they can individually retry

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-2990

Refactors the dynamic validator so that the FPSP/CSP checking is more clearly delineated. Move around some of the implementation so that it's separated more. Attempt to fix up the interfaces a little.

### Test plan for issue:

E2E, CI, people looking real hard at it

N/A


pkg/api/validate/dynamic/dynamic_test.go
s-amann, line 1: nit: since we are splitting out network validation we should also split out its tests to `vnet_tests`

PR #2897, opened 2023-05-08
Remove go vet from validate-go in Makefile

bitoku:

`make validate-go` generates an error when it runs `go vet` because there is only nolint directive for `golangci-lint` not for `go vet`.

```
❯ make validate-go
gofmt -s -w cmd hack pkg test
go run ./vendor/golang.org/x/tools/cmd/goimports -w -local=github.com/Azure/ARO-RP cmd hack pkg test
go run ./hack/validate-imports cmd hack pkg test
go run ./hack/licenses
go vet -tags containers_image_openpgp ./...
# github.com/Azure/ARO-RP/pkg/backend
pkg/backend/openshiftcluster.go:264:15: assignment copies lock value to ocCopy: github.com/Azure/ARO-RP/pkg/api.OpenShiftCluster contains sync.Mutex
make: *** [validate-go] Error 1
```

https://github.com/Azure/ARO-RP/blob/be4bae3d0ce01c2c9ba3cc7fe7bf30b0a1cd78c2/pkg/backend/openshiftcluster.go#L262-L264

`go vet` is now included in `golangci-lint`, and `golangci-lint` can be run already by `lint-go` action in Makefile.
so `go vet` command is no longer required.

https://github.com/Azure/ARO-RP/blob/be4bae3d0ce01c2c9ba3cc7fe7bf30b0a1cd78c2/.golangci.yml#L33

This PR removes `go vet` command from `validate-go` in Makefile.

### Test plan for issue:

This is a part of CI.
I tested it on my local machine.

no tech debt cleanup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,7 +188,6 @@ validate-go:
 	@[ -z "$$(ls pkg/util/*.go 2>/dev/null)" ] || (echo error: go files are not allowed in pkg/util, use a subpackage; exit 1)
 	@[ -z "$$(find -name "*:*")" ] || (echo error: filenames with colons are not allowed on Windows, please rename; exit 1)
 	@sha256sum --quiet -c .sha256sum || (echo error: client library is stale, please run make client; exit 1)
-	go vet -tags containers_image_openpgp ./...
 	go test -tags e2e -run ^$$ ./test/e2e/...
 
 validate-go-action:


facchettos (approved):

cadenmarchese (approved):
makes sense, thanks!

SudoBrendan (commented):
Hmm, as others have said, this does make sense to de-duplicate our CI, but do we really want to push `go vet` errors all the way to CI? Isn't it better for a dev to see those errors on their local machine before pushing?

bitoku:
@SudoBrendan 

> this does make sense to de-duplicate our CI

This PR doesn't de-duplicate our CI.
It just removes inconsistency between local and CI.

`go vet` isn't run on CI. `go vet` is now a part of `golangci-lint` and CI runs `golanci-lint`.
`golanci-lint` can be run locally, so this PR doesn't prevent a dev from running the same lint as CI.

bitoku:
This is what CI does for the linting currently.
https://github.com/Azure/ARO-RP/blob/master/.github/workflows/golint.yml

SudoBrendan (approved):

PR #2899, opened 2023-05-08
Add status conditions to operator MachineSet controller

tsatam:

Fixes [ARO-2997](https://issues.redhat.com/browse/ARO-2997)

Adds status conditions for the ARO Operator MachineSet controller. These will then be propagated to the ARO Cluster Operator resource. 

### Test plan for issue:

Existing unit tests for the controller were modified to assert conditions are changed as expected. 

N/A

### DIFF
--- a/pkg/operator/controllers/base/aro_controller_test.go
+++ b/pkg/operator/controllers/base/aro_controller_test.go
@@ -5,6 +5,7 @@ package base
 
 import (
 	"context"
+	"errors"
 	"testing"
 	"time"
 
@@ -18,7 +19,7 @@ import (
 	utilconditions "github.com/Azure/ARO-RP/test/util/conditions"
 )
 
-func TestSetConditions(t *testing.T) {
+func TestConditions(t *testing.T) {
 	ctx := context.Background()
 
 	controllerName := "Fake"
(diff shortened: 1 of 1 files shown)


pkg/operator/controllers/machineset/machineset_controller.go
```diff
@@ -54,9 +56,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 
 	r.log.Debug("running")
+
+	cnds, err := conditions.GetControllerConditions(ctx, r.client, ControllerName)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
 	modifiedMachineset := &machinev1beta1.MachineSet{}
 	err = r.client.Get(ctx, types.NamespacedName{Name: request.Name, Namespace: machineSetsNamespace}, modifiedMachineset)
```
cadenmarchese, line 66: There are a few additional CRUD operations in this controller as well as this one (list the machinesets, update the machinesets' replicas). Should we also be emitting the same statuses / messages if we get errors there?

tsatam: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/client/fake There does not currently appear to be a way to simulate any errors in the fake client, so adding test coverage for operations like List or Update that would not ever fail due to the presence/absence of resource is not possible.

pkg/operator/controllers/machineset/machineset_controller.go
```diff
@@ -88,11 +102,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 		if err != nil {
+			r.log.Error(err)
+			conditions.SetControllerDegraded(ctx, r.client, cnds, err)
+
 			return reconcile.Result{}, err
 		}
 	}
 
-	return reconcile.Result{}, nil
+	return reconcile.Result{}, conditions.ClearControllerConditions(ctx, r.client, cnds)
```
cadenmarchese, line 112: This confused me a bit, since at first glance it looks like the reconciler is now going to return some blank conditions as an error when there should be no error, but looking deeper I see that we're only returning the result from conditions.ClearControllerConditions as an error, which should be nil.

Do you have thoughts on setting/running the conditions before this final return? Not really necessary, but might improve readability.

cadenmarchese (commented):
Very straightforward, thanks!

pkg/operator/controllers/machineset/machineset_controller.go
tsatam: I think either way is fine, as long as we remain consistent. I'm happy to change this PR either way to establish that standard for future work. 

Perhaps "returning" the result of setting conditions might make more sense if that function was also responsible for crafting the return output? (e.g. make setting conditions a responsibility from exiting the function) I'm not immediately in favor of doing that, as it would conflate two separate responsibilities together, but if it's something we are consistent with it might make sense across the whole operator.

cadenmarchese (commented):
I don't think that we should abort the reconcile if we have an issue setting the conditions, but a few followup comments regarding the error handling.

pkg/operator/controllers/machineset/machineset_controller.go
```diff
@@ -88,11 +102,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		modifiedMachineset.Spec.Replicas = to.Int32Ptr(int32(minSupportedReplicas-replicaCount) + *modifiedMachineset.Spec.Replicas)
 		err := r.client.Update(ctx, modifiedMachineset)
 		if err != nil {
+			r.log.Error(err)
+			conditions.SetControllerDegraded(ctx, r.client, cnds, err)
```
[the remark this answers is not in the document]
tsatam: My intent here was never to shadow "real" errors encountered in the Reconcile loop. The example on line 112 is one where no execution errors were encountered in the reconcile, so I returned the condition handling error in case it exists. 

This is inconsistent behavior, and I will be updating this implementation to never return the errors from status updates. We should never be in a situation where getting the cluster resource fails (the Reconcile function here would have exited early on line 49 in that case), and would not ever want to change our reconcile functionality if updating controller conditions does not proceed as expected.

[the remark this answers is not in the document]
tsatam: I will be logging all errors encountered when handling/updating conditions, even if they are thrown away after. Update to the PR incoming.

jhoreman (dismissed):
As discussed in the PR review meeting, this looks good. Thanks for this!

pkg/operator/controllers/machineset/machineset_controller.go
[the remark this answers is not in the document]
mbarnes: To Kipp's point, I think it's worth logging errors just so they're recorded somewhere, but maybe consider rate limiting them so they don't flood the log?  i.e. if last error occurred less than X seconds ago, skip logging it.

mbarnes (commented):
Not sure where you're perceiving clunkiness (as you stated on Slack), but one suggestion is maybe simplify the base type name to just `AROController`.  There may be other controller commonalities we want to capture here unrelated to conditions, and I wouldn't want to preclude that just because of the type name.  (Nor do I think we want to venture into multiple "inheritance" in Go.)

I'll leave that for you to decide, though.  Not gonna block the PR over it.

tsatam:
Changes have been made and pushed up.

pkg/operator/controllers/base/aro_controller.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (c *AROController) ClearDegraded(ctx context.Context) {
+	c.SetConditions(ctx, c.defaultDegraded())
+}
+
+func (c *AROController) ClearConditions(ctx context.Context) {
+	c.SetConditions(ctx, c.defaultAvailable(), c.defaultProgressing(), c.defaultDegraded())
+}
+
+func (c *AROController) getCluster(ctx context.Context) (*arov1alpha1.Cluster, error) {
```
mbarnes, line 75: Maybe make this method public (`GetCluster`) since it looks like quite a lot of our controllers fetch this object.  This a good example of controller commonalities I mentioned earlier.

mbarnes (approved):
Made one more suggestion below, but this is looking pretty good to me now.

pkg/operator/controllers/base/aro_controller.go
tsatam: Made the change, and adapted the machineset and node controllers to use this method instead of doing the retrieval themselves. There's probably more opportunities for further common functionality refactors along this line (these two controllers both contain the same logic to bail early if their respective feature flags in the cluster resource are not enabled), but I'll leave that for future consideration.

mbarnes (approved):

jhoreman (approved):

kimorris27 (dismissed):

pkg/operator/controllers/base/aro_controller.go
```diff
@@ -0,0 +1,105 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+)
+
+type AROController struct {
+	Log    *logrus.Entry
+	Client client.Client
+	Name   string
```
SudoBrendan, line 22: Do these need to be public, or can we get by with defining a public constructor? [It looks like it should be](https://go.dev/play/p/PMd7FY-82ju?v=goprev), and that would reduce the diff on this PR (and all subsequent PRs) since this has downstream effects, namely on all access to `log` and `client` objects across every controller we manipulate with these changes.

I originally marked this as a nit, but after checking the first 5 controllers (all of them use `log` and `client`) I think this will save us a sufficient amount of work down the line to warrant changing here once now.

SudoBrendan (changes requested):
Overall, this is a really great PR, thanks! The one change I would ask is modifying `AROController` to have private `log` and `client` fields, since that'll require a refactor across nearly our whole codebase for controllers. There's a few other nits you can action or dismiss.

pkg/operator/controllers/base/aro_controller.go
tsatam: My previous ("clunky") version of this, that I unfortunately force-pushed over, did use a public constructor, but this basically involved doubling up the fields on the base controller and component controllers. While this could be abstracted away behind the component controllers' public constructors, it was hard to follow.

pkg/operator/controllers/base/aro_controller_test.go
```diff
@@ -91,39 +80,84 @@ func TestSetControllerConditions(t *testing.T) {
...
+					Type:               "FakeController" + operatorv1.OperatorStatusTypeAvailable,
+					Status:             operatorv1.ConditionTrue,
+					LastTransitionTime: past,
+				},
+			},
+			input: []*operatorv1.OperatorCondition{
+				{
+					Type:   "FakeController" + operatorv1.OperatorStatusTypeAvailable,
+					Status: operatorv1.ConditionTrue,
+				},
```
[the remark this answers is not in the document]
tsatam: I'll do this, thanks!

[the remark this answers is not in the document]
tsatam: I've made this change, and immediately spent the regained readability budget on adding additional test coverage for the wrapper functions around SetConditions. That work is on a separate commit, I'd appreciate a close review of that change so we can decide to keep or toss it.

pkg/operator/controllers/base/aro_controller_test.go
```diff
@@ -43,29 +44,92 @@ func TestSetConditions(t *testing.T) {
...
+		{
+			name:  "ClearProgressing - sets Progressing to false and clears message",
+			start: []operatorv1.OperatorCondition{defaultAvailable, isProgressing, defaultDegraded},
+			action: func(c AROController) {
+				c.ClearProgressing(ctx)
+			},
+			want: []operatorv1.OperatorCondition{defaultAvailable, defaultProgressing, defaultDegraded},
+		},
+		{
+			name:  "SetProgressing - sets Progressing to true with message",
```
kimorris27, line 102: It looks like this is a duplicate of the `SetProgressing` test case two list entries up. Same thing for the `ClearProgressing` test case directly below this one.

tsatam: good catch, thanks

kimorris27 (approved):

PR #2900, opened 2023-05-09
feat: Add retrieval functions to dynamic helper

jaitaiwan:
This PR is to add functionality to the dynamic helper which are needed in order to facilitate the removal of the hive clientset from the repository. Specifically the ability to retrieve resources has been added and existing functions now use the underlying GVRResolver.

This PR is to facilitate the changes required by the hive clientset removal pr (#2806). The changes to the dynamic helper in that PR have been split out into this one, and 2806 will be rebased off of this PR.

### Test plan for issue:
The existing unit tests have been updated and unit tests for newly added functions have been added.

Yes, inline documentation was added as a part of this PR to help facilitate quick understanding of what it does.


pkg/frontend/adminactions/kubeactions.go
```diff
@@ -43,7 +43,7 @@ type kubeActions struct {
 	log *logrus.Entry
 	oc  *api.OpenShiftCluster
 
-	gvrResolver dynamichelper.GVRResolver
+	gvrResolver dynamichelper.GVRInterface
```
jaitaiwan, line 46: Changed so that we use the interface rather than the implementation

pkg/util/dynamichelper/discovery/cache_fallback_discovery_client_mock.go
jaitaiwan, line 1: Extracted from the discovery_client_test.go file so that we can use the mock in other places.

pkg/util/dynamichelper/discovery/cache_fallback_discovery_client_test.go
jaitaiwan, line 1: Changes in this file are simply because we've extracted the mock and the mock has become an exported part of the package.

pkg/util/dynamichelper/dynamichelper.go
```diff
@@ -30,26 +36,35 @@ import (
...
+// Interface defines a set of methods which allow retrieving and updating
+// GVKs from the kubernets api.
 type Interface interface {
-	Refresh() error
-	EnsureDeleted(ctx context.Context, groupKind, namespace, name string) error
-	Ensure(ctx context.Context, objs ...kruntime.Object) error
+	Refresh() error                                                                              // Refresh the in-memory storage of the GVK
+	GetResource(ctx context.Context, groupKind, namespace, name string) (kruntime.Object, error) // Retrieve GVK resource
+	EnsureDeleted(ctx context.Context, groupKind, namespace, name string) error                  // Ensures the deletion of the GVK
+	Ensure(ctx context.Context, objs ...kruntime.Object) error                                   // Ensures the GVK exists and is correct
```
jaitaiwan, line 45: Added inline-docs as well as 3 new function definitions

pkg/util/dynamichelper/dynamichelper.go
```diff
@@ -30,26 +36,35 @@ import (
...
 }
 
-type dynamicHelper struct {
-	GVRResolver
+// DynamicHelper is a wrapper around a kubernetes admin api that provides utility functions
+// for retrieving any GVK with generic types.
+type DynamicHelper struct {
+	GVRInterface
```
jaitaiwan, line 51: Ensure we're using the GVRInterface rather than it's implementation

pkg/util/dynamichelper/dynamichelper.go
```diff
@@ -66,7 +81,25 @@ func New(log *logrus.Entry, restconfig *rest.Config) (Interface, error) {
 	return dh, nil
 }
 
-func (dh *dynamicHelper) EnsureDeleted(ctx context.Context, groupKind, namespace, name string) error {
+// GetResource uses the rest cluster connection to retrieves the GVR from
+// the requested namespace
+func (dh *DynamicHelper) GetResource(ctx context.Context, groupKind, namespace, name string) (kruntime.Object, error) {
+	gvr, err := dh.Resolve(groupKind, "")
```
jaitaiwan, line 87: The empty string is following existing implementations. I'm not a fan of passing an empty string without more context.

pkg/util/dynamichelper/dynamichelper.go
```diff
@@ -92,14 +126,24 @@ func (dh *dynamicHelper) Ensure(ctx context.Context, objs ...kruntime.Object) er
...
 
 	gvk := gvks[0]
+	// ASSUMPTION: The above assumes there will only ever be one gvk for the passed-in object
+	// but there may be a case where the object is not fully formed and therefore may match
+	// multiple gvks. This is a code-path that we don't test for so it may cause unintended
+	// side-effects. Therefore we log a warning for our future selves.
+	if len(gvks) > 1 {
+		_gvk := new.GetObjectKind().GroupVersionKind()
+		dh.log.Warnf("While retrieving kinds for an object \"%s\", more that one version returned which is a codepath that's not fully handled.", _gvk.String())
+	}
```
jaitaiwan, line 144: This was a previously undocumented assumption. I've added a warning and more information in a comment here because I believe in the future when other packages implement this function that this may be the cause of bugs. However it's not clear _what_ the correct functionality should be, which will probably be clearer when future functions make use of dynamic helper's Ensure.

pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -37,7 +45,7 @@ func (gvr mockGVRResolver) Resolve(groupKind, optionalVersion string) (*schema.G
 	return &schema.GroupVersionResource{Group: "metal3.io", Version: "v1alpha1", Resource: "configmap"}, nil
 }
 
-func TestEsureDeleted(t *testing.T) {
+func TestEnsureDeleted(t *testing.T) {
```
jaitaiwan, line 48: Small spelling fix

pkg/util/dynamichelper/generate.go
```diff
@@ -4,5 +4,5 @@ package dynamichelper
 // Licensed under the Apache License 2.0.
 
 //go:generate rm -rf ../mocks/$GOPACKAGE
-//go:generate go run ../../../vendor/github.com/golang/mock/mockgen -destination=../mocks/$GOPACKAGE/$GOPACKAGE.go github.com/Azure/ARO-RP/pkg/util/$GOPACKAGE Interface
+//go:generate go run ../../../vendor/github.com/golang/mock/mockgen -destination=../mocks/$GOPACKAGE/$GOPACKAGE.go github.com/Azure/ARO-RP/pkg/util/$GOPACKAGE Interface,GVRInterface
```
jaitaiwan, line 7: Small update so that the GVRResolver can also be mocked

pkg/util/dynamichelper/gvrresolver_test.go
```diff
@@ -195,17 +199,35 @@ func TestFindGVR(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
 			r := &gvrResolver{apiresources: tt.resources}
 
 			got, err := r.Resolve(tt.kind, "")
-			if !reflect.DeepEqual(err, tt.wantErr) {
-				t.Errorf("got: %#v, expected: %#v", err, tt.wantErr)
+			if err != nil {
+				for _, i := range deep.Equal(err.Error(), tt.wantErr.Error()) {
+					t.Error(i)
+				}
```
jaitaiwan, line 210: As mentioned in https://github.com/Azure/ARO-RP/pull/2806#discussion_r1184194712 `deep.Equal` on the error type is an anti-pattern. This will probably need to be changed before the PR gets approved, however it's not clear to me what the replacement should be. If you know, please comment here!

pkg/util/mocks/dynamichelper/dynamichelper.go
jaitaiwan, line 1: This file is auto-generated and does not require a manual review.

pkg/util/dynamichelper/gvrresolver_test.go
ArrisLee: not sure if I am understanding that in the right way, maybe we should just directly compare the error strings?

```go
if err != nil && testCase.wantErr != nil {
			if !strings.EqualFold(err.Error(), testCase.wantErr.Error()) {
				t.Errorf(...)
			}
			return
}
```

yjst2012:
Thanks for putting this up, there are also quite some changes for dynamichelper in [Guardrails PR](https://github.com/Azure/ARO-RP/pull/2755), looks like some adaptiive changes are inevitable... :joy:

pkg/util/dynamichelper/gvrresolver_test.go
jaitaiwan: The problem is that we shouldn't be doing comparisons on the error message itself from my understanding. We either should be using custom error types or checking only a portion of the error strings which is interesting in-and-of-itself. At least that's what I got from the aforementioned comment's link.

pkg/util/dynamichelper/discovery/cache_fallback_discovery_client_mock.go
hawkowl: Is this mock autogenerated somewhere? I am not sure I see a mirrored generate. We also usually keep mocks in `pkg/util/mocks/`.

jaitaiwan: No this wasn’t autogenerated, as mentioned the code existed in the other file so I extracted it here. Should one be moving it to the util mock package? I imagine there will have to be some refactoring to make that work.

pkg/util/dynamichelper/dynamichelper.go
```diff
@@ -30,26 +36,35 @@ import (
...
+	GetResource(ctx context.Context, groupKind, namespace, name string) (kruntime.Object, error) // Retrieve GVK resource
+	EnsureDeleted(ctx context.Context, groupKind, namespace, name string) error                  // Ensures the deletion of the GVK
+	Ensure(ctx context.Context, objs ...kruntime.Object) error                                   // Ensures the GVK exists and is correct
 }
 
-type dynamicHelper struct {
-	GVRResolver
+// DynamicHelper is a wrapper around a kubernetes admin api that provides utility functions
+// for retrieving any GVK with generic types.
+type DynamicHelper struct {
```
jaitaiwan, line 50: - [ ] Check to make sure that we should be exposing the dynamic helper struct itself outside of the package... this is what we have New functions for.

ArrisLee:
LGTM overall, we may need some more inputs around that error check logic

pkg/util/dynamichelper/dynamichelper_test.go
```diff
@@ -645,3 +653,89 @@ func TestMakeURLSegments(t *testing.T) {
...
+			case "GET":
+
+				switch req.URL.Path {
+				case "/api/v1/event":
+					return &http.Response{
+						StatusCode: http.StatusOK,
+						Header:     h,
+						Body:       io.NopCloser(buf),
+					}, nil
+				}
```
xiaoyu74, line 708: LGTM, may be just add some more test cases (e.g GET request to another URL returns error)
```
case "/api/v1/other-url":
	// New test case: GET request to another URL returns 404 status code
	return &http.Response{
		StatusCode: http.StatusNotFound,
		Header:     h,
		Body:       io.NopCloser(bytes.NewBufferString("Not found")),
	}, nil
}
```

jaitaiwan: Great catch! The reason I'm not testing the not found case is that the Ensure function is designed to create the object, if it doesn't already exist. So if the object can't be found it takes the "create this document" code path - this is an existing codepath that's in production and currently works. I agree that it should be tested, however... At this point in the PR to make this test one would have to mock the restcli or replace the restcli call (the preferred method) which would involve adding the ability to create objects to the GVR resolver. This functionality is actually added by Jeff's PR as I remember, so my intention here is not to duplicate work, or to change existing code that is already in production.

pkg/util/dynamichelper/gvrresolver_test.go
SudoBrendan: @ArrisLee is identifying what we've traditionally done in this codebase, string comparison. Per the original linting PR I linked, that's only somewhat better than deep equal, a final, more mature error state would be error enriching, which is way out of scope.

Granted, let's keep this out of the PR for now, I mostly meant it as an observation. We should create a new work item to track that fix.

pkg/util/dynamichelper/gvrresolver_test.go
```diff
@@ -195,17 +199,35 @@ func TestFindGVR(t *testing.T) {
...
+				return
 			}
+
 			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("got: %#v, expected: %#v", got, tt.want)
+				t.Errorf("%d Result: got: %#v, expected: %#v", index, got, tt.want)
```
SudoBrendan, line 215: Is adding the index helpful? If we fail here, isn't it already telling us "you're on test X, got Y expected Z"?

SudoBrendan (changes requested):
A few comments to add here, the main one is using NewSimpleClientSet, which I'm guessing will eliminate something like 100 lines of code from this PR and keep things, well, simpler.

I'm +1 with Spencer, let's leave the error deep equals for another day.

pkg/util/dynamichelper/dynamichelper.go
jaitaiwan: Will it get picked up when importing the interface?

pkg/util/dynamichelper/dynamichelper.go
jaitaiwan: I don’t want to error out, because if existing code is hitting this case but still currently working, breaking that code in prod seems like a really dangerous idea.

pkg/util/dynamichelper/gvrresolver_test.go
jaitaiwan: During testing if a nil pointer exception occurs it doesn’t show which test actually fails which is why I added the indexing.

facchettos (changes requested):
could we split this into multiple PRs? There are changes that don't really belong in the feature PR and that makes it more complicated to review

jaitaiwan:
@facchettos Can you provide some quantification on your statement? What changes don't belong? What about this PR is too complicated for you to review?

I've provided comments for most of the changes, there are less than 10 files changed one of which is auto-generated. Furthermore every change is documented with a helpful git commit message:
image

Apart from that this PR is about to get a fair sight more complicated as we will be uploading a new version of this PR with the guardrails dynamic helper changes as well as the changes required for this PR under 2806.

facchettos:
changing `dynamichelper.GVRResolver` to `dynamichelper.Interface` isn't really related to the feature right? same for exporting dynamichelper. It is something that makes sense but now it is tied to your feature work and makes the PR review process more cumbersome

pkg/util/dynamichelper/dynamichelper_test.go
facchettos, line 708: why nested switch statement instead of a simpler 
```go
if req.Method == http.MethodGET && req.URL.Path == "/api/v1/event" {
    return ... 
}
``` 
?
btw please prefer the constant from the http file instead of write "GET" yourself

jaitaiwan:
> dynamichelper.GVRResolver to dynamichelper.Interface

It's GVRResolver to GVRInterface and yes it's related. As explained in the inline comments on the PR we had to make the change so that we could generate a mock for the GVR resolver.

pkg/util/dynamichelper/gvrresolver_test.go
facchettos: you should probably use the function in `test/util/error/error.go` for error checking. In any case here I am not sure what `deep.Equal` value added is, since `Error()` returns a string.

pkg/util/dynamichelper/gvrresolver.go
```diff
@@ -13,7 +13,9 @@ import (
 	utildiscovery "github.com/Azure/ARO-RP/pkg/util/dynamichelper/discovery"
 )
 
-type GVRResolver interface {
+// GVRInterface defines the interface for refreshing and resolving Groups and
+// Versions and Resources from the kubenetes api
+type GVRInterface interface {
```
facchettos, line 18: why was this renamed?

pkg/frontend/adminactions/kubeactions.go
facchettos: I am not sure I see what that does, GVRResolver was already an interface, so we were using the interface rather than the implementation

bennerv (commented):
is there a reason we can't use controller-runtime's client that isn't typed?  `"sigs.k8s.io/controller-runtime/pkg/client"` ?  It's should fill the use case for removing hive clientset and has CRUD operations against resources.

jaitaiwan:
Closing this PR out as controller-runtime changes are a significant departure from the current code changes, so there's no use reviewing this branch herein.

PR #2902, opened 2023-05-09
Fix PUCM failure when outboundType is UserDefinedRouting

cadenmarchese:

Fixes an issue where PUCM will fail when outboundType is UserDefinedRouting, because it attempts to `Get` an external load balancer, which doesn't appear to be created in a prod install when UDR is enabled.

PUCM should pass for clusters where [our preview feature to not create a public IP](https://learn.microsoft.com/en-us/azure/openshift/howto-create-private-cluster-4x#create-a-private-cluster-without-a-public-ip-address-preview) is enabled. This PR skips the external load balancer portion of `checkandUpdateNIC()` if the UserDefinedRouting feature is enabled.

### Test plan for issue:

- Create a dev cluster as private, using the [public document](https://learn.microsoft.com/en-us/azure/openshift/howto-create-private-cluster-4x)
- Remove the public load balancer of the dev cluster
- Mimic the prod setup by changing outboundType from `Loadbalancer` to `UserDefinedRouting` in the cluster document
- [Run an admin update](https://github.com/Azure/ARO-RP/blob/master/docs/deploy-development-rp.md#make-admin-action-api-calls-to-a-running-local-rp)

### DIFF
--- a/pkg/cluster/fixssh.go
+++ b/pkg/cluster/fixssh.go
@@ -106,6 +106,10 @@ func (m *manager) checkandUpdateNIC(ctx context.Context, resourceGroup string, i
 			return err
 		}
 
+		if m.doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType == api.OutboundTypeUserDefinedRouting {
+			return nil
+		}
+
 		elbName := infraID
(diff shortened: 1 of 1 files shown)


bennerv (approved):

tony-schndr (approved):
LGTM!

s-amann (approved):

PR #2903, opened 2023-05-09
Update operator version when version.GitCommit changes (2nd approach)

ventifus:
Slightly tweaked version of https://github.com/Azure/ARO-RP/pull/2889. Leverage openshift/library-go `operatorv1helpers.SetOperandVersion` to set the version instead of explicitly modifying `clusterOperatorObj.Status.Versions`.

Modeled on openshift/library-go's example https://github.com/openshift/library-go/blob/d7e4a8ca058837a972a4cdef06334be4f7becdde/pkg/operator/status/status_controller.go#L200


tsatam (approved):
LGTM, thanks for doing this!

I think this scenario reveals an opportunity for improved test coverage on this controller, but this can be a future activity.

cadenmarchese (approved):
Makes sense to leverage the OCP library-go package. Thanks!

petrkotas (approved):
Good work!

PR #2904, opened 2023-05-10
Add an env variable as a rollout flag for CheckAccess

nwnt:
rp-production.json was regenerated with the changes from
- resources_rp.go
- templates_rp.go
- rpVMSS.sh

Fixes ARO-2922

CheckAccess migration will only happen gradually starting with INT and Canary regions.  To orchestrate such gradual release, a new environment variable is introduced to signify if the region will use CheckAccess or not.

### Test plan for issue:
Tested this in INT.


pkg/deploy/generator/templates_rp.go
```diff
@@ -85,6 +85,8 @@ func (g *generator) rpTemplate() *arm.Template {
 			"clustersInstallViaHive",
 			"clusterDefaultInstallerPullspec",
 			"clustersAdoptByHive",
+
```
bennerv, line 88: nit: no extra space.

bennerv (changes requested):
first pass looks great- let's use LiveConfig though so we can leverage it when it's implemented

pkg/deploy/generator/templates_rp.go
bennerv, line 88: nit: won't need this space after we use liveconfig

SudoBrendan (approved):
Overall LGTM! A few nit-picks you can take or dismiss :)

bennerv (approved):
lgtm - if you want to make the changes suggested by Brendan, let me know after and I'll approve + merge again.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -430,18 +430,20 @@ func (c closure) usingListPermissions() (bool, error) {
...
 		scope := c.dv.azEnv.ResourceManagerEndpoint + "/.default"
 		t, err := c.dv.checkAccessSubjectInfoCred.GetToken(c.ctx, policy.TokenRequestOptions{Scopes: []string{scope}})
 		if err != nil {
+			c.dv.log.Error("Unable to get the token from AAD", err)
 			return false, err
 		}
 
 		oid, err := token.GetObjectId(t.Token)
 		if err != nil {
+			c.dv.log.Error("Unable to parse the token oid claim")
```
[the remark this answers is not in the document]
nwnt: Done

pkg/api/validate/dynamic/dynamic_test.go
```diff
@@ -999,8 +999,27 @@ func TestValidateVnetPermissionsWithCheckAccess(t *testing.T) {
...
+							{
+								ActionId:       "Microsoft.Network/virtualNetworks/subnets/join/action",
+								AccessDecision: remotepdp.Allowed,
+							},
+							{
+								ActionId:       "Microsoft.Network/virtualNetworks/subnets/read",
+								AccessDecision: remotepdp.Allowed,
+							},
+							{
+								ActionId:       "Microsoft.Network/virtualNetworks/subnets/write",
```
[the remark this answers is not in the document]
nwnt: Done!  I extracted the variables into corresponding *var* blocks.

bennerv (dismissed):
lgtm

bennerv (approved):

pkg/util/liveconfig/checkaccess.go
```diff
@@ -0,0 +1,18 @@
+package liveconfig
```
petrkotas, line 1: Please document what checkAccess is. There is no explanation of what this option does and what are the implications of using it.

petrkotas (dismissed):
@nwnt Thank you for the PR, it is a clean work. Nicely done. I have one simple ask. Please add a explanation of what CheckAccessV2 is, what does it do and what are the implications of using it. If we do not write it down now it will end up as a tribal knowledge that will be known only to few.
Which will eventually create bottle neck for fixing issues in the code. Thank you! 

PS: I looked at attached Jira tickets, there is no explanation there either.

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -458,9 +461,19 @@ func (c closure) usingCheckAccessV2() (bool, error) {
...
-			c.dv.log.Infof("%s has no access to %s", *c.oid, result.ActionId)
+	for _, action := range c.actions {
+		found := false
+		for _, result := range results.Value {
+			if result.ActionId == action {
+				found = true
+				if result.AccessDecision == remotepdp.Allowed {
+					break
+				}
+				return false, nil
```
s-amann, line 472: is there value in keeping this log here: `c.dv.log.Infof("%s has no access to %s", *c.oid, result.ActionId)`

s-amann (commented):
one question, otherwise lgtm.

One other concern, but I'm not sure if this would happen with 'real' data. When looping through the results, if we don't find a matching action within the results, we will stop looping over actions as `found` will never be set to true.

pkg/util/liveconfig/checkaccess.go
nwnt: Hi @petrkotas, definitely.  I have some info in the ADO wiki [here](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/409184/CheckAccess-API-migration).  Let me know if this is ok or if there are any points missing.

I should also be in a habit of including more info on the jira child tickets.  There are some background info on the parent one in ARO-1800.  Let me know if you have any suggestion on how I can improve it.

pkg/api/validate/dynamic/dynamic.go
nwnt: I'm springing a lot of logging statements here because the API hasn't been GAed yet and there are chances that it might give wrong results.  I plan to remove most of the loggings once we have completely switched over to the new API.  For now, I think the more data to observe how the API behaves, the better.

For the looping question, if we don't find a matching action, then it means the particular action/permission hasn't been granted to the service principal.  In that case, it should stop and go into another retry call.

bennerv:
@petrkotas  if the doc links Nont provided are good enough (note they're in Jira as well), please feel free to give this a thumbs up and merge.  I'm good with it as is.

jhoreman (approved):
Thanks Nont, lgtm!

s-amann (approved):

bennerv:
@petrkotas If you look at the epic it has links to MSFT documentation.  A tl;dr: We currently use an older API to check permissions against identities.  There is a new API that MSFT is currently releasing called CheckAccess, which will have better "freshness" (cache) times and it is handled in a different manner.  It's currently not fully GA in all of public cloud, but undergoing rollout and one of our canary regions is swapped over to the new API to confirm it is working as expected.  

Going to merge.

PR #2914, opened 2023-05-18
Service endpoint removal

kimorris27:

https://issues.redhat.com/browse/ARO-3020 and https://issues.redhat.com/browse/ARO-3021

Implements service endpoint removal according to the following logic:

- If egress lockdown has been enabled or a new cluster is being created: never add or reconcile service endpoints
- If egress lockdown is not enabled: continue to reconcile service endpoints (this is a responsibility of the ARO operator's subnets controller)

### Test plan for issue:

- Unit tests to validate logic
- Spike work to validate that clusters can function without service endpoints
  - https://issues.redhat.com/browse/ARO-2824
  - https://issues.redhat.com/browse/ARO-2953
- Deployed these changes to INT and validated that cluster creation will succeed without service endpoints and that the cluster functioned normally

It looks like the public facing docs at https://learn.microsoft.com/en-us/azure/openshift/ reference service endpoints in multiple places and will need to be updated. Does our team manage these docs?


pkg/cluster/enableendpoints.go
```diff
@@ -8,9 +8,9 @@ import (
 	"fmt"
 )
 
-// enableServiceEndpoints should enable service endpoints on
+// ensureServiceEndpoints should enable service endpoints on
```
cadenmarchese, line 11: You could get away with renaming the file to match the new convention as well.

pkg/util/subnet/subnet.go
```diff
@@ -142,12 +142,17 @@ func (m *manager) GetAll(ctx context.Context, subnetIds []string) ([]*mgmtnetwor
...
+func (m *manager) CreateOrUpdateFromIds(ctx context.Context, subnetIds []string, gatewayEnabled bool) error {
 	subnets, err := m.GetAll(ctx, subnetIds)
 	if err != nil {
 		return err
 	}
 
+	// If egress lockdown is enabled, do not add service endpoints to the subnets.
+	if gatewayEnabled {
+		return m.createOrUpdateSubnets(ctx, subnets)
+	}
```
cadenmarchese, line 154: It's just a nit, but maybe we can reverse this condition. It reads more like "we are defaulting to no service endpoints" that way.

	if !gatewayEnabled { // if the cluster is old :)
		// apply service endpoints if not already there
	}

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
```diff
@@ -12,10 +12,19 @@ import (
...
 
 	"github.com/Azure/ARO-RP/pkg/api"
+	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 	"github.com/Azure/ARO-RP/pkg/util/subnet"
 )
 
 func (r *reconcileManager) ensureSubnetServiceEndpoints(ctx context.Context, s subnet.Subnet) error {
+	// Do not add/reconcile service endpoints if egress lockdown is enabled.
+	if gatewayEnabled(r.instance) {
+		r.log.Info("Skipping service endpoint reconciliation since egress lockdown is enabled")
```
cadenmarchese: Same feedback here - since (most?) clusters are egress lockdown enabled, we might want to reverse this and log on that so we don't add too many new operator logs.

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
```diff
@@ -58,3 +67,7 @@ func (r *reconcileManager) ensureSubnetServiceEndpoints(ctx context.Context, s s
 	}
 	return nil
 }
+
+func gatewayEnabled(cluster *arov1alpha1.Cluster) bool {
+	return len(cluster.Spec.GatewayDomains) > 0
```
cadenmarchese, line 72: Should we use `featureProfile` in the cluster CR to check this instead?
Or even both?

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
```diff
@@ -19,11 +19,11 @@ import (
 func (r *reconcileManager) ensureSubnetServiceEndpoints(ctx context.Context, s subnet.Subnet) error {
 	// Do not add/reconcile service endpoints if egress lockdown is enabled.
 	if gatewayEnabled(r.instance) {
-		r.log.Info("Skipping service endpoint reconciliation since egress lockdown is enabled")
+		r.log.Debug("Skipping service endpoint reconciliation since egress lockdown is enabled")
```
cadenmarchese, line 22: Why use `debug`?

cadenmarchese (commented):
Nice work! A few comments but mostly nits.

petrkotas (commented):
Good work ont his @kimorris27. I have a question, which is actually not covered in workitems. What will happen to clusters that does not have egressLockdown enabled, but will switch to have it enabled? The code currently does not cover the path.

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
s-amann: IMO we should check the featureProfile first, and checking for the list of domains is extra/optional but should also be `>0`.

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
s-amann: I imagine this will be the most common case, since this feature is enabled by default.

s-amann (commented):
overall lgtm, I think updating `gatewayEnabled` to check for the featureProfile flag will align with the previous check we do on the openshiftDoc when deploying the storageAccount.

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
kimorris27: I'm not seeing `featureProfile` in the cluster CR, but to double check that I'm looking at the right thing, the cluster CR is the one you would get with `oc get cluster`, right?

kimorris27:
> Good work ont his @kimorris27. I have a question, which is actually not covered in workitems. What will happen to clusters that does not have egressLockdown enabled, but will switch to have it enabled? The code currently does not cover the path.

@petrkotas , that's a good question. Currently, nothing will happen to them, i.e. their service endpoints will not be deleted but will also not be reconciled. Maybe our runbook for enabling egress lockdown needs to be updated to include a step for removing the service endpoints? Though we wouldn't have permission to do it directly - we'd only be able to ask the customer to do it.

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
cadenmarchese: @kimorris27 you're right, no featureProfile in the cluster CR, that exists only in the cluster doc 🙁 
I'm cool with just checking the gateway domains if it's the most reliable.

cadenmarchese (dismissed):

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
s-amann: my bad, you're correct. What you have is good.

s-amann:
> > Good work ont his @kimorris27. I have a question, which is actually not covered in workitems. What will happen to clusters that does not have egressLockdown enabled, but will switch to have it enabled? The code currently does not cover the path.
> 
> @petrkotas , that's a good question. Currently, nothing will happen to them, i.e. their service endpoints will not be deleted but will also not be reconciled. Maybe our runbook for enabling egress lockdown needs to be updated to include a step for removing the service endpoints? Though we wouldn't have permission to do it directly - we'd only be able to ask the customer to do it.

we could simply add a step to run a full adminUpdate after enabling egress lockdown.

s-amann (dismissed):

kimorris27:
>

kimorris27:
Reopening - I closed the PR by mistake.

kimorris27:
> > > Good work ont his @kimorris27. I have a question, which is actually not covered in workitems. What will happen to clusters that does not have egressLockdown enabled, but will switch to have it enabled? The code currently does not cover the path.
> > 
> > 
> > @petrkotas , that's a good question. Currently, nothing will happen to them, i.e. their service endpoints will not be deleted but will also not be reconciled. Maybe our runbook for enabling egress lockdown needs to be updated to include a step for removing the service endpoints? Though we wouldn't have permission to do it directly - we'd only be able to ask the customer to do it.
> 
> we could simply add a step to run a full adminUpdate after enabling egress lockdown.

Two things to add after looking into this a bit more:

1. There's currently no logic included in the adminUpdate that would remove existing service endpoints - it either adds/reconciles them or doesn't. So if we wanted to remove them in adminUpdate, we would need to add that functionality in this PR.
2. Based on the PRD, removing service endpoints on existing clusters is out of scope. The purpose of this work is just to make sure we don't add them to new clusters.

With this in mind, I think that the logic in this PR is fine as-is. Same with the egress lockdown enablement steps.

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
```diff
@@ -12,49 +12,61 @@ import (
...
+	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 	"github.com/Azure/ARO-RP/pkg/util/subnet"
 )
 
 func (r *reconcileManager) ensureSubnetServiceEndpoints(ctx context.Context, s subnet.Subnet) error {
-	subnetObject, err := r.subnets.Get(ctx, s.ResourceID)
-	if err != nil {
-		return err
-	}
+	if !gatewayEnabled(r.instance) {
```
s-amann, line 20: can we invert this logic to return early instead of nesting the existing logic?

cadenmarchese:
Code changes are looking good, but /hold until we determine if we need to put this behavior behind a feature flag in the short term.

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
kimorris27: I originally wrote it in that "return early" way, but I changed it to the way it is now after Caden pointed out that `!gatewayEnabled` is the exception and not the norm, and so it makes sense to have that logic nested. Having considered both ways, I think I prefer the way it is now.

cadenmarchese (approved):
Based on a conversation with PM, an AFEC feature flag is not a hard requirement for releasing this feature 👍

s-amann (approved):

PR #2926, opened 2023-06-01
ARO-3054 - Set OpenShift Version during Cluster Creation

tphelps93:
Added functionality for clusters creation to use a specified version of OpenShift.

**Testing**
A portion of this code change relies on input from the user in the **RP Pipeline** in **ARO-Pipelines** and that input to be passed as an environment variable (**OS_CLUSTER_VERSION**). That portion has been tested locally and it does successfully take the input and pass it to the script.

The changes made in the **ARO-RP** simply set the _OS_CLUSTER_VERSION_ to a variable (**osClusterVersion**) that is passed in as an argument to the **Create** function call, the _Create_ function calls the **createCluster** function which takes _osClusterVersion_ as an argument, and that is utilized as the **Version** key in the **ClusterProfile** struct. 

We want the functionality to remain the same if no version is passed in by the user. 
This is accomplished by the **omitempty** in the _ClusterProfile_ struct. If the value of the _osClusterVersion_ is null, an empty string, etc then it should not include that key/value. I tested this locally and it does indeed, leave out the version when an empty string is passed.

Next steps would be to merge this PR and test this end-to-end.

[JIRA Task](https://issues.redhat.com/browse/ARO-3054)


tphelps93:
@microsoft-github-policy-service agree company="CloudFit Software"

UlrichSchlueter (approved):
LGTM

PR #2930, opened 2023-06-02
Set a default value for checkaccess

bennerv:

Fixes a deployment issue where the checkAccess RP config var isn't set resulting in the deployment failing.  

### Test plan for issue:

Deploy to FF INT

No


s-amann (approved):

weinong (commented):
thanks!

PR #2931, opened 2023-06-02
aro-3168: retry on conflict and not found for disabling samples

s-amann:

Fixes https://issues.redhat.com/browse/ARO-3168

This step can fail if the RP gains control over the creating cluster before this CR is applied

### Test plan for issue:

Unit tests

no


pkg/cluster/samples.go
```diff
@@ -19,17 +20,20 @@ func (m *manager) disableSamples(ctx context.Context) error {
...
 
-	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		c, err := m.samplescli.SamplesV1().Configs().Get(ctx, "cluster", metav1.GetOptions{})
-		if err != nil {
-			return err
-		}
+	return retry.OnError(retry.DefaultRetry,
+		func(err error) bool {
+			return errors.IsConflict(err) || errors.IsNotFound(err)
+		}, func() error {
```
SudoBrendan, line 26: nit: I was confused by this at first glance, can we fix the formatting to have one parameter per-line considering we're doing in-line anonymous functions (that is, huge, formatted parameters)?

```
return retry.OnError(
    a,
    b,
    c
)
```

SudoBrendan (approved):
Overall LGTM, one nit.

pkg/cluster/samples.go
s-amann: totally, I'll update with re-formatting.

pkg/cluster/samples.go
```diff
@@ -19,17 +20,22 @@ func (m *manager) disableSamples(ctx context.Context) error {
...
+			return errors.IsConflict(err) || errors.IsNotFound(err)
+		},
+		func() error {
+			c, err := m.samplescli.SamplesV1().Configs().Get(ctx, "cluster", metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
 
-		c.Spec.ManagementState = operatorv1.Removed
+			c.Spec.ManagementState = operatorv1.Removed
```
SrinivasAtmakuri, line 34: LGTM, a minor nit, GitHub is highlighting whitespace here, maybe this should be a tab space?

petrkotas (approved):
@s-amann Thanks! LGTM

PR #2932, opened 2023-06-05
[ARO-3194] Vendor dnsmasq into the operator

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-3194 

To remove the installer dependency, we need to vendor in the dnsmasq config code.

### Test plan for issue:

E2E

N/A

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/coreos/go-oidc v2.2.1+incompatible
 	github.com/coreos/go-semver v0.3.0
 	github.com/coreos/go-systemd/v22 v22.3.2
+	github.com/coreos/ignition v0.35.0
 	github.com/coreos/ignition/v2 v2.14.0
 	github.com/coreos/stream-metadata-go v0.2.0
 	github.com/form3tech-oss/jwt-go v3.2.5+incompatible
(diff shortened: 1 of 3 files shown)


SudoBrendan (commented):
To help future reviewers who are a little unfamiliar with where this is coming from, it's basically copy-pasting this file from our vendored installer to remove the dependency: https://github.com/jewzaam/installer-aro/blob/release-4.11-azure/pkg/aro/dnsmasq/dnsmasq.go

There's a few differences with this file here, including the removal of ignition3, which I'm assuming was done because we don't use it here (right?). My biggest concern is how we will keep these files in sync and/or if we need to? Would it be better for us to abstract this away into a 3rd package/repo that is imported by both the installer and the RP so that we prevent drift?

tsatam:
Reviewed, I agree with Brendan's comments. 

I do not believe there to be any practical differences as it stands right now, seeing as the ignition3Config was never used in the result MC, so we should hopefully not have any immediate issues e.g. the installer putting down one version of the MC that then gets replaced by a different one from the operator. This would be of concern if/when more changes are introduced, however.

hawkowl:
Yeah, we don't use ignition3 here, so I removed it.

The work regarding upstream implementation of custom DNS stuff, as well as the ability to "inject" configs and the like into the Installer, will mean that we have to cross the bridge around those differences eventually. Right now, it doesn't particularly matter if the Operator has a different version, as long as *both* versions work.

hawkowl:
@petrkotas this is just a code copy, doesn't change any of the topology or anything :)

PR #2933, opened 2023-06-05
Always log error (if applicable) on failed deployment attempts

kimorris27:

Came up in Slack thread about most recent RP release. When I updated this error handling code a few months ago, I forgot to ensure that the error on failed deployments is always logged regardless of what type of error it is.

Prior to this PR, errors on failed deployments would only be logged if it was the last deployment attempt or the error was known (specifically the known error related to the RP LB). This PR makes it so that the error is always logged for any failed deployment.

### Test plan for issue:

N/A

N/A


SudoBrendan (approved):
LGTM

schiruma:
LGTM

PR #2937, opened 2023-06-06
Remove AZURE_ARM_CLIENT_SECRET helper 

bennerv:

Our local development workflow lays down a role assignment on creation of the cluster resource group to mimic ARM's magical [S2S Resource Group Scoped Authorization](https://armwiki.azurewebsites.net/rbac/authorization/ResourceGroupScopedServicetoServiceAuthorization.html?q=s2s)  

We no longer use client secret in this, so continue using certificate based credentials to lay down this role assignment for development mode.  

Remove stale/old code.  

### Test plan for issue:

E2e will cover everything

nope

### DIFF
--- a/pkg/env/armhelper.go
+++ b/pkg/env/armhelper.go
@@ -75,30 +75,15 @@ func newARMHelper(ctx context.Context, log *logrus.Entry, env Interface) (ARMHel
 	var tokenCredential azcore.TokenCredential
 	var err error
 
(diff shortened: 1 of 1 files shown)


bennerv:
Hold as we don't have access to update the SPs in our CI to remove this.

SudoBrendan (approved):
LGTM as long as we can get CI working eventually.

bennerv:
It got past the failure point from before.  This should work :sunglasses:

PR #2939, opened 2023-06-08
Propagate errors of ARO ImageConfig controller to ARO Operator

SrinivasAtmakuri:
JIRA: https://issues.redhat.com/browse/ARO-3272

Fixes

Based on changes of: https://github.com/Azure/ARO-RP/pull/2899

### DIFF
--- a/pkg/operator/controllers/imageconfig/image_controller.go
+++ b/pkg/operator/controllers/imageconfig/image_controller.go
@@ -74,12 +74,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	imageconfig := &configv1.Image{}
 	err = r.Client.Get(ctx, types.NamespacedName{Name: request.Name}, imageconfig)
 	if err != nil {
+		r.Log.Error(err)
+		r.SetDegraded(ctx, err)
+
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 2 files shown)


SudoBrendan (changes requested):
Hey, thanks for the PR! There's great progress here, but this work isn't complete yet. Moving over to the base controller will now enable you to use the helper functions of the base to merge the errors of this controller into the ARO Operator status. This should happen for anything that would result in error states for this controller.

Here's an example of a call: https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/machineset/machineset_controller.go#L62

tsatam:
Additionally, as Brendan has mentioned, since the goal of this change is to propagate error cases as conditions on the operator, I'd like to see that reflected in the test cases. You can get an example of how to do that in the machineset controller tests as well: https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/machineset/machineset_controller_test.go

SudoBrendan (changes requested):
My re-review was requested, but I don't see the changes suggested implemented. We should see:

1. Status of the controller going to the operator, via the helpers
2. Test cases that showcase how those errors manifest as conditions on the operator.

s-amann (approved):

tsatam (approved):

cadenmarchese (approved):

PR #2942, opened 2023-06-08
Revert "Fix failure in az aro update when missing azure-credentials object"

cadenmarchese:
Reverts Azure/ARO-RP#2887 due to an issue with updating an existing object

https://redhat-internal.slack.com/archives/C02ULBRS68M/p1686225310006819


s-amann (approved):

dem4gus:
Conflicts with #2945 that only reverts using `Apply` and uses `Create`/`Update` instead, but otherwise keeps the rest of the functionality intact.

PR #157, opened 2020-02-06
RP shutdown flow

mjudeikis:
Backends will wait for all workers to complete before sending exiting. If not finished in 60 min, sigkill will be sent from systemd 

Needs to merge first https://github.com/Azure/ARO-RP/pull/154

PR #2946, opened 2023-06-08
update predeploy to restart old VMSS when service secrets rotated

rajdeepc2792:
Jira - https://issues.redhat.com/browse/ARO-3240

- Here's the error scenario described:- 
    **Deployment**
    
    - updates encryption secrets
    - Deploys new VMSS but fails due to RHUI
    
    **Cluster creation**
    
    - Old VMs without updated encryption secrets handle creation
    - Old VMs pass off to hive
    - Hive runs aro-installer image
    - aro-installer image picks up new encryption secrets
    - aro-installer image uses latest encryption secrets to update openshiftcluster doc
    - RP waits for aro-installer to finish
    - aro-installer finishes
    - RP assumes control
    - RP attempts to generate kubeconfigs
    - RP cannot decode existing cluster doc because it's on old keys

- Suggested Fix:-
    
    - Reduces the window of error due to encryption key rotation by restarting gateway and rp on old vmss after key rotation.

### Test plan for issue:
- Reproduced issue in dev using full-RP deployment:-
  - Deployed Full-RP
  - Triggered `make deploy` using changes:- rotated the encryption keys and skipped new VMSS deployment
  - Triggered cluster creation, failed with same `chacha20poly1305` error.
- Performed the above step again, with partial fix:-
  - Deployed Full-RP
  - Triggered `make deploy` using changes:- rotated the encryption keys, restarted aro-rp in the old rp VMSS and skipped new VMSS deployment.
  - Triggered cluster creation, failed with same `chacha20poly1305` error, this time due to aro-gateway not restarted.
- Performed the above step again, with complete fix:-
  - Deployed Full-RP
  - Triggered `make deploy` using changes:- rotated the encryption keys, restarted aro-gateway in the old gateway VMSS, restarted aro-rp in the old rp VMSS,  and skipped new VMSS deployment.
  - Triggered cluster creation
  - Ran the `make deploy` in between the cluster creation again, restarting the aro-rp and aro-gateway services.
    - Successfully restarted the services in no time.
  - Cluster creation completed successfully.

- Unit Tests added for predeploy.go functions.

No
There's a need to watch the lag in rp restart during release in prod because of https://github.com/Azure/ARO-RP/pull/157.
If the lag is frequent and impacts the release process time there's a need to come up with an alternative approach to this PR.


pkg/deploy/predeploy.go
```diff
@@ -376,54 +379,71 @@ func (d *deployer) configureServiceSecrets(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	if isRotated {
+		err = d.restartOldScalesets(ctx, "systemctl restart aro-gateway", d.config.GatewayResourceGroupName)
+		if err != nil {
+			return err
+		}
+		err = d.restartOldScalesets(ctx, "systemctl restart aro-rp", d.config.RPResourceGroupName)
```
SudoBrendan, line 400: IIRC aro-rp has some pretty good control sequences that'll prevent it from being killed if it's in the middle of a job, so this command could hang for as long as a cluster creation takes, right? Or has that gone away with hive? If that's true, this change then introduces a massive lag in our deployment process, where we could be adding something like 60 minutes of time for each rollout. Would it be better if we instead only restarted if we detected that the new VMSS was not coming up in an expected period of time, or would that still introduce the bug we're trying to solve?

SudoBrendan (commented):
Just a quick comment I'd like some discussion on.

pkg/deploy/predeploy.go
SudoBrendan: as a future state, would there be a way for us to instead just refresh our cosmosdb key continuously somehow with what we expect it to be as opposed to only refreshing on boot, or is that just way too hard?

rajdeepc2792: @SudoBrendan , related to aro-rp taking time for getting killed in between the job, I tested the below flow on the fix:-
- Started Cluster Creation
- Restarted aro-rp service before cluster creation completion(by redeploying the RP), took similar time as for no job running on service.
-  Cluster created successfully.

Does that answer that there won't be a lag in the deployment process?

SudoBrendan: I had a talk with Ben about this yesterday. We came to the conclusion that `backend` processes can fail to shutdown, and that can take [60 minutes to resolve at a systemd level by design](https://github.com/Azure/ARO-RP/pull/157). This isn't going to happen each time, but it is a possibility.

SudoBrendan: I guess for now we can just keep this in mind as something to watch out for when deploying. If we find that we're hitting this often, we can revert this change and attempt authoring continuous refreshing as part of the RP and Gateway.

rajdeepc2792: Thanks @SudoBrendan, noted. Makes sense, I will try figuring out the continuous refreshing before hand as an alternate to this PR in case the lag is frequent.

pkg/deploy/predeploy.go
```diff
@@ -439,25 +459,69 @@ func (d *deployer) createSecret(ctx context.Context, kv keyvault.Manager, secret
...
+			errors <- d.vmssvms.RunCommandAndWait(ctx, resourceGroupName, vmssName, id, mgmtcompute.RunCommandInput{
+				CommandID: to.StringPtr("RunShellScript"),
+				Script:    &[]string{script},
+			})
+		}(*vm.InstanceID)
+	}
+
+	d.log.Print("waiting for instances to restart")
+	for range scalesetVMs {
+		err := <-errors
```
cadenmarchese, line 521: This part of the PR will wait for the command to finish executing and returns any errors from `RumCommandAndWait`, but doesn't wait for the full restart of the instances, right?

If so, we might want to also add some logic to wait for the VMs to come back fully healthy before continuing to avoid an issue resulting from this https://github.com/Azure/ARO-RP/pull/2946/files#r1224785534

That way, we won't continue on to deploy the new VMs if the old VMs are stuck restarting.

pkg/deploy/predeploy.go
```diff
@@ -376,54 +379,71 @@ func (d *deployer) configureServiceSecrets(ctx context.Context) error {
...
 
-	return d.ensureSecretKey(ctx, d.portalKeyvault, env.PortalServerSSHKeySecretName)
+	isNew, err := d.ensureSecretKey(ctx, d.portalKeyvault, env.PortalServerSSHKeySecretName)
+	isRotated = isNew || isRotated
+	if err != nil {
+		return err
+	}
+
+	if isRotated {
+		err = d.restartOldScalesets(ctx, "systemctl restart aro-gateway", d.config.GatewayResourceGroupName)
```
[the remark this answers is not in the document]
rajdeepc2792: Moved the `script` conditions to restartOldScaleset() based on vmssName.

pkg/deploy/predeploy.go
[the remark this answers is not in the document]
rajdeepc2792: Moved the `script` conditions to restartOldScaleset() based on vmssName.

pkg/deploy/predeploy.go
```diff
@@ -439,25 +459,69 @@ func (d *deployer) createSecret(ctx context.Context, kv keyvault.Manager, secret
...
+		err = d.restartOldScaleset(ctx, *vmss.Name, script, resourceGroupName)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (d *deployer) restartOldScaleset(ctx context.Context, vmssName string, script string, resourceGroupName string) error {
```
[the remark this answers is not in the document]
rajdeepc2792: Moved the `script` conditions to restartOldScaleset() and it is decided based on vmssName, keeping `restartOldScaleset` and `restartOldScalesets` without `script` passing.

pkg/deploy/predeploy.go
rajdeepc2792: Added a VMSS health check post the restarts as suggested.

pkg/deploy/predeploy.go
```diff
@@ -439,25 +463,91 @@ func (d *deployer) createSecret(ctx context.Context, kv keyvault.Manager, secret
...
+	d.log.Print("waiting for instances to restart")
+	for range scalesetVMs {
+		err := <-errors
+		if err != nil {
+			return err
+		}
+	}
+
+	// wait for load balancer probe to change the health status
+	time.Sleep(30 * time.Second)
```
cadenmarchese, line 545: Do we need to sleep here, or is the `context.WithTimeout` on line 546 sufficient?

pkg/deploy/predeploy.go
```diff
@@ -439,25 +463,91 @@ func (d *deployer) createSecret(ctx context.Context, kv keyvault.Manager, secret
...
+	for _, vm := range scalesetVMs {
+		go func(id string) {
+			errors <- d.vmssvms.RunCommandAndWait(ctx, resourceGroupName, vmssName, id, mgmtcompute.RunCommandInput{
+				CommandID: to.StringPtr("RunShellScript"),
+				Script:    &[]string{restartScript},
+			})
+		}(*vm.InstanceID)
+	}
+
+	d.log.Print("waiting for instances to restart")
```
cadenmarchese, line 536: nit: this should be "waiting for script to execute" or similar if comment https://github.com/Azure/ARO-RP/pull/2946/files#r1228539236 is accurate.

pkg/deploy/predeploy.go
cadenmarchese: Looks good, thanks for using the existing functions for this 👍

pkg/deploy/predeploy.go
rajdeepc2792: As I understand, context.withTimeout is to exit from the long running reporting of unhealthy status, but here after the aro-rp restart there's a chance that rp-probe remains healthy as probe runs with interval 15 seconds and threshold 2. That is rp-probe will start reflecting correct relevant health status for vmss after 30 seconds of restart.
Please correct me if there's something wrong with the understanding.

pkg/deploy/predeploy.go
```diff
@@ -26,9 +27,13 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/keyvault"
 )
 
-// Rotate the secret on every deploy of the RP if the most recent
-// secret is greater than 7 days old
-const rotateSecretAfter = time.Hour * 168
+const (
+	// Rotate the secret on every deploy of the RP if the most recent
+	// secret is greater than 7 days old
+	rotateSecretAfter    = time.Hour * 168
```
dem4gus, line 33: nit: It would be easier to figure out the timespan using `time.Hour * 24 * 7`

dem4gus (approved):
Small suggestion, PR looks good.

pkg/deploy/predeploy.go
rajdeepc2792: Thanks, done.

kimorris27 (approved):

pkg/deploy/predeploy.go
```diff
@@ -439,25 +463,91 @@ func (d *deployer) createSecret(ctx context.Context, kv keyvault.Manager, secret
...
+	switch {
+	case strings.HasPrefix(vmssName, gatewayVMSSPrefix):
+		restartScript = gatewayRestartScript
+		waitForReadiness = d.gatewayWaitForReadiness
+	case strings.HasPrefix(vmssName, rpVMSSPrefix):
+		restartScript = rpRestartScript
+		waitForReadiness = d.rpWaitForReadiness
+	default:
+		return nil
+	}
```
petrkotas, line 518: I am not entirely convinced we need this wait for readiness. I checked the documentation of the command we are calling https://learn.microsoft.com/en-us/rest/api/compute/virtual-machine-scale-set-vms/run-command?tabs=HTTP#runcommandinput and it is supposed to by simple call that run command in the running VM.

Moreover we are not rebooting the VMs, so they will not change they readiness probe. They will always be ready. And this is the state we are checking in the readiness checks:
```
instanceUnhealthy := r.VMHealth != nil && r.VMHealth.Status != nil && r.VMHealth.Status.Code != nil && *r.VMHealth.Status.Code != "HealthState/healthy"
```

Up to my best understanding to the Azure API. We should be fine by simply waiting for the command output, check that and done.

pkg/deploy/predeploy.go
petrkotas: As i mentioned, I am not entirely sure we need this check. The thing we should check is the output of the command we are running.

petrkotas (changes requested):
Hi @rajdeepc2792 this is nice work! I have a small find, that should improve the last check. Would you please look into it?

pkg/deploy/predeploy.go
```diff
@@ -439,25 +463,91 @@ func (d *deployer) createSecret(ctx context.Context, kv keyvault.Manager, secret
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (d *deployer) restartOldScaleset(ctx context.Context, vmssName string, resourceGroupName string) error {
+	var restartScript string
+	var waitForReadiness func(ctx context.Context, vmssName string) error
```
facchettos, line 508: this isn't a pattern that is amazing. the function is quite long this is why you need it imho, that'd be easier if you'd split into smaller pieces

pkg/deploy/predeploy.go
```diff
@@ -439,25 +463,91 @@ func (d *deployer) createSecret(ctx context.Context, kv keyvault.Manager, secret
...
+
+	d.log.Printf("restarting scaleset %s", vmssName)
+	errors := make(chan error, len(scalesetVMs))
+	for _, vm := range scalesetVMs {
+		go func(id string) {
+			errors <- d.vmssvms.RunCommandAndWait(ctx, resourceGroupName, vmssName, id, mgmtcompute.RunCommandInput{
+				CommandID: to.StringPtr("RunShellScript"),
+				Script:    &[]string{restartScript},
+			})
+		}(*vm.InstanceID)
```
facchettos, line 533: does that mean that we are restarting all VMs at the same time? That seems a little dangerous imho

pkg/deploy/predeploy.go
```diff
@@ -439,25 +463,91 @@ func (d *deployer) createSecret(ctx context.Context, kv keyvault.Manager, secret
...
+	}
+
+	d.log.Print("waiting for restart script to complete")
+	for range scalesetVMs {
+		err := <-errors
+		if err != nil {
+			return err
+		}
+	}
```
facchettos, line 542: you are reimplementing `sync.ErrGroup` here

pkg/deploy/predeploy.go
```diff
@@ -439,25 +463,91 @@ func (d *deployer) createSecret(ctx context.Context, kv keyvault.Manager, secret
...
+		return nil
+	}
+
+	scalesetVMs, err := d.vmssvms.List(ctx, resourceGroupName, vmssName, "", "", "")
+	if err != nil {
+		return err
+	}
+
+	d.log.Printf("restarting scaleset %s", vmssName)
+	errors := make(chan error, len(scalesetVMs))
```
facchettos, line 526: you should be using errgroup here
https://pkg.go.dev/golang.org/x/sync/errgroup

facchettos (changes requested):
`No Unit Tests.` You should have some. 
The code needs some simplification, and I am not sure we want to restart all VMs concurrently.

pkg/deploy/predeploy.go
```diff
@@ -439,25 +463,91 @@ func (d *deployer) createSecret(ctx context.Context, kv keyvault.Manager, secret
...
+	var waitForReadiness func(ctx context.Context, vmssName string) error
+	switch {
+	case strings.HasPrefix(vmssName, gatewayVMSSPrefix):
+		restartScript = gatewayRestartScript
+		waitForReadiness = d.gatewayWaitForReadiness
+	case strings.HasPrefix(vmssName, rpVMSSPrefix):
+		restartScript = rpRestartScript
+		waitForReadiness = d.rpWaitForReadiness
+	default:
+		return nil
```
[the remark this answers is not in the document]
rajdeepc2792: I am not sure if we want to stop the RP rollout for a VMSS that isn't expected to be present in the resource group by a check here, this function is supposed to deal with restarting the gateway and rp VMs or skip otherwise.

pkg/deploy/predeploy.go
rajdeepc2792: Yes that makes sense, should be a restart and error handling for each VM, this also removes the need to use errgroup. Thanks for pointing out both concurrent restart and also errgroup usage.

pkg/deploy/predeploy.go
rajdeepc2792: Thanks @petrkotas for providing the api links and insight. I am not sure if the readiness probe will always be the same, according to [load-balancer custom probe doc](https://learn.microsoft.com/en-us/azure/load-balancer/load-balancer-custom-probe-overview#probe-interval), if the probe fails it will mark the backend-pool instance unhealthy. And as there's a LB health probe named rp-probe that checks the rp connection, there can be a chance the `run-command` returns successfully, but the rp connection takes time to become healthy.
Am I thinking correctly?

cadenmarchese: @petrkotas we define a custom probe for RP here https://github.com/Azure/ARO-RP/blob/master/pkg/deploy/assets/rp-production.json#L283-L289 that is used for readiness, so the state of the probe will change depending on the state of the RP service itself.

pkg/deploy/predeploy.go
rajdeepc2792: Thanks, as removed the concurrency no more need of errgroup.

pkg/deploy/predeploy.go
rajdeepc2792: Thanks for pointing the usage.

pkg/deploy/predeploy.go
rajdeepc2792: Updated the code and removed concurrency.

pkg/deploy/predeploy.go
rajdeepc2792: Removed the var usage.

pkg/deploy/predeploy.go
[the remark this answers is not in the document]
cadenmarchese: We may at least want to return an error like `provided vmss does not match RP or gateway prefix` here

cadenmarchese (commented):
Looks good to me, but agreed it would be nice to have some tests before merge.

pkg/deploy/predeploy.go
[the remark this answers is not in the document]
rajdeepc2792: Added an error return as suggested.

pkg/deploy/predeploy.go
```diff
@@ -439,25 +464,97 @@ func (d *deployer) createSecret(ctx context.Context, kv keyvault.Manager, secret
...
+	}
+
+	scalesetVMs, err := d.vmssvms.List(ctx, resourceGroupName, vmssName, "", "", "")
+	if err != nil {
+		return err
+	}
+
+	for _, vm := range scalesetVMs {
+		d.log.Print("waiting for restart script to complete on older vmss %s, instance %s", vmssName, *vm.InstanceID)
```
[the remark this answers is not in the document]
rajdeepc2792: Done, performed the change.

pkg/deploy/predeploy.go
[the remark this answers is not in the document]
rajdeepc2792: Also added test cases for entire predeploy functionality.

pkg/deploy/predeploy_test.go
```diff
@@ -0,0 +1,1949 @@
...
+	gatewayRGDeploymentFailed := func(d *mock_features.MockDeploymentsClient, rg *mock_features.MockResourceGroupsClient, m *mock_msi.MockUserAssignedIdentitiesClient, k *mock_keyvault.MockManager, vmss *mock_compute.MockVirtualMachineScaleSetsClient, vmssvms *mock_compute.MockVirtualMachineScaleSetVMsClient) {
+		d.EXPECT().CreateOrUpdateAndWait(ctx, gatewayRgName, gomock.Any(), gomock.Any()).Return(
+			errors.New("generic error"),
+		)
+	}
+	rpRGDeploymentFailed := func(d *mock_features.MockDeploymentsClient, rg *mock_features.MockResourceGroupsClient, m *mock_msi.MockUserAssignedIdentitiesClient, k *mock_keyvault.MockManager, vmss *mock_compute.MockVirtualMachineScaleSetsClient, vmssvms *mock_compute.MockVirtualMachineScaleSetVMsClient) {
+		d.EXPECT().CreateOrUpdateAndWait(ctx, rpRgName, gomock.Any(), gomock.Any()).Return(
+			errors.New("generic error"),
+		)
+	}
```
facchettos, line 140: This is not how you should do that, in your tests you should have the data that you feed your functions instead of this.

facchettos (dismissed):

pkg/deploy/predeploy.go
kimorris27: I get what @rajdeepc2792 is saying and I see why we would need this `time.Sleep`. I think we may need to adjust the amount of time though. Two reasons:

1. I don't think you accounted for the "timeout period" mentioned here: https://learn.microsoft.com/en-us/azure/load-balancer/load-balancer-custom-probe-overview#probe-interval. I feel like the docs aren't super clear about what the timeout period really is, but it seems like we can assume it is at most 10 seconds. So reusing the logic you used to get to 30 seconds, we would need a sleep of (15 second interval + 10 second timeout period) * 2 probes = 50 seconds if we were to account for the timeout period.
2. I was checking out the health probe on the rp-lb in prod eastus to better understand how it was configured, and I noticed a warning about how there is an Azure bug where the `numberOfProbes` will always be 1 regardless of the configured value. You can find this bug documented on Azure documentation here: https://learn.microsoft.com/en-us/azure/load-balancer/whats-new#known-issues. So with that in mind, we would actually only need 15 second interval + 10 second timeout period = 25 seconds before the health status of the RP is accurately reflected.

So in summary, I think 25 seconds is long enough, but keeping it at 30 seconds won't hurt.

rajdeepc2792: Thanks @kimorris27 for the detailed reasoning, I agree with your analysis. Also to add as pointed out in the comment the `rp-probe` configuration is set [here](https://github.com/Azure/ARO-RP/blob/master/pkg/deploy/generator/resources_rp.go#L298C1-L305C6), and as the [known-issues](https://learn.microsoft.com/en-us/azure/load-balancer/whats-new#known-issues) link suggests the azure product team is working on it, the day it is fixed the waitForReadiness check might stop reflecting the real probe status if kept the timeout to ~30seconds.

kimorris27: Yeah, good point. It may be worth leaving a comment in this code with a link to info about the bug. Or maybe there's a better way to document this.

cadenmarchese (commented):
Thanks @rajdeepc2792 for all of your work with the tests. This test coverage goes above and beyond. I think some of the repetition could be reduced and would like others who are better at reviewing tests to have a look as well.

pkg/deploy/predeploy_test.go
```diff
@@ -0,0 +1,1840 @@
...
+		return func(d *mock_features.MockDeploymentsClient, rg *mock_features.MockResourceGroupsClient, m *mock_msi.MockUserAssignedIdentitiesClient, k *mock_keyvault.MockManager, vmss *mock_compute.MockVirtualMachineScaleSetsClient, vmssvms *mock_compute.MockVirtualMachineScaleSetVMsClient, tp testParams) {
+			d.EXPECT().Get(ctx, resourceGroup, gomock.Any()).Return(deployment, returnError)
+		}
+	}
+	getSecretsMock := func(secretItems []azkeyvault.SecretItem, returnError error) mock {
+		return func(d *mock_features.MockDeploymentsClient, rg *mock_features.MockResourceGroupsClient, m *mock_msi.MockUserAssignedIdentitiesClient, k *mock_keyvault.MockManager, vmss *mock_compute.MockVirtualMachineScaleSetsClient, vmssvms *mock_compute.MockVirtualMachineScaleSetVMsClient, tp testParams) {
+			k.EXPECT().GetSecrets(ctx).Return(secretItems, returnError)
+		}
+	}
+	getSecretMock := func(d *mock_features.MockDeploymentsClient, rg *mock_features.MockResourceGroupsClient, m *mock_msi.MockUserAssignedIdentitiesClient, k *mock_keyvault.MockManager, vmss *mock_compute.MockVirtualMachineScaleSetsClient, vmssvms *mock_compute.MockVirtualMachineScaleSetVMsClient, tp testParams) {
```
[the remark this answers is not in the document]
rajdeepc2792: @cadenmarchese, I agree that some mock function duplications are still present but I think the test cases are already at a point where understanding them is bit hard as there's a need to move between the test cases and mock implementations.
Also, the number of such duplications are low(2-3 functions repeating 2-3 times) and moving these mocks to the outer scope will compromise the readability.
I would like to keep the mocks separated and duplicated for all the test functions as it keeps the implementation scoped.

pkg/deploy/predeploy_test.go
```diff
@@ -0,0 +1,1840 @@
...
+	}
+}
+
+func TestDeployManagedIdentity(t *testing.T) {
+	ctx := context.Background()
+	rgName := "rgTest"
+	existingFileName := generator.FileGatewayProductionPredeploy
+	deploymentName := strings.TrimSuffix(existingFileName, ".json")
+	notExistingFileName := "testFile"
+	genericError := errors.New("generic error")
```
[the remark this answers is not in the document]
rajdeepc2792: Agreed, moved frequently used vars to package level.

pkg/deploy/predeploy_test.go
rajdeepc2792: Updated the test cases with test data getting passed from the test cases.

pkg/deploy/predeploy_test.go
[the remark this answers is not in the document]
cadenmarchese: I agree with you as long as the implementation is different. Thanks for the followup!

cadenmarchese (approved):

kimorris27 (approved):

PR #2947, opened 2023-06-09
ARO-1423: Create rego src, templates, and tests

edisonLcardenas:

(https://issues.redhat.com/browse/ARO-1423)

Block editing of ClusterVersion by regular users. This will reduce amount of ticket we receive due to customers trying to do an unsupported action. We have a policy that will deny regulars users editing of ClusterVersion.

### Test plan for issue:
OPA Tests - Passed
Gator Tests - Passed
Real Cluster Verification - Passed


ArrisLee (changes requested):
ok, so looks like this PR is duplicated with https://github.com/Azure/ARO-RP/pull/2975 ?

xiaoyu74:
> ok, so looks like this PR is duplicated with #2975 ?

Should be different for blocking the ClusterVersion and  UpgradeConfig editing respectively.

PR #2954, opened 2023-06-12
upgrade gatekeeper to version 3.11.1

yjst2012:

Fixes [ARO-3058](https://issues.redhat.com/browse/ARO-3058)

upgrade gatekeeper version from current 3.10.0 to 3.11.1 which is the latest stable

### Test plan for issue:

tested on my local dev env

NA


pkg/operator/controllers/guardrails/guardrails_config.go
```diff
@@ -41,11 +41,11 @@ const (
 
 	defaultManagerRequestsCPU = "100m"
 	defaultManagerLimitCPU    = "1000m"
-	defaultManagerRequestsMem = "256Mi"
+	defaultManagerRequestsMem = "512Mi"
```
SudoBrendan, line 44: out of curiosity - is this following a recommendation per the version bump, or something unrelated?

pkg/operator/controllers/guardrails/staticresources/crd_gk_config_configs.yaml
SudoBrendan, line 1: Is everything under `staticresources` auto-generated or is this something that needs thorough review?

SudoBrendan (commented):
I'm a little disconnected from GK, so I left a few questions - nothing here that is very concerning on it's face, I'm just not really sure what I'm looking at to be honest :p

pkg/operator/controllers/guardrails/guardrails_config.go
petrkotas: Hi @yjst2012 I am also curious about this. Because it will have impact on a scheduler and the masters are of a fixed size.

petrkotas (commented):
Hi @yjst2012 it seems straightforward thanks! Just one question before we can get in in.

pkg/operator/controllers/guardrails/guardrails_config.go
yjst2012: yep, this change is made by gatekeeper. well basically all stuff here is porting from upstream.

pkg/operator/controllers/guardrails/staticresources/crd_gk_config_configs.yaml
yjst2012: no, it is a manual step, ie, compare the changed made to the upstream [gatekeeper.yaml](https://github.com/open-policy-agent/gatekeeper/blob/master/deploy/gatekeeper.yaml) and move it to our corresponding files under staticresources/ if it is proper to us.

petrkotas (approved):

PR #2955, opened 2023-06-13
change pod-security enforcement to baseline in Guardrails namespace to allow gatekeeper pods to start

yjst2012:

Fixes  [gatekeeper pods not starting on openshift4.11](https://redhat-internal.slack.com/archives/C02ULBRS68M/p1686235760473189)

pod security doc for ref: https://kubernetes.io/docs/concepts/security/pod-security-standards/

change pod-security enforcement to baseline in Guardrails namespace to allow gatekeeper pods to start

### Test plan for issue:

tested on my dev cluster, all pods started successfully

NA

### DIFF
--- a/pkg/operator/controllers/guardrails/staticresources/gk_namespace.yaml
+++ b/pkg/operator/controllers/guardrails/staticresources/gk_namespace.yaml
@@ -7,7 +7,9 @@ metadata:
     gatekeeper.sh/system: "yes"
     pod-security.kubernetes.io/audit: restricted
     pod-security.kubernetes.io/audit-version: latest
-    pod-security.kubernetes.io/enforce: restricted
+    # change from restricted to baseline to accommodate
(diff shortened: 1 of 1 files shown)


SrinivasAtmakuri (approved):
LGTM!

SudoBrendan (approved):
For future reviewers, `baseline` is outlined in detail here: https://kubernetes.io/docs/concepts/security/pod-security-standards/#baseline - it seems reasonable to me, though, if possible, it would be nicer to fit all our Pods into `restricted`, but that may not be possible.

cblecker (changes requested):
I think if [these lines](https://github.com/Azure/ARO-RP/commit/c1719cf74a8a56cadb351e17afde3f8465f78b0a#diff-8fb21e01ce44ecdd10aabf9824a5a0d23df646972b75c0f10e75d1d715b5cc1aR87) were uncommented, you wouldn't need this change

PR #2956, opened 2023-06-13
removed csrf

facchettos:
https://issues.redhat.com/browse/ARO-3246
https://issues.redhat.com/browse/ARO-1730

Fixes

removes csrf protection as the library is deprecated and we are not vulnerable to this kind of attack in any case as the portal is read only as of today. 

### Test plan for issue:
should work with existing tests

### DIFF
--- a/pkg/portal/http_test.go
+++ b/pkg/portal/http_test.go
@@ -81,10 +81,6 @@ func (p *testPortal) Request(method string, path string, authenticated bool, ele
 		return nil, err
 	}
 
-	if err != nil {
-		return nil, err
-	}
-
 	if authenticated {
 		var groups []string
 		if elevated {
--- a/pkg/portal/security_test.go
+++ b/pkg/portal/security_test.go
@@ -290,10 +290,6 @@ func TestSecurity(t *testing.T) {
 					t.Fatal(err)
 				}
 
-				if err != nil {
-					t.Fatal(err)
-				}
-
 				if tt2.authenticated {
 					var groups []string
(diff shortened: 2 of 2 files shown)


portal/v2/src/SSHModal.tsx
facchettos, line 1: for some reason we cannot remove the mutablerefobject from this (at least I couldn't with my limited knownledge of the portal frontend). Open to suggestions here

pkg/portal/security_test.go
```diff
@@ -291,7 +290,6 @@ func TestSecurity(t *testing.T) {
 					t.Fatal(err)
 				}
 
-				err = addCSRF(req)
 				if err != nil {
 					t.Fatal(err)
```
cadenmarchese, line 294: Same with this one.

ellis-johnson (approved):
LGTM! Tested locally and it all seems to function normally. Just those two error checks that need to be removed that were already pointed out

pkg/portal/http_test.go
```diff
@@ -81,7 +81,6 @@ func (p *testPortal) Request(method string, path string, authenticated bool, ele
 		return nil, err
 	}
 
-	err = addCSRF(req)
 	if err != nil {
 		return nil, err
 	}
```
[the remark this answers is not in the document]
facchettos: Damn it you're right I should be more careful and check what the result is after my vim macros 😭 will fix it once I'm back from Pto

PR #2959, opened 2023-06-13
Add API Server Reachability Check

niontive:

Fixes https://issues.redhat.com/browse/ARO-3330

1. Scrape metric for API server ping health. This will be used as a signal for Azure Resource Health check.
2. Use API server ping check to determine if other metrics should be scraped. This replaces the current logic of using the API server healthz status code. That status code is now checked after the ping health check.

### Test plan for issue:

INT deployment

Yes:  https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/484208/API-Server-Unreachable

### DIFF
--- a/pkg/monitor/cluster/cluster.go
+++ b/pkg/monitor/cluster/cluster.go
@@ -130,18 +130,18 @@ func (mon *Monitor) Monitor(ctx context.Context) (errs []error) {
 		})
 	}
 
-	// If API is not returning 200, don't need to run the next checks
-	statusCode, err := mon.emitAPIServerHealthzCode(ctx)
-	if err != nil {
+	// If API Server is unreachable, don't need to run the next checks
+	statusCode, err := mon.getAPIServerPingCode(ctx)
+	if err != nil || statusCode != http.StatusOK {
 		errs = append(errs, err)
-		friendlyFuncName := steps.FriendlyName(mon.emitAPIServerHealthzCode)
+		friendlyFuncName := steps.FriendlyName(mon.getAPIServerPingCode)
 		mon.log.Printf("%s: %s", friendlyFuncName, err)
 		mon.emitGauge("monitor.clustererrors", 1, map[string]string{"monitor": friendlyFuncName})
-	}
-	if statusCode != http.StatusOK {
 		return
 	}
+
 	for _, f := range []func(context.Context) error{
+		mon.emitAPIServerHealthzCode,
 		mon.emitAroOperatorHeartbeat,
 		mon.emitAroOperatorConditions,
 		mon.emitNSGReconciliation,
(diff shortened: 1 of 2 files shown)


pkg/monitor/cluster/healthz.go
```diff
@@ -21,5 +21,21 @@ func (mon *Monitor) emitAPIServerHealthzCode(ctx context.Context) (int, error) {
...
+	return err
+}
+
+func (mon *Monitor) getAPIServerPingCode(ctx context.Context) (int, error) {
+	var statusCode int
+	err := mon.cli.Discovery().RESTClient().
+		Get().
+		AbsPath("/healthz/ping").
+		Do(ctx).
+		StatusCode(&statusCode).
```
weinong, line 33: is there any reference to these status codes?

weinong (approved):
LGTM

pkg/monitor/cluster/healthz.go
niontive: We emit a metric so we can track what the status code is.

pkg/monitor/cluster/cluster.go
```diff
@@ -130,18 +130,18 @@ func (mon *Monitor) Monitor(ctx context.Context) (errs []error) {
...
 
-	// If API is not returning 200, don't need to run the next checks
-	statusCode, err := mon.emitAPIServerHealthzCode(ctx)
-	if err != nil {
+	// If API Server is unreachable, don't need to run the next checks
+	statusCode, err := mon.getAPIServerPingCode(ctx)
+	if err != nil || statusCode != http.StatusOK {
 		errs = append(errs, err)
-		friendlyFuncName := steps.FriendlyName(mon.emitAPIServerHealthzCode)
+		friendlyFuncName := steps.FriendlyName(mon.getAPIServerPingCode)
```
cadenmarchese, line 137: What's the advantage of using this metric (/healthz/ping) over what we use now (just /healthz) for the initial check?

The new metric apiserver.healthz.ping.code makes sense to me, just not sure why we would replace the existing metric here.

niontive: We're enabling [Azure Resource Health for ARO](https://learn.microsoft.com/en-us/azure/service-health/resource-health-overview). This requires us to emit metrics (aka "signal") to a Geneva monitor, which alerts Azure Resource Health.

According to the PRD, customers want a signal for cluster reachability. One of the reachability scenarios is API Server reachability, hence this PR. APIServerHealthz code incorporates many tests, and the best one to represent reachability is the ping test.

SrinivasAtmakuri (approved):
Thanks for the PR. LGTM!

s-amann (approved):

cadenmarchese (approved):

PR #2960, opened 2023-06-13
Refactor GA Get-Hive-Cluster-Deployment

SrinivasAtmakuri:
Get Hive Cluster Deployment accepts either ResourceID or ClusterDeployment namespace

JIRA: https://issues.redhat.com/browse/ARO-3308

Fixes

### DIFF
--- a/pkg/frontend/admin_hive_clusterdeployment_get.go
+++ b/pkg/frontend/admin_hive_clusterdeployment_get.go
@@ -39,18 +39,6 @@ func (f *frontend) _getAdminHiveClusterDeployment(ctx context.Context, resourceI
 	if resourceID == "" && clusterDeploymentNamespace == "" {
 		return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", "Parameters resourceID '%s' clusterDeploymentNamespace '%s' are both empty, atleast one should have a valid value.", resourceID, clusterDeploymentNamespace)
 	}
-	if resourceID == "" && clusterDeploymentNamespace != "" {
-		doc = &api.OpenShiftClusterDocument{
-			OpenShiftCluster: &api.OpenShiftCluster{
-				Properties: api.OpenShiftClusterProperties{
-					HiveProfile: api.HiveProfile{
-						Namespace: clusterDeploymentNamespace,
-					},
-				},
-			},
-		}
-	}
-
 	// if resourceID is not null, fetch the clusterDeploymentNamespace using the resourceID
 	// when parameteres resourceID and clusterDeploymentNamespace are not null, clusterDeploymentNamespace will be ignored
 	if resourceID != "" {
@@ -60,6 +48,16 @@ func (f *frontend) _getAdminHiveClusterDeployment(ctx context.Context, resourceI
 		}
 		if doc.OpenShiftCluster.Properties.HiveProfile.Namespace == "" {
 			return nil, api.NewCloudError(http.StatusNoContent, api.CloudErrorCodeResourceNotFound, "", "cluster '%s' is not managed by hive", resourceID)
+		}
+	} else {
+		doc = &api.OpenShiftClusterDocument{
+			OpenShiftCluster: &api.OpenShiftCluster{
+				Properties: api.OpenShiftClusterProperties{
+					HiveProfile: api.HiveProfile{
+						Namespace: clusterDeploymentNamespace,
+					},
+				},
+			},
 		}
 	}
 
--- a/pkg/frontend/admin_hive_clusterdeployment_get.go
+++ b/pkg/frontend/admin_hive_clusterdeployment_get.go
@@ -6,7 +6,6 @@ package frontend
 import (
 	"context"
 	"net/http"
-	"path/filepath"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


SrinivasAtmakuri:
The code is tested.

pkg/frontend/admin_hive_clusterdeployment_get.go
```diff
@@ -19,9 +18,9 @@ import (
 func (f *frontend) getAdminHiveClusterDeployment(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
-	resourceID := strings.TrimPrefix(filepath.Dir(r.URL.Path), "/admin")
-	b, err := f._getAdminHiveClusterDeployment(ctx, resourceID)
-
+	resourceID := r.URL.Query().Get("resourceID")
+	clusterDeploymentNamespace := r.URL.Query().Get("clusterDeploymentNamespace")
```
facchettos, line 22: I guess I am not completely a fan of this pattern, it kind of breaks REST principles. One solution would be to do some url encoding on the resourceID so we could have a route that looks like that. `r.Get("/admin/resourceID/{resourceID}/clusterdeployment")` and `r.Get("/admin/hivenamespace/{namespace}/clusterdeployment")` without having to do tricks with url trimming and stuff (it's what we should have done from the beginning tbh).

jaitaiwan: The object that's trying to be retrieved here is the cluster deployment. It's neither a subdocument of a resource nor a namespace so it doesn't make sense to have the clusterdeployment as a subdocument of either of those things. Both of those things are metadata of the clusterdeployment strictly speaking.

If adhering to typical rest principals was stated goal for portal endpoints, the query would be GET/POST /admin/clusterdeployment/search with a query variable of either resourceID= or namespace=

pkg/frontend/admin_hive_clusterdeployment_get.go
```diff
@@ -30,24 +29,43 @@ func (f *frontend) getAdminHiveClusterDeployment(w http.ResponseWriter, r *http.
...
+	// when parameteres resourceID and clusterDeploymentNamespace are not null, clusterDeploymentNamespace will be ignored
+	if resourceID != "" {
+		doc, err := f.dbOpenShiftClusters.Get(ctx, strings.ToLower(resourceID))
+		if err != nil {
+			return nil, api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeNotFound, "", "cluster '%s' not found", resourceID)
+		}
+		if doc.OpenShiftCluster.Properties.HiveProfile.Namespace == "" {
+			return nil, api.NewCloudError(http.StatusNoContent, api.CloudErrorCodeResourceNotFound, "", "cluster '%s' is not managed by hive", resourceID)
+		}
 	}
```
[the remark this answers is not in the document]
SrinivasAtmakuri: Made the change.

SrinivasAtmakuri:
PR for GenevaAction [repo](https://msazure.visualstudio.com/One/_git/Compute-ARO-GenevaActions/pullrequest/8336307)

pkg/frontend/admin_hive_clusterdeployment_get.go
facchettos: it's the cluster deployment associated with a specific resource ID or a specific namespace. This is not a search, this is a GET operation.

jaitaiwan: Resource ID really isn't strictly an ID though. It's a fully fledged URI path. Sticking an entire path (even if url encoded) inside another path at the very least is aesthetically horrible.

Usually we would retrieve an integer or uuid ID of a specific object if we're doing a GET /resource (e.g. kubernetes object id). Even though we know in practice there's a 1-to-1 relationship between ClusterDeployment documents and namespaces/resourceIds there's technically the ability for there to be n+ ClusterDeployment documents per either of those identifiers... even though we know that's ridiculous with current implementations.

So again despite that, and in the strictest sense, this would be a search operation and if we care that deeply about changing how this GA works (which currently follows the pattern of how other GAs are made) you need to define the line somewhere. There's no reason IMO to stop at resourceID in the url if you're going to be semantic.

I'd be happy with the url `/admin/hivenamespace/{namespace}/clusterdeployment/{clusterdeploymentName}` considering that there can not be more than one with the same name in the same namespace, but then there'd be no consistent comparable way to do resourceID that I can think of. I'd be back at a search endpoint at that point.

facchettos: you are overengineering it, we will deal with that iff there is a change and there can be more than 1 clusterdeployment per cluster which isn't currently (and prob won't be for a long time) the case. The resourceID's appearance takes nothing away from the fact that it is an ID for that cluster, and resources from that cluster should be under its resourceID, and so it's absolutely the same thing than a UUID

petrkotas (approved):
If @SudoBrendan, @facchettos and @jaitaiwan agrees, lets proceed. I honestly do not see reason to make internal tooling work different that it is now.

tsatam (approved):
This PR looks good to me, thank you for doing this! This change will be vital for improving our automation and error handling for customers experiencing failed cluster installations. 

I agree with @petrkotas that this change is fine as-is, but the proposed RESTful URI looks fine as-well for our current state (one ClusterDeployment per namespace, always named `cluster`). If we were to change that state, there would be a few changes necessary, but I do not think we would.

jaitaiwan:
Just noting the attached PR is approved and ready to merge now as well. After fixing conflicts we can merge this - but I assume we'll be doing this after the next release.

SrinivasAtmakuri:

PR #2964, opened 2023-06-14
clarify deploy-full-rp-service-in-dev.md guide

mbukatov:

None.

Improves "Deploy an Entire RP Development Service" guide so that one can avoid problems caused by misunderstanding of the procedure.

### Test plan for issue:

Changes are based on my experience with this procedure. No further testing expected beyond review.

### DIFF
--- a/docs/deploy-full-rp-service-in-dev.md
+++ b/docs/deploy-full-rp-service-in-dev.md
@@ -3,31 +3,61 @@
 ## Prerequisites
 
 1. Your development environment is prepared according to the steps outlined in [Prepare Your Dev Environment](./prepare-your-dev-environment.md)
-
+2. During the deployment, it's recommended to avoid editing files in your
+   ARO-RP repository so that `git status` reports a clean working tree.
+   Otherwise aro container image will have `-dirty` suffix, which can be
+   problematic:
+    - if the working tree becomes dirty during the process (eg. because you
+      create a temporary helper script to run some of the setup), you could end
+      up with different image tag pushed in the azure image registry compared
+      to the tag expected by aro deployer
+    - with a dirty tag, it's not clear what's actually in the image
 
 ## Deploying an int-like Development RP
 
-1. Fetch the most up-to-date secrets with `make secrets`
+1. Fetch the most up-to-date secrets specifying `SECRET_SA_ACCOUNT_NAME` to the
+   name of the storage account containing your shared development environment
+   secrets, eg.:
 
-1. Copy and source your environment file.
-    ```bash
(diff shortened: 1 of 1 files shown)


mbukatov:
Publishing as a draft for visibility, I still need to polish this a bit. Feedback welcome.

docs/deploy-full-rp-service-in-dev.md
```diff
@@ -3,31 +3,56 @@
 ## Prerequisites
 
 1. Your development environment is prepared according to the steps outlined in [Prepare Your Dev Environment](./prepare-your-dev-environment.md)
-
+2. During the deployment, don't edit files in your ARO-RP repository so that
+   `git status` reports a clean working tree. Otherwise aro container image
+   will have `-dirty` suffix, which can cause problems esp. when the repository
+   becomes dirty during the procedure.
```
yehornaumenko, line 9: I guess we need to prevent this programmatically in this scenario because it can't be wanted. Unless somebody wanna deploy an RP using some not committed changes.
I would add some checks or warning messages to the scripts we use or _make_'s targets.
But we can do it in another PR. Does it make sense?

jaitaiwan: What sort of problems does the -dirty suffix cause? Is it only an issue when deploying a shared rp or is it an issue for local dev as well?

SrinivasAtmakuri: One solution would be to set COMMIT/TAG = latest if RP_MODE == development, this will make reruns or testing a lot easier.

cadenmarchese: To me it's more of a recommendation. You can definitely deploy with the dirty tag if you want (I do it and sometimes it's an accident but it still works), but you will have the potential issues mentioned above.

Should we just change the wording to "it's recommended to..." and add the content here to the document? https://github.com/Azure/ARO-RP/pull/2964/files#r1238448125

mbukatov: I updated the text as suggested by @cadenmarchese .

mbukatov:
The changes are ready for review. I addressed the feedback about the dirty image tag, and tweaked the text about container registry used for `make publish-image*` targets.

tiguelu (dismissed):
LGTM

cadenmarchese (dismissed):

mbukatov:
Rebased on top of current master (14e3c5a88) to unblock the github checks.

tiguelu (approved):

PR #2966, opened 2023-06-15
Remove non-current Azure SDK dependencies from being considered

hawkowl:
Minor cleanup from #2817 that doesn't have to be a part of it


ellis-johnson (approved):
LGTM!

PR #2967, opened 2023-06-15
Installer images use sha256

bennerv:

Fixes https://issues.redhat.com/browse/ARO-3276

Allows us to push installer images without having them affect the entire fleet at once and in an inconsistent manner.  

Right now here's how it happens:
1. Push a new installer image
2. imagePullPolicy on hive is `IfNotPresent`
3. If the node has the image cached, it will use the older installer image
4. if the node does not have the image cached, it will use the new installer image

our e2e will be inconsistent after an installer image push, and there's no guarantee around which image is used and when only the newer images are used. 

### Test plan for issue:

1. Roll out this change via `update_ocp_versions`
2. Run an e2e
3. Check the installer image being used is being pulled with image digest

### Is there any documentation that needs to be updated for this 

Yes. TBD

High level:
1. Push new image to all ACRs
5. Update the SHAs used here
6. Build the update_ocp_versions binary
7. Run the ev2 release for update_ocp_versions in SDP manner
8. Profit

There's some docs needed on how to pull the imagedigest for a new image.

### DIFF
--- a/cmd/aro/update_ocp_versions.go
+++ b/cmd/aro/update_ocp_versions.go
@@ -31,11 +31,18 @@ func getLatestOCPVersions(ctx context.Context, log *logrus.Entry) ([]api.OpenShi
 	ocpVersions := []api.OpenShiftVersion{}
 
 	for _, vers := range version.HiveInstallStreams {
+		installerPullSpec := fmt.Sprintf("%s/aro-installer:%s", dstRepo, vers.Version.MinorVersion())
+		digest, ok := version.InstallerImageDigest[vers.Version.MinorVersion()]
+		if !ok {
+			return nil, fmt.Errorf("no digest found for version %s", vers.Version.String())
+		}
+
+		installerPullSpec = fmt.Sprintf("%s@sha256:%s", installerPullSpec, digest)
 		ocpVersions = append(ocpVersions, api.OpenShiftVersion{
 			Properties: api.OpenShiftVersionProperties{
 				Version:           vers.Version.String(),
 				OpenShiftPullspec: vers.PullSpec,
-				InstallerPullspec: fmt.Sprintf("%s/aro-installer:release-%s", dstRepo, vers.Version.MinorVersion()),
+				InstallerPullspec: installerPullSpec,
 				Enabled:           true,
 			},
 		})
--- a/pkg/cluster/version.go
+++ b/pkg/cluster/version.go
@@ -5,14 +5,12 @@ package cluster
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"strings"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/database"
 	"github.com/Azure/ARO-RP/pkg/env"
-	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
 // openShiftClusterDocumentVersioner is the interface that validates and obtains the version from an OpenShiftClusterDocument.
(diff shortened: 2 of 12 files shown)


cmd/aro/update_ocp_versions.go
```diff
@@ -31,11 +31,19 @@ func getLatestOCPVersions(ctx context.Context, log *logrus.Entry) ([]api.OpenShi
 	ocpVersions := []api.OpenShiftVersion{}
 
 	for _, vers := range version.HiveInstallStreams {
+		installerPullSpec := fmt.Sprintf("%s/aro-installer:%s", dstRepo, vers.Version.MinorVersion())
```
hawkowl, line 34: Initially, I was planning that this would come from RP-Config (so that we can update the versions w/o a new code artifact). Should we do that instead, or after this PR?

cmd/aro/update_ocp_versions.go
```diff
@@ -31,11 +31,19 @@ func getLatestOCPVersions(ctx context.Context, log *logrus.Entry) ([]api.OpenShi
...
+		digest, ok := version.InstallerImageDigest[vers.Version.MinorVersion()]
+		if !ok {
```
cadenmarchese, line 36: Should we leave this out? To me it looks like we're still quietly supporting non-tagged when we want to use SHA always.

bennerv: I think that's fine.  Let's just error out if the SHAs are not present.  I don't think anyone is using this for hive and if they are we can get docs updated.

bennerv: Updated to return an error directly if one doesn't exist.

pkg/util/version/const.go
```diff
@@ -25,6 +25,14 @@ const (
 
 var GitCommit = "unknown"
 
+// InstallerImageDigest is the mapping of a minor version to the aro-installer wrapper digest
+// this allows us to utilize SDP instead of pushing up the tag and it rolling out to all regions
+// at once
+var InstallerImageDigest = map[string]string{
+	NewVersion(4, 10).MinorVersion(): "eef5f0d82ab07c866999f99a632edafd845e0dc48a0f06eac9df46b0ab882231",
+	NewVersion(4, 11).MinorVersion(): "82869dd10841046c4b98fc46c6f87030c82b320b82f990610c1fa87150004730",
```
jhoreman, line 33: I believe these are the current SHA values of the installer images. How do we intend to manage these (and the version numbers in the future?) If we have to make an update here every time we want to rollout a new installer image we'd be linking the ARO-Installer code work directly to RP work deploy. Any way to interpret these values?

bennerv: We're going to move them into the RP config.  https://github.com/Azure/ARO-RP/pull/2967#discussion_r1231676683

mbarnes:
Verified Hive is using a SHA-based installer image after deploying this PR to our INT environment.

public-int-installer-image

pkg/cluster/version.go
```diff
@@ -28,28 +26,6 @@ type openShiftClusterDocumentVersionerService struct{}
...
-	// Honor any installer pull spec override
-	installerPullSpec := env.LiveConfig().DefaultInstallerPullSpecOverride(ctx)
-	if installerPullSpec == "" {
-		installerPullSpec = fmt.Sprintf("%s/aro-installer:release-%s", env.ACRDomain(), version.DefaultInstallStream.Version.MinorVersion())
-	}
-
-	// add the default OCP version as we require it as an install target
-	// if this is removed, we need to also update the logic in
-	// pkg/frontend/frontend.go, pkg/frontend/validate.go
-	defaultVersion := &api.OpenShiftVersion{
```
s-amann, line 40: do we need to double check we have versions in all regions?

bennerv: That's my only concern with this change.  Essentially if we don't have any version within a region no clusters will successfully install with this change.  

If we don't have a version though, both the frontend and backend should return that the version doesn't exist, which I think is okay because it should pop up on our cluster install failure dashboard and we can catch it.  

Additionally, e2e will fail to run.

pkg/frontend/validate.go
```diff
@@ -205,13 +205,12 @@ func validateAdminMasterVMSize(vmSize string) error {
 func (f *frontend) validateInstallVersion(ctx context.Context, doc *api.OpenShiftCluster) error {
 	// If this request is from an older API or the user never specified
 	// the version to install we default to the DefaultInstallStream.Version
+	// TODO: We should set default version in cosmosdb instead of hardcoding it in golang code
```
s-amann, line 208: is this TODO already in JIRA?

s-amann (approved):

pkg/frontend/validate.go
bennerv: no i need to create it.

PR #2969, opened 2023-06-19
Add nsg enhancement permission validation for FP/Cluster SPs

nwnt:

Fixes: ARO-3293

Without the validation, only Internal Server Error will be returned to the customer creating the cluster with the private preview NSG enhancement feature.  This validation will make it clearer what needs to be done to make the cluster creation successful

### Test plan for issue:
- Added unit tests
- Also tested in local dev environment and could see the message

N/A


niontive (dismissed):

pkg/api/validate/dynamic/dynamic.go
```diff
@@ -781,6 +783,61 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+				"",
+				errMsgSPHasNoRequiredPermissionsOnNSG,
+				dv.authorizerType,
+				dv.appID,
+				*nsgID,
+			)
+		}
+		if err != nil {
+			return err
+		}
```
facchettos, line 835: can you put this into a function please?

facchettos (dismissed):

pkg/api/validate/dynamic/dynamic.go
nwnt: @facchettos done

s-amann (approved):

PR #2974, opened 2023-06-21
Add tests for clusteroperatoraro controller

tsatam:

None

Add additional test coverage for the clusteroperatoraro controller, responsible for populating the ARO Cluster Operator resource. 

The tests specifically cover the issue found and fixed in #2903 by validating that the version remains set on the resource on all Reconcile scenarios.  

### Test plan for issue:

The PR is just tests

No.


SudoBrendan (approved):
This is really great. Should we add a test to never set Degraded status?

tsatam:
I added a test to ensure we do not set Degraded=True on the operator when controller Degraded=True conditions are present.

jhoreman (approved):
LGTM, thanks for this!

SudoBrendan (approved):

cadenmarchese (approved):

PR #2976, opened 2023-06-22
create monitor for certificate expirations

dem4gus:

Fixes [ARO-2916](https://issues.redhat.com/browse/ARO-2916)

Creates a monitor to report expiration dates of certificates that are managed by the RP. The Geneva certificate is always monitored, ingress and API certificates are only monitored if the cluster's domain is not custom.

### Test plan for issue:

Unit tests for both managed and unmanaged domains.

### DIFF
--- a/pkg/monitor/cluster/certificateexpirationstatuses.go
+++ b/pkg/monitor/cluster/certificateexpirationstatuses.go
@@ -29,12 +29,14 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 	var certs []*x509.Certificate
 
 	mdsdCert, err := mon.getCertificate(ctx, operator.SecretName, operator.Namespace, genevalogging.GenevaCertName)
-	if kerrors.IsNotFound(err) {
-		mon.emitGauge(secretMissingMetricName, int64(1), map[string]string{
-			"secretMissing": operator.SecretName,
-		})
-	} else if err != nil {
+	if err != nil {
+            if !kerrors.IsNotFound(err) {
 		return err
+	    }
+	    mon.emitGauge(secretMissingMetricName, int64(1), map[string]string{
+			"secretMissing": operator.SecretName,
+	    })
+	} 
 	} else {
 		certs = append(certs, mdsdCert)
 	}
--- a/pkg/monitor/cluster/certificateexpirationstatuses.go
+++ b/pkg/monitor/cluster/certificateexpirationstatuses.go
@@ -45,7 +45,7 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 			return err
 		}
 		ingressSecretName := ingressController.Spec.DefaultCertificate.Name
-		for _, secretName := range []string{ingressSecretName, strings.Replace(ingressSecretName, "-ingress", "-apiserver", 1)} {
+		for _, secretName := range []string{ingressSecretName, strings.Replace(ingressSecretName, "-ingress", "-apiserver", 1)} { // certificate name is uuid + "-ingress" or "-apiserver"
 			certificate, err := mon.getCertificate(ctx, secretName, operator.Namespace, corev1.TLSCertKey)
 			if kerrors.IsNotFound(err) {
 				mon.emitGauge(secretMissingMetricName, int64(1), map[string]string{
--- a/pkg/monitor/cluster/certificateexpirationstatuses.go
+++ b/pkg/monitor/cluster/certificateexpirationstatuses.go
@@ -56,7 +56,7 @@ func (mon *Monitor) getCertificate(ctx context.Context, secretName, secretNamesp
 
 	certBlock, _ := pem.Decode(secret.Data[secretKey])
 	if certBlock == nil {
-		return &x509.Certificate{}, fmt.Errorf("certificate data for %s not found", secretName)
+		return &x509.Certificate{}, fmt.Errorf(`certificate "%s" not found on secret "%s"`, secretKey, secretName)
 	}
 	// we only care about the first certificate in the block
 	return x509.ParseCertificate(certBlock.Bytes)
--- a/pkg/monitor/cluster/certificateexpirationstatuses.go
+++ b/pkg/monitor/cluster/certificateexpirationstatuses.go
@@ -71,7 +71,7 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 func (mon *Monitor) getCertificate(ctx context.Context, secretName, secretNamespace, secretKey string) (*x509.Certificate, error) {
 	secret, err := mon.cli.CoreV1().Secrets(secretNamespace).Get(ctx, secretName, metav1.GetOptions{})
 	if err != nil {
-		return &x509.Certificate{}, err
+		return nil, err
 	}
 
 	certBlock, _ := pem.Decode(secret.Data[secretKey])
--- a/pkg/monitor/cluster/certificateexpirationstatuses.go
+++ b/pkg/monitor/cluster/certificateexpirationstatuses.go
@@ -76,7 +76,7 @@ func (mon *Monitor) getCertificate(ctx context.Context, secretName, secretNamesp
 
 	certBlock, _ := pem.Decode(secret.Data[secretKey])
 	if certBlock == nil {
-		return &x509.Certificate{}, fmt.Errorf(`certificate "%s" not found on secret "%s"`, secretKey, secretName)
+		return nil, fmt.Errorf(`certificate "%s" not found on secret "%s"`, secretKey, secretName)
 	}
 	// we only care about the first certificate in the block
 	return x509.ParseCertificate(certBlock.Bytes)
--- a/pkg/monitor/cluster/certificateexpirationstatuses.go
+++ b/pkg/monitor/cluster/certificateexpirationstatuses.go
@@ -28,15 +28,14 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 	// report NotAfter dates for Ingress and API (on managed domains), and Geneva (always)
 	var certs []*x509.Certificate
 
-	mdsdCert, err := mon.getCertificate(ctx, operator.SecretName, operator.Namespace, genevalogging.GenevaCertName)
+	mdsdCert, err := mon.getCertificate(ctx, operator.Namespace, operator.SecretName, genevalogging.GenevaCertName)
 	if err != nil {
-            if !kerrors.IsNotFound(err) {
-		return err
-	    }
-	    mon.emitGauge(secretMissingMetricName, int64(1), map[string]string{
+		if !kerrors.IsNotFound(err) {
+			return err
+		}
+		mon.emitGauge(secretMissingMetricName, int64(1), map[string]string{
 			"secretMissing": operator.SecretName,
-	    })
-	} 
+		})
 	} else {
 		certs = append(certs, mdsdCert)
 	}
@@ -48,7 +47,7 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 		}
 		ingressSecretName := ingressController.Spec.DefaultCertificate.Name
 		for _, secretName := range []string{ingressSecretName, strings.Replace(ingressSecretName, "-ingress", "-apiserver", 1)} { // certificate name is uuid + "-ingress" or "-apiserver"
-			certificate, err := mon.getCertificate(ctx, secretName, operator.Namespace, corev1.TLSCertKey)
+			certificate, err := mon.getCertificate(ctx, operator.Namespace, secretName, corev1.TLSCertKey)
 			if kerrors.IsNotFound(err) {
 				mon.emitGauge(secretMissingMetricName, int64(1), map[string]string{
 					"secretMissing": secretName,
@@ -70,7 +69,7 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 	return nil
 }
 
-func (mon *Monitor) getCertificate(ctx context.Context, secretName, secretNamespace, secretKey string) (*x509.Certificate, error) {
+func (mon *Monitor) getCertificate(ctx context.Context, secretNamespace, secretName, secretKey string) (*x509.Certificate, error) {
 	secret, err := mon.cli.CoreV1().Secrets(secretNamespace).Get(ctx, secretName, metav1.GetOptions{})
 	if err != nil {
 		return nil, err
--- a/pkg/monitor/cluster/certificateexpirationstatuses.go
+++ b/pkg/monitor/cluster/certificateexpirationstatuses.go
@@ -5,6 +5,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"math"
 	"strings"
 	"time"
 
@@ -63,9 +64,11 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 	}
 
 	for _, cert := range certs {
+		daysUntilExpiration := int(math.Round(cert.NotAfter.UTC().Sub(time.Now().UTC()).Hours() / 24))
 		mon.emitGauge(certificateExpirationMetricName, 1, map[string]string{
-			"subject":        cert.Subject.CommonName,
-			"expirationDate": cert.NotAfter.UTC().Format(time.RFC3339),
+			"subject":             cert.Subject.CommonName,
+			"expirationDate":      cert.NotAfter.UTC().Format(time.RFC3339),
+			"daysUntilExpiration": fmt.Sprintf("%d", daysUntilExpiration),
 		})
 	}
 	return nil
--- a/pkg/monitor/cluster/certificateexpirationstatuses.go
+++ b/pkg/monitor/cluster/certificateexpirationstatuses.go
@@ -8,9 +8,10 @@ import (
 	"strings"
 	"time"
 
+	operatorv1 "github.com/openshift/api/operator/v1"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/Azure/ARO-RP/pkg/operator"
 	"github.com/Azure/ARO-RP/pkg/operator/controllers/genevalogging"
@@ -41,12 +42,18 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 	}
 
 	if dns.IsManagedDomain(mon.oc.Properties.ClusterProfile.Domain) {
-		ingressController, err := mon.operatorcli.OperatorV1().IngressControllers("openshift-ingress-operator").Get(ctx, "default", metav1.GetOptions{})
+		ic := &operatorv1.IngressController{}
+		err := mon.clientset.Get(ctx, client.ObjectKey{
+			Namespace: "openshift-ingress-operator",
+			Name:      "default",
+		}, ic)
 		if err != nil {
 			return err
 		}
-		ingressSecretName := ingressController.Spec.DefaultCertificate.Name
-		for _, secretName := range []string{ingressSecretName, strings.Replace(ingressSecretName, "-ingress", "-apiserver", 1)} { // certificate name is uuid + "-ingress" or "-apiserver"
+		ingressSecretName := ic.Spec.DefaultCertificate.Name
+
+		// secret with managed certificates is uuid + "-ingress" or "-apiserver"
+		for _, secretName := range []string{ingressSecretName, strings.Replace(ingressSecretName, "-ingress", "-apiserver", 1)} {
 			certificate, err := mon.getCertificate(ctx, operator.Namespace, secretName, corev1.TLSCertKey)
 			if kerrors.IsNotFound(err) {
 				mon.emitGauge(secretMissingMetricName, int64(1), map[string]string{
@@ -70,7 +77,11 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 }
 
 func (mon *Monitor) getCertificate(ctx context.Context, secretNamespace, secretName, secretKey string) (*x509.Certificate, error) {
-	secret, err := mon.cli.CoreV1().Secrets(secretNamespace).Get(ctx, secretName, metav1.GetOptions{})
+	secret := &corev1.Secret{}
+	err := mon.clientset.Get(ctx, client.ObjectKey{
+		Namespace: secretNamespace,
+		Name:      secretName,
+	}, secret)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/monitor/cluster/certificateexpirationstatuses.go
+++ b/pkg/monitor/cluster/certificateexpirationstatuses.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/pem"
+	"fmt"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -19,30 +20,23 @@ import (
 
 func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error {
 	// report NotAfter dates for Geneva (always), Ingress, and API (on managed domain) certificates
-	var certs []x509.Certificate
+	var certs []*x509.Certificate
 
 	mdsdCert, err := mon.getCertificate(ctx, operator.SecretName, operator.Namespace, genevalogging.GenevaCertName)
 	if err != nil {
 		return err
 	}
-	certs = append(certs, *mdsdCert[0])
+	certs = append(certs, mdsdCert)
 
 	if dns.IsManagedDomain(mon.oc.Properties.ClusterProfile.Domain) {
 		infraID := mon.oc.Properties.InfraID
-		ingressCertificateName := infraID + "-ingress"
-		apiCertificateName := infraID + "-apiserver"
-
-		ingressCertificate, err := mon.getCertificate(ctx, ingressCertificateName, operator.Namespace, corev1.TLSCertKey)
-		if err != nil {
-			return err
+		for _, secretName := range []string{infraID + "-ingress", infraID + "-apiserver"} {
+			certificate, err := mon.getCertificate(ctx, secretName, operator.Namespace, corev1.TLSCertKey)
+			if err != nil {
+				return err
+			}
+			certs = append(certs, certificate)
 		}
-
-		apiCertificate, err := mon.getCertificate(ctx, apiCertificateName, operator.Namespace, corev1.TLSCertKey)
-		if err != nil {
-			return err
-		}
-
-		certs = append(certs, *ingressCertificate[0], *apiCertificate[0])
 	}
 
 	for _, cert := range certs {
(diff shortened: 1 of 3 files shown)


pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,66 @@
...
+		if err != nil {
+			return err
+		}
+
+		apiCertificate, err := mon.getCertificate(ctx, apiCertificateName, operator.Namespace, corev1.TLSCertKey)
+		if err != nil {
+			return err
+		}
+
+		certs = append(certs, *ingressCertificate[0], *apiCertificate[0])
```
facchettos, line 45: here it'd be better with a loop, instead of duplicating code with only a param change
```suggestion
                for _,v := range []string{apiCertificateName, ingressCertificateName}{
                        cert , err := mon.getCertificate(ctx, v , operator.Namespace, corev1.TLSCertKey)
                    	if err != nil {
			         return err
			}    
			certs = append(certs, cert)
                }
```

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (mon *Monitor) getCertificate(ctx context.Context, secretName, secretNamespace, secretKey string) ([]*x509.Certificate, error) {
+	secret, err := mon.cli.CoreV1().Secrets(secretNamespace).Get(ctx, secretName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	certBlock, _ := pem.Decode(secret.Data[secretKey])
+	return x509.ParseCertificates(certBlock.Bytes)
+}
```
facchettos, line 66: It would be better to return directly the `x509.Certificate` instead of the slice here imho, since we always take the first element of that slice without ever ranging over it. (plus a check of the length would be a nice plus to avoid unexpected panics)

facchettos (changes requested):
I left some suggestion to simplify the code. 
On a sidenote, would it make sense to start to alert a little before the expiration so we can do the right thing before some problems happen? (either azcomm or pucm)

kimorris27 (commented):
LGTM other than the outstanding lint error!

s-amann (changes requested):
I like this idea, and great work so far. Some suggestions below.

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,66 @@
...
+		certs = append(certs, *ingressCertificate[0], *apiCertificate[0])
+	}
+
+	for _, cert := range certs {
+		mon.emitGauge("certificate.expirationdate", 1, map[string]string{
+			"subject":        cert.Subject.CommonName,
+			"expirationDate": cert.NotAfter.UTC().Format(time.RFC3339),
+		})
+	}
+	mon.emitGauge("managedCertificates.count", int64(len(certs)), nil)
```
cadenmarchese, line 54: Not sure if needed, since we are already emitting one `certificate.expirationdate` metric per cert. Is there a use case you had in mind?

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,66 @@
...
+		infraID := mon.oc.Properties.InfraID
+		ingressCertificateName := infraID + "-ingress"
+		apiCertificateName := infraID + "-apiserver"
+
+		ingressCertificate, err := mon.getCertificate(ctx, ingressCertificateName, operator.Namespace, corev1.TLSCertKey)
+		if err != nil {
+			return err
+		}
+
+		apiCertificate, err := mon.getCertificate(ctx, apiCertificateName, operator.Namespace, corev1.TLSCertKey)
```
AldoFusterTurpin, line 40: I think it would be a good idea to use a for as Jeremy suggested, we should avoid code duplication.

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,66 @@
...
+	return nil
+}
+
+func (mon *Monitor) getCertificate(ctx context.Context, secretName, secretNamespace, secretKey string) ([]*x509.Certificate, error) {
+	secret, err := mon.cli.CoreV1().Secrets(secretNamespace).Get(ctx, secretName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	certBlock, _ := pem.Decode(secret.Data[secretKey])
```
AldoFusterTurpin, line 64: According to the [docs](https://pkg.go.dev/encoding/pem#Decode):  _If no PEM data is found, p is nil and the whole of the input is returned in rest._

So even though x509.ParseCertificates(nil) will not panic (looking at the implementation), adding a check against nil after line 64 with an early explicit return could be nice as passing nil around is something I would avoid.

pkg/monitor/cluster/certificateexpirationstatuses.go
AldoFusterTurpin, line 45: I think we should either return just one certificate from getCertificate or return the slices as done now (changing the name to get certificate**s**) but always iterate over the returned value as iterating over a nil slice does not panic. I prefer just to return one and not the slice.

***ingressCertificate[0]** and ***apiCertificate[0]** -> I would ensure those are not nil as I am not sure if we are safe against a nil pointer panic here.

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error {
+	// report NotAfter dates for Geneva (always), Ingress, and API (on managed domain) certificates
+	var certs []x509.Certificate
+
+	mdsdCert, err := mon.getCertificate(ctx, operator.SecretName, operator.Namespace, genevalogging.GenevaCertName)
+	if err != nil {
+		return err
+	}
+	certs = append(certs, *mdsdCert[0])
```
AldoFusterTurpin, line 28: I would check against len(mdsdCert) and also ensure *mdsdCert[0] is not nil as this _*mdsdCert[0]_ can panic or maybe we can be appending nil to _certs_ resulting in errors that are really difficult to debug.

AldoFusterTurpin (changes requested):
@dem4gus Good work so far! 😄 

A few suggestions as I think we should check pointers and length of slices in the whole PR (or just return one certificate). I  don't feel pretty safe dereferencing those possible nil pointers.

pkg/monitor/cluster/certificateexpirationstatuses.go
dem4gus: I modeled this after the daemonset monitor, but I don't have a specific case in mind. It can just be removed.

pkg/monitor/cluster/certificateexpirationstatuses.go
dem4gus: I changed it to the singular `x509.ParseCertificate()` since we only care about the first certificate in the chain anyway. That should save us from having to do the length checks at all.

pkg/monitor/cluster/certificateexpirationstatuses.go
dem4gus: I changed the underlying certificate parsing from `x509.GetCertificates` to just `x509.GetCertificate` to help with this. It only parses the first certificate in the chain, but that's the only one we care about checking anyway.

pkg/monitor/cluster/certificateexpirationstatuses.go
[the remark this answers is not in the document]
dem4gus: Done in `getCertificate`.

pkg/monitor/cluster/certificateexpirationstatuses_test.go
```diff
@@ -0,0 +1,139 @@
...
+					"expirationDate": expirationString,
+				},
+			},
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.Background()
+
+			var secrets []runtime.Object
+			_, genevaCert, err := utiltls.GenerateTestKeyAndCertificate("geneva.certificate", nil, nil, false, false, func(template *x509.Certificate) {
```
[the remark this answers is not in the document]
dem4gus: Thank you.

pkg/monitor/cluster/certificateexpirationstatuses_test.go
```diff
@@ -0,0 +1,141 @@
...
+				template.NotAfter = expiration
+			})
```
AldoFusterTurpin, line 67: I think it is a good idea to remove the inline function definition. This will have benefits in the long run. Something like:
replace above chunk with....
```go
tweakTemplateFn := tweakTemplateFn(expiration)
_, genevaCert, err := utiltls.GenerateTestKeyAndCertificate("geneva.certificate", nil, nil, false, false, tweakTemplateFn)
```
and create
```go
func tweakTemplateFn(expiration time.Time) func(*x509.Certificate) {
	return func(template *x509.Certificate) {
		template.NotAfter = expiration
	}
}
```
Also, we can now reuse the function as it was originally hardcoded in two different places.

pkg/monitor/cluster/certificateexpirationstatuses_test.go
```diff
@@ -0,0 +1,141 @@
...
+					Name:      "cluster",
+					Namespace: "openshift-azure-operator",
+				},
+				Data: map[string][]byte{
+					"gcscert.pem": pem.EncodeToMemory(&pem.Block{
+						Type:  "CERTIFICATE",
+						Bytes: genevaCert[0].Raw,
+					}),
+				},
+			})
```
AldoFusterTurpin, line 82: IMHO we can improve the readability if we move the secret to a variable:
```go
s := &corev1.Secret{
	        Name:      "cluster",
	        Namespace: "openshift-azure-operator",
        },
        Data: map[string][]byte{
	        "gcscert.pem": pem.EncodeToMemory(&pem.Block{
		        Type:  "CERTIFICATE",
		        Bytes: genevaCert[0].Raw,
	        }),
        },
}
secrets = append(secrets, s)
```

pkg/monitor/cluster/certificateexpirationstatuses_test.go
```diff
@@ -0,0 +1,141 @@
...
+							Namespace: "openshift-azure-operator",
+						},
+						Data: map[string][]byte{
+							"tls.crt": pem.EncodeToMemory(&pem.Block{
+								Type:  "CERTIFICATE",
+								Bytes: cert[0].Raw,
+							}),
+						},
+					})
+				}
```
AldoFusterTurpin, line 112: IMHO this is really difficult to follow as it defines an anonymous struct, creates a for over an instantiate slice with that struct + calls a function that defines an inline anonymous function. 
We can split it and reuse the previously defined function.

We can replace the above block code with:
```go
certsInfo := []certInfo{
	{
		"foo12-ingress",
		domain,
	},
	{
		"foo12-apiserver",
		"api." + domain,
	},
}
secretsFromCertsInfo := getSecretsFromCertsInfo(certsInfo, tweakTemplateFn)
secrets = append(secrets, secretsFromCertsInfo...)
```
defining also...
```go
type certInfo struct {
	secretName, certSubject string
}

func getSecretsFromCertsInfo(certsInfo []certInfo, tweakTemplateFn func(*x509.Certificate)) []runtime.Object {
	var secrets []runtime.Object
	for _, sec := range certsInfo {
		_, cert, _ := utiltls.GenerateTestKeyAndCertificate(sec.certSubject, nil, nil, false, false, tweakTemplateFn)
		s := &corev1.Secret{
				Name:      sec.secretName,
				Namespace: "openshift-azure-operator",
			},
			Data: map[string][]byte{
				"tls.crt": pem.EncodeToMemory(&pem.Block{
					Type:  "CERTIFICATE",
					Bytes: cert[0].Raw,
				}),
			},
		}
		secrets = append(secrets, s)
	}
	return secrets
}
```
I think after those changes the test is easier to follow 😃

AldoFusterTurpin (changes requested):
@dem4gus Thank you for including the unit tests. Great work so far! 🚀 

I left a few suggestions as I think a chunk of the unit test is not easy to follow. I just propose to split it a bit using some aux functions.

Thanks a lot 😄

AldoFusterTurpin (changes requested):
@dem4gus This looks nice, we are almost done! 🚀 

I have performed a review and I was able to make the tests almost identical with some changes I have added here as a suggestion, in case you think it is useful.

Thank you a lot.

pkg/monitor/cluster/certificateexpirationstatuses_test.go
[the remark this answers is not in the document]
dem4gus: Those are great suggestions, thanks again for the time you've put into this review.

AldoFusterTurpin (approved):
LGTM! Thank you 🚀

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (mon *Monitor) getCertificate(ctx context.Context, secretName, secretNamespace, secretKey string) (*x509.Certificate, error) {
+	secret, err := mon.cli.CoreV1().Secrets(secretNamespace).Get(ctx, secretName, metav1.GetOptions{})
+	if err != nil {
+		return &x509.Certificate{}, err
+	}
+
+	certBlock, _ := pem.Decode(secret.Data[secretKey])
+	if certBlock == nil {
+		return &x509.Certificate{}, fmt.Errorf(`certificate "%s" not found on secret "%s"`, secretKey, secretName)
```
cadenmarchese, line 59: I'm not sure we can/should be logging certificate data in this way into geneva. Using the name of the certificate should be fine since it's one we manage, but we should remove the `secretKey` from the error message

kippmorris7 (approved):

kimorris27 (dismissed):

cadenmarchese (changes requested):
Per an issue that you found, we will need to adjust this so that the secret names are constructed using the right GUID rather than the Infra ID before merge

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,63 @@
...
+
+	mdsdCert, err := mon.getCertificate(ctx, operator.SecretName, operator.Namespace, genevalogging.GenevaCertName)
+	if err != nil {
+		return err
+	}
+	certs = append(certs, mdsdCert)
+
+	if dns.IsManagedDomain(mon.oc.Properties.ClusterProfile.Domain) {
+		infraID := mon.oc.Properties.InfraID
+		for _, secretName := range []string{infraID + "-ingress", infraID + "-apiserver"} {
```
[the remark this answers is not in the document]
cadenmarchese: Never mind - I confirmed with geneva actions that `mon.oc.ID` is the full resource ID, not the GUID you want.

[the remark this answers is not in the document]
cadenmarchese: A few paths forward:

- Range through all secrets in `openshift-azure-operator`, and pick the two with `-ingress` and `-apiserver` suffixes
- Add a new field in the OpenShiftCluster struct that contains the ID that we need to properly construct the secret name (not immediately sure if this is possible)
- Rename the secrets to use something other than the GUID as prefix (may be more long term)

We may have to go with the first option for now.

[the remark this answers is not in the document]
cadenmarchese: Something like this?
```
	if dns.IsManagedDomain(mon.oc.Properties.ClusterProfile.Domain) {
		managedCertificates, err := mon.cli.CoreV1().Secrets("openshift-azure-operator").List(ctx, metav1.ListOptions{})
		if err != nil {
			return err
		}

		for _, secret := range managedCertificates.Items {
			if strings.HasSuffix(secret.Name, "-apiserver") || strings.HasSuffix(secret.Name, "-ingress") {
				cert, err := mon.getCertificate(ctx, secret.Name, "openshift-azure-operator", "gcscert.pem")
				if err != nil {
					return err
				}
				certs = append(certs, cert)
			}
		}
	}
```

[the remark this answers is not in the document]
dem4gus: I went for a different approach. If it's a managed domain, the ingress certificate secret is referenced by name on the `default` IngressController in `openshift-ingress-operator`. I just got that name and replaced `"-ingress"` with `"-apiserver"` for the other secret.

pkg/monitor/cluster/certificateexpirationstatuses.go
dem4gus: I think there's a misunderstanding due to overloading of the term "key". In this case `secretKey` isn't the TLS key for the certificate; those are never accessed anywhere in this code. It's the "key" as in "key/value pair" of the data on the object, and it comes in through the call to `getCertificate()`. It's either `gcscert.pem` in the case of the Geneva certificate (line 31), or `tls.crt` in the case of the native k8s TLS Secret objects (line 50).

pkg/monitor/cluster/cluster.go
```diff
@@ -33,13 +34,14 @@ type Monitor struct {
...
-	m          metrics.Emitter
-	arocli     aroclient.Interface
+	restconfig  *rest.Config
+	cli         kubernetes.Interface
+	configcli   configclient.Interface
+	maocli      machineclient.Interface
+	mcocli      mcoclient.Interface
+	m           metrics.Emitter
+	arocli      aroclient.Interface
+	operatorcli operatorclient.Interface
```
bennerv, line 44: Instead of adding the operatorclient, can we start leveraging the generic client which can query by API type?  This saves us from creating an interface per API version per cluster.  

Example: https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/imageconfig/image_controller.go#L44C1-L44C19

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,83 @@
...
+			"expirationDate": cert.NotAfter.UTC().Format(time.RFC3339),
+		})
+	}
+	return nil
+}
+
+func (mon *Monitor) getCertificate(ctx context.Context, secretName, secretNamespace, secretKey string) (*x509.Certificate, error) {
+	secret, err := mon.cli.CoreV1().Secrets(secretNamespace).Get(ctx, secretName, metav1.GetOptions{})
+	if err != nil {
+		return &x509.Certificate{}, err
```
bennerv, line 74:
```suggestion
		return nil, err

File: pkg/monitor/cluster/certificateexpirationstatuses.go [hunk 15]
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	certBlock, _ := pem.Decode(secret.Data[secretKey])
+	if certBlock == nil {
+		return &x509.Certificate{}, fmt.Errorf(`certificate "%s" not found on secret "%s"`, secretKey, secretName)
```
bennerv, line 79:
```suggestion
		return nil, fmt.Errorf(`certificate "%s" not found on secret "%s"`, secretKey, secretName)

File: pkg/monitor/cluster/certificateexpirationstatuses.go [hunk 16]
```diff
@@ -0,0 +1,83 @@
...
+const (
+	certificateExpirationMetricName = "certificate.expirationdate"
+	secretMissingMetricName         = "certificate.secretnotfound"
+)
+
+func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error {
+	// report NotAfter dates for Geneva (always), Ingress, and API (on managed domain) certificates
+	var certs []*x509.Certificate
+
+	mdsdCert, err := mon.getCertificate(ctx, operator.SecretName, operator.Namespace, genevalogging.GenevaCertName)
```
bennerv, line 31: nit:
standard for parameters for kubernetes is: 
namespace, name  So swapping around the secret name and the namespace would be nice.

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,83 @@
...
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+const (
+	certificateExpirationMetricName = "certificate.expirationdate"
+	secretMissingMetricName         = "certificate.secretnotfound"
+)
+
+func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error {
+	// report NotAfter dates for Geneva (always), Ingress, and API (on managed domain) certificates
```
bennerv, line 28: nit - make it clearer
```suggestion 
	// report NotAfter dates for Ingress and API (on managed domains), and Geneva (always)

File: pkg/monitor/cluster/certificateexpirationstatuses.go [hunk 18]
```diff
@@ -0,0 +1,83 @@
...
+	var certs []*x509.Certificate
+
+	mdsdCert, err := mon.getCertificate(ctx, operator.SecretName, operator.Namespace, genevalogging.GenevaCertName)
+	if kerrors.IsNotFound(err) {
+		mon.emitGauge(secretMissingMetricName, int64(1), map[string]string{
+			"secretMissing": operator.SecretName,
+		})
+	} else if err != nil {
+		return err
```
bennerv, line 37: nit: 
I think in other examples, we nest this.  it allows a go idiomatic approach to return early if there's an error.  

```suggestion
	if err != nil {
            if !kerrors.IsNotFound(err) {
		return err
	    }
	    mon.emitGauge(secretMissingMetricName, int64(1), map[string]string{
			"secretMissing": operator.SecretName,
	    })
	}

File: pkg/monitor/cluster/certificateexpirationstatuses.go [hunk 19]
```diff
@@ -0,0 +1,83 @@
...
+	} else {
+		certs = append(certs, mdsdCert)
+	}
```
bennerv, line 40: with the above, you don't need an `else` block here.

bennerv (changes requested):
Good work - this will go a long way in us reaching out proactively to customers.

pkg/monitor/cluster/cluster.go
dem4gus: Done. I also refactored the other calls in the monitor to use the generic client.

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		if !kerrors.IsNotFound(err) {
+			return err
+		}
+		mon.emitGauge(secretMissingMetricName, int64(1), map[string]string{
+			"secretMissing": operator.SecretName,
+		})
+	} else {
+		certs = append(certs, mdsdCert)
+	}
```
bennerv, line 42: Ok i'm going back on my original statement.  

We should align this with whatever we do https://github.com/Azure/ARO-RP/pull/2976/files#diff-316e632485d745cf8ccd3fb66e5c6dd71b3f33156040db64751d78e9596ad81aR58-R66

If you want to keep original structure, I'm fine with it.  I thought we could return early, but seeing as we can't now, let's just align the two.

bennerv (dismissed):
One tiny nit, otherwise LGTM.

pkg/monitor/cluster/certificateexpirationstatuses.go
bennerv: insert :old-man-yells-at-ben: emoji :laughing:

dem4gus: I think the chained `else if` is how I've done it other places. Maybe a switch statement or something would look better for multiple ways of handling 🤷.

cadenmarchese:
lgtm, but I think we need some type of manual test plan as well. Some ideas:
- Use https://github.com/Azure/ARO-RP/blob/master/docs/unit-testing-for-monitoring-metrics.md to test. You could get rid of the if `dns.IsManagedDomain` check in your PR and then follow https://docs.openshift.com/container-platform/4.13/security/certificates/replacing-default-ingress-certificate.html with a short-lived certificate to get it to fire and ensure metrics are collected as expected.
- Deploy this to INT or PROD with a modification to your PR's validity check so that it will fire even with new signed certs. If we do it this way, we could set up a test monitor in INT to ensure that we're getting the metrics presented the way we want them.
- Just deploy to INT as-is, ensure that we're getting a green signal from clusters at a minimum.

Thoughts?

mhsreddy:
@dem4gus Great work so far. As per our discussions I think it would really nice if we can export the number of days before certificate expiry as additional value inside the metric along with actual date of certificate expiry. It will be easy to configure Alerting in Geneva Monitor as its very limited when it comes to customization.

dem4gus:
@cadenmarchese 
* Just getting rid of the `dns.IsManagedDomain` check wouldn't be enough, because there isn't any guarantee of the secret names and namespaces following the assumptions I've made for managed certificates.
* I updated the code to also calculate the number of days until the certificate expires (thanks to @mhsreddy), which will help when writing the actual Geneva monitor for alerting. This code is just emitting metrics for all the certificates it cares about and not filtering any of them out, so we will still get metrics for newly created and renewed certificates.
* Just deploying as-is is effectively the same as the above.

cadenmarchese:
Given that we now emit days until expiration, it would be easiest to deploy to INT and confirm that we're getting `certificateExpirationMetric` as expected. We can also delete the secrets in the test cluster to confirm that `secretMissingMetric` fires. I agree there's now no need to modify the PR to fake up a test in INT/prod. Thanks for the changes!

cadenmarchese (commented):
Some minor nits but this is looking good, let's do some final validation in INT and then we can merge. Thanks!

SudoBrendan (approved):
This is looking great! I added some nits. The only other thing I want to bring to everyone's attention is `/pkg/util/pem` and `/pkg/util/pki` - I couldn't see anything that warranted de-dupe, but maybe we can leverage those and/or add to them with this change? I'm not sure...

pkg/monitor/cluster/certificateexpirationstatuses_test.go
```diff
@@ -0,0 +1,235 @@
...
+	clusterID := uuid.DefaultGenerator.Generate()
+
+	for _, tt := range []struct {
+		name            string
+		domain          string
+		certsPresent    []certInfo
+		wantExpirations []map[string]string
+		wantWarning     []map[string]string
+		wantErr         string
+	}{
```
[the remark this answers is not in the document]
dem4gus: The first two tests here are happy-path. There is no checking for how close a certificate is, the metric emits for all certificates. It will report 364 days until expiration for a brand new certificate with a year long life. We can use the Geneva monitors to set a threshold for alerting.

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	} else {
+		certs = append(certs, mdsdCert)
+	}
+
+	if dns.IsManagedDomain(mon.oc.Properties.ClusterProfile.Domain) {
+		ic := &operatorv1.IngressController{}
+		err := mon.clientset.Get(ctx, client.ObjectKey{
+			Namespace: "openshift-ingress-operator",
+			Name:      "default",
```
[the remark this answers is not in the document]
dem4gus: Done

pkg/monitor/cluster/cluster.go
```diff
@@ -41,6 +41,7 @@ type Monitor struct {
 	m          metrics.Emitter
 	arocli     aroclient.Interface
 
+	clientset     client.Client
```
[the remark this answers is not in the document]
dem4gus: I went with `ocpclientset`.

pkg/monitor/cluster/certificateexpirationstatuses_test.go
```diff
@@ -0,0 +1,235 @@
...
+			},
+		},
+	}
+
+	clientset := fake.
+		NewClientBuilder().
+		WithObjects(ingressController).
+		WithObjects(secrets...).
+		Build()
+	mon := &Monitor{
```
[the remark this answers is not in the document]
dem4gus: Done

cadenmarchese (approved):

petrkotas (approved):
@dem4gus I see all is addressed. Nice work!

cadenmarchese:
There was one failing e2e check, but it was the admin portal flake being worked on in https://github.com/Azure/ARO-RP/pull/3168. I've merged this so that work on https://github.com/Azure/ARO-RP/pull/3124 can proceed.

PR #2978, opened 2023-06-22
Increase gateway VMSS disk storage

s-fairchild:
This PR increases the OS disk to 256GB, from the default value of 64GB. The root filesystem is grown to a smaller percentage now because 50% of the disk would mostly go unused.

Fixes
[Incident-399550776 Details - IcM (microsofticm.com)](https://portal.microsofticm.com/imp/v3/incidents/details/399550776/home)

This is needed because the gateway container writes directly to the root of it's filesystem, which is mounted in /var/containers. The gateway log is not currently rotated, so an increase in logging can and will result in a full disk halting the VM.

### Test plan for issue:

Creating a full service RP would be a good way to test this. The E2E test coverage should be sufficient, as it creates a new cluster.

No


SudoBrendan (approved):

tony-schndr (approved):
LGTM!

s-amann (approved):

bennerv (approved):

PR #2981, opened 2023-06-23
fix monitoring stop in case of connection failure

bizz001:

Fixes https://portal.microsofticm.com/imp/v3/incidents/details/399924373/home

Prevents the monitor for crashing when there is an issue that prevents Hive connection.
The sequence that leads to the error is better illustrated by this example: https://go.dev/play/p/M9i87kL8smm

### Test plan for issue:

Give the urgency of the easiest way to test would be to deploy to canary and observe.

No


bizz001:
@dem4gus That is a good question. The errors that we saw that eventually lead to monitor failure look something like this:
`Get "https://xxxxxxxxxxxxxxxx/api?timeout=32s": http2: client connection lost`

This change actually changes the intended behavior of the code. Initially this code was intended to work when no hive connection was present, by doing nothing when there is an error. After a change to the vendored package used for the hive client and due to the very specific go language feature explained in the playground example, it caused the monitor to not work when hive connection is not present, and to stop working if there is any fluke in the connection. 

By returning the error we prevent the monitor from emitting metrics when no hive connection is present (but fix the bug for connection flukes). Perhaps a better alternative to this fix would be to modify this `if` to detect that special nil case that we were missing: https://github.com/Azure/ARO-RP/blob/000562dbd993fa4bae0c4158f3111a0ca78f14fd/pkg/monitor/cluster/hiveregistrationstatus.go#L24 That way we maintain monitoring behavior for specific environments without hive.

s-amann (changes requested):
> modify this if to detect that special nil case that we were missing:
That way we maintain monitoring behavior for specific environments without hive.

IMO we should not force the monitor to stop if we cannot connect to hive due to the metrics around hive not being critical to the operation of the cluster. We should continue to have hive metrics be 'optional' by enhancing the guard statement. Once hive based metrics are on the critical path of an operating cluster we will need to reconsider how we use the hiveclientset in the monitor.

bennerv:
To echo Spencer's comments, I'm +1.  The monitor is technically "best-effort" in a sense.  The latter steps are the most unreliable.  I think it would make sense to move any hive steps to the bottom along with the construction of the hive clientset if possible.

pkg/monitor/cluster/cluster.go
jaitaiwan, line 1: From my understanding, we should be modifying line 94 to make the hiveclientset default to nil so that the check in pkg/monitor/cluster/hiveregistrationstatus.go#L24 will actually work as expected.

Returning nil at this point might not be the go because there are still environments in which we don't have hive, such as some development environments.

jaitaiwan (changes requested):

pkg/monitor/cluster/cluster.go
AldoFusterTurpin: @jaitaiwan Exactly. In that case the monitor will not start at all, we need to make hiveclientset compare to nil as expected without the underlying interface type being *client.Client

Do you mean something like that: https://github.com/Azure/ARO-RP/pull/2984 , right?

bizz001:
> hiveclientset will be nil

@AldoFusterTurpin by saying that you are assuming it will be nil. What I am saying is that we should even assume that and discard whatever is returned by a function that gives an error.

AldoFusterTurpin:
> > hiveclientset will be nil
> 
> @AldoFusterTurpin by saying that you are assuming it will be nil. What I am saying is that we should't even assume that and discard whatever is returned by a function that gives an error.

I do not assume anything, the [PR](https://github.com/Azure/ARO-RP/pull/2984) I opened does exactly what I was trying to say: don't use the hiveclientset returned by the function client.New() when this returns an error. That is why I also created Unit Tests to reproduce that panic and to ensure the new function I defined does not provoke a panic when we later use mon.hiveclientset.

AldoFusterTurpin:
@bizz001 let's sync later today to just have one PR and merge efforts. 

IMHO, the simplest thing would be to put ourselves as co-authors on [the other one](https://github.com/Azure/ARO-RP/pull/2984) and close this one as the other contains some unit tests showing the panics/no panics + explanations and a helper function. But we can discuss it later.
😃

PR #2984, opened 2023-06-26
fix monitoring panic when client.New() returns an error and provide unit tests

AldoFusterTurpin:
Co-authored by: @dbizau and @gvanderpotte

### Which issue does this PR address:

Fixes https://portal.microsofticm.com/imp/v3/incidents/details/399924373/home

Prevents the monitor from crashing when client.New() returns an error, without changing the current behavior of the monitor to not have any unexpected impact in prod.

General showcase of the nil interface: 
https://go.dev/play/p/W63APGm-Aik

### Test plan for issue:

Created unit tests to show the panic that we had in production + tests for the fix showing it does not panic anymore.
The unit tests also serve as documentation for this type of issue. 

Unit Tests are provided with explanations and I am finishing an RCA that I will share soon.

### Why we had the error before the PR

Screenshot 2023-06-26 at 09 19 16

### DIFF
--- a/pkg/monitor/cluster/cluster.go
+++ b/pkg/monitor/cluster/cluster.go
@@ -91,14 +91,9 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
 		return nil, err
 	}
 
-	var hiveclientset client.Client
-	if hiveRestConfig != nil {
-		var err error
-		hiveclientset, err = client.New(hiveRestConfig, client.Options{})
-		if err != nil {
-			// TODO(hive): Update to fail once we have Hive everywhere in prod and dev
-			log.Error(err)
-		}
+	hiveclientset, err := getHiveClientSet(hiveRestConfig)
+	if err != nil {
+		log.Error(err)
 	}
 
 	return &Monitor{
@@ -117,6 +112,18 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
 		m:             m,
 		hiveclientset: hiveclientset,
 	}, nil
+}
+
+func getHiveClientSet(hiveRestConfig *rest.Config) (client.Client, error) {
+	if hiveRestConfig == nil {
+		return nil, nil
+	}
+
+	hiveclientset, err := client.New(hiveRestConfig, client.Options{})
+	if err != nil {
+		return nil, err
+	}
+	return hiveclientset, nil
 }
 
 // Monitor checks the API server health of a cluster
--- a/pkg/monitor/cluster/hiveregistrationstatus_panics_test.go
+++ b/pkg/monitor/cluster/hiveregistrationstatus_panics_test.go
@@ -1,0 +1,148 @@
+package cluster
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	utillog "github.com/Azure/ARO-RP/pkg/util/log"
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+// This test file serves as documentation about an interesting behavior of Go,
+// an error we faced in production consisting of
+// evaluating mon.hiveclientset == nil when the interface is initialised from a function call
+// that returned an error and later using the interface and geting a panic.
+
+// TestEmitHiveRegistrationPanics shows that we get a panic when we can not connect to AKS due to us using a interface
+// (hiveclientset) we obtained doing client.Get() which returned an error (but we did not care about the error).
+// The problem is that hiveclientset is an interface with type *client.Client and value nil and hiveclientset == nil evaluates to false
+// and later mon.hiveclientset.Get() panics.
+func TestEmitHiveRegistrationPanics(t *testing.T) {
+	// GIVEN a hiveclientset interface declared which has interface type: nil, interface value: nil
+	var hiveclientset client.Client
+	fmt.Printf("hiveclientset: interface type is %T, interface value is: %v\n", hiveclientset, hiveclientset)
+	fmt.Printf("hiveclientset == nil ? %v\n\n", hiveclientset == nil)
+
+	// AND GIVEN the same hiveclientset assigned the result of a call to client.New() which returned non nil error,
+	// here the interface type gets changed from nil to *client.Client
+	// making future comparison hiveclientset == nil evaulate to FALSE (counter intuitive) and at the same time
+	// mon.hiveclientset.Get panicking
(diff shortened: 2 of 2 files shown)


pkg/monitor/cluster/cluster.go
```diff
@@ -119,6 +114,18 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
...
 }
 
+func getHiveClientSet(hiveRestConfig *rest.Config) (client.Client, error) {
+	if hiveRestConfig == nil {
+		return nil, nil
+	}
+
+	hiveclientset, err := client.New(hiveRestConfig, client.Options{})
+	if err != nil {
+		return nil, err
```
AldoFusterTurpin, line 124: This 
```go
return nil, err
```
is the key to avoiding future panics as the comparison hiveclientset == nil will be evaluated to true (check the unit tests)

pkg/monitor/cluster/hiveregistrationstatus_panics_test.go
```diff
@@ -0,0 +1,151 @@
...
+	// interface type: *client.Client, interface value: nil
+	hiveclientset, err := client.New(&rest.Config{}, client.Options{})
+	fmt.Printf("hiveclientset: interface type is %T, interface value is: %v\n", hiveclientset, hiveclientset)
+	fmt.Printf("hiveclientset == nil ? %v\n\n", hiveclientset == nil)
+	if err == nil {
+		t.Fatalf("this test wants client.New() to return an error to mimic the behavior in production")
+	}
+
+	// Below line modifies the interface value and type, avoiding
+	// hiveclientset == nil evaluating to false which .
```
s-amann, line 88: nit: this sentence doesn't finish. I think we can end on `false.`

pkg/monitor/cluster/hiveregistrationstatus_panics_test.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	_ = mon.emitHiveRegistrationStatus(context.Background()) // does not panic
+}
+
+// TestEmitHiveRegistrationDoesNotPanicWhenWeAssignLiteralNilUsingAuxFunction shows the exact same
+// behavior as TestEmitHiveRegistrationDoesNotPanicWhenWeAssignLiteralNil but encapsulating the
+// initialisation of hiveclientset in a function. The conclusion is the same, assign literal nil
+// to the interface when we get an error: the difference is that in this test this assignment is
+// coming from a function the prod code uses (so it is a good thinkg) while the previous tests was done inline in the test.
```
s-amann, line 116: nit: typo `thinkg` -> `thing`

pkg/monitor/cluster/hiveregistrationstatus_panics_test.go
```diff
@@ -0,0 +1,151 @@
...
+	fmt.Printf("mon.hiveclientset == nil ? %v\n", mon.hiveclientset == nil)
+
+	_ = mon.emitHiveRegistrationStatus(context.Background())
+}
+
+func NewClient() (client.Client, error) {
+	hiveclientset, err := client.New(nil, client.Options{})
+	if err != nil {
+		// The below hardcoded nil returned is key as it returns an interface
+		// with value and type equal to nil, we can later peform hiveclientset == nil obtaining true.
```
s-amann: nit: typo `peform` -> `perform`

s-amann (commented):
lgtm, just a few typos in comments

s-amann (approved):

pkg/monitor/cluster/cluster.go
```diff
@@ -91,14 +91,9 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
 		return nil, err
 	}
 
-	var hiveclientset client.Client
```
jaitaiwan, line 94: I think just setting this to nil would have covered the edge case.

jaitaiwan (approved):
Small note but LGTM

pkg/monitor/cluster/cluster.go
```diff
@@ -91,14 +91,9 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
...
-	if hiveRestConfig != nil {
-		var err error
-		hiveclientset, err = client.New(hiveRestConfig, client.Options{})
-		if err != nil {
-			// TODO(hive): Update to fail once we have Hive everywhere in prod and dev
-			log.Error(err)
-		}
```
yjst2012, line 101: nit, an alternative way is simply removing the outer "hiveRestConfig != nil" check, only keep the inner thing, like all other modules do:

	hiveclientset, err := client.New(hiveRestConfig, client.Options{})
	if err != nil {
		// TODO(hive): Update to fail once we have Hive everywhere in prod and dev
		log.Error(err)
	}


if have to keep "hiveRestConfig != nil" check, it's better to move it to the beginning of the func as other modules before this will crash first in that case, so the check here does not help.

yjst2012 (approved):
LGTM

pkg/monitor/cluster/cluster.go
AldoFusterTurpin: @jaitaiwan yes you are right. This is explained in the RCA. 😃 

The problem is that pkg/monitor/cluster/cluster.go.NewMonitor has implicit dependencies ("k8s.io/client-go/kubernetes", "github.com/openshift/client-go/machine/clientset/versioned", etc.). If we used dependency injection and inversion we would be able to mock the dependencies and test NewMonitor() but as it is not easy to do, creating an aux function enables us to at least test that new function.

pkg/monitor/cluster/cluster.go
AldoFusterTurpin: @yjst2012 Do you mean just removing 
```go
if hiveRestConfig == nil {
```
from pkg/monitor/cluster/cluster.go.NewMonitor() without any other change?

yjst2012: yep, something like this will do the trick I believe, since the problem is introduced by "var hiveclientset hiveclient.Interface".

-       var hiveclientset hiveclient.Interface
-       if hiveRestConfig != nil {
-               var err error
-               hiveclientset, err = hiveclient.NewForConfig(hiveRestConfig)
-               if err != nil {
-                       // TODO(hive): Update to fail once we have Hive everywhere in prod and dev
-                       log.Error(err)
-               }
+       hiveclientset, err := hiveclient.NewForConfig(hiveRestConfig)
+       if err != nil {
+               // TODO(hive): Update to fail once we have Hive everywhere in prod and dev
+               log.Error(err)
        }

yjst2012: Aha, I know the problem, cos I'm looking at a relatively old version that I'm working on. it still calling the old api 

`func NewForConfig(c *rest.Config) (*Clientset, error) {
`

which do not have the problem introduced :joy:

PR #2988, opened 2023-06-26
Add gateway log rotatation

s-fairchild:

Fixes

Improved fix for https://portal.microsofticm.com/imp/v3/incidents/details/399550776/home

The gateway container log is mounted in `/var/log/aro-gateway` to allow `logrotate` access to the file. Logs are rotated based on size, not time. Logs are rotated when they reach 20G in size, and a maximum of 100G total. This is to allow enough space for other logs and system services in /var.

`copytruncate` option prevents the log file from being recreated after being deleted, which would create a condition where the log watcher/shipper thinks they are new.`copytruncate` truncates the log file, leaving it's modification timestamp the same.

### Test plan for issue:

E2E tests, and I believe a full service RP will be required to actually test.
sshing into the gateway VMSS and testing the logrotate configuration manually.

No


facchettos (changes requested):

pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -14,8 +14,8 @@ yum -y -x WALinuxAgent -x WALinuxAgent-udev update --allowerasing
 echo "extending partition table"
 # Linux block devices are inconsistently named
 # it's difficult to tie the lvm pv to the physical disk using /dev/disk files, which is why lvs is used here
-physicalDisk="$(lvs -o devices -a | head -n2 | tail -n1 | cut -d ' ' -f 3 | cut -d \( -f 1 | tr -d '[:digit:]')"
-growpart "$physicalDisk" 2
+physical_disk="$(lvs -o devices -a | head -n2 | tail -n1 | cut -d ' ' -f 3 | cut -d \( -f 1 | tr -d '[:digit:]')"
+growpart "$physical_disk" 2
```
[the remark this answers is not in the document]
s-fairchild: No, just changed for consistency. I can remove if needed.

pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -67,6 +68,19 @@ include /etc/logrotate.d
...
 }
+
+# Maximum log directory size is 100G with this configuration
+# Setting limit to 100G to allow space for other logging services
+# copytruncate is a critical option used to prevent logs from being shipped twice
+${gateway_logdir} {
+    size 20G
+    rotate 5
+    create 0600 root root
+    copytrunate
```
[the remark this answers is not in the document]
s-fairchild: Yes! Thank you

pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -33,7 +33,8 @@ for attempt in {1..5}; do
 done
 
 echo "configuring logrotate"
-cat >/etc/logrotate.conf <<'EOF'
+declare -r gateway_logdir='/var/log/aro-gateway'
```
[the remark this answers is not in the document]
s-fairchild: My thinking is that path is used in 3 different places, and it doesn't change, so a read only variable is an easy way to keep track of this directory if we need to change it in the future.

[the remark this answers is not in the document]
s-fairchild: @facchettos Do you mean add a comment for this?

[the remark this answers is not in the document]
facchettos: both, I was not sure I understood what that did :D

pkg/deploy/generator/scripts/gatewayVMSS.sh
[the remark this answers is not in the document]
facchettos: cool, was just checking

pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -67,6 +72,19 @@ include /etc/logrotate.d
...
+
+# Maximum log directory size is 100G with this configuration
+# Setting limit to 100G to allow space for other logging services
+# copytruncate is a critical option used to prevent logs from being shipped twice
+${gateway_logdir} {
+    size 20G
+    rotate 5
+    create 0600 root root
+    copytruncate
+    nooldir
```
cadenmarchese, line 84: Typo?
```suggestion
    noolddir
```

pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -67,6 +72,19 @@ include /etc/logrotate.d
...
+    missingok
```
cadenmarchese, line 85: What's the justification here? I think it may make sense to log errors in the case of missing log files.

s-amann (dismissed):
@s-fairchild I believe you need to re-run `make generate` and commit/push the results in order for CI to pass.

pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -67,6 +72,19 @@ include /etc/logrotate.d
     create 0600 root utmp
     rotate 1
 }
+
+# Maximum log directory size is 100G with this configuration
+# Setting limit to 100G to allow space for other logging services
+# copytruncate is a critical option used to prevent logs from being shipped twice
+${gateway_logdir} {
+    size 20G
```
cadenmarchese, line 80: How did we decide on 100G?

This is more of an lvm question than a suggestion, but I see that https://github.com/Azure/ARO-RP/pull/2978/files#diff-a7a118c3717408a3425f9209b28c891edad1c68d9b9a67cb41b18545f8ce37e3R21 increased disk size, but also decreased the amount of free space given to the root filesystem (+20%FREE rather than +50%FREE), is that understanding correct?

cadenmarchese (commented):
I went through the logrotate options and things are making sense, just a few comments. Thanks!

pkg/deploy/generator/scripts/gatewayVMSS.sh
s-fairchild: I've been looking for input on what size this should be. Is there a maximum time limit between when a line is written to the log, and when that is read by (MDSD?) If so then maybe a larger size is good, but around 20GB?

Correct, the reason that `/` was reduced to 20% is because it is largely unused by the system. IIRC the root partition is around 45GB in size. /var is being increased to 50% of space, the remainder of the disk. Which should be around 200GB in size. Although I wish I had recorded the numbers I took, with all the partitions on disk I don't remember what is left for / and /var.

pkg/deploy/generator/scripts/gatewayVMSS.sh
s-fairchild: I believe logrotate will fail with an error, and not rotate the rest of the system logs. But I could be wrong it's been a while since I've had to troubleshoot logrotate. It can be finicky, and if it couldn't access the log for rotation I didn't want it to fail.  

> If the log file is missing, go on to the next one without issuing an error message.
The man page looks like it only issues an error and moves on, I'll remove that option.

pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -276,7 +295,8 @@ ExecStart=/usr/bin/docker run \
   -p 443:8443 \
   -v /run/systemd/journal:/run/systemd/journal \
   -v /var/etw:/var/etw:z \
-  $RPIMAGE \
+  -v /ctr.log:${gateway_logdir}:z \
```
kimorris27, line 298: Why is the `:z` needed here?

pkg/deploy/generator/scripts/gatewayVMSS.sh
s-fairchild: Yes, thanks. I should try to copy/paste config options more.

pkg/deploy/generator/scripts/gatewayVMSS.sh
cadenmarchese: I see, so the change in https://github.com/Azure/ARO-RP/pull/2978 is allocating less to `/`, not `/var`, where the logs are. I was misunderstanding, so with that I think we can get away with 100G.

> Is there a maximum time limit between when a line is written to the log, and when that is read by (MDSD?)

I am not sure on this one.

pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -67,6 +72,18 @@ include /etc/logrotate.d
     create 0600 root utmp
     rotate 1
 }
+
+# Maximum log directory size is 100G with this configuration
+# Setting limit to 100G to allow space for other logging services
```
s-amann, line 77: please update this comment to match the size on line 80

s-fairchild: The comment is correct, it's a multiple of `size` and `rotate` options. 5 logs of size 20G will be kept.

cadenmarchese (approved):

PR #2991, opened 2023-06-26
Maintenance Signals

niontive:

Fixes  https://issues.redhat.com/browse/ARO-1933

This PR is needed to allow for ARO SREs to send signals when cluster maintenance is being performed. Those signals will go to resource health check so customers get notified.

Code changes:
1. Add maintenance states to CosmosDB cluster document
2. Add maintenance state admin update task
3. Emit metric via monitoring stack. Metric lists the cluster's maintenance state.

More info can be found here: https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/508199/Maintenance-Signals

### Test plan for issue:

- Unit tests
- Tested in INT


niontive:
Code review:
1. Review design with ben/rogerio
2. Investigate using middleware so you don't have to modify a bunch of ARO-SRE scripts

PR #2992, opened 2023-06-27
HiveNamespace to use docID/clusterID for new Installs

SrinivasAtmakuri:
HiveNamespace currently uses aro-"uuid", this change is an effort to unify UUIDs across cluster doc instead of having multiple, by pointing HiveNamespace to docID so this can be leveraged later.

More Details: https://redhat-internal.slack.com/archives/C02ULBRS68M/p1686806655273309

Fixes

### DIFF
--- a/pkg/hive/manager.go
+++ b/pkg/hive/manager.go
@@ -21,11 +21,10 @@ import (
 	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/util/dynamichelper"
 	utillog "github.com/Azure/ARO-RP/pkg/util/log"
-	"github.com/Azure/ARO-RP/pkg/util/uuid"
 )
 
 type ClusterManager interface {
-	CreateNamespace(ctx context.Context) (*corev1.Namespace, error)
+	CreateNamespace(ctx context.Context, docID string) (*corev1.Namespace, error)
 
 	// CreateOrUpdate reconciles the ClusterDocument and related secrets for an
 	// existing cluster. This may adopt the cluster (Create) or amend the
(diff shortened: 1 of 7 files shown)


pkg/hive/manager.go
```diff
@@ -107,11 +106,11 @@ func NewFromConfig(log *logrus.Entry, _env env.Core, restConfig *rest.Config) (C
 	}, nil
 }
 
-func (hr *clusterManager) CreateNamespace(ctx context.Context) (*corev1.Namespace, error) {
+func (hr *clusterManager) CreateNamespace(ctx context.Context, docID string) (*corev1.Namespace, error) {
 	var namespaceName string
 	var namespace *corev1.Namespace
 	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		namespaceName = "aro-" + uuid.DefaultGenerator.Generate()
+		namespaceName = docID
```
ArrisLee, line 113: does that mean we will not using `aro` prefix as the log searching rule for hive namespaces?

SrinivasAtmakuri: I was 50-50 between using just uuid or aro-uuid and reached out to Amber for inputs, I will make the change to add prefix aro-

pkg/hive/manager.go
```diff
@@ -107,11 +106,11 @@ func NewFromConfig(log *logrus.Entry, _env env.Core, restConfig *rest.Config) (C
 	}, nil
 }
 
-func (hr *clusterManager) CreateNamespace(ctx context.Context) (*corev1.Namespace, error) {
+func (hr *clusterManager) CreateNamespace(ctx context.Context, docID string) (*corev1.Namespace, error) {
```
ArrisLee, line 109: one question here, do we need to modify the place where this function is called by adding an extra param? i.e.,

https://github.com/Azure/ARO-RP/blob/f063e9baaf9286781da38276859b81b08a702562/pkg/cluster/hive.go#L20

facchettos:
CI isn't passing

jaitaiwan (approved):
LGTM!

yjst2012 (approved):
LGTM

s-amann (approved):

petrkotas:
putting hold, we still are looking for the check about hive namespace deletion

SrinivasAtmakuri:
> Hi @SrinivasAtmakuri, thank you for this PR. It really makes sense to pair the hive namespace with the document ID.
> 
> I have one concern that needs to be checked. That is do we delete the hive namespace properly when cluster fails? If we do not delete properly, we might run into the problems. Would you please check?

We are creating a new doc UUID for each run, as per https://github.com/Azure/ARO-RP/blob/aaca34ee0b702e3b5c4f2144fa163440ea02ae41/pkg/frontend/openshiftcluster_putorpatch.go#L75 so if a deletion fails and on retry a new UUID will be generated and not cause an issue. Thank you.

jaitaiwan:
Sorry if I missed it, but is there a test for determining when a cluster is a new install or an old one?

SrinivasAtmakuri:
> Sorry if I missed it, but is there a test for determining when a cluster is a new install or an old one?

No, there is no such test, as in ARO we don't have an option of retry, user needs to delete a failed cluster and create a fresh one again.

mociarain (approved):

PR #2994, opened 2023-06-28
Hack: make eastus2euap non-zonal due to Azure capacity issues

bennerv:

Forces eastus2euap to non-zonal so our e2es successfully run against it.  

Azure currently has capacity issues in eastus2euap AZ1, so force non-zonal so we don't provision across zones.


tsatam (approved):
LGTM

tsatam:
Will this change be mirrored to the installer (https://github.com/openshift/ARO-Installer/blob/19fed938beae727d0ff6b8a0ab599f5ad21a9847/pkg/installer/generateconfig.go#L108) as well?

PR #2995, opened 2023-06-28
Fix resolv.conf during Azure host servicing

s-fairchild:

Fixes

https://issues.redhat.com/browse/ARO-3474
https://issues.redhat.com/browse/ARO-3468

If the network interfaces are brought up/down without restarting NetworkManager, our resolv.conf will be overwritten by NetworkManager. This can occur during [Azure Host Servicing](https://learn.microsoft.com/en-us/azure/developer/intro/hosting-apps-on-azure) events.

See [NetworkManager-dispatcher](https://networkmanager.dev/docs/api/latest/NetworkManager-dispatcher.html) documentation for more information, such as available actions.  

The second commit cleans up the dnsmasq controller package by embedding all Go templates, and moving them into their own files within `pkg/operators/controllers/dnsmasq/scripts`.

**Operator flag has been set to true**

### Test plan for issue:

Tested on a dev cluster with, I have created a machineconfig locally containing the changes and applied them to both master and worker nodes successfully.
The restart script works and has been tested with simulated Azure host servicing in this [jira ticket](https://issues.redhat.com/browse/ARO-3440).

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -79,6 +79,7 @@ func DefaultOperatorFlags() OperatorFlags {
 		"aro.banner.enabled":                       flagFalse,
 		"aro.checker.enabled":                      flagTrue,
 		"aro.dnsmasq.enabled":                      flagTrue,
+		"aro.restartdnsmasq.enabled":               flagFalse,
 		"aro.genevalogging.enabled":                flagTrue,
 		"aro.imageconfig.enabled":                  flagTrue,
 		"aro.ingress.enabled":                      flagTrue,
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -79,7 +79,7 @@ func DefaultOperatorFlags() OperatorFlags {
 		"aro.banner.enabled":                       flagFalse,
 		"aro.checker.enabled":                      flagTrue,
 		"aro.dnsmasq.enabled":                      flagTrue,
-		"aro.restartdnsmasq.enabled":               flagFalse,
+		"aro.restartdnsmasq.enabled":               flagTrue,
 		"aro.genevalogging.enabled":                flagTrue,
 		"aro.imageconfig.enabled":                  flagTrue,
 		"aro.ingress.enabled":                      flagTrue,
--- a/pkg/operator/controllers/dnsmasq/dnsmasq.go
+++ b/pkg/operator/controllers/dnsmasq/dnsmasq.go
@@ -5,6 +5,7 @@ package dnsmasq
 
 import (
 	"bytes"
+	_ "embed"
 	"encoding/json"
 	"fmt"
 	"text/template"
@@ -17,105 +18,17 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-var t = template.Must(template.New("").Parse(`
-
-{{ define "dnsmasq.conf" }}
-resolv-file=/etc/resolv.conf.dnsmasq
-strict-order
-address=/api.{{ .ClusterDomain }}/{{ .APIIntIP }}
-address=/api-int.{{ .ClusterDomain }}/{{ .APIIntIP }}
-address=/.apps.{{ .ClusterDomain }}/{{ .IngressIP }}
-{{- range $GatewayDomain := .GatewayDomains }}
-address=/{{ $GatewayDomain }}/{{ $.GatewayPrivateEndpointIP }}
-{{- end }}
-user=dnsmasq
-group=dnsmasq
-no-hosts
-cache-size=0
-{{ end }}
-
-{{ define "dnsmasq.service" }}
-[Unit]
-Description=DNS caching server.
-After=network-online.target
-Before=bootkube.service
-
-[Service]
-# ExecStartPre will create a copy of the customer current resolv.conf file and make it upstream DNS.
-# This file is a product of user DNS settings on the VNET. We will replace this file to point to
-# dnsmasq instance on the node. dnsmasq will inject certain dns records we need and forward rest of the queries to
-# resolv.conf.dnsmasq upstream customer dns.
-ExecStartPre=/bin/bash /usr/local/bin/aro-dnsmasq-pre.sh
-ExecStart=/usr/sbin/dnsmasq -k
-ExecStopPost=/bin/bash -c '/bin/mv /etc/resolv.conf.dnsmasq /etc/resolv.conf; /usr/sbin/restorecon /etc/resolv.conf'
-Restart=always
-
-[Install]
-WantedBy=multi-user.target
-{{ end }}
-
-{{ define "aro-dnsmasq-pre.sh" }}
-#!/bin/bash
-set -euo pipefail
-
-# This bash script is a part of the ARO DnsMasq configuration
-# It's deployed as part of the 99-aro-dns-* machine config
-# See https://github.com/Azure/ARO-RP
-
-# This file can be rerun and the effect is idempotent, output might change if the DHCP configuration changes
-
-TMPSELFRESOLV=$(mktemp)
-TMPNETRESOLV=$(mktemp)
-
-echo "# Generated for dnsmasq.service - should point to self" > $TMPSELFRESOLV
-echo "# Generated for dnsmasq.service - should contain DHCP configured DNS" > $TMPNETRESOLV
-
-if nmcli device show br-ex; then
-    echo "OVN mode - br-ex device exists"
-    #getting DNS search strings
-    SEARCH_RAW=$(nmcli --get IP4.DOMAIN device show br-ex)
-    #getting DNS servers
-    NAMESERVER_RAW=$(nmcli --get IP4.DNS device show br-ex | tr -s " | " "\n")
-    LOCAL_IPS_RAW=$(nmcli --get IP4.ADDRESS device show br-ex)
-else
-    NETDEV=$(nmcli --get device connection show --active | head -n 1) #there should be only one active device
-    echo "OVS SDN mode - br-ex not found, using device $NETDEV"
-    SEARCH_RAW=$(nmcli --get IP4.DOMAIN device show $NETDEV)
-    NAMESERVER_RAW=$(nmcli --get IP4.DNS device show $NETDEV | tr -s " | " "\n")
-    LOCAL_IPS_RAW=$(nmcli --get IP4.ADDRESS device show $NETDEV)
-fi
-
-#search line
-echo "search $SEARCH_RAW" | tr '\n' ' ' >> $TMPNETRESOLV
-echo "" >> $TMPNETRESOLV
-echo "search $SEARCH_RAW" | tr '\n' ' ' >> $TMPSELFRESOLV
-echo "" >> $TMPSELFRESOLV
-
-#nameservers as separate lines
-echo "$NAMESERVER_RAW" | while read -r line
-do
-    echo "nameserver $line" >> $TMPNETRESOLV
-done
-# device IPs are returned in address/mask format
-echo "$LOCAL_IPS_RAW" | while read -r line
-do
-    echo "nameserver $line" | cut -d'/' -f 1 >> $TMPSELFRESOLV
-done
-
-# done, copying files to destination locations and cleaning up
-/bin/cp $TMPNETRESOLV /etc/resolv.conf.dnsmasq
-chmod 0744 /etc/resolv.conf.dnsmasq
-/bin/cp $TMPSELFRESOLV /etc/resolv.conf
-/usr/sbin/restorecon /etc/resolv.conf
-/bin/rm $TMPNETRESOLV
-/bin/rm $TMPSELFRESOLV
-{{ end }}
-`))
+const (
+	configFileName    = "dnsmasq.conf"
+	unitFileName      = "dnsmasq.service"
+	prescriptFileName = "aro-dnsmasq-pre.sh"
+)
 
 func config(clusterDomain, apiIntIP, ingressIP string, gatewayDomains []string, gatewayPrivateEndpointIP string) ([]byte, error) {
+	t := template.Must(template.New(configFileName).Parse(preScriptFile))
 	buf := &bytes.Buffer{}
 
-	err := t.ExecuteTemplate(buf, "dnsmasq.conf", &struct {
+	err := t.ExecuteTemplate(buf, configFileName, &struct {
 		ClusterDomain            string
 		APIIntIP                 string
 		IngressIP                string
@@ -136,9 +49,10 @@ func config(clusterDomain, apiIntIP, ingressIP string, gatewayDomains []string, 
 }
 
 func service() (string, error) {
+	t := template.Must(template.New(unitFileName).Parse(unitFile))
 	buf := &bytes.Buffer{}
 
-	err := t.ExecuteTemplate(buf, "dnsmasq.service", nil)
+	err := t.ExecuteTemplate(buf, unitFileName, nil)
 	if err != nil {
 		return "", err
 	}
@@ -147,9 +61,10 @@ func service() (string, error) {
 }
 
 func startpre() ([]byte, error) {
+	t := template.Must(template.New(prescriptFileName).Parse(configFile))
 	buf := &bytes.Buffer{}
 
-	err := t.ExecuteTemplate(buf, "aro-dnsmasq-pre.sh", nil)
+	err := t.ExecuteTemplate(buf, prescriptFileName, nil)
 	if err != nil {
 		return nil, err
 	}
@@ -183,7 +98,7 @@ func ignition2Config(clusterDomain, apiIntIP, ingressIP string, gatewayDomains [
 					Node: ign2types.Node{
 						Filesystem: "root",
 						Overwrite:  to.BoolPtr(true),
-						Path:       "/etc/dnsmasq.conf",
+						Path:       "/etc/" + configFileName,
 						User: &ign2types.NodeUser{
 							Name: "root",
 						},
@@ -199,7 +114,7 @@ func ignition2Config(clusterDomain, apiIntIP, ingressIP string, gatewayDomains [
 					Node: ign2types.Node{
 						Filesystem: "root",
 						Overwrite:  to.BoolPtr(true),
-						Path:       "/usr/local/bin/aro-dnsmasq-pre.sh",
+						Path:       "/usr/local/bin/" + prescriptFileName,
 						User: &ign2types.NodeUser{
 							Name: "root",
 						},
@@ -218,34 +133,19 @@ func ignition2Config(clusterDomain, apiIntIP, ingressIP string, gatewayDomains [
 				{
 					Contents: service,
 					Enabled:  to.BoolPtr(true),
-					Name:     "dnsmasq.service",
+					Name:     unitFileName,
 				},
 			},
 		},
 	}
 
 	if restartDnsmasq {
-		restartDnsmasqRaw, err := nmDispatcherRestartDnsmasq()
+		restartDnsmasqScript, err := nmDispatcherRestartDnsmasq()
 		if err != nil {
 			return nil, err
 		}
 
-		ign.Storage.Files = append(ign.Storage.Files, ign2types.File{
-			Node: ign2types.Node{
-				Filesystem: "root",
-				Overwrite:  to.BoolPtr(true),
-				Path:       "/etc/NetworkManager/dispatcher.d/99-dnsmasq-restart",
-				User: &ign2types.NodeUser{
-					Name: "root",
-				},
-			},
-			FileEmbedded1: ign2types.FileEmbedded1{
-				Contents: ign2types.FileContents{
-					Source: dataurl.EncodeBytes(restartDnsmasqRaw),
-				},
-				Mode: to.IntPtr(0744),
-			},
-		})
+		ign.Storage.Files = append(ign.Storage.Files, restartScriptIgnFile(restartDnsmasqScript))
 	}
 
 	return ign, nil
--- a/pkg/operator/controllers/dnsmasq/dnsmasq.go
+++ b/pkg/operator/controllers/dnsmasq/dnsmasq.go
@@ -157,7 +157,7 @@ func startpre() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
-func ignition2Config(clusterDomain, apiIntIP, ingressIP string, gatewayDomains []string, gatewayPrivateEndpointIP string) (*ign2types.Config, error) {
+func ignition2Config(clusterDomain, apiIntIP, ingressIP string, gatewayDomains []string, gatewayPrivateEndpointIP string, restartDnsmasq bool) (*ign2types.Config, error) {
 	service, err := service()
 	if err != nil {
 		return nil, err
@@ -173,7 +173,7 @@ func ignition2Config(clusterDomain, apiIntIP, ingressIP string, gatewayDomains [
 		return nil, err
 	}
 
-	return &ign2types.Config{
+	ign := &ign2types.Config{
 		Ignition: ign2types.Ignition{
 			Version: ign2types.MaxVersion.String(),
 		},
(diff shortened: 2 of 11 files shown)


vendor/github.com/openshift/installer/pkg/aro/dnsmasq/dnsmasq.go
```diff
@@ -108,6 +108,22 @@ chmod 0744 /etc/resolv.conf.dnsmasq
 /bin/rm $TMPNETRESOLV
 /bin/rm $TMPSELFRESOLV
 {{ end }}
+
```
tsatam, line 111: As of #2932, this file has been duplicated directly into the ARO-RP codebase rather than referenced from the vendored dependency.

If I'm understanding this correctly, these changes will need to be made to both:
- ARO-RP: https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/dnsmasq/dnsmasq.go
- installer-aro: https://github.com/jewzaam/installer-aro/blob/release-4.11-azure/pkg/aro/dnsmasq/dnsmasq.go

as this specific vendored file is no longer referenced by our operator reconciler.

s-fairchild: @tsatam Thank you for bringing this to my attention!

s-fairchild:
> Is it possible to emit a log when dnsmasq gets restarted by systemctl? Ultimately ending up in Kusto for observability

@swiencki Yes, I can add that to the script. I'll verify, but `logger` should be installed and we can emit a system log there.
What log level do we want? and any specific information, we can include [anything available here](https://networkmanager.dev/docs/api/latest/NetworkManager-dispatcher.html) in the logs.

swiencki:
> > Is it possible to emit a log when dnsmasq gets restarted by systemctl? Ultimately ending up in Kusto for observability
> 
> @swiencki Yes, I can add that to the script. I'll verify, but `logger` should be installed and we can emit a system log there. What log level do we want? and any specific information, we can include [anything available here](https://networkmanager.dev/docs/api/latest/NetworkManager-dispatcher.html) in the logs.

Oh I  get to choose!

If we can emit logs from [NetworkManager-Dispatcher](https://networkmanager.dev/docs/api/latest/NetworkManager-dispatcher.html) then;

up - The interface has been activated.
down - The interface has been deactivated.
connectivity-change - The network connectivity state has changed (no connectivity, went online, etc).

Maybe reapply, not sure what that is?

env vars we might be able to craft a print similar to;

$TIMESTAMP - Current state: CONNECTIVITY_STATE: $NM_DISPATCHER_ACTION on $DEVICE_IFACE:DEVICE_IP_IFACE (IP4_NUM_ADDRESSES/IP4_GATEWAY)

The following outputs might also be useful:
IP4_NUM_ROUTES | The variable contains the number of IPv4 routes the script may expect.
IP4_NAMESERVERS | The variable contains a space-separated list of the DNS servers.
IP4_DOMAINS | The variable contains a space-separated list of the search domains.
DHCP4_<dhcp-option-name> | If the connection used DHCP for address configuration, the received DHCP configuration is passed in the environment using standard DHCP option names, prefixed with "DHCP4_", like "DHCP4_HOST_NAME=foobar".

cat of; 
dnsmasq.conf
/etc/resolv.conf
/etc/resolv.conf.dnsmasq

All of the above might be a bit verbose but could provide much needed insight..

ventifus:
@swiencki NM normally logs interface up/down events and also when it gets a new DHCP lease:

```
master-0 NetworkManager[1359]: <info>  [1687362059.7572] device (eth0): carrier: link connected
master-0 NetworkManager[1359]: <info>  [1687362059.7577] manager: (eth0): new Ethernet device (/org/freedesktop/NetworkManager/Devices/3)
master-0 NetworkManager[1359]: <info>  [1687362059.7584] manager: (eth0): assume: will attempt to assume matching connection 'Wired Connection' (a9d5d6a9-ba0d-407c-b174-b5e6216cb8b5) (indicated)
master-0 NetworkManager[1359]: <info>  [1687362059.7584] device (eth0): state change: unmanaged -> unavailable (reason 'connection-assumed', sys-iface-state: 'assume')
master-0 NetworkManager[1359]: <info>  [1687362059.7591] device (eth0): state change: unavailable -> disconnected (reason 'connection-assumed', sys-iface-state: 'assume')
master-0 NetworkManager[1359]: <info>  [1687362059.7599] device (eth0): Activation: starting connection 'Wired Connection' (a9d5d6a9-ba0d-407c-b174-b5e6216cb8b5)
master-0 NetworkManager[1359]: <info>  [1687362059.7646] device (eth0): state change: disconnected -> prepare (reason 'none', sys-iface-state: 'assume')
master-0 NetworkManager[1359]: <info>  [1687362059.7650] device (eth0): state change: prepare -> config (reason 'none', sys-iface-state: 'assume')
master-0 NetworkManager[1359]: <info>  [1687362059.7730] device (eth0): state change: config -> ip-config (reason 'none', sys-iface-state: 'assume')
master-0 NetworkManager[1359]: <info>  [1687362059.7735] dhcp4 (eth0): activation: beginning transaction (timeout in 90 seconds)
master-0 NetworkManager[1359]: <info>  [1687362059.7837] dhcp4 (eth0): state changed new lease, address=10.0.0.7
master-0 NetworkManager[1359]: <info>  [1687362059.7840] policy: set 'Wired Connection' (eth0) as default for IPv4 routing and DNS
master-0 NetworkManager[1359]: <info>  [1687362059.7901] device (eth0): state change: ip-config -> ip-check (reason 'none', sys-iface-state: 'assume')
master-0 NetworkManager[1359]: <info>  [1687362059.8752] device (eth0): state change: ip-check -> secondaries (reason 'none', sys-iface-state: 'assume')
master-0 NetworkManager[1359]: <info>  [1687362059.8754] device (eth0): state change: secondaries -> activated (reason 'none', sys-iface-state: 'assume')
master-0 NetworkManager[1359]: <info>  [1687362059.8760] device (eth0): Activation: successful, device activated.
```

swiencki:
> @swiencki NM normally logs interface up/down events and also when it gets a new DHCP lease:

Beauty, +1 on limiting the amount of duplicate logging

s-fairchild:
@swiencki @ventifus I've pushed some changes to the restart script that improve logging information. Here's some example output

```bash
sh-4.4# ./20-dnsmasq-restart eth0 up
+ interface=eth0
+ action=up
+ '[' eth0 == 'eth*' ']'
+ '[' up == up ']'
+ log 'up happened on eth0, connection state is now '
+ logger -i ./20-dnsmasq-restart -t '99-DNSMASQ-RESTART SCRIPT' 'up happened on eth0, connection state is now '
+ log 'restarting dnsmasq now'
+ logger -i ./20-dnsmasq-restart -t '99-DNSMASQ-RESTART SCRIPT' 'restarting dnsmasq now'
+ systemctl restart dnsmasq
+ log 'dnsmasq successfully restarted'
+ logger -i ./20-dnsmasq-restart -t '99-DNSMASQ-RESTART SCRIPT' 'dnsmasq successfully restarted'
++ cat /etc/resolv.conf
+ log '/etc/resolv.conf output
    # Generated for dnsmasq.service - should point to self
search g323k5hfzoau3c2bwxrpslve2f.bx.internal.cloudapp.net 
nameserver 10.0.18.8'
+ logger -i ./20-dnsmasq-restart -t '99-DNSMASQ-RESTART SCRIPT' '/etc/resolv.conf output
    # Generated for dnsmasq.service - should point to self
search g323k5hfzoau3c2bwxrpslve2f.bx.internal.cloudapp.net 
nameserver 10.0.18.8'
++ cat /etc/resolv.conf.dnsmasq
+ log '/etc/resolv.conf.dnsmasq output
    # Generated for dnsmasq.service - should contain DHCP configured DNS
search g323k5hfzoau3c2bwxrpslve2f.bx.internal.cloudapp.net 
nameserver 168.63.129.16'
+ logger -i ./20-dnsmasq-restart -t '99-DNSMASQ-RESTART SCRIPT' '/etc/resolv.conf.dnsmasq output
    # Generated for dnsmasq.service - should contain DHCP configured DNS
search g323k5hfzoau3c2bwxrpslve2f.bx.internal.cloudapp.net 
nameserver 168.63.129.16'
+ exit 0
```
An example of an action not causing a restart:
```bash
sh-4.4# ip link set eth0 down; ip link set eth0 up
Jun 28 20:39:51 steven-cluster-bh8mr-master-0 nm-dispatcher[219385]: + interface=eth0
Jun 28 20:39:51 steven-cluster-bh8mr-master-0 nm-dispatcher[219385]: + action=dhcp4-change
Jun 28 20:39:51 steven-cluster-bh8mr-master-0 nm-dispatcher[219385]: + '[' eth0 == 'eth*' ']'
Jun 28 20:39:51 steven-cluster-bh8mr-master-0 nm-dispatcher[219385]: + '[' dhcp4-change == up ']'
Jun 28 20:39:51 steven-cluster-bh8mr-master-0 nm-dispatcher[219385]: + '[' dhcp4-change == down ']'
Jun 28 20:39:51 steven-cluster-bh8mr-master-0 nm-dispatcher[219385]: + exit 0
```

yjst2012:
LGTM, but I would like another codeowner to approve, re-run e2e to see if it is a flake, thanks for putting this up!

petrkotas (approved):
Nice! I like the explanation which made the review easier. LGTM

UlrichSchlueter:
LGTM but would like to hold this one until we are clear on the rollout impact (PUCM run, worker reboots)

s-fairchild:
Thanks for the reviews everyone!
Latest push has small changes, I  realized I hadn't set the file mode, and improved the if conditional matching of eth* (had all ||, was missing an &&).

I've tested by running `go run -tags aro ./cmd/aro operator master` against my cluster and all works as expected, machineconfigs render and the script executes.

s-amann (approved):

cadenmarchese (approved):

pkg/operator/controllers/dnsmasq/dnsmasq.go
```diff
@@ -110,6 +110,46 @@ chmod 0744 /etc/resolv.conf.dnsmasq
...
+# this will restart dnsmasq, reapplying our /etc/resolv.conf file and overwriting any modifications made by NetworkManager
+
+interface=$1
+action=$2
+
+log() {
+    logger -i "$0" -t '99-DNSMASQ-RESTART SCRIPT' "$@"
+}
+
+if [[ $interface == eth* && $action == "up" ]] || [[ $interface == eth* && $action == "down" ]]; then
```
ventifus, line 127: In the test, the interface that gets unplugged / replugged (the VF interface) is called enP53531s1. Maybe we look for interfaces called `e*` or specifically eth* plus enP*?

pkg/operator/controllers/dnsmasq/dnsmasq.go
```diff
@@ -110,6 +110,46 @@ chmod 0744 /etc/resolv.conf.dnsmasq
...
+    log "$action happened on $interface, connection state is now $CONNECTIVITY_STATE"
+    log "restarting dnsmasq now"
+    if systemctl restart dnsmasq; then
```
ventifus, line 130: `systemctl restart` will start dnsmasq if it's not running. Could we end up racing eth0's DHCP on initial boot potentially ending up with nothing in /etc/resolv.conf.dnsmasq? Let's `systemctl try-restart` here instead so that it will only restart if it's already running.

s-amann (approved):

cadenmarchese:
@s-fairchild mind TAL at the e2e failure? It looks related to your change: https://github.com/Azure/ARO-RP/pull/2995/checks?check_run_id=15211124105

s-fairchild:
@cadenmarchese Yes i'll look into them, may have to run e2e locally on Monday to see what's going on.

pkg/api/defaults.go
```diff
@@ -69,7 +69,7 @@ func DefaultOperatorFlags() OperatorFlags {
 		"aro.azuresubnets.serviceendpoint.managed": flagTrue,
 		"aro.banner.enabled":                       flagFalse,
 		"aro.checker.enabled":                      flagTrue,
-		"aro.dnsmasq.enabled":                      flagTrue,
+		"aro.dnsmasq.enabled":                      flagFalse,
```
bennerv, line 72: I think this will break all the existing clusters that use egress lockdown.  

I was hoping more for an approach where there's a new feature flag and a new file that's laid down, either within the dnsmasq controller, or a separate controller.

pkg/operator/controllers/dnsmasq/dnsmasq.go
```diff
@@ -110,6 +110,52 @@ chmod 0744 /etc/resolv.conf.dnsmasq
 /bin/rm $TMPNETRESOLV
 /bin/rm $TMPSELFRESOLV
 {{ end }}
+
```
bennerv, line 113: move this to a new file or string which is applied ad-hoc.

pkg/operator/controllers/dnsmasq/dnsmasq.go
```diff
@@ -173,6 +230,11 @@ func ignition2Config(clusterDomain, apiIntIP, ingressIP string, gatewayDomains [
 		return nil, err
 	}
 
+	nmDispatcherRestartDnsmasq, err := nmDispatcherRestartDnsmasq()
```
bennerv, line 233: before this, drop in a feature flag above this on whether or not we enable this.

bennerv (dismissed):
I was thinking of a new feature flag.  We shouldn't disable dnsmasq feature for existing clusters, hence the new feature flag to put these behind.  

Then we can apply it ad-hoc to clusters.

s-fairchild:
@bennerv I misunderstood, thanks for explaining. I've updated the branch with a new operator flag to enable/disable these changes specifically.
I want to add an e2e test for the new flag soon so it actually gets tested.

kimorris27 (commented):
Looking good! I made two small recommendations on the restart script.

I also have an open-ended question. I saw in the description of https://issues.redhat.com/browse/ARO-3474 the concern of ensuring that customer changes to the cluster that write to `resolv.conf` are still allowed. The solution implemented in this PR will preserve ARO-specific `resolv.conf` configurations, but do you think it's possible to preserve customer-initiated changes as well?

I'd imagine that when Azure host servicing happens, we could still end up getting DNS-related incidents because of the loss of that configuration. These incidents would be less severe than the one that got this work started though, since the critical stuff from an SRE perspective is the ARO-specific configuration.

s-fairchild:
> Looking good! I made two small recommendations on the restart script.
> 
> I also have an open-ended question. I saw in the description of https://issues.redhat.com/browse/ARO-3474 the concern of ensuring that customer changes to the cluster that write to `resolv.conf` are still allowed. The solution implemented in this PR will preserve ARO-specific `resolv.conf` configurations, but do you think it's possible to preserve customer-initiated changes as well?
> 
> I'd imagine that when Azure host servicing happens, we could still end up getting DNS-related incidents because of the loss of that configuration. These incidents would be less severe than the one that got this work started though, since the critical stuff from an SRE perspective is the ARO-specific configuration.

@ventifus Correct me if I'm wrong here please..
It depends on how customers are making these custom DNS changes. If they are following [this documentation](https://learn.microsoft.com/en-us/azure/openshift/howto-custom-dns) I think this change won't interfere with custom DNS.
Customer's should also be able to set their own operator flags I believe(?), so they could disable this restart script from being added to the dnsmasq machine configs. With the dnsmasq restart flag enabled, this should happen.
If a customer tries to manually set DNS on the nodes via editing `/etc/resolv.conf`, then this will overwrite those changes. If the customer sets custom DNS and DOMAIN settings via `nmcli`, then the next time `dnsmasq` gets restarted those settings will be used. For a short time anyways. Once the DHCP lease get's renewed the DNS settings made with `nmcli` will get overwritten, followed by our Pre scripts getting the new addresses.

To summarize, if the customer sets custom DNS via Azure's DHCP server settings, this change ensures that that address is always used.

An off topic question, would anyone mind if I moved the dnsmasq operator's templates to actual files and then used embed? I'm going to do this in a second commit and I can delete if we don't want that.

kimorris27:
> To summarize, if the customer sets custom DNS via Azure's DHCP server settings, this change ensures that that address is always used.

Cool, this explanation makes sense to me.

> An off topic question, would anyone mind if I moved the dnsmasq operator's templates to actual files and then used embed? I'm going to do this in a second commit and I can delete if we don't want that.

👍🏻 from me on this change.

ventifus:
@kimorris27 If a customer tries to edit resolv.conf via MachineConfig, that won't work well, since NetworkManager rewrites that file on boot, and aro-dnsmasq-pre.sh swaps it with resolv.conf.dnsmasq on service start. I'm not sure if there's any other avenue of modifying that file that wouldn't have conflicts.

kimorris27:
>@kimorris27 If a customer tries to edit resolv.conf via MachineConfig, that won't work well, since NetworkManager rewrites that file on boot, and aro-dnsmasq-pre.sh swaps it with resolv.conf.dnsmasq on service start. I'm not sure if there's any other avenue of modifying that file that wouldn't have conflicts.

Ah, I see. Since modifying the MachineConfigs is unsupported, I don't think I'm worried about that case. As long as the customer's configuration is maintained when they modify it in the supported way, that seems good enough.

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
```diff
@@ -0,0 +1,45 @@
...
+    log "$(echo -n \"/etc/resolv.conf file metadata: \") $(ls -lZ /etc/resolv.conf)"
+
+    log "/etc/resolv.conf.dnsmasq contents
+
+    $(cat /etc/resolv.conf.dnsmasq)"
+
+    log "$(echo -n "/etc/resolv.conf.dnsmasq file metadata: ") $(ls -lZ /etc/resolv.conf.dnsmasq)"
+}
+
+if [[ $interface == eth* && $action == "up" ]] || [[ $interface == eth* && $action == "down" ]] || [[ $interface == enP* && $action == "up" ]] || [[ $interface == enP* && $action == "down" ]]; then
```
cadenmarchese, line 30: It looks like we are proceeding from here no matter what (if the NIC is up, we proceed, if the NIC is down, we proceed), so do we need this `if` statement?

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
```diff
@@ -0,0 +1,45 @@
...
+    log "$action happened on $interface, connection state is now $CONNECTIVITY_STATE"
+    log "Pre dnsmasq restart file information"
+    check_dns_files
```
cadenmarchese, line 33:
```suggestion
    log_dns_files
```
Nit, since it doesn't looks like we are doing a comparison.

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
```diff
@@ -0,0 +1,45 @@
...
+interface=$1
+action=$2
+
+log() {
+    logger -i "$0" -t '99-DNSMASQ-RESTART SCRIPT' "$@"
+}
+
+# log dns configuration information relevant to SRE while troubleshooting
+# The line break used here is important for formatting
+check_dns_files() {
```
cadenmarchese, line 16:
```suggestion
log_dns_files() {
```

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
```diff
@@ -0,0 +1,45 @@
...
+
+    log "$(echo -n "/etc/resolv.conf.dnsmasq file metadata: ") $(ls -lZ /etc/resolv.conf.dnsmasq)"
+}
+
+if [[ $interface == eth* && $action == "up" ]] || [[ $interface == eth* && $action == "down" ]] || [[ $interface == enP* && $action == "up" ]] || [[ $interface == enP* && $action == "down" ]]; then
+    log "$action happened on $interface, connection state is now $CONNECTIVITY_STATE"
+    log "Pre dnsmasq restart file information"
+    check_dns_files
+    log "restarting dnsmasq now"
+    if systemctl try-restart dnsmasq --wait; then
```
cadenmarchese, line 35: Where do we compare the current dnsmasq configuration with the "correct" one? Or, are we just restarting whenever we get a machine config change from the operator?

pkg/api/defaults.go
```diff
@@ -70,6 +70,7 @@ func DefaultOperatorFlags() OperatorFlags {
 		"aro.banner.enabled":                       flagFalse,
 		"aro.checker.enabled":                      flagTrue,
 		"aro.dnsmasq.enabled":                      flagTrue,
+		"aro.restartdnsmasq.enabled":               flagFalse,
```
cadenmarchese, line 73: Since our next PUCM will already cause reboots, we can merge with this enabled:
```suggestion
		"aro.restartdnsmasq.enabled":               flagTrue,
```

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
s-fairchild: There are a number of other things it can be, checkout the documentation for available actions: https://networkmanager.dev/docs/api/latest/NetworkManager-dispatcher.html

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
s-fairchild: We are restarting regardless of the differences, and logging them.

kimorris27:
This LGTM! I'll revisit after E2E is passing.

pkg/operator/controllers/dnsmasq/dnsmasq.go
```diff
@@ -17,105 +18,17 @@ import (
...
-{{ end }}
-`))
+const (
+	configFileName    = "dnsmasq.conf"
+	unitFileName      = "dnsmasq.service"
+	prescriptFileName = "aro-dnsmasq-pre.sh"
+)
 
 func config(clusterDomain, apiIntIP, ingressIP string, gatewayDomains []string, gatewayPrivateEndpointIP string) ([]byte, error) {
+	t := template.Must(template.New(configFileName).Parse(preScriptFile))
```
bennerv, line 28:
```suggestion
	t := template.Must(template.New(configFileName).Parse(configFile))

**Review by bennerv — CHANGES_REQUESTED** [review]

File: pkg/operator/controllers/dnsmasq/machineconfig_controller.go [hunk 13]
```diff
@@ -55,6 +55,10 @@ func (r *MachineConfigReconciler) Reconcile(ctx context.Context, request ctrl.Re
 		return reconcile.Result{}, nil
 	}
 
+	if instance.Spec.OperatorFlags.GetSimpleBoolean(restartDnsmasqEnabled) {
```
kimorris27, line 58: nit: Can we assign the operator flag's value to the intermediate variable `restartDnsmasq` and use it both here and in the call to `reconcileMachineConfigs`? Just so the three controllers that reference the flag match each other in this way.

kimorris27 (commented):
Looking good! One nit.

cadenmarchese (approved):
Steps taken to test this, in addition to the CI checks here:

- Shared cluster
  - Confirmed that we could run this new operator in an existing cluster without any restarts (until a PUCM to update the operator feature flags goes out)
- Local development
  - Built a new cluster with these changes to ensure that creation was successful and the cluster custom resource got the new feature flag
  - [Ran the operator out of cluster](https://github.com/Azure/ARO-RP/blob/master/docs/operators.md#how-to-run-the-operator-locally-out-of-cluster) with these changes to confirm that the dispatcher.d script was created on all nodes (`oc get node -o name | xargs -I {} oc debug {} -- chroot /host sh -c 'cat /etc/NetworkManager/dispatcher.d/99-dnsmasq-restart'`)
- INT
  - Confirmed that the dispatcher.d script was created on all nodes in a new cluster created in INT

pkg/operator/controllers/dnsmasq/machineconfig_controller.go
s-fairchild: Yes, I hadn't really thought about it that way. I choice not to use a variable here simply because it wasn't necessary. But I like your suggestion of continuity, I'll make a follow up PR.

PR #2997, opened 2023-06-29
AutoSized Nodes now reference both masters and workers

bennerv:

https://issues.redhat.com/browse/ARO-3501

Align managed services + allow autonode sizing on masters.  

https://github.com/openshift/managed-cluster-config/blob/master/deploy/kubelet-config/01-kubelet-config.yaml

### Test plan for issue:

Create cluster, enable autonode sizing, ensure it applies to masters & workers.  Profit

no


petrkotas (approved):
LGTM

s-amann (approved):

PR #3003, opened 2023-06-30
Fix MDSD memory issue on RP VMs 

yehornaumenko:

https://issues.redhat.com/browse/ARO-3461

When RP VMs are restarted the Available memory decreases, to the point of turning a VM unhealthy/unresponsive some times. The MDSD process was found to use 50% of the total mem, triggered by Fluentbit suddenly sending all historic logs in the VM.

This PR adds a config change for the RP and Gwy Fluentbit config to keep track of the last log sent to MDSD and in case of a restart resume sending from that point, instead of sending all historic logs again.

We left the change in Canary over the weekend and could see no further memory degradation after restarting VM on Monday. There is a small increase in disk write ops/sec, as Fluentbit updates the cursor of journald logs as they are processed, but it stays constant at approx. 8 ops/sec. See graphs and detailed info in 
[Testing notes](https://microsoft.sharepoint.com/teams/AzureRedHatOpenShift/_layouts/15/Doc.aspx?sourcedoc={b051066f-44da-4884-9d5b-200349b362eb}&action=edit&wd=target%28Untitled%20Section.one%7C5076b5f8-771d-4b77-9730-c3887683f2ca%2FTest%20in%20eastus2euap%20RP%7C9ca29845-a358-4ca6-a91a-18dfbcf6c85d%2F%29&wdorigin=NavigationUrl)

### Test plan for issue:

Manual tests on Canary/Pilot

The investigation performed is described on OneNote
[OneNote notes](https://microsoft.sharepoint.com/:o:/t/AzureRedHatOpenShift/Em8GUbDaRIRInVsgA0mzYusBLJuxfi_zQaSxf_wQQIMz4g?e=9MKRoB)

To be discussed. I believe we should save all the conclusions and add some SOP on ADO Wiki


pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -138,6 +138,7 @@ cat >/etc/fluentbit/fluentbit.conf <<'EOF'
 	Name systemd
 	Tag journald
 	Systemd_Filter _COMM=aro
+	DB /var/lib/fluent/journaldb
```
bizz001, line 141: We should probably also add that to the gateway vmss

tiguelu (approved):
LGTM

SrinivasAtmakuri (approved):
LGTM!

s-amann (approved):

pkg/deploy/generator/scripts/rpVMSS.sh
yehornaumenko: done

PR #3005, opened 2023-06-30
Wrap Timeout errors within Condition steps

tsatam:

Does not fix [ARO-1558](https://issues.redhat.com/browse/ARO-1558), but provides further clarity in log messages when the below issue is encountered

This change will wrap all `ErrWaitTimeout` errors (`"timed out waiting for the condition"`) returned within our steps runner. 

The `wait.PollImmediateUntil()` function used within our Condition checker returns an `ErrWaitTimeout` of its own when the condition is not met within the specified timeout, leading internal errors within the conditions that match `ErrWaitTimeout` to appear identical to conditions that legitimately time out. 

### Test plan for issue:

A unit test was added to the runner tests to handle this specific case. 

No.

### Additional Notes

- The wrapped error will still match `ErrWaitTimeout` (e.g. `errors.Is(err, wait.ErrWaitTimeout)` returns `true`). If we would like to differentiate these errors from legitimate condition timeouts in code, we will need to change this behavior. 
- The same change has been made in openshift/ARO-Installer#51

### Alternatives

- Wrap _all_ errors returned by the condition functions rather than just `ErrWaitTimeout`. 
- Implement handling for any internal `ErrWaitTimeout` within each individual condition function
  - Effort resembling this will likely end up happening within [ARO-1558](https://issues.redhat.com/browse/ARO-1558) long-term, in order to handle timeout errors that should continue condition polling.

### DIFF
--- a/pkg/util/steps/condition.go
+++ b/pkg/util/steps/condition.go
@@ -5,6 +5,7 @@ package steps
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


SudoBrendan (approved):
This seems like a good approach to this problem, where we'll at least differentiate these two (identically looking) issues. Can you submit a PR to ARO-Installer as well?

tsatam:
Equivalent PR submitted to ARO-Installer as well: https://github.com/openshift/ARO-Installer/pull/51

jhoreman (approved):
Thanks for this, lgtm.

PR #3007, opened 2023-07-05
ARO-3536: Add error handling when "CLUSTER" env variable is empty.

edisonLcardenas:

https://issues.redhat.com/browse/ARO-3536

Fixes

Error handling when env variable "CLUSTER" is blank. 
Running hack create script throws a mem issue if it's just an empty "CLUSTER" env variable.

### Test plan for issue:
Run hack create script.

No

### DIFF
--- a/hack/cluster/cluster.go
+++ b/hack/cluster/cluster.go
@@ -44,10 +44,6 @@ func run(ctx context.Context, log *logrus.Entry) error {
 	}
 	clusterName := os.Getenv(Cluster)
 
-	if clusterName == "" {
-		return fmt.Errorf("your environment variable \"CLUSTER\" is empty, please pass a value")
-	}
-
 	osClusterVersion := os.Getenv("OS_CLUSTER_VERSION")
 
 	c, err := cluster.New(log, env, os.Getenv("CI") != "")
--- a/hack/cluster/cluster.go
+++ b/hack/cluster/cluster.go
@@ -44,6 +44,10 @@ func run(ctx context.Context, log *logrus.Entry) error {
 	}
 	clusterName := os.Getenv(Cluster)
 
+	if clusterName == "" {
+		return fmt.Errorf("your environment variable \"CLUSTER\" is empty, please pass a value")
+	}
+
 	osClusterVersion := os.Getenv("OS_CLUSTER_VERSION")
 
 	c, err := cluster.New(log, env, os.Getenv("CI") != "")
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -124,7 +124,7 @@ func IsCI() bool {
 // Otherwise it returns nil.
 func ValidateVars(vars ...string) error {
 	for _, v := range vars {
-		if _, found := os.LookupEnv(v); !found {
+		if v, found := os.LookupEnv(v); !found || v == "" {
 			return fmt.Errorf("environment variable %q unset", v)
 		}
 	}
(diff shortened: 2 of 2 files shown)


hack/cluster/cluster.go
```diff
@@ -44,6 +44,10 @@ func run(ctx context.Context, log *logrus.Entry) error {
 	}
 	clusterName := os.Getenv(Cluster)
 
+	if clusterName == "" {
+		return fmt.Errorf("your environment variable \"CLUSTER\" is empty, please pass a value")
+	}
+
```
AldoFusterTurpin, line 50: I think we can use the already defined variable name Cluster instead of hardcoding it on the error message, I would also change a bit the error message, to something like
```go
if clusterName == "" {
    return fmt.Errorf("the environment variable %v is empty, please use a non empty value", Cluster)
}
```

I also think it would be a good idea to add a similar check in pkg/util/cluster/cluster.go:144 as this funcitonality can be potentially called from other places than the hack script as it is a separate component of a different package.

Something like (pkg/util/cluster/cluster.go:143):
```go
func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName string, osClusterVersion string) error {
if clusterName == "" {
    return fmt.Errorf("clusterName can not be empty")
}
. . .
```

AldoFusterTurpin (changes requested):
@edisonLcardenas  Good work so far 😃 

Just a couple of suggestions to rehuse the variable name that is already defined + proposing to add the same validation in the package cluster to avoid future similar mistakes when using cluster.Cluster from somewhere else.

hack/cluster/cluster.go
yjst2012: how do you feel like move all validations to one place to  **ValidateVars()**, like:

```
		if v, found := os.LookupEnv(v); !found || v == "" {
			return fmt.Errorf("environment variable %q unset or set to empty", v)
		}
```

AldoFusterTurpin: This is a good idea If we are sure that we are not changing the behavior of the code already using ValidateVars().
😃

ArrisLee (approved):
LGTM

AldoFusterTurpin (dismissed):
LGTM.Thanks! 😃 

One last comment as an optional suggestion: I still think that adding also the validation inside pkg/util/cluster/cluster.go:143 in 
_func (c *Cluster) Create_ 
is a good idea as cluster.Create() could probably be called from somewhere else not touching env vars.

pkg/env/env.go
```diff
@@ -124,7 +124,7 @@ func IsCI() bool {
 // Otherwise it returns nil.
 func ValidateVars(vars ...string) error {
 	for _, v := range vars {
-		if _, found := os.LookupEnv(v); !found {
+		if v, found := os.LookupEnv(v); !found || v == "" {
```
yjst2012, line 127: reuse the variable "v" here does not sound very good, sometimes it is dangerous and hard to debug, it have no problem in the case tho. it does not help for a good coding practice, how about change to another one?

yjst2012: oh, it does have problem here, in case the env variable is set to "", the error handling cannot print out the actual variable name, instead it only prints out an "" as v has been re-assigned.

yjst2012 (approved):

petrkotas (approved):
LGTM

PR #3008, opened 2023-07-05
Picking subnets from the unused spaces in the vnet rather than randomly generating them.

anshulvermapatel:
We are usually getting this error while creating a dev cluster.

> \"error\": {\r\n    \"code\": \"NetcfgInvalidSubnet\",\r\n    \"message\": \"Subnet '<cluster>-master' is not valid in virtual network 'dev-vnet'.\",\r\n    \"details\": []\r\n  }\r\n}"}] 
exit status 1

This error comes when a randomly generate subnets from masters and workers are tried to get created in the vnet where the subnets already exist with same address space and hence the installation fails.

This PR changes the way we pick the subnet address space to get the unused space from the vnet and pick the spaces from those spaces rather than generating randomly.

Fixes - not created a JIRA for this yet.

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -49,7 +49,7 @@ import (
 )
 
 const (
-	devVnet = "dev-vnet"
+	devVnetName = "dev-vnet"
 )
 
 type Cluster struct {
@@ -348,7 +348,7 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 
 func (c *Cluster) getUnusedAddressSpaces(ctx context.Context, vnetResourceGroup string) (string, string, string, error) {
 	addressSpaces := make([]string, 0)
-	subnetListResultPage, err := c.subnets.List(ctx, vnetResourceGroup, devVnet)
+	subnetListResultPage, err := c.subnets.List(ctx, vnetResourceGroup, devVnetName)
 	if err != nil {
 		return "", "", "", err
 	}
@@ -357,7 +357,7 @@ func (c *Cluster) getUnusedAddressSpaces(ctx context.Context, vnetResourceGroup 
 		mp[*subnet.AddressPrefix] = true
 	}
 
-	devVnet, err := c.vnets.Get(ctx, c.env.ResourceGroup(), devVnet, "")
+	devVnet, err := c.vnets.Get(ctx, c.env.ResourceGroup(), devVnetName, "")
 	if err != nil {
 		return "", "", "", err
 	}
@@ -370,7 +370,7 @@ func (c *Cluster) getUnusedAddressSpaces(ctx context.Context, vnetResourceGroup 
 	}
 
 	if len(addressSpaces) < 2 {
-		return "", "", "", fmt.Errorf("not enough address space available in the virtual network %s", devVnet)
+		return "", "", "", fmt.Errorf("not enough address space available in the virtual network %s", devVnetName)
 	}
 	addressPrefix := addressSpaces[0][:len(addressSpaces[0])-1] + "3"
 	return addressPrefix, addressSpaces[0], addressSpaces[1], nil
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -9,6 +9,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"net/http"
 	"os"
 	"strings"
@@ -199,9 +200,14 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 		return err
 	}
 
-	addressPrefix, masterSubnet, workerSubnet, err := c.getUnusedAddressSpaces(ctx, vnetResourceGroup)
-	if err != nil {
-		return err
+	var addressPrefix, masterSubnet, workerSubnet string
+	if c.ci {
+		addressPrefix, masterSubnet, workerSubnet = c.generateSubnets()
+	} else {
+		addressPrefix, masterSubnet, workerSubnet, err = c.getUnusedAddressSpaces(ctx, vnetResourceGroup)
+		if err != nil {
+			return err
+		}
 	}
 
 	var kvName string
@@ -344,6 +350,19 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 
 	c.log.Info("done")
 	return nil
+}
+
+func (c *Cluster) generateSubnets() (vnetPrefix string, masterSubnet string, workerSubnet string) {
+	// pick a random 23 in range [10.3.0.0, 10.127.255.0]
+	// 10.0.0.0/16 is used by dev-vnet to host CI
+	// 10.1.0.0/24 is used by rp-vnet to host Proxy VM
+	// 10.2.0.0/24 is used by dev-vpn-vnet to host VirtualNetworkGateway
+	rand.Seed(time.Now().UnixNano())
+	x, y := rand.Intn((124))+3, 2*rand.Intn(128)
+	vnetPrefix = fmt.Sprintf("10.%d.%d.0/23", x, y)
+	masterSubnet = fmt.Sprintf("10.%d.%d.0/24", x, y)
+	workerSubnet = fmt.Sprintf("10.%d.%d.0/24", x, y+1)
+	return
 }
 
 func (c *Cluster) getUnusedAddressSpaces(ctx context.Context, vnetResourceGroup string) (string, string, string, error) {
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -9,7 +9,6 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
-	"math/rand"
 	"net/http"
 	"os"
 	"strings"
@@ -49,6 +48,10 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
+const (
+	devVnet = "dev-vnet"
+)
+
 type Cluster struct {
 	log          *logrus.Entry
 	env          env.Core
@@ -64,6 +67,7 @@ type Cluster struct {
 	openshiftclustersv20220904        redhatopenshift20220904.OpenShiftClustersClient
 	securitygroups                    network.SecurityGroupsClient
 	subnets                           network.SubnetsClient
+	vnets                             network.VirtualNetworksClient
 	routetables                       network.RouteTablesClient
 	roleassignments                   authorization.RoleAssignmentsClient
 	peerings                          network.VirtualNetworkPeeringsClient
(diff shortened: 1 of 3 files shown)


ArrisLee (approved):
LGTM, thanks for the fix!

s-fairchild (approved):
Thanks for fixing this!

s-amann (dismissed):

pkg/util/cluster/cluster.go
```diff
@@ -346,20 +357,44 @@ func (c *Cluster) generateSubnets() (vnetPrefix string, masterSubnet string, wor
...
+		if mp[addressPrefix] {
+			continue
+		}
+		addressSpaces = append(addressSpaces, addressPrefix)
+	}
+
+	if len(addressSpaces) < 2 {
+		return "", "", "", fmt.Errorf("not enough address space available in the virtual network %s", devVnetName)
+	}
+	addressPrefix := addressSpaces[0][:len(addressSpaces[0])-1] + "3"
```
yjst2012, line 394: the algo here seems not very right.
IIUC, for the addressPrefix, it simply picks up the second last char and append a "3" to it, however this only works for /24 subnet. it breaks when subnet mask length changes to other value, eg, 23 or 25, (not sure if this may happen tho). I would suggest to change this to something like:

```
	addr := strings.Split(addressSpaces[0], "/")
	i, _ := strconv.Atoi(addr[1])
	addressPrefix = addr[0] + "/" + strconv.Itoa(i-1)
```

Also, I believe we need to make sure that subnets listed here are consecutive, otherwise the first and second available subnets are probably not merge-able and then the Prefix is invalid. probably need a better algo to calculate it. :thinking:

yjst2012 (changes requested):
Thanks for putting this up, left a comment. pls have a look.

pkg/util/cluster/cluster.go
yjst2012: the lines to comment is a bit messed up, should be this:

```
addressPrefix := addressSpaces[0][:len(addressSpaces[0])-1] + "3"
return addressPrefix, addressSpaces[0], addressSpaces[1], nil
```

pkg/util/cluster/cluster.go
```diff
@@ -346,20 +357,44 @@ func (c *Cluster) generateSubnets() (vnetPrefix string, masterSubnet string, wor
...
+	return addressPrefix, addressSpaces[0], addressSpaces[1], nil
+}
```
AldoFusterTurpin, line 396: I really like the idea of moving all that into its own function. Nevertheless, I think we can improve it a bit as we are defining the method on the type cluster, and it has already a lot of fields and methods. We can improve the cohesion here.

It would be a good idea to just remove the method from the struct and convert it to a standard function and pass just the interfaces and values that we need. In that way, the code is a bit better designed and easier to test as we don't need the whole cluster object or the whole c.env as we just pass from the outside what we need (the interfaces and c.env.ResourceGroup()).  Also, adding the name of the values returned in the function enhances readability as we have more than one return type of the same type (string, string, string).

I had in mind something like:
```go
func getUnusedAddressSpaces(ctx context.Context, subnets network.SubnetsClient, vnets network.VirtualNetworksClient, vnetResourceGroup string, envResourceGroup string) (addressPrefix string, masterSubnet string, workerSubnet string, err error) {
	var addressSpaces []string
	subnetListResultPage, err := subnets.List(ctx, vnetResourceGroup, devVnetName)
	if err != nil {
		return "", "", "", err
	}
	mp := make(map[string]bool)
	for _, subnet := range subnetListResultPage.Values() {
		mp[*subnet.AddressPrefix] = true
	}

	devVnet, err := vnets.Get(ctx, envResourceGroup, devVnetName, "")
	if err != nil {
		return "", "", "", err
	}

	for _, addressPrefix := range *devVnet.AddressSpace.AddressPrefixes {
		if mp[addressPrefix] {
			continue
		}
		addressSpaces = append(addressSpaces, addressPrefix)
	}

	if len(addressSpaces) < 2 {
		return "", "", "", fmt.Errorf("not enough address space available in the virtual network %s", devVnetName)
	}
	addressPrefix = addressSpaces[0][:len(addressSpaces[0])-1] + "3"
	return addressPrefix, addressSpaces[0], addressSpaces[1], nil
}
```
Also, if you want, you can use 
```go
var addressSpaces []string
```
instead of 
```go
addressSpaces := make([]string, 0)
```
as you save one memory allocation and is more idiomatic in Go.

AldoFusterTurpin (changes requested):
@anshulvermapatel Good work so far! I just left a few suggestions as I think we can apply some improvements to make it even better 😃 

Also, maybe it would be a good idea to include some unit tests as part of the PR.

pkg/util/cluster/cluster.go
anshulvermapatel: Yes, thank you @yjst2012 , you are right, I will make the changes.

pkg/util/cluster/cluster.go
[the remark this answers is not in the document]
anshulvermapatel: right.

anshulvermapatel:
We have 3 regions for creating dev cluster - eastus, westeurope, australiaeast.
I see that in westeurope and australiaeast, the address space is 10.0.0.0/9 .. the code was written for generating subnets in the range 10.0.0.0/16 which is not now the case with the above mentioned 2 regions. However, eastus still have the address space of 10.0.0.0/16 which is actually in the form of multiple address spaces in the 10.0.0.0/16 range and lot of them are actually missing as well. I am not sure why it is like this. We should make this uniform and add the 10.0.0.0/9 address space for the dev-vnet in eastus also. Assuming that we will do it, I have made the changes in the code. Since we have 10.0.0.0/9 address spaces which is a massive number, I thing there would not be any issue generating the addresses randomly. 
Currently also, we are only facing issues in eastus because of the address spaces being in the form of multiple small spaces.
[v4-eastus](https://portal.azure.com/#@redhat0.onmicrosoft.com/resource/subscriptions/fe16a035-e540-4ab7-80d9-373fa9a3d6ae/resourceGroups/v4-eastus/providers/Microsoft.Network/virtualNetworks/dev-vnet/addressSpace)
[v4-westeurope](https://portal.azure.com/#@redhat0.onmicrosoft.com/resource/subscriptions/fe16a035-e540-4ab7-80d9-373fa9a3d6ae/resourceGroups/v4-westeurope/providers/Microsoft.Network/virtualNetworks/dev-vnet/addressSpace)

pkg/util/cluster/cluster.go
```diff
@@ -64,6 +64,7 @@ type Cluster struct {
 	openshiftclustersv20220904        redhatopenshift20220904.OpenShiftClustersClient
 	securitygroups                    network.SecurityGroupsClient
 	subnets                           network.SubnetsClient
+	vnets                             network.VirtualNetworksClient
```
AldoFusterTurpin, line 67: This field is only used in the _New_ function, so it is really never used. Can we removed it for now?

pkg/util/cluster/cluster.go
```diff
@@ -119,6 +120,7 @@ func New(log *logrus.Entry, environment env.Core, ci bool) (*Cluster, error) {
 		openshiftclustersv20220904:        redhatopenshift20220904.NewOpenShiftClustersClient(environment.Environment(), environment.SubscriptionID(), authorizer),
 		securitygroups:                    network.NewSecurityGroupsClient(environment.Environment(), environment.SubscriptionID(), authorizer),
 		subnets:                           network.NewSubnetsClient(environment.Environment(), environment.SubscriptionID(), authorizer),
+		vnets:                             network.NewVirtualNetworksClient(environment.Environment(), environment.SubscriptionID(), authorizer),
```
AldoFusterTurpin, line 123: This field is only "used" here (in this case instantiated), so it is really never used. Can we removed for now?

pkg/util/cluster/cluster.go
```diff
@@ -341,19 +340,15 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
...
 	var x, y int
 	rand.Seed(time.Now().UnixNano())
-	// Local Dev clusters are limited to /16 dev-vnet
-	if !c.ci {
-		x, y = 0, 2*rand.Intn(128)
-	} else {
-		x, y = rand.Intn((124))+3, 2*rand.Intn(128)
-	}
+	// Local Dev clusters are limited to /9 dev-vnet
+	x, y = rand.Intn((124))+3, 2*rand.Intn(128)
```
AldoFusterTurpin, line 351: I think there is a redundant parenthesis around _124_.

Also, two things to comment here, just my personal opinion 😃 
- First: What about break the line a bit, it would improve readability and we can delete the line _var x, y int_, something like:
```go
secondOctet := rand.Intn(124) + 3
thirdOctet := 2 * rand.Intn(128)

vnetPrefix = fmt.Sprintf("10.%d.%d.0/23", secondOctet, thirdOctet)
masterSubnet = fmt.Sprintf("10.%d.%d.0/24", secondOctet, thirdOctet)
workerSubnet = fmt.Sprintf("10.%d.%d.0/24", secondOctet, thirdOctet+1)
```
- Second: maybe it would be nice to pass the strings with the form"10.%d.%d.0/23" and add some unit tests to ensure the function is behaving as expected, because right now we are assuming something about the Azure environment that can potentially change. Let's add unit tests to express that + pass the parameters to make it extendable. In other words, I think we should not assume anything about the environment and instead pass values to the function.

AldoFusterTurpin (commented):
Thank you for the comment clarifying the current state in Azure.
I have left a comment regarding the assumptions about details of the environment.
Thanks!

pkg/util/cluster/cluster.go
anshulvermapatel: Yes I will remove it, I added it for the previous changes but now it is not required.

pkg/util/cluster/cluster.go
anshulvermapatel: Same goes with this one. Will remove it.

pkg/util/cluster/cluster.go
anshulvermapatel: Regarding the parenthesis, I  don't know why I have this tendency to not to touch an existing code it being whatever way it is. I did see the redundant parenthesis, but I just kept it as is.

I agree with your both the opinions and I will change the code and add the unit tests.  However, I want to have a discussion first on the `dev-vnet` vnet in `v4-eastus` rg and to have a timeline about when we will be changing it because if this code gets merged, creation of dev cluster will often fail since the subnet is /16 only.

AldoFusterTurpin: Perfect! The idea bout injecting the parameters in the function and adding unit tests also goes in the direction of being able to make the function work in all cases passing from outside /16 or the corresponding one depending on the region.

SudoBrendan:
Hey @anshulvermapatel I've had requests from my team to revisit this, since E2Es and other testing locally will still occasionally fail and lose us an hour+ of time. Re-reviewing this, it looks like this has undergone several revisions. What's the current state, anything I can help with?

cadenmarchese (changes requested):
LGTM, but let's get Aldo's suggestions in before we merge.

kimorris27:
Why did this get closed?

PR #3015, opened 2023-07-10
Move the dynamic validator out of api/

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-3573

The dynamic validator doesn't actually validate API structs and instead checks things in Azure. Since this adds a lot of dependencies, making pkg/api something lighter is more difficult. Therefore, we should move it out.

### Test plan for issue:

Unit tests + E2E
N/A


SrinivasAtmakuri (approved):
LGTM!

petrkotas (approved):
LGTM

PR #3017, opened 2023-07-11
Ping after healthz

s-amann:

Fixes https://issues.redhat.com/browse/ARO-3330

1. Scrape metric for API server ping health if the healthz check returns a non 200. This will be used as a signal for Azure Resource Health check.

### Test plan for issue:

INT deployment

Yes:  https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/484208/API-Server-Unreachable

### DIFF
--- a/pkg/monitor/cluster/cluster.go
+++ b/pkg/monitor/cluster/cluster.go
@@ -137,18 +137,18 @@ func (mon *Monitor) Monitor(ctx context.Context) (errs []error) {
 		})
 	}
 
-	// If API Server is unreachable, don't need to run the next checks
-	statusCode, err := mon.getAPIServerPingCode(ctx)
-	if err != nil || statusCode != http.StatusOK {
+	// If API is not returning 200, don't need to run the next checks
+	statusCode, err := mon.emitAPIServerHealthzCode(ctx)
+	if err != nil {
 		errs = append(errs, err)
-		friendlyFuncName := steps.FriendlyName(mon.getAPIServerPingCode)
+		friendlyFuncName := steps.FriendlyName(mon.emitAPIServerHealthzCode)
 		mon.log.Printf("%s: %s", friendlyFuncName, err)
 		mon.emitGauge("monitor.clustererrors", 1, map[string]string{"monitor": friendlyFuncName})
+	}
+	if statusCode != http.StatusOK {
 		return
 	}
-
 	for _, f := range []func(context.Context) error{
-		mon.emitAPIServerHealthzCode,
 		mon.emitAroOperatorHeartbeat,
 		mon.emitAroOperatorConditions,
 		mon.emitNSGReconciliation,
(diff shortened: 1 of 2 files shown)


pkg/monitor/cluster/cluster.go
```diff
@@ -137,18 +137,23 @@ func (mon *Monitor) Monitor(ctx context.Context) (errs []error) {
 		})
 	}
 
-	// If API Server is unreachable, don't need to run the next checks
-	statusCode, err := mon.getAPIServerPingCode(ctx)
-	if err != nil || statusCode != http.StatusOK {
+	// If API is not returning 200, fallback to checking ping and short circuit the rest of the checks
```
bennerv, line 140: can we put a comment at the top of this saying: "the apiserver healthz MUST always be emitted first as our APIServerHealthzShort monitor requires it to be emitted always".

pkg/monitor/cluster/healthz.go
```diff
@@ -21,10 +21,10 @@ func (mon *Monitor) emitAPIServerHealthzCode(ctx context.Context) error {
 		"code": strconv.FormatInt(int64(statusCode), 10),
 	})
 
-	return err
+	return statusCode, err
 }
 
-func (mon *Monitor) getAPIServerPingCode(ctx context.Context) (int, error) {
```
niontive, line 27: Could we still return the status code? We'll use this in the metric.

pkg/monitor/cluster/cluster.go
```diff
@@ -137,18 +137,24 @@ func (mon *Monitor) Monitor(ctx context.Context) (errs []error) {
...
+	if err != nil {
 		errs = append(errs, err)
-		friendlyFuncName := steps.FriendlyName(mon.getAPIServerPingCode)
+		friendlyFuncName := steps.FriendlyName(mon.emitAPIServerHealthzCode)
 		mon.log.Printf("%s: %s", friendlyFuncName, err)
 		mon.emitGauge("monitor.clustererrors", 1, map[string]string{"monitor": friendlyFuncName})
+	}
+	// If API is not returning 200, fallback to checking ping and short circuit the rest of the checks
+	if statusCode != http.StatusOK {
+		err := mon.emitAPIServerPingCode(ctx)
```
niontive, line 150: Could we additionally emit  gauge metrics for:

1. Status code for ping
2. If `emitAPIServerPingCode` returned an err

This makes it easier to set up monitors for resource health check.

pkg/monitor/cluster/healthz.go
s-amann: The status code is emitted within the `emitAPIServerPingCode` function

pkg/monitor/cluster/cluster.go
s-amann: 1. `emitAPIServerPingCode` emits the statuscode
2. I can add a metric if this call fails

niontive (approved):

cadenmarchese (approved):

PR #3009, opened 2023-07-05
Read installer image digests from an environment variable

mbarnes:
This depends on a couple Azure DevOps pull requests:

- [RP-Config : Add installerImageDigests to global configurations](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/RP-Config/pullrequest/8382631)
- [ARO.Pipelines : Add INSTALLER_IMAGE_DIGESTS env to update-aro-openshift-versions](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines/pullrequest/8382662)

### Which issue this PR addresses:

[[ARO-3526] Store the Default Install Version in the OCP Versions Database](https://issues.redhat.com/browse/ARO-3526)

### What this PR does / why we need it:

Changes `aro update-versions` to read **aro-installer** image digests from an environment variable named "**INSTALLER_IMAGE_DIGESTS**" which contains data from [RP-Config](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/RP-Config).

"**INSTALLER_IMAGE_DIGESTS**" should be a JSON-formatted mapping of `"MAJOR.MINOR"` OpenShift versions to an image digest (e.g. `{"4.10": "sha256:...", ...}`).

### Test plan for issue:

Execute [OneBranch - Update ARO OpenShift Versions - PullRequest](https://msazure.visualstudio.com/AzureRedHatOpenShift/_build?definitionId=287387&_a=summary) pipeline to verify.

### Is there any documentation that needs to be updated for this PR?
[Updating the ARO-Installer image digests](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki?pagePath=/Build%252Dout/ARO%252DInstaller%20Image%20Deployment%20Process&anchor=updating-the-aro-installer-image-digests) procedure will change from modifying **ARO-RP** code to modifying **RP-Config**.

PR #3018, opened 2023-07-11
update_ocp_versions: Code cleanup

mbarnes:
Small followup to https://github.com/Azure/ARO-RP/pull/3009 to address a nit called out in code review.

Split off a new function so the environment variable name `INSTALLER_IMAGE_DIGESTS` only has to be specified once.

[[ARO-3523] Move the ARO-Installer Image Digests to RP-Config](https://issues.redhat.com/browse/ARO-3523)


petrkotas (approved):
LGTM

PR #3019, opened 2023-07-11
bumping mdm mdsd fb versions

jhoreman:

https://issues.redhat.com/browse/ARO-3613
Fixes

bump the fluentbit, mdm, and mdsd images we use so that they don't get flagged as P0/P1/P2 vulnerabilities that need to be resolved.

### Test plan for issue:

Mdsd still ingests logs in INT
fluentbit still ingests logs in INT
mdm still shows metrics coming in INT

no


SudoBrendan (approved):
LGTM - since we're bumping a major version on Fluentbit, we should probably add additional bake time on the rollout to make sure logs still ship after a 24h period or so.

SudoBrendan (approved):
LGTM

PR #3020, opened 2023-07-11
Preview api lb profile

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-3111

This PR adds the LoadBalancerProfile to the preview, admin, and internal API, static validation of the LoadBalancerProfile, and sets default loadBalancerProfile.  I had intended to add Dynamic validation but that is probably going to add too much to this PR and makes more sense to add it during implementation of aro update.

The PreviewAPI includes all fields for this feature though only ManagedOutboundIPs is usable during phase 1 of this feature.  This was done this way to not require multiple API releases.

### Test plan for issue:

Added/updated existing unit tests.  I also manually verified some API calls and validated the expected result in cosmos.  This will be further tested once the aro update part is implemented.

Not yet


pkg/api/v20230701preview/openshiftcluster.go
```diff
@@ -143,6 +181,9 @@ type NetworkProfile struct {
 
 	// The OutboundType used for egress traffic.
 	OutboundType OutboundType `json:"outboundType,omitempty"`
+
+	// The cluster load balancer profile.
+	LoadbalancerProfile *LoadbalancerProfile `json:"loadBalancerProfile,omitempty"`
```
tony-schndr, line 186: Calling out my use of a pointer here, which seems to deviate from the rest of the API structs.  This is a pointer to denote the absence of this struct when outboundType is UserDefinedRouting (no public LB).  I'm open to suggestions if there is a better way.

pkg/api/v20230701preview/openshiftcluster.go
```diff
@@ -133,6 +133,44 @@ const (
 	OutboundTypeLoadbalancer       OutboundType = "Loadbalancer"
 )
 
+// LoadbalancerProfile represents the profile of the cluster public loadbalancer.
+type LoadbalancerProfile struct {
+	// The desired managed outbound IPs for the cluster public load balancer.
+	ManagedOutboundIPs *ManagedOutboundIPs `json:"managedOutboundIps,omitempty" mutable:"true"`
```
tony-schndr, line 139: Calling out my use of a pointer here.  Customer should only be using one of ManagedOutboundIPs, OutboundIPs, or OutboundIPPrefixes.  I could not find a good way to validate if only 1 of these 3 options was within the body of the API request, and validate that ManagedOutboundIPs.Count is between 1-20.  It doesn't seem possible to validate if ManagedOutboundIPs.Count=0 in the body of the API request, or if golang initialized it to its zero value.  I'm open to suggestions if there is a better way to handle this.

pkg/api/v20230701preview/openshiftcluster.go
```diff
@@ -133,6 +133,41 @@ const (
...
+	// The list of effective outbound IP addresses of the public load balancer.
+	EffectiveOutboundIPs []EffectiveOutboundIP `json:"effectiveOutboundIps,omitempty"`
+	// The desired outbound IP resources for the cluster load balancer.
+	OutboundIPs []OutboundIP `json:"outboundIps,omitempty" mutable:"true"`
+	// The desired outbound IP Prefix resources for the cluster load balancer.
+	OutboundIPPrefixes []OutboundIPPrefix `json:"OutboundIpPrefixes,omitempty" mutable:"true"`
+	// The desired number of allocated SNAT ports per VM. Allowed values are in the range of 0 to 64000 (inclusive). The default value is 1024.
+	AllocatedOutboundPorts *int `json:"allocatedOutboundPorts,omitempty" mutable:"true"`
```
tony-schndr, line 153: Using a pointer here again to denote empty value.  0 value here could be ok except that ARO defaults this field to 1024, I don't know of a clean way to determine if 0 was passed in to the API or if golang just initialized it to 0.

pkg/api/v20230701preview/openshiftcluster.go
petrkotas: This is actually a valid way to solve the problem you have described. It is also widely used inside of the Kubernetes (e.g. https://github.com/kubernetes/kubernetes/blob/16534deedf1e3f7301b20041fafe15ff7f178904/staging/src/k8s.io/apiserver/pkg/apis/example/v1/types.go#L84) The only way to be sure that object was not parsed, is to test it's nil.

I am +1 on this

pkg/api/v20230701preview/openshiftcluster.go
petrkotas: Perfectly fine.

pkg/api/v20230701preview/openshiftcluster.go
petrkotas: Perfectly fine for this usecase.

pkg/api/v20230701preview/openshiftcluster_validatestatic.go
```diff
@@ -228,6 +231,44 @@ func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np
...
+		} else if isOutboundIPPrefixes && !isOutboundIPs && !isManagedOutboundIPCount {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".outboundIpPrefixes", "The field outboundIpPrefixes is not implemented at this time, please check back later.")
+		} else if !isManagedOutboundIPCount && !isOutboundIPPrefixes && !isOutboundIPs {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path, "The provided loadBalancerProfile is invalid: must specify one of managedOutboundIps, outboundIps, or outboundIpPrefixes.")
+		} else {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path, "The provided loadBalancerProfile is invalid: can only use one of managedOutboundIps, outboundIps, or outboundIpPrefixes at a time.")
+		}
+
+		if lbp.AllocatedOutboundPorts != nil {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".allocatedOutboundPorts", "The field allocatedOutboundPorts is not implemented at this time, please check back later.")
```
petrkotas, line 263: What is the plan to implement this?

petrkotas (commented):
Hi @tony-schndr thank you, this is an excelent work. I have some minor comments.

pkg/api/v20230701preview/openshiftcluster_validatestatic.go
tony-schndr: This is will be implemented in phase 3, so after `managedOutboundIPs`, `outboundIPs` and `outboundIPPrefixes` are implemented.  There will be some validation on min/max value and allowance of 0 (for dynamic snat port allocation.)

pkg/api/v20230701preview/openshiftcluster.go
s-amann: same here, if we allow no value and the default value and a custom value then I think this is a fine way to go about it. We are at the restriction of JSON, go defaults, and not having a confusing API to make the distinction in another manner.

pkg/api/v20230701preview/openshiftcluster_validatestatic.go
```diff
@@ -228,6 +231,44 @@ func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np
...
+}
+
+func (sv openShiftClusterStaticValidator) validateLoadBalancerProfile(path string, lbp *LoadbalancerProfile, isCreate bool) error {
+	if lbp != nil {
+		var isManagedOutboundIPCount = lbp.ManagedOutboundIPs != nil
+		var isOutboundIPs = lbp.OutboundIPs != nil
+		var isOutboundIPPrefixes = lbp.OutboundIPPrefixes != nil
+
+		if isManagedOutboundIPCount && !isOutboundIPPrefixes && !isOutboundIPs {
+			if !(lbp.ManagedOutboundIPs.Count > 0 && lbp.ManagedOutboundIPs.Count <= 20) {
```
s-amann, line 249: nit: combine this condition with the line above to avoid single path condition nesting.

pkg/api/v20230701preview/openshiftcluster.go
tony-schndr: I appreciate the confirmation that this is OK to do.  I was not sure considering we currently don't use a lot of pointers in the API and I was not sure if there was a way to implement this similar to how maps work in golang (ok is true if key/value is found)

pkg/api/v20230701preview/openshiftcluster_validatestatic.go
SudoBrendan: We've decided to include all fields for all phases of this project in 1 api version so we don't have to do multiple preview API releases to rollout the entirety of this feature. The phases are described in the PRD document linked on the parent Epic for the work item linked at the top of this PR.

pkg/api/v20230701preview/openshiftcluster_validatestatic.go
```diff
@@ -228,6 +231,44 @@ func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np
 			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".outboundType", "The provided outboundType '%s' is invalid: cannot use UserDefinedRouting if either API Server Visibility or Ingress Visibility is public.", np.OutboundType)
 		}
 	}
+
+	if np.OutboundType == OutboundTypeUserDefinedRouting && np.LoadbalancerProfile != nil {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".loadBalancerProfile", "The provided loadBalancerProfile is invalid: cannot use a loadBalancerProfile if outboundType is UserDefinedRouting.")
+	}
```
SudoBrendan, line 237: So today, we have UDR and Loadbalancer, right? Would it be better for us to strictly check for LB type network instead of "not UDR" here in the case where we add additional types later? That seems a little less brittle to me.

pkg/api/v20230701preview/openshiftcluster_validatestatic_test.go
```diff
@@ -517,8 +522,171 @@ func TestOpenShiftClusterStaticValidateNetworkProfile(t *testing.T) {
...
+							ID: "outboundIPID",
+						},
+					},
+					OutboundIPPrefixes: []OutboundIPPrefix{
+						{
+							ID: "outboundIPPrefixID",
+						},
+					},
+				}
+			},
```
SudoBrendan, line 595: I know it's gonna be a lot of copy-pasta, but I'd appreciate having all permutations of this in our tests to give us assurance. No combination of any of these fields should be allowed. I guess I'm assuming that this error would precede the "this field is unsupported" errors.

SudoBrendan (commented):
I generally agree with other reviews so far. This is looking very good! A few comments from me as well, thanks Tony!

pkg/api/v20230701preview/openshiftcluster_validatestatic.go
tony-schndr: So you are thinking something like this?  `np.OutboundType == OutboundTypeLoadbalancer && np.LoadbalancerProfile == nil` then return error?  The issue with this approach is the LoadbalancerProfile is not mandatory during cluster creation, so the cluster doc may not have a LoadbalancerProfile until `api.SetDefaults` is called, which does not happen until AFTER static validation.

pkg/api/v20230701preview/openshiftcluster_validatestatic_test.go
tony-schndr: If more than 1 of these fields is specified at a time will the `else` will catch it and return `"The provided loadBalancerProfile is invalid: can only use one of managedOutboundIps, outboundIps, or outboundIpPrefixes at a time."`.  If a single field is specified (which is correct as these 3 fields are mutually exclusive) but the field is not yet supported it will return "not implemented".  

I will add the additional permutations.

pkg/api/v20230701preview/openshiftcluster_validatestatic.go
SudoBrendan: I think I would rewrite it as `if np.LoadbalancerProfile != nil && np.OutboundType != OutboundTypeLoadbalancer` or whatever it's called.

pkg/api/admin/openshiftcluster.go
```diff
@@ -149,10 +184,11 @@ type NetworkProfile struct {
...
 
-	APIServerPrivateEndpointIP string           `json:"privateEndpointIp,omitempty"`
-	GatewayPrivateEndpointIP   string           `json:"gatewayPrivateEndpointIp,omitempty"`
-	GatewayPrivateLinkID       string           `json:"gatewayPrivateLinkId,omitempty"`
-	PreconfiguredNSG           PreconfiguredNSG `json:"preconfigureNSG,omitempty"`
+	APIServerPrivateEndpointIP string               `json:"privateEndpointIp,omitempty"`
+	GatewayPrivateEndpointIP   string               `json:"gatewayPrivateEndpointIp,omitempty"`
+	GatewayPrivateLinkID       string               `json:"gatewayPrivateLinkId,omitempty"`
+	PreconfiguredNSG           PreconfiguredNSG     `json:"preconfigureNSG,omitempty"`
+	LoadbalancerProfile        *LoadbalancerProfile `json:"loadBalancerProfile,omitempty"`
```
SudoBrendan, line 191: super-nit: we are using `loadBalancerProfile` for this customer-facing option, but OutboundType uses the casing `Loadbalancer`. Might be good to align our casing for customer-facing options.

pkg/api/admin/openshiftcluster.go
```diff
@@ -139,6 +139,41 @@ const (
...
+	// The fully qualified Azure resource id.
+	ID string `json:"id,omitempty"`
+}
+
+// LoadbalancerProfile represents the profile of the cluster public loadbalancer.
+type LoadbalancerProfile struct {
+	// The desired managed outbound IPs for the cluster public load balancer.
+	ManagedOutboundIPs *ManagedOutboundIPs `json:"managedOutboundIps,omitempty"`
+	// The list of effective outbound IP addresses of the public load balancer.
+	EffectiveOutboundIPs []EffectiveOutboundIP `json:"effectiveOutboundIps,omitempty"`
```
cadenmarchese, line 153: Will this be a list of resource IDs, or the IPs themselves? Here it sounds like they are IPs, but the comment on 162 suggests that they will be the resource IDs.

pkg/api/defaults.go
```diff
@@ -46,6 +46,15 @@ func SetDefaults(doc *OpenShiftClusterDocument) {
 		if doc.OpenShiftCluster.Properties.NetworkProfile.PreconfiguredNSG == "" {
 			doc.OpenShiftCluster.Properties.NetworkProfile.PreconfiguredNSG = PreconfiguredNSGDisabled
 		}
+
+		// If OutboundType is Loadbalancer and there is no LoadBalancerProfile, set default one
```
cadenmarchese, line 50: I'm not entirely clear on why `ManagedOutboundIP` needs to be set if the setup is default (OutboundType is Loadbalancer and there is no LoadBalancerProfile). Am I missing something?

cadenmarchese (commented):
The Go looks good to me, I just have a few questions regarding the ManagedOutboundIP / EffectiveOutboundIP setup. Thanks!

pkg/api/admin/openshiftcluster.go
```diff
@@ -139,6 +139,41 @@ const (
...
+// ResourceReference represents a reference to an Azure resource.
+type ResourceReference struct {
+	// The fully qualified Azure resource id.
+	ID string `json:"id,omitempty"`
+}
+
+// LoadbalancerProfile represents the profile of the cluster public loadbalancer.
+type LoadbalancerProfile struct {
+	// The desired managed outbound IPs for the cluster public load balancer.
+	ManagedOutboundIPs *ManagedOutboundIPs `json:"managedOutboundIps,omitempty"`
```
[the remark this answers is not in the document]
tony-schndr: PER the PRD "The existing AKS user experience should be used as guidance for this ARO capability".  I am trying to mirror the [aks rest specs](https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/get?tabs=HTTP#managedclusterloadbalancerprofile).

pkg/api/admin/openshiftcluster.go
tony-schndr: If you follow the references all the way down to `ResourceReference` it clarifies that this is an object containing an azure ID.  The description is mostly verbatim from aks azure rest specs.
```
// ResourceReference represents a reference to an Azure resource.
type ResourceReference struct {
	// The fully qualified Azure resource id.
	ID string `json:"id,omitempty"`
```

pkg/api/defaults.go
tony-schndr: Mainly to be explicit on the configuration of the public load balancer.  If this is not set here then it will be nil and not be represented properly to the customer on the return payload of  a put/patch operation.  I think it would be confusing if the field `effectiveOutboundIPs` will be present and showing the 1 outbound IP on the load balancer but none of the outbound IP options are represented. 

If I don't set it as a default the cluster doc looks like this:
```
"loadBalancerProfile" : {
   "effectiveOutboundIps" : [
      {"resourceIdOfDefaultPublicIP"}
   ]
}
```
If set the cluster doc looks like this:
```
"loadBalancerProfile" : {
   "effectiveOutboundIps" : [
      {"resourceIdOfDefaultPublicIP"}
   ]
   "managedOutboundIps": {
      "count": 1
   }
}
```

I also added some static validation to make sure the customer can't null all the fields, which I think will be bugged if `effectiveOutboundIps` is set but the others aren't.
```
"loadBalancerProfile" : {
   "effectiveOutboundIps" : [
      {"resourceIdOfDefaultPublicIP"}
   ], 
   # Invalid, at least one outboundIP type must be active at all times
   "managedOutboundIps": null,
   "outboundIps" : null,
   "outboundIpPrefixes": null
}
```

pkg/api/admin/openshiftcluster.go
tony-schndr: I'm initially not for this change, but I'm open to changing it if the team wholly agrees.  The PRD states to use the aks user experience as a guide for this feature, so I'm trying to mirror [aks rest specs](https://learn.microsoft.com/en-us/rest/api/aks/managed-clusters/get?tabs=HTTP#containerservicenetworkprofile) convention and I'm hesitant to deviate from that.  Unfortunately the casing for `OutboundTypes` originate from the installer code therefore it was carried into the RP.

pkg/api/v20230701preview/openshiftcluster_validatestatic.go
tony-schndr: It would have to be `np.LoadbalancerProfile != nil && np.OutboundType != OutboundTypeLoadbalancer && np.OutboundType != ""` because OutboundType is not mandated to be set and if it is an empty string it will default to `OutboundTypeLoadbalancer`.  If defaults were ever changed then this would break, but I don't know if/when we would ever change the default value as that would be a breaking change within itself...

pkg/api/defaults.go
cadenmarchese: > I think it would be confusing if the field effectiveOutboundIPs will be present and showing the 1 outbound IP on the load balancer but none of the outbound IP options are represented.

If it would be confusing to set some fields of loadBalancerProfile and not others, wouldn't we want to keep loadBalancerProfile nil If OutboundType is Loadbalancer? To me one of these would make the most sense:

```
"loadBalancerProfile" : {} // is outboundType LoadBalancer
```
OR
```
"loadBalancerProfile" : {
   "effectiveOutboundIps" : [
      {"resourceIdOfDefaultPublicIP"}
   ]
   "managedOutboundIps": {
      "count": 1
   }
} // is outboundType LoadBalancer
```
Whereas now, we have:
```
"loadBalancerProfile" : {
   "managedOutboundIps": {
      "count": 1
   }
} // is outboundType LoadBalancer
```

tony-schndr: Yes, the second one is the ideal state.  The issue is that `effectiveOutboundIps` is unkown at this point of the code.  My plan to add a step to the install process / update process to update `effectiveOutboundIps`.  
Also I have also been using the aks user experience as a guideline here and this is exactly how aks behaves during.

cadenmarchese: Ok - so we are going to populate it, but can't yet. Makes sense to me 👍

pkg/api/v20230701preview/openshiftcluster.go
```diff
@@ -133,6 +133,41 @@ const (
 	OutboundTypeLoadbalancer       OutboundType = "Loadbalancer"
 )
 
+// ResourceReference represents a reference to an Azure resource.
+type ResourceReference struct {
+	// The fully qualified Azure resource id.
+	ID string `json:"id,omitempty"`
+}
+
+// LoadBalancerProfile represents the profile of the cluster public loadbalancer.
```
SudoBrendan, line 142: nit: "load balancer"

SudoBrendan (approved):
While I still have heebie-jeebies about "if it's UDR + LB Profile, that's bad" instead of "if it's anything-but-lb and LB Profile, that's bad", and I'd rather have a test case that makes the networktype "Foo" + LBProfile and see a failure... I won't block merging if I'm the only person that thinks so. Otherwise, LGTM. I left one more nit if desired.

pkg/api/v20230701preview/openshiftcluster_validatestatic.go
SudoBrendan: We talked about this a bit on slack, so bringing the jist of our conversation here - the way that our API is architected, we set defaults after static validation, and if we did the `!= ""` check in static validation here, that would be the first of its kind.

On my side, I think I'm ok to say "if you want an LB Profile, you've gotta strictly supply networkType=LB, instead of relying on our (current) default", though, in AKS, this will work for you (empty networktype, populated LB Profile), I assume because they set defaults before doing validation...

pkg/api/v20230701preview/openshiftcluster.go
tony-schndr: I agree that condition is not the greatest, How do you feel about updating that condition to `np.LoadbalancerProfile != nil && np.OutboundType != OutboundTypeLoadbalancer && np.OutboundType != ""`, assuming the default outboundType will always be Loadbalancer we should be OK.  IF default outboundType was changed, then I think this would break without a unit test failing.

I will fix the nit.

SudoBrendan (dismissed):

SudoBrendan (dismissed):

cadenmarchese (dismissed):

swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/preview/2023-07-01-preview/redhatopenshift.json
```diff
@@ -1846,6 +1846,16 @@
         }
       }
     },
+    "EffectiveOutboundIP": {
```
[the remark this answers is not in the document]
tony-schndr: Just pushed this, though from what I can tell only the parent object `EffectiveOutboundIPs` can be set to read only.

SudoBrendan (dismissed):

swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/preview/2023-07-01-preview/redhatopenshift.json
```diff
@@ -1888,6 +1898,45 @@
...
+        "outboundIpPrefixes": {
+          "description": "The desired outbound IP Prefix resources for the cluster load balancer.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/OutboundIPPrefix"
+          },
+          "x-ms-identifiers": []
+        },
+        "allocatedOutboundPorts": {
+          "description": "The desired number of allocated SNAT ports per VM. Allowed values are in the range of 0 to 64000 (inclusive). The default value is 1024.",
```
bennerv, line 1935: >Load balancer gives [SNAT](https://learn.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections) ports in multiples of 8. If you provide a value not divisible by 8, the configuration operation is rejected. Each load balancing rule and inbound NAT rule consume a range of eight ports. If a load balancing or inbound NAT rule shares the same range of 8 as another, no extra ports are consumed.

Is 0 really a valid value?

bennerv (commented):
I haven't looked at the logic behind the things, but we should confirm before we release the API definitions BU is okay with our descriptions.

SudoBrendan (approved):

PR #3030, opened 2023-07-13
Remove the inbuilt installer

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-1417

Dead code removal :)

Removes the go build tag code also, since that is to enable a handful of things in the installer fork that we no longer use.

### Test plan for issue:
Unit tests+E2E should cover things, but since this code hasn't been used in production for months (and now not e2e either) should be fine.

N/A, just removing code -- although we do need to redocument how install stuff works w/ hive


hawkowl:
Tiny semi-functional change moved to https://github.com/Azure/ARO-RP/pull/3029

ellis-johnson (approved):
LGTM!

hawkowl:
🥳

PR #3036, opened 2023-07-17
Set OpenShift 4.11 as the default install version

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-3675

Changes the default install version to 4.11, updates the PR E2E to do so, and doesn't error on the discoverycache being generated in 4.10.

### Test plan for issue:

E2E

Don't think so?

### DIFF
--- a/pkg/util/version/const.go
+++ b/pkg/util/version/const.go
@@ -25,14 +25,26 @@ const (
 
 var GitCommit = "unknown"
 
-// DefaultInstallStream describes stream we are defaulting to for all new clusters
-var DefaultInstallStream = &Stream{
-	Version:  NewVersion(4, 10, 63),
-	PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:340091aefa0bba06bbb99cc58cb1f2b73404c832f72b83c526b8e7677efbecef",
+// DefaultMinorVersion describes the minor OpenShift version to default to
+var DefaultMinorVersion = 11
+
+// DefaultInstallStreams describes the latest version of our supported streams
+var DefaultInstallStreams = map[int]*Stream{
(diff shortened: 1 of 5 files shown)


jhoreman (approved):
Seems straightforward and E2E passed on PR.

petrkotas (approved):

pkg/util/version/const.go
```diff
@@ -25,14 +25,26 @@ const (
...
+// DefaultInstallStreams describes the latest version of our supported streams
+var DefaultInstallStreams = map[int]*Stream{
+	10: {
+		Version:  NewVersion(4, 10, 63),
+		PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:340091aefa0bba06bbb99cc58cb1f2b73404c832f72b83c526b8e7677efbecef",
+	},
+	11: {
+		Version:  NewVersion(4, 11, 44),
+		PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:52cbfbbeb9cc03b49c2788ac7333e63d3dae14673e01a9d8e59270f3a8390ed3",
+	},
```
bennerv, line 40: I know this is already merged, but why add the 4.10 stuff to a `DefaultInstallStreams` struct if we're not setting 4.10 as a default anymore?  

We can't pass in "4.10" as an install version and it picks the version above.  It might confuse team member IMO.

hawkowl: @bennerv I've done it like this because the discoverycache right now is 4.10 and so I want what the default install stream for 4.10 used to be. Maybe it doesn't make as much sense because you can't ask for "4.10" and get anything. (should we have that, maybe?)

PR #3044, opened 2023-07-19
Remove install word in Condition error messages

SrinivasAtmakuri:
With PR: https://github.com/Azure/ARO-RP/pull/2450 merged, I have added cluster installation failed in the error message and step hiveClusterDeploymentReady will also be run during the Update path, the existing error message will confuse customers.

This PR modifies the error message by removing the install word in the error.

Fixes


petrkotas (dismissed):

SudoBrendan (commented):
I'll be honest, I'm not intuitively seeing how this wording improves the customer experience. These are customer errors, right?

ArrisLee (approved):
LGTM

rogbas:
LGTM but I don't think we need to rush, cut a new release and re-run INT to include this change.

IMO it could wait for next release

petrkotas:
@rogbas I greee, it is LGTM, it makes sense to me, but it can wait.

UlrichSchlueter (approved):
Not sure the error messages are actually all actionable. Might be a good idea to discuss this with the PMs for a next iteration.

cadenmarchese (approved):

PR #3037, opened 2023-07-17
Fix e2e must gather

cadenmarchese:
### Which issue this PR addresses:

Fixes the CI must-gather - a continuation of https://github.com/Azure/ARO-RP/pull/3038

PR #3050, opened 2023-07-21
Add jq for must-gather

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-3640

Similar to #3037 but uses go-jq which is in the e2e tools instead of downloading from gh-actions.

### Test plan for issue:

Will cause a run to fail to test

N/A


hawkowl:
Failed run artifact: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=76820965&view=artifacts&pathAsName=false&type=publishedArtifacts

cadenmarchese:
Nice! In case there's a way to re-use this, I wanted to mention that we're also downloading jq in this pipeline: https://github.com/Azure/ARO-RP/blob/master/.pipelines/templates/template-clean-subscription.yml

hawkowl:
@cadenmarchese I've added that there, now -- what do you think?

PR #3053, opened 2023-07-21
Add worker profile status

gouthamMN:

Fixes https://issues.redhat.com/browse/ARO-2124

OpenShiftCluster GET API returns enriched worker profile data under a new field "workerProfilesStatus". This change is introduced only to new API. More details can be found on [Design Doc](https://microsoft-my.sharepoint.com/:w:/r/personal/gniranjan_microsoft_com/_layouts/15/doc2.aspx?sourcedoc=%7B920CBF46-6A85-484C-93C9-F7DC3EAF7220%7D&file=OpenShift%20cluster%20-%20Create%20API.docx&action=default&mobileredirect=true&DefaultItemOpen=1&ct=1687552228824&wdOrigin=OFFICECOM-WEB.START.EDGEWORTH&cid=0a2f122b-8dfa-4561-8c90-dc131ad9efc2&wdPreviousSessionSrc=HarmonyWeb&wdPreviousSession=c6aedd5b-f157-4414-aec1-aa5fd19c133e&share=IQFGvwyShWpMSJPJ99w-r3IgAXlNyUSBQnB0VSUWgdbjT5I)

Any API call made to endpoints **/openShiftClusters** or **/openShiftCluster** with GET method using newer API versions (>= v2023-09-04) will return the enriched worker profile data under a new field "**workerProfilesStatus**" in the response and "**workerProfiles**" filed will contain the last input that was passed in PUT/PATCH request body for cluster creation or update.

The response and behavior of any older APIs versions (< v2023-09-04) remain unchanged i.e, only "workerProfiles" filed exists in the response.

### Test plan for issue:

Existing unit tests and E2E test copied over from previous API.
As suggested in the comments, performed the Regression Tests:
1. Created a ARO clustered from az cli `**az aro create**`.
2. Performed ARO cluster update `**az aro update**`.

From the test, it was confirmed that there were no changes in the behavior of the existing stable API (v2023-04-01). The new API (v2023-09-04) cannot be tested at this point as the az cli is not updated to use it.
 
N/A


weinong (approved):
LGTM

pkg/swagger/typewalker.go
```diff
@@ -141,6 +141,9 @@ func (tw *typeWalker) schemaFromType(t types.Type, deps map[*types.Named]struct{
 
 				properties := tw.schemaFromType(field.Type(), deps)
 				properties.Description = strings.Trim(node.Doc.Text(), "\n")
+				if field.Name() == "WorkerProfilesStatus" {
+					properties.ReadOnly = true
+				}
```
bennerv, line 146: does ARM prevent the user passing in values for this if we set the swagger specification to ReadOnly?  If not, we should do a server-side check to ensure the user can't actually patch this.

pkg/api/v20200430/openshiftcluster_convert.go
```diff
@@ -51,8 +51,15 @@ func (c openShiftClusterConverter) ToExternal(oc *api.OpenShiftCluster) interfac
...
-		for _, p := range oc.Properties.WorkerProfiles {
+		workerProfiles := oc.Properties.WorkerProfiles
+
+		// Use enriched worker profile data when available
+		if oc.Properties.WorkerProfilesStatus != nil {
+			workerProfiles = oc.Properties.WorkerProfilesStatus
+		}
+
+		out.Properties.WorkerProfiles = make([]WorkerProfile, 0, len(workerProfiles))
+		for _, p := range workerProfiles {
```
bennerv, line 62: Wouldn't this generate new examples for the older API?

pkg/api/v20230904/openshiftcluster.go
```diff
@@ -64,6 +64,9 @@ type OpenShiftClusterProperties struct {
 	// The cluster worker profiles.
 	WorkerProfiles []WorkerProfile `json:"workerProfiles,omitempty"`
 
+	// The cluster worker profiles status.
+	WorkerProfilesStatus []WorkerProfile `json:"workerProfilesStatus,omitempty"`
```
bennerv, line 68: i'm wondering if we should make a new type for this.  In the update code, we never actually set some of the parameters that exist within a workerProfile struct.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -245,6 +245,9 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 	doc.OpenShiftCluster.Properties.ClusterProfile.PullSecret = ""
 	doc.OpenShiftCluster.Properties.ServicePrincipalProfile.ClientSecret = ""
 
+	// We don't return enriched worker profile data on PUT/PATCH operations
+	doc.OpenShiftCluster.Properties.WorkerProfilesStatus = nil
```
bennerv, line 249: If it's a read-only property, why not?

pkg/util/clusterdata/worker_profile.go
```diff
@@ -102,10 +102,11 @@ func (ce machineClientEnricher) Enrich(
 	oc.Lock.Lock()
 	defer oc.Lock.Unlock()
 
-	oc.Properties.WorkerProfiles = workerProfiles
+	oc.Properties.WorkerProfilesStatus = workerProfiles
 	return nil
 }
 
 func (ce machineClientEnricher) SetDefaults(oc *api.OpenShiftCluster) {
 	oc.Properties.WorkerProfiles = nil
```
bennerv, line 110: we need to remove this now right?  We don't want to nil this out.

bennerv (changes requested):
I think we're missing updating any usage to point to the new workerProfilesStatus.  

Anywhere we use kubeSubnets or the workerProfiles field internally may be subject to breakage or not complete / expected functionality.  We should explore them, as the PRs should be merged together to prevent any breaking changes by introducing this change.

pkg/util/clusterdata/worker_profile.go
bennerv:
```
	oc.Properties.WorkerProfiles = nil
```

specifically the above field.  It's not actually called anywhere, but we should be careful not to have it nil out worker profiles by accident.  We should retain any user-provided initial data.

pkg/api/v20200430/openshiftcluster_convert.go
gouthamMN: Nope, because under _openshiftcluster_example.go_ package we set the _WorkerProfilesStatus_ field to nil before calling the converter function (ToExternal). This will ensure that the swagger examples remain unchanged. 

Older APIs swagger examples never showed the enriched worker profile data under _workerProfiles_ field. Do I decided to keep it same.

pkg/api/v20230904/openshiftcluster.go
gouthamMN: WorkerProfilesStatus - I see all the field are set/referenced under _worker_profile.go_ package while performing the enrichment.

pkg/frontend/openshiftcluster_putorpatch.go
gouthamMN: I'm fine doing that, but it was decided to return the _WorkerProfilesStatus_ field only under GET/LIST operation during design meeting.

pkg/swagger/typewalker.go
gouthamMN: Updated _openshiftcluster_validatestatic.go_ to validate _workerProfilesStatus_ filed is nil in the input request. This gets called during preflightvalidation.

gouthamMN:
> I think we're missing updating any usage to point to the new workerProfilesStatus.
> 
> Anywhere we use kubeSubnets or the workerProfiles field internally may be subject to breakage or not complete / expected functionality. We should explore them, as the PRs should be merged together to prevent any breaking changes by introducing this change.

These are the [list of ARO-RP pkgs](https://microsoft-my.sharepoint.com/:w:/r/personal/gniranjan_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7B941C5275-EAB7-4573-9423-639E3B909548%7D&file=terraform%20src%20code%20changes.docx&action=default&mobileredirect=true&DefaultItemOpen=1&login_hint=gniranjan%40microsoft.com&ct=1690489532484&wdOrigin=OFFICECOM-WEB.START.EDGEWORTH&cid=eef8dfe6-c9ff-48ec-a1fd-bf29662eab82&wdPreviousSessionSrc=HarmonyWeb&wdPreviousSession=0fc9ae7c-0271-48dc-9af9-1db30b0b63ac&share=IQF1UhyUt-pzRZQjY547kJVIAZ5vsXYczopWbnfN13q-qNc) that had reference to workerProfiles filed and I had updated whichever needs to be updated to workerProfilesStatus. Let me know if you think I'm missing anything.

s-amann (approved):

pkg/util/clusterdata/worker_profile.go
```diff
@@ -102,10 +102,10 @@ func (ce machineClientEnricher) Enrich(
 	oc.Lock.Lock()
 	defer oc.Lock.Unlock()
 
-	oc.Properties.WorkerProfiles = workerProfiles
```
s-amann, line 105: after further thinking (thanks pto), I believe this will break our admin api. Since the doc will not have `workerProfiles` set, when the conversion is made, workerProfiles is nil and so the output will not have workerProfiles since we first check for workerProfiles and then check for workerProfileStatus. Any automation in place which relies on the workerProfiles via admin API would need to be changed.

I think an alternative approach, would be to leave the doc the same (always have it be enriched via workerProfiles) and for new API versions we can do the conversion.

s-amann (changes requested):
I believe we need to at least rethink the conversion logic so it can account for nil workerProfileStatus before workerProfiles since the enrichment in this PR will never populate workerProfiles. See comments below.

bennerv:
I haven't had the chance to review the list of places it's been used yet, but we need to be careful this change doesn't break existing functionality.  

Think this:
1. Essentially this change makes the workerProfile immutable.  the customer sets it on creation, but moving forward, they actually can't update the field and the RP does anything about it.  Probably not the best behavior, but the reality 
2. The workerProfilesStatus field represents the in-cluster state of the world.  We use this in multiple places: to gather all subnets and ensure we have the proper RBAC on the vnets.
3. When a customer performs day2 operations, they can technically bring worker nodes from a completely different subnet
4. Because of 3, we need to update our usage everywhere workerProfiles is used.  We always want to reconcile or use the state of the in-cluster machines/machinesets, otherwise this change may have undesired consequences. 


I would say generally we can proceed by:
1. Anywhere `workerProfiles` is used, change it to `workerProfilesStatus` if `workerProfilesStatus != nil || len(workerProfilesStatus) > 0`.  Otherwise use `workerProfiles`.  
2. When we update the azure-cli to use the latest API version, we need to take the intersection of `workerProfiles` and  `workerProfilesStatus` to ensure we have the proper FPSP and CSP RBAC to work.

gouthamMN:
> I haven't had the chance to review the list of places it's been used yet, but we need to be careful this change doesn't break existing functionality.
> 
> Think this:
> 
> 1. Essentially this change makes the workerProfile immutable.  the customer sets it on creation, but moving forward, they actually can't update the field and the RP does anything about it.  Probably not the best behavior, but the reality
> 2. The workerProfilesStatus field represents the in-cluster state of the world.  We use this in multiple places: to gather all subnets and ensure we have the proper RBAC on the vnets.
> 3. When a customer performs day2 operations, they can technically bring worker nodes from a completely different subnet
> 4. Because of 3, we need to update our usage everywhere workerProfiles is used.  We always want to reconcile or use the state of the in-cluster machines/machinesets, otherwise this change may have undesired consequences.
> 
> I would say generally we can proceed by:
> 
> 1. Anywhere `workerProfiles` is used, change it to `workerProfilesStatus` if `workerProfilesStatus != nil || len(workerProfilesStatus) > 0`.  Otherwise use `workerProfiles`.
> 2. When we update the azure-cli to use the latest API version, we need to take the intersection of `workerProfiles` and  `workerProfilesStatus` to ensure we have the proper FPSP and CSP RBAC to work.

made changes to the workerProfiles references as suggessted.

pkg/util/clusterdata/worker_profile.go
gouthamMN: Updated workerProfiles references as suggessted by Ben.

pkg/api/util/subnet/subnet.go
```diff
@@ -33,7 +33,14 @@ func NetworkSecurityGroupID(oc *api.OpenShiftCluster, subnetID string) (string,
 		infraID = "aro"
 	}
 	isWorkerSubnet := false
-	for _, s := range oc.Properties.WorkerProfiles {
+	workerProfiles, _ := api.GetEnrichedWorkerProfiles(oc.Properties)
+
+	// Use enriched worker profile data when available
```
gouthamMN, line 38: remove this if condition

petrkotas (approved):
Hi @gouthamMN thank you for the PR. I believe it should work as is, but this has potential to go in regression. Meaning, I approve, but we need at least one more approve before merging. @bennerv ?

gouthamMN:
@bennerv - I have added JIRA item https://issues.redhat.com/browse/ARO-4016 to work on updating the azure-cli to take the intersection of workerProfiles and workerProfilesStatus to ensure we have the proper FPSP and CSP RBAC to work.

s-amann (approved):
lgtm, considering the risk of regression and the importance of workerprofiles, can we deploy this branch to INT, create a cluster and run an az aro update on it, PUCM it, and see what the output of the GetCluster GA is?

gouthamMN:
> lgtm, considering the risk of regression and the importance of workerprofiles, can we deploy this branch to INT, create a cluster and run an az aro update on it, PUCM it, and see what the output of the GetCluster GA is?

As suggested in the comments performed the Regression Tests:
1. Created a ARO clustered from az cli `**az aro create**`.
2. Performed ARO cluster update `**az aro update**`.

From the test it was confirmed there were no changes in the behavior of the existing stable API (v2023-04-01). The new API (v2023-09-04) cannot be tested at this point as the az cli is not updated to use it.

PR #3054, opened 2023-07-21
Add more information to on conflict for gateway PE linkId

bennerv:

Will help identify issues with cluster errors for linkId conflicts.

https://issues.redhat.com/browse/ARO-3750

This will have more verbose error logging for us to be able to connect if a linkId is being reused on cluster installation failures, or if Azure has a bug with linkId reuse.  

### Background

We're still seeing cluster installations fail because of this error message, and we don't log the linkId anywhere so we don't know which cluster it is conflicting with.  This will explicitly error out which cluster has the conflict and what the linkID is, so we can go in and examine if there's a race condition in our code, or if we need to file a bug with the NRP team.


s-amann (approved):
lgtm, :shipit:

PR #3055, opened 2023-07-21
fix API Skeleton file permissions

tony-schndr:

Fixes file permissions on newly generated API Skeleton

No changes, just fixes the file permissions on the generated API Skeleton


bennerv (approved):
does it make sense to have a test for this so we don't accidentally pr it with wrong permissions in the future?

tony-schndr:
> does it make sense to have a test for this so we don't accidentally pr it with wrong permissions in the future?

I think this makes sense, maybe we can run `make client` and then do the git status check like we do with `make generate`?

tony-schndr:
https://github.com/Azure/ARO-RP/blob/master/.pipelines/ci.yml#L90-L95, we could update `make generate` to `make client`.  `make client` already includes the `generate` target

petrkotas (approved):
LGTM

petrkotas:
I agree on the check. I will merge this PR, and we add the check in a new PR

PR #3057, opened 2023-07-24
Fix NSG e2e error

nwnt:
Original discussion thread:
https://redhat-external.slack.com/archives/C03F6AA3HDH/p1689891751922919

Fixes: the spring 239 release failure, ARO-3293

It also includes the changes reverted by #3058

Fix the subtle error found during the e2e test in Dev release.  The error occurred when the new NSG validate function is called for the FP before the flag alternation in `checkPreconfiguredNSG` called by validateSubnets.  In e2e, the subnets will not have any NSG, but the new NSG validation accesses it.  That's why it's giving a nil pointer reference error.

### Test plan for issue:

e2e and CI are now passed.

N/A


pkg/validate/openshiftcluster_validatedynamic.go
```diff
@@ -263,6 +258,11 @@ func (dv *openShiftClusterDynamicValidator) Dynamic(ctx context.Context) error {
 		return err
 	}
 
+	err = fpDynamic.ValidatePreConfiguredNSGs(ctx, dv.oc, subnets)
```
bennerv, line 261: The goal was to group the specific types of validation together.  We should try to keep the service-principal validations together as much as possible.  I would rather move the SP validations to the top than move a singular FP validation to the bottom.

bennerv (changes requested):
Can we elaborate on the issue, why it's an issue, and what this PR does to fix it?

pkg/validate/openshiftcluster_validatedynamic.go
nwnt: Agreed.  I added more description on how the error currently occurs (TLDR: any calls to `ValidatePreconfiguredNSG` needs to happen after `ValidateSubnets`.  This should no longer be the case after GA as the logic in `ValidateSubnets` will go away).  

I moved all the SP validations to the top and FP's lower to keep them grouped.

bennerv (dismissed):
the e2e test `ARO Operator - Azure Subnet Reconciler` needs updating to detect if it's a preconfigured NSG cluster.  If so, we should skip the e2e test for reconciling the NSG.

s-amann (dismissed):

petrkotas (dismissed):
Thanks @nwnt for the update. It seems all is addressed now.

SudoBrendan (dismissed):

petrkotas (approved):
@nwnt thank you for the updates, I believe all comments are addressed now.

test/e2e/operator.go
```diff
@@ -342,6 +342,15 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
 	}
 
 	BeforeEach(func(ctx context.Context) {
+		// TODO remove this when GA
```
petrkotas, line 345: @bennerv I think this is what you were asking for.

PR #3059, opened 2023-07-24
Validate ACR token validity after token password rotation

ventifus:

Fixes [ARO-3018](https://issues.redhat.com/browse/ARO-3018)

When we rotate ACR tokens via PUCM, it is possible for the token to end up in an ambiguous state, possibly due to a race condition. See slack thread for more context: https://redhat-internal.slack.com/archives/C02ULBRS68M/p1689709898669219

### Test plan for issue:

TODO

TODO


pkg/cluster/acrtoken.go
```diff
@@ -72,16 +74,24 @@
 	if err != nil {
 		return err
 	}
+	log = log.WithField("tokenManager", tokenManager)
 
 	registryProfile := tokenManager.GetRegistryProfile(m.doc.OpenShiftCluster)
+	log = log.WithField("registryProfile", registryProfile)
```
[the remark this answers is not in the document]
ventifus: fixed

pkg/cluster/adminupdate_test.go
```diff
@@ -82,7 +82,9 @@ func TestAdminUpdateSteps(t *testing.T) {
 				"[Action fixMCSCert-fm]",
 				"[Action fixMCSUserData-fm]",
 				"[Action ensureGatewayUpgrade-fm]",
+				"[Condition aroDeploymentReady-fm, timeout 5m0s]",
 				"[Action rotateACRTokenPassword-fm]",
+				"[Action validateACRToken-fm]",
```
cadenmarchese, line 87: Since `rotateACRTokenPassword-fm` just updates the secret in the operator namespace and doesn't wait for the operator to add that secret to the global pull secret, we could hit a race condition here if we add a new PUCM step `validateACRToken-fm`, since our operator might not be finished reconciling yet.

I also think that `rotateACRTokenPassword` should confirm that the operator is running and ready before doing anything, since we depend on it.

It might make sense to have it all in the same PUCM task and wait for the operator to reconcile, then validate before returning. It might also make sense to have this validation built in to the operator instead, since we use it to update the cluster pull secret. What are your thoughts?

ventifus: I was hoping to cover the "operator is running and ready" by using the Condition aroDeploymentReady rather than adding a wait-loop to rotateACRTokenPassword, since Conditions already have the ability to timeout. What if I made validateACRToken also a Condition and have it wait for the operator to finish reconciliation?

pkg/util/acrtoken/acrtoken.go
```diff
@@ -142,6 +152,33 @@ func (m *manager) RotateTokenPassword(ctx context.Context, rp *api.RegistryProfi
 	return nil
 }
 
+// ValidateToken attempts to verify that the current ACR token functions
+func (m *manager) ValidateToken(ctx context.Context, rp *api.RegistryProfile) error {
+	creds := clusterauthorizer.Credentials{
+		ClientID:     []byte(rp.Username),
+		ClientSecret: []byte(rp.Password),
```
dem4gus, line 159: It looks like you're using the registry profile that's stored on the cluster doc in CosmosDB to get the token password. I chose not to even worry about updating that password during the rotation process because the only time the cluster cares about that password during the initial cluster installation — that's how the RP passes the username (token name) and password to the cluster during the bootstrap process.

The cluster's "source of truth" for its current ACR password is the `cluster` secret in `openshift-azure-operator`. That's what it uses during the pullsecret controller's reconciliation to ultimately construct the pull secret that nodes use to get container images. The cluster never accesses the password on the cluster document after installation, so I opted not to as well in order to retain a single source of truth for the password.

As far as testing the validity of the new credentials, it would make sense to me to use the exact same credentials the cluster uses in order to minimize potential differences in data. That means getting the secret from the cluster instead of the RP, and deserializing the token stored on it.

pkg/util/acrtoken/acrtoken.go
```diff
@@ -142,6 +152,33 @@ func (m *manager) RotateTokenPassword(ctx context.Context, rp *api.RegistryProfi
...
+		TenantID:     []byte(m.env.TenantID()),
+	}
+	token, err := m.getTokenCredential(m.env.Environment(), &creds)
+	if err != nil {
+		return err
+	}
+
+	client, err := m.newAzAcrClient(m.env, fmt.Sprintf("https://%s", m.env.ACRDomain()), token, nil)
```
dem4gus, line 167: The container registry's URL is stored as part of the .dockerconfigjson string. You should be able to use that to get a connection to the ACR without having to use the environment interface at all.

pkg/cluster/adminupdate_test.go
cadenmarchese: I think that would work!

dem4gus (commented):
I understand the desire to separate the token validation from the pull-secret controller's reconciliation process, which can possibly error for unrelated reasons, but we will need to keep in mind that what we're validating here isn't actually the thing the cluster is using to authenticate to the container registry. Ideally it eventually *becomes* that (or part of it), but there is much more that goes into the pull secret during the controller's reconciliation that could affect the final pull secret. Getting it from here is probably a better option though, because there's no chance of accessing customer secrets that could also be on the pull-secret object.

We will also want to keep in mind that we're never *re*validating the token, which may become a problem if (or when) password expirations are introduced to the process. However, that is something we can revisit when the time comes.

pkg/cluster/adminupdate_test.go
```diff
@@ -130,7 +130,9 @@ func TestAdminUpdateSteps(t *testing.T) {
 				"[Action fixMCSCert-fm]",
 				"[Action fixMCSUserData-fm]",
 				"[Action ensureGatewayUpgrade-fm]",
+				"[Condition aroDeploymentReady-fm, timeout 5m0s]",
 				"[Action rotateACRTokenPassword-fm]",
+				"[Condition ensureValidACRToken-fm, timeout 5m0s]",
```
cadenmarchese, line 135: Do you have thoughts on consolidating these? i.e, convert rotateACRTokenPassword-fm to a Condition with timeout, have your validate path be triggered as part of `rotateACRTokenPassword-fm` instead of a separate step?

pkg/operator/controllers/pullsecret/pullsecret_controller.go
```diff
@@ -192,10 +192,22 @@ func (r *Reconciler) ensureGlobalPullSecret(ctx context.Context, operatorSecret,
...
 		return secret, err
 	}
 
 	err = r.client.Update(ctx, secret)
+	if err != nil {
+		return secret, err
+	}
+
+	rc := pullsecret.NewRegistryClient()
+	err = rc.ValidatePullSecret(ctx, secret)
```
cadenmarchese, line 210: Not exactly related to your change, but we might want to rename this whole function `rotateAndValidateGlobalPullSecret` or something similar since from your change it's clear that we're always going to rotate + validate it.

cadenmarchese: Also, we should update operator statuses (progressing / degraded / etc) when we get an error validating.

pkg/util/pullsecret/pullsecret.go
```diff
@@ -164,3 +169,41 @@ func Build(oc *api.OpenShiftCluster, ps string) (string, error) {
...
+}
+
+func NewRegistryClient() registryClient {
+	return registryClient{
+		checkAuth: dockerregistry.CheckAuth,
+	}
+}
+
+// ValidatePullSecret validates a passed in pull secret by attempting to log in to the registry
+func (r *registryClient) ValidatePullSecret(ctx context.Context, secret *corev1.Secret) error {
```
cadenmarchese, line 184: +1 on keeping this in util - we could potentially configure a monitor that executes this as well if we decide we need that later.

cadenmarchese (commented):
The ValidatePullSecret logic looks good to me! I think the only thing to iron out is when exactly to run it. Right now, were running it on PUCM as well as in the operator, which I think I agree with because we'd get both an operator status and a PUCM failure, but I can see why we'd only really need one or the other since we're always doing the rotation.

pkg/cluster/adminupdate_test.go
ventifus: That makes sense, I'll combine it all into a single condition.

pkg/cluster/acrtoken.go
```diff
@@ -162,8 +176,17 @@ func (m *manager) rotateOpenShiftConfigSecret(ctx context.Context, encodedDocker
...
+	if err != nil {
+		return err
+	}
+
 	return retryOperation(func() error {
 		_, err = m.kubernetescli.CoreV1().Secrets(pullSecretName.Namespace).Apply(ctx, applyConfiguration, metav1.ApplyOptions{FieldManager: "aro-rp", Force: true})
+		if err != nil {
+			return err
+		}
+		err = m.validateACRToken(ctx)
```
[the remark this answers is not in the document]
cadenmarchese: I believe I contradicted myself on this one here: https://github.com/Azure/ARO-RP/pull/3059#discussion_r1379199571 going to resolve this comment.

pkg/cluster/adminupdate_test.go
```diff
@@ -130,7 +130,7 @@ func TestAdminUpdateSteps(t *testing.T) {
 				"[Action fixMCSCert-fm]",
 				"[Action fixMCSUserData-fm]",
 				"[Action ensureGatewayUpgrade-fm]",
-				"[Action rotateACRTokenPassword-fm]",
+				"[Condition rotateAndValidateACRTokenPassword-fm, timeout 5m0s]",
```
cadenmarchese, line 133: Since we aren't waiting on a reconciliation here, do you think we should just make this an `Action` with no timeout instead of leaving it as `Condition`?

pkg/util/pullsecret/pullsecret.go
```diff
@@ -164,3 +169,41 @@ func Build(oc *api.OpenShiftCluster, ps string) (string, error) {
...
+		}
+
+		auth := strings.SplitN(string(authDecoded), ":", 2)
+		if len(auth) != 2 {
+			err = fmt.Errorf("credentials format error: %s", registry)
+			return err
+		}
+		if err != nil {
+			return err
+		}
```
cadenmarchese, line 202: Redundant? It looks like we already returned any possible error on line 198.

pkg/util/pullsecret/pullsecret.go
```diff
@@ -164,3 +169,41 @@ func Build(oc *api.OpenShiftCluster, ps string) (string, error) {
...
+	for registry, authBase64 := range dockerConfig {
+		authDecoded, err := base64.StdEncoding.DecodeString(authBase64)
+		if err != nil {
+			return err
+		}
+
+		auth := strings.SplitN(string(authDecoded), ":", 2)
+		if len(auth) != 2 {
+			err = fmt.Errorf("credentials format error: %s", registry)
+			return err
```
cadenmarchese, line 198: Nit - might as well consolidate these lines since you did it on line 205:
```suggestion
			return fmt.Errorf("credentials format error: %s", registry)
```

cadenmarchese (commented):
Few small comments, but for the most part, looking good!

pkg/util/pullsecret/pullsecret.go
```diff
@@ -164,3 +169,41 @@ func Build(oc *api.OpenShiftCluster, ps string) (string, error) {
...
+	}
+}
+
+// ValidatePullSecret validates a passed in pull secret by attempting to log in to the registry
+func (r *RegistryClient) ValidatePullSecret(ctx context.Context, secret *corev1.Secret) error {
+	dockerConfig, err := UnmarshalSecretData(secret)
+	if err != nil {
+		return err
+	}
+	for registry, authBase64 := range dockerConfig {
```
dem4gus, line 189: This will check every registry in openshift-config/pull-secret, including user-defined credentials. Are we okay with doing that? And if we are, errors returned from this loop will short-circuit the validation. For example, if there is a pull-secret with both a user-defined registry and the ARO ACR defined and the user secret is checked first and has some sort of error, this will early return on that error and the ARO ACR token will never be validated. I don't know how we feel about using goroutines to check credentials concurrently, but since one credential's validity shouldn't have a bearing on another cred that may be a solution. Alternatively, we could extract the ARO ACR token and only check that one.

pkg/util/pullsecret/pullsecret.go
dem4gus: +1

pkg/util/pullsecret/pullsecret_test.go
```diff
@@ -373,3 +376,79 @@ func TestUnmarshalSecretData(t *testing.T) {
...
+			return nil
+		},
+	}
+	test := []struct {
+		name     string
+		ps       *corev1.Secret
+		wantAuth map[string]string
+		wantErr  string
+		client   RegistryClient
+	}{
```
dem4gus, line 398: If we're going to validate all tokens in the secret then we'll need a test case for a valid ACR token being checked after an invalid user token. I don't remember if go slices are guaranteed to be handled in order in a `for` loop but it would be good at least attempt it.

dem4gus (commented):
Some questions about which tokens we're validating in the pull secret.

pkg/cluster/install.go
```diff
@@ -102,7 +102,7 @@ func (m *manager) adminUpdate() []steps.Step {
 	if isEverything {
 		toRun = append(toRun,
 			steps.Action(m.ensureGatewayUpgrade),
-			steps.Action(m.rotateACRTokenPassword),
+			steps.Condition(m.rotateAndValidateACRTokenPassword, 5*time.Minute, true),
```
kimorris27, line 105: Don't we need to ensure the operator is running before moving forward with the rotation? You can make sure the operator is ready using the `aroDeploymentReady` Condition function.

pkg/cluster/install.go
```diff
@@ -195,7 +195,7 @@ func (m *manager) Update(ctx context.Context) error {
 		steps.Action(m.createOrUpdateDenyAssignment),
 		steps.Action(m.startVMs),
 		steps.Condition(m.apiServersReady, 30*time.Minute, true),
-		steps.Action(m.rotateACRTokenPassword),
+		steps.Condition(m.rotateAndValidateACRTokenPassword, 5*time.Minute, true),
```
kimorris27, line 198: Same question/comment as in `adminUpdate`.

pkg/util/pullsecret/pullsecret.go
kimorris27: I'm +1 to extracting the ARO ACR token for validation and disregarding the other tokens.

mociarain:
What's the current state of this and can I help get it moving again?

PR #3060, opened 2023-07-25
Unplanned Maintenance Signals

niontive:

Fixes https://issues.redhat.com/browse/ARO-1933

For resource health check, we have to emit a metric when we're doing unplanned maintenance on a cluster.  This code changes accomplishes this by:

1. Adding a maintenance signal middleware. For now, this just emits the "unplanned maintenance" signal.
2. Using that middleware for write admin operations (ex: start VM).

### Test plan for issue:

- Unit tests
- INT test

No - everything is automated


pkg/frontend/middleware/maintenance.go
```diff
@@ -0,0 +1,51 @@
...
+
+		// Use a do-while loop to ensure we emit the metric at least once
+		mm.emitMaintenanceSignal("unplanned", resourceID)
+		go func(ctx context.Context, resourceID string) {
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+					time.Sleep(1 * time.Minute)
```
niontive, line 36: Check using time.after instead of time.sleep

pkg/frontend/middleware/maintenance.go
```diff
@@ -0,0 +1,51 @@
...
+// Emit metric for unplanned maintenance
+func (mm MaintenanceMiddleware) UnplannedMaintenanceSignal(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx, cancel := context.WithCancel(r.Context())
+		defer cancel()
+
+		resourceID := strings.TrimPrefix(filepath.Dir(r.URL.Path), "/admin")
+
+		// Use a do-while loop to ensure we emit the metric at least once
+		mm.emitMaintenanceSignal("unplanned", resourceID)
```
cadenmarchese, line 29: I'm wondering if we want to emit a second metric for planned maintenance as well, or just have a single maintenance signal that emits when maintenance is occurring generally (regardless of if planned / unplanned)

pkg/frontend/frontend.go
```diff
@@ -320,23 +317,23 @@ func (f *frontend) chiAuthenticatedRoutes(router chi.Router) {
...
-				r.Post("/startvm", f.postAdminOpenShiftClusterStartVM)
+				r.With(f.maintenanceMiddleware.UnplannedMaintenanceSignal).Post("/startvm", f.postAdminOpenShiftClusterStartVM)
 
-				r.Post("/upgrade", f.postAdminOpenShiftUpgrade)
+				r.With(f.maintenanceMiddleware.UnplannedMaintenanceSignal).Post("/upgrade", f.postAdminOpenShiftUpgrade)
 
 				r.Get("/skus", f.getAdminOpenShiftClusterVMResizeOptions)
 
-				r.Post("/resize", f.postAdminOpenShiftClusterVMResize)
+				r.With(f.maintenanceMiddleware.UnplannedMaintenanceSignal).Post("/resize", f.postAdminOpenShiftClusterVMResize)
```
cadenmarchese, line 330: My only suggestion would be to not emit an unplanned maintenance signal for the resize endpoint, since this is always something we come to agreement on with the customer before performing. This might change in the future, but for now the resize is planned maintenance only.

pkg/frontend/middleware/maintenance.go
```diff
@@ -0,0 +1,50 @@
...
+		}(ctx, resourceID)
+
+		h.ServeHTTP(w, r)
+	})
+}
+
+func (mm MaintenanceMiddleware) emitMaintenanceSignal(maintenanceType, resourceID string) {
+	maintenanceMetric := "frontend.maintenance." + maintenanceType
+	mm.EmitGauge(maintenanceMetric, 1, map[string]string{
+		"resourceID": resourceID,
```
s-amann, line 48: please change to `resourceId` as we have set it up this way for cluster monitor metrics here: https://github.com/Azure/ARO-RP/blob/6cfcd6eb324806efede2d8f18399d53580c558b8/pkg/monitor/cluster/cluster.go#L63

pkg/frontend/middleware/maintenance_test.go
```diff
@@ -0,0 +1,53 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			controller := gomock.NewController(t)
+			defer controller.Finish()
+
+			m := mock_metrics.NewMockEmitter(controller)
+
+			maintenanceMiddleware := MaintenanceMiddleware{m}
+
+			m.EXPECT().EmitGauge("frontend.maintenance.unplanned", int64(1), map[string]string{
+				"resourceID": tt.resourceID,
```
s-amann, line 37: this test needs to be updated to reflect the change to `resourceId`

pkg/frontend/middleware/maintenance.go
niontive: I agree having one signal "source" is ideal. The main issue is unplanned maintenance only uses the synchronous, frontend operations. PUCM uses asynchronous, backend operations.

cadenmarchese (approved):

jhoreman (approved):

niontive:
Rebased with master due to merge conflict.

cadenmarchese (approved):

PR #3006, opened 2023-07-04
Add v20230701preview API changes for NSG Enhancement

nwnt:
### Which issue this PR addresses:

Fixes ARO-3494

### What this PR does / why we need it:

Add a new API field called preconfiguredNSG for NSG Enhancement

### Test plan for issue:

Tested using the local development environment.  Could see that the new field was sent and received properly with the changes included in the PR.

### Is there any documentation that needs to be updated for this PR?

N/A

PR #3062, opened 2023-07-26
Add nsg to api

nwnt:

Fixes [ARO-3494](https://issues.redhat.com/browse/ARO-3494)

With this, #3006 will be abandoned because we no longer add the changes for 2023-07-01-preview.

Add a new API field called preconfiguredNSG for NSG Enhancement

Test plan for issue:
Tested using the local development environment. Could see that the new field was sent and received properly with the changes included in the PR.

N/A


swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/stable/2023-09-04/redhatopenshift.json
```diff
@@ -2224,6 +2228,14 @@
         "modelAsString": true
       }
     },
+    "PreconfiguredNSG": {
```
bennerv, line 2231: I think we want to use the x-ms-enum value here as well, similar to OUtboundType.  I can't remember why this is the way it is, but i think there's a good reason.

nwnt: Agreed.  Made the required changes and regenerated to have `x-ms-enum` included.

bennerv (approved):
lgtm

PR #3063, opened 2023-07-26
[Draft] 4.12 cluster installation test from CI/PR

tiguelu:
https://issues.redhat.com/browse/ARO-3198

This PR is a draft in order to test 4.12 cluster creation and E2E from PRs. This is **NOT** intended to be merged. Please disregard.

### DIFF
--- a/pkg/util/version/const.go
+++ b/pkg/util/version/const.go
@@ -58,6 +58,10 @@ var HiveInstallStreams = []*Stream{
 		Version:  NewVersion(4, 11, 26),
 		PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:1c3913a65b0a10b4a0650f54e545fe928360a94767acea64c0bd10faa52c945a",
 	},
+	{
+		Version:  NewVersion(4, 12, 24),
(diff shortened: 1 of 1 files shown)


pkg/util/version/const.go
```diff
@@ -38,6 +38,10 @@ var DefaultInstallStreams = map[int]*Stream{
 		Version:  NewVersion(4, 11, 44),
 		PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:52cbfbbeb9cc03b49c2788ac7333e63d3dae14673e01a9d8e59270f3a8390ed3",
 	},
+	12: {
+		Version:  NewVersion(4, 12, 17),
+		PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:7ca5f8aa44bbc537c5a985a523d87365eab3f6e72abc50b7be4caae741e093f4",
+	},
```
bennerv, line 44: I know it's not meant to be merged, but can we set the min version in 4.12 we're installing to 4.12.23+?  The FIPS fixes are in that version and beyond so we shouldn't offer any lower version as an install target.

tiguelu: Changed to 4.12.25

PR #3065, opened 2023-07-26
Dynamic validation of "EncryptionAtHost" feature subscription level registration at the RP. [DO NOT MERGE]

schiruma:

[ARO-3211](https://issues.redhat.com/browse/ARO-3211)

When customers are attempting to set EncryptionAtHost, but that feature is not enabled for their Subscriptions, an error message describing about the issue and its resolution is displayed to them. This is achieved by dynamically validating their Subscription documents.

### Test plan for issue:

E2E

N/A


pkg/api/validate/dynamic/encryptionathost.go
```diff
@@ -46,3 +53,20 @@ func (dv *dynamic) validateEncryptionAtHostSupport(VMSize api.VMSize, path strin
 
 	return nil
 }
+
+func (dv *dynamic) IsRegisteredForFeature(sub *api.SubscriptionProperties) error {
```
tsatam, line 57: minor nit:

The function name here doesn't make it clear _which_ feature we are checking for registration for. Given that it's a method on the generic dynamic validator, I would expect it to either be named specifically for the EncryptionAtHost feature (e.g. `IsSubRegisteredForEncryptionAtHostFeature`), or take in the feature name as a parameter if there are places to reuse it.

pkg/api/validate/dynamic/encryptionathost_test.go
```diff
@@ -19,30 +19,85 @@ import (
...
+				},
+				WorkerProfiles: []api.WorkerProfile{{
+					EncryptionAtHost: api.EncryptionAtHostEnabled,
+					VMSize:           WorkerVMSize,
+				}},
+			},
+		}
+	}
+
+	EncryptionAtHostEnabledOrDisabled := func(MasterEncryptionAtHost api.EncryptionAtHost, WorkerEncryptionAtHost api.EncryptionAtHost) *api.OpenShiftCluster {
```
tsatam, line 49: I like the refactor here to keep the actual test cases focused on their intent, but I think the function name and syntax here is a little confusing and ends up obscuring some important information when looking at the test cases themselves, e.g. it's unclear from just reading `EncryptionAtHostEnabledOrDisabled(api.EncryptionAtHostEnabled, api.EncryptionAtHostDisabled)` what that test case is trying to set up. 

I wonder if it might be cleaner to establish static master/workerprofile definitions, and have each test case assemble the cluster definition by composing those static members (e.g. `func(masterProfile api.MasterProfile, workerProfile api.WorkerProfile)`), or to just make a single, four-arg function that handles both cases.

bennerv (changes requested):
This won't actually work.  

The subscription document only holds features registered under the RP's namespace (Microsoft.RedHatOpenShift) and the Microsoft.Resources namespaces.  

>All AFEC features that the subscriptions has been registered under RP namespace and platform namespace (Microsoft.Resources).

bennerv (changes requested):
Quick summary of actions needed to be taken:
1. Update our FPSP Built-In Service Role to include `Microsoft.Features/features/read` and `Microsoft.Features/providers/features/read`
2. The above follows SDP
3. Update our local custom role in our dev environment to include the above permissions as well
4. Update this PR to use the features client instead of reading from the subscription document
5. Roll it out :shipit:

PR #3066, opened 2023-07-27
Don't use the uid in containerinstall code

hawkowl:

Fixes the need for further docs in https://issues.redhat.com/browse/ARO-3710

It doesn't matter if this runs as 0, since rootless podman should do it fine.

### Test plan for issue:

CI/E2E

N/A


ellis-johnson (dismissed):
LGTM!

ellis-johnson (approved):
LGTM! (again)

PR #2965, opened 2023-06-14
Propagate errors of ARO DNSMasqMachineConfig controller

SrinivasAtmakuri:
JIRA: https://issues.redhat.com/browse/ARO-3254

### Which issue this PR addresses:

Fixes

PR #3068, opened 2023-07-27
Aro 2998 monitoring errors

jhoreman:
https://issues.redhat.com/browse/ARO-2998

Aggregate errors of the monitoring controller through the aro controller. Precedent and reference here:

https://github.com/SrinivasAtmakuri/ARO-RP/blob/b025e278676744aeea0a01095c7854bf2f[…]9c/pkg/operator/controllers/dnsmasq/machineconfig_controller.go
10:31
https://github.com/Azure/ARO-RP/pull/2965/files

### Test plan for issue:

Local tests pass, error propagation will happen and be checked in INT

No

### DIFF
--- a/pkg/operator/controllers/monitoring/monitoring_controller.go
+++ b/pkg/operator/controllers/monitoring/monitoring_controller.go
@@ -25,10 +25,11 @@ import (
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+	"github.com/Azure/ARO-RP/pkg/operator/controllers/base"
 )
 
 const (
-	ControllerName = "Monitoring"
+	MonitoringControllerName = "Monitoring"
 
 	controllerEnabled = "aro.monitoring.enabled"
 )
@@ -55,7 +56,7 @@ type Config struct {
 }
 
 type Reconciler struct {
-	log *logrus.Entry
+	base.AROController
 
 	client client.Client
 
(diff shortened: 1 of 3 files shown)


pkg/operator/controllers/monitoring/monitoring_controller.go
```diff
@@ -54,52 +55,54 @@ type Config struct {
...
 	}
 }
 
-func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
-	instance := &arov1alpha1.Cluster{}
-	err := r.client.Get(ctx, types.NamespacedName{Name: arov1alpha1.SingletonClusterName}, instance)
+func (r *MonitoringReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	instance, err := r.GetCluster(ctx)
 	if err != nil {
 		return reconcile.Result{}, err
```
cadenmarchese, line 78: Why not `SetDegraded` here? I noticed this is a convention in all of the controllers.

To me, it makes sense for the operator to report degraded as soon as it isn't able to Get the cluster CR for any reason. I'm curious why we decided to not do this. But if we decided to do it, it would probably be in a separate PR, so I'm good to approve this one.

cadenmarchese (approved):
Very straightforward change, thanks! I have question about how we handle degraded statuses when we can't get the cluster CR.

petrkotas (approved):
LGTM, we can talk about the degraded state in next arch call. Putting it to the agenda.

pkg/operator/controllers/monitoring/monitoring_controller.go
tsatam: The simple reason is that these controller statuses are saved on the cluster CR itself, so if we're unable to retrieve it we'd be unable to set statuses anyways.

PR #3069, opened 2023-07-27
ARO-3112: Set DisableOutbound SNAT to true on Load Balancers

lranjbar:

Fixes ARO-3112

Adds a controller the ARO operator to watch and maintain the `cloud-provider-config` in the `openshift-config` namespace.

As a part of the roll out Public IP functionality changes we need to set the setting disableOutboundSNAT to true on load balancers. This can be managed using the `cloud-provider-config`.

### Test plan for issue:

Unit tests and e2e tests were added for this controller.
Functional testing was done by doing the following:

1. Created a new ARO cluster in Azure
2. Ran the operator locally against this new cluster with different settings in `cloud-provider-config`

Tested the following scenarios:
- cloud-provider-config has `disableOutboundSNAT: null` -> Set to true
- cloud-provider-config has `disableOutboundSNAT: false` -> Set to true
- cloud-provider-config has `disableOutboundSNAT: true` -> No change
- cloud-provider-config has the this value changed from true to false while the controller is running -> Set to true

Yes. Updating the `cloud-provider-config` causes a graceful node reboot of all OpenShift nodes. We should have a doc writer add docs to inform the customer that changing this setting will cause two graceful reboots. First when they manually change it and second when this is remediated.

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -99,5 +99,6 @@ func DefaultOperatorFlags() OperatorFlags {
 		"rh.srep.muo.managed":                      flagTrue,
 		"aro.guardrails.enabled":                   flagFalse,
 		"aro.guardrails.deploy.managed":            flagFalse,
+		"aro.cloudproviderconfig.enabled":          flagTrue,
 	}
 }
--- a/pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
+++ b/pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
@@ -1,0 +1,259 @@
+package cloudproviderconfig
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+	"github.com/Azure/ARO-RP/pkg/operator/controllers/base"
+)
+
+const (
+	ControllerName    = "CloudProviderConfig"
+	controllerEnabled = "aro.cloudproviderconfig.enabled"
+)
+
+var cloudProviderConfigName = types.NamespacedName{Name: "cloud-provider-config", Namespace: "openshift-config"}
+
+// / This is a very old version of the CloudProviderConfig struct from k8s upstream
+// / OpenShift diverged from the upstream cluster API and we have a depreciated version of this struct
+// / I was unable to use the recent version of this struct in the upstream k8s Azure Cloud Provider
+// / This version of the struct can be found here:
+// / https://github.com/kubernetes/kubernetes/blob/v1.13.5/pkg/cloudprovider/providers/azure/azure.go#L81
+type azCloudProviderConfig struct {
+	// The cloud environment identifier. Takes values from https://github.com/Azure/go-autorest/blob/ec5f4903f77ed9927ac95b19ab8e44ada64c1356/autorest/azure/environments.go#L13
+	Cloud string `json:"cloud" yaml:"cloud"`
+	// The AAD Tenant ID for the Subscription that the cluster is deployed in
+	TenantID string `json:"tenantId" yaml:"tenantId"`
+	// The ClientID for an AAD application with RBAC access to talk to Azure RM APIs
+	AADClientID string `json:"aadClientId" yaml:"aadClientId"`
+	// The ClientSecret for an AAD application with RBAC access to talk to Azure RM APIs
+	AADClientSecret string `json:"aadClientSecret" yaml:"aadClientSecret"`
+	// The path of a client certificate for an AAD application with RBAC access to talk to Azure RM APIs
+	AADClientCertPath string `json:"aadClientCertPath" yaml:"aadClientCertPath"`
+	// The password of the client certificate for an AAD application with RBAC access to talk to Azure RM APIs
+	AADClientCertPassword string `json:"aadClientCertPassword" yaml:"aadClientCertPassword"`
+	// Use managed service identity for the virtual machine to access Azure ARM APIs
+	UseManagedIdentityExtension bool `json:"useManagedIdentityExtension" yaml:"useManagedIdentityExtension"`
+	// UserAssignedIdentityID contains the Client ID of the user assigned MSI which is assigned to the underlying VMs. If empty the user assigned identity is not used.
+	// More details of the user assigned identity can be found at: https://docs.microsoft.com/en-us/azure/active-directory/managed-service-identity/overview
+	// For the user assigned identity specified here to be used, the UseManagedIdentityExtension has to be set to true.
+	UserAssignedIdentityID string `json:"userAssignedIdentityID" yaml:"userAssignedIdentityID"`
+	// The ID of the Azure Subscription that the cluster is deployed in
+	SubscriptionID string `json:"subscriptionId" yaml:"subscriptionId"`
+	// ResourceManagerEndpoint is the cloud's resource manager endpoint. If set, cloud provider queries this endpoint
+	// in order to generate an autorest.Environment instance instead of using one of the pre-defined Environments.
+	ResourceManagerEndpoint string `json:"resourceManagerEndpoint,omitempty" yaml:"resourceManagerEndpoint,omitempty"`
+	// The name of the resource group that the cluster is deployed in
+	ResourceGroup string `json:"resourceGroup" yaml:"resourceGroup"`
+	// The location of the resource group that the cluster is deployed in
+	Location string `json:"location" yaml:"location"`
+	// The name of the VNet that the cluster is deployed in
+	VnetName string `json:"vnetName" yaml:"vnetName"`
+	// The name of the resource group that the Vnet is deployed in
+	VnetResourceGroup string `json:"vnetResourceGroup" yaml:"vnetResourceGroup"`
+	// The name of the subnet that the cluster is deployed in
+	SubnetName string `json:"subnetName" yaml:"subnetName"`
+	// The name of the security group attached to the cluster's subnet
+	SecurityGroupName string `json:"securityGroupName" yaml:"securityGroupName"`
+	// (Optional in 1.6) The name of the route table attached to the subnet that the cluster is deployed in
+	RouteTableName string `json:"routeTableName" yaml:"routeTableName"`
+	// (Optional) The name of the availability set that should be used as the load balancer backend
+	// If this is set, the Azure cloudprovider will only add nodes from that availability set to the load
+	// balancer backend pool. If this is not set, and multiple agent pools (availability sets) are used, then
+	// the cloudprovider will try to add all nodes to a single backend pool which is forbidden.
+	// In other words, if you use multiple agent pools (availability sets), you MUST set this field.
+	PrimaryAvailabilitySetName string `json:"primaryAvailabilitySetName" yaml:"primaryAvailabilitySetName"`
+	// The type of azure nodes. Candidate values are: vmss and standard.
+	// If not set, it will be default to standard.
+	VMType string `json:"vmType" yaml:"vmType"`
+	// The name of the scale set that should be used as the load balancer backend.
+	// If this is set, the Azure cloudprovider will only add nodes from that scale set to the load
+	// balancer backend pool. If this is not set, and multiple agent pools (scale sets) are used, then
+	// the cloudprovider will try to add all nodes to a single backend pool which is forbidden.
+	// In other words, if you use multiple agent pools (scale sets), you MUST set this field.
+	PrimaryScaleSetName string `json:"primaryScaleSetName" yaml:"primaryScaleSetName"`
+	// Enable exponential backoff to manage resource request retries
+	CloudProviderBackoff bool `json:"cloudProviderBackoff" yaml:"cloudProviderBackoff"`
+	// Backoff retry limit
+	CloudProviderBackoffRetries int `json:"cloudProviderBackoffRetries" yaml:"cloudProviderBackoffRetries"`
+	// Backoff exponent
+	CloudProviderBackoffExponent float64 `json:"cloudProviderBackoffExponent" yaml:"cloudProviderBackoffExponent"`
+	// Backoff duration
+	CloudProviderBackoffDuration int `json:"cloudProviderBackoffDuration" yaml:"cloudProviderBackoffDuration"`
+	// Backoff jitter
+	CloudProviderBackoffJitter float64 `json:"cloudProviderBackoffJitter" yaml:"cloudProviderBackoffJitter"`
+	// Enable rate limiting
+	CloudProviderRateLimit bool `json:"cloudProviderRateLimit" yaml:"cloudProviderRateLimit"`
+	// Rate limit QPS (Read)
+	CloudProviderRateLimitQPS float32 `json:"cloudProviderRateLimitQPS" yaml:"cloudProviderRateLimitQPS"`
+	// Rate limit Bucket Size
+	CloudProviderRateLimitBucket int `json:"cloudProviderRateLimitBucket" yaml:"cloudProviderRateLimitBucket"`
+	// Rate limit QPS (Write)
+	CloudProviderRateLimitQPSWrite float32 `json:"cloudProviderRateLimitQPSWrite" yaml:"cloudProviderRateLimitQPSWrite"`
+	// Rate limit Bucket Size
+	CloudProviderRateLimitBucketWrite int `json:"cloudProviderRateLimitBucketWrite" yaml:"cloudProviderRateLimitBucketWrite"`
+
+	// Use instance metadata service where possible
+	UseInstanceMetadata bool `json:"useInstanceMetadata" yaml:"useInstanceMetadata"`
+
+	// Sku of Load Balancer and Public IP. Candidate values are: basic and standard.
+	// If not set, it will be default to basic.
+	LoadBalancerSku string `json:"loadBalancerSku" yaml:"loadBalancerSku"`
+	// ExcludeMasterFromStandardLB excludes master nodes from standard load balancer.
+	// If not set, it will be default to true.
+	ExcludeMasterFromStandardLB *bool `json:"excludeMasterFromStandardLB" yaml:"excludeMasterFromStandardLB"`
+	// DisableOutboundSNAT disables the outbound SNAT for public load balancer rules.
+	// It should only be set when loadBalancerSku is standard. If not set, it will be default to false.
+	DisableOutboundSNAT *bool `json:"disableOutboundSNAT" yaml:"disableOutboundSNAT"`
+
+	// Maximum allowed LoadBalancer Rule Count is the limit enforced by Azure Load balancer
+	MaximumLoadBalancerRuleCount int `json:"maximumLoadBalancerRuleCount" yaml:"maximumLoadBalancerRuleCount"`
+}
+
+// CloudProviderConfigReconciler reconciles the openshift-config/cloud-provider-config ConfigMap
+type CloudProviderConfigReconciler struct {
+	base.AROController
+}
+
+func NewReconciler(Log *logrus.Entry, client client.Client) *CloudProviderConfigReconciler {
+	return &CloudProviderConfigReconciler{
+		AROController: base.AROController{
+			Log:    Log,
+			Client: client,
+			Name:   ControllerName,
+		},
+	}
+}
+
+// Reconcile makes sure that the cloud-provider-config is healthy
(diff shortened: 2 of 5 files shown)


petrkotas (changes requested):
HI @lranjbar thank you for the PR. I have a humble ask. Would you run the make generate and update the PR so it the test can pass?

lranjbar:
> HI @lranjbar thank you for the PR. I have a humble ask. Would you run the make generate and update the PR so it the test can pass?

Thanks for taking a look at the draft PR.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -0,0 +1,100 @@
...
+		Named(ControllerName).
+		Complete(r)
+}
+
+func (r *CloudProviderConfigReconciler) updateCloudProviderConfig(ctx context.Context) error {
+	cm := &corev1.ConfigMap{}
+	err := r.Client.Get(ctx, cloudProviderConfigName, cm)
+	if err != nil {
+		return fmt.Errorf("the ConfigMap openshift-config/cloud-provider-config is missing")
+	}
```
SudoBrendan, line 80: It's probably best to return the real error instead of masking. We can get more than 404 here. If we want to special case 404, check out the `kerrors.IsNotFound(err)` helper.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -0,0 +1,100 @@
...
+
+	jsonConfig, ok := cm.Data["config"]
+	if !ok {
+		return fmt.Errorf("field config in ConfigMap openshift-config/cloud-provider-config is missing")
```
SudoBrendan, line 84: Again, I think I'd rather not mask errors completely, can we append the original error to the new one?

pkg/util/jsonutils/jsonutils.go
```diff
@@ -0,0 +1,38 @@
...
+		// check for key in map
+		if val, ok := jsonMap[k]; ok {
+			// update existing key
+			if val != v {
+				jsonMap[k] = v
+				changed = true
+			}
+		} else {
+			// key does not exist
+			return jsonString, changed, fmt.Errorf("key does not exist in json string")
```
SudoBrendan, line 31: probably best to enrich the error with the key that was missing.

pkg/util/jsonutils/jsonutils_test.go
```diff
@@ -0,0 +1,48 @@
...
+	type test struct {
+		name             string
+		jsonString       string
+		jsonKeyValuePair map[string]string
+		expectedString   string
+		expectedChanged  bool
+	}
+	jsonStringByte, _ := json.Marshal(map[string]string{"test": "test"})
+
+	for _, tt := range []*test{
```
SudoBrendan, line 21: It would be good to include some cases for error cases as well (can't parse json, key did not exist)

SudoBrendan (changes requested):
Thanks for the PR! It's looking pretty good, I'd like to see some improvements on errors and handling to help us troubleshoot in prod.

SudoBrendan:
One more idea - it would probably be good to have an E2E test for this that we revert changes to this value. See https://github.com/Azure/ARO-RP/blob/master/test/e2e/operator.go#L158 for an example.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	updateMap := map[string]string{"disableOutboundSNAT": "true"}
```
tony-schndr, line 87: From my understanding this will be encoded to a json string type like this `\"disableOutboundSNAT\": \"true\",`?  The underlying config map uses a bool value for this field and when testing the field as a string the load balancing rules do not get updated like they would if this was a json bool.  Is there a way to encode this as a boolean or maybe use some struct from upstream so we can marshal/unmarshal this data?

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -0,0 +1,100 @@
...
+	changed := false
+	cm.Data["config"], changed, err = jsonutils.UpdateJsonString(jsonConfig, updateMap)
+	if err != nil {
+		return err
+	}
+
+	if changed {
+		r.Log.Debug("openshift-config/cloud-provider-config was updated")
```
kimorris27, line 95: Could we consider making this `Info` instead of `Debug` (or whatever level we would need to make it for it to show up without having to adjust operator's log level)? It would be nice to see when this change is made, and I think it would happen infrequently enough that we wouldn't end up with an excessive number of these messages.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
lranjbar: The marshalling and unmarshalling of the data is happening in the utility function. This data is stored as a json string overall.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
lranjbar: I agree it shouldn't happen to frequently.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
tony-schndr: Maybe I'm missing something but from my testing this has to be rendered as a boolean in the config map like `\"disableOutboundSNAT\": true`.  When I tested this as a string like `\"disableOutboundSNAT\": \"true\"` there was an issue with the updating the nodes and the nodes don't recover.  Maybe there is a struct upstream that we can use to marshal/unmarshal?

lranjbar: The config map has the set up of:

```json
{ "config": { "disableOutboundSNAT": "true" } }
```

lranjbar: FYI the upstream struct is missing this boolean setting.

https://github.com/openshift/cluster-api-provider-azure/blob/master/controllers/helpers.go#L318

lranjbar: I did more research into this particular issue today. Even using a struct from the k8s upstream it does not unmarshal into a bool.

See: https://github.com/lranjbar/ARO-RP/pull/1

lranjbar: Turns out we have a depreciated version of this struct in OpenShift in general. I had to paste in the struct from old upstream it get it work. All of the above conversation is moot. 🤦🏼‍♀️

jhoreman:
Ran checks.

petrkotas (approved):
@lranjbar  thank you for the work, I have one non-blocking question.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -0,0 +1,239 @@
...
+)
+
+const (
+	ControllerName    = "CloudProviderConfig"
+	controllerEnabled = "aro.cloudproviderconfig.enabled"
+)
+
+var cloudProviderConfigName = types.NamespacedName{Name: "cloud-provider-config", Namespace: "openshift-config"}
+
+// / This is a very old version of the CloudProviderConfig struct from k8s upstream
```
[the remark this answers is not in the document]
lranjbar: Given the history I'm not sure if there is much risk in this breaking in a few releases. 

This is more of a question of when OpenShift updates this API and we would want to pivot to use the new API. The other thing to note is that in the version of this struct we have in the OpenShift upstream, there are fields missing from that struct. Which would allow us to use the upstream k8s API. So it would be possible to bring in the upstream k8s API at that point. The missing fields are included in this struct I've provided here so it wont actually break.

To be clear this fork and divergence happened a long time ago... some time in 2018 from what I can gather. There is a large update to the cluster API going on in OpenShift right now so it is possible they update it soon but it might just stay the same too.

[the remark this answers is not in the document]
lranjbar: Also if it isn't clear in the reply above the OpenShift cluster API actually moonlights at the upstream k8s API. The reason I can't just pull in the upstream k8s API is because the imports get confused since they both identify as the same Golang import module.

This works as expected when the OpenShift cluster API and the upstream k8s API are in sync but since the OpenShift cluster API is missing fields it causes errors when unmarshalling the JSON even if I pull in a newer version of k8s API that includes the fields.

bennerv:
Removing /hold so we can merge on this release.  

We can merge, but shouldn't PUCM clusters with the feature flag to enable the controller until we work with BU for dates on PUCM.  

Otherwise we can enable for new clusters out of the box.

pkg/api/defaults.go
```diff
@@ -99,5 +99,6 @@ func DefaultOperatorFlags() OperatorFlags {
 		"rh.srep.muo.managed":                      flagTrue,
 		"aro.guardrails.enabled":                   flagFalse,
 		"aro.guardrails.deploy.managed":            flagFalse,
+		"aro.cloudproviderconfig.enabled":          flagTrue,
```
SudoBrendan, line 102: @bennerv based on our discussion, does this need to be `flagFalse` to merge?

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -0,0 +1,239 @@
...
+
+func (r *CloudProviderConfigReconciler) updateCloudProviderConfig(ctx context.Context) error {
+	r.Log.Debug("checking openshift-config/cloud-provider-config")
+
+	cm := &corev1.ConfigMap{}
+	err := r.Client.Get(ctx, cloudProviderConfigName, cm)
+	if err != nil {
+		if kerrors.IsNotFound(err) {
+			r.Log.Debug("the ConfigMap cloud-provider-config was not found in the openshift-config namespace")
+		}
```
SudoBrendan, line 195: nit: based on the test suite, this error (CM not found) is already resulting in an Error-level log message, do we need the Debug level one as well, or can this be removed?

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -0,0 +1,239 @@
...
+	changed := false
+	var cpc azCloudProviderConfig
+	err = json.Unmarshal([]byte(jsonConfig), &cpc)
+	if err != nil {
+		return err
+	}
+
+	if cpc.DisableOutboundSNAT != nil && !*cpc.DisableOutboundSNAT {
+		r.Log.Info("openshift-config/cloud-provider-config disableOutboundSNAT is set to false")
+		r.Log.Info("updating openshift-config/cloud-provider-config disableOutboundSNAT to true")
```
SudoBrendan, line 213: nit: every log message of every container we manage in every cluster goes through the Gateway, a regional service, so we try to optimize bits in logs as much as we can. Can we compress these into one message, something like "updating disableOutboundSNAT from false to true"? Same for below?

SudoBrendan (commented):
A very lovely PR, especially the tests! Thank you! I added one comment based on some Slack threads we have for this. The PR LGTM outside the `defaults.go`, which I just have a question on :)

pkg/api/defaults.go
lranjbar: My understanding is that this should only be for new cluster installs by default because its a new flag (?)

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
lranjbar: The error message that it emits doesn't specify the namespace the CM is in, so I left this in on the debug channel. We can remove it, just feel it enriches the debugging experience a little bit.

pkg/api/defaults.go
bennerv: True should be fine; it'll be set on new clusters only, and when we PUCM existing clusters we'll patch in their flag to true since it won't be set.

SudoBrendan (approved):

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -0,0 +1,236 @@
...
+	} else {
+		r.Log.Debug("openshift-config/cloud-provider-config disableOutboundSNAT is set to true no changes needed")
+	}
+
+	jsonStringByte, err := json.Marshal(cpc)
+	if err != nil {
+		return err
+	}
+
+	cm.Data["config"] = string(jsonStringByte)
```
tony-schndr, line 229: This data was read as bytes, but is being written as a string, is that ok?  I did test this change and the load balancer was updated as expected and everything worked great, I'm just wanting to double check this with you as I noticed the config map went from this
```
data:
  config: "{\n\t\"cloud\": \"AzurePublicCloud\",\n\t\"tenantId\": \"64dc69e4-d083-49fc-9569-ebece1dd1408\",\n\t\"aadClientId\":
    \"\",\n\t\"aadClientSecret\": \"\",\n\t\"aadClientCertPath\": \"\",\n\t\"aadClientCertPassword\":
    \"\",\n\t\"useManagedIdentityExtension\": false,\n\t\"userAssignedIdentityID\":
    \"\",\n\t\"subscriptionId\": \"fe16a035-e540-4ab7-80d9-373fa9a3d6ae\",\n\t\"resourceGroup\":
    \"aro-reo2fmqn\",\n\t\"location\": \"eastus\",\n\t\"vnetName\": \"aro-vnet\",\n\t\"vnetResourceGroup\":
    \"tschneid-02\",\n\t\"subnetName\": \"worker-subnet\",\n\t\"securityGroupName\":
    \"tschneid-npmdm-nsg\",\n\t\"routeTableName\": \"tschneid-npmdm-node-routetable\",\n\t\"primaryAvailabilitySetName\":
    \"\",\n\t\"vmType\": \"\",\n\t\"primaryScaleSetName\": \"\",\n\t\"cloudProviderBackoff\":
    true,\n\t\"cloudProviderBackoffRetries\": 0,\n\t\"cloudProviderBackoffExponent\":
    0,\n\t\"cloudProviderBackoffDuration\": 6,\n\t\"cloudProviderBackoffJitter\":
    0,\n\t\"cloudProviderRateLimit\": false,\n\t\"cloudProviderRateLimitQPS\": 0,\n\t\"cloudProviderRateLimitBucket\":
    0,\n\t\"cloudProviderRateLimitQPSWrite\": 0,\n\t\"cloudProviderRateLimitBucketWrite\":
    0,\n\t\"useInstanceMetadata\": true,\n\t\"loadBalancerSku\": \"standard\",\n\t\"excludeMasterFromStandardLB\":
    false,\n\t\"disableOutboundSNAT\": null,\n\t\"maximumLoadBalancerRuleCount\":
    0\n}\n"
```
to this
```
data:
  config: '{"cloud":"AzurePublicCloud","tenantId":"64dc69e4-d083-49fc-9569-ebece1dd1408","aadClientId":"","aadClientSecret":"","aadClientCertPath":"","aadClientCertPassw
ord":"","useManagedIdentityExtension":false,"userAssignedIdentityID":"","subscriptionId":"fe16a035-e540-4ab7-80d9-373fa9a3d6ae","resourceGroup":"aro-reo2fmqn","location"
:"eastus","vnetName":"aro-vnet","vnetResourceGroup":"tschneid-02","subnetName":"worker-subnet","securityGroupName":"tschneid-npmdm-nsg","routeTableName":"tschneid-npmdm-
node-routetable","primaryAvailabilitySetName":"","vmType":"","primaryScaleSetName":"","cloudProviderBackoff":true,"cloudProviderBackoffRetries":0,"cloudProviderBackoffEx
ponent":0,"cloudProviderBackoffDuration":6,"cloudProviderBackoffJitter":0,"cloudProviderRateLimit":false,"cloudProviderRateLimitQPS":0,"cloudProviderRateLimitBucket":0,"
cloudProviderRateLimitQPSWrite":0,"cloudProviderRateLimitBucketWrite":0,"useInstanceMetadata":true,"loadBalancerSku":"standard","excludeMasterFromStandardLB":false,"disa
bleOutboundSNAT":true,"maximumLoadBalancerRuleCount":0}'
```

SudoBrendan: we might be able to test this out functionally in an E2E? Attempt to modify the CM and then see that it gets reverted? Granted, idk what kind of timeframe this would take :/

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -0,0 +1,236 @@
...
+)
+
+var cloudProviderConfigName = types.NamespacedName{Name: "cloud-provider-config", Namespace: "openshift-config"}
+
+// / This is a very old version of the CloudProviderConfig struct from k8s upstream
+// / OpenShift diverged from the upstream cluster API and we have a depreciated version of this struct
+// / I was unable to use the recent version of this struct in the upstream k8s Azure Cloud Provider
+// / This version of the struct can be found here:
+// / https://github.com/kubernetes/kubernetes/blob/v1.13.5/pkg/cloudprovider/providers/azure/azure.go#L81
+type azCloudProviderConfig struct {
```
niontive, line 39: It looks like we only use a subset of these fields. Since we're unmarshalling a `[]byte` into this struct, we should be good to omit the fields we don't need ([unmarshal](https://pkg.go.dev/encoding/json#Unmarshal) ignore missing fields).

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -0,0 +1,236 @@
...
+		r.Log.Info("updating openshift-config/cloud-provider-config disableOutboundSNAT from false to true")
+		*cpc.DisableOutboundSNAT = true
+		changed = true
+	} else if cpc.DisableOutboundSNAT == nil {
+		r.Log.Info("updating openshift-config/cloud-provider-config disableOutboundSNAT from nil to true")
+		truePointer := true
+		cpc.DisableOutboundSNAT = &truePointer
+		changed = true
+	} else {
+		r.Log.Debug("openshift-config/cloud-provider-config disableOutboundSNAT is set to true no changes needed")
```
niontive, line 221: Could we just return after this debug statement? This way we don't have to keep the `changed` variable.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
lranjbar: I tested this scenario with the functional testing.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
lranjbar: Omitting fields causes JSON marshal errors.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
lranjbar: Removing the changed variable would make this more wordy. We'd be marshaling JSON and catching errors in every branch.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+}
+
+// Reconcile makes sure that the cloud-provider-config is healthy
+func (r *CloudProviderConfigReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	r.Log.Debug("reconcile ConfigMap openshift-config/cloud-provider-config")
+
+	instance, err := r.GetCluster(ctx)
+	if err != nil {
+		r.Log.Info(err)
```
s-amann, line 151: nit: should this be logged at error level since we couldn't pull the cluster?

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
s-amann: I think returning nil here and removing changed simplies the code. There would be three paths, two of which need to continue on (changed = true) and the third which can exit early. Since the two non terminal paths also have changed = true, there are really only two paths: we need to update or we need to return. The update code can be left outside of the conditional block and we always just return `r.Client.Update()` as the last line of this method. IMO we should try to avoid using tracking booleans such as changed as it adds mental load when reading the code. I tried this change locally and the tests still passed :)

That is of course, unless we need to also check if we can marshal the unchanged `cpc`, which I think we can omit if everything is fine (changed = false).

s-amann (approved):
some nits around logging level and some clean up, but LGTM functionally.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
lranjbar: I do not agree that having a changed boolean adds mental load to the code. IMO it makes this more readable.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
lranjbar: This was put in for debugging purposes. I'll remove it.

kimorris27 (approved):
I looked into the issue with the Azure cloud provider config struct a bit, but I'm still trying to piece together what's going on - it's complicated as you said.

I'm wondering if we could eventually find a way to import the struct rather than having to copy/paste it into our code, but I don't want to block this PR over it.

pkg/api/defaults.go
gvanderpotte: on top of that, after PUCM, we should not see a change for existing cluster unless the services are recreated. That being said, it might be worth sending notification about the possible change of outbound ip if services are recreated

SudoBrendan (approved):

PR #3071, opened 2023-07-28
remove gateway restart during rp predeploy

rajdeepc2792:

Jira - https://issues.redhat.com/browse/ARO-3240

Fixes

Remove the gateway restart during the predeploy which was added as part of PR https://github.com/Azure/ARO-RP/pull/2946 because the rotated encryptions secrets are not used by gateway and the gateway service doesn't need the restart during predeploy

### Test plan for issue:
Updated Unit Test Cases

No

### DIFF
--- a/pkg/deploy/predeploy.go
+++ b/pkg/deploy/predeploy.go
@@ -401,11 +401,7 @@ func (d *deployer) configureServiceSecrets(ctx context.Context) error {
 	}
 
 	if isRotated {
-		err = d.restartOldScalesets(ctx, d.config.GatewayResourceGroupName)
-		if err != nil {
-			return err
-		}
-		err = d.restartOldScalesets(ctx, d.config.RPResourceGroupName)
+		err = d.restartOldRPScalesets(ctx)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


pkg/deploy/predeploy.go
```diff
@@ -506,35 +502,29 @@ func (d *deployer) restartOldScalesets(ctx context.Context, resourceGroupName st
 	return nil
 }
 
-func (d *deployer) restartOldScaleset(ctx context.Context, vmssName string, resourceGroupName string) error {
-	var restartScript string
-	switch {
-	case strings.HasPrefix(vmssName, gatewayVMSSPrefix):
-		restartScript = gatewayRestartScript
```
cadenmarchese, line 513: Can we also remove the `gatewayRestartScript` var at the beginning of the file?

pkg/deploy/predeploy.go
```diff
@@ -490,14 +486,14 @@ func (d *deployer) ensureSecretKey(ctx context.Context, kv keyvault.Manager, sec
 	})
 }
 
-func (d *deployer) restartOldScalesets(ctx context.Context, resourceGroupName string) error {
-	scalesets, err := d.vmss.List(ctx, resourceGroupName)
+func (d *deployer) restartOldRPScalesets(ctx context.Context) error {
```
cadenmarchese, line 489: I'm not sure you need to rename the functions to include `RP`, to it makes sense to keep them the same as they were in case we need to reuse them in the future.

cadenmarchese (commented):
Thanks for your quick work @rajdeepc2792 !

pkg/deploy/predeploy.go
rajdeepc2792: Done.

pkg/deploy/predeploy.go
rajdeepc2792: Reverted the function rename.

cadenmarchese (approved):

petrkotas (approved):
Thank you, LGTM

PR #3072, opened 2023-07-28
Ignore vendor directory while scanning golang code

SudoBrendan:

Fixes


mociarain:
No longer needed as we removed these unused templates: https://github.com/Azure/ARO-RP/pull/3108/files

Please do re-open if I've misunderstood

PR #3073, opened 2023-07-31
`az aro update` `--load-balancer-managed-outbound-ip-count` flag support

tsatam:

Fixes [ARO-3103](https://issues.redhat.com/browse/ARO-3103)
Implicitly fixes [ARO-2467](https://issues.redhat.com/browse/ARO-2467) as well

Implements support within the `az aro update` command in the ARO extension for the managedOutboundIps.Count property on the cluster load balancer profile, implemented in the 2023-07-01 preview API. 

### Test plan for issue:

On a provisioned ARO cluster, managed by an RP with the 2023-07-01 preview API (e.g. local dev):

- Run `az aro update --help` and validate that the `--load-balancer-managed-outbound-ip-count` flag and information is present
- Run `az aro update --load-balancer-managed-outbound-ip-count $NUMBER` for numbers != 1 and ensure that the resulting cluster document has property `openShiftCluster.properties.networkProfile.loadBalancerProfile.managedOutboundIps.count` set to `$NUMBER` instead of the previous value. 

None for this specific PR (overall doc updates captured in [ARO-3513](https://issues.redhat.com/browse/ARO-3513))

### Additional Notes:

- Azure CLI command flags require at least one option to be present that is overall shorter than 22 characters. As a result, I have added the `--lb-ip-count` alias for this flag. Open to suggestions for name changes to this shorthand flag (as long as it is within the 22 character limit)
- This command will not yet perform any observable changes in ARO clusters, outside of changing the property within the cluster document, as [ARO-3110](https://issues.redhat.com/browse/ARO-3110) needs to be completed in order to implement the functionality in the RP.

### DIFF
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -68,8 +68,6 @@ def load_arguments(self, _):
         c.argument('outbound_type',
                    help='Outbound type of cluster. Must be "Loadbalancer" (default) or "UserDefinedRouting".',
                    validator=validate_outbound_type)
-        c.argument('enable_preconfigured_nsg', arg_type=get_three_state_flag(),
-                   help='Use Preconfigured NSGs. Allowed values: false, true.')
         c.argument('disk_encryption_set',
                    help='ResourceID of the DiskEncryptionSet to be used for master and worker VMs.',
                    validator=validate_disk_encryption_set)
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -17,6 +17,7 @@ from azext_aro._validators import validate_worker_vm_disk_size_gb
 from azext_aro._validators import validate_refresh_cluster_credentials
 from azext_aro._validators import validate_version_format
 from azext_aro._validators import validate_outbound_type
+from azext_aro._validators import validate_load_balancer_managed_outbound_ip_count
 from azure.cli.core.commands.parameters import name_type
 from azure.cli.core.commands.parameters import get_enum_type, get_three_state_flag
 from azure.cli.core.commands.parameters import resource_group_name_type
@@ -122,6 +123,11 @@ def load_arguments(self, _):
                    help='Refresh cluster application credentials.',
                    options_list=['--refresh-credentials'],
                    validator=validate_refresh_cluster_credentials)
+        c.argument('load_balancer_managed_outbound_ip_count',
+                   type=int,
+                   help='Count of ARO-managed public IP addresses to attach to the public load balancer',
+                   validator=validate_load_balancer_managed_outbound_ip_count,
+                   options_list=['--load-balancer-managed-outbound-ip-count', '--lb-ip-count'])
     with self.argument_context('aro get-admin-kubeconfig') as c:
         c.argument('file',
                    help='Path to the file where kubeconfig should be saved. Default: kubeconfig in local directory',
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -125,7 +125,7 @@ def load_arguments(self, _):
                    validator=validate_refresh_cluster_credentials)
         c.argument('load_balancer_managed_outbound_ip_count',
                    type=int,
-                   help='Count of ARO-managed public IP addresses to attach to the public load balancer',
+                   help='The desired number of IPv4 outbound IPs created and managed by Azure for the cluster public load balancer.',  # pylint: disable=line-too-long
                    validator=validate_load_balancer_managed_outbound_ip_count,
                    options_list=['--load-balancer-managed-outbound-ip-count', '--lb-ip-count'])
     with self.argument_context('aro get-admin-kubeconfig') as c:
--- a/python/az/aro/azext_aro/_validators.py
+++ b/python/az/aro/azext_aro/_validators.py
@@ -271,3 +271,9 @@ def validate_refresh_cluster_credentials(namespace):
 def validate_version_format(namespace):
     if namespace.version is not None and not re.match(r'^[4-9]{1}\.[0-9]{1,2}\.[0-9]{1,2}$', namespace.version):
         raise InvalidArgumentValueError('--version is invalid')
+
+
+def validate_load_balancer_managed_outbound_ip_count(namespace):
+    if namespace.load_balancer_managed_outbound_ip_count is not None:
+        if namespace.load_balancer_managed_outbound_ip_count < 1 or namespace.load_balancer_managed_outbound_ip_count > 20:  # pylint: disable=line-too-long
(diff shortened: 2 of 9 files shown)


SudoBrendan:
Adding `hold` until we get BU approval on the shorthand.

python/az/aro/azext_aro/_params.py
```diff
@@ -122,6 +123,11 @@ def load_arguments(self, _):
                    help='Refresh cluster application credentials.',
                    options_list=['--refresh-credentials'],
                    validator=validate_refresh_cluster_credentials)
+        c.argument('load_balancer_managed_outbound_ip_count',
+                   type=int,
+                   help='Count of ARO-managed public IP addresses to attach to the public load balancer',
```
SudoBrendan, line 128: Can we align this closer to the swagger spec? https://github.com/Azure/ARO-RP/blob/master/swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/preview/2023-07-01-preview/redhatopenshift.json#L2009

python/az/aro/azext_aro/_validators.py
```diff
@@ -274,3 +275,9 @@ def validate_refresh_cluster_credentials(namespace):
 def validate_version_format(namespace):
     if namespace.version is not None and not re.match(r'^[4-9]{1}\.[0-9]{1,2}\.[0-9]{1,2}$', namespace.version):
         raise InvalidArgumentValueError('--version is invalid')
+
+
+def validate_load_balancer_managed_outbound_ip_count(namespace):
+    if namespace.load_balancer_managed_outbound_ip_count is not None:
+        if namespace.load_balancer_managed_outbound_ip_count < 1 or namespace.load_balancer_managed_outbound_ip_count > 20:  # pylint: disable=line-too-long
+            raise InvalidArgumentValueError('--load-balancer-managed-outbound-ip-count must be between 1 and 20')  # pylint: disable=line-too-long
```
SudoBrendan, line 283: probably worth adding `(inclusive)` at the end of this message.

SudoBrendan (commented):
Just a few comments on customer experience things, might be considered nits, but it would be good to align our messaging with the API.

SudoBrendan (commented):
LGTM, but waiting on approval until I talk with the BU. Great work, thank you!

tsatam:
NOTE: For developer convenience reasons, this PR should also be on hold until the 20230701preview API is released to all RP regions in prod, as it not being available will break local extension usage.

python/az/aro/azext_aro/_params.py
```diff
@@ -122,6 +123,11 @@ def load_arguments(self, _):
                    help='Refresh cluster application credentials.',
                    options_list=['--refresh-credentials'],
                    validator=validate_refresh_cluster_credentials)
+        c.argument('load_balancer_managed_outbound_ip_count',
+                   type=int,
+                   help='The desired number of IPv4 outbound IPs created and managed by Azure for the cluster public load balancer.',  # pylint: disable=line-too-long
+                   validator=validate_load_balancer_managed_outbound_ip_count,
+                   options_list=['--load-balancer-managed-outbound-ip-count', '--lb-ip-count'])
```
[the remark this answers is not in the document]
tsatam: I saw the shorthand alternative rule as a linter failure from the `azdev` linter commands. It's possible that this rule is ignorable, or that AKS implemented this flag before that rule was imposed and the rule only applies to new and/or changed flags. In any case I'm not sure how we could validate the rule being required until we attempt to upstream this change. 

+1 on aligning the description with AKS, I will make that change.

python/az/aro/azext_aro/_params.py
```diff
@@ -113,6 +114,12 @@ def load_arguments(self, _):
         c.argument('worker_subnet',
                    help='Name or ID of worker vnet subnet.  If name is supplied, `--vnet` must be supplied.',
                    validator=validate_subnet('worker_subnet'))
+        c.argument('load_balancer_managed_outbound_ip_count',
+                   type=int,
```
jhoreman, line 118: Comment: Upstream specifically wanted int32 formats here. Should work just fine but just calling it out.

SudoBrendan: I think that Tanmay had mentioned that we should likely hold on this until we do the downstreaming of all the REST Specs work, then rebase this PR on that.

tsatam: A quick peek at AKS's equivalent flag shows they've defined the type as `int` as well: https://github.com/Azure/azure-cli/blob/d419ab357481e803d2f2a84a40760fa6e6af7c8b/src/azure-cli/azure/cli/command_modules/acs/_params.py#L334

I'm not sure if the API specificity on the data type necessarily correlates to the CLI/Python-side requirements here.

tony-schndr: native python3 doesn't have a int32 type

tsatam:
Updated PR to point to new `azext-aro-v20230701preview` branch, in order to not conflict with the in-flight efforts on the stable v20230901 API and corresponding CLI changes.

SudoBrendan (approved):

python/az/aro/azext_aro/custom.py
```diff
@@ -147,7 +154,11 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
             pod_cidr=pod_cidr or '10.128.0.0/14',
             service_cidr=service_cidr or '172.30.0.0/16',
             outbound_type=outbound_type or '',
+<<<<<<< HEAD
```
tony-schndr, line 157: Missed during rebase I'm guessing

python/az/aro/azext_aro/_params.py
SudoBrendan: are we good to resolve this?

SudoBrendan (changes requested):
Other than the comment Tony left, LGTM.

python/az/aro/azext_aro/_params.py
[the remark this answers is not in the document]
SudoBrendan: This looks resolved as well, can we resolve?

python/az/aro/azext_aro/custom.py
tsatam: fixed, thanks.

python/az/aro/azext_aro/_params.py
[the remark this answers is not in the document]
tsatam: The description is not fixed, but we will be moving that effort to a post-release activity as it will require us to change how the entire `az aro` command is documented (moving user-facing documentation from the parameter definitions themselves into `_help.py`).

tony-schndr (approved):
lgtm, thanks Tanmay!

SudoBrendan (approved):

PR #3074, opened 2023-08-01
fix the gatekeeper doesnot start on 4.10 problem

yjst2012:

Fixes [the gatekeeper doesnot start on 4.10 problem](https://issues.redhat.com/browse/ARO-3843) 

It seems the default seccomp profiles need to be enabled on 4.10 in order to make it work, ref: [Enabling the default seccomp profile for all pods](https://docs.openshift.com/container-platform/4.10/security/seccomp-profiles.html#configuring-default-seccomp-profile_configuring-seccomp-profiles)
however for 4.11:
In OpenShift Container Platform 4.11 the ability to add the pod annotations [seccomp.security.alpha.kubernetes.io/pod](http://seccomp.security.alpha.kubernetes.io/pod): runtime/default and [container.seccomp.security.alpha.kubernetes.io/](http://container.seccomp.security.alpha.kubernetes.io/)<container_name>: runtime/default is deprecated.
https://docs.openshift.com/container-platform/4.11/security/seccomp-profiles.html
which explains why it works with the section for 4.11 but not for 4.10.

### Test plan for issue:

pods are able to start and constrains work 

NA

### DIFF
--- a/pkg/operator/controllers/guardrails/staticresources/gk_audit_controller_deployment.yaml
+++ b/pkg/operator/controllers/guardrails/staticresources/gk_audit_controller_deployment.yaml
@@ -100,8 +100,8 @@ spec:
             - ALL
           readOnlyRootFilesystem: true
           runAsNonRoot: true
-          seccompProfile:
-            type: RuntimeDefault
+          # seccompProfile:
+          #   type: RuntimeDefault
         volumeMounts:
         - mountPath: /certs
           name: cert
--- a/pkg/operator/controllers/guardrails/staticresources/gk_controller_manager_deployment.yaml
+++ b/pkg/operator/controllers/guardrails/staticresources/gk_controller_manager_deployment.yaml
@@ -112,8 +112,8 @@ spec:
             - ALL
           readOnlyRootFilesystem: true
           runAsNonRoot: true
-          seccompProfile:
-            type: RuntimeDefault
+          # seccompProfile:
+          #   type: RuntimeDefault
         volumeMounts:
         - mountPath: /certs
           name: cert


hawkowl (approved):

pkg/operator/controllers/guardrails/staticresources/gk_audit_controller_deployment.yaml
```diff
@@ -100,8 +100,8 @@ spec:
             - ALL
           readOnlyRootFilesystem: true
           runAsNonRoot: true
-          seccompProfile:
-            type: RuntimeDefault
+          # seccompProfile:
```
petrkotas, line 103: I do apologize. I fail to see how removing the secompProfile enables the default.
I have read the linked documentation, which explains how to add it and define. I do not see this done in this PR.

How does this work on 4.10? How does this work on 4.11?

petrkotas (changes requested):
Hi @yjst2012 I have a question before we merge it. The change is straightforward, but it really deserves deeper explanation. SecurityContext is hard topic, and providing the explanation will simplify our life in the future. Thank you.

pkg/operator/controllers/guardrails/staticresources/gk_audit_controller_deployment.yaml
petrkotas: To clarify, I am not asking for code change, I am asking for explanation to go in as a part of the PR

yjst2012: Fist of all, thanks for looking at the change, really need more eyes on guardrails related changes to prevent any potential risky code from being merged.
TBH, I'm not very sure about such security related issues, most of the changes (if not all) are due to a rising problem.
As to seccompProfile, my understanding is, for 4.10 it requires some changes described in the doc in order to make it work, otherwise the pods wont start which is the same as what I encountered recently in my tests.
and for 4.11, this has been deprecated, so with or without it wont make much difference IMO.
So far we are mostly plan to do some tests on limited clusters and it wont be enforced or largely deployed, the change should be fine for this purpose.
And I also tested another change which should be a bit securer than this, however, I'm not vert sure about that.
I was asking in OPA channel to get more insights, hopefully could address this completely.

s-amann (changes requested):
I don't think we should merge commented out code to be deployed into clusters. Rather can we template out the yaml files to be version dependant? If a field is deprecated in 4.11 then it seems reasonable eventually it will not be supported.

SudoBrendan:
FYI: Might be a dupe of https://github.com/Azure/ARO-RP/pull/3080 ?

yjst2012:
close this as it is replaced by #3080

PR #3075, opened 2023-08-01
Monitor SP Expiry Date

rhamitarora:

[Alert on loss of access rights and permissions on cluster / network resources](https://issues.redhat.com/browse/ARO-1987)

### DIFF
--- a/cmd/aro/monitor.go
+++ b/cmd/aro/monitor.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"os"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/go-autorest/tracing"
 	"github.com/sirupsen/logrus"
 	kmetrics "k8s.io/client-go/tools/metrics"
@@ -126,7 +127,18 @@ func monitor(ctx context.Context, log *logrus.Entry) error {
 		return err
 	}
 
-	mon := pkgmonitor.NewMonitor(log.WithField("component", "monitor"), dialer, dbMonitors, dbOpenShiftClusters, dbSubscriptions, m, clusterm, liveConfig)
+	options := _env.Environment().EnvironmentCredentialOptions()
+	spTokenCredential, err := azidentity.NewEnvironmentCredential(options)
+	if err != nil {
+		return err
+	}
+
+	spGraphClient, err := _env.Environment().NewGraphServiceClient(spTokenCredential)
+
+	if err != nil {
+		return err
+	}
+	mon := pkgmonitor.NewMonitor(log.WithField("component", "monitor"), dialer, dbMonitors, dbOpenShiftClusters, dbSubscriptions, m, clusterm, liveConfig, spGraphClient)
 
 	return mon.Run(ctx)
 }
--- a/pkg/monitor/cluster/serviceprincipalexpiration.go
+++ b/pkg/monitor/cluster/serviceprincipalexpiration.go
@@ -1,0 +1,27 @@
+package cluster
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+
+	utilgraph "github.com/Azure/ARO-RP/pkg/util/graph"
+)
+
+/****************************************************
+	Monitor the Service Prinicpal exipiration date
+****************************************************/
+
+func (mon *Monitor) emitSpExpiration(ctx context.Context) error {
+	expiryDate, err := utilgraph.GetServicePrincipalExpiryByAppID(ctx, mon.spGraphClient, mon.oc.Properties.ServicePrincipalProfile.ClientID)
+	if err != nil {
+		return err
+	}
+
+	mon.emitGauge("cluster.serviceprincipal.expiration", 1, map[string]string{
+		"expiryDate": expiryDate.String(),
+	})
+
+	return nil
+}
--- a/pkg/monitor/cluster/serviceprincipalexpiration_test.go
+++ b/pkg/monitor/cluster/serviceprincipalexpiration_test.go
@@ -1,0 +1,59 @@
+package cluster
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+
+	"github.com/Azure/ARO-RP/pkg/api"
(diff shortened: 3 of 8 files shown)


pkg/monitor/cluster/serviceprincipalexpiration.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		return err
+	}
+
+	expiryDate, err := time.Parse("2006-01-02", *date) // Convert string date in Date Time Format
+	if err != nil {
+		return err
+	}
+
+	if int(time.Until(expiryDate).Hours()/24) <= 30 { // Alert if no. of days left on SP expiration is less than or equal to 30
```
dem4gus, line 30: Instead of hard coding the number of days for the alert here, why don't we just emit the number of days until expiration (Geneva only does `int` from what I've seen) and let the Geneva monitors do the work of deciding whether or not it's close enough to raise the alert on? That way if we decide to adjust the alert in the future we just need to change the stuff in Geneva, and not the code.

pkg/monitor/cluster/serviceprincipalexpiration_test.go
```diff
@@ -0,0 +1,59 @@
...
+						ClientID: tt.appId,
+					},
+				},
+			}
+			mon := &Monitor{
+				m:  m,
+				oc: oc,
+			}
+
+			m.EXPECT().EmitGauge("cluster.serviceprincipal.expiration", int64(1), map[string]string{
```
dem4gus, line 49: In order to test this you'll need a mock for your call to `utilgraph.GetServicePrincipalExpiryByAppID` to return what you need. The `mock_graph` package is in pkg/util/mocks/graph.

pkg/util/graph/graph.go
```diff
@@ -38,3 +40,37 @@ func GetServicePrincipalIDByAppID(ctx context.Context, graph *msgraph.GraphServi
 		return nil, fmt.Errorf("%d service principals have appId '%s'", len(matches), appId)
 	}
 }
+
+// GetServicePrincipalExpiryByAppID returns a service principal's expiry date from
+// an application (client) ID.
+func GetServicePrincipalExpiryByAppID(ctx context.Context, graph *msgraph.GraphServiceClient, appId string) (*string, error) {
```
dem4gus, line 46: nit: return a `time.Time` here and then let whatever is calling this function manipulate and format it to what it needs.

dem4gus (changes requested):
Good first pass, some suggestions for utility and reducing the footprint of future changes.

pkg/util/graph/graph.go
```diff
@@ -38,3 +40,37 @@ func GetServicePrincipalIDByAppID(ctx context.Context, graph *msgraph.GraphServi
...
+	if err != nil {
+		return nil, err
+	}
+
+	var endDateTime string
+	matches := result.GetValue()
+
+	for _, element := range matches[0].GetPasswordCredentials() {
+		e := element.(*models.PasswordCredential)
+		endDateTime = e.GetEndDateTime().Format("2006-01-02")
```
dem4gus, line 65:
```suggestion
		endDateTime = e.GetEndDateTime()
```

pkg/monitor/cluster/serviceprincipalexpiration.go
```diff
@@ -0,0 +1,36 @@
...
+		mon.emitGauge("cluster.serviceprincipal.expiration", 1, map[string]string{
+			"expiry-date": *date,
```
dem4gus, line 32: We use camelCase on our monitor field names.
```suggestion
			"expiryDate": *date,
```

cmd/aro/monitor.go
```diff
@@ -126,7 +127,18 @@ func monitor(ctx context.Context, log *logrus.Entry) error {
...
 	}
 
-	mon := pkgmonitor.NewMonitor(log.WithField("component", "monitor"), dialer, dbMonitors, dbOpenShiftClusters, dbSubscriptions, m, clusterm, liveConfig)
+	options := _env.Environment().EnvironmentCredentialOptions()
+	spTokenCredential, err := azidentity.NewEnvironmentCredential(options)
+	if err != nil {
+		return err
+	}
+
+	spGraphClient, err := _env.Environment().NewGraphServiceClient(spTokenCredential)
```
bennerv, line 136: Have we verified this will work?  

This environment credential option reads a service principal from environment variables, and we're not setting them on the monitor.  

Additionally, how do we plan on accomplishing this?  If we use the customer's cluster SP, it may not have the proper permissions to read its own expiry.

bennerv (changes requested):

SudoBrendan (changes requested):
I'm agreeing with other comments, adding a few more. Great first pass, thank you!

pkg/monitor/cluster/serviceprincipalexpiration.go
rhamitarora: Done

pkg/util/graph/graph.go
rhamitarora: Done

pkg/util/graph/graph.go
rhamitarora: Done

pkg/monitor/cluster/serviceprincipalexpiration.go
rhamitarora: Done

cmd/aro/monitor.go
rhamitarora: The spGraphClient is passed from Monitor to Cluster package via pkg/monitor/worker.go.

Currently using msgraph-sdk-go we are fetching SP Id using APP Id in Graph Package. I created new function GetServicePrincipalExpiryByAppID using above logic under Graph package and made the API call to msgraph-sdk-go to fetch SP Expiry date using APP Id. If there is a Permission restriction it will impact both.

bennerv: I'm not concerned with the passing of the graph client, I'm concerned with how it's constructed.  

>This environment credential option reads a service principal from environment variables, and we're not setting them on the monitor. 

If you look at how the monitor is constructed from a systemd unit, we're not passing any [environment variables to the monitor process.](https://github.com/Azure/ARO-RP/blob/861c83319e8e7c59dab4011a8cc4c9f997509bc8/pkg/deploy/generator/scripts/rpVMSS.sh#L386-L423)

How are we testing this actually works as intended?  Additionally, if we use the first party service principal to check the expiration, the FPSP doesn't have the proper Microsoft.Graph permissions within the customer's tenant in order to check this expiration.

PR #3076, opened 2023-08-02
Add timeout for go generate

s-fairchild:
NA

Fixes

NA

This will prevent excessively long go generate runs from occurring in the Build and Push Binary and Container pipelines. Timeout sends a list of exit codes we can expect if something goes wrong. Here they are used to help provide more information about the failure. go generate doesn't support a timeout value directly.

### Test plan for issue:

[Build and Push Binary and Container pipelines - Pull Request pipeline for testing.](https://dev.azure.com/msazure/AzureRedHatOpenShift/_build/results?buildId=77505047&view=results)

No

### DIFF
--- a/hack/ci-utils/go_generate.sh
+++ b/hack/ci-utils/go_generate.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+set -o errexit
+
+# Variables checked for timeout values, defaults used if unset
+soft_timeout="${GO_GENERATE_SOFT_TIMEOUT:-2h}"
+hard_timeout="${GO_GENERATE_HARD_TIMEOUT:-2h}"
+
+if [[ -n $CI ]]; then
+    set -x
+fi
+
+
+declare timeout_exitcode go_generate
+# go generate doesn't support timeout itself, so --preserve-status wouldn't help us as it would simply be the status timeout sent to go.
+go_generate="go generate ./..."
+echo "Running ${go_generate} sending SIGTERM after ${soft_timeout} and SIGKILL after ${hard_timeout}"
+echo "${go_generate}"
+# shellcheck disable=SC2086
+timeout --foreground --kill-after="${hard_timeout}" -v -s SIGTERM "${soft_timeout}" ${go_generate}
+timeout_exitcode=$?
+
+# Capture expected exit codes
+case $timeout_exitcode in
+124)
+    echo "Command ${go_generate} timed out."
+    ;;
+125)
+    echo "The command \"timeout\" failed trying to run ${go_generate}. Note that ${go_generate} did not fail."
+    ;;
+126)
+    echo "Command found but cannot be invoked."
+    ;;
+127)
+    echo "Command not found recieved."
+    ;;
+137)
+    echo "timeout or command recieved kill signal 9) SIGKILL."
+    ;;
+-)
+    echo "Command go generate ./... exited with exit_code code ${-}."
+    ;;
+0)
+    echo "Command ${go_generate} completed successfully, exit code ${timeout_exitcode}."
+    ;;
+*)
+    echo "Unexpected error code received: ${timeout_exitcode}."
+    ;;
+esac
+
+exit "${timeout_exitcode}"
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ discoverycache:
 	$(MAKE) generate
 
 generate:
-	go generate ./...
+	hack/ci-utils/go_generate.sh
 
 generate-guardrails:
 	cd pkg/operator/controllers/guardrails/policies && ./scripts/generate.sh > /dev/null


hack/ci-utils/go_generate.sh
```diff
@@ -0,0 +1,64 @@
...
+    fi
+
+
+    # Declare local before using command substitution so that local doesn't cover up timeout's exit code.
+    local timeout_exitcode go_generate
+    # go generate doesn't support timeout itself, so --preserve-status wouldn't help us as it would simply be the status timeout sent to go.
+    go_generate="go generate ./..."
+    echo "Running ${go_generate} sending SIGTERM after ${soft_timeout} and SIGKILL after ${hard_timeout}"
+    echo "${go_generate}"
+     # shellcheck disable=SC2086
```
s-fairchild, line 23: This is warning about variable expansion, which we actually want here so that COMMAND and ARGS are separated.

hack/ci-utils/go_generate.sh
```diff
@@ -0,0 +1,64 @@
...
+
+    -)
+        echo "Command go generate ./... exited with exit_code code ${-}"
+        ;;
+    0)
+        echo "Command ${go_generate} completed successfully, exit code ${timeout_exitcode}"
+        ;;
+
+    *)
+        echo "Unexpected error code recieved: ${timeout_exitcode}"
```
cadenmarchese, line 57:
```suggestion
        echo "Unexpected error code received: ${timeout_exitcode}"
```

cadenmarchese (commented):
Thanks for proposing this PR! Overall I'm not sure we need to handle all of these cases. I think we could simplify this a little bit by just including these two cases:
```
        echo "Command ${go_generate} completed successfully, exit code ${timeout_exitcode}"
        echo "Unexpected error code received: ${timeout_exitcode}"
```
I'm also not sure about the timeout / imposing a timeout in general - I left a comment on that, let me know what you think.

hack/ci-utils/go_generate.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+set -o errexit
+
+
+main() {
+
+    # Variables checked for timeout values, defaults used if unset
+    soft_timeout="${GO_GENERATE_SOFT_TIMEOUT:-20m}"
+    hard_timeout="${GO_GENERATE_HARD_TIMEOUT:-21m}"
```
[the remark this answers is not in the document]
s-fairchild: I landed on 20/21 minutes by incrementing the time required to pass the CI pipeline. My thinking here is we want it to take as little time as possible ideally.
I'm not sure what the longest average successful `go generate` time is. We don't want it to kill early if a few more minutes would've been successful.
How does 30 minutes sound?

When it fails I believe the root cause is slow VM IO performance, since `go generate` is so IO intensive.

hack/ci-utils/go_generate.sh
```diff
@@ -0,0 +1,64 @@
...
+    125)
+        echo "The command \"timeout\" failed trying to run ${go_generate}. Note that ${go_generate} did not fail."
+        ;;
+
+    126)
+        echo "Command found but cannot be invoked."
+        ;;
+
+    127)
+        echo "Command not found recieved."
```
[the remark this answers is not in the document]
s-fairchild: Probably not, but I figured it's easier to print these messages for known error codes rather than having the pipeline display the exit code for us to figure out.

hack/ci-utils/go_generate.sh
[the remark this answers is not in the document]
cadenmarchese: Maybe a broader question, why fail out the whole pipeline on a timeout when it will eventually be successful or fail out? If we need a pipeline run executed quickly and believe it's stuck, would it be better to just kick off a new pipeline run?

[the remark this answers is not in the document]
s-fairchild: So the default timeout value for `go test` is 10 minutes. I believe the thinking here is, your unit tests shouldn't take long, and if they do something is probably wrong.
I think the delay of waiting for a `go generate` run that never completes is more problematic than failing after a known "too long" point. I haven't seen `go generate` succeed past 1 hour, so maybe that could be a timeout value.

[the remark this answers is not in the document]
jhoreman: What is the current behaviour should the `go generate` hang or fail? Does it get stuck indefinitely? We can probably settle on 1 hour here to be on the safe side.

hack/ci-utils/go_generate.sh
```diff
@@ -0,0 +1,64 @@
...
+set -o errexit
+
+
+main() {
+
+    # Variables checked for timeout values, defaults used if unset
+    soft_timeout="${GO_GENERATE_SOFT_TIMEOUT:-20m}"
+    hard_timeout="${GO_GENERATE_HARD_TIMEOUT:-21m}"
+
+    if [[ -n $CI ]]; then
```
cadenmarchese, line 12: You can probably leave out the `if` statement since I think it's assumed that this script is specifically for CI (hack/ci-utils/*) and we'd always expect this to be set.

hack/ci-utils/go_generate.sh
cadenmarchese: Should this be at the top of the file?

hack/ci-utils/go_generate.sh
```diff
@@ -0,0 +1,64 @@
...
+
+    *)
+        echo "Unexpected error code received: ${timeout_exitcode}."
+        ;;
+    esac
+
+    exit "${timeout_exitcode}"
+}
+
+main
```
cadenmarchese, line 64: My 0.02 - we can remove this `main` function / cmd since we're not calling this logic anywhere outside of this file.

hack/ci-utils/go_generate.sh
```diff
@@ -0,0 +1,64 @@
...
+     # shellcheck disable=SC2086
+    timeout --foreground --kill-after="${hard_timeout}" -v -s SIGTERM "${soft_timeout}" ${go_generate}
+    timeout_exitcode=$?
+
+    # Capture expected exit codes
+    case $timeout_exitcode in
+    124)
+        echo "Command ${go_generate} timed out."
+        ;;
+
```
cadenmarchese, line 32:
```suggestion
```
Ultimate nit - there's some extra whitespace around the script I think can be removed.

cadenmarchese (commented):
I think we can take this approach:
- Timeout after an hour (because at that point, we know it's stuck, and we want to free up the agent)
- Handle the error codes
- Get a few pipeline runs in to make sure it's working as expected

hack/ci-utils/go_generate.sh
s-fairchild: No, shellcheck's linter looks for comments directly above the error (I believe so anyways).

hack/ci-utils/go_generate.sh
s-fairchild: I typically create a main function in most of my bash programs, the reason is it puts the "main" logic at the top of the script. Since bash executes every line sequentially it can be harder to read this way.
I'm happy to pull it out of the function though if you think we should.

s-fairchild:
@cadenmarchese The error codes being translated here are directly from timeout, not necessarily go generate. Go generate doesn't appear to have any exit codes documented, I'm assuming it may use the exit code of that specific generator, but again it's not documented.

> If any generator returns an error exit status, "go generate" skips
all further processing for that package.

So if we are thinking a timeout value isn't useful for us, I can close this PR. If it maybe useful, I think we should raise the timeout value to something like 2 hours. After running the build pipeline several times now I've seen it take a variable amount of time, but over an hour isn't unusual.

hack/ci-utils/go_generate.sh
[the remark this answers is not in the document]
s-fairchild: @jhoreman It "hangs", or continues to execute very slowly until it's done. On a VM with very poor performance this could go on for quite some time.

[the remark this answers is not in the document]
SudoBrendan: I guess then the question is - is it "faster" for us to:

1. fail CI due to a slow disk (and then likely have a PR sit around for ~24h?) until a dev looks at logs, just to run `azp ci` again and gets a 10 minute run on a faster disk? Or
2. accept that we will occasionally get a long 60m `go generate` that eventually completes, or even never completes and has the CI stuck on the PR?

My guess is that even if we fail generate runs often, option 2 will be faster than option 1 for our FTS team, since most people reviewing PRs see :x: on CI and think "that's a YOU problem" not "that's a VM disk problem, I should re-run CI" :p

Alternatively, I'd be onboard with a timeout if we had automatic retry logic built in - because that will 100% of the time be the answer to getting unlucky like this.

hack/ci-utils/go_generate.sh
SudoBrendan: ah, I'm a little sad to see this removed, honestly... I use the same pattern @s-fairchild, and appreciate seeing someone else using it too! I think `main` in bash looks a lot cleaner and makes the code more organized into `set`s, `main`, and other various helpers.

s-fairchild: @SudoBrendan There's still time, I can add `main` back :)
It is a good design pattern, even for small scripts like this I think it helps.

hack/ci-utils/go_generate.sh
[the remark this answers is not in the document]
s-fairchild: @SudoBrendan I like the idea of retry logic. I can update this PR to shift priority to retrying `go generate` upon failure, rather than failing the job run.  

Are you suggesting retry logic to retry go generate in the same job agent, or starting a new job run on a new agent?

If this is due to slow disk performance, is it possible to start the job on a new agent, from within the job run itself? I think the answer is yes, with github/ADO APIs.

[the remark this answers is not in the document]
SudoBrendan: I would say if disk is the issue, we should spawn a new Job and/or retry the current Job, if that is possible.

hack/ci-utils/go_generate.sh
SudoBrendan: If anyone still isn't sold, I think the biggest win of `main` is that you don't have to scroll through "how the sausage is made" in your script before you hit the top-level workflow. That is, in bash, a (helper) function has to be defined before it is used. That inevitably leads to you defining a bunch of seemingly "random" helpers until you suddenly have top-level code that uses it, something like:

```
#!/bin/bash

func foo() { ... }

func bar() { ... }

func biz() { ... }

foo()
bar()
biz()
```

With `main`, you can put that all at the top and (hopefully) define easily legible top-level logic flow with some cool function names, and code starts to look like:

```
#!/bin/bash

func main() {
    foo()
    bar()
    biz()
}

func foo() { ... }

func bar() { ... }

func biz() { ... }

main() # kinda icky to have this "floating" top-level call, but hey, it's the best we can do. There's exactly 1 of these calls, and it ideally happens as the last line of every file...
```

hack/ci-utils/go_generate.sh
```diff
@@ -0,0 +1,51 @@
...
+    echo "The command \"timeout\" failed trying to run ${go_generate}. Note that ${go_generate} did not fail."
+    ;;
+126)
+    echo "Command found but cannot be invoked."
+    ;;
+127)
+    echo "Command not found recieved."
+    ;;
+137)
+    echo "timeout or command recieved kill signal 9) SIGKILL."
```
kimorris27, line 38: Nit: recieved -> received

kimorris27 (commented):
I'm +1 to the idea that automatic retry logic is the "true" solution to the problem, given that the `go generate` does sometimes complete on its own in longer-than-normal-but-still-not-unreasonable times (like, for example, 1.5 hr). The retry logic is probably outside of the scope of this PR though.

With that in mind, a 2 hour timeout would make sense to me.

s-fairchild:
This pull request generated some ideas to think about to improve our release process.

PR #3077, opened 2023-08-02
adding missing fields for upstream process

jhoreman:

It was found during the upstream process that these fields were missing and actively blocked linting/validation. They should be added here in the generator so we can properly upstream them. 

### Test plan for issue:

CI pass + upstream process functions. This should only affect the 2023-07-01 preview api. 

No


tsatam (approved):
LGTM

jhoreman:
Labeling as work in progress. Upstream process found a few more issues we have to correct downstream.

jhoreman:
Closing this in favour of new expanded PR.

PR #3070, opened 2023-07-28
Move pkg/api dependencies on util/subnet to api/util/subnet

hawkowl:
### Which issue this PR addresses:

Fixes https://issues.redhat.com/browse/ARO-3574

### What this PR does / why we need it:

Moves `pkg/api/` dependencies of `pkg/util/subnet/` into `pkg/api/util/subnet` so that `pkg/api` does not have cross-package dependencies.

### Test plan for issue:

Unit tests should continue to pass

### Is there any documentation that needs to be updated for this PR?

N/A, refactor

PR #3079, opened 2023-08-03
Move pkg/api dependencies on util/immutable to api/util/immutable

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-3864

Moves a dependency that is only used in pkg/api/ into pkg/api. Depends on #3070 

### Test plan for issue:
Unit tests

N/A


tsatam (approved):
Straightforward change, LGTM!

I'm trusting CI being 🟢 to have validated that the package isn't being used outside of `pkg/api`, and that there are no changes to generated content as a result of this.

SudoBrendan (approved):

PR #3080, opened 2023-08-03
use restricted scc for gatekeeper-manager-role

yjst2012:
Fixes [the gatekeeper doesnot start on 4.10 problem](https://issues.redhat.com/browse/ARO-3843)
It replaces https://github.com/Azure/ARO-RP/pull/3074

Get back SCC for gatekeeper-manager-role, however set it to the most strict one as per https://docs.openshift.com/container-platform/4.11/authentication/managing-security-context-constraints.html

the seccomp profiles is a different set from the SCC that is currently being used, and it does not work well on ocp4.10 without mutating the base cluster which is not very feasible, as per [Enabling the default seccomp profile for all pods](https://docs.openshift.com/container-platform/4.10/security/seccomp-profiles.html#configuring-default-seccomp-profile_configuring-seccomp-profiles)

for 4.11, the restricted-v2 already includes "runtime/default seccompProfile" by default, so no need to have it.

Test plan for issue:
pods are able to start and constrains work

NA


SudoBrendan:
> It replaces https://github.com/Azure/ARO-RP/pull/3074

Can we close that PR then?

pkg/operator/controllers/guardrails/staticresources/gk_controller_manager_deployment.yaml
```diff
@@ -112,8 +112,8 @@ spec:
             - ALL
           readOnlyRootFilesystem: true
           runAsNonRoot: true
-          seccompProfile:
-            type: RuntimeDefault
+          # seccompProfile:
+          #   type: RuntimeDefault
```
SudoBrendan, line 116: same as above.

pkg/operator/controllers/guardrails/staticresources/gk_audit_controller_deployment.yaml
```diff
@@ -100,8 +100,8 @@ spec:
             - ALL
           readOnlyRootFilesystem: true
           runAsNonRoot: true
-          seccompProfile:
-            type: RuntimeDefault
+          # seccompProfile:
+          #   type: RuntimeDefault
```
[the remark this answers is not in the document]
yjst2012: left it commented here is to indicate this is an intended modification from upstream, not a neglected one, so it wont be mistakenly added back when upgrading gatekeeper version in future.

yjst2012:
> > It replaces #3074
> 
> Can we close that PR then?

I thought merging of #3080 will automatically close this one, which makes the relations clearer? 
is it true, or not?

pkg/operator/controllers/guardrails/staticresources/gk_audit_controller_deployment.yaml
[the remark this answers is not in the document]
jaitaiwan: @yjst2012 I think that should be put into a comment in the file then.

jaitaiwan (commented):
One small request and then LGTM

pkg/operator/controllers/guardrails/staticresources/gk_audit_controller_deployment.yaml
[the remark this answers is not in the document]
yjst2012: done

pkg/operator/controllers/guardrails/staticresources/gk_audit_controller_deployment.yaml
```diff
@@ -100,8 +100,8 @@ spec:
             - ALL
           readOnlyRootFilesystem: true
           runAsNonRoot: true
-          seccompProfile:
-            type: RuntimeDefault
+          # seccompProfile: #intended deviation from upstream
```
hlipsig, line 103: Why are we deviating from upstream? Do we have anything that explains our reasoning any what we gain vs what we risk? This looks low risk, but I want to make sure.

hlipsig (commented):
Just have 1 probably silly question about deviation from upstream. This is probably obvious to folks who have been around a while.

pkg/operator/controllers/guardrails/staticresources/gk_audit_controller_deployment.yaml
yjst2012: well, these is a "long" story regarding this, tl;dr is with this enabled, gatekeeper does not start on 4.11 and onwards. more details can be found in  #3074, which is the initial fix for the problem. 
and after some more investigation, this PR came up as a better solution, which applies more strict SCCs then the upstream requires on the gatekeeper roles.

yjst2012: and this deviation is actually suggested by gatekeeper in order to run on openshift 4.x: https://open-policy-agent.github.io/gatekeeper/website/docs/v3.11.x/vendor-specific/#running-on-openshift-4x

hlipsig: Thanks this is really helpful!

PR #3085, opened 2023-08-07
Updating Makefile to install wheel

Shivkumar13:
### Which issue does this PR address:

Whenever `make az` is fired under the newly cloned ARO-RP repo with the necessary exported environment variables, we can see the below errors. This was not the case earlier and `make az` used to work fine. This PR address the same issue.

```
removing 'build/bdist.linux-x86_64/egg' (and everything under it)
Wheel is not available, disabling bdist_wheel hook
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: invalid command 'bdist_wheel'
```

This PR adds the installation for a pre-built binary package format `wheel` in the Makefile,  to address the `Wheel is not available` issue. 

### Test plan for issue:

Tested manually in the local dev environment, `make az` is successful on a newly cloned repo.

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,9 @@ runlocal-rp:
 
 az: pyenv
 	. pyenv/bin/activate && \
-	cd python/az/aro && \
+	cd pyenv/bin && \
+	pip install wheel && \
+	cd ../../python/az/aro && \
 	python3 ./setup.py bdist_egg && \
 	python3 ./setup.py bdist_wheel || true && \
 	rm -f ~/.azure/commandIndex.json # https://github.com/Azure/azure-cli/issues/14997


dem4gus (commented):
We should probably just add these dependencies to `requirements.txt` and have the `make az` command install those.

https://pip.pypa.io/en/stable/reference/requirements-file-format/

dem4gus:
I believe the actual problem is the lack of version pinning in the `make pyenv` command, which runs before `make az`.

Our virtual environment creation is just using `python3 -m venv`, which uses whatever `python3` is on the system. On a lot of fully updated *nix workstations, that ends up being 3.11 which is not supported by `az` at this time. On my dev machine I changed the `pyenv` command to `python3.10 -m venv ...` and was able to run `make az` successfully.

Instead of installing `wheel` explicitly, we need to use a virtual environment creator that allows pinning the python version. The basic `venv` is not capable of that, but [`pyenv`](https://github.com/pyenv/pyenv) is.

dem4gus (changes requested):
See comments regarding Python version.

mociarain:
Personally, I'm a big +1 to tightening up our python stuff (pipenv is great, [poetry](https://python-poetry.org/) is awesome too) but I think this PR is dead. @dem4gus & @Shivkumar13 I was thinking to card up something to effect of "Pin our python version (pipenv/poetry" and closing this.

What do you think?

mociarain:
The intention here is captured in this card: https://issues.redhat.com/browse/ARO-9377

PR #3087, opened 2023-08-08
Remove gvrresolver use in the frontend

hawkowl:

Part of cleanups for https://issues.redhat.com/browse/ARO-3812

Gets rid of one usage of our custom GVRResolver, and instead uses the one from controller-runtime which is capable of doing the refreshing we need.

This means we don't use the cache, but in situations where that matters we often didn't have much luck getting the APIServer to do anything.

### Test plan for issue:

E2E?

N/A


cblecker (dismissed):

cblecker (dismissed):

yjst2012 (dismissed):
LGTM

yjst2012 (approved):

cblecker (approved):

PR #3088, opened 2023-08-08
Update RegistryProfile with new ACR token password once rotated

cadenmarchese:

Fixes https://issues.redhat.com/browse/ARO-3897

Currently, the RegistryProfile in the cluster doc, which stores the original ACR token credentials, is left unchanged when we run PUCM to rotate the ACR token. Because of this, a incomplete PUCM or OperatorUpdate will overwrite the new token password in the cluster with the old one, and depending on that old password's validity, can break clusters from pulling from our prod ACR until another full PUCM is run.

This PR patches the cluster doc with the new RegistryProfile before continuing on to rotate the token in-cluster.

### Test plan for issue:

Steps taken to test:

- Added a quick hack to print the token password on GetCluster (we don't currently return it):
```
	if oc.Properties.RegistryProfiles != nil {
		out.Properties.RegistryProfiles = make([]RegistryProfile, len(oc.Properties.RegistryProfiles))
		for i, v := range oc.Properties.RegistryProfiles {
			out.Properties.RegistryProfiles[i].Name = v.Name
			out.Properties.RegistryProfiles[i].Username = string(v.Password) // hack - fill this field with the password instead so we can see if it's getting updated
		}
	}
```
- Deployed this change + the hack above to INT
- Created a cluster and ran 1 PUCM and several `az aro update` https://portal.microsoftgeneva.com/s/DDE621D0 
- Used GetCluster to confirm that the token password was changing
- Checked in the cluster's `cluster` secret in `openshift-azure-update` to confirm the .dockerconfigjson was changing
- Upgraded the cluster to confirm the cluster could still pull images (this was after rotating the token ~6 times)

No

### DIFF
--- a/pkg/cluster/acrtoken.go
+++ b/pkg/cluster/acrtoken.go
@@ -68,16 +68,25 @@ func (m *manager) rotateACRTokenPassword(ctx context.Context) error {
 		return nil
 	}
 
-	tokenManager, err := acrtoken.NewManager(m.env, m.localFpAuthorizer)
+	token, err := acrtoken.NewManager(m.env, m.localFpAuthorizer)
 	if err != nil {
 		return err
 	}
 
-	registryProfile := tokenManager.GetRegistryProfile(m.doc.OpenShiftCluster)
-	err = tokenManager.RotateTokenPassword(ctx, registryProfile)
+	registryProfile := token.GetRegistryProfile(m.doc.OpenShiftCluster)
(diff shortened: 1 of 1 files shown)


petrkotas (approved):
ok took me a while before I figured the whole flow. I see, when we deploy the operator we pull the cluster document and build on it. During regular pucm the update will happen as a last step, therefore the issue.

PR #3091, opened 2023-08-09
Remove upgrades from the Admin API

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-3925

We haven't used this endpoint in a long time, and if we do ever need to do mass OpenShift upgrades, we should utilise something like MUO instead. Either way, this endpoint doesn't serve any use.

### Test plan for issue:

Unit tests, E2E

Removal of any docs from the wiki?


mbarnes (dismissed):
LGTM, nice cleanup.

cblecker (approved):

PR #3092, opened 2023-08-09
Remove old upgradestreams logic from the monitor & old parts of the RP

hawkowl:

Removing the last bit of stuff that isn't used after https://issues.redhat.com/browse/ARO-3925

Builds off #3091

We don't update the UpgradeStreams meaningfully, so we should just remove it from the codebase. This includes one or two small uses, and mainly the emitted metric (which is fairly useless tbh).

### Test plan for issue:

E2E, unit tests

N/A


mbarnes (approved):
LGTM.  I see #3091 is embedded here so this will need re-approved after rebasing.

mbarnes (approved):
With #3091 merged, LGTM again.

s-amann (approved):
lgtm, +1 for removing bloat

PR #3094, opened 2023-08-12
Move default openshift version

mbarnes:

[[ARO-3896] Read default OCP version from CosmosDB](https://issues.redhat.com/browse/ARO-3896)

Depends on a couple ADO pull requests:
- [RP-Config: Add openShiftVersions data](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/RP-Config/pullrequest/8603646) :heavy_check_mark: _merged_
- [ARO.Pipelines: Read OpenShiftVersions from YAML configuration](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines/pullrequest/8603681) :heavy_check_mark: _merged_

Make CosmosDB the single source of truth for available OpenShift versions, including the default version.  Currently it's a mix: CosmosDB has the available versions but only ARO RP knows the default version.

#### Local Development Mode

Local development mode has been a subject of discussion within my functional team (Loki).  Initially I had planned to implement a small hack to keep local development mode working as is in lieu of an internal default OpenShift version and pullspec.  But the team has grander plans for local development mode, so I'm going to defer that issue here and leave the internal default version in place.  But it's ONLY to be used for local development mode, _not_ in the frontend or in any production pipelines.

### Test plan for issue:

Combination of local development mode and probably deployment to INT to test `aro update-versions` in a pipeline.

The wiki page [Multi-Version OpenShift Installations](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki?anchor=openshift-versions&pagePath=/Build%252Dout/Multi%252DVersion%20OpenShift%20Installations) should direct readers to [RP-Config](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/RP-Config) rather than [pkg/util/version/const.go](https://github.com/Azure/ARO-RP/blob/master/pkg/util/version/const.go) for updating install streams.

### DIFF
--- a/cmd/aro/update_ocp_versions.go
+++ b/cmd/aro/update_ocp_versions.go
@@ -172,10 +172,10 @@ func getVersionsDatabase(ctx context.Context, log *logrus.Entry) (database.OpenS
 
 	m := statsd.New(ctx, log.WithField("component", "update-ocp-versions"), _env, os.Getenv("MDM_ACCOUNT"), os.Getenv("MDM_NAMESPACE"), os.Getenv("MDM_STATSD_SOCKET"))
 
-	if err := env.ValidateVars(KeyVaultPrefix); err != nil {
+	if err := env.ValidateVars(envKeyVaultPrefix); err != nil {
 		return nil, err
 	}
-	keyVaultPrefix := os.Getenv(KeyVaultPrefix)
+	keyVaultPrefix := os.Getenv(envKeyVaultPrefix)
 	serviceKeyvaultURI := keyvault.URI(_env, env.ServiceKeyvaultSuffix, keyVaultPrefix)
 	serviceKeyvault := keyvault.NewManager(msiKVAuthorizer, serviceKeyvaultURI)
 
@@ -184,11 +184,11 @@ func getVersionsDatabase(ctx context.Context, log *logrus.Entry) (database.OpenS
 		return nil, err
 	}
 
-	if err := env.ValidateVars(DatabaseAccountName); err != nil {
+	if err := env.ValidateVars(envDatabaseAccountName); err != nil {
 		return nil, err
 	}
 
-	dbAccountName := os.Getenv(DatabaseAccountName)
+	dbAccountName := os.Getenv(envDatabaseAccountName)
 	clientOptions := &policy.ClientOptions{
 		ClientOptions: _env.Environment().ManagedIdentityCredentialOptions().ClientOptions,
 	}
--- a/cmd/aro/update_ocp_versions.go
+++ b/cmd/aro/update_ocp_versions.go
@@ -51,7 +51,7 @@ func getEnvironmentData(envKey string, envData any) error {
 }
 
 func getOpenShiftVersions() (*OpenShiftVersions, error) {
-	const envKey = "OPENSHIFT_VERSIONS"
+	const envKey = envOpenShiftVersions
 	var openShiftVersions OpenShiftVersions
 
 	if err := getEnvironmentData(envKey, &openShiftVersions); err != nil {
@@ -72,7 +72,7 @@ func getInstallerImageDigests() (map[string]string, error) {
 	// the aro-installer wrapper digest.  This allows us to utilize
 	// Azure Safe Deployment Practices (SDP) instead of pushing the
 	// version tag and deploying to all regions at once.
-	const envKey = "INSTALLER_IMAGE_DIGESTS"
+	const envKey = envInstallerImageDigests
 	var installerImageDigests map[string]string
 
 	if err := getEnvironmentData(envKey, &installerImageDigests); err != nil {
--- a/cmd/aro/update_ocp_versions.go
+++ b/cmd/aro/update_ocp_versions.go
@@ -22,6 +22,12 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
+// Corresponds to configuration.openShiftVersions in RP-Config
+type OpenShiftVersions struct {
+	DefaultStream  map[string]string
+	InstallStreams map[string]string
+}
+
 func getEnvironmentData(envKey string, envData any) error {
 	var err error
 
@@ -44,6 +50,23 @@ func getEnvironmentData(envKey string, envData any) error {
 	return nil
 }
 
+func getOpenShiftVersions() (*OpenShiftVersions, error) {
+	const envKey = "OPENSHIFT_VERSIONS"
+	var openShiftVersions OpenShiftVersions
+
+	if err := getEnvironmentData(envKey, &openShiftVersions); err != nil {
+		return nil, err
+	}
+
+	// The DefaultStream map must have exactly one entry.
+	numDefaultStreams := len(openShiftVersions.DefaultStream)
+	if numDefaultStreams != 1 {
+		return nil, fmt.Errorf("%s: DefaultStream must have exactly 1 entry, found %d", envKey, numDefaultStreams)
+	}
+
+	return &openShiftVersions, nil
+}
+
 func getInstallerImageDigests() (map[string]string, error) {
 	// INSTALLER_IMAGE_DIGESTS is the mapping of a minor version to
 	// the aro-installer wrapper digest.  This allows us to utilize
@@ -59,6 +82,34 @@ func getInstallerImageDigests() (map[string]string, error) {
 	return installerImageDigests, nil
 }
 
+func appendOpenShiftVersions(ocpVersions []api.OpenShiftVersion, installStreams map[string]string, installerImageName string, installerImageDigests map[string]string, isDefault bool) ([]api.OpenShiftVersion, error) {
+	for fullVersion, openShiftPullspec := range installStreams {
+		openShiftVersion, err := version.ParseVersion(fullVersion)
+		if err != nil {
+			return nil, err
+		}
+		fullVersion = openShiftVersion.String() // trimmed of whitespace
+		minorVersion := openShiftVersion.MinorVersion()
+		installerDigest, ok := installerImageDigests[minorVersion]
+		if !ok {
+			return nil, fmt.Errorf("no installer digest for version %s", minorVersion)
+		}
+		installerPullspec := fmt.Sprintf("%s:%s@%s", installerImageName, minorVersion, installerDigest)
+
+		ocpVersions = append(ocpVersions, api.OpenShiftVersion{
+			Properties: api.OpenShiftVersionProperties{
+				Version:           fullVersion,
+				OpenShiftPullspec: openShiftPullspec,
+				InstallerPullspec: installerPullspec,
+				Enabled:           true,
+				Default:           isDefault,
+			},
+		})
+	}
+
+	return ocpVersions, nil
+}
+
 func getLatestOCPVersions(ctx context.Context, log *logrus.Entry) ([]api.OpenShiftVersion, error) {
 	env, err := env.NewCoreForCI(ctx, log)
 	if err != nil {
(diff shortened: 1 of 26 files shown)


pkg/api/v20191231preview/openshiftcluster_validatestatic_test.go
```diff
@@ -51,7 +50,7 @@ func validOpenShiftCluster() *OpenShiftCluster {
 			ClusterProfile: ClusterProfile{
 				PullSecret:      `{"auths":{"registry.connect.redhat.com":{"auth":""},"registry.redhat.io":{"auth":""}}}`,
 				Domain:          "cluster.location.aroapp.io",
-				Version:         version.DefaultInstallStream.Version.String(),
+				Version:         "4.10.0",
```
jhoreman, line 53: Hey Matt, why are we looking at 4.10 specifically here?

mbarnes: No real reason, I just needed _some_ version to plug into `ClusterProfile` structs.

cmd/aro/update_ocp_versions.go
```diff
@@ -33,57 +38,111 @@ func getInstallerImageDigests(envKey string) (map[string]string, error) {
...
 
-	return installerImageDigests, nil
+	return nil
 }
 
-func getLatestOCPVersions(ctx context.Context, log *logrus.Entry) ([]api.OpenShiftVersion, error) {
-	env, err := env.NewCoreForCI(ctx, log)
-	if err != nil {
+func getOpenShiftVersions() (*OpenShiftVersions, error) {
+	const envKey = "OPENSHIFT_VERSIONS"
```
SudoBrendan, line 53: nit: would it be better to move this to a package-level const so it appears at the top of the file? Same for `INSTALLER_IMAGE_DIGESTS` below?

pkg/frontend/openshiftversions_list.go
```diff
@@ -42,13 +41,5 @@ func (f *frontend) getEnabledInstallVersions(ctx context.Context) []*api.OpenShi
 	}
 	f.mu.RUnlock()
 
-	if len(versions) == 0 {
```
SudoBrendan, line 45: This is an interesting symptom of this change - I feel like we should have this case outlined somewhere, right? Maybe we decide that it lands in the RP release pipeline or something? We need to guarantee that we'll always return at least one version (defaultversion) to the customer at all times and prevent rollouts that would make this possible (we could also do this at an RP level, for example, we exit `1` if no default version is found, and that happens way before a request would reach `frontend`).

pkg/monitor/cluster/clusterversions.go
```diff
@@ -51,13 +51,12 @@ func (mon *Monitor) emitClusterVersions(ctx context.Context) error {
 	mon.emitGauge("cluster.versions", 1, map[string]string{
 		"actualVersion":                        actualVersion,
 		"desiredVersion":                       desiredVersion(cv),
-		"provisionedByResourceProviderVersion": mon.oc.Properties.ProvisionedBy,                     // last successful Put or Patch
-		"resourceProviderVersion":              version.GitCommit,                                   // RP version currently running
-		"operatorVersion":                      operatorVersion,                                     // operator version in the cluster
-		"availableVersion":                     availableVersion(cv, version.UpgradeStreams),        // current available version for upgrade from stream
-		"availableRP":                          availableRP,                                         // current RP version available for document update, empty when none
-		"latestGaMinorVersion":                 version.DefaultInstallStream.Version.MinorVersion(), // Latest GA in ARO Minor version
```
SudoBrendan, line 59: is there a reason we've dropped the `latestGaMinorVersion` metric?

SudoBrendan (commented):
Overall this is looking pretty solid, a few questions. Thanks for the PR!

cmd/aro/update_ocp_versions.go
mbarnes: +0, I can do that.

pkg/monitor/cluster/clusterversions.go
mbarnes: I couldn't find any evidence of its use, and `version.DefaultInstallStream` is strictly for local development only now.

pkg/frontend/openshiftversions_list.go
mbarnes: That brings up an important point: we need to run the **Update ARO OpenShift Versions** pipelines to deploy the default flag _before_ we release an RP with this change so that the scenario you're talking about can't happen.  If we're releasing the RP next week, then we should either start a version update rollout this week or hold this PR until September.

I did add a [check to "aro update-versions"](https://github.com/Azure/ARO-RP/blob/14f7393dd0c1a12bd85d79d8293bc6706701d30b/cmd/aro/update_ocp_versions.go#L60-L64) to ensure the `defaultStreams` map has one and only one item.  We can't really block RP rollouts over this but we can block OCP version updates.

mbarnes: Additionally, the absence of a default version in CosmosDB is something I think SREs should be alerted to so the database can be corrected.  Maybe a followup task would be to add an `OpenShiftVersions` changefeed listener in the metrics code and emit a default version metric that we can craft an alert around?

cmd/aro/update_ocp_versions.go
mbarnes: It turns out we have a whole file (`/cmd/aro/const.go`) for environment variable names, but it needed a touch of rework.  See additional commits.

mbarnes:
E2E will fail until [ARO.Pipelines: Read OpenShiftVersions from YAML configuration](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines/pullrequest/8603681) is merged.

mbarnes:
Needed a final commit to use the internal default OCP version for running E2E tests for CI.

SudoBrendan (dismissed):

mbarnes:
Placing a **Hold** on this while I test a concern I had today.  The code here is technically correct, but getting this all deployed without breaking anything may require more steps than I expected.

mbarnes:
Confirmed my concern, unfortunately.  If we deploy the default flag to CosmosDB before the RP is aware of the flag, it will break the changefeed process in the RP frontend with the error below and the RP will fail to read valid OCP versions from CosmosDB.

```
pkg/frontend/changefeed.go:39 frontend.(*frontend).updateFromIterator() json decode error [pos 626]:
no matching struct field found when decoding stream map with key default  component=frontend
```

That means we need to make the RP aware of the default flag _before_ deploying the default flag.  So not one but _two_ RP releases:

1. Release an RP with only the default flag added to `OpenShiftVersionProperties`.  In other words, just the [1st commit](https://github.com/Azure/ARO-RP/pull/3094/commits/be25278a370c98bab995daf05e08988dcbf83611).
2. Merge this pull request and run the "Update ARO OpenShift Versions" pipeline to deploy the default version flag out to all regions.
3. Release an RP with the rest of the commits here to utilize the default version flag.

mbarnes:
Split the 1st commit into https://github.com/Azure/ARO-RP/pull/3202.

pkg/util/version/const.go
```diff
@@ -30,44 +30,12 @@ type Stream struct {
...
-		Version:  NewVersion(4, 10, 40),
-		PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:b9fad814fb4442e7e852b0614d9bb4e2ebc5e1a2fa51623aa838b4ee0e4a5369",
-	},
-	DefaultInstallStreams[11],
-	{
-		Version:  NewVersion(4, 11, 26),
-		PullSpec: "quay.io/openshift-release-dev/ocp-release@sha256:1c3913a65b0a10b4a0650f54e545fe928360a94767acea64c0bd10faa52c945a",
-	},
-	DefaultInstallStreams[12],
+// Install stream data for production and INT has moved to RP-Config.
```
petrkotas, line 33: I am wondering, should we also move other image versions to a database?

pkg/frontend/openshiftversions_list.go
petrkotas: This is something that would not be changing continuously. I believe only simple test added to our e2e suite verifying the presence would be sufficient. Running e2e is mandatory step after rollout. This would alert us soon enough making manual corrective action possible.

petrkotas (approved):
Hi @mbarnes to clarify, the process of updating versions in production stays the same. The twist is we need make sure we run it before release. Otherwise, this is really nice simplification. Thank you.

pkg/util/version/const.go
mbarnes: I think it makes sense; it's configuration data.  Downsides are having to repeat the data in each of the four RP-Config cloud files, and it potentially complicates local development mode if it depends on any of those images.

SudoBrendan:
Hey, this was tagged as `next-release` but I think we're waiting on a release of https://github.com/Azure/ARO-RP/pull/3202 before this can merge, right?

mbarnes:
> Hey, this was tagged as next-release but I think we're waiting on a release of https://github.com/Azure/ARO-RP/pull/3202 before this can merge, right?

You're right, I guess we haven't cut a release since that merged.  So this needs to wait til **next-next-release**.

mbarnes:
@SudoBrendan and @petrkotas had previously approved this.

cmd/aro/update_ocp_versions.go
```diff
@@ -34,57 +39,111 @@ func getInstallerImageDigests(envKey string) (map[string]string, error) {
...
 		jsonData, err = base64.StdEncoding.DecodeString(string(jsonData))
 		if err != nil {
-			return nil, fmt.Errorf("%s: Failed to decode base64: %v", envKey, err)
+			return fmt.Errorf("%s: Failed to decode base64: %v", envKey, err)
 		}
 	}
 
-	if err = json.Unmarshal(jsonData, &installerImageDigests); err != nil {
-		return nil, fmt.Errorf("%s: Failed to parse JSON: %v", envKey, err)
+	if err = json.Unmarshal(jsonData, envData); err != nil {
```
SudoBrendan, line 46: Looking another time, `data` feels a little too generic... Can we make this something like `envInstallerImageDigests` or similar, or is it too late?

cmd/aro/update_ocp_versions.go
```diff
@@ -34,57 +39,111 @@ func getInstallerImageDigests(envKey string) (map[string]string, error) {
 	if !env.IsLocalDevelopmentMode() {
 		jsonData, err = base64.StdEncoding.DecodeString(string(jsonData))
 		if err != nil {
-			return nil, fmt.Errorf("%s: Failed to decode base64: %v", envKey, err)
+			return fmt.Errorf("%s: Failed to decode base64: %v", envKey, err)
```
SudoBrendan, line 42: scope creep, I know... Feel free to :hear_no_evil: and we can groom additional work...

...but while we're touching this, can we maybe make these log messages something that would be useful in production, even adding a bit of context like "Failed to base64 decode install image digests" and use `%w` instead of `%v`?

cmd/aro/update_ocp_versions.go
```diff
@@ -34,57 +39,111 @@ func getInstallerImageDigests(envKey string) (map[string]string, error) {
...
 		}
 	}
 
-	if err = json.Unmarshal(jsonData, &installerImageDigests); err != nil {
-		return nil, fmt.Errorf("%s: Failed to parse JSON: %v", envKey, err)
+	if err = json.Unmarshal(jsonData, envData); err != nil {
+		return fmt.Errorf("%s: Failed to parse JSON: %v", envKey, err)
```
SudoBrendan, line 47: :hear_no_evil: same scope-creep as above, make this generic error useful by including a reference to image digests, and use `%w`

SudoBrendan (approved):
left some comments/nits. Thanks!

cmd/aro/update_ocp_versions.go
mbarnes: It's meant to be generic because it's not specific to installer image digests.

I guess it's easy to miss in the diff, but this logic is now part of a helper function named `getEnvironmentData` that's used for both installer image digests and openshift versions.

cmd/aro/update_ocp_versions.go
mbarnes: Same answer as above, though note it does print the environment variable name first.

So it's going to be one of:
```
INSTALLER_IMAGE_DIGESTS: Failed to parse JSON: ...

OPENSHIFT_VERSIONS: Failed to parse JSON: ...
```

I can change `%v` to `%w` although I'm not sure I understand the difference.

mbarnes: Updated the "Refactoring" commit to change `%v` to `%w`.

kimorris27 (commented):
Looks good to me! Have the changes been tested in INT yet?

jhoreman (approved):
Was able to personally test this in INT. Updating the default version in the cosmosDB document for OpenShiftVersions changes the default install version as expected and desired.

SudoBrendan (approved):

kimorris27 (approved):
Reviewed this back in December but am giving my official green check mark now that it's been tested in INT 🙂

cmd/aro/update_ocp_versions.go
```diff
@@ -22,8 +22,13 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
-func getInstallerImageDigests(envKey string) (map[string]string, error) {
-	var installerImageDigests map[string]string
+// Corresponds to configuration.openShiftVersions in RP-Config
+type OpenShiftVersions struct {
+	DefaultStream  map[string]string
+	InstallStreams map[string]string
+}
```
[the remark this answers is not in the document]
mbarnes: FWIW, [ARO.Pipelines has the full struct and parser](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO.Pipelines?path=/ev2/generator/rpconfig_struct.go) you're looking for.  What's happening in this context is the parser is extracting the version bits we need from RP-Config and passing it to the RP as a JSON blob by way of an environment variable.  All we're parsing here is the environment varaible, not RP-Config directly.

PR #3095, opened 2023-08-14
multiple ARO managed public IPs per Load balancer during cluster update

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-3110

This PR allows for adding additional rp managed ips to outbound-rule-v4 via az aro update to allow for scaling clusters past 62 nodes.  The overall process looks like this:

1. Get the public load balancer and if needed update the cluster docs "effectiveOutboundIPs"
2. Get desired Outbound IPs, creating additional managed IPs if needed.
3. Determine if an update is needed by comparing the desired IPs to effectiveOutboundIPs.
4. Update the load balancer with new frontend ip config / outbound-rule-v4 frontend ip config
6. Patch effectiveOutboundIPs
7. Delete any unused rp managed IP addresses.

The default outbound IP is also shared with the api server when the api server visibility is public, therefore the api server frontend ip config will always be saved when api server is public.  The default outbound IP/api server is always reused when updating managedOutboundIPs.Count.
NOTE:  In future iterations when transitioning from ManagedIPs -> Customer provider IPs/Prefixes the default outbound IP will be deleted if the API Server is not public.

I tried to write this in such a way to allow for expansion when adding customer provided public IPs and public IP prefixes.  It may have been easy enough to add it to the code at this time but it would require a lot of additional test cases.

### Test plan for issue:

Added unit tests and tested this manually.

Testing this functionality with a live cluster is fairly simple.  Create a cluster via local-rp and use the aro_ext with PR    https://github.com/Azure/ARO-RP/pull/3073 checked out.  Add / remove outbound IPs with  `az aro update -g <resourcegroup> -n <clustername> --load-balancer-managed-outbound-ip-count 4`

Yes, to be completed by doc writer.


pkg/validate/dynamic/loadbalancerprofile.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	netUsages, err := dv.spNetworkUsage.List(ctx, oc.Location)
+	if err != nil {
+		return err
+	}
+
+	for _, netUsage := range netUsages {
+		if *netUsage.Name.Value == "PublicIPAddresses" {
+			if int64(requestedIPs) > (*netUsage.Limit - *netUsage.CurrentValue) {
```
tsatam, line 58: Validate that pointer dereferencing does not fail from nil pointers

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+// Remove all frontend ip config in use by outbound-rule-v4.  Frontend IP config that is used by load balancer rules will be saved.
+func removeOutboundIPsFromLB(lb mgmtnetwork.LoadBalancer) {
+	// get all outbound rule fip config to remove
+	var obRuleV4FIPConfigMap = make(map[string]mgmtnetwork.SubResource)
+	for _, obRule := range *lb.LoadBalancerPropertiesFormat.OutboundRules {
+		if *obRule.Name == "outbound-rule-v4" {
+			for i := 0; i < len(*obRule.OutboundRulePropertiesFormat.FrontendIPConfigurations); i++ {
+				obRuleV4FIPConfigMap[*(*obRule.OutboundRulePropertiesFormat.FrontendIPConfigurations)[i].ID] = (*obRule.OutboundRulePropertiesFormat.FrontendIPConfigurations)[i]
```
tsatam, line 95: extract keys out to a variable to make this easier to read

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,386 @@
...
+
+// Reconcile the outbound rule "outbound-rule-v4" frontend IP Config.
+func (m *manager) reconcileOBRuleV4OBIPs(ctx context.Context, lb mgmtnetwork.LoadBalancer) error {
+	m.log.Info("reconciling outbound-rule-v4")
+	defer func() {
+		err := m.deleteUnusedManagedIPs(ctx)
+		if err != nil {
+			m.log.Error("failed to cleanup unused managed IPs, error: %w", err)
+		}
+	}()
```
SudoBrendan, line 48: this might update IPs, but not patch cosmos EffectiveIPs, right? What happens when we hit the first error case, and we can't run patchEffectiveOutboundIPs for the first time, then we attempt to delete managedIPs?

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,386 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// Remove all frontend ip config in use by outbound-rule-v4.  Frontend IP config that is used by load balancer rules will be saved.
+func removeOutboundIPsFromLB(lb mgmtnetwork.LoadBalancer) {
+	// get all outbound rule fip config to remove
+	var obRuleV4FIPConfigMap = make(map[string]mgmtnetwork.SubResource)
```
SudoBrendan, line 91: nit: `ConfigMap` has other meanings in our industry so this is slightly confusing, even something like `ConfigsMap` is a little clearer IMO, or you can stick with just `Configs`, since golang will give us the struct type.

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,386 @@
...
+		}
+		savedFIPConfig = append(savedFIPConfig, fipConfig)
+	}
+	lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations = &savedFIPConfig
+}
+
+// Adds IPs or IPPrefixes to the load balancer outbound rule "outbound-rule-v4".
+func addOutboundIPsToLB(resourceGroupID string, lb mgmtnetwork.LoadBalancer, obIPsOrIPPrefixes []api.ResourceReference) {
+	// map out frontendConfig to ID of public IP addresses for quick lookup
+	var frontendIPConfigMap = make(map[string]mgmtnetwork.FrontendIPConfiguration, len(*lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations))
```
SudoBrendan, line 122: nit: same `ConfigMap` naming as above

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,386 @@
...
+	var obRuleV4FIPConfigMap = make(map[string]mgmtnetwork.SubResource)
+	for _, obRule := range *lb.LoadBalancerPropertiesFormat.OutboundRules {
+		if *obRule.Name == "outbound-rule-v4" {
+			for i := 0; i < len(*obRule.OutboundRulePropertiesFormat.FrontendIPConfigurations); i++ {
+				fipConfigID := *(*obRule.OutboundRulePropertiesFormat.FrontendIPConfigurations)[i].ID
+				fipConfig := (*obRule.OutboundRulePropertiesFormat.FrontendIPConfigurations)[i]
+				obRuleV4FIPConfigMap[fipConfigID] = fipConfig
+			}
+			// clear outbound-rule-v4 frontend ip config
+			*obRule.FrontendIPConfigurations = []mgmtnetwork.SubResource{}
```
SudoBrendan, line 100: nit: can optimize with a `break` here, if we're only expecting one Rule with this `Name`.

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,386 @@
...
+		doc.OpenShiftCluster.Properties.NetworkProfile.LoadBalancerProfile.EffectiveOutboundIPs = effectiveOutboundIPs
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func frontendIPConfig(name string, id string, publicIPorIPPrefixID string) mgmtnetwork.FrontendIPConfiguration {
```
SudoBrendan, line 342: nit: probably best to have this start with `new` since it's a constructor.

SudoBrendan (commented):
Left some comments/nits/questions. Mostly though, I'm struggling to understand most of this because most functions here are quite complex - it's often that any given function has at least 2 `for` loops, which makes it a bit difficult to parse exactly what's happening. I'm not sure how to address that complexity here, just I guess saying "geeze, this is hard!". I'll have to sleep on this and come back tomorrow with any suggestions I come up with, if any.

For future reviewers, I think that `loadbalancerprofile.go` is definitely the majority of complexity here, even if things are honestly pretty simple. I spent some time making a "code map" of what is happening at a high level, which makes this a lot easier to understand:

```
1. ReconcileLBProfile
    A. reconcileOBRuleV4OBIPs
        I. first time? yes
            a. cosmos.patchEffectiveOutboundIPs
        II. anything to resolve? yes
            a. removeOutboundIPsFromLB
                i. remove "outbound-rule-v4" OutboundRule FIP Configs
                ii. remove "outbound-rule-v4" FIP configs from LBProperties FIP configs
            b. addOutboundIPstoLB
                i. foreach provided IP
                    - if it isn't already a FIP Config, create it
                    - add FIP Config to LBProperties
                ii. set "outbound-rule-v4" FIP Configs to the new FIP Configs
            c. lb.CreateOrUpdateAndWait
            d. cosmos.patchEffectiveOutboundIPs
        III. deleteUnusedManagedIPs
```

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,386 @@
...
+
+// Adds IPs or IPPrefixes to the load balancer outbound rule "outbound-rule-v4".
+func addOutboundIPsToLB(resourceGroupID string, lb mgmtnetwork.LoadBalancer, obIPsOrIPPrefixes []api.ResourceReference) {
+	// map out frontendConfig to ID of public IP addresses for quick lookup
+	var frontendIPConfigMap = make(map[string]mgmtnetwork.FrontendIPConfiguration, len(*lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations))
+	for i := 0; i < len(*lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations); i++ {
+		fipConfigIPID := *(*lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations)[i].FrontendIPConfigurationPropertiesFormat.PublicIPAddress.ID
+		fipConfig := (*lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations)[i]
+		frontendIPConfigMap[fipConfigIPID] = fipConfig
+	}
```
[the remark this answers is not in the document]
petrkotas: Yes please, with clean to understand naming

petrkotas (dismissed):
Hi @tony-schndr thank you for the PR, it is a really good work. I believe when the small nits addressed and questions answered, we can merge it. And thank you for providing a lot of explaining comments.

pkg/cluster/loadbalancerprofile.go
tony-schndr: What is "this" in your question?  If you are talking about the process as a whole, the load balancer could be updated with a change in managed IPs but then cosmos patch may fail.  This would run and cleanup unused managed IPs but the value of `effectiveOutboundIPs` would be wrong.  If "this" is the `m.deleteUnusedManagedIPs` then only IPs that are not attached to the loadbalancer are cleaned up.  

Overall this tries to prevent orphaning IP addresses if the process fails at any point.

SudoBrendan: sorry - "this" meaning the entire `defer`. The path I outline is:

1. New cluster, need to patch cosmos as step 1
2. Cosmos write fails (e.g., cosmos is unavailabe?), returns an error, but before we can return, we do this `defer` statement

Does that path make sense to execute on?

SudoBrendan:
Back from sleep. Common patterns I'm seeing here include:

1. Find matching values that meet criteria (e.g. find all FIPConfig.ID that match "outbound-rule-v4")
2. Compare StructA's FOO field with StructB's BAR field (e.g. LBProfile.FIPConfig.ID to OBRule.FIPConfig.ID), this could probably be a function
3. Remove a Struct's FOO field based on a match (e.g. remove LBProfile.FIPConfig.ID by ID value)

I think this could simplify a lot of these functions, because we could do something like:

```
idsToRemove, err := findFIPConfigIdsByRuleName(lb, "outbound-rule-v4"))
err := removeLBProfileConfigsByIds(lb, idsToRemove)
err := removeOutboundProfileConfigsByIds(lb, idsToRemove)
```

I don't think we need to do this to get it merged, but might be a good follow-up PR.

pkg/validate/dynamic/loadbalancerprofile.go
```diff
@@ -0,0 +1,102 @@
...
+
+	totalBackendInstances := len(*backendPools.BackendAddressPoolPropertiesFormat.BackendIPConfigurations)
+	// TODO: update once allocatedOutboundPorts is implemented
+	allocatedOutboundPorts := 1024
+	var desiredNumIPs int
+	// TODO: add OutboundIPs and OutboundIPPrefixes
+	if oc.Properties.NetworkProfile.LoadBalancerProfile.ManagedOutboundIPs != nil {
+		desiredNumIPs = oc.Properties.NetworkProfile.LoadBalancerProfile.ManagedOutboundIPs.Count
+	}
+	totalSNATPorts := desiredNumIPs * 63992
```
niontive, line 89: Where does 63992 come from?

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+func newOutboundRuleFrontendIPConfig(id string) mgmtnetwork.SubResource {
+	return mgmtnetwork.SubResource{
+		ID: &id,
+	}
+}
+
+// Reports if two []api.ResourceReference are equal.
+func areResourceRefsEqual(a, b []api.ResourceReference) bool {
```
niontive, line 372: Could we use `reflect.DeepEqual` here?

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,396 @@
...
+			}
+		}
+	}
+
+	return outboundIPs
+}
+
+func (m *manager) patchEffectiveOutboundIPs(ctx context.Context, outboundIPs []api.ResourceReference) error {
+	m.log.Info("patching effectiveOutboundIPs")
+	effectiveOutboundIPs := make([]api.EffectiveOutboundIP, 0, len(outboundIPs))
```
niontive, line 337: I noticed in line 56, we check if `effectiveOutboundIPs` is nil. If we use `make` here, this will never be nil, so is the nil case only for new clusters? If a cluster has no effective outbound IPs, we just represent this as an empty slice?

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+
+	return publicIPAddress, nil
+}
+
+func getOutboundIPsFromLB(lb mgmtnetwork.LoadBalancer) []api.ResourceReference {
+	var outboundIPs []api.ResourceReference
+	fipConfigs := make(map[string]mgmtnetwork.FrontendIPConfiguration, len(*lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations))
+
+	for i := 0; i < len(*lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations); i++ {
```
niontive, line 315: Could we get a nil pointer deref?

pkg/cluster/loadbalancerprofile.go
tony-schndr: I did change this function to return the result of `reflect.DeepEqual` but I found the lists must be sorted and lowercased for reflect.DeepEqual to work correctly.

pkg/cluster/loadbalancerprofile.go
tony-schndr: The nil case is only for new clusters.  If a cluster has a outboundType of `Loadbalancer` there will always be at least 1 effective outbound IP.  A cluster with outboundType of `UserDefinedRouting` will not have a `LoadBalancerProfile` thus no `effectiveOutboundIPs`.

pkg/cluster/loadbalancerprofile.go
tony-schndr: There should always be data in  `*lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations`.  There should  be at least 1 FrontendIPConfiguration containing the IP used by the outbound rule `outbound-rule-v4`.

pkg/validate/dynamic/loadbalancerprofile.go
tony-schndr: From what I can tell digging through docs/code we shouldn't hit any nil references here, the netUsage is present it should have fields populated.

Also we have been running this same operation on the frontend for some time now with out issue. 
 https://github.com/Azure/ARO-RP/blob/f26f8f41d034618b2042002a90e255f21be1fcc5/pkg/frontend/quota_validation.go#L90-L100

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,412 @@
...
+	if cleanupErrors != nil {
+		return fmt.Errorf("failed to cleanup unused managed ips\n%s", strings.Join(cleanupErrors, "\n"))
+	}
+
+	return nil
+}
+
+// Returns the desired RP managed outbound publicIPAddresses.  Additional Managed Outbound IPs
+// will be created as required to satisfy ManagedOutboundIP.Count.
+func (m *manager) getDesiredOutboundIPs(ctx context.Context) ([]api.ResourceReference, error) {
```
SudoBrendan, line 218: nit: similar to earlier - this `get` function is mutating since reconcileDesiredManagedIPs does things like create new IP addresses.

SudoBrendan (approved):
Thanks for fixing the cosmos bug, I think this LGTM and if we want additional changes/refactors they can land in subsequent PRs.

pkg/cluster/loadbalancerprofile.go
tony-schndr: I will address this in a follow up PR.

SudoBrendan (approved):

rogbas (approved):
LGTM

s-amann (approved):

pkg/cluster/loadbalancerprofile.go
```diff
@@ -0,0 +1,374 @@
...
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func frontendIPConfig(name string, id string, publicIPorIPPrefixID string) mgmtnetwork.FrontendIPConfiguration {
+	// TODO: add check for publicIPorIPPrefixID
```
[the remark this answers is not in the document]
SudoBrendan: This'll land in a subsequent phase - details are in the PRD linked in the Epic.

PR #3097, opened 2023-08-15
Fix local make admin.kubeconfig

cadenmarchese:

Fixes an issue with running `make admin.kubeconfig` locally. The issue was introduced in an earlier PR.

We should no longer need to specify path when invoking jq since we now have https://github.com/Azure/ARO-RP/pull/3050. Specifying the path to jq fails locally when running make admin.kubeconfig unless a jq binary is present in the current working directory.

### Test plan for issue:

- CI should pass.
- Local `make admin.kubeconfig` should pass.

### DIFF
--- a/hack/get-admin-kubeconfig.sh
+++ b/hack/get-admin-kubeconfig.sh
@@ -6,7 +6,7 @@ if [[ "$#" -ne 1 ]]; then
 fi
 
 if [[ $CI ]]; then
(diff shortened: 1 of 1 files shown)


tsatam (approved):

jhoreman (approved):
LGTM

ross-bryan (approved):

PR #3098, opened 2023-08-15
Add hack script to build hive install-log-regexes from ARO definition

tsatam:
> This is a spike PR to evaluate feasibility/desirability of a "single source of truth" for our install failure definitions, which can then be used by both our Hive setup as well as the RP. It _should not_ be merged as-is for now.  

Building block for [ARO-3220](https://issues.redhat.com/browse/ARO-3220)

- Adds a definition for ARO-specific install failure reasons, maintained in Go, that defines specific regular expressions to identify reasons from a failed install's log. 
- Adds a hack script to build a Hive `additional-install-log-regexes` ConfigMap, to supersede the [default rules](https://github.com/openshift/hive/blob/master/config/configmaps/install-log-regexes-configmap.yaml) provided by Hive with ARO/Azure-specific cases, populated from the above definitions. 

This work item will eventually involve building RP-side handling for these install failures, to directly alert customers when the reason is customer-inflicted, that will also build off of these definitions. The work will either be a part of a future PR or future commits on this branch, TBD. 

### Test plan for issue:

- Unit tests have been added for the regular expressions themselves (against "fake" install logs), as well as for the generator's conversion logic.
- The generated `additional-install-log-regexes` ConfigMap has been tested in a Hive cluster and ARO cluster creations have been performed that reproduce the scenarios within this PR, and validate that Hive has propagated the expected Reasons. 

No.

### DIFF
--- a/hack/genhiveconfig/genhiveconfig.go
+++ b/hack/genhiveconfig/genhiveconfig.go
@@ -1,0 +1,84 @@
+package main
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"os"
+
+	"github.com/ghodss/yaml"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	failure "github.com/Azure/ARO-RP/pkg/hive/failure"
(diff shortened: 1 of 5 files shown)


tsatam:
Changes included in https://github.com/Azure/ARO-RP/pull/3116

PR #3099, opened 2023-08-16
ARO-3455: Added Gator Tests for all AccessModes

edisonLcardenas:
[ARO-3455](https://issues.redhat.com/browse/ARO-3455)

Fixes

Added gator tests for all AccessModes of PV.

### Test plan for issue:
Gator Test

No.


SudoBrendan:
I started a slack thread to discuss this internally. I'm against a merge, not because of the code, but because of the feature itself. I'm looking forward to that discussion. Thank you!

yjst2012 (changes requested):
Hey @edisonLcardenas looks like the casts are a bit improper as we are not stopping the creation of PVs backed by r/w host path, can you please either revise or withdraw it? thanks.

PR #3103, opened 2023-08-16
Use CodeQL3000 tasks in OneBranch pipeline

tsatam:

Part of the fix for [ARO-3957](https://issues.redhat.com/browse/ARO-3957)

Uses Microsoft CodeQL ADO tasks for code scanning on master branch in our Onebranch pipeline, rather than our custom task across all branches/PRs.

### Test plan for issue:

This can unfortunately only be tested after merge, as we cannot trigger this task from a pull request, and the only branch configured for OneBranch is master. 

N/A (CI fix)

### DIFF
--- a/.pipelines/ci.yml
+++ b/.pipelines/ci.yml
@@ -1,17 +1,4 @@
 # Azure DevOps Pipeline running CI
-#
-# Note: This pipeline uses a secret variable "github_codeql_upload_token".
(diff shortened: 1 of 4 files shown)


cblecker:
@tsatam Once this is merged, will it still be something we can evaluate on PRs?

tsatam:
Not immediately. We're trying to see if we can get a slimmed down version of the CodeQL Go scans running in a GitHub Actions runner but no luck thus far.

PR #3105, opened 2023-08-16
Fix CodeQL GitHub Action

cblecker:
Fixes ARO-3957

* Removes Golang CodeQL check from ADO CI pipeline
* Adds Golang CodeQL check to the existing GitHub Action, but uses a hosted 1ES pool for golang

### Test plan for issue:
This is a presubmit CI job, so a green test here means :shipit: 

Yes, will need to document the 1ES pool

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,29 +1,21 @@
-# For most projects, this workflow file will not need changing; you simply need
-# to commit it to your repository.
-#
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
-#
-# ******** NOTE ********
-# We have attempted to detect the languages in your repository. Please check
-# the `language` matrix defined below to confirm you have the correct set of
-# supported CodeQL languages.
-#
 name: "CodeQL"
 
 on:
   push:
-    branches: [master]
+    tags:
+      - v*
+    branches:
+      - master
   pull_request:
-    # The branches below must be a subset of the branches above
-    branches: [master]
+    branches:
+      - master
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -32,38 +24,43 @@ jobs:
...
-    # ℹ️ Command-line programs to run using the OS shell.
-    # 📚 https://git.io/JvXDl
-
-    # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
-    #    and modify them (or add more) to build your code if your project
-    #    uses a compiled language
+    - name: Build Go
+      if: matrix.language == 'go'
+      run: |
+        go build -v ./cmd/aro
```
bennerv, line 63: now that we have runners, do we want to let autobuild do everything for us?

bennerv (dismissed):
lgtm - one question

.github/workflows/codeql-analysis.yml
cblecker: Will autobuild work for this? I haven't tested it, but can.

bennerv: yeah that's what we previously used.  

https://github.com/mbarnes/ARO-RP/blob/4094ce29905a13c399413cb48cc98a9ee4440f74/.github/workflows/codeql-analysis.yml#L35-L57

cblecker: Removed the specific build step. All I'm doing now that's go specific is ensuring we have the right go version, and installing the Linux headers we need

bennerv (approved):

PR #3110, opened 2023-08-18
multiple ARO managed public IPs per Load balancer during cluster create

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-3102

This PR adds ability to specify multiple outbound IPs during a cluster creation.  Additional OutboundIPs are added to the "storage" arm template.  The API Server IP is reused when the API Server is public and using Managed Outbound IPs.

### Test plan for issue:
Manually created cluster with public and private API Server and validated load balancer frontend IP configuration is correct.

Yes, to be done by doc writer.

### DIFF
--- a/pkg/cluster/deploybaseresources_additional.go
+++ b/pkg/cluster/deploybaseresources_additional.go
@@ -556,9 +556,15 @@ func (m *manager) networkPublicLoadBalancer(azureRegion string, outboundIPs []ap
 		APIVersion: azureclient.APIVersion("Microsoft.Network"),
 		DependsOn:  []string{},
 	}
+
+	if m.doc.OpenShiftCluster.Properties.NetworkProfile.LoadBalancerProfile.ManagedOutboundIPs == nil && m.doc.OpenShiftCluster.Properties.APIServerProfile.Visibility == api.VisibilityPublic {
+		armResource.DependsOn = append(armResource.DependsOn, "Microsoft.Network/publicIPAddresses/"+m.doc.OpenShiftCluster.Properties.InfraID+"-pip-v4")
+	}
+
 	for _, ip := range outboundIPs {
 		ipName := stringutils.LastTokenByte(ip.ID, '/')
 		armResource.DependsOn = append(armResource.DependsOn, "Microsoft.Network/publicIPAddresses/"+ipName)
 	}
+
 	return armResource
 }
--- a/pkg/cluster/deploybaseresources_additional.go
+++ b/pkg/cluster/deploybaseresources_additional.go
@@ -16,6 +16,7 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/arm"
 	"github.com/Azure/ARO-RP/pkg/util/azureclient"
 	"github.com/Azure/ARO-RP/pkg/util/rbac"
+	"github.com/Azure/ARO-RP/pkg/util/stringutils"
 )
 
 func (m *manager) denyAssignment() *arm.Resource {
@@ -452,22 +453,13 @@ func (m *manager) networkInternalLoadBalancer(azureRegion string) *arm.Resource 
 	}
 }
 
-func (m *manager) networkPublicLoadBalancer(azureRegion string) *arm.Resource {
+func (m *manager) networkPublicLoadBalancer(azureRegion string, outboundIPs []api.ResourceReference) *arm.Resource {
 	lb := &mgmtnetwork.LoadBalancer{
 		Sku: &mgmtnetwork.LoadBalancerSku{
 			Name: mgmtnetwork.LoadBalancerSkuNameStandard,
 		},
 		LoadBalancerPropertiesFormat: &mgmtnetwork.LoadBalancerPropertiesFormat{
-			FrontendIPConfigurations: &[]mgmtnetwork.FrontendIPConfiguration{
-				{
-					FrontendIPConfigurationPropertiesFormat: &mgmtnetwork.FrontendIPConfigurationPropertiesFormat{
-						PublicIPAddress: &mgmtnetwork.PublicIPAddress{
-							ID: to.StringPtr("[resourceId('Microsoft.Network/publicIPAddresses', '" + m.doc.OpenShiftCluster.Properties.InfraID + "-pip-v4')]"),
-						},
-					},
-					Name: to.StringPtr("public-lb-ip-v4"),
-				},
-			},
+			FrontendIPConfigurations: &[]mgmtnetwork.FrontendIPConfiguration{},
 			BackendAddressPools: &[]mgmtnetwork.BackendAddressPool{
 				{
 					Name: to.StringPtr(m.doc.OpenShiftCluster.Properties.InfraID),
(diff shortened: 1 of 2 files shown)


tony-schndr:
PR is branched from https://github.com/Azure/ARO-RP/pull/3095 so only the changes from "[add managed ips to arm templates](https://github.com/Azure/ARO-RP/pull/3110/commits/794c93e0f75bebbab3de47388f343cc4fe55efdf)" need to be reviewed.

pkg/cluster/deploybaseresources_additional.go
```diff
@@ -499,6 +487,15 @@ func (m *manager) networkPublicLoadBalancer(azureRegion string) *arm.Resource {
 	}
 
 	if m.doc.OpenShiftCluster.Properties.APIServerProfile.Visibility == api.VisibilityPublic {
```
SudoBrendan, line 489: bug? We are now only conditionally adding `public-lb-ip-v4` FIP config and outbound rule on Public clusters. Before, it appears we did this for public and private. Is this intentional?

pkg/cluster/deploybaseresources_additional.go
```diff
@@ -532,11 +529,36 @@ func (m *manager) networkPublicLoadBalancer(azureRegion string) *arm.Resource {
...
+		for i := len(*lb.FrontendIPConfigurations); i < m.doc.OpenShiftCluster.Properties.NetworkProfile.LoadBalancerProfile.ManagedOutboundIPs.Count; i++ {
+			resourceGroupID := m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID
+			frontendIPConfigName := stringutils.LastTokenByte(outboundIPs[i].ID, '/')
+			frontendConfigID := fmt.Sprintf("%s/providers/Microsoft.Network/loadBalancers/%s/frontendIPConfigurations/%s", resourceGroupID, *lb.Name, frontendIPConfigName)
+			*lb.FrontendIPConfigurations = append(*lb.FrontendIPConfigurations, newFrontendIPConfig(frontendIPConfigName, frontendConfigID, outboundIPs[i].ID))
+		}
+
+		for i := 0; i < m.doc.OpenShiftCluster.Properties.NetworkProfile.LoadBalancerProfile.ManagedOutboundIPs.Count; i++ {
+			resourceGroupID := m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID
+			if i == 0 && m.doc.OpenShiftCluster.Properties.APIServerProfile.Visibility == api.VisibilityPublic {
```
SudoBrendan, line 542: re `i == 0`, is there another way we can check if this is the `public-lb-ip-v4` rule? this seems prone to flakiness/buginess in the future since currently array order matters.

SudoBrendan (commented):
Great initial work on this! I left a few questions/nits/potential bugs that need confirmation. Thank you

pkg/cluster/deploybaseresources_additional.go
tony-schndr: This is intentional.  My thought behind this is when customer provided outbound IPs/Prefixes are being used this IP would be deleted if it does not have a load balancing rule associated with it.

pkg/cluster/deploybaseresources_additional.go
tony-schndr: There should be no bug/flake.  If this IP is present it will always be at index 0 as this IP is added to the `outboundIPs` slice first, it is only present when the api server is public like you called out in your other comment.  This if check is only to get the correct Frontend IP Config name when the api server is public since it is named `public-lb-ip-v4`, when the api server is private the frontend IP Config name does not matter so it will use the name of the public IP Address.
https://github.com/Azure/ARO-RP/blob/613b5f9fdd42640dc0b31fe3595a3ba446b203b8/pkg/cluster/deploybaseresources.go#L149-L164

SudoBrendan:
Hey, I know this was originally proposed before ARM was fully released. Now that it is, can we get some E2Es added? Do we need https://github.com/Azure/ARO-RP/pull/3183 to merge first for those to be included here?

cadenmarchese (approved):
LGTM

tony-schndr:
>Hey, I know this was originally proposed before ARM was fully released. Now that it is, can we get some E2Es added? Do we need https://github.com/Azure/ARO-RP/pull/3183 to merge first for those to be included here?

@SudoBrendan Yes we would, then a rebase, but I am not sure it makes a difference here.  All cluster creation code I changed/added here was executed during the e2e test run of this PR.  In addition I tested cluster creation with API Server = private with localrp and had a successful create there as well.

SudoBrendan (approved):

cadenmarchese (dismissed):

jhoreman (dismissed):
Looks good, thank you!

pkg/cluster/deploybaseresources_additional.go
```diff
@@ -532,11 +529,36 @@ func (m *manager) networkPublicLoadBalancer(azureRegion string) *arm.Resource {
...
 		Resource:   lb,
 		APIVersion: azureclient.APIVersion("Microsoft.Network"),
-		DependsOn: []string{
-			"Microsoft.Network/publicIPAddresses/" + m.doc.OpenShiftCluster.Properties.InfraID + "-pip-v4",
-		},
+		DependsOn:  []string{},
+	}
+	for _, ip := range outboundIPs {
+		ipName := stringutils.LastTokenByte(ip.ID, '/')
+		armResource.DependsOn = append(armResource.DependsOn, "Microsoft.Network/publicIPAddresses/"+ipName)
```
kimorris27, line 561: I think there's a bug here. If API visibility is public and `ManagedOutboundIPs` is nil, `outboundIPs` will also be `nil`, so the LB won't be set up to depend on the `-pip-v4` PIP, which it still needs. Does this need to be reworked to ensure that the LB depends on the `-pip-v4` PIP whenever API visibility is public?

tony-schndr: Good catch, when adding outboundIPs/outboundIPPrefixes to this feature there will be a bug if that IP is not added.  I'll update this to prevent that from happening.

SudoBrendan (approved):

hlipsig:
Code changes look fine. I'm not seeing any unit tests or e2e tests, and that concerns me. Do we have those in a different PR?

petrkotas (approved):
This makes sense, I would like to see this covered by tests, would it be unit or e2e. 
For sake of not blocking the release, @tony-schndr would you please add test as a followup?

tony-schndr:
@petrkotas @hlipsig I will add unit tests in a follow up.  This install is partially validated with passing e2e tests.  I also tested multiple installations, here is the test scenario breakdown:

Public ingress was used for all tests

Installed cluster with private API Server and 5 additional managed IPs, install succeeded and all COs healthy.

Installed cluster with public API Server and 5 managed IPs, install succeeded and validated COs healthy in cluster console..  
- API Server IP is 1 of the 5 managed IPs.

Installed cluster with private API Server and 1 Managed IP, install succeeded and validated COs healthy.

Installed cluster with public API Server and 1 Managed IP, install succeeded and validated COs healthy.  This scenario is also the default install for e2e tests.

PR #3112, opened 2023-08-21
Enable preconfigured NSG flag on azure-cli

gouthamMN:

Fixes: https://issues.redhat.com/browse/ARO-2591

Bumps api version used and adds parameter --enable-preconfigured-nsg to az aro cli extension for using preconfigured NSGs. Allowed values for the flag are true or false and by default, preconfigured NSG will be disabled. This should probably not be merged until the next stable API is released.

Yes, the [Azure Doc](https://learn.microsoft.com/en-us/azure/openshift) will need to be updated though that will be a separate effort.

### DIFF
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -68,6 +68,8 @@ def load_arguments(self, _):
         c.argument('outbound_type',
                    help='Outbound type of cluster. Must be "Loadbalancer" (default) or "UserDefinedRouting".',
                    validator=validate_outbound_type)
+        c.argument('enable_preconfigured_nsg',
+                   help='Use Preconfigured NSGs. Allowed values: false, true.')
         c.argument('disk_encryption_set',
                    help='ResourceID of the DiskEncryptionSet to be used for master and worker VMs.',
                    validator=validate_disk_encryption_set)
--- a/python/az/aro/azext_aro/custom.py
+++ b/python/az/aro/azext_aro/custom.py
@@ -7,7 +7,7 @@ import os
 from base64 import b64decode
 import textwrap
 
-import azext_aro.vendored_sdks.azure.mgmt.redhatopenshift.v2023_04_01.models as openshiftcluster
+import azext_aro.vendored_sdks.azure.mgmt.redhatopenshift.v2023_09_04.models as openshiftcluster
 
 from azure.cli.command_modules.role import GraphError
 from azure.cli.core.commands.client_factory import get_mgmt_service_client
@@ -44,6 +44,7 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
                worker_subnet,
                vnet=None,  # pylint: disable=unused-argument
                vnet_resource_group_name=None,  # pylint: disable=unused-argument
+               enable_preconfigured_nsg=None,
                location=None,
                pull_secret=None,
                domain=None,
@@ -141,6 +142,7 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
             pod_cidr=pod_cidr or '10.128.0.0/14',
             service_cidr=service_cidr or '172.30.0.0/16',
             outbound_type=outbound_type or '',
+            preconfigured_nsg='Enabled' if enable_preconfigured_nsg else 'Disabled',
         ),
         master_profile=openshiftcluster.MasterProfile(
             vm_size=master_vm_size or 'Standard_D8s_v3',
--- a/python/az/aro/azext_aro/_client_factory.py
+++ b/python/az/aro/azext_aro/_client_factory.py
@@ -4,7 +4,7 @@
 import urllib3
 
 from azext_aro.custom import rp_mode_development
-from azext_aro.vendored_sdks.azure.mgmt.redhatopenshift.v2023_04_01 import AzureRedHatOpenShiftClient
+from azext_aro.vendored_sdks.azure.mgmt.redhatopenshift.v2023_09_04 import AzureRedHatOpenShiftClient
 from azure.cli.core.commands.client_factory import get_mgmt_service_client
 
 
--- a/python/az/aro/azext_aro/commands.py
+++ b/python/az/aro/azext_aro/commands.py
@@ -11,7 +11,7 @@ from azext_aro._help import helps  # pylint: disable=unused-import
 
 def load_command_table(self, _):
     aro_sdk = CliCommandType(
-        operations_tmpl='azext_aro.vendored_sdks.azure.mgmt.redhatopenshift.v2023_04_01.operations#OpenShiftClustersOperations.{}',  # pylint: disable=line-too-long
+        operations_tmpl='azext_aro.vendored_sdks.azure.mgmt.redhatopenshift.v2023_09_04.operations#OpenShiftClustersOperations.{}',  # pylint: disable=line-too-long
         client_factory=cf_aro)
 
     with self.command_group('aro', aro_sdk, client_factory=cf_aro) as g:


python/az/aro/azext_aro/custom.py
```diff
@@ -73,6 +74,14 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
...
             raise UnauthorizedError('Microsoft.RedHatOpenShift provider is not registered.',
                                     'Run `az provider register -n Microsoft.RedHatOpenShift --wait`.')
+        if enable_preconfigured_nsg:
+            feature_client = get_mgmt_service_client(
+                cmd.cli_ctx, ResourceType.MGMT_RESOURCE_FEATURES)
+            feature = feature_client.features.get('Microsoft.RedHatOpenShift', 'PreconfiguredNSG')
+            if feature.properties.state != 'Registered':
+                raise ValidationError('PreconfiguredNSG feature is not registered. \
+                                      Run `az feature register --namespace Microsoft.RedHatOpenShift \
+                                      -n PreconfiguredNSG`.')
```
weinong, line 84: as it is for GA, feature flag shouldn't be needed.

python/az/aro/azext_aro/custom.py
```diff
@@ -458,6 +474,19 @@ def get_network_resources_from_subnets(cli_ctx, subnets, fail):
...
+            if subnet.get("networkSecurityGroup", None):
+                subnet_resources.add(subnet['networkSecurityGroup']['id'])
+                subnet_with_nsg_preattached = subnet_with_nsg_preattached + 1
+
+    # when preconfiguredNSG feature is Enabled we either have all subnets NSG attached or none.
+    if oc.network_profile.preconfigured_nsg == 'Enabled' and \
+        subnet_with_nsg_preattached != 0 and \
+            subnet_with_nsg_preattached != len(subnets):
+        raise ValidationError("""(ValidationError) preconfiguredNSG feature is enabled but NSG not attached for subnets.
+                                  Please makesure all the subnets have network security group attached and retry.
```
weinong, line 487:
```suggestion
                                  Please make sure all the subnets have network security group attached and retry.
```

python/az/aro/azext_aro/custom.py
```diff
@@ -458,6 +474,19 @@ def get_network_resources_from_subnets(cli_ctx, subnets, fail):
...
+                                  If issue persists: raise azure support ticket""")
```
weinong, line 488: this is user error, i don't think we should encourage them to open support request?

weinong: i think a better error message is to tell them which subnet is missing nsg configured?

python/az/aro/azext_aro/_params.py
```diff
@@ -68,6 +68,8 @@ def load_arguments(self, _):
         c.argument('outbound_type',
                    help='Outbound type of cluster. Must be "Loadbalancer" (default) or "UserDefinedRouting".',
                    validator=validate_outbound_type)
+        c.argument('enable_preconfigured_nsg',
```
s-amann, line 71: can you make this `arg_type=get_three_state_flag()`? this will align with other true/false/none flags we have.

petrkotas (approved):
Hi @gouthamMN good work. LGTM. I think we can merge as 1) the API should be already released, 2) this will affect only extension which we use for development. The same work will need to be backported to upstream: https://github.com/Azure/azure-cli to make it available to the public.

s-amann (approved):

gouthamMN:
> Hi @gouthamMN good work. LGTM. I think we can merge as 1) the API should be already released, 2) this will affect only extension which we use for development. The same work will need to be backported to upstream: https://github.com/Azure/azure-cli to make it available to the public.

ARM manifest rollout and upstreaming azure-rest-api-specs swagger is still in progress.

SudoBrendan:
> ARM manifest rollout and upstreaming azure-rest-api-specs swagger is still in progress.

I'd agree with Petr. We should be safe to merge into this repo's `/python` directory, since ARO engineering should be the only ones using this code. We have historically merged here early, then upstreamed changes once REST Specs and ARM Manifest Rollout are complete. This process is super unclear, so I [drew up some diagrams](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki?anchor=introduction&pageId=240029&friendlyName=API-Release-Process) about it recently.

python/az/aro/azext_aro/custom.py
```diff
@@ -145,6 +147,7 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
             pod_cidr=pod_cidr or '10.128.0.0/14',
             service_cidr=service_cidr or '172.30.0.0/16',
             outbound_type=outbound_type or '',
+            preconfigured_nsg='Enabled' if enable_preconfigured_nsg else 'Disabled',
```
SudoBrendan, line 150: this is good to merge since it follows suit, but I created a follow-up item for someone on our team to tackle: https://issues.redhat.com/browse/ARO-4250

python/az/aro/azext_aro/custom.py
```diff
@@ -202,14 +206,18 @@ def validate(cmd,  # pylint: disable=too-many-locals,too-many-statements
...
+                preconfigured_nsg='Enabled' if preconfigured_nsg else 'Disabled'
+            )
             self.master_profile = openshiftcluster.MasterProfile(
                 subnet_id=master_subnet_id,
                 disk_encryption_set_id=disk_encryption_id
             )
             self.worker_profiles = [openshiftcluster.WorkerProfile(
                 subnet_id=worker_subnet_id
             )]
+            self.worker_profiles_status = None
```
SudoBrendan, line 220: q: I'm sure this is fine, but it's unintuitive at a glace and now I'm curious: why was this needed?

python/az/aro/azext_aro/custom.py
```diff
@@ -455,6 +464,20 @@ def get_network_resources_from_subnets(cli_ctx, subnets, fail):
...
+            if subnet.get("networkSecurityGroup", None):
+                subnet_resources.add(subnet['networkSecurityGroup']['id'])
+            else:
+                subnets_with_no_nsg_attached.add(sn)
+
+    # when preconfiguredNSG feature is Enabled we either have all subnets NSG attached or none.
+    if oc.network_profile.preconfigured_nsg == 'Enabled' and \
+        len(subnets_with_no_nsg_attached) != 0 and \
+            len(subnets_with_no_nsg_attached) != len(subnets):
+        raise ValidationError(f"(ValidationError) preconfiguredNSG feature is enabled but NSG not attached for subnets.\
```
SudoBrendan, line 477: we should probably fix up the grammar a bit on this:

```
(ValidationError) The preconfiguredNSG feature is enabled but an NSG is not attached for all required subnets.
```

python/az/aro/azext_aro/custom.py
```diff
@@ -455,6 +464,20 @@ def get_network_resources_from_subnets(cli_ctx, subnets, fail):
...
+                                  Please make sure all the following subnets have network security group attached and retry.\
```
SudoBrendan, line 478: one more grammar issue, should be either:

1. `have a network` not `have network` (if only one NSG is to be used), or
2. `groups` not `group` (if multiple NSGs can be used)

SudoBrendan (commented):
This is nearly there, I left a few grammar nits and one question. I think it would be good to check with the BU on all the error messaging before we merge.

python/az/aro/azext_aro/custom.py
gouthamMN: Not Setting/defining the field worker_profiles_status here will result in exception when this function makes a call to get_cluster_network_resources() since it loops through the worker_profiles_status. During az create we don't pass any values to this filed as it's readOnly.

gouthamMN:
> > ARM manifest rollout and upstreaming azure-rest-api-specs swagger is still in progress.
> 
> I'd agree with Petr. We should be safe to merge into this repo's `/python` directory, since ARO engineering should be the only ones using this code. We have historically merged here early (because we can use localhost RP instances to test), then upstreamed changes once REST Specs and ARM Manifest Rollout are complete. This process is super unintuitive, so I [drew up some diagrams](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki?anchor=introduction&pageId=240029&friendlyName=API-Release-Process) about it recently.

@SudoBrendan I'm fine merging the changes. Thank you for updating that wiki.

SudoBrendan (approved):

python/az/aro/azext_aro/custom.py
s-amann: IMO the RP code should be able to handle the case of worker_profiles_status being nil?

s-amann (approved):

PR #3113, opened 2023-08-22
Fix credential rotation failure

dem4gus:

Fixes [ARO-2898](https://issues.redhat.com/browse/ARO-2898)

Recreates the `azure-credentials` secret if it has been deleted.

The implementation uses [server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) to accomplish this task. Since SSA as a concept is new to the code base I am going to annotate the PR a bit more than normal.

### Test plan for issue:

Unit tests in `clusterserviceprincipal_test.go` updated for new behavior.

Previous code failed because of missing `Force` parameter in apply action (see annotations).

### DIFF
--- a/pkg/cluster/clusterserviceprincipal.go
+++ b/pkg/cluster/clusterserviceprincipal.go
@@ -185,27 +185,7 @@ func (m *manager) updateOpenShiftSecret(ctx context.Context) error {
 		// azure_tenant_id: tenant_id
 		secret, err := m.kubernetescli.CoreV1().Secrets(clusterauthorizer.AzureCredentialSecretNameSpace).Get(ctx, clusterauthorizer.AzureCredentialSecretName, metav1.GetOptions{})
 		if kerrors.IsNotFound(err) {
-			// rebuild secret
-			resourceGroupID := m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID
-			secret = &corev1.Secret{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      clusterauthorizer.AzureCredentialSecretName,
-					Namespace: clusterauthorizer.AzureCredentialSecretNameSpace,
-				},
-				Type: corev1.SecretTypeOpaque,
-				Data: map[string][]byte{
-
-					"azure_subscription_id": []byte(m.subscriptionDoc.ID),
-					"azure_resource_prefix": []byte(m.doc.OpenShiftCluster.Properties.InfraID),
-					"azure_resourcegroup":   []byte(resourceGroupID[strings.LastIndex(resourceGroupID, "/")+1:]),
-					"azure_region":          []byte(m.doc.OpenShiftCluster.Location),
-					// values set below
-					"azure_client_id":     []byte(""),
-					"azure_client_secret": []byte(""),
-					"azure_tenant_id":     []byte(""),
-				},
-			}
-			recreate = true
+			secret = m.newAzureCredentialSecret()
 		} else if err != nil {
 			return err
 		}
@@ -258,4 +238,26 @@ func (m *manager) updateOpenShiftSecret(ctx context.Context) error {
 		m.log.Error(err)
 	}
 	return nil
+}
+
+func (m *manager) newAzureCredentialSecret() *corev1.Secret {
+
+	resourceGroupID := m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      clusterauthorizer.AzureCredentialSecretName,
+			Namespace: clusterauthorizer.AzureCredentialSecretNameSpace,
+		},
+		Type: corev1.SecretTypeOpaque,
+		Data: map[string][]byte{
+
+			"azure_subscription_id": []byte(m.subscriptionDoc.ID),
+			"azure_resource_prefix": []byte(m.doc.OpenShiftCluster.Properties.InfraID),
+			"azure_resourcegroup":   []byte(resourceGroupID[strings.LastIndex(resourceGroupID, "/")+1:]),
+			"azure_region":          []byte(m.doc.OpenShiftCluster.Location),
+			"azure_client_id":       []byte(""),
+			"azure_client_secret":   []byte(""),
+			"azure_tenant_id":       []byte(""),
+		},
+	}
 }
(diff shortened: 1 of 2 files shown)


pkg/cluster/clusterserviceprincipal.go
```diff
@@ -173,50 +177,50 @@ func (m *manager) updateAROSecret(ctx context.Context) error {
...
+			},
+			Data: map[string][]byte{},
+		}
+	} else if err != nil {
 		return err
 	}
 
+	resourceGroupID := m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID
+	spp := m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile
+	desiredData := map[string][]byte{
```
dem4gus, line 199: Server-side apply allows for a more declarative approach to object management. Instead of getting the existing object and mutating the data on it, we can just define what we want the data to look like. Most of the time we could just apply this data without ever fetching the original object, but since this data requires a restart of the `cloud-credential-operator` pods there's a bit extra work to do.

pkg/cluster/clusterserviceprincipal.go
```diff
@@ -173,50 +177,50 @@ func (m *manager) updateAROSecret(ctx context.Context) error {
...
+	var changed bool
+	for key, val := range desiredData {
+		changed = bytes.Compare(val, secret.Data[key]) != 0 || changed
+	}
 	if !changed {
 		return nil
 	}
 
+	secretApplyConfig := applyv1.Secret(secret.Name, secret.Namespace).WithData(desiredData)
+	_, err = m.kubernetescli.CoreV1().Secrets(clusterauthorizer.AzureCredentialSecretNameSpace).Apply(ctx, secretApplyConfig, metav1.ApplyOptions{FieldManager: "aro-rp", Force: true})
```
dem4gus, line 219: There are two new fields in `metav1.ApplyOptions{}`. `FieldManager` is metadata about the entity to last change the affected fields, in this case the RP. The `Force` parameter is required for use in controllers ([ref]()https://kubernetes.io/blog/2022/10/20/advanced-server-side-apply/#controller-authors).

pkg/cluster/clusterserviceprincipal.go
```diff
@@ -173,50 +177,50 @@ func (m *manager) updateAROSecret(ctx context.Context) error {
...
-			}
-		}
-		return nil
-	})
-	if err != nil {
+	//data:
+	// azure_client_id: secret_id
+	// azure_client_secret: secret_value
+	// azure_tenant_id: tenant_id
+	secret, err := m.kubernetescli.CoreV1().Secrets(clusterauthorizer.AzureCredentialSecretNameSpace).Get(ctx, clusterauthorizer.AzureCredentialSecretName, metav1.GetOptions{})
```
dem4gus, line 184: The only reason to `Get` the secret with server-side apply is because of the pod restarts required when the service principal changes. We could also further simplify the logic here by not even caring about whether the credentials are changing. We would define the data we want on the secret and just apply it and restart the `cloud-credential-operator` pods every time this function is run. This code only runs when the cluster admin runs `az aro update`, so restarting the pods shouldn't make much of a difference.

pkg/cluster/clusterserviceprincipal_test.go
```diff
@@ -297,9 +301,13 @@ func getFakeOpenShiftSecret() corev1.Secret {
 	name := "azure-credentials"
 	namespace := "kube-system"
 	data := map[string][]byte{
```
dem4gus, line 303: `updateOpenShiftSecret` now sets these fields, so they need to be present on the fake object.

pkg/cluster/clusterserviceprincipal_test.go
```diff
@@ -438,3 +458,44 @@ func TestUpdateOpenShiftSecret(t *testing.T) {
 		})
 	}
 }
+
```
dem4gus, line 461: See linked issue kubernetes/client-go#1184 for more information on why this is required. Potential upstream contribution if someone wants to tackle it.

pkg/cluster/clusterserviceprincipal.go
cadenmarchese: If we plan to re-use server side apply with the force flag elsewhere in the codebase, we may want to declare `FieldManager` somewhere else rather than hardcoding it so that we can re-use this value.

I am imagining a scenario where we use server side apply with this resource somewhere else in the codebase, but `FieldManager: "aro-rp"` is set differently (maybe `FieldManager: "rp"`) and we get a conflict.

It doesn't seem to matter as long as we use force: true, but maybe we could use some global solution.

dem4gus: > I am imagining a scenario where we use server side apply with this resource somewhere else in the codebase, but FieldManager: "aro-rp" is set differently (maybe FieldManager: "rp") and we get a conflict.

That's exactly the reason you want to use the force flag in a controller. From what I understood from the documents, the FieldManager exists to tell user-run apply operations that the field was set by a different entity and the user could then decide how to handle that information, but when using it in some sort of controller (which this essentially is, just one that's triggered manually) you want to just apply your information. We could even set different FieldManagers depending on the RP operation that's setting the field, which may be more useful for traceability. This one could be something like `"az-aro-update"`.

pkg/cluster/clusterserviceprincipal.go
dem4gus: I removed it because previously it was building both the `Secret` object and defining the data, which was a bit much to have inline. With this we're defining the data regardless and only creating a bare secret to operate on. If we just `Apply` every time we don't even need to create the secret.

pkg/cluster/clusterserviceprincipal.go
```diff
@@ -173,50 +177,50 @@ func (m *manager) updateAROSecret(ctx context.Context) error {
 }
 
 func (m *manager) updateOpenShiftSecret(ctx context.Context) error {
-	var changed bool
-	spp := m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile
-	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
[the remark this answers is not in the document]
dem4gus: We'll know it fails because we're returing any error after the `Apply` on line 195. It's less "passing the buck" to the server and more using a feature that's explicity intended to simplify controller logic and make them more declarative ((ref)[https://kubernetes.io/docs/reference/using-api/server-side-apply/#comparison-with-client-side-apply]). Part of the simplification of logic is that we no longer have to worry about retrying on conflict, because the server is just going to force our changes on a conflict. It's the same end result.

s-amann (approved):

PR #3115, opened 2023-08-22
Update documentation for testing the ARO operator

lranjbar:

N/A

This PR updates the documentation around the operator and Fedora 37. I found some issues with the docs while creating a new controller for ARO-3112.

### Test plan for issue:

N/A

No. This is to update the developer docs. 😄

### DIFF
--- a/docs/operators.md
+++ b/docs/operators.md
@@ -66,13 +66,24 @@ Make sure KUBECONFIG is set:
 make admin.kubeconfig
 export KUBECONFIG=$(pwd)/admin.kubeconfig
 ```
+
+Alternatively you can set the KUBECONFIG by logging into your test cluster using `oc`:
+
+```sh
+oc login ${cluster-api-url} -u kubeadmin -p ${kubeadmin-password}
+```
+
 If you are using a private cluster, you need to connect to the respective VPN of your region. For example for eastus:
 ```sh
 sudo openvpn --config secrets/vpn-eastus.ovpn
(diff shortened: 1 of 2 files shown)


SudoBrendan (approved):

jhoreman (approved):
Small comment but won't hold approval over it:

`/bin/sh` is often symlinked to another shell depending on the flavour of your distro. As long as we don't truly expect `bash` style implementation (and `dash` works just fine, for example) then this looks fine to me!

docs/operators.md
```diff
@@ -66,13 +66,24 @@ Make sure KUBECONFIG is set:
...
-Then do:
+Scale the operator:
 ```sh
 oc scale -n openshift-azure-operator deployment/aro-operator-master --replicas=0
+```
+
+Build the operator binary and run it locally (as if it was running a master node)
+```sh
 make generate
 go run ./cmd/aro operator master
```
cadenmarchese, line 88: One more thing I think we can add here that's required for MacOS / M1 users (it isn't required on Linux, but still works just as well):
```suggestion
go run -tags containers_image_openpgp ./cmd/aro operator master
```
We also build ARO with this tag, but it looks like we forgot to update the docs here.

petrkotas (approved):
Thanks @lranjbar !

petrkotas:
Merging the docs update, ignoring e2e because this PR does not modify any code.

PR #3117, opened 2023-08-24
Rest Spec API upstream ARO-RP updates

jhoreman:

https://issues.redhat.com/browse/ARO-4029

Rest Spec API upstream ARO-RP updates

Fixed: 

Integers set to int32 format. (this is a must-fix, but also a "breaking change")
Added "OutboundType" to x-ms-enums for preview API (linting)
Added description on "OutboundType" (PR nit)
Fixed readme to include 2023-07-01-preview (must-fix)
Updated description of resource reference (PR nit)
Added "Cancelled" status to "ProvisioningState" (linting)

This should make future api upstream efforts easier, as this should bring the code base closer to being able to pass upstream linting.

### Test plan for issue:

<!--

These changes were recently merged into the azure-rest-spec-api upstream branch here:

https://github.com/Azure/azure-rest-api-specs/pull/25182

Yes, wiki docs are being worked on to reflect the process changes for upstreaming


tsatam (dismissed):

pkg/swagger/typewalker.go
```diff
@@ -92,6 +92,7 @@ func (tw *typeWalker) schemaFromType(t types.Type, deps map[*types.Named]struct{
 			s.Type = "boolean"
 		case types.Int:
 			s.Type = "integer"
+			s.Format = "int32"
```
cblecker, line 95: Is this a breaking change? This seems to impact the stable API too, not just the preview one

jhoreman: This was indeed flagged as a breaking change - but the requirement of the int32 format on all integers was listed as a "must-fix" on the CI during the upstream process. As such this must-fix introduces the "breaking change" but we can't get approval to upstream without fixing the must fix. One thing to also consider is that we also only define bools, ints and strings once on our side in the generator code - for all iterations of the aforementioned. Definitely open to suggestions on this one.

pkg/api/v20230701preview/openshiftcluster.go
```diff
@@ -78,6 +78,7 @@ type ProvisioningState string
 const (
 	ProvisioningStateCreating      ProvisioningState = "Creating"
 	ProvisioningStateUpdating      ProvisioningState = "Updating"
+	ProvisioningStateCancelled     ProvisioningState = "Cancelled"
```
bennerv, line 81: What's the criteria for "Cancelled"?  We should add it to our top-level openshiftcluster doc.  We should also add logic or a TODO comment stating we do not use it yet.

jhoreman: ARM defines only 3 terminal states. Succeed, Failed, Cancelled. For linting purposes upstream requires at least these three to be present. More info here: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations

We can certainly add a TODO comment - but I don't think we ever intend to use this status in our implementation, but it should be included so that upstreaming doesn't nit or have its linting fail. Let me know if you think we should add it still.

bennerv: Pull it into the top-level and throw a todo around it please :). 

Just in case we ever end up using it, our converters won't go nuts.

pkg/api/v20230701preview/openshiftcluster.go
```diff
@@ -135,7 +138,7 @@ const (
 
 // ResourceReference represents a reference to an Azure resource.
 type ResourceReference struct {
-	// The fully qualified Azure resource id.
+	// The fully qualified Azure of an IP address resource.
```
SudoBrendan, line 141: Seems like a small typo compared to what we upstreamed: https://github.com/Azure/azure-rest-api-specs/pull/25182/files#diff-9e37deed88559aa01045e43a7eb621ea865edb3a79cd0e90b8e947ba2879410cR1854

SudoBrendan (commented):
One typo, otherwise LGTM

SudoBrendan (dismissed):

pkg/api/v20230701preview/openshiftcluster.go
```diff
@@ -126,6 +128,7 @@ type ServicePrincipalProfile struct {
 	ClientSecret string `json:"clientSecret,omitempty" mutable:"true"`
 }
 
+// The outbound routing strategy used to provide your cluster egress to the internet.
```
bennerv, line 131: Should be set in `pkg/api/openshiftcluster.go` as well.

pkg/api/v20230701preview/openshiftcluster.go
```diff
@@ -135,7 +138,7 @@ const (
 
 // ResourceReference represents a reference to an Azure resource.
 type ResourceReference struct {
-	// The fully qualified Azure resource id.
+	// The fully qualified Azure resource id of an IP address resource.
```
bennerv, line 141: Same -> set in `pkg/appi/openshiftcluster.go`

bennerv (changes requested):
two small changes otherwise lgtm

pkg/api/v20230701preview/openshiftcluster.go
jhoreman: In /pkg/api/openshiftcluster.go we have the following:

// OutboundType represents the type of routing a cluster is using
type OutboundType string

We can change it but technically this is new in the 07 01 preview API.

bennerv (approved):

PR #3118, opened 2023-08-25
Use Kiota to generate the Graph SDK, instead of using msgraphsdk

hawkowl:
Fixes https://issues.redhat.com/browse/ARO-4054

This removes the msgraphsdk dependency with built versions based on a cut-down OpenAPI specification. This allows us to remove a lot of code and cuts CI time in half and a good 10 minutes off E2E (from building).


pkg/util/graph/graphsdk/applications/application_item_request_builder.go
```diff
@@ -0,0 +1,186 @@
+package applications
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+
+	i2ae4187f7daee263371cb1c977df639813ab50ffa529013b7437480d1ec0158f "github.com/microsoft/kiota-abstractions-go"
```
lranjbar, line 9: Is there any way to make this generated API more readable?

lranjbar (dismissed):
Due to the amount of changes in this PR - I request this be broken into two PRs. One with the generated graph and vendored changes and one with the changes to use the new graph SDK.

Beyond that putting the changes for vendor in to a seperate commit will allow for review in Github much easier.

lranjbar:
I'd also like to see a knowledge share around Kiota, maybe in the architecture review forum? This is a big change with a neat new tool. 😃

hawkowl:
> I'd also like to see a knowledge share around Kiota, maybe in the architecture review forum? This is a big change with a neat new tool. 😃 

Kiota is the same tool used to generate the official SDK, it just uses the whole openapi specification instead of the cut down one I've put in. Functionally, the only changes are imports, initialisers from the SDK, and some error handling function name changes to match newer Kiota versions (since I used the latest stable).

pkg/util/graph/graphsdk/applications/application_item_request_builder.go
hawkowl: not really 🥲

hawkowl:
> Due to the amount of changes in this PR - I request this be broken into two PRs. One with the generated graph and vendored changes and one with the changes to use the new graph SDK.
> 
> Beyond that putting the changes for vendor in to a seperate commit will allow for review in Github much easier. 

@lranjbar I've already got each of the changes in separate commits for review -- it should be possible to step through them individually and skip the go mod vendor ones.

We also can't break master, so it would have to be one of adding the generated but unused code, and then another switching to the new code and deleting the old stuff -- the moment we stop using it, go mod vendor will trim the import tree and cause all of the vendor changes.

lranjbar:
> > Due to the amount of changes in this PR - I request this be broken into two PRs. One with the generated graph and vendored changes and one with the changes to use the new graph SDK.
> > Beyond that putting the changes for vendor in to a seperate commit will allow for review in Github much easier.
> 
> @lranjbar I've already got each of the changes in separate commits for review -- it should be possible to step through them individually and skip the go mod vendor ones.
> 
> We also can't break master, so it would have to be one of adding the generated but unused code, and then another switching to the new code and deleting the old stuff -- the moment we stop using it, go mod vendor will trim the import tree and cause all of the vendor changes.

Yeah I'm aware it would create a chain of events like that. This change seems to be large enough where it warrants multiple PRs. I did walk through the commits yesterday, but it is not really easy to review in this state. I'll let others decide their own opinion on that but as it stands it's just too much change in one PR IMO.

hawkowl:
@lranjbar I've rebased the PR to make some of the commits clearer, apart from the generation and vendoring it's ~20LoC of non-copied or non-generated golang code change, which should be a bit clearer.

mbarnes (approved):
LGTM, I appreciate the rebase.  Looks like you've addressed only my concern from the initial set of commits, which was how to update the generated code in the future.  But if I'm not mistaken it appears all the necessary steps are under `make generate-kiota` now.

Being the one that added **msgraph-sdk-go** originally, I'm happy to see us ditch that monstrosity for this much slimmer SDK.

Makefile
```diff
@@ -93,6 +93,14 @@ generate:
 generate-guardrails:
 	cd pkg/operator/controllers/guardrails/policies && ./scripts/generate.sh > /dev/null
 
+generate-kiota:
```
SudoBrendan, line 96: idea: can we add this as a dependent target of `generate` it feels like a good fit there, and would mean it gets tied into all our CI/etc.

pkg/util/graph/graphsdk/kiota-lock.json
```diff
@@ -0,0 +1,31 @@
+{
```
SudoBrendan, line 1: I think you had mentioned in our 1:1 yesterday "oh shoot, this should be .gitignored!" - so leaving you a todo and/or comment to immediately resolve because I'm way off base? :)

pkg/util/graph/adapter.go
```diff
@@ -0,0 +1,36 @@
+package graph
```
SudoBrendan, line 1: I might be a little confused due to commit order, but would these files be required for the `make` target to work? If so it might be clearer to squash those two together as the first commit. I'm trying to figure out what is a root dependency, generated content, or other changes we need to use the generated SDK. This feels like a root dependency for generation since it's importing `kiota` stuff?

SudoBrendan (commented):
Thanks Amber, this looks really cool! I'm +1 on a demo and/or other documentation and/or links for breadcrumbs on future maintenance of this SDK. Otherwise, all I have are a few nits/questions. Thank you for the PR!

pkg/util/graph/graphsdk/kiota-lock.json
hawkowl: The log file (which I fixed) should be gone now. This is a metadata file that's used for incremental updates, in case we need them.

Makefile
hawkowl: I don't think this is a good idea, as the Kiota generator is something that's probably a bit annoying to install and will need to actually be run every once in a blue moon, so I just added some docs for when it's needed instead.

pkg/util/graph/adapter.go
hawkowl: Not particularly -- none of this works until it's all done :). These files wrap the generated code for us. The kiota generator outputs go but is not go, so it doesn't matter if the codebase is in an inconsistent state between these commits.

pkg/util/graph/graphsdk/applications/application_item_request_builder.go
petrkotas: This is generated, right?

petrkotas (approved):
@hawkowl nice one. I must admit I am not an expert on this topic, but I do not see anything that would immediately not work. I have one question to confirm please.
Otherwise LGTM.

PR #3119, opened 2023-08-25
Revendor hive to 70b666ec8

2uasimojo:
Hive needs to be vendored at the same commit level as it is deployed in ARO. One reason, as described in the linked card, is that changes in APIs can lead to unintended edits during round-trip Get()/Update() flows.

### Which issue this PR addresses:

Partially Fixes ARO-3801

### What this PR does / why we need it:

Vendored hive code needs to be at the same commit level as what's deployed. This will avoid many issues; but the one in play here is [round tripping with Get() and Update()](https://groups.google.com/a/redhat.com/g/aos-devel/c/xs96y-oQdDA/m/tmEkFkPcAQAJ)

### Test plan for issue:

Build. Deploy. Try again.

### Is there any documentation that needs to be updated for this PR?
Is there documentation on the CI/CD process? It should be updated to note that ARO-RP needs to revendor hive to match whatever commit level is going to be deployed.

PR #3123, opened 2023-08-28
Remove the unneeded replace for hive

hawkowl:
Followup from https://github.com/Azure/ARO-RP/pull/3119

We don't actually need this anymore since we moved to controller-runtime

### Test plan for issue:

N/A


cblecker (approved):

2uasimojo:
Rockin. I was wondering why that import was needed, glad we could get rid of it.

PR #3125, opened 2023-08-29
Reduce diffs between our `az aro` extension and the upstream Azure CLI command

tsatam:

N/A

Small, nitpicky changes to our `az aro` extension code to better align it with upstream and reduce the diffs between the two

### Test plan for issue:

Ensure everything works as expected (no functionality changes should be present as a part of this)

AzureCLI extension and upstreaming documentation pages will need to reduce/change the documented diffs within them


tony-schndr (approved):
Thanks for this!  This will be really helpful when upstreaming changes.

Shivkumar13 (approved):
Good catch with unnecessary if block, overall looks good to me!

bennerv (approved):

PR #3129, opened 2023-08-29
Reduce flakes in TestLoadBalancerProfile()

lranjbar:

Fixes N/A

Fixes a discovered flake around the tests for load balancer profiles.

The test `managed ip cleanup errors are propagated when cleanup fails` deletes two load balancers using gomock. This action order is not guaranteed, so the error messages can be either order. (ie: uuid1 and then uuid2, or uuid2 and then uuid1)

### Test plan for issue:

Unit tests were updated to be less flaky.

No, test flake clean up.

### DIFF
--- a/pkg/cluster/loadbalancerprofile_test.go
+++ b/pkg/cluster/loadbalancerprofile_test.go
@@ -454,7 +454,7 @@ func TestReconcileLoadBalancerProfile(t *testing.T) {
 			loadBalancersClient *mock_network.MockLoadBalancersClient,
 			publicIPAddressClient *mock_network.MockPublicIPAddressesClient,
 			ctx context.Context)
-		expectedErr error
+		expectedErr []error
 	}{
 		{
 			name:  "default managed ips",
@@ -792,7 +792,7 @@ func TestReconcileLoadBalancerProfile(t *testing.T) {
 					},
 				},
 			},
-			expectedErr: fmt.Errorf("lb update failed"),
+			expectedErr: []error{fmt.Errorf("lb update failed")},
 		},
 		{
 			name:  "managed ip cleanup errors are propagated when cleanup fails",
@@ -866,7 +866,7 @@ func TestReconcileLoadBalancerProfile(t *testing.T) {
 					},
 				},
 			},
-			expectedErr: fmt.Errorf("failed to cleanup unused managed ips\ndeletion of unused managed ip uuid1-outbound-pip-v4 failed with error: error\ndeletion of unused managed ip uuid2-outbound-pip-v4 failed with error: error"),
+			expectedErr: []error{fmt.Errorf("failed to cleanup unused managed ips\ndeletion of unused managed ip uuid1-outbound-pip-v4 failed with error: error\ndeletion of unused managed ip uuid2-outbound-pip-v4 failed with error: error"), fmt.Errorf("failed to cleanup unused managed ips\ndeletion of unused managed ip uuid2-outbound-pip-v4 failed with error: error\ndeletion of unused managed ip uuid1-outbound-pip-v4 failed with error: error")},
 		},
 		{
 			name:  "all errors propagated",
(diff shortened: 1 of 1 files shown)


s-amann (approved):

kimorris27 (approved):

pkg/cluster/loadbalancerprofile_test.go
```diff
@@ -866,7 +866,7 @@ func TestReconcileLoadBalancerProfile(t *testing.T) {
 					},
 				},
 			},
-			expectedErr: fmt.Errorf("failed to cleanup unused managed ips\ndeletion of unused managed ip uuid1-outbound-pip-v4 failed with error: error\ndeletion of unused managed ip uuid2-outbound-pip-v4 failed with error: error"),
+			expectedErr: []error{fmt.Errorf("failed to cleanup unused managed ips\ndeletion of unused managed ip uuid1-outbound-pip-v4 failed with error: error\ndeletion of unused managed ip uuid2-outbound-pip-v4 failed with error: error"), fmt.Errorf("failed to cleanup unused managed ips\ndeletion of unused managed ip uuid2-outbound-pip-v4 failed with error: error\ndeletion of unused managed ip uuidi-outbound-pip-v4 failed with error: error")},
```
s-amann, line 869: typo: uuidi -> uuid1

SudoBrendan (commented):
Thanks for the PR and quick turnaround! The implementation here solves the bug and I'd be happy to merge.

Considering that the test fails because gomock EXPECT calls are non-deterministic, an alternative to consider would be to change the mocks to have deterministic order of deletion on IPs, that is, lines 856-857 become:

```golang
gomock.InOrder(
    publicIPAddressClient.EXPECT().DeleteAndWait(gomock.Any(), "clusterRG", "uuid1-outbound-pip-v4").Return(fmt.Errorf("error")),
    publicIPAddressClient.EXPECT().DeleteAndWait(gomock.Any(), "clusterRG", "uuid2-outbound-pip-v4").Return(fmt.Errorf("error")),
)
```

This deterministic ordering would mimic production code functionality (IPs are deleted in a certain order, and the log message reflects that ordering). I think there are actually other bugs we might not have caught yet, considering that other EXPECT calls are also not in guaranteed order (e.g. `publicIPAddressClient.EXPECT().List` happens multiple times in this test too). We likely need to follow up and ensure those other ordering issues aren't causing flakes. I can create a new Jira card for this.

Granted, I think the implementation as written improves current state and reduces flakiness for this case (as long as we fix the typo Spencer mentioned), so I wouldn't hold a merge on it. I only mean to provide info - that [`gomock` is not deterministic in ordering of EXPECT calls](https://pkg.go.dev/github.com/golang/mock/gomock#pkg-overview), and this is one way they suggest to accomplish that ordering when needed :)

pkg/cluster/loadbalancerprofile_test.go
lranjbar: Thanks!

lranjbar:
Updated to fix the typo.

SudoBrendan (approved):

kimorris27 (approved):

pkg/cluster/loadbalancerprofile_test.go
```diff
@@ -866,7 +866,7 @@ func TestReconcileLoadBalancerProfile(t *testing.T) {
 					},
 				},
 			},
-			expectedErr: fmt.Errorf("failed to cleanup unused managed ips\ndeletion of unused managed ip uuid1-outbound-pip-v4 failed with error: error\ndeletion of unused managed ip uuid2-outbound-pip-v4 failed with error: error"),
+			expectedErr: []error{fmt.Errorf("failed to cleanup unused managed ips\ndeletion of unused managed ip uuid1-outbound-pip-v4 failed with error: error\ndeletion of unused managed ip uuid2-outbound-pip-v4 failed with error: error"), fmt.Errorf("failed to cleanup unused managed ips\ndeletion of unused managed ip uuid2-outbound-pip-v4 failed with error: error\ndeletion of unused managed ip uuid1-outbound-pip-v4 failed with error: error")},
```
cblecker, line 869: Will the errors always be in this order? Slice ordering is not always guaranteed

cblecker (approved):

PR #3132, opened 2023-08-31
Exclude some more log namespaces

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-1414

Reduces logging by removing more customer-owned openshift- namespaces that we don't need to collect.

### Test plan for issue:

E2E should tell us if it breaks things :)

N/A

### DIFF
--- a/pkg/operator/controllers/genevalogging/const.go
+++ b/pkg/operator/controllers/genevalogging/const.go
@@ -73,7 +73,7 @@ const (
 [FILTER]
 	Name grep
 	Match containers
-	Regex NAMESPACE ^(?:default|kube-.*|openshift|(?!openshift-(logging|gitops|user-workload-monitoring))(openshift-.*))$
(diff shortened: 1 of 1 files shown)


pkg/operator/controllers/genevalogging/const.go
```diff
@@ -73,7 +73,7 @@ const (
 [FILTER]
 	Name grep
 	Match containers
-	Regex NAMESPACE ^(?:default|kube-.*|openshift|(?!openshift-(logging|gitops|user-workload-monitoring))(openshift-.*))$
+	Regex NAMESPACE ^(?:default|kube-.*|openshift|(?!openshift-(logging|gitops|user-workload-monitoring|adp|distributed-tracing|cluster-samples-operator|cnv|serverless|pipelines|nfd))(openshift-.*))$
```
cblecker, line 76: The samples operator is a part of the platform

cblecker (dismissed):

pkg/operator/controllers/genevalogging/const.go
hawkowl: Fair enough. I haven't seen a time we've needed to use the logs of it while on-call, but that's probably a different discussion (trimming cx things vs trimming things that aren't useful)

cblecker (approved):

PR #3133, opened 2023-08-31
increase machine health check node unready timeout to 15m

s-amann:

Fixes https://issues.redhat.com/browse/ARO-4040


bennerv (approved):
lgtm

The 15m change is due to customers having issues during upgrade where nodes would be NotReady while they are rebooting and running `rpm-ostree` commands to upgrade to the latest RHCOS versions.  

MHC was killing worker nodes causing some disruption during upgrades.  This should prevent that.  It will "negatively" impact the removal of NotReady nodes in a sense they will take longer to replace. 

We should confirm the Node Not Ready monitor shouldn't fire within this 15m timeframe, but a bit after (30m or so).

cadenmarchese (approved):

PR #3134, opened 2023-08-31
Dynamic validation of "EncryptionAtHost" feature subscription level registration at the RP.

schiruma:

[ARO-3211](https://issues.redhat.com/browse/ARO-3211)

When customers are attempting to set EncryptionAtHost, but that feature is not enabled for their Subscriptions, an error message describing about the issue and its resolution is displayed to them. This is achieved by dynamically validating their Subscription in the RP's frontend.
### Test plan for issue:

E2E

N/A


python/az/aro/azext_aro/aaz/__init__.py
```diff
@@ -1,3 +1,6 @@
+# Copyright (c) Microsoft Corporation.
```
tsatam, line 1: This looks like an inadvertent application of the RP's license to Azure CLI code.

If I'm understanding this correctly, this PR should not be changing any Python code, so this entire folder can be backed out.

pkg/frontend/encryptionathost_validation.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	subFeatureRegistrationsClient, err := armfeatures.NewSubscriptionFeatureRegistrationsClient(subscriptionID, credential, nil)
+	if err != nil {
+		return err
+	}
+	return validateEncryptionAtHost(ctx, subFeatureRegistrationsClient, oc)
+}
+
+func validateEncryptionAtHost(ctx context.Context, subFeatureRegistrationsClient features.SubscriptionFeatureRegistrationsClient, oc *api.OpenShiftCluster) error {
```
tsatam, line 36: (discussed during synchronous review)

Suggestion: This can be simplified as follows

```

func validateEncryptionAtHost(ctx context.Context, subFeatureRegistrationsClient features.SubscriptionFeatureRegistrationsClient, oc *api.OpenShiftCluster) error {
	if clusterRequiresEncryptionAtHost(oc.Properties.MasterProfile, oc.properties.WorkerProfiles) {
		return IsRegisteredForEncryptionAtHostFeature(ctx, subFeatureRegistrationsClient)
	}

	return nil
}

func clusterRequiresEncryptionAtHost(mp api.MasterProfile, wps []api.WorkerProfile) bool {
	if mp.EncryptionAtHost == api.EncryptionAtHostEnabled {
		return true
	}

	for _, wp := range workerProfiles {
		if wp.EncryptionAtHost == api.EncryptionAtHostEnabled {
			return true
		}
	}

	return false
}
```

SudoBrendan (commented):
As we talked about in review - this code looks great! I agree with Tanmay's comments. In addition, I'm going to put a `hold` on this until we have the production permissions changes in place required to make these feature API calls. Nice work!

python/az/aro/azext_aro/aaz/__init__.py
schiruma: reverted these changes

pkg/frontend/encryptionathost_validation.go
schiruma: As discussed earlier, I have implemented an alternate logic. The suggested one looks great, but tried to achieve the same without having to implement another func here.

jboutaud:
error message looks good to me

tsatam (dismissed):

go.mod
```diff
@@ -4,8 +4,8 @@ go 1.18
 
 require (
 	github.com/Azure/azure-sdk-for-go v63.1.0+incompatible
-	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.7.1
-	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.3.1
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.6.1
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.3.0
```
cadenmarchese, line 8: Is it needed? @kimorris27 is currently working towards moving the ARO operator to a chained credential which may require this version of `azidentity`.

pkg/frontend/encryptionathost_validation.go
```diff
@@ -0,0 +1,71 @@
...
+
+func IsRegisteredForEncryptionAtHostFeature(ctx context.Context, subFeatureRegistrationsClient features.SubscriptionFeatureRegistrationsClient) error {
+	response, err := subFeatureRegistrationsClient.Get(ctx, "Microsoft.Compute", "EncryptionAtHost", nil)
+	if err != nil {
+		return err
+	}
+	if *response.Properties.State == armfeatures.SubscriptionFeatureRegistrationStateRegistered {
+		return nil
+	}
+	return &api.CloudError{
```
cadenmarchese, line 63: This is kind of a nit, but I think it reads better to `return nil` as default, flip the conditional on line 60 to `if *response.Properties.State != armfeatures.SubscriptionFeatureRegistrationStateRegistered`, and then return the cloud error there instead.

cadenmarchese (commented):
The validation code looks good to me, I'd just make sure we have an approval from @kimorris27 before proceeding since it makes a change to the vendored azidentity.

go.mod
kimorris27: I think we can avoid changing the version in this PR. I cloned this branch and changed `azidentity` back to `v1.3.1`, and a `go mod tidy` didn't change anything else aside from setting `azcore` back to `v1.7.1`.

@schiruma, would you mind changing the versions for `azcore` and `azidentity` back to the values on the master branch and then making sure your code continues to work? I don't think using 1.3.0 instead of 1.3.1 will cause problems, but it would be nice to stick with the newer one since I had already updated us to it. Might as well keep those patch contents added in 1.3.1.

pkg/frontend/encryptionathost_validation.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return nil
+}
+
+func IsRegisteredForEncryptionAtHostFeature(ctx context.Context, subFeatureRegistrationsClient features.SubscriptionFeatureRegistrationsClient) error {
+	response, err := subFeatureRegistrationsClient.Get(ctx, "Microsoft.Compute", "EncryptionAtHost", nil)
+	if err != nil {
+		return err
+	}
+	if *response.Properties.State == armfeatures.SubscriptionFeatureRegistrationStateRegistered {
```
kimorris27, line 60: Just to make sure we don't get any surprise nil pointer dereferences, let's check for nil pointers as needed before evaluating `*response.Properties.State`.

kimorris27 (changes requested):
The validation logic looks good! I requested small changes in a few different places. Let me know if I can further clarify any of those comments.

pkg/frontend/encryptionathost_validation.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+type EncryptionAtHostValidator interface {
+	ValidateEncryptionAtHost(ctx context.Context, subscriptionID string, oc *api.OpenShiftCluster) error
+}
+
+type encryptionAtHostValidator struct{}
+
+func (e encryptionAtHostValidator) ValidateEncryptionAtHost(ctx context.Context, subscriptionID string, oc *api.OpenShiftCluster) error {
+	credential, err := azidentity.NewDefaultAzureCredential(nil)
```
bennerv, line 24: This should be using client certificate credentials, not default azure credentials.

In dev, it will use our shared tooling SP, which has access to everything I believe.  In INT/Prod, it will use the MSI's service principal.  
```
	credential, err := azidentity.NewClientCertificateCredential(tenantID, d.fpClientID, fpCertificates, fpPrivateKey, options)

**Review by bennerv — CHANGES_REQUESTED** [review]

File: go.mod (hunk 3 above)
Comment (schiruma) [reply go.mod]: @kimorris27 made changes as mentioned.

File: pkg/frontend/encryptionathost_validation.go (hunk 4 above)
Comment (schiruma) [reply pkg/frontend/encryptionathost_validation.go]: @cadenmarchese This was a good call out. Made changes as mentioned.

File: pkg/frontend/encryptionathost_validation.go [hunk 7]
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return err
+	}
+
+	subFeatureRegistrationsClient, err := armfeatures.NewSubscriptionFeatureRegistrationsClient(subscriptionID, credential, nil)
```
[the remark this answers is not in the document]
schiruma: made changes as suggested.

pkg/util/azureclient/mgmt/features/subscriptionfeatureregistrations.go
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armfeatures"
+)
+
+// SubscriptionFeatureRegistrationsClient is a minimal interface for azure SubscriptionFeatureRegistrationsClient
+type SubscriptionFeatureRegistrationsClient interface {
```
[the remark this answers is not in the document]
schiruma: made changes as suggested.

pkg/frontend/encryptionathost_validation.go
schiruma: @bennerv made changes as mentioned.

pkg/frontend/encryptionathost_validation.go
schiruma: In the new commit, since we are returning nil by default it shouldn't cause any nil pointer dereferences. Let me know your thoughts.

kimorris27: I see that you flipped the conditional in response to @cadenmarchese 's comment, but that won't stop us from potentially encountering a nil pointer dereference here. I would still check for nil pointers before trying `if *response.Properties.State ...`.

kimorris27 (changes requested):
Looking good, but I hadn't noticed when I looked last week that you had imported the subscription features client from `azure-sdk-for-go`'s `sdk` package. My requested change is to import the client from the `services` package, which I gave a link to in the more detailed comment.

pkg/util/azureclient/mgmt/features/subscriptionfeatureregistrations.go
```diff
@@ -0,0 +1,35 @@
...
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armfeatures"
```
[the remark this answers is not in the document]
bennerv: As part of the track2 migration  (which we haven't started but at some point will), we will move over clients one-by-one to the new track2.  I would prefer any net-new work uses the newer track2 sdk package.  

Otherwise we're going to come back to this and "reimplement" it using the newer track2 sdk.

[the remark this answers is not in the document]
schiruma: @bennerv thank you for this context as it will make sense to anyone looking at this code.

SudoBrendan:
Closing this and moving to an updated branch in https://github.com/Azure/ARO-RP/pull/3826

PR #3137, opened 2023-08-31
GenevaAction for etcd Certificate Renewal

SrinivasAtmakuri:

JIRA: https://issues.redhat.com/browse/ARO-3989
KCS[1]: https://access.redhat.com/solutions/7025431

The Geneva Action applies the KCS[1] with extra validations and tries recovery incase of issue.

Fixes

### DIFF
--- a/pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
+++ b/pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
@@ -1,0 +1,451 @@
+package frontend
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/Azure/go-autorest/autorest/azure"
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/sirupsen/logrus"
+	"github.com/ugorji/go/codec"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	kruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
+	"github.com/Azure/ARO-RP/pkg/frontend/adminactions"
+	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+	utilcert "github.com/Azure/ARO-RP/pkg/util/cert"
+	utilpem "github.com/Azure/ARO-RP/pkg/util/pem"
+	"github.com/Azure/ARO-RP/pkg/util/steps"
+	"github.com/Azure/ARO-RP/pkg/util/version"
+)
+
+type etcdrenew struct {
+	log           *logrus.Entry
+	k             adminactions.KubeActions
+	doc           *api.OpenShiftClusterDocument
+	secretNames   []string
+	mode          string
+	backupSecrets map[string][]byte
+	lastRevision  int32
+	timeout       time.Duration
+}
+
+var etcdOperatorControllerConditionsExpected = map[string]operatorv1.ConditionStatus{
+	"EtcdCertSignerControllerDegraded": operatorv1.ConditionFalse,
+}
+
+var etcdOperatorConditionsExpected = map[configv1.ClusterStatusConditionType]configv1.ConditionStatus{
+	configv1.OperatorAvailable: configv1.ConditionTrue,
+	configv1.OperatorDegraded:  configv1.ConditionFalse,
+}
+
+func (f *frontend) postAdminOpenShiftClusterEtcdCertificateRenew(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
+	r.URL.Path = filepath.Dir(r.URL.Path)
+	resourceID := strings.TrimPrefix(r.URL.Path, "/admin")
+
+	err := f._postAdminOpenShiftClusterEtcdCertificateRenew(ctx, resourceID, log, 30*time.Minute)
+
+	adminReply(log, w, nil, nil, err)
+}
+
+// validate cluster is <4.9 and etcd operator is in expected state
+// Secrets exists, unexpired and close to expiry
+// backup and delete secrets, if backupAndDelete is set True
+func (e *etcdrenew) validateEtcdAndBackupDeleteSecretOnFlagSet(ctx context.Context, backupAndDelete bool) error {
+	s := []steps.Step{
+		steps.Action(e.validateEtcdOperatorControllersState),
+		steps.Action(e.validateEtcdOperatorState),
+		steps.Action(e.validateEtcdCertsExistsAndExpiry),
+	}
+
+	if backupAndDelete {
+		s = append(s,
+			steps.Action(e.fetchEtcdCurrentRevision),
+			steps.Action(e.backupEtcdSecrets),
+			steps.Action(e.deleteEtcdSecrets),
+		)
+	}
+
+	_, err := steps.Run(ctx, e.log, 10*time.Second, s, nil)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Etcd secrets are deleted or updated, a new revision is will put and applied
+// This function polls if a new revision is applied successfully
+func (e *etcdrenew) isEtcDRootCertRenewed(ctx context.Context) error {
+	s := []steps.Step{
+		steps.Condition(e.isEtcdRevised, e.timeout, true),
+	}
+	_, err := steps.Run(ctx, e.log, 30*time.Second, s, nil)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (e *etcdrenew) revertChanges(ctx context.Context) error {
+	s := []steps.Step{
+		steps.Action(e.fetchEtcdCurrentRevision),
+		steps.Action(e.recoverEtcdSecrets),
+		steps.Condition(e.isEtcdRevised, 30*time.Minute, true),
+	}
+	_, err := steps.Run(ctx, e.log, 10*time.Second, s, nil)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// runs the etcd renewal and recovery
+func (e *etcdrenew) run(ctx context.Context) error {
+	if err := e.validateClusterVersion(ctx); err != nil {
+		return err
+	}
+
+	// Fetch secretNames using nodeNames
+	for i := 0; i < 3; i++ {
+		nodeName := e.doc.OpenShiftCluster.Properties.InfraID + "-master-" + strconv.Itoa(i)
+		for _, prefix := range []string{"etcd-peer-", "etcd-serving-", "etcd-serving-metrics-"} {
+			e.secretNames = append(e.secretNames, prefix+nodeName)
+		}
+	}
+
+	// validate etcd and certificates, backup and delete secrets
+	if err := e.validateEtcdAndBackupDeleteSecretOnFlagSet(ctx, true); err != nil {
+		return err
+	}
+
+	// Once secrets are deleted, the operator recreates the secrets and a new etcd revision is applied
+	// On failure, proceed for recovery by applying the backupsecrets on the cluster again
+	// On success, verify the etcd state, certificates
+	err := e.isEtcDRootCertRenewed(ctx)
+	if err != nil {
+		e.log.Infoln("Attempting to recover from backup, and wait for new revision to be applied after recovery")
+		if err = e.revertChanges(ctx); err != nil {
+			return err
+		}
+		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "etcd renewal failed, recovery performed to revert the changes.")
+	}
+
+	e.log.Infoln("Etcd certificates are renewed and new revision is applied, verifying.")
+	e.mode = "renewed"
+	return e.validateEtcdAndBackupDeleteSecretOnFlagSet(ctx, false)
+}
+
+func (f *frontend) _postAdminOpenShiftClusterEtcdCertificateRenew(ctx context.Context, resourceID string, log *logrus.Entry, timeout time.Duration) error {
+	r, err := azure.ParseResourceID(resourceID)
+	if err != nil {
+		return err
+	}
+	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	switch {
+	case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
+		return api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, "", "The Resource '%s/%s' under resource group '%s' was not found.", r.ResourceType, r.ResourceName, r.ResourceGroup)
+	case err != nil:
+		return err
+	}
+
+	k, err := f.kubeActionsFactory(log, f.env, doc.OpenShiftCluster)
+	if err != nil {
+		return err
+	}
+
+	e := &etcdrenew{
+		log:           log,
+		k:             k,
+		doc:           doc,
+		secretNames:   nil,
+		mode:          "",
+		backupSecrets: make(map[string][]byte),
+		lastRevision:  0,
+		timeout:       timeout,
+	}
+
+	if err := e.run(ctx); err != nil {
+		log.Errorf("Geneva Action run failed with error %s", err.Error())
+		return err
+	}
+
+	log.Infoln("Done")
+	return nil
+}
+
+func (e *etcdrenew) validateClusterVersion(ctx context.Context) error {
+	e.log.Infoln("validating cluster version now")
+	rawCV, err := e.k.KubeGet(ctx, "ClusterVersion.config.openshift.io", "", "version")
+	if err != nil {
+		return err
+	}
+	cv := &configv1.ClusterVersion{}
+	err = codec.NewDecoderBytes(rawCV, &codec.JsonHandle{}).Decode(cv)
+	if err != nil {
+		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", fmt.Sprintf("failed to decode clusterversion, %s", err.Error()))
+	}
+	clusterVersion, err := version.GetClusterVersion(cv)
+	if err != nil {
+		return err
+	}
+	// ETCD ceritificates are autorotated by the operator when close to expiry for cluster running 4.9+
+	if !clusterVersion.Lt(version.NewVersion(4, 9)) {
+		return api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden, "", "etcd certificate renewal is not needed for cluster running version 4.9+")
+	}
+	e.log.Infof("validated: cluster version is %s", clusterVersion)
+
+	return nil
+}
+
+func (e *etcdrenew) validateEtcdOperatorControllersState(ctx context.Context) error {
+	e.log.Infoln("validating etcdOperator Controllers state now")
+	rawEtcd, err := e.k.KubeGet(ctx, "etcd.operator.openshift.io", "", "cluster")
+	if err != nil {
+		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
+	}
+	etcd := &operatorv1.Etcd{}
+	err = codec.NewDecoderBytes(rawEtcd, &codec.JsonHandle{}).Decode(etcd)
+	if err != nil {
+		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", fmt.Sprintf("failed to decode etcd object, %s", err.Error()))
+	}
+	for _, c := range etcd.Status.Conditions {
+		if _, ok := etcdOperatorControllerConditionsExpected[c.Type]; !ok {
+			continue
+		}
+		if etcdOperatorControllerConditionsExpected[c.Type] != c.Status {
+			return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "%s is in state %s, quiting.", c.Type, c.Status)
+		}
+	}
+	e.log.Infoln("EtcdOperator Controllers state is validated.")
+
+	return nil
+}
+
+func (e *etcdrenew) validateEtcdOperatorState(ctx context.Context) error {
+	e.log.Infoln("validating Etcd Operator state")
+	rawEtcdOperator, err := e.k.KubeGet(ctx, "ClusterOperator.config.openshift.io", "", "etcd")
+	if err != nil {
+		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
+	}
+	etcdOperator := &configv1.ClusterOperator{}
+	err = codec.NewDecoderBytes(rawEtcdOperator, &codec.JsonHandle{}).Decode(etcdOperator)
+	if err != nil {
+		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", fmt.Sprintf("failed to decode etcd operator, %s", err.Error()))
+	}
+	for _, c := range etcdOperator.Status.Conditions {
+		if _, ok := etcdOperatorConditionsExpected[c.Type]; !ok {
+			continue
+		}
+		if etcdOperatorConditionsExpected[c.Type] != c.Status {
+			return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "Etcd Operator is not in expected state, quiting.")
+		}
+	}
+	e.log.Infoln("Etcd operator state validated.")
+
+	return nil
+}
+
+func (e *etcdrenew) validateEtcdCertsExistsAndExpiry(ctx context.Context) error {
+	e.log.Infoln("validating if etcd certs exists and expiry")
+
+	for _, secretname := range e.secretNames {
+		e.log.Infof("validating secret %s", secretname)
+		cert, err := e.k.KubeGet(ctx, "Secret", namespaceEtcds, secretname)
+		if err != nil {
+			return err
+		}
+
+		var u unstructured.Unstructured
+		var secret corev1.Secret
+		if err = json.Unmarshal(cert, &u); err != nil {
+			return err
+		}
+		err = kruntime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &secret)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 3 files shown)


pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,391 @@
...
+	if err != nil {
+		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", fmt.Sprintf("failed to decode clusterversion, %s", err.Error()))
+	}
+	clusterVersion, err := version.GetClusterVersion(cv)
+	if err != nil {
+		return err
+	}
+	// ETCD ceritificates are autorotated by the operator when close to expiry for cluster running 4.9+
+	if !clusterVersion.Lt(version.NewVersion(4, 9)) {
+		return api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden, "", "etcd certificate renewal is not needed for cluster running version 4.9+")
```
jhoreman, line 201: Hey Srinivas. 4.9 is currently already considered EOL, correct? Can you elaborate a bit on how this action would work as our first suggestion for 4.9 clusters with a support issue should be to upgrade. Would this GA be used prior in order to facilitate said upgrade?

hlipsig: Hey Jory! TL's made the decision that we should do this pro-actively even though we know these clusters are technically out of support. The reasoning was that recovering the clusters after they expired took 2 people 6 hours, and pro-actively rotating the certificates by hand took 2 people 3 minutes. We reasoned that creating an automation that allows 1 person to accomplish this (no need to elevate permissions) was the absolute best route, and so we created an initiative and several epics for this. See: https://issues.redhat.com/browse/ARO-3983

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,391 @@
...
+
+	resourceID := strings.TrimPrefix(r.URL.Path, "/admin")
+
+	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	switch {
+	case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
+		return api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, "", "The Resource '%s/%s' under resource group '%s' was not found.", resourceType, resourceName, resourceGroupName)
+	case err != nil:
+		return err
+	}
```
rhamitarora, line 118: **Default Case is missing**, it is not mandatory to declare the default clause in a switch statement. It is entirely optional. Without it, the code will still be valid. However, as a good programming practice, you should always declare a default clause in the switch expression

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,391 @@
...
+		e.log.Infof("Deleting secret %s", secretname)
+		err := e.k.KubeDelete(ctx, "Secret", namespaceEtcds, secretname, false, nil)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (e *etcdrenew) isRevisied(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
SrinivasAtmakuri: Done.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,387 @@
...
+	if err = e.validateClusterVersion(ctx); err != nil {
+		return err
+	}
+
+	// Fetch secretNames using nodeNames
+	masterNodeNames, err := fetchNodeNames(ctx, k, log)
+	if err != nil {
+		return err
+	}
+	if len(masterNodeNames) != 3 {
```
jaitaiwan, line 137: Is there a case where a customer might have 5 nodes? I would have this set to less than 3 maybe?

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,387 @@
...
+		e.log.Infoln("Done")
+	}
+
+	if e.mode == "recovery" {
+		e.log.Println("Attempting to recover from backup")
+		if err = e.recoverEtcdSecrets(ctx); err != nil {
+			return err
+		}
+		e.log.Infoln("Recovered")
+	}
```
jaitaiwan, line 172: These two if statements can be combined by the looks of things

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,387 @@
...
+	log.Infoln("fetching node names")
+
+	nodeList, err := k.KubeList(ctx, "node", "")
+	if err != nil {
+		return nil, err
+	}
+	if err = json.Unmarshal(nodeList, &u); err != nil {
+		return nil, err
+	}
+	err = kruntime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &nodes)
```
jaitaiwan, line 213: Why are we converting the response to an unstructured type and then again to a structured type? Can't we just convert directly?

jaitaiwan (dismissed):
Couple of nits, overall LGTM

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
petrkotas: We setup masters, no more than 3. When less than 3 we get alert and need to recover the missing one.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+
+	for _, nodeName := range masterNodeNames {
+		for _, prefix := range []string{"etcd-peer-", "etcd-serving-", "etcd-serving-metrics-"} {
+			e.secretNames = append(e.secretNames, prefix+nodeName)
+		}
+	}
+
+	// validate, backup and delete secrets
+	if err = e.validateAndBackup(ctx, true); err != nil {
```
petrkotas, line 148: What is the flow? How does the certs actually get renewed? Do they get auto-created when they are deleted?

petrkotas (changes requested):
Hi @SrinivasAtmakuri the PR is good work! I am missing some more information that needs to be in place before we merge it. Can you please update according the comments?
Most importantly, the renewing logic needs to added to the code as form of comment. It is not clear how it works. Thank you!

SudoBrendan:
@SrinivasAtmakuri can you please update the description of the PR so I know what acceptance criteria I'm trying to review? If I'm looking correctly, it's this, right? https://issues.redhat.com/browse/ARO-3989

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
SudoBrendan: I left a comment to update the PR description, which would make this more clear. After digging through Jira, I was led to the KCS here - https://access.redhat.com/solutions/7025431

In short, "yes". The etcd operator auto-recovers the secret if it is deleted.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,387 @@
...
+		return err
+	}
+
+	if err = e.isRenewed(ctx); err != nil {
+		e.mode = "recovery"
```
SudoBrendan, line 153: Having multiple "modes" like this generally leads to over-complexity, it's similar to old-timey `goto` calls - they're hard to follow, review for bugs, and maintain later, because for the rest of the review, I now have to think "what state am I in right now?" and "what states COULD I be in right now?". Generally, it's safer and easier to keep things stateless.

Can we restructure the code to not require modes, so it's more declarative right in this `if` what needs to happen for each case? For example, something pseudo like:

```
if err != nil {
    e.log.Println("Attempting Recovery of etcd cluster")
    // all the functions required to recover, maybe wrapped in a giant e.RecoverEtcdCluster() function or similar...
} else {
    e.log.Println("Attempting to renew etcd certificates")
    // all the functions required to renew certs, maybe wrapped in a giant e.RenewEtcdCertificates() function or similar...
}
e.log.Println("Done")

File: pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go [hunk 9]
```diff
@@ -0,0 +1,387 @@
...
+	} else {
+		e.mode = "renewed"
+	}
+
+	if e.mode == "renewed" {
+		// only validate
+		if err = e.validateAndBackup(ctx, false); err != nil {
```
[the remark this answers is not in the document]
SudoBrendan: I agree. It would likely be clearer to separate these into two unique functions. Generally when I see `and` in a function name, that's the moment at which I scratch my head and say "should this be two functions?"

In addition, with this function name, it's hard to tell _what_ is being validated and/or backed up, so adding some qualifiers would be helpful - are we validating secrets? certificates? etcd readiness?

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,387 @@
...
+		if _, ok := etcdOperatorControllerConditionsExpected[c.Type]; !ok {
+			continue
+		}
+		if etcdOperatorControllerConditionsExpected[c.Type] != c.Status {
+			return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "%s is in state %s, quiting.", c.Type, c.Status)
+		}
+	}
+	if e.mode == "renew" {
+		e.lastRevision = etcd.Status.LatestAvailableRevision
+	}
```
SudoBrendan, line 248: Please correct me if I'm wrong, but I think this is another case of state creeping into something that could be stateless (and in a function starting with `validate` - something I'd expect to not modify state and only retrieve/read values).

Can we determine `lastRevision` earlier in our reconciliation, perhaps unconditionally? If we do so, then we could use it when/if it's relevant, right? If not, why is it important to only determine last revision _now_ in the code? I'm unclear if so.

SudoBrendan (changes requested):
Thanks for your work on this so far, I think it's nearly there. As other commenters are pointing out, the code is a bit complex. I think that's because it is unnecessarily adding state into parts of the workflow. Can we restructure the code to not require this state?

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
SrinivasAtmakuri: Made the changes, and introduced a new function to capture the lastRevision. Please review.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,387 @@
...
+
+	if e.mode == "renewed" {
+		// only validate
+		if err = e.validateAndBackup(ctx, false); err != nil {
+			return err
+		}
+		e.log.Infoln("Done")
+	}
+
+	if e.mode == "recovery" {
```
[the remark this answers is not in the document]
SrinivasAtmakuri: Changes made.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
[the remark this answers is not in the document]
SrinivasAtmakuri: I made changes by introducing an explicit name to the function, please take a look.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+
+	_, err := steps.Run(ctx, e.log, 10*time.Second, s, nil)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (e *etcdrenew) isRenewed(ctx context.Context) error {
```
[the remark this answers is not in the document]
SrinivasAtmakuri: Made the change.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,387 @@
...
+	backupSecrets map[string][]byte
+	lastRevision  int32
+}
+
+var etcdOperatorControllerConditionsExpected = map[string]operatorv1.ConditionStatus{
+	"EtcdCertSignerControllerDegraded": operatorv1.ConditionFalse,
+	// "EtcdMembersAvailable":             operatorv1.ConditionTrue,
+	// "NodeInstallerProgressing":         operatorv1.ConditionFalse,
+	// "NodeControllerDegraded":           operatorv1.ConditionFalse,
+	// "EtcdMembersProgressing":           operatorv1.ConditionFalse,
```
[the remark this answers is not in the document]
SrinivasAtmakuri: Done, removed the dead code.

petrkotas (dismissed):
Hi @SrinivasAtmakuri I see all comments are addressed. This is a really good work. Thank you

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
cadenmarchese: I think it's good as-is here since it's the convention we have in other adminactions, and the purpose with the switch/case here is to handle errors from cosmosdb only, not run the subsequent code.

pkg/util/cert/cert.go
```diff
@@ -0,0 +1,27 @@
+package cert
```
SudoBrendan, line 1: nit: might be good in a follow-up to put these in either `pkg/util/pem` or `/pkg/util/pki` - these feel like a good fit there.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+
+	e.log.Infoln("Etcd certificates are renewed and new revision is applied, verifying.")
+	e.mode = "renewed"
+	return e.validateEtcdAndBackupDeleteSecretOnFlagSet(ctx, false)
+}
+
+func (e *etcdrenew) validateClusterVersion(ctx context.Context) error {
+	e.log.Infoln("validating cluster version now")
+	rawCV, err := e.k.KubeGet(ctx, "ClusterVersion.config.openshift.io", "", "version")
```
cadenmarchese, line 184: nit, but the extra `name` in this call to KubeGet is optional:
```
oc get clusterversion.config.openshift.io
NAME      VERSION                              AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.14.0-0.nightly-2023-08-08-222204   True        False         35d     Cluster version is 4.14.0-0.nightly-2023-08-08-222204

oc get clusterversion.config.openshift.io version
NAME      VERSION                              AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.14.0-0.nightly-2023-08-08-222204   True        False         35d     Cluster version is 4.14.0-0.nightly-2023-08-08-222204
```

cadenmarchese, line 184:
```suggestion
	rawCV, err := e.k.KubeGet(ctx, "ClusterVersion.config.openshift.io", "", "")
```

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,431 @@
...
+	// ETCD ceritificates are autorotated by the operator when close to expiry for cluster running 4.9+
+	if !clusterVersion.Lt(version.NewVersion(4, 9)) {
+		return api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden, "", "etcd certificate renewal is not needed for cluster running version 4.9+")
+	}
+	e.log.Infof("validated: cluster version is %s", clusterVersion)
+
+	return nil
+}
+
+func fetchNodeNames(ctx context.Context, k adminactions.KubeActions, log *logrus.Entry) ([]string, error) {
```
cadenmarchese, line 206: Instead of having this separate function to check master node names, what's to stop us from just constructing the expected master node names like we did with the secret names? For example, the three master node names will always be clustername+infraID+master-*, unless there has been an unsupported change or master replacement. For example:
```
oc get nodes
NAME                                           STATUS   ROLES                  AGE   VERSION
steven-wi-cluster-pzqvm-master-0               Ready    control-plane,master   35d   v1.27.3+e123787
steven-wi-cluster-pzqvm-master-1               Ready    control-plane,master   35d   v1.27.3+e123787
steven-wi-cluster-pzqvm-master-2               Ready    control-plane,master   35d   v1.27.3+e123787
```

SudoBrendan (dismissed):
I didn't have time to complete a full review, sorry, but one issue that's worth bringing up now, I'll return in a bit.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,431 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (e *etcdrenew) revertChanges(ctx context.Context) error {
+	s := []steps.Step{
+		steps.Action(e.fetchEtcdCurrentRevision),
+		steps.Action(e.recoverEtcdSecrets),
```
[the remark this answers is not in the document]
petrkotas: We have discussed this with @SrinivasAtmakuri. We decided not to dump the secret for following reasons:

- The code is only short lived, once we have all cluster on 4.9, we can remove it. Meaning no need for over engineering
- There is a KCS, which Srini can link, that has step by step process to mitigate issue when certificate is expired or lost
- Even with moderate risk more cluster would work and not fail

pkg/util/cert/cert.go
SrinivasAtmakuri: Will take this up in a different PR, thank you.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
SrinivasAtmakuri: Made the changes, thank you.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
SrinivasAtmakuri: I have added "version" to make it more readable :)

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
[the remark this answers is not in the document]
SrinivasAtmakuri: @SudoBrendan @petrkotas I think the following KCS: https://access.redhat.com/solutions/7023254 can be applied if we fail during recovery and the secrets go missing. The KCS is applied when all etcd secrets of the cluster are expired and certificates need to be manually created. (P.S. This was not tested)

I have taken some time out today and have added changes to copy/delete the certificates as blobs in "aro" containers of the cluster Storage Account.

The code doesn't try reading the storage account during recovery and fails, but will let SREs perform a manual recovery if this goes sideways.

Please review. Thank you.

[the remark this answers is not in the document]
SrinivasAtmakuri: Also, I have made a few functions as Conditional steps, to have more retries as sugggested.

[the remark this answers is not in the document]
hawkowl: wrt retries -- if we retry unconditionally for up to a minute or so, we should be able to do it fine. We shouldn't have it as Conditions, though, since those are meant to be idempotent.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew_test.go
```diff
@@ -0,0 +1,508 @@
...
+}
+
+func tweakTemplateFn(notBefore time.Time, notAfter time.Time) func(*x509.Certificate) {
+	return func(template *x509.Certificate) {
+		template.NotBefore = notBefore
+		template.NotAfter = notAfter
+	}
+}
+
+func CreateCertSecret(t *testing.T, notBefore time.Time, notAfter time.Time) []byte {
```
jaitaiwan, line 488: Nit: Not sure why this function is uppercase

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -0,0 +1,429 @@
...
+	e.log.Infoln("backing up etcd secrets done")
+	return nil
+}
+
+// delete the etcd secrets and on successful deletion,
+// valid secrets will be recreated and a new revision will be applied by the etcd operator
+func (e *etcdrenew) deleteEtcdSecrets(ctx context.Context) error {
+	e.log.Infoln("deleting etcd secrets now")
+	for _, secretname := range e.secretNames {
+		return retry.OnError(wait.Backoff{
```
hawkowl, line 354: this shouldn't be a return as it only does one secret

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew_test.go
```diff
@@ -0,0 +1,508 @@
...
+				},
+			},
+			notBefore: time.Now(),
+			notAfter:  time.Now().AddDate(3, 0, 0),
+			mocks: func(tt *test, k *mock_adminactions.MockKubeActions) {
+				k.EXPECT().
+					KubeGet(gomock.Any(), "ClusterVersion.config.openshift.io", "", "version").MaxTimes(1).
+					Return(encodeClusterVersion(t, tt.version), nil)
+				k.EXPECT().
+					KubeGet(gomock.Any(), "etcd.operator.openshift.io", "", "cluster").AnyTimes().
```
hawkowl, line 209: I think you want to check that it is called at least once, I think that AnyTimes() can mean 0

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew_test.go
```diff
@@ -0,0 +1,508 @@
...
+					},
+				},
+			},
+			mocks: func(tt *test, k *mock_adminactions.MockKubeActions) {
+				k.EXPECT().
+					KubeGet(gomock.Any(), "ClusterVersion.config.openshift.io", "", "version").
+					Return(encodeClusterVersion(t, tt.version), nil)
+			},
+			wantStatusCode: http.StatusForbidden,
+			wantError:      "403: Forbidden: : etcd certificate renewal is not needed for cluster running version 4.9+",
```
jaitaiwan, line 74: Nit: 403 usually indicates that an authenticated user doesn't have the right privileges to access the resource. I think a better status code would be 208 which means request successful but the resource was already processed.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew_test.go
```diff
@@ -0,0 +1,508 @@
...
+					KubeGet(gomock.Any(), "etcd.operator.openshift.io", "", "cluster").MaxTimes(2).
+					Return(encodeEtcdOperatorController(t, tt.etcdoperator), nil)
+				k.EXPECT().
+					KubeGet(gomock.Any(), "ClusterOperator.config.openshift.io", "", "etcd").AnyTimes().
+					Return(encodeEtcdOperator(t, tt.etcdCO), nil)
+				k.EXPECT().
+					KubeGet(gomock.Any(), "Secret", namespaceEtcds, gomock.Any()).AnyTimes().
+					Return(CreateCertSecret(t, tt.notBefore, tt.notAfter), nil)
+				d := k.EXPECT().
+					KubeDelete(gomock.Any(), "Secret", namespaceEtcds, gomock.Any(), false, nil).MaxTimes(9).
```
hawkowl, line 358: Don't we want it to be specifically 9 times?

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew_test.go
```diff
@@ -0,0 +1,508 @@
...
+							Type:   configv1.OperatorDegraded,
+							Status: configv1.ConditionFalse,
+						},
+					},
+				},
+			},
+			notBefore: time.Now().AddDate(-2, -8, 0),
+			notAfter:  time.Now().Add((1 * time.Hour)),
+			mocks: func(tt *test, k *mock_adminactions.MockKubeActions) {
+				k.EXPECT().
```
hawkowl, line 345: We should be testing the specific number of calls, since this ought to be deterministic?

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew_test.go
```diff
@@ -0,0 +1,508 @@
...
+					KubeGet(gomock.Any(), "ClusterVersion.config.openshift.io", "", "version").MaxTimes(1).
+					Return(encodeClusterVersion(t, tt.version), nil)
+				k.EXPECT().
+					KubeGet(gomock.Any(), "etcd.operator.openshift.io", "", "cluster").
+					Return(encodeEtcdOperatorController(t, tt.etcdoperator), nil)
+			},
+			wantStatusCode: http.StatusInternalServerError,
+			wantError:      "500: InternalServerError: : EtcdCertSignerControllerDegraded is in state True, quiting.",
```
jaitaiwan, line 112: Nit: 500 is usually reserved for when a server-caused error occurs. I'd be more inclined to use 422 which means "unprocessable content"

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew_test.go
```diff
@@ -0,0 +1,508 @@
...
+					Return(encodeEtcdOperatorController(t, tt.etcdoperator), nil)
+				k.EXPECT().
+					KubeGet(gomock.Any(), "ClusterOperator.config.openshift.io", "", "etcd").AnyTimes().
+					Return(encodeEtcdOperator(t, tt.etcdCO), nil)
+				k.EXPECT().
+					KubeGet(gomock.Any(), "Secret", namespaceEtcds, gomock.Any()).AnyTimes().
+					Return(CreateCertSecret(t, tt.notBefore, tt.notAfter), nil)
+			},
+			wantStatusCode: http.StatusInternalServerError,
+			wantError:      "500: InternalServerError: : secret etcd-peer-cluster-aro-master-0 is not near expiry, quitting.",
```
jaitaiwan, line 219: Nit: I'd see this as a 208/207 http response

jaitaiwan (approved):
I had a couple of Nits but nothing important

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew_test.go
hawkowl: not sure if geneva actions handles that

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
SrinivasAtmakuri: Done, made the change.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew_test.go
SrinivasAtmakuri: Modified.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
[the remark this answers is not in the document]
SrinivasAtmakuri: FYI, I have removed the code to save the secrets in the cluster storage account after discussions with @hawkowl @petrkotas

petrkotas (approved):
@SrinivasAtmakuri Thank you for the PR, I see that the retry is reducing the chance of fail even further. This is a good work. Lets get it in

petrkotas:
I am going to merge the PR for the sake of speeding up the release. For the following reasons:

- The code is testing by unit, that passed
- The code is affecting part of ARO that is not covered by E2E, therefore E2E will not uncover fail
- We need to merge to proceed with the release

PR #3140, opened 2023-09-01
Remove Ignition V2 from the ARO Operator

tsatam:

Part of [ARO-4065](https://issues.redhat.com/browse/ARO-4065) (changes will need to be mirrored to the installer as well)

Eliminates all usage of ignitionv2 in the ARO Operator and replaces with ignitionv3. 

This will introduce a dependency on the operator to run on OCP versions > 4.6, so PUCM has been updated to only perform operator updates on cluster versions 4.7 or greater. 

Commits are structured for easy review. 

### Test plan for issue:

- Existing Unit tests continue to pass
- New unit tests added to validate PUCM does not update the operator on clusters older than 4.7
- Clusters running the updated operator successfully update `99-aro-{master,worker}-dns` MachineConfigs to ignitionv3 and roll out node updates successfully (tested manually)

Updates to SRE PUCM documentation to indicate operator updates will not occur on pre-4.7 clusters


bennerv (dismissed):
lgtm;

few things that might be useful to test:
1. Upgrading shared cluster ignition v2 -> ignitionv3 via the operator
2. Taking a 4.7 cluster and making sure the ignitionv3 version set works for it
3. Taking a cluster born <4.7 and ensure the ignitionv3 types are compatible on 4.7+.  There's some stuff with the way the original host boots up needing it's original RHCOS & ignition config to be compatible.  I believe it should based on examining older clusters born on <4.6, but would be good to be sure.  

I know it's a lot to ask.  Sudharsan has spun up a couple 4.8 clusters already so if you don't mind testing those scenarios and have time, he might be able to create or show some steps on how to create older clusters.  Otherwise I think i have the steps documented somewhere.

tsatam:
@bennerv I can test these scenarios if still necessary; but as far as I am aware, the MCO has been up-converting these ARO Dnsmasq MachineConfigs to 3.2 in the `rendered-*` MachineConfigs [since OpenShift 4.7](https://github.com/openshift/machine-config-operator/blob/release-4.7/pkg/server/api.go#L146), so the MC that actually gets applied to nodes has always been 3.2. 

I have set the Ignition version on this MachineConfig to be the max version provided by our vendored Ignition library, but perhaps we want to pin that to 3.2 instead and track `the default version supported by MCO on the earliest cluster version we support` instead? 

I'm going to rerun the test I did last week (run the updated operator on a freshly-created 4.12 cluster) and ensure that the diff between the rendered machineconfigs is trivial.

bennerv:
>I have set the Ignition version on this MachineConfig to be the max version provided by our vendored Ignition library, but perhaps we want to pin that to 3.2 instead and track the default version supported by MCO on the earliest cluster version we support instead?
Sounds good.  

When I looked at a cluster born on ~4.4, all the machine configs for kubelet and containerruntimeconfigs were created at 3.2, so it shouldn't be an issue.  I'm fine with your plan proceeding.

tsatam:
Created a 4.12 cluster, and ran the operator from this PR against it. 

The only diff between the old and new rendered MCs as a result of this appears to be the order of the applied dnsmasq files (actual script contents replaced for brevity):

```diff
7c7
<   creationTimestamp: "2023-09-05T14:33:57Z"
---
>   creationTimestamp: "2023-09-05T15:11:00Z"
9c9
<   name: rendered-master-45bcbf1ade58de0480f3bc08816f6793
---
>   name: rendered-master-cc633e2578a9aaeca0c4f3de3f417a04
17,18c17,18
<   resourceVersion: "30891"
<   uid: a6f2769d-aff9-450c-a174-72c9df1d70f6
---
>   resourceVersion: "56736"
>   uid: e6dcf9e9-9aa3-46fc-b8b6-08b2d89d17b1
208,209c208,209
<           source: data:text/plain;charset=utf-8;base64,${DNSMASQ_RESTART}
<         mode: 484
---
>           source: data:text/plain;charset=utf-8;base64,${DNSMASQ_CONF}
>         mode: 420
211c211
<         path: /etc/NetworkManager/dispatcher.d/99-dnsmasq-restart
---
>         path: /etc/dnsmasq.conf
222,223c222,223
<           source: data:text/plain;charset=utf-8;base64,${DNSMASQ_CONF}
<         mode: 420
---
>           source: data:text/plain;charset=utf-8;base64,${DNSMASQ_RESTART}
>         mode: 484
225c225
<         path: /etc/dnsmasq.conf
---
>         path: /etc/NetworkManager/dispatcher.d/99-dnsmasq-restart
```

tsatam:
Correct. All the disparate Ignition configs across all MachineConfigs for the node type are merged into a single Ignition config at the desired version (Ignition 3.2 for OCP 4.7-4.13).

s-amann (dismissed):

jhoreman (approved):

pkg/operator/controllers/dnsmasq/dnsmasq.go
```diff
@@ -88,51 +89,55 @@ func ignition2Config(clusterDomain, apiIntIP, ingressIP string, gatewayDomains [
...
 					},
-					FileEmbedded1: ign2types.FileEmbedded1{
-						Contents: ign2types.FileContents{
-							Source: dataurl.EncodeBytes(startpre),
+					FileEmbedded1: ign3types.FileEmbedded1{
+						Contents: ign3types.Resource{
+							Source: ignutil.StrToPtr(dataurl.EncodeBytes(startpre)),
 						},
-						Mode: to.IntPtr(0744),
+						Mode: ignutil.IntToPtr(0744),
```
s-fairchild, line 131: I believe there was an effort to remove our dependency on ignition util not too long ago. Any reason to use it over `to`?

pkg/operator/controllers/dnsmasq/dnsmasq_restart.go
```diff
@@ -7,8 +7,8 @@ import (
 	"bytes"
 	"text/template"
 
-	"github.com/Azure/go-autorest/autorest/to"
-	ign2types "github.com/coreos/ignition/config/v2_2/types"
+	ignutil "github.com/coreos/ignition/v2/config/util"
```
s-fairchild, line 10: Same as comment above

pkg/operator/controllers/dnsmasq/dnsmasq.go
tsatam: This was just the result of me carrying over the existing ignition3 config from installer-aro, which used these conversion functions: https://github.com/jewzaam/installer-aro/blob/96c4ea6a1ddb8d3885bae9fe7648462a9d6580d1/pkg/aro/dnsmasq/dnsmasq.go#L275 

I'm happy to switch these back over.

pkg/cluster/install.go
```diff
@@ -156,6 +161,17 @@ func (m *manager) adminUpdate() []steps.Step {
 	return toRun
 }
 
+func (m *manager) shouldUpdateOperator() bool {
+	runningVersion, err := semver.NewVersion(m.doc.OpenShiftCluster.Properties.ClusterProfile.Version)
```
cadenmarchese, line 165: I know it's convention to use the cluster doc, but Is it possible to look directly at the cluster version operator resources to get the running version rather than use the doc?

My concern is just around potential drift or a race between the version in-cluster and the version reporting in the cluster doc. If the cluster version in the doc is guaranteed to be current by the time this step runs, I think we're ok to leave it as-is. It's not totally clear to me looking at code where we update the doc with the current running version.

pkg/operator/controllers/dnsmasq/dnsmasq.go
tsatam: Made the change

pkg/cluster/install.go
tsatam: I had that thought too. I can investigate this further.

tsatam: So the version check here is done at the time the steps are defined, not when they are executed, so it's basically guaranteed that it will not be up to date since the check is done before any actions are performed. 

Because of this, I'm also hesitant to add any more complicated checks to e.g. the CVO resource at this stage. I think a better approach would be to do these checks within each action at run-time, and/or potentially grouping the operator reconciliation steps into a single Action with its own sub-task runner, that does the checks before execution? 

Either way would be a larger change, so I'm curious to explore the impact of the current implementation to see if we can make this improvement a post-merge activity. As far as I see it, if we PUCM a cluster that was on <=4.6 during our last PUCM but has since been upgraded, they would lose the operator upgrade from that PUCM and would have to wait for the next PUCM to be performed in order to upgrade.

tsatam: Never mind, the version update is done by the frontend, before the backend even starts defining which steps to run: 
- (frontend enrichment) https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/openshiftcluster_putorpatch.go#L120 
- (actual version update) https://github.com/Azure/ARO-RP/blob/master/pkg/util/clusterdata/cluster_version.go#L22

pkg/cluster/install.go
```diff
@@ -156,6 +161,17 @@ func (m *manager) adminUpdate() []steps.Step {
...
+	if err != nil {
+		return false
+	}
+
+	cutoffVersion := semver.New(operatorCutoffVersion)
+
+	return cutoffVersion.LessThan(*runningVersion)
```
SudoBrendan, line 172: nit? `LessThan` will return false if `runningVersion` is `4.7.0`, right? So we're excluding 4.7.0? is that intentional?

SudoBrendan (commented):
Thanks for the quick turnaround on this! Code overall looks very good to merge to me - I think I might have found a small rounding error? Otherwise LGTM!

pkg/cluster/install.go
tsatam: Confirmed. I'll get a fix up for this.

tsatam: fixed

SudoBrendan (approved):
LGTM, thanks!

petrkotas (approved):
Thanks @tsatam , I believe we can get it in. I am not very concerned about the drift. As you pointed out it done by frontend and moreover we have not observed drift in past 3 years.

PR #3144, opened 2023-09-02
Fix kube versions in go.mod

cblecker:
Follow up to #3143 

There's a few versions in the require section of go.mod that show incorrect/incompatible versions that are then being overridden by replace.. but because they are incorrect in the require section, our code vuln tools aren't showing that they have been updated.

This change should be strictly cosmetic.

### Test plan for issue:
Green CI == :shipit:

No


bennerv (approved):
lgtm

PR #3147, opened 2023-09-06
Update mirroring targets to 4.10+ and add UBI9

hawkowl:
### What this PR does / why we need it:

This updates the mirroring to be 4.10+ and includes UBI9 base images, as well as Mariner azure-cli.

It also updates the MUO mirroring to the latest promoted release.

### Test plan for issue:

merge, see if int works?

PR #3149, opened 2023-09-06
Bumping up MUO to v0.1.952

tiguelu:
https://issues.redhat.com/browse/ARO-4088
This PR follows PR #3147

Bump up MUO image to include recent bugfixes

### Test plan for issue:
**NOTE** E2E tests for the PR have passed (see [here](https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=79463173&view=logs&j=10f66467-59e5-53bc-0c8a-42cc89770de4&t=8530e8d4-fb46-5697-903b-6bef07c9756d)), however the pipeline was cancelled by Azure during cleanup phase.

Smoke test run manually:
- Create new 4.11.44 cluster (no upgrade history), which runs **aro-b4** MUO image
- Create UpgradeConfig aiming for 4.12.25 and confirm MUO pods CrashLoopBackoff
- Upgrade MUO to **v0.1.952** and confirm pods get into Running and upgrade starts
- Wait for Upgrade completion and check pods are Running and history shows in UpgradeConfig
- Create UpgradeConfig aiming for 4.13.9 and confirm MUO pods don't fail and upgrade starts (cluster with upgrade history)
- Wait for Upgrade completion and check pods are still Running and history shows in UpgradeConfig

A separate Epic [ARO-2732](https://issues.redhat.com/browse/ARO-2732) will document MUO lifecycle in ARO.

### DIFF
--- a/pkg/util/version/const.go
+++ b/pkg/util/version/const.go
@@ -89,7 +89,7 @@ func MdsdImage(acrDomain string) string {
 
 // MUOImage contains the location of the Managed Upgrade Operator container image
 func MUOImage(acrDomain string) string {
-	return acrDomain + "/managed-upgrade-operator:v0.1.952-44b631a"
+	return acrDomain + "/app-sre/managed-upgrade-operator:v0.1.952-44b631a"
 }
 
 // GateKeeperImage contains the location of the GateKeeper container image
--- a/pkg/util/version/const.go
+++ b/pkg/util/version/const.go
@@ -89,7 +89,7 @@ func MdsdImage(acrDomain string) string {
 
 // MUOImage contains the location of the Managed Upgrade Operator container image
 func MUOImage(acrDomain string) string {
-	return acrDomain + "/app-sre/managed-upgrade-operator:v0.1.952-44b631a"
+	return acrDomain + "/managed-upgrade-operator:v0.1.952-44b631a"
 }
 
 // GateKeeperImage contains the location of the GateKeeper container image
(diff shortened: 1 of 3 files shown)


pkg/util/version/const.go
```diff
@@ -89,7 +89,7 @@ func MdsdImage(acrDomain string) string {
 
 // MUOImage contains the location of the Managed Upgrade Operator container image
 func MUOImage(acrDomain string) string {
-	return acrDomain + "/managed-upgrade-operator:aro-b4"
+	return acrDomain + "/app-sre/managed-upgrade-operator:v0.1.952-44b631a"
```
bennerv, line 92: why did the `/app-sre/` change?

bennerv (approved):
one question - otherwise lgtm

pkg/util/version/const.go
tiguelu: Because the [upstream image builds in quay](https://quay.io/repository/app-sre/managed-upgrade-operator?tab=tags) are under that path, and our mirror code is later replicating it into ACR.

On the other hand, the previous aro-b4 image was a custom ARO build (needed for FIPS support at that time), which was specifically pushed to ACR, and not published under /app-sre repo.

s-amann (approved):

cadenmarchese (approved):
Thanks for fixing! LGTM

petrkotas (approved):
lgtm

PR #3150, opened 2023-09-06
Fix nil pointer dereference in subnet validation

bennerv:

Fixes a nil pointer dereference in cluster creation due to addressPrefix not being set during cluster creation.  

Switch between addressPrefix and addressPrefixes on subnets.  

### Test plan for issue:

e2e & green ci 

nope

### DIFF
--- a/pkg/validate/dynamic/dynamic.go
+++ b/pkg/validate/dynamic/dynamic.go
@@ -51,6 +51,8 @@ var (
 	errMsgInvalidVNetLocation               = "The vnet location '%s' must match the cluster location '%s'."
 )
 
+const minimumSubnetMaskSize int = 27
+
 type Subnet struct {
 	// ID is a resource id of the subnet
 	ID string
@@ -764,23 +766,41 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 			)
 		}
 
-		_, net, err := net.ParseCIDR(*ss.AddressPrefix)
-		if err != nil {
-			return err
-		}
-
-		ones, _ := net.Mask.Size()
-		if ones > 27 {
-			return api.NewCloudError(
-				http.StatusBadRequest,
-				api.CloudErrorCodeInvalidLinkedVNet,
-				s.Path,
-				errMsgSubnetInvalidSize,
-				s.ID,
-			)
+		// Handle both addressPrefix & addressPrefixes
+		if ss.AddressPrefix == nil {
+			for _, address := range *ss.AddressPrefixes {
+				if err = validateSubnetSize(s, address); err != nil {
+					return err
+				}
+			}
+		} else {
+			if err = validateSubnetSize(s, *ss.AddressPrefix); err != nil {
+				return err
+			}
 		}
 	}
 
+	return nil
+}
+
+// validateSubnetSize checks if the subnet mask is >27, and returns
+// an error if so as it is too small for OCP
+func validateSubnetSize(s Subnet, address string) error {
+	_, net, err := net.ParseCIDR(address)
+	if err != nil {
+		return err
+	}
+
+	ones, _ := net.Mask.Size()
+	if ones > minimumSubnetMaskSize {
+		return api.NewCloudError(
+			http.StatusBadRequest,
+			api.CloudErrorCodeInvalidLinkedVNet,
+			s.Path,
+			errMsgSubnetInvalidSize,
+			s.ID,
+		)
+	}
 	return nil
 }
 
--- a/pkg/validate/dynamic/dynamic_test.go
+++ b/pkg/validate/dynamic/dynamic_test.go
@@ -6,6 +6,7 @@ package dynamic
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 	"testing"
 	"time"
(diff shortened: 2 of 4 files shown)


pkg/validate/dynamic/dynamic.go
```diff
@@ -764,23 +764,41 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
 
-		_, net, err := net.ParseCIDR(*ss.AddressPrefix)
-		if err != nil {
-			return err
+		// Handle both addressPrefix & addressPrefixes
+		if ss.AddressPrefix == nil {
+			for _, address := range *ss.AddressPrefixes {
+				if err = validateSubnetSize(s, address); err != nil {
+					return err
+				}
```
cblecker, line 772: If this is validating a range, do we want to error if the first element of the slice doesn't validate?

bennerv: Honestly I'm not sure. 

I'm not even sure how openshift will handle this.  You can technically specify multiple IP address ranges on a subnet now (wasn't that the point of a subnet to begin with!??). 

For now, we've only seen customers with a single one defined.  You have to opt into the MSFT AFEC feature flag `Microsoft.Network/AllowMultipleAddressPrefixesOnSubnet` in order to use it for now.

kimorris27: >I'm not even sure how openshift will handle this. You can technically specify multiple IP address ranges on a subnet now (wasn't that the point of a subnet to begin with!??).
>
>For now, we've only seen customers with a single one defined. You have to opt into the MSFT AFEC feature flag Microsoft.Network/AllowMultipleAddressPrefixesOnSubnet in order to use it for now.

I poked around in the installer code and nothing I found indicated that the installer will be unhappy if you provide a subnet with multiple address ranges. Complicated stuff though, so it would be good to ask the installer folks to verify.

If OCP doesn't support multiple address ranges, we can error out here if the cx has put more than one and then the original question in this thread is moot.

pkg/validate/dynamic/dynamic.go
```diff
@@ -764,23 +764,41 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
...
+// validateSubnetSize checks if the subnet mask is >27, and returns
+// an error if so as it is too small for OCP
+func validateSubnetSize(s Subnet, address string) error {
+	_, net, err := net.ParseCIDR(address)
+	if err != nil {
+		return err
 	}
 
+	ones, _ := net.Mask.Size()
+	if ones > 27 {
```
AldoFusterTurpin, line 793: Maybe we can move that 27 to a const on top of the file as it is a hard-coded config.

pkg/validate/dynamic/dynamic_test.go
```diff
@@ -1727,3 +1728,39 @@ func TestCheckBYONsg(t *testing.T) {
...
+			err := validateSubnetSize(tt.subnet, tt.address)
+			if err == nil {
+				if tt.wantErr != "" {
+					t.Error(err)
+				}
+			} else {
+				if err.Error() != tt.wantErr {
+					t.Error(err)
+				}
+			}
```
AldoFusterTurpin, line 1763: This can be replaced by
```go
utilerror.AssertErrorMessage(t, err, tt.wantErr)
```

AldoFusterTurpin (dismissed):
Thanks for this! 
One minor suggestion and the usage of an existing aux function to check errors.

pkg/validate/dynamic/dynamic.go
bennerv: Thanks for the investigation.  

I also poked around in machine-config-provider-azure and there is never an explicit subnet address assignment used during NIC provisioning, so it should be handled transparently.  The only thing that is specified during creation is the subnet ID.  

So I think the proper validation here would be to add up all the address spaces and make sure the total address space is >27 mask, also taking into account the number of reserved IP addresses in smaller address spaces may have an impact.

pkg/validate/dynamic/dynamic.go
bennerv: done

pkg/validate/dynamic/dynamic_test.go
bennerv: done - good catch i didn't know that was a thing.

pkg/validate/dynamic/dynamic.go
bennerv: It's not perfect, but instead of adding up individual ranges, I'm going to validate that at least one address prefix has a /27 or greater and error out if not.  

If we have a bunch of /30s it's not going to make sense as there isn't enough address space for anything there.  The cases around this are going to get crazy, and at least we're returning a valid message to the customer.

pkg/validate/dynamic/dynamic_test.go
```diff
@@ -1727,3 +1728,31 @@ func TestCheckBYONsg(t *testing.T) {
...
+		wantErr string
+	}{
+		{
+			name:    "subnet size is too small",
+			address: "10.0.0.0/32",
+			subnet:  Subnet{ID: subnetId, Path: subnetPath},
+			wantErr: fmt.Sprintf("400: InvalidLinkedVNet: %s: The provided subnet '%s' is invalid: must be /27 or larger.", subnetPath, subnetId),
+		},
+		{
+			name:    "subnet size is gucci gang",
```
s-amann, line 1748: :rocket:

s-amann (approved):

kimorris27 (approved):

pkg/validate/dynamic/dynamic.go
```diff
@@ -764,23 +766,41 @@ func (dv *dynamic) ValidateSubnets(ctx context.Context, oc *api.OpenShiftCluster
 			)
 		}
 
-		_, net, err := net.ParseCIDR(*ss.AddressPrefix)
-		if err != nil {
-			return err
+		// Handle both addressPrefix & addressPrefixes
+		if ss.AddressPrefix == nil {
+			for _, address := range *ss.AddressPrefixes {
```
AldoFusterTurpin, line 771: Stupid question: why we don't need
```go
if ss.AddressPrefixes != nil {
```
before iterating *ss.AddressPrefixes (note the plural)?

bennerv: One or the other will always be set.

cadenmarchese (approved):

PR #3151, opened 2023-09-06
[TEST] This should panic on AddressPRefixes used for master subnets

bennerv:
Do not merge

Should panic; testing theory with e2e.


bennerv:
panics as expected.  Test failed successfully :)

PR #3154, opened 2023-09-07
feature: deny crd modification policy

ArrisLee:
https://issues.redhat.com/browse/ARO-3043

Do not allow modification of cluster generated CRDs
checking rule should be based on a white list of generated/init CRDs

Test plan for issue:
Rego unit test + Gator test

Need to update the design doc / customer doc for guardrail policies


pkg/operator/controllers/guardrails/policies/gktemplates-src/library/common.rego
```diff
@@ -139,4 +143,127 @@ privileged_ns = {
   "openshift-sdn",
   "openshift-service-ca",
   "openshift-service-ca-operator"
-}
\ No newline at end of file
+}
+
+critical_crds = {
```
yjst2012, line 148: I'm a bit hesitate to add long list to common.rego, as it will appear on all templates that use common lib, feel like it is probably a little better to keep it in the template src file until another policy wants to invoke it, wdyt?

yjst2012 (changes requested):
thanks for putting this up, generally looks good to me, I just have one suggestion.

PR #3156, opened 2023-09-07
Add NSG monitoring for preconfiguredNSG

nwnt:

Fixes: ARO-3391

Adding monitoring for NSG and emitting metrics when invalid NSG configurations have been detected, including when the FP (the one performing the monitoring) loses access to the NSG.

### Test plan for issue:

- Unit tests are added
- More tests to be performed with INT the environment.

N/A


pkg/monitor/worker.go
```diff
@@ -246,11 +246,19 @@ func (mon *monitor) workOne(ctx context.Context, log *logrus.Entry, doc *api.Ope
 		log.Warnf("no hiveShardConfigs set for shard %d", shard)
 	}
 
+	// Need to add the code here
+	// if hourlyRun and preconfiguredNSG
+	// 	create a new client
+	// 	go run new client monitor
```
niontive, line 252: Why do we need to run preconfigured NSG check as separate monitor from existing cluster monitor?

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,155 @@
...
+func (n *NSGMonitor) Monitor(ctx context.Context) error {
+	masterSubnet, err := n.toSubnetConfig(ctx, n.oc.Properties.MasterProfile.SubnetID)
+	if err != nil {
+		// FP has no access to the subnet
+		return err
+	}
+
+	workerSubnets := make([]subnetNSGConfig, 0, len(n.oc.Properties.WorkerProfiles))
+	workerPrefixes := make([]netip.Prefix, 0, len(n.oc.Properties.WorkerProfiles))
+	for _, wp := range n.oc.Properties.WorkerProfiles {
```
gouthamMN, line 106: You will also need to look at workerProfilesStatus to make sure no wp are missed.

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// to make sure each NSG is processed only once
+	nsgSet := map[string]*sdknetwork.SecurityGroup{
+		*masterSubnet.nsg.ID: masterSubnet.nsg,
+	}
+	for _, w := range workerSubnets {
+		nsgSet[*w.nsg.ID] = w.nsg
+	}
+
+	for nsgID, nsg := range nsgSet {
```
niontive, line 124: Just curious - do we have different rules for worker NSG vs. master NSG?

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,155 @@
...
+			nsgResource, err := arm.ParseResourceID(nsgID)
+			if err != nil {
+				n.log.Errorf("Unable to parse NSG resource ID: %s. %s", nsgID, err)
+				continue
+			}
+
+			r := newRuleChecker(n.log, masterSubnet.prefix, workerPrefixes, &rule)
+
+			if r.isInvalidDenyRule() {
+				dims := map[string]string{
```
niontive, line 140: For resource health check, I'd assume you'd want to use this as your health metric. Regardless, make sure to emit (1) location and (2) cluster resource ID.

pkg/monitor/azure/nsg/nsg.go
gouthamMN: With BYO-NSG, I think it's possible.

pkg/monitor/azure/nsg/rulechecker.go
```diff
@@ -0,0 +1,180 @@
...
+		return prefix, nil
+	}
+	return netip.Prefix{}, fmt.Errorf("invalid IP address or CIDR range: %s", value)
+}
+
+// toPrefixes converts a slice of addresses into a slice of Prefixes
+func toPrefixes(log *logrus.Entry, addresses []string) []netip.Prefix {
+	prefixes := make([]netip.Prefix, 0, len(addresses))
+	for _, address := range addresses {
+		if address == "*" || address == "VirtualNetwork" {
```
niontive, line 166: Nit: `address == virtualNetwork`

pkg/monitor/azure/nsg/nsg.go
nwnt: Basically each subnet can be attached with a different NSG altogether.

pkg/monitor/azure/nsg/nsg.go
nwnt: Good call.

pkg/monitor/azure/nsg/rulechecker.go
nwnt: Nice catch!  I somehow overlooked that when refactoring.

pkg/monitor/worker.go
nwnt: This will be an hourly run instead of every minute like what cluster monitor currently does.

pkg/monitor/azure/nsg/nsg.go
nwnt: Changed to using api.GetEnrichedWorkerProfiles

pkg/monitor/worker.go
niontive: Cluster monitor does keep track of hourly run. Here is an example: https://github.com/Azure/ARO-RP/blob/8ffcf691bad8e70182f34550f982e414fccde21f/pkg/monitor/cluster/cluster.go#L140

pkg/monitor/azure/nsg/nsg.go
niontive, line 1: We have a util pkg for subnets in `pkg/util/subnet`. Would it make sense to either (1) refactor some of the NSG specific code to this pkg or (2) create a new util (ex: `pkg/util/nsg`)? I'm assuming some of this NSG stuff may be used by future RP components, so having a common util would be helpful.

pkg/monitor/worker.go
nwnt: that's what I actually did in my update after the original comment.

kimorris27 (changes requested):
Nice work!

I'm concerned about getting nil pointer dereferences in `NSGMonitor.Monitor`, particularly when accessing `subnet.AddressPrefixes`, and pointed to an example of such an issue and the code changes that had to be made to address it. My other recommendations are just nits.

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/metrics"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/network"
+)
+
+const (
```
bennerv, line 23: some of these are used within the cluster monitor as additional dimmensions we pass when we construct the cluster monitor.  Can we pull these into a single place and make sure they're consistent?  

i.e. resourceGroup, subscriptionid, clusterResourceId

bennerv: https://github.com/Azure/ARO-RP/blob/22a6d6063c3c18ec3a1e39bcdc107c7b403ac9d5/pkg/monitor/cluster/cluster.go#L65-L69

Make sure the naming is consistent as well.

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,176 @@
...
+func (n *NSGMonitor) Done() <-chan error {
+	return n.done
+}
+
+func NewNSGMonitor(log *logrus.Entry, oc *api.OpenShiftCluster, subscriptionID string, environment *azureclient.AROEnvironment, cred autorest.Authorizer, emitter metrics.Emitter) *NSGMonitor {
+	return &NSGMonitor{
+		log:     log,
+		emitter: emitter,
+		oc:      oc,
+
```
bennerv, line 66: In the cluster monitor, we construct a set of dimmensions specific to the monitor itself.  I.e. 
resourceId, subscriptionId, etc.  This means that every emit gauge or metric within the monitor itself will inherit these dimmensions, and save us a few lines of code.

bennerv: https://github.com/Azure/ARO-RP/blob/22a6d6063c3c18ec3a1e39bcdc107c7b403ac9d5/pkg/monitor/cluster/cluster.go#L65-L69

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,176 @@
...
+	MetricSubnetAcessResponseCode = "preconfiguredNSG.subnetaccess.responsecode"
+	MetricSubnetAccessForbidden   = "preconfigurednsg.subnetaccess.forbidden"
+	MetricUnsuccessfulFPCreation  = "preconfigurednsg.fpcreation.unsuccessful"
+	MetricNSGMonitoringTimedOut   = "preconfigurednsg.monitoring.timedout"
+
+	expandNSG = "NetworkSecurityGroup"
+)
+
+// NSGMonitor is responsible for performing NSG rule validations when preconfiguredNSG is enabled
+type NSGMonitor struct {
```
bennerv, line 48: I'm wondering if we should have a monitor interface we create as part of this which adheres to a set of functions.  

Then any new monitors we introduce down the line would function the same and be expected to implement that interface.

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,176 @@
...
+	DimNSGRulePriority     = "priority"
+	DimClusterResourceID   = "clusterresourceid"
+	DimLocation            = "location"
+	DimTenantID            = "tenantid"
+
+	MetricInvalidDenyRule         = "preconfigurednsg.invalid.denyrule"
+	MetricSubnetAcessResponseCode = "preconfiguredNSG.subnetaccess.responsecode"
+	MetricSubnetAccessForbidden   = "preconfigurednsg.subnetaccess.forbidden"
+	MetricUnsuccessfulFPCreation  = "preconfigurednsg.fpcreation.unsuccessful"
+	MetricNSGMonitoringTimedOut   = "preconfigurednsg.monitoring.timedout"
```
bennerv, line 42: Any metrics we emit typically take the form: 
`<component>.<subcomponent>.<metric>` or some form of that.  

I would do something like for most of these 
`monitor.nsg. ...`

Naming things is hard :) - dealers choice

pkg/monitor/worker.go
```diff
@@ -246,11 +247,47 @@ func (mon *monitor) workOne(ctx context.Context, log *logrus.Entry, doc *api.Ope
...
+			// Not stopping here just because can't monitor NSG
+			log.Error("Unable to create FP Authorizer for NSG monitoring.", err)
+			mon.m.EmitGauge(nsg.MetricUnsuccessfulFPCreation, int64(1), map[string]string{
+				nsg.DimClusterResourceID: doc.OpenShiftCluster.ID,
+				nsg.DimLocation:          doc.OpenShiftCluster.Location,
+				nsg.DimSubscriptionID:    sub.ID,
+				nsg.DimTenantID:          sub.Subscription.Properties.TenantID,
+			})
+		} else {
+			nsgMon = nsg.NewNSGMonitor(log, doc.OpenShiftCluster, sub.ID, mon.env.Environment(), fpAuthorizer, mon.m)
```
bennerv, line 263:
```suggestion
			nsgMon = nsg.NewNSGMonitor(log, doc.OpenShiftCluster, sub.ID, mon.env.Environment(), fpAuthorizer, mon.clusterm)

File: pkg/monitor/azure/nsg/nsg.go [hunk 11]
```diff
@@ -0,0 +1,176 @@
...
+		if subnet.StatusCode == http.StatusForbidden {
+			n.emitter.EmitGauge(MetricSubnetAccessForbidden, int64(1), dims)
+		}
+		n.log.Errorf("Error while getting subnet %s. %s", subnetID, err)
+		return subnetNSGConfig{}, err
+	}
+
+	n.emitter.EmitGauge(MetricSubnetAcessResponseCode, int64(subnet.StatusCode), dims)
+
+	cidr, err := toPrefix(*subnet.AddressPrefix)
```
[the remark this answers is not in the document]
bennerv: Good catch Kipp!

pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -411,6 +419,14 @@ ExecStart=/usr/bin/docker run \
...
   --rm \
   --cap-drop net_raw \
+  -e AZURE_FP_CLIENT_ID \
+  -e DOMAIN_NAME \
+  -e CLUSTER_MDSD_ACCOUNT \
+  -e CLUSTER_MDSD_CONFIG_VERSION \
+  -e GATEWAY_DOMAINS \
+  -e GATEWAY_RESOURCEGROUP \
+  -e MDSD_ENVIRONMENT \
+  -e CLUSTER_MDSD_NAMESPACE \
```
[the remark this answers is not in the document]
bennerv: I'm not sure if this is possible without a major refactor so might not make sense.  But these are unused (with the exception of the `AZURE_FP_CLIENT_ID` in the monitor.

bennerv (changes requested):
few nits & changes - good work so far!

pkg/deploy/generator/scripts/rpVMSS.sh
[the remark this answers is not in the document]
cadenmarchese: If the monitor component can't start without these variables but only one of them is actually used, it might be worth leaving a comment for cleanup when / if we get to the refactor.

pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -662,13 +678,16 @@ EOF
...
 
 for scan in baseline clamav software; do
-  /usr/local/bin/azsecd config -s $scan -d P1D
+	/usr/local/bin/azsecd config -s $scan -d P1D
 done
 
 echo "rebooting"
 restorecon -RF /var/log/*
-(sleep 30; reboot) &
+(
```
cadenmarchese, line 690: This is one example of a style change that I think we can leave out of the rpVMSS.sh script unless it's needed. We're creating differences between the gateway and RP scripts by only editing one, and not sure it's required for your changes.

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/network"
+)
+
+const (
+	DimSubnet              = "subnet"
+	DimVnet                = "vNet"
+	DimResourceGroup       = "resourcegroup"
+	DimSubscriptionID      = "subscriptionid"
+	DimSecurityGroup       = "networksecuritygroup"
```
cadenmarchese, line 28: I recommend this naming convention to keep it consistent with the following 5 consts:
```suggestion
	DimNSG       = "networksecuritygroup"
```

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,176 @@
...
+	DimNSGRuleName         = "rulename"
+	DimNSGRuleSources      = "sources"
+	DimNSGRuleDestinations = "destinations"
+	DimNSGRuleDirection    = "direction"
+	DimNSGRulePriority     = "priority"
+	DimClusterResourceID   = "clusterresourceid"
+	DimLocation            = "location"
+	DimTenantID            = "tenantid"
```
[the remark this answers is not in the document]
cadenmarchese: `location` is also a dimension we might not need.

pkg/monitor/azure/nsg/nsg.go
cadenmarchese: +1, we should store the `dims` in a separate map, and then the codes/metrics in the const.

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/metrics"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/network"
+)
+
+const (
+	DimSubnet              = "subnet"
+	DimVnet                = "vNet"
+	DimResourceGroup       = "resourcegroup"
```
cadenmarchese, line 26: Can we use NetworkResourceGroup or VnetResourceGroup instead of just ResourceGroup, to avoid confusion with the managed resource group?

pkg/monitor/worker.go
```diff
@@ -227,8 +232,36 @@ out:
...
+func (mon *monitor) newNSGMonitor(log *logrus.Entry, oc *api.OpenShiftCluster, subscriptionID, tenantID string, e metrics.Emitter, dims map[string]string, wg *sync.WaitGroup) Monitor {
+	token, err := mon.env.FPNewClientCertificateCredential(tenantID)
+	if err != nil {
+		log.Error("Unable to create FP Authorizer for NSG monitoring.", err)
+		mon.clusterm.EmitGauge(nsg.MetricUnsuccessfulFPCreation, int64(1), dims)
+		return noOpMonitor{}
+	}
+	client, err := armnetwork.NewSubnetsClient(subscriptionID, token, nil)
+	if err != nil {
+		log.Error("Unable to create the subnet client for NSG monitoring", err)
```
bennerv, line 256: do we want to emit a metric here as well?

pkg/monitor/worker.go
```diff
@@ -227,8 +232,36 @@ out:
...
+
+func (no noOpMonitor) Monitor(context.Context) []error {
+	return []error{}
+}
+
+func (mon *monitor) newNSGMonitor(log *logrus.Entry, oc *api.OpenShiftCluster, subscriptionID, tenantID string, e metrics.Emitter, dims map[string]string, wg *sync.WaitGroup) Monitor {
+	token, err := mon.env.FPNewClientCertificateCredential(tenantID)
+	if err != nil {
+		log.Error("Unable to create FP Authorizer for NSG monitoring.", err)
+		mon.clusterm.EmitGauge(nsg.MetricUnsuccessfulFPCreation, int64(1), dims)
```
bennerv, line 251: maybe make this metric more generic -> `failedNsgMonitorCreation` and use it below as well?  curious on your thoughts.

pkg/monitor/worker.go
```diff
@@ -255,5 +302,24 @@ func (mon *monitor) workOne(ctx context.Context, log *logrus.Entry, doc *api.Ope
 		return
 	}
 
-	c.Monitor(ctx)
+	monitors = append(monitors, c)
+	//c.Monitor(ctx)
```
bennerv, line 306: remove comment

pkg/monitor/worker.go
```diff
@@ -255,5 +302,24 @@ func (mon *monitor) workOne(ctx context.Context, log *logrus.Entry, doc *api.Ope
...
+	allJobsDone := make(chan bool)
+	go execute(ctx, allJobsDone, wg, monitors)
+
+	select {
+	case <-allJobsDone:
+	case <-ctx.Done():
+		log.Infof("NSG Monitoring processing for cluster %s has timed out", doc.OpenShiftCluster.ID)
```
bennerv, line 313: this could be related to the normal monitor failing as well, correct?

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,168 @@
...
+	emitter metrics.Emitter
+	oc      *api.OpenShiftCluster
+
+	wg *sync.WaitGroup
+
+	subnetClient *armnetwork.SubnetsClient
+	dims         map[string]string
+}
+
+func NewNSGMonitor(log *logrus.Entry, oc *api.OpenShiftCluster, subscriptionID string, subnetClient *armnetwork.SubnetsClient, emitter metrics.Emitter, wg *sync.WaitGroup) *NSGMonitor {
```
bennerv, line 47: We should force these to fail to compile if they do not implement the monitor interface. 

Please do the same for the cluster monitor.  

e.g.
```
var _ http.RoundTripper = (*tracerRoundTripper)(nil)
```

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+type subnetNSGConfig struct {
+	// subnet CIDR range
+	prefix []netip.Prefix
+	// The rules from the subnet NSG
+	nsg *armnetwork.SecurityGroup
+}
+
+func (n *NSGMonitor) toSubnetConfig(ctx context.Context, subnetID string) (subnetNSGConfig, error) {
```
bennerv, line 71: My biggest question(s): 
- Will we create a monitor which alerts SRE if it's improperly configured?  -> **Answer: NO**
    - If so, how will we construct the monitor?  
- How will we alert SRE should the nsgs be misconfigured if an IcM is opened to us?  -> I don't think we need to answer this today, I think ASI is the proper path forward, but need to get there as this is relevant info. 
    - I.e. how can we determine which geneva health signals are firing against a cluster?

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,168 @@
...
+		// FP has no access to the subnet
+		return []error{err}
+	}
+
+	// need this to get the right workerProfiles
+	workerProfiles, _ := api.GetEnrichedWorkerProfiles(n.oc.Properties)
+	workerSubnets := make([]subnetNSGConfig, 0, len(workerProfiles))
+	workerPrefixes := make([]netip.Prefix, 0, len(workerProfiles))
+	for _, wp := range workerProfiles {
+		s, err := n.toSubnetConfig(ctx, wp.SubnetID)
```
bennerv, line 121: I _think_ it's possible the subnetId can be empty should a customer misconfigure it.  We may want to have a check to skip should the subnet id be empty.  

We should test this to see if we can create a bad machineset with an invalid subnet and see if the enriched worker profiles has an empty subnet id.

bennerv: empty subnets should be ignored as the machinesets are invalid.

bennerv (dismissed):

pkg/monitor/azure/nsg/nsg.go
bennerv: Maybe just create a story for the second item.  No real action here.

nwnt: ARO-4502 has been created.

pkg/monitor/azure/nsg/nsg.go
nwnt: Confirmed that the subnetID can be empty

```
            {
                "name": "created-with-an-invalid-subnet",
                "count": 1
            },
            {
                "name": "machine-set-name",
                "vmSize": "Standard_D4s_v3",
                "diskSizeGB": 128,
                "subnetId": "subnet-value",
                "count": 1,
                "encryptionAtHost": "Disabled"
            },
```

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,183 @@
...
+	DimNSGRuleSources      = "sources"
+	DimNSGRuleDestinations = "destinations"
+	DimNSGRuleDirection    = "direction"
+	DimNSGRulePriority     = "priority"
+	DimClusterResourceID   = "clusterresourceid"
+	DimLocation            = "location"
+	DimTenantID            = "tenantid"
+
+	MetricInvalidDenyRule         = "preconfigurednsg.invalid.denyrule"
+	MetricSubnetAcessResponseCode = "preconfiguredNSG.subnetaccess.responsecode"
```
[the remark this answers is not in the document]
nwnt: done

pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -29,8 +29,8 @@ rpm --import https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-8
 rpm --import https://packages.microsoft.com/keys/microsoft.asc
 
 for attempt in {1..5}; do
-  yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && break
-  if [[ ${attempt} -lt 5 ]]; then sleep 10; else exit 1; fi
+	yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && break
```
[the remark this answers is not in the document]
nwnt: This was unintentional.  The linters I have somehow added these things in.  I have reverted the unintentional changes from the commits.

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -0,0 +1,176 @@
...
+			n.done <- err
+			return
+		}
+		workerSubnets = append(workerSubnets, s)
+		workerPrefixes = append(workerPrefixes, s.prefix)
+	}
+
+	// to make sure each NSG is processed only once
+	nsgSet := map[string]*mgmtnetwork.SecurityGroup{
+		*masterSubnet.nsg.ID: masterSubnet.nsg,
```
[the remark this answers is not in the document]
nwnt: Thanks Kipp.  I have added the guards.

pkg/monitor/azure/nsg/nsg.go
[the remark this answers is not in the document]
nwnt: Removed `tenantID`.  However, we need `location` as it's required for the RHC.

pkg/deploy/generator/scripts/rpVMSS.sh
nwnt: This was unintentionally done by one of my linters and I have removed them.

pkg/monitor/azure/nsg/nsg.go
nwnt: done

pkg/monitor/azure/nsg/nsg.go
nwnt: done

pkg/monitor/worker.go
nwnt: done

pkg/monitor/worker.go
nwnt: made this into a general message.

pkg/monitor/azure/nsg/nsg.go
nwnt: done

pkg/monitor/azure/nsg/nsg.go
nwnt: add the logic to skip when a blank subnet is found.

kimorris27 (approved):

kimorris27 (approved):

cadenmarchese (approved):
lgtm, thanks for addressing the comments!

kimorris27 (approved):

pkg/monitor/worker.go
```diff
@@ -255,5 +298,23 @@ func (mon *monitor) workOne(ctx context.Context, log *logrus.Entry, doc *api.Ope
 		return
 	}
 
-	c.Monitor(ctx)
+	monitors = append(monitors, c)
+	allJobsDone := make(chan bool)
+	go execute(ctx, allJobsDone, &wg, monitors)
```
petrkotas, line 303: So we essentially introduce concurent mechanism to run multiple monitors.
We might as well migrate the rest to use this pattern, because sequential run we have in the original code is error prone.

petrkotas (approved):
This PR is a good work. I like the fact that it opens the door to make the other monitors run concurently. 

I am tentative  approving this one. Good work and I do not see anything that would break exiting code.

petrkotas (approved):
This PR is a good work. I like the fact that it opens the door to make the other monitors run concurently. 

I am tentative  approving this one. Good work and I do not see anything that would break exiting code.

PR #3157, opened 2023-09-07
Multi ip per load balancer update code followup 

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-5251
Follow up for PR #3095

- Change dynamic validation to use FP SP
- Refactors IP Creation/Deletion and adds concurrency to create/update, IPs are now deleted and created in parallel which speeds up the update process. 

### Test plan for issue:
E2E tests
Existing unit tests

No

### DIFF
--- a/pkg/cluster/loadbalancerprofile.go
+++ b/pkg/cluster/loadbalancerprofile.go
@@ -179,6 +179,11 @@ func setOutboundRuleV4(lb mgmtnetwork.LoadBalancer, outboundRuleV4FrontendIPConf
 	}
 }
 
+type deleteResult struct {
+	name string
+	err  error
+}
+
 // Delete all managed outbound IPs that are not in use by the load balancer.
 // The default outbound ip is saved if the api server is public.
 func (m *manager) deleteUnusedManagedIPs(ctx context.Context) error {
@@ -188,20 +193,37 @@ func (m *manager) deleteUnusedManagedIPs(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	ch := make(chan deleteResult)
+	defer close(ch)
 	var cleanupErrors []string
+
 	for _, id := range ipsToDelete {
 		ipName := stringutils.LastTokenByte(id, '/')
-		m.log.Infof("deleting managed public IP Address: %s", ipName)
-		err := m.publicIPAddresses.DeleteAndWait(ctx, resourceGroupName, ipName)
-		if err != nil {
-			cleanupErrors = append(cleanupErrors, fmt.Sprintf("deletion of unused managed ip %s failed with error: %v", ipName, err))
+		go m.deleteIPAddress(ctx, resourceGroupName, ipName, ch)
+	}
+
+	for range ipsToDelete {
+		result := <-ch
+		if result.err != nil {
+			cleanupErrors = append(cleanupErrors, fmt.Sprintf("deletion of unused managed ip %s failed with error: %v", result.name, result.err))
 		}
 	}
+
 	if cleanupErrors != nil {
 		return fmt.Errorf("failed to cleanup unused managed ips\n%s", strings.Join(cleanupErrors, "\n"))
 	}
 
 	return nil
+}
+
+func (m *manager) deleteIPAddress(ctx context.Context, resourceGroupName string, ipName string, ch chan<- deleteResult) {
+	m.log.Infof("deleting managed public IP Address: %s", ipName)
+	err := m.publicIPAddresses.DeleteAndWait(ctx, resourceGroupName, ipName)
+	ch <- deleteResult{
+		name: ipName,
+		err:  err,
+	}
 }
 
 func (m *manager) getManagedIPsToDelete(ctx context.Context) ([]string, error) {
--- a/pkg/cluster/loadbalancerprofile.go
+++ b/pkg/cluster/loadbalancerprofile.go
@@ -194,7 +194,7 @@ func setOutboundRuleV4(lb mgmtnetwork.LoadBalancer, outboundRuleV4FrontendIPConf
 func (m *manager) deleteUnusedManagedIPs(ctx context.Context) error {
 	resourceGroupName := stringutils.LastTokenByte(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
 
-	ipsToDelete, err := m.getManagedIPsToDelete(ctx)
+	unusedManagedIPs, err := m.getUnusedManagedIPs(ctx)
 	if err != nil {
 		return err
 	}
@@ -203,12 +203,12 @@ func (m *manager) deleteUnusedManagedIPs(ctx context.Context) error {
 	defer close(ch)
 	var cleanupErrors []string
 
-	for _, id := range ipsToDelete {
+	for _, id := range unusedManagedIPs {
 		ipName := stringutils.LastTokenByte(id, '/')
 		go m.deleteIPAddress(ctx, resourceGroupName, ipName, ch)
 	}
 
-	for range ipsToDelete {
+	for range unusedManagedIPs {
 		result := <-ch
 		if result.err != nil {
 			cleanupErrors = append(cleanupErrors, fmt.Sprintf("deletion of unused managed ip %s failed with error: %v", result.name, result.err))
@@ -231,7 +231,7 @@ func (m *manager) deleteIPAddress(ctx context.Context, resourceGroupName string,
 	}
 }
 
-func (m *manager) getManagedIPsToDelete(ctx context.Context) ([]string, error) {
+func (m *manager) getUnusedManagedIPs(ctx context.Context) ([]string, error) {
 	resourceGroupName := stringutils.LastTokenByte(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
 	infraID := m.doc.OpenShiftCluster.Properties.InfraID
 
@@ -250,17 +250,17 @@ func (m *manager) getManagedIPsToDelete(ctx context.Context) ([]string, error) {
 	for i := 0; i < len(outboundIPs); i++ {
 		outboundIPMap[strings.ToLower(outboundIPs[i].ID)] = outboundIPs[i]
 	}
-	var ipsToDelete []string
+	var unusedManagedIPs []string
 	for _, ip := range managedIPs {
 		// don't delete api server ip
 		if *ip.Name == infraID+"-pip-v4" && m.doc.OpenShiftCluster.Properties.APIServerProfile.Visibility == api.VisibilityPublic {
 			continue
 		}
 		if _, ok := outboundIPMap[strings.ToLower(*ip.ID)]; !ok && strings.Contains(strings.ToLower(*ip.ID), strings.ToLower(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID)) {
-			ipsToDelete = append(ipsToDelete, *ip.ID)
+			unusedManagedIPs = append(unusedManagedIPs, *ip.ID)
 		}
 	}
-	return ipsToDelete, nil
+	return unusedManagedIPs, nil
 }
 
 // Returns the desired RP managed outbound publicIPAddresses.  Additional Managed Outbound IPs
--- a/pkg/cluster/loadbalancerprofile.go
+++ b/pkg/cluster/loadbalancerprofile.go
@@ -19,6 +19,16 @@ import (
 
 const outboundRuleV4 = "outbound-rule-v4"
 
+type deleteIPResult struct {
+	name string
+	err  error
+}
+
+type createIPResult struct {
+	ip  mgmtnetwork.PublicIPAddress
+	err error
+}
+
 func (m *manager) reconcileLoadBalancerProfile(ctx context.Context) error {
 	if m.doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType != api.OutboundTypeLoadbalancer || m.doc.OpenShiftCluster.Properties.ArchitectureVersion == api.ArchitectureVersionV1 {
 		return nil
@@ -179,11 +189,6 @@ func setOutboundRuleV4(lb mgmtnetwork.LoadBalancer, outboundRuleV4FrontendIPConf
 	}
 }
 
-type deleteResult struct {
-	name string
-	err  error
-}
-
 // Delete all managed outbound IPs that are not in use by the load balancer.
 // The default outbound ip is saved if the api server is public.
 func (m *manager) deleteUnusedManagedIPs(ctx context.Context) error {
@@ -194,7 +199,7 @@ func (m *manager) deleteUnusedManagedIPs(ctx context.Context) error {
 		return err
 	}
 
-	ch := make(chan deleteResult)
+	ch := make(chan deleteIPResult)
 	defer close(ch)
 	var cleanupErrors []string
 
@@ -217,10 +222,10 @@ func (m *manager) deleteUnusedManagedIPs(ctx context.Context) error {
 	return nil
 }
 
-func (m *manager) deleteIPAddress(ctx context.Context, resourceGroupName string, ipName string, ch chan<- deleteResult) {
+func (m *manager) deleteIPAddress(ctx context.Context, resourceGroupName string, ipName string, ch chan<- deleteIPResult) {
 	m.log.Infof("deleting managed public IP Address: %s", ipName)
 	err := m.publicIPAddresses.DeleteAndWait(ctx, resourceGroupName, ipName)
-	ch <- deleteResult{
+	ch <- deleteIPResult{
 		name: ipName,
 		err:  err,
 	}
@@ -267,11 +272,6 @@ func (m *manager) reconcileOutboundIPs(ctx context.Context) ([]api.ResourceRefer
 		return m.reconcileDesiredManagedIPs(ctx)
 	}
 	return nil, nil
-}
-
-type createIPResult struct {
-	ip  mgmtnetwork.PublicIPAddress
-	err error
 }
 
 // Returns RP managed outbound ips to be added to the outbound rule.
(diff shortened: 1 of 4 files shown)


pkg/cluster/loadbalancerprofile.go
```diff
@@ -20,7 +20,7 @@ import (
 const outboundRuleV4 = "outbound-rule-v4"
```
s-amann, line 21: with this file being kinda massive, I wonder if we should have a `loadbalancerprofile` package instead and have different files for creation, deletion, and reconciliation code. This would also benefit the tests, as we could separate them out based on their purpose. wdyt?

pkg/cluster/loadbalancerprofile.go
```diff
@@ -95,9 +93,34 @@ func (m *manager) reconcileOutboundRuleV4IPsInner(ctx context.Context, lb mgmtne
 	return nil
 }
 
-// Remove all frontend ip config in use by outbound-rule-v4.  Frontend IP config that is used by load balancer rules will be saved.
+func replaceOutboundRuleV4OutboundIPs(resourceGroupID string, lb mgmtnetwork.LoadBalancer, desiredOutboundIPs []api.ResourceReference) {
```
s-amann, line 96: unless we plan on testing this function separately from `reconcileLoadBalancerProfile`, I don't see the need to split these two calls into their own func.

The same goes for `removeOutboundIPsFromLB`, since all the contained functions are acting on side effects, IMO it makes more sense to have them be called directly by the parent and not nest them two levels.

s-amann: so this would mean line 80 inside `reconcileLoadBalancerProfile` calls the now three functions in a row:
```
removeOutboundRuleV4FrontendIPConfig(lb)

setOutboundRuleV4(lb, []mgmtnetwork.SubResource{})

addOutboundIPsToLB(resourceGroupID, lb, desiredOutboundIPs)
```

s-amann (commented):
this lgtm for the most part, I think `loadbalancerprofile` is getting too large (almost 500 loc), some suggestions below.

pkg/cluster/loadbalancerprofile.go
tony-schndr: I split this up to maybe make it more readable, but I can agree on that its not really needed.

pkg/cluster/loadbalancerprofile.go
tony-schndr: So I am piggy backing off the manager for the required cluster doc and azure clients for much of this code.  I am not sure how I would go about pulling any of the manager methods out of the cluster package.

There are plenty of functions that don't need the manager, I could pull those out into pkg/util.

Thoughts?

tony-schndr:
The only change here that really needs to go out in the next release is the check for Architecture v1 clusters.  I can break this up into two PRs if that would be better.

pkg/cluster/loadbalancerprofile.go
s-amann: Pulling out the non manager functions would be a good start in helping to narrow the scope of this file. I like that idea.

tony-schndr:
Opened new PR just for skipping Arch V1 checks https://github.com/Azure/ARO-RP/pull/3162

pkg/util/loadbalancer/loadbalancer.go
```diff
@@ -8,9 +8,16 @@ import (
...
 
 	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/util/stringutils"
 )
 
 func RemoveFrontendIPConfiguration(lb *mgmtnetwork.LoadBalancer, resourceID string) error {
+	if lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations == nil {
+		return fmt.Errorf("FrontendIPConfigurations in nil")
```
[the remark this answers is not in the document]
tony-schndr: Thanks for catching this, fixed.

pkg/cluster/loadbalancerprofile.go
tony-schndr: I just wrapped up the refactor to move non-manager functions out of this file.  I also refactored testing a bit.  If you have a moment let me know what you think.  Thanks!

pkg/util/loadbalancer/loadbalancer.go
```diff
@@ -28,3 +37,124 @@ func RemoveFrontendIPConfiguration(lb *mgmtnetwork.LoadBalancer, resourceID stri
...
+func getOutboundRuleV4FIPConfigs(lb mgmtnetwork.LoadBalancer) map[string]mgmtnetwork.SubResource {
+	var obRuleV4FIPConfigs = make(map[string]mgmtnetwork.SubResource)
+	for _, obRule := range *lb.LoadBalancerPropertiesFormat.OutboundRules {
+		if *obRule.Name == OutboundRuleV4 {
+			for i := 0; i < len(*obRule.OutboundRulePropertiesFormat.FrontendIPConfigurations); i++ {
+				fipConfigID := *(*obRule.OutboundRulePropertiesFormat.FrontendIPConfigurations)[i].ID
+				fipConfig := (*obRule.OutboundRulePropertiesFormat.FrontendIPConfigurations)[i]
+				obRuleV4FIPConfigs[fipConfigID] = fipConfig
+			}
+			break
```
SudoBrendan, line 72: There's a lot of refactors at once, so I want to be sure we're not breaking things...

In the original code, /pkg/cluster/loadbalancerprofile.go L110, we had:

```
            // clear outbound-rule-v4 frontend ip config
			*obRule.FrontendIPConfigurations = []mgmtnetwork.SubResource{}
            break
```

why was this removed here?

SudoBrendan:
@tony-schndr I really appreciate the follow-up here, there are some very exciting changes, but I worry about the size of this PR, and the fact we are changing tests _and_ code - it makes it hard for me to determine if the tests still pass because we have the same functionality, or if the tests still pass because they have been changed to accommodate changes in the code. Would you mind splitting this up into two PRs?

- PR 1 would include all refactors to tests, which pass given the current implementation
- PR 2 would include all code changes, which pass given the refactored tests

is this possible to do?

SudoBrendan (commented):
I'd prefer to have this PR split into test/implementation refactors, but I also had one comment on the implementation I don't want lost.

pkg/util/loadbalancer/loadbalancer.go
tony-schndr: In multiple locations throughout this file, any time I needed to update the outbound rule frontendIPConfiguration I was iterating on outbound rules, matching on "outbound-rule-v4", and then setting the frontendIPConfiguration.  I created a setter function to do this now 
https://github.com/Azure/ARO-RP/pull/3157/files#diff-d87786de7031b1f89a1bce960b3176383b16135b6d8cf887898443fc348f2c21R173

pkg/cluster/loadbalancerprofile.go
```diff
@@ -224,29 +269,38 @@ func (m *manager) getDesiredOutboundIPs(ctx context.Context) ([]api.ResourceRefe
 	return nil, nil
 }
 
+type createIPResult struct {
```
mociarain, line 272: Do we have any convention around grouping type definitions at the top of a file or do we just define them before they are used like this as the norm?

tony-schndr: I agree that these should be at the top, I'll move them.

pkg/cluster/loadbalancerprofile.go
tony-schndr: I broke this PR up and the new home for these functions is in `pkg/util/loadbalancer`, see https://github.com/Azure/ARO-RP/pull/3379

s-fairchild (dismissed):

cadenmarchese (commented):
I'm still catching up on the logic that was there before, but I have a few questions and am wondering if we can simplify a few things and still keep the concurrency.

pkg/validate/openshiftcluster_validatedynamic.go
```diff
@@ -249,5 +244,10 @@ func (dv *openShiftClusterDynamicValidator) Dynamic(ctx context.Context) error {
 		return err
 	}
 
+	err = fpDynamic.ValidateLoadBalancerProfile(ctx, dv.oc)
```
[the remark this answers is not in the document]
SudoBrendan: We're updating to use the FPSP in validation of the resources, since that is the identity that will be used to modify them later.

SudoBrendan (dismissed):
LGTM, I trust the unit tests, and FPSP is the correct choice on dynamic validation, good work.

pkg/cluster/loadbalancerprofile.go
```diff
@@ -174,48 +193,79 @@ func addOutboundIPsToLB(resourceGroupID string, lb mgmtnetwork.LoadBalancer, obI
...
-	}
-
-	return nil
+	return ipsToDelete, nil
 }
 
 // Returns the desired RP managed outbound publicIPAddresses.  Additional Managed Outbound IPs
 // will be created as required to satisfy ManagedOutboundIP.Count.
-func (m *manager) getDesiredOutboundIPs(ctx context.Context) ([]api.ResourceReference, error) {
+func (m *manager) reconcileOutboundIPs(ctx context.Context) ([]api.ResourceReference, error) {
```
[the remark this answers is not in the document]
tony-schndr: outboundIPs are customer provided, they won't be used until phase 2 of the project.  Managed IPs are RP provided.

pkg/cluster/loadbalancerprofile.go
```diff
@@ -174,48 +193,79 @@ func addOutboundIPsToLB(resourceGroupID string, lb mgmtnetwork.LoadBalancer, obI
 // The default outbound ip is saved if the api server is public.
 func (m *manager) deleteUnusedManagedIPs(ctx context.Context) error {
 	resourceGroupName := stringutils.LastTokenByte(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
+
+	ipsToDelete, err := m.getManagedIPsToDelete(ctx)
```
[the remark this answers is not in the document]
tony-schndr: I like that name change, I'll update this.

My reason for the additional function was to try and keep the functions at a reasonable length.  If I combined these functions it would result in about 60 lines of code.  Instead I opted to keep the logic separate and used a function name to convey the logic to the reader.

pkg/cluster/loadbalancerprofile.go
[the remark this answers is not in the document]
tony-schndr: >Should we do away with this function? It seems like it only exists to execute and pass along any errors from reconcileDesiredManagedIPs().

When I originally wrote this I was thinking about the additional phases.  If/when we get to phase 2 then this function could return outboundIPs, outboundIPPrefixes, or managedOutboundIPs.

pkg/cluster/loadbalancerprofile.go
```diff
@@ -230,23 +280,27 @@ func (m *manager) getDesiredOutboundIPs(ctx context.Context) ([]api.ResourceRefe
...
-	for i := 0; i < numToCreate; i++ {
-		ipAddress, err := m.createPublicIPAddress(ctx)
+
+	if numToCreate > 0 {
+		err = m.createPublicIPAddresses(ctx, ipAddresses, numToCreate)
 		if err != nil {
 			return nil, err
 		}
-		ipAddresses[*ipAddress.Name] = ipAddress
 	}
```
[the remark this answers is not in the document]
tony-schndr: I was considering the edge case where IPs could get created and the loadbalancer fails to be updated resulting in dangling IP addresses.  That is why the cleanup is at the end of the process.  There is also the potential for phase 2 in the future where managed IP addresses could be replaced by customer IP addresses and would need to be cleaned up outside of the creation process.

cadenmarchese (dismissed):
Thanks for helping me with questions. https://github.com/Azure/ARO-RP/pull/3157#discussion_r1504763612 is still open but is not blocking approval. lgtm!

jaitaiwan (approved):
LGTM, just a small thank you as a response to someone else's comment

pkg/cluster/loadbalancerprofile.go
```diff
@@ -263,8 +317,31 @@ func (m *manager) reconcileDesiredManagedIPs(ctx context.Context) ([]api.Resourc
...
+}
 
-	return desiredIPAddresses, nil
+func (m *manager) createPublicIPAddresses(ctx context.Context, ipAddresses map[string]mgmtnetwork.PublicIPAddress, numToCreate int) error {
+	ch := make(chan createIPResult)
+	defer close(ch)
+	var errResults []string
+	// create additional IPs if needed
+	for i := 0; i < numToCreate; i++ {
+		go m.createPublicIPAddress(ctx, ch)
```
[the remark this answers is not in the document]
AldoFusterTurpin: More precisely, _pkg/cluster/loadbalancerprofile_test.go:114_ is what is modifying the global variable and the fake generator is not concurrent safe. Even this test line was not introduced in this PR, we are now using it concurrently.

The quick fix is to make fake generator concurrent safe by using a mutex:

```go
package fake

import (
	"sync"

)

type fakeGenerator struct {
	words      []string
	currentPos int
	mu         *sync.Mutex
}

func NewGenerator(predefinedWords []string) uuid.Generator {
	return &fakeGenerator{
		words: predefinedWords,
		mu:    &sync.Mutex{},
	}
}

func (f *fakeGenerator) movePos() {
	f.currentPos++
}

func (f *fakeGenerator) Generate() string {
	f.mu.Lock()
	defer f.mu.Unlock()

	defer f.movePos()
	if len(f.words) < f.currentPos {
		return ""
	}
	return f.words[f.currentPos]
}

```

AldoFusterTurpin (changes requested):
Possible merge blocker. There is a race condition I have detected by doing
$ go test -race ./pkg/cluster
which is not present in master. 

It is introduced by an already existing unit test that modifies the global uuid.Default generator assigning it the fake one, which is not concurrent safe. I have added the details and a quick fix in the review. 🙂

pkg/cluster/loadbalancerprofile.go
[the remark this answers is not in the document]
mociarain: >I checked locally the PR and executed
> `go test -race ./pkg/cluster` 

Should we be doing this as part of our normal test suite?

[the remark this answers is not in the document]
tony-schndr: > Should we be doing this as part of our normal test suite? 
That is my question too, I was unaware of this check so it may be best to add this to CI.

kimorris27 (approved):

pkg/cluster/loadbalancerprofile.go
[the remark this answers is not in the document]
AldoFusterTurpin: Good question 🙂 
I mentioned in the past that we have some race conditions in some test helper functions and utilities. Adding the -race flag is something that should be done when working with concurrent code, but we can not include it as a default step in the CI as it will fail with those "expected" race conditions we have in the repo. We would ideally remove the race conditions and later make that -race option a default. 
This is the [PR](https://github.com/Azure/ARO-RP/pull/2466) where I fixed some, but I was not able to fix all of them in a reasonable time.

pkg/cluster/loadbalancerprofile.go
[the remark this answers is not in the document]
AldoFusterTurpin: My humble and personal opinion is that the separate function helps a lot to understand the code. I am a strong advocate of small functions. 🙂

pkg/cluster/loadbalancerprofile.go
```diff
@@ -263,8 +317,31 @@ func (m *manager) reconcileDesiredManagedIPs(ctx context.Context) ([]api.Resourc
...
+	}
+
+	for i := 0; i < numToCreate; i++ {
+		result := <-ch
+		if result.err != nil {
+			errResults = append(errResults, fmt.Sprintf("creation of ip address %s failed with error: %s", *result.ip.Name, result.err.Error()))
+		} else {
+			ipAddresses[*result.ip.Name] = result.ip
```
AldoFusterTurpin, line 337: I have to admit that the fact that this function is modifying the ipAddresses map at the same time that it checks for the result of the ch channel is a bit unexpected/confusing for me. 

I think we could make "createPublicIPAddress" write into two different channels, the resultsChannels where we just write the ones that did not fail and an error channel that just receives and expects errors in there. In that way, we do not need to check with an if in "createPublicIPAddresses" if the operation succeded for the result of the channel as it is not easy to follow. We could simply iterate over the two channels, having a good separation of concerns. I think that would simplify the logic and would be more idiomatic and maintainable. I have worked on something that is working (happy the tests passed after some fight) and I would be happy to talk about it if you are interested in 🙂

pkg/cluster/loadbalancerprofile.go
```diff
@@ -263,8 +317,31 @@ func (m *manager) reconcileDesiredManagedIPs(ctx context.Context) ([]api.Resourc
...
+		}
+	}
+
+	if len(errResults) > 0 {
```
AldoFusterTurpin, line 341: I think we can check directly if errResult != nil, a bit more idiomatic and shorter 🙂

AldoFusterTurpin (commented):
Thank you for this. 🙂 
I have not seen anything wrong blocking the PR, but I have added some minor refactor comments (some are clearly optional) and a possible suggestion to the concurrent code that could be more idiomatic and easy to follow, something we could comment privately if you wish.
Thank you!

pkg/cluster/loadbalancerprofile.go
AldoFusterTurpin: Update: we already discussed it privately. Nice chat.

tony-schndr:
Closed to cutover to a branch in #3508

PR #3159, opened 2023-09-08
Enable autonodezing by default

petrkotas:

Fixes https://issues.redhat.com/browse/ARO-2545

Enables autonode sizing by default

### Test plan for issue:
option is already tested several times in production without issues.


bennerv (approved):

bennerv (commented):
There's a few tests which rely on operator flags set to specific values. Can you go clean or swap that up.?

Shivkumar13 (approved):

bennerv (approved):

PR #3160, opened 2023-09-08
Remove rhcos publishing doc, can be found in ARO-Installer

SudoBrendan:

Removes a duplicated doc, which currently lives in https://github.com/openshift/ARO-Installer/blob/main/docs/publish-rhcos-image.md after all the rhcos mirroring things have moved there.


cblecker (approved):

bennerv (approved):
docs dont' rely on e2e, approving and merging.

PR #3162, opened 2023-09-11
skip lb profile reconcile when arch is v1

tony-schndr:

Skips reconciling when architecture is version 1 as this feature is only needed/tested against v2.
### Test plan for issue:

Unit Tests


SudoBrendan (approved):
Thanks for pulling out this change for easier review, LGTM!

petrkotas (approved):
LGTM thanks

PR #3166, opened 2023-09-13
Test removal of context on eventually

bennerv:
This should hopefully time out at 5m on the logs test.


bennerv:
Don't think this is doing anything.  The tests are failing immediately now not waiting at all.

PR #3169, opened 2023-09-14
fix: add priv user check for machine config & master toleration policies

ArrisLee:
Add  priv user check for existing policies:

- machine config deny policy
- master toleration taint deny policy


yjst2012 (approved):
LGTM

PR #3170, opened 2023-09-14
Allow metricalerts, actiongroups, and activitylogalerts for monitoring in the denyassignment

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-4241

Allows setting ActionGroups (which contain what happens when alerts happen, like sending an email) and MetricAlerts/ActivityLogAlerts on the protected resource group. This allows some alerting on the Azure-level VM resources, while we don't have them all flowing through the OpenShiftCluster object

### Test plan for issue:
Not sure how to test denyassignments without going to prod :thinking: 

N/A?


SudoBrendan (commented):
The code changes seem simple enough, but after reading the cards and the comments here, I'm not sure why they're necessary. Can we update the Jira to include acceptance criteria so we understand why this change is needed and what it is attempting to solve?

hawkowl:
@SudoBrendan I've updated the ticket. Basically, this allows customers to alert on the Azure Monitor metrics they can see in the locked resource group, without having to set up Prometheus forwarding (which is reliant on prometheus)

jaitaiwan (approved):
LGTM

PR #3171, opened 2023-09-14
Default to 4.12 for new installations

tiguelu:
https://issues.redhat.com/browse/ARO-2720

Make 4.12 the default install stream in ARO

### DIFF
--- a/pkg/util/version/const.go
+++ b/pkg/util/version/const.go
@@ -31,7 +31,7 @@ type Stream struct {
 }
 
 // DefaultMinorVersion describes the minor OpenShift version to default to
-var DefaultMinorVersion = 11
+var DefaultMinorVersion = 12
 
 // DefaultInstallStreams describes the latest version of our supported streams
 var DefaultInstallStreams = map[int]*Stream{


tiguelu:
For the record, [RHELPLAN-167394](https://issues.redhat.com/browse/RHELPLAN-167394) is still open for 4.12. 

Might not be a blocker for making 4.12 the default version. There is a resolution path if any cluster was found to be impacted by upgrading to 4.13.11. Also, a temporary workaround is to reboot nodes, which buys time until the next leak happens.

tiguelu:
Adding hold label until we get confirmation from BU that it is ok to proceed with this change despite [RHELPLAN-167394](https://issues.redhat.com/browse/RHELPLAN-167394)

tiguelu:
> Adding `hold` label until we get confirmation from BU that it is ok to proceed with this change despite [RHELPLAN-167394](https://issues.redhat.com/browse/RHELPLAN-167394)

The change has been approved to go ahead. Removing `hold`.

UlrichSchlueter (approved):

PR #3177, opened 2023-09-15
Propagate errors of ARO MachineHealthCheckController to ARO Operator

safwank97:

Fixes - https://issues.redhat.com/browse/ARO-3256

Aggregates errors from the MachineHealthCheck controller and propagates to the ARO Operator.

### Test plan for issue:

As of now I think it must pass the local test cases after spinning up a test cluster and if case scenarios falls under the test conditions defined in code then error propagation will happen.

None that I'm aware of.

### DIFF
--- a/cmd/aro/operator.go
+++ b/cmd/aro/operator.go
@@ -192,9 +192,9 @@ func operator(ctx context.Context, log *logrus.Entry) error {
 			return fmt.Errorf("unable to create controller %s: %v", autosizednodes.ControllerName, err)
 		}
 		if err = (machinehealthcheck.NewMachineHealthCheckReconciler(
-			log.WithField("controller", machinehealthcheck.MHCControllerName),
+			log.WithField("controller", machinehealthcheck.ControllerName),
 			client, dh)).SetupWithManager(mgr); err != nil {
-			return fmt.Errorf("unable to create controller %s: %v", machinehealthcheck.MHCControllerName, err)
+			return fmt.Errorf("unable to create controller %s: %v", machinehealthcheck.ControllerName, err)
 		}
 		if err = (ingress.NewReconciler(
 			log.WithField("controller", ingress.ControllerName),
--- a/cmd/aro/operator.go
+++ b/cmd/aro/operator.go
@@ -191,7 +191,7 @@ func operator(ctx context.Context, log *logrus.Entry) error {
 			client)).SetupWithManager(mgr); err != nil {
 			return fmt.Errorf("unable to create controller %s: %v", autosizednodes.ControllerName, err)
 		}
-		if err = (machinehealthcheck.NewMachineHealthCheckReconciler(
+		if err = (machinehealthcheck.NewReconciler(
 			log.WithField("controller", machinehealthcheck.ControllerName),
 			client, dh)).SetupWithManager(mgr); err != nil {
 			return fmt.Errorf("unable to create controller %s: %v", machinehealthcheck.ControllerName, err)
--- a/cmd/aro/operator.go
+++ b/cmd/aro/operator.go
@@ -191,10 +191,10 @@ func operator(ctx context.Context, log *logrus.Entry) error {
 			client)).SetupWithManager(mgr); err != nil {
 			return fmt.Errorf("unable to create controller %s: %v", autosizednodes.ControllerName, err)
 		}
-		if err = (machinehealthcheck.NewReconciler(
-			log.WithField("controller", machinehealthcheck.ControllerName),
+		if err = (machinehealthcheck.NewMachineHealthCheckReconciler(
+			log.WithField("controller", machinehealthcheck.MHCControllerName),
 			client, dh)).SetupWithManager(mgr); err != nil {
-			return fmt.Errorf("unable to create controller %s: %v", machinehealthcheck.ControllerName, err)
+			return fmt.Errorf("unable to create controller %s: %v", machinehealthcheck.MHCControllerName, err)
 		}
 		if err = (ingress.NewReconciler(
 			log.WithField("controller", ingress.ControllerName),
--- a/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
@@ -70,12 +70,13 @@ func (r *MachineHealthCheckReconciler) Reconcile(ctx context.Context, request ct
 
 	r.Log.Debug("running")
 	if !instance.Spec.OperatorFlags.GetSimpleBoolean(MHCManaged) {
-		r.SetProgressing(ctx, "We are disabling the MHCHealthCheck because this cluster at current moment is NOT MHC Managed.")
+		r.SetProgressing(ctx, "MHC and it's alerts are disabled because this cluster is not MHC managed.")
 
 		err := r.dh.EnsureDeleted(ctx, "MachineHealthCheck", "openshift-machine-api", "aro-machinehealthcheck")
 		if err != nil {
 			r.Log.Error(err)
 			r.SetDegraded(ctx, err)
+			r.ClearProgressing(ctx)
 
 			return reconcile.Result{RequeueAfter: time.Hour}, err
 		}
@@ -84,6 +85,7 @@ func (r *MachineHealthCheckReconciler) Reconcile(ctx context.Context, request ct
 		if err != nil {
 			r.Log.Error(err)
 			r.SetDegraded(ctx, err)
+			r.ClearProgressing(ctx)
 
 			return reconcile.Result{RequeueAfter: time.Hour}, err
 		}
--- a/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
@@ -37,14 +37,14 @@ const (
 	enabled        string = "aro.machinehealthcheck.enabled"
 )
 
-type MachineHealthCheckReconciler struct {
+type Reconciler struct {
 	base.AROController
 
 	dh dynamichelper.Interface
 }
 
-func NewMachineHealthCheckReconciler(log *logrus.Entry, client client.Client, dh dynamichelper.Interface) *MachineHealthCheckReconciler {
-	return &MachineHealthCheckReconciler{
+func NewReconciler(log *logrus.Entry, client client.Client, dh dynamichelper.Interface) *Reconciler {
+	return &Reconciler{
 		AROController: base.AROController{
 			Log:    log,
 			Client: client,
@@ -56,7 +56,7 @@ func NewMachineHealthCheckReconciler(log *logrus.Entry, client client.Client, dh
 
 // Reconcile watches MachineHealthCheck objects, and if any changes,
 // reconciles the associated ARO MachineHealthCheck object
-func (r *MachineHealthCheckReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
 	instance, err := r.GetCluster(ctx)
 
 	if err != nil {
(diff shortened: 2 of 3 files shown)


pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -32,50 +32,63 @@ var machinehealthcheckYaml []byte
...
+		dh: dh,
 	}
 }
 
-func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
-	instance := &arov1alpha1.Cluster{}
-	err := r.client.Get(ctx, types.NamespacedName{Name: arov1alpha1.SingletonClusterName}, instance)
+// Reconcile watches MachineHealthCheck objects, and if any changes,
+// reconciles the associated ARO MachineHealthCheck object
+
```
SudoBrendan, line 59: I'm not 100% sure, but I think this extra newline might mess up godoc on this function.

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -32,50 +32,63 @@ var machinehealthcheckYaml []byte
...
 		}
 
 		err = r.dh.EnsureDeleted(ctx, "PrometheusRule", "openshift-machine-api", "mhc-remediation-alert")
 		if err != nil {
+			r.Log.Error(err)
+			r.SetDegraded(ctx, err)
+
 			return reconcile.Result{RequeueAfter: time.Hour}, err
 		}
+
```
SudoBrendan, line 91: bug: needs a `r.ClearConditions` to clear `Progressing` status here.

SudoBrendan (dismissed):
Looks good so far, thanks for your work on this. One bug to fix. The last thing I think is to add some E2E tests for validating the conditions of the operator during each test execution.

As a final super-nit, there are quite a few newlines added that I don't think add value (at least 6). It's likely best to delete those to reduce the diff.

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
safwank97: So should  I just remove the blank lines or the comment also? Because other ones are just comments on it.

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller_test.go
```diff
@@ -19,24 +19,35 @@ import (
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 	mock_dynamichelper "github.com/Azure/ARO-RP/pkg/util/mocks/dynamichelper"
 	_ "github.com/Azure/ARO-RP/pkg/util/scheme"
+	utilconditions "github.com/Azure/ARO-RP/test/util/conditions"
 	utilerror "github.com/Azure/ARO-RP/test/util/error"
+	operatorv1 "github.com/openshift/api/operator/v1"
```
mbarnes, line 24: Move this line up to the 2nd grouping of packages, under `ctrlfake`.

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -32,50 +32,65 @@ var machinehealthcheckYaml []byte
 var mhcremediationalertYaml []byte
 
 const (
-	ControllerName string = "MachineHealthCheck"
-	managed        string = "aro.machinehealthcheck.managed"
-	enabled        string = "aro.machinehealthcheck.enabled"
+	MHCControllerName string = "MachineHealthCheck"
+	MHCManaged        string = "aro.machinehealthcheck.managed"
+	MHCEnabled        string = "aro.machinehealthcheck.enabled"
```
SudoBrendan, line 37: nit: I think we can remove the `MHC` prefix changes here, since these constants will already be namespaced by the package name - we'd have `machinehealthcheck.MHCControllerName` which feels redundant compared to `machinehealthcheck.ControllerName`. I think this also breaks a standard pattern we have for all other controllers (seen in `operator.go`)

SudoBrendan (commented):
Overall, LGTM. I think removing the `MHC` prefix changes would reduce the diff and keep this controller in-line with other controller standard patterns. Nice work!

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller_test.go
safwank97: TYSM

safwank97:
Removed the MHC Prefix.

safwank97:
Done.

cmd/aro/operator.go
```diff
@@ -190,7 +190,7 @@ func operator(ctx context.Context, log *logrus.Entry) error {
 			client)).SetupWithManager(mgr); err != nil {
 			return fmt.Errorf("unable to create controller %s: %v", autosizednodes.ControllerName, err)
 		}
-		if err = (machinehealthcheck.NewReconciler(
+		if err = (machinehealthcheck.NewMachineHealthCheckReconciler(
```
SudoBrendan, line 193: This refactor is also going against our other controller patterns of always using the constructor name `NewReconciler`, see above.

SudoBrendan (changes requested):
Logic and tests are looking great! I found a few more breaks from convention with a few struct/function names. Keeping our controllers and properties named consistently makes it easier to work on each of them. Excellent work, thanks Mohammed!

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -37,45 +37,60 @@ const (
 	enabled        string = "aro.machinehealthcheck.enabled"
 )
 
-type Reconciler struct {
-	log *logrus.Entry
+type MachineHealthCheckReconciler struct {
```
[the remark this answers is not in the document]
safwank97: Sure currently I'm testing it by running a operator locally once that is done I'll update this as final refactor.

safwank97:
I'll take care of that part you mentioned. Thank you very much.

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -37,45 +37,60 @@ const (
...
 	if !instance.Spec.OperatorFlags.GetSimpleBoolean(enabled) {
-		r.log.Debug("controller is disabled")
+		r.Log.Debug("controller is disabled")
 		return reconcile.Result{}, nil
 	}
 
-	r.log.Debug("running")
+	r.Log.Debug("running")
 	if !instance.Spec.OperatorFlags.GetSimpleBoolean(managed) {
+		r.SetProgressing(ctx, "MHC and it's alerts are disabled because this cluster is not MHC managed.")
```
cadenmarchese, line 73: I would spell this out to help make it more clear to the customer:
```suggestion
		r.SetProgressing(ctx, "Managed MachineHealthCheck and associated alerts have been disabled in this cluster - reconciling the MachineHealthCheck object.")
```

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -37,45 +37,60 @@ const (
...
+
 		err := r.dh.EnsureDeleted(ctx, "MachineHealthCheck", "openshift-machine-api", "aro-machinehealthcheck")
 		if err != nil {
+			r.Log.Error(err)
+			r.SetDegraded(ctx, err)
+			r.ClearProgressing(ctx)
```
cadenmarchese, line 79: Ordering?
```suggestion
			r.ClearProgressing(ctx)
			r.SetDegraded(ctx, err)
```

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -37,45 +37,60 @@ const (
...
+
 			return reconcile.Result{RequeueAfter: time.Hour}, err
 		}
 
 		err = r.dh.EnsureDeleted(ctx, "PrometheusRule", "openshift-machine-api", "mhc-remediation-alert")
 		if err != nil {
+			r.Log.Error(err)
+			r.SetDegraded(ctx, err)
+			r.ClearProgressing(ctx)
```
cadenmarchese, line 88: Same comment as above.

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
tsatam: Order shouldn't actually matter but I agree with the change for consistency's sake

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
tsatam: We should not be setting a Progressing status here (or at all in this controller, in its current state). 

Progressing statuses should indicate to us that a reconciliation is actively in progress. In this controller, all the changes are performed in one go (which is an issue in and of itself that will need to be addressed in follow-up work, but not here). 

If the goal is to indicate to SRE that the MHC is unmanaged, the spec block of the cluster resource that these statuses ultimately end up on contains the `aro.machinehealthcheck.managed` flag within its spec.

safwank97: I'll update it by eod today.

tsatam (changes requested):
Looks good, there are just some minor fixups to make, awesome work!

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
safwank97: The way it is written here if the MHC alerts and Prometheus Rule aren't deleted then set Degraded so consider this way if they are deleted so it will not be in degraded but it has to clear the Progressing field mentioned above where it says not MHC Managed for both MHC alerts and Prometheus Rules and this clear Progressing should be for both of them individually.

safwank97:
Done.

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -38,44 +38,59 @@ const (
...
 	if !instance.Spec.OperatorFlags.GetSimpleBoolean(enabled) {
-		r.log.Debug("controller is disabled")
+		r.Log.Debug("controller is disabled")
 		return reconcile.Result{}, nil
 	}
 
-	r.log.Debug("running")
+	r.Log.Debug("running")
 	if !instance.Spec.OperatorFlags.GetSimpleBoolean(managed) {
+		r.SetProgressing(ctx, "Not MHC Managed for cluster maintenance purpose.")
```
SudoBrendan, line 73: Leaving a comment here based on our discussion yesterday, it seems like `Progressing` isn't necessary here. We should remove the `SetProgresing` and `ClearProgressing` calls.

SudoBrendan (changes requested):
This is looking really clean! Only one comment based on our discussion. Thanks for your hard work on this!

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
jhoreman: Agree here. `Progressing` is not required and will solve a few of the other requested changes on this PR by removing it.

SudoBrendan (approved):
LGTM, just need all CI/E2E green :)

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
safwank97: Removed those calls.

cadenmarchese (approved):
Nice work!

jhoreman (approved):
LGTM Thanks for this!

tsatam (approved):

PR #3181, opened 2023-09-19
Preflight update validation

carvalhe:

Currently we have preflight set up for create operations but also need it for update. These changes allow for both put/patch operations on preflight so that it can leave the testing stage and go live.

### Test plan for issue:
Created additional unit tests that are similar to put/patch tests. Kusto logs also show preflight success and failures with details.


pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -88,23 +101,34 @@ func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry,
...
-	converter.ToInternal(ext, oc)
-	if err := staticValidator.Static(ext, nil, f.env.Location(), f.env.Domain(), f.env.FeatureIsSet(env.FeatureRequireD2sV3Workers), resourceID); err != nil {
-		return api.ValidationResult{
-			Status: api.ValidationStatusFailed,
-			Error: &api.ManagementErrorWithDetails{
-				Message: to.StringPtr(err.Error()),
-			},
+	if isCreate {
+		log.Print("preflight static create validation")
+		if err := staticValidator.Static(ext, nil, f.env.Location(), f.env.Domain(), f.env.FeatureIsSet(env.FeatureRequireD2sV3Workers), resourceID); err != nil {
```
niontive, line 106: Nit: you can use the `err` from `f.dbOpenShiftClusters.Get(ctx, resourceID)`

niontive, line 106: Given the frontend PUT/PATCH code path uses `staticValidator.Static`, what's the advantage of putting this here? I understand pre-flight validation occurs before deployment, but if we return the same error message for both, what's the advantage of using preflight validation?

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -58,6 +58,8 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 		return nil, err
 	}
 
+	println("PATH")
```
niontive, line 61: Why are we printing this? I could maybe see value in logging the `path`

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: As stated in the meeting this morning, the goal is to shift validation from deployment to preflight. Once everything looks good on preflight side, we can remove the validation from PUT/PATCH (or at least the ones that can be performed in preflight)

carvalhe:
> In `func (f *frontend) ValidateNewCluster()`, there are several other validators that aren't performed by preflight validation. Will we add support for those validators, or are those out of scope?

Those other validators are out of scope for preflight, and will stay on the deployment side.

niontive (dismissed):

pkg/frontend/openshiftcluster_preflightvalidation_test.go
```diff
@@ -20,6 +21,58 @@ import (
...
+	preflightPayload := []byte(`
+	{
+		"apiVersion": "2022-04-01",
+		"id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resourceGroup/providers/Microsoft.RedHatOpenShift/openShiftClusters/resourceName",
+		"name": "resourceName",
+		"type": "microsoft.redhatopenshift/openshiftclusters",
+		"location": "eastus",
+		"properties": {
+			"clusterProfile": {
+			"domain": "default",
```
tony-schndr, line 33: nit: fix indentation in json to make it easier to read.

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -88,23 +101,34 @@ func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry,
...
+			return api.ValidationResult{
+				Status: api.ValidationStatusFailed,
+				Error: &api.ManagementErrorWithDetails{
+					Code:    to.StringPtr(api.CloudErrorCodeInvalidParameter),
+					Message: to.StringPtr(err.Error()),
+				},
+			}
+		}
+	} else {
+		log.Print("preflight static update validation")
```
tony-schndr, line 124: I noticed the story indicates to add update and delete but here it looks like this only covers updates?

tony-schndr (commented):
Nice work here!  Just a few questions/comments for you.

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: Correct, I after looking into preflight we saw that delete isn't a path thats going to be used by ARM and so we won't be adding that. Ill adjust the ticket.

pkg/frontend/openshiftcluster_preflightvalidation_test.go
carvalhe: Fixed the indentation

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -68,6 +70,17 @@ func (f *frontend) preflightValidation(w http.ResponseWriter, r *http.Request) {
 }
 
 func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry, raw json.RawMessage, apiVersion string, resourceID string) api.ValidationResult {
+	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	if err != nil && !cosmosdb.IsErrorStatusCode(err, http.StatusNotFound) {
```
SudoBrendan, line 74: It took me a hot second to understand why we perform this specific callout of `http.StatusNotFound` here but not with `isCreate` below.

For future reviewers: in my reading, our `Get` implementation always returns `nil` when `err` is not `nil` (including NotFound), so there's a case here where we get `nil` + `http.StatusNotFound`, which translates into `isCreate := true`, so the logic is sound with how `Get` happens to be implemented today - you just have to read all of `Get` to know it.

However, even though the current logic is sound as `Get` is written, I think `isCreate := cosmosdb.IsErrorStatusCode(err, http.StatusNotFound)` is a clearer and less brittle way to show that this specific error case is the only time where isCreate should be `true`:

```
doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
isCreate := cosmosdb.IsErrorStatusCode(err, http.StatusNotFound)
if err != nill && !isCreate {
    // return err
}

// rest of the changes
```

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -88,23 +101,34 @@ func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry,
...
-	}
-
-	if err := f.validateInstallVersion(ctx, oc); err != nil {
-		return api.ValidationResult{
-			Status: api.ValidationStatusFailed,
-			Error: &api.ManagementErrorWithDetails{
-				Code:    to.StringPtr(api.CloudErrorCodeInvalidParameter),
-				Message: to.StringPtr(err.Error()),
-			},
+		if err = f.validateInstallVersion(ctx, oc); err != nil {
```
SudoBrendan, line 114: bug? In the original code, this `oc` has been `converter.ToInternal`'d - are we now using the External version, and that works for all API versions?

pkg/frontend/openshiftcluster_preflightvalidation_test.go
```diff
@@ -20,6 +21,58 @@ import (
 func TestPreflightValidation(t *testing.T) {
 	ctx := context.Background()
 	mockSubID := "00000000-0000-0000-0000-000000000000"
+	preflightPayload := []byte(`
```
cadenmarchese, line 24: To DRY this up a bit, we could parameterize some of these mock IDs (such as `00000000-0000-0000-0000-000000000000`) with fmt.Sprintf - here's an example in another test: https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/machine/machine_controller_test.go#L243-L254

pkg/frontend/openshiftcluster_preflightvalidation.go
weinong: @carvalhe let's add a test case that would have failed here and what brendan said is correct, it should be done using internal data model

petrkotas (approved):
Hi @carvalhe with my limited understanding of the validation, I believe it is good. But Please @SudoBrendan could you please double check?

petrkotas (approved):
Hi @carvalhe with my limited understanding of the validation, I believe it is good. But Please @SudoBrendan could you please double check?

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: Made the suggested changes

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: Added a version test case, and correctly use the converter.ToInternal as well

SudoBrendan: I'm still unsure about this, to my eye, all test cases rely on the `apiVersion` static value, so everything still appears to be hitting `apiVersion := "2022-04-01"` in the tests, or maybe I'm missing the apiversion case you mentioned? In the original code, we called `toInternal` before `staticValidator.Static`, which is what is catching my attention here. TBH, I'm not entirely sure if that is required or not, just noting it's different/standing out to me that we've changed order of calls. If we have confirmation that this code works as expected in multiple RP API versions, then I'm fine with the change but I don't see that in the tests atm.

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -68,6 +70,18 @@ func (f *frontend) preflightValidation(w http.ResponseWriter, r *http.Request) {
 }
 
 func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry, raw json.RawMessage, apiVersion string, resourceID string) api.ValidationResult {
+	log.Infof("running preflight validation on resource: %s", resourceID)
+	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	isCreate := cosmosdb.IsErrorStatusCode(err, http.StatusNotFound)
+	if err != nil && !cosmosdb.IsErrorStatusCode(err, http.StatusNotFound) {
+		return api.ValidationResult{
```
bennerv, line 77: Why is there a new api.ValidationResult type instead of using the existing `pkg/api/error.go` -> `CloudError` struct?

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -52,9 +53,10 @@ func (f *frontend) preflightValidation(w http.ResponseWriter, r *http.Request) {
 			continue
```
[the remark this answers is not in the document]
bennerv: I guess api-version isn't set.  hmm wondering if there's a better way to do this with existing structs.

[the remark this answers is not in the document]
bennerv: `ProviderResourceType` model has api-versions though.  Not sure if it translates.

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -88,26 +102,36 @@ func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry,
...
-	converter.ToInternal(ext, oc)
-	if err := staticValidator.Static(ext, nil, f.env.Location(), f.env.Domain(), f.env.FeatureIsSet(env.FeatureRequireD2sV3Workers), resourceID); err != nil {
-		return api.ValidationResult{
-			Status: api.ValidationStatusFailed,
-			Error: &api.ManagementErrorWithDetails{
-				Message: to.StringPtr(err.Error()),
-			},
+	switch {
+	case isCreate:
+		if err = staticValidator.Static(ext, nil, f.env.Location(), f.env.Domain(), f.env.FeatureIsSet(env.FeatureRequireD2sV3Workers), resourceID); err != nil {
```
bennerv, line 107: In our RP openshiftcluster_putorpatch, we do quota validation as well on create.  would/should we consider this?

pkg/frontend/openshiftcluster_preflightvalidation.go
bennerv: +@SudoBrendan 's comments.  It may not have any impact actually, we'd have to examine which fields are being dropped on external conversion and if we're performing static validation on them to figure out if there's an actual issue with the representation of what we're statically validating.  

I'd prefer consistency with `openshiftcluster_putorpatch.go` though.

bennerv (commented):
Few comments, some suggestions.  

Makes sense though.  Great work so far.

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -52,9 +53,10 @@ func (f *frontend) preflightValidation(w http.ResponseWriter, r *http.Request) {
 			continue
 		}
 		if strings.EqualFold(typeMeta.Type, "Microsoft.RedHatOpenShift/openShiftClusters") {
-			res := f._preflightValidation(ctx, log, raw, typeMeta.APIVersion, typeMeta.Id)
+			res := f._preflightValidation(ctx, log, raw, typeMeta.APIVersion, strings.ToLower(typeMeta.Id))
 			if res.Status == api.ValidationStatusFailed {
 				log.Warningf("preflight validation failed")
+				log.Warning(to.String(res.Error.Message))
```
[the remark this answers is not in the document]
s-amann: +1 can we consolidate the two log lines

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: So the current test I have is for an incorrect apiversion which gets a failed case we desire. I tried other valid apiversions that seemed to work fine. I understand wanting the consistency with putorpath so I can move the toInternal call before static validate gets called (at least for create as update doesn't call toInternal before static.)

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: This was looked into for the previous PR and unfortunately this isn't something that can be done during preflight.

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: The documentation described the response for preflight to look like:
`type response = {
  status: "Failed" | "Succeeded";
  error: error;
}

type error = {
  code: string;
  message: string;
  target: string;
  details: error[];
}`

I saw that clouderror has a status code field, I didn't want that potentially interfering. The error looks to be the same as clouderrorbody so I could use that instead of making a new one.

pkg/frontend/openshiftcluster_preflightvalidation.go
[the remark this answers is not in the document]
carvalhe: Adjusted to one log line

pkg/frontend/openshiftcluster_preflightvalidation.go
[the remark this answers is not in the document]
carvalhe: Possibly. I think for now since we want to get this in for release the ResourceTypeMeta might be ok since the api-version is important.

pkg/api/error.go
```diff
@@ -43,6 +46,15 @@ type CloudErrorBody struct {
 	Details []CloudErrorBody `json:"details,omitempty"`
 }
 
+type ValidationStatus string
```
bennerv, line 49: Ok so I may have mislead you partially previously.  

So ARM dictates we must respond with a [specific error response content](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/common-api-details.md#error-response-content).  This doesn't include the newly added `status` field on the CloudError struct.  

So my suggestion is to create a new type, but nest the CloudError struct within it and add the field `status` necessary within it. 

This shouldn't generate new swagger specifications because that would require an API change.  

The thing that confuses me is ARM expects all error responses as [ODatav4.0 json format response](http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091), but for this specific one it expects the typical error response plus an additional field.  I would have hoped it would be common.  

nesting the struct within a new struct should solve our problem and introducing the additional `status` field should be sufficient here though.  

Apologies for the misleading previously.

PR #3187, opened 2023-09-21
rotate the openshift-config pull-secret on acr token rotation

s-amann:

Fixes https://issues.redhat.com/browse/ARO-4304

### Test plan for issue:
Due to the nature of acr token passwords, this will be tested manually

### DIFF
--- a/pkg/cluster/acrtoken.go
+++ b/pkg/cluster/acrtoken.go
@@ -149,12 +149,26 @@ func (m *manager) rotateOpenShiftConfigSecret(ctx context.Context, encodedDocker
 		if err != nil && !kerrors.IsNotFound(err) {
 			return err
 		}
+
+		// attempt to merge if we can, defaults to the created pull secret
+		if openshiftConfigSecret != nil && openshiftConfigSecret.Data != nil {
+			previousConfigData, previousConfigDataExists := openshiftConfigSecret.Data[corev1.DockerConfigJsonKey]
+			if previousConfigDataExists {
+				mergedPullSecretData, _, err := pullsecret.Merge(string(previousConfigData), string(encodedDockerConfigJson))
+				if err == nil {
+					recreatedSecret.Data[corev1.DockerConfigJsonKey] = []byte(mergedPullSecretData)
+				} else {
+					m.log.Error("Could not merge openshift config pull secret, overriding with new acr token", err)
+				}
+			}
+		}
 		return retryOperation(func() error {
 			_, err = m.kubernetescli.CoreV1().Secrets(pullSecretName.Namespace).Create(ctx, recreatedSecret, metav1.CreateOptions{})
 			return err
 		})
 	}
 
+	// update flow
 	mergedPullSecretData, _, err := pullsecret.Merge(string(openshiftConfigSecret.Data[corev1.DockerConfigJsonKey]), string(encodedDockerConfigJson))
 	openshiftConfigSecret.Data[corev1.DockerConfigJsonKey] = []byte(mergedPullSecretData)
 
--- a/pkg/cluster/acrtoken.go
+++ b/pkg/cluster/acrtoken.go
@@ -125,36 +125,30 @@ func (m *manager) rotateOpenShiftConfigSecret(ctx context.Context, encodedDocker
 	if err != nil {
 		return err
 	}
-	updatedSecret := corev1.Secret{}
 
-	recreate := openshiftConfigSecret == nil ||
+	recreationOfSecretRequired := openshiftConfigSecret == nil ||
 		(openshiftConfigSecret.Type != corev1.SecretTypeDockerConfigJson || openshiftConfigSecret.Data == nil) ||
 		(openshiftConfigSecret.Immutable != nil && *openshiftConfigSecret.Immutable)
 
-	if recreate {
-		updatedSecret = corev1.Secret{
+	if recreationOfSecretRequired {
+		recreatedSecret := &corev1.Secret{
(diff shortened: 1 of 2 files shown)


pkg/cluster/acrtoken.go
```diff
@@ -108,6 +112,49 @@ func (m *manager) rotateACRTokenPassword(ctx context.Context) error {
...
+				Namespace: pullSecretName.Namespace,
+			},
+			Type: corev1.SecretTypeDockerConfigJson,
+			Data: make(map[string][]byte),
+		}
+	} else {
+		updatedSecret = *openshiftConfigSecret.DeepCopy()
+	}
+
+	updatedSecret.Data[corev1.DockerConfigJsonKey] = encodedDockerConfigJson
```
dem4gus, line 147: This will overwrite all the data on the global pull secret, including customer data. We need to use [`pullsecret.Merge`](https://github.com/Azure/ARO-RP/blob/bfddd185cea94c279b14d87374d7933aeb25ecaa/pkg/util/pullsecret/pullsecret.go#L86)` at some point like the [controller code](https://github.com/Azure/ARO-RP/blob/bfddd185cea94c279b14d87374d7933aeb25ecaa/pkg/operator/controllers/pullsecret/pullsecret_controller.go#L173).

dem4gus (dismissed):
We need to merge the existing data in the global pull secret with the new token.

pkg/cluster/acrtoken.go
s-amann: we only need to merge if the original openshift-config pull secret had data in it, if we needed to recreate it, there is nothing to merge. I'll add the merge in for when we can update the secret.

pkg/cluster/acrtoken.go
```diff
@@ -108,6 +115,60 @@ func (m *manager) rotateACRTokenPassword(ctx context.Context) error {
...
+	}
 
 	return nil
 }
+
+func (m *manager) rotateOpenShiftConfigSecret(ctx context.Context, encodedDockerConfigJson []byte) error {
+	openshiftConfigSecret, err := m.kubernetescli.CoreV1().Secrets(pullSecretName.Namespace).Get(ctx, pullSecretName.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
```
bennerv, line 130: If the secret doesn't exist, do we want to attempt to recreate?

bennerv: We should have logic for kerrors is not found.  

```
if err != nil {
			if kerrors.IsNotFound(err) { // we are not in control if secret is not present
				return nil
			}
			return err
		}

**Review by SudoBrendan — CHANGES_REQUESTED** [review]
We already have [a bug in this controller](https://issues.redhat.com/browse/ARO-4021) (and others) surrounding multiple write operations of the same object in a single reconcile loop, and I wonder if we are introducing more of that here on this `Secret` resource. Generally, every time we modify any resources, we need to end reconciliation to prevent data from going stale/out of date in our reconciliation loop. Ideally we write 1 object at a time to move towards a desired state, then requeue reconciliation to happen again immediately. For example, a StatefulSet reconciliation regarding scaling would only create/delete 1 Pod at a time, even if the scale was off by 4 to begin with; that just means reconciliation is executed 4 times. I see at least 3 write operations on the same Secret in this PR, and I wonder if this will lead to inconsistent states and lots of errors in logs.

**Comment by SudoBrendan** [comment]
> This bug does not apply as we are not in a reconciliation loop on the cluster object, this is done during admin update. The logic in this PR will only call update on the openshift-config secret 1 time (if successful). For the update path we call delete and then create on the secret for 2 write calls.

I think then, proper controller structure would be something pseudo like:

```
func Reconcile() {
    if secretDoesNotExist() {
        createSecret()
        return nil, requeue(time.now())
    }
    if secretIsOutOfDate() {
        deleteSecret()
        return nil, requeue(time.now())
    }
}
```

bennerv:
@SudoBrendan 
The logic change isn't a controller.  It's included as part of PUCM.

SudoBrendan:
> @SudoBrendan The logic change isn't a controller. It's included as part of a step in PUCM.

I am blind. ty

bennerv:
@SudoBrendan

In regards to the controller reconciliation failures.  We should probably be using a PATCH instead of an Update to the aro cluster CR as during an Update, it's essentially a PUT operation to the resource.  A PATCH instead of Update will only update that given field.  

I think that would cause less contention from an error log perspective in the ARO operator.  Ideally we would use PATCH everywhere unless we want a specific configuration laid down.

tsatam (approved):
LGTM!

Some things @lranjbar and I noticed during review that might help future reviewers:

- The logic contained here is duplicated across here and the pullsecret controller in the ARO Operator. There are some differences in behavior (e.g. how a notfound error is handled when getting the existing pull secret). IMO this is fine to merge as is, with a future effort to try and deduplicate these with common logic in a shared third location. 
- The rotateACRTokenPassword function appears to recreate the pull secret if it needs to, using the contents of an available registryProfile within the cluster document. The code does not perform any nil checks and we were initially concerned that this function would cause a nil pointer dereference when attempting to read the username off this profile. After digging through the code further, we believe that all production clusters should have at least one registryProfile present in the clusterdoc, as the above ensureACRToken function, which is called during bootstrap, will ensure that a registryProfile is created if not already present.

hlipsig (approved):
I think the oustanding feedback has been addressed. No new concerns from me.

lranjbar (approved):

PR #3193, opened 2023-09-25
Rename workerProfilesStatus to ActiveWorkerProfiles

gouthamMN:

Fixes: https://issues.redhat.com/browse/ARO-4319

During the [review of upstreaming azure-rest-api-spec](https://github.com/Azure/azure-rest-api-specs/pull/25711) for stable api v2023-09-04, ARM team requested to make the following changes.

1. Rename WorkerProfilesStatus to ActiveWorkerProfiles. All instances of WorkerProfilesStatus have been changed.
3. Add cluster ProvisioningState enum "Canceled".

### Test plan for issue:
Existing tests were updated to refer ActiveWorkerProfiles to make sure the changes are verified.

No


bennerv:
We _probably_ need to leave the `workerProfileStatus` field in the internal API and nil it out everywhere because it's already shipped to production for at least one release before removing the internal representation.  

Otherwise our `MissingFields` field in the internal representation is going to leave it in there for all cluster docs and confuse us later.  

or at least if we can confirm the converter ToInternal removes it from the internal representation, we should be good.  will require some testing.

gouthamMN:
> We _probably_ need to leave the `workerProfileStatus` field in the internal API and nil it out everywhere because it's already shipped to production for at least one release before removing the internal representation.
> 
> Otherwise our `MissingFields` field in the internal representation is going to leave it in there for all cluster docs and confuse us later.
> 
> or at least if we can confirm the converter ToInternal removes it from the internal representation, we should be good. will require some testing.

Updated to keep WorkerProfilesStatus.

PR #3194, opened 2023-09-27
Don't Put Cluster in Update State after PUCM Pending

niontive:

Fixes bug that causes customer confusion when they randomly get update message.

Prevent customer confusion.

### Test plan for issue:

INT testing

https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/457767/ARO-Signals-for-Azure-Monitor


bennerv (approved):
lgtm - ty for the fix

s-amann (approved):

PR #3198, opened 2023-10-03
Add customer data regression tests for get/create/update/delete k8s admin api

cblecker:
None

Adds additional unit tests to the admin API to ensure that we validate we don't allow access to customer objects on get/create/update/delete k8s objects so that we don't regress these functions in the future.

### Test plan for issue:
Green CI == :shipit: 

No


bennerv (approved):
lgtm

PR #3201, opened 2023-10-04
Fix Construction of Monitor for Clusters with kube-apiservers Down

bennerv:

Story to be created. 

This fixes an issue during monitor creation resulting in a failure to emit metrics for clusters whose apiservers are down.  Similar to #2984 

### Test plan for issue:

Ran the monitor locally.  

We need to as follow-up create a monitor which will alert on the new metric we're introducing if it exceeds a threshold.  Currently, we have no way to catch this, but the introduction of a new metric will allow us to.

### DIFF
--- a/pkg/monitor/cluster/cluster.go
+++ b/pkg/monitor/cluster/cluster.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/metrics"
@@ -92,7 +93,15 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
 		return nil, err
 	}
 
-	ocpclientset, err := client.New(restConfig, client.Options{})
+	// lazy discovery will not attempt to reach out to the apiserver immediately
+	mapper, err := apiutil.NewDynamicRESTMapper(restConfig, apiutil.WithLazyDiscovery)
+	if err != nil {
+		return nil, err
+	}
+
+	ocpclientset, err := client.New(restConfig, client.Options{
+		Mapper: mapper,
+	})
 	if err != nil {
 		return nil, err
 	}
@@ -126,7 +135,15 @@ func getHiveClientSet(hiveRestConfig *rest.Config) (client.Client, error) {
 		return nil, nil
 	}
 
-	hiveclientset, err := client.New(hiveRestConfig, client.Options{})
+	// lazy discovery will not attempt to reach out to the apiserver immediately
+	mapper, err := apiutil.NewDynamicRESTMapper(hiveRestConfig, apiutil.WithLazyDiscovery)
+	if err != nil {
+		return nil, err
+	}
+
+	hiveclientset, err := client.New(hiveRestConfig, client.Options{
+		Mapper: mapper,
+	})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


niontive (approved):

hlipsig (approved):

pkg/monitor/cluster/cluster.go
```diff
@@ -92,7 +93,15 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
 		return nil, err
 	}
 
-	ocpclientset, err := client.New(restConfig, client.Options{})
+	// lazy discovery will not attempt to reach out to the apiserver immediately
+	mapper, err := apiutil.NewDynamicRESTMapper(restConfig, apiutil.WithLazyDiscovery)
```
petrkotas, line 97: I am not sure this will deal with the underlying issue of death kube-apiserver. The DynamicRESTMapper first talks to the API to get the resources during creation time. The calls that follow are:

- https://github.com/bennerv/ARO-RP/blob/375cc54a2964d979ecf0b4e831bcc64b730fbbfb/vendor/sigs.k8s.io/controller-runtime/pkg/client/apiutil/dynamicrestmapper.go#L85-L86 
- https://github.com/bennerv/ARO-RP/blob/375cc54a2964d979ecf0b4e831bcc64b730fbbfb/vendor/k8s.io/client-go/restmapper/discovery.go#L148
- https://github.com/bennerv/ARO-RP/blob/375cc54a2964d979ecf0b4e831bcc64b730fbbfb/vendor/k8s.io/client-go/discovery/discovery_client.go#L216
- https://github.com/bennerv/ARO-RP/blob/375cc54a2964d979ecf0b4e831bcc64b730fbbfb/vendor/k8s.io/client-go/discovery/discovery_client.go#L244
- https://github.com/bennerv/ARO-RP/blob/375cc54a2964d979ecf0b4e831bcc64b730fbbfb/vendor/k8s.io/client-go/discovery/discovery_client.go#L161

The laze loading kicks in later on.

petrkotas (commented):
Hi @bennerv I have few pointers, I might be understanding the issue differently than you, so apologies for poking. 

I am thinking, if the API server is down, every client we use will fail to reach the cluster. This leads to the whole monitoring to start reporting errors. Which leads to the fact, that we need to handle errors gracefully and not postpone it for later. Is that correct?

If yes, than I would prefer doing that than postponing error with lazy loading. 

As for the lazy loading, I dug into the k8s code, where I think it does not do what was the intent. The very first step in creation of the DynamicRESTMapper is the call to the API server. It is done gracefully, but it still can fail. 

I provided the call graph to make the finding easier for confirmation. I am not too sure whether it is the real problem or not, going back to problem description.

pkg/monitor/cluster/cluster.go
bennerv: @petrkotas in our monitoring stack, we run a series of checks.  [The first check is to call the apiserver](https://github.com/Azure/ARO-RP/blob/master/pkg/monitor/cluster/cluster.go#L147-L161).  If that call fails, the rest of the checks will not run as we return early and emit a metric.  

We don't actually care if the apiserver is up during client construction, what we do care about is that we're able to successfully construct the clients in order to check if the apiserver is up and emit the metric as the first portion of our monitor.  

If the apiserver is down, all other checks will fail anyway and client construction isn't relevant.

bennerv: >The DynamicRESTMapper first talks to the API to get the resources during creation time.
If you look at that piece of code linked, it's actually not calling the apiserver.  [It's creating a function which would call the apiserver to discover resources](https://github.com/bennerv/ARO-RP/blob/375cc54a2964d979ecf0b4e831bcc64b730fbbfb/vendor/sigs.k8s.io/controller-runtime/pkg/client/apiutil/dynamicrestmapper.go#L82-L91).  The actual call to the apiserver happens within the [`setStaticMapper` function](https://github.com/bennerv/ARO-RP/blob/375cc54a2964d979ecf0b4e831bcc64b730fbbfb/vendor/sigs.k8s.io/controller-runtime/pkg/client/apiutil/dynamicrestmapper.go#L98).

petrkotas (approved):
LGTM remarks are not a concern.

s-amann (approved):

PR #3203, opened 2023-10-05
fix monitor reusing aux function

AldoFusterTurpin:
Work performed by @bizz001 and me.
WIP

Fixes the panic in monitor when using ocpclientset fileld of the monitor.


mociarain:
What's the current state of this?

mociarain:
No longer needed after speaking with author

PR #3183, opened 2023-09-20
Add e2e tests for multiple ips per loadbalancer

tony-schndr:
### Which issue this PR addresses:

Fixes https://issues.redhat.com/browse/ARO-4178

### What this PR does / why we need it:

This PR adds e2e tests for increasing and decreasing load balancer managed outbound IPs

### Test plan for issue:

e2e tests

### Is there any documentation that needs to be updated for this PR?
N/A

PR #3204, opened 2023-10-06
BUG: Fix golang client PUT for GA

mbarnes:

Fixes [[ARO-4223] BUG: Fix golang client PUT for GA](https://issues.redhat.com/browse/ARO-4223)

This pull request allows clients to successfully make PUT requests to update a cluster's `properties.networkProfile.loadBalancerProfile.managedOutboundIps` value.

The bug has to do with the `effectiveOutboundIps` field, also in the `loadBalancerProfile`.  This field is tagged as "read-only" in the API's Swagger specification, which means generated API clients will omit this field when making POST or PUT requests.  But due to its omission, the request was failing static validation because the comparison logic treated this as an attempt to modify the read-only field.

This pull request presents a generalized method for handling "read-only" API fields by using a `swagger:"readOnly"` tag in the Golang struct definition for applicable API versions.  The comparison logic in the `github.com/Azure/ARO-RP/pkg/api/util/immutable` package acts on the `swagger:"readOnly"` struct tag by checking only that the incoming request body omits the read-only field.  More precisely in Golang terms, it checks that the read-only field is equal to the field type's "zero-value", which for slices like `effectiveOutboundIps` is `nil`.

Furthermore, this pull request takes steps to ensure the `effectiveOutboundIps` field is preserved in the CosmosDB document during a long-running cluster update operation, even though it will be updated at the end of the operation.  This is important for two reasons:

1. The frontend still responds to GET requests during a long-running operation, and the caller may expect `effectiveOutboundIps` to be present and accurate (pending completion of the operation).
2. The long-running operation may fail, preventing the `effectiveOutboundIps` field from being updated with new data.  If the old data is not preserved then the field will remain empty until the next PUT request.

### Test plan for issue:

I piggy-backed on @tony-schndr's pull request to add E2E tests for these fields (https://github.com/Azure/ARO-RP/pull/3183), modifying the relevant tests to use a PUT request instead of PATCH.  I did not include the test modifications here to avoid a dependency on Tony's pull request.

Update: Since https://github.com/Azure/ARO-RP/pull/3183 is merged, I've appended an E2E test modification commit.

No, this is an internal bug fix.

### DIFF
--- a/pkg/api/admin/openshiftcluster_convert.go
+++ b/pkg/api/admin/openshiftcluster_convert.go
@@ -252,7 +252,16 @@ func (c openShiftClusterConverter) ToInternal(_oc interface{}, out *api.OpenShif
 	out.Properties.NetworkProfile.GatewayPrivateEndpointIP = oc.Properties.NetworkProfile.GatewayPrivateEndpointIP
 	out.Properties.NetworkProfile.GatewayPrivateLinkID = oc.Properties.NetworkProfile.GatewayPrivateLinkID
 	if oc.Properties.NetworkProfile.LoadBalancerProfile != nil {
-		out.Properties.NetworkProfile.LoadBalancerProfile = &api.LoadBalancerProfile{}
+		loadBalancerProfile := api.LoadBalancerProfile{}
+
+		// EffectiveOutboundIPs is a read-only field, so it will never be present in requests.
+		// Preserve the slice from the pre-existing internal object.
+		if out.Properties.NetworkProfile.LoadBalancerProfile != nil {
+			loadBalancerProfile.EffectiveOutboundIPs = make([]api.EffectiveOutboundIP, len(out.Properties.NetworkProfile.LoadBalancerProfile.EffectiveOutboundIPs))
+			copy(loadBalancerProfile.EffectiveOutboundIPs, out.Properties.NetworkProfile.LoadBalancerProfile.EffectiveOutboundIPs)
+		}
+
+		out.Properties.NetworkProfile.LoadBalancerProfile = &loadBalancerProfile
 
 		if oc.Properties.NetworkProfile.LoadBalancerProfile.AllocatedOutboundPorts != nil {
 			out.Properties.NetworkProfile.LoadBalancerProfile.AllocatedOutboundPorts = oc.Properties.NetworkProfile.LoadBalancerProfile.AllocatedOutboundPorts
--- a/pkg/api/admin/openshiftcluster_convert.go
+++ b/pkg/api/admin/openshiftcluster_convert.go
@@ -351,3 +351,12 @@ func (c openShiftClusterConverter) ToInternal(_oc interface{}, out *api.OpenShif
 	// with filling the out.Properties.RegistryProfiles[i].Password as default is "" which erases the original value.
 	// Workaround would be filling the password when receiving request, but it is array and the logic would be to complex.
 }
+
+// ExternalNoReadOnly removes all read-only fields from the external representation.
+func (c openShiftClusterConverter) ExternalNoReadOnly(_oc interface{}) {
+	oc := _oc.(*OpenShiftCluster)
+	oc.Properties.WorkerProfilesStatus = nil
+	if oc.Properties.NetworkProfile.LoadBalancerProfile != nil {
+		oc.Properties.NetworkProfile.LoadBalancerProfile.EffectiveOutboundIPs = nil
+	}
+}
--- a/pkg/api/util/immutable/immutable.go
+++ b/pkg/api/util/immutable/immutable.go
@@ -153,6 +153,14 @@ func validate(path string, v, w reflect.Value, ignoreCase bool) error {
 			}
 			subpath += name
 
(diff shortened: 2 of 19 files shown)


SudoBrendan:
I think our best chance to avoid errors is to add the E2E tests - let's get #3183 merged, and add a single `PUT` test in the suite to prove the functionality here.

pkg/api/admin/openshiftcluster_convert.go
```diff
@@ -252,7 +252,16 @@ func (c openShiftClusterConverter) ToInternal(_oc interface{}, out *api.OpenShif
...
 	out.Properties.NetworkProfile.GatewayPrivateLinkID = oc.Properties.NetworkProfile.GatewayPrivateLinkID
 	if oc.Properties.NetworkProfile.LoadBalancerProfile != nil {
-		out.Properties.NetworkProfile.LoadBalancerProfile = &api.LoadBalancerProfile{}
+		loadBalancerProfile := api.LoadBalancerProfile{}
+
+		// EffectiveOutboundIPs is a read-only field, so it will never be present in requests.
+		// Preserve the slice from the pre-existing internal object.
+		if out.Properties.NetworkProfile.LoadBalancerProfile != nil {
+			loadBalancerProfile.EffectiveOutboundIPs = make([]api.EffectiveOutboundIP, len(out.Properties.NetworkProfile.LoadBalancerProfile.EffectiveOutboundIPs))
+			copy(loadBalancerProfile.EffectiveOutboundIPs, out.Properties.NetworkProfile.LoadBalancerProfile.EffectiveOutboundIPs)
```
s-amann, line 261: nice use of copy

SudoBrendan:
@mbarnes https://github.com/Azure/ARO-RP/pull/3183 just merged, can we add in the PUT E2E test?

s-amann (approved):

SudoBrendan (approved):

tony-schndr:
@mbarnes @SudoBrendan would it be worth to cut over the OperationsClient and OpenShiftClustersClient to use the preview API instead of having the additional client now that the bug is fixed?

https://github.com/Azure/ARO-RP/blob/365413c3576bd42faac61b77d565186a6f21a39f/test/e2e/setup.go#L58-L60

mbarnes:
> @mbarnes @SudoBrendan would it be worth to cut over the OperationsClient and OpenShiftClustersClient to use the preview API instead of having the additional client now that the bug is fixed?

No strong opinion but happy to add another commit if you and @SudoBrendan are in agreement.

pkg/api/admin/openshiftcluster_convert.go
```diff
@@ -342,3 +351,12 @@ func (c openShiftClusterConverter) ToInternal(_oc interface{}, out *api.OpenShif
 	// with filling the out.Properties.RegistryProfiles[i].Password as default is "" which erases the original value.
 	// Workaround would be filling the password when receiving request, but it is array and the logic would be to complex.
 }
+
+// ExternalNoReadOnly removes all read-only fields from the external representation.
+func (c openShiftClusterConverter) ExternalNoReadOnly(_oc interface{}) {
```
petrkotas, line 356: This can be done using the reflection package to make us of the swagger field. Same is done for RP config here https://github.com/mbarnes/ARO-RP/blob/452d153e8b413ad28efadb9b7e5b989d89c59aac/pkg/deploy/config.go#L166-L188. 

No need to block on this, but I would like to see done as a follow up, please. This has the potential to be forgotten over time when setting field in config structure and forgetting it here.

pkg/api/util/immutable/immutable.go
```diff
@@ -151,7 +153,15 @@ func validate(path string, v, w reflect.Value, ignoreCase bool) error {
 			}
 			subpath += name
 
-			ic := ignoreCase || strings.EqualFold(v.Type().Field(i).Tag.Get("mutable"), "case")
+			// Read-only properties should be omitted from PUT/POST requests.
+			if strings.EqualFold(structField.Tag.Get("swagger"), "readOnly") {
+				if !v.FieldByIndex([]int{i}).IsZero() {
```
petrkotas, line 158: This is a bit confusing. I believe this can be changed to 

```suggestion
if !v.Field(i).IsZero() {
```

as this is what happends internally if there is only one value in the array.

petrkotas (approved):
Hi @mbarnes thank you for the PR. It looks good. I have two nonblocking comments. I think it is good to go in.

pkg/api/admin/openshiftcluster_convert.go
mbarnes: I think it's appropriate to do as a follow up.  I looked into doing this with `reflect` but it's going to require recursion since these tags can be several levels deep in nested struct types.

What might be best is see if I can factor out the recursive type walking in [immutable.go](https://github.com/Azure/ARO-RP/blob/master/pkg/api/util/immutable/immutable.go), apply the [visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern) through a callback parameter, and make some kind of reusable type inspection utility.

PR #3205, opened 2023-10-06
Pause ARO MachineHealthCheck during cluster upgrades

tsatam:

Fixes [ARO-4041](https://issues.redhat.com/browse/ARO-4041)

Adds the `cluster.x-k8s.io/paused=""` annotation to the `aro-machinehealthcheck` resource during cluster upgrades, and ensures the MachineHealthCheck controlle responsible for this resource requeues (every hour) during a cluster upgrade, in order to ensure the annotation is removed after the upgrade is complete. 

Updates DynamicHelper to preserve the Status when updating MachineHealthChecks. This fixes an existing bug where the MachineHealthCheck controller would delete the Status on the aro-machinehealthcheck on every reconciliation. 

### Test plan for issue:

- Unit tests were updated to ensure the controller requeues during cluster upgrades
  - Due to limitations in the mock DynamicHelper used in the unit tests, it is not possible to ensure the annotation is actually set/unset when expected in these unit tests.
- A manual test was performed successfully on a dev cluster running an operator with this change:
  - Ensure the annotation is not present on the `aro-machinehealthcheck` resource
  - Initiate a cluster upgrade
  - Ensure the annotation is added to the `aro-machinehealthcheck` resource
  - After the upgrade is complete, ensure the annotation is no longer present 

Internal SRE documentation ([link](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/185635/MachineHealthCheck#)) on ARO MachineHealthCheck 

### Additional Notes

- The RequeueAfter may not be necessary here, as this controller now watches the clusterversion resource which will be constantly undergoing changes throughout an upgrade.

### DIFF
--- a/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
@@ -9,16 +9,20 @@ import (
 	"strings"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/sirupsen/logrus"
 	kruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 	"github.com/Azure/ARO-RP/pkg/operator/controllers/base"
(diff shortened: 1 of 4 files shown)


pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -101,6 +110,23 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+				r.SetDegraded(ctx, err)
+
+				return reconcile.Result{}, err
+			}
+
+			if isUpgrading {
+				mhc.ObjectMeta.Annotations = map[string]string{
+					MHCPausedAnnotation: "",
+				}
+				shouldRequeue = true
```
s-amann, line 126: instead of using a boolean here, can we declare `requeueAfter` on line 100 and then set it to 1hr here?

Unless we had other plans for `shouldRequeue` I don't see any other use for it now.

s-amann (commented):
lgtm, please consider removing the use of a 'tracking boolean' below.

pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
tsatam: I can make that change. I also think we can just remove the requeueing entirely, since the clusterversion resource will update itself during an upgrade, causing the reconciler here to fire again. I can test this.

tsatam: I've confirmed that requeueing is completely unnecessary for cluster upgrades since the updated clusterversion resource will invoke a reconcile anyways. 

I'll push up a change to remove it entirely after I get some better test coverage around it.

cadenmarchese (approved):

SudoBrendan (approved):

cadenmarchese (approved):
lgtm!

PR #3207, opened 2023-10-10
Fail cluster deletion if RP is not authorized to delete managed resource group

tsatam:

Fixes [ARO-4436](https://issues.redhat.com/browse/ARO-4436)

Currently, during cluster deletion, the RP will silently succeed if it does not have the permissions to delete the ARO-managed resource group for the cluster. This PR changes the behavior to fail the delete operation. 

The PR also refactors the deletion logic to be simpler, taking advantage of new permissions added to the RP FPSP to read resource groups. 

### Test plan for issue:

Unit tests have been added for some of the logic around deciding when to delete, and how to handle deletion errors, though the entire deletion flow is not unit tested. 

N/A


bennerv (approved):
lgtm

PR #3211, opened 2023-10-12
Network acl fix for storageaccounts on Create or AdminUpdate

bennerv:

Fixes [ARO-4450](https://issues.redhat.com/browse/ARO-4450)

to properly use `azurefile-csi` storage class that comes out of the box in openshift, the worker / master virtual machine must be able to talk to the cluster storage account.  

Set Network ACLs on the cluster storage account during Creation & AdminUpdate only if the Microsoft.Storage service endpoint is set on the subnets.  

### Test plan for issue:

There will be an e2e test that will run against production clusters to ensure this works based on the storage account controller.  

We can test this as follows though: 
- Disable the storage account controller
- Disable the service endpoints controller
- Disable `ensureServiceEndpoints` step in RP for dev clusters only
- Remove NACLs from cluster storage account
- Set service endpoints on one or all cluster subnets
- NACLs should only include subnets which have Microsoft.Storage service endpoints.  

We need to publish something along the lines of:
- the default `azurefile-csi` storage class will not work if you don't have Microsoft.Storage service endpoints set on the subnets your OCP cluster uses.  The way around this is to create your own storage class.

### DIFF
--- a/pkg/cluster/deploybaseresources.go
+++ b/pkg/cluster/deploybaseresources.go
@@ -178,6 +178,45 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
 	return arm.DeployTemplate(ctx, m.log, m.deployments, resourceGroup, "storage", t, nil)
 }
 
+// subnetsWithServiceEndpoint returns a unique slice of subnet resource IDs that have the corresponding
+// service endpoint
+func (m *manager) subnetsWithServiceEndpoint(ctx context.Context, serviceEndpoint string) ([]string, error) {
+	subnetsMap := map[string]struct{}{}
+
+	subnetsMap[m.doc.OpenShiftCluster.Properties.MasterProfile.SubnetID] = struct{}{}
+	workerProfiles, _ := api.GetEnrichedWorkerProfiles(m.doc.OpenShiftCluster.Properties)
+	for _, v := range workerProfiles {
+		// don't fail empty worker profiles/subnet IDs as they're not valid
+		if v.SubnetID == "" {
+			continue
+		}
+
+		subnetsMap[v.SubnetID] = struct{}{}
+	}
+
+	subnets := []string{}
+	for subnetId := range subnetsMap {
+		// We purposefully fail if we can't fetch the subnet as the FPSP most likely
+		// lost read permission over the subnet.
+		subnet, err := m.subnet.Get(ctx, subnetId)
+		if err != nil {
+			return nil, err
+		}
+
+		if subnet.SubnetPropertiesFormat == nil || subnet.ServiceEndpoints == nil {
+			continue
+		}
+
+		for _, endpoint := range *subnet.ServiceEndpoints {
+			if endpoint.Service != nil && strings.EqualFold(*endpoint.Service, serviceEndpoint) {
+				subnets = append(subnets, subnetId)
+			}
+		}
+	}
+
+	return subnets, nil
+}
+
 func (m *manager) attachNSGs(ctx context.Context) error {
 	if m.doc.OpenShiftCluster.Properties.NetworkProfile.PreconfiguredNSG == api.PreconfiguredNSGEnabled {
 		return nil
--- a/pkg/cluster/deploybaseresources.go
+++ b/pkg/cluster/deploybaseresources.go
@@ -25,6 +25,8 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/stringutils"
 )
 
+const storageServiceEndpoint = "Microsoft.Storage"
+
 func (m *manager) createDNS(ctx context.Context) error {
 	return m.dns.Create(ctx, m.doc.OpenShiftCluster)
 }
@@ -132,11 +134,16 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
 	clusterStorageAccountName := "cluster" + m.doc.OpenShiftCluster.Properties.StorageSuffix
 	azureRegion := strings.ToLower(m.doc.OpenShiftCluster.Location) // Used in k8s object names, so must pass DNS-1123 validation
 
+	ocpSubnets, err := m.subnetsWithServiceEndpoint(ctx, storageServiceEndpoint)
+	if err != nil {
+		return err
+	}
+
 	resources := []*arm.Resource{
-		m.storageAccount(clusterStorageAccountName, azureRegion, true),
+		m.storageAccount(clusterStorageAccountName, azureRegion, ocpSubnets, true),
 		m.storageAccountBlobContainer(clusterStorageAccountName, "ignition"),
 		m.storageAccountBlobContainer(clusterStorageAccountName, "aro"),
-		m.storageAccount(m.doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName, azureRegion, true),
+		m.storageAccount(m.doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName, azureRegion, ocpSubnets, true),
 		m.storageAccountBlobContainer(m.doc.OpenShiftCluster.Properties.ImageRegistryStorageAccountName, "image-registry"),
 		m.clusterNSG(infraID, azureRegion),
 		m.clusterServicePrincipalRBAC(),
@@ -191,7 +198,7 @@ func (m *manager) subnetsWithServiceEndpoint(ctx context.Context, serviceEndpoin
 			continue
 		}
 
-		subnetsMap[v.SubnetID] = struct{}{}
+		subnetsMap[strings.ToLower(v.SubnetID)] = struct{}{}
 	}
 
 	subnets := []string{}
@@ -208,8 +215,12 @@ func (m *manager) subnetsWithServiceEndpoint(ctx context.Context, serviceEndpoin
 		}
 
 		for _, endpoint := range *subnet.ServiceEndpoints {
-			if endpoint.Service != nil && strings.EqualFold(*endpoint.Service, serviceEndpoint) {
-				subnets = append(subnets, subnetId)
+			if endpoint.Service != nil && strings.EqualFold(*endpoint.Service, serviceEndpoint) && endpoint.Locations != nil {
+				for _, loc := range *endpoint.Locations {
+					if loc == "*" || strings.EqualFold(loc, m.doc.OpenShiftCluster.Location) {
+						subnets = append(subnets, subnetId)
+					}
+				}
 			}
 		}
 	}
--- a/pkg/cluster/deploybaseresources_test.go
+++ b/pkg/cluster/deploybaseresources_test.go
@@ -8,6 +8,8 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"reflect"
+	"sort"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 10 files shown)


pkg/cluster/deploybaseresources.go
```diff
@@ -178,6 +185,54 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
...
+
+	// TODO: service endpoints can be region specific.  Will this cause a problem
+	// if serviceendpoint is on regionA but storage account is region B?
+	subnets := []string{}
+	for subnetId := range subnetsMap {
+		// TODO: do we want to fail if we can't fetch a subnet?
+		// customers could configure machinesets to invalid subnets, and could cause this to fail
+		// it wouldn't work anyway, so might make sense to be best-effort.
+
+		// Does subnet GET expand service endpoints properly?
```
s-amann, line 212: I sure hope so

pkg/cluster/deploybaseresources.go
```diff
@@ -178,6 +185,54 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
 	return arm.DeployTemplate(ctx, m.log, m.deployments, resourceGroup, "storage", t, nil)
 }
 
+// subnetsWithServiceEndpoint returns a unique slice of subnet resource IDs that have the corresponding
+// service endpoint
+func (m *manager) subnetsWithServiceEndpoint(ctx context.Context, serviceEndpoint string) ([]string, error) {
+	subnetsMap := map[string]struct{}{}
```
s-amann, line 191: why a map here

pkg/cluster/deploybaseresources.go
```diff
@@ -178,6 +185,54 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
...
+	for _, v := range workerProfiles {
+		// don't fail empty worker profiles/subnet IDs as they're not valid
+		if v.SubnetID == "" {
+			continue
+		}
+
+		subnetsMap[strings.ToLower(v.SubnetID)] = struct{}{}
+	}
+
+	// TODO: service endpoints can be region specific.  Will this cause a problem
```
cadenmarchese, line 204: I think so - the wording [here](https://learn.microsoft.com/en-us/azure/storage/common/storage-network-security?toc=%2Fazure%2Fvirtual-network%2Ftoc.json&tabs=azure-portal#grant-access-from-a-virtual-network) suggests that when connecting a vnet in region A to a storage account in region B, you need a cross-region service endpoint.

GIven that these just went GA a few months back, I am assuming we are using service endpoints classic which are region specific.

pkg/cluster/deploybaseresources.go
bennerv: returns a set of subnets.  

if you try to create NACLs with two of the same subnet it will fail.

pkg/cluster/deploybaseresources.go
bennerv: it does after some testing :+1:

pkg/cluster/deploybaseresources.go
bennerv: I accounted for both - need to remove the comments.

pkg/cluster/deploybaseresources_test.go
```diff
@@ -400,3 +402,209 @@ func TestEnsureInfraID(t *testing.T) {
...
+								SubnetID: masterSubnet,
+							},
+							WorkerProfiles: workerProfiles,
+						},
+					},
+				},
+				subnet: subnet,
+			}
+
+			// hopefully setting a seed here means it passes consistently :)
```
s-amann, line 596: nothing like betting on rng

bennerv: :game_die: 

I don't think we actually need seeding since we're not using random at all.  I'll remove.

pkg/cluster/deploybaseresources.go
s-amann: yeah, the alternative is to loop over the slice looking for a subnet already added..

pkg/cluster/deploybaseresources.go
cadenmarchese: For my own knowledge then - in [this loop](https://github.com/Azure/ARO-RP/pull/3211/files#diff-757380ec654f144c2f36b9c7f8c8ce2ca911fb9d054d258aa81c447c47ff7de6R222) we append location if resource type is `Microsoft.Storage`, but just skip updating it if resource type is `Microsoft.Storage.Global` since it doesn't need a location - and this handles both service endpoint types. Is that roughly correct?

pkg/operator/controllers/storageaccounts/storageaccounts.go
```diff
@@ -22,21 +21,37 @@ func (r *reconcileManager) reconcileAccounts(ctx context.Context) error {
 
 	serviceSubnets := r.instance.Spec.ServiceSubnets
 
-	// Only include the master and worker subnets in the storage accounts' virtual
-	// network rules if egress lockdown is not enabled.
-	if !operator.GatewayEnabled(r.instance) {
-		subnets, err := r.kubeSubnets.List(ctx)
+	subnets, err := r.kubeSubnets.List(ctx)
+
```
bennerv, line 25: nit - new line shouldn't be here

pkg/operator/controllers/storageaccounts/storageaccount_controller.go
```diff
@@ -103,6 +104,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 
 		client:      r.client,
 		kubeSubnets: subnet.NewKubeManager(r.client, resource.SubscriptionID),
+		mgmtSubnets: subnet.NewManager(&azEnv, resource.SubscriptionID, authorizer),
```
bennerv, line 107: nit - typically we just call this `subnet` or `subnets`.

pkg/operator/controllers/storageaccounts/storageaccounts.go
```diff
@@ -22,21 +21,37 @@ func (r *reconcileManager) reconcileAccounts(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Check each of the cluster subnets for the Microsoft.Storage service endpoint. If the subnet has
+	// the service endpoint, it needs to be included in the storage account vnet rules.
+	for _, subnet := range subnets {
+		hasStorageEndpoint := false
+		mgmtSubnet, err := r.mgmtSubnets.Get(ctx, subnet.ResourceID)
```
bennerv, line 34: nit: just call this subnet

pkg/operator/controllers/storageaccounts/storageaccounts.go
```diff
@@ -22,21 +21,37 @@ func (r *reconcileManager) reconcileAccounts(ctx context.Context) error {
...
+
```
bennerv, line 35: nit - no new line

pkg/operator/controllers/storageaccounts/storageaccounts.go
```diff
@@ -22,21 +21,37 @@ func (r *reconcileManager) reconcileAccounts(ctx context.Context) error {
...
 		if err != nil {
 			return err
 		}
 
-		for _, subnet := range subnets {
+		if mgmtSubnet.SubnetPropertiesFormat != nil && mgmtSubnet.SubnetPropertiesFormat.ServiceEndpoints != nil {
+			for _, serviceEndpoint := range *mgmtSubnet.SubnetPropertiesFormat.ServiceEndpoints {
+				if serviceEndpoint.Service != nil && *serviceEndpoint.Service == "Microsoft.Storage" {
+					hasStorageEndpoint = true
```
bennerv, line 43: do we want to confirm location matches for the service endpoint as well?

bennerv: also, no need for a boolean, just append right here and break out of your loop

test/e2e/cluster.go
```diff
@@ -115,22 +243,34 @@ var _ = Describe("Cluster", func() {
...
-	oc, _ := clients.OpenshiftClusters.Get(ctx, vnetResourceGroup, clusterName)
-	installVersion, _ := version.ParseVersion(*oc.ClusterProfile.Version)
+// clusterSubnets returns a slice containing all of the cluster subnets' resource IDs
+func clusterSubnets(oc redhatopenshift.OpenShiftCluster) []string {
+	subnetMap := map[string]struct{}{}
+	subnetMap[*oc.OpenShiftClusterProperties.MasterProfile.SubnetID] = struct{}{}
+
+	// TODO: change to workerProfileStatuses when we bump the API to 20230904 stable
+	for _, p := range *oc.OpenShiftClusterProperties.WorkerProfiles {
+		subnetMap[*p.SubnetID] = struct{}{}
```
AldoFusterTurpin, line 253: I might be wrong, but I think it would be safer to convert to lowercase when inserting on the map, not when inserting on the slice. Because right now we are converting to lowercase when inserting the values on the slice, which could lead to duplicates in the slice itself as the map can contain {A, a} and then slice later will contain {a, a}.

Maybe there is a precondition that this can not happen but I prefer to mention it even I am wrong.
😃  

I think it would be safer to do:
```go
func clusterSubnets(oc redhatopenshift.OpenShiftCluster) []string {
	subnetMap := map[string]struct{}{}
	subnetMap[*oc.OpenShiftClusterProperties.MasterProfile.SubnetID] = struct{}{}

	// TODO: change to workerProfileStatuses when we bump the API to 20230904 stable
	for _, p := range *oc.OpenShiftClusterProperties.WorkerProfiles {
		s := strings.ToLower(*p.SubnetID) // change -> ToLower here to ensure no duplicates
		subnetMap[s] = struct{}{}
	}

	subnets := []string{}

	for subnet := range subnetMap {
		subnets = append(subnets, subnet) //change -> ToLower NOT here
	}

	return subnets
}
```

AldoFusterTurpin (commented):
I left a comment as I am not sure if this is causing the e2e test failure we have in INT.

That function is just used in the e2e test that is failing. I might be wrong but just in case...
Thanks!

test/e2e/cluster.go
kimorris27: I don't think this is the reason E2E failed, but that makes sense and I've made the change you suggested 👍🏻

test/e2e/cluster.go
```diff
@@ -52,18 +60,150 @@ var _ = Describe("Cluster", func() {
...
+		err := createStatefulSet(ctx, clients.Kubernetes, storageClass)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("verifying the stateful set is ready")
+		Eventually(func(g Gomega, ctx context.Context) {
+			s, err := clients.Kubernetes.AppsV1().StatefulSets(testNamespace).Get(ctx, fmt.Sprintf("busybox-%s", storageClass), metav1.GetOptions{})
+			g.Expect(err).NotTo(HaveOccurred())
+
+			g.Expect(ready.StatefulSetIsReady(s)).To(BeTrue(), "expect stateful to be ready")
+		}).WithContext(ctx).Should(Succeed())
```
cblecker, line 82:
```suggestion
		}).Should(Succeed())
```

test/e2e/cluster.go
```diff
@@ -52,18 +60,150 @@ var _ = Describe("Cluster", func() {
...
+				if rule.Action == storage.Allow && rule.VirtualNetworkResourceID != nil {
+					nAclSubnets = append(nAclSubnets, strings.ToLower(*rule.VirtualNetworkResourceID))
+				}
+			}
+
+			for _, subnet := range ocpSubnets {
+				g.Expect(nAclSubnets).To(ContainElement(strings.ToLower(subnet)))
+			}
+
+		}).WithContext(ctx).Should(Succeed())
```
cblecker, line 172:
```suggestion
		}).Should(Succeed())
```

test/e2e/cluster.go
```diff
@@ -52,18 +60,150 @@ var _ = Describe("Cluster", func() {
...
 		Expect(err).NotTo(HaveOccurred())
 
 		By("verifying the stateful set is ready")
 		Eventually(func(g Gomega, ctx context.Context) {
-			s, err := clients.Kubernetes.AppsV1().StatefulSets(testNamespace).Get(ctx, "busybox", metav1.GetOptions{})
+			s, err := clients.Kubernetes.AppsV1().StatefulSets(testNamespace).Get(ctx, fmt.Sprintf("busybox-%s", storageClass), metav1.GetOptions{})
 			g.Expect(err).NotTo(HaveOccurred())
 
 			g.Expect(ready.StatefulSetIsReady(s)).To(BeTrue(), "expect stateful to be ready")
 		}).WithContext(ctx).Should(Succeed())
```
cblecker, line 185:
```suggestion
		}).Should(Succeed())
```

hlipsig (approved):

cblecker (approved):

cblecker:
Based on the conversation I've had with @hlipsig this morning, merging this..

PR #3213, opened 2023-10-14
[Draft] Remove bad storageClass in clusergo for hotfix test

hlipsig:
Remove non-existent storage class as from test.

Fixes

I think the test was timing out waiting for the stateful set to become ready because the storage class didn't exist. I removed the bad storage class, that was wrapped in an if statement, and now we need to run e2e again.

### Test plan for issue:

e2e passes

No


hlipsig:
If was for Less than 4.11 so we want it, Closing.

PR #3220, opened 2023-10-18
Additional Waiting for API Server to become ready

rhamitarora:

https://issues.redhat.com/browse/ARO-4404

A long discussion is here https://redhat-internal.slack.com/archives/C02ULBRS68M/p1697570801196159?thread_ts=1695641115.974799&cid=C02ULBRS68M

### Test plan for issue:

None

None

### DIFF
--- a/pkg/cluster/install.go
+++ b/pkg/cluster/install.go
@@ -343,6 +343,7 @@ func (m *manager) Install(ctx context.Context) error {
 			steps.Action(m.initializeOperatorDeployer), // depends on kube clients
 			steps.Action(m.removeBootstrap),
 			steps.Action(m.removeBootstrapIgnition),
+			steps.Condition(m.apiServersReady, 30*time.Minute, true), // waiting for API Server to become ready
 			steps.Action(m.configureAPIServerCertificate),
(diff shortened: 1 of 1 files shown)


pkg/cluster/install.go
```diff
@@ -343,6 +343,7 @@ func (m *manager) Install(ctx context.Context) error {
 			steps.Action(m.initializeOperatorDeployer), // depends on kube clients
 			steps.Action(m.removeBootstrap),
 			steps.Action(m.removeBootstrapIgnition),
+			steps.Condition(m.apiServersReady, 30*time.Minute, true), // waiting for API Server to become ready
```
bennerv, line 346: Should we put a comment in here stating the apiserver will occasionally be disrupted and the step to configure certificates will fail?  Currently under investigation.

bennerv (approved):
lgtm - wondering if we should add a comment though.

Shivkumar13 (approved):
LGTM

pkg/cluster/install.go
rhamitarora: Comment updated

Shivkumar13 (approved):

bennerv (approved):
lgtm

PR #3221, opened 2023-10-18
Propagate errors of GenevaLogging controller

mbarnes:

[[ARO-2994] Propagate errors of GenevaLogging controller](https://issues.redhat.com/browse/ARO-2994)
under epic [[ARO-1627] Propagate Errors and Statuses of ARO Controllers to ARO Cluster Operator](https://issues.redhat.com/browse/ARO-1627)

Adds a `GenevaLoggingReady` "check" condition to the `cluster.aro.openshift.io/v1alpha1` singleton object that exposes controller errors.

### Test plan for issue:

Manual testing using a custom operator image.

None that I'm aware of.

### DIFF
--- a/pkg/operator/controllers/genevalogging/genevalogging_controller.go
+++ b/pkg/operator/controllers/genevalogging/genevalogging_controller.go
@@ -102,9 +102,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	err = r.ensureResources(ctx, instance)
 	if err != nil {
 		r.Log.Error(err)
+		r.SetDegraded(ctx, err)
 		return reconcile.Result{}, err
 	}
 
+	r.ClearConditions(ctx)
 	return reconcile.Result{}, nil
 }
 
--- a/pkg/operator/controllers/genevalogging/genevalogging_controller.go
+++ b/pkg/operator/controllers/genevalogging/genevalogging_controller.go
@@ -51,6 +51,40 @@ func NewReconciler(log *logrus.Entry, client client.Client, dh dynamichelper.Int
 	}
 }
 
+func (r *Reconciler) ensureResources(ctx context.Context, instance *arov1alpha1.Cluster) error {
+	operatorSecret := &corev1.Secret{}
+	operatorSecretName := types.NamespacedName{
+		Namespace: operator.Namespace,
+		Name:      operator.SecretName,
+	}
+	err := r.Client.Get(ctx, operatorSecretName, operatorSecret)
+	if err != nil {
+		return err
+	}
+
+	resources, err := r.resources(ctx, instance, operatorSecret.Data[GenevaCertName], operatorSecret.Data[GenevaKeyName])
+	if err != nil {
+		return err
+	}
+
+	err = dynamichelper.SetControllerReferences(resources, instance)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/operator/controllers/genevalogging/genevalogging_controller.go
```diff
@@ -34,66 +36,104 @@ const (
...
+	if !instance.Spec.OperatorFlags.GetSimpleBoolean(controllerEnabled) {
+		r.log.Debug("controller is disabled")
+		return r.reconcileDisabled(ctx)
+	}
+
+	r.log.Debug("running")
+
+	reconcileErr := r.ensureResources(ctx, instance)
+
+	err = conditions.SetCondition(ctx, r.client, r.condition(reconcileErr), r.role)
```
SudoBrendan, line 104: I think for consistency, the way to implement this would be to use the helpers/interfaces Tanmay wrote that we've been using for all other controllers, which will eliminate some duplicate code here. We'd be looking to have our Reconciler implement the `AROController`, and then use the `r.SetDegraded(ctx, err)` helper here.

ref:
https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/base/aro_controller.go

SudoBrendan (changes requested):
This code looks good, but I think we should follow our standard pattern for propoating the controller status conditions by using `base.AROController`; it'll save us some duplicate code and make things easier to maintain over time.

mbarnes:
I rewrote my changes from scratch since `AROController` handles conditions quite a bit differently than my initial attempt.  Hardest part of [ARO-2994](https://issues.redhat.com/browse/ARO-2994) was understanding what was being asked: the breadcrumb there was misleading, but I gleaned the pattern from the other controllers.

tsatam:
> I rewrote my changes from scratch since `AROController` handles conditions quite a bit differently than my initial attempt. Hardest part of [ARO-2994](https://issues.redhat.com/browse/ARO-2994) was understanding what was being asked: the breadcrumb there was misleading, but I gleaned the pattern from the other controllers.

The PR linked in the breadcrumb is admittedly not the most up-to-date example of how to do this - that initial PR predates the existence of the base controller and helper functions. Perhaps we should update any future stories to point to a more recent example, such as this PR itself?

pkg/operator/controllers/genevalogging/genevalogging_test.go
```diff
@@ -154,8 +155,10 @@ func TestGenevaLoggingDaemonset(t *testing.T) {
 			}
```
tsatam, line 155: I think we should update these tests to also assert the expected conditions are set for the controller. PRs such as https://github.com/Azure/ARO-RP/pull/2965 have an example of how to use the helper functions defined for this purpose.

tsatam (changes requested):
Changes to the controller LGTM, I like the refactor to split the "work" into a simple function and let Reconcile just focus on controller things. 

I think we should capture our expected conditions in the tests, similar to other PRs such as https://github.com/Azure/ARO-RP/pull/2965.

mbarnes:
Added condition checks to the unit tests, although there's currently no test cases that cause an error.

tsatam (approved):
Tests LGTM. The lack of test coverage over error scenarios is concerning, but that can be a follow-up effort.

SudoBrendan (approved):

PR #3222, opened 2023-10-18
Return errors from AuthorizationRetryingAction if timeout is reached

cadenmarchese:

Fixes ARO-4024

Currently, we have a lot of `timed out waiting for the condition` error messages during installs which are associated with timing out actions during validateResources. validateResources is wrapped in a `wait.PollImmediateUntil` as an AuthorizationRetryingAction step. We should not store the timeout error from `wait.PollImmediateUntil`, and instead retain the real error in all cases, and return that if timeout is reached.

Also returns a cloud error if access token can not be requested.

### Test plan for issue:

I don't have a plan to test this at this time - I am open to ideas on how we can test AuthorizationRetryingActions or simulate a timeout here somehow.

No

### DIFF
--- a/pkg/util/steps/refreshing.go
+++ b/pkg/util/steps/refreshing.go
@@ -37,8 +37,11 @@ type authorizationRefreshingActionStep struct {
 }
 
 func (s *authorizationRefreshingActionStep) run(ctx context.Context, log *logrus.Entry) error {
-	var pollInterval time.Duration
-	var retryTimeout time.Duration
+	var (
+		err          error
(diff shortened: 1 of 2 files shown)


tsatam (approved):
The code LGTM, I agree with the testing plan to validate changes in Int and/or Canary.

SrinivasAtmakuri (approved):
LGTM!

Shivkumar13 (approved):

petrkotas (approved):
This is straightforward change. LGTM

PR #3223, opened 2023-10-19
Fix panic on addressPrefix vs addressPrefixes on subnet

bennerv:

ARO-4516

We currently panic & dequeue 6 times when trying to fetch an IP address in a subnet where AddressPrefixes is used.  

### Test plan for issue:

Unit tests cover this case in the future.  

nope

### DIFF
--- a/pkg/util/subnet/subnet.go
+++ b/pkg/util/subnet/subnet.go
@@ -75,7 +75,14 @@ func (m *manager) GetHighestFreeIP(ctx context.Context, subnetID string) (string
 		return "", err
 	}
 
-	_, subnetCIDR, err := net.ParseCIDR(*subnet.AddressPrefix)
+	// grab the first addresPrefix in the subnet
+	var subnetCIDR *net.IPNet
+	if subnet.AddressPrefix == nil {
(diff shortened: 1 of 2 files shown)


tsatam (approved):

pkg/util/subnet/subnet.go
```diff
@@ -75,7 +75,14 @@ func (m *manager) GetHighestFreeIP(ctx context.Context, subnetID string) (string
 		return "", err
 	}
 
-	_, subnetCIDR, err := net.ParseCIDR(*subnet.AddressPrefix)
+	// grab the first addresPrefix in the subnet
+	var subnetCIDR *net.IPNet
+	if subnet.AddressPrefix == nil {
+		_, subnetCIDR, err = net.ParseCIDR((*subnet.AddressPrefixes)[0])
```
s-amann, line 81: is this always one or the other? and addressprefixes always has at least one element?

bennerv: yes

SudoBrendan (approved):

PR #3226, opened 2023-10-20
Fix e2e test for azurefile-csi storage class that uses ARO-managed storage account on OCP 4.11

kimorris27:

https://issues.redhat.com/browse/ARO-4488

This is a follow-up PR on the hotfix because we had trouble getting the test working before the hotfix release. We want this test to work so we have continuous validation that our code can handle this situation properly.

**Edit:** I unfortunately found that the reason that the reason the test wasn't working in CI is because CI clusters are FIPS-enabled, and Azure File storage does not work on FIPS-enabled clusters. So the plan for this PR is changing from fixing the test -> leaving the test pending but leaving comments saying it should be enabled when/if FIPS becomes toggleable post-install.

If you look at the E2E run for commit [8e39971](https://github.com/Azure/ARO-RP/pull/3226/commits/8e399716140d6389762320f1428b5dd561e40518), you can see that the test passes with FIPS disabled (though the overall E2E result was a failure for some unrelated reason).

I also fixed an issue related to the way the test gets the permanent volume - it was using the wrong name.

### Test plan for issue:

The PR is for a test 😋

No

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -445,7 +445,7 @@ func (c *Cluster) createCluster(ctx context.Context, vnetResourceGroup, clusterN
 			ClusterProfile: api.ClusterProfile{
 				Domain:               strings.ToLower(clusterName),
 				ResourceGroupID:      fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", c.env.SubscriptionID(), "aro-"+clusterName),
-				FipsValidatedModules: api.FipsValidatedModulesEnabled,
+				FipsValidatedModules: api.FipsValidatedModulesDisabled,
 				Version:              osClusterVersion,
 			},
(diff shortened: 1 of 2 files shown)


kimorris27:
I determined that the reason the test wasn't working in CI is because CI clusters are FIPS-enabled, and Azure File storage does not work on FIPS-enabled clusters. So the plan for this PR is changing from fixing the test -> leaving the test pending but leaving comments saying it should be enabled when/if FIPS becomes toggleable post-install.

If you look at the E2E run for commit [8e39971](https://github.com/Azure/ARO-RP/pull/3226/commits/8e399716140d6389762320f1428b5dd561e40518), you can see that the test passes with FIPS disabled (though the overall E2E result was a failure for some unrelated reason).

cadenmarchese (approved):
lgtm!

PR #3227, opened 2023-10-20
Fix dnsmasq restart script for Azure Host Servicing workaround

s-fairchild:

Issues that arose in [this ICM](https://portal.microsofticm.com/imp/v3/incidents/incident/433805667/summary)

Fixes

NA

systemctl try-restart no longer supports --wait. --wait can now only be used with start or restart.
This PR changes the logic to check if dnsmasq is running before attempting to restart, rather than relying on `try-restart`.
There is also a sanity check to see if the files we are attempting to fix exist.

### DIFF
--- a/pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
+++ b/pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
@@ -27,12 +27,29 @@ log_dns_files() {
     log "$(echo -n "/etc/resolv.conf.dnsmasq file metadata: ") $(ls -lZ /etc/resolv.conf.dnsmasq)"
 }
 
+sanity_check() {
+    if [[ ! -f /etc/resolv.conf ]]; then
+        log "/etc/resolv.conf is missing. Sanity check failed."
+        exit 1
+    elif [[ ! -f /etc/resolv.conf.dnsmasq ]]; then
+        log "/etc/resolv.conf.dnsmasq is missing. Sanity check failed."
+        exit 1
+    fi
+}
+
 if [[ $interface == eth* && $action == "up" ]] || [[ $interface == eth* && $action == "down" ]] || [[ $interface == enP* && $action == "up" ]] || [[ $interface == enP* && $action == "down" ]]; then
+    sanity_check
     log "$action happened on $interface, connection state is now $CONNECTIVITY_STATE"
     log "Pre dnsmasq restart file information"
     log_dns_files
     log "restarting dnsmasq now"
-    if systemctl try-restart dnsmasq --wait; then
+    
+    if systemctl is-active dnsmasq --quiet; then
+        log "dnsmasq is inactive, refusing to restart"
+        exit 1
+    fi
+
+    if systemctl restart dnsmasq --wait; then
         log "dnsmasq successfully restarted"
         log "Post dnsmasq restart file information"
         log_dns_files
--- a/pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
+++ b/pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
@@ -1,5 +1,5 @@
 {{ define "99-dnsmasq-restart" }}
-#!/bin/sh
+#!/bin/bash
 # This is a NetworkManager dispatcher script to restart dnsmasq
 # in the event of a network interface change (e. g. host servicing event https://learn.microsoft.com/en-us/azure/developer/intro/hosting-apps-on-azure)
 # this will restart dnsmasq, reapplying our /etc/resolv.conf file and overwriting any modifications made by NetworkManager
@@ -32,11 +32,15 @@ restart_dnsmasq() {
     exit 0
 }
 
+# log logs all provided input to journald
+# stack_level can be set to change the function name used as the syslog identifier postfix
+# The default stack_level is the function calling log
 log() {
-    logger -i "$0" -t '99-DNSMASQ-RESTART SCRIPT' "$@"
+    logger -i "$0" -t "99-DNSMASQ-RESTART ${FUNCNAME[${stack_level:-1}]}" "$@"
 }
 
 abort() {
+    stack_level=2
     log "$@"
     exit 1
 }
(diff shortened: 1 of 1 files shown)


pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
```diff
@@ -27,12 +27,29 @@ log_dns_files() {
...
+
 if [[ $interface == eth* && $action == "up" ]] || [[ $interface == eth* && $action == "down" ]] || [[ $interface == enP* && $action == "up" ]] || [[ $interface == enP* && $action == "down" ]]; then
+    sanity_check
     log "$action happened on $interface, connection state is now $CONNECTIVITY_STATE"
     log "Pre dnsmasq restart file information"
     log_dns_files
     log "restarting dnsmasq now"
-    if systemctl try-restart dnsmasq --wait; then
+    
+    if systemctl is-active dnsmasq --quiet; then
```
s-amann, line 47: just for my understanding, we are expecting dnsmasq service to not be running due to the host servicing, so if its active then we exit out? should we try to kill the service and then restart it?

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
```diff
@@ -27,12 +27,29 @@ log_dns_files() {
...
 
+sanity_check() {
+    if [[ ! -f /etc/resolv.conf ]]; then
+        log "/etc/resolv.conf is missing. Sanity check failed."
+        exit 1
+    elif [[ ! -f /etc/resolv.conf.dnsmasq ]]; then
+        log "/etc/resolv.conf.dnsmasq is missing. Sanity check failed."
+        exit 1
+    fi
+}
```
SudoBrendan, line 38: It might be worth leaving a comment in the code on why this is necessary. If the files are missing, what would be next steps to resolve that?

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
s-fairchild: This script is to a fix `/etc/resolv.conf` when it has been modified after `dnsmasq` is running.
IIRC this is to prevent a restart loop of dnsmasq, I'll update this comment once I find it.
@ventifus will also have background knowledge on why it's done this way.

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
s-fairchild: `/etc/resolv.conf.dnsmasq` should only be missing on the bootstrapper. Ideally I'd like to disable this portion of the controller from being applied to the bootstrapper, and this in theory shouldn't happen.

My thinking here is to just exit early incase of the unexpected, rather than letting the script execute and log file not found errors.

On a worker or master node, if `/etc/resolv.conf.dnsmasq` doesn't exist, we probably want to restart dnsmasq anyways? So that the systemd unit `ExecStartPre` executes `aro-dnsmasq-pre.sh`.

I'd like others thoughts on this as well?

hlipsig: +1

ventifus: Another case where /etc/resolv.conf.dnsmasq is missing is on node boot. eth0 comes up before dnsmasq.service, so the script gets called before dnsmasq has run for the first time.

s-fairchild: @ventifus I can update to check if this is the first node boot, then exit without here without performing anything and logging that information. I can also add a check to see if `dnsmasq` has ever been started and log the results appropriately.  

If `/etc/resolv.conf.dnsmasq` doesn't exist **and** `dnsmasq` hasn't been started, this is expected. If `dnsmasq` has been started, and `/etc/resolv.conf.dnsmasq` is missing, we have a problem that's unaccounted for.

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
s-fairchild: @Spencer-z During the node boot process eth0 will come up/down before `dnsmasq` does. This check is to prevent attempting to start `dnsmasq` too early in the boot process.

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
```diff
@@ -27,19 +58,37 @@ log_dns_files() {
...
+    # If this condition is met, dnsmasq will be restarted in main
+    elif [[ ! -f $dnsmasq_resolv ]] && [[ $dnsmasq_state -eq 0 ]]; then
+        log "$dnsmasq_resolv is missing and dnsmasq is active. Something has removed $dnsmasq_resolv. Sanity check failed.
+        Restarting dnsmasq to restore $dnsmasq_resolv"
     fi
-fi
 
-exit 0
+    # dnsmasq won't be running during initial boot, which will trigger this script
+    if [[ ! $dnsmasq_state -eq 0 ]]; then
```
cadenmarchese, line 87: Is this something we want to check first in the sanity_check(), rather than last?

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
```diff
@@ -7,12 +7,43 @@
 interface=$1
 action=$2
 
+main() {
+    if [[ $interface == eth* && $action == "up" ]] || [[ $interface == eth* && $action == "down" ]] || [[ $interface == enP* && $action == "up" ]] || [[ $interface == enP* && $action == "down" ]]; then
+        log "$action happened on $interface, connection state is now $CONNECTIVITY_STATE"
+        sanity_check
```
cadenmarchese, line 13: Super duper nit, but we might want to just call it what it is to avoid some jumping around in the code:
```suggestion
        check_resolv_files
```
Or something along those lines.

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
```diff
@@ -7,12 +7,43 @@
 interface=$1
 action=$2
 
+main() {
+    if [[ $interface == eth* && $action == "up" ]] || [[ $interface == eth* && $action == "down" ]] || [[ $interface == enP* && $action == "up" ]] || [[ $interface == enP* && $action == "down" ]]; then
```
cadenmarchese, line 11: Do we have a manual test plan for this PR / way to simulate this behavior before we release it? I think confidence is key since it takes a lot of effort and communication to get these changes out to customers.

Although the host servicing events are random, maybe we can remove the resolv.conf in a test cluster to confirm expected logging, and restart the NIC with network manager to confirm that dnsmasq service restarts without issues in all supported ARO OCP versions (should only be 2-3 versions).

pkg/operator/controllers/dnsmasq/scripts/99-dnsmasq-restart.gotmpl
s-amann: thank you for the clarification

kimorris27 (commented):
LGTM but I am +1 to testing it, manually if needed, before we release it.

s-amann (approved):

bennerv:
Pending discussion on whether we should merge this or not.

s-fairchild:
Closing this based on our discussion [over what possibly happened](https://docs.google.com/document/d/1Hr70hx1JJBwzeJcTKwW1HCeosm9l_qM6IGIF-G2hUmg/edit?usp=sharing).

I do want to mention this may have been a path issue all along, @ventifus was able to successfully test the original script successfully by adding `PATH` to the environment.

Issue #10, opened 2019-12-17
Prevent provisioned cluster communicating with Red Hat services


jim-minter:
Ensure that insights, telemeter and links to upstream Cincinnati are suitably disabled.

PR #3230, opened 2023-10-24
Use an enum for cluster maintenance states 

niontive:

Fixes https://issues.redhat.com/browse/ARO-4365

Right now, if a PUCM fails, SREs have to run PUCM again to get the cluster to emit "no maintenance ongoing" to customers via resource health. After speaking with @bennerv and @s-amann, we need to support the scenario where SREs move the cluster to "no maintenance ongoing" after (1) failed PUCM and (2) no re-PUCM needed.

In [this state diagram](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/595193/Maintenance-Signals-State-Diagram?anchor=new-state-diagram), this PR implements transition #10. 

### Test plan for issue:

- Added unit tests
- Will test in INT prior to merge

ADO [doc](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/595193/Maintenance-Signals-State-Diagram?anchor=new-state-diagram)


niontive:
Comments from code review session:
- Best to not use PUCM terminology in code. Use something like "admin update". Sync with Amber on this.

pkg/api/admin/openshiftcluster.go
```diff
@@ -82,13 +82,29 @@ const (
...
-	MaintenanceTaskRenewCerts  MaintenanceTask = "CertificatesRenewal"
-	MaintenanceTaskPucmPending MaintenanceTask = "PucmPending"
+	MaintenanceTaskEverything MaintenanceTask = "Everything"
+	MaintenanceTaskOperator   MaintenanceTask = "OperatorUpdate"
+	MaintenanceTaskRenewCerts MaintenanceTask = "CertificatesRenewal"
+
+	// Maintenance tasks for updating customer maintenance signals
+	// None signal should only be used when (1) admin update fails and (2) SRE fixes the failed admin update without running another admin updates
+	// Admin update success should automatically set the cluster into None state
+	MaintenanceTaskPending MaintenanceTask = "PucmPending"
```
s-amann, line 106: Is it possible to change this to `MaintenancePending`?

niontive: We can, but I'll need to do a corresponding code change in our PowerShell script: [link](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO-Scripts?path=/powerShellActions/pucm_pending.ps1&version=GBmaster&line=157&lineEnd=158&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

Since we need to do an RP release to deploy this code change, we'll need to coordinate code changes as we do maintenance prior to an RP release.  I say we:
1. Change the maintenance task name to `MaintenancePending` as you suggested.
2. We'll do maintenance pending + RP release.
3. After the next RP release, I'll create a PR to update the PowerShell script.

Thoughts?

pkg/frontend/openshiftcluster_putorpatch_test.go
```diff
@@ -679,8 +690,202 @@ func TestPutOrPatchOpenShiftClusterAdminAPI(t *testing.T) {
...
+					MasterProfile: admin.MasterProfile{
+						EncryptionAtHost: admin.EncryptionAtHostDisabled,
+					},
+					OperatorFlags:    admin.OperatorFlags(api.DefaultOperatorFlags()),
+					MaintenanceState: admin.MaintenanceStatePlanned,
+				},
+			},
+		},
+		{
+			name: "patch a cluster with maintenance none request",
```
kimorris27, line 826: In addition to this test case where the request causes the `MaintenanceState` to transition from empty -> none, how about one where it transitions from planned -> none?

pkg/monitor/cluster/maintenance.go
```diff
@@ -10,75 +10,59 @@ import (
...
 
 const (
-	pucmNone      pucmState = "none"
-	pucmPending   pucmState = "pending"
-	pucmPlanned   pucmState = "planned"
-	pucmUnplanned pucmState = "unplanned"
+	none      maintenanceState = "none"
+	pending   maintenanceState = "pending"
+	planned   maintenanceState = "planned"
+	unplanned maintenanceState = "unplanned"
```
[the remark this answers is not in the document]
niontive: Previously, there wasn't a 1:1 mapping between maintenance states and metrics types. There's another bug fix I'm going to work on, so I can take another look at this in my next PR.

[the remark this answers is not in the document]
niontive: For now, I'm ok with decoupling our metric names from our maintenance states names.

pkg/frontend/openshiftcluster_putorpatch_test.go
```diff
@@ -1119,7 +1119,6 @@ func TestPutOrPatchOpenShiftClusterAdminAPI(t *testing.T) {
 		{
 			name: "patch a none maintenance state cluster with maintenance unplanned request",
 			request: func(oc *admin.OpenShiftCluster) {
-				oc.Properties.MaintenanceTask = admin.MaintenanceTaskEverything
```
kimorris27: I'm a little confused by the fact that you had to remove this line to fix the test. To help me understand: what's the conceptual difference between sending an empty PATCH request and sending a PATCH request that includes `MaintenanceTaskEverything`?

niontive: It doesn't matter. When you don't set the maintenance task, the default is everything: https://github.com/Azure/ARO-RP/blob/9a9edacf6be058d0832c5827e837ecb15a692e17/pkg/api/defaults.go#L37

kimorris27 (approved):

cadenmarchese (approved):

PR #3231, opened 2023-10-24
Return actionable service principal errors to the user

cadenmarchese:

Continuation of PR #3222 and ARO-4024

- Reverts the change previously made in #3222 to ensureAccessTokenClaims(), since we need the raw Azure error for `AuthorizationRetryingAction` to do the error matching and work as intended.
- Returns the actionable errors related to service principal credentials to the user as CloudError, rather than `timed out waiting for condition` in the RP and Internal Server Error

### Test plan for issue:
- We can test it by ensuring that if provided with invalid client ID or client secret values, `az aro create` will return them to the user

### DIFF
--- a/pkg/util/steps/refreshing.go
+++ b/pkg/util/steps/refreshing.go
@@ -90,15 +90,11 @@ func (s *authorizationRefreshingActionStep) run(ctx context.Context, log *logrus
 	switch err != nil {
 	case azureerrors.IsUnauthorizedClientError(err):
 		return s.servicePrincipalCloudError(
-			"The provided service principal application ID was not found in the tenant. Please ensure that the provided clientID and client secret are correct.",
+			"The provided service principal application (client) ID was not found in the directory (tenant). Please ensure that the provided application (client) id and client secret value are correct.",
 		)
-	case azureerrors.HasAuthorizationFailedError(err):
+	case azureerrors.HasAuthorizationFailedError(err) || azureerrors.IsInvalidSecretError(err):
 		return s.servicePrincipalCloudError(
-			"Authorization using provided credentials failed. Please ensure that the provided clientID and client secret are correct.",
-		)
-	case azureerrors.IsInvalidSecretError(err):
-		return s.servicePrincipalCloudError(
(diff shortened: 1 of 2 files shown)


pkg/util/steps/refreshing.go
```diff
@@ -84,7 +86,23 @@ func (s *authorizationRefreshingActionStep) run(ctx context.Context, log *logrus
...
+		)
+	case azureerrors.HasAuthorizationFailedError(err):
+		return s.servicePrincipalCloudError(
+			"Authorization using provided credentials failed. Please ensure that the provided clientID and client secret are correct.",
+		)
+	case azureerrors.IsInvalidSecretError(err):
+		return s.servicePrincipalCloudError(
+			"Invalid client secret provided. Please ensure that the provided clientID and client secret are correct.",
+		)
+	}
```
bennerv, line 103: I think one of these is legacy from before we were using MSGraph.  I think they're related to the same error message though.  I would just say: `Invalid Client secret`.  May want to rope in Jerome, Oren, or Puneet for customer-facing message though.  I'll ping them.

cadenmarchese: So just to confirm, we want to change the error message from `"Invalid client secret provided. Please ensure that the provided clientID and client secret are correct."` to just `"Invalid client secret"?

cadenmarchese: I changed this to combine these two, and use the error messages recommended by Puneet 👍

s-amann (dismissed):

pkg/util/steps/refreshing.go
```diff
@@ -84,6 +86,19 @@ func (s *authorizationRefreshingActionStep) run(ctx context.Context, log *logrus
 		return true, err
 	}, timeoutCtx.Done())
 
+	// After timeout, return any actionable errors to the user
+	switch err != nil {
```
bennerv, line 90: switch has to be on a var.  we're getting a nil pointer exception on this.   I'd just suggest to use if else statements.

s-amann (approved):

bennerv (approved):

PR #3232, opened 2023-10-25
etcd-renew - Add extra validation and remove IsLessThanMinDuration

SrinivasAtmakuri:
Added extra validation to verify if etcd certificates are renewed Also, removed isLessThanMinDuration code block so we can even renew clusters whose expiry is more than 6 months duration

Fixes

### DIFF
--- a/pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
+++ b/pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
@@ -337,7 +337,7 @@ func (e *etcdrenew) validateEtcdCertsRenewed(ctx context.Context) error {
 	}
 
 	if isError {
-		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "etcd certificates renewal not successful, as atleast one or all certificates are not renewed")
+		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "etcd certificates renewal not successful, as at least one or all certificates are not renewed")
 	}
 
 	e.log.Infoln("etcd certificates are successfully renewed")
--- a/pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
+++ b/pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
@@ -294,6 +294,9 @@ func (e *etcdrenew) validateEtcdCertsExistsAndExpiry(ctx context.Context) error 
 		if err != nil {
 			return err
 		}
+		if len(certData) < 1 {
+			return fmt.Errorf("invalid cert data when parsing secret: %s", secret.Name)
+		}
 		if utilcert.IsCertExpired(certData[0]) {
 			return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "secret %s is already expired, quitting.", secretname)
 		}
--- a/pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
+++ b/pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
@@ -41,7 +41,6 @@ type etcdrenew struct {
 	k             adminactions.KubeActions
 	doc           *api.OpenShiftClusterDocument
 	secretNames   []string
-	mode          string
 	backupSecrets map[string][]byte
 	lastRevision  int32
 	timeout       time.Duration
@@ -52,8 +51,9 @@ var etcdOperatorControllerConditionsExpected = map[string]operatorv1.ConditionSt
 }
 
 var etcdOperatorConditionsExpected = map[configv1.ClusterStatusConditionType]configv1.ConditionStatus{
-	configv1.OperatorAvailable: configv1.ConditionTrue,
-	configv1.OperatorDegraded:  configv1.ConditionFalse,
+	configv1.OperatorAvailable:   configv1.ConditionTrue,
+	configv1.OperatorProgressing: configv1.ConditionFalse,
+	configv1.OperatorDegraded:    configv1.ConditionFalse,
 }
 
 func (f *frontend) postAdminOpenShiftClusterEtcdCertificateRenew(w http.ResponseWriter, r *http.Request) {
@@ -150,8 +150,13 @@ func (e *etcdrenew) run(ctx context.Context) error {
 	}
 
 	e.log.Infoln("Etcd certificates are renewed and new revision is applied, verifying.")
-	e.mode = "renewed"
-	return e.validateEtcdAndBackupDeleteSecretOnFlagSet(ctx, false)
+	err = e.validateEtcdAndBackupDeleteSecretOnFlagSet(ctx, false)
+	if err != nil {
+		return err
+	}
+
+	// validates if the etcd certificates are renewed
+	return e.validateEtcdCertsRenewed(ctx)
 }
 
 func (f *frontend) _postAdminOpenShiftClusterEtcdCertificateRenew(ctx context.Context, resourceID string, log *logrus.Entry, timeout time.Duration) error {
@@ -177,7 +182,6 @@ func (f *frontend) _postAdminOpenShiftClusterEtcdCertificateRenew(ctx context.Co
 		k:             k,
 		doc:           doc,
 		secretNames:   nil,
-		mode:          "",
 		backupSecrets: make(map[string][]byte),
 		lastRevision:  0,
 		timeout:       timeout,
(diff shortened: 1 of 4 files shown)


SrinivasAtmakuri:
Currently testing these changes using a test cluster.

petrkotas (approved):
Nice simplification @srinivasAtmakuri

petrkotas (dismissed):
Nice simplification @srinivasAtmakuri

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -299,6 +300,44 @@ func (e *etcdrenew) validateEtcdCertsExistsAndExpiry(ctx context.Context) error
 	return nil
 }
 
+func (e *etcdrenew) validateEtcdCertsRenewed(ctx context.Context) error {
```
s-amann, line 303: I don't see any unit tests for this new method.

s-amann (dismissed):
can we add tests to ensure `validateEtcdCertsRenewed` works when 1, all, and none of the certs are renewed?

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -299,6 +303,47 @@ func (e *etcdrenew) validateEtcdCertsExistsAndExpiry(ctx context.Context) error
...
+		// etcd operator renews certificates for another 3 years, 1000+ days (3*365)
+		e.log.Infof("certificate '%s' expiration date is '%s'", secretname, certData[0].NotAfter)
+		if utilcert.DaysUntilExpiration(certData[0]) < 1000 {
+			isError = true
+			e.log.Errorf("certificate %s is not renewed successfully.", secretname)
+		}
+	}
+
+	if isError {
+		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "etcd certificates renewal not successful, as atleast one or all certificates are not renewed")
```
s-amann, line 340:
```suggestion
		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "etcd certificates renewal not successful, as at least one or all certificates are not renewed")
```

kimorris27 (commented):
Nice work! I agree with @s-amann's request for unit tests for the new method but don't have any other comments.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
SrinivasAtmakuri: Added a new test case. Thanks.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
SrinivasAtmakuri: Made the change

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
kimorris27: The added test cases look good, but I was expecting a new test function for `validateEtcdCertsRenewed` (i.e., `TestValidateEtcdCertsRenewed`, to _just_ test this single method). Are you open to adding that as well?

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -299,6 +303,47 @@ func (e *etcdrenew) validateEtcdCertsExistsAndExpiry(ctx context.Context) error
 	return nil
 }
 
+func (e *etcdrenew) validateEtcdCertsRenewed(ctx context.Context) error {
```
AldoFusterTurpin, line 306: I see that there is a commit for adding a new test to pkg/frontend/admin_openshiftcluster_etcdcertificaterenew_test.go
but I think it would be a better idea to add a unit test for this function.

The problem is that we are testing the functionality of the business logic (certs validation) creating the whole infra structs and mocks when it could be simplified a bit.

From a testing perspective, in validateEtcdCertsRenewed(...) you could first take all the secrets, store them in an slice, or in a map (from secretName to secret value) and then iterate over the map and do the whole validation. Doing that, we make the function more "pure" avoiding mocking e.k and we can just focus on the inputs and outputs and not side effects. We can also, if we want, get a counter of how many certs were not correctly renewed or even the names of those. The important part is making the function more testable without the mock overhead, in the below example I use the counter variant but we can use whatever we want (like names using a slice of strings for the invalid ones).

something like
```go
//definition of your original function
func (e *etcdrenew) validateEtcdCertsExistsAndExpiry(ctx context.Context) error { 
	e.log.Infoln("validating if etcd certs exists and expiry")

	for _, secretname := range e.secretNames {
		e.log.Infof("validating secret %s", secretname)
		cert, err := e.k.KubeGet(ctx, "Secret", namespaceEtcds, secretname)
		if err != nil {
			return err
		}

		var u unstructured.Unstructured
		var secret corev1.Secret
		if err = json.Unmarshal(cert, &u); err != nil {
			return err
		}
		err = kruntime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &secret)
		if err != nil {
			return err
		}
		_, certData, err := utilpem.Parse(secret.Data[corev1.TLSCertKey])
		if err != nil {
			return err
		}
		if utilcert.IsCertExpired(certData[0]) {
			return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "secret %s is already expired, quitting.", secretname)
		}
	}
	e.log.Infoln("Etcd certs exits, are not expired")

	return nil
}

// this is the pure function we really want to test without the mocks
func (e *etcdrenew) validateEtcdCertsRenewed(ctx context.Context) error { 
	e.log.Infoln("validating if etcd certs are renewed")

	certs := make(map[string][]byte)
	for _, secretname := range e.secretNames {
		e.log.Infof("getting secret %s", secretname)
		cert, err := e.k.KubeGet(ctx, "Secret", namespaceEtcds, secretname)
		if err != nil {
			return err
		}
		certs[secretname] = cert
	}

	notRenewedCerts, err := getNotRenewedCerts(e.log, certs)
	if err != nil {
		return err
	}

	if notRenewedCerts > 0 {
		errorMsg := fmt.Sprintf("%v certificates were not renewed", notRenewedCerts)
		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", errorMsg)
	}

	e.log.Infoln("etcd certificates are successfully renewed")
	return nil
}
```
You will need "encodeSecret" from ~/git/ARO-RP/pkg/frontend/admin_openshiftcluster_etcdcertificaterenew_test.go to test the previous function.

So the main idea is to try to avoid testing functionality in code that includes mocks and just use the mocks to ensure that given X I get Y, but NOT given X I get Y and the functionality should be Z, that last thing should be tested independently from the mocks if possible. 

I think in that way our codebase would be much healthier as we would need less mocks to test business logic.
🙂

AldoFusterTurpin (changes requested):
Thank you for this! I just left some optional suggestions and one important comment about testability, in case you find it useful.

Thanks! 
🙂

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -287,9 +294,6 @@ func (e *etcdrenew) validateEtcdCertsExistsAndExpiry(ctx context.Context) error
 		if err != nil {
 			return err
 		}
-		if !utilcert.IsLessThanMinimumDuration(certData[0], utilcert.DefaultMinDurationPercent) && e.mode != "renewed" {
-			return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "secret %s is not near expiry, quitting.", secretname)
-		}
```
jaitaiwan, line 292: Would be good to understand why this was removed?

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -299,6 +303,47 @@ func (e *etcdrenew) validateEtcdCertsExistsAndExpiry(ctx context.Context) error
 	return nil
 }
 
+func (e *etcdrenew) validateEtcdCertsRenewed(ctx context.Context) error {
+	e.log.Infoln("validating if etcd certs are renewed")
+	isError := false
+
+	for _, secretname := range e.secretNames {
+		e.log.Infof("validating secret %s", secretname)
+		cert, err := e.k.KubeGet(ctx, "Secret", namespaceEtcds, secretname)
```
jaitaiwan, line 312: Perhaps a retry on this would be good?

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -299,6 +303,47 @@ func (e *etcdrenew) validateEtcdCertsExistsAndExpiry(ctx context.Context) error
...
+		if err != nil {
+			return err
+		}
+		_, certData, err := utilpem.Parse(secret.Data[corev1.TLSCertKey])
+		if err != nil {
+			return err
+		}
+
+		// etcd operator renews certificates for another 3 years, 1000+ days (3*365)
+		e.log.Infof("certificate '%s' expiration date is '%s'", secretname, certData[0].NotAfter)
```
jaitaiwan, line 332: `certData[0]` is a potential index out of range error. We're not testing to make sure there is at least 1 item in the certData array.

jaitaiwan (dismissed):
There's a single place where there's a potential for an index out of range panic to happen. Once that's resolved I vote we merge and immediately do another PR to encompass the feedback received here. I don't think any of the feedback is a merge blocker for getting the functionality in the release.

jaitaiwan:
I was testing out the push feature before running tests locally to see if my code had worked, but I've added the catch for the potential index panic, so if the CI passes this PR is ready to merge.

yjst2012 (approved):
Same as @hlipsig with the believe of this brings more benefits, I'm approving this PR now, @SrinivasAtmakuri pls address the suggestions in a follow up session, thank you.

PR #3236, opened 2023-10-26
fix docs: puslish aro operator to quay

mbukatov:

No issue/task related to this problem.

Makefile targets `publish-image-aro*` are using env variable `ARO_IMAGE` as a destination, so setting just `DST_QUAY` variable (as currently explained in the docs) is not enough. Updating `ARO_IMAGE`  accordingly fixed the problem.

### Test plan for issue:

No further testing needed, I just followed the procedure to push my locally build aro image to quay.


bennerv (approved):

bennerv:
e2e failed but this is docs update - merging.

PR #3237, opened 2023-10-26
Default installation target to 4.12

bennerv:

[ARO-2720](https://issues.redhat.com/browse/ARO-2720)
[ARO-4306](https://issues.redhat.com/browse/ARO-4306)

Since we merged a fix for #3220, we can now change the default install target to 4.12.  

### Test plan for issue:

e2e + we've already tested this in production and it's available as an installation target.  

No

### DIFF
--- a/pkg/util/version/const.go
+++ b/pkg/util/version/const.go
@@ -31,7 +31,7 @@ type Stream struct {
 }
 
 // DefaultMinorVersion describes the minor OpenShift version to default to
-var DefaultMinorVersion = 11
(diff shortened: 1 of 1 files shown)


bennerv:
Added next-release as we're close to pushing out 4.13, which starts EoL of 4.11 which needs to no longer be the default install version.  

Additionally we believe the fix in #3220 will fix i/o timeout installation issues.

s-amann (approved):

PR #3238, opened 2023-10-26
Remove 4.10 as an installation target

bennerv:

Fixes https://issues.redhat.com/browse/ARO-4573

We can now remove 4.10 as an installation target.  The support fell off in September, but we've held onto it because of FIPS related things.  

### Test plan for issue:

e2e, but no-op since 4.10 is no longer available.  

nope


petrkotas (approved):
FArily straightforward. tHanks

s-amann (approved):

PR #3239, opened 2023-10-26
add unit tests for multiple IP aro create

tony-schndr:
Adds unit tests for adding multiple managed IPs to the public load balancer during arm template generation

Refactored public load balancer ARM template generation by isolating it into a separate function.  Added unit tests for the new function that tests the changes in #3110 

### Test plan for issue:

Unit tests
Manual test: 5 managed IPs, API server public
Manual test: 5 managed IPs, API Server private
Manual test: 1 managed IP, API Server private
E2E cluster creation: 1 managed IP, API Server public

No

### DIFF
--- a/pkg/cluster/deploybaseresources.go
+++ b/pkg/cluster/deploybaseresources.go
@@ -154,6 +154,13 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
 	// Create a public load balancer routing if needed
 	if m.doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType == api.OutboundTypeLoadbalancer {
 		m.newPublicLoadBalancer(ctx, &resources)
+
+		// If the cluster is public we still want the default public IP address
+		if m.doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPublic {
+			resources = append(resources,
+				m.networkPublicIPAddress(azureRegion, infraID+"-default-v4"),
+			)
+		}
 	}
 
 	if m.doc.OpenShiftCluster.Properties.FeatureProfile.GatewayEnabled {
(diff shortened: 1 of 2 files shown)


SudoBrendan (approved):

pkg/cluster/deploybaseresources.go
```diff
@@ -201,6 +175,39 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
...
+	}
+	if m.doc.OpenShiftCluster.Properties.NetworkProfile.LoadBalancerProfile.ManagedOutboundIPs != nil {
+		for i := len(outboundIPs); i < m.doc.OpenShiftCluster.Properties.NetworkProfile.LoadBalancerProfile.ManagedOutboundIPs.Count; i++ {
+			ipName := genManagedOutboundIPName()
+			*resources = append(*resources, m.networkPublicIPAddress(azureRegion, ipName))
+			outboundIPs = append(outboundIPs, api.ResourceReference{ID: m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID + "/providers/Microsoft.Network/publicIPAddresses/" + ipName})
+		}
+	}
+	m.patchEffectiveOutboundIPs(ctx, outboundIPs)
+	// Normal private clusters still need a public load balancer
```
kimorris27, line 199: I realize that this comment was just moved into a function and you didn't change it as part of this PR, but would you mind clarifying what it means either in this discussion or directly in the code (as appropriate)? It reads to me as if this code has been moved from a context where the comment made more sense.

pkg/cluster/deploybaseresources.go
```diff
@@ -201,6 +175,39 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
...
+	*resources = append(*resources,
+		m.networkPublicLoadBalancer(azureRegion, outboundIPs),
+	)
+	// If the cluster is public we still want the default public IP address
+	if m.doc.OpenShiftCluster.Properties.IngressProfiles[0].Visibility == api.VisibilityPublic {
```
kimorris27, line 204: I didn't see a test case for this scenario (public ingress -> default PIP needed). If it makes sense to you, you could add one.

pkg/cluster/deploybaseresources.go
tony-schndr: Agreed, I removed this comment.  I think it is clear in the code why we need a public load balancer (outbound routing).

pkg/cluster/deploybaseresources.go
tony-schndr: Now that you point it out, it doesn't make sense to be in the `newPublicLoadBalancer` function.  It doesn't get added to the load balancer in the ARM template so I removed it from the function I created.

kimorris27 (approved):

SudoBrendan (approved):

cadenmarchese (approved):
lgtm!

PR #3241, opened 2023-10-26
Fix CheckAccess unmarshal bug

nwnt:
From the schema, the status code returned is an integer.

Fixes: https://msazure.visualstudio.com/One/_workitems/edit/15937293#26804649

Currently there's an error when unmarshaling the body of the error payload returned by CheckAccess.  The schema of the error indicates that the field is an integer while the current struct uses a string.

### Test plan for issue:
N/A This fixes the unmarshal error returned with message *"Unexpected error when calling CheckAccessV2: json: cannot unmarshal number into Go struct field CheckAccessErrorResponse.statusCode of type string"*

N/A

### DIFF
--- a/pkg/util/azureclient/authz/remotepdp/client.go
+++ b/pkg/util/azureclient/authz/remotepdp/client.go
@@ -6,6 +6,7 @@ package remotepdp
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
@@ -87,6 +88,6 @@ func newCheckAccessError(r *http.Response) error {
 	return &azcore.ResponseError{
 		StatusCode:  r.StatusCode,
 		RawResponse: r,
-		ErrorCode:   checkAccessError.Message,
+		ErrorCode:   fmt.Sprint(checkAccessError.StatusCode),
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/util/azureclient/authz/remotepdp/client.go
```diff
@@ -87,6 +88,6 @@ func newCheckAccessError(r *http.Response) error {
 	return &azcore.ResponseError{
 		StatusCode:  r.StatusCode,
 		RawResponse: r,
-		ErrorCode:   checkAccessError.Message,
+		ErrorCode:   fmt.Sprint(checkAccessError.StatusCode),
```
bennerv, line 91: is this intentional?  Changing errorcode to point to the status code instead of the message on the error?

bennerv (approved):
lgtm - just one question

pkg/util/azureclient/authz/remotepdp/client.go
nwnt: Hmm I was kinda in two minds about this.  In the end, I made this change because according the description of ResponseError field shown below.  I assume the code might be used somewhere like being returned from cli, and it should be a code not a message.  The raw response is still included in the error so the message is not actually lost.

image

But let me know what you think. I can change it back if returning the message is beter.

petrkotas: I tend to agree with Nont on this. The documentation says code. I would try it and see if it is bad or not.

petrkotas (approved):
LGTM

petrkotas (approved):
LGTM

PR #3243, opened 2023-10-27
Hive Minimal Install Annotation on ClusterDeployment

bennerv:

Use hive minimal install mode (or set ourselves up for).

https://github.com/openshift/hive/pull/2157

As part of installation through hive, the following is done:
1. Hive pulls the ocp-release images
2. Hive pulls the oc-cli image (discovers this from the ocp-release image)
    - this is used for must-gather, which we' don't configure
3. Hive pulls the installer image (discovers from the ocp-release image unless `installerImageOverride` is set)
    - we set `installerImageOverride` in ARO. 

That means that if we allow the installation of older versions, older images are pulled down, which we don't necessarily want.  

### Test plan for issue:

This will have no affect until we get to a hive version which has https://github.com/openshift/hive/pull/2157/files merged in. 

Testing will be manual and in INT environment. 

no

### DIFF
--- a/pkg/hive/install.go
+++ b/pkg/hive/install.go
@@ -13,6 +13,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/Azure/go-autorest/autorest/to"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	hivev1azure "github.com/openshift/hive/apis/hive/v1/azure"
 	corev1 "k8s.io/api/core/v1"
@@ -191,9 +192,6 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 				createdByHiveLabelKey:                "true",
 			},
 			Annotations: map[string]string{
-				"hive.openshift.io/try-install-once":                "true",
-				"hive.openshift.io/cli-domain-from-installer-image": "true",
-
 				// https://github.com/openshift/hive/pull/2157
 				// Will not pull ocp-release and oc-cli images
 				"hive.openshift.io/minimal-install-mode": "true",
@@ -215,6 +213,7 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 				APIServerIPOverride: doc.OpenShiftCluster.Properties.NetworkProfile.APIServerPrivateEndpointIP,
 				APIURLOverride:      fmt.Sprintf("api-int.%s:6443", clusterDomain),
 			},
+			InstallAttemptsLimit: to.Int32Ptr(1),
 			PullSecretRef: &corev1.LocalObjectReference{
 				Name: pullsecretSecretName,
 			},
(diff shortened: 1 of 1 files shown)


bennerv:
@2uasimojo if you want to take a quick glance / sanity check it.

pkg/hive/install.go
```diff
@@ -193,6 +193,10 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 			Annotations: map[string]string{
 				"hive.openshift.io/try-install-once":                "true",
 				"hive.openshift.io/cli-domain-from-installer-image": "true",
```
2uasimojo, line 195: Guess you don't need this anymore.

pkg/hive/install.go
```diff
@@ -193,6 +193,10 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 			Annotations: map[string]string{
 				"hive.openshift.io/try-install-once":                "true",
```
2uasimojo, line 194: (unrelated/latent) Suggest using `[ClusterDeployment.Spec.InstallAttemptsLimit](https://github.com/openshift/hive/blob/903546686f6109e4c0389d72893cf0272a00e2cb/vendor/github.com/openshift/hive/apis/hive/v1/clusterdeployment_types.go#L192)=1` instead. Does the same thing, but is the official/documented/supported mechanism.

2uasimojo (commented):
/lgtm

hawkowl:
I wasn't sure about `"hive.openshift.io/cli-domain-from-installer-image": "true",` being removed, but it appears like that is only if we use ImageSets, which we don't. :+1: from me.

hawkowl (approved):

2uasimojo:
> I wasn't sure about `"hive.openshift.io/cli-domain-from-installer-image": "true",` being removed, but it appears like that is only if we use ImageSets, which we don't. 👍 from me.

ImageSet or releaseImage, both of which are bypassed by this new thing.

pkg/hive/install.go
```diff
@@ -211,6 +213,7 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 				APIServerIPOverride: doc.OpenShiftCluster.Properties.NetworkProfile.APIServerPrivateEndpointIP,
 				APIURLOverride:      fmt.Sprintf("api-int.%s:6443", clusterDomain),
 			},
+			InstallAttemptsLimit: to.Int32Ptr(1),
```
kimorris27, line 216: What's the motivation for this change? I'm just curious since a quick scan over the Hive code made it seem to me like this could potentially increase our install failure rate by never attempting to retry. IDK how many retry-iable  errors our Hive installs hit though.

tsatam: We have never retried failed installs in ARO, this is just changing how we ensure that only one attempt is performed to  (see removed `"hive.openshift.io/try-install-once":                "true"` annotation above)

kimorris27 (approved):

petrkotas (approved):
lgtm

PR #3249, opened 2023-10-30
Skip failing pki test for now

hawkowl:

CI is failing because the PKI tests call an external webservice which is presently down.

### Test plan for issue:

Should fix CI.

don't know?


petrkotas (approved):
LGTM

PR #3254, opened 2023-10-31
Make env know what the service component its running is

hawkowl:

Part of https://issues.redhat.com/browse/ARO-4595

Makes env/ understand what service it is running for, which allows us to standardise some of the "what" about the service running (e.g. knowing what dbtoken permission we're asking for, what logger we want to use, etc)

### Test plan for issue:

CI/E2E

N/A, cleanup

### DIFF
--- a/pkg/env/msiauthorizer.go
+++ b/pkg/env/msiauthorizer.go
@@ -20,14 +20,22 @@ const (
 	MSIContextGateway MSIContext = "GATEWAY"
 )
 
-func (c *core) NewMSITokenCredential(msiContext MSIContext) (azcore.TokenCredential, error) {
+func (c *core) NewMSITokenCredential() (azcore.TokenCredential, error) {
 	if !c.IsLocalDevelopmentMode() {
 		options := c.Environment().ManagedIdentityCredentialOptions()
 		return azidentity.NewManagedIdentityCredential(options)
 	}
+
+	var msiContext string
+	if c.component == COMPONENT_GATEWAY {
+		msiContext = string(MSIContextGateway)
+	} else {
+		msiContext = string(MSIContextRP)
+	}
+
 	tenantIdKey := "AZURE_TENANT_ID"
-	azureClientIdKey := "AZURE_" + string(msiContext) + "_CLIENT_ID"
-	azureClientSecretKey := "AZURE_" + string(msiContext) + "_CLIENT_SECRET"
+	azureClientIdKey := "AZURE_" + msiContext + "_CLIENT_ID"
+	azureClientSecretKey := "AZURE_" + msiContext + "_CLIENT_SECRET"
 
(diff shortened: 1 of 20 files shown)


kimorris27 (commented):
The general concept behind the changes makes sense, but I want to understand it a little better.

In the Jira where you discuss deduplicating things, are you referring to things like https://github.com/Azure/ARO-RP/blob/master/cmd/aro/rp.go#L110-L123 and https://github.com/Azure/ARO-RP/blob/master/cmd/aro/monitor.go#L90-L103 and saying that these changes are a step toward helping us deduplicate that code? What would things look like after the deduplication is complete?

hawkowl:
> The general concept behind the changes makes sense, but I want to understand it a little better.
> 
> In the Jira where you discuss deduplicating things, are you referring to things like [`master`/cmd/aro/rp.go#L110-L123](https://github.com/Azure/ARO-RP/blob/master/cmd/aro/rp.go?rgh-link-date=2023-10-31T21%3A05%3A48Z#L110-L123) and [`master`/cmd/aro/monitor.go#L90-L103](https://github.com/Azure/ARO-RP/blob/master/cmd/aro/monitor.go?rgh-link-date=2023-10-31T21%3A05%3A48Z#L90-L103) and saying that these changes are a step toward helping us deduplicate that code? What would things look like after the deduplication is complete?

https://github.com/Azure/ARO-RP/pull/3210/files#diff-41cd00a4f558bca78edde4a4923b345945f34697605af2490b007c1082d4980e < this is kind of what I aim to do in followups :)

kimorris27 (dismissed):

kimorris27 (approved):

jaitaiwan:
Rerunning the failed job

pkg/env/msiauthorizer.go
```diff
@@ -20,14 +20,22 @@ const (
...
 		options := c.Environment().ManagedIdentityCredentialOptions()
 		return azidentity.NewManagedIdentityCredential(options)
 	}
+
+	var msiContext string
+	if c.component == COMPONENT_GATEWAY {
+		msiContext = string(MSIContextGateway)
+	} else {
+		msiContext = string(MSIContextRP)
+	}
```
SudoBrendan, line 34: This feels a little prone to failure later on in weird ways. If we want envs to be our source of truth, would it be better to put this all in objects there? So we get a `component` type with `name` and `msiContext` attributes? That way we're forced to always define what context each component gets, and it's immediately obvious in a single location?

SudoBrendan (approved):
Ultimately LGTM, one comment.

pkg/env/msiauthorizer.go
hawkowl: Yeah, this is a good idea for further work on it, especially as MIMO will have its own component contexts and permissions.

PR #3256, opened 2023-10-31
add muo alerts

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-4338

Adds muo alerts to aro operator that will be picked up by RP monitor process and shipped to Geneva.  Alerts are based on the set of [metrics](https://github.com/openshift/managed-upgrade-operator/blob/4efa725c960b07674c0cd4952898439f3b27a2db/docs/metrics.md?plain=1#L7-L13) exposed by the MUO operator and OSD/ROSAs  [current alerting configuration]( https://github.com/openshift/managed-cluster-config/blob/master/deploy/sre-prometheus/100-managed-upgrade-operator.PrometheusRule.yaml)
I opted to not add an alert for the metric  `upgradeoperator_scaling_failed` because the inability of scaling does not block cluster upgrades so the alert would be invalid, see [OSD-10740](https://issues.redhat.com//browse/OSD-10740).

### Test plan for issue::
I created a cluster, started an upgrade via MUO, killed kubelet on one control plane node during the upgrade, and observed `UpgradeControlPlaneUpgradeTimeoutSRE` was firing in the OpenShift console.

~~I am also looking into how to test these~~ Tested these alerts against a production cluster/geneva.

Documentation is being created in a separate work item.

### DIFF
--- a/pkg/operator/controllers/muo/staticresources/muo_alerts.yaml
+++ b/pkg/operator/controllers/muo/staticresources/muo_alerts.yaml
@@ -63,3 +63,13 @@ spec:
       annotations:
         summary: "UpgradeConfig has not successfully synced in 4 hours."
         description: "This clusters UpgradeConfig has not been synced in 4 hours and may be out of date"
+    - alert: UpgradeClusterCheckFailedSRE
+      # Alert if the cluster has set its pre/post-upgrade health check failure metric for a ten-minute average window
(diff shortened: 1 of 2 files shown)


tsatam (approved):
looks awesome, thanks!

tsatam (approved):

cadenmarchese (approved):

PR #3260, opened 2023-11-02
Bump Hive Version + Minimal Install

bennerv:

ARO-4545

Bumps the hive version to resolve vulns as well as moves us to a hive minimal version which allows us not to pull ocp-release and oc-cli images.

### Test plan for issue:

Full int environment

Nope


s-amann (approved):
:bee:

PR #3263, opened 2023-11-02
Introduce an interface for controllers to improve consistency

mbarnes:

Fixes


mbarnes:
With https://github.com/Azure/ARO-RP/pull/3221 merged I incorporated the `GenevaLogging` controller into these commits.

pkg/operator/controllers/base/aro_controller.go
```diff
@@ -13,16 +13,21 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 )
 
 type AROReconciler interface {
+	reconcile.Reconciler
 	GetName() string
 	SetupWithManager(ctrl.Manager) error
+	ReconcileEnabled(context.Context, ctrl.Request, *arov1alpha1.Cluster) (ctrl.Result, error)
```
tsatam, line 25: nit: these names may convey that these functions return a boolean (e.g. `isReconcileEnabled`). I can't think of an immediately better name, but this function and `ReconcileDisabled` may need a better name.

tsatam (approved):
LGTM - one comment on the names for the `ReconcileEnabled` and `ReconcileDisabled` functions, but the structure of the change itself looks great, thank you!

pkg/operator/controllers/base/aro_controller.go
mbarnes: Best I can think of is flip-flop it to `EnabledReconcile` but I'm not sure if that's any clearer.
Naming is hard.  I'll submit to the consensus of opinion.

SudoBrendan: Java uses `is` as a standard prefix :coffee: https://www.comp.nus.edu.sg/~cs2103/AY1516S2/contents/coding-standards-java.html - since that was my first language, it's my preference too. It is longer though, and we don't have a coding standard :bangbang: so for now I'd say stick with what feels right to you.

mbarnes: I take that to mean the method name is clear enough as is then, if I'm understanding you right, because the expectation for a simple "is reconcile enabled?" test is that it would be named `IsReconcileEnabled`.  Whereas this method is what the controller implements for "reconcile when enabled flag is true". (Note also the companion "reconcile when enabled flag is false" method on the next line down).

We can always rename it later if it proves confusing to others, but I think in the proper context the confusion @tsatam was calling out should be avoided.

pkg/operator/controllers/base/aro_controller.go
```diff
@@ -33,6 +38,27 @@ func (c *AROController) GetName() string {
...
+		c.Log.Debug("controller is disabled")
+		return c.Reconciler.ReconcileDisabled(ctx, request, cluster)
+	}
+
+	c.Log.Debug("running")
+	return c.Reconciler.ReconcileEnabled(ctx, request, cluster)
+}
+
+func (c *AROController) ReconcileDisabled(ctx context.Context, request ctrl.Request, cluster *arov1alpha1.Cluster) (ctrl.Result, error) {
+	c.ClearConditions(ctx)
```
jaitaiwan, line 58: Do we want to clear out the current conditions when disabling? I'm wondering if that could cause issues with debugging, in that we might want to keep the conditions around after it's disabled. I don't know of any reason why we couldn't leave the conditions as they are if there's an enabled/disabled condition?

pkg/operator/controllers/base/aro_controller.go
jaitaiwan: I think that we should keep `Reconcile` as is and use `context.WithValue` instead because it seems to me that it fits this use case. In fact one could use `context.WithValue` to pass along whether or not the `Reconcile` function should be acting in an enabled/disabled mode too.

Imho this would simplify the implementation a bit and return control of how the enabled/disabled logic worked back to each reconcile function.

jaitaiwan (commented):
I have a couple of thoughts on this PR, one of which talks about another way of implementing which would take advantage of the existing call structure and put control back into the hands of each reconcile function.

PR #3265, opened 2023-11-07
e2e: Spew ODataError details before panicing

mbarnes:

Errors returned from MS Graph APIs are useless as an `error` type.  They just use a [hard-coded message](https://github.com/microsoft/kiota-abstractions-go/blob/main/api_error.go#L16):
```
error status code received from the API
```

To get to the real information requires casting the `error` as an `ODataError` and unpacking the details, which I usually just "spew" to a log.

This is to help debug such errors we're currently seeing in E2E tests.

### DIFF
--- a/test/e2e/setup.go
+++ b/test/e2e/setup.go
@@ -18,6 +18,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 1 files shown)


bennerv (approved):

SudoBrendan:
Hah - so, if E2E fails and showcases the new functionality, is that :x: actually a :heavy_check_mark: ? :joy:

mbarnes:
Rebased on master with (hopefully) the E2E fix from https://github.com/Azure/ARO-RP/pull/3269.

bennerv (approved):

PR #3266, opened 2023-11-07
e2e: Spew ODataError details before panicing

bennerv:

Fixes


bennerv:
Fails on Posting of application.  Request struct has a displayName in it.

PR #3274, opened 2023-11-08
Update ARO operator Azure auth scheme to use a DefaultAzureCredential

kimorris27:

- https://issues.redhat.com/browse/ARO-3992
- https://issues.redhat.com/browse/ARO-3987
- https://issues.redhat.com/browse/ARO-3988

This PR updates the ARO operator to authenticate to Azure using a [DefaultAzureCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential), which will work with both a CSP and WI.

To get it working with a CSP, I had to make a few other changes:
- Have the ARO operator get the Azure authentication details from Secret values that are put into the Pod via environment variables
- Add a CredentialsRequest to the ARO operator deployment, which tells the CCO (cloud credential operator) to create an Azure creds Secret in the ARO operator's namespace
- Restart the ARO operator (master only) when customers rotate their CSP credentials (to ensure the new Secret value is read in). The `az aro update` process will also now verify that the CredentialsRequest mentioned in the bullet above has been reconciled (and thus the Secret has been updated) before doing this restart.

Note that to get the ARO operator working on an ARO WI cluster later on, we'll need a follow-up PR to have the operator Pod 1) use `AZURE_FEDERATED_TOKEN_FILE` instead of `AZURE_CLIENT_SECRET` and 2) mount a bound service account token (more info in testing steps below because I had to do it manually; the testing steps could also be a useful reference for the engineer who makes the follow-up PR).

### Test plan for issue:

There were a few different things to test:

-  The new authentication scheme works with a CSP cluster
    - Tested by provisioning a dev cluster and observing the operator successfully doing stuff to Azure resources in response to changes I was making on the cluster
    - Will also be covered in E2E
- The operator still authenticates to Azure successfully and continues to work after rotating the CSP's client secret via an `az aro update --refresh-credentials`
    1. I ran an `az aro update --refresh-credentials` against my dev cluster and observed the ARO operator master Pod get replaced during the Deployment rollout.
    2. I then used the Portal to remove the Microsoft.Storage service endpoint from both cluster subnets.
    3. Then I scaled down one of the worker MachineSets to trigger the subnets controller. The subnets controller added the service endpoint back successfully.
    - I observed some 401 Azure authentication errors immediately after the new operator master Pod started up. It seems like there is short lag before the correct Secret value makes it into the Pod, but it eventually worked with no intervention (per my test results). Just mentioning here because it was interesting to me.
 - The new authentication scheme works with a WI (tested on a vanilla OCP WI cluster)
     1. I provisioned a WI cluster using the instructions at [1], but before running `ccoctl azure create-all`, I copied the ARO operator's CredentialsRequest manifest into the `--credentials-request-dir` so that ccoctl would create a WI and a Secret manifest for the ARO operator. I also removed this Secret manifest from the install manifest directory before installing the cluster and saved the manifest so I could apply it myself later.
     2. After the cluster finished installing, I added the ARO Cluster CRD to the cluster and created a minimal Cluster CR with just the attributes needed to make the subnets controller work. I included these attributes:
         - azEnvironment 
         - clusterResourceGroupId
         - location
         - operatorFlags (disabled everything except the subnets controller and service endpoint reconciliation)
         - resourceId (just put any resource ID that has the right subscription; the subnet controller only reads this to get the subscription ID)
         - serviceSubnets:
         - storageSuffix
         - vnetId
     3. I then set up the ARO operator (master only, since only the master Pod makes Azure API calls) on the cluster using the manifests in `pkg/operator/deploy` and the Secret manifest that I had saved during step 1. I did make a few changes to the Deployment manifest: 1) replaced `AZURE_CLIENT_SECRET` with `AZURE_FEDERATED_TOKEN_FILE` and 2) mounted a bound service account token (see [2] for the things to add).

[1] https://github.com/openshift/cloud-credential-operator/blob/master/docs/azure_workload_identity.md

[2] Two things to add in the ARO operator Deployment spec to mount the bound service account token:

1. Add this to the _Pod_ spec:

```
volumes:
- name: bound-sa-token
  projected:
  defaultMode: 420
  sources:
  - serviceAccountToken:
      audience: openshift
      expirationSeconds: 3600
      path: token
```

2. Add this to the _container_ spec:

```
volumeMounts:
- mountPath: /var/run/secrets/openshift/serviceaccount
  name: bound-sa-token
  readOnly: true
```

No


pkg/util/clusterauthorizer/authorizer.go
```diff
@@ -70,12 +63,8 @@ func (a *azRefreshableAuthorizer) NewRefreshableAuthorizerToken(ctx context.Cont
...
-	return azidentity.NewClientSecretCredential(
-		string(credentials.TenantID),
-		string(credentials.ClientID),
-		string(credentials.ClientSecret),
-		environment.ClientSecretCredentialOptions())
+func GetTokenCredential(environment *azureclient.AROEnvironment) (azcore.TokenCredential, error) {
+	return azidentity.NewDefaultAzureCredential(environment.DefaultAzureCredentialOptions())
 }
 
 // AzCredentials gets Cluster Service Principal credentials from the Kubernetes secrets
```
cadenmarchese, line 70: Do we need `AzCredentials()` anymore? Browsing through the code base, It looks like we've replaced all uses of it with `GetTokenCredential()`, except for the unit test of `AzCredentials()`.

cadenmarchese (commented):
First pass, looks great! I just have a question about the continued use of an old function.

pkg/cluster/install.go
```diff
@@ -201,6 +201,8 @@ func (m *manager) Update(ctx context.Context) error {
 		steps.Action(m.renewMDSDCertificate),
 		steps.Action(m.updateOpenShiftSecret),
 		steps.Action(m.updateAROSecret),
+		steps.Action(m.restartAROOperatorMaster), // depends on m.updateOpenShiftSecret; the point of restarting is to pick up any changes made to the secret
+		steps.Condition(m.aroDeploymentReady, 20*time.Minute, true),
```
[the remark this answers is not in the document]
kimorris27: I never stopped to think about it - I just used 20 minutes because it's same timing we use in other places (admin update and install).

But since this is a context where we're not doing any huge updates and it's not during cluster install times, why don't we change it to 5 minutes? Probably still longer than needed, but will allow some time in case something weird is going on.

pkg/util/clusterauthorizer/authorizer.go
```diff
@@ -70,29 +64,33 @@ func (a *azRefreshableAuthorizer) NewRefreshableAuthorizerToken(ctx context.Cont
...
-	err := client.Get(ctx, types.NamespacedName{Namespace: AzureCredentialSecretNameSpace, Name: AzureCredentialSecretName}, clusterSPSecret)
 	if err != nil {
 		return nil, err
 	}
 
+	return credential, nil
+}
+
+// AzCredentials gets Cluster Service Principal credentials from the Kubernetes Secret. It returns nil if the Secret is not found.
+func AzCredentials(ctx context.Context, kubernetescli kubernetes.Interface) (Credentials, error) {
```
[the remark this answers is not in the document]
kimorris27: You're right, it's not being used anywhere else and we don't need it anymore. I left it in because it seems like it might be useful in the future and we already have tests for it and everything, but I'll remove it if we think it makes sense to.

pkg/util/clusterauthorizer/authorizer.go
kimorris27: Answered in other discussion thread about this topic.

pkg/util/clusterauthorizer/authorizer.go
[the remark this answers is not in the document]
cadenmarchese: +1 on cleaning it up - it seems duplicate now :)

pkg/cluster/install.go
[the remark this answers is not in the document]
cadenmarchese: +1, 5 minutes sounds good.

pkg/util/kubernetes/deployments.go
```diff
@@ -0,0 +1,31 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	appsv1client "k8s.io/client-go/kubernetes/typed/apps/v1"
+)
+
+// Restart restarts the given Deployment by performing the equivalent of a `kubectl rollout restart deployment <deployment-name>` against it.
+func Restart(ctx context.Context, cli appsv1client.DeploymentInterface, deploymentName string) error {
+	d, err := cli.Get(ctx, deploymentName, metav1.GetOptions{})
+
```
bennerv, line 17: nit - no new line

pkg/util/kubernetes/deployments.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	if d.Spec.Template.ObjectMeta.Annotations == nil {
+		d.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+	}
+
+	d.Spec.Template.ObjectMeta.Annotations["kubectl.kubernetes.io/restartedAt"] = time.Now().Format(time.RFC3339)
+
+	_, err = cli.Update(ctx, d, metav1.UpdateOptions{})
+
```
bennerv, line 29: nit - no new line

bennerv: should we patch this?  We've been having issues with running updates instead of patches.

pkg/operator/deploy/deploy.go
```diff
@@ -405,6 +407,18 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
 	return true, nil
 }
 
+func (o *operator) Restart(ctx context.Context, deploymentNames []string) error {
+	for _, dn := range deploymentNames {
+		err := utilkubernetes.Restart(ctx, o.kubernetescli.AppsV1().Deployments(pkgoperator.Namespace), dn)
+
```
bennerv, line 413: nit - no new line

pkg/cluster/install.go
```diff
@@ -201,6 +201,8 @@ func (m *manager) Update(ctx context.Context) error {
 		steps.Action(m.renewMDSDCertificate),
 		steps.Action(m.updateOpenShiftSecret),
 		steps.Action(m.updateAROSecret),
+		steps.Action(m.restartAROOperatorMaster), // depends on m.updateOpenShiftSecret; the point of restarting is to pick up any changes made to the secret
```
bennerv, line 204: do we need to wait for the reconciliation of the credentialsRequest before we restart it?

pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go
```diff
@@ -22,18 +22,14 @@ type servicePrincipalChecker interface {
 type checker struct {
 	log *logrus.Entry
 
-	credentials        func(ctx context.Context) (*clusterauthorizer.Credentials, error)
-	getTokenCredential func(azEnv *azureclient.AROEnvironment, credentials *clusterauthorizer.Credentials) (azcore.TokenCredential, error)
+	getTokenCredential func(azEnv *azureclient.AROEnvironment) (azcore.TokenCredential, error)
```
bennerv, line 25: what's the purpose of this as a struct?  So that the controller can fetch a new token credential every time it needs to run the checker, rather than once on startup?

bennerv (commented):
Great work - just a quick first pass on this

pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go
bennerv: I'm also wondering if this is "legacy".  The newer azure-sdk-for-go no longer requires us to refresh the token iirc.

pkg/cluster/install.go
kimorris27: That's a good point. I think we want to, but I'll have to think about how to check that the reconciliation is done.

pkg/operator/controllers/checkers/serviceprincipalchecker/checker.go
kimorris27: >what's the purpose of this as a struct? So that the controller can fetch a new token credential every time it needs to run the checker, rather than once on startup?

I'm not 100% sure, but it seems that way to me.

>I'm also wondering if this is "legacy". The newer azure-sdk-for-go no longer requires us to refresh the token iirc.

I wasn't aware of this. I'll have to look into it a bit, and then maybe we can remove unneeded stuff from the struct.

kimorris27: After some digging, I found that you're right - as long as we're using azidentity, our tokens will get auto-refreshed before they expire. So I'll see about refactoring this struct a bit.

For context, https://github.com/Azure/ARO-RP/pull/2667 switched us over to `azidentity`, which uses MSAL under the hood, and https://learn.microsoft.com/en-us/entra/identity-platform/msal-overview states that MSAL refreshes tokens automatically.

kimorris27: Well, after looking at this code in more detail and considering how we might refactor it, I think we might as well leave it how it is.

It looks like for all of the checkers under `pkg/operator/controllers/checkers`, there's a pattern of defining a `SomethingChecker` interface and then creating a struct type that implements the interface. So it logically makes sense to have that struct type encapsulate anything related to the `Check` that the controller in question does. I think `getTokenCredential` and `newSPValidator` are included in the struct more because its logical than because of a goal of getting a new token each time the controller reconciles (which is what I was originally thinking).

Additionally, if you look at the storage accounts controller for example (specifically https://github.com/Azure/ARO-RP/blob/ac7cf2f35c24c1e1d39960cb1b382de7def8f9ba/pkg/operator/controllers/storageaccounts/storageaccount_controller.go#L95), it also "manually" creates a new token credential each time it `Reconcile`s. So if we wanted to refactor to only create a token credential on controller startup, we'd probably want to do that to all of the controllers that leverage Azure clients and not just the service principal checker.

Let me know what you think!

pkg/operator/deploy/deploy.go
```diff
@@ -405,6 +407,18 @@ func (o *operator) IsReady(ctx context.Context) (bool, error) {
 	return true, nil
 }
 
+func (o *operator) Restart(ctx context.Context, deploymentNames []string) error {
+	for _, dn := range deploymentNames {
+		err := utilkubernetes.Restart(ctx, o.kubernetescli.AppsV1().Deployments(pkgoperator.Namespace), dn)
```
s-amann, line 412: should this consider attempting to restart the rest of the supplied `deploymentNames` instead of erroring out on the first failure? - might be out of scope for the PR since we are only using it for one name so far.

pkg/util/kubernetes/deployments.go
s-amann: +1 to making this a patch call, especially considering we just want to ensure an annotation is added.

s-amann (commented):
first pass lgtm, some comments below. If you haven't already done so, please rebase.

pkg/cluster/install.go
s-fairchild: What would happen if the credentialsRequest hadn't reconciled? I think that would cause the ARO operator to crashloop briefly until it's finished.

pkg/operator/deploy/deploy.go
s-fairchild: I think it's safe to add a `//TODO` comment here for later. [For the moment](https://github.com/Azure/ARO-RP/blob/ad9145187387efb7ed5f7b15cf1bb345ea6efbfa/pkg/cluster/arooperator.go#L58C76-L58C76) only a singly deployment name is being passed in.

s-fairchild (changes requested):
LGTM, other than some additional test coverage if possible. If that's too much for this PR then maybe in a follow up PR.
 Thanks for the neat commit structure.

pkg/util/kubernetes/deployments.go
kimorris27: I became aware while changing this over to a patch call that there are multiple ways to patch. After looking at the different options, I went with server-side apply. Let me know if we should do it a different way though.

pkg/operator/deploy/deploy.go
kimorris27: This makes sense and isn't a difficult change. I've already made it.

pkg/cluster/arooperator.go
```diff
@@ -53,3 +53,7 @@ func (m *manager) ensureAROOperatorRunningDesiredVersion(ctx context.Context) (b
 func (m *manager) renewMDSDCertificate(ctx context.Context) error {
 	return m.aroOperatorDeployer.RenewMDSDCertificate(ctx)
 }
+
+func (m *manager) restartAROOperatorMaster(ctx context.Context) error {
```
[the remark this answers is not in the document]
kimorris27: Unit testing didn't make sense to me here because the ARO operator deployer's `Restart` method basically just calls already-tested code in a loop, but I added an E2E test. Let me know what you think.

pkg/cluster/install.go
kimorris27: I did somewhat of a deep dive while figuring out how to go about this. The CredentialsRequest's [status](https://docs.openshift.com/container-platform/4.12/rest_api/security_apis/credentialsrequest-cloudcredential-openshift-io-v1.html#status) has a `lastSyncCloudCredsSecretResourceVersion` that we could check, but it feels hacky because we would have to save the old version prior to updating the secret so that we can see what it used to be. I'm thinking we can just use the `lastSyncTimestamp` instead by checking to see if it's something within the past 5 minutes. Your thoughts?

It's also tricky because CredentialsRequests haven't made their way into openshift/client-go. Based on my reading, I'll have to use https://pkg.go.dev/k8s.io/client-go/dynamic#DynamicClient to work with the CredentialsRequest from within `pkg/cluster`, but let me know if there's a better way.

kimorris27: I pushed a solution that uses the `lastSyncTimestamp`, and I feel like this should be fine.

test/e2e/update.go
```diff
@@ -15,11 +15,39 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	mgmtredhatopenshift20220904 "github.com/Azure/ARO-RP/pkg/client/services/redhatopenshift/mgmt/2022-09-04/redhatopenshift"
 	mgmtredhatopenshift20230701preview "github.com/Azure/ARO-RP/pkg/client/services/redhatopenshift/mgmt/2023-07-01-preview/redhatopenshift"
 	"github.com/Azure/ARO-RP/pkg/util/stringutils"
 )
 
 var _ = Describe("Update clusters", func() {
+	It("must restart the aro-operator-master Deployment", func(ctx context.Context) {
```
s-fairchild, line 24: This is good, thanks!

s-fairchild (dismissed):
Thanks for adding test coverage, lgtm.

pkg/cluster/clusterserviceprincipal.go
```diff
@@ -77,47 +78,64 @@ func (m *manager) createOrUpdateClusterServicePrincipalRBAC(ctx context.Context)
...
+	if changed {
+		data, err := yaml.Marshal(cf)
+		if err != nil {
+			return false, nil, err
+		}
+		secret.Data["cloud-config"] = data
+		return true, secret, nil
+	}
+
+	return false, nil, nil
```
s-amann, line 127: fwiw: if we always return `true` with an object and `false` with nil, we don't need the boolean.

pkg/cluster/clusterserviceprincipal.go
```diff
@@ -77,47 +78,64 @@ func (m *manager) createOrUpdateClusterServicePrincipalRBAC(ctx context.Context)
 	return nil
 }
 
-func (m *manager) updateAROSecret(ctx context.Context) error {
+// servicePrincipalUpdated checks whether the CSP has been updated by comparing the cluster doc's
+// ServicePrincipalProfile to the contents of the kube-system/azure-cloud-provider Secret. If the CSP
+// has changed, it returns true along with a new corev1.Secret to use to update the Secret to match
+// what's in the cluster doc.
+func (m *manager) servicePrincipalUpdated(ctx context.Context) (bool, *corev1.Secret, error) {
 	var changed bool
```
s-amann, line 86: if possible I'd rather see the code that is in the `if changed` block extracted to its own helper function and used along with a `return` when we need to set `changed=true`. this avoids the use of a tracking boolean and reduces complexity.

kimorris27: Yeah, I see what you mean. I've pushed a commit that I believe implements this the way you have in mind.

s-amann (approved):

cadenmarchese (approved):
I've tested both a new install and PUCM with these changes to confirm behavior looks correct - lgtm, thanks for addressing all of the changes!

PR #3275, opened 2023-11-08
[ARO-4637] Migrate keyvault to track2 SDK

jonathan34c:

Migrate keyvault to track2 SDK. 

Replace [original keyvault](https://github.com/niontive/ARO-RP/blob/master/pkg/util/azureclient/mgmt/keyvault/vaults.go) with newly created keyvault using [track2 SDK]("github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/keyvault/armkeyvault")


pkg/util/cluster/cluster.go
```diff
@@ -34,9 +34,9 @@ import (
 	"github.com/Azure/ARO-RP/pkg/deploy/generator"
 	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/util/arm"
+	keyvaultclient "github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/armkeyvault"
```
nwnt, line 37: Nit: let's use just armkeyvault for azureclient/azuresdk.

pkg/util/cluster/cluster.go
```diff
@@ -104,6 +106,15 @@ func New(log *logrus.Entry, environment env.Core, ci bool) (*Cluster, error) {
 	scopes := []string{environment.Environment().ResourceManagerScope}
 	authorizer := azidext.NewTokenCredentialAdapter(spTokenCredential, scopes)
 
+	valutClient, err := keyvaultclient.NewVaultsClient(environment.SubscriptionID(), spTokenCredential, &policy.ClientOptions{
+		ClientOptions: azcore.ClientOptions{
+			Cloud: environment.Environment().Cloud,
+		},
```
nwnt, line 112: Sorry, I forgot to update my doc on this.  For the client options, let's use something like this or its equivalent instead of creating our own from scratch.

```
	environment.Environment().EnvironmentCredentialOptions()
```

nwnt (changes requested):
Some minor changes.  Should be ready for review after addressing them.

portal/v2/package.json
```diff
@@ -39,7 +39,7 @@
     "prettier": "^2.7.1",
     "process": "^0.11.10",
     "react-app-rewired": "^2.2.1",
-    "react-scripts": "^5.0.1"
+    "react-scripts": "^3.0.1"
```
niontive, line 42: downgrade?

pkg/util/cluster/cluster.go
```diff
@@ -121,7 +126,7 @@ func New(log *logrus.Entry, environment env.Core, ci bool) (*Cluster, error) {
 		routetables:                       network.NewRouteTablesClient(environment.Environment(), environment.SubscriptionID(), authorizer),
 		roleassignments:                   authorization.NewRoleAssignmentsClient(environment.Environment(), environment.SubscriptionID(), authorizer),
 		peerings:                          network.NewVirtualNetworkPeeringsClient(environment.Environment(), environment.SubscriptionID(), authorizer),
-		vaultsClient:                      keyvaultclient.NewVaultsClient(environment.Environment(), environment.SubscriptionID(), authorizer),
+		vaultsClient:                      valutClient,
```
nwnt, line 129: nit: typo.

nwnt (dismissed):

pkg/util/cluster/cluster.go
```diff
@@ -212,7 +217,8 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 	if c.ci {
 		// name is limited to 24 characters, but must be globally unique, so we generate one and try if it is available
 		kvName = "kv-" + uuid.DefaultGenerator.Generate()[:21]
-		result, err := c.vaultsClient.CheckNameAvailability(ctx, mgmtkeyvault.VaultCheckNameAvailabilityParameters{Name: &kvName, Type: to.StringPtr("Microsoft.KeyVault/vaults")})
+
+		result, err := c.vaultsClient.CheckNameAvailability(ctx, sdkkeyvault.VaultCheckNameAvailabilityParameters{Name: &kvName, Type: to.StringPtr("Microsoft.KeyVault/vaults")}, nil)
 		if err != nil {
 			return err
```
SudoBrendan, line 223: Just wondering if we need to be considering anything re: https://github.com/Azure/azure-sdk-for-go/blob/main/documentation/MIGRATION_GUIDE.md#error-handling ? It doesn't seem like it at a glance but just want to be sure errors are in the correct format as they are spat up?

SudoBrendan (commented):
One question, but otherwise LGTM.

pkg/util/cluster/cluster.go
jonathan34c: I think we should be good here, since we are not checking the status code, but just the general error

hawkowl:
@jonathan34c There's some unrelated changes related to the portal in this PR, could you remove these?

hawkowl (dismissed):
unrelated JS package changes

mociarain (dismissed):

package-lock.json
```diff
@@ -0,0 +1,10 @@
+{
```
tsatam, line 1: This file isn't related to the changes, looks like a mistaken addition?

tsatam (changes requested):
Looks like there's still a remaining package-lock.json that shouldn't be included as a part of this change, otherwise LGTM.

package-lock.json
jonathan34c: deleted the file

cadenmarchese (approved):

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.9.1
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.3.1
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/cosmos/armcosmos/v2 v2.5.0
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/keyvault/armkeyvault v1.3.0
```
kimorris27, line 10: Is there any particular reason for using `v1.3.0` when there's a `v1.4.0` available? (See https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/keyvault/armkeyvault?tab=versions)

pkg/util/azureclient/azuresdk/armkeyvault/vaults.go
```diff
@@ -0,0 +1,37 @@
...
+	CheckNameAvailability(ctx context.Context, vaultName armkeyvault.VaultCheckNameAvailabilityParameters, options *armkeyvault.VaultsClientCheckNameAvailabilityOptions) (armkeyvault.VaultsClientCheckNameAvailabilityResponse, error)
+}
+
+type vaultsClient struct {
+	*armkeyvault.VaultsClient
+}
+
+var _ VaultsClient = &vaultsClient{}
+
+func NewVaultsClient(subscriptionID string, credential azcore.TokenCredential, options *azidentity.EnvironmentCredentialOptions) (VaultsClient, error) {
```
kimorris27, line 27: I noticed that this track 2 SDK client takes in an `azidentity.EnvironmentCredentialsOption` here rather than an `arm.ClientOptions` like the [subnets client](https://github.com/Azure/ARO-RP/blob/master/pkg/util/azureclient/azuresdk/armnetwork/subnets.go), the first one that was migrated to track 2, does.

I see why you did it this way - because in the context of `pkg/util/cluster`, you are given an `azidentity.EnvironmentCredentialOptions` to work with - but I think all of the track 2 SDK clients should take in the same type for the `options` parameter for consistency's sake.

@cadenmarchese , since you were the one who did that subnet client migration, what do you think?

go.mod
jonathan34c: i can update to 1.4.0, no particular reason to use the old version

pkg/util/azureclient/azuresdk/armkeyvault/vaults.go
jonathan34c: thank you for pointing out. i agree that a consistant interface is important, therefore, i have changed the PR.

kimorris27 (commented):
Thanks for the responses to my feedback! I think I'm approving once all CI checks have passed.

tsatam (approved):

kimorris27 (approved):
Not sure why the Azure Pipelines aren't running even though I tried to manually trigger them, but I'm happy for now if the other checks are passing.

mociarain (approved):

UlrichSchlueter (approved):
LGTM

PR #3280, opened 2023-11-10
Watch MachineSets for worker subnet changes instead of Machines

tsatam:

- Potential fix for [ARO-4632](https://issues.redhat.com/browse/ARO-4632)
- Fixes [ARO-4468](https://issues.redhat.com/browse/ARO-4468) as the test flakiness blocks merge of this PR

- Updates the ARO Operator Subnets and StorageAccounts controller to only watch master Machines, and to watch worker MachineSets instead
- Updates the underlying cluster_subnets List() function to also retrieve worker subnets via MachineSets rather than Machines
- Updates the e2e test for subnet NSG reconciliation, as it was previously dependent on the subnet controller reconciling frequently and unnecessarily in order to pass. 

This should (hopefully) cut down on Reconcile loops for the above two controllers when individual worker nodes change but their underlying definition and therefore subnet do not. 

### Test plan for issue:

- Unit tests for cluster_subnets have been updated to use a worker MachineSet
- E2E test has been updated out of necessity, ensure it passes consistently against this change
- Manual test against a real cluster to ensure the underlying ARM calls are reduced

No

### DIFF
--- a/pkg/operator/controllers/storageaccounts/storageaccount_controller.go
+++ b/pkg/operator/controllers/storageaccounts/storageaccount_controller.go
@@ -9,7 +9,6 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	"github.com/sirupsen/logrus"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -116,11 +115,15 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	aroClusterPredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
 		return o.GetName() == arov1alpha1.SingletonClusterName
 	})
+	masterMachinePredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		role, ok := o.GetLabels()["machine.openshift.io/cluster-api-machine-role"]
+		return ok && role == "master"
+	})
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&arov1alpha1.Cluster{}, builder.WithPredicates(aroClusterPredicate)).
-		Watches(&source.Kind{Type: &machinev1beta1.Machine{}}, &handler.EnqueueRequestForObject{}). // to reconcile on machine replacement
-		Watches(&source.Kind{Type: &corev1.Node{}}, &handler.EnqueueRequestForObject{}).            // to reconcile on node status change
+		Watches(&source.Kind{Type: &machinev1beta1.Machine{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(masterMachinePredicate)). // to reconcile on master machine replacement
+		Watches(&source.Kind{Type: &machinev1beta1.MachineSet{}}, &handler.EnqueueRequestForObject{}).                                              // to reconcile on worker machinesets
 		Named(ControllerName).
 		Complete(r)
 }
--- a/pkg/operator/controllers/storageaccounts/storageaccounts.go
+++ b/pkg/operator/controllers/storageaccounts/storageaccounts.go
@@ -13,6 +13,7 @@ import (
 	imageregistryv1 "github.com/openshift/api/imageregistry/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"github.com/Azure/ARO-RP/pkg/util/azureerrors"
 	"github.com/Azure/ARO-RP/pkg/util/stringutils"
 )
 
@@ -32,6 +33,10 @@ func (r *reconcileManager) reconcileAccounts(ctx context.Context) error {
 	for _, subnet := range subnets {
 		mgmtSubnet, err := r.subnets.Get(ctx, subnet.ResourceID)
 		if err != nil {
+			if azureerrors.IsNotFoundError(err) {
+				r.log.Infof("Subnet %s not found, skipping", subnet.ResourceID)
+				break
+			}
 			return err
 		}
 
--- a/pkg/util/subnet/cluster_subnets.go
+++ b/pkg/util/subnet/cluster_subnets.go
@@ -45,8 +45,8 @@ func (m *kubeManager) List(ctx context.Context) ([]Subnet, error) {
 
 	// select all workers by the  machine.openshift.io/cluster-api-machine-role: not equal to master Label
 	selector, _ := labels.Parse("machine.openshift.io/cluster-api-machine-role!=master")
-	machines := &machinev1beta1.MachineList{}
-	err := m.client.List(ctx, machines, &client.ListOptions{
+	machineSets := &machinev1beta1.MachineSetList{}
+	err := m.client.List(ctx, machineSets, &client.ListOptions{
 		Namespace:     machineSetsNamespace,
 		LabelSelector: selector,
 	})
@@ -54,8 +54,8 @@ func (m *kubeManager) List(ctx context.Context) ([]Subnet, error) {
 		return nil, err
 	}
 
-	for _, machine := range machines.Items {
-		subnetDesc, err := m.getDescriptorFromProviderSpec(machine.Spec.ProviderSpec.Value)
+	for _, machineSet := range machineSets.Items {
+		subnetDesc, err := m.getDescriptorFromProviderSpec(machineSet.Spec.Template.Spec.ProviderSpec.Value)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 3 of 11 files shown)


pkg/operator/controllers/storageaccounts/storageaccounts.go
```diff
@@ -32,6 +33,10 @@ func (r *reconcileManager) reconcileAccounts(ctx context.Context) error {
 	for _, subnet := range subnets {
 		mgmtSubnet, err := r.subnets.Get(ctx, subnet.ResourceID)
 		if err != nil {
+			if azureerrors.IsNotFoundError(err) {
+				r.log.Infof("Subnet %s not found, skipping", subnet.ResourceID)
+				break
+			}
```
SudoBrendan, line 39: while we're here, should we add `429` handling, or are we waiting for SDK Track 2?

pkg/util/subnet/cluster_subnets.go
```diff
@@ -44,26 +44,27 @@ func (m *kubeManager) List(ctx context.Context) ([]Subnet, error) {
 	subnetMap := []Subnet{}
 
 	// select all workers by the  machine.openshift.io/cluster-api-machine-role: not equal to master Label
+	// TODO: is this even necessary? Masters should not have a MachineSet
```
SudoBrendan, line 47: ...yet. I think this is eventually coming though?

SudoBrendan (approved):
Overall, LGTM. Left some comments.

pkg/util/subnet/cluster_subnets.go
hlipsig: As far as I know Masters doe have a machineset. https://docs.openshift.com/container-platform/4.12/machine_management/control_plane_machine_management/cpmso-about.html

pkg/operator/controllers/storageaccounts/storageaccounts.go
tsatam: We perhaps ought to add 429 handling, either before we update to Track 2, or even after (if we want our controllers to requeue themselves when throttled rather than running and waiting for eventual completion), but I think it can and should be done separately. I've included the change in [this PR](https://github.com/Azure/ARO-RP/pull/3272) to the StorageAccounts controller, and we can bring the same refactor to the Subnets controller, or refactor both controllers to perform their work within a single shared Reconcile and share whatever ARM API GET Resource calls they are able to. 

Handling Retry-After is good for dealing with the operator hitting API quota limits - the change in this PR is trying to prevent us from even getting to that state in the first place.

pkg/util/subnet/cluster_subnets.go
tsatam: verified on shared cluster (4.13) that this is the case, though I don't know if that is representative of a fully "active" control plane machine management example, as the CPMS is Inactive.

tsatam:
temp hold: focusing flaky e2e test in CI to ensure it is fixed

tsatam:
re-adding hold, e2e test still appears to be flaky. Do not merge until this test is confirmed consistent.

pkg/operator/controllers/storageaccounts/storageaccount_controller.go
```diff
@@ -116,11 +115,15 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	masterMachinePredicate := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		role, ok := o.GetLabels()["machine.openshift.io/cluster-api-machine-role"]
+		return ok && role == "master"
+	})
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&arov1alpha1.Cluster{}, builder.WithPredicates(aroClusterPredicate)).
-		Watches(&source.Kind{Type: &machinev1beta1.Machine{}}, &handler.EnqueueRequestForObject{}). // to reconcile on machine replacement
-		Watches(&source.Kind{Type: &corev1.Node{}}, &handler.EnqueueRequestForObject{}).            // to reconcile on node status change
+		Watches(&source.Kind{Type: &machinev1beta1.Machine{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(masterMachinePredicate)). // to reconcile on master machine replacement
```
cadenmarchese, line 125: What would be the impact of putting in a `Watches` for control plane machinesets now (before we have them in many/most ARO clusters) so that we don't have to go and change that later? I'm assuming the end goal is to only watch machinesets and control plane machinesets, and watching master machines is a temporary solution in the meantime.

test/e2e/operator.go
```diff
@@ -302,27 +303,51 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
 		location = *oc.Location
 
-		vnet, masterSubnet, err := apisubnet.Split(*oc.OpenShiftClusterProperties.MasterProfile.SubnetID)
-		Expect(err).NotTo(HaveOccurred())
-		_, workerSubnet, err := apisubnet.Split((*(*oc.OpenShiftClusterProperties.WorkerProfiles)[0].SubnetID))
+		vnet, workerSubnet, err := apisubnet.Split((*(*oc.OpenShiftClusterProperties.WorkerProfiles)[0].SubnetID))
 		Expect(err).NotTo(HaveOccurred())
 
 		subnetsToReconcile = map[string]*string{
-			masterSubnet: to.StringPtr(""),
```
cadenmarchese, line 318: What's the justification from removing the master subnet from e2e scope? It seems like we should be testing that since it's still part of the controller's functionality, even though we've changed it a bit.

test/e2e/operator.go
```diff
@@ -302,27 +303,51 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
 		r, err := azure.ParseResourceID(vnet)
 		Expect(err).NotTo(HaveOccurred())
 		resourceGroup = r.ResourceGroup
 		vnetName = r.ResourceName
+
+		// Store the existing NSGs for the cluster before the test runs, in order to ensure we clean up
+		// after the test finishes, success or failure.
+		// This is expensive but will prevent flakes.
+		By("gathering existing subnet NSGs")
+		for subnet := range subnetsToReconcile {
```
cadenmarchese, line 330: I don't believe we'd need this range anymore if we get rid of the master subnet portion of the test (there's only one subnet to check).

test/e2e/operator.go
```diff
@@ -302,27 +303,51 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
+			subnetObject, err := clients.Subnet.Get(ctx, resourceGroup, vnetName, subnet, "")
+			Expect(err).NotTo(HaveOccurred())
+
+			subnetsToReconcile[subnet] = subnetObject.NetworkSecurityGroup.ID
+		}
+	}
+
+	cleanUpSubnetNSGs := func(ctx context.Context) {
+		By("cleaning up subnet NSGs")
+		for subnet := range subnetsToReconcile {
```
cadenmarchese, line 340: Same comment as above.

cadenmarchese (commented):
Thanks for going over your PR yesterday! I have some followup questions and comments, but this is looking good.

pkg/operator/controllers/storageaccounts/storageaccount_controller.go
tsatam: Assuming there are no frequent changes to the ControlPlaneMachineSet resources, I don't imagine there would be a wider negative impact. However, the code to actually extract the in-use cluster subnets will, due to the timing you called out, only be reading master Machines and worker MachineSets anyways, so we'd be adding a Watch to a resource we never actually read. 

I think if we get to a point where we only ever watch MachineSets and ControlPlaneMachineSets, we'd need a similar PR to this one to update both our Watches and our actual subnet retrieval code. 

I'm personally in favor of leaving this as-is, especially since we've discussed internally about the possibility of merging these two controllers into a single one anyways (to cut down on duplicate GET Subnet Azure calls).

test/e2e/operator.go
tsatam: The primary goal was flake defense - I was noticing that if these e2e tests failed for any reason, they would leave the fake/garbage NSG on the subnets, and would effectively hose the cluster. I think, with the additional guards in place on this test, that it may be safe to reintroduce the master subnet to this test. I can try that out and hammer the test to make sure it doesn't bring back any flakes or otherwise degrade the test cluster.

test/e2e/operator.go
tsatam: I'll remove the ranges if I find we need to keep the master subnet out of the test. 

My original goal here was to create a dedicated MachineSet (or multiple) with their own isolated subnets, effectively implementing https://learn.microsoft.com/en-us/azure/openshift/howto-segregate-machinesets in an E2E test, but given that we already have constraints on requesting the two master/worker subnets in our shared dev-vnet I held off on this for now.

test/e2e/operator.go
tsatam: Re-introduced master subnet, e2e test passed locally w/o flakes (ran 50 times successfully in a row).

test/e2e/operator.go
tsatam: I brought the master subnet back so the range is necessary again.

cadenmarchese (approved):

kimorris27 (approved):
LGTM! These are exciting toil-reducing changes.

hawkowl:
I like this :)

PR #3283, opened 2023-11-15
Remove journald fields that aren't needed

hawkowl:
TBA

Reduces a bunch of usually empty or useless columns

### Test plan for issue:

Not sure?


bennerv (approved):
Should these `Remove` directives be added to the installer fork as well?  Since the bootstrap VM has a different fluent.config

PR #3284, opened 2023-11-17
Fixing cert metric emitting condition param

sankur-codes:

Fixes the parameter for condition check of managed domains [ARO-4741](https://issues.redhat.com/browse/ARO-4741)

This is required so that check for managed domain is more precise and accurate for emitting metrics of apiserver and ingress tls certificates

### Test plan for issue:

Tests are refactored and passing

No


pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -46,7 +47,12 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 		})
 	}
 
-	if dns.IsManagedDomain(mon.oc.Properties.ClusterProfile.Domain) {
+	host, err := getHostFromAPIURL(mon.oc.Properties.APIServerProfile.URL)
+	if err != nil {
+		return err
+	}
+
+	if dns.IsManagedDomain(host) {
```
cadenmarchese, line 55: Do we need any changes to `dns.IsManagedDomain()` as well, or is the change just moving from checking `mon.oc.Properties.ClusterProfile.Domain` to `mon.oc.Properties.APIServerProfile.URL`?

sankur-codes: There isn't any changes required to `dns.IsManaged()` as it just checks if the provided string (host of API URL in our case) ends with a managed domain. The change to `mon.oc.Properties.APIServerProfile.URL` was required as `mon.oc.Properties.ClusterProfile.Domain` would not always end with domain at end of it.

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -98,6 +104,14 @@ func secretMissingMetric(namespace, name string) map[string]string {
 	}
 }
 
+func getHostFromAPIURL(apiURL string) (string, error) {
```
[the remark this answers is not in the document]
sankur-codes: We will require this function to extract the hostname (which will always have the domain at the end) because if its just API URL, it will end with `:6443` so `dns.IsManaged()` will return a `false` every time as it checks for string ending with managed domain name.

SrinivasAtmakuri (approved):
LGTM!

jaitaiwan (approved):
LGTM!

s-amann (approved):

Shivkumar13 (approved):

cadenmarchese (approved):

PR #3286, opened 2023-11-17
Admin action to delete a cluster managed resource

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-4425

This PR adds an admin action to delete cluster managed resources. All resources are deleted using DeleteByIDAndWait except for FrontendIPConfiguration.  FrontendIPConfiguration cannot be deleted via http DELETE, instead loadbalancer config is retrieved, manipulated to not include the FrontendIPConfiguration and applied with `CreateOrUpdateAndWait`.

### Test plan for issue:
Unit tests
E2E tests added verifying that LoadBalancer loadbalancingrules, frontendIPConfiguration, and PublicIPAddresses can be deleted.

Yes, will add command example to `docs/deploy-development-rp.md`.

### DIFF
--- a/docs/deploy-development-rp.md
+++ b/docs/deploy-development-rp.md
@@ -269,6 +269,12 @@ After that, when you [create](https://github.com/Azure/ARO-RP/blob/master/docs/d
   curl -X PATCH -k "https://localhost:8443/admin/subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$RESOURCEGROUP/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER/etcdrecovery" 
   ```
 
+* Delete a managed resource
+  ```bash
+  RESOURCEID=<id of managed resource to delete>
+  curl -X POST -k "https://localhost:8443/admin/subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$RESOURCEGROUP/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER/deletemanagedresource?resourceid=$RESOURCEID"
+  ```
+
 ## OpenShift Version
 
 * We have a cosmos container which contains supported installable OCP versions, more information on the definition in `pkg/api/openshiftversion.go`.
--- a/docs/deploy-development-rp.md
+++ b/docs/deploy-development-rp.md
@@ -271,8 +271,8 @@ After that, when you [create](https://github.com/Azure/ARO-RP/blob/master/docs/d
 
 * Delete a managed resource
   ```bash
-  RESOURCEID=<id of managed resource to delete>
-  curl -X POST -k "https://localhost:8443/admin/subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$RESOURCEGROUP/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER/deletemanagedresource?resourceid=$RESOURCEID"
+  MANAGED_RESOURCEID=<id of managed resource to delete>
+  curl -X POST -k "https://localhost:8443/admin/subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$RESOURCEGROUP/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER/deletemanagedresource?managedresourceid=$MANAGED_RESOURCEID"
   ```
 
 ## OpenShift Version
--- a/docs/deploy-development-rp.md
+++ b/docs/deploy-development-rp.md
@@ -272,7 +272,7 @@ After that, when you [create](https://github.com/Azure/ARO-RP/blob/master/docs/d
 * Delete a managed resource
   ```bash
   MANAGED_RESOURCEID=<id of managed resource to delete>
-  curl -X POST -k "https://localhost:8443/admin/subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$RESOURCEGROUP/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER/deletemanagedresource?managedresourceid=$MANAGED_RESOURCEID"
+  curl -X POST -k "https://localhost:8443/admin/subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$RESOURCEGROUP/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER/deletemanagedresource?managedResourceID=$MANAGED_RESOURCEID"
   ```
 
 ## OpenShift Version
--- a/pkg/frontend/admin_openshiftcluster_delete_managedresource.go
+++ b/pkg/frontend/admin_openshiftcluster_delete_managedresource.go
@@ -9,13 +9,13 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/Azure/go-autorest/autorest"
 	"github.com/go-chi/chi/v5"
 	"github.com/sirupsen/logrus"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
 	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+	"github.com/Azure/go-autorest/autorest"
 )
 
 func (f *frontend) postAdminOpenShiftDeleteManagedResource(w http.ResponseWriter, r *http.Request) {
--- a/pkg/frontend/admin_openshiftcluster_delete_managedresource.go
+++ b/pkg/frontend/admin_openshiftcluster_delete_managedresource.go
@@ -1,0 +1,67 @@
+package frontend
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"net/http"
+	"path/filepath"
+	"strings"
+
+	"github.com/go-chi/chi/v5"
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
+	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+	"github.com/Azure/go-autorest/autorest"
+)
+
+func (f *frontend) postAdminOpenShiftDeleteManagedResource(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
+	r.URL.Path = filepath.Dir(r.URL.Path)
+
+	err := f._postAdminOpenShiftClusterDeleteManagedResource(ctx, r, log)
+	adminReply(log, w, nil, nil, err)
+}
+
+func (f *frontend) _postAdminOpenShiftClusterDeleteManagedResource(ctx context.Context, r *http.Request, log *logrus.Entry) error {
+	resType, resName, resGroupName := chi.URLParam(r, "resourceType"), chi.URLParam(r, "resourceName"), chi.URLParam(r, "resourceGroupName")
+	managedResourceID := r.URL.Query().Get("resourceid")
+	resourceID := strings.TrimPrefix(r.URL.Path, "/admin")
+
+	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	switch {
+	case cosmosdb.IsErrorStatusCode(err, http.StatusNotFound):
+		return api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, "", "The Resource '%s/%s' under resource group '%s' was not found.", resType, resName, resGroupName)
+	case err != nil:
+		return err
+	}
+
+	if !strings.Contains(strings.ToLower(managedResourceID), strings.ToLower(doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID)) {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", "The resource %s is not within the cluster's managed resource group %s.", managedResourceID, doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID)
+	}
+
+	subscriptionDoc, err := f.getSubscriptionDocument(ctx, doc.Key)
+	if err != nil {
+		return err
+	}
+
+	a, err := f.azureActionsFactory(log, f.env, doc.OpenShiftCluster, subscriptionDoc)
+	if err != nil {
+		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
+	}
+
+	err = a.ResourceDeleteAndWait(ctx, managedResourceID)
+	if err != nil {
+		if detailedErr, ok := err.(autorest.DetailedError); ok &&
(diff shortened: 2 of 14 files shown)


docs/deploy-development-rp.md
```diff
@@ -269,6 +269,12 @@ After that, when you [create](https://github.com/Azure/ARO-RP/blob/master/docs/d
   curl -X PATCH -k "https://localhost:8443/admin/subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$RESOURCEGROUP/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER/etcdrecovery" 
   ```
 
+* Delete a managed resource
+  ```bash
+  RESOURCEID=<id of managed resource to delete>
```
s-amann, line 274: nit: consider naming this `MANAGED_RESOURCEID` so it is not confused with the resourceID which is part of the url / request

pkg/frontend/admin_openshiftcluster_delete_managedresource.go
```diff
@@ -0,0 +1,67 @@
...
+	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
+	r.URL.Path = filepath.Dir(r.URL.Path)
+
+	err := f._postAdminOpenShiftClusterDeleteManagedResource(ctx, r, log)
+	adminReply(log, w, nil, nil, err)
+}
+
+func (f *frontend) _postAdminOpenShiftClusterDeleteManagedResource(ctx context.Context, r *http.Request, log *logrus.Entry) error {
+	resType, resName, resGroupName := chi.URLParam(r, "resourceType"), chi.URLParam(r, "resourceName"), chi.URLParam(r, "resourceGroupName")
+	managedResourceID := r.URL.Query().Get("resourceid")
```
s-amann, line 32: nit: along with the above comment, can we name this query param to `managedResourceId`

s-amann (commented):
really like the real world e2e, just some nits below for consideration

pkg/frontend/adminactions/delete_managedresource.go
```diff
@@ -0,0 +1,53 @@
...
+
+	apiVersion := azureclient.APIVersion(strings.ToLower(idParts.Provider + "/" + idParts.ResourceType))
+
+	_, err = a.resources.GetByID(ctx, resourceID, apiVersion)
+	if err != nil {
+		return err
+	}
+
+	// FrontendIPConfiguration cannot be deleted with DeleteByIDAndWait (DELETE method is invalid on frontendIPConfiguration resourceID)
+	if strings.Contains(strings.ToLower(resourceID), "frontendipconfigurations") {
```
cadenmarchese, line 30: We might want to do a more accurate check on the resource type here using `idParts`, something like:
```
if strings.ToLower(idParts[5]) == "frontendipconfigurations" {
     return a.deleteFrontendIPConfiguration(ctx, resourceID)
}
```
There could be a scenario where something else has a string `frontendipconfigurations` that causes unintended behavior.

pkg/frontend/admin_openshiftcluster_delete_managedresource.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	subscriptionDoc, err := f.getSubscriptionDocument(ctx, doc.Key)
+	if err != nil {
+		return err
+	}
+
+	a, err := f.azureActionsFactory(log, f.env, doc.OpenShiftCluster, subscriptionDoc)
+	if err != nil {
+		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
```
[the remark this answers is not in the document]
SudoBrendan: This is a great question, and got me thinking, because I was also in the space of "you're right, this does seem weird".

My guess is Tony was thinking in HTTP best practice. In this case, we've got a dependent, external service that fails - what's the appropriate thing to return to the caller of _our_ API? Consensus differs, but generally people say 500/503 is a general standard if you don't have a cache to pull from.

I think our team has thought about this case a bit differently, because we're in `/admin`. Only technologically-literate people (although, I know _my_ status is up for debate) and automation will be calling this API, right? Why wrap errors in "niceties"? Well, I think the question turns to: "if we automate this, is it better for me to return 500 or 429 when I see a dependent service returns 429?"... and I think I fall on Tony's side for this. 500 seems more appropriate - I think errors should be handled as "close" as possible to the original client (retries, caching, etc).

Looking at the code block right below this I think adds some weight to this too. For a 404 case - we see Tony specifically call it out because callers of _our_ API, even automation, would care that the ResourceID they sent does not exist. I think this is good to merge, and might even be a good example for us to improve the rigidity of our `/admin` endpoints.

TL;DR - I'm +1, excellent work Tony!

SudoBrendan (commented):
Still reviewing but need a break... left some comments.

pkg/frontend/admin_openshiftcluster_delete_managedresource.go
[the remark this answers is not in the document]
tony-schndr: Now that you bring it up, I probably just saw this done in one of the many Geneva Actions already written and brought it in to this one without much thought.  I think it makes more sense to just return err and not wrap the error in `NewCloudError` since the adminReply() looks to do that by default.

pkg/frontend/adminactions/delete_managedresource.go
```diff
@@ -0,0 +1,53 @@
...
+		return a.deleteFrontendIPConfiguration(ctx, resourceID)
+	}
+
+	return a.resources.DeleteByIDAndWait(ctx, resourceID, apiVersion)
```
SudoBrendan, line 34: As programmed, this is following the acceptance criteria, but I wonder if those acceptance criteria need to be modified. I think we should at least blacklist a few resources:

1. Private Endpoints - deleting this means SRE can never access the cluster again
2. Storage Accounts - potential for customer data
3. Disks - potential for customer data

s-amann: +1 to not allow PE deletion, good catch.

pkg/util/loadbalancer/loadbalancer.go
```diff
@@ -0,0 +1,30 @@
...
+	"strings"
+
+	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
+)
+
+func RemoveFrontendIPConfiguration(lb *mgmtnetwork.LoadBalancer, resourceID string) error {
+	newFrontendIPConfig := make([]mgmtnetwork.FrontendIPConfiguration, 0, len(*lb.FrontendIPConfigurations))
+	for _, fipConfig := range *lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations {
+		if strings.EqualFold(*fipConfig.ID, resourceID) {
+			if isFrontendIPConfigReferenced(fipConfig) {
```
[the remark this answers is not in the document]
SudoBrendan: I think that messes up the flow of `continue` though?

pkg/frontend/adminactions/delete_managedresource.go
tony-schndr: I updated this to just use a regex.

pkg/frontend/adminactions/delete_managedresource.go
```diff
@@ -0,0 +1,57 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/Azure/go-autorest/autorest/azure"
+
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+	"github.com/Azure/ARO-RP/pkg/util/loadbalancer"
+)
+
+var frontendIPConfigurationPattern = `(?i)^/subscriptions/(.+)/resourceGroups/(.+)/providers/Microsoft\.Network/loadBalancers/(.+)/frontendIPConfigurations/([^/]+)$`
```
cadenmarchese, line 17: This is an improvement IMO, thanks for making the change!
I agree with others that we should never delete private endpoints and storage accounts - if possible, before merge we should add a check before line 38 that bails out if one of these resources is attempted to be deleted.

pkg/frontend/adminactions/delete_managedresource.go
tony-schndr: I added a deny list that will prevent deletion of the Private Link Service and anything under `Microsoft.Storage`. 

I also added a e2e that attempts to delete the PLS, or is that to risky?  LMK

kimorris27: Is there a `GET` that occurs on the resource before this `DELETE`? If not, you could have the E2E test use the resource ID of a non-existent PLS to test the safeguarding logic in a less risky way.

tony-schndr: That will work, I'll update with a bad name.

cadenmarchese (dismissed):

pkg/util/loadbalancer/loadbalancer.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"fmt"
+	"strings"
+
+	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
+)
+
+func RemoveFrontendIPConfiguration(lb *mgmtnetwork.LoadBalancer, resourceID string) error {
+	newFrontendIPConfig := make([]mgmtnetwork.FrontendIPConfiguration, 0, len(*lb.FrontendIPConfigurations))
+	for _, fipConfig := range *lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations {
```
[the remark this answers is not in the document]
tony-schndr: With the current flow, there is a GET against the resource ID of the frontend IP Configuration, so the pointer should never be nil at this point.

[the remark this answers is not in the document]
tony-schndr: Though it doesn't prevent other calling code from not passing in populated data, but IMO if there is a loadbalancer that doesn't have frontend IP configuration there is a bigger problem.

pkg/frontend/admin_openshiftcluster_delete_managedresource_test.go
```diff
@@ -0,0 +1,139 @@
...
+			wantStatusCode: http.StatusNotFound,
+			wantError:      fmt.Sprintf("404: NotFound: : The resource '%s' could not be found.", fmt.Sprintf("/subscriptions/%s/resourceGroups/test-cluster/providers/Microsoft.Network/publicIPAddresses/infraID-adce98f85c7dd47c5a21263a5e39c083", mockSubID)),
+		},
+		{
+			name:              "cannot delete resources in the deny list",
+			resourceID:        testdatabase.GetResourcePath(mockSubID, "resourceName"),
+			managedResourceID: fmt.Sprintf("/subscriptions/%s/resourceGroups/test-cluster/providers/Microsoft.Network/publicIPAddresses/infraID", mockSubID),
+			mocks: func(tt *test, a *mock_adminactions.MockAzureActions) {
+				a.EXPECT().ResourceDeleteAndWait(gomock.Any(), tt.managedResourceID).Return(api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "",
+					fmt.Sprintf("deletion of resource /subscriptions/%s/resourcegroups/clusterRG/providers/Microsoft.Network/privateLinkServices/infraID is forbidden", mockSubID)),
```
SudoBrendan, line 76: bug? ResourceIDs don't match.

pkg/frontend/adminactions/delete_managedresource.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+	"github.com/Azure/ARO-RP/pkg/util/loadbalancer"
+)
+
+var (
+	frontendIPConfigurationPattern = `(?i)^/subscriptions/(.+)/resourceGroups/(.+)/providers/Microsoft\.Network/loadBalancers/(.+)/frontendIPConfigurations/([^/]+)$`
+	denyList                       = []string{
+		`(?i)^/subscriptions/(.+)/resourceGroups/(.+)/providers/Microsoft\.Network/privateLinkServices/([^/]+)$`,
```
SudoBrendan, line 22: We want both PLS and PE (/providers/Microsoft.Network/privateEndpoints/)

SudoBrendan (changes requested):
Nearly there, a few quick fixes.

pkg/frontend/admin_openshiftcluster_delete_managedresource_test.go
SudoBrendan: it's even scarier, because CI passed :eye:

pkg/frontend/adminactions/delete_managedresource.go
tony-schndr: From my initial search the Private Endpoint resource does not exist in the cluster managed resource group, just the private link service (which references the PE)

Please double check.

pkg/util/loadbalancer/loadbalancer.go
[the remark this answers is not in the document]
kimorris27: That's true, but I feel we should keep in mind that even if we can't control what Azure's APIs are returning to us we can control how we respond. I agree that missing data here would definitely be an indicator of a bigger problem, but that theoretical problem will be easier to handle in prod if we error out elegantly instead of `panic`-ing.

pkg/frontend/adminactions/delete_managedresource.go
tony-schndr: Added Private Endpoints to the deny list.

SudoBrendan (approved):

pkg/frontend/adminactions/delete_managedresource.go
bennerv: PE would be the PE for the gateway, PLS would be our entrypoint into the cluster.  

Both should be blocked from deletion.

s-amann (approved):

pkg/util/loadbalancer/loadbalancer.go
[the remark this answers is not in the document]
AldoFusterTurpin: Minor detail FYI, we can avoid the nesting inverting the first if and using first the continue. (test still green :) )

```go
func RemoveFrontendIPConfiguration(lb *mgmtnetwork.LoadBalancer, resourceID string) error {
	newFrontendIPConfig := make([]mgmtnetwork.FrontendIPConfiguration, 0, len(*lb.FrontendIPConfigurations))
	for _, fipConfig := range *lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations {
		if !strings.EqualFold(*fipConfig.ID, resourceID) { //this was inverted
			newFrontendIPConfig = append(newFrontendIPConfig, fipConfig)
			continue
		}
		if isFrontendIPConfigReferenced(fipConfig) { // no need to nest
			return fmt.Errorf("frontend IP Configuration %s has external references, remove the external references prior to removing the frontend IP configuration", resourceID)
		}
	}
	lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations = &newFrontendIPConfig
	return nil
}
```

pkg/util/loadbalancer/loadbalancer.go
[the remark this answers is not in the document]
tony-schndr: @kimorris27 & @AldoFusterTurpin I'm addressing this in a follow up PR `https://github.com/Azure/ARO-RP/pull/3157`.  There are other public functions that are added that assume the pointers are safe but with the concerns raised here I'll make sure to check for nil where appropriate.

PR #3288, opened 2023-11-23
Fix not returning ARM template deploy errors

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-4753

Fixes a regression in #3222 where err was added outside the polling loop but the ``err := `` inside the polling loop was meaning that the external err was never actually set and used/returned. This meant that non-auth errors protected by the polling function would silently eat errors.

### Test plan for issue:

Manual testing locally to replicate the bug + fixing the bug
n/a


pkg/util/steps/refreshing.go
```diff
@@ -83,6 +83,7 @@ func (s *authorizationRefreshingActionStep) run(ctx context.Context, log *logrus
 			err = s.auth.Rebuild()
```
bennerv, line 83: should we suppress this err from being returned potentially?

PR #3290, opened 2023-11-24
Dynamic validation on CIDR

rhamitarora:
[ARO-3219](https://issues.redhat.com/browse/ARO-3219)

The failure of cluster provisioning occurs when an invalid network address is supplied for the CIDR. This code will introduce dynamic validation for both the cluster network (pod) CIDR range and service CIDR.

None

### DIFF
--- a/pkg/api/v20220904/openshiftcluster_validatestatic.go
+++ b/pkg/api/v20220904/openshiftcluster_validatestatic.go
@@ -193,31 +193,45 @@ func (sv openShiftClusterStaticValidator) validateServicePrincipalProfile(path s
 }
 
 func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np *NetworkProfile) error {
-	_, pod, err := net.ParseCIDR(np.PodCIDR)
+	podIP, pod, err := net.ParseCIDR(np.PodCIDR)
+
 	if err != nil {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided pod CIDR '%s' is invalid: '%s'.", np.PodCIDR, err)
 	}
+
 	if pod.IP.To4() == nil {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided pod CIDR '%s' is invalid: must be IPv4.", np.PodCIDR)
 	}
-	{
-		ones, _ := pod.Mask.Size()
-		if ones > 18 {
-			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided vnet CIDR '%s' is invalid: must be /18 or larger.", np.PodCIDR)
-		}
+
+	ones, _ := pod.Mask.Size()
+	if ones > 18 {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided vnet CIDR '%s' is invalid: must be /18 or larger.", np.PodCIDR)
 	}
-	_, service, err := net.ParseCIDR(np.ServiceCIDR)
+
+	nip := podIP.Mask(pod.Mask)
+
+	if nip.String() != podIP.String() {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidNetworkAddress, path+".podCidr", "The provided pod CIDR '%s' is invalid, expecting: '%s/%d'.", np.PodCIDR, nip.String(), ones)
+	}
+
+	serviceIP, service, err := net.ParseCIDR(np.ServiceCIDR)
 	if err != nil {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".serviceCidr", "The provided service CIDR '%s' is invalid: '%s'.", np.ServiceCIDR, err)
 	}
+
 	if service.IP.To4() == nil {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".serviceCidr", "The provided service CIDR '%s' is invalid: must be IPv4.", np.ServiceCIDR)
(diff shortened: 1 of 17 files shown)


pkg/validate/dynamic/dynamic.go
```diff
@@ -576,10 +577,25 @@ func (dv *dynamic) validateCIDRRanges(ctx context.Context, subnets []Subnet, add
...
+		}
+
+		nip := IP.Mask(cidr.Mask)
+		if nip.String() != IP.String() {
+			return api.NewCloudError(
+				http.StatusBadRequest,
+				api.CloudErrorCodeInvalidNetworkAddress,
+				"",
+				errMsgInvalidCIDR,
+				err,
```
cadenmarchese, line 596: Your error message contains 2 formatting directives (%s and %s), but I only see one string passed in here (err).

Also, I don't think that the var `err` will be set here, since you are just doing a comparison. Your error would say something like `Invalid network address got "", expecting %s.`

Maybe we need something like this?
```
	return api.NewCloudError(
				http.StatusBadRequest,
				api.CloudErrorCodeInvalidNetworkAddress,
				"",
				errMsgInvalidCIDR,
				nip.String(),
                                IP.String()
```

bennerv (changes requested):
Azure does validation of the subnets created under a vnet to ensure they're created successfully.

We only have to do validation of the Pod and Service CIDRs passed in by the customer.  We should be doing this as part of static validation, not dynamic.
`pkg/api/v202XXXXX/openshiftcluster_validatestatic.go` `validateNetworkProfile`.  That's the only place we need changes.

pkg/validate/dynamic/dynamic.go
rhamitarora: Done

bennerv (changes requested):
All the API definitions should have this static validation.

pkg/api/v20220904/openshiftcluster_validatestatic.go
```diff
@@ -193,10 +193,17 @@ func (sv openShiftClusterStaticValidator) validateServicePrincipalProfile(path s
 }
 
 func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np *NetworkProfile) error {
-	_, pod, err := net.ParseCIDR(np.PodCIDR)
+	podIP, pod, err := net.ParseCIDR(np.PodCIDR)
 	if err != nil {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided pod CIDR '%s' is invalid: '%s'.", np.PodCIDR, err)
 	}
+	{
```
bennerv, line 200: Why is this in an empty block?

pkg/api/v20220904/openshiftcluster_validatestatic.go
```diff
@@ -206,10 +213,18 @@ func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np
 			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided vnet CIDR '%s' is invalid: must be /18 or larger.", np.PodCIDR)
 		}
 	}
-	_, service, err := net.ParseCIDR(np.ServiceCIDR)
+
+	serviceIP, service, err := net.ParseCIDR(np.ServiceCIDR)
 	if err != nil {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".serviceCidr", "The provided service CIDR '%s' is invalid: '%s'.", np.ServiceCIDR, err)
 	}
+	{
```
bennerv, line 221: same here, no need to put these in blocks

pkg/api/v20220904/openshiftcluster_validatestatic_test.go
```diff
@@ -486,6 +486,13 @@ func TestOpenShiftClusterStaticValidateNetworkProfile(t *testing.T) {
 			},
 			wantErr: "400: InvalidParameter: properties.networkProfile.serviceCidr: The provided vnet CIDR '10.0.0.0/23' is invalid: must be /22 or larger.",
 		},
+		{
+			name: "podCidr invalid network",
+			modify: func(oc *OpenShiftCluster) {
+				oc.Properties.NetworkProfile.PodCIDR = "10.254.0.0/14"
+			},
+			wantErr: "400: InvalidNetworkAddress: properties.networkProfile.podCidr: The provided pod CIDR '10.254.0.0/14' is invalid, expecting '10.252.0.0'.",
```
bennerv, line 494: We may want to keep the same `/14` mask on the end of it.

pkg/api/v20220904/openshiftcluster_validatestatic_test.go
```diff
@@ -486,6 +486,13 @@ func TestOpenShiftClusterStaticValidateNetworkProfile(t *testing.T) {
...
+		},
```
bennerv, line 495: We should also have a test case for Service CIDR.

pkg/api/v20220904/openshiftcluster_validatestatic.go
rhamitarora: Done removed

pkg/api/v20220904/openshiftcluster_validatestatic.go
rhamitarora: Done removed

pkg/api/v20220904/openshiftcluster_validatestatic_test.go
rhamitarora: Done

bennerv (dismissed):
lgtm

bennerv (changes requested):
We need this change reflected in all our API versions which support the `validateNetworkProfile` function in our static validation.  Right now it's only in v20220904, but it should be in all our API versions.

bennerv (changes requested):
should add preview API validation as well

bennerv (approved):
lgtm

cadenmarchese (approved):
thanks for the additional validation step!

PR #3292, opened 2023-11-27
ARO Operator - Reduce unnecessary reconciles by limiting watched resources/changes

tsatam:

Another proposed fix for [ARO-4632](https://issues.redhat.com/browse/ARO-4632)
May help with [ARO-5216](https://issues.redhat.com/browse/ARO-5216) as well

Various changes to how the ARO operator watches resources:

- Extract all common predicates used by controllers to shared package
- Only watch for generation change in ARO cluster resource on most controllers (avoid Reconciles when just status changes)

### Test plan for issue:

- Unit tests continue to pass
- E2E tests continue to pass
- Operator functionality not covered by existing tests but likely to be impacted by the change have new tests (TODO)

Note - I am expecting some E2E tests to become _more_ flaky because of this change, as we have many E2E tests that are dependent on the operator reconciling more often than is necessary, and expecting an operator controller to perform changes without actually modifying any of the resources that controller is actually watching. I am planning to fix up any observed occurrences of such test flakes with individual commits on this branch for each test/controller, and each commit will likely contain changes to both the tests as well as the watched resources for that controller. 

N/A

### Additional ideas not implemented in this PR (yet?)
- [ ] For the majority of controllers - only watch for changes on a specific set of cluster flags on the ARO cluster resource, rather than the whole spec
- [ ] Similar to above, implement more targeted Watches for other watched resources that only reconcile when the specific fields actually used by the controller change

### DIFF
--- a/pkg/operator/controllers/autosizednodes/autosizednodes_controller.go
+++ b/pkg/operator/controllers/autosizednodes/autosizednodes_controller.go
@@ -16,6 +16,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/Azure/ARO-RP/pkg/operator"
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
@@ -103,7 +104,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	// Controller adds ControllerManagedBy to KubeletConfit created by this controller.
 	// Any changes will trigger reconcile, but only for that config.
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&arov1alpha1.Cluster{}, builder.WithPredicates(predicates.AROCluster)).
+		For(&arov1alpha1.Cluster{}, builder.WithPredicates(predicate.And(predicates.AROCluster, predicate.GenerationChangedPredicate{}))).
 		Owns(&mcv1.KubeletConfig{}).
 		Named(ControllerName).
 		Complete(r)
--- a/pkg/operator/controllers/autosizednodes/autosizednodes_controller.go
+++ b/pkg/operator/controllers/autosizednodes/autosizednodes_controller.go
@@ -6,7 +6,6 @@ package autosizednodes
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	"github.com/Azure/go-autorest/autorest/to"
 	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
@@ -17,10 +16,10 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/Azure/ARO-RP/pkg/operator"
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+	"github.com/Azure/ARO-RP/pkg/operator/predicates"
 )
 
 type Reconciler struct {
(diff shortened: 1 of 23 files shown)


SudoBrendan (approved):

tsatam:
I'm validating any flake impact to e2e on this, and will include any required fixups on this PR. Adding hold until that is complete.

pkg/operator/controllers/autosizednodes/autosizednodes_controller.go
```diff
@@ -101,19 +101,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
-	})
-
-	b := ctrl.NewControllerManagedBy(mgr).
-		For(&arov1alpha1.Cluster{}, builder.WithPredicates(clusterPredicate))
-
-		// Controller adds ControllerManagedBy to KubeletConfit created by this controller.
-		// Any changes will trigger reconcile, but only for that config.
-	return b.
-		Named(ControllerName).
+	// Controller adds ControllerManagedBy to KubeletConfit created by this controller.
```
mociarain, line 104:
```suggestion
	// Controller adds ControllerManagedBy to KubeletConfig created by this controller.
```
I assume

pkg/operator/controllers/storageaccounts/storageaccount_controller.go
```diff
@@ -111,18 +112,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
-		return ok && role == "master"
-	})
-
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&arov1alpha1.Cluster{}, builder.WithPredicates(aroClusterPredicate)).
-		Watches(&source.Kind{Type: &machinev1beta1.Machine{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(masterMachinePredicate)). // to reconcile on master machine replacement
-		Watches(&source.Kind{Type: &machinev1beta1.MachineSet{}}, &handler.EnqueueRequestForObject{}).                                              // to reconcile on worker machinesets
+		For(&arov1alpha1.Cluster{}, builder.WithPredicates(predicate.And(predicates.AROCluster, predicate.GenerationChangedPredicate{}))).
+		Watches(&source.Kind{Type: &machinev1beta1.Machine{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(predicates.MachineRoleMaster)).    // to reconcile on master machine replacement
+		Watches(&source.Kind{Type: &machinev1beta1.MachineSet{}}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(predicates.MachineRoleWorker)). // to reconcile on worker machines
```
mociarain, line 118: [Borderline Rant - Not Blocking Anything :upside_down_face:  ] 

These 2 lines are very long. I know Go doesn't have a max line length but we live in a physical world with fixed screens.  A quick `grep '.\{120\}' -r ./pkg/*/*.go | wc -l` tells me we have > 1000 lines longer than 120.

- Have we ever considered adding a max line length?
- Has this ever bothered anyone else?
- Do I need to stop worrying and learn to love some line wrapping plugin???

mociarain (approved):
Looks great. 2 thoughts that don't block this at all

- A quick grep shows me that we have 29 instances of `NewControllerManagedBy` in the repo and this PR will make more than half use the `GenerationChangedPredicate`. I don't have a great sense of how big a problem all these extra calls were/are/will be but did we ever consider wrapping the controllerBuilder and changing the default from "load" to "quiet"?
- I don't usually like this pattern but what would you say about wrapping `GenerationChangedPredicate` in something that lives with the other `predicates` for discoverability? Since the default with the controllers is to shout and we need to add this force them to only fire on "real" changes making that option very clear seems like it might be worth it. What do you think?

tsatam:
>     * A quick grep shows me that we have 29 instances of `NewControllerManagedBy` in the repo and this PR will make more than half use the `GenerationChangedPredicate`. I don't have a great sense of how big a problem all these extra calls were/are/will be but did we ever consider wrapping the controllerBuilder and changing the default from "load" to "quiet"?

I'm not sure if it's possible or even desirable to make this change a "default", since GenerationChangedPredicate here is being applied specifically to the ARO cluster resource, not all resources. I also personally prefer to keep the composition of watch rules within each controller, so you can refer to them at-a-glance when making changes to them, rather than having to refer to multiple layers of abstraction to put together the rules. I think your second point is probably more preferable for reusability and deduplication here. 

>     * I don't usually like this pattern but what would you say about wrapping `GenerationChangedPredicate` in something that lives with the other `predicates` for discoverability? Since the default with the controllers is to shout and we need to add this force them to only fire on "real" changes making that option very clear seems like it might be worth it. What do you think?

Along the same lines as my personal preference above, I think just extracting out the `GenerationChangedPredicate` wouldn't help discoverability here. I do have some ideas for extracting out this specific predicate, but I need to determine a path forward and choose one:
- Extract `predicate.And(predicates.AROCluster, predicate.GenerationChangedPredicate{}))` into a shared/reusable token (simplest option) 
- Make more opinionated functions that return predicates (names can be workshopped, they're intentionally verbose for now):
  - `ChangesInAROFeatureFlags(flags []string)`
  - `ChangesInSpecificAROSpecProperty()` (for the few controllers that do act on specific fields in the CRD)
  - `ChangesInAROStatus()` (only used by ClusterOperatorARO controller to propagate controller statuses to operator status)

mociarain:
> I'm not sure if it's possible or even desirable to make this change a "default", since GenerationChangedPredicate here is being applied specifically to the ARO cluster resource, not all resources.

Sorry I wasn't clear. I was suggesting something like a `NewAROControllerManagedBy` that just adds `GenerationChangedPredicate` but I take your point all the same.  

> Along the same lines as my personal preference above, I think just extracting out the GenerationChangedPredicate wouldn't help discoverability here. 

Yeah I buy this. I thought after this PR we just used `predicates` and `predicate.GenerationChangedPredicate` and could collapse them to one place but `predicate` is all over. 

Thanks for explaining that and FWIW your suggestions make sense to me.

mociarain:
This changes look great. What's stopping us merging it, since the E2E tests passed? Would we like to run the E2E pipeline against it to get a larger sample size? If so we have a new joiner and this would be a nice task for them to see some parts of the system.

What do you think @tsatam?

tsatam:
> This changes look great. What's stopping us merging it, since the E2E tests passed? Would we like to run the E2E pipeline against it to get a larger sample size? If so we have a new joiner and this would be a nice task for them to see some parts of the system.
> 
> What do you think @tsatam?

I think this change is mergeable, there are some future follow-up ideas in the description of the PR but they can be done later. This work was ultimately de-prioritized due to being fairly low on the impact/effort ratio, we ended up remediating the issue this PR seeked to reduce through more direct changes, but it is still a net positive IMO.

PR #3298, opened 2023-11-29
CIF - ServerError: storage.AccountsClient#ListAccountSAS: TooManyRequests

LiniSusan:
https://issues.redhat.com/browse/ARO-3804

Fixes

To report the cluster installation failure error related to storage.AccountsClient#ListAccountSAS: Failure responding to request: StatusCode=429

### DIFF
--- a/pkg/util/storage/manager.go
+++ b/pkg/util/storage/manager.go
@@ -5,6 +5,7 @@ package storage
 
 import (
 	"context"
+	"net/http"
 	"net/url"
 	"time"
 
@@ -13,6 +14,7 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/date"
 
+	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/storage"
 )
@@ -33,6 +35,30 @@ func NewManager(env env.Core, subscriptionID string, authorizer autorest.Authori
 	}
 }
 
+func getCorrectErrWhenTooManyRequests(err error) error {
+	detailedError, ok := err.(autorest.DetailedError)
+	if !ok {
+		return err
+	}
+	if detailedError.StatusCode != http.StatusTooManyRequests {
+		return err
+	}
+	msg := "Requests are being throttled due to Azure Storage limits being exceeded. Please visit https://learn.microsoft.com/en-us/azure/openshift/troubleshoot#exceeding-azure-storage-limits for more details."
+	cloudError := &api.CloudError{
+		StatusCode: http.StatusTooManyRequests,
+		CloudErrorBody: &api.CloudErrorBody{
+			Code:    api.CloudErrorCodeThrottlingLimitExceeded,
+			Message: "ThrottlingLimitExceeded",
+			Details: []api.CloudErrorBody{
+				{
+					Message: msg,
+				},
+			},
+		},
+	}
+	return cloudError
+}
+
 func (m *manager) BlobService(ctx context.Context, resourceGroup, account string, p mgmtstorage.Permissions, r mgmtstorage.SignedResourceTypes) (*azstorage.BlobStorageClient, error) {
 	t := time.Now().UTC().Truncate(time.Second)
 	res, err := m.storageAccounts.ListAccountSAS(ctx, resourceGroup, account, mgmtstorage.AccountSasParameters{
@@ -44,7 +70,7 @@ func (m *manager) BlobService(ctx context.Context, resourceGroup, account string
 		SharedAccessExpiryTime: &date.Time{Time: t.Add(24 * time.Hour)},
 	})
 	if err != nil {
-		return nil, err
+		return nil, getCorrectErrWhenTooManyRequests(err)
 	}
 
 	v, err := url.ParseQuery(*res.AccountSasToken)
--- a/pkg/api/error.go
+++ b/pkg/api/error.go
@@ -103,6 +103,7 @@ const (
 	CloudErrorCodeScopeLocked                        = "ScopeLocked"
 	CloudErrorCodeRequestDisallowedByPolicy          = "RequestDisallowedByPolicy"
 	CloudErrorCodeInvalidNetworkAddress              = "InvalidNetworkAddress"
+	CloudErrorCodeThrottlingLimitExceeded            = "ThrottlingLimitExceeded"
 )
 
 // NewCloudError returns a new CloudError


pkg/util/steps/runner.go
```diff
@@ -54,6 +56,14 @@ func Run(ctx context.Context, log *logrus.Entry, pollInterval time.Duration, ste
 
 		if err != nil {
 			log.Errorf("step %s encountered error: %s", step, err.Error())
+			if strings.Contains(err.Error(), "ThrottlingLimitExceeded") {
+				err = api.NewCloudError(
+					http.StatusTooManyRequests,
+					api.CloudErrorCodeThrottlingLimitExceeded,
+					"",
+					"Requests are being throttled due to Azure Storage limits being exceeded. Please visit https://learn.microsoft.com/en-us/azure/openshift/troubleshoot for more details.")
```
swiencki, line 64: Can you update the link to this PR when it lands - https://github.com/MicrosoftDocs/azure-docs/pull/117572

pkg/util/storage/manager.go
```diff
@@ -43,7 +46,23 @@ func (m *manager) BlobService(ctx context.Context, resourceGroup, account string
...
 	if err != nil {
+		if strings.Contains(err.Error(), "TooManyRequests") { 
+				err = &api.CloudError{
+					StatusCode: http.StatusTooManyRequests,
+					CloudErrorBody: &api.CloudErrorBody{
+						Code:    api.CloudErrorCodeThrottlingLimitExceeded,
+						Message: "ThrottlingLimitExceeded",
+						Details: []api.CloudErrorBody{
+							{
+								Message: string("Requests are being throttled due to Azure Storage limits being exceeded. Please visit https://learn.microsoft.com/en-us/azure/openshift/troubleshoot for more details."),
```
swiencki, line 59: Can you update the link to https://learn.microsoft.com/en-us/azure/openshift/troubleshoot#exceeding-azure-storage-limits

pkg/util/steps/runner.go
LiniSusan: Updated

pkg/util/storage/manager.go
LiniSusan: updated. Thanks @swiencki

pkg/util/steps/runner.go
```diff
@@ -54,6 +57,16 @@ func Run(ctx context.Context, log *logrus.Entry, pollInterval time.Duration, ste
 
 		if err != nil {
 			log.Errorf("step %s encountered error: %s", step, err.Error())
+			if strings.Contains(err.Error(), "ThrottlingLimitExceeded") || strings.Contains(err.Error(), "TooManyRequests") {
+				correlation_id := fmt.Sprintf("%v", log.Data["correlation_id"])
+				message := fmt.Sprintf("Requests are being throttled due to Azure Storage limits being exceeded. Please visit https://learn.microsoft.com/en-us/azure/openshift/troubleshoot#exceeding-azure-storage-limits for more details. CorrelationId: " + correlation_id)
```
SrinivasAtmakuri, line 62:
```suggestion
				message := fmt.Sprintf("Requests are throttled due to Azure Storage limits being exceeded. Please visit https://learn.microsoft.com/en-us/azure/openshift/troubleshoot#exceeding-azure-storage-limits for more details. CorrelationId: " + correlation_id)
```

pkg/util/storage/manager.go
```diff
@@ -43,7 +46,22 @@ func (m *manager) BlobService(ctx context.Context, resourceGroup, account string
...
 	if err != nil {
+		if strings.Contains(err.Error(), "TooManyRequests") {
+			err = &api.CloudError{
+				StatusCode: http.StatusTooManyRequests,
+				CloudErrorBody: &api.CloudErrorBody{
+					Code:    api.CloudErrorCodeThrottlingLimitExceeded,
+					Message: "ThrottlingLimitExceeded",
+					Details: []api.CloudErrorBody{
+						{
+							Message: string("Requests are being throttled due to Azure Storage limits being exceeded. Please visit https://learn.microsoft.com/en-us/azure/openshift/troubleshoot#exceeding-azure-storage-limits for more details."),
```
SrinivasAtmakuri, line 59: LGTM, a small nit,

```suggestion
							Message: string("Requests are throttled due to Azure Storage limits being exceeded. Please visit https://learn.microsoft.com/en-us/azure/openshift/troubleshoot#exceeding-azure-storage-limits for more details."),
```

pkg/util/steps/runner.go
swiencki: Any details on why the error is happening should be within the link. I think this is too wordy

rhamitarora (approved):
LGTM

pkg/util/steps/runner.go
jaitaiwan: `Requests are being throttled` is colloquial english. So that was fine wording. To shorten it you might be able to do something like `Requests Throttled. Visit () for more info. CorrelationId: `

jaitaiwan (approved):
I'm fine with longer errors, I prefer verbosity unless we can fit more information in less characters. Gave a suggestion for a shorter error message to satisfy Simon's request. LGTM!

pkg/util/storage/manager.go
```diff
@@ -44,6 +47,20 @@ func (m *manager) BlobService(ctx context.Context, resourceGroup, account string
 		SharedAccessExpiryTime: &date.Time{Time: t.Add(24 * time.Hour)},
 	})
 	if err != nil {
+		if strings.Contains(err.Error(), "TooManyRequests") {
```
mociarain, line 50: Why match on a sting in the error message and not the `res.Response.StatusCode`?

LiniSusan: I dont think there is much impact when it is changed to `if err.(autorest.DetailedError).StatusCode == "TooManyRequests"`

mociarain: I was thinking more about matching against the number. Like we do here: https://github.com/azure/ARO-RP/blob/7cbf95b811d79677594809f90c8e45681a382b49/pkg/database/cosmosdb/zz_generated_cosmosdb.go#L73

This way if they change "TooManyRequest" -> "Too Many Requests" or whatever we're still good. IMO Matching on the statusCode is more robust. 

What do you think?

pkg/util/storage/manager.go
LiniSusan: We generated this error message with reference to the original message that we go from Azure aside : "The request is being throttled as the limit has been reached for operation type - Read_ObservationWindow_00:05:00. For more information, see - https://aka.ms/srpthrottlinglimits"

pkg/util/steps/runner.go
LiniSusan: We generated this error message with reference to the original message that we go from Azure errors : "The request is being throttled as the limit has been reached for operation type - Read_ObservationWindow_00:05:00. For more information, see - https://aka.ms/srpthrottlinglimits"

pkg/util/storage/manager.go
LiniSusan: @mociarain : could you please review the code now? 
let me know if this is more convincing or not

mociarain (approved):
LGTM :)

pkg/util/storage/manager.go
mociarain: That's exactly what I had in mind

pkg/util/storage/manager.go
```diff
@@ -44,6 +46,22 @@ func (m *manager) BlobService(ctx context.Context, resourceGroup, account string
...
+						Message: "ThrottlingLimitExceeded",
+						Details: []api.CloudErrorBody{
+							{
+								Message: string("Requests are being throttled due to Azure Storage limits being exceeded. Please visit https://learn.microsoft.com/en-us/azure/openshift/troubleshoot#exceeding-azure-storage-limits for more details."),
+							},
+						},
+					},
+				}
+			}
+		}
```
AldoFusterTurpin, line 64: I think we could extract that to an auxiliar function and leverage the concept of early return to simplify the levels of nesting of the block of code. It is pretty common in Go to exit as soon as possible and I think we can take advantage of it 🙂 
Line 65 could become 
```go
return nil, getCorrectErrWhenTooManyRequests(err)
```
and getCorrectErrWhenTooManyRequests would be something like
```go
func getCorrectErrWhenTooManyRequests(err error) error {
	detailedError, ok := err.(autorest.DetailedError)
	if !ok {
		return err
	}

	if detailedError.StatusCode != http.StatusTooManyRequests {
		return err
	}

	msg := "Requests are being throttled due to Azure Storage limits being exceeded. Please visit https://learn.microsoft.com/en-us/azure/openshift/troubleshoot#exceeding-azure-storage-limits for more details."
	cloudError := &api.CloudError{
		StatusCode: http.StatusTooManyRequests,
		CloudErrorBody: &api.CloudErrorBody{
			Code:    api.CloudErrorCodeThrottlingLimitExceeded,
			Message: "ThrottlingLimitExceeded",
			Details: []api.CloudErrorBody{
				{
					Message: string(msg),
				},
			},
		},
	}

	return cloudError
}
```

AldoFusterTurpin (dismissed):
@LiniSusan  That's good. Thank you for this. 
I left a suggestion to leverage early returns to make the code a bit more Go-ish, reducing the cyclomatic complexity.
🙂

pkg/util/storage/manager.go
LiniSusan: updated the code with the mentioned changes..

rhamitarora:
LGTM

rhamitarora (approved):

pkg/util/storage/manager.go
```diff
@@ -33,6 +35,32 @@ func NewManager(env env.Core, subscriptionID string, authorizer autorest.Authori
...
 
+func getCorrectErrWhenTooManyRequests(err error) error {
+	detailedError, ok := err.(autorest.DetailedError)
+	if !ok {
+		return err
+	}
+
+	if detailedError.StatusCode != http.StatusTooManyRequests {
+		return err
+	} else {
```
AldoFusterTurpin, line 46: One last comment before LGTM, the else is not necessary due to the previous return (we can avoid one nested level then). One of the benefits of having a separate function for this 🙂

AldoFusterTurpin (changes requested):
One last minor thing and then LGTM.

pkg/util/storage/manager.go
AldoFusterTurpin, line 46: We can get rid of this else but just a minor thing.

AldoFusterTurpin (approved):
LGTM. One minor suggestions but definitely not a blocker 🙂

swiencki:
Looks good :+1:

jaitaiwan (approved):
LGTM

SrinivasAtmakuri (approved):

Shivkumar13 (approved):
LGTM

Shivkumar13:
Waiting for CI to go green, post that good to merge.

mociarain (approved):

PR #3299, opened 2023-11-29
Addressing registry conflicts in Makefile

Shivkumar13:

Fixes the issue with the `RP_IMAGE_ACR` env, as when `$(RP_IMAGE_ACR)`  is empty, the `$ARO_IMAGE_BASE` would be `.azurecr.io/aro`, which will fail any operations that we do with the image.

Encountered this issue while the `RP_IMAGE_ACR` is not set, and also the else if conditions seems unnecessary as we are defaulting to registry.access.redhat.com in the else block so no need to check `RP_IMAGE_ACR` explicitly  for the empty value. 

### Test plan for issue:
N/A

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 SHELL = /bin/bash
 TAG ?= $(shell git describe --exact-match 2>/dev/null)
 COMMIT = $(shell git rev-parse --short=7 HEAD)$(shell [[ $$(git status --porcelain) = "" ]] || echo -dirty)
-ARO_IMAGE_BASE = ${RP_IMAGE_ACR}.azurecr.io/aro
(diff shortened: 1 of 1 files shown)


jhoreman (approved):
Seems straightforward enough. Thanks for this.

anshulvermapatel (approved):
/lgtm

SudoBrendan (changes requested):
I worry about this change because it has wider implications to our pipelines, where this variable is set and overridden, e.g. https://github.com/Azure/ARO-RP/blob/master/.pipelines/e2e.yml#L42

What is the pain we are trying to mitigate here? I feel like if someone doesn't run `make secrets` + `. ./env` other issues will also arise outside this.

mociarain:
Closing this since it's stale and has no associated ticket. Push back if this seems worth pursuing

PR #3310, opened 2023-12-04
Delete the deny assignment first during cluster deletion

cadenmarchese:

https://issues.redhat.com/browse/ARO-4427

If a cluster deletion fails for whatever reason, SRE needs to break glass to clean up cluster resources due to our deny assignment preventing manual resource deletion.

https://github.com/Azure/ARO-RP/pull/3286 added the ability for SRE to delete resources without breaking glass, and this PR is meant to enable customers to clean up resources on their end without SRE intervention.

azure-sdk-for-go doesn't provide any client to do this, so this PR creates a client that does it similarly to our manual process: https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/587166/Delete-Deny-Assignment 

### Test plan for issue:

Deploy to canary, confirm the deny assignment is deleted first. Since this relies on a cluster being in a deleting state to work, we won't be able to test this with e2e.

I should be able to test this once the current release is complete, but I want to put up this draft so it can get some eyes and feedback on the client.

### DIFF
--- a/pkg/util/azureclient/mgmt/authorization/denyassignment_addons.go
+++ b/pkg/util/azureclient/mgmt/authorization/denyassignment_addons.go
@@ -12,18 +12,18 @@ import (
 	policy "github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	runtime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
-	"github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-09-01-preview/authorization"
+	mgmtauthorization "github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-09-01-preview/authorization"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 )
 
 // DenyAssignmentClientAddons contains addons for DenyAssignmentClient
 type DenyAssignmentClientAddons interface {
-	ListForResourceGroup(ctx context.Context, resourceGroupName string, filter string) (result []authorization.DenyAssignment, err error)
+	ListForResourceGroup(ctx context.Context, resourceGroupName string, filter string) (result []mgmtauthorization.DenyAssignment, err error)
 	DeleteDenyAssignment(ctx context.Context, fpTokenCredential *azidentity.ClientCertificateCredential, subscriptionDoc *api.SubscriptionDocument, doc *api.OpenShiftClusterDocument) error
 }
 
-func (c *denyAssignmentClient) ListForResourceGroup(ctx context.Context, resourceGroupName string, filter string) (result []authorization.DenyAssignment, err error) {
+func (c *denyAssignmentClient) ListForResourceGroup(ctx context.Context, resourceGroupName string, filter string) (result []mgmtauthorization.DenyAssignment, err error) {
 	page, err := c.DenyAssignmentsClient.ListForResourceGroup(ctx, resourceGroupName, filter)
 	if err != nil {
 		return nil, err
@@ -77,7 +77,7 @@ func (c *denyAssignmentClient) DeleteDenyAssignment(ctx context.Context, fpToken
 	}
 
 	// Get the deny assignment
-	var denyAssignment authorization.DenyAssignment
+	var denyAssignment mgmtauthorization.DenyAssignment
 	get, err := c.getDenyAssignmentRequest(ctx, *client, subscriptionDoc, doc)
 	if err != nil {
 		return err
@@ -88,7 +88,7 @@ func (c *denyAssignmentClient) DeleteDenyAssignment(ctx context.Context, fpToken
 	}
 	defer httpResp.Body.Close()
 
-	// Marshall response into type authorization.DenyAssignment
+	// Marshal response into type authorization.DenyAssignment
 	err = json.NewDecoder(httpResp.Body).Decode(&denyAssignment)
 	if err != nil {
 		return err
--- a/pkg/util/azureclient/mgmt/authorization/denyassignment_addons.go
+++ b/pkg/util/azureclient/mgmt/authorization/denyassignment_addons.go
@@ -5,9 +5,8 @@ package authorization
 
 import (
 	"context"
-	"encoding/json"
-	"fmt"
 	"net/http"
+	"strings"
 
 	policy "github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	runtime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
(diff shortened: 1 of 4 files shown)


pkg/util/azureclient/mgmt/authorization/denyassignment_addons.go
```diff
@@ -5,16 +5,25 @@ package authorization
...
 	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
 
-	mgmtauthorization "github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-09-01-preview/authorization"
+	policy "github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	runtime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-09-01-preview/authorization"
```
kimorris27, line 15: Any particular reason for renaming this import?

pkg/util/azureclient/mgmt/authorization/denyassignment_addons.go
```diff
@@ -30,3 +39,71 @@ func (c *denyAssignmentClient) ListForResourceGroup(ctx context.Context, resourc
 
 	return result, nil
 }
+
+// getCreateRequest creates the deny assignment get request.
+func (c *denyAssignmentClient) getDenyAssignmentRequest(ctx context.Context, client DenyAssignmentsARMClient, subscriptionDoc *api.SubscriptionDocument, doc *api.OpenShiftClusterDocument) (*policy.Request, error) {
```
kimorris27, line 44: Even though the sdk's client doesn't have a delete method, could we still use it's get method?

https://pkg.go.dev/github.com/Azure/azure-sdk-for-go@v63.1.0+incompatible/services/preview/authorization/mgmt/2018-09-01-preview/authorization#DenyAssignmentsClient.Get

pkg/util/azureclient/mgmt/authorization/denyassignment_addons.go
```diff
@@ -30,3 +39,71 @@ func (c *denyAssignmentClient) ListForResourceGroup(ctx context.Context, resourc
...
+	if err != nil {
+		return err
+	}
+	httpResp, err := client.internal.Pipeline().Do(get)
+	if err != nil {
+		return err
+	}
+	defer httpResp.Body.Close()
+
+	// Marshall response into type authorization.DenyAssignment
```
kimorris27, line 91: Super nitpicky:`Marshall` -> `Marshal`

kimorris27 (changes requested):
Good idea to copy what the legit sdk does internally, it made the new methods easy to review.

Requesting changes because I think we can use azure-sdk-for-go's get method instead of rolling our own.

pkg/util/azureclient/mgmt/authorization/denyassignment_addons.go
kimorris27: Actually, is this get call before the delete call even needed? What's it for?

cadenmarchese: It looks like I was wrong about this. We randomly generate a name (guid) for the deny assignment using this ARM function: https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-string#guid and we don't store it anywhere, so we do need a Get to determine the deny assignment name

cadenmarchese: I removed the custom `Get` and replaced it with our custom `ListForResourceGroup`. Let me know what you think!

kimorris27: Yeah, that works too! Just wanted to find a way to remove as much custom stuff as we could.

pkg/util/azureclient/mgmt/authorization/denyassignment_addons.go
```diff
@@ -30,3 +38,50 @@ func (c *denyAssignmentClient) ListForResourceGroup(ctx context.Context, resourc
...
+
+	managedResourceGroupName := strings.Split(doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, "/")[4]
+	denyAssignmentList, err := c.ListForResourceGroup(ctx, managedResourceGroupName, "")
+	if err != nil {
+		return err
+	}
+
+	// There should never be more than one deny assignment, but in case there is, don't proceed
+	if len(denyAssignmentList) > 1 {
+		return nil
```
kimorris27, line 56: I'm not sure how we would ever end up with more than one deny assignment, but just in case, I think it would be wise to return an error here to ensure something gets logged.

kimorris27 (changes requested):

kimorris27:
Other than the small change I requested, LGTM assuming the test in canary goes well.

pkg/util/azureclient/mgmt/authorization/denyassignment_addons.go
```diff
@@ -30,3 +38,50 @@ func (c *denyAssignmentClient) ListForResourceGroup(ctx context.Context, resourc
 
 	return result, nil
 }
+
+func (c *denyAssignmentClient) DeleteDenyAssignment(ctx context.Context, fpTokenCredential *azidentity.ClientCertificateCredential, subscriptionDoc *api.SubscriptionDocument, doc *api.OpenShiftClusterDocument) error {
```
dkeohane, line 42: minor: could just pass in the subscriptionDoc.ID directly instead of the entire document given it's the only thing from it which is used

dkeohane: Same could apply to doc for creating:
managedResourceGroupName := strings.Split(doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, "/")[4]

pkg/util/azureclient/mgmt/authorization/denyassignment_addons.go
```diff
@@ -30,3 +38,50 @@ func (c *denyAssignmentClient) ListForResourceGroup(ctx context.Context, resourc
...
+	deleteClient, err := NewDenyAssignmentsARMClient(subscriptionDoc.ID, fpTokenCredential, nil)
+	if err != nil {
+		return err
+	}
+
+	managedResourceGroupName := strings.Split(doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, "/")[4]
```
dkeohane, line 48: sorry, I'm entirely new, so just want to double check - can we be certain that the ResourceGroupID when split will have 5 or more results?

PR #3313, opened 2023-12-05
Disable dnsmasq restart script for new clusters

cadenmarchese:

Fixes https://issues.redhat.com/browse/ARO-4803

Disables the restart dnsmasq operator flag so that the dnsmasq script isn't applied to new clusters. The path forward for clusters experiencing dnsmasq issues from Azure host servicing events will be to upgrade to OpenShift 4.11.30+ which contains the fix.

### Test plan for issue:

- [x] Install a new cluster with the changes, confirm that the script is not applied
- [ ] Install a new cluster without the changes, apply these changes to confirm that a new machineconfig rollout doesn't occur

The plan for the time being is to send comms to affected clusters.


s-amann (approved):

bennerv (approved):
lgtm

kimorris27 (approved):

bennerv:
putting a /hold on this for now while we figure out the path forward on outstanding OCPBUG.

PR #3318, opened 2023-12-07
Update purge script (hack/clean/clean.go)

cadenmarchese:

Fixes https://issues.redhat.com/browse/ARO-4856

- Delete clusters with a managedBy field
  - Our purge script is not deleting resource groups that have the managedBy field. Previously, we were assuming that any resource group with a managedBy field was a prod cluster, but dev clusters can/do have that field, and we should be deleting them.
  - From now on, we use `purge:true` only as the source of truth for which clusters are dev
- Add `shared-cluster` to the deny list, so it doesn't get wiped out by this change
  - Prod clusters that we want to persist should already be tagged `persist: true` and not have the `purge` tag, so they shouldn't get deleted. 
- Move to infof in the script, so we see these messages during a pipeline run (if something isn't getting deleted, we want to know why)

### Test plan for issue:

- Run this change against the dev subscription to see if it works as intended and deletes what it should.

### DIFF
--- a/hack/clean/clean.go
+++ b/hack/clean/clean.go
@@ -153,24 +153,24 @@ func (s settings) shouldDelete(resourceGroup mgmtfeatures.ResourceGroup, log *lo
 
 	for t := range resourceGroup.Tags {
 		if strings.ToLower(t) == defaultKeepTag {
-			log.Debugf("Group %s is to persist. SKIP.", *resourceGroup.Name)
+			log.Infof("Group %s is to persist. SKIP.", *resourceGroup.Name)
 			return false
 		}
 	}
 
 	// azure tags is not consistent with lower/upper cases.
 	if _, ok := resourceGroup.Tags[s.createdTag]; !ok {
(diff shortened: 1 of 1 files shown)


hack/clean/clean.go
```diff
@@ -31,6 +31,7 @@ var denylist = []string{
 	"images",
 	"secrets",
 	"dns",
+	"shared-cluster",
```
kimorris27, line 34: The shared cluster's managed RG should be safe since it doesn't have a `purge` tag, but you could add it here just for extra safety if you think it's worth it (`aro-dsfx17my`).

kimorris27 (approved):

hack/clean/clean.go
cadenmarchese: I could do that, but I believe that the deny assignment + the lack of purge tag should protect us here. Plus, if we ever had to reinstall the shared cluster, the managed resource group ID would change, and we'd need a new PR.

s-amann (approved):

UlrichSchlueter (approved):

PR #3321, opened 2023-12-11
Documentation - Update Go dependency management doc

s-fairchild:

N/A

Fixes

Update documentation to remove instructions to run make vendor. This script was previously required when the ARO installer was vendored, as well as other OCP 4.x dependencies.

### Test plan for issue:

Updated Go module dependencies recently with these steps.

N/A

### DIFF
--- a/docs/updating-dependencies.md
+++ b/docs/updating-dependencies.md
@@ -1,5 +1,8 @@
 # Updating dependencies
 
+  - Go Documentation for [Managing dependencies](https://go.dev/doc/modules/managing-dependencies)
+  - [Go Module Awareness](https://go.dev/blog/go116-module-changes)
+
 To update golang dependencies one has to choose between two options:
 
 1) update the whole dependency tree
@@ -9,37 +12,23 @@ To update golang dependencies one has to choose between two options:
 ## Update the whole dependency tree
 
 Done when the need to catch upstream dependencies arises, it can be done by
-calling
+calling:
 
 ```bash
-make vendor
-```
-
-in root folder, which calls `hack/update-go-module-dependencies.sh`.
-
-The reason for calling script instead of directly calling:
-
-```bash
+# Ensure go module-aware mode is set to auto
+export GO111MODULE=auto
 go get -u ./...
 go mod tidy -compat=1.18
 go mod vendor
 ```
 
-is that packages modified in this script do not fully support modules and
-semantic versioning via tags. Therefore the proper version is parsed from the version
-branch and fixed using replace directive. Otherwise it will upgrade every time
-the command is started.
-
-When upgrading to a newer version of OpenShift, this script have to be updated to
-reflect the proper release.
-
-
 ## Update only required dependencies
 
 When adding new dependencies or updating old ones, based on the requirement of
-the PR, one can simply call
+the PR, one can simply call:
 
 ```bash
+export GO111MODULE=auto
 go get <module>@<release> OR
 go get -u <module>@<release>
 


docs/updating-dependencies.md
```diff
@@ -9,37 +12,23 @@ To update golang dependencies one has to choose between two options:
 ## Update the whole dependency tree
 
 Done when the need to catch upstream dependencies arises, it can be done by
-calling
-
-```bash
-make vendor
```
cadenmarchese, line 15: Should the makefile target be deprecated / removed altogether? I'm curious on if we still need it for anything if we're removing it from docs.

s-fairchild: I was asking myself the same question. I see Ben updated the script a few weeks ago [PR #3301](https://github.com/Azure/ARO-RP/pull/3301)  

He said it still has some uses, so I don't think we should remove it. Documentation on it's specific purposes and uses would be nice.

kimorris27 (commented):
Don't certain OCP repos still need to be vendored using the script? I'm looking at https://github.com/Azure/ARO-RP/blob/21f0e23ad3245c7b7b62deab26122438b5696886/go.mod#L740 on master and that's why I'm thinking the script is still needed.

But if the premise that we no longer need the script holds, the doc updates LGTM.

s-fairchild:
> Don't certain OCP repos still need to be vendored using the script? I'm looking at
> 
> https://github.com/Azure/ARO-RP/blob/21f0e23ad3245c7b7b62deab26122438b5696886/go.mod#L740
> on master and that's why I'm thinking the script is still needed.
> 
> But if the premise that we no longer need the script holds, the doc updates LGTM.

We flushed out that this documentation is outdated [in this thread](https://redhat-internal.slack.com/archives/C02ULBRS68M/p1701290900456899).  

Maybe I should remove that comment from `go.mod` too?

kimorris27:
> > Don't certain OCP repos still need to be vendored using the script? I'm looking at
> > https://github.com/Azure/ARO-RP/blob/21f0e23ad3245c7b7b62deab26122438b5696886/go.mod#L740
> > 
> > on master and that's why I'm thinking the script is still needed.
> > But if the premise that we no longer need the script holds, the doc updates LGTM.
> 
> We flushed out that this documentation is outdated [in this thread](https://redhat-internal.slack.com/archives/C02ULBRS68M/p1701290900456899).
> 
> Maybe I should remove that comment from `go.mod` too?

Gotcha, I read through the thread. Based on https://github.com/Azure/ARO-RP/pull/3301, it looks like @bennerv didn't remove the whole script, just the part that previously vendored in the custom installer. And reading through the comments at the top of the script (see below), it's not clear to me why we're considering the script deprecated, since the logic explained there seems that it would still apply despite the recent installer-related changes.

So I'm not even sure about the `go.mod` comment tbh. My intention in going back and forth here is just to help guide us to a point where we're fully confident we're not prematurely deprecating something that we still need.

https://github.com/Azure/ARO-RP/blob/21f0e23ad3245c7b7b62deab26122438b5696886/hack/update-go-module-dependencies.sh#L5-L21

docs/updating-dependencies.md
```diff
@@ -9,37 +12,23 @@ To update golang dependencies one has to choose between two options:
...
-```
-
-in root folder, which calls `hack/update-go-module-dependencies.sh`.
-
-The reason for calling script instead of directly calling:
+calling:
 
 ```bash
+# Ensure go module-aware mode is set to auto
+export GO111MODULE=auto
```
SudoBrendan, line 19: My work laptop is fubar so I can't test this myself, but I am 99% sure this breaks re: mac vs linux. Can we get an SRE from each side of the aisle to attempt this change?

docs/updating-dependencies.md
```diff
@@ -9,37 +12,23 @@ To update golang dependencies one has to choose between two options:
...
-
-
 ## Update only required dependencies
 
 When adding new dependencies or updating old ones, based on the requirement of
-the PR, one can simply call
+the PR, one can simply call:
 
 ```bash
+export GO111MODULE=auto
```
SudoBrendan, line 31: Same as above.

SudoBrendan (commented):
I think this introduces some changes that will impact SREs on various OS. Can we get a +1 from both sides?

s-fairchild:
@bennerv Can you review this PR when you have a chance?
There's a [question in this comment](https://github.com/Azure/ARO-RP/pull/3321#issuecomment-1856393652) I think you maybe able to answer.

docs/updating-dependencies.md
SudoBrendan: Woops, ok, so maybe it was a Linux thing - have my laptop back up and I, for reasons unknown along my dev journey on this team, have changed my `GO111MODULE=""`, I can test this now, so I'll give it a go and see what happens :)

s-fairchild:
No longer needed with the removal of vendoring.

PR #3322, opened 2023-12-11
ARO-4773 cluster install failure fix

kimorris27:

https://issues.redhat.com/browse/ARO-4773

I used Kusto to search for this error, and there are quite a few just in the US regions even within the past week. Fortunately, they are all the same, and I think I see how we can fix them.

All of the failures are occurring at the attachNSGs bootstrap step, specifically on this line: https://github.com/Azure/ARO-RP/blob/91fd9d00ac7573a71d78852c85e5d68cbb6fd030/pkg/cluster/deploybaseresources.go#L300.

Given that the code, as currently written, attempts to attach the NSG immediately after it gets created by the previous bootstrap step, I'll bet that what's happening is that there's just a little bit of lag in Azure where the NSG is not ready to be referenced in a subnet attachment yet. My proposed fix is to make attachNSGs a retrying step to allow some time for this lag.

Note that I've coded attachNSGs such that it only retries if it encounters the particular "NSG not found" error we're concerned about here - other errors are still returned directly.

### Test plan for issue:

Unit tests and E2E

No

### DIFF
--- a/pkg/util/steps/condition.go
+++ b/pkg/util/steps/condition.go
@@ -117,7 +117,7 @@ func enrichConditionTimeoutError(f conditionFunction, originalErr error) error {
 	return api.NewCloudError(
 		http.StatusInternalServerError,
 		api.CloudErrorCodeDeploymentFailed,
-		"", message+"Please retry, if issue persists: raise Azure support ticket",
+		"", message+" Please retry, and if the issue persists, raise an Azure support ticket",
 	)
 }
 
--- a/pkg/util/steps/condition.go
+++ b/pkg/util/steps/condition.go
@@ -23,6 +23,7 @@ import (
 // Instead of providing Hive-specific error messages to customers, the below will send a timeout error message.
 // The below functions are run during Install, Update, AdminUpdate.
 var timeoutConditionErrors = map[string]string{
+	"attachNSGs":                             "Failed to attach the ARO NSG to the cluster subnets.",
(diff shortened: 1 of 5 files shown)


cadenmarchese (dismissed):
This looks like a straightforward / safe fix - thanks!

pkg/util/steps/condition.go
```diff
@@ -116,7 +117,7 @@ func enrichConditionTimeoutError(f conditionFunction, originalErr error) error {
 	return api.NewCloudError(
 		http.StatusInternalServerError,
 		api.CloudErrorCodeDeploymentFailed,
-		"", message+"Please retry, if issue persists: raise azure support ticket",
+		"", message+"Please retry, if issue persists: raise Azure support ticket",
```
dem4gus, line 120: nit: while we're capitalizing this, can we get a space between `message` and `Please retry ...`? Or to make it even more grammatically correct, `<message>, please retry. If the issue persists raise an Azure support ticket.`

dem4gus (commented):
Just a small nit

pkg/util/steps/condition.go
kimorris27: Good point, I should done this in the first place. I did a slight variation on what you suggested to avoid removing the period at the end of each error message, so let me know what you think.

s-amann (approved):

cadenmarchese (approved):

PR #3323, opened 2023-12-12
Fixes in disk-encryption-set.md

mbukatov:

None

The RP is addressing:

- difference between des naming scheme used in e2e tests and the disk-encryption-set.md document
- issues with `az keyvault show` reporting just created keyvault as not found

### Test plan for issue:

No

No further edits needed.

### DIFF
--- a/docs/disk-encryption-set.md
+++ b/docs/disk-encryption-set.md
@@ -45,7 +45,7 @@ az keyvault key create --vault-name $KEYVAULT_NAME \
                        -n $KEYVAULT_KEY_NAME \
                        --protection software
 
-KEYVAULT_ID=$(az keyvault show --name $KEYVAULT_NAME --query "[id]" -o tsv)
+KEYVAULT_ID=$(az keyvault show --name $KEYVAULT_NAME -g $RESOURCEGROUP --query "[id]" -o tsv)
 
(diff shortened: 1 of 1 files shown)


docs/disk-encryption-set.md
```diff
@@ -32,7 +32,7 @@ Follow [tutorial-create-cluster](https://docs.microsoft.com/en-us/azure/openshif
 ```bash
 export KEYVAULT_NAME=$USER-enckv
 export KEYVAULT_KEY_NAME=$USER-key
-export DISK_ENCRYPTION_SET_NAME=$USER-des
+export DISK_ENCRYPTION_SET_NAME=$RESOURCEGROUP-disk-encryption-set
```
niontive, line 35: `$RESOURCEGROUP` is not specified above - does this doc assume `$RESOURCEGROUP` is set?

niontive, line 35: Does this doc assume `$RESOURCEGROUP` is already set?

mbukatov: Yes, we assume a reader will check ARO tutorial where this is mentioned (among all the other assumptions):

> Follow [tutorial-create-cluster](https://docs.microsoft.com/en-us/azure/openshift/tutorial-create-cluster) but don't run the az aro create command, instead proceed as follows ...

tiguelu (approved):

PR #3325, opened 2023-12-13
`az aro update` CredentialsRequest hotfix

kimorris27:

Issue where `az aro update --refresh-credentials` fails because `credentialsrequest/openshift-azure-operator` doesn't exist   (for clusters not yet updated to use the new ARO operator Azure auth scheme; see https://github.com/Azure/ARO-RP/pull/3274)

- Adds a step that reconciles the CredentialsRequest upon every `az aro update` regardless of the auth scheme the operator is using or whether the update is a CSP refreshing update
- Fixes the `aroCredentialsRequestReconciled` function to retry until the timeout if the CredentialsRequest is not found (as opposed to erroring out immediately)

### Test plan for issue:

- On a dev cluster, manually tested the code changes by deleting the CredentialsRequest and then running a successful `az aro update --refresh-credentials`
- Added an E2E test that tests whether `az aro update` recreates a missing CredentialsRequest (did not include the CSP refresh in this test)

No

### DIFF
--- a/pkg/cluster/condition.go
+++ b/pkg/cluster/condition.go
@@ -6,6 +6,7 @@ package cluster
 import (
 	"context"
 	"errors"
+	"regexp"
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -13,6 +14,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+var credentialsRequestNotFoundErrorRegex = regexp.MustCompile(`credentialsrequests[.]cloudcredential[.]openshift[.]io.*not found`)
 
 const minimumWorkerNodes = 2
 
@@ -113,6 +116,11 @@ func (m *manager) aroCredentialsRequestReconciled(ctx context.Context) (bool, er
 
 	u, err := m.dynamiccli.Resource(CredentialsRequestGroupVersionResource).Namespace("openshift-cloud-credential-operator").Get(ctx, "openshift-azure-operator", metav1.GetOptions{})
 	if err != nil {
+		// If the CredentialsRequest is not found, it may have just recently been reconciled.
+		// Return nil to retry until we hit the condition timeout.
+		if credentialsRequestNotFoundErrorRegex.MatchString(err.Error()) {
+			return false, nil
+		}
 		return false, err
 	}
 
--- a/pkg/cluster/condition.go
+++ b/pkg/cluster/condition.go
@@ -104,7 +104,7 @@ func isOperatorAvailable(operator *configv1.ClusterOperator) bool {
 // is true if the CredentialsRequest has been reconciled within the past 5 minutes.
 // Checking for a change to the lastSyncCloudCredsSecretResourceVersion attribute of the CredentialRequest's status would be a neater way of checking
 // whether it was reconciled, but we would would have to save the value prior to updating the kube-system/azure-credentials Secret so that we'd have
-// and old value to compare to.
+// an old value to compare to.
 func (m *manager) aroCredentialsRequestReconciled(ctx context.Context) (bool, error) {
 	// If the CSP hasn't been updated, the CredentialsRequest does not need to be reconciled.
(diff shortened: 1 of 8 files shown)


pkg/cluster/condition.go
```diff
@@ -113,6 +116,11 @@ func (m *manager) aroCredentialsRequestReconciled(ctx context.Context) (bool, er
 
 	u, err := m.dynamiccli.Resource(CredentialsRequestGroupVersionResource).Namespace("openshift-cloud-credential-operator").Get(ctx, "openshift-azure-operator", metav1.GetOptions{})
 	if err != nil {
+		// If the CredentialsRequest is not found, it may have just recently been reconciled.
+		// Return nil to retry until we hit the condition timeout.
```
bennerv, line 120: There's a pattern we use `kerrors.IsNotFound(err)` which may be better than using a regex.

pkg/operator/deploy/deploy.go
```diff
@@ -368,6 +369,28 @@ func (o *operator) CreateOrUpdate(ctx context.Context) error {
 	return nil
 }
 
+// CreateOrUpdateCredentialsRequest just creates/updates the ARO operator's CredentialsRequest
+// rather than doing all of the operator's associated Kubernetes resources.
+func (o *operator) CreateOrUpdateCredentialsRequest(ctx context.Context) error {
+	templ, err := template.ParseFS(embeddedFiles, "staticresources/credentialsrequest.yaml")
```
bennerv, line 375: If this path changes, will this break in some way?  

I guess our e2e will catch it since our e2e runs an update.  Just nervous about hardcoded strings in relation to path.

pkg/cluster/condition.go
```diff
@@ -113,6 +116,11 @@ func (m *manager) aroCredentialsRequestReconciled(ctx context.Context) (bool, er
...
+		if credentialsRequestNotFoundErrorRegex.MatchString(err.Error()) {
```
dem4gus, line 121: Use [`errors.IsNotFound()`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsNotFound) here (we import this as `kerrors` elsewhere).

```suggestion
		if kerrors.IsNotFound(err) {
```

test/e2e/update.go
```diff
@@ -17,10 +17,25 @@ import (
...
 	mgmtredhatopenshift20220904 "github.com/Azure/ARO-RP/pkg/client/services/redhatopenshift/mgmt/2022-09-04/redhatopenshift"
 	mgmtredhatopenshift20230701preview "github.com/Azure/ARO-RP/pkg/client/services/redhatopenshift/mgmt/2023-07-01-preview/redhatopenshift"
+	"github.com/Azure/ARO-RP/pkg/cluster"
 	"github.com/Azure/ARO-RP/pkg/util/stringutils"
 )
 
 var _ = Describe("Update clusters", func() {
+	It("must replace the ARO operator's CredentialsRequest if it has been deleted", func(ctx context.Context) {
+		By("deleting the CredentialsRequest")
+		err := clients.DynamicKubernetes.Resource(cluster.CredentialsRequestGroupVersionResource).Namespace("openshift-cloud-credential-operator").Delete(ctx, "openshift-azure-operator", metav1.DeleteOptions{})
```
bennerv, line 27: We shouldn't need the dynamic client here.  We can use the `controller-runtime` client. We don't need to deal with unstructured objects in our e2e, we can just reference them as long as they're recognized by the apiserver schema.

bennerv (commented):
few nits - good first pass though :)

pkg/operator/deploy/deploy.go
kimorris27: Yep, E2E would break if the path were to change.

The only alternative I thought of to hardcoding the path was constructing the object in code (as opposed to reading in YAML from a file) and creating the CredentialsRequest over in `pkg/cluster` with the client-go dynamic client, but then we're duplicating the same CR content in different places, which seems worse than hardcoding a file path.

bennerv (approved):
lgtm

cadenmarchese (approved):

PR #3326, opened 2023-12-14
Server side apply

dem4gus:

Fixes [https://issues.redhat.com/browse/ARO-4537](ARO-4537)

Updates pullsecret controller to use server-side apply to simplify logic flow and avoid race conditions with RP updates. 

### Test plan for issue:
Unit tests updated. Manual dev cluster tests ongoing.

### DIFF
--- a/pkg/operator/controllers/pullsecret/pullsecret_controller.go
+++ b/pkg/operator/controllers/pullsecret/pullsecret_controller.go
@@ -22,7 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	applyv1 "k8s.io/client-go/applyconfigurations/core/v1"
 	"k8s.io/client-go/kubernetes"
-	typedv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -51,7 +51,7 @@ var rhKeys = []string{"registry.redhat.io", "cloud.openshift.com", "registry.con
 type Reconciler struct {
 	base.AROController
 
-	secretsClient typedv1.SecretInterface
+	secretsClient corev1client.SecretInterface
 }
 
 func NewReconciler(log *logrus.Entry, client client.Client, kubernetescli kubernetes.Interface) *Reconciler {
--- a/pkg/operator/controllers/pullsecret/pullsecret_controller.go
+++ b/pkg/operator/controllers/pullsecret/pullsecret_controller.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/Azure/ARO-RP/pkg/operator"
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+	"github.com/Azure/ARO-RP/pkg/operator/controllers/base"
 	"github.com/Azure/ARO-RP/pkg/util/pullsecret"
 )
 
@@ -45,15 +46,16 @@ var rhKeys = []string{"registry.redhat.io", "cloud.openshift.com", "registry.con
 
 // Reconciler reconciles a Cluster object
 type Reconciler struct {
-	log *logrus.Entry
-
-	client client.Client
+	base.AROController
 }
 
 func NewReconciler(log *logrus.Entry, client client.Client) *Reconciler {
 	return &Reconciler{
-		log:    log,
-		client: client,
+		AROController: base.AROController{
+			Log:    log,
+			Client: client,
+			Name:   ControllerName,
+		},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
dem4gus, line 1: Two reasons for most of the changes in this file:

1. The `k8s.io/client-go/kubernetes/fake` package does not interact with the `ResourceVersion` field at all.
1. The `Apply()` function adds `TypeMeta` fields.

One other change is to change which mock store the final object is being retrieved from to compare to the expected changes. Other than that the test logic and expected outcomes have been left intact.

pkg/operator/controllers/pullsecret/pullsecret_controller.go
```diff
@@ -145,58 +153,42 @@ func (r *Reconciler) ensureGlobalPullSecret(ctx context.Context, operatorSecret,
...
-			Type: corev1.SecretTypeDockerConfigJson,
-			Data: make(map[string][]byte),
+	// recreate := false
+	secretData := make(map[string][]byte)
+
+	var invalidSecretType bool
+	if userSecret != nil {
+		// when userSecret has broken type, recreates it with proper type
+		// unfortunately the type field is immutable, therefore the whole secret has to be deleted and recreated
+		if userSecret.Type != corev1.SecretTypeDockerConfigJson {
```
dem4gus, line 163: I managed to test this particular functionality by scaling the ARO operator to 0, deleting `openshift-config/pull-secret` and recreating it with only `type: Opaque` changed, then re-scaling the operator to 1 and patching `openshift-azure-operator/cluster` secret with `{}`. I don't know how else the pull-secret would become Opaque, but this functionality is still intact.

pkg/operator/controllers/pullsecret/pullsecret_controller.go
```diff
@@ -145,58 +153,42 @@ func (r *Reconciler) ensureGlobalPullSecret(ctx context.Context, operatorSecret,
...
+			invalidSecretType = true
+			err := r.secretsClient.Delete(ctx, pullSecretName.Name, metav1.DeleteOptions{})
```
cadenmarchese, line 165: Similar to above - do we need to delete an invalid secret if our plan is to apply unconditionally at the end? I believe any invalid fields would be overridden / replaced.

pkg/operator/controllers/pullsecret/pullsecret_controller.go
```diff
@@ -145,58 +153,42 @@ func (r *Reconciler) ensureGlobalPullSecret(ctx context.Context, operatorSecret,
...
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      pullSecretName.Name,
-				Namespace: pullSecretName.Namespace,
-			},
-			Type: corev1.SecretTypeDockerConfigJson,
-			Data: make(map[string][]byte),
+	// recreate := false
+	secretData := make(map[string][]byte)
+
+	var invalidSecretType bool
```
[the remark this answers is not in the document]
cadenmarchese: I think the idea I have would just be:
1) Define a secret like so: https://github.com/Azure/ARO-RP/pull/3326/files#diff-daed81f3f056e85b1265f2dc476197ca52b05529e2c6a5e186ca62370f0f727aL158-L164
2) Apply that exactly: https://github.com/Azure/ARO-RP/pull/3326/files#diff-daed81f3f056e85b1265f2dc476197ca52b05529e2c6a5e186ca62370f0f727aR187-R191

If we know exactly what secret fields should/will be, we don't need to extract any of the current object fields or assert validity/non validity. Or am I wrong?

cadenmarchese (commented):
Thanks for your efforts to move this controller to apply! I am still familiarizing myself with the current state of the pull secret controller so I'll follow up with another review, but I have some first-pass questions about the old logic we are partially carrying over.

I think that an e2e should be in-scope for this PR since it's one of the most (if not the most) critical controller. We could have something like `pull secret should be recreated correctly if deleted` and `pull secret should be recreated correctly if modified`.

pkg/operator/controllers/pullsecret/pullsecret_controller.go
```diff
@@ -68,19 +77,18 @@ func NewReconciler(log *logrus.Entry, client client.Client) *Reconciler {
...
-		r.log.Debug("controller is disabled")
+		r.AROController.Log.Debug("controller is disabled")
 		return reconcile.Result{}, nil
 	}
 
-	r.log.Debug("running")
-	userSecret := &corev1.Secret{}
-	err = r.client.Get(ctx, pullSecretName, userSecret)
+	r.AROController.Log.Debug("running")
+	userSecret, err := r.secretsClient.Get(ctx, pullSecretName.Name, metav1.GetOptions{})
```
jaitaiwan, line 91: It's confusing to me why we need `secretsClient` and why we aren't just using `r.AROController.Client`

pkg/operator/controllers/pullsecret/pullsecret_controller.go
```diff
@@ -89,7 +97,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	// detects if the global pull secret is broken and fixes it by using backup managed by ARO operator
 	if instance.Spec.OperatorFlags.GetSimpleBoolean(controllerManaged) {
 		operatorSecret := &corev1.Secret{}
-		err = r.client.Get(ctx, types.NamespacedName{Namespace: operator.Namespace, Name: operator.SecretName}, operatorSecret)
+		err = r.AROController.Client.Get(ctx, types.NamespacedName{Namespace: operator.Namespace, Name: operator.SecretName}, operatorSecret)
```
jaitaiwan, line 100: Previously we used `r.secretsClient` to get the secret, how come we're not doing that here?

pkg/operator/controllers/pullsecret/pullsecret_controller.go
jaitaiwan: I don't think unconditional apply will work if the user secret being of the wrong type as per the code comments.

pkg/operator/controllers/pullsecret/pullsecret_controller.go
```diff
@@ -145,58 +153,42 @@ func (r *Reconciler) ensureGlobalPullSecret(ctx context.Context, operatorSecret,
...
-
-	if recreate {
-		secret = &corev1.Secret{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      pullSecretName.Name,
-				Namespace: pullSecretName.Namespace,
-			},
-			Type: corev1.SecretTypeDockerConfigJson,
-			Data: make(map[string][]byte),
+	// recreate := false
```
jaitaiwan, line 156: I'm not sure why this commented code is here

pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
```diff
@@ -761,13 +780,20 @@ func TestEnsureGlobalPullSecret(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			ctx := context.Background()
 
+			var kubernetesFake *fake.Clientset
 			clientBuilder := ctrlfake.NewClientBuilder()
 			if tt.initialSecret != nil {
 				clientBuilder = clientBuilder.WithObjects(tt.initialSecret)
+				kubernetesFake = serversideapply.CliWithApply([]string{"secrets"}, tt.initialSecret)
+			} else {
+				kubernetesFake = serversideapply.CliWithApply([]string{"secrets"})
```
jaitaiwan, line 789: Can this be simplified? I'm wondering if an empty `tt.initialSecret` is treated the same as `tt.initialSecret` not being supplied to `CliWithApply` that way we don't need the else branch

jaitaiwan (changes requested):
I think this is a net good change. I have a couple queries and one question about branching code. I also agree that an e2e would be good here.

PR #3327, opened 2023-12-15
Fixing magic strings on operator flags

azoppiserpa:

There isn't a work item to be referenced for this change. The issue was found searching for `// TODO` in the code. It's a small refactor.

- Fixes magic strings related to operator flags, and places them in a new `flags.go` file that can be reused instead of us having a bunch of magic strings that could be wrongly changed easily everywhere.

### Test plan for issue:

- Unit tests: done, passed.
- Integration tests: manually tested on INT through the following steps:

1. Deployed this PR on INT pipeline.
2. Created a test cluster.
3. Checked that operator flags on the test cluster were set as expected, comparing them to the flags on the PR.
4. Also checked that operator flags matched the expected on CosmosDB.
5. On both cases the flags were set as expected. 

✅ 

No documentation changes needed.


niontive:
Let's revert all changes to the Python code. That code already has the licenses stuff.

test/e2e/cluster.go
```diff
@@ -144,11 +145,11 @@ var _ = Describe("Cluster", Serial, func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			// Poke the ARO storageaccount controller to reconcile
-			cluster.Spec.OperatorFlags["aro.storageaccounts.enabled"] = "false"
+			cluster.Spec.OperatorFlags[operator.StorageAccountsEnabled] = "false"
```
niontive, line 148: Add your new `operator.FlagTrue` and `operator.FlagFalse` here.

test/e2e/cluster.go
```diff
@@ -144,11 +145,11 @@ var _ = Describe("Cluster", Serial, func() {
...
 			cluster, err = clients.AROClusters.AroV1alpha1().Clusters().Update(ctx, cluster, metav1.UpdateOptions{})
 			Expect(err).NotTo(HaveOccurred())
 
-			cluster.Spec.OperatorFlags["aro.storageaccounts.enabled"] = "true"
+			cluster.Spec.OperatorFlags[operator.StorageAccountsEnabled] = "true"
```
niontive, line 152: `operator.FlagFalse`

test/e2e/operator.go
```diff
@@ -375,7 +376,7 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
 		By("checking if preconfiguredNSG is enabled")
 		co, err := clients.AROClusters.AroV1alpha1().Clusters().Get(ctx, "cluster", metav1.GetOptions{})
 		Expect(err).NotTo(HaveOccurred())
-		if co.Spec.OperatorFlags["aro.azuresubnets.nsg.managed"] == "false" {
+		if co.Spec.OperatorFlags[operator.AzureSubnetsNsgManaged] == "false" {
```
niontive, line 379: Add `operator.FlagFalse` here

pkg/util/mocks/env/core.go
```diff
@@ -8,12 +8,13 @@ import (
 	context "context"
 	reflect "reflect"
 
-	env "github.com/Azure/ARO-RP/pkg/env"
-	azureclient "github.com/Azure/ARO-RP/pkg/util/azureclient"
-	liveconfig "github.com/Azure/ARO-RP/pkg/util/liveconfig"
 	azcore "github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	autorest "github.com/Azure/go-autorest/autorest"
 	gomock "github.com/golang/mock/gomock"
+
```
niontive, line 14: No need to make this change - we're not touching this code in this PR.

niontive:
We can get rid of `controllerEnabled` consts now since we have a central place where these are defined.

pkg/util/mocks/env/core.go
azoppiserpa: fixed

azoppiserpa:
A summary of what the commits I added did:

- Reverted unintended changes to the Python code.
- Replaced the strings `"true"` and `"false"` with `operator.FlagTrue` and `operator.FlagFalse` as discussed with Nic.
- Removed the `controllerEnabled` consts and replaced their usage with the adequate flag from the newly created `flags.go` file.
- Ran `make validate-go` because the automatic check pointed out a failure. With that, it became clear the failure was due to a missing copyright comment on `flags.go`, which was added on the last commit.

azoppiserpa:
Removed the `controllerEnabled` constant from the `pkg/operator/controllers/checkers/common/const.go`, and then the file itself since it became an empty file after removing the constant.

(I left this removal for later because the constant had more usages so it seemed more complex than the previous ones)

pkg/operator/controllers/alertwebhook/alertwebhook_controller_test.go
```diff
@@ -180,13 +181,13 @@ func TestSetAlertManagerWebhook(t *testing.T) {
 				},
 				Spec: arov1alpha1.ClusterSpec{
 					OperatorFlags: arov1alpha1.OperatorFlags{
-						controllerEnabled: "false",
+						operator.AlertWebhookEnabled: operator.FlagFalse,
 					},
 				},
 			}
 
 			if tt.controllerEnabled {
```
gouthamMN, line 189: Should variable "controllerEnabled" have more meaningful name now like alertWebhookEnabled?

pkg/operator/controllers/guardrails/guardrails_config.go
```diff
@@ -18,9 +18,7 @@ import (
 
 const (
```
gouthamMN, line 19: Can we define all const vars with value starting with "aro.guardrails.*" under operator package and then refer them from it?

pkg/operator/flags.go
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the Apache License 2.0.
+
+const (
+	AlertWebhookEnabled                = "aro.alertwebhook.enabled"
+	AzureSubnetsEnabled                = "aro.azuresubnets.enabled"
+	AzureSubnetsNsgManaged             = "aro.azuresubnets.nsg.managed"
+	AzureSubnetsServiceEndpointManaged = "aro.azuresubnets.serviceendpoint.managed"
+	BannerEnabled                      = "aro.banner.enabled"
+	CheckerEnabled                     = "aro.checker.enabled"
+	DnsMasqEnabled                     = "aro.dnsmasq.enabled"
```
dem4gus, line 13: nit: Dnsmasq isn't camelCase. It's technically `dnsmasq` but Go wants exported things to start with a capital letter.

dem4gus (commented):
Thank you for this change, it was very needed.

dem4gus (dismissed):

pkg/operator/flags.go
azoppiserpa: I've changed it both for `DnsmasqEnabled` and `RestardDnsmasqEnabled`, thanks for pointing out :)

pkg/operator/controllers/guardrails/guardrails_config.go
azoppiserpa: Hi, asking just to make sure I understand what you're suggesting :) Do you mean doing something like this for the variables that start with `"aro.guardrails"` in this file?

```
package main

import "fmt"

const (
	prefix = "foo"
	suffix = prefix + ".bar"
)

func main() {
	fmt.Println(suffix)
}
```

Or do you mean something else?

pkg/operator/controllers/alertwebhook/alertwebhook_controller_test.go
azoppiserpa: In case you're referring to the variable of line 183, we're removing it, so it won't be present anymore, since we can use the `operator.AlertWebhookEnabled` flag directly. So we wouldn't have to rename it, just remove it.
 
If you're referring to other occurrences of `controllerEnabled` local variables that haven't been removed (I looked it up and there are some local variables with this name on `alertwebhook_controller_test.go` and `ingress_controller_test.go`): yes, I agree it could be changed for a more meaningful name. Given the existing scope of this PR, we can fix this in a future refactor.

hawkowl (dismissed):
I'm working on making pkg/api/ independent of the rest of the codebase so that it can be imported in the Installer Wrapper, MIMO, etc without dragging in potentially all the RP's dependencies. Is there a way we can do this without pkg/api/ having a hard dependency on pkg/operator/?

pkg/operator/controllers/guardrails/guardrails_config.go
gouthamMN: Not really, for example do you see how _**aro.guardrails.enabled**_ is defined in operator package and then referred here as _**operator.GuardrailsEnabled**_? Something like that if that's possible.
Another example is _**aro.guardrails.deploy.manged**_ which is _**operator.GuardrailsDeployManaged**_.

gouthamMN: My bad, this is already in the operator package, sorry for the confusion :)

dofinn (approved):
lgtm nice clean up

PR #3329, opened 2023-12-18
Use networkmanager-provisioned dnsmasq for 4.13+

hawkowl:
proof of concept


ventifus:
According to some notes from the 3.x days, `dns=dnsmasq` results in a resolv.conf of `nameserver 127.0.0.1`, which won't work when copied to pods' resolv.conf; we need the node IP there instead.

https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_node/files/networkmanager/99-origin-dns.sh

ventifus:
Here's a crazier idea: configure dns=resolvconf and ship a resolveconf program that implements our logic of updating /etc/resolv.conf.dnsmasq instead (or openresolv if it can do what we want).

PR #3332, opened 2023-12-19
Update shared cluster usage

mociarain:

When replacing the shared-cluster I noticed that we have some scripts and docs related to the shared-cluster that are stale now.

This PR just tidies up the secrets around the shared-cluster.

### Test plan for issue:
I could recreate the shared-cluster to test it but I think it's probably okay.

Note:
I don't have access to a B-account so there might be some style guides etc that I'm missing so get pedantic and let me know if there's anything like that ;)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,7 +247,7 @@ shared-cluster-login:
 	@oc login ${SHARED_CLUSTER_API} -u kubeadmin -p ${SHARED_CLUSTER_KUBEADMIN_PASSWORD}
 
 shared-cluster-create:
-	./hack/shared-cluster.sh crerate
+	./hack/shared-cluster.sh create
 
 shared-cluster-delete:
 	./hack/shared-cluster.sh delete
--- a/Makefile
+++ b/Makefile
@@ -246,6 +246,12 @@ test-python: pyenv az
 shared-cluster-login:
 	@oc login ${SHARED_CLUSTER_API} -u kubeadmin -p ${SHARED_CLUSTER_KUBEADMIN_PASSWORD}
 
+shared-cluster-create:
+	./hack/shared-cluster.sh crerate
+
+shared-cluster-delete:
+	./hack/shared-cluster.sh delete
+
 unit-test-python:
 	hack/unit-test-python.sh
 
--- a/docs/shared-cluster.md
+++ b/docs/shared-cluster.md
@@ -37,32 +37,24 @@ export KUBECONFIG=$PWD/secrets/shared-cluster.kubeconfig
 
 ## Creating / Deleting the Shared Cluster
 
-The shared cluster has the following attributes:
-
-* Tenant: AME
-
-* Subscription: ARO CI/E2E
-
-* Region: Westcentralus
-
-* Name: shared-cluster
-
-* Resource group name: shared-cluster
-
-* Cluster resource group name: aro-shared-cluster
+The shared cluster is for general use and experimentation. Note that there is no guarantee that it is in a "good state".
 
 
 ### Create / Delete
 To create/ delete/ administer the cluster from az cli you must have proper permissions (JIT in the case of AME).
 
+The following commands assume the same secrets as above.
+
 * Create:
 
 ```bash
-./hack/shared-cluster.sh create
+
+make shared-cluster-create
+
 ```
 
 * Delete:
 
 ```bash
-./hack/shared-cluster.sh delete
+make shared-cluster-delete
 ```
--- a/docs/shared-cluster.md
+++ b/docs/shared-cluster.md
@@ -37,7 +37,7 @@ export KUBECONFIG=$PWD/secrets/shared-cluster.kubeconfig
 
 ## Creating / Deleting the Shared Cluster
 
-The shared cluster is for general use and experimentation.
+The shared cluster is for general use and experimentation. All the related details for it can be found in the secrets (look for the prefix `SHARED_CLUSTER_`).
 Note that there is no guarantee that it is in a "good state".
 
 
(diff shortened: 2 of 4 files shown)


cadenmarchese (commented):
If we're no longer hardcoding the shared cluster name, we might also want to work with https://github.com/Azure/ARO-RP/blob/master/hack/clean/clean.go#L34 to remove `shared-cluster` from that, or parameterize it.

It's technically not required to have a production cluster on the deny list, since our clean script + pipeline only deletes dev clusters (clusters with `purge: true` tag). So maybe we should just remove it?

dem4gus (commented):
Just a small consideration/suggestion, but looks good otherwise.

mociarain:
> If we're no longer hardcoding the shared cluster name, we might also want to work with https://github.com/Azure/ARO-RP/blob/master/hack/clean/clean.go#L34 to remove `shared-cluster` from that, or parameterize it.
> 
> It's technically not required to have a production cluster on the deny list, since our clean script + pipeline only deletes dev clusters (clusters with `purge: true` tag). So maybe we should just remove it?

Good call! I've gone though that and I've got 2 actions:
- For this PR I want to add the `persist:true` tag to the create script
- More broadly I'll open another discussion/PR with my thoughts on it. I think the logic is a little broken and over-complicated plus I need to practice Go :D

mociarain:
> @mociarain the command you issued was incorrect. Please try again.
> 
> Examples are:
> 
> ```
> @microsoft-github-policy-service agree
> ```
> 
> and
> 
> ```
> @microsoft-github-policy-service agree company="your company"
> ```

@microsoft-github-policy-service agree

tiguelu (dismissed):
/lgtm

UlrichSchlueter (approved):

hack/shared-cluster.sh
```diff
@@ -1,63 +1,86 @@
...
 if [ "$#" -ne 1 ]; then
 	usage
 
 elif [[ $1 == create ]]; then
     echo "creating resource group and network"
-    az group create --name $RESOURCEGROUP --location $LOCATION
+    az group create \
+    --name $SHARED_CLUSTER_RESOURCE_GROUP_NAME \
+    --location $SHARED_CLUSTER_LOCATION \
+    --tags persist:true  # This tag stops the RG being cleaned up
```
cadenmarchese, line 32: Nit - we don't need this tag in the shared cluster script since the purge pipeline (clean.go) will only clean up resource groups with `purge:true`, which is automatically applied to any cluster created by a dev RP, but not applied to any cluster created by a prod RP. Since the shared cluster is a prod cluster without `purge`, it doesn't need to be tagged with `persist`, the clean.go script will ignore it either way.

If the intent is to change that behavior and clean up both prod and dev clusters, maybe we can handle that + the shared cluster tagging in a separate PR?

Overall, I think that the `purge` vs `persist` tagging conventions we use are confusing and we should probably just use one or the other. This could be a future PR / enhancement / doc.

hack/shared-cluster.sh
```diff
@@ -1,63 +1,86 @@
...
       --master-subnet master-subnet \
       --worker-subnet worker-subnet
 
+    CLUSTER_RESOURCE_GROUP_ID=az aro show \
+      --name $SHARED_CLUSTER_CLUSTER_NAME \
+      --resource-group $SHARED_CLUSTER_RESOURCE_GROUP_NAME \
+      | jq .clusterProfile.resourceGroupId
+
+    # This tag stops the RG being cleaned up
+    az tag create \
```
cadenmarchese, line 75: Same comment as above.

cadenmarchese (commented):
Nice work! One nitpick / suggestion for a future change related to our conversation about the purge pipeline.

hack/shared-cluster.sh
mociarain: > Overall, I think that the purge vs persist tagging conventions we use are confusing and we should probably just use one or the other. This could be a future PR / enhancement / doc.

Couldn't agree more. I added it here for what seemed like consistency to me i.e. by supporting both and not choosing either in this PR we can decide to drop to just one in the future "easier". Happy to drop the tag here if you feel strongly but IMO (fresh and mostly context free as it is ;) ) we should look at all this together and decide how much we actually care. 

I actually wrote up some notes on this whole topic [here](https://docs.google.com/document/d/1UKHAoSRt2qLU53at9kdYg_3B3BDQW39FCoqYbZ3yJ0o/edit) that you might find interesting.

Makefile
```diff
@@ -246,6 +246,12 @@ test-python: pyenv az
 shared-cluster-login:
 	@oc login ${SHARED_CLUSTER_API} -u kubeadmin -p ${SHARED_CLUSTER_KUBEADMIN_PASSWORD}
 
+shared-cluster-create:
+	./hack/shared-cluster.sh create
+
+shared-cluster-delete:
+	./hack/shared-cluster.sh delete
+
```
jaitaiwan, line 254: Very helpful!

docs/shared-cluster.md
```diff
@@ -37,32 +37,25 @@ export KUBECONFIG=$PWD/secrets/shared-cluster.kubeconfig
...
-
-* Subscription: ARO CI/E2E
-
-* Region: Westcentralus
-
-* Name: shared-cluster
-
-* Resource group name: shared-cluster
-
-* Cluster resource group name: aro-shared-cluster
```
jaitaiwan, line 52: It might be good to replace these details with generalized details about where we deploy the shared cluster to.

jaitaiwan (commented):
I agree with splitting up the PR if there's an intention to change how the shared cluster pruning works. Otherwise LGTM

SudoBrendan (approved):

petrkotas (commented):
LGTM, good improvement.

UlrichSchlueter (approved):
lgtm

PR #3334, opened 2023-12-20
Add E2E test for host servicing

kimorris27:

https://issues.redhat.com/browse/ARO-4824

The test simulates host servicing on all three worker nodes based on [instructions from Azure docs on how to do so](https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-how-it-works#disabling-or-enabling-accelerated-networking-in-a-nonrunning-vm). It then waits for the worker nodes to become Ready again and reads each worker node's `resolv.conf` to confirm that the contents are still correct after the "host servicing" has taken place.

A concern that the team should provide feedback on: I did my best to do only what's necessary, but the test takes > 10 minutes just because of the nature of what we have to do. Is this unreasonably long considering that we will likely only continue to add to the length of E2E in the future?

### Test plan for issue:

This work is itself a test 😋

N/A

### DIFF
--- a/test/e2e/dns.go
+++ b/test/e2e/dns.go
@@ -399,7 +399,7 @@ func workerNodesReady(ctx context.Context, cli kubernetes.Interface) error {
 	}
 
 	for _, node := range nodeList.Items {
-		if isWorkerNode(node) && !ready.NodeIsReady(node) {
+		if isWorkerNode(node) && !ready.NodeIsReady(&node) {
 			return fmt.Errorf("a worker node is not yet ready")
 		}
 	}
--- a/test/e2e/dns.go
+++ b/test/e2e/dns.go
@@ -34,6 +34,7 @@ var (
 )
 
 const (
+	maxObjNameLen           = 63
 	resolvConfContainerName = "read-resolv-conf"
 )
 
@@ -250,7 +251,11 @@ func nicName(nodeName string) string {
 }
 
 func resolvConfJobName(nodeName string) string {
-	return fmt.Sprintf("read-resolv-conf-%s", nodeName)
+	jobName := fmt.Sprintf("read-resolv-conf-%s", nodeName)
+	if len(jobName) > maxObjNameLen {
+		jobName = jobName[:maxObjNameLen]
+	}
+	return jobName
 }
 
 func createResolvConfJob(ctx context.Context, cli kubernetes.Interface, nodeName string, namespace string) error {
--- a/test/e2e/dns.go
+++ b/test/e2e/dns.go
@@ -1,0 +1,406 @@
+package e2e
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"regexp"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/Azure/go-autorest/autorest/to"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/network"
+	"github.com/Azure/ARO-RP/pkg/util/stringutils"
+	"github.com/Azure/ARO-RP/test/util/project"
+)
+
+var (
+	nameserverRegex              = regexp.MustCompile("nameserver [0-9.]*")
+	verifyResolvConfTimeout      = 30 * time.Second
+	verifyResolvConfPollInterval = 1 * time.Second
+	nodesReadyPollInterval       = 2 * time.Second
+	nicUpdateWaitTime            = 5 * time.Second
+)
+
+const (
+	resolvConfContainerName = "read-resolv-conf"
+)
+
+var _ = Describe("ARO cluster DNS", func() {
+	It("must not be adversely affected by Azure host servicing", func(ctx context.Context) {
+		By("creating a test namespace")
+		testNamespace := fmt.Sprintf("test-e2e-%d", GinkgoParallelProcess())
+		p := project.NewProject(clients.Kubernetes, clients.Project, testNamespace)
+		err := p.Create(ctx)
+		Expect(err).NotTo(HaveOccurred(), "Failed to create test namespace")
+
+		By("verifying the namespace is ready")
+		Eventually(func(ctx context.Context) error {
+			return p.Verify(ctx)
+		}).WithContext(ctx).WithTimeout(DefaultEventuallyTimeout).Should(BeNil())
+
+		DeferCleanup(func(ctx context.Context) {
+			By("deleting the test namespace")
+			err := p.Delete(ctx)
+			Expect(err).NotTo(HaveOccurred(), "Failed to delete test namespace")
+
+			By("verifying the namespace is deleted")
+			Eventually(func(ctx context.Context) error {
+				return p.VerifyProjectIsDeleted(ctx)
+			}).WithContext(ctx).WithTimeout(DefaultEventuallyTimeout).Should(BeNil())
+		})
+
+		By("listing all cluster nodes to retrieve the names of the worker nodes")
+		workerNodes := map[string]string{}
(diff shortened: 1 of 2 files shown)


kimorris27:
E2E failure looks like a flake. I captured the flake in a Jira: https://issues.redhat.com/browse/ARO-4956

test/e2e/dns.go
```diff
@@ -0,0 +1,409 @@
...
+
+const (
+	maxObjNameLen           = 63
+	resolvConfContainerName = "read-resolv-conf"
+)
+
+var _ = Describe("ARO cluster DNS", func() {
+	It("must not be adversely affected by Azure host servicing", func(ctx context.Context) {
+		By("creating a test namespace")
+		testNamespace := fmt.Sprintf("test-e2e-%d", GinkgoParallelProcess())
```
cadenmarchese, line 44: Not required for this PR, but I think it would be an improvement to use this convention everywhere we create a test namespace in e2e. I see there are a few more examples of us creating test namespaces in the e2e suite.

test/e2e/dns.go
```diff
@@ -0,0 +1,409 @@
...
+		})
+
+		By("listing all cluster nodes to retrieve the names of the worker nodes")
+		workerNodes := map[string]string{}
+		nodeList, err := clients.Kubernetes.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		for _, node := range nodeList.Items {
+			name := node.ObjectMeta.Name
+			if strings.Contains(name, "worker") {
```
cadenmarchese, line 72: We should try to use the node role / label here, if possible, to avoid possible flakes from the node name.

test/e2e/dns.go
```diff
@@ -0,0 +1,409 @@
...
+				workerNodes[name] = ""
+			}
+		}
+		Expect(workerNodes).To(HaveLen(3))
```
cadenmarchese, line 76: Is this required for the rest of the test?

test/e2e/dns.go
```diff
@@ -0,0 +1,409 @@
...
+
+				return nil
+			}).WithContext(ctx).
+				WithTimeout(10 * time.Second).
+				WithPolling(1 * time.Second).
+				Should(Succeed())
+		})
+
+		By("verifying each worker node's resolv.conf via a one-shot Job per node")
+		for wn, ip := range workerNodes {
```
cadenmarchese, line 156: I'm wondering if we need to check all three, or if we can save some time by only simulating and checking one node.

test/e2e/dns.go
```diff
@@ -0,0 +1,409 @@
...
+
+func toggleAcceleratedNetworking(ctx context.Context, interfaces network.InterfacesClient, clusterResourceGroup string, nodeName string, enabled bool) error {
+	nic, err := interfaces.Get(ctx, clusterResourceGroup, nicName(nodeName), "")
+	if err != nil {
+		return err
+	}
+
+	nic.EnableAcceleratedNetworking = to.BoolPtr(enabled)
+	err = clients.Interfaces.CreateOrUpdateAndWait(ctx, clusterResourceGroup, nicName(nodeName), nic)
+	return err
```
cadenmarchese, line 384: Instead of doing the `sleep` after each toggle in the tests, can we just sleep here in the function itself, and/or leverage `Eventually` or something else in ginkgo to handle the wait?

test/e2e/dns.go
```diff
@@ -0,0 +1,409 @@
...
+}
+
+func nodesReady(ctx context.Context, cli kubernetes.Interface) error {
+	nodeList, err := cli.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, node := range nodeList.Items {
```
cadenmarchese, line 393: Do we need to check readiness on all nodes or just the ones we're testing (worker(s))?

test/e2e/dns.go
```diff
@@ -0,0 +1,409 @@
...
+	if err != nil {
+		return err
+	}
+
+	nic.EnableAcceleratedNetworking = to.BoolPtr(enabled)
+	err = clients.Interfaces.CreateOrUpdateAndWait(ctx, clusterResourceGroup, nicName(nodeName), nic)
+	return err
+}
+
+func nodesReady(ctx context.Context, cli kubernetes.Interface) error {
```
cadenmarchese, line 387: We leverage `github.com/Azure/ARO-RP/pkg/util/ready` in the VM redeploy test - can we leverage it in this test as well?

cadenmarchese (commented):
Great work so far! There might be an opportunity or two to save some time during the test run, such as running the test on a single node only.

In the short term, I think that adding a little time to e2e to check for a regression is definitely worth it, but it seems we'd only really need to run this when we update the default OCP version used in `az aro create`. In the long term, we might want to look at an improvement to toggle e2e tests in RP-Config so that we can better tune e2e to what's being released, and skip tests such as this if there's no version change. This would definitely need to be its own PR/proposal :)

test/e2e/dns.go
kimorris27: I thought about this too. I could go either way. I wrote the test to check all three workers because it more accurately simulates an actual occurrence of host servicing in prod, and I don't think it will take _that_ much longer than just doing one worker.

The long bottleneck is the time spent waiting for the nodes to become Ready again after the restart, but it's mostly happening "in parallel" - the actions of starting and stopping don't take long per node.

I would leave it as-is, but I'm open to changing it to just do one worker if you think saving time is a priority.

test/e2e/dns.go
kimorris27: Oh, I took this from `cluster.go`, which I thought was the only other place where we create a test namespace. Where else did you see us creating a test namespace?

test/e2e/dns.go
cadenmarchese: If we are sleeping for the purposes of waiting for a nic update, and we are updating all of the nics, I think it makes sense to sleep for each one. Do we have an idea of how much time it takes to let the change settle on a per-nic basis, or are we just making a ballpark estimate to avoid flakes?

test/e2e/dns.go
cadenmarchese: This is what I found searching the code base during my last review: https://github.com/Azure/ARO-RP/blob/master/test/e2e/adminapi_cluster_getlogs.go#L35

But looking again, it doesn't look like we ever create that namespace - just confirm that we can't get logs from a pod there if we make an attempt. I think it's fine as-is!

test/e2e/dns.go
```diff
@@ -0,0 +1,408 @@
...
+}
+
+func workerNodesReady(ctx context.Context, cli kubernetes.Interface) error {
+	nodeList, err := cli.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, node := range nodeList.Items {
+		if isWorkerNode(node) && !ready.NodeIsReady(node) {
```
cadenmarchese, line 402:
```suggestion
		if isWorkerNode(*node) && !ready.NodeIsReady(*node) {
```
Pointers needed?

test/e2e/dns.go
kimorris27: The time I chose is a ballpark estimate to avoid flakes, unfortunately. I wanted to make the test resilient to flakes, I'm not sure how we'd determine exactly how long it takes for the change to truly take effect. I'm thinking that even if we `GET` the NIC and the field is updated, there might still be some stuff on the Network RP backend that hasn't happened yet.

What do we gain by sleeping for each one though? Assuming that it takes 5 seconds for a NIC update to settle, I can just iteratively update all three and then sleep for 5 seconds. If I sleep for 5 seconds after each update, I'm wasting 10 seconds in the middle.

test/e2e/dns.go
kimorris27: I made the `node` param to `isWorkerNode` a struct/value rather than a pointer since we don't need to modify the node that's passed in. I think you would only want to use a pointer in this scenario for performance, and only if the node struct was _really_ big.

kimorris27: You must be referring to the `ci-go` error here. I fixed it by using the `&` operator when we pass the node into `ready.NodeIsReady`, which needs a pointer.

I left `isWorkerNode` how it is since it doesn't need to modify the node that's passed in.

test/e2e/dns.go
cadenmarchese: I think it's fine since we can only ballpark. It's not super ideal to have a sleep in a test but we might not have enough visibility on how long this will take in each region we plan on running this test

cadenmarchese (approved):
lgtm! Third review's the charm. I'm personally good to merge this on green e2e.

PR #3335, opened 2023-12-20
Added FP Check for validation

carvalhe:

Fixes https://issues.redhat.com/browse/ARO-4734

Users are getting "Internal Server Error" when they have a deleted ServicePrincipal, and we want to give them a better and more detailed response. We noticed that the SP validation checks happen that would catch this and return the right error, but FP doesn't so we are doing the same validation step for FP as well.

### Test plan for issue:
Looking into ways to test this as it would require removing a SP on this branch in an INT env.

### DIFF
--- a/pkg/backend/openshiftcluster.go
+++ b/pkg/backend/openshiftcluster.go
@@ -340,9 +340,9 @@ func (ocb *openShiftClusterBackend) asyncOperationResultLog(log *logrus.Entry, i
 		return
 	}
 
-	if strings.Contains(backendErr.Error(), "one of the claims 'puid' or 'altsecid' or 'oid' should be present") {
+	if strings.Contains(strings.ToLower(backendErr.Error()), "one of the claims 'puid' or 'altsecid' or 'oid' should be present") {
 		backendErr = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalClaims,
-			"properties.servicePrincipleProfile", "The provided service principal does not give an access token with at least one of the claims 'altsecid', 'oid' or 'puid'. Please make sure service principal is properly created in the tenant.")
+			"properties.servicePrincipleProfile", "The Azure Red Hat Openshift resource provider service principal has been removed from your tenant. To restore, please unregister and then re-register the Azure Red Hat OpenShift resource provider.")
 	}
 
 	_, ok := backendErr.(*api.CloudError)
--- a/pkg/backend/openshiftcluster.go
+++ b/pkg/backend/openshiftcluster.go
@@ -342,7 +342,7 @@ func (ocb *openShiftClusterBackend) asyncOperationResultLog(log *logrus.Entry, i
 
 	if strings.Contains(strings.ToLower(backendErr.Error()), "one of the claims 'puid' or 'altsecid' or 'oid' should be present") {
 		backendErr = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalClaims,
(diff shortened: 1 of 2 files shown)


nwnt:
Looks good to me, but we will have to test this to validate.  As discussed, let's use the provided test tenant to simulate and see whether we get a correct error message.

nwnt:
On a second thought, I think we should adjust the error message for invalid token that the service principal couldn't be found too.

pkg/backend/openshiftcluster.go
```diff
@@ -339,6 +340,11 @@ func (ocb *openShiftClusterBackend) asyncOperationResultLog(log *logrus.Entry, i
 		return
 	}
 
+	if strings.Contains(backendErr.Error(), "one of the claims 'puid' or 'altsecid' or 'oid' should be present") {
+		backendErr = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalClaims,
+			"properties.servicePrincipleProfile", "The provided service principal does not give an access token with at least one of the claims 'altsecid', 'oid' or 'puid'. Please make sure service principal is properly created in the tenant.")
```
nwnt, line 345: We could lower the case of the error message in backendErr so that the case is always the same.

The returned error message could be something along the lines of "The ARO service principal has been removed from your tenant.  To restore, please unregister and then re-register the Azure Red Hat OpenShift resource provider.".  But please check with the PM for the exact message that should be returned.

carvalhe: Sent a message to Puneet to see what they want the error message to be, will adjust accordingly

pkg/validate/openshiftcluster_validatedynamic.go
```diff
@@ -96,7 +96,7 @@ func ensureAccessTokenClaims(ctx context.Context, spTokenCredential azcore.Token
 		http.StatusBadRequest,
 		api.CloudErrorCodeInvalidServicePrincipalClaims,
 		"properties.servicePrincipleProfile",
```
tsatam, line 98: typo: servicePrincipleProfile -> servicePrincipalProfile

s-fairchild (changes requested):
I think this looks good, let some comments. Maybe a good idea to turn that string into a constant to improve reuse in the future.

SudoBrendan (approved):

pkg/backend/openshiftcluster.go
```diff
@@ -339,6 +340,11 @@ func (ocb *openShiftClusterBackend) asyncOperationResultLog(log *logrus.Entry, i
 		return
 	}
 
+	if strings.Contains(strings.ToLower(backendErr.Error()), "one of the claims 'puid' or 'altsecid' or 'oid' should be present") {
+		backendErr = api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidServicePrincipalClaims,
+			"properties.servicePrincipalProfile", "The Azure Red Hat Openshift resource provider service principal has been removed from your tenant. To restore, please unregister and then re-register the Azure Red Hat OpenShift resource provider.")
```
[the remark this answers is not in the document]
carvalhe: Looking at other cloud errors I don't see a set up for constant values, but I think in the future that would be a good change to make and create constants for the cloud errors that use the same messages, placing them in the same spot.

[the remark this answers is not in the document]
s-fairchild: @carvalhe Makes sense, I agree this would be a good future change to make as a whole.

s-fairchild (approved):

PR #3336, opened 2023-12-21
Hopefully fix E2E flake related to deletion of a test Service

kimorris27:

https://issues.redhat.com/browse/ARO-4956

I encountered a test failure in a different PR where E2E errored out on the line of the call to delete the Service. This PR attempts to fix that error by handling the Service deletion in a (hopefully) more robust way.

### Test plan for issue:

N/A

N/A

### DIFF
--- a/test/e2e/adminapi_delete_managedresource.go
+++ b/test/e2e/adminapi_delete_managedresource.go
@@ -54,10 +54,7 @@ var _ = Describe("[Admin API] Delete managed resource action", func() {
 			// wait for deletion to prevent flakes on retries
 			Eventually(func(g Gomega, ctx context.Context) {
 				err := clients.Kubernetes.CoreV1().Services("default").Delete(ctx, "test", metav1.DeleteOptions{})
-				Expect(err).NotTo(HaveOccurred())
-
-				_, err = clients.Kubernetes.CoreV1().Services("default").Get(ctx, "test", metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


mociarain (dismissed):
LGTM

tony-schndr (dismissed):
LGTM, Thanks for fixing this!

mociarain (commented):
I'm not sure about this. As it was, we were certain that at the end of the `defer` that the cluster was back to an "empty" state. Now, we just know the delete request was accepted. I'm not sure how our test isolation works so maybe this is fine?

kimorris27:
> > I think in this case it's probably fine the way I wrote it in this PR because there's no reason that this particular test Service would have any time-consuming finalizers that would block it from getting deleted immediately. But the point you make is valid and could apply to other Kubernetes resources we create/delete throughout the E2E test suite.
> 
> This may be true now but could change in future. IMO the certainty of asserting the object is indeed not there is worth a lot (do challenge this). Maybe to step back for a second. Can you explain to me why this change is expected to be less flaky than the initial get request?

@mociarain - Point taken about how asserting that the object is no longer there is worthwhile. I put the `Get` call to confirm that the object is gone in its own `Eventually` like we discussed 1:1. This way we fully confirm that the Service is gone within the `defer` function, preventing retries from flaking because the cluster isn't in a clean state after each attempt.

mociarain (commented):
LGTM

mociarain (approved):

cadenmarchese (approved):

PR #3337, opened 2023-12-22
Add unit tests for ensureAccessTokenClaims

nwnt:

Fixes: N/A, just adding unit tests.

Adding unit tests for the ensureAccessTokenClaims function.  Also adding why the function is needed.

### Test plan for issue:
Unit tests ran successfully.

N/A


jaitaiwan (approved):
LGTM

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/Azure/azure-sdk-for-go v63.1.0+incompatible
-	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.7.1
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.9.1
```
cadenmarchese, line 7: Hmm, are the version bumps for `azure-sdk-for-go` needed for your unit test? I see that our current version provides `TokenCredential` as well, which I believe is what your test depends on.

SudoBrendan (approved):

go.mod
nwnt: Hi Caden, I had to upgrade this to 1.9.1 for using fake in my unit tests.  I wanted to start using this as Petr once mentioned we should rely less on mocks and this is the official way to do it per the SDK.

PR #3343, opened 2024-01-04
Make NSG Monitor run every 10 minutes

nwnt:
Fixes ARO-4854

This makes the NSG Monitor run every 10 minutes instead of every hour.  It accomplishes this by maintaining a time.Ticker of 10*time.Minute duration in the ~~monitor object~~ workOne function and using the tick as a signal to monitor during the main monitoring loop.

### Test plan for issue:
Unit tests 

N/A

### DIFF
--- a/pkg/monitor/azure/nsg/nsg.go
+++ b/pkg/monitor/azure/nsg/nsg.go
@@ -11,6 +11,7 @@ import (
 	"net/netip"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
@@ -18,6 +19,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/metrics"
 	"github.com/Azure/ARO-RP/pkg/monitor/dimension"
 	"github.com/Azure/ARO-RP/pkg/monitor/emitter"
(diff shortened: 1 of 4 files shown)


jaitaiwan (dismissed):
LGTM

pkg/monitor/monitor.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/liveconfig"
 )
 
+var NSGMonitorFrequency = 10 * time.Minute
```
niontive, line 27: Since this is specific to NSG monitor, maybe we can put this in `pkg/monitor/azure/nsg/nsg.go`?

mociarain (commented):
~~If I've understood this correctly we have a monitor that runs every X (now 10mins) and asserts that no invalid rule is present in a bunch of NSG's we care about. Do I have this right?~~
 
~~If so then I have the following questions~~:
- ~~It's not clear to me how this is addressing the linked issue. How does help with a Timing Side Channel attack?~~
- ~~The linked issue mentions that this is fixed in Go 1.20 which we should move to very shortly: https://github.com/Azure/ARO-RP/pull/3350. Does that have any bearing on this?~~

pkg/monitor/monitor.go
nwnt: I think this is more like a config for the monitor than something specific to the nsg monitoring package.

nwnt:
@mociarain seems like I linked to a wrong story, meant to link ARO-4854 (fixed now, my bad.. sorry about that).  Maybe that wrongly linked issue misled you into having those questions, but if not, can you elaborate more about your first question?

mociarain:
> @mociarain seems like I linked to a wrong story, meant to link [ARO-4854](https://issues.redhat.com/browse/ARO-4854) (fixed now, my bad.. sorry about that). Maybe that wrongly linked issue misled you into having those questions, but if not, can you elaborate more about your first question?

That explains it :D No worries. This looks great

mociarain (approved):

pkg/monitor/azure/nsg/nsg_test.go
```diff
@@ -92,6 +98,29 @@ var (
 	nsgOutbound = armnetwork.SecurityRuleDirectionOutbound
 )
 
+func createOC() api.OpenShiftCluster {
```
SudoBrendan, line 101: What's the major difference between this func and `ocFactory` below? They seem very similar. Would it be possible to parameterize/merge the diffs into a single func?

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -49,20 +51,45 @@ type NSGMonitor struct {
 	dims         map[string]string
 }
 
-func NewNSGMonitor(log *logrus.Entry, oc *api.OpenShiftCluster, subscriptionID string, subnetClient sdknetwork.SubnetsClient, emitter metrics.Emitter, wg *sync.WaitGroup) *NSGMonitor {
+func NewMonitor(log *logrus.Entry, oc *api.OpenShiftCluster, e env.Interface, subscriptionID string, tenantID string, emitter metrics.Emitter, dims map[string]string, wg *sync.WaitGroup, trigger <-chan time.Time) monitoring.Monitor {
+	if oc.Properties.NetworkProfile.PreconfiguredNSG != api.PreconfiguredNSGEnabled {
```
SudoBrendan, line 55: nil pointer check?

SudoBrendan (commented):
I'm gonna be a stickler about nil pointer checks now, let's get those in. LGTM otherwise.

pkg/monitor/monitor.go
bennerv, line 27: this isn't exported, correct?  It doens't need capitalization then.

bennerv (commented):
I think the nsg ticker is shared between goroutines, which I'm not sure what happens when multiple NSG monitors are waiting on the same ticker.  Assuming a single channel shared among all goroutines, only one will pass every 10 minutes for each RP monitor process (3 per region).

nwnt:
> I think the nsg ticker is shared between goroutines, which I'm not sure what happens when multiple NSG monitors are waiting on the same ticker. Assuming a single channel shared among all goroutines, only one will pass every 10 minutes for each RP monitor process (3 per region).

This is definitely a bug.  I have fixed this by having the ticker created within `workOne`.  Each workOne call will be associated with one goroutine invoked by `fixDoc`.  In other words, each monitoring go routine will maintain an NSG ticker and it will not be shared by any other monitoring routines.

pkg/monitor/azure/nsg/nsg_test.go
```diff
@@ -547,3 +566,99 @@ func wait(wg *sync.WaitGroup, done chan<- any) {
 	wg.Wait()
 	done <- nil
 }
+
+func isOfType[T any](mon monitoring.Monitor) bool {
```
mociarain, line 570: Nice to see more generics :) 

2 things:
- This type of 1-2 line function seems a bit Java-y to me (not a bad thing). Is it common to do this type of thing in Go too?
- Potentially of interest/curious to hear your thoughts on this: https://github.com/Azure/ARO-RP/pull/3376#discussion_r1478472271

mociarain: I get you. I like this and agree it's the best choice. Thanks for giving context :)

mociarain:
For reference the specific E2E errors here should be addressed by:
- https://github.com/Azure/ARO-RP/pull/3366 
- https://github.com/Azure/ARO-RP/pull/3388

When the second one is merged a rebase should get this unstuck

tsatam (approved):
Code change LGTM. I'd like to see more reviews from other people as I am not fully up-to-speed with the implications this may have across our entire monitor.

UlrichSchlueter (approved):
LGTM

PR #3344, opened 2024-01-05
Explicitly reference 1.18 only

mociarain:
This PR just updates the docs to make it clear what version of GO we support


bennerv:
@mociarain those docs are vendored in as part of go modules.  We shouldn't be updating them as we don't own the repositories which host them, and it will fail CI as a result of updating them.   Any README.md s at the root level or in `docs/` should be applicable.

jaitaiwan:
Thanks heaps for the first go @mociarain ; As per Ben's comment I'm going to close this out.

PR #3345, opened 2024-01-08
Add clienthelper, a replacement for dynamichelper

hawkowl:

Fixes [ARO-3930](https://issues.redhat.com/browse/ARO-3930)

Replacing dynamichelper in-place is basically impossible. This implements the changes that I'd like to do in it, so we can port code that doesn't have special weird requirements over to it.

### Test plan for issue:

Has unit tests

N/A


pkg/util/clienthelper/createorder.go
```diff
@@ -0,0 +1,39 @@
...
+
+	reflect.TypeOf(&corev1.Namespace{}): -4, // before namespaced resources
+
+	// namespaced resources
+	reflect.TypeOf(&corev1.ConfigMap{}):      -3, // before workload resources
+	reflect.TypeOf(&corev1.Secret{}):         -2, // before workload resources
+	reflect.TypeOf(&corev1.ServiceAccount{}): -1, // before workload resources
+
+	// everything else defaults to 0
+}
```
jaitaiwan, line 33: Ngl I am surprised this is allowed outside of an init() func

jaitaiwan:
I'm not sure if I understand all the changes and why they're in this PR compared with the related Jira issue. Most of the helpers seem helpful to me. I need more time to look into it. There were a couple of changes to the machine health check stuff that felt like it had been leaked into this PR so maybe something to check @hawkowl ?

pkg/util/clienthelper/clienthelper.go
```diff
@@ -0,0 +1,273 @@
...
+	_ "github.com/Azure/ARO-RP/pkg/util/scheme"
+)
+
+type Interface interface {
+	EnsureDeleted(ctx context.Context, gvk schema.GroupVersionKind, key types.NamespacedName) error
+	Ensure(ctx context.Context, objs ...kruntime.Object) error
+	GetOne(ctx context.Context, key types.NamespacedName, obj kruntime.Object) error
+}
+
+type dynamicHelper struct {
```
jaitaiwan, line 43: I think we should rename this to clientHelper. Might be a bit confusing that we're "removing dynamichelper" and replacing it with this and yet here is a dynamichelper.

pkg/util/clienthelper/clienthelper.go
```diff
@@ -0,0 +1,273 @@
...
+
+// merge takes the existing (old) and desired (new) objects.  It compares them
+// to see if an update is necessary, fixes up the new object if needed, and
+// returns the difference for debugging purposes.
+func merge(old, new client.Object) (client.Object, bool, string, error) {
+	if reflect.TypeOf(old) != reflect.TypeOf(new) {
+		return nil, false, "", fmt.Errorf("types differ: %T %T", old, new)
+	}
+
+	// 1. Set defaults on new.  This gets rid of many false positive diffs.
```
jaitaiwan, line 153: I think the explanation on this needs to be a little more explicit. I think what you're saying is that it does things like strips away "creationDate" etc. ?

pkg/util/clienthelper/clienthelper.go
```diff
@@ -0,0 +1,273 @@
...
+			caBundle, ext := old.Data["ca-bundle.crt"]
+			if ext {
+				new.Data["ca-bundle.crt"] = caBundle
+			}
+		}
+
+	case *machinev1beta1.MachineHealthCheck:
+		old, new := old.(*machinev1beta1.MachineHealthCheck), new.(*machinev1beta1.MachineHealthCheck)
+		new.Status = old.Status
+	}
```
jaitaiwan, line 245: Should we warn on a default case?

pkg/util/clienthelper/clienthelper.go
```diff
@@ -0,0 +1,273 @@
...
+
+	var diff string
+	if _, ok := old.(*corev1.Secret); !ok { // Don't show a diff if kind is Secret
+		diff = strings.Join(deep.Equal(old, new), "\n")
+	} else {
+		diff = "<scrubbed>"
```
jaitaiwan, line 251: Can we show more than scrubbed? e.g. the keys of the object?

pkg/util/clienthelper/clienthelper_test.go
```diff
@@ -0,0 +1,1502 @@
...
+			if obj.GetName() == "test-name-2" {
+				return fmt.Errorf("error on %s", obj.GetName())
+			}
+			return nil
+		})
+
+	dh := &dynamicHelper{
+		client: client,
+		log:    logrus.NewEntry(logrus.StandardLogger()),
+	}
```
jaitaiwan, line 58: Shouldn't we just use NewWithClient here?

jaitaiwan (dismissed):
Couple nits. LGTM

pkg/util/clienthelper/clienthelper.go
```diff
@@ -0,0 +1,273 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
```
jaitaiwan, line 27: I know we're trying to get away from using client-go. Is there a away to not use client-go's scheme?

jaitaiwan (approved):
LGTM; one concern about using client-go. Others are nits

pkg/util/clienthelper/clienthelper.go
hawkowl: It's used by controller-runtime as well -- some of client-go is still required for the purposes of this, we're just not using one specific part of it. https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/client.go#L31C1-L31C38

PR #3346, opened 2024-01-09
Min worker node errors

LiniSusan:

Fixes


pkg/cluster/condition.go
```diff
@@ -28,65 +32,85 @@ func (m *manager) bootstrapConfigMapReady(ctx context.Context) (bool, error) {
...
 				readyWorkers++
+			} else {
+				messageString := fmt.Sprintf("%+v - Status:%+v, Message: %+v\n", node, cond.Status, cond.Message)
+				message += messageString
 			}
 		}
 	}
 
-	return readyWorkers >= minimumWorkerNodes, nil
+	if readyWorkers >= minimumWorkerNodes {
```
AldoFusterTurpin, line 69: We could extract _readyWorkers >= minimumWorkerNodes_ to a temporal boolean variable as we are writing twice the expression.

pkg/cluster/condition.go
```diff
@@ -28,65 +32,85 @@ func (m *manager) bootstrapConfigMapReady(ctx context.Context) (bool, error) {
...
+		return readyWorkers >= minimumWorkerNodes, false, nil
+	} else {
+		if message == "" {
+			message = "Check the config and versions"
+		}
+		return false, true, api.NewCloudError(
+			http.StatusInternalServerError,
+			api.CloudErrorCodeDeploymentFailed,
+			"", message+" Minimum number of worker nodes have not been successfully created. Please retry, and if the issue persists, raise an Azure support ticket",
```
AldoFusterTurpin, line 78: The string " Minimum number of worker nodes ... ticket" could be extracted to a package variable (or at the beginning of this function) as it is long and used twice.

pkg/cluster/hive.go
```diff
@@ -34,14 +34,24 @@ func (m *manager) hiveEnsureResources(ctx context.Context) error {
...
-func (m *manager) hiveClusterDeploymentReady(ctx context.Context) (bool, error) {
+func (m *manager) hiveClusterDeploymentReady(ctx context.Context) (bool, bool, error) {
 	m.log.Info("waiting for cluster deployment to become ready")
-	return m.hiveClusterManager.IsClusterDeploymentReady(ctx, m.doc)
+	ok, err := m.hiveClusterManager.IsClusterDeploymentReady(ctx, m.doc)
+	retry := true
+	if err != nil {
+		retry = false
+	}
+	return ok, retry, err
```
AldoFusterTurpin, line 44: I think we can simplify the function if we make use of named return parameters + we remove the temporal _retry_ variable (we also return as soon as possible in case of an error). Something like:
```go
func (m *manager) hiveClusterDeploymentReady(ctx context.Context) (deploymentReady bool, retry bool, err error) {
	m.log.Info("waiting for cluster deployment to become ready")
	deploymentReady, err = m.hiveClusterManager.IsClusterDeploymentReady(ctx, m.doc)
	if err != nil {
		return deploymentReady, false, err
	}
	return deploymentReady, true, err
}
```

pkg/cluster/hive.go
```diff
@@ -34,14 +34,24 @@ func (m *manager) hiveEnsureResources(ctx context.Context) error {
...
 }
 
-func (m *manager) hiveClusterInstallationComplete(ctx context.Context) (bool, error) {
+func (m *manager) hiveClusterInstallationComplete(ctx context.Context) (bool, bool, error) {
```
AldoFusterTurpin, line 47: Same comment as in hiveClusterDeploymentReady to simplify the function.

AldoFusterTurpin (changes requested):
Thank you for this. 

I left an important comment about adding named return parameters to not confuse return values of the same type (in all functions returning 2 booleans) and some optional suggestions that could simplify the code (in case you find them useful).

Thanks!

jaitaiwan (changes requested):
I see a change to the release-note action which I don't believe belongs in this PR

LiniSusan:
Closing this PR as wrong commits were made here... 
Changes are updated in PR: https://github.com/Azure/ARO-RP/pull/3464

PR #3349, opened 2024-01-11
Add Go 1.20 images (and ose-tools-rhel8 for 4.14)

hawkowl:

chore

Updates the images so we can use 1.20

### Test plan for issue:

N/A

N/A


jaitaiwan (approved):
LGTM

PR #3351, opened 2024-01-12
Trim the dependency tree

hawkowl:

Chore -- updating dependencies

Unpins some dependencies that shouldn't need to be pinned, minor version updates, more explicit exclusions to clean up the dependency tree. Also excluded all the unused cloud.google APIs which should lighten the dependency tree math significantly.

### Test plan for issue:

CI, E2E

N/A


jaitaiwan (approved):
LGTM

go.mod
```diff
@@ -74,14 +74,14 @@ require (
 	github.com/tebeka/selenium v0.9.9
 	github.com/ugorji/go/codec v1.2.7
 	github.com/vincent-petithory/dataurl v1.0.0
-	golang.org/x/crypto v0.17.0
-	golang.org/x/net v0.19.0
+	golang.org/x/crypto v0.18.0
+	golang.org/x/net v0.20.0
 	golang.org/x/oauth2 v0.10.0
```
SudoBrendan, line 79: Am I not reading this correctly, is the PR missing something, or is the PR comment out of date? The only `oauth2` changes are found in `exclude` blocks. Is OAuth2 getting a version bump?

hawkowl: I reverted it because it seemed to cause some problems by being too new -- it looks like we need Go 1.20 first for it.

PR #3353, opened 2024-01-15
Apply mc-name-suffix annotation to kubeletconfig

SrinivasAtmakuri:
JIRA: https://issues.redhat.com/browse/ARO-5028

We want to proactively add MCO's `mc-name-suffix` annotation set to empty string "" to the kubeletconfig dynamic-node. This will make the resulting rendered machineconfig 99-worker-generated-kubelet have no suffix, resulting in the lowest priority of kubeletconfigs.

Fixes


pkg/operator/controllers/autosizednodes/autosizednodes_controller.go
```diff
@@ -32,6 +32,10 @@ type Reconciler struct {
 const (
 	ControllerName = "AutoSizedNodes"
 	configName     = "dynamic-node"
+
+	// machineconfiguration annotation added on the kubelet CRD - dynamic-node
+	mcAnnotationName  = "machineconfiguration.openshift.io/mc-name-suffix"
+	mcAnnotationValue = ""
```
SudoBrendan, line 38: I would comment why we use `""` here just so we remember in a while, it's unintuitive.

SudoBrendan (approved):

pkg/operator/controllers/autosizednodes/autosizednodes_controller.go
```diff
@@ -90,7 +94,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return ctrl.Result{}, fmt.Errorf("could not fetch KubeletConfig: %w", err)
 	}
 
-	// If already exists, update the spec
+	// If already exists, update the spec and annotations
+	if config.Annotations == nil {
+		config.Annotations = defaultConfig.Annotations
+	}
+	if val, ok := config.Annotations[mcAnnotationName]; !ok || val != mcAnnotationValue {
```
cblecker, line 101: If this is already set, we shouldn't be changing it regardless of value, as it will disconnect the kubeletconfig from the already rendered machineconfig

cblecker (changes requested):

pkg/operator/controllers/autosizednodes/autosizednodes_controller.go
```diff
@@ -90,7 +94,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return ctrl.Result{}, fmt.Errorf("could not fetch KubeletConfig: %w", err)
 	}
 
-	// If already exists, update the spec
+	// If already exists, update the spec and annotations
+	if config.Annotations == nil {
```
dofinn, line 98: nit; but should be using `len (config.Annotations == 0)` to determine if the map is empty https://go.dev/ref/spec#Length_and_capacity

pkg/operator/controllers/autosizednodes/autosizednodes_controller.go
```diff
@@ -90,7 +94,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		config.Annotations = defaultConfig.Annotations
+	}
+	if val, ok := config.Annotations[mcAnnotationName]; !ok || val != mcAnnotationValue {
+		config.Annotations[mcAnnotationName] = mcAnnotationValue
```
dofinn, line 102: What happens if a different `kubeletconfig` (maybe one provided by the customer) already his this annotation and value?

pkg/operator/controllers/autosizednodes/autosizednodes_controller.go
dofinn: This is true, we understand MCO behaviour in this condition.

pkg/operator/controllers/workaround/const.go
```diff
@@ -16,6 +16,10 @@ const (
 	workerMachineConfigPoolName = "worker"
 	memReserved                 = "2000Mi"
 
+	// 	machineconfiguration annotation added on the kubelet CRD - aro-limits
+	mcAnnotationName  = "machineconfiguration.openshift.io/mc-name-suffix"
```
dofinn, line 20: These are already declared in nodeautoresize in this PR. This should be moved to consistent spot, https://github.com/Azure/ARO-RP/blob/master/pkg/operator/const.go

pkg/operator/controllers/workaround/systemreserved.go
```diff
@@ -87,6 +87,10 @@ func (sr *systemreserved) Ensure(ctx context.Context) error {
 			kc.Labels = make(map[string]string)
 		}
 		kc.Labels[labelName] = labelValue
+		if kc.Annotations == nil {
```
dofinn, line 90: This will blindly `Update` and won't check for existence of other kubeletconfigs that may have the same annotation. This will put MCO in an unknown condition.

PR #3356, opened 2024-01-17
Adjust the NSG validation logic for GA

nwnt:

Fixes: ARO-2590

To adjust the validation logic of NSG validation for GA.  Previously in private preview, we allow no nsgs to be attached to the subnets which the FP then assumes that the feature shouldn't be used.  In GA, we make it explicit so that when the preconfigured nsg flag is enabled and no subnets are attached, the FP will error out instead.

### Test plan for issue:
Unit tests.

no


mociarain (commented):
LGTM

jaitaiwan (approved):
LGTM

cadenmarchese (approved):

PR #3357, opened 2024-01-17
Delete AFEC flag behavior for preconfiguredNSG

nwnt:

Fixes: ARO-2593

For GA, we will no longer rely on AFEC flag.  So deleting this part in the code.

### Test plan for issue:
Unit tests, e2e tests.

No

### DIFF
--- a/pkg/frontend/openshiftcluster_putorpatch.go
+++ b/pkg/frontend/openshiftcluster_putorpatch.go
@@ -22,7 +22,6 @@ import (
 	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
 	"github.com/Azure/ARO-RP/pkg/operator"
-	"github.com/Azure/ARO-RP/pkg/util/feature"
 	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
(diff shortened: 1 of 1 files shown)


mociarain (commented):
Makes sens to me but a quick grep showed me that there are 2 other files that directly reference `byonsg` :

- https://github.com/Azure/ARO-RP/blob/016ba7b4b962a1398111822566b79d570df897fd/pkg/cluster/validate.go#L22
- https://github.com/Azure/ARO-RP/blob/6b122e52acf4f5bb99ea1498221f5df52a8ebe21/pkg/validate/dynamic/dynamic_test.go


for `PreconfiguredNSG` there are tonnes of references.

Should we deal with these also or can they be safely ignore?

nwnt:
> Makes sense to me but a quick grep showed me that there are 2 other files that directly reference `byonsg` :
> 
> * https://github.com/Azure/ARO-RP/blob/016ba7b4b962a1398111822566b79d570df897fd/pkg/cluster/validate.go#L22
> * https://github.com/Azure/ARO-RP/blob/6b122e52acf4f5bb99ea1498221f5df52a8ebe21/pkg/validate/dynamic/dynamic_test.go
> 
> and for `PreconfiguredNSG` there are tonnes of references.
> 
> Should we deal with these also or can they be safely ignore?

Good call on the first one, thank you.  For the second one, I'll address it in #3356   to avoid requiring rebasing.  Note that BYO NSG was renamed to Preconfigured NSG, so the language is correct.  If you see where the AFEC flag is used, feel free to let me know.

mociarain (approved):
Love seeing PR's that are 100% deletion

cadenmarchese (approved):
lgtm! Do we have this feature / change documented for customers somewhere?

nwnt:
@cadenmarchese that'll be part of the official GA.  We have a story for that.

PR #3359, opened 2024-01-19
Update deploy-development-rp.md to include a heads up.

anshulvermapatel:
This is to give a heads up that this podman issue can occur if the podman user socket is not enabled.


Shivkumar13 (approved):
LGTM

PR #3360, opened 2024-01-19
accept pull-secret when we use hack script to create a clsuter

bitoku:

N/A

There is no way to specify pull-secret when we use the hack script to create a cluster.
When testing the installer, if the installer image is hosted in quay.io, hive doesn't changes the domains of other images required during installation to ACR domain and it causes Authentication error if it doesn't have the pull-secret.

This pull-secret enables us to access to those images in quay.io and make the installer test convenient.

### Test plan for issue:

I was able to create a cluster with ACR domain installer and quay.io domain installer.

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -447,6 +447,7 @@ func (c *Cluster) createCluster(ctx context.Context, vnetResourceGroup, clusterN
 				ResourceGroupID:      fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", c.env.SubscriptionID(), "aro-"+clusterName),
 				FipsValidatedModules: api.FipsValidatedModulesEnabled,
 				Version:              osClusterVersion,
+				PullSecret:           api.SecureString(os.Getenv("USER_PULL_SECRET")),
 			},
 			ServicePrincipalProfile: api.ServicePrincipalProfile{
 				ClientID:     clientID,


mociarain (approved):
LGTM

pkg/util/cluster/cluster.go
```diff
@@ -447,6 +447,7 @@ func (c *Cluster) createCluster(ctx context.Context, vnetResourceGroup, clusterN
 				ResourceGroupID:      fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", c.env.SubscriptionID(), "aro-"+clusterName),
 				FipsValidatedModules: api.FipsValidatedModulesEnabled,
 				Version:              osClusterVersion,
+				PullSecret:           api.SecureString(os.Getenv("USER_PULL_SECRET")),
```
Shivkumar13, line 450: Do you want to validate if Pull Secret content is valid? I'm still okay with the approach here.

bitoku: I think we don't have to check it because this function is only used  for hack cluster creation.
And PULL_SECRET is set in the secret env file, so as long as we run `. ./env`, we won't set it explicitly.

Shivkumar13 (approved):

pkg/util/cluster/cluster.go
Shivkumar13: Makes sense

Shivkumar13:
LGTM!

pkg/util/cluster/cluster.go
cadenmarchese, line 450: Can we include an accompanying docs change in this PR?

We can update this one with instructions on passing in a pull secret: https://github.com/Azure/ARO-RP/blob/master/docs/deploy-development-rp.md#run-the-rp-and-create-a-cluster

bitoku: I don't think we need it, because `USER_PULL_SECRET` is already configured in the secret, so we don't need to set it explicitly.

And also it is not meant as a general way of configuring the pull-secret.
If we want to use other pull-secret then we can set it after creation, so we don't need to use this `USER_PULL_SECRET` variable. 
It is only needed when we want to use a installer image stored in other place than ACR.

bitoku:
close this PR because it continues to fail E2E.
I think this is because I made the PR from my fork.

PR #3361, opened 2024-01-19
Validate required customer service principal permissions

rhamitarora:

[Alert on loss of access rights and permissions on cluster / network resources](https://issues.redhat.com/browse/ARO-1987)

### DIFF
--- a/pkg/monitor/cluster/validatepermissions.go
+++ b/pkg/monitor/cluster/validatepermissions.go
@@ -1,0 +1,46 @@
+package cluster
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+
+	"github.com/Azure/ARO-RP/pkg/validate/dynamic"
+)
+
+/***************************************************************
+	Monitor the Cluster Service Prinicpal required permissions:
+	- Network Contributor role on vnet
+****************************************************************/
+
+func (mon *Monitor) emitValidatePermissions(ctx context.Context) error {
+	subnets := []dynamic.Subnet{{
+		ID:   mon.oc.Properties.MasterProfile.SubnetID,
+		Path: "properties.masterProfile.subnetId",
+	}}
+
+	err := mon.validator.ValidateVnet(ctx, mon.oc.Location, subnets, mon.oc.Properties.NetworkProfile.PodCIDR,
+		mon.oc.Properties.NetworkProfile.ServiceCIDR)
+
+	if err != nil {
+		mon.emitGauge("cluster.validateVnet.permissions", 1, map[string]string{
+			"vnetError": err.Error(),
+		})
+	}
+
+	err = mon.validator.ValidateSubnets(ctx, mon.oc, subnets)
+	if err != nil {
+		mon.emitGauge("cluster.validateSubnets.permissions", 1, map[string]string{
+			"subnetError": err.Error(),
+		})
+	}
+
+	err = mon.validator.ValidateDiskEncryptionSets(ctx, mon.oc)
+	if err != nil {
+		mon.emitGauge("cluster.validateDiskEncryptionSets.permissions", 1, map[string]string{
+			"diskEncryptionSetError": err.Error(),
+		})
+	}
+	return nil
+}
--- a/pkg/monitor/cluster/validatepermissions_test.go
+++ b/pkg/monitor/cluster/validatepermissions_test.go
@@ -1,0 +1,115 @@
+package cluster
+
+// Copyright (c) Microsoft Corporation.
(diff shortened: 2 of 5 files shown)


AldoFusterTurpin (changes requested):
I know it is in draft mode. Just a quick comment about the panic for now. I will do a full review when it is in non draft mode. 🙂

pkg/monitor/cluster/validatepermissions_test.go
```diff
@@ -0,0 +1,115 @@
...
+			oc := &api.OpenShiftCluster{}
+
+			mon := &Monitor{
+				m:         m,
+				oc:        oc,
+				validator: validatorMock,
+			}
+
+			err := validatorMock.ValidateVnet(ctx, tt.location, tt.subnet, tt.ServiceCIDR, tt.PodCIDR)
+			fmt.Println(err.Error())
```
AldoFusterTurpin, line 102: I know this is not a definitive version, but I think the panic you mentioned on Slack is coming from that line as err is nil at that point and then err.Error panics. I sent on Slack a fix, in case it helps. 🙂

pkg/monitor/cluster/validatepermissions_test.go
```diff
@@ -0,0 +1,115 @@
...
+			os.Exit(-1)
+
+			m.EXPECT().EmitGauge("cluster.validateVnet.permissions", int64(1), map[string]string{
+				"vnetError": "nil",
+			})
```
AldoFusterTurpin, line 107: This is also causing the test to fail as not all expectations from the mock are present. I have added it on the message I mentioned before 🙂

pkg/monitor/cluster/validatepermissions.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/Azure/ARO-RP/pkg/validate/dynamic"
+)
+
+/***************************************************************
+	Monitor the Cluster Service Prinicpal required permissions:
+	- Network Contributor role on vnet
+****************************************************************/
+
+func (mon *Monitor) emitValidatePermissions(ctx context.Context) error {
```
AldoFusterTurpin, line 17: I would also check against nil on the pointers that can be nil (like mon.oc, mon.validator...) and return an error if they are nil (I have added that to our Slack conversation).

pkg/monitor/cluster/validatepermissions_test.go
```diff
@@ -0,0 +1,115 @@
...
+			PodCIDR:       "10.128.0.0/14",
+			ServiceCIDR:   "0.0.0.0/0",
+			subnet:        []dynamic.Subnet{},
+			SubnetId:      fmt.Sprintf("/subscriptions/%s/resourceGroups/vnet/providers/Microsoft.Network/virtualNetworks/test-vnet/subnets/master", subscriptionID),
+			expectedState: nil,
+			validator: func(controller *gomock.Controller) dynamic.Dynamic {
+				validator := mock_dynamic.NewMockDynamic(controller)
+				validator.EXPECT().ValidateVnet(ctx, gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any())
+				return validator
+			},
```
AldoFusterTurpin, line 65: The test also fails due to the fact that we are missing some expectations in the mock (the other calls to mon.validator).

AldoFusterTurpin (commented):
Some comments pointing to things that could be causing you problems when testing it 🙂

pkg/monitor/cluster/validatepermissions.go
```diff
@@ -0,0 +1,46 @@
...
+	subnets := []dynamic.Subnet{{
+		ID:   mon.oc.Properties.MasterProfile.SubnetID,
+		Path: "properties.masterProfile.subnetId",
+	}}
+
+	err := mon.validator.ValidateVnet(ctx, mon.oc.Location, subnets, mon.oc.Properties.NetworkProfile.PodCIDR,
```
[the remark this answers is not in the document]
rhamitarora: done

pkg/monitor/cluster/validatepermissions_test.go
rhamitarora: done

pkg/monitor/cluster/validatepermissions.go
rhamitarora: done

pkg/monitor/cluster/validatepermissions_test.go
rhamitarora: done

jaitaiwan:
Some of the comments on this PR remain unresolved despite being marked as such. I don't know why I'm being notified about these PRs still... but I figured I might as well take a look since I'm bored.

PR #3362, opened 2024-01-22
Remove Dockerfile.aro and the image-aro Makefile

hawkowl:

Breaking up some of the Go 1.20 changes that aren't strictly related.

Fixes https://issues.redhat.com/browse/ARO-5135

Removes the image-aro makefile target and the Dockerfile. This build target is legacy and does the wrong thing in general, and could lead to inconsistent results depending on the host Golang.

### Test plan for issue:

CI, E2E, should validate it's not used anywhere

Updated the remaining reference

### DIFF
--- a/Dockerfile.aro
+++ b/Dockerfile.aro
@@ -1,7 +1,0 @@
-ARG REGISTRY
-FROM ${REGISTRY}/ubi8/ubi-minimal
-RUN microdnf update && microdnf clean all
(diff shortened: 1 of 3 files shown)


SudoBrendan (approved):
Love it. Probably worth a callout in the release channel so people are aware.

cadenmarchese (approved):

PR #3365, opened 2024-01-22
Added Preflight Update Checks

carvalhe:

Currently we have preflight set up for create operations but also need it for update. These changes allow for both put/patch operations on preflight so that it can leave the testing stage and go live. This PR also cleans up the way we validate for preflight and adds new checks for it as well. There was a previous PR for this here: https://github.com/Azure/ARO-RP/pull/3181 but it was getting messy, and issues were arising so creating this new PR instead.

### Test plan for issue:
Created additional unit tests that are similar to put/patch tests. Kusto logs also show preflight success and failures with more details than before.

### DIFF
--- a/pkg/api/preflight.go
+++ b/pkg/api/preflight.go
@@ -14,19 +14,8 @@ type PreflightRequest struct {
 
 // ValidationResult is the validation result to return in the deployment preflight response body
 type ValidationResult struct {
-	Status ValidationStatus            `json:"status"`
-	Error  *ManagementErrorWithDetails `json:"error,omitempty"`
-}
-
-type ManagementErrorWithDetails struct {
-	// Code - The error code returned from the server.
-	Code *string `json:"code,omitempty"`
-	// Message - The error message returned from the server.
-	Message *string `json:"message,omitempty"`
-	// Target - The target of the error.
-	Target *string `json:"target,omitempty"`
-	// Details - Validation error.
-	Details *[]ManagementErrorWithDetails `json:"details,omitempty"`
+	Status ValidationStatus `json:"status"`
+	Error  *CloudErrorBody  `json:"error,omitempty"`
 }
 
 // ResourceTypeMeta is the Typemeta inside request body of preflight
--- a/pkg/frontend/openshiftcluster_preflightvalidation.go
+++ b/pkg/frontend/openshiftcluster_preflightvalidation.go
@@ -9,10 +9,10 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/sirupsen/logrus"
 
 	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
 	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
 )
@@ -36,8 +36,8 @@ func (f *frontend) preflightValidation(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		b = marshalValidationResult(api.ValidationResult{
 			Status: api.ValidationStatusFailed,
-			Error: &api.ManagementErrorWithDetails{
-				Message: to.StringPtr(err.Error()),
+			Error: &api.CloudErrorBody{
+				Message: err.Error(),
 			},
 		})
 		reply(log, w, header, b, statusCodeError(http.StatusOK))
@@ -52,9 +52,9 @@ func (f *frontend) preflightValidation(w http.ResponseWriter, r *http.Request) {
 			continue
 		}
 		if strings.EqualFold(typeMeta.Type, "Microsoft.RedHatOpenShift/openShiftClusters") {
-			res := f._preflightValidation(ctx, log, raw, typeMeta.APIVersion, typeMeta.Id)
+			res := f._preflightValidation(ctx, log, raw, typeMeta.APIVersion, strings.ToLower(typeMeta.Id))
 			if res.Status == api.ValidationStatusFailed {
-				log.Warningf("preflight validation failed")
+				log.Warningf("preflight validation failed with err: '%s'", res.Error.Message)
 				b = marshalValidationResult(res)
 				reply(log, w, header, b, statusCodeError(http.StatusOK))
 				return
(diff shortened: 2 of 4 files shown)


pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -9,10 +9,10 @@ import (
 	"net/http"
```
mociarain, line 9: Generic question. Why do have this flat structure in the FE directory? A lot of files have a prefix that makes me think they should be in a sub-package... What am I missing?

mociarain (approved):
LGTM and sorry for being the person who gets my random questions :D

pkg/api/preflight.go
```diff
@@ -14,19 +14,8 @@ type PreflightRequest struct {
...
-	// Code - The error code returned from the server.
-	Code *string `json:"code,omitempty"`
-	// Message - The error message returned from the server.
-	Message *string `json:"message,omitempty"`
-	// Target - The target of the error.
-	Target *string `json:"target,omitempty"`
-	// Details - Validation error.
-	Details *[]ManagementErrorWithDetails `json:"details,omitempty"`
+	Status ValidationStatus `json:"status"`
+	Error  *CloudErrorBody  `json:"error,omitempty"`
```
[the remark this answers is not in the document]
carvalhe: This was more from the older PR, where we discussed making a ValidationResult instead of adding to the existing CloudErrorBody, and just use inheritance. The format of CloudErrorBody is the standard but might be called a different name.

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -68,6 +68,17 @@ func (f *frontend) preflightValidation(w http.ResponseWriter, r *http.Request) {
...
 
 func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry, raw json.RawMessage, apiVersion string, resourceID string) api.ValidationResult {
+	log.Infof("running preflight validation on resource: %s", resourceID)
+	doc, err := f.dbOpenShiftClusters.Get(ctx, resourceID)
+	isCreate := cosmosdb.IsErrorStatusCode(err, http.StatusNotFound)
+	if err != nil && !cosmosdb.IsErrorStatusCode(err, http.StatusNotFound) {
+		return api.ValidationResult{
+			Status: api.ValidationStatusFailed,
+			Error: &api.CloudErrorBody{
+				Message: err.Error(),
```
SudoBrendan, line 78: If we're special-casing cluster documents not found here, and these errors get sent back to customers, is it better to return a well-defined string, something like "Custer not found for resourceID %s" instead of what I assume will be something related to CosmosDB?

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -79,35 +90,45 @@ func (f *frontend) _preflightValidation(ctx context.Context, log *logrus.Entry,
...
 	staticValidator := f.apis[apiVersion].OpenShiftClusterStaticValidator
 	ext := converter.ToExternal(oc)
-	if err := json.Unmarshal(raw, &ext); err != nil {
+	if err = json.Unmarshal(raw, &ext); err != nil {
 		return api.ValidationResult{
 			Status: api.ValidationStatusFailed,
-			Error: &api.ManagementErrorWithDetails{
-				Message: to.StringPtr(err.Error()),
+			Error: &api.CloudErrorBody{
+				Message: err.Error(),
```
SudoBrendan, line 97: Same as above, if I'm reading correctly these are returned to customers - should we be sending them "Could not parse L35 C15 '}' expected ']'" errors, or is it better to log that message and return a 5xx error because our API is fundamentally broken?

pkg/frontend/openshiftcluster_preflightvalidation_test.go
```diff
@@ -108,49 +143,259 @@ func TestPreflightValidation(t *testing.T) {
...
+				}
+			},
+			wantStatusCode: http.StatusOK,
+			wantResponse: &api.ValidationResult{
+				Status: api.ValidationStatusFailed,
+				Error: &api.CloudErrorBody{
+					Message: "400: PropertyChangeNotAllowed: properties.clusterProfile.domain: Changing property 'properties.clusterProfile.domain' is not allowed.",
 				},
 			},
 		},
```
SudoBrendan, line 393: Can we add test cases for the errors I point out (cosmos, json parse), perhaps I'm wrong on what those errors would be formatted as, or it would otherwise showcase expected customer experience.

SudoBrendan (commented):
Overall LGTM, I worry about a few internal-error-looking cases being returned to customers directly.

pkg/api/preflight.go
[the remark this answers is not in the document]
bennerv: @mociarain All errors in Azure are dictated to use the OData error format, but in this specific API they expect an additional field. 

http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091

pkg/frontend/openshiftcluster_preflightvalidation.go
bennerv: no reason, feel free to propose a refactor if you'd like :)

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: For preflight, it expects a 200 response each time. We could log the message for it though

jaitaiwan (commented):
LGTM apart from Brendan's concerns about returning internal errors to our customers. I am also concerned that based on recently experiences, that a 404 might just mean "it's not in the db _yet_". It might be good to make sure that cosmosdb does indeed return 404 for "it's not there and you shouldn't wait".

pkg/frontend/openshiftcluster_preflightvalidation_test.go
carvalhe: I changed the errors to what you mentioned to show internalServerError (and log out a more detailed error) for the json parse, as well as changed the message for the cosmos case. I made a unit test for the cosmos one but for the unmarashalling json there are a few similar checks done before that, so I didn't make a test as I'm not sure how to hit this specific mark without turning it into a function. 

Sorry this took so long to address, got wrapped up in MIWI work and coming back to finish this task while I have a moment. Hopefully things are ok.

SudoBrendan (changes requested):
Hey Brian, sorry to take so long on re-review. I really like the changes, thank you!

I mostly have nits here - feel free to ignore/action them. I think it is worth reverting/rewriting the Subscription test data so we don't conflate Subscription docs with OpenShiftCluster docs.

pkg/frontend/openshiftcluster_preflightvalidation_test.go
```diff
@@ -6,96 +6,131 @@ package frontend
...
+			name: "Successful Preflight Create",
 			fixture: func(f *testdatabase.Fixture) {
 				f.AddSubscriptionDocuments(&api.SubscriptionDocument{
-					ID: mockSubID,
+					ID: api.ExampleOpenShiftClusterDocument().ID,
 					Subscription: &api.Subscription{
 						State: api.SubscriptionStateRegistered,
 						Properties: &api.SubscriptionProperties{
-							TenantID: "11111111-1111-1111-1111-111111111111",
+							TenantID: api.ExampleOpenShiftClusterDocument().ClientIDKey,
```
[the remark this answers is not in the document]
carvalhe: Created ExampleSubscriptionDocument() and use that now instead

pkg/frontend/openshiftcluster_preflightvalidation.go
```diff
@@ -52,9 +52,9 @@ func (f *frontend) preflightValidation(w http.ResponseWriter, r *http.Request) {
 			continue
```
cadenmarchese, line 52: Won't all subsequent validation fail if we continue here?

cadenmarchese (commented):
The e2e failure should be resolved before merge (the branch will need to be rebased) but the changes look good to me! I left a comment on something I noticed during review we might want to fix while we're here.

pkg/frontend/openshiftcluster_preflightvalidation.go
carvalhe: Not necessarily as we are unmarshalling each resource separately. If one fails to unmarshall it doesn't mean the rest will, and according to preflight, is not considered a validation failure either.

PR #3366, opened 2024-01-23
Wrap K8s client in E2E with retry logic

mociarain:

Fixes https://issues.redhat.com/browse/ARO-4489#L67 and 

This PR addresses the first part of initiative 1 outlined [here](https://docs.google.com/document/d/1mVWg2K0sct8fcYrbrCY18tEJ8HxX4VWvgWSVRuNFQOQ/edit#heading=h.i1hq6tedaz2).

TODO:
- [ ] (Assuming the approach is okay'd) Apply expand this PR to fully cover the k8s API calls.
- [ ] ~~(Assuming the approach is okay'd) Start another PR/PR's for the other clients.~~ Not doing. See [here](https://docs.google.com/document/d/1mVWg2K0sct8fcYrbrCY18tEJ8HxX4VWvgWSVRuNFQOQ/edit#heading=h.h9k88a9fc1uv) for details.
- [ ] Bump ginkgo to 2.8.x (We're on 2.7.0) at least to get the [GinkoHelper](https://github.com/onsi/ginkgo/blob/master/CHANGELOG.md#280) and apply that to the helper methods. ~~Ideally, I would do this here as part of this PR. Is bumping the version of a dependency a big deal? AFAIK it's used (nearly) entirely in the E2E tests so I'm hoping it will be smooth.~~ Will do in a subsequent PR. This one is meaty enough.

### Test plan for issue:

Run them all.

Very possibly


test/e2e/helpers.go
```diff
@@ -0,0 +1,35 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	DefaultTimeout  = 5 * time.Second
+	PollingInterval = 250 * time.Millisecond
+)
```
mociarain, line 19: Unclear to me at this point what these numbers should be and if we should expose them in the API for users to mess with

kimorris27 (commented):
I like this approach!

Re: your concern about where/when to bump the dependency, I don't see any problem with doing it in the same PR as this work. I think that's typically what we do unless it's an especially scary/large set of dependency bumps.

tsatam (dismissed):
I'm a big fan of this approach and implementation, I love the usage of generics here. 

I'm in favor of having this PR, if acceptable to the wider team, merge as-is, and to make the effort of rolling this out across all our E2E tests the work of future stories, that are spread across the team. It'd be slower, but it would get more people involved in the change to hopefully normalize this approach across future tests as well.

mociarain:
> I'm in favor of having this PR, if acceptable to the wider team, merge as-is, and to make the effort of rolling this out across all our E2E tests the work of future stories, that are spread across the team. It'd be slower, but it would get more people involved in the change to hopefully normalize this approach across future tests as well.

I like this approach but I would suggest that splitting the work by client would be a more natural division. I can expand this PR to cover the k8s client and version bump and then people could pick up each of the other clients? 

What do you think?

(I actually have a second proposal that builds on this concept that I thought would be useful to spread around for the same reasons you outline but I think you're right and we can start introducing it from here)

tsatam:
I think that works too, my goal is to split work on this across the team, whichever way we split it is fine by me as long as it gets enough eyes on the change.

tsatam (dismissed):
👍🏼

kimorris27 (dismissed):

test/util/project/project.go
```diff
@@ -1,90 +0,0 @@
-package project
```
mociarain, line 1: This was only being used by `cluster.go` so I just co-located it.

test/e2e/adminapi_cluster_getlogs.go
```diff
@@ -45,21 +45,23 @@ func testGetPodLogsOK(ctx context.Context, containerName, podName, namespace str
 	expectedLog := "mock-pod-logs"
 
 	By("creating a test pod in openshift-azure-operator namespace with some known logs")
-	pod := mockPod(containerName, podName, namespace, expectedLog)
-	pod, err := clients.Kubernetes.CoreV1().Pods(namespace).Create(ctx, pod, metav1.CreateOptions{})
-	Expect(err).NotTo(HaveOccurred())
+	podDefinition := mockPod(containerName, podName, namespace, expectedLog)
```
tsatam, line 48: do we need to maintain a separate reference to the old definition after creation here?

tsatam (approved):
LGTM - one minor nit but it's not a blocker

test/e2e/cluster.go
```diff
@@ -27,36 +27,31 @@ import (
...
 	"github.com/Azure/ARO-RP/pkg/util/version"
-	"github.com/Azure/ARO-RP/test/util/project"
 )
 
 const (
 	testPVCName = "e2e-test-claim"
 )
 
 var _ = Describe("Cluster", Serial, func() {
-	var p project.Project
```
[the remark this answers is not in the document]
mociarain: :musical_note:  I like, long, names and I cannot lie :musical_note: ;)

test/e2e/adminapi_cluster_getlogs.go
tsatam: we don't ever use `podDefinition` after creating the `pod`, and they share the same type (pointer to `corev1.Pod`) so I think the old implementation where we reuse the same variable name/pointer makes more sense.

mociarain: I thought that's what you meant. I find this re-using of the same variable name weird but I think I just need to adjust to it :D. I'll make that change now

tsatam: like I said, it's a minor nit and I'm fine with it merging as-is, just want to make sure I wasn't missing any specific intent here for handling the two values separately.

mociarain: No no. I'm keen to get find the standard we have so do go for it with conventions and things like this.

Ty

mociarain:
@tsatam & @kimorris27 Sorry. AFAIK I didn't dismiss your reviews... tbh I have no idea how I triggered this action. This has actually happened me in a few PRs. Is there some implicit workflow or something I'm unaware of?

test/e2e/helpers.go
```diff
@@ -0,0 +1,188 @@
...
+	p.VerifyProjectIsDeleted(ctx)
+}
+
+func (p Project) Delete(ctx context.Context) {
+	deleteCall := p.projectClient.ProjectV1().Projects().Delete
+	DeleteK8sObjectWithRetry(ctx, deleteCall, p.Name, metav1.DeleteOptions{})
+}
+
+// VerifyProjectIsReady verifies that the project and relevant resources have been created correctly and returns error otherwise
+func (p Project) Verify(ctx context.Context) error {
```
AldoFusterTurpin, line 147: Outdated comment or method name (VerifyProjectIsReady vs Verify).

 Also, I would say ..."and returns an error if any" as maybe the creation can be successful but other errors can arise.

AldoFusterTurpin (approved):
LGTM. Thank you for this @mociarain. This is a good improvement.

I just left some optional suggestion in case you find any useful.

test/e2e/adminapi_csr.go
```diff
@@ -37,16 +37,17 @@ var _ = Describe("[Admin API] CertificateSigningRequest action", func() {
 		By("creating mock CSRs via Kubernetes API")
 		for i := 0; i < csrCount; i++ {
 			csr := mockCSR(prefix+strconv.Itoa(i), namespace, csrData)
-			_, err := clients.Kubernetes.CertificatesV1().CertificateSigningRequests().Create(ctx, csr, metav1.CreateOptions{})
-			Expect(err).NotTo(HaveOccurred())
+
+			createCall := clients.Kubernetes.CertificatesV1().CertificateSigningRequests().Create
```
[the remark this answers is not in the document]
mociarain: Good call ( :new_moon_with_face: ) . I went back and forth on Call/Method/something else.  I'll switch to *Func

test/e2e/adminapi_redeployvm.go
```diff
@@ -108,25 +109,30 @@ func getNodeUptime(g Gomega, ctx context.Context, node string) (time.Time, error
...
-		if err != nil {
-			log.Error("Could not delete test Pod")
-		}
+		By("deleting the uptime pod via Kubernetes API")
+		deleteCall := clients.Kubernetes.CoreV1().Pods(namespace).Delete
+		DeleteK8sObjectWithRetry(ctx, deleteCall, podName, metav1.DeleteOptions{})
+
+		// To avoid flakes, we need it to be completely deleted before we can use it again
+		// in a separate run or in a separate It block
+		By("waiting for uptime pod to be deleted")
```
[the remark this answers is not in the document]
mociarain: I'll leave it as is for now since this comment is getting nuked in a subsequent PR that I'll open after this: https://github.com/mociarain/ARO-RP/pull/1.

test/e2e/cluster.go
```diff
@@ -290,14 +280,14 @@ func clusterSubnets(oc redhatopenshift.OpenShiftCluster) []string {
 	return subnets
 }
 
-func createStatefulSet(ctx context.Context, cli kubernetes.Interface, namespace, storageClass string) (string, error) {
+func createStatefulSet(ctx context.Context, cli kubernetes.Interface, namespace, storageClass string) string {
 	pvcStorage, err := resource.ParseQuantity("2Gi")
 	if err != nil {
-		return "", err
+		Fail("Could not parse '2Gi' when creating a stateful set.")
```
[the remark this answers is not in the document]
mociarain: Done

test/e2e/helpers.go
```diff
@@ -0,0 +1,188 @@
...
+	DefaultTimeout  = 10 * time.Minute
+	PollingInterval = 250 * time.Millisecond
+)
+
+type K8sGetFunc[T kruntime.Object] func(ctx context.Context, name string, options metav1.GetOptions) (T, error)
+type K8sListFunc[T kruntime.Object] func(ctx context.Context, options metav1.ListOptions) (T, error)
+type K8sCreateFunc[T kruntime.Object] func(ctx context.Context, object T, options metav1.CreateOptions) (T, error)
+type K8sDeleteFunc func(ctx context.Context, name string, options metav1.DeleteOptions) error
+
+// This function takes a get function like clients.Kubernetes.CertificatesV1().CertificateSigningRequests().Get
```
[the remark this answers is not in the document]
mociarain: I like this.

test/e2e/helpers.go
```diff
@@ -0,0 +1,188 @@
...
+// and the parameters for it. It then makes the call with some retry logic and returns the result after
+// asserting there were no errors.
+//
+// By default the call is retried for 5s with a 250ms interval.
+func GetK8sObjectWithRetry[T kruntime.Object](
+	ctx context.Context, get K8sGetFunc[T], name string, options metav1.GetOptions,
```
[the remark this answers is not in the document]
mociarain: Agreed

AldoFusterTurpin (approved):
LGTM. Thanks!

tsatam (approved):

tsatam:
Marking this PR as high priority - I believe this change will go a long way towards fixing known flakes and currently-broken e2e tests such as the one being debugged in #3384

PR #3368, opened 2024-01-24
Use the right option for the cluster client

nwnt:

Fixes: ARO-3315

Make sure the az identity client use the right option.  At the moment it's set with nil.

### Test plan for issue:
E2E

No

### DIFF
--- a/pkg/validate/openshiftcluster_validatedynamic.go
+++ b/pkg/validate/openshiftcluster_validatedynamic.go
@@ -140,7 +140,7 @@ func (dv *openShiftClusterDynamicValidator) Dynamic(ctx context.Context) error {
 			dv.subscriptionDoc.Subscription.Properties.TenantID,
 			spp.ClientID,
 			string(spp.ClientSecret),
-			nil,
+			dv.env.Environment().ClientSecretCredentialOptions(),
 		)
 		if err != nil {
 			return err


pkg/validate/openshiftcluster_validatedynamic.go
```diff
@@ -140,7 +140,7 @@ func (dv *openShiftClusterDynamicValidator) Dynamic(ctx context.Context) error {
 			dv.subscriptionDoc.Subscription.Properties.TenantID,
 			spp.ClientID,
 			string(spp.ClientSecret),
-			nil,
+			dv.env.Environment().ClientSecretCredentialOptions(),
```
mociarain, line 143: This is used further down in the same function (158). Can we lift the definition higher up in the function and use it multiple times?

mociarain (commented):
Small suggestion on it as it. On a general note I think we could consider breaking this function up into a few parts. It's quite long and essentially there are 2 parts labelled `// X Validation` that could be broken out easily. 

What do you think?

mociarain:
Also does (this)[https://github.com/Azure/ARO-RP/blob/714feb2c1730dd07ce738cb19282ee31ad0ac178/pkg/validate/openshiftcluster_validatedynamic.go#L131] have any bearing for the linked epic?

pkg/validate/openshiftcluster_validatedynamic.go
nwnt: No, this is conditioned so that we only create if checkaccess is turned on.

nwnt:
ARO-2248 is the work item to remove that after all the releases have been completed.  As for the refactoring suggestion, I think it's better to keep or track that as a separate task as this one is aimed to fix the apparent bug.  IMHO, that's going to be a non-trivial task because there's not a single unit test that validates this function.  Probably that's a required step prior to any refactoring efforts.

cadenmarchese:
@nwnt @mociarain as discussed in review today, I think we could handle the refactor in another PR since this is a bugfix.

mociarain:
> @nwnt @mociarain as discussed in review today, I think we could handle the refactor in another PR since this is a bugfix.

Np. Happy to do the refactor myself to get the practice in. Just let me know :)

nwnt:
No longer need this PR as the same fix is applied in #3335.

pkg/validate/openshiftcluster_validatedynamic.go
mociarain: Suggestion was applied [here[(https://github.com/Azure/ARO-RP/pull/3335).

PR #3369, opened 2024-01-24
ARO-4745: Update cluster lifecycle API fields to readOnly

lranjbar:

Fixes ARO-4745

This PR sets a handful of fields in the API to be read only because they don't change once set. Changing the values in these fields would create unexpected functionality for the RP.


pkg/api/v20231122/openshiftcluster.go
```diff
@@ -50,7 +50,7 @@ type OpenShiftClusterProperties struct {
 	ClusterProfile ClusterProfile `json:"clusterProfile,omitempty"`
 
 	// The console profile.
-	ConsoleProfile ConsoleProfile `json:"consoleProfile,omitempty"`
+	ConsoleProfile ConsoleProfile `json:"consoleProfile,omitempty" swagger:"readOnly"`
```
SudoBrendan, line 53: I think instead of making the whole `Profile` readonly, we just want the URL (which, yea, is the only thing in there... but... (see L122)

pkg/api/v20231122/openshiftcluster.go
```diff
@@ -68,7 +68,7 @@ type OpenShiftClusterProperties struct {
 	WorkerProfilesStatus []WorkerProfile `json:"workerProfilesStatus,omitempty" swagger:"readOnly"`
 
 	// The cluster API server profile.
-	APIServerProfile APIServerProfile `json:"apiserverProfile,omitempty"`
+	APIServerProfile APIServerProfile `json:"apiserverProfile,omitempty" swagger:"readOnly"`
```
SudoBrendan, line 71: Same as above, instead of `Profile`, let's do the specific field of the URL (L252)

SudoBrendan (changes requested):

SudoBrendan (dismissed):
LGTM, thanks!

mociarain (dismissed):
LGTM

bennerv (commented):
Can we update the existing field which are readonly in swagger to have the swagger `readOnly` tag?

tsatam:
Follow #3380 for the implementation of this change

PR #3370, opened 2024-01-25
Make the code DRY-er

mociarain:
Just to demonstrate and idea for my understanding


mociarain:
No longer needed as the suggestion was applied [here](https://github.com/Azure/ARO-RP/pull/3335).

PR #3371, opened 2024-01-26
Return error on certificateexpirationstatus monitor when IngressController is invalid rather than panic

tsatam:

Fixes [ARO-5227](https://issues.redhat.com/browse/ARO-5227)

The ARO Monitor's certificateexpirationstatus monitor expects the IngressController resource on-cluster to be well-formed, namely it requires a default certificate name to be present within `.spec.defaultCertificate.name`. 

The Go type definition for IngressController allows the `DefaultCertificate` property to be nil, and we do not currently perform a nil check here.

This causes the monitor service to panic and terminate when evaluating a cluster that has such an invalid IngressController resource. 

### Test plan for issue:

- Unit test added for this specific case
- All existing unit and e2e tests continue to pass

No

### DIFF
--- a/pkg/monitor/cluster/certificateexpirationstatuses.go
+++ b/pkg/monitor/cluster/certificateexpirationstatuses.go
@@ -61,6 +61,11 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 		if err != nil {
 			return err
 		}
+
+		if ic.Spec.DefaultCertificate == nil {
+			return fmt.Errorf("ingresscontroller spec invalid, unable to get default certificate name")
+		}
+
 		ingressSecretName := ic.Spec.DefaultCertificate.Name
 
 		// secret with managed certificates is uuid + "-ingress" or "-apiserver"
--- a/pkg/monitor/cluster/certificateexpirationstatuses_test.go
+++ b/pkg/monitor/cluster/certificateexpirationstatuses_test.go
@@ -42,18 +42,32 @@ func TestEmitCertificateExpirationStatuses(t *testing.T) {
 	//expirationString := expiration.UTC().Format(time.RFC3339)
 	clusterID := "00000000-0000-0000-0000-000000000000"
 
+	defaultIngressController := &operatorv1.IngressController{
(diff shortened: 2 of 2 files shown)


mociarain (commented):
LGTM

petrkotas (approved):
Nice catch. LGTM

pkg/monitor/cluster/certificateexpirationstatuses.go
```diff
@@ -61,6 +61,11 @@ func (mon *Monitor) emitCertificateExpirationStatuses(ctx context.Context) error
 		if err != nil {
 			return err
 		}
+
+		if ic.Spec.DefaultCertificate == nil {
+			return fmt.Errorf("ingresscontroller spec invalid, unable to get default certificate name")
```
AldoFusterTurpin, line 66: Stupid minor question: would it make sense to clearly state that the default certificate is nil (or empty) in the returned error -same in the test- rather than saying "spec invalid" ? (as we need to be more knowledgeable to know what "invalid" means in this context)

I think nil or empty is clearer and straightforward, and would be less confusing in case the spec had a config that is not valid but the DefaultCertificate is not nil (in case if any).

Also, that statement can fail due to _spec_ being empty (and then _DefaultCertificate_ nil for sure) or _spec_ not being empty and _DefaultCertificate_ being nil. Are we happy with one statement for both cases ? (maybe yes)🙂

AldoFusterTurpin (approved):
LGTM. Thanks for this!

Just a minor question, NOT blocking this to get merged.

pkg/monitor/cluster/certificateexpirationstatuses.go
tsatam: Happy to change this error message, I didn't put too much thought into it and somewhat-intentionally left it generic while writing the test in case I discovered other scenarios that could cause a nil pointer dereference on the IngressController. I didn't end up seeing any so we could make this message more specific.

AldoFusterTurpin: Nice. Makes sense, thanks!

cadenmarchese (approved):

PR #3372, opened 2024-01-26
Do not exit NSG monitor when unable to get subnet

tsatam:

Proposed behavioral change for the refactor performed in #3358 

Original Jiras for feature: [ARO-4892](https://issues.redhat.com/browse/ARO-4892) and [ARO-4967](https://issues.redhat.com/browse/ARO-4967)

In the NSG monitor, we currently exit early if we encounter an error when trying to retrieve a subnet. This PR changes the behavior to continue trying to evaluate the subnets and to ignore ones with errors. 

The implementation as-is was causing flaky test failures, as the order of subnet evaluation was not consistent, leading to mocked Subnet GET calls not always occurring as expected. 

### Test plan for issue:

- Existing unit tests continue to pass
  - were also run multiple times to ensure the flake is gone

no

### DIFF
--- a/pkg/monitor/azure/nsg/nsg.go
+++ b/pkg/monitor/azure/nsg/nsg.go
@@ -113,15 +113,23 @@ func (n *NSGMonitor) toSubnetConfig(ctx context.Context, subnetID string) (subne
 func (n *NSGMonitor) Monitor(ctx context.Context) []error {
 	defer n.wg.Done()
 
+	errors := []error{}
+
+	// to make sure each NSG is processed only once
+	nsgSet := map[string]*armnetwork.SecurityGroup{}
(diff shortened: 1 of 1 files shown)


tsatam:
@nwnt I'd like to get your review on this approach and if this makes sense to do within the monitor, or if we should evaluate a different approach to handle this test flake.

nwnt:
@tsatam I think this is definitely better than it currently is.  Thank you very much for the improvement.

nwnt (approved):

mociarain (approved):
LGTM

cadenmarchese (approved):

SudoBrendan (approved):

PR #3376, opened 2024-02-02
Refactor adminUpdate steps

mociarain:

A refactor to lay the groundwork for: https://issues.redhat.com/browse/ARO-5327

I've smashed all the various `if` blocks to give us another view on what this is doing and hopefully, this is clearer. I've preserved the step execution order (mostly) because it's not clear to me what exactly what dependencies are involved.   Functionally, I believe this is identical to what we had before but now there is some extra repeats of certain idempotent steps. 

Note: I noticed that in the `everything` path weren't actually running the `m.renewMDSDCertificate` step. This PR fixes this.

Some questions:

- ~~Is it valid to run this with all flags set to false? Technically, it would have done bootstrap 0 & 1 and step 0 but here I set it to panic.~~
It is valid according to the unit tests.

- ~~Is the hive stuff dependant on `initializeOperatorDeployer`?~~
No.

Related PUCM scipt PR: https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO-Scripts/pullrequest/9515547

### Test plan for issue:

I've tested this and it's using my local RP to prove that:
- [x] The everything step runs everything happily.
- [x] The individual "branches" can be ran successfully.

https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/WIKI-Designs/pullrequest/9515543

### Future Ideas/Thoughts:

1. Don't grow `GeneralFixesSteps`
This is now a kind of dumping ground of random things that we should stop adding to IMO. It's only going to snowball and hopefully what I have done here shows how we could define extra groups and add them with correct context.
1. Break up  `GeneralFixesSteps`
This would involve understanding the dependencies between steps and more understanding than me to group them better conceptually.

### DIFF
--- a/pkg/cluster/adminupdate_test.go
+++ b/pkg/cluster/adminupdate_test.go
@@ -96,7 +96,7 @@ func TestAdminUpdateSteps(t *testing.T) {
 				doc.OpenShiftCluster.Properties.MaintenanceTask = api.MaintenanceTaskOperator
 				return doc, true
 			},
-			shouldRunSteps: concatMultipleSlices(zerothSteps, certificateRenewalSteps, operatorUpdateSteps),
+			shouldRunSteps: concatMultipleSlices(zerothSteps, operatorUpdateSteps),
 		},
 		{
 			name: "ARO Operator Update on <= 4.6 cluster does not update operator",
@@ -107,7 +107,7 @@ func TestAdminUpdateSteps(t *testing.T) {
 				doc.OpenShiftCluster.Properties.ClusterProfile.Version = "4.6.62"
 				return doc, true
 			},
-			shouldRunSteps: concatMultipleSlices(zerothSteps, certificateRenewalSteps),
+			shouldRunSteps: concatMultipleSlices(zerothSteps),
 		},
 		{
 			name: "ARO Operator Update on 4.7.0 cluster does update operator",
@@ -118,7 +118,7 @@ func TestAdminUpdateSteps(t *testing.T) {
 				doc.OpenShiftCluster.Properties.ClusterProfile.Version = "4.7.0"
 				return doc, true
 			},
-			shouldRunSteps: concatMultipleSlices(zerothSteps, certificateRenewalSteps, operatorUpdateSteps),
+			shouldRunSteps: concatMultipleSlices(zerothSteps, operatorUpdateSteps),
 		},
 		{
 			name: "Everything update and adopt Hive.",
--- a/pkg/cluster/adminupdate_test.go
+++ b/pkg/cluster/adminupdate_test.go
@@ -96,7 +96,7 @@ func TestAdminUpdateSteps(t *testing.T) {
 				doc.OpenShiftCluster.Properties.MaintenanceTask = api.MaintenanceTaskOperator
 				return doc, true
 			},
-			shouldRunSteps: concatMultipleSlices(zerothSteps, operatorUpdateSteps),
+			shouldRunSteps: concatMultipleSlices(zerothSteps, certificateRenewalSteps, operatorUpdateSteps),
 		},
 		{
 			name: "ARO Operator Update on <= 4.6 cluster does not update operator",
@@ -107,7 +107,7 @@ func TestAdminUpdateSteps(t *testing.T) {
 				doc.OpenShiftCluster.Properties.ClusterProfile.Version = "4.6.62"
 				return doc, true
 			},
-			shouldRunSteps: zerothSteps,
+			shouldRunSteps: concatMultipleSlices(zerothSteps, certificateRenewalSteps),
 		},
 		{
 			name: "ARO Operator Update on 4.7.0 cluster does update operator",
@@ -118,7 +118,7 @@ func TestAdminUpdateSteps(t *testing.T) {
 				doc.OpenShiftCluster.Properties.ClusterProfile.Version = "4.7.0"
 				return doc, true
 			},
-			shouldRunSteps: concatMultipleSlices(zerothSteps, operatorUpdateSteps),
+			shouldRunSteps: concatMultipleSlices(zerothSteps, certificateRenewalSteps, operatorUpdateSteps),
 		},
 		{
 			name: "Everything update and adopt Hive.",
--- a/pkg/cluster/adminupdate_test.go
+++ b/pkg/cluster/adminupdate_test.go
@@ -265,6 +265,7 @@ func TestAdminUpdateSteps(t *testing.T) {
 				"[Action populateRegistryStorageAccountName-fm]",
 				"[Action ensureMTUSize-fm]",
 				"[Action initializeOperatorDeployer-fm]",
+				"[Action renewMDSDCertificate-fm]",
 				"[Action ensureAROOperator-fm]",
 				"[Condition aroDeploymentReady-fm, timeout 20m0s]",
 				"[Condition ensureAROOperatorRunningDesiredVersion-fm, timeout 5m0s]",
--- a/pkg/cluster/adminupdate_test.go
+++ b/pkg/cluster/adminupdate_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/go-test/deep"
 
 	"github.com/Azure/ARO-RP/pkg/api"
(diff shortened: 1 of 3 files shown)


mociarain:
I've smashed all the various if blocks that I could together to give us another view on what this is doing. 

Some questions:
- Is it valid to run this with all flags set to false? Technically, it would have done `bootstrap 0 & 1 and step 0` but here I set it to panic. 


Some thoughts:
- Let's make the 3 paths independent. We can restructure the `all/everything` steps -> to 3 distinct steps that do:
  - GeneralMaintenance
  - RenewCertificates 
  - UpdateOperator
Initially, we can run these in series with a failure in one causing that "group" to fail and allowing any subsequent groups to carry on

pkg/cluster/install.go
```diff
@@ -46,90 +46,110 @@ func (m *manager) adminUpdate() []steps.Step {
...
+
+		// Block 7
+		steps.Action(m.populateRegistryStorageAccountName),
+		steps.Action(m.ensureMTUSize),
+
+		// Bootstrap 2
+		steps.Action(m.initializeOperatorDeployer),
+
+		// Block 8
+		steps.Action(m.renewMDSDCertificate),
```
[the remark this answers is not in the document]
mociarain: (I think it was a bug that it wasn't in everything)

pkg/cluster/install.go
```diff
@@ -31,9 +31,19 @@ import (
 )
 
 const (
-	operatorCutoffVersion = "4.7.0" // OCP versions older than this will not receive ARO operator updates
+	operatorCutOffVersion = "4.7.0" // OCP versions older than this will not receive ARO operator updates
 )
 
+func concatMultipleSlices[T any](slices ...[]T) []T {
```
mociarain, line 37: Is there a better place for super generic helpers like this?

dkeohane: Maybe somewhere like https://github.com/Azure/ARO-RP/tree/master/pkg/util/stringutils but with a rename to utils, but that nested utils folder makes me sad, so dunno

pkg/cluster/adminupdate_test.go
```diff
@@ -31,9 +31,61 @@ func TestAdminUpdateSteps(t *testing.T) {
 		}
 	}
 
+	zerothSteps := []string{
+		"[Action initializeKubernetesClients-fm]",
+		"[Action ensureBillingRecord-fm]",
+		"[Action ensureDefaults-fm]",
+		"[AuthorizationRetryingAction fixupClusterSPObjectID-fm]",
+		"[Action startVMs-fm]",
+		"[Condition apiServersReady-fm, timeout 30m0s]",
```
mociarain, line 40: These now happen earlier in the ordering.

pkg/cluster/install.go
```diff
@@ -31,9 +31,19 @@ import (
...
+	result := []T{}
```
nwnt, line 38: Nit: if the final size of the result array is known, we could use `make` to preallocate the capacity.  This will avoid unnecessary new slice memory allocations.

pkg/cluster/install.go
nwnt: Looks like a good use of generics here.

pkg/cluster/install.go
mociarain: So I considered something like this but I think I would need to loop over the slices twice. Once, to get the length, then build the array, then loop over the slices again to populate the array. My heuristic is that more looping is bad and dynamically adding to a slice would be better.

1. Did you have a different implementation in mind?
2. Is my heuristic just plain wrong?

pkg/cluster/install.go
```diff
@@ -57,109 +94,73 @@ func (m *manager) adminUpdate() []steps.Step {
...
-		toRun = append(toRun,
-			steps.Action(m.ensureResourceGroup), // re-create RP RBAC if needed after tenant migration
-			steps.Action(m.createOrUpdateDenyAssignment),
-			steps.Action(m.ensureServiceEndpoints),
-			steps.Action(m.populateRegistryStorageAccountName), // must go before migrateStorageAccounts
-			steps.Action(m.migrateStorageAccounts),
-			steps.Action(m.fixSSH),
-			// steps.Action(m.removePrivateDNSZone), // TODO(mj): re-enable once we communicate this out
-		)
+	// Generic fix-up actions that are fairly safe to always take, and don't require a running cluster
```
tiguelu, line 101: Curious about why the `fixInfraID` action is outside the list above. The actions in that list also comply with the description of the comment for this action (in the old code, the comment was actually for all of them).

pkg/cluster/install.go
```diff
@@ -57,109 +94,73 @@ func (m *manager) adminUpdate() []steps.Step {
...
-	if isEverything {
-		toRun = append(toRun,
-			steps.Action(m.fixSREKubeconfig),
-			steps.Action(m.fixUserAdminKubeconfig),
-			steps.Action(m.createOrUpdateRouterIPFromCluster),
-		)
-	}
+		steps.Action(m.fixSREKubeconfig),
+		steps.Action(m.fixUserAdminKubeconfig),
+		steps.Action(m.createOrUpdateRouterIPFromCluster),
```
tiguelu, line 121: The action`populateDatabaseIntIP` that populates the internal LB IP for the API into the DB was run before these 3 actions in the old code. I wonder if there was any requirement for that, since these 3 steps will require the API to be up.

Same wonder with the `apiServersReady` condition after starting the VMs which is now in `getZerothSteps` function before the DB IP fix runs, but used to be after.

pkg/cluster/install.go
```diff
@@ -46,9 +56,36 @@ func (m *manager) adminUpdate() []steps.Step {
...
+		}
+	} else if isRenewCerts {
+		stepsToRun = append(stepsToRun, m.getCertificateRenewalSteps()...)
+	}
+
+	// We don't run this on an operator-only deploy as PUCM scripts then cannot
+	// determine if the cluster has been fully admin-updated
+	// Run this last so we capture the resource provider only once the upgrade has been fully performed
+	if isEverything {
+		stepsToRun = append(stepsToRun, steps.Action(m.updateProvisionedBy))
```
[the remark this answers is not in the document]
mociarain: I think so. Done

pkg/cluster/install.go
mociarain: My thinking was that before there were several generic tasks that happened for every path. These were all clearly bootstrapping tasks apart from this one which is actually changing the state of the cluster.

pkg/cluster/install.go
mociarain: The reason I moved this down is because `populateDatabaseIntIP` patches  `doc.OpenShiftCluster.Properties.APIServerProfile.IntIP`. This property in only 2 other tasks: 
- [fixMCSCert](https://github.com/mociarain/ARO-RP/blob/60758295008fbd905dab0b3e485996d8e4024a9b/pkg/cluster/fixmcscert.go#L27)
- [fixMCSUserData](https://github.com/mociarain/ARO-RP/blob/60758295008fbd905dab0b3e485996d8e4024a9b/pkg/cluster/fixmcsuserdata.go#L176)


For the `startVMs` task I moved that back into bootstrapping because it seems to fit there better conceptually and I'm assuming the tasks that come before it in the if blocks can't be pre-requisites because we start the VMs in each path but only run some tasks before we start the VMs in `isEverything`.

pkg/cluster/install.go
jaitaiwan: Golang 1.22 includes the new `slices.Concat` method which would replace this function. I recommend we include the [implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/slices/slices.go;l=502) ourselves for now. Nothing stands out immediately as something that can't be done in Go 1.20.

As far as function placement is concerned, I'd recommend `pkg/util` maybe under a folder like `generics` or `helpers` although that can end up being a bucket for everything.

jaitaiwan (changes requested):
Mostly LGTM, just one point on the slice concat. Noting that I think it should be a public function.

mociarain:
Closing in favour of: https://github.com/Azure/ARO-RP/pull/3449

PR #3377, opened 2024-02-05
Revendor hive to 5fbe0d158ba9

s-fairchild:

[ARO-5059](https://issues.redhat.com/browse/ARO-5059)

1. Hive needs to be vendored at the same commit level as it is deployed in ARO. One reason, as described in the linked card, is that changes in APIs can lead to unintended edits during round-trip Get()/Update() flows. 
1. Increase verbosity for `golangci-lint` to improve general debug ability
1. Change `golangci-lint` output format to be github-actions format 

Updating [Hive Dependency Update Wiki](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/366470/Hive-Deploy#)


hack/lint-go.sh
```diff
@@ -14,5 +14,5 @@ then
     echo "ERROR: golangci-lint could not be found, Go linting aborted. To install it visit https://golangci-lint.run/usage/install/#local-installation"
     exit 1
 else
-    golangci-lint run
+    golangci-lint run --verbose
```
mociarain, line 17: Just to check my understanding: Is this change and the change in `.golangci.yml` in fact separate to the main thrust of this PR and actually just a nice general improvement for us?

Not opposed to making things more verbose but just checking this is intentional as I didn't expect it based on the PR description.

s-fairchild: @mociarain Yes, the changes in `hack/lint-go.sh` and `golangci.yml` are simply general improvements that don't have to be associated with this PR.  Upping the verbosity helped me troubleshoot my install of `golangci-lint` and may help us in future pipeline run troubleshooting.

mociarain (approved):
LGTM. (Absolute pain we need to do this ever few weeks :D Thanks for handling it)

go.mod
```diff
@@ -59,7 +59,7 @@ require (
 	github.com/openshift/client-go v0.0.0-20220525160904-9e1acff93e4a
 	github.com/openshift/cloud-credential-operator v0.0.0-00010101000000-000000000000
 	github.com/openshift/console-operator v0.0.0-20220407014945-45d37e70e0c2
-	github.com/openshift/hive/apis v0.0.0-20231116161336-9dd47f8bfa1f
+	github.com/openshift/hive/apis v0.0.0-20240131003045-5fbe0d158ba9
```
cadenmarchese, line 62: For next time, we should have a more clear step-by-step on this part of the doc:
```
Revendor Hive in the RP to match the version that we intend to upgrade.
Example PR can be found https://github.com/Azure/ARO-RP/pull/3119 
```
Is this the only field that gets updated for the revendor?

cadenmarchese (changes requested):
I'm good to hit approve when docs are updated!

go.mod
s-fairchild: Yes, that example shows updates to kube-api which I think may have been required when updating hive then.

[Wiki Page Updated](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/366470/Hive-Deploy#)

cadenmarchese (approved):

PR #3380, opened 2024-02-05
Update cluster lifecycle API fields to readOnly (different approach) 

tsatam:

Different attempt at fixing [ARO-4745]. 

This PR sets a handful of fields in the API's OpenShiftCluster to be read only because they don't change once set. Changing the values in these fields would create unexpected functionality for the RP. 

Fields changed to read-only:
- `.properties.consoleProfile.url`
- `.properties.apiserverProfile.url`
- `.systemData`

This PR differs from #3369 in that we set the entire `systemData` top level property on the resource to read-only. This will prevent any modifications to all fields on this subresource by clients performing any PUT/POST operations. 

### Test plan for issue:

- Existing unit tests and E2E continue to pass
- Static validation tests have been updated to account for these fields becoming read only

No

### Additional notes

This PR seeks to solve the same problem as #3369. Only one of these PRs should be merged.


SudoBrendan (approved):

Shivkumar13 (approved):
LGTM

tony-schndr (approved):
LGTM

PR #3388, opened 2024-02-07
Add explicit step to wait for job completion in ARO cluster DNS e2e test

tsatam:

Fixes [ARO-5379](https://issues.redhat.com/browse/ARO-5379)

Adds an explicit step to the `ARO cluster DNS - must not be adversely affected by Azure host servicing` E2E step to wait for completion on the `resolv.conf` check Job that is deployed on-cluster. 

### Test plan for issue:

- E2E stops flaking 

No.

### Additional notes

This PR has a hard dependency on #3366, and currently includes the changes from that PR. This PR should be rebased on `master` once #3366 is merged, and should only include the last commit.


test/e2e/cluster.go
```diff
@@ -27,36 +27,31 @@ import (
...
 		testNamespace := fmt.Sprintf("test-e2e-%d", GinkgoParallelProcess())
-		p = project.NewProject(clients.Kubernetes, clients.Project, testNamespace)
-		err := p.Create(ctx)
-		Expect(err).NotTo(HaveOccurred(), "Failed to create test namespace")
+		project = BuildNewProject(ctx, clients.Kubernetes, clients.Project, testNamespace)
 
 		By("verifying the namespace is ready")
 		Eventually(func(ctx context.Context) error {
-			return p.Verify(ctx)
+			return project.VerifyProjectIsReady(ctx)
```
cadenmarchese, line 46: Do we need to keep the `Eventually` in the test itself wherever `VerifyProjectIsReady()` is used? I ask because there is already an `Eventually` with timeout in the `CreateK8sObjectWithRetry` function, which `VerifyProjectIsReady()` calls.

cadenmarchese: Resolving as I now understand this commit to be part of https://github.com/Azure/ARO-RP/pull/3366 :)

mociarain (approved):
Nice

mociarain:
> This PR is only necessary to merge if https://github.com/Azure/ARO-RP/pull/3366 on its own does not fix the flakiness in this E2E test.

IMO this is addressing something else that the linked PR doesn't. We have similar logic other tests for a pod to come up and ensure it's in a succeed state before continuing.

jaitaiwan (approved):
LGTM

kimorris27 (approved):
IMO this can and should merge even if the other PR fixes the flakes. This change makes the test even more robust and won't cause any harm.

cadenmarchese (approved):
thanks for the enhancement!

PR #3389, opened 2024-02-07
refactor package pki to be unit testeable

AldoFusterTurpin:

Makes the package pki unit testeable decoupling the HTTP calls with the business logic. It also removes the unused state that was maintained just for the sake of the previous unit test (technically not a unit test as was touching HTPP and was flaky).

We still need to decide what needs to be tested and if the unit test that we will create is enough.

The main idea is to change from: 
- function which performs HTTP call + business logic 
to 
- a public function that makes the HTTP call + another function that is unit testeable that uses the previous result.

I think caMap, mu, getCaCertPoolFromMap and setCaCertPoolInMap can be deleted.

**Slack [thread](https://redhat-internal.slack.com/archives/C02ULBRS68M/p1707324423230289) with context.**

I have also refactored _newPipeline_ that now calls _getClientOptions_ which contains the logic to build the clientOptions in one function as IMHO the current usage of the initialised cp to nil is not straightforward to reason about. i,e: we are calling _NewClientOptions(nil)_ implicitly when options == nil and issuerUrlTemplate == "" || caName == "", something that now is explicit and clear.

I have also introduced an early return in the functions to make things a bit clearer.

We need to stop having a flaky hidden integration test as part of our regular unit tests. Unit tests should never touch the external world (I/O, netowrking, etc.). This PR also removes the unnecessary global state that was created just to be used in the test of the HTTP Get calls.

### Test plan for issue:

I have removed the "fake" unit test that existed in pkg/util/pki/pki.go because it was calling the HTTP endpoint and checking that the result was not nil. It was not even testing explicitly any result of business logic after performing the HTTP calls, so I think at this point it does not bring much value. On top of that, we are most of the time skipping this test in the CI (using the env var) because it is almost always failing. 

One last note: I have refactored the code to be unit-testable. If we really want to test the objects without the external HTTP calls we can do that in the near future, but a different kind of test would be more valuable.

N/A


pkg/util/azureclient/applens/applens_client.go
```diff
@@ -61,16 +61,20 @@ func NewClient(endpoint, issuerUrlTemplate, caName, scope string, cred azcore.To
 }
 
 func newPipeline(authPolicy []policy.Policy, options *ClientOptions, issuerUrlTemplate, caName string) (*runtime.Pipeline, error) {
-	var cp *x509.CertPool = nil
```
AldoFusterTurpin, line 64: Move it to a narrow scope + delete the "nil" literal as it is already implicit in the declaration.

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -61,16 +61,20 @@ func NewClient(endpoint, issuerUrlTemplate, caName, scope string, cred azcore.To
...
-	var err error = nil
```
AldoFusterTurpin, line 65: Can be removed.

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -61,16 +61,20 @@ func NewClient(endpoint, issuerUrlTemplate, caName, scope string, cred azcore.To
...
 	if options == nil {
+		var cp *x509.CertPool
 		// if provided pki info fetch the correct cert pool
 		// otherwise use the default of nil
 		if issuerUrlTemplate != "" && caName != "" {
-			cp, err = pki.GetTlsCertPool(issuerUrlTemplate, caName)
+			data, err := pki.FetchDataFromGetIssuerPki(issuerUrlTemplate, caName)
```
AldoFusterTurpin, line 69: Explicit HTTP call.

pkg/util/azureclient/applens/applens_client.go
```diff
@@ -61,16 +61,20 @@ func NewClient(endpoint, issuerUrlTemplate, caName, scope string, cred azcore.To
...
 			if err != nil {
 				return nil, err
 			}
+
+			cp, err := pki.BuildCertPoolForCaName(data)
+			if err != nil || cp == nil {
```
AldoFusterTurpin, line 75: I think the _cp == nil_ part is redundant and can be deleted.

mociarain: In the current implementation this is definitely true but more generally I would expect a builder that is unable to build -> invalid state so it should never give a `nil, nil`.

AldoFusterTurpin: Changed the approach to just return the NewClientOptions from inside the function as we just talked about 🙂

pkg/util/pki/fetchdata.go
```diff
@@ -0,0 +1,35 @@
+package pki
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
```
ehvs, line 9: Using go 1.21 this package was deprecated and fails the golangci-lint test. Changing to `io` only, works.
```io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.  (staticcheck)```

AldoFusterTurpin: Thank you. Done.

AldoFusterTurpin:
Closing in favor of the one in upstream: https://github.com/Azure/ARO-RP/pull/3708

PR #3393, opened 2024-02-08
Use ginkgo helper

mociarain:

This is a small PR to improve the readability of E2E test output by using the GinkgoHelper(). I noticed this feature while doing https://issues.redhat.com/browse/ARO-5321. The details are [here](https://onsi.github.io/ginkgo/#the-offset-decorator) but the tl;dr is we get more informative/clearer stack traces if there is an error in the helpers.

To get access to the GinkgoHelper() I needed to bump Ginkgo to 2.8.0 (currently at 2.7.0) and while I was at it I went all the way to 2.14.0.

Note: 2.15.0 is out but requires Go 1.20.0.

### Test plan for issue:

The current E2E tests.

No. I think having this in the existing helpers is enough documentation.


SudoBrendan (dismissed):

AldoFusterTurpin (approved):
LGTM!

cadenmarchese:
@mociarain I think you'll need to adjust your target branch to `master` to get e2e going on this. I assume the intention is to merge to master?

mociarain:
> @mociarain I think you'll need to adjust your target branch to `master` to get e2e going on this. I assume the intention is to merge to master?

Nope. I wanted to try a workflow of breaking PR's in smaller pieces to help review but with our current Git setup I'm not sure it'll actually work. Contex [here](https://redhat-internal.slack.com/archives/C02ULBRS68M/p1707302710755289) if you're curious.

cadenmarchese:
@mociarain two cents - I think breaking up large changes into smaller PRs is always preferable, but I'd target `master` in both PRs to keep the convention that we're used to and logically separate the enhancements. It seems that the two PRs (https://github.com/Azure/ARO-RP/pull/3393 and https://github.com/Azure/ARO-RP/pull/3392) are related but are solving different problems, and approvals from this PR won't carry over to the other, so it's duplicated review and merge work in a way. I'd personally be good to LGTM this.

PR #3395, opened 2024-02-08
Update `az aro` permissions validation to mirror RP frontend validation

tsatam:

Fixes [ARO-5075](https://issues.redhat.com/browse/ARO-5075)

Updates the dynamic validation performed by the Azure CLI `az aro` command to mirror the validation performed by the RP itself. This also fixes incorrect behavior in the CLI handling of the `not_actions` list on permissions, which were previously being treated as full deny permissions, rather than negations on only the sibling `actions` list of its parent permission. 

Note that this will change the error reported to users - previously users would see a `${ACTION} permission is disabled` error message, which is untrue (since permissions cannot be "disabled"). The only error message from this validation will be `${ACTION} permission is missing`. 

### Test plan for issue:

Unit tests were added for the dynamic validator's `can_do_action` function. These tests mirror the equivalent tests for the RP's validation. 

No

### Review breadcrumbs

Code to compare against:
- [RP implementation](https://github.com/Azure/ARO-RP/blob/master/pkg/util/permissions/permissions.go)
- [RP tests](https://github.com/Azure/ARO-RP/blob/master/pkg/util/permissions/permissions_test.go)

### DIFF
--- a/python/az/aro/azext_aro/_dynamic_validators.py
+++ b/python/az/aro/azext_aro/_dynamic_validators.py
@@ -27,16 +27,27 @@ log_entry_type = {'warn': 'Warning', 'error': 'Error'}
 
 def can_do_action(perms, action):
     for perm in perms:
-        for not_action in perm.not_actions:
-            match = re.escape(not_action)
-            match = re.match("(?i)^" + match.replace(r"\*", ".*") + "$", action)
-            if match:
-                return f"{action} permission is disabled"
+        matched = False
+
         for perm_action in perm.actions:
             match = re.escape(perm_action)
             match = re.match("(?i)^" + match.replace(r"\*", ".*") + "$", action)
             if match:
-                return None
+                matched = True
+                break
+
+        if not matched:
+            continue
+
+        for not_action in perm.not_actions:
+            match = re.escape(not_action)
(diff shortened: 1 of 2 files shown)


mociarain (changes requested):
So functionally I agree that this works but it's very odd to me that `can_do_action` is returning essentially an `Optional[ErrorString]` and not a boolean.

IMO this would be clearer and more idiomatic if it returned a boolean and then we added the error message up a level in the `validate_resource` on the basis of that. 

What do you think?

mociarain:
> given that we're actively investigating moving _all_ of this validation logic into the RP and having the `az aro validate` command that uses this validation to just call an API endpoint instead.

Oh in that case then I get it. Thanks for the explication and :ship:  ;)

mociarain (approved):

tsatam:
> > given that we're actively investigating moving _all_ of this validation logic into the RP and having the `az aro validate` command that uses this validation to just call an API endpoint instead.
> 
> Oh in that case then I get it. Thanks for the explication and 🚢 ;)

I changed my mind, I pushed up this refactor on a new commit.

mociarain (approved):
Double approved ;)

SudoBrendan (approved):

s-amann (approved):
:shipit:

PR #3396, opened 2024-02-09
Adding segregation of resultType for CloudError

sankur-codes:

Fixes [ARO-5112](https://issues.redhat.com/browse/ARO-5112)

As of now, all CloudErrors are termed as UserError.  This PR will check upon the statusCode of the error and provide resultType accordingly. 

### Test plan for issue:

Done : Manually Tested mapping of CloudError `StatusCode` with `ResultType`.
Also, included test coverage for the functions

No

### DIFF
--- a/pkg/backend/openshiftcluster.go
+++ b/pkg/backend/openshiftcluster.go
@@ -347,7 +347,10 @@ func (ocb *openShiftClusterBackend) asyncOperationResultLog(log *logrus.Entry, i
 
 	err, ok := backendErr.(*api.CloudError)
 	if ok {
-		if int(err.StatusCode) < 500 {
+		if err.StatusCode < 300 && err.StatusCode >= 200 {
+			log.Info("long running operation succeeded")
+			return
+		} else if err.StatusCode < 500 {
 			log = log.WithField("resultType", utillog.UserErrorResultType)
 		} else {
 			log = log.WithField("resultType", utillog.ServerErrorResultType)
--- a/pkg/backend/openshiftcluster.go
+++ b/pkg/backend/openshiftcluster.go
@@ -345,9 +345,13 @@ func (ocb *openShiftClusterBackend) asyncOperationResultLog(log *logrus.Entry, i
 			"properties.servicePrincipalProfile", "The Azure Red Hat Openshift resource provider service principal has been removed from your tenant. To restore, please unregister and then re-register the Azure Red Hat OpenShift resource provider.")
 	}
 
-	_, ok := backendErr.(*api.CloudError)
+	err, ok := backendErr.(*api.CloudError)
 	if ok {
-		log = log.WithField("resultType", utillog.UserErrorResultType)
+		if int(err.StatusCode) < 500 {
+			log = log.WithField("resultType", utillog.UserErrorResultType)
+		} else {
+			log = log.WithField("resultType", utillog.ServerErrorResultType)
+		}
 	} else {
 		log = log.WithField("resultType", utillog.ServerErrorResultType)
 	}
--- a/pkg/frontend/frontend.go
+++ b/pkg/frontend/frontend.go
@@ -475,6 +475,14 @@ func frontendOperationResultLog(log *logrus.Entry, method string, err error) {
 
 	switch err := err.(type) {
 	case *api.CloudError:
+		if int(err.StatusCode) < 300 && int(err.StatusCode) >= 200 {
+			log.Info("front end operation succeeded")
+			return
+		} else if int(err.StatusCode) < 500 {
+			log = log.WithField("resultType", utillog.UserErrorResultType)
+		} else {
+			log = log.WithField("resultType", utillog.ServerErrorResultType)
+		}
 		log = log.WithField("resultType", utillog.UserErrorResultType)
 	case statusCodeError:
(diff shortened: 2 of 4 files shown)


pkg/frontend/frontend.go
```diff
@@ -475,6 +475,14 @@ func frontendOperationResultLog(log *logrus.Entry, method string, err error) {
 
 	switch err := err.(type) {
 	case *api.CloudError:
+		if int(err.StatusCode) < 300 && int(err.StatusCode) >= 200 {
```
mociarain, line 478: Can we really have a "successful" CloudError?

mociarain (commented):
Looks good but I have a small suggestion. Would we consider putting this into the `CloudError` itself?

We could have a:
```
func (err *CloudError) GetResultType() ResultType {...}
```

This way the logic of mapping stausCode to ResultType lives in one place i.e. the thing that holds the statusCode.

pkg/frontend/frontend.go
sankur-codes: Considering the condition, there could not be a "real" error but a CloudError that has a statusCode between 200 & 300. 
Eg : https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/admin_hive_clusterdeployment_get.go#L45

mociarain: An error with a 204 seems like a bug to me and IMO we should make that impossible. Definitely not for this PR though. 

Also after a quick grep I think you might have the only instance of the a CloudError with a non 4XX/5XX status code in the whole RP so maybe the change I'm implying wouldn't be so painful. If you have sometime maybe we could pair on this as a subsequent task? I don't think I've met any of the APAC folks yet.

sankur-codes:
> Looks good but I have a small suggestion. Would we consider putting this into the `CloudError` itself?
> 
> We could have a:
> 
> ```
> func (err *CloudError) GetResultType() ResultType {...}
> ```
> 
> This way the logic of mapping stausCode to ResultType lives in one place i.e. the thing that holds the statusCode.

sankur-codes:
> Looks good but I have a small suggestion. Would we consider putting this into the `CloudError` itself?
> 
> We could have a:
> 
> ```
> func (err *CloudError) GetResultType() ResultType {...}
> ```
> 
> This way the logic of mapping stausCode to ResultType lives in one place i.e. the thing that holds the statusCode.

Putting a high level view on the suggestion: To include `ResultType` as part of or use it in ClousError, `github.com/Azure/ARO-RP/pkg/util/log` will have to be imported in `pkg/api/error.go` which might create a circular dependency leading to import cycle. 
Another option would be to have `ResultType` completely under CloudError and defined in `pkg/api/error.go` itself but that might not be totally accurate as `ResultType` is meant to be part of **logging** & thus placed in `pkg/util/log`.
I might be missing on something, do help me with further suggestions.

mociarain:
> Putting a high level view on the suggestion: To include `ResultType` as part of or use it in ClousError, `github.com/Azure/ARO-RP/pkg/util/log` will have to be imported in `pkg/api/error.go` which might create a circular dependency leading to import cycle. Another option would be to have `ResultType` completely under CloudError and defined in `pkg/api/error.go` itself but that might not be totally accurate as `ResultType` is meant to be part of **logging** & thus placed in `pkg/util/log`. I might be missing on something, do help me with further suggestions.

No stress. We could put something in `log` that maps `statusCode int` to `ResultType` instead. IMO so long as this logic is in one place I think that's a win. We can do this as part of the pairing on Friday actually (or as a separate PR) since it's gonna branch out beyond this PR

mociarain (approved):

pkg/backend/openshiftcluster.go
```diff
@@ -345,9 +345,16 @@ func (ocb *openShiftClusterBackend) asyncOperationResultLog(log *logrus.Entry, i
...
+			log.Info("long running operation succeeded")
+			return
+		} else if err.StatusCode < 500 {
+			log = log.WithField("resultType", utillog.UserErrorResultType)
+		} else {
+			log = log.WithField("resultType", utillog.ServerErrorResultType)
+		}
 	} else {
 		log = log.WithField("resultType", utillog.ServerErrorResultType)
 	}
```
jaitaiwan, line 360: This conditional could be simplified down.

jaitaiwan (approved):
LGTM small nit - the if statement seems unnecessarily complex

pkg/backend/openshiftcluster.go
sankur-codes: Based on the discussion above, the `if` statement might go away entirely once we discuss and check on the only instance of `CloudError` with 3XX `StatusCode`.

SrinivasAtmakuri (dismissed):

SrinivasAtmakuri:
Can we please add a few test cases coverage here? Thank you.

jaitaiwan (changes requested):
A good point by Srini. Lets add some tests to go with this PR

kartikaysharma27:
LGTM

mociarain (approved):
LGTM

SrinivasAtmakuri (approved):
LGTM.

jaitaiwan (approved):
LGTM

PR #3397, opened 2024-02-09
Add CI containerfile for portal javascript build

SudoBrendan:

Fixes https://issues.redhat.com/browse/ARO-5089

In short, this PR creates a new `make` target that takes all these diverging bits and pieces of build code related to the SRE portal and 1.) makes `podman` the only build dependency 2.) consolidates them into a single location that's easier to maintain 3.) guarantees that you don't skip any steps locally or in CI, and 4.) improves build time performance through caching:

- [portal lint (ci)](https://github.com/Azure/ARO-RP/blob/master/.pipelines/ci.yml#L107-L115)
    - [make lint-admin-portal](https://github.com/Azure/ARO-RP/blob/6ed5f52e0f264529734e436e11d9c096876a8149/Makefile#L232)
        - [Dockerfile for portal linting](https://github.com/Azure/ARO-RP/blob/6ed5f52e0f264529734e436e11d9c096876a8149/Dockerfile.portal_lint#L1-L9)
- [portal audit (ci)](https://github.com/Azure/ARO-RP/blob/6ed5f52e0f264529734e436e11d9c096876a8149/.github/workflows/npm-audit.yml#L15-L29)
    - [portal audit shell script](https://github.com/Azure/ARO-RP/blob/master/hack/github-actions/npm_audit.sh)
- [portal build (ci)](https://github.com/Azure/ARO-RP/blob/6ed5f52e0f264529734e436e11d9c096876a8149/.github/workflows/npm-audit.yml#L31-L48)
- [make build-portal](https://github.com/Azure/ARO-RP/blob/6ed5f52e0f264529734e436e11d9c096876a8149/Makefile#L162)

You can try it yourself locally:

```
# full build/etc
make ci-portal
```

This change supports adding a new variable to your `./env` (`NO_CACHE=false`) to optimize local development.

### Test plan for issue:

This is artifact generation, not a new feature. I did validate that breaking the audit and/or build does exit non-zero on the make target.

I updated `env.example` to showcase the new Makefile variable, and set it to a reasonable default (to leverage podman caching while building locally).

This change will impact dev flows more eventually. For now, this effort will run in parallel to the present-day CI. Subsequent PRs will tie this all together in pipelines and dev flows.

### DIFF
--- a/Dockerfile.ci-portal
+++ b/Dockerfile.ci-portal
@@ -1,0 +1,13 @@
+ARG REGISTRY
+
+FROM ${REGISTRY}/ubi8/nodejs-16 AS portal-base
+WORKDIR /build/portal/v2
+USER root
+
+FROM portal-base AS portal-build
+COPY /portal/v2/package*.json ./
+RUN npm ci
+RUN npm audit --omit=dev
+COPY /portal/v2/ ./
+RUN npm run lint
+RUN npm run build
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ COMMIT = $(shell git rev-parse --short=7 HEAD)$(shell [[ $$(git status --porcela
 ARO_IMAGE_BASE = ${RP_IMAGE_ACR}.azurecr.io/aro
 E2E_FLAGS ?= -test.v --ginkgo.v --ginkgo.timeout 180m --ginkgo.flake-attempts=2 --ginkgo.junit-report=e2e-report.xml
 GO_FLAGS ?= -tags=containers_image_openpgp,exclude_graphdriver_btrfs,exclude_graphdriver_devicemapper
+NO_CACHE ?= true
 
 export GOFLAGS=$(GO_FLAGS)
 
@@ -74,6 +75,9 @@ clean:
 
 client: generate
 	hack/build-client.sh "${AUTOREST_IMAGE}" 2020-04-30 2021-09-01-preview 2022-04-01 2022-09-04 2023-04-01 2023-07-01-preview 2023-09-04 2023-11-22 2024-08-12-preview
+
+ci-portal:
+	docker build . -f Dockerfile.ci-portal --build-arg REGISTRY=$(REGISTRY) --no-cache=$(NO_CACHE)
 
 # TODO: hard coding dev-config.yaml is clunky; it is also probably convenient to
 # override COMMIT.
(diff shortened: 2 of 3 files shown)


SudoBrendan:
As a point of discussion that's relevant, what is our plan for `/portal/v1`? It's been "dead" for a while, so I didn't bother adding it here. Is it safe to say we can just throw that out (as another item of work?).

Dockerfile.ci-portal
```diff
@@ -0,0 +1,17 @@
+ARG REGISTRY
+
+FROM ${REGISTRY}/ubi8/nodejs-16 AS portal-base
+WORKDIR /build
+USER root
+
+FROM portal-base AS portal-build
+COPY /portal/v2 /build/
+RUN npm ci
```
tsatam, line 9: I think `npm ci` will be a prerequisite for lint and audit as well, and would need to be moved into a base step. 

EDIT: wait I see this whole build is a prereq for the other two. I don't think that is strictly necessary, lint and audit should run against the code itself and do not require any built artifacts, and the full app build is a generally more expensive/slower step to run than the lint/audit steps will be.

tsatam (approved):
I think this is functional and fine to merge as is, but there's an opportunity to slim down some steps to avoid unnecessarily slow JS build times on the lint and audit steps.

Dockerfile.ci-portal
SudoBrendan: We discussed this offline - I've pushed a change that makes the local container build cache the way to optimize this away. In a local env, devs will be able to set `NO_CACHE=false` to get quicker builds.

tsatam (approved):

mociarain:
> As a point of discussion that's relevant, what is our plan for `/portal/v1`? It's been "dead" for a while, so I didn't bother adding it here. Is it safe to say we can just throw that out (as another item of work?).

I was poking around with portal a few weeks ago and wondered the same. Looks like it might still technically work (it basically just uses `/api/clusters`) but we do actually [build both](https://github.com/azure/ARO-RP/blob/ed94c2834683bb033254631da313a3f50ceefc85/Makefile#L163). 

I had a look in the dashboards and could find a metric for the portal FE version so not sure how we could fully rule it out.

Makefile
```diff
@@ -75,6 +76,9 @@ clean:
 client: generate
 	hack/build-client.sh "${AUTOREST_IMAGE}" 2020-04-30 2021-09-01-preview 2022-04-01 2022-09-04 2023-04-01 2023-07-01-preview 2023-09-04 2023-11-22 2024-08-12-preview
 
+ci-portal:
+	docker build . -f Dockerfile.ci-portal --build-arg REGISTRY=$(REGISTRY) --no-cache=$(NO_CACHE)
```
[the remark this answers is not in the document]
SudoBrendan: Yea... I honestly don't know. I'll start a slack thread. I'm +1 for it's removal, I think we're way past the point of no return on it but maybe others feel differently?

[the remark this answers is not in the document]
SudoBrendan: ref https://github.com/Azure/ARO-RP/pull/3445

Dockerfile.ci-portal
```diff
@@ -0,0 +1,13 @@
...
+WORKDIR /build/portal/v2
+USER root
+
+FROM portal-base AS portal-build
+COPY /portal/v2/package*.json ./
+RUN npm ci
+RUN npm audit --omit=dev
+COPY /portal/v2/ ./
+RUN npm run lint
+RUN npm run build
```
kimorris27, line 13: I'm not too familiar with JS and npm, but I was tracing through the build process using the links you provided in the PR description, and I noticed that the ["old" build command](https://github.com/Azure/ARO-RP/blob/f20df8471736ba501a9bf86a660451e8dd26a5ac/Makefile#L162) includes an `npm install` that's not in this Containerfile. Why'd you remove that?

SudoBrendan: `npm install` is a near-duplicate (dev version) of `npm ci`, which performs all the actions of `npm install` but without any caching. Honestly, considering how the containerfile is written (only copy `package.json` files, then install deps) there wouldn't be a cache to pull from anyways... but `npm ci` is a more robust version of `npm install` in any case, so I chose it instead.

kimorris27 (approved):

s-amann (approved):

PR #3401, opened 2024-02-12
 Don't add security context on <4.11 as OpenShift restricted SCCs do not tolerate it

bennerv:

In <4.11, the SCCs don't align with the secCompProfile set on the pods, which result in the pods with the added `secCompProfile` failing to schedule / run.  

### Test plan for issue:

Unit tests + manual testing of 4.10 version

Testing instructions:
1. Create a 4.10 cluster at this aro operator version + RP version
2. Make sure MUO, aro-operator (worker and master), and geneva logging pods are working as intended.
3. Check the presence of: aro-worker and aro-master operators (secCompProfile), geneva logging namespace (should not have labels), and MUO deployment (secCompProfile)
4. Upgrade to 4.11
5. AdminUpdate the cluster
6. All the items in step 3 should have the appropriate things now

**Changes**
- Namespace: 
```
	"pod-security.kubernetes.io/enforce": "privileged",
	"pod-security.kubernetes.io/audit":   "privileged",
	"pod-security.kubernetes.io/warn":    "privileged",
```
- Pods: under `.spec.template.spec`
```
      securityContext:
        runAsNonRoot: true
        seccompProfile:
          type: RuntimeDefault
```

- Pods: under `.spec.template.spec.containers[]`
```
          securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop:
              - ALL
            runAsNonRoot: true
```

nope

### DIFF
--- a/pkg/operator/helpers.go
+++ b/pkg/operator/helpers.go
@@ -4,9 +4,40 @@ package operator
 // Licensed under the Apache License 2.0.
 
 import (
+	"context"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+	"github.com/Azure/ARO-RP/pkg/util/version"
 )
+
+var clusterVersionForPodSecurityStandard = version.NewVersion(4, 11)
 
 func GatewayEnabled(cluster *arov1alpha1.Cluster) bool {
 	return len(cluster.Spec.GatewayDomains) > 0
+}
+
+// ShouldUsePodSecurityStandard is an admissions controller
+// for pods which replaces pod security policies, enabled on
+// OpenShift 4.11 and up
+func ShouldUsePodSecurityStandard(ctx context.Context, client client.Client) (bool, error) {
+	cv := &configv1.ClusterVersion{}
+
+	err := client.Get(ctx, types.NamespacedName{Name: "version"}, cv)
+	if err != nil {
+		return false, err
+	}
+	vers, err := version.GetClusterVersion(cv)
+	if err != nil {
+		return false, err
+	}
+
+	if vers.Lt(clusterVersionForPodSecurityStandard) {
+		return false, nil
+	}
+
+	return true, nil
 }
--- a/pkg/operator/helpers.go
+++ b/pkg/operator/helpers.go
@@ -30,6 +30,7 @@ func ShouldUsePodSecurityStandard(ctx context.Context, client client.Client) (bo
 	if err != nil {
 		return false, err
 	}
+
 	vers, err := version.GetClusterVersion(cv)
 	if err != nil {
 		return false, err
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -20,6 +20,7 @@ import (
 	extensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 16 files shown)


pkg/operator/helpers.go
```diff
@@ -4,9 +4,39 @@ package operator
...
+// OpenShift 4.11 and up
+func PodSecurityAdmissionControl(ctx context.Context, client client.Client) (bool, error) {
+	cv := &configv1.ClusterVersion{}
+
+	err := client.Get(ctx, types.NamespacedName{Name: "version"}, cv)
+	if err != nil {
+		return false, err
+	}
+
+	vers, err := version.ParseVersion(cv.Spec.DesiredUpdate.Version)
```
SrinivasAtmakuri, line 32: This should be as below. we want to get the current cluster version instead of the desired one to prevent a panic.

`vers, err := version.GetClusterVersion(cv)`

pkg/operator/helpers.go
```diff
@@ -4,9 +4,39 @@ package operator
...
 )
 
 func GatewayEnabled(cluster *arov1alpha1.Cluster) bool {
 	return len(cluster.Spec.GatewayDomains) > 0
 }
+
+// PodSecurityAdmissionControl is an admissions controller
+// for pods which replaces pod security policies, enabled on
+// OpenShift 4.11 and up
+func PodSecurityAdmissionControl(ctx context.Context, client client.Client) (bool, error) {
```
AldoFusterTurpin, line 24: A minor optional thing to clarify function usage: we could name that ShouldUsePodSecurityAdmission or similar.

pkg/operator/helpers.go
AldoFusterTurpin: Not sure if Ben used the _cv.Spec.DesiredUpdate.Version_ on purpose (maybe we need to use it to not apply the privilegedNamespaceLabels once it is too late. If we check for the cluster version once cluster was updated to 4.11+ maybe we have already failed to schedule the pods -but maybe it is not a problem-).

If it was done on purpose I think we need at least a 
```go
if cv.Spec.DesiredUpdate != nil 
```
to avoid a possible panic. 

Another option would be to check first for a possible desired update (with the current code + protection to nil) to 4.11 and if not, check the version with version.GetClusterVersion(cv).

pkg/operator/helpers.go
```diff
@@ -4,9 +4,39 @@ package operator
...
+	if err != nil {
+		return false, err
+	}
+
+	vers, err := version.ParseVersion(cv.Spec.DesiredUpdate.Version)
+	if err != nil {
+		return false, err
+	}
+
+	if vers.Lt(version.NewVersion(4, 11)) {
```
AldoFusterTurpin, line 37: Optional thing: we can extract the version.NewVersion(4, 11) to a package variable on top of this file like 
```go
var targetClusterVersion = version.NewVersion(4, 11)
```
and then just do
```go
if vers.Lt(targetClusterVersion)
```
We can then remove the hardcoded value from the function and the function's comment.

AldoFusterTurpin (commented):
One concern about a possible panic and a couple of optional minor suggestions.
Thank you!

pkg/operator/helpers.go
bennerv: moved it top-level.

pkg/operator/helpers.go
bennerv: Updated.

pkg/operator/helpers.go
bennerv: Using the correct function now.  Should be good as I dont' see where a panic can happen unless the CV object does not exist entirely, but then `err != nil`

AldoFusterTurpin: Nice. The panic I was referring to was on the direct access to _cv.Spec.DesiredUpdate_, now all safe 🙂

jaitaiwan (approved):
LGTM

cadenmarchese (approved):

AldoFusterTurpin (approved):
LGTM! Thanks for the minor tweaks.

PR #3403, opened 2024-02-13
Fix failures provisioning a cluster in dev due to NSG attaching

hawkowl:
fixes a thing I encountered trying to do local testing

Makes the NSG attaching code robust to authentication failures as well as the "not ready" bug

### Test plan for issue:

fixes local dev for me

N/A

### DIFF
--- a/pkg/cluster/deploybaseresources.go
+++ b/pkg/cluster/deploybaseresources.go
@@ -256,11 +256,25 @@ func (m *manager) subnetsWithServiceEndpoint(ctx context.Context, serviceEndpoin
 	return subnets, nil
 }
 
+// attachNSGs attaches NSGs to the cluster subnets, if preconfigured NSG is not
+// enabled. This method is suitable for use with steps, and has default
+// timeout/polls set.
 func (m *manager) attachNSGs(ctx context.Context) error {
-	return m._attachNSGs(ctx, 3*time.Minute)
+	// Since we need to guard against the case where NSGs are not ready
+	// immediately after creation, we can have a relatively short retry period
+	// of 30s and timeout of 3m. These numbers were chosen via a
+	// highly-non-specific and data-adjacent process (picking them because they
+	// seemed decent enough).
+	//
+	// If we get the NSG not-ready error after 3 minutes, it's unusual enough
+	// that we should be raising it as an issue rather than tolerating it.
+	return m._attachNSGs(ctx, 3*time.Minute, 30*time.Second)
 }
 
-func (m *manager) _attachNSGs(ctx context.Context, timeout time.Duration) error {
+// _attachNSGs attaches NSGs to the cluster subnets, if preconfigured NSG is not
+// enabled. timeout and pollInterval are provided as arguments for testing
+// reasons.
+func (m *manager) _attachNSGs(ctx context.Context, timeout time.Duration, pollInterval time.Duration) error {
 	if m.doc.OpenShiftCluster.Properties.NetworkProfile.PreconfiguredNSG == api.PreconfiguredNSGEnabled {
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


pkg/cluster/install.go
```diff
@@ -283,7 +283,7 @@ func (m *manager) bootstrap() []steps.Step {
 		steps.Action(m.ensureServiceEndpoints),
 		steps.Action(m.setMasterSubnetPolicies),
 		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.deployBaseResourceTemplate),
-		steps.Condition(m.attachNSGs, 3*time.Minute, true),
+		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.attachNSGs),
```
mociarain, line 286: To check my understanding: This changes from
- retry on `m.attachNSGs` failure for 3 mins
to
- run `m.attachNSGs` (and in it we run for 3 mins with 30s intervals)

If so why is this better? Fewer calls is not nothing but it's not clear to me how this addresses the main points you mentioned.

hawkowl: It's possible for the NSGs to not pass the authorisation checks immediately, so we need to also refresh that -- the existing code doesn't do that.

pkg/cluster/deploybaseresources.go
```diff
@@ -254,65 +256,84 @@ func (m *manager) subnetsWithServiceEndpoint(ctx context.Context, serviceEndpoin
 	return subnets, nil
 }
 
-func (m *manager) attachNSGs(ctx context.Context) (bool, error) {
+func (m *manager) attachNSGs(ctx context.Context) error {
+	return m._attachNSGs(ctx, 3*time.Minute)
+}
+
+func (m *manager) _attachNSGs(ctx context.Context, timeout time.Duration) error {
```
jaitaiwan, line 263: Curious as to why you've avoided changing the function signature of attachNSGs when it comes to the args but changed the return values anyway? Also I think the reasoning for 3m should be documented in a comment or if it's a standard value put somewhere else as a const

pkg/cluster/deploybaseresources.go
```diff
@@ -254,65 +256,84 @@ func (m *manager) subnetsWithServiceEndpoint(ctx context.Context, serviceEndpoin
...
+				if s.SubnetPropertiesFormat == nil {
+					s.SubnetPropertiesFormat = &mgmtnetwork.SubnetPropertiesFormat{}
+				}
 
-		nsgID, err := apisubnet.NetworkSecurityGroupID(m.doc.OpenShiftCluster, subnetID)
-		if err != nil {
-			return false, err
-		}
+				nsgID, err := apisubnet.NetworkSecurityGroupID(m.doc.OpenShiftCluster, subnetID)
+				if err != nil {
```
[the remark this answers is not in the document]
jaitaiwan: I don't think there's a good reason to switch

pkg/cluster/deploybaseresources.go
```diff
@@ -254,65 +256,84 @@ func (m *manager) subnetsWithServiceEndpoint(ctx context.Context, serviceEndpoin
...
-		// TODO: there is probably an undesirable race condition here - check if etags can help.
-
-		s, err := m.subnet.Get(ctx, subnetID)
-		if err != nil {
-			return false, err
-		}
+	timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	_ = wait.PollImmediateUntil(30*time.Second, func() (bool, error) {
```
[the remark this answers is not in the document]
jaitaiwan: It would be good to document the significance of 30 seconds with a comment. I don't think this should go inside the for loop - there's only two subnets being ranged over and the point seems to be that there may be errors at different stages of the process that require retrying from the top.

jaitaiwan (approved):
A few nits, LGTM

jaitaiwan (approved):
LGTM!

PR #3404, opened 2024-02-13
Create oic storage account in dev

cadenmarchese:

https://issues.redhat.com/browse/ARO-5029

- Creates storage account in dev environments, named after the resource group so that we have unique names
- Uses template function to remove dashes from storage account names - our resource groups commonly have dashes (v4-eastus, aro-australiaeast, aro-shared-cluster, etc...) and the storage account names don't allow these.
  - Also ensures storage account name doesn't exceed 24 characters.
- Creates role assignment on the RP MSI in a full env, or the RP service principal in a shared env
- Updates docs
- Keeps deployment of oic resources separate from other ARM templates to maintain reusability

Uses AKS and AZWI storage account examples as inspiration.

### Test plan for issue:

- [x] Tested applying this template: https://portal.azure.com/#view/HubsExtension/DeploymentDetailsBlade/~/overview/id/%2Fsubscriptions%2Ffe16a035-e540-4ab7-80d9-373fa9a3d6ae%2FresourceGroups%2Foictest%2Fproviders%2FMicrosoft.Resources%2Fdeployments%2Frp-oic
- [x] Tested a full RP service in dev to confirm that the role assignment is attached to the right MSI: https://portal.azure.com/#@redhat0.onmicrosoft.com/resource/subscriptions/fe16a035-e540-4ab7-80d9-373fa9a3d6ae/resourceGroups/marches-aro-eastus/providers/Microsoft.ManagedIdentity/userAssignedIdentities/aro-rp-eastus/azure_resources

Once the PR is approved and merged I can go and apply this to our shared environments.

Yes - updated.


pkg/deploy/assets/rp-oic.json
```diff
@@ -0,0 +1,41 @@
...
+    "resources": [
+        {
+            "sku": {
+                "name": "Standard_LRS"
+            },
+            "kind": "StorageV2",
+            "properties": {
+                "accessTier": "Hot",
+                "supportsHttpsTrafficOnly": true,
+                "allowBlobPublicAccess": true,
```
gouthamMN, line 18: Should this be set to False?

pkg/deploy/generator/resouces_oic.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+func (g *generator) oicStorageAccount() *arm.Resource {
+	storageAccount := &mgmtstorage.Account{
+		Kind: mgmtstorage.StorageV2,
+		Sku: &mgmtstorage.Sku{
+			Name: "Standard_LRS",
+		},
+		AccountProperties: &mgmtstorage.AccountProperties{
+			AllowBlobPublicAccess:  to.BoolPtr(true),
```
gouthamMN, line 31: Should we set this to False?

pkg/deploy/assets/rp-oic.json
cadenmarchese: My assumption is that in prod, yes (because we'll have Azure Front Door and won't need blobs publicly accessible), but in dev we wont have Azure Front Door so we need the public access. Is that right?

gouthamMN (approved):

niontive (approved):

hlipsig (approved):
LGTM

pkg/deploy/assets/rp-oic.json
SudoBrendan: In my read of the documentation - this setting allows anonymous/unauthenticated access to the SA: https://learn.microsoft.com/en-us/azure/storage/blobs/anonymous-read-access-configure?tabs=portal#set-the-storage-accounts-allowblobpublicaccess-property

If unauthenticated access is what we need, this is the right setting, I just want to be sure that is what we need?

SudoBrendan (commented):
If anonymous/unauthenticated access to this SA is a requirement, LGTM - I'm not sure if we need that or not?

pkg/deploy/assets/rp-oic.json
SudoBrendan: Caden and I talked offline, he pointed me to https://github.com/openshift/cloud-credential-operator/blob/master/docs/azure_workload_identity.md - the blob only holds public keys, LGTM.

SudoBrendan (approved):

PR #3405, opened 2024-02-13
[ARO-4562] Migrate DNS to track2 SDK

jonathan34c:

Migrate DNS to track2 SDK.

Replace [original dns](github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/dns) with newly createdDNS using [track2 SDK]("github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/dns/armdns")


bitoku:
We should do this migration carefully since the scope is too big.
It would be better not to replace the old one at a time but add the new client and migrate little by little with multiple small PRs.

jonathan34c:
@bitoku understood. I will close this PR for now and open one with new client

PR #3406, opened 2024-02-13
add archv1 check to multiple ip static validation

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-5484

Adds a check to api v2023-11-22 static validation to prevent updating managedOutboundIPs.count to anything greater than 1 for architecture v1 clusters.
### Test plan for issue:

Unit tests
No

### DIFF
--- a/pkg/api/v20230701preview/openshiftcluster_validatestatic.go
+++ b/pkg/api/v20230701preview/openshiftcluster_validatestatic.go
@@ -18,6 +18,7 @@ import (
 	"github.com/Azure/ARO-RP/pkg/api/validate"
 	"github.com/Azure/ARO-RP/pkg/util/pullsecret"
 	"github.com/Azure/ARO-RP/pkg/util/uuid"
+	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
 type openShiftClusterStaticValidator struct {
@@ -35,12 +36,16 @@ func (sv openShiftClusterStaticValidator) Static(_oc interface{}, _current *api.
 	sv.domain = domain
 	sv.requireD2sV3Workers = requireD2sV3Workers
 	sv.resourceID = resourceID
+	var architectureVersion api.ArchitectureVersion
 
 	oc := _oc.(*OpenShiftCluster)
 
 	var current *OpenShiftCluster
 	if _current != nil {
+		architectureVersion = _current.Properties.ArchitectureVersion
 		current = (&openShiftClusterConverter{}).ToExternal(_current).(*OpenShiftCluster)
+	} else {
+		architectureVersion = version.InstallArchitectureVersion
 	}
 
 	var err error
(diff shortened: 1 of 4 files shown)


pkg/api/v20231122/openshiftcluster_validatestatic.go
```diff
@@ -35,11 +35,12 @@ func (sv openShiftClusterStaticValidator) Static(_oc interface{}, _current *api.
 	sv.domain = domain
 	sv.requireD2sV3Workers = requireD2sV3Workers
 	sv.resourceID = resourceID
-
+	var architectureVersion *api.ArchitectureVersion
```
tony-schndr, line 38: I am using a pointer for the new variable `architectureVersion`, I chose to do it this way as there is no Architecture version on cluster create (because there is no `_current` cluster).  Also, I didn't think it would be a good idea to use the default 0 value for the variable as it could be mistaken for ArchitectureVersionV1

mociarain (commented):
LGTM but why are we only making this change in this version (`20231122`)?

tony-schndr:
> LGTM but why are we only making this change in this version (`20231122`)?

Good callout, this check can also be in the 2023-07-1-preview api.

tsatam (approved):

s-fairchild (approved):

mociarain (approved):

pkg/api/v20230701preview/openshiftcluster_validatestatic.go
```diff
@@ -35,12 +36,16 @@ func (sv openShiftClusterStaticValidator) Static(_oc interface{}, _current *api.
 	sv.domain = domain
 	sv.requireD2sV3Workers = requireD2sV3Workers
 	sv.resourceID = resourceID
+	var architectureVersion api.ArchitectureVersion
```
bennerv, line 39:
```suggestion
	architectureVersion := version.InstallArchitectureVersion

File: pkg/api/v20230701preview/openshiftcluster_validatestatic.go [hunk 3]
```diff
@@ -35,12 +36,16 @@ func (sv openShiftClusterStaticValidator) Static(_oc interface{}, _current *api.
...
 
 	oc := _oc.(*OpenShiftCluster)
 
 	var current *OpenShiftCluster
 	if _current != nil {
+		architectureVersion = _current.Properties.ArchitectureVersion
 		current = (&openShiftClusterConverter{}).ToExternal(_current).(*OpenShiftCluster)
+	} else {
+		architectureVersion = version.InstallArchitectureVersion
```
bennerv, line 48: No need for this, just the arch version to what's here.  
```suggestion
```

bennerv: Then can make this consistent throughout all the api versions that support this.

bennerv (commented):
just nits mostly.  Looks good otherwise.

bennerv (approved):

cadenmarchese (approved):
This is needed for next release, so we can address comments in a follow up PR.

tsatam:
Follow-up PR: https://github.com/Azure/ARO-RP/pull/3412

PR #3410, opened 2024-02-16
Refactor result type

mociarain:

This PR is a small refactor that comes out of a pairing session with myself and @sankur-codes. The originating PR is [here](https://github.com/Azure/ARO-RP/pull/3396).

### Test plan for issue:

I've added some unit tests.

Nope

### DIFF
--- a/pkg/util/log/log.go
+++ b/pkg/util/log/log.go
@@ -49,11 +49,11 @@ const (
 func MapStatusCodeToResultType(statusCode int) ResultType {
 	if statusCode < 300 && statusCode >= 200 {
 		return SuccessResultType
-	} else if statusCode < 500 {
-		return UserErrorResultType
-	} else {
-		return ServerErrorResultType
 	}
+	if statusCode < 500 {
+		return UserErrorResultType
+	}
+	return ServerErrorResultType
 }
 
 func getBaseLogger() *logrus.Logger {
--- a/pkg/util/log/log.go
+++ b/pkg/util/log/log.go
@@ -46,6 +46,16 @@ const (
 	ServerErrorResultType ResultType = "InternalServerError"
 )
 
+func MapStatusCodeToResultType(statusCode int) ResultType {
+	if statusCode < 300 && statusCode >= 200 {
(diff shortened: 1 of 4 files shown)


sankur-codes:
LGTM.

sankur-codes (approved):
LGTM.

pkg/util/log/log.go
```diff
@@ -46,6 +46,16 @@ const (
 	ServerErrorResultType ResultType = "InternalServerError"
 )
 
+func MapStatusCodeToResultType(statusCode int) ResultType {
```
AldoFusterTurpin, line 49: I think we can remove the else's{} to simplify the function (pretty common in Go).

pkg/util/log/log_test.go
```diff
@@ -11,6 +11,41 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func TestMapStatusCodeToResultType(t *testing.T) {
```
AldoFusterTurpin, line 14: This is just a minor thing that may not be worth it, but we could remove the expectedResult (eg. "SuccessResultType") from the test name to not couple the test name with the implementation (if the name of the result type changes, the test name should not change, because the behavior is the same no matter if it is called "UserErrorResultType" or "ErrorFromTheUserResultType").

pkg/util/log/log.go
AldoFusterTurpin:
```go
func MapStatusCodeToResultType(statusCode int) ResultType {
	if statusCode < 300 && statusCode >= 200 {
		return SuccessResultType
	}
	if statusCode < 500 {
		return UserErrorResultType
	}
	return ServerErrorResultType
}
```

AldoFusterTurpin (commented):
One simplification removing else's + an optional thing about the test. Otherwise, LGTM! Nice refactor. 🙂

AldoFusterTurpin (approved):
LGTM. Thanks!

sankur-codes (approved):
LGTM

tiguelu (approved):

PR #3385, opened 2024-02-07
Update images to Go 1.20

hawkowl:
### Which issue this PR addresses:

Fixes https://issues.redhat.com/browse/ARO-5320

### What this PR does / why we need it:
Updates images to Go 1.20, fixes build problems and CI/lint issues from Go 1.20.

### Test plan for issue:

Should be covered by e2e, unit tests, etc

### Is there any documentation that needs to be updated for this PR?

N/A

PR #3413, opened 2024-02-20
Clean up some of the CI/linting/generate

hawkowl:
Split out work of https://github.com/Azure/ARO-RP/pull/3385

Migrates the Github Actions to proper containerised steps, updates the linter and fixes some issues, and moves out the client-gen in generate since it doesn't work if you are out of GOROOT and causes hard to debug failures in generate.

### Test plan for issue:

Unit tests, e2e, etc

N/A


jaitaiwan (approved):
LGTM - not merging immediately because it deletes files and want to make sure that it's not something others need.

hawkowl:
@jaitaiwan I'm comfortable enough saying that these files _shouldn't_ be used, so if people are using them then breaking it so that we can get them using a better way is a good thing :)

PR #3415, opened 2024-02-20
Revert pkg/util/cluster creation to use 20230904 API

tsatam:

Fixes ongoing release

As part of the effort to eventually release the ARO 20231122 API, we had previously updated `pkg/util/cluster` to use this new API. Unfortunately, this breaks e2e tests against production regions, as this API does not yet officially exist. 

This PR temporarily updates pkg/util/cluster to use the 20230904 API, until we are able to update. 

This is intended to be a "hotfix" patch for the ongoing release, and as such the change is intended to be as small as possible. I think there is an opportunity for a more involved refactor to `pkg/util/cluster` to make changes like this easier, which I will evaluate separately. 

### Test plan for issue:

- [X] Manually ran `go run ./hack/cluster create` and ensured the request made it to my locally running RP
- [X] E2E tests in CI can still create a cluster
- [X] E2E tests should continue to pass (waiting for green e2e before merge)

- No (existing docs on API release process already call out the correct order of operations here)

When merging this PR, please squash this so we can cherry pick it easily.

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -42,7 +42,6 @@ import (
 	redhatopenshift20210901preview "github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/redhatopenshift/2021-09-01-preview/redhatopenshift"
 	redhatopenshift20220401 "github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/redhatopenshift/2022-04-01/redhatopenshift"
 	redhatopenshift20230904 "github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/redhatopenshift/2023-09-04/redhatopenshift"
-	redhatopenshift20231122 "github.com/Azure/ARO-RP/pkg/util/azureclient/mgmt/redhatopenshift/2023-11-22/redhatopenshift"
 	utilgraph "github.com/Azure/ARO-RP/pkg/util/graph"
 	"github.com/Azure/ARO-RP/pkg/util/rbac"
 	"github.com/Azure/ARO-RP/pkg/util/uuid"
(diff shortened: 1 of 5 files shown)


SudoBrendan (dismissed):

pkg/util/cluster/cluster.go
```diff
@@ -60,7 +61,7 @@ type Cluster struct {
 	openshiftclustersv20200430        redhatopenshift20200430.OpenShiftClustersClient
 	openshiftclustersv20210901preview redhatopenshift20210901preview.OpenShiftClustersClient
 	openshiftclustersv20220401        redhatopenshift20220401.OpenShiftClustersClient
-	openshiftclustersv20220904        redhatopenshift20231122.OpenShiftClustersClient
+	openshiftclustersv20230904        redhatopenshift20230904.OpenShiftClustersClient
 	openshiftclustersv20231122        redhatopenshift20231122.OpenShiftClustersClient
```
kimorris27, line 65: I don't see why simply instantiating this client without using it would break anything, but we could consider removing it completely to stay on the safe side.

kimorris27 (dismissed):

pkg/util/cluster/cluster.go
tsatam: The specific failures we saw during our RP deployment were when we attempted to call CreateOrUpdate on the 20231122 client, not during instantiation, so I think this is safe. I'll remove it anyways though. 

My overarching concern here is that I don't think we need more than one client here at all, but I'll tackle that in a more involved refactor that we don't hotfix.

s-fairchild (dismissed):

kimorris27 (approved):

jhoreman (approved):

bennerv (approved):

PR #3416, opened 2024-02-20
Update shared cluster login

mociarain:

This PR simplifies the `shared-cluster-loging` target by using az directly rather than using secrets we cache and share. More context [here](https://redhat-internal.slack.com/archives/C02ULBRS68M/p1708442218652829?thread_ts=1708427098.104239&cid=C02ULBRS68M). 

Note: An upside of this is that the re-creation of the shared cluster leads to zero changes in the secrets with these changes... automated shared-cluster rebuilds anyone??

I think we can do the following:
- Remove `secrets/shared-cluster.kubeconfig`
  - This file is old and very out of data
  - AFAIK nothing is using it
- Remove from `secrets/env`:
  - SHARED_CLUSTER_API
  - SHARED_CLUSTER_KUBEADMIN_PASSWORD

If people agree I'll implement the above.

### Test plan for issue:

- [X] test the updated target.

- [ ] review documentation around the shared cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -233,6 +233,7 @@ test-python: pyenv az
 
 
 shared-cluster-login:
+	RP_MODE="production" \
 	az aro get-admin-kubeconfig \
 		--name ${SHARED_CLUSTER_CLUSTER_NAME} \
 		--resource-group ${SHARED_CLUSTER_RESOURCE_GROUP_NAME} \
--- a/Makefile
+++ b/Makefile
@@ -233,7 +233,10 @@ test-python: pyenv az
 
 
 shared-cluster-login:
-	@oc login ${SHARED_CLUSTER_API} -u kubeadmin -p ${SHARED_CLUSTER_KUBEADMIN_PASSWORD}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -233,7 +233,11 @@ test-python: pyenv az
 
 
 shared-cluster-login:
-	@oc login ${SHARED_CLUSTER_API} -u kubeadmin -p ${SHARED_CLUSTER_KUBEADMIN_PASSWORD}
+	RP_MODE="production" \
+	az aro get-admin-kubeconfig \
```
tsatam, line 237: If we want to preserve the existing behavior of this command, we could instead use `az aro list-credentials`, which outputs username/password as JSON, and then invoke `oc login` with those parameters. 

IMO I prefer to have a dedicated kubeconfig for shared cluster on my machine, so I like the current implementation more, though I'd suggest changing the file name to something like `shared-cluster.admin.kubeconfig` to not clobber any dev cluster's kubeconfigs retrieved via `make admin.kubeconfig`.

PR #3418, opened 2024-02-22
Do not run Azure Host Servicing test in non-development environments

tsatam:

Fixes ongoing rollout

Skips the Azure Host Servicing E2E test when not in development environments. This is required as this test performs direct actions on the cluster VMs, which are protected by ARO's deny assignment in production regions. This causes this E2E test to fail when used e.g. to validate a production region during our deployment process. 

Note that this is intended to be a temporary fix - we should evaluate a more permanent solution that allows us to validate this behavior in all regions during our rollout.  

### Test plan for issue:

- [x] E2E tests continue to work in CI for PRs 
  - [X] Azure Host Servicing test should still run
- [ ] E2E tests continue to work against Int
  - [ ] Azure Host Servicing test should still run 
- [ ] E2E tests work again in production regions
  - [ ] Azure Host Servicing test should be skipped

No.

### DIFF
--- a/test/e2e/dns.go
+++ b/test/e2e/dns.go
@@ -39,6 +39,8 @@ const (
 )
 
 var _ = Describe("ARO cluster DNS", func() {
+	BeforeEach(skipIfNotInDevelopmentEnv)
+
 	It("must not be adversely affected by Azure host servicing", func(ctx context.Context) {
 		By("creating a test namespace")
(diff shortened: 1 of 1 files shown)


mociarain (approved):
Quick fix LGTM

cadenmarchese (approved):
makes sense, thanks!

PR #3422, opened 2024-02-26
reduce the amount of package names in the logs

hawkowl:

Part of https://github.com/Azure/ARO-RP/pull/3287/

Reduces the number of redundant package names in the logs

### Test plan for issue:

Unit tests

N/A


pkg/util/steps/runner.go
```diff
@@ -18,7 +18,12 @@ import (
 
 // FriendlyName returns a "friendly" stringified name of the given func.
 func FriendlyName(f interface{}) string {
-	return runtime.FuncForPC(reflect.ValueOf(f).Pointer()).Name()
+	return strings.TrimSuffix(runtime.FuncForPC(reflect.ValueOf(f).Pointer()).Name(), "-fm")
+}
+
+// FriendlyName returns a "friendly" stringified name of the given func.
+func ShortFriendlyName(f interface{}) string {
+	return strings.TrimPrefix(FriendlyName(f), "github.com/Azure/ARO-RP/")
```
cadenmarchese, line 26: If [this](https://github.com/Azure/ARO-RP/pull/3422/files#diff-4013a8e6a28b1efd731c440735d77847e53bb080cf7d1b9a79abbc2beb80d668R110) is the only place we are still using the old FriendlyName() function, should we just incorporate your changes directly to `FriendlyName()` rather than introduce this new function, and just adjust the logic in `enrichConditionTimeoutError()` if needed?

cadenmarchese (commented):
This will be nice to have for reading logs 👍

pkg/util/steps/runner.go
```diff
@@ -18,7 +18,12 @@ import (
...
 }
 
 func shortName(fullName string) string {
```
jaitaiwan, line 29: I think while we're here also changing shortName to something a more descriptive might be helpful to understand the diff between `FriendlyName` and `shortName`. Something like `StripPath`

pkg/util/steps/runner.go
jaitaiwan: I agree with Caden here, plus the request on the shortName change and this will really clean it up.

jaitaiwan (changes requested):
A couple small nits

mociarain:
I think this PR can be closed as these changes were merged as part of this: https://github.com/Azure/ARO-RP/pull/3630.

What do you think?

hawkowl:
@mociarain oh yeah, that one replaced this (pushing to main repo instead of fork), good catch

PR #3423, opened 2024-02-26
Make a package out of ARO-RP/pkg/api

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-3571

This makes ARO-RP/pkg/api a golang package, similar to how Hive's API package is used. This means we can reuse the ARO-RP API definitions in other repos, such as the ARO-Installer-Wrapper repo.

### Test plan for issue:

Unit tests should suffice

Not entirely sure -- the vendor check should catch if the vendoring doesn't occur after pkg/api changes.


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/Azure/ARO-RP
 go 1.18
 
 require (
+	github.com/Azure/ARO-RP/pkg/api v0.0.0-00010101000000-000000000000
```
mociarain, line 6: Why such a strange suffix or is this just some Go convention?

pkg/api/test/error/error.go
```diff
@@ -0,0 +1,18 @@
+package error
+
+import "testing"
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+// AssertErrorMessage asserts that err.Error() is equal to wantMsg.
+func AssertErrorMessage(t *testing.T, err error, wantMsg string) {
```
mociarain, line 9: Why duplicate this and the `CloudError` tests? Couldn't we drop the corresponding `/test/.../` files in favour of just these?

go.mod
hawkowl: That's a Go convention :(

pkg/api/test/error/error.go
hawkowl: I'm of two minds about this, honestly... perhaps it is just better to have everything use this.

go.mod
mociarain: Sometimes Go makes me sad

SudoBrendan:
I absolutely love this change, I had investigated it in a [repo restructure doc](https://docs.google.com/document/d/1aRoT_x2Ab5nDq3GKDhXAYccckyAAHb0CkW9ygH-XASo/edit#heading=h.9dv5btjqepxo) and it was my favorite choice (monorepo, multi-module). With it, however, we now have to question if `./vendor` is the right choice, because going down this pattern will eventually lead to that effectively being a duplicate of all code everywhere - that is, PRs now for `/pkg/api` are always going to be twice as big as they need to be, for no reason.

Like I said, I'm +1 for this, and +1 for removing vendoring (to be replaced with golang proxy server). Are there plans to make that happen?

hawkowl:
Supersceded by https://github.com/Azure/ARO-RP/pull/3423

PR #3425, opened 2024-02-27
Adding new virtual network to the list of cosmosdb nets

azoppiserpa:

Fixes https://issues.redhat.com/browse/ARO-5646

ARO-RP's pipeline has been failing since we tried to add a new virtual network (`ev2-vnet`), as described in the Jira issue. We'd like to see if adding the new virtual network to the list of networks for `cosmosdb` fixes the issue and allows the deployment to occur as intended.

### Test plan for issue:

Run `make generate` locally and observe if it executes without any errors.
Test this pull request on INT and observe if the deploy succeeds.
Check with Red Hat if these two steps are sufficient or if there are extra steps.
Try to deploy to production again.

`make generate` showed no errors.
Will test on INT soon and update result here.

I don't know. Will check with Red Hat. At the moment I'm creating the PR to share my proposed fix with others, and to be able to test on INT. If there is docs mentioning the virtual networks of `cosmosb`, it is likely that it requires updating, as we added a new virtual network.

### DIFF
--- a/pkg/deploy/assets/rp-production.json
+++ b/pkg/deploy/assets/rp-production.json
@@ -13,7 +13,7 @@
                 "id": "[resourceId('Microsoft.Network/virtualNetworks/subnets', 'aks-net', 'ClusterSubnet-001')]"
             },
             {
-                "id": "[resourceId('Microsoft.Network/virtualNetworks/subnets', 'ev2-vnet', 'ev2-subnet')]"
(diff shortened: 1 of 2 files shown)


azoppiserpa:
Running on INT: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=88362672&view=results

mociarain:
This issue has been solved in other work so I think this is no longer needed. closing for now but do re-open if I've misunderstood something.

PR #3427, opened 2024-02-28
Switch to using the secondary key

mociarain:
~~This is blocked until the keys are rotated in prod~~

Rollout is now underway and this can be merged.

Full context is [here](https://docs.google.com/document/d/1ectS3jrOAR6Y7dBF2S0zuSlH8DeqHOuL2cdRTIhm5q4/edit?usp=sharing) but it implements https://issues.redhat.com/browse/ARO-5491

Change RP to use Secondary RO/RW Cosmos DB Tokens

### Test plan for issue:

This will get tested correctly as part of the release.

Done as part of this PR

### DIFF
--- a/docs/dbtoken-service.md
+++ b/docs/dbtoken-service.md
@@ -8,12 +8,13 @@ In brief, there are three options:
 
 1. use r/w or r/o primary keys, which grant access to the whole database account
 2. implement a service which transforms (1) into scoped resource tokens
-3. a third AAD RBAC-based model is in preview.
+3. a third AAD RBAC-based model with Entra Ids.
 
 Currently, the RP, monitoring and portal service share the same security
 boundary (the RP VM) and use option 1.  The dbtoken service, which also runs on
-the RP VM, is our implementation of option 2.  As and when option 3 goes GA, it
-may be possible to retire the dbtoken service.
+the RP VM, is our implementation of option 2.  Option 3 is now in GA and there are plans
+[here](https://issues.redhat.com/browse/ARO-5512) to implement it and replace option 1
+and 2.
 
 The purpose of the dbtoken service at its implementation time is to enable the
 gateway component (which handles end-user traffic) to access the service Cosmos
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -68,7 +68,7 @@ func NewMasterKeyAuthorizer(ctx context.Context, token azcore.TokenCredential, c
 		return nil, err
 	}
 
-	return cosmosdb.NewMasterKeyAuthorizer(*keys.PrimaryMasterKey)
+	return cosmosdb.NewMasterKeyAuthorizer(*keys.SecondaryMasterKey)
 }
 
 func NewJSONHandle(aead encryption.AEAD) (*codec.JsonHandle, error) {
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -55,7 +55,7 @@ func NewDatabaseClient(log *logrus.Entry, _env env.Core, authorizer cosmosdb.Aut
 	return cosmosdb.NewDatabaseClient(log, c, h, databaseAccountName+"."+_env.Environment().CosmosDBDNSSuffix, authorizer), nil
 }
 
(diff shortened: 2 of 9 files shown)


tiguelu:
Added `hold`label as this PR cannot be merged until the secondary keys are rotated. We will remove it when they are rotated.

AldoFusterTurpin (commented):
Created a PR with a possible idea for the problem you mentioned 🙂
https://github.com/Azure/ARO-RP/pull/3458

pkg/database/database.go
```diff
@@ -68,7 +68,13 @@ func NewMasterKeyAuthorizer(ctx context.Context, token azcore.TokenCredential, c
...
 	}
 
-	return cosmosdb.NewMasterKeyAuthorizer(*keys.PrimaryMasterKey)
+	return cosmosdb.NewMasterKeyAuthorizer(getDatabaseKey(keys, log))
+}
+
+func getDatabaseKey(keys sdkcosmos.DatabaseAccountsClientListKeysResponse, log *logrus.Entry) string {
+	keyName := "SecondaryMasterKey"
+	log.Infof("Using %s to authenticate with CosmosDB", keyName)
+	return *keys.SecondaryMasterKey
```
AldoFusterTurpin, line 77: Thinking a bit more about this, maybe it could be good to instead of passing the log here, this function could return both the key and the "contextMessage" and do the log after calling the function. In that way we can simplify the test case just comparing the expected values without needing to deal with assertions in logs (we also avoid touching I/O in the function that retrieves the key).
Something like:
```go
func getDatabaseKey(keysResponse sdkcosmos.DatabaseAccountsClientListKeysResponse) (key, contextInfo string) {
	key = *keysResponse.SecondaryMasterKey

	keyName := "SecondaryMasterKey"
	contextInfo = fmt.Sprintf("Using %s to authenticate with CosmosDB", keyName)

	return key, contextInfo
}
```
or even
```go
func getDatabaseKey(keysResponse sdkcosmos.DatabaseAccountsClientListKeysResponse) (key, contextInfo string) {
	return *keysResponse.SecondaryMasterKey, fmt.Sprintf("Using %s to authenticate with CosmosDB", "SecondaryMasterKey")
}
```
and in func NewMasterKeyAuthorizer() ... on line 71 do
```go
        key, contextInfo := getDatabaseKey(keys)
	log.Infof(contextInfo)

	return cosmosdb.NewMasterKeyAuthorizer(key)
```

pkg/database/database_test.go
AldoFusterTurpin, line 1: Idea: we could avoid the testlog if we return directly the message (as mentioned above) and just assert on literal strings.

AldoFusterTurpin (dismissed):
LGTM. Thank you for this. I just left some optional suggestions in case any make sense to you (definitely not blocking the merge). 🙂

pkg/database/database.go
mociarain: I'm going to leave it as since it's working and we'll drop all this in the near future.

tsatam: I'm +1 to both the suggested change as well as the decision to not pursue it within this PR.

tsatam (dismissed):

tsatam (approved):

tiguelu (approved):

cadenmarchese (approved):

docs/dbtoken-service.md
```diff
@@ -8,12 +8,13 @@ In brief, there are three options:
 
 1. use r/w or r/o primary keys, which grant access to the whole database account
 2. implement a service which transforms (1) into scoped resource tokens
-3. a third AAD RBAC-based model is in preview.
+3. a third AAD RBAC-based model with Entra Ids.
```
kimorris27, line 11: What did you mean here by "with Entra Ids"?

I would have written this line as:

```
3. a third Microsoft Entra ID RBAC-based model
```

or maybe to make it a bit less formal:

```
3. a third Azure RBAC-based model
```

kimorris27 (commented):
The code changes LGTM! I just had one question/suggestion on the documentation update.

bennerv (approved):

docs/dbtoken-service.md
bennerv: Can address as part of follow-up as it's documentation.

mociarain: That's fair. PR here: https://github.com/Azure/ARO-RP/pull/3480

PR #3430, opened 2024-03-05
Remove unneeded OpenShift pins & imports

hawkowl:
See https://github.com/Azure/ARO-RP/pull/3428

Splitout of some of the dependency pinning work

Removes two dependencies that we don't need to carry, as well as the update scripts that mention them. This also means we don't need to carry the pre-4.10 machine provider spec garbage either.

### Test plan for issue:

E2E, unit tests

N/A


pkg/cluster/fixmcsuserdata.go
```diff
@@ -121,15 +132,15 @@ func (m *manager) fixMCSUserData(ctx context.Context) error {
...
+				return fmt.Errorf("failed decoding secret userData: %w", err)
 			}
 
 			var changed bool
 			for i, a := range userData.Ignition.Config.Merge {
 				var _changed bool
 				a.Source, _changed, err = m.fixSource(a.Source)
 				if err != nil {
-					return err
+					return fmt.Errorf("failed fixing domain source: %w", err)
```
mociarain, line 143:
```suggestion
					return fmt.Errorf("Failed fixing domain source when merging fields: %w", err)
```

pkg/cluster/fixmcsuserdata.go
```diff
@@ -141,7 +152,7 @@ func (m *manager) fixMCSUserData(ctx context.Context) error {
...
```
mociarain, line 155:
```suggestion
					return fmt.Errorf("Failed fixing domain source when appending fields: %w", err)
```

pkg/cluster/fixmcsuserdata.go
```diff
@@ -82,24 +82,35 @@ func getUserDataSecretReference(objMeta *metav1.ObjectMeta, spec *machinev1beta1
...
+
+	secret := &corev1.SecretReference{}
+
+	userDataSecretNamespace, ok, err := unstructured.NestedString(machineProviderSpec, "userDataSecret", "namespace")
+	if err != nil {
+		return nil, fmt.Errorf("failed getting secret reference: %w", err)
+	} else if !ok || userDataSecretNamespace == "" {
+		secret.Namespace = objMeta.Namespace
+	} else {
+		secret.Namespace = userDataSecretNamespace
```
mociarain, line 101: I think this path is missing a test.

mociarain (dismissed):
I think there's one test case missing but LGTM

mociarain:
Out of my own curiosity why change from your forked PR to this? AFAIK they're the same set of changes?

pkg/cluster/fixmcsuserdata.go
hawkowl: Errors are meant to have a lower case first letter, not entirely sure why, but it's a lint error otherwise :(

pkg/cluster/fixmcsuserdata.go
hawkowl: fixed

jaitaiwan (approved):
LGTM

PR #3434, opened 2024-03-05
Remove autosizednodes flag

mociarain:

Fixes: https://issues.redhat.com/browse/ARO-1663

This PR removes an old unneeded feature flag

### Test plan for issue:

Just the unit tests and the E2E.

I think not.

## Questions:

- Do we have docs on flags in other places I don't know about?
- Do we have a mimimum version of clusters we will support? The only other workaround is fixed after [4.4.11](https://github.com/azure/ARO-RP/blob/4ea4c0a23d8ae9302acca91fa1fe09402fd59f9c/pkg/operator/controllers/workaround/ifreload.go#L28). Can we remove it?

### DIFF
--- a/pkg/monitor/cluster/clusterflagsandbanner_test.go
+++ b/pkg/monitor/cluster/clusterflagsandbanner_test.go
@@ -51,12 +51,12 @@ func generateDefaultFlags() arov1alpha1.OperatorFlags {
 	return df
 }
 
-func generateNonStandardFlags(nonDefualtFlagNames []string) arov1alpha1.OperatorFlags {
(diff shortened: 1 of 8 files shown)


hawkowl:
Not sure this is actually unused on all clusters?

mociarain:
> Not sure this is actually unused on all clusters?

Good call. I'll confirm this

mociarain:
This PR isn't correct. I'll pull out the useful parts and open another soon.

mociarain:
> Not sure this is actually unused on all clusters?

For reference this is very used. I took away the wrong things from the ticket.

PR #3435, opened 2024-03-05
compare mdm/mdsd certs by serial number

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-5903

This pr updates rp/gateway vmss scripts, specifically the systemd service that checks for new mdm/mdsd certificates.  Instead of comparing based on the exit code of `diff` I am using openssl to compare the serial numbers of the new/current certificate. 

I originally considered using the certificates enddates, if the new cert end date is greater than the current cert then replace.  I decided against this because I could imagine a scenario where we need to do a cert renewal, where the new the certs are on a tighter renewal interval and the end date is earlier than the current.  In this scenario the cert would not be able to be renewed without a release.  But maybe that scenario is not realistic?

### Test plan for issue:
I tested this change locally by running the commands and pointing to local cert files.  This won't be truly tested until it is released so it would be good to monitor this once it is released to canary.


hlipsig (approved):

pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -581,7 +581,11 @@ if [ -f \$NEW_CERT_FILE ]; then
   else
     sed -i -ne '1,/END CERTIFICATE/ p' \$NEW_CERT_FILE
   fi
-  if ! diff $NEW_CERT_FILE $CURRENT_CERT_FILE >/dev/null 2>&1; then
+
+  NEW_CERT_SN=$(openssl x509 -in $NEW_CERT_FILE -noout -serial | awk -F= '{print $2}')
```
s-fairchild, line 585: A few suggestions.
1. We should wrap command substituions in quotes to prevent globbing
2. Ideally we should be using lowercase variable names for variables that aren't export, e. g. used only here. It's a little nit picky, but helps us clearly see what is intended for local use vs. exported use
3. The variables inside of command substitution should be quoted, to prevent globbing

pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -581,7 +581,11 @@ if [ -f \$NEW_CERT_FILE ]; then
...
+  CURRENT_CERT_SN=$(openssl x509 -in $CURRENT_CERT_FILE -noout -serial | awk -F= '{print $2}')
+  if [ "$NEW_CERT_SN" != "$CURRENT_CERT_SN" ]; then
+    echo updating certificate for \$COMPONENT
```
s-fairchild, line 588: What's the purpose of escaping `COMPONENT` here? I think we want to display the variable's value?

s-fairchild (changes requested):
This looks good, just a few change requests to improve reliability and some naming convention suggestions.

pkg/deploy/generator/scripts/rpVMSS.sh
s-fairchild: Also if you make use of `test` with double square brackets, you don't have to quote the left variable.  
The right is still expanded, has some good use cases.  
```bash
if [[ $NEW_CERT_SN != "$CURRENT_CERT_SN" ]]; then
    do stuff
fi
```

pkg/deploy/generator/scripts/rpVMSS.sh
s-fairchild: Also if you make use of `test` with double square brackets, you don't have to quote the left variable.  
The right is still expanded, has some good use cases.  
```bash
if [[ $NEW_CERT_SN != "$CURRENT_CERT_SN" ]]; then
    do stuff
fi
```

pkg/deploy/generator/scripts/rpVMSS.sh
tony-schndr: Because it is being added inline with cat > filename <<EOF .... and we want the $ sign to be included.  Which points out I missed a few escapes.

s-fairchild: Makes sense, I don't use `cat > file <<EOF` often so it looked out of place.

s-fairchild (approved):

pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -392,7 +392,11 @@ if [ -f \$NEW_CERT_FILE ]; then
   else
     sed -i -ne '1,/END CERTIFICATE/ p' \$NEW_CERT_FILE
   fi
-  if ! diff $NEW_CERT_FILE $CURRENT_CERT_FILE >/dev/null 2>&1; then
+
+  new_cert_sn="\$(openssl x509 -in "\$NEW_CERT_FILE" -noout -serial | awk -F= '{print \$2}')"
+  current_cert_sn="\$(openssl x509 -in "\$CURRENT_CERT_FILE" -noout -serial | awk -F= '{print \$2}')"
+  if [[ -n \$new_cert_sn ]] && [[ \$new_cert_sn != "\$current_cert_sn" ]]; then
```
SudoBrendan, line 398: I think that `! -z` (not null or empty) is a little safer than `-n` (not null), because we may end up with just `new_cert_sn=""` somehow

SudoBrendan (approved):
Overall LGTM, I tested with some self-signed certs, I think `-z` is slightly safer, but not a blocker on merge I think.

pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -581,7 +581,11 @@ if [ -f \$NEW_CERT_FILE ]; then
   else
     sed -i -ne '1,/END CERTIFICATE/ p' \$NEW_CERT_FILE
   fi
-  if ! diff $NEW_CERT_FILE $CURRENT_CERT_FILE >/dev/null 2>&1; then
+
+  new_cert_sn="\$(openssl x509 -in "\$NEW_CERT_FILE" -noout -serial | awk -F= '{print \$2}')"
+  current_cert_sn="\$(openssl x509 -in "\$CURRENT_CERT_FILE" -noout -serial | awk -F= '{print \$2}')"
+  if [[ -n \$new_cert_sn ]] && [[ \$new_cert_sn != "\$current_cert_sn" ]]; then
```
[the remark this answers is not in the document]
tony-schndr: Made this change

pkg/deploy/generator/scripts/gatewayVMSS.sh
tony-schndr: Made this change

jaitaiwan (approved):
LGTM but inquired about testing, as there's a lack of automated tests for this. Noting manual testing my multiple SREs: Tanmay and Tony have tested the bash script in INT and from a previous comment it seems that Brendan has also tested.

tiguelu (approved):

PR #3437, opened 2024-03-06
Remove ifreload workaround

mociarain:

When investigating [this](https://issues.redhat.com/browse/ARO-1663) I found that the other workaround we have can be removed. The `ifReload` workaround could only get used by a cluster older than `4.4.10` and the "youngest" cluster we have is 4.4.33: https://portal.microsoftgeneva.com/dashboard/share/AEA451FF.

### Test plan for issue:

Existing tests

Nope


tiguelu (approved):
lgtm

tsatam (approved):
LGTM - we're no longer updating the ARO operator on clusters that old anyways so this change should have no impact even if we did have clusters eligible for the workaround. 

I appreciate the cleanup and hope we can apply this to our other "workaround" controllers in due time.

mociarain:
> I appreciate the cleanup and hope we can apply this to our other "workaround" controllers in due time.

The other workaround is needed for clusters older than `4.8.0` and we still have a few of those hanging around :(

PR #3439, opened 2024-03-06
Minimal Python container to build `az aro` extension

tsatam:

Small step for [ARO-5088](https://issues.redhat.com/browse/ARO-5088)

- Creates a new Makefile target to build the `az aro` extension with no other dependencies
- Adds a new containerfile to create a build container for the above

### Test plan for issue:

Manually invoke target and ensure it creates an `aro-${VERSION}-py2.py3-none-any.whl` file within `./python/az/aro/dist`. 

Eventually, https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/188187/AzureCLI-preview-extension-publishing# should be updated to include this target. 

### Additional Notes
- This container does not perform the verification steps performed by the `azdev` command (`style`/`linter`/`test`). `azdev` requires that it be run within the context of a Git repository, as well as from within a specific virtualenv.


Dockerfile.aro-azext
```diff
@@ -0,0 +1,7 @@
+FROM mcr.microsoft.com/azure-cli:2.58.0
+
+RUN pip install pytest
+COPY /python /data/
+
+WORKDIR /data/az/aro
+CMD ["/bin/sh", "-c", "pytest --ignore=azext_aro/tests/latest/integration; python3 setup.py bdist_wheel"]
```
SudoBrendan, line 7: I would prefer if the `podman build` itself got us our extension built and ready to go, rather than requiring both `build` and `run` commands to generate an artifact. Can we:

1. do a `COPY` for all the files required to build (now, that's a `-v` in the makefile)
2. Make this `CMD` a `RUN` clause instead

This means our makefile becomes a single `podman build`, and by the end, we have a guaranteed known-good artifact (docker image) with all our changes. If we want those changes locally, we can issue a `podman cp` to extract it, but for now I'd like to just keep things 100% in the container space (and prevent artifact dumping in our git).

SudoBrendan (changes requested):

Dockerfile.aro-azext
SudoBrendan, line 1: nit: other containerfiles in this initiative are being named `Dockerfile.ci-*`, can we perhaps rename this file to be `Dockerfile.ci-aro-azext`?

SudoBrendan (changes requested):

SudoBrendan (approved):
While I think we'd all prefer for the style/lint/etc to be included, we can explore that in subsequent PRs since it seems like a lot of additional work. This is a great MVP for us to iterate on. Thanks Tanmay!

tsatam:
Rebased and opened replacement PR w/ in-repo branch for CI checks: https://github.com/Azure/ARO-RP/pull/3490

PR #3447, opened 2024-03-08
[ARO-5445] DNS checker logging custom DNS server as an error

rolandmkunkel:

Fixes https://issues.redhat.com/browse/ARO-5445

Customer reported ARO operator errors in the logs when custom DNS servers are configured in forward plugin:
https://portal.microsofticm.com/imp/v3/incidents/incident/469705806/summary

This is a [supported setup as per our docs](https://learn.microsoft.com/en-us/azure/openshift/dns-forwarding) and should not be logged as an error.

The DNS checks had been setup initially to be included as cluster conditions for SRE awareness (this did not log the status conditions as errors):
https://github.com/Azure/ARO-RP/blob/3e69b7e742af5aeee471a8d781d518c72fd235dd/pkg/operator/controllers/checker/clusterdns.go

A refactoring was done [in this commit](https://github.com/Azure/ARO-RP/commit/340ebb161f590eaf4ee069d519e6cfdd9d5c5f01). Since then, when the status now needs to be updated, the new logic is returning an [Error to the Reconciler](https://github.com/Azure/ARO-RP/blob/340ebb161f590eaf4ee069d519e6cfdd9d5c5f01/pkg/operator/controllers/checkers/clusterdnschecker/checker.go#L50), which in turn always [returns the error](https://github.com/Azure/ARO-RP/blob/340ebb161f590eaf4ee069d519e6cfdd9d5c5f01/pkg/operator/controllers/checkers/clusterdnschecker/controller.go#L82) to K8s controller-runtime, which is where it is being [logged as an error](https://github.com/Azure/ARO-RP/blob/master/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go#L317).

### DIFF
--- a/pkg/operator/controllers/checkers/clusterdnschecker/checker.go
+++ b/pkg/operator/controllers/checkers/clusterdnschecker/checker.go
@@ -14,7 +14,12 @@ import (
 )
 
 type clusterDNSChecker interface {
-	Check(ctx context.Context) error
+	Check(ctx context.Context) (result, error)
+}
+
+type result struct {
+	success bool
+	message string
 }
 
 type checker struct {
@@ -27,11 +32,11 @@ func newClusterDNSChecker(client client.Client) *checker {
 	}
 }
 
-func (r *checker) Check(ctx context.Context) error {
+func (r *checker) Check(ctx context.Context) (result, error) {
 	dns := &operatorv1.DNS{}
 	err := r.client.Get(ctx, types.NamespacedName{Name: "default"}, dns)
 	if err != nil {
-		return err
+		return result{}, err
 	}
 
 	var upstreams []string
@@ -40,8 +45,8 @@ func (r *checker) Check(ctx context.Context) error {
 			if z == "." {
 				// If "." is set as a zone, bail out and warn about the
 				// malformed config, as this will prevent CoreDNS from rolling
-				// out
-				return fmt.Errorf("malformed config: %q in zones", z)
+				// out.
+				return result{false, fmt.Sprintf("malformed config: %q in zones", z)}, nil
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/operator/controllers/checkers/clusterdnschecker/controller_test.go
```diff
@@ -39,24 +39,43 @@ func TestReconcile(t *testing.T) {
...
 		{
-			name:                 "no errors",
+			name: "no errors",
+			checkerReturnResult: result{
+				pass:    true,
+				message: "no in-cluster upstream DNS servers",
+			},
 			wantConditionStatus:  operatorv1.ConditionTrue,
-			wantConditionMessage: "No in-cluster upstream DNS servers",
+			wantConditionMessage: "no in-cluster upstream DNS servers",
```
mociarain, line 55: Our log convention is to start with lower case?

pkg/operator/controllers/checkers/clusterdnschecker/checker.go
```diff
@@ -14,7 +14,12 @@ import (
 )
 
 type clusterDNSChecker interface {
-	Check(ctx context.Context) error
+	Check(ctx context.Context) (result, error)
+}
+
+type result struct {
+	pass    bool
```
mociarain, line 21: Weak suggestion for `isSuccess` or just `success` instead of `pass`.

mociarain (approved):
LGTM. Small suggestion and a question but nothing blocking.

pkg/operator/controllers/checkers/clusterdnschecker/checker.go
rolandmkunkel: done

mociarain (approved):

rolandmkunkel:
superseded by https://github.com/Azure/ARO-RP/pull/3453 
E2E won't run from fork

PR #3458, opened 2024-03-14
add keysprovider package to ensure usage of specific key and not access them directly

AldoFusterTurpin:
After our short chat, this is something simple I came across .

We basically ensure we are using the appropriates key/s by using a new type (in a different package) that encapsulates the access to the keys using a public method. We just provide the implementation of the method of the key that we want to expose (the _keys: field is unexported to force the usage of the public method and use the key we want). 

So instead of:
get cosmosdb keys -> select explicitly a concrete key (error prone) -> use that key
do:
get cosmosdb keys -> instantiate our new keysprovider struct -> use that component to access the correct key.

Quick note, the fact that _ContextInfo_ in pkg/database/keysprovider/keysprovider.go is defined as a new type with _type ContextInfo string_ is interesting to ensure we do not use the keycontextinfo instead of keyvalue by mistake (it is satisfied at compile time) and we can easily convert it to string with _string(keycontextInfoValue)_. 

Also, for the testing part we don't need to deal with logrus Entries and just use plain strings (no more do something + log(I/O) at the same time, also good ?).

Maybe something nice would be to define masterkey parameter in _func NewMasterKeyAuthorizer(masterKey string) (Authorizer, error)_ in  pkg/database/cosmosdb/zz_generated_authorizer.go:34 as its own data type and make our new struct that I have defined be the only one returning a value of that type to ensure at compile time that we are passing the expected key(but that code is autogenerated AFAIK). 

This approach could be really useful if there were a lot of places where we use directly the keys and we change it to use this new type passing it as a reference instead of the raw keys. Right now is just one usage but I still see value on using it to prevent exactly that in the future.

This PR is just something simple I thought about instead of using reflection or something obscure.
🙂

### DIFF
--- a/docs/dbtoken-service.md
+++ b/docs/dbtoken-service.md
@@ -8,12 +8,13 @@ In brief, there are three options:
 
 1. use r/w or r/o primary keys, which grant access to the whole database account
 2. implement a service which transforms (1) into scoped resource tokens
-3. a third AAD RBAC-based model is in preview.
+3. a third AAD RBAC-based model with Entra Ids.
 
 Currently, the RP, monitoring and portal service share the same security
 boundary (the RP VM) and use option 1.  The dbtoken service, which also runs on
-the RP VM, is our implementation of option 2.  As and when option 3 goes GA, it
-may be possible to retire the dbtoken service.
+the RP VM, is our implementation of option 2.  Option 3 is now in GA and there are plans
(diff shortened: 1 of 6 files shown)


AldoFusterTurpin:
Good chat about the key stuff. Closing it for now as this secondary key thingy is just a temporary change that will be replaced with something different in the future.

PR #3464, opened 2024-03-15
Changes for retry logic - Min Worker Node Err

LiniSusan:

Fixes

### DIFF
--- a/pkg/util/steps/condition.go
+++ b/pkg/util/steps/condition.go
@@ -40,7 +40,7 @@ var timeoutConditionErrors = map[string]string{
 // condition has been met and an error.
 //
 // Suitable for polling external sources for readiness.
-type conditionFunction func(context.Context) (bool, error)
+type conditionFunction func(context.Context) (bool, bool, error)
 
 // Condition returns a Step suitable for checking whether subsequent Steps can
 // be executed.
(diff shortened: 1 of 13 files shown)


mociarain:
FYI the CI pipeline won't run on forks anymore. I'm not sure if people are looking into this but AFAIK you'll need to switch this to a branch on the main RP repo

pkg/util/steps/condition.go
```diff
@@ -86,8 +86,8 @@ func (c conditionStep) run(ctx context.Context, log *logrus.Entry) error {
 		// We use the outer context, not the timeout context, as we do not want
 		// to time out the condition function itself, only stop retrying once
 		// timeoutCtx's timeout has fired.
-		cnd, cndErr := c.f(ctx)
-		if errors.Is(cndErr, wait.ErrWaitTimeout) {
+		cnd, retry, cndErr := c.f(ctx)
+		if errors.Is(cndErr, wait.ErrWaitTimeout) && !retry {
```
mociarain, line 90: Why do we need this extra `retry` over the existing retry logic this has?

mociarain:
Also this might be helpful with the CI: https://redhat-internal.slack.com/archives/C02ULBRS68M/p1710774666598769

bitoku (commented):
I think it lacks the explanation how is the `retry` handled.
At first I thought that it indicates whether the error is retryable or not, but the logic doesn't seem so.

If you want to tell the `wait.Poll*` function if the condition is retryable, `ConditionFunc` can return (false, nil) for retryable error and (false, err) for non-retryable error.
https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediateUntil

mociarain:
What's the current state of this? Can I help?

LiniSusan:
Closing this as it is no more required.

PR #3467, opened 2024-03-19
Consistent storage account naming between dev and prod

cadenmarchese:

N/a

https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO-Pipelines/pullrequest/9703384?_a=files changed the way that the prod storage accounts are named (using storageAccountDomain + "oic" rather than prefix + "oidc" + region). This PR is meant to align our dev storage accounts to that as much as possible so that they follow the same naming convention.

In the case of the shared dev environments (v4-eastus, v4-westeurope, etc) those will likely have to stay as-is using the resource group name since in a local or shared dev environment there is no storageAccountDomain or RP-config defined for those environments.

### DIFF
--- a/hack/devtools/deploy-shared-env.sh
+++ b/hack/devtools/deploy-shared-env.sh
@@ -56,7 +56,8 @@ deploy_oic_dev() {
         -n rp-oic \
         --template-file pkg/deploy/assets/rp-oic.json \
         --parameters \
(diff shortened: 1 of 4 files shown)


cadenmarchese:
Closing in favor of https://github.com/Azure/ARO-RP/pull/3468.

PR #3468, opened 2024-03-19
Consistent storage account naming between dev and prod

cadenmarchese:

I've had to re-open this PR a few times due to e2e failures resulting from invalid tenant. I've moved over to the main trunk repo and hopefully e2e will succeed.

https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO-Pipelines/pullrequest/9703384?_a=files changed the way that the prod storage accounts are named (using storageAccountDomain + "oic" rather than prefix + "oidc" + region). This PR is meant to align our dev storage accounts to that as much as possible so that they follow the same naming convention.

In the case of the shared dev environments (v4-eastus, v4-westeurope, etc) those will likely have to stay as-is using the resource group name since in a local or shared dev environment there is no storageAccountDomain or RP-config defined for those environments.

### DIFF
--- a/hack/devtools/deploy-shared-env.sh
+++ b/hack/devtools/deploy-shared-env.sh
@@ -56,8 +56,8 @@ deploy_oic_dev() {
         -n rp-oic \
         --template-file pkg/deploy/assets/rp-oic.json \
         --parameters \
-            "rpServicePrincipalId=$(az ad sp list --filter "appId eq '$AZURE_RP_CLIENT_ID'" --query '[].id' -o tsv)" >/dev/null \
-            "storageAccountDomain=$(echo ${RESOURCEGROUP//-})"
+            "rpServicePrincipalId=$(az ad sp list --filter "appId eq '$AZURE_RP_CLIENT_ID'" --query '[].id' -o tsv)" \
+            "storageAccountDomain=$(echo ${RESOURCEGROUP//-})" >/dev/null
 }
 
 deploy_aks_dev() {
--- a/hack/devtools/deploy-shared-env.sh
+++ b/hack/devtools/deploy-shared-env.sh
@@ -56,7 +56,8 @@ deploy_oic_dev() {
         -n rp-oic \
         --template-file pkg/deploy/assets/rp-oic.json \
         --parameters \
-            "rpServicePrincipalId=$(az ad sp list --filter "appId eq '$AZURE_RP_CLIENT_ID'" --query '[].id' -o tsv)" >/dev/null 
+            "rpServicePrincipalId=$(az ad sp list --filter "appId eq '$AZURE_RP_CLIENT_ID'" --query '[].id' -o tsv)" >/dev/null \
+            "storageAccountDomain=$(echo ${RESOURCEGROUP//-})"
 }
 
 deploy_aks_dev() {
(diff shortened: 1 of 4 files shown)


hack/devtools/deploy-shared-env.sh
```diff
@@ -56,7 +56,8 @@ deploy_oic_dev() {
         -n rp-oic \
         --template-file pkg/deploy/assets/rp-oic.json \
         --parameters \
-            "rpServicePrincipalId=$(az ad sp list --filter "appId eq '$AZURE_RP_CLIENT_ID'" --query '[].id' -o tsv)" >/dev/null 
+            "rpServicePrincipalId=$(az ad sp list --filter "appId eq '$AZURE_RP_CLIENT_ID'" --query '[].id' -o tsv)" >/dev/null \
```
bitoku, line 59: I think `>/dev/null` should be one line below.

pkg/deploy/generator/resources_oic.go
```diff
@@ -16,8 +16,8 @@ import (
 
 var (
 	// Storage accounts must not contain dashes or be more than 24 characters
-	// Name it after the resource group + 'oic'
-	storageAccountName         string = "concat(take(replace(resourceGroup().name, '-', ''), 21), 'oic')"
+	// Append "oidc" to the pre-existing storage account prefix.
```
kimorris27, line 19: Should this be `Append "oic"...`?

hack/devtools/deploy-shared-env.sh
```diff
@@ -88,7 +89,8 @@ deploy_oic_for_dedicated_rp() {
         -n rp-oic \
         --template-file pkg/deploy/assets/rp-oic.json \
         --parameters \
-            "rpServicePrincipalId=$(az identity show -g $RESOURCEGROUP -n aro-rp-$LOCATION | jq -r '.["principalId"]')"
+            "rpServicePrincipalId=$(az identity show -g $RESOURCEGROUP -n aro-rp-$LOCATION | jq -r '.["principalId"]')" \
+            "storageAccountDomain=$(yq '.rps[].configuration.storageAccountDomain' dev-config.yaml | cut -d '.' -f1)"
```
kimorris27, line 93: Two questions here:

1. Could you run me through why `deploy_oic_dev` uses the resource group name but `deploy_aks_dev` uses `yq` to grab it from a config file? I think you intended to address this in the PR description, but I'm missing some connections between the dots.
2. Answering Q1 may help me understand this one as well, but isn't this shell stuff (`yq` output piped to `cut`) duplicating the changes made in `pkg/deploy/generator/resources_oic.go`? Why is that?

cadenmarchese: For your first question, we have 3 different environments to account for:
1) Prod - gets the storage account name from the storageAccountDomain, which is in RP-Config
2) Full RP service in dev - gets the storage account name from the storageAccountDomain, which is in dev-config.yaml (basically the same behavior as 1, or as similar as we can get)
3) Shared dev / local dev - there isn't a dev-config/RP-config available so we have to get the name from somewhere else. Since we need unique names, the only clear path forward that I can see is to use the resource group name, but I'm definitely open to ideas.

For your second question, good catch. I'll change to just `yq '.rps[].configuration.storageAccountDomain' dev-config.yaml` which should mirror the prod behavior more.

kimorris27 (approved):

hawkowl (approved):

PR #3473, opened 2024-03-21
Only perform machineconfig reconciliation during OpenShift upgrades

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-6250

Only performs MachineConfig reconciliation when OpenShift is being upgraded. 

### Test plan for issue:

Unit tests + E2E should cover it.

Yes, SOPs for manually forcing reconciliation

E2E + testing should cover known cases, but might require further testing.

### DIFF
--- a/pkg/operator/controllers/base/aro_controller.go
+++ b/pkg/operator/controllers/base/aro_controller.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/Azure/ARO-RP/pkg/operator"
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 	"github.com/Azure/ARO-RP/pkg/util/version"
 )
@@ -115,4 +116,20 @@ func (c *AROController) IsClusterUpgrading(ctx context.Context) (bool, error) {
 	}
 
 	return version.IsClusterUpgrading(cv), nil
+}
+
+func (c *AROController) AllowRebootCausingReconciliation(ctx context.Context, instance *arov1alpha1.Cluster) (bool, error) {
+	// If reconciliation is forced, perform it
+	if instance.Spec.OperatorFlags.GetSimpleBoolean(operator.ForceReconciliation) {
+		c.Log.Debugf("allowing reconciliation of %s because reconciliation forced", c.Name)
+		return true, nil
+	}
+
+	// Allow the reconciliation if the cluster is upgrading
+	isClusterUpgrading, err := c.IsClusterUpgrading(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	return isClusterUpgrading, nil
 }
--- a/pkg/operator/controllers/base/aro_controller.go
+++ b/pkg/operator/controllers/base/aro_controller.go
@@ -5,6 +5,7 @@ package base
 
 import (
 	"context"
+	"fmt"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
@@ -111,7 +112,8 @@ func (c *AROController) IsClusterUpgrading(ctx context.Context) (bool, error) {
 	cv := &configv1.ClusterVersion{}
 	err := c.Client.Get(ctx, types.NamespacedName{Name: "version"}, cv)
 	if err != nil {
-		c.Log.Errorf("error getting the ClusterVersion: %v", err)
+		err = fmt.Errorf("error getting the ClusterVersion: %w", err)
+		c.Log.Error(err)
 		return false, err
 	}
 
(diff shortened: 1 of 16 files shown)


hlipsig:
I added the "next release" tag, but I mean next next release. Not the one that happens after we finally release everything that's been waiting since January.

UlrichSchlueter:
On hold for now, preference is to try a hive based approach first

mociarain:
What happened to the Hive based approach mentioned above? Is this still on hold?

pkg/operator/controllers/base/aro_controller.go
```diff
@@ -103,3 +107,31 @@ func (c *AROController) defaultDegraded() *operatorv1.OperatorCondition {
...
+	cv := &configv1.ClusterVersion{}
+	err := c.Client.Get(ctx, types.NamespacedName{Name: "version"}, cv)
+	if err != nil {
+		err = fmt.Errorf("error getting the ClusterVersion: %w", err)
+		c.Log.Error(err)
+		return false, err
+	}
+
+	return version.IsClusterUpgrading(cv), nil
+}
```
SudoBrendan, line 121: ok - just a thought I had - is this singular case an appropriate one for us to set the ARO Operator into Progressing until all MCs are successfully applied and validated? If I understand correctly, this would put the upgrade on "pause" until we are 100% reconciled.

pkg/util/version/clusterversion.go
```diff
@@ -51,3 +51,7 @@ func ClusterVersionIsLessThan4_4(ctx context.Context, configcli configclient.Int
 	// 4.3 uses SRV records for etcd
 	return v.Lt(NewVersion(4, 4)), nil
 }
+
+func IsClusterUpgrading(cv *configv1.ClusterVersion) bool {
+	return cv.Spec.DesiredUpdate != nil
```
wking, line 56: This is probably not what you want to check.  For example, in a random cluster I have access to:

```console
$ date --utc --iso=m
2024-08-27T16:06+00:00
$ oc get -o json clusterversion version | jq '{desiredUpdate: .spec.desiredUpdate, progressing: ([.status.conditions[] | select(.type == "Progressing")][0]), history: .status.history[0]}'
{
  "desiredUpdate": {
    "architecture": "",
    "force": false,
    "image": "quay.io/openshift-release-dev/ocp-release@sha256:6a80ac72a60635a313ae511f0959cc267a21a89c7654f1c15ee16657aafa41a0",
    "version": "4.16.9"
  },
  "progressing": {
    "lastTransitionTime": "2024-08-27T10:47:19Z",
    "message": "Cluster version is 4.16.9",
    "status": "False",
    "type": "Progressing"
  },
  "history": {
    "completionTime": "2024-08-27T10:47:19Z",
    "image": "quay.io/openshift-release-dev/ocp-release@sha256:6a80ac72a60635a313ae511f0959cc267a21a89c7654f1c15ee16657aafa41a0",
    "startedTime": "2024-08-27T09:40:53Z",
    "state": "Completed",
    "verified": true,
    "version": "4.16.9"
  }
}
```

That cluster has been on 4.16.9 for hours, but `desiredUpdate` is still populated with the 4.16.9 ask, because... that ask is still valid, regardless of whether the cluster-version operator has delivered it yet or not.

You probably want to follow `oc adm upgrade ...` and [check the `status` of the `Progressing` condition][1].

[1]: https://github.com/openshift/oc/blob/96c567af10356e8461c02361cdd243693a6ea53c/pkg/cli/admin/upgrade/upgrade.go#L629-L630

hlipsig: We've updated the code to follow this recommendation if you feel like taking another look.

pkg/operator/controllers/base/aro_controller.go
hlipsig: It's this case and also for the full deprecation of DNSMasq. What is the concern about the operator status remaining in progressing for a prolonged period of time? Is it that it may obfuscate other issues?

tsatam (dismissed):

pkg/operator/controllers/dnsmasq/cluster_controller.go
```diff
@@ -109,5 +131,19 @@ func reconcileMachineConfigs(ctx context.Context, instance *arov1alpha1.Cluster,
...
+	} else {
+		for _, i := range resources {
+			err := c.Create(ctx, i.(client.Object))
+			// Since we are only creating, ignore AlreadyExists
+			if err != nil && !kerrors.IsAlreadyExists(err) {
+				return fmt.Errorf("error creating client object: %w", err)
+			}
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
hawkowl: Yeah, we need to update it for entirely-new MCs, since that was outside of the scope of the original PR in March.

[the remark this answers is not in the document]
tsatam: I've had this question for a while - why do we care about customers creating new MachineConfigPools at all for either the current dnsmasq or new /etc/hosts MachineConfigs? It should be sufficient to create MachineConfigs that apply to role: master and role: worker, as every node should be a member of one of those two roles, even if they belong to a custom MCP.

SudoBrendan (approved):
I trust the tests here, thanks Amber.

jaitaiwan (approved):
I finished QA'ing and it was working fine. Approved from my standpoint as well.

PR #3477, opened 2024-03-21
Add git hooks for branch name validation

SudoBrendan:

Fixes [ARO-6316](https://issues.redhat.com/browse/ARO-6316)

See [ARO Branching Strategy](https://docs.google.com/document/d/1wDe1vBCyzjul6ZXJM5ZFMa8a7-qmiSuWueis4S0tyA0/edit).

### Test plan for issue:

Local git - attempted to use "bad" branch names and ensured they failed to create commits locally. `git branch --move some-bad-name some-good-name` works and passes the pre-commit hooks.

Yes, included in this PR. I will send notifications to all SREs both in slack and email of this change.

Local dev changes, not deployed to prod.

### DIFF
--- a/hack/git/hooks/pre-commit
+++ b/hack/git/hooks/pre-commit
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+LC_ALL=C
+
+# e.g. "ARO-1234/short-description" or "hotfix/v20240321.00"
+valid_branch_regex="^(ARO-[0-9]{4}|hotfix)\/[a-z0-9._-]+$"
+local_branch="$(git rev-parse --abbrev-ref HEAD)"
+
+if [[ ! $local_branch =~ $valid_branch_regex ]]
+then
+    echo "There is something wrong with your branch name. Branch names in this project must adhere to this contract: $valid_branch_regex. Your commit will be rejected. Please rename your branch (git branch --move) to a valid name and try again."
+    exit 1
+fi
+
+exit 0
--- a/docs/prepare-your-dev-environment.md
+++ b/docs/prepare-your-dev-environment.md
@@ -138,3 +138,9 @@ Make sure that `PKG_CONFIG_PATH` contains the pkgconfig files of the above packa
     ```bash
     cd ${GOPATH:-$HOME/go}/src/github.com/Azure/ARO-RP
     ```
+
+1. Add standard git hooks
+
+    ```bash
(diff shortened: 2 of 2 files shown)


kimorris27:
I'm cool with this as-is, but I think it would also be neat to have the hook setup added as a Makefile target like @mbarnes pointed out in the Google Doc.

hack/git/hooks/pre-commit
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+LC_ALL=C
+
+# e.g. "ARO-1234/short-description" or "hotfix/v20240321.00"
+valid_branch_regex="^(ARO-[0-9]{4}|hotfix)\/[a-z0-9._-]+$"
```
cadenmarchese, line 5: The proposal looks good to me, but it seems like we now are requiring a JIRA for all PRs. This seems like a barrier for quick fix PRs, or contributors outside of Red Hat. Do you have thoughts on how we can do something like this without requiring a JIRA link?

Maybe we can add a regex for `nojira`: 
```suggestion
valid_branch_regex="^(nojira|ARO-[0-9]{4}|hotfix)\/[a-z0-9._-]+$"
```

cadenmarchese: We should probably include an update here in this PR as well: https://github.com/Azure/ARO-RP/blob/master/README.md#contributing

mociarain: Along the same lines why not just:
```suggestion
valid_branch_regex="[a-z0-9._-]*\/[a-z0-9._-]+$"
```

As stated in the doc we just need difference and the names don't matter so why not just keep it open? I'm usually a fan of consistency but being simple seems even nicer here.

SudoBrendan: I liked the idea of using github usernames as a prefix for this (pointed out in the doc), because the core thing we want to avoid is people squashing other IC's commits with a --force (unfortunately required in our repo today without a simple fix). I had to close this branch to match that new schema.

The point on open source contribs I was originally on-board with; that made more sense in a fork world, but now we're in a complex situation - do we plan to open up `create branch` permissions to the internet on this repo?

SudoBrendan: I like this discussion let's move it to https://github.com/Azure/ARO-RP/pull/3479 (sorry; I would have preferred to keep this one open under a new branch but apparently that's not possible in GH or I messed up the flow)

PR #3478, opened 2024-03-22
Add new initial fields to v20240812preview

cadenmarchese:

Fixes https://issues.redhat.com/browse/ARO-4381 && https://issues.redhat.com/browse/ARO-4377, unblocks https://issues.redhat.com/browse/ARO-3979

This is the first of 3-ish PRs to add new fields to the preview API + converter. There will be additional fields added to support OCP upgrades as well as Cluster MSI.


pkg/api/v20240812preview/openshiftcluster.go
```diff
@@ -55,6 +58,9 @@ type OpenShiftClusterProperties struct {
 	// The cluster service principal profile.
 	ServicePrincipalProfile ServicePrincipalProfile `json:"servicePrincipalProfile,omitempty"`
 
+	// The workload identity profile.
+	PlatformWorkloadIdentityProfile PlatformWorkloadIdentityProfile `json:"platformWorkloadIdentityProfile,omitempty"`
```
kimorris27, line 62: This field is a pointer in the design doc to allow for it to be `nil` for CSP clusters.

pkg/api/v20240812preview/openshiftcluster.go
```diff
@@ -276,6 +288,33 @@ type IngressProfile struct {
 	IP string `json:"ip,omitempty" swagger:"readOnly"`
 }
 
+type PlatformWorkloadIdentityProfile struct {
+	PlatformWorkloadIdentities []PlatformWorkloadIdentity `json:"platformWorkloadIdentities,omitempty"`
+}
+
+// PlatformWorkloadIdentity stores information representing a single workload identity.
+type PlatformWorkloadIdentity struct {
+	Name       string `json:"name,omitempty"`
```
kimorris27, line 297: This was originally `Name`, but we later changed it to `OperatorName` in the design doc to make the purpose of the field more clear.

pkg/api/openshiftcluster.go
```diff
@@ -122,6 +123,8 @@ type OpenShiftClusterProperties struct {
 
 	ServicePrincipalProfile ServicePrincipalProfile `json:"servicePrincipalProfile,omitempty"`
 
+	PlatformWorkloadIdentityProfile PlatformWorkloadIdentityProfile `json:"platformWorkloadIdentityProfile,omitempty"`
```
kimorris27, line 126: This field is a pointer in the design doc to allow for it to be `nil` for CSP clusters.

pkg/api/openshiftcluster.go
```diff
@@ -761,3 +767,38 @@ type HiveProfile struct {
...
+	MissingFields
+
+	PlatformWorkloadIdentities []PlatformWorkloadIdentity `json:"platformWorkloadIdentities,omitempty"`
+}
+
+// PlatformWorkloadIdentity stores information representing a single workload identity.
+type PlatformWorkloadIdentity struct {
+	MissingFields
+
+	Name       string `json:"name,omitempty"`
```
kimorris27, line 781: This was originally `Name`, but we later changed it to `OperatorName` in the design doc to make the purpose of the field more clear.

kimorris27 (changes requested):
I suggested some small changes to make sure we're lining up with what the team agreed on in the design doc.

kimorris27 (approved):

kimorris27 (changes requested):
Just noticed that we're missing the admin API changes. Let's add those into this PR.

kimorris27 (dismissed):

pkg/api/admin/openshiftcluster.go
```diff
@@ -24,27 +24,29 @@ type OpenShiftCluster struct {
...
+	MaintenanceTask                 MaintenanceTask                  `json:"maintenanceTask,omitempty" mutable:"true"`
+	OperatorFlags                   OperatorFlags                    `json:"operatorFlags,omitempty" mutable:"true"`
+	OperatorVersion                 string                           `json:"operatorVersion,omitempty" mutable:"true"`
+	CreatedAt                       time.Time                        `json:"createdAt,omitempty"`
+	CreatedBy                       string                           `json:"createdBy,omitempty"`
+	ProvisionedBy                   string                           `json:"provisionedBy,omitempty"`
+	ClusterProfile                  ClusterProfile                   `json:"clusterProfile,omitempty"`
+	FeatureProfile                  FeatureProfile                   `json:"featureProfile,omitempty"`
+	ConsoleProfile                  ConsoleProfile                   `json:"consoleProfile,omitempty"`
+	ServicePrincipalProfile         ServicePrincipalProfile          `json:"servicePrincipalProfile,omitempty"`
```
tsatam, line 46: Are we planning on turning this field into a pointer, to allow it to be nil for WIMI clusters (inverse of `PlatformWorkloadIdentityProfile` below)? 

Making this change would require us to ensure nil checks across all of our references to the field throughout the codebase, so I'm in favor of it being left out of this PR and in a dedicated one if we do go down that path.

tsatam (approved):
LGTM, I've left one question about handling the existing ServicePrincipalProfile struct on WIMI clusters for my own curiosity, but it shouldn't block this PR and it would likely need to be the subject of its own dedicated PR if we want to tackle it.

pkg/api/admin/openshiftcluster.go
kimorris27: I'm personally in favor of making that change, and I agree that it should go in a dedicated PR.

cadenmarchese: Agreed, we have talked about doing this as well. I created a dedicated card for this: https://issues.redhat.com/browse/ARO-6623

pkg/api/admin/openshiftcluster.go
```diff
@@ -409,6 +415,33 @@ type IngressProfile struct {
 	IP         string     `json:"ip,omitempty"`
 }
 
+type PlatformWorkloadIdentityProfile struct {
```
s-amann, line 418: should we also add some doc comments here?

pkg/api/openshiftcluster.go
```diff
@@ -248,6 +251,8 @@ const (
 	FipsValidatedModulesDisabled FipsValidatedModules = "Disabled"
 )
 
+type OIDCIssuer string
```
s-amann, line 254: we have doc comments in the admin version, should we also add some doc comments here?

pkg/api/openshiftcluster.go
```diff
@@ -761,3 +767,38 @@ type HiveProfile struct {
 	// ClusterDeployments.
 	CreatedByHive bool `json:"createdByHive,omitempty"`
 }
+
+type PlatformWorkloadIdentityProfile struct {
```
s-amann, line 771: should we also add some doc comments here?

pkg/api/v20240812preview/openshiftcluster.go
```diff
@@ -276,6 +288,33 @@ type IngressProfile struct {
 	IP string `json:"ip,omitempty" swagger:"readOnly"`
 }
 
+type PlatformWorkloadIdentityProfile struct {
```
s-amann, line 291: should we also add some doc comments here?

s-amann (approved):
lgtm. Some minor comments below which could be in a new PR

cadenmarchese:
@s-amann should be good now, thanks for pointing that out!

kimorris27 (approved):

kimorris27 (changes requested):
I checked out the CI failure, and it looks like the full results of `make client` haven't yet been pushed since the most recent changes.

tsatam (approved):

s-amann (approved):

kimorris27 (approved):

PR #3479, opened 2024-03-22
Add git hooks for branch name validation

SudoBrendan:

Fixes [ARO-6316](https://issues.redhat.com/browse/ARO-6316) - moved from #3477 

See [ARO Branching Strategy](https://docs.google.com/document/d/1wDe1vBCyzjul6ZXJM5ZFMa8a7-qmiSuWueis4S0tyA0/edit). Now that branching is required over forks, we need to keep our repo organized. This PR introduces strict naming conventions of branches to:

1. Minimize risk of development efforts getting squashed via --force pushing
2. Make CODEOWNER lives a bit easier while managing branches in rare situations like branches going stale, employees leaving the company, etc.

### Test plan for issue:

Local git - attempted to use "bad" branch names and ensured they failed to create commits locally. `git branch --move some-bad-name some-good-name` works and passes the pre-commit hooks.

Yes, included in this PR. I will send notifications to all SREs both in slack and email of this change.

Local dev changes, not deployed to prod.

### DIFF
--- a/hack/git/hooks/pre-commit
+++ b/hack/git/hooks/pre-commit
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+set -e
+LC_ALL=C
+
+git_username_lower="$(git config github.user | tr '[:upper:]' '[:lower:]')"
+if [[ -z "${git_username_lower}" ]]
+then
+    echo "Please set github.user (git config github.user) locally or globally before issuing commits to this repo."
+    exit 1
+fi
+
+# e.g. "USERNAME/ARO-1234", "USERNAME/hotfix-v20240321.00", or "USERNAME/gh-issue-123"
+valid_branch_regex="^${git_username_lower}\/(ARO-[0-9]{4}[a-z0-9._-]*|hotfix-[a-z0-9._-]+|gh-issue-[0-9]+[a-z0-9._-]*)$"
+
+local_branch="$(git rev-parse --abbrev-ref HEAD)"
+
+if [[ ! $local_branch =~ $valid_branch_regex ]]
+then
+    echo "There is something wrong with your branch name. Branch names in this project must adhere to this contract: $valid_branch_regex. Your commit will be rejected. Please rename your branch (git branch --move) to a valid name and try again."
+    exit 1
+fi
+
+exit 0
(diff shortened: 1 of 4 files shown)


SudoBrendan:
cc @cadenmarchese @kimorris27 @mociarain thank you for your review of the old PR, I had to close it in favor of this one, and I hope I addressed all the feedback left in subsequent changes. LMK what you think.

tiguelu (approved):

hack/git/hooks/pre-commit
```diff
@@ -0,0 +1,23 @@
...
+
+git_username_lower="$(git config github.user | tr '[:upper:]' '[:lower:]')"
+if [[ -z "${git_username_lower}" ]]
+then
+    echo "Please set github.user (git config github.user) locally or globally before issuing commits to this repo."
+    exit 1
+fi
+
+# e.g. "USERNAME/ARO-1234", "USERNAME/hotfix-v20240321.00", or "USERNAME/gh-issue-123"
+valid_branch_regex="^${git_username_lower}\/(ARO-[0-9]{4}[a-z0-9._-]*|hotfix-[a-z0-9._-]+|gh-issue-[0-9]+[a-z0-9._-]*)$"
```
mbarnes, line 13: Minor objection to the `$` at the end.  Could we allow for an optional 3rd free-form segment for cards/issues requiring multiple PRs.  Like:
```
USERNAME "/" TICKET ["/" .*]
```

I'm just imagining a followup PR for something that was missed or deferred in the initial review.  e.g. `mbarnes/ARO-1234/bugfix`

SudoBrendan: is `ARO-1234-bugfix` not acceptable in this case? I can add `/` to the regex if that is considered a blocker.

tsatam: I think, for our goals with naming consistency here, we should only really need to enforce the prefix of the name, if all the required information is present at the start of the name for easy identification, it shouldn't matter what other suffixes we might need to include on the branch, be it identifications for split units of work, easy clarification to identify branches without memorizing Jira IDs, etc.

mbarnes: My bad, I didn't read the regex close enough.  A 2nd slash would make the branch name easier to parse if there's ever a need to do that.  But not important enough to block on.

kimorris27 (approved):

PR #3485, opened 2024-03-28
Return image mirroring errors to calling function

tiguelu:

Fixes
https://issues.redhat.com/browse/ARO-6456

The logs from the mirroring command are huge. Ev2 pipelines and Ev2 Portal trim the output from the mirroring command and errors mirroring images at the beginning of the execution are lost and not observable.

The mirroring function doesn't immediately fail and return an error if an image cannot be mirrored, it just prints it in the log. This is by design, to continue mirroring any other remaining images without returning if there is a failure.

However, if one or more images fail to be mirrored, we just return a "failed to mirror" error to the calling function, instead of a collection of failed images/errors. The log is trimmed by Ev2 due to its big size, and we lose visibility of what failed.

### Test plan for issue:

Ev2 pipeline is currently failing and we can't see which images are failing to be mirrored, blocking 4.14 release and important OCP bugfix version updates for customers.

We'll rerun the pipeline and check the errors are printed after merging the PR

No.

Rerun mirroring pipelines and check logs are working and errors displayed at the end of the execution.

### DIFF
--- a/cmd/aro/mirror.go
+++ b/cmd/aro/mirror.go
@@ -76,10 +76,13 @@ func mirror(ctx context.Context, log *logrus.Entry) error {
 
 	// Geneva allows anonymous pulls
 	var srcAuthGeneva *types.DockerAuthConfig
-	var errorOccurred bool
+
+	// We can lose visibility of early image mirroring errors because logs are trimmed in the output of Ev2 pipelines.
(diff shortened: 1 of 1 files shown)


UlrichSchlueter (approved):
LGTM

PR #3486, opened 2024-03-28
accept pull-secret when we use hack script to create a clsuter

bitoku:

This is the same PR as #3360 but it was made from a branch not a fork to pass the E2E.

There is no way to specify pull-secret when we use the hack script to create a cluster.
When testing the installer, if the installer image is hosted in quay.io, hive doesn't changes the domains of other images required during installation to ACR domain and it causes Authentication error if it doesn't have the pull-secret.

This pull-secret enables us to access to those images in quay.io and make the installer test convenient.

### Test plan for issue:

I was able to create a cluster with ACR domain installer and quay.io domain installer.


kimorris27 (commented):
Have you tested what happens if you leave the environment variable empty? Does cluster creation still work?

bitoku:
yes it does. also `USER_PULL_SECRET` is set in the secret so we normally don't have to care about it.

kimorris27 (approved):

tiguelu (approved):

tiguelu:
LGTM, as a follow-up, can you add a comment to our doc to document the use of this new var:
https://github.com/Azure/ARO-RP/blob/master/docs/deploy-development-rp.md#run-the-rp-and-create-a-cluster

PR #3489, opened 2024-03-29
multi staged CI via make and docker build

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-5087

Creates a new make target and container file that builds/tests the aro-rp stages.  Stages are ran in parallel to improve efficiency.

### Test plan for issue:

Run `make ci-rp` and verify final aro-rp image is produced.

NA

The first stage and last stage of this build are nearly identical to `Dockerfile.aro-multistage`, the difference being moving FIPs checks to its own build stage.

### DIFF
--- a/Dockerfile.ci-rp
+++ b/Dockerfile.ci-rp
@@ -1,0 +1,22 @@
+ARG REGISTRY
+FROM ${REGISTRY}/ubi8/go-toolset:1.20.10 as builder
+
+USER root
+ENV GOPATH=/root/go
+WORKDIR /app
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/local/bin v1.55.2
+COPY . /app
+RUN make aro RELEASE=${IS_OFFICIAL_RELEASE} -o generate && make e2e.test &&\
+  go mod tidy -compat=1.20 && \
+  go mod vendor && \
+  hack/ci-utils/isClean.sh && \
+  make lint-go && \
+  ARO_RUN_PKI_TESTS=nope make unit-test-go && \
+  make validate-fips
+FROM ${REGISTRY}/ubi8/ubi-minimal
+RUN microdnf update && microdnf clean all
+COPY --from=builder /app/aro /app/e2e.test /usr/local/bin/
+ENTRYPOINT ["aro"]
+EXPOSE 2222/tcp 8080/tcp 8443/tcp 8444/tcp 8445/tcp
+USER 1000
+
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,9 @@ clean:
 client: generate
 	hack/build-client.sh "${AUTOREST_IMAGE}" 2020-04-30 2021-09-01-preview 2022-04-01 2022-09-04 2023-04-01 2023-07-01-preview 2023-09-04 2023-11-22 2024-08-12-preview
 
+ci-rp:
+	docker build -f Dockerfile.ci-rp --build-arg REGISTRY=$(REGISTRY) --no-cache=$(NO_CACHE)
+
 ci-portal:
 	docker build . -f Dockerfile.ci-portal --build-arg REGISTRY=$(REGISTRY) --no-cache=$(NO_CACHE)
 
@@ -273,4 +276,4 @@ vendor:
 install-go-tools:
 	go install ${GOTESTSUM}
 
-.PHONY: admin.kubeconfig aks.kubeconfig aro az ci-portal clean client deploy dev-config.yaml discoverycache generate image-aro-multistage image-fluentbit image-proxy init-contrib lint-go runlocal-rp proxy publish-image-aro-multistage publish-image-fluentbit publish-image-proxy secrets secrets-update e2e.test tunnel test-e2e test-go test-python vendor build-all validate-go unit-test-go coverage-go validate-fips install-go-tools
+.PHONY: admin.kubeconfig aks.kubeconfig aro az ci-portal ci-rp clean client deploy dev-config.yaml discoverycache generate image-aro-multistage image-fluentbit image-proxy init-contrib lint-go runlocal-rp proxy publish-image-aro-multistage publish-image-fluentbit publish-image-proxy secrets secrets-update e2e.test tunnel test-e2e test-go test-python vendor build-all validate-go unit-test-go coverage-go validate-fips install-go-tools


Dockerfile.ci-rp
```diff
@@ -0,0 +1,38 @@
+ARG REGISTRY
+FROM ${REGISTRY}/ubi8/go-toolset:1.20.10 as builder
+
+USER root
+ENV GOPATH=/root/go
+RUN mkdir -p /app
```
SudoBrendan, line 6: we can delete this; `WORKDIR` will create if it doesn't exist.

Dockerfile.ci-rp
```diff
@@ -0,0 +1,38 @@
...
+
+FROM builder as vendor-check
+WORKDIR /app
+COPY --from=builder /app/aro /app/e2e.test /usr/local/bin/
+RUN go mod vendor && go mod tidy -compat=1.18 && hack/ci-utils/isClean.sh
+
+FROM builder as lint-go
+WORKDIR /app
+COPY --from=builder /app/aro /app/e2e.test /usr/local/bin/
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/local/bin v1.55.2
```
SudoBrendan, line 19: We should put this before any `COPY` operations so we don't have to download/install this each time we make a code change. Can we put this at the top of the `builder` stage?

Dockerfile.ci-rp
```diff
@@ -0,0 +1,38 @@
...
+
+USER root
+ENV GOPATH=/root/go
+RUN mkdir -p /app
+WORKDIR /app
+COPY . /app
+RUN make aro RELEASE=${IS_OFFICIAL_RELEASE} -o generate
+
+FROM builder as vendor-check
+WORKDIR /app
```
SudoBrendan, line 12: since we use `FROM builder`, `WORKDIR` is already set - same for all below.

SudoBrendan (changes requested):
I'd recommend some quick fixes to make this more cache friendly and to reduce extra comands, great work so far, thank you!

Dockerfile.ci-rp
razo7, line 19: Linting is important in the build process, but wouldn't be easier to download the linter locally, add `lint-go` in the `ci-rp` target ([line 79](https://github.com/Azure/ARO-RP/pull/3489/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R79)) instead of doing that in the Dockerfile? 
IMO it would simplify the _Dockefile_, and make better code separation.
I am not sure where eventually the `ci-rp` target will be called. But it would be nice to test this target in a GitHub CI, this way you test linting per every PR in a CI image (e.g., Ubuntu).

Dockerfile.ci-rp
```diff
@@ -0,0 +1,38 @@
...
+COPY --from=builder /app/aro /app/e2e.test /usr/local/bin/
+RUN go mod vendor && go mod tidy -compat=1.18 && hack/ci-utils/isClean.sh
```
razo7, line 14: Any reason to check [compatibility](https://go.dev/ref/mod#go-mod-tidy) for v1.18 and not the latest Golang version in the repo? Golang [v1.20](https://github.com/Azure/ARO-RP/blob/master/go.mod#L3) was also used in [this file](https://github.com/Azure/ARO-RP/pull/3489/files#diff-d189eef96a9753a0ef4403084a625155b556091a2b336d98e20f24139b7ac3acR2).

Dockerfile.ci-rp
SudoBrendan: What you suggest is our current process (`make lint-go` already exists and expects this CLI to be installed locally). However, this has limitations:

1. This linter isn't the only dependency required - there are close to a dozen, so setup for a new dev can get incredibly difficult
2. It is difficult to get 50+ engineers in sync when we do things like bump the version of any of these tools

This is where this change comes in. They hope to make `podman` the only local dependency required to build and test our codebase. All you need day 1 and day 100 is `podman`, and we can easily keep everyone in sync with all our build dependencies with a `git pull`. It also ensures that every step of our code checks (builds/tests/linting/etc) happen locally the same way they happen during build processes.

Dockerfile.ci-rp
SudoBrendan: Good callout; the bump to 1.20 was a very recent change. @tony-schndr can you rebase and get this working for 1.20?

Dockerfile.ci-rp
tony-schndr: One of the goals behind this change was to have all CI "Just work" on developer machines, moving the lint out of the container may cause issues between dev setups.

Dockerfile.ci-rp
tony-schndr: bumped

Dockerfile.ci-rp
razo7: >What you suggest is our current process (make lint-go already exists and expects this CLI to be installed locally)

Correct, I was referring to adding the `lint-go` (or [`test-go`](https://github.com/Azure/ARO-RP/blob/master/Makefile#L208)) before building the container as you can see below. This way the test is done locally and not in the _Dockerfile.ci-rp_. 

```
ci-rp: lint-go
	docker build -f Dockerfile.ci-rp --build-arg REGISTRY=$(REGISTRY) --no-cache=$(NO_CACHE) --skip-unused-stages=false --jobs=0
```

I am not sure how the above limitations are relevant to the suggested change, but I do agree that it is "safer" to do lint in the Dockerfile rather than in the makefile target.
If it will be called mainly using the makefile target, then it seems pretty safe IMO

> One of the goals behind this change was to have all CI "Just work" on developer machines, moving the lint out of the container may cause issues between dev setups.

I am unaware of the usual dev setups that you folks do, but I guess the building procedure (multiple "tests" in the build process prior to the [actual container build](https://github.com/Azure/ARO-RP/pull/3489/files#diff-d189eef96a9753a0ef4403084a625155b556091a2b336d98e20f24139b7ac3acR27-R32)) in the Dockerfile has been discussed. I just wanted to share my two cents here.

razo7: >I am not sure where eventually the ci-rp target will be called. But it would be nice to test this target in a GitHub CI, this way you test linting per every PR in a CI image (e.g., Ubuntu)

When is the `ci-rp` target supposed to be called in dev workflow?

hawkowl (changes requested):
We shouldn't be running tests in the dockerfile build -- is there a reason we want to do it in the builds?

Dockerfile.ci-rp
tony-schndr: The idea here is that everything everywhere can just run these build containers from a make target.  So this would be ran in CI running in ADO pipelines, developers locally, etc...

tony-schndr:
> We shouldn't be running tests in the dockerfile build -- is there a reason we want to do it in the builds?

@hawkowl, the idea is that this container is used as CI everywhere and developers can run ci exactly the same locally as ADO pipelines would.  I have taken care to build this using multiple stages, the test stages are not used to produce the final image so there should be any excess "ci" layers added to the final image.

SudoBrendan:
While a great end goal, I think for now with the [state of asynchronous builds in podman](https://github.com/Azure/ARO-RP/pull/3525), that we should stick to sequential builds, and rely on build caching for optimization in the meantime. Not optimal, but still better than what we have today.

SudoBrendan:
> We shouldn't be running tests in the dockerfile build -- is there a reason we want to do it in the builds?

Yes. The reason is consistency everywhere, all the time across local, PR validation, and CI for production grade artifacts. Docker builds are uniquely positioned to give us consistency everywhere in that they give us things no other platform I'm aware of can do:

1. A consistent filesystem, with correct versions of all tools and dependencies. For example, whenever we bump our linter/golang/etc version, inevitably 10 unique slack threads crop up in FTS about "why is X broken?". Each of our teammates must then figure out what version they need to install locally, how to install it in their unique host system, as well as what might have broken across versions (config/etc). In a `docker build` world, this whole category of work goes away - the fix will always be "build again, but without caching"
2. docker gives us a build caching mechanism that enables us to create build processes in a way that are optimized for local development and fast feedback. E.g. if we just did `install the linter` as a makefile target or wrapped in a shell script, that install would happen every time unless we came up with an esoteric caching mechanism, and would not be guaranteed to work across our various dev envs.
3. Dockerfiles are now the single place you need to look for how we build/test/etc all the things. One file, with about a dozen lines, goes through in exact detail what needs to happen and in what order to accomplish a successfully validated and production grade artifact. You can't skip steps, you can't mix up versions, you can't diverge pipelines from local, etc. It guarantees consistency of build process in a way no other platform provides.

Dockerfile.ci-rp
```diff
@@ -0,0 +1,33 @@
...
+COPY --from=builder /app/aro /app/e2e.test /usr/local/bin/
+RUN make lint-go
+
+FROM builder as unit-test-go
+COPY --from=builder /app/aro /app/e2e.test /usr/local/bin/
+RUN ARO_RUN_PKI_TESTS=nope make unit-test-go
+
+FROM builder as validate-fips
+COPY --from=builder /app/aro /app/e2e.test /usr/local/bin/
+RUN make validate-fips
```
SudoBrendan, line 25:
```suggestion
RUN make lint-go \
    && make aro RELEASE=${IS_OFFICIAL_RELEASE} -o generate \
    && ARO_RUN_PKI_TESTS=nope make unit-test-go \
    && make e2e.test \
    && make validate-fips
```

Dockerfile.ci-rp
```diff
@@ -0,0 +1,33 @@
+ARG REGISTRY
+FROM ${REGISTRY}/ubi8/go-toolset:1.20.10 as builder
+
+USER root
```
SudoBrendan, line 4:
```suggestion
ARG REGISTRY

FROM ${REGISTRY}/ubi8/go-toolset:1.20.10 as builder
USER root
```

Makefile
```diff
@@ -76,6 +76,9 @@ clean:
 client: generate
 	hack/build-client.sh "${AUTOREST_IMAGE}" 2020-04-30 2021-09-01-preview 2022-04-01 2022-09-04 2023-04-01 2023-07-01-preview 2023-09-04 2023-11-22 2024-08-12-preview
 
+ci-rp:
+	docker build -f Dockerfile.ci-rp --build-arg REGISTRY=$(REGISTRY) --no-cache=$(NO_CACHE) --skip-unused-stages=false --jobs=0
```
SudoBrendan, line 80:
```suggestion
	docker build -f Dockerfile.ci-rp --build-arg REGISTRY=$(REGISTRY) --no-cache=$(NO_CACHE)
```

SudoBrendan (changes requested):
Added some suggestions to remove concurrency and avoid podman bugs. For me locally, an uncached build takes 5:51 - not great, but not terrible as a final "pre push" check either IMO or a way for someone to avoid waiting in line for hours for a pipeline agent.

We could definitely do more to improve build time and cache hits by modifying the `COPY . /app` line, and instead specify what files exactly are needed (for example, making a CLI and/or hack script change would invalidate the cache today)... but I think this can be a follow-up.

Another thing we may consider doing is removing any references to `make` in the containerfile, because we are in effect creating a possible loop: makefile > containerfile > makefile. It may be preferable and more clear for us to have the convention of makefile > containerfile > shell to remove complexity from the build process and make it very very clear what dependencies are needed to perform a build from a single file (dockerfile). To showcase what I mean - with this containerfile as written, can we say if our build uses the `/hack` directory or not? Again, not sure if this is the correct choice, and also probably out-of-scope and/or a follow-up PR.

One final note - our pipelines are eventually going to want to get access to the `report.xml` junit file for test results, so we'll need to consider how to get that out of the container and into the pipeline agent's filesystem - open to ideas here, but my guess is it'll involve a `podman cp` command.

SudoBrendan:
@tony-schndr with your permission I'd like to close this as a dupe of https://github.com/Azure/ARO-RP/pull/3558, I included your commit and kept you as the author in that PR alongside some other updates.

PR #3491, opened 2024-04-02
ARO-5325: Use cloud error whenever the run encounters an error so it …

edisonLcardenas:
…is propagated to user.

Fixes
https://issues.redhat.com/browse/ARO-5325

Propagate error to user.

### Test plan for issue:
Updated runner_test.go 

No, tech debt.

Unit test.

### DIFF
--- a/pkg/util/steps/runner.go
+++ b/pkg/util/steps/runner.go
@@ -56,7 +56,6 @@ func Run(ctx context.Context, log *logrus.Entry, pollInterval time.Duration, ste
 		err := step.run(ctx, log)
 
 		if err != nil {
-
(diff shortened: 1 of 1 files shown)


yjst2012:
thanks for the fix, you may want to fix the test failures caused by the err -> NewCloudError change, eg, in install_test.go.

mociarain:
FYI you'll need to move this to a branch off the main repo to get the CI to run: https://redhat-internal.slack.com/archives/C02ULBRS68M/p1710987980698209?thread_ts=1710774666.598769&cid=C02ULBRS68M

PR #3503, opened 2024-04-04
log outbound correlation data

AldoFusterTurpin:

Fixes [Log outbound requests correlationIds and/or clientRequestIds](https://issues.redhat.com/browse/ARO-3406) 

We encourage you to check the [SPIKE](https://docs.google.com/document/d/1RbnKKPNjw7kJZeR-2me4euM5CClGqF62FkaHVlWGxtM/edit?usp=sharing) we created to understand what this is about as it is not a trivial PR.

During investigations on RP issues, we faced difficulties to find out why some messages fail to be sent by the ARO-RP to ARM. Logs of messages (correlation data) sent by ARO RP to ARM are missing to help know what actually happened to the requests made by the RP, and in particular if they were actually sent to ARM.

The purpose of this is to have at least correlationID of outbound requests done by the ARO RP logged in ARO RP logs so that troubleshooting is possible for such complex cases.

The PR includes changes about the config of the log of outgoing requests to its own Geneva/Kusto table using a different marker like [this](https://github.com/Azure/ARO-RP/blob/a03c6ebd5f187b0028eb777098d89ee07c8cb06f/pkg/backend/openshiftcluster.go#L325), similar to HttpIncoming and Outgoing request tables in ARM log (modifcation of the fluentbit config in pkg/deploy/generator/scripts/rpVMSS.sh), also explained in the Spike document mentioned before.

### Test plan for issue:

Unit Tests, e2e tests and tested this PR in INT with the corresponding Geneva config changes.

The [spike document](https://docs.google.com/document/d/1RbnKKPNjw7kJZeR-2me4euM5CClGqF62FkaHVlWGxtM/edit?usp=sharing) contains detailed information that can be used as reference for specific topics as it was written in general terms


AldoFusterTurpin:
This PR was originally from my [Fork](https://github.com/Azure/ARO-RP/pull/2886), just for reference.

mociarain (dismissed):

mociarain:
FYI this is also going to conflict with this: https://github.com/Azure/ARO-RP/pull/3520

mociarain (approved):

bitoku:
Can we split this PR into smaller PRs? If we can, it can get more reviews.

pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -169,6 +169,18 @@ cat >/etc/fluentbit/fluentbit.conf <<'EOF'
...
+	Match journald
+	Rule $LOGKIND outboundRequests outboundRequests false
+
+[FILTER]
+	Name modify
+	Match  outboundRequests
+	Remove CLIENT_PRINCIPAL_NAME
+	Remove FILE
+	Remove COMPONENT
+
```
AldoFusterTurpin, line 183: FYI: Fluentbit config to log to a different Geneva table (explained in the Splike)

mociarain:
> Can we split this PR into smaller PRs? If we can, it can get more reviews.

IIRC I had the same thought when I first reviewed it :D I agree with the sentiment but I think the nature of this PR makes it hard.

pkg/util/azureclient/roundtripper.go
AldoFusterTurpin, line 1: FYI: We use this component in all the Go Azure clients to intercept HTTP requests and log useful info. (explained in the Spike).

AldoFusterTurpin:
> Can we split this PR into smaller PRs? If we can, it can get more reviews.

Hello Ayato, this is a good question. I do like small PRs, but in this case, it was pretty difficult to do so because of the nature of the problem. I thought about it in the past but my conclusion was that it would not be beneficial as this PR is the minimum chunk of related changes that make sense:
- Create a custom round tripper
- Modify all the Go clients with the new round tripper
- Update some fluentbit config to send the new logs to a different Geneva table

IMHO Any possible change makes things more confusing in this case without the overall picture.

I think that the Spike can help to understand what the PR is about.

Thank you.

pkg/util/azureclient/roundtripper.go
```diff
@@ -0,0 +1,81 @@
...
+
+	return l
+}
+
+func updateCorrelationDataAndEnrichLogWithResponse(correlationData *api.CorrelationData, l *logrus.Entry, res *http.Response, requestTime time.Time) *logrus.Entry {
+	if res == nil {
+		return l.WithFields(logrus.Fields{
+			"response_status_code":   "0",
+			"contentLength":          "-1",
+			"durationInMilliseconds": time.Since(requestTime).Milliseconds(),
```
bitoku, line 69: We should have consistency in the field naming.

https://github.com/Azure/ARO-RP/blob/cf51f260ac5611121d98655baee7c8ea34fd6eb8/pkg/backend/openshiftcluster.go#L334-L339

https://github.com/Azure/ARO-RP/blob/6d18e6449f661ad198f8c64f8f4f2de3011abfd5/pkg/frontend/middleware/log.go#L150-L156

Although there are already multiple rules, we should consistent in this file.

pkg/util/azureclient/roundtripper.go
```diff
@@ -0,0 +1,81 @@
...
+		})
+	}
+
+	correlationData.CorrelationID = res.Header.Get("X-Ms-Correlation-Request-Id")
```
bitoku, line 73: nit: this might be worth adding, or in the `CorrelationData` struct?

https://docs.google.com/document/d/1RbnKKPNjw7kJZeR-2me4euM5CClGqF62FkaHVlWGxtM/edit
>  Technically speaking, the client performs a request to ARM with x-ms-client-request-id, ARM receives the request and in the response, it adds the x-ms-correlation-request-id (the flow is client->ARM->ARO-RP->ARM) but it was simplified in Figure 1.

bitoku (commented):
Nice work and investigation!
Almost LGTM, but can we see some examples of what logs look like to get better understanding?

pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -169,6 +169,18 @@ cat >/etc/fluentbit/fluentbit.conf <<'EOF'
...
+[FILTER]
+	Name rewrite_tag
+	Match journald
+	Rule $LOGKIND outboundRequests outboundRequests false
+
+[FILTER]
+	Name modify
+	Match  outboundRequests
+	Remove CLIENT_PRINCIPAL_NAME
+	Remove FILE
```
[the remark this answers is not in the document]
AldoFusterTurpin: Yes, in that table, we don't need/want this field as we just want to see the correlation_id, the message, request time, etc but we are not interested in the file. This table works like some sort of "join table" to correlate inbound requests with outgoing ones so that info is not used 🙂

pkg/util/azureclient/roundtripper.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	utillog "github.com/Azure/ARO-RP/pkg/util/log"
+)
+
+// genevaOutboundRequestsTable is the table name we configure in Geneva
+// to send logs of outgoing requests from ARO-RP to ARM.
+// https://docs.google.com/document/d/1RbnKKPNjw7kJZeR-2me4euM5CClGqF62FkaHVlWGxtM/edit?usp=sharing
+const genevaOutboundRequestsTable = "outboundRequests"
```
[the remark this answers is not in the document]
AldoFusterTurpin: Sure. Done.

pkg/util/azureclient/roundtripper.go
AldoFusterTurpin: I agree with that, but I prefer the _durationInMilliseconds_ to make it explicit. For the consistency part, I totally agree, but those fields are used in Geneva tables and I prefer not to touch them to not break anything as this is big enough.

pkg/util/azureclient/roundtripper.go
AldoFusterTurpin: This footnote in the Spike exists because the picture with all the details was too big. This is just a comment to clarify that there should be one more "box" in the picture. As the Spike exists, I didn't want to put many comments in the code to not pollute it. 🙂

AldoFusterTurpin: I included the link to the Spike in pkg/util/azureclient/roundtripper.go to not loose the spike in the ocean of files we have and have it as a reference 🙂

AldoFusterTurpin:
> Nice work and investigation! Almost LGTM, but can we see some examples of what logs look like to get better understanding?

Thank you! 

In the Spike, there is a demo where I show what the logs look like in Geneva after the changes and how it can be used to correlate incoming requests with outgoing ones in a cluster installation in INT. I didn't want to put much non-related code stuff in the PR to not confuse people, the Spike is a good source of information about all that IMHO
🙂

pkg/util/azureclient/roundtripper.go
bitoku: This is just a nit so you don't have to reflect it, but I don't think leaving a small comment pollutes the code.
You don't have to explain all your design, but adding comments to some "magic" variables is useful as you did here.

https://github.com/Azure/ARO-RP/blob/4b06345c878558a06d4207db4ea3fd1c651f964f/pkg/util/azureclient/roundtripper.go#L16-L19

pkg/util/azureclient/roundtripper.go
```diff
@@ -0,0 +1,81 @@
...
+	})
+
+	return l
+}
+
+func updateCorrelationDataAndEnrichLogWithResponse(correlationData *api.CorrelationData, l *logrus.Entry, res *http.Response, requestTime time.Time) *logrus.Entry {
+	if res == nil {
+		return l.WithFields(logrus.Fields{
+			"response_status_code":   "0",
+			"contentLength":          "-1",
```
bitoku, line 68: Can we omit this field?
Because contentLength is quantity value different from status code, we should omit if we can.
Otherwise we need some effort to get statistic values from here.

pkg/util/azureclient/roundtripper.go
AldoFusterTurpin: I meant that I didn't want to change log fields in code that was already present (which I was not touching). 

I have now updated the fields for the new Round Tripper to be sneak case for consistency 🙂 
Thanks!

pkg/util/azureclient/roundtripper.go
AldoFusterTurpin: Sure. Done. thanks!

mociarain (approved):

PR #3510, opened 2024-04-09
Update dbtoken-service.md

anshulvermapatel:
The link was not pointing to where it was supposed to point. Corrected it.


mociarain (approved):
Good catch. (IIRC I did this in the first place :smile: )

Note: I would have expected this to fail the new branch name validation we have... hmmm :thinking:

rhamitarora (approved):

cadenmarchese (approved):

PR #3511, opened 2024-04-09
adding retry logic on yum commands to avoid resource locking

jhoreman:

This PR attempts to address the latest failure in the rpVMSS.sh script by adding retry logic to the yum update commands found therein. Failure can be observed in the ARO production pipeline e2e step.

Adds retry logic to yum update commands in rpVMSS.sh in order to allow for subsequent attempts if a resource lock is present,.

### Test plan for issue:

<!--

E2E will have to prove the logic allows for successful subsequent yum commands. 

No, essentially a knock-on fix of a recent change

rpVMSS.sh will complete successfully.


pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -5,11 +5,18 @@ echo "setting ssh password authentication"
 sed -i 's/PasswordAuthentication no/PasswordAuthentication yes/g' /etc/ssh/sshd_config
 systemctl reload sshd.service
 
-echo "running RHUI fix"
-yum update -y --disablerepo='*' --enablerepo='rhui-microsoft-azure*'
+#Adding retry logic to yum commands in order to avoid stalling out on resource locks
+for attempt in {1..5}; do
+  echo "running RHUI fix"
```
tsatam, line 10: nit: we should echo this line outside of the loop

pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -5,11 +5,18 @@ echo "setting ssh password authentication"
...
+  yum update -y --disablerepo='*' --enablerepo='rhui-microsoft-azure*'
+  if [[ ${attempt} -lt 5 ]]; then sleep 10; else exit 1; fi
+done
 
-echo "running yum update"
-yum -y -x WALinuxAgent -x WALinuxAgent-udev update --allowerasing
+for attempt in {1..5}; do
+  echo "running yum update"
```
tsatam, line 16: same change as above

pkg/deploy/generator/scripts/rpVMSS.sh
jhoreman: Agreed. Changed and ran make generate

pkg/deploy/generator/scripts/rpVMSS.sh
jhoreman: Agreed. Changed and ran make generate

s-fairchild (approved):
LGTM, legibility changes can be made in a follow up improvement PR.

hlipsig (approved):

PR #3512, opened 2024-04-10
Refactor/multiple ip using multiple channels.

AldoFusterTurpin:
**[WIP as I was planning to do a demo.]**

Co-authored with @tony-schndr.

Refactor PR to introduce some common Go concurrency patterns that the team could be interested in.

This PR shows a different approach to handle concurrency in the pkg/cluster/loadbalancerprofile.go file in regards to the creation of PublicIP Addresses (**createPublicIPAddresses**). 

This will serve as material in the next Azure Demos session to explain some Go concurrency concepts for those who are not aware of as of today but could be interested in (different people have different backgrounds).

I discussed these suggestions with @tony-schndr and we find it insightful enough to share it with the team to see a different approach to handle the same problem.

Some benefits of this approach:
- separation of clear concerns between the different small functions.
- a channel for errors and a channel for non error results to simplify the reasoning behind the code.
- waiting and reading from the channels decoupled from the number of writes (not explicit _for_ iterations).
- able to scale if new channels are needed to process more/different stuff. 

(I am working on a simple diagram to explain it)

### Test plan for issue:

Already Unit Tests in place, no behavior changed.

N/A

Already Unit Tests in place.

### DIFF
--- a/pkg/cluster/loadbalancerprofile.go
+++ b/pkg/cluster/loadbalancerprofile.go
@@ -9,6 +9,7 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
 
 	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
 
@@ -22,11 +23,6 @@ const outboundRuleV4 = "outbound-rule-v4"
 type deleteIPResult struct {
 	name string
 	err  error
-}
-
-type createIPResult struct {
-	ip  mgmtnetwork.PublicIPAddress
-	err error
 }
 
 func (m *manager) reconcileLoadBalancerProfile(ctx context.Context) error {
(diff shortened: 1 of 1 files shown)


pkg/cluster/loadbalancerprofile.go
```diff
@@ -24,11 +25,6 @@ type deleteIPResult struct {
 	err  error
 }
 
-type createIPResult struct {
-	ip  mgmtnetwork.PublicIPAddress
-	err error
-}
-
```
AldoFusterTurpin, line 31: No longer needed as we can use the ip type directly in the channel.

pkg/cluster/loadbalancerprofile.go
```diff
@@ -321,27 +317,51 @@ func getDesiredOutboundIPs(managedOBIPCount int, ipAddresses map[string]mgmtnetw
...
-	if len(errResults) > 0 {
-		return fmt.Errorf("failed to create required IPs\n%s", strings.Join(errResults, "\n"))
+func consumResults(resultCh <-chan mgmtnetwork.PublicIPAddress, errCh <-chan error, doneCh <-chan bool, ipAddresses map[string]mgmtnetwork.PublicIPAddress) error {
+	var errResults []string
+	for {
+		select {
+		case err := <-errCh:
+			if err != nil {
+				errResults = append(errResults, err.Error())
+			}
```
AldoFusterTurpin, line 353: we could do 
```go
case err, ok := <-errCh:
	if ok {
		errResults = append(errResults, err.Error())
	}
```
instead, but checking against nil (the zeroed value of error) is more straightforward and easy to read.

Same for the "ip := <-resultCh:" case.

pkg/cluster/loadbalancerprofile.go
```diff
@@ -321,27 +317,51 @@ func getDesiredOutboundIPs(managedOBIPCount int, ipAddresses map[string]mgmtnetw
...
+	resultCh := make(chan mgmtnetwork.PublicIPAddress)
+	errCh := make(chan error)
+	doneCh := make(chan bool)
 
+	m.processIPAddresses(ctx, numToCreate, resultCh, errCh, &wg)
+
+	go func() {
+		wg.Wait()
+		close(resultCh)
+		close(errCh)
```
bitoku, line 331: These channels should be closed after finishing `consumeResults`, I guess.

According to [this](https://gobyexample.com/closing-channels#:~:text=Reading%20from%20a%20closed%20channel,channel%20is%20closed%20and%20empty.)
> Reading from a closed channel succeeds immediately, returning the zero value of the underlying type.

Also accoding to [this](https://go.dev/tour/concurrency/5)
> A select blocks until one of its cases can run, then it executes that case. It chooses one at random if multiple are ready.

That means, while `resultCh` and `errCh` are closed and `doneCh` is not closed, `for` loop in `consumeResults` never stops and reads the zero value from those channels.
And also even after `doneCh` is closed, it's not guaranteed that it stops. 

I wrote a small example. https://go.dev/play/p/jXqoRK-5V4b
If you run the program a few times, you probably see there are some `0`s printed as output.

I suggest those `close(resultCh)` and `close(errCh)` go into `defer` block, and in this goroutine we should just close `doneCh`, which makes sure that after `doneCh` is handled in `consumeResults`, it closes the other two channels.

pkg/cluster/loadbalancerprofile.go
bitoku: If we defer closing `errCh` and `resultCh`, can we remove nil check, because it won't read the zero value from the channels anymore?

pkg/cluster/loadbalancerprofile.go
```diff
@@ -321,27 +317,51 @@ func getDesiredOutboundIPs(managedOBIPCount int, ipAddresses map[string]mgmtnetw
...
-		} else {
-			ipAddresses[*result.ip.Name] = result.ip
-		}
+		go m.createPublicIPAddress(ctx, resultCh, errCh, wg)
 	}
+}
 
-	if len(errResults) > 0 {
-		return fmt.Errorf("failed to create required IPs\n%s", strings.Join(errResults, "\n"))
+func consumResults(resultCh <-chan mgmtnetwork.PublicIPAddress, errCh <-chan error, doneCh <-chan bool, ipAddresses map[string]mgmtnetwork.PublicIPAddress) error {
```
bitoku, line 346:
```suggestion
func consumeResults(resultCh <-chan mgmtnetwork.PublicIPAddress, errCh <-chan error, doneCh <-chan bool, ipAddresses map[string]mgmtnetwork.PublicIPAddress) error {
```

PR #3513, opened 2024-04-10
Do not perform explicit cluster deletion in prod e2e

tsatam:

Fixes no Jira yet

Skips explicit cluster deletion when CI=true and RP_MODE!=development, and let deletion of the resource group take care of cluster deletion for us. This is required in production e2e contexts to avoid a race condition between us and ARM performing cluster deletion simultaneously and causing one of the requests to fail. 

### Test plan for issue:

- E2E should still work on this PR

No

- E2E should still work in releases

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -390,15 +390,23 @@ func (c *Cluster) Delete(ctx context.Context, vnetResourceGroup, clusterName str
 		if err != nil {
 			errs = append(errs, err)
 		}
-
-		c.log.Print("deleting cluster")
-		err = c.openshiftclustersv20200430.DeleteAndWait(ctx, vnetResourceGroup, clusterName)
-		if err != nil {
-			errs = append(errs, err)
-		}
 	}
 
 	if c.ci {
+		// Only perform explicit cluster deletion when in local development mode, otherwise
+		// we let the resource group deletion performed below clean up the cluster.
+		//
+		// When the cluster resource is known by ARM (e.g. prod e2e), deleting the cluster
+		// here can result in race conditions when deleting the resource group below,
+		// as ARM will also attempt to delete the cluster.
+		if env.IsLocalDevelopmentMode() {
+			c.log.Print("deleting cluster")
+			err = c.openshiftclustersv20200430.DeleteAndWait(ctx, vnetResourceGroup, clusterName)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/util/cluster/cluster.go
```diff
@@ -388,22 +388,33 @@ func (c *Cluster) Delete(ctx context.Context, vnetResourceGroup, clusterName str
...
-		c.log.Print("deleting cluster")
-		err = c.openshiftclustersv20200430.DeleteAndWait(ctx, vnetResourceGroup, clusterName)
-		if err != nil {
+			c.log.Errorf("error when deleting role assignments: %v", err)
 			errs = append(errs, err)
 		}
 	}
 
 	if c.ci {
+		// Only perform explicit cluster deletion when in local development mode, otherwise
```
niontive, line 397: Why doesn't deleting the resource group only work in local development mode?

tsatam: Discussed during sync call: The implementation here, with the two boolean flags (CI and localDevelopmentMode) leads to confusion on which flags are set and what behavior we expect for each permutation. We should follow-up this change with a refactor to make it more explicit and clear what steps are being performed in what contexts, even if it does lead to code duplication.

SudoBrendan (approved):
LGTM - nit: would it be better to use `%w` vs `%v` in all these cases? idk if it matters in a logging context or not?

tsatam:
> nit: would it be better to use `%w` vs `%v` in all these cases? idk if it matters in a logging context or not?

`%w` only has special behavior if within a `fmt.Errorf` call - it will implement the Unwrap method on the returned error that returns the error(s) included in any `%w`s specified, otherwise it behaves identically to `%v`. https://pkg.go.dev/fmt#Errorf 

Perhaps a better change here is to wrap the errors themselves using `fmt.Errorf` and both log and return the wrapped error, but without being 100% certain what downstream consumers of the error are expecting I did not want to make that change.

tsatam:
I pushed up an opinionated refactor to address some legibility/intent concerns - we can back this change out or defer it to a follow-up PR if desired.

mociarain (approved):

UlrichSchlueter (approved):
Might be good to switch the Prod E2E case to use the deleteCluster call again in a future PR. We don't test our cluster deletion code the way it is. 
Approving anyway to unblock the next release.

PR #3514, opened 2024-04-10
Add identityURL to internal apis for CMSI usage

cadenmarchese:

Fixes ARO-6086

The cluster doc needs to persist identityURL in order for Cluster MSI (CUAMSI) token refreshing to work. The identityURL is a header that is provided by ARM to the RP and is used for token refreshing purposes. For more info, see the design doc: https://docs.google.com/document/d/1dtgp6B-VYyXUmPsMX9f9MdlAE9sON4OuH1Cw9Ij0mmg/edit that @rajdeep wrote.

### Test plan for issue:

API calls should succeed in persisting identityURL header when a put or patch call to RP is made.

### DIFF
--- a/pkg/api/admin/openshiftcluster.go
+++ b/pkg/api/admin/openshiftcluster.go
@@ -442,6 +442,7 @@ type UserAssignedIdentities map[string]ClusterUserAssignedIdentity
 type Identity struct {
 	Type                   string                 `json:"type,omitempty"`
 	UserAssignedIdentities UserAssignedIdentities `json:"userAssignedIdentities,omitempty"`
+	IdentityURL            string                 `json:"identityURL,omitempty"`
 }
 
 // Install represents an install process.
--- a/pkg/api/admin/openshiftcluster.go
+++ b/pkg/api/admin/openshiftcluster.go
@@ -442,7 +442,6 @@ type UserAssignedIdentities map[string]ClusterUserAssignedIdentity
 type Identity struct {
 	Type                   string                 `json:"type,omitempty"`
 	UserAssignedIdentities UserAssignedIdentities `json:"userAssignedIdentities,omitempty"`
-	IdentityURL            string                 `json:"identityURL,omitempty"`
 }
 
 // Install represents an install process.
--- a/pkg/api/openshiftcluster.go
+++ b/pkg/api/openshiftcluster.go
@@ -803,4 +803,5 @@ type Identity struct {
 
 	Type                   string                 `json:"type,omitempty"`
 	UserAssignedIdentities UserAssignedIdentities `json:"userAssignedIdentities,omitempty"`
+	IdentityURL            string                 `json:"identityURL,omitempty"`
 }
--- a/pkg/api/openshiftcluster.go
+++ b/pkg/api/openshiftcluster.go
@@ -803,5 +803,5 @@ type Identity struct {
 
 	Type                   string                 `json:"type,omitempty"`
 	UserAssignedIdentities UserAssignedIdentities `json:"userAssignedIdentities,omitempty"`
-	IdentityURL            string                 `json:"identityURL,omitempty"`
+	IdentityURL            string                 `json:"identityURL,omitempty" mutable:"true"`
 }
--- a/pkg/frontend/openshiftcluster_putorpatch.go
+++ b/pkg/frontend/openshiftcluster_putorpatch.go
@@ -25,6 +25,8 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
+var errMissingIdentityURL error = fmt.Errorf("identityURL not provided but required for workload identity cluster")
+
 func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
@@ -41,10 +43,12 @@ func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Req
 	subId := chi.URLParam(r, "subscriptionId")
 	resourceProviderNamespace := chi.URLParam(r, "resourceProviderNamespace")
 
+	identityURL := r.Header.Get("x-ms-identity-url")
+
 	apiVersion := r.URL.Query().Get(api.APIVersionKey)
 	err := cosmosdb.RetryOnPreconditionFailed(func() error {
 		var err error
-		b, err = f._putOrPatchOpenShiftCluster(ctx, log, body, correlationData, systemData, r.URL.Path, originalPath, r.Method, referer, &header, f.apis[apiVersion].OpenShiftClusterConverter, f.apis[apiVersion].OpenShiftClusterStaticValidator, subId, resourceProviderNamespace, apiVersion)
+		b, err = f._putOrPatchOpenShiftCluster(ctx, log, body, correlationData, systemData, r.URL.Path, originalPath, r.Method, referer, &header, f.apis[apiVersion].OpenShiftClusterConverter, f.apis[apiVersion].OpenShiftClusterStaticValidator, subId, resourceProviderNamespace, apiVersion, identityURL)
 		return err
 	})
 
@@ -52,7 +56,7 @@ func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Req
 	reply(log, w, header, b, err)
 }
 
-func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.Entry, body []byte, correlationData *api.CorrelationData, systemData *api.SystemData, path, originalPath, method, referer string, header *http.Header, converter api.OpenShiftClusterConverter, staticValidator api.OpenShiftClusterStaticValidator, subId, resourceProviderNamespace string, apiVersion string) ([]byte, error) {
+func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.Entry, body []byte, correlationData *api.CorrelationData, systemData *api.SystemData, path, originalPath, method, referer string, header *http.Header, converter api.OpenShiftClusterConverter, staticValidator api.OpenShiftClusterStaticValidator, subId, resourceProviderNamespace string, apiVersion string, identityURL string) ([]byte, error) {
 	subscription, err := f.validateSubscriptionState(ctx, path, api.SubscriptionStateRegistered)
 	if err != nil {
 		return nil, err
@@ -86,9 +90,15 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 				},
 			},
 		}
+
 		if !f.env.IsLocalDevelopmentMode() /* not local dev or CI */ {
 			doc.OpenShiftCluster.Properties.FeatureProfile.GatewayEnabled = true
 		}
+	}
+
+	err = validateIdentityUrl(doc.OpenShiftCluster, identityURL, isCreate)
+	if err != nil {
+		return nil, err
 	}
 
 	doc.CorrelationData = correlationData
@@ -286,6 +296,26 @@ func enrichClusterSystemData(doc *api.OpenShiftClusterDocument, systemData *api.
 	if systemData.LastModifiedByType != "" {
 		doc.OpenShiftCluster.SystemData.LastModifiedByType = systemData.LastModifiedByType
 	}
+}
+
+func validateIdentityUrl(cluster *api.OpenShiftCluster, identityURL string, isCreate bool) error {
+	// Don't persist identity URL in non-wimi clusters
+	if cluster.Properties.ServicePrincipalProfile != nil || cluster.Identity == nil {
+		return nil
+	}
+
+	if identityURL == "" {
+		if isCreate {
+			return errMissingIdentityURL
+		}
+		return nil
+	}
+
+	if cluster.Identity != nil {
+		cluster.Identity.IdentityURL = identityURL
+	}
+
+	return nil
 }
 
 func (f *frontend) ValidateNewCluster(ctx context.Context, subscription *api.SubscriptionDocument, cluster *api.OpenShiftCluster, staticValidator api.OpenShiftClusterStaticValidator, ext interface{}, path string) error {
--- a/pkg/frontend/openshiftcluster_putorpatch.go
+++ b/pkg/frontend/openshiftcluster_putorpatch.go
@@ -311,9 +311,7 @@ func validateIdentityUrl(cluster *api.OpenShiftCluster, identityURL string, isCr
 		return nil
 	}
 
-	if cluster.Identity != nil {
-		cluster.Identity.IdentityURL = identityURL
-	}
+	cluster.Identity.IdentityURL = identityURL
 
 	return nil
 }
--- a/pkg/frontend/openshiftcluster_putorpatch_test.go
+++ b/pkg/frontend/openshiftcluster_putorpatch_test.go
@@ -3305,3 +3305,84 @@ func TestEnrichClusterSystemData(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateIdentityUrl(t *testing.T) {
+	for _, tt := range []struct {
+		name        string
+		identityURL string
+		cluster     *api.OpenShiftCluster
+		expected    *api.OpenShiftCluster
+		isCreate    bool
+		wantError   error
+	}{
+		{
+			name:        "identity URL is empty, is not wi/mi cluster create",
+			identityURL: "",
+			cluster:     &api.OpenShiftCluster{},
+			expected:    &api.OpenShiftCluster{},
+			isCreate:    false,
+		},
+		{
+			name:        "identity URL is empty, is wi/mi cluster create",
+			identityURL: "",
+			cluster:     &api.OpenShiftCluster{},
+			expected:    &api.OpenShiftCluster{},
+			isCreate:    true,
+			wantError:   errMissingIdentityURL,
+		},
+		{
+			name:        "cluster is not wi/mi, identityURL passed",
+			identityURL: "http://foo.bar",
+			cluster: &api.OpenShiftCluster{
+				Properties: api.OpenShiftClusterProperties{
+					ServicePrincipalProfile: &api.ServicePrincipalProfile{},
+				},
+			},
+			expected: &api.OpenShiftCluster{
+				Properties: api.OpenShiftClusterProperties{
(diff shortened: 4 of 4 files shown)


rajdeepc2792:
Left a few comments, all other changes seems good for identityURL.

.sha256sum
```diff
@@ -1,9 +1 @@
...
-e4e80ae293dce1a6acfde17fcbd1399487a2fa3587babe6bc69c4ebbdabaa570  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/stable/2022-04-01/redhatopenshift.json
-b1f1de0fe40d05de90742b17928968923b936adc294000f58974f50a297581dd  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/stable/2022-09-04/redhatopenshift.json
-01ba9562a8dac2824998ff0ad0d2465f79e6a66597bdb321e9409b9f2d12d222  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/stable/2023-04-01/redhatopenshift.json
-c023515341196746454c0ae7af077d40d3ec13f6b88b33cb558f0a7ab17a5a24  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/preview/2023-07-01-preview/redhatopenshift.json
-440748951dd1c3b34b5ccbdcb7cd966e3b89490887a1f1d64429561fad789515  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/stable/2023-09-04/redhatopenshift.json
-74a46fdde6ceb0121fe1515c7e11e902dd921b54cffe693307fb02b3dc88f26e  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/stable/2023-11-22/redhatopenshift.json
```
kimorris27, line 8: These SHA sums for the older API versions need to be added back in 😉

kimorris27 (changes requested):
The functionality to add the `IdentityURL` looks good. Two points to address:

- The SHA sums for older API versions need to be added back in.
- I think you could add unit test cases for the changes to `openshiftcluster_putorpatch.go`.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -276,6 +288,17 @@ func enrichClusterSystemData(doc *api.OpenShiftClusterDocument, systemData *api.
 	}
 }
 
+func (f *frontend) persistIdentityUrl(cluster *api.OpenShiftCluster, identityURL []string) error {
+	if identityURL == nil {
+		return nil
+	} else if len(identityURL) != 1 {
+		return fmt.Errorf("identityURL cannot contain multiple values: %v", identityURL)
+	}
+	cluster.Identity.IdentityURL = identityURL[0]
```
rajdeepc2792, line 297: This function relies on identityURL for checking if the cluster is MIWI or not, IMO, MIWI check can be done by a nil-check on `cluster.Identity`.
And error is returned if identityURL is nil or have more than 1 values.
Alternative approach can be one from the [Preview API design](https://docs.google.com/document/d/1AIuf6H9y0Jjboywp0NgD8-9lU1jCj5WrE83jh58rEg4/edit#heading=h.ds2n08h4tc02) too
WDYT?

pkg/frontend/openshiftcluster_putorpatch_test.go
```diff
@@ -3300,3 +3300,137 @@ func TestEnrichClusterSystemData(t *testing.T) {
...
+		expected    *api.OpenShiftCluster
+		wantError   string
+	}{
+		{
+			name:        "pass - identity URL is nil",
+			identityURL: nil,
+			expected:    &api.OpenShiftCluster{},
+			wantError:   "",
+		},
+		// {
```
rajdeepc2792, line 3401: This test fails because when `persistIdentityUrl` is called value for `Identity` in cluster object is nil. Thus it fails for `IdentityURL` assignment here:- `cluster.Identity.IdentityURL = identityURL[0]`

It would be a good idea to perform a MIWI cluster check on the basis of `PlatformWorkloadIdentityProfile` or `cluster.Identity`. And a test case for cluster object without Identity profile and the for already existing in this file should have the Identity profile in the cluster object like:-
```
cluster := &api.OpenShiftCluster{
    Identity: &api.Identity{},
}
```

rajdeepc2792 (changes requested):
Left few comments on the test cases and identityURL header const value.

pkg/api/admin/openshiftcluster.go
```diff
@@ -441,6 +441,7 @@ type UserAssignedIdentities map[string]ClusterUserAssignedIdentity
 type Identity struct {
 	Type                   string                 `json:"type,omitempty"`
 	UserAssignedIdentities UserAssignedIdentities `json:"userAssignedIdentities,omitempty"`
+	IdentityURL            string                 `json:"identityURL,omitempty"`
```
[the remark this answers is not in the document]
cadenmarchese: Would it ever be useful in a Geneva Action context? I think that would be the use case for keeping it admin API, but I'm not sure of others.

cadenmarchese:
I'm adding `hold` while we work on ARO-6623 which is required to resolve the open comments.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -288,6 +300,22 @@ func enrichClusterSystemData(doc *api.OpenShiftClusterDocument, systemData *api.
...
+	if len(identityURL) != 1 {
+		return fmt.Errorf("identityURL cannot contain multiple values: %v", identityURL)
+	}
+
+	if cluster.Identity != nil && cluster.Properties.ServicePrincipalProfile == nil {
+		cluster.Identity.IdentityURL = identityURL[0]
+	}
+
+	return nil
+}
```
kimorris27, line 317: I think the code will proceed without erroring if the cluster is MIWI and no IdentityURL is passed. Is that intentional?

cadenmarchese: I think we don't want to block an update on having that header provided or not provided, just persist when provided. What if MSI-RP behavior around secret headers changes in the future? I'm open to opinions on that.

kimorris27: Ensuring that we don't block updates makes sense, but what about creates? Is the IdentityURL necessary there?

cadenmarchese: If cluster is wi/mi and no identityURL is provided by ARM that would fail an installation because the RP wouldn't have permissions to federate identities. I can think of how to represent that in this logic and adjust things.

kimorris27 (dismissed):

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -41,18 +42,23 @@ func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Req
 	subId := chi.URLParam(r, "subscriptionId")
 	resourceProviderNamespace := chi.URLParam(r, "resourceProviderNamespace")
 
+	identityUrlHeader := "x-ms-identity-url"
```
s-amann, line 45: nit: did a quick search and we don't generally create constants for headers we set/get, might be worth considering as a later refactoring effort.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -31,6 +31,7 @@ func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Req
 
 	var header http.Header
 	var b []byte
+	var identityURL []string
```
s-amann, line 34: nit: considering this is an array, might be better named `identityURLs`

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -288,6 +300,28 @@ func enrichClusterSystemData(doc *api.OpenShiftClusterDocument, systemData *api.
 	}
 }
 
+func persistIdentityUrl(cluster *api.OpenShiftCluster, identityURL []string, isCreate bool) error {
```
s-amann, line 303: nit: since this function isn't actually persisting data, perhaps it could be renamed `validateIdentityURL` and return the valid `identityURL` to then be set on `doc.OpenShiftCluster` instead of passing it in as a side effect.

s-amann (dismissed):
minor nits, nothing critical. lgtm

pkg/api/admin/openshiftcluster.go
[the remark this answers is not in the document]
niontive: To clarify:
- We need to persist to fetch the credentials object
- After we fetch the credentials object, we store in key vault. That object has the identity URL (field ClientSecretURL): https://github.com/Azure/msi-dataplane/blob/557894f3863f5591445903674cc6ba90e2a26fb7/pkg/dataplane/zz_generated_models.go#L21
- Credential refresher will use the URL stored in key vault for fetching future credentials objects.

There are a few gotchas for line 445:
1. MSI team reserves the right to change the identity URL after refresh, so if we persist this guy, we should at least have a code comment clarifying when and when not to use this field. https://eng.ms/docs/products/arm/rbac/managed_identities/msicredentialshandlingguideline
2. Instead of relying on Geneva Action to see if this value got persisted, couldn't we just rely on log statements: (1) did we get a valid identity URL and (2) did we persist it

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -41,18 +42,23 @@ func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Req
 	subId := chi.URLParam(r, "subscriptionId")
 	resourceProviderNamespace := chi.URLParam(r, "resourceProviderNamespace")
 
+	identityUrlHeader := "x-ms-identity-url"
+	if r.Header.Get(identityUrlHeader) != "" {
```
niontive, line 46: `Header.Get()` returns the first value associated with the key. Since we expect only one key, we should be good to just pass a string to `_putOrPatchOpenShiftCluster` instead of a slice of strings.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -41,18 +42,23 @@ func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Req
...
+		identityURL = r.Header.Values(identityUrlHeader)
+	}
+
 	apiVersion := r.URL.Query().Get(api.APIVersionKey)
 	err := cosmosdb.RetryOnPreconditionFailed(func() error {
 		var err error
-		b, err = f._putOrPatchOpenShiftCluster(ctx, log, body, correlationData, systemData, r.URL.Path, originalPath, r.Method, referer, &header, f.apis[apiVersion].OpenShiftClusterConverter, f.apis[apiVersion].OpenShiftClusterStaticValidator, subId, resourceProviderNamespace, apiVersion)
+		b, err = f._putOrPatchOpenShiftCluster(ctx, log, body, correlationData, systemData, r.URL.Path, originalPath, r.Method, referer, &header, f.apis[apiVersion].OpenShiftClusterConverter, f.apis[apiVersion].OpenShiftClusterStaticValidator, subId, resourceProviderNamespace, apiVersion, identityURL)
```
niontive, line 53: Out of scope for this PR but just leaving this here:
Refactor this function to not accept so many params

pkg/frontend/openshiftcluster_putorpatch_test.go
```diff
@@ -3305,3 +3305,107 @@ func TestEnrichClusterSystemData(t *testing.T) {
...
+			isCreate:    false,
+			wantError:   "",
+		},
+		{
+			name:        "identity URL is nil, is wi/mi cluster create",
+			identityURL: nil,
+			cluster:     &api.OpenShiftCluster{},
+			expected:    &api.OpenShiftCluster{},
+			isCreate:    true,
+			wantError:   "identityURL not provided but required for workload identity cluster",
```
niontive, line 3332: Nit: in `openshiftcluster_putorpatch.go`, define an error there so we don't have to repeat the same string twice

pkg/frontend/openshiftcluster_putorpatch.go
cadenmarchese: I tried this, but it reduces the scope of the unit test coverage, since we're no longer testing that the correct identity URL was set in the cluster document and only testing that the identity URL was validated correctly.

pkg/frontend/openshiftcluster_putorpatch.go
cadenmarchese: I'm going to go with this suggestion: https://github.com/Azure/ARO-RP/pull/3514#discussion_r1603751020 which should resolve this comment.

pkg/api/openshiftcluster.go
```diff
@@ -803,4 +803,5 @@ type Identity struct {
 
 	Type                   string                 `json:"type,omitempty"`
 	UserAssignedIdentities UserAssignedIdentities `json:"userAssignedIdentities,omitempty"`
+	IdentityURL            string                 `json:"identityURL,omitempty"`
```
rajdeepc2792, line 806: Again this is the candidate for being `Mutable` here in InternalAPI only, as the field would be changed to an empty string after the bootstrap certificates are fetched.
Although that can be done when the story around the certificate fetching is performed later.
WDYT?

pkg/api/admin/openshiftcluster.go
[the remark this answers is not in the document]
rajdeepc2792: Yes, it's better to rely on the Service Logs for this purpose, and this will only be faced for the Cluster Installs. Since, this field would be anyways removed after the bootstrap MSI certificate fetching.

pkg/api/openshiftcluster.go
rajdeepc2792: Also will it fail here:- https://github.com/Azure/ARO-RP/pull/3514/files#diff-c0dceb05e2a59300f2e11380f3601ea682171f3f4837bb769689fc79e78b1a70R315
If the field is not `Mutable` for the internalAPI?

rajdeepc2792 (commented):
Commented the doubts, rather than any change request.
Other than the concerns, changes LGTM.

pkg/api/openshiftcluster.go
cadenmarchese: I don't think so, since we're bailing out of the function early if the identity field isn't present: https://github.com/Azure/ARO-RP/pull/3514/files#diff-c0dceb05e2a59300f2e11380f3601ea682171f3f4837bb769689fc79e78b1a70R303 

Although, I could be wrong on that.

pkg/api/admin/openshiftcluster.go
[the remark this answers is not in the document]
cadenmarchese: I'll remove this from the admin API for now.

pkg/api/openshiftcluster.go
cadenmarchese: I think I understand now - the concern is that we might not be able to modify IdentityURL once populated without that tag. I can add it since the assumption is that it's needed for fields that change (such as ServicePrincipalProfile, whose subfields use the mutable tag).

kimorris27: Since we're talking about the internal API here, I don't think this tag is relevant; it can be present or absent and it will never be checked. The only place this tag gets checked is during static validation. See this function and its doc comment: https://github.com/Azure/ARO-RP/blob/master/pkg/api/util/immutable/immutable.go#L21-L24. The `mutable` tag is a thing we added to our codebase and not a Go thing.

I do see some fields with the `mutable` tag in the internal API though, so... keep it just to be safe?

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -288,6 +298,26 @@ func enrichClusterSystemData(doc *api.OpenShiftClusterDocument, systemData *api.
...
+	}
+
+	if identityURL == "" {
+		if isCreate {
+			return errMissingIdentityURL
+		}
+		return nil
+	}
+
+	if cluster.Identity != nil {
```
kimorris27, line 314: This conditional is unneeded since we exit the function early on line 303 if the `Identity` is `nil`.

kimorris27 (changes requested):
Requesting one more small change for readability's sake.

rajdeepc2792 (approved):

rajdeepc2792 (approved):

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -86,11 +90,17 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 				},
 			},
 		}
+
 		if !f.env.IsLocalDevelopmentMode() /* not local dev or CI */ {
 			doc.OpenShiftCluster.Properties.FeatureProfile.GatewayEnabled = true
 		}
 	}
 
+	err = validateIdentityUrl(doc.OpenShiftCluster, identityURL, isCreate)
```
niontive, line 99: Nit: in general, it's not good to have flag arguments: https://martinfowler.com/bliki/FlagArgument.html

So maybe to follow the paradigm in this function, we can change validation based on `isCreate`:

```
if isCreate {
...
} else {
...
}
```

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -41,18 +43,20 @@ func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Req
 	subId := chi.URLParam(r, "subscriptionId")
 	resourceProviderNamespace := chi.URLParam(r, "resourceProviderNamespace")
 
+	identityURL := r.Header.Get("x-ms-identity-url")
```
niontive, line 46: I'll add a todo in msi dataplane repo to make this an exportable const: https://github.com/Azure/msi-dataplane/blob/557894f3863f5591445903674cc6ba90e2a26fb7/pkg/dataplane/constants.go#L13

kimorris27 (approved):

pkg/frontend/openshiftcluster_putorpatch.go
cadenmarchese: Do you think it would be better to split validateIdentityUrl into two functions? I ask because either way, we will need the `isCreate` logic in place as you pointed out, it's just a matter of where it goes.

hlipsig: I'd like us to clean this up before GA (or soon after) but not hold up overall progress. Since this a current norm in our codebase we can acknowledge that it should be improved and move on to mission critical pieces. A good fist issue for a new joiner would be this refactor.

hlipsig (approved):

PR #3493, opened 2024-04-02
MDM/MDSD/Fluentbit Image Bumps

tsatam:
### Which issue this PR addresses:

Part of [ARO-6628](https://issues.redhat.com/browse/ARO-6628)

### What this PR does / why we need it:

Updates images for MDM/MDSD/Fluentbit utilized by the RP infra and by clusters to the latest supported versions. 

Details about specific coordinates included in the story card linked above. 

### Test plan for issue:

- [ ] Unit tests should pass
- [ ] E2E tests should pass

### Is there any documentation that needs to be updated for this PR?

N/A 

### How do you know this will function as expected in production? 

Tests above ensure the specified versions function as expected within _most_ contexts - we will need to pay close attention when releasing the RP with this version to ensure things are not broken in int before pushing forward.

PR #3516, opened 2024-04-11
Specify the correct MDSD image

mociarain:
This PR specifies the correct image for mirroring that we use. 

I think with all the bumps we missed it in here: https://github.com/Azure/ARO-RP/pull/3493


tiguelu (approved):
LGTM. Waiting for CI tests to complete.

PR #3517, opened 2024-04-11
Use the correct MDSD image

mociarain:
- Revert changes introduced here: https://github.com/Azure/ARO-RP/pull/3516
- Specify a "good" image for MDSD that is present in our CR.


tiguelu (approved):
LGTM

tsatam (approved):

PR #3518, opened 2024-04-11
add export in createapp.env

bitoku:

N/A

The [doc](https://github.com/Azure/ARO-RP/blob/master/docs/deploy-development-rp.md#run-the-rp-and-create-a-cluster) says we need to run `source clusterapp.env` but it doesn't work because `clusterapp.env` doesn't have `export` with each env, and the env variables aren't passed to `hack/cluster create`.

### Test plan for issue:

Test it locally

N/A
Change the code according to the doc.

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -152,7 +152,7 @@ func (c *Cluster) CreateApp(ctx context.Context, clusterName string) error {
 		return err
 	}
 
-	return os.WriteFile("clusterapp.env", []byte(fmt.Sprintf("AZURE_CLUSTER_SERVICE_PRINCIPAL_ID=%s\nAZURE_CLUSTER_APP_ID=%s\nAZURE_CLUSTER_APP_SECRET=%s", spID, appID, appSecret)), 0o600)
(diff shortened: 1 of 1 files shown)


pkg/util/cluster/cluster.go
```diff
@@ -152,7 +152,7 @@ func (c *Cluster) CreateApp(ctx context.Context, clusterName string) error {
 		return err
 	}
 
-	return os.WriteFile("clusterapp.env", []byte(fmt.Sprintf("AZURE_CLUSTER_SERVICE_PRINCIPAL_ID=%s\nAZURE_CLUSTER_APP_ID=%s\nAZURE_CLUSTER_APP_SECRET=%s", spID, appID, appSecret)), 0o600)
+	return os.WriteFile("clusterapp.env", []byte(fmt.Sprintf("export AZURE_CLUSTER_SERVICE_PRINCIPAL_ID=%s\nexport AZURE_CLUSTER_APP_ID=%s\nexport AZURE_CLUSTER_APP_SECRET=%s", spID, appID, appSecret)), 0o600)
```
AldoFusterTurpin, line 155: Now that we are touching this line, it could be a good idea to make a small change and replace this single line with something easier to read (the current line is long and contains multiple parentheses)
```go
content := fmt.Sprintf("export AZURE_CLUSTER_SERVICE_PRINCIPAL_ID=%s\nexport AZURE_CLUSTER_APP_ID=%s\nexport AZURE_CLUSTER_APP_SECRET=%s", spID, appID, appSecret)
contentBytes := []byte(content)
return os.WriteFile("clusterapp.env", contentBytes, 0o600)
```

AldoFusterTurpin (approved):
LGTM! Thanks. I left an optional simple suggestion to improve the readability as that line is really long

tiguelu (approved):

PR #3519, opened 2024-04-12
Change the upgrademode of VMSS to rolling

nwnt:

Fixes: ARO-6610 (Upgrade Policy Mode part)

AzSecPack team policy requires that we don't use `Manual` upgrade policy.  Automatic shouldn't be used as it can introduce disruption and is not suitable for production workload.  Changing this to rolling also means accepting the default values of the [RollingUpgradePolicy](https://learn.microsoft.com/en-us/azure/templates/microsoft.compute/2020-12-01/virtualmachinescalesets?pivots=deployment-language-arm-template#rollingupgradepolicy-1), namely 20% maxBatchInstancePercent, maxUnhealthyInstancePercent, maxUnhealthyUpgradeInstancePercent and 0 pauseTimeBetweenBatches.

### Test plan for issue:
VMSSes (rp, gateway) can still be created and have upgradepolicy set to `rolling`

N/A

VMSSes (rp, gateway) can still be created and have upgradepolicy set to `rolling`

### DIFF
--- a/pkg/deploy/assets/env-development.json
+++ b/pkg/deploy/assets/env-development.json
@@ -224,7 +224,7 @@
             },
             "properties": {
                 "upgradePolicy": {
-                    "mode": "Manual"
+                    "mode": "Rolling"
                 },
                 "virtualMachineProfile": {
                     "osProfile": {
--- a/pkg/deploy/assets/gateway-production.json
+++ b/pkg/deploy/assets/gateway-production.json
@@ -226,7 +226,7 @@
             },
             "properties": {
                 "upgradePolicy": {
-                    "mode": "Manual"
+                    "mode": "Rolling"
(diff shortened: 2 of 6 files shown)


hlipsig (approved):

cadenmarchese (approved):
makes sense, thanks!

kimorris27 (approved):

tsatam:
I deployed this PR to int, and the deployment succeeded.

rajdeepc2792 (approved):

SudoBrendan (approved):
LGTM. To be clear, our software now must support N-1 versions, correct? If this is the case, this is worth an email at large to the team.

UlrichSchlueter (approved):
LGTM

PR #3255, opened 2023-10-31
Migrate documentdb client from sdk track 1 to track 2 client

nwnt:
### Which issue this PR addresses:

Fixes: ARO-4565

### What this PR does / why we need it:
Making a progress towards migrating to track 2 go sdk clients.

### Test plan for issue:

Unit tests, E2E test and INT test

### Is there any documentation that needs to be updated for this PR?
N/A

PR #3520, opened 2024-04-12
Setting TLS1.2 as min for cosmoDB

sankur-codes:

Fixes [ARO-6842](https://issues.redhat.com/browse/ARO-6842)

This PR set TLS 1.2 as minimum requirement for cosmodb as required by Microsoft Compliance 

### Test plan for issue:

Tested using full-rp-dev-setup. Details added in JIRA ticket. 

NA

Tested using full-rp-dev-setup. Details added in JIRA ticket.

### DIFF
--- a/pkg/deploy/assets/databases-development.json
+++ b/pkg/deploy/assets/databases-development.json
@@ -11,160 +11,146 @@
     },
     "resources": [
         {
+            "apiVersion": "2023-04-15",
+            "location": "[resourceGroup().location]",
+            "name": "[concat(parameters('databaseAccountName'), '/', parameters('databaseName'))]",
             "properties": {
-                "resource": {
-                    "id": "[parameters('databaseName')]"
-                },
                 "options": {
                     "autoscaleSettings": {
                         "maxThroughput": 1000
                     }
+                },
+                "resource": {
+                    "id": "[parameters('databaseName')]"
                 }
             },
-            "name": "[concat(parameters('databaseAccountName'), '/', parameters('databaseName'))]",
-            "type": "Microsoft.DocumentDB/databaseAccounts/sqlDatabases",
-            "location": "[resourceGroup().location]",
-            "apiVersion": "2021-01-15"
+            "type": "Microsoft.DocumentDB/databaseAccounts/sqlDatabases"
         },
         {
+            "apiVersion": "2023-04-15",
+            "location": "[resourceGroup().location]",
+            "name": "[concat(parameters('databaseAccountName'), '/', parameters('databaseName'), '/AsyncOperations')]",
             "properties": {
+                "options": {},
                 "resource": {
+                    "defaultTtl": 604800,
                     "id": "AsyncOperations",
                     "partitionKey": {
+                        "kind": "Hash",
                         "paths": [
                             "/id"
-                        ],
-                        "kind": "Hash"
-                    },
-                    "defaultTtl": 604800
-                },
-                "options": {}
+                        ]
+                    }
+                }
             },
-            "name": "[concat(parameters('databaseAccountName'), '/', parameters('databaseName'), '/AsyncOperations')]",
-            "type": "Microsoft.DocumentDB/databaseAccounts/sqlDatabases/containers",
-            "location": "[resourceGroup().location]",
-            "apiVersion": "2021-01-15",
-            "dependsOn": [
-                "[resourceId('Microsoft.DocumentDB/databaseAccounts/sqlDatabases', parameters('databaseAccountName'), parameters('databaseName'))]"
-            ]
+            "type": "Microsoft.DocumentDB/databaseAccounts/sqlDatabases/containers"
         },
         {
+            "apiVersion": "2023-04-15",
+            "location": "[resourceGroup().location]",
+            "name": "[concat(parameters('databaseAccountName'), '/', parameters('databaseName'), '/OpenShiftVersions')]",
             "properties": {
+                "options": {},
                 "resource": {
+                    "defaultTtl": -1,
                     "id": "OpenShiftVersions",
                     "partitionKey": {
+                        "kind": "Hash",
                         "paths": [
                             "/id"
-                        ],
-                        "kind": "Hash"
-                    },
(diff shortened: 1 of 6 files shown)


pkg/deploy/assets/databases-development.json
```diff
@@ -11,160 +11,146 @@
     },
```
mociarain, line 11: The JSON diff is quite confusing. Was it sorted and if so can you revert that?

Huge +1 alphabetical sorting but in a separate PR

pkg/deploy/generator/resources_rp.go
```diff
@@ -8,8 +8,8 @@ import (
 	"fmt"
 	"strings"
 
+	sdkcosmos "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/cosmos/armcosmos/v2"
 	mgmtcompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
-	mgmtdocumentdb "github.com/Azure/azure-sdk-for-go/services/cosmos-db/mgmt/2021-01-15/documentdb"
```
mociarain, line 12: We are still using this "old" library in 2 other places. Do we need to change those too?

mociarain (commented):
Nice change just a few questions to help with the review

pkg/deploy/assets/databases-development.json
sankur-codes: The json is generated programatically with `make generate`. I think it isn't specifically sorted.

pkg/deploy/generator/resources_rp.go
sankur-codes: The reason for using sdk track 2 here is because the required parameters for this change isn't available in sdk track 1. I suppose, changes in rest of the places would be better done in a seperate PR that would be specifically for migration from sdk track 1 to sdk track 2.

hlipsig (approved):

pkg/deploy/generator/resources_rp.go
hlipsig: We still need to do track 2 migration in several places. I know Dan was tyring to track (ha pun) that down a month or so ago but I haven't had a chance to look deeply into it.

pkg/util/azureclient/apiversions.go
```diff
@@ -19,7 +19,7 @@ var apiVersions = map[string]string{
 	"microsoft.compute/snapshots":              "2020-05-01",
 	"microsoft.containerregistry":              "2020-11-01-preview",
 	"microsoft.resources/deployments":          "2021-04-01",
-	"microsoft.documentdb":                     "2021-01-15",
+	"microsoft.documentdb":                     "2023-04-15",
```
anshulvermapatel, line 22: How did you come up with this apiversion?
I just checked this [Documentation Page](https://learn.microsoft.com/en-us/azure/templates/microsoft.documentdb/databaseaccounts?pivots=deployment-language-arm-template) and didn't see this particular version there.

pkg/deploy/generator/resources_rp.go
```diff
@@ -978,25 +978,31 @@ func (g *generator) rpServiceKeyvault() *arm.Resource {
...
-		Kind: mgmtdocumentdb.GlobalDocumentDB,
-		DatabaseAccountCreateUpdateProperties: &mgmtdocumentdb.DatabaseAccountCreateUpdateProperties{
-			ConsistencyPolicy: &mgmtdocumentdb.ConsistencyPolicy{
-				DefaultConsistencyLevel: mgmtdocumentdb.Strong,
-			},
-			Locations: &[]mgmtdocumentdb.Location{
+	dbType := sdkcosmos.DatabaseAccountKindGlobalDocumentDB
+	consistency := sdkcosmos.DefaultConsistencyLevelStrong
+	backupPolicy := sdkcosmos.BackupPolicyTypePeriodic
+	minTLSVersion := sdkcosmos.MinimalTLSVersionTls12
```
anshulvermapatel, line 984: nit: since you are using constants here from the library, I think you don't have to put them in variables explicitly, you can directly put them in the following assignments.

pkg/deploy/generator/resources_rp.go
anshulvermapatel: There was this [PR](https://github.com/Azure/ARO-RP/pull/3255) which migrated documentdb client to track 2 sdk, it didn't cover this part though. A separate PR for that would be a better option.

anshulvermapatel (commented):
Awesome work. I have just put in some comments.

pkg/util/azureclient/apiversions.go
sankur-codes: There is a drop down menu for API Versions which has this version mentioned. 
[Here](https://learn.microsoft.com/en-us/azure/templates/microsoft.documentdb/2023-04-15/databaseaccounts?pivots=deployment-language-arm-template) you go.

pkg/deploy/generator/resources_rp.go
sankur-codes: Agree. However, the cosmodb object being created were already nested(ugly) so to make it simple to understand I have assigned values to variables and used them instead of directly getting values in the object creation. (Also, to match the way `hashPartitionKey` is being used in other different objects.)

anshulvermapatel: If you see on other func definitions like [this](https://github.com/Azure/ARO-RP/blob/1be3f99fdac41a45dc6b0494487f427922b748d6/pkg/deploy/generator/resources_rp.go#L103-L111) , you see that the consts are directly placed while creating/defining an object.
Having variables is a good thing, however in this particular case, someone looking at the object will have to again look up to know the variable initialisation. 
I still think you should put the consts directly rather than through variables.

anshulvermapatel (approved):
LGTM

pkg/deploy/generator/resources_rp.go
sankur-codes: Discussed & agreed upon necessity of variables over call.

PR #3522, opened 2024-04-15
MIWI preview API - Add PlatformWorkloadIdentityRoleSet proxy resource type 

kimorris27:

https://issues.redhat.com/browse/ARO-6082

This PR adds API types, converters, and validators for the new `PlatformWorkloadIdentityRoleSet` proxy resource that's being added to support the MIWI feature.

**Note:** I commented out the admin API converter and validator because the linter didn't like them to exist without being used... is it okay to merge them like this for now? This PR is going to grow very large if we add in all of the stuff that will use this converter and validator.

### Test plan for issue:

These API changes will have to be tested once we have created a Cosmos DB container to store the new proxy resources.

N/A

Unit tests and other tests as applicable will be implemented or performed as part of later tickets once the actual new functionality has been added to the RP.

### DIFF
--- a/pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
+++ b/pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
@@ -1,0 +1,67 @@
+package admin
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+/*
+TODO: Uncomment once API endpoints have been implemented and this code is being used.
+
+type platformWorkloadIdentityRoleSetStaticValidator struct{}
+
+func (sv platformWorkloadIdentityRoleSetStaticValidator) Static(_new interface{}, _current *api.PlatformWorkloadIdentityRoleSet) error {
+	new := _new.(*PlatformWorkloadIdentityRoleSet)
+
+	var current *PlatformWorkloadIdentityRoleSet
+	if _current != nil {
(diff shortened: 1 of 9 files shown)


cadenmarchese (approved):
This looks good to me and is in line with the way I understand your design doc 👍 
committing the commented code isn't ideal but it will save some time later on when we onboard the new endpoints.

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
```diff
@@ -0,0 +1,67 @@
...
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "properties.openShiftVersion", "Must be provided")
+	}
+
+	if new.Properties.PlatformWorkloadIdentityRoles == nil || len(new.Properties.PlatformWorkloadIdentityRoles) == 0 {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "properties.platformWorkloadIdentityRoles", "Must be provided and must be non-empty")
+	}
+
+	errs := []error{}
+
+	for i, r := range new.Properties.PlatformWorkloadIdentityRoles {
```
rajdeepc2792, line 42: Is check on serviceaccounts not required?

kimorris27: Good catch! I'll add in a check for the service accounts.

rajdeepc2792 (approved):

cadenmarchese (approved):

PR #3523, opened 2024-04-16
Add command to rotate e2e credentials

hawkowl:

Fixes ARO-6742

Generates/rotates passwords for the SPNs

### Test plan for issue:
Tested in dev, don't have permissions for INT, will require a TL to do that

SOPs for rotation

I don't, but it should be re-runnable...


hack/e2e/cmd/generateServicePrincipals.go
```diff
@@ -0,0 +1,218 @@
...
+			return err
+		}
+		return nil
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(generateServicePrincipalsCmd)
+
+	generateServicePrincipalsCmd.Flags().String("keyvault", "aro-e2e-principals", "keyvault to use")
```
mociarain, line 42: I like this pattern. Maybe I'm extrapolating too much (I find the cobra docs hard to parse) but if I understand it correctly we could replace a lot of the `go run ......` stuff with this and "proper" CLI commands?

hack/e2e/cmd/generateServicePrincipals.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/Azure/go-autorest/autorest/date"
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/jongio/azidext/go/azidext"
+	"github.com/spf13/cobra"
+
+	"github.com/Azure/ARO-RP/pkg/env"
+	msgraph_apps "github.com/Azure/ARO-RP/pkg/util/graph/graphsdk/applications"
+	msgraph_models "github.com/Azure/ARO-RP/pkg/util/graph/graphsdk/models"
+	"github.com/Azure/ARO-RP/pkg/util/graph/graphsdk/serviceprincipals"
+	"github.com/Azure/ARO-RP/pkg/util/keyvault"
```
bitoku, line 21: We may want to use the new keyvault track2 SDK instead of the old one since the keyvault SDK track2 is merged https://github.com/Azure/ARO-RP/pull/3275/files

hack/e2e/cmd/generateServicePrincipals.go
```diff
@@ -0,0 +1,218 @@
...
+
+	keyvaultURI := keyvault.URI(_env, keyvaultParam, "")
+	keyvaultManager := keyvault.NewManager(cred, keyvaultURI)
+
+	graphClient, err := _env.Environment().NewGraphServiceClient(azcred)
+	if err != nil {
+		return err
+	}
+
+	for i := 1; i == spnCountParam; i++ {
```
AldoFusterTurpin, line 86: Shouldn't this be
``` 
for i := 0; i < spnCountParam; i++ {
```
but not "==" ? 
Maybe I am missing something.

hack/e2e/cmd/generateServicePrincipals.go
```diff
@@ -0,0 +1,218 @@
...
+			return fmt.Errorf("%d applications found for spn '%s'", len(apps), spnName)
+		}
+
+		log.Infof("found IDs for '%s': ObjectID='%s', AppID='%s'", spnName, spnObjectID, spnAppID)
+
+		log.Info("Ensuring SPN ID in keyvault...")
+
+		err = keyvaultManager.SetSecret(ctx, fmt.Sprintf("%s-app-id", spnName), azkeyvault.SecretSetParameters{
+			Value: to.StringPtr(spnAppID),
+		})
```
AldoFusterTurpin, line 123: IMHO extracting into a variable inline structs that are > 1 line long is a good thing (talking about azkeyvault.SecretSetParameters{)

hack/e2e/cmd/generateServicePrincipals.go
```diff
@@ -0,0 +1,218 @@
...
+		if err != nil {
+			return err
+		}
+
+		if len(apps[0].GetPasswordCredentials()) > 0 {
```
AldoFusterTurpin, line 128: We are using _apps[0].GetPasswordCredentials()_ in 3 different places, maybe we can extract it to a variable, more performant and easier to read.

hack/e2e/cmd/generateServicePrincipals.go
```diff
@@ -0,0 +1,218 @@
...
+			return err
+		}
+
+		log.Info("setting password credential for SPN in keyvault")
+		err = keyvaultManager.SetSecret(ctx, fmt.Sprintf("%s-secret-value", spnName), azkeyvault.SecretSetParameters{
+			Value: resp.GetSecretText(),
+			SecretAttributes: &azkeyvault.SecretAttributes{
+				Expires: &expiryTime,
+			},
+		})
```
AldoFusterTurpin, line 167: Same comment as before, I think it would be great to extract _azkeyvault.SecretSetParameters_ into a variable as this is not easy to read (at least for me).

hack/e2e/cmd/generateServicePrincipals.go
```diff
@@ -0,0 +1,218 @@
...
+		if err != nil {
+			return err
+		}
+
+		log.Info("checking for service principal")
+
+		filter = fmt.Sprintf("appId eq '%s'", spnAppID)
+		spResult, err := graphClient.ServicePrincipals().Get(ctx, &serviceprincipals.ServicePrincipalsRequestBuilderGetRequestConfiguration{
```
AldoFusterTurpin, line 175: Same as before for _serviceprincipals.ServicePrincipalsRequestBuilderGetRequestConfiguration_

hack/e2e/cmd/generateServicePrincipals.go
```diff
@@ -0,0 +1,218 @@
...
+			spID = *spAddResult.GetId()
+		case 1:
+			spID = *sps[0].GetId()
+			log.Infof("found SP ID %s", spID)
+		default:
+			return fmt.Errorf("%d sps found for spn '%s'", len(sps), spnName)
+		}
+
+		log.Info("setting service principal ID in keyvault")
+		err = keyvaultManager.SetSecret(ctx, fmt.Sprintf("%s-sp-id", spnName), azkeyvault.SecretSetParameters{
```
AldoFusterTurpin, line 207: Same as before for _azkeyvault.SecretSetParameters_. Even here is not that important, I still see value in extracting it.

hack/e2e/cmd/generateServicePrincipals.go
```diff
@@ -0,0 +1,218 @@
...
+		err := generateKeys(cmd)
+		if err != nil {
+			cmd.PrintErr(err)
+			return err
+		}
+		return nil
+	},
+}
+
+func init() {
```
AldoFusterTurpin, line 39: Sadly, Cobra forces the developer to use the init() function to register the commands/subcommands and "fun" problems can arise with the usage of init() in Go. But what can we do ...? 😆

AldoFusterTurpin (commented):
Thank you for this @hawkowl. I just left some optional suggestions and a question about a for loop.

hawkowl:
I don't think we need this anymore.

PR #3530, opened 2024-04-23
[ARO-7181] Upgrade the ARO Operator, then update the Cluster object

hawkowl:

Fixes ARO-7181

Fixes a potential bug (and something that affects #6660) where a flag change *and* an ARO Operator upgrade would cause the flag to take effect on the old operator, rather than the new one.

### Test plan for issue:

Unit tests should cover the AdminUpdate behaviour, E2E tests should cover the installation behaviour. It will also undergo QA as a part of #6660.

N/A

Testing :)

### DIFF
--- a/pkg/cluster/adminupdate_test.go
+++ b/pkg/cluster/adminupdate_test.go
@@ -77,6 +77,14 @@ func TestAdminUpdateSteps(t *testing.T) {
 		"[Action ensureAROOperator]",
 		"[Condition aroDeploymentReady, timeout 20m0s]",
 		"[Condition ensureAROOperatorRunningDesiredVersion, timeout 5m0s]",
+		"[Action syncClusterObject]",
+	}
+
+	syncClusterObjectSteps := []string{
+		"[Action startVMs]",
+		"[Condition apiServersReady, timeout 30m0s]",
+		"[Action initializeOperatorDeployer]",
+		"[Action syncClusterObject]",
 	}
 
 	hiveSteps := []string{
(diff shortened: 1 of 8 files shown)


pkg/operator/deploy/deploy.go
```diff
@@ -277,31 +301,33 @@ func (o *operator) resources(ctx context.Context) ([]kruntime.Object, error) {
...
+func (o *operator) SyncClusterObject(ctx context.Context) error {
+	resource, err := o.clusterObject(ctx)
+	if err != nil {
+		return err
+	}
+	return o.dh.Ensure(ctx, resource)
 }
 
-func (o *operator) CreateOrUpdate(ctx context.Context) error {
+func (o *operator) CreateOrUpdate(ctx context.Context, isInstall bool) error {
```
mociarain, line 315: I think we should split this into 2 separate functions: Create and Update with a common internal method.

pkg/cluster/adminupdate_test.go
```diff
@@ -56,6 +56,7 @@ func TestAdminUpdateSteps(t *testing.T) {
 				"[Action ensureAROOperator-fm]",
 				"[Condition aroDeploymentReady-fm, timeout 20m0s]",
 				"[Condition ensureAROOperatorRunningDesiredVersion-fm, timeout 5m0s]",
+				"[Action syncClusterObject-fm]",
```
mociarain, line 59: I think it's worth adding a test for just the `MaintenanceTaskSyncClusterObject=SyncClusterObject`.

pkg/operator/deploy/deploy.go
```diff
@@ -207,6 +208,29 @@ func (o *operator) resources(ctx context.Context) ([]kruntime.Object, error) {
 		return nil, err
 	}
 
+	domain := o.oc.Properties.ClusterProfile.Domain
+	if !strings.ContainsRune(domain, '.') {
+		domain += "." + o.env.Domain()
+	}
```
mociarain, line 214: Why is this needed?

mociarain (changes requested):
Few questions and suggestions

pkg/cluster/adminupdate_test.go
hawkowl: done

pkg/operator/deploy/deploy.go
hawkowl: done

pkg/operator/deploy/deploy.go
hawkowl: incorrect copy-paste, fixed

mociarain (dismissed):
Nice. Looks great

jaitaiwan (approved):
One small nit, don't think it needs actioning in this PR. Approving and merging

PR #3531, opened 2024-04-23
Update some clienthelper hooks for better testing

hawkowl:

Split out of #6660

Improves the clienthelper hooks for pre/post behaviour, to cover being able to spawn errors as well as keep track of changes.

### Test plan for issue:

Part of test code

N/A

Purely test code

### DIFF
--- a/test/util/clienthelper/hookclient.go
+++ b/test/util/clienthelper/hookclient.go
@@ -18,59 +18,110 @@ type hookFunc func(obj client.Object) error
 type HookingClient struct {
 	f client.WithWatch
 
-	getHook    []getFunc
-	deleteHook []hookFunc
-	createHook []hookFunc
-	updateHook []hookFunc
-	patchHook  []hookFunc
+	preGetHook    []getFunc
+	preDeleteHook []hookFunc
+	preCreateHook []hookFunc
+	preUpdateHook []hookFunc
+	prePatchHook  []hookFunc
+
+	postGetHook    []getFunc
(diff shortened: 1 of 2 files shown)


test/util/clienthelper/hookclient.go
```diff
@@ -18,59 +18,110 @@ type hookFunc func(obj client.Object) error
...
 	}
 }
 
 func (c *HookingClient) WithGetHook(f getFunc) *HookingClient {
-	c.getHook = append(c.getHook, f)
+	c.postGetHook = append(c.postGetHook, f)
 	return c
 }
 
 func (c *HookingClient) WithDeleteHook(f hookFunc) *HookingClient {
```
mociarain, line 63: If we have `WithPreDeleteHook` I would be in favour of `WithPostDeleteHook` for symmetry and discoverability. Same with the others here.

What do you think?

mociarain (commented):
I really like this, thanks for finding the time and making the improvement. Just a small suggestion on naming.

test/util/clienthelper/hookclient.go
jaitaiwan: I agree, doesn't block PR though

jaitaiwan (approved):
LGTM

test/util/clienthelper/hookclient.go
mociarain: PR for it here: https://github.com/Azure/ARO-RP/pull/3549

PR #3534, opened 2024-04-23
Restructure VMSS bootstrap bash scripts for increased reliability and debugability

s-fairchild:

[ARO-6773](https://issues.redhat.com/browse/ARO-6773)

Fixes

**I will be closing this PR once [PR #3741](https://github.com/Azure/ARO-RP/pull/3741) is merged**.

This is needed for a resilent to failure RP deployment.
Our current implementation has shown some faults over time with RP deployments, and need improvements in general.  

[Custom scripts Tips](https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/custom-script-linux#tips) Recommends scheduling a reboot if it is needed. This is to allow all other VM extensions to complete their setup. If the VM is rebooted while they are executing setup steps, they will fail to finish.
Therefore, I have made all system services start at the end of the script, made firewall rules effective in runtime, and scheduled a reboot 1 hour after the script finishes. 

Move all shared code into a commonly shared file to be sourced by all
bootstrapping scripts. This allows for code reuse, minimal duplication.

### Test plan for issue:

Tested bootstrapping scripts in an Azure VM using image RedHat:RHEL:8-lvm-gen2:latest.  
Mainly to catch missed syntax errors, and ensure files written are formatted as expected.  

Fully tested with deployment to INT and FF INT. [Successful Deployment and E2E](https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=95439277&view=results)

[Second INT test run](https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=97407813&view=results)

If you would like to test these bootstrapping scripts out yourself, here are some helper functions.

```bash
# testing_vm.env

export RESOURCEGROUP="${USER}-bootstrap-test"
export VM_NAME="$RESOURCEGROUP"
```

```bash
#!/bin/bash
# manage_testing_vm.sh

vm_create() {
    check_vars

    # os disk size is set to 256GB to allow testing of configure_disk_partitions
    az vm create --nic-delete-option Delete --public-ip-address pubIP --public-ip-address-allocation dynamic -g $RESOURCEGROUP -n $VM_NAME --image RedHat:RHEL:8-lvm-gen2:latest --ssh-key-values ~/.ssh/id_rsa.pub --public-ip-address-allocation static --os-disk-size-gb 256
}

vm_delete() {
    check_vars

    az vm delete -g $RESOURCEGROUP -n $VM_NAME -y
}

check_vars() {
    echo "RESOURCEGROUP=${RESOURCEGROUP:?Please export RESOURCEGROUP}"
    echo "VM_NAME=${VM_NAME:?Please excport VM_NAME}"
}
```

Here is a helper environment file, to avoid failures for unset env variables.  

```bash
# rp_boostrap_testing_variables.env
# rpVMSS.sh
export RPMDSDCONFIGVERSION=empty RPMDSDNAMESPACE=empty MDSDCERTIFICATESAN=empty LOCATION=eastus RPMDSDACCOUNT=empty MDSDENVIRONMENT=empty PORTALELEVATEDGROUPIDS=empty DATABASEACCOUNTNAME=empty KEYVAULTPREFIX=empty RPMDMACCOUNT=empty RPPARENTDOMAINNAME=empty RPIMAGE=empty AZURECLOUDNAME=empty MDMIMAGE=empty FLUENTBITIMAGE=empty LOGKIND=empty MDMFRONTENDURL=empty MDMSOURCEENVIRONMENT=empty MDMSOURCEROLE=empty MDMSOURCEROLEINSTANCE=empty KEYVAULTDNSSUFFIX=empty ACRRESOURCEID=empty ADMINAPICLIENTCERTCOMMONNAME=empty ARMAPICLIENTCERTCOMMONNAME=empty ARMCLIENTID=empty FPCLIENTID=empty FPSERVICEPRINCIPALID=empty BILLINGE2ESTORAGEACCOUNTID=empty CLUSTERMDMACCOUNT=empty CLUSTERMDSDACCOUNT=empty CLUSTERMDSDCONFIGVERSION=empty CLUSTERMDSDNAMESPACE=empty CLUSTERPARENTDOMAINNAME=empty GATEWAYDOMAINS=empty GATEWAYRESOURCEGROUPNAME=empty RPFEATURES=empty CLUSTERSINSTALLVIAHIVE=empty CLUSTERDEFAULTINSTALLERPULLSPEC=empty CLUSTERSADOPTBYHIVE=empty USECHECKACCESS=empty DBTOKENCLIENTID=empty GATEWAYSERVICEPRINCIPALID=empty PORTALACCESSGROUPIDS=empty PORTALCLIENTID=empty

base64_empty="ZW1wdHkK"
export ADMINAPICABUNDLE="$base64_empty"
export ARMAPICABUNDLE="$base64_empty"

# devProxyVMSS.sh
export PROXYIMAGE=empty PROXYIMAGEAUTH=empty AZURECLOUDNAME=empty
export PROXYCERT="$base64_empty"
export PROXYKEY="$base64_empty"
export PROXYCLIENTCERT="$base64_empty"

# gatewayVMSS.sh
export DBTOKENURL=empty GATEWAYFEATURES=empty GATEWAYMDSDCONFIGVERSION=empty AZURESECPACKVSATENANTID=empty AZURESECPACKQUALYSURL=empty
```

Simply `scp` the files you would like to test to your newly created VM, and execute them.  

```bash
. testing_vm.env

vm_create
RESOURCEGROUP=steven-bootstrap-test
VM_NAME=steven-bootstrap-test

bootstrapper_vm_ip="Replace with VM IP address output from above"

scp pkg/deploy/generator/scripts/devProxyVMSS.sh "$bootstrapper_vm_ip":
scp pkg/deploy/generator/scripts/commonVMSS.sh "$bootstrapper_vm_ip":
scp pkg/deploy/generator/scripts/gatewayVMSS.sh"$bootstrapper_vm_ip":
scp pkg/deploy/generator/scripts/rpVMSS.sh "$bootstrapper_vm_ip":
scp ~/Documents/bootstrap_vmss_testing/rp_boostrap_testing_variables.env "$bootstrapper_vm_ip":
```

When you're done testing, delete the vm.

```bash
vm_delete
RESOURCEGROUP=steven-bootstrap-test
VM_NAME=steven-bootstrap-test
```

##### Abort

```bash
Error: Unable to find a match: azsec-clamav azsec-monitor azure-cli azure-mdsd azure-security
dnf_install_pkgs: failed to install required packages
```

##### Log

```bash
configure_sshd: setting ssh password authentication
configure_disk_partitions: extending partition table
```

No

Per testing mentioned in the test plan.


tony-schndr (commented):
Nice work!  There is a lot to review here, just one question regarding reboots so far.

pkg/deploy/generator/scripts/commonVMSS.sh
```diff
@@ -0,0 +1,1259 @@
...
+# reboot_vm restores all selinux file contexts, then schedules a reboot for one hour later
+# Reboots should scheduled after all VM extensions have had time to complete
+# Reference: https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/custom-script-linux#tips
+reboot_vm() {
+    log "starting"
+
+    configure_selinux "true"
+    
+    hour="$(date -d "1 hour" +%H:%M)"
+    shutdown -r "$hour" "Post deployment reboot is happening now"
```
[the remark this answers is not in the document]
s-fairchild: Good call out.
I imagine the three would reboot around the same time, and cause a short outage.  

Trying to think about the best way to make the reboots cascade, is probably to parse the VM's hostname for the numerical postfix. Then time reboots 5 minutes apart based on the number.
1. _0 reboots 1h later
2. _1 reboots 1h 5m later
3. _2 reboots 1h 10m later  

Thoughts?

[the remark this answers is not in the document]
s-fairchild: I pushed a commit with the changes I suggested.  
[INT Deploy Pipeline](https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=95609542&view=results)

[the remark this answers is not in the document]
tony-schndr: This works, but IMO the ideal scenario is the old VMSS isn't torn down until reboots finish on the new VMSS, right?  Also, in a sense a reboot can inform you if there are any changes that aren't persistent, in this case we wouldn't learn that till after the old VMSS is torn down, by then it's too late and we potentially cause an outage.

Is there any other driver to this change besides Azure documentation best practices?  Have we seen issues rebooting immediately preventing extension installation from completing?  If not then maybe it's best to continue doing reboots how we have been.

My other idea on this is if we are going to give other extensions time to finish installing before rebooting, we check health probes to ensure we only bring down one node at a time, that might be hard/impossible to coordinate at a node level though.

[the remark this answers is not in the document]
s-fairchild: There's no other driver for this.

Yes, although reviewing the code I think that's just based on timing.
The reboot occurs at the end of the script, and as it is in master now, the healthprobe will fail until the VMSS reboots, starting the enabled aro-rp service.

The instances are configured to be fully functional without rebooting in this PR, the reboot remains simply to ensure our config persists across reboots.

With the delay, there is the danger that a reboot could break the running config on instances. With the deploy being completed, we can't rollback to the previous VMSS, so it would just break and possibly cause a service outage.

I think we maybe able to coordinate health probe checking with `az cli` or possibly even `curl` to check the other instance's health probe statuses. Then use this to determine if they are all healthy before going forward with a reboot.  
In the case that one is unhealthy, the reboot doesn't happen (or polls until it's healthy), and we get an alert that the VMSS is degraded.

Other options are...
1. Force the deployment to wait for the reboots to occur
2. Ignore the extension setup, reboot upon completion anyways
3. Just don't reboot?
  - If we don't reboot instances, and a reboot occurs causing one to go unhealthy, remediation would mean scaling the VMSS up and down.

niontive:
Could we consider moving to managed run commands: https://learn.microsoft.com/en-us/azure/virtual-machines/linux/run-command-managed? We can file a JIRA issue and get to this later.

Custom script extension is on its way out, and run/managed run commands offer several more benefits (see docs).

pkg/deploy/generator/scripts/commonVMSS.sh
```diff
@@ -0,0 +1,1281 @@
...
+# Reference: https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/custom-script-linux#tips
+reboot_vm() {
+    log "starting"
+
+    configure_selinux "true"
+    
+    local -i offset
+    get_reboot_offset offset
+    hour="$(date -d "1 hour" +%H:%M + $offset minutes)"
+    shutdown -r "$hour" "Post deployment reboot is happening at $hour"
```
niontive, line 456: Just a heads up - Microsoft recommends not putting reboots into CSE scripst: https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/custom-script-linux

This is more of a tip than a requirement, but maybe it's something we can clean up later.

pkg/deploy/generator/scripts.go
```diff
@@ -15,3 +15,6 @@ var scriptGatewayVMSS []byte
 
 //go:embed scripts/rpVMSS.sh
 var scriptRpVMSS []byte
+
+//go:embed scripts/commonVMSS.sh
+var scriptCommonVMSS []byte
```
niontive, line 20: `commonVMSS.sh` is a massive file - over a thousand lines!

Could we do the following:
- Refactor `commonVMSS.sh` into separate scripts
- Have a function that concats all those scripts into a byte slice called `scriptCommonVMSS`
- Then append the other scripts (ex: `rpVMSS.sh`) for deployment.

This way we can make `commonVMSS.sh` more readable.

pkg/deploy/generator/scripts/commonVMSS.sh
s-fairchild: Technically a reboot isn't happening in this script. It's being scheduled for later. The existing scripts have the reboot occurring in the script.

With the changes in this script, everything is functional in the present runtime. With the existing scripts in master, they are persisted to disk, but not started, such as enabling but not starting services, firewalld rules.

The point in rebooting now is simply to ensure a reboot doesn't impact the running state. Incase an instance were to reboot.
If we don't test that during deployment, Can we safely assume the instances will never reboot?

pkg/deploy/generator/scripts.go
s-fairchild: Appending will not work, because the functions won't be declared in `rpVMSS.sh`.
The functions would be called before they have been created.

s-fairchild:
I'm placing this PR under a hold until I can work out the ideal way to reboot and monitor the VMSS instances.
Ideally, they will reboot in a rolling fashion while also checking the health probes.
Currently I have the script performing a rolling reboot, but there is no status/health check to determine if the other instances are healthy, before rebooting.

pkg/deploy/generator/scripts/commonVMSS.sh
s-fairchild: @niontive I've broken `commonVMSS.sh` up into 4 different `util-*.sh` files.

pkg/deploy/generator/scripts.go
s-fairchild: @niontive I've broken `commonVMSS.sh` up into 4 different `util-*.sh` files.
I'm testing it in INT now.

PR #3535, opened 2024-04-24
Limiting KEYVAULT_PREFIX to 20 Chars

anshulvermapatel:
We have set the char limit to 20 in our deployment template. Hence this env's value should also be atmost 20 char. It exceeds with a region with big name is chosed like australiaeast.

### DIFF
--- a/env-int.example
+++ b/env-int.example
@@ -5,5 +5,6 @@ export RESOURCEGROUP=$USER-aro-$LOCATION
 export DATABASE_ACCOUNT_NAME=$USER-aro-$LOCATION
 export DATABASE_NAME=ARO
 export KEYVAULT_PREFIX=$USER-aro-$LOCATION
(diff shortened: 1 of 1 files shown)


mociarain:
Some questions:
- Out of curiosity why did we set the limit to 20 characters in the deployment template? I'm assuming we didn't have one before so why add it?
- Can you link the PR that changes the deployment template?

anshulvermapatel:
I don't know why it was the limit. I saw this last time also when I deploy full rp service. 
this is the PR which added the template - https://github.com/Azure/ARO-RP/pull/2128
@mociarain

mociarain (approved):
Okay, makes sense but it seems weird to me that we've gotten this far without this. I'm assuming colleagues in longer named regions have been dealing with this ad-hoc. 

Do I have this right? (not blocking just curious)

hawkowl (approved):

PR #3536, opened 2024-04-24
extended maximum wait time before yum update

bizz001:

Fixes the issue of yum update conflicting with running AzSecPack VM extension. Increases the total wait time before running a yum update command to 30 minutes in order to allow the 20 minute timeout of the VM extension to end. More info here https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/features-linux?tabs=azure-cli#common-reasons-for-extension-failures 

### Test plan for issue:

Run the deployment.

Deployment works.


UlrichSchlueter (approved):
LGTM

PR #3538, opened 2024-04-25
Bug fix - clear ReadOnly fields on PUT or PATCH for 2023-11-22 and 2024-08-12-preview APIs

tony-schndr:

Fixes https://issues.redhat.com/browse/ARO-7192

Clears the ReadOnly fields on PUT or PATCH requests in 2023-11-22 and 2024-08-12-preview APIs.  The addition of clearing the fields on PUT requests is due to the frontend prepopulating fields it "thinks" are expected, toInternal() preserves these fields.

### Test plan for issue:
Ran e2e against api 2023-11-22
Existing unit tests
created cluster and ran az aro update against 2023-11-22

### DIFF
--- a/pkg/frontend/openshiftcluster_putorpatch.go
+++ b/pkg/frontend/openshiftcluster_putorpatch.go
@@ -149,8 +149,9 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 	// Patch should be used for updating individual fields of the document.
 	case http.MethodPatch:
 		ext = converter.ToExternal(doc.OpenShiftCluster)
-		converter.ExternalNoReadOnly(ext)
 	}
+
+	converter.ExternalNoReadOnly(ext)
 
 	err = json.Unmarshal(body, &ext)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cadenmarchese (commented):
I pulled these changes into #3529 and ran an `az aro update` + `az aro update --refresh-credentials`, and it works! Let me know if you need any additional manual smoke tests for sanity's sake.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -149,9 +149,10 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 	// Patch should be used for updating individual fields of the document.
 	case http.MethodPatch:
 		ext = converter.ToExternal(doc.OpenShiftCluster)
-		converter.ExternalNoReadOnly(ext)
 	}
 
+	converter.ExternalNoReadOnly(ext)
```
SudoBrendan, line 154: This change seems odd at a glance, but maybe it is correct - is the intended behavior that readonly fields shouldn't be returned on GET either, or am I reading this incorrectly? I could see this affecting things that need to read SystemData/etc

cadenmarchese (approved):

cadenmarchese:
I've tested this change in lower environments as part of #3529 so I'm going to go ahead and get it merged. Thanks again!

PR #3542, opened 2024-04-26
Make a package from pkg/api

hawkowl:
Supersedes #3423

Fixes https://issues.redhat.com/browse/ARO-3571

This makes ARO-RP/pkg/api a golang package, similar to how Hive's API package is used. This means we can reuse the ARO-RP API definitions in other repos, such as the ARO-Installer-Wrapper repo.

### Test plan for issue:

Unit tests should suffice

Not entirely sure -- the vendor check should catch if the vendoring doesn't occur after pkg/api changes.


hawkowl:
Closing this for now.

PR #3543, opened 2024-04-26
Drop ErrorIfNoField

mociarain:
This PR drops the use of the ErrorIfNoField flag for our decoding. This flag: ` If true, returns an error if a map in the stream has a key which does not map to any field; else read and discard the key and value in the stream and proceed to the next.` (http://ugorji.net/blog/go-codec-primer)

The principal behind this is that we shouldn't error if we receive more data than expected.

### DIFF
--- a/pkg/api/pretty.go
+++ b/pkg/api/pretty.go
@@ -15,13 +15,7 @@ import (
 var secretPreservingJSONHandle *codec.JsonHandle
 
 func init() {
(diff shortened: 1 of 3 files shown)


bennerv (approved):
lgtm

s-fairchild (approved):

AldoFusterTurpin (approved):

PR #3549, opened 2024-04-29
Rename clienthelpers hooks

mociarain:
Small refactor from https://github.com/Azure/ARO-RP/pull/3531 for symmetry and discoverability.


tiguelu (approved):
LGTM

PR #3492, opened 2024-04-02
Add new makefile targets for go mod verification

razo7:
### Which issue this PR addresses:

None. It is a small enhancement. 

### What this PR does / why we need it:
Run 'go mod' functions to search for [tidy](https://go.dev/ref/mod#go-mod-tidy), and [vendor](https://go.dev/ref/mod#go-mod-vendor) package changes and then [verify](https://go.dev/ref/mod#go-mod-verify) it.

### Is there any documentation that needs to be updated for this PR?

Yes, in the second commit

PR #3548, opened 2024-04-29
update go version to 1.20.12

ArrisLee:
**Which issue this PR addresses:**
https://issues.redhat.com/browse/ARO-6716

**Fixes**
What this PR does / why we need it:
This fixes some CVEs in the standard library which govulncheck notes:
https://pkg.go.dev/vuln/GO-2023-2382
https://pkg.go.dev/vuln/GO-2023-2186
among others patched in go-toolset

**Test plan for issue:**
local go unit test with `/pkg/containerinstall/install_test.go` and E2E

**Is there any documentation that needs to be updated for this PR?**
N/A

**How do you know this will function as expected in production?**
should be fine as long as E2E passes

PR #3550, opened 2024-04-30
Add new makefile targets for go mod verification

razo7:

None. It is a small enhancement. Duplicate of #3492.

Run 'go mod' functions to search for [tidy](https://go.dev/ref/mod#go-mod-tidy), and [vendor](https://go.dev/ref/mod#go-mod-vendor) package changes and then [verify](https://go.dev/ref/mod#go-mod-verify) it.

Yes, in the second commit

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,9 @@ AUTOREST_IMAGE = quay.io/openshift-on-azure/autorest:${AUTOREST_VERSION}
 GATEKEEPER_VERSION = v3.15.1
 GOTESTSUM = gotest.tools/gotestsum@v1.11.0
 
+# Golang version go mod tidy compatibility
+GOLANG_VERSION ?= 1.20
+
 ifneq ($(shell uname -s),Darwin)
     export CGO_CFLAGS=-Dgpgme_off_t=off_t
 endif
@@ -334,10 +337,23 @@ admin.kubeconfig:
 aks.kubeconfig:
 	hack/get-admin-aks-kubeconfig.sh
 
+.PHONY: go-tidy
+go-tidy: # Run go mod tidy - add missing and remove unused modules.
+	go mod tidy -compat=${GOLANG_VERSION}
+
+.PHONY: go-vendor
+go-vendor:  # Run go mod vendor - only modules that are used in the source code will be vendored in (make vendored copy of dependencies). 
+	go mod vendor
+
+.PHONY: go-verify
+go-verify: go-tidy go-vendor # Run go mod verify - verify dependencies have expected content
+	go mod verify
+
 .PHONY: vendor
 vendor:
 	# See comments in the script for background on why we need it
 	hack/update-go-module-dependencies.sh
+	$(MAKE) go-verify
 
 .PHONY: install-go-tools
 install-go-tools:
--- a/.github/workflows/ci-go.yml
+++ b/.github/workflows/ci-go.yml
@@ -37,8 +37,7 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v4
     - run: |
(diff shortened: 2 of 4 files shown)


tiguelu (changes requested):

anshulvermapatel (commented):
Everything looks good to me, except what @tiguelu already pointed about. Once thats changed. LGTM

Makefile
```diff
@@ -266,9 +269,22 @@ admin.kubeconfig:
 aks.kubeconfig:
 	hack/get-admin-aks-kubeconfig.sh
 
+.PHONY: go-tidy
```
mociarain, line 272: Why each individual PHONY and not added to the overall PHONY at the bottom of the file?

Makefile
```diff
@@ -18,6 +18,9 @@ GATEKEEPER_VERSION = v3.10.0
 GATEKEEPER_IMAGE ?= ${RP_IMAGE_ACR}.azurecr.io/gatekeeper:$(GATEKEEPER_VERSION)
 GOTESTSUM = gotest.tools/gotestsum@v1.11.0
 
+# Golang version go mod tidy compatibility
+GOLANG_VERSION ?= 1.20
```
mociarain, line 22: Should we specify 1.20.12? (Still finding my feet with Go, I have no idea how strict it is with versions)

For reference: https://github.com/Azure/ARO-RP/pull/3548

mociarain (commented):
Nice change. Just a few questions

Makefile
```diff
@@ -167,7 +170,7 @@ run-portal:
 	go run -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=$(VERSION)" ./cmd/aro portal
 
 build-portal:
-	cd portal/v2 && npm install && npm run build
+	cd portal/v1 && npm install && npm run build && cd ../v2 && npm install && npm run build
```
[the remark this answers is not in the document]
razo7: Yes, not sure how it was even got in...

Makefile
razo7: go mod tidy is expecting an x.y version, thus we can't do that.

> go mod tidy -compat=1.20.12
invalid value "1.20.12" for flag -compat: expecting a Go version like "1.20"
usage: go mod tidy [-e] [-v] [-x] [-go=version] [-compat=version]
Run 'go help mod tidy' for details.
make: *** [Makefile:283: go-tidy] Error 2

Makefile
razo7: Do you mean why we need to create three targets (with one PHONY per target) instead of one target (and one PHONY)?

Each go module command has a different purpose, and by separating them, I believe, we get better code readability and flow of the targets. They could be grouped together as they are used today, but separation empowers more flexibility :) Having said, I don't have a strong opinion on that.

Makefile
hlipsig: Let's just double check this is the version we need. I remember some CVE issues so even though we just moved to 1.20 we need to make sure we have the version with the xnet fixes in.

hlipsig (approved):
LTGM except the go version. Let's just get confirmation that we've not altering it to a higher version anywher else first.

Makefile
jaitaiwan: Go 1.20 is the best version we can support right now, we'll definitely need to move to 1.22 soon though. My understanding is that we have many underlying components we have to upgrade as well if we do, so 1.20 min is our first best step towards compliance.

jaitaiwan (approved):
LGTM, if there's no objections from @hlipsig I think we should be able to just merge this.

tiguelu (approved):

PR #3555, opened 2024-05-02
MISE Middleware

SrinivasAtmakuri:

Fixes


pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -303,6 +303,199 @@ WantedBy=multi-user.target"
...
+FPTENANTID='$FPTENANTID'
+MISEIMAGE='$image'
+MISEVALIDAUDIENCES='$MISEVALIDAUDIENCES'
+MISEVALIDAPPIDS='$MISEVALIDAPPIDS'
+LOGININSTANCE='$LOGININSTANCE'"
+
+    write_file aro_mise_service_conf_filename aro_mise_service_conf_file true
+
+    mkdir -p /app/mise
+    cat >/app/mise/appsettings.json <<EOF
```
s-fairchild, line 329: Make use of `write_file`. You should be able to use single quotes `''` to allow for using double quotes `""` within the variable that you pass to `write_file`.  
Worst case, you can always escape them with `\"`, but you shouldn't need to do that.

pkg/deploy/generator/scripts/util-system.sh
```diff
@@ -165,7 +165,7 @@ pull_container_images() {
 
     # This name is used in the case that az acr login searches for this in it's environment
     export REGISTRY_AUTH_FILE="/root/.docker/config.json"
-
```
s-fairchild, line 168: Were these just 4 spaces left behind?

s-fairchild (commented):
The bootstrapper code looks LGTM. Thank you for following the new conventions laid out in the bash scripts!

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -303,6 +303,199 @@ WantedBy=multi-user.target"
...
+[Service]
+RestartSec=1s
+EnvironmentFile=/etc/sysconfig/aro-mise
+ExecStartPre=-/usr/bin/docker rm -f %N
+ExecStart=/usr/bin/docker run \
+  -p 5000:5000 \
+  -v /app/mise/appsettings.json:/app/appsettings.json:z \
+  --hostname %H \
+  --name %N \
+  --net=aro-net \
```
s-fairchild, line 405: Can you use `$network` passed in on line 312?

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -303,6 +303,199 @@ WantedBy=multi-user.target"
...
```
s-fairchild, line 486: Same as above, use `$network` argument provided.

s-fairchild (changes requested):

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -292,17 +301,218 @@ ExecStart=/usr/bin/podman run \
...
+StartLimitIntervalSec=0
+[Service]
+RestartSec=1s
+EnvironmentFile=/etc/sysconfig/aro-otel-collector
+ExecStartPre=-/usr/bin/docker rm -f %N
+ExecStart=/usr/bin/docker run \
+  --hostname %H \
+  --name %N \
+  --rm \
+  --network=${PODMAN_NETWORK} \
```
s-fairchild, line 502: Can you add a commit to make this change for the other service files? That way it's all consistent.
I'm fine with putting the network in the systemd env file, but it should be consistent with the rest of the services.

pkg/deploy/config.go
```diff
@@ -63,6 +63,7 @@ type Configuration struct {
 	ExtraServiceKeyvaultAccessPolicies []interface{}          `json:"extraServiceKeyvaultAccessPolicies,omitempty" value:"required"`
 	FluentbitImage                     *string                `json:"fluentbitImage,omitempty" value:"required"`
 	FPClientID                         *string                `json:"fpClientId,omitempty" value:"required"`
+	FPTENANTID                         *string                `json:"fpTenantId,omitempty" value:"required"`
```
bennerv, line 66: Nit: not all caps

pkg/deploy/config.go
```diff
@@ -78,6 +79,8 @@ type Configuration struct {
 	KeyvaultPrefix                     *string                `json:"keyvaultPrefix,omitempty" value:"required"`
 	MDMFrontendURL                     *string                `json:"mdmFrontendUrl,omitempty" value:"required"`
 	MDSDEnvironment                    *string                `json:"mdsdEnvironment,omitempty" value:"required"`
+	MISEVALIDAUDIENCES                 []string               `json:"miseValidAudiences,omitempty"`
+	MISEVALIDAPPIDs                    []string               `json:"miseValidAppIDs,omitempty"`
```
bennerv, line 83: nit: not all caps

pkg/deploy/devconfig.go
```diff
@@ -174,6 +175,13 @@ func DevConfig(_env env.Core) (*Config, error) {
 			MDMFrontendURL:              ptr.To("https://global.ppe.microsoftmetrics.com/"),
 			MDSDEnvironment:             ptr.To(version.DevGenevaLoggingEnvironment),
 			MsiRpEndpoint:               ptr.To("https://iamaplaceholder.com"),
+			MISEVALIDAUDIENCES: []string{
+				"https://management.core.windows.net/",
```
bennerv, line 179: is this consistent every cloud?  Should this be a config?

pkg/deploy/devconfig.go
```diff
@@ -174,6 +175,13 @@ func DevConfig(_env env.Core) (*Config, error) {
...
+				_env.Environment().ResourceManagerEndpoint,
+			},
+			MISEVALIDAPPIDs: []string{
+				"2187cde1-7e28-4645-9104-19edfa500053",
```
bennerv, line 183: can we read this in as config.

pkg/deploy/devconfig.go
SrinivasAtmakuri: This is for generating devconfig for full RP deployment, so there is no ARM involved, which means MISE will not be needed. I have added a test detail to make it pass validation.

pkg/deploy/devconfig.go
SrinivasAtmakuri: This is for generating devconfig for full RP deployment, so there is no ARM involved, which means MISE will not be needed. I have added a test detail to make it pass validation.

pkg/deploy/devconfig.go
```diff
@@ -152,6 +152,7 @@ func DevConfig(_env env.Core) (*Config, error) {
 			},
 			FluentbitImage:       ptr.To(version.FluentbitImage(azureUniquePrefix + "aro." + _env.Environment().ContainerRegistryDNSSuffix)),
 			FPClientID:           ptr.To(os.Getenv("AZURE_FP_CLIENT_ID")),
+			FPTENANTID:           ptr.To(os.Getenv("AZURE_TENANT_ID")),
```
sankur-codes, line 155:
```suggestion
			FPTenantID:           ptr.To(os.Getenv("AZURE_TENANT_ID")),
```

sankur-codes (changes requested):
Missed change.

sankur-codes:
The PR has been rebased and worked upon & tested further in different PR (#4070)

PR #3557, opened 2024-05-02
Do not perform explicit CSP roleassignment deletion in prod e2e

tsatam:

Fixes no Jira yet

Along the same lines as #3513, this PR skips explicit roleassignment deletion during cluster cleanup when we are going to delete the resources within the roleassignment scope anyways. 

If the resource within the scope of a roleassignment is deleted, or any parent resources are deleted (e.g. resource group, or top-level resource of a sub-resource scope), the role assignments get cleaned up as well. 

This change avoids "failing" E2E in scenarios where we fail to perform explicit roleassignment deletion, but the roleassignments get cleaned up anyways through implicit deletion alongside their resources.

### Test plan for issue:

- PR E2E succeeds and no roleassignments are left behind on the service principal
- Prod E2E succeeds and no roleassignments are left behind on the service principal

N/A 

See test plan above - we should see green E2E when all tests pass and not error during cleanup.

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -389,16 +389,12 @@ func (c *Cluster) Delete(ctx context.Context, vnetResourceGroup, clusterName str
 	switch {
 	case c.ci && env.IsLocalDevelopmentMode(): // PR E2E
 		errs = append(errs,
-			c.deleteRoleAssignments(ctx, vnetResourceGroup, clusterName),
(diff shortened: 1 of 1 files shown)


SudoBrendan (approved):

jhoreman (approved):
Looks good to me. Thanks for the quick turnaround.

swiencki:
lgtm!

PR #3532, opened 2024-04-23
upgrade to go1.20.12

yjst2012:
### Which issue this PR addresses:

https://issues.redhat.com/browse/ARO-6717

Fixes

### What this PR does / why we need it:

This fixes some CVEs in the standard library which govulncheck notes:
https://pkg.go.dev/vuln/GO-2023-2382
https://pkg.go.dev/vuln/GO-2023-2186
among others patched in go-toolset

### Test plan for issue:

basic checks plus ci & e2e

### Is there any documentation that needs to be updated for this PR?

NA
### How do you know this will function as expected in production? 

should be fine as long as e2e passes

PR #3558, opened 2024-05-02
`make ci` support for golang (RP) components

SudoBrendan:

Fixes [ARO-5087](https://issues.redhat.com/browse/ARO-5087). Special shout-out to #tony-schndr for his amazing work in getting the massive initial hurdles out of the way!

* Adds a Makefile target `make ci-rp` that issues a `podman build` command to consolidate 100% of required validation steps into a container build process (read: in the containerfile) so that it runs anywhere (local, pr, and on trunk commits in CI).
* Creates a clear line of logic and removes complex inter-dependencies during the build process. I hope I have clarified the exact build requirements by only moving commands in one direction flow: Makefile >> docker build >> shell commands. There are no longer complex chains of logic to parse out (e.g. "what env vars do I need to set for this command to work?" "what does shell script X do, and what are it's dependencies?" etc. I hope the "magic" of our build process is now demystified in a single place that can be read clearly from top to bottom and used as a source of truth (the Dockerfile). As an added benefit, by moving away from `make`, we are now allowed to disassociate from `git`, meaning we can actually use the docker build cache (alternative being that we `COPY .git .git`, which will always invalidate the cache on each build)
* As with `ci-portal`, this change accommodates local podman build caching to improve performance so it only rebuilds when specific changes are present (e.g. changes in `/python` or `/docs` will not invalidate the cache). I would generally recommend taking advantage of this by including `NO_CACHE=false` in your `env` file. At the same time, it's important to build without caching every once in a while to be sure your changes work with the most recent versions of things (e.g. linux packages pulled in the image with dnf).

### Test plan for issue:

Everyone, everywhere, all the time, should be able to run `make ci-rp` and it "just works". I need feedback from people globally and on various OS (I have tested on Fedora 39, Podman 4.9.4).

Yes, this affects local development flows, **in particular for Mac users**:

1. We re-discovered that `golangci-lint` is a HOG, and needs something like 4GB of memory to execute successfully. Ensuring that `podman machine` is properly configured with enough memory is critical here, and there are documentation updates for how to do that.
2. MacOS, with a case-insensitive filesystem, may build `vendor` differently locally in a way not compatible with Linux. We added a Makefile target to fix this automatically in the meantime, but I think the removal of `vendor` is the long-term solution to consistent builds everywhere.

Currently not a production change - this will be a parallel CI/CD effort that for now does not impact prod. Eventually, these container builds will be used to generate consistent artifacts and create a chain of custody into prod - improving our consistency and resiliency during release. All of this work will be reviewable in future PRs, and are not included here.

### DIFF
--- a/Dockerfile.ci-rp
+++ b/Dockerfile.ci-rp
@@ -1,0 +1,22 @@
+ARG REGISTRY
+FROM ${REGISTRY}/ubi8/go-toolset:1.20.10 as builder
+
+USER root
+ENV GOPATH=/root/go
+WORKDIR /app
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/local/bin v1.55.2
+COPY . /app
+RUN make aro RELEASE=${IS_OFFICIAL_RELEASE} -o generate && make e2e.test &&\
+  go mod tidy -compat=1.20 && \
+  go mod vendor && \
+  hack/ci-utils/isClean.sh && \
+  make lint-go && \
+  ARO_RUN_PKI_TESTS=nope make unit-test-go && \
+  make validate-fips
+FROM ${REGISTRY}/ubi8/ubi-minimal
+RUN microdnf update && microdnf clean all
+COPY --from=builder /app/aro /app/e2e.test /usr/local/bin/
+ENTRYPOINT ["aro"]
+EXPOSE 2222/tcp 8080/tcp 8443/tcp 8444/tcp 8445/tcp
+USER 1000
+
--- a/Dockerfile.ci-rp
+++ b/Dockerfile.ci-rp
@@ -1,22 +1,48 @@
 ARG REGISTRY
-FROM ${REGISTRY}/ubi8/go-toolset:1.20.10 as builder
+ARG VERSION
 
+###############################################################################
+# builder is responsible for all compilation and validation of the RP
+###############################################################################
+FROM ${REGISTRY}/ubi8/go-toolset:1.20.12-5 as builder
 USER root
-ENV GOPATH=/root/go
 WORKDIR /app
-RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/local/bin v1.55.2
-COPY . /app
-RUN make aro RELEASE=${IS_OFFICIAL_RELEASE} -o generate && make e2e.test &&\
-  go mod tidy -compat=1.20 && \
-  go mod vendor && \
-  hack/ci-utils/isClean.sh && \
-  make lint-go && \
-  ARO_RUN_PKI_TESTS=nope make unit-test-go && \
-  make validate-fips
-FROM ${REGISTRY}/ubi8/ubi-minimal
+
+# golang config
+ENV GOPATH=/root/go
+ENV GOFLAGS="-tags=containers_image_openpgp,exclude_graphdriver_btrfs,exclude_graphdriver_devicemapper"
+
+# install general build tools
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/local/bin v1.56.2
+
+# copy from least changed -> most changed to improve cache hits
+COPY go.mod go.sum ./
+COPY vendor vendor
+COPY swagger swagger
+COPY .golangci.yml ./
+COPY hack hack
+COPY cmd cmd
+COPY pkg pkg
+COPY test test
+
+# lint
+RUN golangci-lint run --verbose
+
+# build
+RUN go generate ./...
+RUN go build -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=${VERSION}" ./cmd/aro
+RUN go test ./test/e2e/... -tags e2e,codec.safe -c -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=${VERSION}" -o e2e.test
+
+# test
+RUN ARO_RUN_PKI_TESTS=nope go run gotest.tools/gotestsum@v1.11.0 --format pkgname --junitfile report.xml -- -coverprofile=cover.out ./...
+RUN hack/fips/validate-fips.sh ./aro
+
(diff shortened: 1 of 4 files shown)


shubhadapaithankar (dismissed):

kimorris27 (changes requested):
I'd like to suggest a little hack to improve quality of life for Mac users.

Given the lack of case sensitivity in the Mac filesystem and the issues it causes with the vendored `Microsoft` directory, I suggest adding a Make target to check if the system is a Mac and rename the Microsoft directory if so. This Make target could be added as a prereq for all of the image build targets to save Mac users from having to remember to do the rename.

I think a good way to check if the system is a Mac is to run `uname -s`. If the system is a Mac, the output should be `Darwin`. If it's `Darwin`, then the Make target should rename `vendor/github.com/Microsoft` to `vendor/github.com/microsoft`.

I'm happy to test it out after you've made this change if you decide to go forward with it 🙇🏻‍♂️

Makefile
```diff
@@ -92,6 +95,11 @@ discoverycache:
 	KUBECONFIG=admin.kubeconfig go run ./hack/gendiscoverycache
 	$(MAKE) generate
 
+fix-macos-vendor:
+ifeq ($(shell uname -s),Darwin)
+	mv ./vendor/github.com/Microsoft ./vendor/github.com/microsoft || true
```
kimorris27, line 100:
```suggestion
	mv ./vendor/github.com/Microsoft ./vendor/github.com/temp-microsoft && mv ./vendor/github.com/temp-microsoft ./vendor/github.com/microsoft || true
```

Suggesting this change because it doesn't work on Mac as-is. If we try to `mv` the directory without an intermediate rename, `mv` thinks we want to put the directory inside of itself.

kimorris27 (changes requested):

shubhadapaithankar (approved):
On macOS latest changes are working fine without an error .

kimorris27 (dismissed):

shubhadapaithankar (dismissed):

Makefile
```diff
@@ -76,6 +76,9 @@ clean:
 client: generate
 	hack/build-client.sh "${AUTOREST_IMAGE}" 2020-04-30 2021-09-01-preview 2022-04-01 2022-09-04 2023-04-01 2023-07-01-preview 2023-09-04 2023-11-22 2024-08-12-preview
 
+ci-rp: fix-macos-vendor
+	docker build -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) --no-cache=$(NO_CACHE)
```
anshulvermapatel, line 80: There is a dot missing after build, its failing because of that.

shubhadapaithankar (dismissed):

kimorris27 (dismissed):

s-fairchild (approved):
Tested on my Fedora 39 workstation.

rajdeepc2792 (dismissed):

Makefile
```diff
@@ -76,6 +76,9 @@ clean:
 client: generate
 	hack/build-client.sh "${AUTOREST_IMAGE}" 2020-04-30 2021-09-01-preview 2022-04-01 2022-09-04 2023-04-01 2023-07-01-preview 2023-09-04 2023-11-22 2024-08-12-preview
 
+ci-rp: fix-macos-vendor
+	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) --no-cache=$(NO_CACHE)
```
pepedocs, line 80: In the past, a couple of times we (non-ARO-SRE team) have been put into an inconvenient situation where we have to ensure podman also works while a lot of code/commands were already built that specifically used docker. While in theory podman can be a drop-in replacement of docker, the reality is that when docker is used, it's possible to unintentionally accumulate docker-specific features. In this case, when switching to podman, the amount of work necessary can be low to high, in my experience it was high or near high.

I'm not sure about the implications of this but I would highly recommend make the container engine a variable so both podman/docker usage are supporeted and tested.

Dockerfile.ci-rp
```diff
@@ -0,0 +1,48 @@
+ARG REGISTRY
+ARG VERSION
+
+###############################################################################
+# builder is responsible for all compilation and validation of the RP
+###############################################################################
+FROM ${REGISTRY}/ubi8/go-toolset:1.20.10 as builder
```
yjst2012, line 7: go toolset has been bumped recently to 1.20.12-5 in another [PR](https://github.com/Azure/ARO-RP/pull/3532/files) which has been merged, it is probably better to bump it here as well.

yjst2012:
Works on my ubuntu2204

yjst2012 (dismissed):
Thanks for all the efforts of putting this together, it was awesome!

Makefile
SudoBrendan: This is a good point. [Our current documentation for local development setup only includes instructions for podman + podman-docker](https://github.com/Azure/ARO-RP/blob/master/docs/prepare-your-dev-environment.md#software-required), so my guess is that nearly all devs are in a podman ecosystem and testing there first. At the same time, I imagine that our eventual CI pipeline runs will be using docker, because MSFT pipelines and gold-standard templates and build agents have been configured in that engine. Do you anticipate podman having features docker will not be able to support, or only the other way around?

s-fairchild (approved):

shubhadapaithankar (approved):

anshulvermapatel (approved):

kimorris27 (approved):

cadenmarchese (approved):

yjst2012 (approved):

Makefile
pepedocs: @SudoBrendan 

>Do you anticipate podman having features docker will not be able to support, or only the other way around? 

I cannot provide an exact answer for this. However at least in my experience, it's always the other way around. For example, the assumption of rootfull in docker that will not work in podman.

SudoBrendan: I think that's mostly what I've seen as well - where at least podman has all functionality that docker has. If this changes in the future I think we can reassess - but agree this is something to keep in mind as a potential risk.

PR #3559, opened 2024-05-03
Added option to only give ipv4 as output.

anshulvermapatel:
In the steps where we need to white list the local ISP IP by creating a NSG rule, it fails because `$(curl --silent ipecho.net/plain)/32` gives ipv6 IP. Fixed that by adding the option to give ipv4 IP.


SrinivasAtmakuri (approved):

yjst2012 (approved):
LGTM

sankur-codes (approved):
Good catch.
LGTM

mociarain (approved):

PR #3561, opened 2024-05-06
Disable cosmosdb local auth

sankur-codes:

Fixes [ARO-5514](https://issues.redhat.com/browse/ARO-5514)

This PR will disable local authentication on CosmosDB to force use of managed identities for the same implemented under  [ARO-7195](https://issues.redhat.com/browse/ARO-7195).

### Test plan for issue:

NA

NA

NA

### DIFF
--- a/pkg/deploy/assets/rp-development.json
+++ b/pkg/deploy/assets/rp-development.json
@@ -70,6 +70,7 @@
                     "defaultConsistencyLevel": "Strong"
                 },
                 "databaseAccountOfferType": "Standard",
+                "disableLocalAuth": true,
                 "locations": [
                     {
                         "locationName": "[resourceGroup().location]"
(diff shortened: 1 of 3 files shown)


SudoBrendan (approved):

anshulvermapatel (approved):

bennerv (commented):
lgtm, but how are we retrofitting all our environments? 

Have we manually updated role assignments on the aro service principals in all environments?  Who's expected to make the change to the existing coosmosdb instances.

weinong:
@bennerv my understanding is that ARO RP already uses MSI to access cosmosDB. The only thing that didn't was billing which got updated to use MSI in August.

weinong:
https://github.com/Azure/ARO-RP/blob/master/pkg/database/fromenv.go#L25-L36

bennerv:
> @bennerv my understanding is that ARO RP already uses MSI to access cosmosDB. The only thing that didn't was billing which got updated to use MSI in August.

I thought we had a few cosmosdb instances hanging around that we missing permissions in our local dev.  It's possible we still do, and we'll need to retrofit them to work and probably disable local auth on them.  We should also remove usages of master key authorizers because it will no longer work with this change.  

i'm fine with it now though because everything should mostly be in place.

bennerv (approved):

PR #3562, opened 2024-05-06
Migrate dev-proxy-vmss to Azure Linux and enable rolling OS update

nwnt:

Fixes: ARO-7329

Security Sprint/Waves requires that 1p apps start using Azure Linux as its base OS and enable VMSS auto image upgrade.

To enable VMSS rolling upgrades, we need an LB with a healthcheck.  Dev proxy doesn't have a healthcheck endpoint so I added a simple one on port 8080.

### Test plan for issue:
Ensure the deployment is done and e2e runs successfully in INT.  The changes will not be applied to Prod.

No

N/A

### DIFF
--- a/Dockerfile.proxy
+++ b/Dockerfile.proxy
@@ -7,11 +7,13 @@ ENV GOPATH=/root/go
 RUN mkdir -p /app
 WORKDIR /app
 
+COPY . /app
 RUN make proxy
 
 FROM ${REGISTRY}/ubi8/ubi-minimal
 RUN microdnf update && microdnf clean all
-COPY --from=builder /go/src/github.com/Azure/ARO-RP/proxy /usr/local/bin/
+COPY --from=builder /app/proxy /usr/local/bin/
 ENTRYPOINT ["proxy"]
 EXPOSE 8443/tcp
+EXPOSE 8080/tcp
 USER 1000
--- a/pkg/deploy/assets/env-development.json
+++ b/pkg/deploy/assets/env-development.json
@@ -401,9 +401,18 @@
                 },
                 "overprovision": false
             },
+            "identity": {
+                "type": "UserAssigned",
+                "userAssignedIdentities": {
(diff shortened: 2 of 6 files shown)


mociarain (commented):
In principle I think this makes sense but I have some questions:
- What do we do in prod and why do we need a dev-proxy?
- Should docs be updated as part of this?
- As far as I can make out, this code is used in quite a narrow set of places and I'm not even sure we use those any more. Possibly this can all be gutted?

s-fairchild:
> In principle I think this makes sense but I have some questions:
> 
>     * What do we do in prod and why do we need a dev-proxy?
> 
>     * Should docs be updated as part of this?
> 
>     * As far as I can make out this code is in a quite narrow set of places and I'm not even sure we use those anymore. Possibly this can all be gutted?

@mociarain Here's it's purpose, from the ARO Wiki. [Shared environment for local RP development](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/446528/Shared-environment-for-local-RP-development)

> dev-proxy-vmss - This is needed as a proxy for your local PR to access the k8s control plane endpoints while the cluster is being provisioned.

I agree, more documentation on this VMSS would be nice.

nwnt:
@mociarain production is not needed as this is used in dev/INT only.  dev-proxy-vmss can't be gutted yet mainly because it's still being used by non-prod environments, especially CI-e2e.  

to be honest, I'm not an expert about this component myself.... not sure if I'm the right person to write the doc for it.  Can we track doc improvement in a separate ticket instead?

hlipsig (approved):

kimorris27:
@nwnt @mociarain , I think the proxy VMSS is sufficiently covered in documentation already. See here: https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/166757/Debug-E2E

I also added a link to this "Debug-E2E" page on the one that @s-fairchild linked to.

hlipsig:
I'm still just waiting for CI to report successful. It's okay to merge whenever that happens.

kimorris27 (approved):

PR #3563, opened 2024-05-07
ARO-6425 v20240812preview validation

yithian:

For ARO-6425

Adds static validation for new API fields added in v20240812preview

### Test plan for issue:

Includes unit tests for the new static validation features.

No, this is just static validation to ensure inputs are as expected.

Unit test data has been written to match expected inputs but nonetheless this should be tested in a lower environment before being pushed to production

### DIFF
--- a/pkg/api/v20240812preview/openshiftcluster_convert.go
+++ b/pkg/api/v20240812preview/openshiftcluster_convert.go
@@ -139,12 +139,6 @@ func (c openShiftClusterConverter) ToExternal(oc *api.OpenShiftCluster) interfac
 		}
 	}
 
-	if oc.Properties.ServicePrincipalProfile != nil {
-		out.Properties.ServicePrincipalProfile = &ServicePrincipalProfile{}
-		out.Properties.ServicePrincipalProfile.ClientID = oc.Properties.ServicePrincipalProfile.ClientID
-		out.Properties.ServicePrincipalProfile.ClientSecret = string(oc.Properties.ServicePrincipalProfile.ClientSecret)
-	}
-
 	if oc.Properties.PlatformWorkloadIdentityProfile != nil && oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities != nil {
 		out.Properties.PlatformWorkloadIdentityProfile = &PlatformWorkloadIdentityProfile{}
 		out.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities = make([]PlatformWorkloadIdentity, len(oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities))
--- a/pkg/api/v20240812preview/openshiftcluster_convert.go
+++ b/pkg/api/v20240812preview/openshiftcluster_convert.go
@@ -345,8 +345,10 @@ func (c openShiftClusterConverter) ExternalNoReadOnly(_oc interface{}) {
 	for i := range oc.Properties.IngressProfiles {
 		oc.Properties.IngressProfiles[i].IP = ""
 	}
-	for i := range oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
-		oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities[i].ClientID = ""
-		oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities[i].ObjectID = ""
+	if oc.Properties.PlatformWorkloadIdentityProfile != nil {
+		for i := range oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
+			oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities[i].ClientID = ""
+			oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities[i].ObjectID = ""
+		}
 	}
 }
--- a/pkg/api/v20240812preview/openshiftcluster_convert.go
+++ b/pkg/api/v20240812preview/openshiftcluster_convert.go
@@ -139,6 +139,12 @@ func (c openShiftClusterConverter) ToExternal(oc *api.OpenShiftCluster) interfac
 		}
 	}
 
+	if oc.Properties.ServicePrincipalProfile != nil {
+		out.Properties.ServicePrincipalProfile = &ServicePrincipalProfile{}
+		out.Properties.ServicePrincipalProfile.ClientID = oc.Properties.ServicePrincipalProfile.ClientID
+		out.Properties.ServicePrincipalProfile.ClientSecret = string(oc.Properties.ServicePrincipalProfile.ClientSecret)
+	}
+
 	if oc.Properties.PlatformWorkloadIdentityProfile != nil && oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities != nil {
 		out.Properties.PlatformWorkloadIdentityProfile = &PlatformWorkloadIdentityProfile{}
 		out.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities = make([]PlatformWorkloadIdentity, len(oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities))
--- a/pkg/api/v20240812preview/openshiftcluster_validatestatic.go
+++ b/pkg/api/v20240812preview/openshiftcluster_validatestatic.go
@@ -449,14 +449,11 @@ func (sv openShiftClusterStaticValidator) validatePlatformWorkloadIdentityProfil
 
 func (sv openShiftClusterStaticValidator) validateIdentityXORServiceCredentials(pwip *PlatformWorkloadIdentityProfile, spPath string, spp *ServicePrincipalProfile) error {
 	// nothing to validate if the platform workload identity profile is nil
-	if pwip == nil {
-		if spp == nil {
-			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, spPath, "Must provide either an identity or service principal credentials.")
-		}
-		return nil
+	if pwip == nil && spp == nil {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, spPath, "Must provide either an identity or service principal credentials.")
 	}
 
-	if spp != nil && (spp.ClientID != "" || spp.ClientSecret != "") {
+	if pwip != nil && spp != nil && (spp.ClientID != "" || spp.ClientSecret != "") {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, spPath, "Cannot use identities and service principal credentials at the same time.")
 	}
 
--- a/pkg/api/v20240812preview/openshiftcluster_validatestatic.go
+++ b/pkg/api/v20240812preview/openshiftcluster_validatestatic.go
@@ -79,7 +79,7 @@ func (sv openShiftClusterStaticValidator) validate(oc *OpenShiftCluster, isCreat
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "location", "The provided location '%s' is invalid.", oc.Location)
 	}
 
-	if err := sv.validatePlatformIdentitiesExist("identity", oc); err != nil {
+	if err := sv.validatePlatformIdentities(oc); err != nil {
 		return err
 	}
 
@@ -116,9 +116,6 @@ func (sv openShiftClusterStaticValidator) validateProperties(path string, p *Ope
 		return err
 	}
 	if err := sv.validatePlatformWorkloadIdentityProfile(path+".platformWorkloadIdentityProfile", p.PlatformWorkloadIdentityProfile); err != nil {
-		return err
-	}
-	if err := sv.validateIdentityXORServiceCredentials(p.PlatformWorkloadIdentityProfile, path+".servicePrincipalProfile", p.ServicePrincipalProfile); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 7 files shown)


pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -412,3 +423,60 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
...
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, spPath, "Must provide either an identity or service principal credentials.")
+		}
+		return nil
+	}
+
+	if spp != nil && (spp.ClientID != "" || spp.ClientSecret != "") {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, spPath, "Cannot use identities and service principal credentials at the same time.")
+	}
+
+	return nil
```
bitoku, line 463: This might be easier to understand that this is XOR, isn't it?

```suggestion
	if pwip == nil && spp == nil {
		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, spPath, "Must provide either an identity or service principal credentials.")
	}

	if pwip != nil && spp != nil && (spp.ClientID != "" || spp.ClientSecret != "") {
		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, spPath, "Cannot use identities and service principal credentials at the same time.")
	}

	return nil
```

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -412,3 +423,57 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
...
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, fmt.Sprintf("%s.PlatformWorkloadIdentities[%d].resourceID", path, n), "Resource must be a user assigned identity.")
+		}
+	}
+	return nil
+}
+
+func (sv openShiftClusterStaticValidator) validateIdentityXORServiceCredentials(pwip *PlatformWorkloadIdentityProfile, spPath string, spp *ServicePrincipalProfile) error {
+	// nothing to validate if the platform workload identity profile is nil
+	if pwip == nil && spp == nil {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, spPath, "Must provide either an identity or service principal credentials.")
```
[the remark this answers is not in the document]
yithian: added a unit test for this

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -412,3 +423,57 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
...
+	}
+
+	// Validate the PlatformWorkloadIdentities
+	for n, p := range pwip.PlatformWorkloadIdentities {
+		if !stringutils.IsResourceIDFormatted(p.ResourceID) {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, fmt.Sprintf("%s.PlatformWorkloadIdentities[%d].resourceID", path, n), "ResourceID %s formatted incorrectly.", p.ResourceID)
+		}
+
+		if p.OperatorName == "" {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, fmt.Sprintf("%s.PlatformWorkloadIdentities[%d].resourceID", path, n), "Operator name is empty.")
```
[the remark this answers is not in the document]
yithian: added a unit test for this

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -412,3 +423,57 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
...
+	}
+
+	if pwip != nil && spp != nil && (spp.ClientID != "" || spp.ClientSecret != "") {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, spPath, "Cannot use identities and service principal credentials at the same time.")
+	}
+
+	return nil
+}
+
+func (sv openShiftClusterStaticValidator) validatePlatformIdentitiesExist(path string, oc *OpenShiftCluster) error {
```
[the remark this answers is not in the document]
yithian: I combined these two functions into one. It's not the prettiest code I've ever written but it should get the job done

pkg/api/v20240812preview/openshiftcluster_convert.go
```diff
@@ -139,6 +139,12 @@ func (c openShiftClusterConverter) ToExternal(oc *api.OpenShiftCluster) interfac
 		}
 	}
 
+	if oc.Properties.ServicePrincipalProfile != nil {
```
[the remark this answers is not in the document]
yithian: Removed!

pkg/util/stringutils/stringutils.go
```diff
@@ -18,3 +21,18 @@ func Contains(list []string, value string) bool {
...
+	// the regexp is statically defined and hard-coded, we shouldn't need to check for errors
+	match, _ := regexp.MatchString(resourceIDPattern, id)
+
+	return match
+}
+
+// ResourceType takes a resourceID as a string and returns the resource type as a string
+func ResourceType(resourceID string) string {
+	return strings.Split(resourceID, "/")[7]
+}
```
bitoku, line 38: It might be a nit pick but `stringutil` sounds like utilities for general string operations (like `LastTokenByte`).
These two functions are domain-specific string operations, and it seems natural for me to be written in a different place.

yithian: Fair enough; I removed those functions since they were only being used in one place (each) anyways. They're simple enough that it doesn't really affect readability of the code blocks where they were being called

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -412,3 +420,58 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
...
+		for _, pwi := range pwip.PlatformWorkloadIdentities {
+			if pwi.OperatorName != "" {
+				operatorRolePresent = true
+			}
+		}
+	}
+
+	if clusterIdentityPresent != operatorRolePresent {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "identity", "Cluster identity and operator roles require each other.")
+	}
```
bitoku, line 475: sorry for sporadic comments.
This is just a question, but in what case `operatorName` would be empty?

pkg/api/v20240812preview/openshiftcluster_validatestatic_test.go
```diff
@@ -1161,3 +1161,152 @@ func TestOpenShiftClusterStaticValidateDelta(t *testing.T) {
...
+			},
+		},
+		{
+			name: "valid workloadIdentityProfile",
+			modify: func(oc *OpenShiftCluster) {
+				oc.Properties.PlatformWorkloadIdentityProfile = &PlatformWorkloadIdentityProfile{
+					PlatformWorkloadIdentities: []PlatformWorkloadIdentity{
+						{
+							OperatorName: "FAKE-OPERATOR",
+							ResourceID:   "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/a-fake-group/providers/Microsoft.RedHatOpenShift/userAssignedIdentities/fake-cluster-name",
```
bitoku, line 1180: should this be `Microsoft.ManagedIdentity/userAssignedIdentities`?
same for other test cases.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
yithian: I'm not sure, to be honest, but this scenario was called out as something to validate for. I'm happy to make changes to this validation if you have feedback though!

pkg/api/v20240812preview/openshiftcluster_validatestatic_test.go
yithian: The regexp the resourceID is being compared against is at [pkg/api/v20240812preview/openshiftcluster_validatestatic.go line 430](https://github.com/Azure/ARO-RP/pull/3563/files/ca9536ab55648929c0b9e424f31978bcbd1baada#diff-5a28605fb092ea2ef41d370f17daac8ce3c72e7c060a83d62f8aa0366bdfdd4aR430)

The specific provider for the resource ID isn't being checked, just that the string is formatted like a resource ID at all. I can change the value in the test cases to `Microsoft.ManagedIdentity` if you'd like

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
cadenmarchese: We want to ensure that if a user passes in identities for an operator during a cluster create, the top-level cluster identity exists, otherwise the installation itself will fail because without the top-level cluster identity (`oc.Identity`) the RP won't be able to generate federated credentials for each operator identity.

This does raise an interesting case, though, which is the `az aro update`. With this validation present, I think that the user will have to also pass in the cluster identity along with it if providing a new identity to an existing cluster, even if the cluster identity details haven't changed.

@kimorris27 @tsatam do you have thoughts on how to best proceed?

kimorris27: My reading of this conversation is that there are actually two different topics that ended up in this same thread, so I'll address them one at a time.

>This is just a question, but in what case operatorName would be empty?

In the context of this function, `validatePlatformIdentities` I think we can skip the check for empty `OperatorName` fields because we will be checking the same thing in `validatePlatformWorkloadIdentityProfile`. With that in mind, the section Ayato asked about:

```
operatorRolePresent := false
if pwip != nil {
	for _, pwi := range pwip.PlatformWorkloadIdentities {
		if pwi.OperatorName != "" {
			operatorRolePresent = true
		}
	}
}
```

can be replaced with `operatorRolePresent := pwip != nil` to simplify the code and avoid checking the same thing twice.

>This does raise an interesting case, though, which is the az aro update. With this validation present, I think that the user will have to also pass in the cluster identity along with it if providing a new identity to an existing cluster, even if the cluster identity details haven't changed.

>@kimorris27 @tsatam do you have thoughts on how to best proceed?

I think this code is fine as-is and we won't encounter this issue, since the RP responds to `PATCH` requests by getting the cluster doc from Cosmos DB and merging in any new fields the customer has provided prior to static validation. See https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/openshiftcluster_putorpatch.go#L61-L172.

yithian: I've pushed a commit to simply the code as suggested; thanks!

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -412,3 +420,51 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
...
 	return nil
 }
+
+func (sv openShiftClusterStaticValidator) validatePlatformWorkloadIdentityProfile(path string, pwip *PlatformWorkloadIdentityProfile) error {
+	// PlatformWorkloadIdentityProfile being empty is acceptable
+	if pwip == nil {
+		return nil
+	}
+
+	resourceIDPattern, _ := regexp.Compile("/subscriptions/.*/resourceGroups/.*/providers/.*/.*/.*")
```
tsatam, line 430: I think the regex-based validation is fine as it is, but we also maintain a parser that converts a resourceID into a typed struct: https://github.com/Azure/ARO-RP/blob/master/pkg/util/arm/resources.go

cadenmarchese: +1 on using the shared code at https://github.com/Azure/ARO-RP/blob/master/pkg/util/arm/resources.go here.

pkg/util/arm/resources.go
```diff
@@ -43,24 +42,29 @@ func (r ArmResource) String() string {
 // ParseArmResourceId take the resourceID of a child resource to an OpenShiftCluster
 // TODO refactor this function to support an additional layer of child resources if we ever get to that point, right now only supports 1 child resource
```
kimorris27, line 43: This `TODO` comment can be removed since the function now supports an additional level of subresource nesting.

pkg/util/arm/resources.go
```diff
@@ -43,24 +42,29 @@ func (r ArmResource) String() string {
 // ParseArmResourceId take the resourceID of a child resource to an OpenShiftCluster
```
kimorris27, line 42:
```suggestion
// ParseArmResourceId takes in an ARM resource ID and returns an ArmResource object representing that resource. It supports up to two levels of subresource nesting.
```

Not related to your work in this PR, but I noticed this comment wasn't a full sentence, so I'm proposing a better doc comment.

pkg/util/arm/resources.go
```diff
@@ -43,24 +42,29 @@ func (r ArmResource) String() string {
...
 // TODO refactor this function to support an additional layer of child resources if we ever get to that point, right now only supports 1 child resource
 func ParseArmResourceId(resourceId string) (*ArmResource, error) {
```
kimorris27, line 44: Please update the unit test cases for this function to reflect the added support for two levels of subresource nesting.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -412,3 +420,49 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
...
+
+	if pwip != nil && spp != nil && (spp.ClientID != "" || spp.ClientSecret != "") {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "properties.servicePrincipalProfile", "Cannot use identities and service principal credentials at the same time.")
+	}
+
+	clusterIdentityPresent := oc.Identity != nil
+	operatorRolePresent := pwip != nil
+
+	if clusterIdentityPresent != operatorRolePresent {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "identity", "Cluster identity and operator roles require each other.")
```
kimorris27, line 465: Kind of a nit since it doesn't affect functionality, but how about changing the terminology used here from "operator roles" to "platform workload identities" to align with the terminology used in the API? The error message may be confusing to customers otherwise.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -412,3 +420,49 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
...
 }
+
+func (sv openShiftClusterStaticValidator) validatePlatformWorkloadIdentityProfile(path string, pwip *PlatformWorkloadIdentityProfile) error {
+	// PlatformWorkloadIdentityProfile being empty is acceptable
+	if pwip == nil {
+		return nil
+	}
+
+	// Validate the PlatformWorkloadIdentities
+	for n, p := range pwip.PlatformWorkloadIdentities {
```
kimorris27, line 431: @cadenmarchese @rajdeepc2792, should we be validating `ClientID` and `ObjectID` within this loop too? I think we determined a while back that the RP will be able to fill these two fields in on its own given the MI's resource ID, but I'm still a bit unsure - even if we do fill in these fields on behalf of the cx, is that magic going to happen in the RP or the CLI? If it happens in the CLI, we should add the two fields to static validation.

@tsatam , maybe you have some thoughts on this too?

kimorris27 (changes requested):
Made another review pass since there are a lot of details to inspect and you've made some changes since my last review. I suggested a few small changes and started a group discussion about a piece that I'm unsure about.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
tsatam: CLI is planning to accept only resourceIDs - while we _could_ infer client/objectIDs for a given MI resource ID CLI-side, it would be better if it was handled RP-side for consistency across non-CLI usage as well as correctness guarantees.

kimorris27: Okay, then we can leave them out of static validation, since it wouldn't make sense to validate a field that the CLI won't be passing to the RP.

No changes needed here @yithian , and I'll resolve the conversation.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
yithian: Good idea, done

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -412,3 +420,51 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
...
+		match := resourceIDPattern.Match([]byte(p.ResourceID))
+		if !match {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, fmt.Sprintf("%s.PlatformWorkloadIdentities[%d].resourceID", path, n), "ResourceID %s formatted incorrectly.", p.ResourceID)
+		}
+
+		if p.OperatorName == "" {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, fmt.Sprintf("%s.PlatformWorkloadIdentities[%d].resourceID", path, n), "Operator name is empty.")
+		}
+
+		if strings.Split(p.ResourceID, "/")[7] != "userAssignedIdentities" {
```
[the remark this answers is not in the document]
yithian: Summarizing slack conversation here for posterity:

The Azure SDK is now being used to parse the resource ID [above](https://github.com/Azure/ARO-RP/pull/3563/files#diff-5a28605fb092ea2ef41d370f17daac8ce3c72e7c060a83d62f8aa0366bdfdd4aR432) and the resourcetype [is being validated](https://github.com/Azure/ARO-RP/pull/3563/files#diff-5a28605fb092ea2ef41d370f17daac8ce3c72e7c060a83d62f8aa0366bdfdd4aR441) using the Azure SDK resource types (golang types, that is)

pkg/api/v20240812preview/openshiftcluster_convert.go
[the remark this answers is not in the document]
cadenmarchese: @yithian mind revisiting this comment? I still see this code in the diff as well as the branch.

kimorris27 (dismissed):
LGTM, thanks for being flexible and responding quickly to all of the suggestions!

kimorris27:
Thanks for being flexible and responding to all of the suggestions quickly. Will LGTM once Caden's last comment is addressed.

pkg/api/v20240812preview/openshiftcluster_convert.go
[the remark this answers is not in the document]
yithian: Sorry, overlooked this. Removed for real this time

kimorris27 (approved):

PR #3564, opened 2024-05-07
Create and populate OIDC blob store for the cluster

rajdeepc2792:

Jira issue :- [ARO-4373](https://issues.redhat.com/browse/ARO-4373)
Related Docs:-
https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/603739/OIDC-Traffic-Flow
Similar Implementation:-
https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/tree/master/pkg/aws/cloudcredentialbuilder

- The RP needs to generate a keypair
- The private key needs to be stored in cosmos, It needs to be encrypted if it's stored in cosmos
- The private key will be passed to the installer-aro-wrapper via a secret as part of the existing cluster document in CosmosDB(Not implemented here)
- Set / populate OIDC issuerURL in the internal API
- The RP should generate OIDC docs
- Should populate them in regional storage accounts
- The OIDC doc needs to be resolvable through the regional AFD instances and should follow a hashing pattern

### Test plan for issue:
- [x] Unit Tests
- [x] Full RP Cluster Creation
- [x] CI
- [x] CI e2e
- [x] Local cluster installation
- [x] INT Deployment / e2e

- Doc for setting up the full RP. (Part of the PR)

Testing the implementation in all the environments.


pkg/util/storage/manager.go
```diff
@@ -82,3 +86,41 @@ func (m *manager) BlobService(ctx context.Context, resourceGroup, account string
...
+	}
+
+	_, err = m.blobContainers.Create(
+		context.Background(),
+		resourceGroup,
+		account,
+		container,
+		mgmtstorage.BlobContainer{
+			ContainerProperties: &mgmtstorage.ContainerProperties{
+				PublicAccess: mgmtstorage.PublicAccessNone,
```
cadenmarchese, line 115: I believe we need to keep public access open if development mode is set, because we won't have an AFD instance. Otherwise, we should keep `PublicAccessNone`.

pkg/util/storage/manager.go
```diff
@@ -82,3 +86,41 @@ func (m *manager) BlobService(ctx context.Context, resourceGroup, account string
 
 	return &blobcli, nil
 }
+
+func (m *manager) CreateBlobContainer(ctx context.Context, resourceGroup string, account string, container string) error {
```
cadenmarchese, line 90: Do we need to create a storage account in our e2e environment (RP health subscription)? I believe they were only created in shared dev, canary, and INT environments.

cadenmarchese (commented):
not a full review, just a few questions as you're working through the draft and testing. looks great!

pkg/util/storage/manager.go
rajdeepc2792: I tried testing it in the local dev rp and it worked, maybe because I used the localFPAuthorizer, but I do have this access noted that it has to tested and be applied as per the environments. Starting with the lowest access available.
Thanks for noting this.

pkg/util/storage/manager.go
rajdeepc2792: Yes good point, we will need the storage account in e2e too. Currently updating the secrets for e2e to add `STORAGE_ACCOUNT_DOMAIN` to form the correct storage account name.

pkg/api/openshiftcluster.go
```diff
@@ -258,12 +258,13 @@ type OIDCIssuer string
...
-	Version              string               `json:"version,omitempty"`
-	ResourceGroupID      string               `json:"resourceGroupId,omitempty"`
-	FipsValidatedModules FipsValidatedModules `json:"fipsValidatedModules,omitempty"`
-	OIDCIssuer           OIDCIssuer           `json:"oidcIssuer,omitempty"`
+	PullSecret                    SecureString         `json:"pullSecret,omitempty"`
+	Domain                        string               `json:"domain,omitempty"`
+	Version                       string               `json:"version,omitempty"`
+	ResourceGroupID               string               `json:"resourceGroupId,omitempty"`
+	FipsValidatedModules          FipsValidatedModules `json:"fipsValidatedModules,omitempty"`
+	OIDCIssuer                    OIDCIssuer           `json:"oidcIssuer,omitempty"`
```
bennerv, line 266: should this be a pointer?

pkg/api/openshiftcluster.go
```diff
@@ -258,12 +258,13 @@ type OIDCIssuer string
...
+	BoundServiceAccountSigningKey SecureString         `json:"boundServiceAccountSigningKey,omitempty"`
```
bennerv, line 267: we're not going to populate this on older clusters, correct?  So this field should only be set on WI clusters?

pkg/cluster/deploybaseresources.go
```diff
@@ -35,6 +37,35 @@ func (m *manager) createDNS(ctx context.Context) error {
 	return m.dns.Create(ctx, m.doc.OpenShiftCluster)
 }
 
+func (m *manager) createOIDC(ctx context.Context) error {
+	blobContainerName := env.OIDCBlobContainerPrefix + m.doc.ID
```
bennerv, line 41: Missing condition at top for idempotency.  
```
if m.doc.OpenShiftCluster.Properties.ClusteRProfile.OIDCIssuer != nil {
    return nil
}

Comment (bennerv) [reply pkg/cluster/deploybaseresources.go:41]: This may need to change if we support rotating the SA signing key.  I would say out of scope for now, but to consider for future.

File: pkg/cluster/install.go [hunk 6]
```diff
@@ -280,6 +280,8 @@ func (m *manager) bootstrap() []steps.Step {
 		steps.Action(m.populateMTUSize),
 
 		steps.Action(m.createDNS),
+		// TODO: Add a condition on ServicePrincipalProfile to run createOIDC after testing
+		steps.Action(m.createOIDC),
```
bennerv, line 284: Are we allowing customers to update their clusters to consume OIDC / workload identities that the customer can specify?  IF so, are we doing that here?  

`az aro update --enable-oidc` something like that.

pkg/env/env.go
```diff
@@ -58,6 +58,7 @@ const (
 	ServiceKeyvaultSuffix            = "-svc"
 	RPPrivateEndpointPrefix          = "rp-pe-"
 	ProxyHostName                    = "PROXY_HOSTNAME"
+	OIDCBlobContainerPrefix          = "oic-"
```
bennerv, line 61: is this already decided on for name?  All our other storage accounts are `-<subject>` rather than `<subject>-`. 

No worries if it's already done.

pkg/frontend/asyncoperationresult_get.go
```diff
@@ -64,6 +64,7 @@ func (f *frontend) _getAsyncOperationResult(ctx context.Context, r *http.Request
 
 	asyncdoc.OpenShiftCluster.Properties.ClusterProfile.PullSecret = ""
 	asyncdoc.OpenShiftCluster.Properties.ServicePrincipalProfile.ClientSecret = ""
+	doc.OpenShiftCluster.Properties.ClusterProfile.BoundServiceAccountSigningKey = ""
```
bennerv, line 67: `doc` or `asyncdoc` ?

pkg/util/oidcbuilder/jwks.go
```diff
@@ -0,0 +1,136 @@
...
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"gopkg.in/square/go-jose.v2"
+)
+
+func CreateKeyPair() (encPrivateKey []byte, encPublicKey []byte, err error) {
+	bitSize := 4096
```
bennerv, line 21: is this the bit size & recommendation of OpenShift for SA private keys?

pkg/api/openshiftcluster.go
rajdeepc2792: This comes from the changes proposed in [Preview API design doc](https://docs.google.com/document/d/1AIuf6H9y0Jjboywp0NgD8-9lU1jCj5WrE83jh58rEg4/edit#heading=h.cifwpl8od05m), and were pushed as part of [Preview API fields addition PR](https://github.com/Azure/ARO-RP/pull/3478/files#diff-c7bfe4dfde8a1a575263edc226847cbab8ac7f60f04193f531a17d73571a3398R81).
Not sure why this field need to be a pointer. The MIWI cluster check would be done on ServicePrincipalProfile field which is changed to pointer.

pkg/api/openshiftcluster.go
rajdeepc2792: Yes this value will only be generated and populated for MIWI clusters as part of OIDC blob creations for the clusters.

pkg/cluster/deploybaseresources.go
rajdeepc2792: Since during the cluster installation OIDCIssuer is set as part of the createOIDC function, below check wouldn't allow the OIDCIssuer configuration.
```
if m.doc.OpenShiftCluster.Properties.ClusteRProfile.OIDCIssuer != nil {
    return nil
}
```

The check if the cluster is MIWI was supposed to be performed here:- https://github.com/Azure/ARO-RP/pull/3564/files#diff-71a1dd610bf07a9e4ece9f782af2bad11e6f2d0e03df442019088f8c47265497R283

With a nil check on `m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile`.
I will now make the add the check in the function itself:-

```
if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile != nil {
    return nil
}
```

pkg/cluster/install.go
rajdeepc2792: I don't think we have the scope around customer providing the OIDC IssuerURL, it will always be managed and set by RP.
@cadenmarchese can confirm.

pkg/env/env.go
rajdeepc2792: Storage account name is already determined by the STORAGE_ACCOUNT_DOMAIN env variable.
This prefix is used here for creating a container inside the OIDC storage account. But this naming still needs a decision.
Decision here is made from the container naming convention used for bootdiagnostic files in arogwy<region> storage accounts. They have a naming convention as `bootdiagnostics-gatwayvm-<ID>`, also the Private Endpoints do have one-one mapping between cluster and PE and they also uses naming like `rp-pe-<docID>`.

pkg/frontend/asyncoperationresult_get.go
rajdeepc2792: Changing it. Thanks.

pkg/util/oidcbuilder/jwks.go
rajdeepc2792: Creation of jwks is mostly copied from [Cloud-credential-operator](https://github.com/openshift/cloud-credential-operator/blob/master/pkg/cmd/provisioning/create_key_pair.go#L35) same as [ROSA STS](https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/tree/master/pkg/aws/cloudcredentialbuilder)

pkg/util/storage/manager.go
rajdeepc2792: This has been taken care of and it will always be `PublicAccessNone` for non-dev, otherwise `PublicAccessNone`. It can only be tested once we create the MIWI cluster after all the changes ready.

pkg/cluster/delete.go
```diff
@@ -441,6 +441,12 @@ func (m *manager) Delete(ctx context.Context) error {
 		return err
 	}
 
+	m.log.Printf("deleting OIDC configuration")
+	err = m.rpBlob.DeleteBlobContainer(ctx, m.env.ResourceGroup(), m.env.OIDCStorageAccountName(), env.OIDCBlobContainerPrefix+m.doc.ID)
```
niontive, line 445: For service principal clusters, wouldn't this return an error?

rajdeepc2792: Thanks, yes this needs the same condition as during creation. Will change it after the remaining testing it in CI e2e.

pkg/env/prod.go
```diff
@@ -261,6 +273,19 @@ func (p *prod) ACRDomain() string {
 	return p.acrDomain
 }
 
+func (p *prod) OIDCStorageAccountName() string {
```
niontive, line 276: My main concern is we'll have to maintain the storage account name in two separate places. Ex: https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO-Pipelines?path=/rp/oidc/Region/Templates/modules/storage.bicep

Would you be good if we:
- Define the storage account names in `rp-config`
- Have ARO-Pipelines and ARO-RP consume that name from `rp-config`

pkg/util/oidcbuilder/jwks.go
niontive: 4096 should be ok for FIPS compliance as well: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Br2.pdf

niontive: Also - would it be possible to have the cloud credential operator folks publish functions for generating key pairs, so that way ROSA and ARO can consume those directly instead of repeating the same code?

pkg/util/oidcbuilder/jwks.go
```diff
@@ -0,0 +1,136 @@
...
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"gopkg.in/square/go-jose.v2"
```
niontive, line 17: Looks like Square deprecated this and it's maintained here now: https://github.com/go-jose/go-jose/blob/main/go.mod

pkg/util/oidcbuilder/jwks.go
```diff
@@ -0,0 +1,136 @@
+package oidcbuilder
```
niontive, line 1: No `jwks_test.go`?

pkg/env/prod.go
rajdeepc2792: Thanks Nic for raising this, I agree and back your idea for the consistency. It was hard anyways to get the container naming using the assumption present in the function.
Although I would like to get the consensus on the this.
cc: @cadenmarchese

pkg/util/oidcbuilder/jwks.go
rajdeepc2792: The majority of the flow was already tested via the oidcbuilder functions, thus kept only a single test file for this package.

pkg/util/oidcbuilder/jwks.go
cadenmarchese: @niontive they do actually: https://github.com/openshift/cloud-credential-operator/blob/master/docs/ccoctl.md#creating-rsa-keys-1 It's a CLI tool though, it's a bit different for us since we own the OIDC stuff.

pkg/cluster/install.go
cadenmarchese: It might be in-scope eventually, but for now I think we can table it.

pkg/api/openshiftcluster.go
bennerv: Do you want to make it a pointer then?

rajdeepc2792: Changed it SecureBytes now.

pkg/cluster/deploybaseresources.go
rajdeepc2792: Uncommented the condition now for making `createOIDC` idempotent.

pkg/cluster/delete.go
rajdeepc2792: Added the condition for the MIWI cluster for OIDC deletion.

pkg/env/prod.go
rajdeepc2792: As per discussion, this is going to be handle as a separate issue after the PR is merged. I will post the Jira story for the same here when created.

pkg/util/oidcbuilder/jwks.go
rajdeepc2792: Avoided the direct dependency on square but the `github.com/coreos/go-oidc` still uses it indirectly.

pkg/util/oidcbuilder/oidcbuilder.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (b *OIDCBuilder) GetPrivateKey() string {
+	return string(b.privateKey)
+}
+
+func (b *OIDCBuilder) GetBlobContainerURL() string {
+	return b.blobContainerURL
+}
+
+func populateOidcFolder(ctx context.Context, body string, jwks []byte, azBlobClient utilazblob.AZBlobClient) error {
```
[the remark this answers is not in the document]
rajdeepc2792: Changed `body` -> `discoveryDocument`.

pkg/cluster/cluster.go
```diff
@@ -169,6 +176,11 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 		return nil, err
 	}
 
+	rpBlob, err := azblob.NewManager(_env, _env.SubscriptionID(), msiCredential)
```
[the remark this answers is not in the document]
rajdeepc2792: Changed it to `rpBlob, err := azblob.NewManager(_env.Environment(), _env.SubscriptionID(), msiCredential)`
Haven't passed `_env` directly as there can be a use case where different SubscriptionID is needed.

pkg/util/azureclient/mgmt/storage/blobcontainers.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"context"
+
+	mgmtstorage "github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-06-01/storage"
+	"github.com/Azure/go-autorest/autorest"
+
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+)
+
+type BlobContainersClient interface {
```
kimorris27, line 15: Is this being used? I couldn't find any references to it. Could it be removed?

pkg/cluster/deploybaseresources_test.go
```diff
@@ -1394,3 +1399,214 @@ func TestNewPublicLoadBalancer(t *testing.T) {
 		})
 	}
 }
+
+func TestCreateOIDC(t *testing.T) {
```
kimorris27, line 1403: You could consider adding two more test cases:

- Fail to create `OIDCBuilder`
- `OIDCBuilder` fails to populate OIDC blob

pkg/util/oidcbuilder/jwks.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to generate private key")
+	}
+	encodedPrivateKey := pem.EncodeToMemory(&pem.Block{
+		Type:    "RSA PRIVATE KEY",
+		Headers: nil,
+		Bytes:   x509.MarshalPKCS1PrivateKey(privateKey),
+	})
+
+	// Generate public key from private keypair
```
kimorris27, line 34: I would reword this comment for clarity's sake. IIUC what's happening here is that the keypair has already been generated and you are simply putting the public key data into a byte array.
```suggestion
	// Serialize public key into a byte array to prepare to store it in the OIDC storage blob
```

pkg/util/oidcbuilder/jwks.go
```diff
@@ -0,0 +1,136 @@
...
+	pubKeyBytes, err := x509.MarshalPKIXPublicKey(&privateKey.PublicKey)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to generate public key from private")
```
kimorris27, line 37:
```suggestion
		return nil, nil, errors.Wrapf(err, "failed to serialize public key")
```

See my comment a few lines above this for the reasoning behind this suggestion.

pkg/util/oidcbuilder/oidcbuilder_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/golang/mock/gomock"
+
+	mock_azblob "github.com/Azure/ARO-RP/pkg/util/mocks/azblob"
+	utilerror "github.com/Azure/ARO-RP/test/util/error"
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+func TestEnsureOIDCDocs(t *testing.T) {
```
kimorris27, line 22: You could consider adding a few more test cases here to cover some other ways in which the code could fail:

- `Public key is not of type RSA` (the `default` case in the `switch`)
- `Failed to fetch key ID from public key`
- `JSON encoding of web key set failed` (if you can - I'm not sure if you can easily test this)

kimorris27 (commented):
Nice work! I just made some small suggestions here and there.

pkg/cluster/deploybaseresources.go
```diff
@@ -35,6 +38,47 @@ func (m *manager) createDNS(ctx context.Context) error {
...
+		return nil
+	}
+
+	blobContainerName := env.OIDCBlobContainerPrefix + m.doc.ID
+
+	publicAccess := azstorage.PublicAccessNone
+	// Public access on OIDC Container needed for development environments because of no AFD avaialbility
+	if m.env.IsLocalDevelopmentMode() {
+		publicAccess = azstorage.PublicAccessBlob
+	}
```
bennerv, line 52: This would be better as a feature flag on the RP rather than an `IsLocalDevelopmentMode`.  

It would allow us to toggle the functionality in INT if needed or if we needed to decouple from AFD in the future.

pkg/util/azblob/manager.go
```diff
@@ -0,0 +1,125 @@
...
+		resourceGroup,
+		accountName,
+		containerName,
+		&azstorage.BlobContainersClientGetOptions{},
+	)
+	if err != nil {
+		if !bloberror.HasCode(err, bloberror.ContainerNotFound) {
+			return err
+		}
+		needToCreateBlobContainer = true
```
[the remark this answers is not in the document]
rajdeepc2792: `nil` can't be returned here, as the container is need to be created if the error is `ContainerNotFound`.

pkg/util/oidcbuilder/oidcbuilder_test.go
rajdeepc2792: Thanks!!
Added unit tests for the top 2 suggestions, but was not able to find any easy way to test the third `JSON encoding of web key set failed`.

pkg/util/oidcbuilder/jwks.go
rajdeepc2792: Done.

pkg/cluster/deploybaseresources_test.go
rajdeepc2792: Was not able to find an easy way to test `Fail to create OIDCBuilder` as there's no mock present for the `rsa.GenerateKey`

pkg/util/azureclient/mgmt/storage/blobcontainers.go
rajdeepc2792: Thanks for pointing, it is removed now.

pkg/util/azblob/manager.go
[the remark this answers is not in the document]
kimorris27: Oops, I got a little confused and made an incorrect recommendation, but I think you can still make this function a little easier by simplifying the logic down. Here's my recommendation (I tried it out locally to make sure it passes unit tests):

```
func (m *manager) CreateBlobContainer(ctx context.Context, resourceGroup string, accountName string, containerName string, publicAccess azstorage.PublicAccess) error {
	_, err := m.blobContainer.Get(
		ctx,
		resourceGroup,
		accountName,
		containerName,
		&azstorage.BlobContainersClientGetOptions{},
	)

	if err != nil && !bloberror.HasCode(err, bloberror.ContainerNotFound) { // we hit an error other than ContainerNotFound
		return err
	} else if err == nil { // we find that the container already exists
		return nil
	}

	_, err = m.blobContainer.Create(
		ctx,
		resourceGroup,
		accountName,
		containerName,
		azstorage.BlobContainer{
			ContainerProperties: &azstorage.ContainerProperties{
				PublicAccess: to.Ptr(publicAccess),
			},
		},
		&azstorage.BlobContainersClientCreateOptions{},
	)

	return err
}
```

pkg/api/openshiftcluster.go
cadenmarchese: In a recent code review did we decide to make this a pointer after all just for the purpose of being able to make it nil, or am I getting this field confused with `BoundServiceAccountSigningKey` just below?

pkg/api/openshiftcluster.go
rajdeepc2792: Changed it to *SecureString now, please review the [commit](https://github.com/Azure/ARO-RP/pull/3564/commits/6067a9dc670b123a2845e812d0dfa3c23c76acc7)

pkg/api/openshiftcluster.go
rajdeepc2792: `OIDCIssuer` has a similar behavior as `BoundServiceAccountSigningKey` and will always be set when it is a MIWI cluster. Otherwise not set.
@kimorris27 can confirm if it should be a pointer or not, as he is the owner of  [Preview API design doc](https://docs.google.com/document/d/1AIuf6H9y0Jjboywp0NgD8-9lU1jCj5WrE83jh58rEg4/edit#heading=h.cifwpl8od05m). Also the `BoundServiceAccountSigningKey` has been changed to *SecureString.

kimorris27: I don't see why either of these fields (`OIDCIssuer` and `BoundServiceAccountSigningKey`) _must_ be pointers. I feel like the difference is mostly semantic. However, in order to be consistent with how things are done in existing code, I would make both of these fields values rather than pointers if it were up to me.

For example, `ClusterProfile.PullSecret` is optional, but it's a value rather than a pointer.

I'd defer to @bennerv though since he raised the question in the first place.

pkg/util/pointerutils/pointerutils.go
```diff
@@ -0,0 +1,6 @@
+package pointerutils
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+func ToPtr[T any](t T) *T { return &t }
```
kimorris27, line 6: This was a good idea 👍🏻

pkg/api/openshiftcluster.go
cadenmarchese: I felt the same. Approving either way.

cadenmarchese (approved):

pkg/env/prod.go
rajdeepc2792: Created a story for the same:- https://issues.redhat.com/browse/ARO-8405

pkg/api/openshiftcluster.go
rajdeepc2792: After some discussion, it seems that since the `OIDCIssuer` being a non-pointer will be visible to the customer when doing `az aro show`. So it has been decided to convert `OIDCIssuer` & `BoundServiceAccountSigningKey` to pointer.
Commit Ref:-  https://github.com/Azure/ARO-RP/pull/3564/commits/2d223560cd9435dea2202a462bfddb63cfd71b13

cadenmarchese (approved):

pkg/cluster/deploybaseresources.go
```diff
@@ -35,6 +39,47 @@ func (m *manager) createDNS(ctx context.Context) error {
 	return m.dns.Create(ctx, m.doc.OpenShiftCluster)
 }
 
+func (m *manager) createOIDC(ctx context.Context) error {
+	if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile != nil || m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile == nil {
+		return nil
+	}
+
+	blobContainerName := env.OIDCBlobContainerPrefix + m.doc.ID
```
gouthamMN, line 47: Few days back we found out that $web container under storage account will let us access data keeping anonymous blob access disabled when you enable static website. This should comply with the ongoing MSFT security wave and fit with our need. I would suggest using $web container here. Here is the more [details on how this works](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-static-website#impact-of-setting-the-access-level-on-the-web-container) and [how to setup this.](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-static-website-host). You can also look at the [OIDC infra-PR to see how this is done on INT and Prod](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO-Pipelines/pullrequest/10275068)
cc - @weinong

cadenmarchese: I'm adding a hold label here while we address this comment.

pkg/api/openshiftcluster.go
kimorris27: That choice makes sense to me 👍🏻

pkg/cluster/deploybaseresources.go
```diff
@@ -35,6 +38,48 @@ func (m *manager) createDNS(ctx context.Context) error {
...
 }
 
+func (m *manager) createOIDC(ctx context.Context) error {
+	if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile != nil || m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile == nil {
+		return nil
+	}
+
+	// OIDC Storage Web Endpoint need to be determined for Development environments
+	var oidcEndpoint string
+	if m.env.FeatureIsSet(env.FeatureRequireOIDCStorageWebEndpoint) {
```
rajdeepc2792, line 48: This is needed for the Full and Local RP as the generated web endpoint is a DNS Zone present in it which cannot be determined without getting from the Storage Account Properties.
Please provide any other approach that can be taken to provide the Web Endpoint URL.

Thought around fitting it as a secret, it can be done in local dev. But for Full RP, the storage account creation depends on the Resource Group creation that is, RG Creation -> Storage Account Creation -> Fetch Web Endpoint -> Pass the Value to RP as environment variable(Can't think of a good way to do this for this flow.).

gouthamMN (commented):
lgtm

pkg/api/openshiftcluster.go
```diff
@@ -258,12 +258,13 @@ type OIDCIssuer string
...
-	ResourceGroupID      string               `json:"resourceGroupId,omitempty"`
-	FipsValidatedModules FipsValidatedModules `json:"fipsValidatedModules,omitempty"`
-	OIDCIssuer           OIDCIssuer           `json:"oidcIssuer,omitempty"`
+	PullSecret                    SecureString         `json:"pullSecret,omitempty"`
+	Domain                        string               `json:"domain,omitempty"`
+	Version                       string               `json:"version,omitempty"`
+	ResourceGroupID               string               `json:"resourceGroupId,omitempty"`
+	FipsValidatedModules          FipsValidatedModules `json:"fipsValidatedModules,omitempty"`
+	OIDCIssuer                    *OIDCIssuer          `json:"oidcIssuer,omitempty"`
+	BoundServiceAccountSigningKey *SecureString        `json:"boundServiceAccountSigningKey,omitempty"`
```
niontive, line 267: Why do we make the signing key a pointer to `SecureString`, but we make pull secret a `SecureString`?

kimorris27: It's been discussed here: https://github.com/Azure/ARO-RP/pull/3564#discussion_r1638456307

pkg/cluster/deploybaseresources.go
kimorris27: Resolving this comment thread since Rajdeep has committed these changes.

pkg/api/openshiftcluster.go
kimorris27: Resolving this thread since the change has been made and the motivation for the change is clear.

pkg/api/openshiftcluster.go
kimorris27: Resolving this thread since the change has been made and the motivation for the change is clear.

pkg/cluster/deploybaseresources.go
kimorris27: Resolving this thread since the change has been made and the motivation for the change is clear.

pkg/cluster/deploybaseresources.go
kimorris27: I don't see any problem with the current approach. I agree that ideally we would store these types of configuration values in env variables, but since this DNS name is "dynamically" determined, I see why we can't do that here.

pkg/util/azblob/manager.go
```diff
@@ -0,0 +1,71 @@
...
+	azstorage "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+
+	"github.com/Azure/ARO-RP/pkg/util/azureclient"
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/armstorage"
+)
+
+type Manager interface {
+	GetContainerProperties(ctx context.Context, resourceGroupName string, accountName string, containerName string) (azstorage.AccountsClientGetPropertiesResponse, error)
+	GetAZBlobClient(blobContainerURL string, options *azblob.ClientOptions) (AZBlobClient, error)
```
kimorris27, line 19: I think to avoid issues in FairFax (or other non-public clouds as the ARO service expands), the cloud probably needs to be included in the `ClientOptions` wherever this function is called. I see two places:

1. https://github.com/Azure/ARO-RP/blob/66f608e2ed092e7f26556dffb585bf481f7436fe/pkg/cluster/delete.go#L449
2. https://github.com/Azure/ARO-RP/blob/66f608e2ed092e7f26556dffb585bf481f7436fe/pkg/cluster/deploybaseresources.go#L64

One way I see that you could do this is to pass a reusable `ClientOptions` to the `azblob.Manager` on https://github.com/Azure/ARO-RP/blob/66f608e2ed092e7f26556dffb585bf481f7436fe/pkg/cluster/cluster.go#L195 and then use that `ClientOptions` whenever you instantiate a new client with `GetAZBlobClient`. You can use this one: https://github.com/Azure/ARO-RP/blob/66f608e2ed092e7f26556dffb585bf481f7436fe/pkg/cluster/cluster.go#L173.

Let me know what you think about this.

kimorris27 (changes requested):
~~I'm suggesting one change to ensure this code works in FairFax, and I left a few smaller comments/questions.~~

Edit: I think we're all good WRT the FairFax thing, so I just have some small questions and nits.

pkg/util/azblob/manager.go
kimorris27: More info about the cloud configuration: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud

kimorris27: Wait, false alarm. I see now where the cloud is being pulled from the ARO env: https://github.com/Azure/ARO-RP/blob/66f608e2ed092e7f26556dffb585bf481f7436fe/pkg/util/azureclient/azuresdk/armstorage/accounts.go#L31.

I think we're good here.

pkg/util/oidcbuilder/oidcbuilder.go
```diff
@@ -0,0 +1,105 @@
...
+	return b.blobContainerURL
+}
+
+func populateOidcFolder(ctx context.Context, directory string, discoveryDocument string, jwks []byte, azBlobClient utilazblob.AZBlobClient) error {
+	err := azBlobClient.UploadBuffer(
+		ctx,
+		"",
+		DocumentKey(directory, DiscoveryDocumentKey),
+		[]byte(discoveryDocument),
+	)
```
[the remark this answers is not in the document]
rajdeepc2792: Yes, it's somewhat tricky.
During the initialization of azBlobClient, the URL was passed which was already at the container level. If we pass anything instead of "", it is being added at top of the URL passed during initialization.
Eg:- https://github.com/openshift/cloud-credential-operator/blob/release-4.14/pkg/cmd/provisioning/azure/create_oidc_issuer.go#L375C1-L383C3

pkg/util/oidcbuilder/jwks.go
```diff
@@ -0,0 +1,136 @@
...
+			"sub",
+			"iat",
+			"iss",
+			"sub"
+		]
+	}`
+)
+
+func GenerateDiscoveryDocument(bucketURL string) string {
+	return fmt.Sprintf(discoveryDocumentTemplate, bucketURL, bucketURL)
```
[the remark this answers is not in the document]
rajdeepc2792: This is mostly copied from the https://github.com/openshift/cloud-credential-operator/blob/release-4.14/pkg/cmd/provisioning/azure/create_oidc_issuer.go#L319. ROSA STS also uses the same approach to create a discoveryDocument.

pkg/deploy/generator/resources_oic.go
```diff
@@ -17,21 +17,23 @@ import (
 var (
 	// Storage accounts must not contain dashes or be more than 24 characters
 	// Append "oic" to the pre-existing storage account prefix.
-	storageAccountName         string = "concat(take(substring(parameters('storageAccountDomain'), 0, indexOf(parameters('storageAccountDomain'), '.')), 21), 'oic')"
+	storageAccountName         string = "parameters('oidcStorageAccountName')"
```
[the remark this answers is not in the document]
rajdeepc2792: Removed comments!

pkg/util/azblob/manager.go
```diff
@@ -0,0 +1,71 @@
...
+	return m.account.GetProperties(ctx, resourceGroupName, accountName, &azstorage.AccountsClientGetPropertiesOptions{})
+}
+
+func (m *manager) GetAZBlobClient(blobContainerURL string, options *azblob.ClientOptions) (AZBlobClient, error) {
+	return NewAZBlobClient(blobContainerURL, m.cred, options)
+}
+
+type AZBlobClient interface {
+	UploadBuffer(ctx context.Context, containerName string, blobName string, buffer []byte) error
+	DeleteBlob(ctx context.Context, containerName string, directoryName string) error
```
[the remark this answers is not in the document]
rajdeepc2792: Used `blobName` everywhere now!

kimorris27:
LGTM, but my recommendation is to wait until we have E2E available before merging because of the size and scope of the PR.

kimorris27 (approved):

cadenmarchese (approved):

PR #3565, opened 2024-05-08
Add hack script to install multiple clusters at one time.

bitoku:

Fixes N/A
As a part of https://issues.redhat.com/browse/ARO-6198

Once we have the install in prod regions, we need to install some clusters to make sure that there is no increase in installer failure rate.
This hack script can help to test multiple installations at one time in a region. 

### Test plan for issue:

Ran in local.

yes I updated the doc.

You can check the output of the script to see the result.

### DIFF
--- a/docs/upstream-differences.md
+++ b/docs/upstream-differences.md
@@ -202,3 +202,20 @@ In the ARO-RP codebase:
 Publish RHCOS image to the Azure Cloud Partner Portal:
 1. Publish RHCOS image. See [this document](https://github.com/openshift/installer-aro-wrapper/blob/main/docs/publish-rhcos-image.md).
 1. After this point, you should be able to create a dev cluster using the RP and it should use the new release.
+
+# Install Test
+
+Once the installer is deployed in each region, we need to test installation to detect if there is any increase in installation failure.
+
+You can install multiple clusters at one time in a region by running this script (**`tmux` is required**):
+
+```bash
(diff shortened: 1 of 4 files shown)


mociarain (commented):
LGTM but I wonder why this stuff isn't in the installer repo?

bitoku:
Good question. To be honest, I didn't think of that.

If I had to say, the installer repo has scripts that are needed when developing the installer, but the process like [ARO-6198](https://issues.redhat.com/browse/ARO-6198) is conducted after we deployed a new release in each region (I mean not when we are developing it).

But I don't have the preference. I want to know what others think.

tiguelu:
> @mociarain Good question. To be honest, I didn't think of that.
> 
> If I had to say, the installer repo has scripts that are needed when developing the installer, but the process like [ARO-6198](https://issues.redhat.com/browse/ARO-6198) is conducted after we deployed a new release in each region. Also ARO-RP is more accessible because everyone clones it.
> 
> But I don't have the preference. I want to know what others think.

I didn't think of it either, but it makes more sense to have the script in the wrapper repo. Probably even the .md file with the instructions too. If we move the content from upstream-difference.md too, we probably should keep the old file just with a link to the new one (since it is linked from other docs, jiras and wikis).

bitoku:
Thank you all!
I will close this PR and will make new ones to here and the installer repo to migrate the `upstream-difference.md`.

PR #3567, opened 2024-05-08
Add nil check on ServicePrincipalProfile before populating clientsecrets for different frontend routes

rajdeepc2792:
Additional changes for https://issues.redhat.com/browse/ARO-6623
Previous PR:- https://github.com/Azure/ARO-RP/pull/3529

Adds a condition on ServicePrincipalProfile before setting the client secret. This required for the MIWI clusters which will have the ServicePrincipalProfile as nil, and setting ClientSecret on the nil pointer would cause an error.

### Test plan for issue:
e2e and CI should pass

No

Additional checks on already tested and working flow.


kimorris27 (approved):

shubhadapaithankar (approved):

bitoku (commented):
we want to unit tests for the change maybe?

cadenmarchese (approved):
thanks!

PR #3568, opened 2024-05-09
Add track2 API Client to Cluster struct (loadbalancer and interfaces)

bitoku:

Part of https://issues.redhat.com/browse/ARO-4665

This PR adds new track2 clients for later modifications.
The reason for creating new methods (`armLoadbalancers` and `armInterfaces`) instead of modifying the current ones (`Loadbalancers` and `Interfaces`)  is to make the impact of each change as small as possible.

After merging this PR, I will work on the actual API changes.

https://github.com/Azure/azure-sdk-for-go/blob/main/documentation/MIGRATION_GUIDE.md

### Test plan for issue:

There is no change that needs new tests.
CI is enough.

cleanup N/A

This PR doesn't make behaviour change.


bitoku:
The API change would be like this for example.

https://github.com/Azure/ARO-RP/blob/ed7799db13d8df1435b8cb4284e5213be135bd9e/pkg/cluster/ipaddresses.go#L170-L200

This PR contains just the basic change.
When we want to add a retry policy, it would be like this.

https://github.com/Azure/ARO-RP/blob/ed7799db13d8df1435b8cb4284e5213be135bd9e/pkg/util/azureclient/azuresdk/armnetwork/client_factory.go#L16-L54

pkg/cluster/cluster.go
```diff
@@ -66,12 +67,16 @@ type manager struct {
...
-	interfaces            network.InterfacesClient
-	publicIPAddresses     network.PublicIPAddressesClient
+	spGraphClient   *utilgraph.GraphServiceClient
+	disks           compute.DisksClient
+	virtualMachines compute.VirtualMachinesClient
+	// TODO: use armInterfaces instead.
+	interfaces        network.InterfacesClient
+	armInterfaces     armnetwork.InterfacesClient
+	publicIPAddresses network.PublicIPAddressesClient
+	// TODO: use armLoadBalancers instead.
```
mociarain, line 77: Can you reset the formatting here?

pkg/util/azureclient/azuresdk/armnetwork/interfaces.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+type interfacesClient struct {
+	*armnetwork.InterfacesClient
+}
+
+var _ InterfacesClient = &interfacesClient{}
+
+// NewInterfacesClient creates a new InterfacesClient
+func NewInterfacesClient(environment *azureclient.AROEnvironment, subscriptionID string, credential azcore.TokenCredential) (InterfacesClient, error) {
```
mociarain, line 29: We only pass the FPCredential above. Does this expect the FPCredential always?

mociarain (approved):
2 small things but LGTM

pkg/util/azureclient/azuresdk/armnetwork/interfaces.go
bitoku: The question is whether `credential` can be different from `fpCredential`?

It is always `fpCredential` for `LoadBalancersClient` and `InterfacesClient` as the previous implementation always uses fpAuthorizer.
This `credential` is a replacement of `authorizer` for the track1.
https://github.com/Azure/azure-sdk-for-go/blob/main/documentation/MIGRATION_GUIDE.md#authentication

When we replace the SDK of `fpPrivateEndpoints`, `rpPrivateLinkServices`, or `dns`, the credential should be different from `fpCredential` because it uses the different authorizer (see `pkg/cluster/cluster.go`).

mociarain: Cool. Then if I understood you correctly we should rename the variable to fpCredential. 

As a separate future PR I think we could extend  azidentity.ClientCertificateCredential to ServicePrincipalCredential and FirstPartyCredential to give us some more assurance and type-safety. What do you think?

bitoku: Mostly I agree with the idea.
My concern is that MIWI feature might use the different credential type.
In that case, we should keep it as it is, but I'm not sure about it though.

bitoku: So my initial answer was incorrect. sorry about that.
> It is always fpCredential for LoadBalancersClient and InterfacesClient as the previous implementation always uses fpAuthorizer.

This is correct just for now.
This can be `WorkloadIdentityCredential`, maybe, in the future.
not sure what the implementation would be like though

https://docs.google.com/document/d/1AIuf6H9y0Jjboywp0NgD8-9lU1jCj5WrE83jh58rEg4/edit
https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.11.1/internal/exported#TokenCredential
https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#WorkloadIdentityCredential

mociarain (approved):

tiguelu (approved):

PR #3573, opened 2024-05-10
Update document about custom installer

bitoku:

None 

Related to #3486 
This PR adds document about how to create a cluster with custom installer.

### Test plan for issue:

N/A document update

updated

N/A

### DIFF
--- a/docs/deploy-development-rp.md
+++ b/docs/deploy-development-rp.md
@@ -153,6 +153,26 @@
       alias rp-logs-all='journalctl _COMM=aro -o json -e | jq -r '\''select (.COMPONENT != null and (.COMPONENT | contains("access"))|not) | .MESSAGE'\'''
       EOF
       ```
+     
+### Use a custom installer
+
+Sometimes you want to use a custom installer, for example, when you want to test a new OCP version's installer.
+You can create a cluster with the new installer following these steps:
+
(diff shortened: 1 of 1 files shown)


docs/deploy-development-rp.md
```diff
@@ -153,6 +153,32 @@
...
+
+Sometimes you want to use a custom installer, for example, when you want to test a new OCP version's installer.
+You can create a cluster with the new installer following these steps:
+
+1. Push the installer image to somewhere accessible from Hive AKS.
+    
+    [quay.io](https://quay.io/) would be one of the options.
+    You need pull-secret to use the repositories other than `arointsvc.azurecr.io`.
+    It must be configured in the secrets.
+    If you are using the hack script, you don't have to care about it because the script uses `PULL_SECRET` automatically.
```
tiguelu, line 167: I believe the quay.io pull-secret is in USER_PULL_SECRET:

```suggestion
    If you are using the hack script, you don't have to care about it because the script uses `USER_PULL_SECRET` automatically.
```

tiguelu (changes requested):
Thank you for adding this information. Just added a comment about a small change.

docs/deploy-development-rp.md
bitoku: thanks! yes it is

tiguelu (approved):

PR #3575, opened 2024-05-10
Allow customers to annotate CloudCredential resource on update

cadenmarchese:

Fixes ARO-7348

Clusters with manual credentials mode need to be annotated in order to be upgraded: https://docs.openshift.com/container-platform/4.14/updating/preparing_for_updates/preparing-manual-creds-update.html#cco-manual-upgrade-annotation_preparing-manual-creds-update

This PR adds a new API field that the user can patch on update, which will then annotate the cluster itself with the version.

Requires https://github.com/Azure/ARO-RP/pull/3563 as this will also require some validation to ensure that the user is passing in a version and not some other string.


pkg/api/v20240812preview/openshiftcluster.go
```diff
@@ -290,9 +290,13 @@ type IngressProfile struct {
 
 // PlatformWorkloadIdentityProfile encapsulates all information that is specific to workload identity clusters.
 type PlatformWorkloadIdentityProfile struct {
+	UpgradeableTo              *UpgradeableTo             `json:"upgradeableTo,omitempty"`
```
rajdeepc2792, line 293: Same as the conversation here:- https://redhat-external.slack.com/archives/C06FRQKUMFF/p1716217461235529
Do we need to add a struct tag `mutable` for this field too?

It has to be only for the external APIs, I suppose.

pkg/operator/deploy/deploy.go
```diff
@@ -432,6 +436,34 @@ func (o *operator) RenewMDSDCertificate(ctx context.Context) error {
 	return nil
 }
 
+func (o *operator) EnsureUpgradeAnnotations(ctx context.Context) error {
+	if o.oc.Properties.PlatformWorkloadIdentityProfile == nil ||
+		o.oc.Properties.ServicePrincipalProfile != nil {
```
rajdeepc2792, line 441: Why both conditions are required here? Why can't we rely on either of the 2 conditions?

rajdeepc2792 (commented):
Left 2 comments:-
- I believe a `Mutable` tag would be needed for the ExternalAPI.
Everything else seems good.

pkg/operator/deploy/deploy.go
cadenmarchese: I was thinking belt and braces, but which one do you think makes more sense to rely on?

cadenmarchese: The more I think about this, the more I think we should have a utility function that we can call everywhere in the codebase that runs this check so we don't have to duplicate the logic everywhere. I think we could probably handle this here or in a separate PR.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -442,6 +443,14 @@ func (sv openShiftClusterStaticValidator) validatePlatformWorkloadIdentityProfil
 			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, fmt.Sprintf("%s.PlatformWorkloadIdentities[%d].resourceID", path, n), "Resource must be a user assigned identity.")
 		}
 	}
+
+	if pwip.UpgradeableTo != nil {
+		matches, err := regexp.MatchString(`^4\.[0-9]{2}\.`, string(*pwip.UpgradeableTo))
```
kimorris27, line 448: You can use the CoreOS `semver` package here instead of rolling your own regex: https://pkg.go.dev/github.com/coreos/go-semver/semver#NewVersion. Just call this function and check for an error.

We already have this package vendored in, so no new imports needed.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -442,6 +443,14 @@ func (sv openShiftClusterStaticValidator) validatePlatformWorkloadIdentityProfil
...
+		if !matches || err != nil {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, fmt.Sprintf("%s.UpgradeableTo[%v]", path, *pwip.UpgradeableTo), "UpgradeableTo must be a valid OpenShift version.")
```
kimorris27, line 450:
```suggestion
			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, fmt.Sprintf("%s.UpgradeableTo[%v]", path, *pwip.UpgradeableTo), "UpgradeableTo must be a valid OpenShift version in the format 'x.y.z'.")
```

Just to make it clear to customers that they have to include the z-stream in the version number. I checked https://docs.openshift.com/container-platform/4.14/updating/preparing_for_updates/preparing-manual-creds-update.html#cco-manual-upgrade-annotation_preparing-manual-creds-update to verify that the annotation should include the z-stream.

pkg/cluster/arooperator.go
```diff
@@ -65,6 +65,10 @@ func (m *manager) ensureCredentialsRequest(ctx context.Context) error {
 	return m.aroOperatorDeployer.CreateOrUpdateCredentialsRequest(ctx)
 }
 
+func (m *manager) ensureUpgradeAnnotations(ctx context.Context) error {
```
kimorris27, line 68:
```suggestion
func (m *manager) ensureUpgradeAnnotation(ctx context.Context) error {
```

Suggesting this change for clarity's sake since there's only one annotation.

pkg/operator/deploy/deploy.go
```diff
@@ -56,6 +57,7 @@ type Operator interface {
 	Restart(context.Context, []string) error
 	IsRunningDesiredVersion(context.Context) (bool, error)
 	RenewMDSDCertificate(context.Context) error
+	EnsureUpgradeAnnotations(context.Context) error
```
kimorris27, line 60:
```suggestion
	EnsureUpgradeAnnotation(context.Context) error
```

Suggesting this change for clarity's sake since there's only one annotation.

kimorris27 (changes requested):
The logic LGTM, but I suggested some small changes.

pkg/operator/deploy/deploy.go
rajdeepc2792: Yes, I agree, and better to have a standardized way(util function) to check this part.

cadenmarchese: I'll card up the utility function and have Sanjana or Alex handle :)

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
cadenmarchese: Nice, thanks for the tip.

.sha256sum
```diff
@@ -1,9 +1 @@
...
-01ba9562a8dac2824998ff0ad0d2465f79e6a66597bdb321e9409b9f2d12d222  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/stable/2023-04-01/redhatopenshift.json
-c023515341196746454c0ae7af077d40d3ec13f6b88b33cb558f0a7ab17a5a24  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/preview/2023-07-01-preview/redhatopenshift.json
-440748951dd1c3b34b5ccbdcb7cd966e3b89490887a1f1d64429561fad789515  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/stable/2023-09-04/redhatopenshift.json
-74a46fdde6ceb0121fe1515c7e11e902dd921b54cffe693307fb02b3dc88f26e  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/stable/2023-11-22/redhatopenshift.json
-a27184734436629e24b344c3b5c015437f144e18e7eddce7e252a1ed4cda7bca  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/preview/2024-08-12-preview/redhatopenshift.json
+f1e9d42d7c4c0081282e065e7845455db28ed6924687f1acecafb5fbc43ae0c3  swagger/redhatopenshift/resource-manager/Microsoft.RedHatOpenShift/preview/2024-08-12-preview/redhatopenshift.json
```
kimorris27, line 1: Need to add the old SHA sums back in :wink:

kimorris27 (changes requested):

pkg/operator/deploy/deploy.go
rajdeepc2792: I would also go with `||`, we don't want to perform any MIWI operations until all the conditions we think are required to be termed it as a MIWI cluster are `True`.

rajdeepc2792 (approved):

kimorris27 (approved):

hlipsig (approved):

PR #3581, opened 2024-05-15
Add public blob access exception tag to dev OIDC storage account

cadenmarchese:

Creating storage accounts with public blob access in MSFT tenants without this tag will cause the storage account to be flagged. We want to keep the storage account blob access public because it will save effort required to onboard Azure Front Door in development and e2e environments, and save cost of having long-running Azure Front Door instances.

The longer term goal is to migrate to a different subscription for CI e2e where this rule won't apply.

### Test plan for issue:

Nic has created a storage account with this tag in the MSFT tenant already, we are monitoring it to see if it will prevent the account from being flagged.

### DIFF
--- a/pkg/deploy/assets/rp-oic.json
+++ b/pkg/deploy/assets/rp-oic.json
@@ -21,6 +21,9 @@
                 "allowBlobPublicAccess": true,
                 "minimumTlsVersion": "TLS1_2"
             },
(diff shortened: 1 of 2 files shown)


cadenmarchese:
Closing in favor of https://github.com/Azure/ARO-RP/pull/3585.

PR #3583, opened 2024-05-15
Use single ARO API/client version in pkg/util/cluster

tsatam:

Fixes no Jira

Updates pkg/util/cluster/cluster.go to use a single ARO API version and client, replacing the 4 different API versions previously in use. 

This version is currently set to the v20230904 API - the latest public-facing API as of now. This allows the functionality in pkg/util/cluster to continue to be used against production RP instances, e.g. in production E2E contexts. 

### Test plan for issue:

- [x] Local cluster creation kicked off successfully
- [x] Local cluster deletion kicked off successfully
- [x] PR E2E cluster creation/deletion works as expected

No. 

N/A (this functionality is only used during local development and when running E2E)

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -31,8 +31,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/Azure/ARO-RP/pkg/api"
-	v20230904 "github.com/Azure/ARO-RP/pkg/api/v20230904"
-	mgmtredhatopenshift20230904 "github.com/Azure/ARO-RP/pkg/client/services/redhatopenshift/mgmt/2023-09-04/redhatopenshift"
+	"github.com/Azure/ARO-RP/pkg/api/v20240812preview"
+	mgmtredhatopenshift20240812preview "github.com/Azure/ARO-RP/pkg/client/services/redhatopenshift/mgmt/2024-08-12-preview/redhatopenshift"
 	"github.com/Azure/ARO-RP/pkg/deploy/assets"
 	"github.com/Azure/ARO-RP/pkg/deploy/generator"
 	"github.com/Azure/ARO-RP/pkg/env"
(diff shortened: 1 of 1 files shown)


pkg/util/cluster/cluster.go
```diff
@@ -169,9 +160,9 @@ func (c *Cluster) DeleteApp(ctx context.Context) error {
 }
 
 func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName string, osClusterVersion string) error {
-	clusterGet, err := c.openshiftclustersv20230904.Get(ctx, vnetResourceGroup, clusterName)
+	clusterGet, err := c.openshiftclusters.Get(ctx, vnetResourceGroup, clusterName)
```
cadenmarchese, line 163: I'm sure you've thought of this, but for the time being, does it make sense to just wrap this into an `if isDevelopment` or similar block, and have prod e2es use the old client? It would make for cleanup later, but not sure how else we'd merge this before the v20240812preview API is available everywhere.

This has the downside of e2e testing the new API in dev but the old API in prod, which might be useful for smoke testing the new API, but would create a slight difference between the e2e runs depending on environment.

tsatam: Team Loki talked about this in our sync PR review session, this proposed solution is somewhere on the spectrum between the current implementation in this PR, and a long-term idea we have to make the API version used here configurable (e.g. allow us to run E2E against multiple API versions), something that may become more desirable for us as we progress further along the CI/CD initiative. 

I would like to implement this proposed solution in the near-term, but I think it would require some thought into re-usability for that potential later step, and thus I think it makes the most sense in a follow-up PR. 

For now, I'm thinking the steps are:
- This PR: downgrade the single API version used to our latest routeable API (20230904)
- Follow-up PR: implement some sort of determination between the latest routeable and in-progress APIs, e.g. through an abstracted client implementation that accepts requests and provides responses with the internal clusterdoc representation, and converts to the specific external representation before actually amking the API call (which is how pkg/util/cluster operates today anyways)
- Future effort, depending on need: allow pkg/util/cluster to be instantiated with a specific API version, implement a configurable parameter in both PR and prod E2E to specify the version to use (default being the behavior above)

SudoBrendan (approved):

cadenmarchese (approved):

PR #3584, opened 2024-05-16
Removed DBToken. Made the components uses Managed Identity

anshulvermapatel:
As part of the [ARO-5512](https://issues.redhat.com/browse/ARO-5512) , we needed to disable local auth for cosmosdb. In order to disable local authentication, this PR is to make our components rely on managed identities/entra id rather than keys for authentication/authorizing cosmosdb.

This is currently a draft PR as I will be testing it in the INT env first before going forward, however, reviews are appreciated.

~~I have manually changed some generated code here. That was because the [PR](https://github.com/jewzaam/go-cosmosdb/pull/22) I created to have the generated code changed is not merged yet. Once that is merged, I will generate the code from the go-cosmosdb repo itself just like how we usually do.~~

Fixes
[ARO-7195](https://issues.redhat.com/browse/ARO-7195)
[ARO-7314](https://issues.redhat.com/browse/ARO-7314)

### Test plan for issue:

There are multiple ways this can be tested.
1. Checkout to this PR and try running Local RP and create a cluster using it.
2. If you have a full dev RP service running. Try deploying this code and try creating a cluster post that.
3. Try deploying this PR in INT env.


pkg/database/billing.go
```diff
@@ -29,48 +36,61 @@ type Billing interface {
...
+		}
+		ctx, cancel := context.WithTimeout(ctx, time.Minute*5)
+		defer cancel()
+
+		poller, err := sqlResourceClient.BeginCreateUpdateSQLTrigger(ctx, resourceGroup, dbAccountName, dbName, billingContainerName, *triggerResource.ID, createUpdateSQLTriggerParameters, nil)
+		if err != nil {
+			return nil, err
+		}
+		_, err = poller.PollUntilDone(ctx, nil)
+		if err != nil {
```
bennerv, line 93: the error logic handling is changing here.  Should it be?  Should we only ignore errors of `http.StatusConflict` from cosmosdb?

pkg/database/billing.go
```diff
@@ -29,48 +36,61 @@ type Billing interface {
...
+	for _, triggerResource := range triggerResources {
+		createUpdateSQLTriggerParameters := armcosmos.SQLTriggerCreateUpdateParameters{
+			Properties: &armcosmos.SQLTriggerCreateUpdateProperties{
+				Options:  &armcosmos.CreateUpdateOptions{},
+				Resource: triggerResource,
+			},
+			Location: &location,
+		}
+		ctx, cancel := context.WithTimeout(ctx, time.Minute*5)
+		defer cancel()
```
bennerv, line 86: Do we have an over-arching context assocaited with this function already?  is the old method `triggerc.Create(ctx, trigger)` blocking vs now we have to poll?

pkg/database/database.go
```diff
@@ -71,6 +73,22 @@ func NewMasterKeyAuthorizer(ctx context.Context, log *logrus.Entry, token azcore
 	return cosmosdb.NewMasterKeyAuthorizer(getDatabaseKey(keys, log))
 }
 
+func NewTokenAuthorizer(ctx context.Context, log *logrus.Entry, cred azcore.TokenCredential, databaseAccountName string, scopes []string) (cosmosdb.Authorizer, error) {
+	scopes = append(scopes, fmt.Sprintf("https://%s.documents.azure.com/.default", databaseAccountName))
```
bennerv, line 77: This scope is incorrect for government cloud (and any non-Azure Public Cloud)

We should read the url from the env struct we use everywhere to be cloud aware.

https://learn.microsoft.com/en-us/azure/azure-government/compare-azure-government-global-azure

pkg/database/openshiftclusters.go
```diff
@@ -23,6 +26,8 @@ const (
 	OpenshiftClustersPrefixQuery        = `SELECT * FROM OpenShiftClusters doc WHERE STARTSWITH(doc.key, @prefix)`
 	OpenshiftClustersClientIdQuery      = `SELECT * FROM OpenShiftClusters doc WHERE doc.clientIdKey = @clientID`
 	OpenshiftClustersResourceGroupQuery = `SELECT * FROM OpenShiftClusters doc WHERE doc.clusterResourceGroupIdKey = @resourceGroupID`
+
+	openShiftClusterContainerName = "OpenShiftClusters"
```
bennerv, line 30: if we're pulling this out, we should leverage it in our above queries.

pkg/database/subscriptions.go
```diff
@@ -8,13 +8,20 @@ import (
...
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/database/cosmosdb"
 	"github.com/Azure/ARO-RP/pkg/util/uuid"
 )
 
-const SubscriptionsDequeueQuery string = `SELECT * FROM Subscriptions doc WHERE (doc.deleting ?? false) AND (doc.leaseExpires ?? 0) < GetCurrentTimestamp() / 1000`
+const (
+	SubscriptionsDequeueQuery string = `SELECT * FROM Subscriptions doc WHERE (doc.deleting ?? false) AND (doc.leaseExpires ?? 0) < GetCurrentTimestamp() / 1000`
+	subscriptionContainerName string = "Subscriptions"
```
bennerv, line 23: same as above `OpenShiftClusters`.  if we're pulling it out, leverage it within our queries within this.

pkg/deploy/devconfig.go
```diff
@@ -157,9 +157,9 @@ func DevConfig(_env env.Core) (*Config, error) {
 				"qos.ppe.warm.ingest.monitor.core.windows.net",
 				"test1.diagnostics.monitoring.core.windows.net",
 			},
-			GatewayFeatures: []string{
-				"InsecureSkipVerifyDBTokenCertificate",
-			},
+			// GatewayFeatures: []string{
```
bennerv, line 160: Let's just rip this out completely if it's not going to be used.  No need to comment it out.

pkg/database/database.go
```diff
@@ -71,6 +73,22 @@ func NewMasterKeyAuthorizer(ctx context.Context, log *logrus.Entry, token azcore
 	return cosmosdb.NewMasterKeyAuthorizer(getDatabaseKey(keys, log))
 }
 
+func NewTokenAuthorizer(ctx context.Context, log *logrus.Entry, cred azcore.TokenCredential, databaseAccountName string, scopes []string) (cosmosdb.Authorizer, error) {
```
bennerv, line 76: I think there's a problem inherently with this function.  

The token we request from AAD with the scope is a short-lived credential (~12 hrs max I believe).  That means that at some point, the token will expire both from AAD MSI perspective and cosmosdb perspective.  As a result, we may need some refreshing of the actual database client in all places where it's used. 

Previously, we used the primary database keys which never expire but are rotatable.  That's why we didn't need to implement any database client reconstruction due to token expiration.  Now we more than likely do. 

The best way to test this would be to run an RP for a long time (or get the expiration of the token from this function) and see what happens when you get past the expiration time.

bennerv (changes requested):
Also need to remove: 
- `rm docs/dbtoken-service.md` 
- The load balancing rule to allow gateway to talk to the dbtoken service from `pkg/deploy` (should be in templates)

We probably don't need to peer the gateway and RP vnets anymore.  We might be able to remove that peering as well.  Which would be good for overall security posture.

pkg/database/billing.go
anshulvermapatel: I ran this multiple times on the same cosmosdb database and it didn't return `http.StatusConflict` even when the triggers were already created. However, it will be good that I just add the logic to ignore `http.StatusConflict` err.

pkg/database/database.go
anshulvermapatel: The azcore.TokenCredential, it is indeed an object for tokens but it does not statically contains the token. It has a `getToken()` method which gives a refreshed token every time we call the method.
There is a token refresh logic in place in this go-cosmosdb code in this [PR](https://github.com/jewzaam/go-cosmosdb/pull/22) which cache the token. Every time we make a request, it is check is the current token is expired or 5 mins left to expire, if so it calls `getToken()` to get an updated token else use the cached ones.

pkg/database/openshiftclusters.go
anshulvermapatel: yeah point, I will do that.

pkg/deploy/devconfig.go
anshulvermapatel: Sure

pkg/database/database.go
anshulvermapatel: Gotcha! .. sorry for that, I'll correct it.

docs/deploy-full-rp-service-in-dev.md
jaitaiwan, line 1: Are there any changes in this file that aren't formatting? A quick review seemed like there wasn't any meaningful changes to this file which make me think it should be removed from the PR.

pkg/database/database.go
jaitaiwan: Is it possible on quieter regions that we might not have a request for over 12 hours? Would that have any effect?

pkg/deploy/deploy.go
```diff
@@ -43,13 +43,13 @@ type deployer struct {
...
-	deployments            features.DeploymentsClient
-	groups                 features.ResourceGroupsClient
-	loadbalancers          network.LoadBalancersClient
+	globaldeployments features.DeploymentsClient
+	globalgroups      features.ResourceGroupsClient
+	globalrecordsets  dns.RecordSetsClient
+	globalaccounts    storage.AccountsClient
+	deployments       features.DeploymentsClient
+	groups            features.ResourceGroupsClient
+	// loadbalancers          network.LoadBalancersClient
```
jaitaiwan, line 52: We should just remove this entirely I think

pkg/deploy/deploy.go
```diff
@@ -95,13 +94,13 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Core, config *RPConfig
...
-		deployments:            features.NewDeploymentsClient(_env.Environment(), config.SubscriptionID, authorizer),
-		groups:                 features.NewResourceGroupsClient(_env.Environment(), config.SubscriptionID, authorizer),
-		loadbalancers:          network.NewLoadBalancersClient(_env.Environment(), config.SubscriptionID, authorizer),
+		globaldeployments: features.NewDeploymentsClient(_env.Environment(), *config.Configuration.GlobalSubscriptionID, authorizer),
+		globalgroups:      features.NewResourceGroupsClient(_env.Environment(), *config.Configuration.GlobalSubscriptionID, authorizer),
+		globalrecordsets:  dns.NewRecordSetsClient(_env.Environment(), *config.Configuration.GlobalSubscriptionID, authorizer),
+		globalaccounts:    storage.NewAccountsClient(_env.Environment(), *config.Configuration.GlobalSubscriptionID, authorizer),
+		deployments:       features.NewDeploymentsClient(_env.Environment(), config.SubscriptionID, authorizer),
+		groups:            features.NewResourceGroupsClient(_env.Environment(), config.SubscriptionID, authorizer),
+		// loadbalancers:          network.NewLoadBalancersClient(_env.Environment(), config.SubscriptionID, authorizer),
```
jaitaiwan, line 103: We should just remove this entirely I think

jaitaiwan (changes requested):
A few small comments. Nothing critical. Seems like this PR really needs folks to manually test - is there instructions for how one might do this?

pkg/database/database.go
anshulvermapatel: No, because everything a request goes to cosmosdb, there's a method `Authorize()` which gets called, and this method has the logic to acquire the token if the token is expired. The same logic which we were discussing lately.

pkg/deploy/deploy.go
anshulvermapatel: remove what? line 52 `// loadbalancers          network.LoadBalancersClient` ?
I was not sure about if this is affecting something or not, so I just commented it out for now. I will remove that as well once I have more peer reviews on this.

pkg/deploy/deploy.go
anshulvermapatel: same thing

docs/deploy-full-rp-service-in-dev.md
anshulvermapatel: Ahh, I think prettier plugin changed this file. I will remove this changes.
I did remove this portion though:

> az keyvault certificate import \
    --vault-name "$KEYVAULT_PREFIX-dbt" \
    --name dbtoken-server \
    --file secrets/localhost.pem >/dev/null

niontive:
Regarding impact for billing service:

1) Billing uses master key authorization and doesn't go through DB token service. Consequently, we can remove db token service and not affect billing.

2) Billing is using an older ARO-RP module release, so we can update the latest one and not affect billing.

3) For rollout, don't disable local auth for CosmosDB. We'll migrate billing service to not use master key authorization, rollout billing, then we should be good to disable local auth for Cosmos DB.

cmd/aro/gateway.go
```diff
@@ -44,38 +38,25 @@ func gateway(ctx context.Context, log *logrus.Entry) error {
...
-	for _, feature := range strings.Split(os.Getenv("GATEWAY_FEATURES"), ",") {
-		if feature == "InsecureSkipVerifyDBTokenCertificate" {
-			insecureSkipVerify = true
-			break
-		}
-	}
-
-	url, err := getURL(_env.IsLocalDevelopmentMode())
+	dbAccountName := os.Getenv(envDatabaseAccountName)
+	scope := []string{fmt.Sprintf("https://%s.%s", dbAccountName, _env.Environment().CosmosDBDNSSuffixScope)}
```
rajdeepc2792, line 54: There's no `CosmosDBDNSSuffixScope` variable for any environment. Is `CosmosDBDNSSuffix` needed here?This can be probably the reason of VMSS bootstrap getting timed out.

pkg/database/database.go
```diff
@@ -97,3 +114,28 @@ func NewJSONHandle(aead encryption.AEAD) (*codec.JsonHandle, error) {
 
 	return h, nil
 }
+
+func createTriggers(ctx context.Context, sqlResourceClient *sdkcosmos.SQLResourcesClient, triggerResources []*sdkcosmos.SQLTriggerResource, resourceGroup, dbName, dbAccountName, location, containerName string) error {
```
nwnt, line 118: I believe this function and all its callers are the culprit of the failures in e2e and INT deployment.  This is called by NewXXX (monitors, database, subscriptions and billing), so each time those clients are created, the code sends PUT and GET of `Microsoft.DocumentDB/databaseAccounts/sqlDatabases/containers/triggers`.  In fact, your first e2e from yesterday had 28 (14 PUTs, 14 GETs), hammering ARM.

Instead, what you should do is to have these triggers created by the arm template like those generated by `pkg/deploy`.

anshulvermapatel: So e2e passed in INT, however, I do agree, it is indeed hammering ARM, and I am aware of it. I think it is the polling stuff which is doing this.
Thanks for the suggestions, Let me give it more thought to come up with a solution for this.

anshulvermapatel: There are 5 trigger to be created so 5 PUT request is expected, I don't understand how come 7 PUT requests are sent.

anshulvermapatel: Okay, If we create these triggers through arm templates. Still these will be created when an RP deployment would be done. However, we directly run the RP when running locally and while running CI e2e tests.

nwnt: Sent you what ARM receives.  I don't really understand your last part, but like an infrastructure change, you don't test that kind of change on a local rp either.

anshulvermapatel: Okay, so implemented the ARM templates based solution. The code is pushed. Have a look.

cmd/aro/monitor.go
```diff
@@ -88,11 +88,12 @@ func monitor(ctx context.Context, log *logrus.Entry) error {
 	}
 
 	dbAccountName := os.Getenv(envDatabaseAccountName)
-	clientOptions := &policy.ClientOptions{
-		ClientOptions: _env.Environment().ManagedIdentityCredentialOptions().ClientOptions,
-	}
+	// clientOptions := &policy.ClientOptions{
```
cadenmarchese, line 91: Can we remove line 91-93 if no longer used?

cmd/aro/dbtoken.go
```diff
@@ -1,130 +0,0 @@
...
-	}
-
-	dbName, err := DBName(_env.IsLocalDevelopmentMode())
-	if err != nil {
-		return err
-	}
-
-	userc := cosmosdb.NewUserClient(dbc, dbName)
-
-	err = pkgdbtoken.ConfigurePermissions(ctx, dbName, userc)
```
cadenmarchese, line 86: The original dbtoken only gains permissions over the cluster database, not the entire database account, if i'm understanding the original purpose of dbtoken originally. We are now using an MSI credential which suggests that we won't be scoping our permissions down to the cluster database anymore, but instead granting permissions over the whole database account. Is this a correct understanding, or am I misunderstanding the scope?

cmd/aro/gateway.go
```diff
@@ -44,38 +38,25 @@ func gateway(ctx context.Context, log *logrus.Entry) error {
...
+	logrusEntry := log.WithField("component", "database")
 
-	// Access token GET request needs to be:
-	// http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=$AZURE_DBTOKEN_CLIENT_ID
-	//
-	// In this context, the "resource" parameter is passed to azidentity as a
-	// "scope" argument even though a scope normally consists of an endpoint URL.
-	scope := os.Getenv("AZURE_DBTOKEN_CLIENT_ID")
-	msiRefresherAuthorizer, err := _env.NewMSIAuthorizer(scope)
+	dbName, err := DBName(_env.IsLocalDevelopmentMode())
```
cadenmarchese, line 48: I see we still instantiate the dbname (cluster database), but how is this used to scope permissions of the MSI authorizer?

cadenmarchese (commented):
The removal / cleanup itself looks good to me, I just want to ensure that the scope of permissions isn't changing with the move to MSI.

Would be good to deploy this to canary and leave a cluster running for a bit, as well, just to be extra safe :)

Dockerfile.aro-multistage
```diff
@@ -15,5 +15,5 @@ FROM ${REGISTRY}/ubi8/ubi-minimal
 RUN microdnf update && microdnf clean all
 COPY --from=builder /app/aro /app/e2e.test /usr/local/bin/
 ENTRYPOINT ["aro"]
-EXPOSE 2222/tcp 8080/tcp 8443/tcp 8444/tcp 8445/tcp
+EXPOSE 2222/tcp 8080/tcp 8443/tcp 8444/tcp
```
SudoBrendan, line 18: Can we add this to Dockerfile.ci-rp as well please :)

cmd/aro/rp.go
```diff
@@ -108,13 +107,11 @@ func rp(ctx context.Context, log, audit *logrus.Entry) error {
...
 	}
-
 	dbAccountName := os.Getenv(envDatabaseAccountName)
-	clientOptions := &policy.ClientOptions{
-		ClientOptions: _env.Environment().ManagedIdentityCredentialOptions().ClientOptions,
-	}
+
 	logrusEntry := log.WithField("component", "database")
-	dbAuthorizer, err := database.NewMasterKeyAuthorizer(ctx, logrusEntry, msiToken, clientOptions, _env.SubscriptionID(), _env.ResourceGroup(), dbAccountName)
+	scope := []string{fmt.Sprintf("https://%s.%s", dbAccountName, _env.Environment().CosmosDBDNSSuffixScope)}
```
SudoBrendan, line 113: nit: I've seen this same code 3 times in this PR so far, would it be worth extracting into something well named?

pkg/deploy/deploy.go
SudoBrendan: do we have enough data now to remove this? What data would we need to make a decision to keep/remove this comment?

pkg/deploy/generator/resources_rp.go
```diff
@@ -334,69 +308,6 @@ func (g *generator) rpLB() *arm.Resource {
...
-				},
-				LoadBalancingRules: &[]mgmtnetwork.LoadBalancingRule{
-					{
-						LoadBalancingRulePropertiesFormat: &mgmtnetwork.LoadBalancingRulePropertiesFormat{
-							FrontendIPConfiguration: &mgmtnetwork.SubResource{
-								ID: to.StringPtr("[resourceId('Microsoft.Network/loadBalancers/frontendIPConfigurations', 'rp-lb-internal', 'dbtoken-frontend')]"),
-							},
-							BackendAddressPool: &mgmtnetwork.SubResource{
-								ID: to.StringPtr("[resourceId('Microsoft.Network/loadBalancers/backendAddressPools', 'rp-lb-internal', 'rp-backend')]"),
-							},
```
SudoBrendan, line 368: This is the only removal that concerns me here and generally with the LB, it could just be an odd naming choice, but to be sure, this isn't removing the LB config for anything related to the RP Backend service, right?

SudoBrendan (commented):
I don't think I have many concerns - can you drop me a quick comment on if removing the `rp-backend` rules from the internal LB is expected to have an impact, and/or why it's OK to delete?

pkg/deploy/generator/resources_rp.go
anshulvermapatel: `rp-lb-internal` is currently being used for gateway to request `dbtoken` service in `RP`, apart from that in my knowledge it is not used for anything else. Now since dbtoken service is being removed, there is no use of this LB

Dockerfile.aro-multistage
anshulvermapatel: Sure

cmd/aro/monitor.go
anshulvermapatel: Yeah, I am sorry for that.

cmd/aro/dbtoken.go
anshulvermapatel: So these permission are for the data plane only not the account plane.
However I think I can narrow it down by specifying scope in the role assignment to be of the ARO database.

cmd/aro/rp.go
anshulvermapatel: sure

anshulvermapatel: Well, this was the way it was previously, I think lets address this in a separate PR

cmd/aro/dbtoken.go
anshulvermapatel: Added the scope of the database rather than it being the database account. Please have a look.

hlipsig:
Pretty sure e2e failed on the cert re-running. 

/azp run e2e

hlipsig (approved):

cmd/aro/dbtoken.go
cadenmarchese: thank you!

SudoBrendan (approved):

cadenmarchese (approved):

hlipsig:
> * `rm docs/dbtoken-service.md`
> * The load balancing rule to allow gateway to talk to the dbtoken service from `pkg/deploy` (should be in templates)

Ben I'm going to make a follow on JIra and PRs for these because you're right but I don't think they should block the main body of work.

PR #3585, opened 2024-05-17
Add exemption tags for OIDC dev and RP version storage accounts

niontive:

Fixes https://issues.redhat.com/browse/ARO-7388

- Add exempt tags to storage accounts that have public blob anonymous access. This prevents an s360 violation.
- The two storage accounts are dev OIDC and RP version. These are meant to have public access

### Test plan for issue:

These are infra changes - will test via rollout in INT

https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/675077/S360-Disable-anonymous-public-blob-access

We're going to test in INT and cross fingers the exemption goes through. Also, this change just adds a tag - it's not affecting the configuration of the storage accounts.

### DIFF
--- a/pkg/deploy/assets/rp-oic.json
+++ b/pkg/deploy/assets/rp-oic.json
@@ -21,6 +21,9 @@
                 "allowBlobPublicAccess": true,
                 "minimumTlsVersion": "TLS1_2"
             },
+            "tags": {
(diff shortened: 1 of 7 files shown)


bennerv (approved):
lgtm

cadenmarchese (approved):
thanks!

PR #3586, opened 2024-05-17
Adding Azure policy exemptions to the deny assignment

shane-snyder:

Fixes https://issues.redhat.com/browse/XCMSTRAT-681

Currently, customers that need to submit Azure policy exemptions in the ARO resource group are unable to, as they're blocked by the deny assignment. For instance, many have an Azure Policy that states "Storage accounts should disable public network access", but are blocked from applying an exemption to that policy in Azure Policy. For some, doing so is needed for compliance.

### Test plan for issue:

I'm unsure how we can properly test the deny assignment prior to pushing this change.
No

This changed is just scoped to policyExemptions
https://learn.microsoft.com/en-us/azure/governance/policy/concepts/exemption-structure#required-permissions

### DIFF
--- a/pkg/cluster/deploybaseresources_additional.go
+++ b/pkg/cluster/deploybaseresources_additional.go
@@ -34,6 +34,8 @@ func (m *manager) denyAssignment() *arm.Resource {
 							"*/write",
 						},
 						NotActions: &[]string{
+							"Microsoft.Authorization/policyExemptions/write",
+							"Microsoft.Authorization/policyExemptions/delete",
 							"Microsoft.Compute/disks/beginGetAccess/action",
 							"Microsoft.Compute/disks/endGetAccess/action",
 							"Microsoft.Compute/disks/write",


SudoBrendan (changes requested):
Shane, thanks so much for this PR and for the willingness to dive in! I really do like the concept of this change, it's customer centric and I think will alleviate a LOT of toil for SREs too by empowering customers to do what is right for them and their environments.

The solution here is going in the right direction, but it's partial.

Documentation absolutely needs a lot of changes here. If nothing else, I think this change should include an adjustment to our [Support Policy](https://learn.microsoft.com/en-us/azure/openshift/support-policies-v4#cluster-management), which currently states:

```
Don't place policies within your subscription or management group that prevent SREs from performing normal maintenance against the Azure Red Hat OpenShift cluster. For example, don't require tags on the Azure Red Hat OpenShift RP-managed cluster resource group.
```

If we are now allowing customers to make policy exemptions in the managed cluster resource group, we should probably point them to do so for the sake of clarity, and to save us some tickets. One suggestiong may be something like:

```
Don't place policies within your subscription or management group that prevent SREs from performing normal maintenance against the Azure Red Hat OpenShift cluster. For example, don't require tags on the Azure Red Hat OpenShift RP-managed cluster resource group. If you have a policy that would prevent these actions, you must make the ARO cluster and its child resources exempt from these policies.
```

We will also need to update internal documentation on things like AzComms to customers when Policies are affecting resources (PUCM/Install/etc), and probably need to make changes to the RP Frontend so that errors on Policy during cluster operations don't result in 500s, and instead return some form of 4xx, and instead ask the customer to add a policy exemption to the CRG before issuing a cluster delete/recreate. Some of these actions can only be made by engineers with MSFT accounts due to how we must store our operational documentation.

All of that said, this is one of those "small" development changes that is actually quite large operationally in regards to customer vs SRE responsibility and what "ARO" as a product is. Again, I really like the change an initiative here, and I'd like to work with you to get this merged. To do so, we need to be sure this is globally communicated, fully implemented across the codebase, tested, and operationally complete before shipping it.

SREs have development environments to test such a change - but to be frank it's quite a complicated environment. I don't know where this work is currently sitting compared to other priorities on the team. I'd suggest reaching out to Team Leads and/or the BU in forum-aro to see about getting an SRE assigned to this effort (you can cc me too), unless you really feel like getting onboarded into our dev env, which I can see about helping with.

Thanks for the PR! IMO, let's work together to get it operationally in line with our production requirements and :ship:!

EDIT: for future reviewers, I was informed that these Exemptions only affect Policies `assigned with effect set to audit` - so actually, I need to rethink this stance, because I don't think SRE would even ever be aware of a policy violation in this case (it would only show up in a report for the customer).

PR #3587, opened 2024-05-18
Automate Local Image Pruning

safwank97:

Fixes https://issues.redhat.com/browse/ARO-7189

This PR helps us automatically prune local container images generated by `make ci`, so that we as developers do not run out of local disk space in our machines. Additionally it also does labeling for all the intermediate images in a multistage Docker file which could help us at the time of pruning.

### Test plan for issue:

Run `make ci-rp` to produce `aro-*` images and later run `make ci-clean` to verify and validate whether it is pruning those images to free up our local disk space.

None that I'm aware of.

### DIFF
--- a/Dockerfile.ci-rp
+++ b/Dockerfile.ci-rp
@@ -65,7 +65,7 @@ RUN hack/fips/validate-fips.sh ./aro
 # Stage 3: final is our slim image with minimal layers and tools
 ###############################################################################
 FROM ${REGISTRY}/ubi8/ubi-minimal AS final
-LABEL final=true
+LABEL aro-final=true
 RUN microdnf update && microdnf clean all
 COPY --from=builder /app/aro /app/e2e.test /usr/local/bin/
 ENTRYPOINT ["aro"]
--- a/Dockerfile.ci-rp
+++ b/Dockerfile.ci-rp
@@ -23,15 +23,8 @@ RUN npm run lint && npm run build
 # Stage 2: Compile the Golang RP code
 ###############################################################################
 FROM ${REGISTRY}/ubi8/go-toolset:1.20.12-5 AS builder
-<<<<<<< HEAD
-<<<<<<< HEAD
 ARG ARO_VERSION
-=======
-LABEL builder=true
->>>>>>> 8c2de5475 (adding labels to each stage in docker file ci-rp and building them individually)
-=======
 LABEL aro-builder=true
->>>>>>> 88c3b811f (manually fixing few things which got missed at the time of resolving the merge conflicts)
 USER root
 WORKDIR /app
 
--- a/Dockerfile.ci-rp
+++ b/Dockerfile.ci-rp
@@ -22,7 +22,11 @@ RUN npm run lint && npm run build
 # Stage 2: Compile the Golang RP code
 ###############################################################################
 FROM ${REGISTRY}/ubi8/go-toolset:1.20.12-5 AS builder
+<<<<<<< HEAD
 ARG ARO_VERSION
+=======
+LABEL builder=true
+>>>>>>> 8c2de5475 (adding labels to each stage in docker file ci-rp and building them individually)
 USER root
 WORKDIR /app
 
@@ -61,6 +65,7 @@ RUN hack/fips/validate-fips.sh ./aro
 # Stage 3: final is our slim image with minimal layers and tools
 ###############################################################################
 FROM ${REGISTRY}/ubi8/ubi-minimal AS final
+LABEL final=true
 RUN microdnf update && microdnf clean all
 COPY --from=builder /app/aro /app/e2e.test /usr/local/bin/
 ENTRYPOINT ["aro"]
--- a/Dockerfile.ci-rp
+++ b/Dockerfile.ci-rp
@@ -6,6 +6,7 @@ ARG ARO_VERSION
 # builder is responsible for all compilation and validation of the RP
 ###############################################################################
 FROM ${REGISTRY}/ubi8/nodejs-16  as portal-build
+LABEL aro-portal-build=true
 WORKDIR /build/portal/v2
 USER root
 
@@ -23,10 +24,14 @@ RUN npm run lint && npm run build
 ###############################################################################
 FROM ${REGISTRY}/ubi8/go-toolset:1.20.12-5 AS builder
 <<<<<<< HEAD
+<<<<<<< HEAD
 ARG ARO_VERSION
 =======
 LABEL builder=true
 >>>>>>> 8c2de5475 (adding labels to each stage in docker file ci-rp and building them individually)
+=======
+LABEL aro-builder=true
+>>>>>>> 88c3b811f (manually fixing few things which got missed at the time of resolving the merge conflicts)
 USER root
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -77,11 +77,16 @@ client: generate
...
-	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) --no-cache=$(NO_CACHE)
+	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) --target builder --no-cache=$(NO_CACHE) -t aro-builder
+	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) --cache-from builder -t aro-final
 
 ci-portal:
 	docker build . -f Dockerfile.ci-portal --build-arg REGISTRY=$(REGISTRY) --no-cache=$(NO_CACHE)
 
+ci-clean:
+	docker image prune --force --all --filter="label=builder=true"
+	docker image prune --force --all --filter="label=final=true"
```
SudoBrendan, line 88: If `--force` is required, I think putting a comment in the code explaining that would be helpful in the future. If it is not required, it should be removed.

SudoBrendan (changes requested):
The direction of the PR is really good. Let's align on image tags, and document/remove `--force`.

Dockerfile.ci-rp
```diff
@@ -5,6 +5,7 @@ ARG VERSION
 # builder is responsible for all compilation and validation of the RP
 ###############################################################################
 FROM ${REGISTRY}/ubi8/go-toolset:1.20.12-5 as builder
+LABEL builder=true
```
geowa4, line 8: Small request: could you prefix the label names with something like "aro-"? I think the use of labels here makes sense, but I think this small change will reduce the likelihood of a collision with other images.

Makefile
```diff
@@ -77,11 +77,16 @@ client: generate
 	hack/build-client.sh "${AUTOREST_IMAGE}" 2020-04-30 2021-09-01-preview 2022-04-01 2022-09-04 2023-04-01 2023-07-01-preview 2023-09-04 2023-11-22 2024-08-12-preview
 
 ci-rp: fix-macos-vendor
-	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) --no-cache=$(NO_CACHE)
+	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) --target builder --no-cache=$(NO_CACHE) -t aro-builder
```
geowa4, line 80: Is this step needed? With this line, it should build all the images in the multistage Dockerfile, correct?
```
docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) -t aro-final
```

SudoBrendan: I specifically asked for this, which is maybe a bit out of scope of this individual PR, but I was thinking a bit ahead to the point where we are building this dockerfile in PR builds on ephemeral CI agents without a docker cache. If we want to hit the cache from the previous build, we'll need to pull last build's `builder` image to save time, and to do that, we need to build/tag/push that stage individually.

geowa4: Ack, agreed on the comment to state the purpose.

Dockerfile.ci-rp
safwank97: Yes even I thought about the same when I released this draft pr but will complete the mentioned changes in following commits

Makefile
safwank97: We do not need to use --force flag here. It just suppresses the dialogue box of acknowledgement before deleting any images. So we can remove --force here. I'll make the changes in my upcoming commit.

Makefile
SudoBrendan: after a review, and with Shubhada's recent changes merged (which involved a scope increase) I talked with Mohammed and we decided to pull this change out - my bad, poor planning on my part.

I groomed additional work for this: https://issues.redhat.com/browse/ARO-7789

safwank97: I've done the roll back operation. And based on our discussion from last week I've committed my changes will rebase and try to resolve the merge conflicts first. And after that this PR will be ready for review.

safwank97: Regarding this https://issues.redhat.com/browse/ARO-7789 me and Brendan had discussion and we will track it with another PR and after rebase there will be some additional work to resolve merge conflicts as we got rid of ci-portal and consolidated all of those under one file which is ci-rp. And I'll take care of that.

shubhadapaithankar (dismissed):
Looks good to me

tsatam (approved):

SudoBrendan (approved):

Makefile
```diff
@@ -79,6 +79,9 @@ client: generate
 ci-rp: fix-macos-vendor
 	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) --no-cache=$(NO_CACHE)
 
+ci-clean:
```
SudoBrendan, line 82: since we are adding a new make target, we should add it to `.PHONY` at the bottom of the file.

SudoBrendan (changes requested):
Just a quick fix - we need to mark the new target as `.PHONY` for makefile reasons, since we aren't compiling a C program with this target.

Makefile
safwank97: Done :)

SudoBrendan (approved):

Makefile
```diff
@@ -278,4 +281,4 @@ vendor:
 install-go-tools:
 	go install ${GOTESTSUM}
 
-.PHONY: admin.kubeconfig aks.kubeconfig aro az ci-portal ci-rp clean client deploy dev-config.yaml discoverycache fix-macos-vendor generate image-aro-multistage image-fluentbit image-proxy init-contrib lint-go runlocal-rp proxy publish-image-aro-multistage publish-image-fluentbit publish-image-proxy secrets secrets-update e2e.test tunnel test-e2e test-go test-python vendor build-all validate-go unit-test-go coverage-go validate-fips install-go-tools
+.PHONY: admin.kubeconfig aks.kubeconfig aro az ci-rp ci-clean clean client deploy dev-config.yaml discoverycache fix-macos-vendor generate image-aro-multistage image-fluentbit image-proxy init-contrib lint-go runlocal-rp proxy publish-image-aro-multistage publish-image-fluentbit publish-image-proxy secrets secrets-update e2e.test tunnel test-e2e test-go test-python vendor build-all validate-go unit-test-go coverage-go validate-fips install-go-tools
```
tsatam, line 284: Looks like you removed `ci-portal` from here by accident?

tsatam (changes requested):

Makefile
safwank97: No it got added there by accident that's why I had to remove it.

tsatam: ...nvm, this change is desired (we removed the `ci-portal` target but never removed it from the `.PHONY` block here).

tsatam (approved):

Makefile
safwank97: Nothing to change

shubhadapaithankar (approved):

PR #3588, opened 2024-05-20
[deprecated] Initial ansible proof of concept

ventifus:

### Test plan for issue:

Then run

```
$ make smoketest
```

This will first build the required aro-smoketest container image if it doesn't exist, then it will execute `ansible-playbook` in the container. By default, the /smoketest files are baked into the container image to ensure when this is automated in the future the ansible playbook is in a known state. To test local changes, use the
```
$ make smoketest-dev
```
target; this will mount your /smoketest tree into the container.

To personalize the resulting resource groups, set the CLUSTERPREFIX as desired. For example

```
$ make smoketest CLUSTERPREFIX=${USER}
```

The default region used is `eastus`. Set the REGION parameter to choose a different region:

```
$ make smoketest REGION=centraluseuap
```

To choose one or more cluster configurations, set the CLUSTERPATTERN parameter to a wildcard string that matches the cluster scenarios you wish to test:

```
$ make smoketest CLUSTERPATTERN=smoketest-udr
```

- `smoketest-basic`: Simplest cluster, nothing fancy
- `smoketest-private`: Simple cluster with apiserver and ingress visibility set to private
- `smoketest-enc`: Encryption-at-host enabled
- `smoketest-udr`: UserDefinedRouting with a blackhole Route Table
- `smoketest-byok`: Disk encryption using bring-your-own-key

Yes

### DIFF
--- a/Dockerfile.smoketest
+++ b/Dockerfile.smoketest
@@ -1,0 +1,45 @@
+ARG REGISTRY
+ARG VERSION
+
+###############################################################################
+# final is our slim image with minimal layers and tools
+###############################################################################
+# FROM ${REGISTRY}/ubi9/python-312:1-10
+FROM ${REGISTRY}/ubi9/python-311:1-62 AS final
+ARG PIPX_VERSION=1.5.0 \
+    ANSIBLE_VERSION=9.5.1 \
+    ANSIBLE_LINT_VERSION=24.6.0 \
+    ANSIBLE_AZCOLLECTION_VERSION=2.3.0
+
+# Cause Ansible to print task timing information
+ENV ANSIBLE_CALLBACKS_ENABLED=profile_tasks
+USER root
+WORKDIR /smoketest
+
+COPY smoketest /smoketest
+
+# Using pipx here because ansible and azure-cli have differing required core Azure modules
+# They each need a separate venv to avoid collisions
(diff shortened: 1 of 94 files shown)


smoketest/regions/dev/hosts.yaml
```diff
@@ -0,0 +1,15 @@
+---
+aro_clusters:
+  hosts:
+    aro-adenton:
```
s-amann, line 4: nothing against you, :) but can this be made to be more generic?

ventifus: Oops yes, I did that but missed one place :D

pepedocs:
Looks good, I ran `make smoketest` and I was getting the following error.

```
docker run --rm -t -v ${AZURE_CONFIG_DIR:-~/.azure}:/opt/app-root/src/.azure:O aro-smoketest:83860bc -i regions/eastus -l smoketest-basic* -e CLUSTERPREFIX=smoketest  smoketest.yaml                                                                                                     
docker: Error response from daemon: invalid mode: O.
See 'docker run --help'.
make: *** [Makefile:304: smoketest] Error 125
```

I think the error is caused by docker volume mount mode to `O` which it doesn't support? I thought it was meant for podman overlay.

PR #3590, opened 2024-05-21
Retry failed ARM template deployments during cluster installation when FPSP is missing roleassignments

tsatam:

Fixes [ARO-7598](https://issues.redhat.com/browse/ARO-7598)

Handle ARM deployment preflight errors containing `"Authorization failed for template resource"` as authorization errors within our `AuthorizationRetryingAction` step. 

These errors occur in local development environments, where we fake the behavior of the first-party service principal automatically being granted Owner permissions over a cluster's managed resource group, by explicitly and manually granting this role assignment after MRG creation. This role assignment can take time to propagate, and may not be recognized by the time we apply ARM templates to the managed resource group.

ARM returns an InvalidTemplateDeployment error (HTTP 400), which was not previously considered to be an authorization error and we would thus fail the install when this occurred. 

### Test plan for issue:

- [X] Unit tests were added for the error matching functionality
- [x] E2E tests (incl. cluster creation) continue to pass
- [x] Manual test of the new functionality in a full-service RP dev environment confirmed to be experiencing the issue

No

We have not observed this specific error in production, so functionality should stay the same.

### DIFF
--- a/pkg/util/arm/deploy.go
+++ b/pkg/util/arm/deploy.go
@@ -28,13 +28,18 @@ func DeployTemplate(ctx context.Context, log *logrus.Entry, deployments features
 		},
 	})
 
+	if err == nil {
+		return nil
+	}
+
 	if azureerrors.IsDeploymentActiveError(err) {
 		log.Printf("waiting for %s template to be deployed", deploymentName)
 		err = deployments.Wait(ctx, resourceGroupName, deploymentName)
 	}
 
 	if azureerrors.HasAuthorizationFailedError(err) ||
-		azureerrors.HasLinkedAuthorizationFailedError(err) {
+		azureerrors.HasLinkedAuthorizationFailedError(err) ||
+		azureerrors.IsDeploymentMissingPermissionsError(err) {
 		return err
 	}
 
--- a/pkg/util/arm/deploy.go
+++ b/pkg/util/arm/deploy.go
@@ -28,10 +28,6 @@ func DeployTemplate(ctx context.Context, log *logrus.Entry, deployments features
 		},
 	})
 
-	if err == nil {
-		return nil
-	}
-
 	if azureerrors.IsDeploymentActiveError(err) {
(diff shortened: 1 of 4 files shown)


pkg/util/arm/deploy.go
```diff
@@ -28,13 +28,18 @@ func DeployTemplate(ctx context.Context, log *logrus.Entry, deployments features
 		},
 	})
 
+	if err == nil {
```
kimorris27, line 31: What's this particular `if` statement here for? I think you called it out in the PR review meeting, but I forgot.

tsatam: I basically added a short-circuit here, mainly just to reproduce the flaky role assignment propagation issues in my local environment when running with a debugger attached. The rest of this function handles a `nil` err the way we would have wanted, and the function will correctly return `nil` upon completion, this just allows the function to exit early upon success rather than performing a bunch of error checks against a `nil` error.

kimorris27: Ah, makes sense. As far as keeping it in the codebase goes, I see it as an unnecessary optimization that's confusing for someone else reading the code, so I'd recommend removing it. It's an opinion thing though, so no big deal either way. 🙂

kimorris27 (dismissed):

pkg/util/arm/deploy.go
tsatam: I'm happy to remove it, I'll leave this open for a bit for more thoughts though.

SudoBrendan (dismissed):

rajdeepc2792 (dismissed):

pkg/util/arm/deploy.go
tsatam: I removed the short-circuit check.

jaitaiwan (approved):
LGTM.

rajdeepc2792 (approved):

SudoBrendan (approved):

PR #3595, opened 2024-05-28
Adding hack for `allowSharedKeyAccess = false` on storage accounts

azoppiserpa:

Fixes https://issues.redhat.com/browse/ARO-7791

We need to set `allowSharedAccessKey` property to `false` on our `rp-production.json` ARM template. This PR changes `pkg/deploy/generator/templates.go` so that this happens when we run `make generate`.  

### Test plan for issue:

Test steps described in Jira issue. Local test (seeing how `rp-production.json` looks like after running `make generate` on my terminal) performed. INT tests and VMSS test yet to be done.

Q: I guess not? Please correct me if I am wrong and/or teach me how I'd find out there isn't, so that I know this for the future.

We do need to make sure that, before rolling out this, we don't have any storage accounts relying on SAS that don't have an alternative authentication method.

Q: How could we do that?

### DIFF
--- a/pkg/deploy/assets/rp-production.json
+++ b/pkg/deploy/assets/rp-production.json
@@ -541,7 +541,7 @@
             },
             "location": "[resourceGroup().location]",
             "name": "[substring(parameters('storageAccountDomain'), 0, indexOf(parameters('storageAccountDomain'), '.'))]",
-            "type": "Microsoft.Storage/storageAccounts",
+            "type": "Microsoft.Storage/storageAccounts", { "properties": "allowSharedAccessKey: false" },
             "apiVersion": "2019-06-01"
         },
         {
--- a/pkg/deploy/generator/generators.go
+++ b/pkg/deploy/generator/generators.go
@@ -114,7 +114,13 @@ func (g *generator) Artifacts() error {
 }
 
 func (g *generator) writeTemplate(t *arm.Template, output string) error {
-	b, err := g.templateFixup(t)
+
+	sharedAccessKeyHack := false
+	if output == FileRPProduction {
+		sharedAccessKeyHack = true
+	}
+
+	b, err := g.templateFixup(t, sharedAccessKeyHack)
 	if err != nil {
 		return err
 	}
--- a/pkg/deploy/generator/templates.go
+++ b/pkg/deploy/generator/templates.go
@@ -39,7 +39,7 @@ func max(is ...int) int {
 	return max
 }
 
-func (g *generator) templateFixup(t *arm.Template) ([]byte, error) {
+func (g *generator) templateFixup(t *arm.Template, sharedAccessKeyHack bool) ([]byte, error) {
 	b, err := json.MarshalIndent(t, "", "    ")
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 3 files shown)


pkg/deploy/assets/rp-production.json
```diff
@@ -541,7 +541,7 @@
             },
             "location": "[resourceGroup().location]",
             "name": "[substring(parameters('storageAccountDomain'), 0, indexOf(parameters('storageAccountDomain'), '.'))]",
-            "type": "Microsoft.Storage/storageAccounts",
+            "type": "Microsoft.Storage/storageAccounts", { "properties": "allowSharedAccessKey: false" },
```
azoppiserpa, line 544: Note to self: I can add `\n` to make this better indented. This was just a proof of concept / draft.

pkg/deploy/generator/generators.go
```diff
@@ -114,7 +114,13 @@ func (g *generator) Artifacts() error {
 }
 
 func (g *generator) writeTemplate(t *arm.Template, output string) error {
-	b, err := g.templateFixup(t)
+
+	sharedAccessKeyHack := false
+	if output == FileRPProduction {
```
azoppiserpa, line 119: For Nont and Weinong:

Without this `if`, some other files besides `rp-production.json` get modified, such as `rp-oicd.json`, among others. 

As I am not sure yet which set of the files we want to modify, I added this to help us control that.

I'm thinking probably all files, as we don't want any more SAS keys anywhere. But, still, being able to control it feels safer - I don't know if all the storage accounts have other authentication ways and are ready for SAS to be disabled.

pkg/deploy/generator/templates.go
```diff
@@ -68,6 +68,32 @@ func (g *generator) templateFixup(t *arm.Template) ([]byte, error) {
 		b = bytes.Replace(b, []byte(`"sourceAddressPrefixes": []`), []byte(`"sourceAddressPrefixes": "[parameters('rpNsgPortalSourceAddressPrefixes')]"`), 1)
 	}
 
+	if sharedAccessKeyHack {
+		b = bytes.ReplaceAll(b, []byte(`"type": "Microsoft.Storage/storageAccounts"`), []byte(`"type": "Microsoft.Storage/storageAccounts", { "properties": "allowSharedAccessKey: false" }`))
+	}
+
+	// TO-DO:
```
azoppiserpa, line 75: For Weinong and Nont: 

Added this so that we don't forget to consider doing the SDK migration instead of keeping this hack here for too long.

Also added a mention to this on the SDK migration wiki itself: https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/AzureRedHatOpenShift.wiki/604062/Azure-SDK-track-2-migration

pkg/deploy/generator/generators.go
```diff
@@ -114,7 +114,13 @@ func (g *generator) Artifacts() error {
 }
 
 func (g *generator) writeTemplate(t *arm.Template, output string) error {
-	b, err := g.templateFixup(t)
```
azoppiserpa, line 117: Note to self: fix this due to the linter failure.

azoppiserpa:
Abandoned because we won't be using this fix, we'll be updating the `mgmt/storage` package instead, and this will be done in a different PR

PR #3596, opened 2024-05-28
Updating mgmt/storage package version for rpStorageAccount() so that we can set SAS key permission to false

azoppiserpa:

Fixes https://issues.redhat.com/browse/ARO-7791 in an alternative (probably better) way than [this previous draft PR](https://github.com/Azure/ARO-RP/pull/3595).

The PR updates from a 2019 to a 2021 version of the `mgmt/storage` package of Azure SDK for Go, so that we can have a new field available in the API (`allowSharedKeyAccess`) and set it to `false` for the RP storage account struct, so that when the ARM template is generated (`rp_production.json`) it has the property set to `false` accordingly.

### Test plan for issue:

Test steps for local, INT and VMSS related validation available on the Jira issue.

Not sure, I suppose not.

Answered this on this PR description, will paste here once I know more about this or once one of the PRs is chosen, as they're both proposed fixes for the same thing: https://github.com/Azure/ARO-RP/pull/3595


pkg/deploy/assets/rp-production.json
```diff
@@ -539,6 +539,9 @@
             "sku": {
                 "name": "Standard_LRS"
             },
+            "properties": {
+                "allowSharedKeyAccess": false
+            },
             "location": "[resourceGroup().location]",
             "name": "[substring(parameters('storageAccountDomain'), 0, indexOf(parameters('storageAccountDomain'), '.'))]",
             "type": "Microsoft.Storage/storageAccounts",
```
weinong, line 547: should we update the apiVersion in L548 as well?

pkg/deploy/generator/resources.go
```diff
@@ -8,7 +8,7 @@ import (
 	mgmtkeyvault "github.com/Azure/azure-sdk-for-go/services/keyvault/mgmt/2019-09-01/keyvault"
 	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
 	mgmtinsights "github.com/Azure/azure-sdk-for-go/services/preview/monitor/mgmt/2018-03-01/insights"
-	mgmtstorage "github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-06-01/storage"
+	mgmtstorage "github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2021-09-01/storage"
```
weinong, line 11: L29 needs to be update as well

pkg/deploy/assets/rp-production.json
azoppiserpa: Yes! Now that I took a better look at `apiVersions`, I understood that it must be updated.

pkg/deploy/generator/resources.go
weinong: Sorry, I meant the line with Microsoft.Storage

azoppiserpa: Ok, thanks! Then it's done.
I created a new PR to do the whole migration thing as I was scared to break things:
https://github.com/Azure/ARO-RP/pull/3603
Will cleanup this PR later.

nwnt:
The api version of the generated arm template is still the old one.

PR #3597, opened 2024-05-28
Add explicit `ARG ARO_VERSION` to RP build stage in ci-rp

tsatam:

Fixes no Jira

The `go-toolset` image used to build the ARO binary in `make ci-rp` sets its own `VERSION` environment variable, overriding the `VERSION` arg we pass in. This results in the `version.GitCommit` ldflag being set to an incorrect value (version of Go in the go-toolset image). 

This PR changes the name of this argument to `ARO_VERSION` to avoid this naming conflict, and additionally explicitly sets this arg on the RP build stage, since all ARGs are set to nil at the completion of each stage in a multi-stage build. 

### Test plan for issue:

- Ran `make ci-rp` with this change in place, and ensured the resulting binary has the correct ldflags set for `version.GitCommit`. 

No

N/A (output of `make ci-rp` not yet used in production)

### DIFF
--- a/Dockerfile.ci-rp
+++ b/Dockerfile.ci-rp
@@ -1,5 +1,5 @@
 ARG REGISTRY
-ARG VERSION
+ARG ARO_VERSION
 
 ###############################################################################
 # Stage 1: Build the SRE Portal Assets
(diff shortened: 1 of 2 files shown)


anshulvermapatel (approved):
Good catch. LGTM

shubhadapaithankar (approved):
Looks good to me

rajdeepc2792 (approved):

SudoBrendan (approved):

PR #3603, opened 2024-05-29
Making Gateway and RP Managed Boot Diagnostics

azoppiserpa:

Fixes https://issues.redhat.com/browse/ARO-7792 and https://issues.redhat.com/browse/ARO-7791

- We are making Gateway and RP Boot Diagnostics use managed storage accounts instead of the previously used custom storage accounts, which will be deleted.
- Before, we needed to disable SAS for such storage accounts. As they will be deleted, we no longer need that, just migrating to managed Boot Diagnostics is sufficient.

### Test plan for issue:

✅ 1. Local test: I ran `make generate` and `rp-production.json` and `gateway-production.json` had the desired change on the storage account resource. I also ran `make unit-test-go` and `make build-all` to be sure I hadn't broken anything and it was compiling fine. All these commands were executed in the root of `ARO-RP`.
✅  2. INT test: deployment to int was successful, and desired settings were shown in the Portal (managed Boot Diagnostics).
✅  3. VMSS rebooting validation: rebooted the INT VMSS, and the logs had a new entry correctly added about this reboot, that could be seen from the Boot Diagnostics UI in the Portal.

I guess not. Let me know if I am wrong.

I will do a manual check in the Portal, like I did for INT, about the desired settings.

### DIFF
--- a/pkg/deploy/generator/resources_gateway.go
+++ b/pkg/deploy/generator/resources_gateway.go
@@ -346,8 +346,7 @@ func (g *generator) gatewayVMSS() *arm.Resource {
 					},
 					DiagnosticsProfile: &mgmtcompute.DiagnosticsProfile{
 						BootDiagnostics: &mgmtcompute.BootDiagnostics{
-							Enabled:    to.BoolPtr(true),
-							StorageURI: to.StringPtr("[concat('https://', parameters('gatewayStorageAccountDomain'), '/')]"),
+							Enabled: to.BoolPtr(true),
 						},
 					},
 				},
--- a/pkg/deploy/assets/gateway-production.json
+++ b/pkg/deploy/assets/gateway-production.json
@@ -294,8 +294,7 @@
                     },
                     "diagnosticsProfile": {
                         "bootDiagnostics": {
-                            "enabled": true,
-                            "storageUri": "[concat('https://', parameters('gatewayStorageAccountDomain'), '/')]"
+                            "enabled": true
                         }
                     },
                     "extensionProfile": {
--- a/pkg/deploy/assets/rp-production.json
+++ b/pkg/deploy/assets/rp-production.json
@@ -493,8 +493,7 @@
                     },
                     "diagnosticsProfile": {
                         "bootDiagnostics": {
-                            "enabled": true,
-                            "storageUri": "[concat('https://', parameters('storageAccountDomain'), '/')]"
+                            "enabled": true
                         }
                     },
                     "extensionProfile": {
--- a/pkg/deploy/generator/resources_rp.go
+++ b/pkg/deploy/generator/resources_rp.go
@@ -634,8 +634,7 @@ func (g *generator) rpVMSS() *arm.Resource {
 					},
 					DiagnosticsProfile: &mgmtcompute.DiagnosticsProfile{
 						BootDiagnostics: &mgmtcompute.BootDiagnostics{
-							Enabled:    to.BoolPtr(true),
-							StorageURI: to.StringPtr("[concat('https://', parameters('storageAccountDomain'), '/')]"),
+							Enabled: to.BoolPtr(true),
 						},
 					},
 				},


pkg/deploy/generator/resources_gateway.go
```diff
@@ -421,5 +422,8 @@ func (g *generator) gatewayRBAC() []*arm.Resource {
 }
 
 func (g *generator) gatewayStorageAccount() *arm.Resource {
-	return g.storageAccount("[substring(parameters('gatewayStorageAccountDomain'), 0, indexOf(parameters('gatewayStorageAccountDomain'), '.'))]", nil, nil)
+	accountProperties := mgmtstorage.AccountProperties{
```
nwnt, line 425: Duplicate

azoppiserpa: Updated the PR making this change in a way that is more clean and reusable, as per our discussion on Teams. Thanks!

cadenmarchese:
FYI on https://github.com/Azure/ARO-RP/pull/3564 which we are working on to interact with regional storage accounts (using some new packages). In it's current state, it doesn't look to conflict with this PR, but it would be worth taking a look to make sure we're in alignment on it.

pkg/deploy/generator/resources_oic.go
```diff
@@ -23,15 +23,15 @@ var (
 
 func (g *generator) oicStorageAccount() *arm.Resource {
 	storageAccount := &mgmtstorage.Account{
-		Kind: mgmtstorage.StorageV2,
+		Kind: mgmtstorage.KindStorageV2,
 		Sku: &mgmtstorage.Sku{
 			Name: "Standard_LRS",
 		},
 		AccountProperties: &mgmtstorage.AccountProperties{
```
rajdeepc2792, line 30: Are we scoping the OIC storage accounts too for this story?
If yes, then we need to disable the SAS keys here too.
The OIC storage accounts will be accessed via EntraID and will not need the SAS keys, thus can be disabled.
The INT and PROD pipelines for deploying the OIC storage accounts will also need a change for disabling the SAS keys.

azoppiserpa: Hi @rajdeepc2792 , to the best of my knowledge, this story doesn't have OIC in scope.
At least it didn't, at the time I got it with @nwnt .
The reason why I changed from `StorageV2` to `KindStorageV2` here was to make it compile.
The `mgmt/storage` package was upgraded to a different version for the whole repo as a pre-requisite for RP and Gateway, and there was a usage of it there in OIC that I had to change as part of this upgrade.
So it's more of a coincidence, not a SAS change that was left behind for this story.

Please correct me if I'm wrong @nwnt . Thank you!

azoppiserpa:
@cadenmarchese regarding this
> FYI on #3564 which we are working on to interact with regional storage accounts (using some new packages). In it's current state, it doesn't look to conflict with this PR, but it would be worth taking a look to make sure we're in alignment on it.

I looked at it, and actually there might be a conflict.
That PR is adding `armstorage` to the packages.
`armstorage` is a newer version of `mgmt/storage`, from SDK track 2 instead of SDK track 1.

It makes sense to me to change my PR once #3564  is merged, so that I can use `armstorage` instead of `mgmt/storage`, as the migration to `armstorage` will be done in #3564 .

@nwnt What do you think?

azoppiserpa:
@cadenmarchese @rajdeepc2792 Updating after discussion with @nwnt :

Actually I don't need to wait your PR.
I can adjust mine so that it doesn't disable SAS (as we're deleting the RP and Gateway SAs later anyways), just migrates to managed boot diagnostics.
This way there won't be a conflict with your PR, nor a dependency.

rajdeepc2792 (approved):

azoppiserpa:
Note: I will be deleting the gateway and RP storage accounts on a separate PR

hlipsig (approved):

hlipsig:
My question in slack was if no definition of a location URI causes a default to the managed storage. The answer is yes and that's been validated in Int so I'm fine to merge this PR.

PR #3605, opened 2024-05-30
change envvar name to skip pki unit tests

bitoku:

Fixes N/A

Currently to skip the pki tests, which always fail in dev env, you have to set `ARO_RUN_PKI_TESTS`.
Contrary to the name, if you set it, it **skips** the tests.

This PR changes the name to the more appropriate and understandable name, and also adds the envvar to env.example because it should be set in the dev env.

### Test plan for issue:

Run `make unit-test-go` in the devenv and confirm that the pki tests are skipped.
Confirm that CI succeeds.

N/A

N/A


tsatam:
Can you update the reference to this env var in the CI-RP Dockerfile as well? https://github.com/Azure/ARO-RP/blob/6a973ea1f10e64bf86ff97319b57d0ff6399752c/Dockerfile.ci-rp#L56

tiguelu (approved):

PR #3608, opened 2024-05-31
MIWI API endpoints

kimorris27:

https://issues.redhat.com/browse/ARO-6448

This PR introduces the API endpoints for the `PlatformWorkloadIdentityRoleSet` proxy resources as discussed in MIWI design documents. The only deviation from the design docs is that I did not implement an admin `PATCH` - it's extra work, and we are hardly if ever going to use it (refer to `update_ocp_versions.go` to see how we do `OpenShiftVersions` today and talk directly to Cosmos DB without even using the `PUT` API endpoint). But if the team reviews and thinks a `PATCH` is worth adding, we can add it.

The three new endpoints are:
- External `GET` (list)
- Admin `GET` (list)
- Admin `PUT`

### Test plan for issue:

- Unit tests
- Tested the endpoints with a full service dev RP
- (Should I test in INT as well?)

No (I think?)

Tested the endpoints with a full service RP, reused existing battle-tested frontend code, and was thorough in writing unit test cases.


pkg/api/v20240812preview/platformworkloadidentityroleset_convert.go
```diff
@@ -57,9 +60,12 @@ func (c platformWorkloadIdentityRoleSetConverter) ToInternal(_new interface{}, o
 	out.Properties.OpenShiftVersion = new.Properties.OpenShiftVersion
 	out.Properties.PlatformWorkloadIdentityRoles = make([]api.PlatformWorkloadIdentityRole, 0, len(new.Properties.PlatformWorkloadIdentityRoles))
 
-	for i, r := range new.Properties.PlatformWorkloadIdentityRoles {
-		out.Properties.PlatformWorkloadIdentityRoles[i].OperatorName = r.OperatorName
-		out.Properties.PlatformWorkloadIdentityRoles[i].RoleDefinitionName = r.RoleDefinitionName
-		out.Properties.PlatformWorkloadIdentityRoles[i].RoleDefinitionID = r.RoleDefinitionID
+	for _, r := range new.Properties.PlatformWorkloadIdentityRoles {
+		role := api.PlatformWorkloadIdentityRole{
```
cadenmarchese, line 64: Same comment as above.

pkg/frontend/changefeed.go
```diff
@@ -12,23 +12,38 @@ import (
...
 
 	t := time.NewTicker(10 * time.Second)
 	defer t.Stop()
 
-	f.updateFromIterator(ctx, t, frontendIterator)
+	f.updateFromIteratorOcpVersions(ctx, t, ocpVersionsIterator)
 }
 
-func (f *frontend) updateFromIterator(ctx context.Context, ticker *time.Ticker, frontendIterator cosmosdb.OpenShiftVersionDocumentIterator) {
+func (f *frontend) changefeedRoleSets(ctx context.Context) {
```
cadenmarchese, line 31: Could you explain (or leave a comment) on the changefeed/iterator concept?

pkg/frontend/changefeed.go
```diff
@@ -75,3 +90,47 @@ func (f *frontend) updateOcpVersions(docs []*api.OpenShiftVersionDocument) {
 		}
 	}
 }
+
+func (f *frontend) updateFromIteratorRoleSets(ctx context.Context, ticker *time.Ticker, frontendIterator cosmosdb.PlatformWorkloadIdentityRoleSetDocumentIterator) {
```
cadenmarchese, line 94: Is there an opportunity to re-use the existing `updateFromIterator()` function?

cadenmarchese (commented):
mostly lgtm! Just a few questions / comments. thanks!

cmd/aro/rp.go
```diff
@@ -173,7 +178,7 @@ func rp(ctx context.Context, log, audit *logrus.Entry) error {
 	if err != nil {
 		return err
 	}
-	f, err := frontend.NewFrontend(ctx, audit, log.WithField("component", "frontend"), _env, dbAsyncOperations, dbClusterManagerConfiguration, dbOpenShiftClusters, dbSubscriptions, dbOpenShiftVersions, api.APIs, metrics, clusterm, feAead, hiveClusterManager, adminactions.NewKubeActions, adminactions.NewAzureActions, clusterdata.NewParallelEnricher(metrics, _env))
+	f, err := frontend.NewFrontend(ctx, audit, log.WithField("component", "frontend"), _env, dbAsyncOperations, dbClusterManagerConfiguration, dbOpenShiftClusters, dbSubscriptions, dbOpenShiftVersions, dbPlatformWorkloadIdentityRoleSets, api.APIs, metrics, clusterm, feAead, hiveClusterManager, adminactions.NewKubeActions, adminactions.NewAzureActions, clusterdata.NewParallelEnricher(metrics, _env))
```
niontive, line 181: Nit: let's add a TODO to refactor `frontend.NewFrontend`. This function is getting out of control!

pkg/api/admin/platformworkloadidentityroleset_convert.go
```diff
@@ -21,12 +20,17 @@ func (c platformWorkloadIdentityRoleSetConverter) ToExternal(s *api.PlatformWork
...
 	}
 
-	for i, r := range s.Properties.PlatformWorkloadIdentityRoles {
-		out.Properties.PlatformWorkloadIdentityRoles[i].OperatorName = r.OperatorName
-		out.Properties.PlatformWorkloadIdentityRoles[i].RoleDefinitionName = r.RoleDefinitionName
-		out.Properties.PlatformWorkloadIdentityRoles[i].RoleDefinitionID = r.RoleDefinitionID
-		out.Properties.PlatformWorkloadIdentityRoles[i].ServiceAccounts = make([]string, 0, len(r.ServiceAccounts))
-		out.Properties.PlatformWorkloadIdentityRoles[i].ServiceAccounts = append(out.Properties.PlatformWorkloadIdentityRoles[i].ServiceAccounts, r.ServiceAccounts...)
+	for _, r := range s.Properties.PlatformWorkloadIdentityRoles {
+		role := PlatformWorkloadIdentityRole{
```
niontive, line 24: Nit: space between `PlatformWorkloadIdentityRole` and `{`

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
```diff
@@ -37,27 +42,25 @@ func (sv platformWorkloadIdentityRoleSetStaticValidator) validate(new *PlatformW
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "properties.platformWorkloadIdentityRoles", "Must be provided and must be non-empty")
 	}
 
-	errs := []error{}
-
 	for i, r := range new.Properties.PlatformWorkloadIdentityRoles {
 		if r.OperatorName == "" {
-			errs = append(errs, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, fmt.Sprintf("properties.platformWorkloadIdentityRoles[%d].operatorName", i), "Must be provided"))
```
niontive, line 44: Instead of comparing struct fields individually, consider using this: https://github.com/go-playground/validator

hlipsig:
I agree that patch seems unnecessary. This PR is already large, if we decide we need patch after a while then let's do it in its own PR. Still reviewing code may have more comments later.

pkg/api/admin/platformworkloadidentityroleset_convert.go
```diff
@@ -56,12 +60,16 @@ func (c platformWorkloadIdentityRoleSetConverter) ToInternal(_new interface{}, o
 	out.Properties.OpenShiftVersion = new.Properties.OpenShiftVersion
 	out.Properties.PlatformWorkloadIdentityRoles = make([]api.PlatformWorkloadIdentityRole, 0, len(new.Properties.PlatformWorkloadIdentityRoles))
 
-	for i, r := range new.Properties.PlatformWorkloadIdentityRoles {
-		out.Properties.PlatformWorkloadIdentityRoles[i].OperatorName = r.OperatorName
-		out.Properties.PlatformWorkloadIdentityRoles[i].RoleDefinitionName = r.RoleDefinitionName
-		out.Properties.PlatformWorkloadIdentityRoles[i].RoleDefinitionID = r.RoleDefinitionID
-		out.Properties.PlatformWorkloadIdentityRoles[i].ServiceAccounts = make([]string, 0, len(r.ServiceAccounts))
-		out.Properties.PlatformWorkloadIdentityRoles[i].ServiceAccounts = append(out.Properties.PlatformWorkloadIdentityRoles[i].ServiceAccounts, r.ServiceAccounts...)
+	for _, r := range new.Properties.PlatformWorkloadIdentityRoles {
```
hlipsig, line 63: Why are s.Properties and new.properties basically identical? Is there a way we can implement this once such as using generics?

hlipsig (commented):
half done review for discussion points.

pkg/api/admin/platformworkloadidentityroleset_convert.go
```diff
@@ -21,12 +20,17 @@ func (c platformWorkloadIdentityRoleSetConverter) ToExternal(s *api.PlatformWork
...
 
-	for i, r := range s.Properties.PlatformWorkloadIdentityRoles {
-		out.Properties.PlatformWorkloadIdentityRoles[i].OperatorName = r.OperatorName
-		out.Properties.PlatformWorkloadIdentityRoles[i].RoleDefinitionName = r.RoleDefinitionName
-		out.Properties.PlatformWorkloadIdentityRoles[i].RoleDefinitionID = r.RoleDefinitionID
-		out.Properties.PlatformWorkloadIdentityRoles[i].ServiceAccounts = make([]string, 0, len(r.ServiceAccounts))
-		out.Properties.PlatformWorkloadIdentityRoles[i].ServiceAccounts = append(out.Properties.PlatformWorkloadIdentityRoles[i].ServiceAccounts, r.ServiceAccounts...)
+	for _, r := range s.Properties.PlatformWorkloadIdentityRoles {
+		role := PlatformWorkloadIdentityRole{
+			OperatorName:       r.OperatorName,
```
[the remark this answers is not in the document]
hlipsig: Darn that should have been for the line below.

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
```diff
@@ -37,27 +42,25 @@ func (sv platformWorkloadIdentityRoleSetStaticValidator) validate(new *PlatformW
...
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, fmt.Sprintf("properties.platformWorkloadIdentityRoles[%d].operatorName", i), "Must be provided")
```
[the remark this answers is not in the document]
kimorris27: See the changes I made and the commit message for this commit and let me know what you think: https://github.com/Azure/ARO-RP/pull/3608/commits/214350273529cbe16efd0553b2ed36eba7623cc8

pkg/api/v20240812preview/platformworkloadidentityroleset_convert.go
```diff
@@ -24,10 +24,13 @@ func (c platformWorkloadIdentityRoleSetConverter) ToExternal(s *api.PlatformWork
 		},
 	}
 
-	for i, r := range s.Properties.PlatformWorkloadIdentityRoles {
-		out.Properties.PlatformWorkloadIdentityRoles[i].OperatorName = r.OperatorName
-		out.Properties.PlatformWorkloadIdentityRoles[i].RoleDefinitionName = r.RoleDefinitionName
-		out.Properties.PlatformWorkloadIdentityRoles[i].RoleDefinitionID = r.RoleDefinitionID
+	for _, r := range s.Properties.PlatformWorkloadIdentityRoles {
+		role := PlatformWorkloadIdentityRole{
```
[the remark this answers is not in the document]
kimorris27: See the message on this commit and let me know what you think: https://github.com/Azure/ARO-RP/pull/3608/commits/f8deb78edf13bd908393ed9e9a027cd177899747

pkg/api/v20240812preview/platformworkloadidentityroleset_convert.go
kimorris27: Similarly to the other one, I think this commit message address this concern: https://github.com/Azure/ARO-RP/pull/3608/commits/f8deb78edf13bd908393ed9e9a027cd177899747

pkg/frontend/changefeed.go
kimorris27: I don't think so, at least not without a non-trivial refactor.

The issue is that the generated database clients, specifically the ones for `PlatformWorkloadIdentityRoleSets` and `OpenShiftVersions` in this case, don't implement a common interface that we could leverage to deduplicate this code. I think it would require a refactor of the database client generation to add such an interface, and after doing that we would have to regenerate the clients and rewrite this code to use the interface.

pkg/frontend/changefeed.go
kimorris27: This doc goes through it in detail: https://learn.microsoft.com/en-us/azure/cosmos-db/change-feed.

But the general idea is that rather than issuing a `DELETE` to get rid of an entry in the database, you `PUT/PATCH` it to set the `Deleting` field, and then Cosmos DB knows to remove the entry from the database after the TTL has passed. So when we are updating the role sets that are in the frontend cache, we look at each database entry's `Deleting` to see if the entry is mid-deletion, and if so, we remove it/refrain from adding it to the cache.

Is that what you were asking?

pkg/api/admin/platformworkloadidentityroleset_convert.go
kimorris27: I'd put a space before the curly brace if for example this were line 23 and I had left out the space after the `range` expression, but I think it's normal Go style to avoid leaving blank space in that spot in a struct literal, and that's what's done throughout the codebase.

pkg/api/admin/platformworkloadidentityroleset_convert.go
kimorris27: I think in order to leverage generics here `api.PlatformWorkloadIdentityRoleSet`, the internal type, and `admin.PlatformWorkloadIdentityRoleSet`, the admin API type, would have to implement a common interface. If we could do that and wanted to do that, I think that refactor would be outside of the scope of this PR since we'd want to makesimilar improvements across all of the various API converters (like https://github.com/Azure/ARO-RP/blob/master/pkg/api/admin/openshiftcluster_convert.go for example).

cmd/aro/rp.go
kimorris27: Done!

pkg/api/v20240812preview/platformworkloadidentityroleset_convert.go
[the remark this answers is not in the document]
cadenmarchese: Ah, I also thought that. Thanks!

pkg/frontend/changefeed.go
cadenmarchese: Yep, thanks for the doc - this tracks with our current CosmosDB usage.

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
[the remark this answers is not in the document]
cadenmarchese: I see. I think it's fine to fail fast versus having to adjust the error-catching logic. Maybe we could put in a TODO if you think it's worth the cleanup later, but up to you.

pkg/frontend/admin_platformworkloadidentityroleset_list_test.go
```diff
@@ -14,7 +14,7 @@ import (
 	testdatabase "github.com/Azure/ARO-RP/test/database"
 )
 
-func TestPlatformWorkloadIdenittyRoleSetList(t *testing.T) {
```
cadenmarchese: I keep doing the same thing :)

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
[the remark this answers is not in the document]
kimorris27: I'd be cool with cleaning this up now except that I'm not sure how. Do you have any ideas in mind?

pkg/frontend/admin_platformworkloadidentityroleset_list_test.go
kimorris27: I caught this one and fixed it myself before opening the PR. I think you're commenting on the content of one of my older commits.

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
kimorris27: I played around with it, but I don't think it's good for our static validators simply because it doesn't give the user friendly error messages that we want going to those making API requests. For example:

```
Key: 'PlatformWorkloadIdentityRole.OperatorName' Error:Field validation for 'OperatorName' failed on the 'required' tag
```

This error message would be perfectly fine for me as a regular developer of the RP, but it doesn't even include the fully qualified path of the API field presenting the issue, which won't be great for external users. (Well, I say "external users", but this is the admin API... :slightly_smiling_face:)

I think adding it in just for this PR would create an inconsistency with respect to our error messaging and that with that in mind it would be best considered as a future refactor that we do across all of the static validators. Your thoughts?

pkg/api/admin/platformworkloadidentityroleset_convert.go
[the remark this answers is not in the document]
hlipsig: OperatorName:       r.OperatorName, the new line between these two pieces. It matches the following line style:

RoleDefinitionName: 
r.RoleDefinitionName,


I'm just not sure why it's different, but it could be totally correct that's why this is a question.

pkg/api/admin/platformworkloadidentityroleset_convert.go
hlipsig: Thanks for the clarification. We'll put that on the maybe someday list.

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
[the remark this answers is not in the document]
bitoku: How about wrapping the joined errors like this?

```golang
return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, errors.Join(errs...).Error())
```

In the first place, we should use `errors.Is` to compare errors instead of type assertion.
If we use it, we can compare wrapped errors.

https://pkg.go.dev/errors#Join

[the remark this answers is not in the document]
bitoku: I made a ticket of that https://issues.redhat.com/browse/ARO-7922

pkg/api/admin/platformworkloadidentityroleset_convert.go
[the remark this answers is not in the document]
kimorris27: There's no newline between `RoleDefinitionName:` and `r.RoleDefinitionName,`, and none of the struct fields are done that way. The GitHub UI sometimes inserts newlines and indents to get things to fit on the screen... maybe it confused you here by doing that?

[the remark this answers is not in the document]
hlipsig: Yep it's the UI then. Thanks I didn't know it did that (thanks GH).

pkg/frontend/admin_platformworkloadidentityroleset_list_test.go
```diff
@@ -0,0 +1,159 @@
...
+		name           string
+		fixture        func(f *testdatabase.Fixture)
+		wantStatusCode int
+		wantResponse   *admin.PlatformWorkloadIdentityRoleSetList
+		wantError      string
+	}
+
+	for _, tt := range []*test{
+		{
+			name:           "empty",
```
SudoBrendan, line 30: nit: it is clearer to name test cases with expected prereqs and outcomes. For example, 6mo down the line when this fails, `empty` is a lot less clear than `empty list of WorkloadIdentityRoleSet cosmos documents returns StatusOK` due to some other change in the code elsewhere. I would add the same nit for all other test cases below.

pkg/frontend/admin_platformworkloadidentityroleset_list_test.go
```diff
@@ -0,0 +1,159 @@
...
+									ServiceAccounts: []string{
+										"openshift-cloud-controller-manager:cloud-controller-manager",
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
```
SudoBrendan, line 128: I see some code paths in the function that are not tested, I'd suggest adding a few more test cases here:

1. cosmosdb errors result in 500 Internal Server Error
2. the returned result is sorted in an expected way, even if cosmos returns things not sorted
3. errors in json parsing result in an error

By adding these tests, we'll codify the requirements of the code and make it clear what our needs are (in plain English sentences with some good test names), while also ensuring those features remain in place over time.

pkg/frontend/admin_platformworkloadidentityroleset_list.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/coreos/go-semver/semver"
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/api/admin"
+	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+)
+
+func (f *frontend) getAdminPlatformWorkloadIdentityRoleSets(w http.ResponseWriter, r *http.Request) {
```
SudoBrendan, line 20: nit: the file is named `list`, and the only function in it is a `get` (albeit, a plural one ending in `s`). Keeping consistent names helps identify and clearly map code to files and their test cases, so I'd suggest picking one verb for everything. I think you did this perfectly in the `put` files, code, and their tests - can we do that here too?

pkg/frontend/admin_platformworkloadidentityroleset_list.go
```diff
@@ -0,0 +1,49 @@
...
+	ctx := r.Context()
+	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
+	r.URL.Path = filepath.Dir(r.URL.Path)
+
+	converter := f.apis[admin.APIVersion].PlatformWorkloadIdentityRoleSetConverter
+
+	docs, err := f.dbPlatformWorkloadIdentityRoleSets.ListAll(ctx)
+	if err != nil {
```
SudoBrendan, line 28: this error is getting swallowed - can we at least log it if we aren't returning it directly? Is there a reason to _not_ return the error in the HTTP response on `/admin`?

pkg/frontend/admin_platformworkloadidentityroleset_put.go
```diff
@@ -0,0 +1,93 @@
...
+
+	var ext *admin.PlatformWorkloadIdentityRoleSet
+	err := json.Unmarshal(body, &ext)
+	if err != nil {
+		api.WriteError(w, http.StatusBadRequest, api.CloudErrorCodeInvalidRequestContent, "", "The request content could not be deserialized: "+err.Error())
+		return
+	}
+
+	docs, err := f.dbPlatformWorkloadIdentityRoleSets.ListAll(ctx)
+	if err != nil {
```
SudoBrendan, line 40: error swallowing.

pkg/frontend/admin_platformworkloadidentityroleset_put.go
```diff
@@ -0,0 +1,93 @@
...
+		adminReply(log, w, nil, []byte{}, err)
+		return
+	}
+
+	converter.ToInternal(ext, roleSetDoc.PlatformWorkloadIdentityRoleSet)
+
+	if isCreate {
+		roleSetDoc, err = f.dbPlatformWorkloadIdentityRoleSets.Create(ctx, roleSetDoc)
+		if err != nil {
```
SudoBrendan, line 74: error swallowing.

pkg/frontend/admin_platformworkloadidentityroleset_put.go
```diff
@@ -0,0 +1,93 @@
...
+			api.WriteError(w, http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "Internal server error.")
+			return
+		}
+	} else {
+		roleSetDoc, err = f.dbPlatformWorkloadIdentityRoleSets.Update(ctx, roleSetDoc)
+		if err != nil {
```
SudoBrendan, line 80: error swallowing.

pkg/frontend/admin_platformworkloadidentityroleset_put_test.go
```diff
@@ -0,0 +1,790 @@
...
+		body           *admin.PlatformWorkloadIdentityRoleSet
+		wantStatusCode int
+		wantResponse   *admin.PlatformWorkloadIdentityRoleSet
+		wantError      string
+		wantDocuments  []*api.PlatformWorkloadIdentityRoleSetDocument
+	}
+
+	for _, tt := range []*test{
+		{
+			name: "updating known version",
```
SudoBrendan, line 32: nit: would again be good to get a full sentence as the name - updating (I would probably rename this to `putting` or `PUT on known version` or similar for consistency) a known version is doing what exactly? Same for a few other tests in this file.

pkg/frontend/changefeed_test.go
```diff
@@ -194,3 +194,345 @@ func TestUpdateFromIterator(t *testing.T) {
...
+
+func TestUpdateFromIteratorRoleSets(t *testing.T) {
+	for _, tt := range []struct {
+		name           string
+		docsInIterator []*api.PlatformWorkloadIdentityRoleSetDocument
+		roleSets       map[string]*api.PlatformWorkloadIdentityRoleSet
+		wantRoleSets   map[string]*api.PlatformWorkloadIdentityRoleSet
+	}{
+		{
+			name: "add to empty",
```
SudoBrendan, line 206: nit: same naming conventions as noted above would be beneficial here.

pkg/frontend/admin_platformworkloadidentityroleset_list.go
SudoBrendan: ok, reviewing the rest of our code I am seeing we do this _everywhere_ e.g. `openshiftclusters_list` with a `getOpenshiftClusters` function inside :face_exhaling: Gonna keep the comment for posterity, but nothing to do. Good work following convention.

pkg/frontend/openshiftcluster_preflightvalidation_test.go
```diff
@@ -158,14 +158,14 @@ func TestPreflightValidation(t *testing.T) {
...
-			f, err := NewFrontend(ctx, ti.audit, ti.log, ti.env, ti.asyncOperationsDatabase, ti.clusterManagerDatabase, ti.openShiftClustersDatabase, ti.subscriptionsDatabase, ti.openShiftVersionsDatabase, api.APIs, &noop.Noop{}, &noop.Noop{}, nil, nil, nil, nil, nil)
+			f, err := NewFrontend(ctx, ti.audit, ti.log, ti.env, ti.asyncOperationsDatabase, ti.clusterManagerDatabase, ti.openShiftClustersDatabase, ti.subscriptionsDatabase, ti.openShiftVersionsDatabase, ti.platformWorkloadIdentityRoleSetsDatabase, api.APIs, &noop.Noop{}, &noop.Noop{}, nil, nil, nil, nil, nil)
 			if err != nil {
 				t.Fatal(err)
 			}
 			oc := tt.preflightRequest()
 
 			go f.Run(ctx, nil, nil)
-			f.mu.Lock()
+			f.ocpVersionsMu.Lock()
```
SudoBrendan, line 168: This might be ok - but whenever I see changes to concurrency I kinda freak out and get antsy :p is it required to redo how we perform frontend mutex locking for this change, or is it a premature optimization outside the acceptance criteria? These changes feel like something, given the scope of other changes, that might be better in their own PR so it is more clear exactly what changes are needed for this concurrency updates, and what the benefits are, if at all possible.

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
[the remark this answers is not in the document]
kimorris27: I pushed a commit to wrap multiple validation issues with the `PlatformWorkloadIdentityRoles` together. `api.NewCloudError` works a little bit differently from in your example, so I had to tweak it a bit. Let me know what you all think.

pkg/frontend/admin_platformworkloadidentityroleset_put.go
```diff
@@ -0,0 +1,93 @@
...
+			api.WriteError(w, http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "Internal server error.")
+			return
+		}
+	}
+
+	b, err := json.MarshalIndent(converter.ToExternal(roleSetDoc.PlatformWorkloadIdentityRoleSet), "", "    ")
+	if err == nil {
+		if isCreate {
+			err = statusCodeError(http.StatusCreated)
```
SudoBrendan, line 89: nit: idk if there's any way to make this read better, but this line of code hurts my brain - we are returning a 201, do we gotta have `statusCodeError` to set that?

pkg/frontend/platformworkloadidentityrolesets_list.go
```diff
@@ -0,0 +1,45 @@
...
+	"net/http"
+
+	"github.com/go-chi/chi/v5"
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+)
+
+func (f *frontend) listPlatformWorkloadIdentityRoleSets(w http.ResponseWriter, r *http.Request) {
```
SudoBrendan, line 18: :cry: well now I have to tell you to make this a `get`.

SudoBrendan (commented):
Alright - some changes got pushed in the middle of my review, I hope stuff is still relevant :joy: ...

What a fun PR to review! Thanks so much for your efforts here Kipp. Left some comments, overall, stuff is super clean and the changes overall make sense. Left some comments and nits, but I don't think any of them are blocking? If it looks janky, just dismiss and give me a ping so I can re-review.

pkg/frontend/admin_platformworkloadidentityroleset_list.go
kimorris27: Yep - even though it's a little odd, I did it for the sake of following the convention.

pkg/frontend/admin_platformworkloadidentityroleset_list.go
```diff
@@ -0,0 +1,49 @@
...
+			roleSets = append(roleSets, doc.PlatformWorkloadIdentityRoleSet)
+		}
+	}
+
+	sort.Slice(roleSets, func(i, j int) bool {
+		appendPatch := func(in string) string {
+			return in + ".0"
+		}
+		return semver.New(appendPatch(roleSets[i].Properties.OpenShiftVersion)).LessThan(*semver.New(appendPatch(roleSets[j].Properties.OpenShiftVersion)))
+	})
```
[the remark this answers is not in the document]
kimorris27: I had intended to do that and I forgot! Thanks for catching it.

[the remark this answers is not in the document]
kimorris27: I pushed a commit to address this, but you'll have to let me know if it's close to what you had in mind. I think we could get fancier if we introduced generics for API types, but my goal with this solution was to make the code more readable and document my logic without letting the scope creep on an already-large PR.

pkg/frontend/admin_platformworkloadidentityroleset_list.go
kimorris27: To be frank, I copied and pasted `pkg/frontend/admin_openshiftversion_list.go` and tweaked it, and this is a less-than-ideal piece that got copied over that I didn't pay much attention to.

+1 to your comment though, and I'll fix it in this file, but I'm not going to fix it in the other file to avoid scope creep. We can add a Jira to the backlog for making similar logging improvements to other endpoints.

pkg/frontend/admin_platformworkloadidentityroleset_put.go
kimorris27: Same response as https://github.com/Azure/ARO-RP/pull/3608#discussion_r1626650921.

pkg/frontend/admin_platformworkloadidentityroleset_put.go
kimorris27: Same response as https://github.com/Azure/ARO-RP/pull/3608#discussion_r1626650921.

pkg/frontend/admin_platformworkloadidentityroleset_put.go
kimorris27: Same response as https://github.com/Azure/ARO-RP/pull/3608#discussion_r1626650921.

pkg/frontend/platformworkloadidentityrolesets_list.go
kimorris27: Similar comment to some other places - I copy/pasted this from `pkg/frontend/openshiftversions_list.go` and that's how it turned out like this :sweat_smile:

This is another one that I'd like to leave as-is to avoid scope creep if that works for you, even though I agree with your idea.

pkg/frontend/openshiftcluster_preflightvalidation_test.go
kimorris27: The reason for the rename of this mutex lock and the addition of a second one for the `PlatformWorkloadIdentityRoleSets` is that it logically doesn't make sense for one goroutine to be unable to edit the `PlatformWorkloadIdentityRoleSets` because another is editing the `OpenShiftVersions` in Cosmos DB.

I could use one mutex for this PR and add the second in a follow-up if that would be better, but we have to modify concurrency stuff no matter what in this PR because we have to use _a mutex lock_ around accesses to the `PlatformWorkloadIdentityRoleSets` - it's just a matter of whether it's the same old one or the new one I added.

pkg/frontend/admin_platformworkloadidentityroleset_put.go
kimorris27: I took this from existing code, and it's done this way in difference places throughout the frontend package, so I'd like to leave this one as-is to avoid scope creep if that's okay.

pkg/frontend/admin_platformworkloadidentityroleset_list_test.go
kimorris27: I agree. I added the first two cases you recommended, but I had trouble figuring out a good way to do the third one... I think it might require using mocks for the API converters or something like that.

Unless you see a quick way to do it, I'd like to leave it for a future refactor, since I couldn't find that case being tested in any other endpoints (and therefore don't have a similar instance to go off of).

pkg/frontend/openshiftcluster_preflightvalidation_test.go
SudoBrendan: correct - I'm just trying to wrap my head around concurrent writes to different tables. I understand that from a Cosmos perspective, we should be able to write to both containers at the same time, but my question is, can the RP handle that? Would it be safer code-wise to keep all locks as the same thing, even if it is inefficient, so we don't risk a future change mistakingly using the wrong lock, because the RP doesn't write enough to Cosmos for us to care? ... Maybe I'm over-thinking this. :sweat_smile:

SudoBrendan (approved):
LGTM. I left one final comment I honestly don't know the answer to - just want to get a thought out there. Great work!

cadenmarchese (approved):
lgtm!

pkg/frontend/openshiftcluster_preflightvalidation_test.go
kimorris27: Oh, I see what you're saying. I agree that functionally, for us, one mutex lock for both containers would be fine since we edit both containers very rarely and they each contain a small amount of data and hence each won't one doesn't take long to update.

rajdeepc2792 (approved):

PR #3611, opened 2024-06-03
Add validation for default CIDR range

rhamitarora:
[CIF - Prevent creating OVNKubernetes Clusters with VNET ranges of 100.64.0.0/16](https://issues.redhat.com/browse/ARO-4743)

OVN-Kubernetes, the default network provider in OpenShift Container Platform 4.11 and later, uses the _100.64.0.0/16_ IP address range internally. Added validation that Pod and Service CIDR must not use above IP address.

**Reference Documentation:** https://docs.openshift.com/container-platform/4.14/networking/cidr-range-definitions.html
### Test plan for issue:

- Unit test added for this specific case
- All existing unit and e2e tests continue to pass

No

### DIFF
--- a/pkg/api/openshiftcluster.go
+++ b/pkg/api/openshiftcluster.go
@@ -367,6 +367,15 @@ type NetworkProfile struct {
 	LoadBalancerProfile        *LoadBalancerProfile `json:"loadBalancerProfile,omitempty"`
 }
 
+// IP address ranges internally used by ARO
+var (
+	JoinCIDRRange []string = []string{
+		"100.64.0.0/16",
+		"169.254.169.0/29",
+		"100.88.0.0/16",
+	}
+)
+
 // PreconfiguredNSG represents whether customers want to use their own NSG attached to the subnets
 type PreconfiguredNSG string
 
--- a/pkg/api/v20191231preview/openshiftcluster_validatestatic.go
+++ b/pkg/api/v20191231preview/openshiftcluster_validatestatic.go
@@ -189,13 +189,19 @@ func (sv openShiftClusterStaticValidator) validateServicePrincipalProfile(path s
 
 func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np *NetworkProfile) error {
 	podIP, pod, err := net.ParseCIDR(np.PodCIDR)
-
 	if err != nil {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided pod CIDR '%s' is invalid: '%s'.", np.PodCIDR, err)
 	}
 
 	if pod.IP.To4() == nil {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided pod CIDR '%s' is invalid: must be IPv4.", np.PodCIDR)
+	}
+
+	for _, s := range api.JoinCIDRRange {
+		_, cidr, _ := net.ParseCIDR(s)
+		if cidr.Contains(pod.IP) || pod.Contains(cidr.IP) {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidCIDRRange, path, "Azure Red Hat OpenShift uses 100.64.0.0/16, 169.254.169.0/29, and 100.88.0.0/16 IP address ranges internally. Do not include this '%s' IP address range in any other CIDR definitions in your cluster.", np.PodCIDR)
+		}
 	}
 
 	ones, _ := pod.Mask.Size()
@@ -216,6 +222,13 @@ func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np
 
 	if service.IP.To4() == nil {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".serviceCidr", "The provided service CIDR '%s' is invalid: must be IPv4.", np.ServiceCIDR)
+	}
+
+	for _, s := range api.JoinCIDRRange {
+		_, cidr, _ := net.ParseCIDR(s)
+		if cidr.Contains(service.IP) || service.Contains(cidr.IP) {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidCIDRRange, path, "Azure Red Hat OpenShift uses 100.64.0.0/16, 169.254.169.0/29, and 100.88.0.0/16 IP address ranges internally. Do not include this '%s' IP address range in any other CIDR definitions in your cluster.", np.ServiceCIDR)
+		}
 	}
 
 	ones, _ = service.Mask.Size()
--- a/pkg/api/v20191231preview/openshiftcluster_validatestatic_test.go
+++ b/pkg/api/v20191231preview/openshiftcluster_validatestatic_test.go
@@ -461,6 +461,48 @@ func TestOpenShiftClusterStaticValidateNetworkProfile(t *testing.T) {
 			},
 			wantErr: "400: InvalidNetworkAddress: properties.networkProfile.serviceCidr: The provided service CIDR '10.0.150.0/16' is invalid, expecting: '10.0.0.0/16'.",
 		},
+		{
+			name: "podCidr invalid CIDR-1",
+			modify: func(oc *OpenShiftCluster) {
+				oc.Properties.NetworkProfile.PodCIDR = "100.64.0.0/18"
+			},
+			wantErr: "400: InvalidCIDRRange: properties.networkProfile: Azure Red Hat OpenShift uses 100.64.0.0/16, 169.254.169.0/29, and 100.88.0.0/16 IP address ranges internally. Do not include this '100.64.0.0/18' IP address range in any other CIDR definitions in your cluster.",
+		},
+		{
+			name: "podCidr invalid CIDR-2",
+			modify: func(oc *OpenShiftCluster) {
+				oc.Properties.NetworkProfile.PodCIDR = "169.254.169.0/29"
+			},
+			wantErr: "400: InvalidCIDRRange: properties.networkProfile: Azure Red Hat OpenShift uses 100.64.0.0/16, 169.254.169.0/29, and 100.88.0.0/16 IP address ranges internally. Do not include this '169.254.169.0/29' IP address range in any other CIDR definitions in your cluster.",
+		},
+		{
+			name: "podCidr invalid CIDR-3",
+			modify: func(oc *OpenShiftCluster) {
+				oc.Properties.NetworkProfile.PodCIDR = "100.88.0.0/16"
+			},
+			wantErr: "400: InvalidCIDRRange: properties.networkProfile: Azure Red Hat OpenShift uses 100.64.0.0/16, 169.254.169.0/29, and 100.88.0.0/16 IP address ranges internally. Do not include this '100.88.0.0/16' IP address range in any other CIDR definitions in your cluster.",
+		},
+		{
+			name: "serviceCidr invalid CIDR-1",
+			modify: func(oc *OpenShiftCluster) {
+				oc.Properties.NetworkProfile.ServiceCIDR = "100.64.0.0/16"
+			},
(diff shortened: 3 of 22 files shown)


pkg/api/v20191231preview/openshiftcluster_validatestatic.go
```diff
@@ -188,6 +188,10 @@ func (sv openShiftClusterStaticValidator) validateServicePrincipalProfile(path s
 }
 
 func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np *NetworkProfile) error {
+	if np.PodCIDR == api.JoinCIDR || np.ServiceCIDR == api.JoinCIDR {
```
bitoku, line 191: It can't check the overlap with the different subnet mask.
We may want to use this https://pkg.go.dev/github.com/apparentlymart/go-cidr/cidr#VerifyNoOverlap.

rolandmkunkel (dismissed):

pkg/api/v20191231preview/openshiftcluster_validatestatic.go
rajdeepc2792: Just to add on top of Bitoku comments, the unit test cases for all the internalAPI versions can also use the podCIDR/serviceCIDR such that it covers the reserved ranges partially. For Eg. podCIDR: 100.64.0.0/18

rhamitarora: The requested changes completed. New approach https://stackoverflow.com/a/34729915 used

pkg/api/openshiftcluster.go
```diff
@@ -367,6 +367,11 @@ type NetworkProfile struct {
 	LoadBalancerProfile        *LoadBalancerProfile `json:"loadBalancerProfile,omitempty"`
 }
 
+// In ARO 4.11 and beyond, OVN-Kubernetes, the default network provider, internally operates within the IP address range of 100.64.0.0/16.
+const (
+	JoinCIDR = "100.64.0.0/16"
```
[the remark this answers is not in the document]
rhamitarora: Other ranges added.

pkg/api/v20191231preview/openshiftcluster_validatestatic_test.go
```diff
@@ -461,6 +461,48 @@ func TestOpenShiftClusterStaticValidateNetworkProfile(t *testing.T) {
 			},
 			wantErr: "400: InvalidNetworkAddress: properties.networkProfile.serviceCidr: The provided service CIDR '10.0.150.0/16' is invalid, expecting: '10.0.0.0/16'.",
 		},
+		{
+			name: "podCidr invalid CIDR-1",
+			modify: func(oc *OpenShiftCluster) {
+				oc.Properties.NetworkProfile.PodCIDR = "100.64.0.0/16"
```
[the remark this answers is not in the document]
rhamitarora: 3 test cases updated with partial overlapping.

pkg/api/v20191231preview/openshiftcluster_validatestatic.go
```diff
@@ -198,6 +197,13 @@ func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided pod CIDR '%s' is invalid: must be IPv4.", np.PodCIDR)
 	}
 
+	for _, s := range api.JoinCIDRRange {
+		_, cidr, _ := net.ParseCIDR(s)
+		if pod.Contains(cidr.IP) || cidr.Contains(pod.IP) {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidCIDRRange, path, "Azure Red Hat OpenShift uses 100.64.0.0/16, 169.254.169.0/29, 100.88.0.0/16, fd98::/64, fd69::/125, and fd97::/64 IP address range internally. Do not include this '%s' IP address range in any other CIDR definitions in your cluster.", np.PodCIDR)
```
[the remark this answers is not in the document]
rhamitarora: Statement updated and IPV6 removed as we are handling them via other condition.

pkg/api/v20191231preview/openshiftcluster_validatestatic_test.go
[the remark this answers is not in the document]
bitoku: It seems this file has the same test cases.
Others seem ok.

pkg/api/v20191231preview/openshiftcluster_validatestatic.go
```diff
@@ -198,6 +197,13 @@ func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided pod CIDR '%s' is invalid: must be IPv4.", np.PodCIDR)
 	}
 
+	for _, s := range api.JoinCIDRRange {
+		_, cidr, _ := net.ParseCIDR(s)
+		if cidr.Contains(pod.IP) {
```
bitoku, line 202: I don't think this works.
For example, when they set `100.0.0.0/8` as podCidr, `100.0.0.1`, which is the first IP of it, is not contained by any `JoinCIDRRange`, but it should be rejected because `100.0.0.0/8` overlaps with (completely includes) `100.64.0.0/16`.

we need the both-way check.

pkg/api/v20191231preview/openshiftcluster_validatestatic_test.go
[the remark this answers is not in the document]
rhamitarora: Test case updated.

pkg/api/v20191231preview/openshiftcluster_validatestatic.go
rhamitarora: I modified the condition for both-way check, however add condition to skip 0.0.0.0/0 also

bitoku: Why do we have to skip 0.0.0.0/0? Should 0.0.0.0/0 be allowed?
If it's because of the test failures, we have to change the existing test cases instead.

https://github.com/Azure/ARO-RP/blob/a69d5854e4f248c1e7b224330a4cb5ee5d6744f7/pkg/api/v20191231preview/openshiftcluster_validatestatic_test.go#L828-L836

bitoku: Why do we skip 0.0.0.0/0? Should it be allowed?
If we skip it because of the test failures, the tests are something that should be fixed.

https://github.com/Azure/ARO-RP/blob/a69d5854e4f248c1e7b224330a4cb5ee5d6744f7/pkg/api/v20191231preview/openshiftcluster_validatestatic_test.go#L827-L836

It would be better to change this 0.0.0.0/0 to other cidr that should be allowed so that we can check `PropertyChangeNotAllowed`.

pkg/api/v20210901preview/openshiftcluster_validatestatic.go
```diff
@@ -201,6 +200,16 @@ func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided pod CIDR '%s' is invalid: must be IPv4.", np.PodCIDR)
 	}
 
+	switch np.SoftwareDefinedNetwork {
```
bitoku, line 203: nit: we can make it `if`.

This is not a change request, but I'll write it down what I thought.
I thought we should have the same validation in other versions.
If someone tries to change podCIDR to 100.64.0.0/16 in the old cluster with old SDN, it will get the overlap error, but it should be `PropertyChangeNotAllowed`.
However only this version has the `SoftwareDefinedNetwork` field, so other versions don't have a way to know which SDN is being used.
Changing the fields would be too much effort, so I think we're good to go.

pkg/api/v20210901preview/openshiftcluster_validatestatic.go
```diff
@@ -221,6 +230,16 @@ func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".serviceCidr", "The provided service CIDR '%s' is invalid: must be IPv4.", np.ServiceCIDR)
 	}
 
+	switch np.SoftwareDefinedNetwork {
```
bitoku, line 233: ditto

bitoku (approved):
LGTM. I left nit comments, but you don't have to fix them.

anshulvermapatel (dismissed):
LGTM

pkg/api/v20191231preview/openshiftcluster_validatestatic.go
```diff
@@ -198,6 +197,13 @@ func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".podCidr", "The provided pod CIDR '%s' is invalid: must be IPv4.", np.PodCIDR)
 	}
 
+	for _, s := range api.JoinCIDRRange {
+		_, cidr, _ := net.ParseCIDR(s)
+		if cidr.Contains(pod.IP) || pod.Contains(cidr.IP) {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidCIDRRange, path, "Azure Red Hat OpenShift uses 100.64.0.0/16, 169.254.169.0/29, and 100.88.0.0/16 IP address range internally. Do not include this '%s' IP address range in any other CIDR definitions in your cluster.", np.PodCIDR)
```
SrinivasAtmakuri, line 203: nit;
```suggestion
			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidCIDRRange, path, "Azure Red Hat OpenShift uses 100.64.0.0/16, 169.254.169.0/29, and 100.88.0.0/16 IP address ranges internally. Do not include this '%s' IP address range in any other CIDR definitions in your cluster.", np.PodCIDR)
```

pkg/api/v20191231preview/openshiftcluster_validatestatic.go
```diff
@@ -218,6 +224,13 @@ func (sv openShiftClusterStaticValidator) validateNetworkProfile(path string, np
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, path+".serviceCidr", "The provided service CIDR '%s' is invalid: must be IPv4.", np.ServiceCIDR)
 	}
 
+	for _, s := range api.JoinCIDRRange {
+		_, cidr, _ := net.ParseCIDR(s)
+		if cidr.Contains(service.IP) || service.Contains(cidr.IP) {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidCIDRRange, path, "Azure Red Hat OpenShift uses 100.64.0.0/16, 169.254.169.0/29, and 100.88.0.0/16 IP address range internally. Do not include this '%s' IP address range in any other CIDR definitions in your cluster.", np.ServiceCIDR)
```
SrinivasAtmakuri, line 230:
```suggestion
			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidCIDRRange, path, "Azure Red Hat OpenShift uses 100.64.0.0/16, 169.254.169.0/29, and 100.88.0.0/16 IP address ranges internally. Do not include this '%s' IP address range in any other CIDR definitions in your cluster.", np.ServiceCIDR)
```

SrinivasAtmakuri (commented):
LGTM! Re-ran E2E as it failed.

pkg/api/v20191231preview/openshiftcluster_validatestatic.go
rhamitarora: Done

pkg/api/v20191231preview/openshiftcluster_validatestatic.go
rhamitarora: Done

anshulvermapatel (dismissed):

SrinivasAtmakuri (dismissed):
LGTM

jaitaiwan (dismissed):
I agree with @bitoku's nit. There's only one case in the switch so this really should be an if statement.

anshulvermapatel (approved):

SrinivasAtmakuri (approved):

PR #3612, opened 2024-06-04
Update coordinates for MDM/MDSD images

tsatam:

Fixes [ARO-7837](https://issues.redhat.com/browse/ARO-7837)

Updates our coordinates for Geneva MDM/MDSD images:
- Removes the Geneva image mirroring step from `cmd/aro/mirror.go`, as we have a dedicated pipeline for this purpose now
  - The MDM/MDSD versions used in this PR correspond to the currently-configured versions in this new pipeline. 
- Explicitly use the distroless variant of these images
- Ensures all references/usages to MDM/MDSD images across the RP/Gateway and on-cluster are compatible with the `/distroless/` prefix on image names
  - Updates RP/Gateway VMSS scripts to ensure they preserve the `/distroless/` prefix

### Test plan for issue:

- [x] All unit and E2E tests continue to pass
- [X] Manual test of the RP and Gateway VMSS deployments in a full-service environment
- [X] Cluster creation successfully brings up MDSD on the bootstrap node
  - This is validating the equivalent change to the installer wrapper: https://github.com/openshift/installer-aro-wrapper/pull/122 
- [x] ARO operator on-cluster successfully deploys MDSD daemonset

- Our internal docs on MDM/MDSD image refresh will need to be updated.

See testing done above. We have existing E2E tests to ensure the health of the MDSD daemonset on-cluster to ensure the change works in production as well.


pkg/deploy/generator/scripts/gatewayVMSS.sh
```diff
@@ -150,7 +150,7 @@ touch /etc/containers/nodocker
 mkdir -p /root/.docker
 REGISTRY_AUTH_FILE=/root/.docker/config.json az acr login --name "$(sed -e 's|.*/||' <<<"$ACRRESOURCEID")"
 
-MDMIMAGE="${RPIMAGE%%/*}/${MDMIMAGE##*/}"
+MDMIMAGE="${RPIMAGE%%/*}/${MDMIMAGE#*/}"
```
bitoku, line 153: It's just nit, but I'm wondering if we can set it directly without substring removal. (maybe we need more code changes)
It was a bit difficult to imagine what `MDMIMAGE` would be like.

bitoku (approved):

pkg/deploy/generator/scripts/gatewayVMSS.sh
tsatam: This was difficult to understand for me as well, I agree that we potentially should do something different here. My understanding of the intent here is as follows:
- We set the `MDMIMAGE` env var on the VMSS to a value without a specified registry, e.g. `/genevamdm:${TAG}` before this change, and `/distroless/genevamdm:${TAG}` after. We don't know what image registry we're pulling from, as it's different depending on the environment the RP is running in. 
- In the code here, we extract the registry from the `RPIMAGE` env var, and prefix it to the MDM image name from above. The code before would only keep everything in the `MDMIMAGE` value "after the last `/`", my change here is to keep everything "after the first `/`" (to preserve the `/distroless/` prefix). 

I'll bring this up with @s-fairchild, as he is working on #3534 right now as a more comprehensive restructure of these VMSS scripts.

bitoku: Thank you for clarifying!
good to know that there is an ongoing refactor task.

rajdeepc2792 (approved):

SudoBrendan (approved):
I trust the tests here - assuming we are still seeing logs/metrics in Geneva in those tests, LGTM.

cadenmarchese (approved):

PR #3615, opened 2024-06-05
Removing the RP and Gateway SA creation functions

azoppiserpa:

Fixes https://issues.redhat.com/browse/ARO-7792 and https://issues.redhat.com/browse/ARO-7791

Removes gateway and RP storage accounts that are used to store Boot Diagnostics logs, because we're now using Managed Boot Diagnostics and thus these two accounts are no longer necessary.

Previous PR making the Boot Diagnostics managed has already been merged: https://github.com/Azure/ARO-RP/pull/3603

### Test plan for issue:

Unit tests:
- Ran `go test` on `pkg/deploy` and all passed.

- Ran `make generate` and it generated the JSONs without the removed resources and parameters and no other undesired changes.

Integration tests: not applicable.

If these two accounts are listed anywhere in our docs, we need to remove them from the list, as they will no longer exist, and explaining that our boot diagnostics is now of managed type. Are there any such documents? I can update them.

This is only removing resources, there isn't much to be verified I guess?


hlipsig (approved):

hlipsig:
Would like a passing e2e before this merges. e2e currently broken we need to regenerate a certificate it uses. Working on getting it fixed.

hlipsig (approved):

PR #3617, opened 2024-06-06
New cmd to populate PlatformWorkloadIdentityRoleSets in dev and prod

cadenmarchese:

Fixes https://issues.redhat.com/browse/ARO-6449

Adds a new cmd for populating our new proxy resource (PlatformWorkloadIdentityRoleSetDocument) with role sets in prod, and adds the hack code to populate it in dev too. This resource is used to determine identity requirements for supported OCP versions.

### Test plan for issue:

- [x] CI
- [x] e2e
- [x] Run a dev RP ARO install using these changes, confirm that the role set is populated in cosmosDB
- [x] Confirm prod script works using local environment variables set
  - Steps to test:
  1) `export PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS='[{"openShiftVersion":"4.14","platformWorkloadIdentityRoles":[{"operatorName":"CloudControllerManager","roleDefinitionName":"Azure RedHat OpenShift Cloud Controller Manager Role","roleDefinitionId":"/providers/Microsoft.Authorization/roleDefinitions/a1f96423-95ce-4224-ab27-4e3dc72facd4","serviceAccounts":["openshift-cloud-controller-manager:cloud-controller-manager"]},{"operatorName":"ClusterIngressOperator","roleDefinitionName":"Azure RedHat OpenShift Cluster Ingress Operator Role","roleDefinitionId":"/providers/Microsoft.Authorization/roleDefinitions/0336e1d3-7a87-462b-b6db-342b63f7802c","serviceAccounts":["openshift-ingress-operator:ingress-operator"]},{"operatorName":"MachineApiOperator","roleDefinitionName":"Azure RedHat OpenShift Machine API Operator Role","roleDefinitionId":"/providers/Microsoft.Authorization/roleDefinitions/0358943c-7e01-48ba-8889-02cc51d78637","serviceAccounts":["openshift-machine-api:machine-api-operator"]},{"operatorName":"StorageOperator","roleDefinitionName":"Azure RedHat OpenShift Storage Operator Role","roleDefinitionId":"/providers/Microsoft.Authorization/roleDefinitions/5b7237c5-45e1-49d6-bc18-a1f62f400748","serviceAccounts":["openshift-cluster-csi-drivers:azure-disk-csi-driver-operator","openshift-cluster-csi-drivers:azure-disk-csi-driver-controller-sa"]},{"operatorName":"NetworkOperator","roleDefinitionName":"Azure RedHat OpenShift Network Operator Role","roleDefinitionId":"/providers/Microsoft.Authorization/roleDefinitions/be7a6435-15ae-4171-8f30-4a343eff9e8f","serviceAccounts":["openshift-cloud-network-config-controller:cloud-network-config-controller"]},{"operatorName":"ImageRegistryOperator","roleDefinitionName":"Azure RedHat OpenShift Image Registry Operator Role","roleDefinitionId":"/providers/Microsoft.Authorization/roleDefinitions/8b32b316-c2f5-4ddf-b05b-83dacd2d08b5","serviceAccounts":["openshift-image-registry:cluster-image-registry-operator","openshift-image-registry:registry"]},{"operatorName":"AzureFilesStorageOperator","roleDefinitionName":"Azure RedHat OpenShift Azure Files Storage Operator Role","roleDefinitionId":"/providers/Microsoft.Authorization/roleDefinitions/0d7aedc0-15fd-4a67-a412-efad370c947e","serviceAccounts":["openshift-cluster-csi-drivers:azure-file-csi-driver-operator","openshift-cluster-csi-drivers:azure-file-csi-driver-controller-sa","openshift-cluster-csi-drivers:azure-file-csi-driver-node-sa"]},{"operatorName":"ServiceOperator","roleDefinitionName":"Azure RedHat OpenShift Service Operator","roleDefinitionId":"/providers/Microsoft.Authorization/roleDefinitions/4436bae4-7702-4c84-919b-c4069ff25ee2","serviceAccounts":["openshift-azure-operator:aro-operator-master"]}]}]'`
  2) Source regular env vars, and then run `go run ./cmd/aro update-role-sets` and observe that the role set above is populated.

### DIFF
--- a/cmd/aro/const.go
+++ b/cmd/aro/const.go
@@ -4,9 +4,10 @@ package main
 // Licensed under the Apache License 2.0.
 
 const (
-	envDatabaseName          = "DATABASE_NAME"
-	envDatabaseAccountName   = "DATABASE_ACCOUNT_NAME"
-	envKeyVaultPrefix        = "KEYVAULT_PREFIX"
-	envOpenShiftVersions     = "OPENSHIFT_VERSIONS"
-	envInstallerImageDigests = "INSTALLER_IMAGE_DIGESTS"
+	envDatabaseName                     = "DATABASE_NAME"
+	envDatabaseAccountName              = "DATABASE_ACCOUNT_NAME"
+	envKeyVaultPrefix                   = "KEYVAULT_PREFIX"
+	envOpenShiftVersions                = "OPENSHIFT_VERSIONS"
+	envInstallerImageDigests            = "INSTALLER_IMAGE_DIGESTS"
+	envPlatformWorkloadIdentityRoleSets = "PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS"
 )
--- a/cmd/aro/update_role_sets.go
+++ b/cmd/aro/update_role_sets.go
@@ -20,9 +20,9 @@ import (
 )
 
 // 1 - Get env data from agent VMs (with getEnvironmentData)
-func getRoleSetsFromEnv() ([]api.PlatformWorkloadIdentityRoleSet, error) {
+func getRoleSetsFromEnv() ([]api.PlatformWorkloadIdentityRoleSetProperties, error) {
 	const envKey = envPlatformWorkloadIdentityRoleSets
-	var roleSets []api.PlatformWorkloadIdentityRoleSet
+	var roleSets []api.PlatformWorkloadIdentityRoleSetProperties
 
 	// Unmarshal env data into type api.PlatformWorkloadIdentityRoleSet
 	if err := getEnvironmentData(envKey, &roleSets); err != nil {
@@ -103,9 +103,9 @@ func updatePlatformWorkloadIdentityRoleSetsInCosmosDB(ctx context.Context, dbPla
 		return err
 	}
 
-	newRoleSets := make(map[string]api.PlatformWorkloadIdentityRoleSet)
+	newRoleSets := make(map[string]api.PlatformWorkloadIdentityRoleSetProperties)
 	for _, doc := range incomingRoleSets {
-		newRoleSets[doc.Properties.OpenShiftVersion] = doc
+		newRoleSets[doc.OpenShiftVersion] = doc
 	}
 
 	// 3 - Put/patch the new role sets to the doc, overwriting whatever is there for that version, or adding if new
@@ -113,16 +113,16 @@ func updatePlatformWorkloadIdentityRoleSetsInCosmosDB(ctx context.Context, dbPla
 	for _, doc := range existingRoleSets.PlatformWorkloadIdentityRoleSetDocuments {
 		existing, found := newRoleSets[doc.PlatformWorkloadIdentityRoleSet.Properties.OpenShiftVersion]
 		if found {
-			log.Printf("Found Version %q, patching", existing.Properties.OpenShiftVersion)
+			log.Printf("Found Version %q, patching", existing.OpenShiftVersion)
 			_, err := dbPlatformWorkloadIdentityRoleSets.Patch(ctx, doc.ID, func(inFlightDoc *api.PlatformWorkloadIdentityRoleSetDocument) error {
-				inFlightDoc.PlatformWorkloadIdentityRoleSet = &existing
+				inFlightDoc.PlatformWorkloadIdentityRoleSet.Properties = existing
 				return nil
 			})
 			if err != nil {
 				return err
 			}
-			log.Printf("Version %q found", existing.Properties.OpenShiftVersion)
-			delete(newRoleSets, existing.Properties.OpenShiftVersion)
+			log.Printf("Version %q found", existing.OpenShiftVersion)
+			delete(newRoleSets, existing.OpenShiftVersion)
 			continue
 		}
 
@@ -140,11 +140,14 @@ func updatePlatformWorkloadIdentityRoleSetsInCosmosDB(ctx context.Context, dbPla
 	}
 
 	for _, doc := range newRoleSets {
-		log.Printf("Version %q not found in database, creating", doc.Properties.OpenShiftVersion)
+		log.Printf("Version %q not found in database, creating", doc.OpenShiftVersion)
 		newDoc := api.PlatformWorkloadIdentityRoleSetDocument{
-			ID:                              dbPlatformWorkloadIdentityRoleSets.NewUUID(),
-			PlatformWorkloadIdentityRoleSet: &doc,
+			ID: dbPlatformWorkloadIdentityRoleSets.NewUUID(),
+			PlatformWorkloadIdentityRoleSet: &api.PlatformWorkloadIdentityRoleSet{
+				Properties: doc,
+			},
 		}
+
 		_, err := dbPlatformWorkloadIdentityRoleSets.Create(ctx, &newDoc)
 		if err != nil {
 			return err
--- a/cmd/aro/update_role_sets.go
+++ b/cmd/aro/update_role_sets.go
@@ -25,11 +25,7 @@ func getRoleSetsFromEnv() ([]api.PlatformWorkloadIdentityRoleSetProperties, erro
 	var roleSets []api.PlatformWorkloadIdentityRoleSetProperties
 
 	// Unmarshal env data into type api.PlatformWorkloadIdentityRoleSet
-	if err := getEnvironmentData(envKey, &roleSets); err != nil {
-		return nil, err
-	}
-
-	return roleSets, nil
+	return roleSets, getEnvironmentData(envKey, &roleSets)
 }
 
 func getPlatformWorkloadIdentityRoleSetDatabase(ctx context.Context, log *logrus.Entry) (database.PlatformWorkloadIdentityRoleSets, error) {
@@ -82,12 +78,8 @@ func getPlatformWorkloadIdentityRoleSetDatabase(ctx context.Context, log *logrus
 	if err != nil {
 		return nil, err
 	}
-	dbPlatformWorkloadIdentityRoleSets, err := database.NewPlatformWorkloadIdentityRoleSets(ctx, dbc, dbName)
-	if err != nil {
-		return nil, err
-	}
 
-	return dbPlatformWorkloadIdentityRoleSets, nil
+	return database.NewPlatformWorkloadIdentityRoleSets(ctx, dbc, dbName)
 }
 
 func updatePlatformWorkloadIdentityRoleSetsInCosmosDB(ctx context.Context, dbPlatformWorkloadIdentityRoleSets database.PlatformWorkloadIdentityRoleSets, log *logrus.Entry) error {
--- a/cmd/aro/update_role_sets.go
+++ b/cmd/aro/update_role_sets.go
@@ -19,7 +19,7 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/keyvault"
 )
 
-// 1 - Get env data from agent VMs (with getEnvironemntData) and write to types created in step 1
+// 1 - Get env data from agent VMs (with getEnvironmentData) and write to types created in step 1
 func getPlatformWorkloadIdentityRoleSet() (*api.PlatformWorkloadIdentityRoleSet, error) {
 	const envKey = envPlatformWorkloadIdentityRoleSets
 	var PlatformWorkloadIdentityRoleSet api.PlatformWorkloadIdentityRoleSet
@@ -47,7 +47,7 @@ func getRoleSetFromEnv() ([]api.PlatformWorkloadIdentityRoleSet, error) {
 // 2 - Get the existing role set documents, if existing
 // Mostly copied from update_ocp_versions.go
 func getPlatformWorkloadIdentityRoleSetDatabase(ctx context.Context, log *logrus.Entry) (database.PlatformWorkloadIdentityRoleSets, error) {
-	_env, err := env.NewCore(ctx, log, env.COMPONENT_UPDATE_OCP_VERSIONS)
+	_env, err := env.NewCore(ctx, log, env.COMPONENT_UPDATE_ROLE_SETS)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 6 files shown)


cmd/aro/update_role_sets.go
```diff
@@ -0,0 +1,290 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("MSI Authorizer failed with: %s", err.Error())
+	}
+
+	msiKVAuthorizer, err := _env.NewMSIAuthorizer(_env.Environment().KeyVaultScope)
+	if err != nil {
+		return nil, fmt.Errorf("MSI KeyVault Authorizer failed with: %s", err.Error())
+	}
+
+	// is this needed in development, or just prod?
```
kimorris27, line 166: I think the OCP versions script is only used in prod, and we can assume this one will only be used in prod too. For dev, you can just make API calls to set up whatever OCP versions or role sets you need.

kimorris27 (changes requested):
I'm not clear on why there are multiple env vars instead of just one, but as a draft it LGTM so far aside from the API endpoint needing to be fixed in `pkg/util/cluster`.

cmd/aro/const.go
```diff
@@ -4,10 +4,16 @@ package main
...
+	envKeyVaultPrefix                = "KEYVAULT_PREFIX"
+	envDBTokenUrl                    = "DBTOKEN_URL"
+	envOpenShiftVersions             = "OPENSHIFT_VERSIONS"
+	envInstallerImageDigests         = "INSTALLER_IMAGE_DIGESTS"
+	envPlatformWorkloadIdentityRoles = "PLATFORM_WORKLOAD_IDENTITY_ROLES"
+	envOpenShiftVersion              = "OPENSHIFT_VERSION"
+	envOperatorName                  = "OPERATOR_NAME"
+	envRoleDefinitionName            = "ROLE_DEFINITION_NAME"
+	envRoleDefinitionId              = "ROLE_DEFINITION_ID"
+	envServiceAccounts               = "SERVICE_ACCOUNTS"
```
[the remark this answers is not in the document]
cadenmarchese: Changed to use the single env var, thanks! I had those because I thought they'd be needed to populate the fields.

cmd/aro/update_role_sets.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database"
+	"github.com/Azure/ARO-RP/pkg/env"
+	"github.com/Azure/ARO-RP/pkg/metrics/statsd"
+	"github.com/Azure/ARO-RP/pkg/util/encryption"
+	"github.com/Azure/ARO-RP/pkg/util/keyvault"
+)
+
+// 1 - Get env data from agent VMs (with getEnvironemntData) and write to types created in step 1
```
kimorris27, line 22: Nit: `getEnvironemntData` -> `getEnvironmentData`

cmd/aro/update_role_sets.go
```diff
@@ -0,0 +1,191 @@
...
+func getPlatformWorkloadIdentityRoleSet() (*api.PlatformWorkloadIdentityRoleSet, error) {
```
kimorris27, line 23: RP-Config will store a list of role sets as opposed to a single role set, so we should modify this function accordingly (have it read in a list of role sets from the env variable).

cmd/aro/update_role_sets.go
```diff
@@ -0,0 +1,191 @@
...
+	finalRoleSet := []api.PlatformWorkloadIdentityRoleSet{}
+	finalRoleSet = append(finalRoleSet, *roleSet)
+
+	return finalRoleSet, nil
+}
+
+// 2 - Get the existing role set documents, if existing
+// Mostly copied from update_ocp_versions.go
+func getPlatformWorkloadIdentityRoleSetDatabase(ctx context.Context, log *logrus.Entry) (database.PlatformWorkloadIdentityRoleSets, error) {
+	_env, err := env.NewCore(ctx, log, env.COMPONENT_UPDATE_OCP_VERSIONS)
```
kimorris27, line 50: We can add a new component and use it here.

cmd/aro/update_role_sets.go
```diff
@@ -0,0 +1,191 @@
...
+	dbc, err := database.NewDatabaseClient(log.WithField("component", "database"), _env, dbAuthorizer, m, aead, dbAccountName)
+	if err != nil {
+		return nil, err
+	}
+
+	dbName, err := DBName(_env.IsLocalDevelopmentMode())
+	if err != nil {
+		return nil, err
+	}
+	dbPlatformWorkloadIdentityRoleSetsDocument, err := database.NewPlatformWorkloadIdentityRoleSets(ctx, dbc, dbName)
```
kimorris27, line 99: Nit: this is a client and not a document, so we could remove `Document` from the name for clarity.

cmd/aro/update_role_sets.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, err
+	}
+
+	return dbPlatformWorkloadIdentityRoleSetsDocument, nil
+}
+
+// 3 - Put/patch the new role sets to the doc, overwriting whatever is there for that version, or adding if new
+// Mostly copied from update_ocp_versions.go
+func updatePlatformWorkloadIdentityRoleSetsInCosmosDB(ctx context.Context, dbPlatformWorkloadIdentityRoleSets database.PlatformWorkloadIdentityRoleSets, log *logrus.Entry) error {
+	dbPlatformWorkloadIdentityRoleSet, err := dbPlatformWorkloadIdentityRoleSets.ListAll(ctx)
```
kimorris27, line 110: Nit: We could rename this to something like `existingRoleSets` for clarity.

kimorris27 (changes requested):
It's just about done, but we need to tweak it to read in a list of role sets as opposed to just one role set.

kimorris27:
I made the necessary changes to have the script read in a list of role sets and tested locally. If anyone reading would like to test, follow Caden's original instructions from the PR description but add a set of square brackets (`[ ]`) around the contents of the JSON-formatted environment variable contents to make it into a list with one element in it.

kimorris27 (dismissed):

cadenmarchese (commented):
Tested @kimorris27's changes and they LGTM.

kimorris27:
Sticking a hold label on this while we work out how to address an issue I encountered while testing the Azure pipeline that will use this script.

kimorris27 (approved):
The approach of unmarshaling the data into the properties makes sense to me. I also confirmed that it works in the context of the Azure pipeline/reading from RP-Config.

bennerv (commented):
I need to make another pass but just a few things I've noticed.

pkg/api/v20230904/clustermanager.go
```diff
@@ -1,4 +1,4 @@
-package v20230904
+package v20240812preview
```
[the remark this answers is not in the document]
cadenmarchese: Yes, it was cherry-picked over for some smoke testing and shouldn't have been pushed - I'll roll it back.

pkg/util/rolesets/const.go
```diff
@@ -0,0 +1,84 @@
+package rolesets
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"github.com/Azure/ARO-RP/pkg/api"
+)
+
+var DefaultPlatformWorkloadIdentityRoleSet = api.PlatformWorkloadIdentityRoleSet{
```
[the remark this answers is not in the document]
cadenmarchese: Should it be a Go embed or an environment variable in the secrets? I mostly copied the approach from the way OCP versions handles it now, but I'm definitely open to a better idea.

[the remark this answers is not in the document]
cadenmarchese: Since the role sets are stored in RP-Config but we don't have RP-Config in dev, I think it would have to be sourced in the secrets or left as-is.

cmd/aro/update_role_sets.go
```diff
@@ -0,0 +1,174 @@
...
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database"
+	"github.com/Azure/ARO-RP/pkg/env"
+	"github.com/Azure/ARO-RP/pkg/metrics/statsd"
+	"github.com/Azure/ARO-RP/pkg/util/encryption"
+	"github.com/Azure/ARO-RP/pkg/util/keyvault"
+)
+
+// 1 - Get env data from agent VMs (with getEnvironmentData)
```
mociarain, line 22: Why the 1,2 & 3 here and below? If there's some ordering as to how these should be called is that seems better placed in the place where they are called or in a general overview of the flow IMO

cmd/aro/const.go
```diff
@@ -4,9 +4,10 @@ package main
...
-	envDatabaseAccountName   = "DATABASE_ACCOUNT_NAME"
-	envKeyVaultPrefix        = "KEYVAULT_PREFIX"
-	envOpenShiftVersions     = "OPENSHIFT_VERSIONS"
-	envInstallerImageDigests = "INSTALLER_IMAGE_DIGESTS"
+	envDatabaseName                     = "DATABASE_NAME"
+	envDatabaseAccountName              = "DATABASE_ACCOUNT_NAME"
+	envKeyVaultPrefix                   = "KEYVAULT_PREFIX"
+	envOpenShiftVersions                = "OPENSHIFT_VERSIONS"
+	envInstallerImageDigests            = "INSTALLER_IMAGE_DIGESTS"
+	envPlatformWorkloadIdentityRoleSets = "PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS"
```
mociarain, line 12: Since these are being exported shouldn't they all be PascalCase? https://go.dev/ref/spec#Exported_identifiers

cmd/aro/update_role_sets.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		log.Printf("Version %q not found, deleting", doc.PlatformWorkloadIdentityRoleSet.Properties.OpenShiftVersion)
+		// Delete via changefeed
+		_, err := dbPlatformWorkloadIdentityRoleSets.Patch(ctx, doc.ID,
+			func(d *api.PlatformWorkloadIdentityRoleSetDocument) error {
+				d.PlatformWorkloadIdentityRoleSet.Deleting = true
+				d.TTL = 60
+				return nil
+			})
```
mociarain, line 128: It's not clear to me why it's being deleted via a Patch rather than using the actual delete method. Can you flesh this out for me?

pkg/env/core.go
```diff
@@ -26,6 +26,7 @@ const (
 	COMPONENT_MIRROR              ServiceComponent = "MIRROR"
 	COMPONENT_PORTAL              ServiceComponent = "PORTAL"
 	COMPONENT_UPDATE_OCP_VERSIONS ServiceComponent = "UPDATE_OCP_VERSIONS"
+	COMPONENT_UPDATE_ROLE_SETS    ServiceComponent = "UPDATE_ROLE_SETS"
 	COMPONENT_DEPLOY              ServiceComponent = "DEPLOY"
 	COMPONENT_TOOLING             ServiceComponent = "TOOLING"
```
mociarain, line 31: Since these are being exported shouldn't they all be PascalCase? https://go.dev/ref/spec#Exported_identifiers

mociarain (commented):
I left some comments around style but some about context.

cmd/aro/update_role_sets.go
cadenmarchese: Sure - we don't ever delete docs from CosmosDB directly, but instead patch them with Deleting and a TTL where it is then cleaned up by CosmosDB itself. We have this convention in the update_ocp_versions script as well.

cmd/aro/update_role_sets.go
cadenmarchese: It's more of a logical order of steps that I used to help guide writing / reading the script. I can remove code comments if you don't feel like they are needed, or I can remove the numbering to avoid ordering confusion.

cmd/aro/const.go
cadenmarchese: These consts are only used in package `main`, so I don't see a reason to export them.

pkg/env/core.go
cadenmarchese: These consts are only used in package main, so I don't see a reason to export them.

cmd/aro/update_role_sets.go
mociarain: IMO the comments can go. The code is pretty clear for me

cmd/aro/const.go
mociarain: Ahh. It's package scope not file scope

cmd/aro/update_role_sets.go
mociarain: Hmmm. Why the delete method then? From a quick search we never use the Delete method. I wonder would wrapping the "deleting patch" in the Delete method be clearer? Doesn't block this at all but I've made a note for myself

mociarain (dismissed):
LGTM

cmd/aro/update_role_sets.go
cadenmarchese: done!

pkg/util/rolesets/const.go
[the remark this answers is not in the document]
kimorris27: In the long term, a Go embed sounds nice because you'd be able to copy-paste the one you want to use in your dev environment from RP-Config into the file and then have the script read it in. But I feel like it would make sense to leave that as a refactor for later so that it can be done to this script and the OCP versions script at the same time.

kimorris27 (approved):

cmd/aro/update_role_sets.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+
+	newRoleSets := make(map[string]api.PlatformWorkloadIdentityRoleSetProperties)
+	for _, doc := range incomingRoleSets {
+		newRoleSets[doc.OpenShiftVersion] = doc
+	}
+
+	for _, doc := range existingRoleSets.PlatformWorkloadIdentityRoleSetDocuments {
+		existing, found := newRoleSets[doc.PlatformWorkloadIdentityRoleSet.Properties.OpenShiftVersion]
```
tiguelu, line 101: Would it make sense to rename `existing`? We are using that term for the role sets which are already in the DB in line 85, and this var will have the value of an incoming role set, not the current existing value in DB.

cadenmarchese: I can change it to `incoming` if that works.

pkg/util/rolesets/const.go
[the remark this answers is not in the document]
cadenmarchese: Agreed, I'm going to resolve this comment with justification that any refactor here should also include the openShiftVersions script (since this is mostly copied over from that) and is something that we could tackle separately.

kimorris27 (approved):

pkg/util/rolesets/const.go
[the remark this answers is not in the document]
tiguelu: This could be a `good-first-item` jira to have in the backlog. Has it been carded?

tiguelu (approved):

mociarain (approved):

kimorris27 (approved):

PR #3616, opened 2024-06-06
update SDK to track2 in ensureGatewayCreate

bitoku:
### Which issue this PR addresses:

Fixes a part of https://issues.redhat.com/browse/ARO-4665 and https://issues.redhat.com/browse/ARO-7316

This is blocked until https://github.com/Azure/ARO-RP/pull/3638 is released.

### What this PR does / why we need it:

This PR makes `ensureGateway` action use track2 SDK.

### Test plan for issue:

unittest for ensureGateway.
e2e

### Is there any documentation that needs to be updated for this PR?

tech debt cleanup N/A

### How do you know this will function as expected in production? 

cluster installation test

PR #3618, opened 2024-06-06
update SDK to track2 in ensureServiceEndpoints

bitoku:

Fixes a part of https://issues.redhat.com/browse/ARO-4665 and https://issues.redhat.com/browse/ARO-7316

This is blocked by https://github.com/Azure/ARO-RP/pull/3616

This PR makes `ensureServiceEndpoints` action use track2 SDK.

It also contains the change of subnet manager, and I'm going to remove the subnet manager completely in the future.
I want to remove the subnet manager because some methods (`CreateOrUpdateFromIds`, `CreateOrUpdateSubnets`, and `GetHighestFreeIP`) are used only once, and I think it adds unnecessary complexity.
The merit to have it which I can think of is that it can be mocked and can keep the unit tests simple, but after removing it, the test is not that complex and I think the suggested change would make the behaviour clearer.

Any comments about removal of the subnet manager are welcome.

### Test plan for issue:

unit tests for ensureServiceEndpoints.
e2e

tech debt cleanup N/A

cluster installation test


pkg/cluster/ensureendpoints.go
```diff
@@ -14,12 +19,35 @@ import (
...
+			return err
+		}
+		shouldUpdate := addEndpointsToSubnet(api.SubnetsEndpoints, &subnet.Subnet)
+		if !shouldUpdate {
+			continue
+		}
+		err = m.armSubnets.CreateOrUpdateAndWait(ctx, r.ResourceGroupName, r.Parent.Name, r.Name, subnet.Subnet, nil)
+		if err != nil {
+			return err
+		}
```
AldoFusterTurpin, line 48: Just a comment off the top of my head.

There is nothing wrong with this, but a slightly different approach that could also work would be to divide that in the following way: 
- first get all the subnets that should be updated (in a function returning a slice)
- pass those subnets that should be updated (return value from the previous function) to another function that would simply do _CreateOrUpdateAndWait_ for each value in the slice.

I don't think is much better than what the PR does, but it sticks to the principle of not iterating + doing stuff with what you are iterating, but instead, first do some processing and once that is processed update the values in Azure. Not asking to change that, just a comment about a pattern that sometimes can be useful, even to simplify unit tests in some cases (as you don't need to mock _m.armSubnets_ because you just want to test the function that does the business logic which is filtering the subnets).

🙂

AldoFusterTurpin (dismissed):
Providing review as requested 🙂  
LGTM. Just left a comment about one chunk of code to share an idea, NOT asking to change it. Thank you.

PR #3619, opened 2024-06-06
Dynamic validation for workload identity permissions and requirements

rajdeepc2792:

JIRA: [ARO-4376](https://issues.redhat.com/browse/ARO-4376)

For the MIWI Cluster feature, this PR performs dynamic validation for the Platform Workload Identities and Cluster MSI.
Uses CheckAccessV2 to validate the actions for all the Platform Workload Identities and Cluster MSI.

During Cluster Creation:-
- Fetch PlatformWorkloadIdentityRoleSet for the requested OCP version.
- Match the count and OperatorNames associated with customer provided Platform Workload Identities and fetched PlatformWorkloadIdentityRoleSet.
- Check permissions on the network resources for all the Platform Workload Identities using CheckAccessV2
- Check the count of Cluster MSI(User Assigned Identity), it should be one.
- Check the permission/actions(RoleAzureRedHatOpenShiftFederatedCredentialRole) on Platform Workload Identity for Cluster MSI.

### Test plan for issue:
- [x] Unit Tests
- [x] CI
- [x] CI e2e
- [x] Local cluster installation

No

None of the environment should have an impact due to this change as the additional validation will only happen for MIWI clusters.


pkg/portal/middleware/aad.go
```diff
@@ -308,7 +309,7 @@ func (a *aad) callback(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	groupsIntersect := GroupsIntersect(a.allGroups, claims.Groups)
+	groupsIntersect := stringutils.GroupsIntersect(a.allGroups, claims.Groups)
```
cadenmarchese, line 312: thanks for the cleanup!

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -0,0 +1,108 @@
...
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+func (dv *dynamic) ValidatePlatformWorkloadIdentityProfile(ctx context.Context, oc *api.OpenShiftCluster, platformWorkloadIdentityRoles []api.PlatformWorkloadIdentityRole, roleDefinitions armauthorization.RoleDefinitionsClient) error {
+	dv.log.Print("ValidatePlatformWorkloadIdentityProfile")
+
+	platformIdentitiesActionsMap := map[string][]string{}
+
+	requiredOperatorIdentities := []string{}
+	recievedOperatorIdentities := []string{}
```
cadenmarchese, line 25: Just a super nit, totally optional:
```
var (
	platformIdentitiesActionsMap map[string][]string{}
	requiredOperatorIdentities []string{}
	recievedOperatorIdentities []string{}
)
```
Since you have three next to each other I think it reads cleaner like this.

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -0,0 +1,108 @@
...
+	for _, role := range platformWorkloadIdentityRoles {
+		requiredOperatorIdentities = append(requiredOperatorIdentities, role.OperatorName)
+	}
+	for _, role := range oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
+		recievedOperatorIdentities = append(recievedOperatorIdentities, role.OperatorName)
+	}
+	ok := stringutils.ElementsMatch(requiredOperatorIdentities, recievedOperatorIdentities)
+	if !ok {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePlatformWorkloadIdentityMismatch,
```
cadenmarchese, line 34: Can we also return the required identities here? Something like `"properties.ValidatePlatformWorkloadIdentityProfile.PlatformWorkloadIdentities", "There's a mismatch between the required and expected set of platform workload identities for the requested OpenShift version '%s'. The required identities are '%v'", oc.Properties.ClusterProfile.Version, requiredOperatorIdentities)`
Or something like that.

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -0,0 +1,108 @@
...
+		if len(definition.Properties.Permissions) <= 0 {
+			return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInvalidClusterMSICount,
+				"dynamic.ValidatePlatformWorkloadIdentityProfile", "No Permissions exists for the role %s", role.RoleDefinitionID)
+		}
+
+		actions := []string{}
+		for _, action := range definition.Properties.Permissions[0].Actions {
+			actions = append(actions, *action)
+		}
+		platformIdentitiesActionsMap[role.OperatorName] = actions
```
cadenmarchese, line 58: This code is duplicated in `validateClusterMSIPermissions()` - should we make a new shared function for both?

pkg/validate/openshiftcluster_validatedynamic.go
```diff
@@ -159,10 +164,22 @@ func (dv *openShiftClusterDynamicValidator) Dynamic(ctx context.Context) error {
 		pdpClient,
 	)
 
-	// SP validation
-	err = spDynamic.ValidateServicePrincipal(ctx, spClientCred)
-	if err != nil {
-		return err
+	if dv.oc.Properties.PlatformWorkloadIdentityProfile == nil || dv.oc.Properties.ServicePrincipalProfile != nil {
```
cadenmarchese, line 167: cc @slawande2 - here's another place we want to be able to use a shared function to determine if a cluster is workload identity enabled or not.

cadenmarchese (commented):
the draft is looking good! I left a few small comments / nits on the validation logic.

pkg/validate/dynamic/platformworkloadidentityprofile.go
rajdeepc2792: Added a shared function for the same.

pkg/validate/dynamic/platformworkloadidentityprofile.go
rajdeepc2792: Now only using platformIdentitiesActionsMap, so don't need it now.

pkg/util/azureclient/azuresdk/armauthorization/roledefinitions.go
```diff
@@ -0,0 +1,26 @@
...
+	GetByID(ctx context.Context, roleID string, options *armauthorization.RoleDefinitionsClientGetByIDOptions) (armauthorization.RoleDefinitionsClientGetByIDResponse, error)
+}
+
+type roleDefinitionsClient struct {
+	armauthorization.RoleDefinitionsClient
+}
+
+var _ RoleDefinitionsClient = &roleDefinitionsClient{}
+
+func NewRoleDefinitionsClient(credential azcore.TokenCredential, options *arm.ClientOptions) (RoleDefinitionsClient, error) {
```
niontive, line 24: In general, we should return structs, not interfaces. It makes code more flexible and robust. https://duncanleung.com/go-idiom-accept-interfaces-return-types/

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
```diff
@@ -0,0 +1,49 @@
...
+	GetPlatformWorkloadIdentityRoles() []api.PlatformWorkloadIdentityRole
+}
+
+// platformWorkloadIdentityRolesByVersionService is the default implementation of the PlatformWorkloadIdentityRolesByVersion interface.
+type platformWorkloadIdentityRolesByVersionService struct {
+	platformWorkloadIdentityRoles []api.PlatformWorkloadIdentityRole
+}
+
+func NewPlatformWorkloadIdentityRolesByVersion(ctx context.Context, oc *api.OpenShiftCluster, dbPlatformWorkloadIdentityRoleSets database.PlatformWorkloadIdentityRoleSets) (PlatformWorkloadIdentityRolesByVersion, error) {
+	if oc.Properties.PlatformWorkloadIdentityProfile == nil || oc.Properties.ServicePrincipalProfile != nil {
```
niontive, line 25: Another place for shared function to check if MIWI cluster

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -0,0 +1,117 @@
...
+
+	for _, role := range oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
+		platformIdentitiesActionsMap[role.OperatorName] = nil
+	}
+
+	for _, role := range platformWorkloadIdentityRoles {
+		_, ok := platformIdentitiesActionsMap[role.OperatorName]
+		if !ok {
+			requiredOperatorIdentities := []string{}
+			for _, role := range platformWorkloadIdentityRoles {
```
kimorris27, line 31: You could rename this loop variable to something other than `role`, since it shadows the `role` variable declared for the outer loop and makes the code a little confusing to read.

Or alternatively you could do something like this:

```
roleIsMissing := false
requiredOperatorIdentities := []string{}
for _, role := range platformWorkloadIdentityRoles {
    requiredOperatorIdentities = append(requiredOperatorIdentities, role.OperatorName)
    if role not in platformIdentitiesActionsMap {
        roleIsMissing = true
    }
}

if roleIsMissing || len(platformWorkloadIdentityRoles) != len(platformIdentitiesActionsMap) {
    return the error
}
```

Note that I'm also checking the length of the cx's role set, which I think your original implementation is missing; in addition to making sure they haven't missed any, we want to make sure they aren't providing extras either.

pkg/validate/dynamic/dynamic.go
```diff
@@ -438,22 +424,28 @@ func (c closure) usingCheckAccessV2() (bool, error) {
...
-			c.dv.log.Infof("The result didn't include permission %s", action)
 			return false, nil
 		}
 	}
+	if len(actionsToFind) > 0 {
+		missingActions := []string{}
+		for missingAction := range actionsToFind {
+			missingActions = append(missingActions, missingAction)
+		}
+		c.dv.log.Infof("The result didn't include permissions %v", missingActions)
```
kimorris27, line 446: We should also log the name or object ID of the SP/identity being validated so that SREs reading the RP logs will know which identity is missing the permissions when they read this message.

pkg/validate/dynamic/dynamic.go
```diff
@@ -438,22 +424,28 @@ func (c closure) usingCheckAccessV2() (bool, error) {
...
-		if !found {
-			c.dv.log.Infof("The result didn't include permission %s", action)
 			return false, nil
 		}
 	}
+	if len(actionsToFind) > 0 {
+		missingActions := []string{}
+		for missingAction := range actionsToFind {
+			missingActions = append(missingActions, missingAction)
+		}
```
kimorris27, line 445: IMO we don't need to declare a new slice and copy over the `actionsToFind` just to effectively "rename" the slice. We could use `actionsToFind` directly in the log statement.

pkg/validate/dynamic/dynamic.go
```diff
@@ -386,12 +372,12 @@ func (dv *dynamic) validateNatGatewayPermissions(ctx context.Context, s Subnet)
 	return err
 }
 
-func (dv *dynamic) validateActions(ctx context.Context, r *azure.Resource, actions []string) error {
+func (dv *dynamic) validateActionsByOID(ctx context.Context, r *azure.Resource, actions []string, oid *string) error {
```
kimorris27, line 375: I recommend adding a doc comment on this function.

pkg/validate/dynamic/dynamic.go
```diff
@@ -780,13 +772,31 @@ func (dv *dynamic) ValidatePreConfiguredNSGs(ctx context.Context, oc *api.OpenSh
 	return nil
 }
 
+func (dv *dynamic) validateActions(ctx context.Context, r *azure.Resource, actions []string) (*string, error) {
```
kimorris27, line 775: I recommend adding a doc comment on this function, especially to make it easier for readers to understand what the string pointer return value is for.

pkg/util/azureclient/azuresdk/armauthorization/roledefinitions.go
rajdeepc2792: Throughout the codebase this is the norm to use the Interface to hide the Client function implementation that is not needed. I am happy to use the concrete struct type, but can you please point how to approach this here?

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -0,0 +1,117 @@
...
+				requiredOperatorIdentities = append(requiredOperatorIdentities, role.OperatorName)
+			}
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePlatformWorkloadIdentityMismatch,
+				"properties.ValidatePlatformWorkloadIdentityProfile.PlatformWorkloadIdentities", "There's a mismatch between the required and expected set of platform workload identities for the requested OpenShift version '%s'. The required platform workload identities are '%v'", oc.Properties.ClusterProfile.Version, requiredOperatorIdentities)
```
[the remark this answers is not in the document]
rajdeepc2792: Done.

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -0,0 +1,117 @@
...
+
+func getActionsForRoleDefinition(ctx context.Context, roleDefinitionID string, roleDefinitions armauthorization.RoleDefinitionsClient, errorStatusCode int) ([]string, error) {
+	definition, err := roleDefinitions.GetByID(ctx, roleDefinitionID, &sdkauthorization.RoleDefinitionsClientGetByIDOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(definition.RoleDefinition.Properties.Permissions) <= 0 {
+		return nil, api.NewCloudError(errorStatusCode, api.CloudErrorCodeInvalidClusterMSICount,
+			"dynamic.validateClusterMSIPermissions", "No Permissions exists for the role %s", rbac.RoleAzureRedHatOpenShiftFederatedCredentialRole)
```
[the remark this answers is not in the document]
rajdeepc2792: Logs are added for each role.

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -0,0 +1,117 @@
...
+	return nil
+}
+
+func getActionsForRoleDefinition(ctx context.Context, roleDefinitionID string, roleDefinitions armauthorization.RoleDefinitionsClient, errorStatusCode int) ([]string, error) {
+	definition, err := roleDefinitions.GetByID(ctx, roleDefinitionID, &sdkauthorization.RoleDefinitionsClientGetByIDOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(definition.RoleDefinition.Properties.Permissions) <= 0 {
```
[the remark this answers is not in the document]
rajdeepc2792: Changed the condition to `==`

pkg/validate/dynamic/platformworkloadidentityprofile.go
rajdeepc2792: Thanks for providing the snippet, used it as it is. Yes, the length check was added in the local branch, but used your provided approach.

pkg/validate/dynamic/dynamic.go
rajdeepc2792: When using `actionsToFind` directly, it gives the error as:-
`The result didn't include permissions map[Microsoft.Network/natGateways/write:{}] for object ID: billyqeillor`

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
```diff
@@ -0,0 +1,49 @@
...
+
+	for _, doc := range docs.PlatformWorkloadIdentityRoleSetDocuments {
+		if requestedInstallVersion == doc.PlatformWorkloadIdentityRoleSet.Properties.OpenShiftVersion {
+			return &platformWorkloadIdentityRolesByVersionService{
+				platformWorkloadIdentityRoles: doc.PlatformWorkloadIdentityRoleSet.Properties.PlatformWorkloadIdentityRoles,
+			}, nil
+		}
+	}
+
+	return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "properties.clusterProfile.version", "No PlatformWorkloadIdentityRoleSet found for the requested OpenShift version '%s'.", requestedInstallVersion)
```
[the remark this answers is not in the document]
rajdeepc2792: This is changed to just pointing the cluster creator to open an Azure Support Ticket.

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
```diff
@@ -0,0 +1,49 @@
...
+
+	for _, doc := range docs.PlatformWorkloadIdentityRoleSetDocuments {
+		if requestedInstallVersion == doc.PlatformWorkloadIdentityRoleSet.Properties.OpenShiftVersion {
+			return &platformWorkloadIdentityRolesByVersionService{
+				platformWorkloadIdentityRoles: doc.PlatformWorkloadIdentityRoleSet.Properties.PlatformWorkloadIdentityRoles,
+			}, nil
+		}
+	}
+
+	return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "properties.clusterProfile.version", "No PlatformWorkloadIdentityRoleSet found for the requested OpenShift version '%s'. Please retry with different OpenShift version, and if the issue persists, raise an Azure support ticket", requestedInstallVersion)
```
kimorris27, line 44:
```suggestion
	return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "properties.clusterProfile.version", "No PlatformWorkloadIdentityRoleSet found for the requested OpenShift version '%s'. Please either raise an Azure support ticket or retry with a different OpenShift version", requestedInstallVersion)
```

I'm suggesting this tweak because I don't think asking the customer to retry with a different OpenShift version provides a good UX. We could also connect with PM to see what they think about the messaging here.

pkg/validate/dynamic/platformworkloadidentityprofile.go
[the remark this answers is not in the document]
kimorris27: Where did you add the log statements? I'm having trouble finding them.

pkg/validate/dynamic/platformworkloadidentityprofile.go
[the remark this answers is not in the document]
kimorris27: I wasn't necessarily asking for a change here. I was just curious what made you choose to use `<= 0` rather than `!= 1`, which makes the most sense to me if we're assuming that the slice should contain exactly 1 element.

pkg/validate/dynamic/dynamic.go
kimorris27: It looks like you changed it to align with my suggestion though. How did you end up fixing that error after you encountered it?

kimorris27 (commented):
I made one more pass with some follow-up questions on my initial comments.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -499,5 +499,10 @@ func (sv openShiftClusterStaticValidator) validatePlatformIdentities(oc *OpenShi
 	if clusterIdentityPresent != operatorRolePresent {
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "identity", "Cluster identity and platform workload identities require each other.")
 	}
+
+	if operatorRolePresent && len(oc.Identity.UserAssignedIdentities) != 1 {
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "identity", "The provided set of user assigned identities is invalid; there should be exactly one.")
```
hlipsig, line 504: Is there any documentation link we should be returning to users here?

pkg/validate/dynamic/dynamic.go
rajdeepc2792: So the output/error-message in my last comment is the output of a unit-test.
Yes, I changed it to your suggested way. So question here is to keep it like this or revert to the earlier way?
The difference is in the error message format:-
Current:- `The result didn't include permissions map[Microsoft.Network/natGateways/write:{}] for object ID: billyqeillor`
Revert:- `The result didn't include permissions [Microsoft.Network/natGateways/write] for object ID: billyqeillor`

pkg/cluster/cluster.go
```diff
@@ -270,10 +286,11 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
...
 
-		installViaHive:                    installViaHive,
-		adoptViaHive:                      adoptByHive,
-		hiveClusterManager:                hiveClusterManager,
-		now:                               func() time.Time { return time.Now() },
-		openShiftClusterDocumentVersioner: new(openShiftClusterDocumentVersionerService),
+		installViaHive:                         installViaHive,
+		adoptViaHive:                           adoptByHive,
+		hiveClusterManager:                     hiveClusterManager,
+		now:                                    func() time.Time { return time.Now() },
```
hlipsig, line 292: Do we need Now on line 128 and here?

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
rajdeepc2792: I used the following usage of `support ticket`  when returning for other error in the backend:- https://github.com/Azure/ARO-RP/blob/master/pkg/util/steps/condition.go#L120
I can reach region leads for the better framing/detailing of this error message.

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
rajdeepc2792: Yes good idea. @hlipsig raised to add a link to sop to open a support ticket as a ref here.

pkg/validate/dynamic/platformworkloadidentityprofile.go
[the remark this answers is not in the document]
rajdeepc2792: It is here:- https://github.com/Azure/ARO-RP/pull/3619/commits/c5cdb7bdadc8e943bcb07e596d8203149551e5b3#diff-37d477b0fa9b192907f2c28202177e7551e270d8473c804b223075610b310d53R82

pkg/validate/dynamic/platformworkloadidentityprofile.go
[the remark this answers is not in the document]
rajdeepc2792: I don't have any idea why `<=` was used previously when the PR was in draft, also since there were 2 weeks of on-call in between it doesn't help to remember.
I agree that it makes no sense at all to have `<=0` and it is supposed to be `!=1`.

pkg/cluster/cluster.go
rajdeepc2792: Yes I think so, this part hasn't been touched in this PR. And now is being used here:- https://github.com/Azure/ARO-RP/blob/master/pkg/cluster/install.go#L415

pkg/validate/dynamic/dynamic.go
kimorris27: Oh... I missed that `actionToFind` is a map and `missingActions` was a slice. I originally thought one slice was just being copied into another slice here.

Now that I properly get what's going on, I'd say to revert this change so we have the nicer error message format.

pkg/validate/dynamic/platformworkloadidentityprofile.go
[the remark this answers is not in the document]
kimorris27: Looks good then!

pkg/validate/dynamic/platformworkloadidentityprofile.go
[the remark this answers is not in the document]
kimorris27: > also since there were 2 weeks of on-call in between it doesn't help to remember

:laughing::laughing::laughing:

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
```diff
@@ -0,0 +1,49 @@
...
+
+	requestedInstallVersion := oc.Properties.ClusterProfile.Version
+
+	docs, err := dbPlatformWorkloadIdentityRoleSets.ListAll(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, doc := range docs.PlatformWorkloadIdentityRoleSetDocuments {
+		if requestedInstallVersion == doc.PlatformWorkloadIdentityRoleSet.Properties.OpenShiftVersion {
```
rajdeepc2792, line 37: Don't use the patch version to locate the rolesets entry for the version, as the rolesets are defined in db as per minor version. For Eg:- 4.14, 4.15 etc.

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, err
+	}
+
+	if oc.Properties.PlatformWorkloadIdentityProfile.UpgradeableTo != nil {
+		upgradeableVersion, err := version.ParseVersion(string(*oc.Properties.PlatformWorkloadIdentityProfile.UpgradeableTo))
+		if err != nil {
+			return nil, err
+		}
+		upgradeableMinorVersion := upgradeableVersion.MinorVersion()
+		if currentMinorVersion != upgradeableMinorVersion && !currentOpenShiftVersion.Lt(upgradeableVersion) {
```
cadenmarchese, line 50: Do we want to handle cases where `currentOpenShiftVersion.Lt(upgradeableVersion)` or just ignore it?

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
```diff
@@ -0,0 +1,81 @@
...
+	currentMinorVersion := currentOpenShiftVersion.MinorVersion()
+	requiredMinorVersions := map[string]bool{currentMinorVersion: false}
+	platformWorkloadIdentityRoles := []api.PlatformWorkloadIdentityRole{}
+
+	docs, err := dbPlatformWorkloadIdentityRoleSets.ListAll(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if oc.Properties.PlatformWorkloadIdentityProfile.UpgradeableTo != nil {
```
cadenmarchese, line 44: IMO, this whole file could benefit from some code comments for reviewers and readability in general. I can mostly follow the logic but for others who might not be aware of the assumptions we are making around upgradeableTo and what specifically we are validating I think it would help!

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"testing"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/database"
+	testdatabase "github.com/Azure/ARO-RP/test/database"
+	"github.com/Azure/ARO-RP/test/util/deterministicuuid"
+	utilerror "github.com/Azure/ARO-RP/test/util/error"
+)
+
+const fakeClusterSPObjectId = "00000000-0000-0000-0000-000000000000"
```
cadenmarchese, line 17: nit - IMO the const isn't really needed, it's only used once.

pkg/validate/openshiftcluster_validatedynamic.go
cadenmarchese: Should we move over to the shared function here?

cadenmarchese (commented):
great work! I only have small comments and questions.

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
rajdeepc2792: We can ignore it, as it will mean that the cluster is upgraded and will be ignored as part of the validation.

pkg/validate/openshiftcluster_validatedynamic.go
```diff
@@ -112,51 +120,67 @@ func (dv *openShiftClusterDynamicValidator) Dynamic(ctx context.Context) error {
...
+	} else {
+		// PlatformWorkloadIdentity and ClusterMSIIdentity Validation
+		spDynamic = dynamic.NewValidator(
+			dv.log,
+			dv.env,
+			dv.env.Environment(),
+			dv.subscriptionDoc.ID,
+			nil,
+			nil,
+			dynamic.AuthorizerClusterServicePrincipal,
```
kimorris27, line 176: This authorizer type gets included in error messages, so if we use the existing `AuthorizerClusterServicePrincipal`, we're going to print some misleading stuff in error messages about a "service principal" that's missing permissions. I think it might be a good idea to define a new `AuthorizerType` for managed identities and pass it in here. Your thoughts?

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -0,0 +1,130 @@
...
+		if oc.Properties.PlatformWorkloadIdentityProfile.UpgradeableTo != nil {
+			upgradeableVersion, err := version.ParseVersion(string(*oc.Properties.PlatformWorkloadIdentityProfile.UpgradeableTo))
+			if err != nil {
+				return err
+			}
+			if !upgradeableVersion.Eq(currentOpenShiftVersion) && !currentOpenShiftVersion.Lt(upgradeableVersion) {
+				v = fmt.Sprintf("%s or %s", v, string(*oc.Properties.PlatformWorkloadIdentityProfile.UpgradeableTo))
+			}
+		}
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePlatformWorkloadIdentityMismatch,
```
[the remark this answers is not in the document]
rajdeepc2792: So the failure here can be because of the platform identity validation failures. Those missing platform identities can be of either the current or upgradeableTo version.

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -0,0 +1,130 @@
...
+	for _, pwi := range oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
+		_, ok := platformWorkloadIdentityRolesByRoleName[pwi.OperatorName]
+		if ok {
+			dv.platformIdentitiesActionsMap[pwi.OperatorName] = nil
+			dv.platformIdentities = append(dv.platformIdentities, pwi)
+		}
+	}
+
+	// Check if any required platform identity is missing
+	if len(dv.platformIdentities) != len(platformWorkloadIdentityRolesByRoleName) {
```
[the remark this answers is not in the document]
rajdeepc2792: Yes, because the customer can bring workload identities for their own workloads.

pkg/validate/dynamic/platformworkloadidentityprofile.go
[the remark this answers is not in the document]
kimorris27: Okay, I see. So the assumption is that the cx should include `UpgradeableTo` and any new platform identities for that new OCP version in the same `PUT`/`PATCH` request?

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
tsatam: We might have an opportunity for reusable logic around "what OCP version do I need to handle upgradeability to" across multiple steps in our cluster update process. I think we might benefit from extracting this logic here into a common function and setting up these kinds of checks so we have a single consistent version:
- UpgradeableTo not specified (return the current version)
- UpgradeableTo specified but less than current version (return the current version)
- UpgradeableTo specified and greater than current version (return the UpgradeableTo value)

pkg/validate/dynamic/platformworkloadidentityprofile.go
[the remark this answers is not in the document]
rajdeepc2792: the cx should include UpgradeableTo only when all the new platform identities for that new OCP version are present in the cluster doc either via previous or the same PUT/PATCH request?

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
```diff
@@ -0,0 +1,81 @@
...
+			if version == doc.PlatformWorkloadIdentityRoleSet.Properties.OpenShiftVersion {
+				platformWorkloadIdentityRoles = append(platformWorkloadIdentityRoles, doc.PlatformWorkloadIdentityRoleSet.Properties.PlatformWorkloadIdentityRoles...)
+				requiredMinorVersions[version] = true
+			}
+		}
+	}
+
+	for version, exists := range requiredMinorVersions {
+		if !exists {
+			return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", "No PlatformWorkloadIdentityRoleSet found for the requested or upgradeable OpenShift minor version '%s'. Please retry with different OpenShift version, and if the issue persists, raise an Azure support ticket", version)
```
tsatam, line 66: I think we'll need a more clarifying error message here, as the message here as-is might be confusing to users. Is this returned when an invoking user does not provide a required identity for a given OCP x.y.z version but passes that version in as UpgradeableTo?

pkg/validate/dynamic/dynamic.go
```diff
@@ -25,28 +25,34 @@ import (
...
+	errMsgSPHasMissingReaderRole               = "The %s service principal (Application ID: %s) does not have Reader permission"
+	errMsgSPHasMissingNetworkContributorRole   = "The %s service principal (Application ID: %s) does not have Network Contributor role"
+	errMsgPlatformIdentityHasMissingPermission = "The %s platform managed identity does not have required permissions"
+	errMsgNSGAttached                          = "The provided subnet '%s' is invalid: must not have a network security group attached."
+	errMsgOriginalNSGNotAttached               = "The provided subnet '%s' is invalid: must have network security group '%s' attached."
+	errMsgNSGNotAttached                       = "The provided subnet '%s' is invalid: must have a network security group attached."
+	errMsgNSGNotProperlyAttached               = "When the enable-preconfigured-nsg option is specified, both the master and worker subnets should have network security groups (NSG) attached to them before starting the cluster installation."
+	errMsgMissingPermissionsOnNSG              = "on network security group '%s'. This is required when the enable-preconfigured-nsg option is specified."
```
tsatam, line 43: This error message was truncated, it looks like it was a mistake? In my opinion we should just maintain separate error messages for "CSP missing required permissions" vs "managed identity missing required permissions", or just maintain a generic "the identity 'foo' is missing required permissions" for both cases.

tsatam (dismissed):

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
rajdeepc2792: This logic works for getting the PlatformWorkloadIdentityRoleSet for the OCP minor version stated in clusterprofile or UpgradeableTo. It has nothing to do if required identities are provided or not.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
rajdeepc2792: Done!! Thanks.

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
rajdeepc2792: Used minor version everywhere.

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
rajdeepc2792: Added a comment to the top of the function.

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -0,0 +1,130 @@
...
+			return err
+		}
+
+		return dv.validateClusterMSIPermissions(ctx, identity.PrincipalID, oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities, roleDefinitions)
+	}
+
+	return nil
+}
+
+// Validate all Platform Identities such that Cluster MSI can perform actions present in role AzureRedHatOpenShiftFederatedCredentialRole
```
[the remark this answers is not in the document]
rajdeepc2792: Done. Thanks!!

pkg/validate/dynamic/platformworkloadidentityprofile_test.go
```diff
@@ -0,0 +1,518 @@
...
+				mockTokenCredential(tokenCred)
+				msiAuthReq := createAuthorizationRequest(dummyObjectId, platformIdentity1, msiRequiredPermissionsList...)
+				pdpClient.EXPECT().CheckAccess(gomock.Any(), msiAuthReq).Do(func(arg0, arg1 interface{}) {
+					cancel()
+				}).Return(&msiNotAllowedActions, errors.New("Generic Error")).AnyTimes()
+			},
+			wantErr: "Generic Error",
+		},
+		{
+			name:                  "Fail - An action is missing for a platform identity",
```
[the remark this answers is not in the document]
rajdeepc2792: Updated the handling of these test cases, previously they were incorrectly returning the Generic Error.

pkg/validate/openshiftcluster_validatedynamic.go
rajdeepc2792: Added a new `AuthorizerType`.

pkg/validate/dynamic/dynamic.go
rajdeepc2792: Working on this now.

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, err
+	}
+
+	if oc.Properties.PlatformWorkloadIdentityProfile.UpgradeableTo != nil {
+		upgradeableVersion, err := version.ParseVersion(string(*oc.Properties.PlatformWorkloadIdentityProfile.UpgradeableTo))
+		if err != nil {
+			return nil, err
+		}
+		upgradeableMinorVersion := upgradeableVersion.MinorVersion()
+		if currentMinorVersion != upgradeableMinorVersion && currentOpenShiftVersion.Lt(upgradeableVersion) {
```
kimorris27, line 50: This can be simplified; if we know that the current version is less than the upgrade version, we know they aren't equal:

```suggestion
		if currentOpenShiftVersion.Lt(upgradeableVersion) {
```

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -0,0 +1,130 @@
...
+		currentOpenShiftVersion, err := version.ParseVersion(oc.Properties.ClusterProfile.Version)
+		if err != nil {
+			return err
+		}
+		if oc.Properties.PlatformWorkloadIdentityProfile.UpgradeableTo != nil {
+			upgradeableVersion, err := version.ParseVersion(string(*oc.Properties.PlatformWorkloadIdentityProfile.UpgradeableTo))
+			if err != nil {
+				return err
+			}
+			if !upgradeableVersion.Eq(currentOpenShiftVersion) && currentOpenShiftVersion.Lt(upgradeableVersion) {
```
kimorris27, line 50: This can be simplified; if we know that the current version is less than the upgrade version, we know they aren't equal:

```suggestion
			if currentOpenShiftVersion.Lt(upgradeableVersion) {
```

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion_test.go
```diff
@@ -62,6 +63,123 @@ func TestNewPlatformWorkloadIdentityRolesByVersion(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				)
+			},
+			wantPlatformIdentities: []string{"Dummy1", "Dummy2"},
+		},
+		{
+			name: "Success - The role set document found for the cluster and upgradeable version(with lesser than current version)",
```
kimorris27, line 145: I'm confused why we need this case. I thought that if the upgrade version is less than the current version then we skip getting the upgrade version's role set?

kimorris27 (changes requested):

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion_test.go
rajdeepc2792: Yes, that's what the test is doing here, it is not considering the upgradeable version's identities to be the desired one.

kimorris27: Oh, then it's the description of the test case that threw me off. I think this would help clarify what this case is for:

```suggestion
			name: "Success - Role set document found for cluster version; UpgradeableTo version ignored because it is less than cluster version",
```

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
rajdeepc2792: So the check here is to make sure that the Minor version is also not same. For eg, if your suggested change is performed, it will be recurring to consider upgradeableVersion for validation upgradeableVersion = 4.14.60 and currentVersion = 4.14.50 .

pkg/validate/dynamic/platformworkloadidentityprofile.go
rajdeepc2792: So the check here is to make sure that the Minor version is also not same. For eg, if your suggested change is performed, it will be recurring to consider upgradeableVersion for validation upgradeableVersion = 4.14.60 and currentVersion = 4.14.50 .
Here the equality check is incorrect, it should be done between the minor versions. Performing the change here.

rajdeepc2792: Done, updated the comparison and also added unit test cases for it.

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion_test.go
rajdeepc2792: Done, changed the name!

pkg/validate/dynamic/dynamic.go
rajdeepc2792: Separated error messaging based on the authorizerType.

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, nil
+	}
+
+	currentOpenShiftVersion, err := version.ParseVersion(oc.Properties.ClusterProfile.Version)
+	if err != nil {
+		return nil, err
+	}
+	currentMinorVersion := currentOpenShiftVersion.MinorVersion()
+	requiredMinorVersions := map[string]bool{currentMinorVersion: false}
+	platformWorkloadIdentityRoles := []api.PlatformWorkloadIdentityRole{}
```
SudoBrendan, line 37: It appears that OCP versions are in some way important here. Can we update the Jira to include acceptance criteria to showcase this as a requirement? It's unclear now why this is important.

SudoBrendan (commented):
I'm flying a bit blind on this, but did my best :sweat_smile:

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
```diff
@@ -0,0 +1,81 @@
...
+	GetPlatformWorkloadIdentityRolesByRoleName() map[string]api.PlatformWorkloadIdentityRole
+}
+
+// platformWorkloadIdentityRolesByVersionService is the default implementation of the PlatformWorkloadIdentityRolesByVersion interface.
+type platformWorkloadIdentityRolesByVersionService struct {
+	platformWorkloadIdentityRoles []api.PlatformWorkloadIdentityRole
+}
+
+// NewPlatformWorkloadIdentityRolesByVersion aims to populate platformWorkloadIdentityRoles for current OpenShift minor version and also for UpgradeableTo minor version if provided and is greater than the current version
+func NewPlatformWorkloadIdentityRolesByVersion(ctx context.Context, oc *api.OpenShiftCluster, dbPlatformWorkloadIdentityRoleSets database.PlatformWorkloadIdentityRoleSets) (PlatformWorkloadIdentityRolesByVersion, error) {
```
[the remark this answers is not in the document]
rajdeepc2792: Updated the implementation of NewPlatformWorkloadIdentityRolesByVersion and RoleDefinitionsClient as asked and suggested.
Thanks for the suggestion.
The only change that wasn't performed was:-

> 4. in our case, we mockgen the interface and use the mocked implementation in the tests (the "hard way" would be to create a second implementation of the interface manually in the test suite that does whatever functionality we want)

Can you please suggest, if this is needed as for now the implemented unit test cases worked as desired. Also the goto way for implementing test cases for ARM clients using fakes:- https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/storage/armstorage/fake_example_test.go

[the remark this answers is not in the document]
rajdeepc2792: This is done, separated the main logic to another function.

pkg/util/azureclient/azuresdk/armauthorization/generate.go
```diff
@@ -0,0 +1,8 @@
+package armauthorization
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+//go:generate rm -rf ../../../../util/mocks/$GOPACKAGE
+//go:generate go run ../../../../../vendor/github.com/golang/mock/mockgen -destination=../../../../util/mocks/azureclient/azuresdk/$GOPACKAGE/$GOPACKAGE.go github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/$GOPACKAGE RoleDefinitionsClient
```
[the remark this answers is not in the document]
rajdeepc2792: Will wait for either PR to get merged to see which one needs attenction.

SudoBrendan (commented):
Again, I can only really speak to golang style here - this PR is huge and a lot of this is unfamiliar territory since I'm not close to MIWI. Here's a few more comments - I'm relying on people closer to the feature to review and say if this is meeting acceptance criteria or not.

pkg/util/platformworkloadidentity/platformworkloadidentityrolesbyversion.go
```diff
@@ -0,0 +1,87 @@
...
+	GetPlatformWorkloadIdentityRolesByRoleName() map[string]api.PlatformWorkloadIdentityRole
+	PopulatePlatformWorkloadIdentityRolesByVersion(ctx context.Context, oc *api.OpenShiftCluster, dbPlatformWorkloadIdentityRoleSets database.PlatformWorkloadIdentityRoleSets) error
+}
+
+// platformWorkloadIdentityRolesByVersionService is the default implementation of the PlatformWorkloadIdentityRolesByVersion interface.
+type PlatformWorkloadIdentityRolesByVersionService struct {
+	platformWorkloadIdentityRoles []api.PlatformWorkloadIdentityRole
+}
+
+var _ PlatformWorkloadIdentityRolesByVersion = &PlatformWorkloadIdentityRolesByVersionService{}
```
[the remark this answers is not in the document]
rajdeepc2792: Makes sense, it shouldn't be needed.

pkg/util/stringutils/stringutils.go
```diff
@@ -20,3 +20,16 @@ func Contains(list []string, value string) bool {
...
+	for _, a := range as {
+		for _, b := range bs {
+			if a == b {
+				gs = append(gs, a)
+				break
+			}
+		}
+	}
+
+	return gs
```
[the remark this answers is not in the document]
rajdeepc2792: Added unit test cases.

pkg/validate/dynamic/dynamic.go
```diff
@@ -438,22 +481,24 @@ func (c closure) usingCheckAccessV2() (bool, error) {
...
+		_, ok := actionsToFind[result.ActionId]
+		if ok {
+			delete(actionsToFind, result.ActionId)
+			if result.AccessDecision == remotepdp.Allowed {
+				continue
 			}
-		}
-		if !found {
-			c.dv.log.Infof("The result didn't include permission %s", action)
 			return false, nil
```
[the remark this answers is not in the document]
rajdeepc2792: Made the changes as per suggestion.

pkg/util/azureclient/azuresdk/armauthorization/generate.go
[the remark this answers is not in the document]
SudoBrendan: .bingo has merged - this PR needs a rebase, and this is now a requirement to fix here

cmd/aro/rp.go
```diff
@@ -175,7 +175,7 @@ func rp(ctx context.Context, log, audit *logrus.Entry) error {
 		return err
 	}
 
-	b, err := backend.NewBackend(ctx, log.WithField("component", "backend"), _env, dbAsyncOperations, dbBilling, dbGateway, dbOpenShiftClusters, dbSubscriptions, dbOpenShiftVersions, aead, metrics)
+	b, err := backend.NewBackend(ctx, log.WithField("component", "backend"), _env, dbAsyncOperations, dbBilling, dbGateway, dbOpenShiftClusters, dbSubscriptions, dbOpenShiftVersions, dbPlatformWorkloadIdentityRoleSets, aead, metrics)
```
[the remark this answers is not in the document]
rajdeepc2792: Added it as a note for dbPlatformWorkloadIdentityRoleSets.

pkg/util/azureclient/azuresdk/armauthorization/generate.go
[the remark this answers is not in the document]
rajdeepc2792: Done!!

pkg/cluster/cluster.go
```diff
@@ -233,6 +238,19 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 		return nil, err
 	}
 
+	armRoleDefinitionsClient, err := armauthorization.NewArmRoleDefinitionsClient(fpCredClusterTenant, &clientOptions)
```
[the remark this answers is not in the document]
rajdeepc2792: Tested it following the steps:-

1. Created an Application/Service Principal `test`
2. Assigned `Network Contributor` to `test` on a network resource group
3. Performed:-
```curl --location 'https://management.azure.com/providers/Microsoft.Authorization/roleDefinitions/0336e1d3-7a87-462b-b6db-342b63f7802c?api-version=2022-04-01' \
--header 'Authorization: Bearer <REDACTED>'```

The above request returned the Role Definition successfully. 0336e1d3-7a87-462b-b6db-342b63f7802c is Azure Red Hat OpenShift Cluster Ingress Operator Role

Created JIRA, to have the backup ready to handle the case if this fails.

SudoBrendan (approved):
I think all my comments have been addressed :)

cadenmarchese (approved):

kimorris27 (approved):

PR #3623, opened 2024-06-07
Architecting Local `make ci-rp` File To Allow Caching For Later

safwank97:

Fixes https://issues.redhat.com/browse/ARO-7789

This PR helps us build intermediate docker images individually by running `make ci-rp` in ephemeral CI agents without a docker cache, such that, it has the capability of building a docker cache from it's previous run build stage. This will help us optimize our build time in the ephemeral agent.

### Test plan for issue:

Run `make ci-rp` to verify and validate whether it is building all the intermediate docker images individually from a multi-staged container file `Dockerfile.ci-rp`.

None that I'm aware of.


Makefile
```diff
@@ -77,7 +79,9 @@ client: generate
 	hack/build-client.sh "${AUTOREST_IMAGE}" 2020-04-30 2021-09-01-preview 2022-04-01 2022-09-04 2023-04-01 2023-07-01-preview 2023-09-04 2023-11-22 2024-08-12-preview
 
 ci-rp: fix-macos-vendor
-	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg ARO_VERSION=$(VERSION) --no-cache=$(NO_CACHE)
+	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg ARO_VERSION=$(VERSION) --target portal-build --no-cache=$(NO_CACHE) -t $(ARO_PORTAL_BUILD_IMAGE)
+	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg ARO_VERSION=$(VERSION) --target builder --cache-from portal-build -t $(ARO_BUILDER_IMAGE)
+	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg ARO_VERSION=$(VERSION) --cache-from builder -t $(ARO_IMAGE)
```
bitoku, line 84: same as above

Makefile
```diff
@@ -77,7 +79,9 @@ client: generate
 	hack/build-client.sh "${AUTOREST_IMAGE}" 2020-04-30 2021-09-01-preview 2022-04-01 2022-09-04 2023-04-01 2023-07-01-preview 2023-09-04 2023-11-22 2024-08-12-preview
 
 ci-rp: fix-macos-vendor
-	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg ARO_VERSION=$(VERSION) --no-cache=$(NO_CACHE)
+	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg ARO_VERSION=$(VERSION) --target portal-build --no-cache=$(NO_CACHE) -t $(ARO_PORTAL_BUILD_IMAGE)
+	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg ARO_VERSION=$(VERSION) --target builder --cache-from portal-build -t $(ARO_BUILDER_IMAGE)
```
bitoku, line 83: https://docs.docker.com/reference/cli/docker/image/build/#cache-from
https://docs.podman.io/en/stable/markdown/podman-build.1.html
According to the doc, `--cache-from` should point to an image name.
I'm not sure if it works with the stage name (maybe not?).

And also it wants to have `--no-cache` here, otherwise it can use the cache if the previous build has the same result.

```suggestion
	docker build . -f Dockerfile.ci-rp --ulimit=nofile=4096:4096 --build-arg REGISTRY=$(REGISTRY) --build-arg ARO_VERSION=$(VERSION) --target builder --no-cache=$(NO_CACHE) --cache-from $(ARO_PORTAL_BUILD_IMAGE) -t $(ARO_BUILDER_IMAGE)
```

bitoku: I was wrong about `--no-cache`. Please forget about it.

safwank97: Okay no worries.

safwank97: Updated --cache-from to point to an image name

geowa4:
Hey @safwank97, can you include your testing strategy on this? Or is that in the ticket? I just want to ensure the changes in response to @bitoku's feedback are sufficient.

safwank97:
> Hey @safwank97, can you include your testing strategy on this? Or is that in the ticket? I just want to ensure the changes in response to @bitoku's feedback are sufficient.

There are not any direct ways to test whether a particular docker image is built using `--cache-from` flag other than pulling above code changes and running `make ci-rp` directly, because this uses variables defined in `env.example` and one has to make canges to their `env` file as specified in `env.example` file and then also source that particular file to run the above specified builts. And this will explain the changes in response to @bitoku's feedback because in my previous it used a tag and in this it is properly pointing to an image name instead of a tag.

Makefile
```diff
@@ -38,6 +38,8 @@ else
 	REGISTRY = $(RP_IMAGE_ACR)
 endif
 
+ARO_PORTAL_BUILD_IMAGE ?= $(ARO_IMAGE_BASE):$(VERSION)
+ARO_BUILDER_IMAGE ?= $(ARO_IMAGE_BASE):$(VERSION)
```
SudoBrendan, line 42: These images are currently the same, and are the same as ARO_IMAGE as well below. We likely need all 3 to be unique. I like the naming convention used in the env file, can we replicate those names here?

SudoBrendan (changes requested):

Makefile
tsatam: I'd suggest explicitly pointing these to localhost images (similar to what you did in env.example below):

```suggestion
ARO_PORTAL_BUILD_IMAGE ?= localhost/aro-portal-build:$(VERSION)
ARO_BUILDER_IMAGE ?= localhost/aro-builder:$(VERSION)
```

env.example
```diff
@@ -1,4 +1,6 @@
 export LOCATION=eastus
+export ARO_PORTAL_BUILD_IMAGE=arointsvc.azurecr.io/aro-portal-build
+export ARO_BUILDER_IMAGE=arointsvc.azurecr.io/aro-builder
```
tsatam, line 3: If the change made above is done, I think this can be removed. 

```suggestion
```

tsatam (changes requested):
This looks good, just a few suggestions to keep the implementation here simple.

Makefile
```diff
@@ -2,7 +2,7 @@ SHELL = /bin/bash
 TAG ?= $(shell git describe --exact-match 2>/dev/null)
 COMMIT = $(shell git rev-parse --short=7 HEAD)$(shell [[ $$(git status --porcelain) = "" ]] || echo -dirty)
 ARO_IMAGE_BASE = ${RP_IMAGE_ACR}.azurecr.io/aro
-INTERMEDIATE_IMAGE_LOCAL = ${INTERMEDIATE_IMAGE_ACR_LOCAL}.azurecr.io/aro
+RP_IMAGE_LOCAL = localhost/aro
```
tsatam, line 5: Can we move this definition to right before the portal/builder images so the context is more apparent? 

Additionally, should we set this variable only if it is already unset? e.g. 
```suggestion
RP_IMAGE_LOCAL ?= localhost/aro
```

tsatam (changes requested):

Makefile
safwank97: We can, I'll update it

Makefile
safwank97: Done. Not added version for building local images because --cache-from functionality neither uses tags nor it uses digests.

Makefile
safwank97: Done 👍🏼

tsatam (approved):

shubhadapaithankar (approved):
LGTM

PR #3624, opened 2024-06-10
WI/MI CLI Phase 1 - Base API functionality

tsatam:

Fixes [ARO-6453](https://issues.redhat.com/browse/ARO-6453)
Fixes [ARO-4398](https://issues.redhat.com/browse/ARO-4398)

Implements functionality into the `az aro` extension to allow creating ARO clusters with a set of managed identities, instead of a service principal. 

As a consequence, this PR also includes changes to update the version of the `az aro` extension, as well as changes to upgrade the used ARO SDK/API to v20240812preview. 

### Test plan for issue:

- [X] Unit tests have been added for validation logic implemented across the three flags: `--enable-managed-identity, --assign-platform-workload-identity, --mi-user-assigned`
- [x] `az aro create` with the MIWI flags results in an API call being made to the RP with the neccessary fields populated
  - Example invocation:
    ```sh
    az aro create \
      --subscription ${SUBSCRIPTION_ID} \
      --resource-group ${CLUSTER_RESOURCE_GROUP} \
      --name azcli-testcluster \
      --version 4.14.16 
      --vnet aro-vnet --master-subnet master-subnet --worker-subnet worker-subnet \ 
      --enable-managed-identity \
      --assign-platform-workload-identity foo asdf \
      --assign-platform-workload-identity bar ghjkl \
      --mi-user-assigned baz
    ```
  - Produces request body:
    <details>
      <summary>Open</summary>
      
      ```json
      {
          "id": "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${CLUSTER_RESOURCE_GROUP}/providers/Microsoft.RedHatOpenShift/openShiftClusters/azcli-testcluster",
          "location": "eastus",
          "identity": {
              "type": "UserAssigned",
              "userAssignedIdentities": {
                  "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${CLUSTER_RESOURCE_GROUP}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/baz": {}
              }
          },
          "properties": {
              "clusterProfile": {
                  "pullSecret": "",
                  "domain": "fv0w9cvw",
                  "version": "4.14.16",
                  "resourceGroupId": "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/aro-fv0w9cvw",
                  "fipsValidatedModules": "Disabled"
              },
              "platformWorkloadIdentityProfile": {
                  "platformWorkloadIdentities": [
                      {
                          "operatorName": "foo",
                          "resourceId": "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${CLUSTER_RESOURCE_GROUP}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/asdf"
                      },
                      {
                          "operatorName": "bar",
                          "resourceId": "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${CLUSTER_RESOURCE_GROUP}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/ghjkl"
                      }
                  ]
              },
              "networkProfile": {
                  "podCidr": "10.128.0.0/14",
                  "serviceCidr": "172.30.0.0/16",
                  "outboundType": "",
                  "preconfiguredNSG": "Disabled"
              },
              "masterProfile": {
                  "vmSize": "Standard_D8s_v3",
                  "subnetId": "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${CLUSTER_RESOURCE_GROUP}/providers/Microsoft.Network/virtualNetworks/aro-vnet/subnets/master-subnet",
                  "encryptionAtHost": "Disabled"
              },
              "workerProfiles": [
                  {
                      "name": "worker",
                      "vmSize": "Standard_D2s_v3",
                      "diskSizeGB": 128,
                      "subnetId": "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${CLUSTER_RESOURCE_GROUP}/providers/Microsoft.Network/virtualNetworks/aro-vnet/subnets/worker-subnet",
                      "count": 3,
                      "encryptionAtHost": "Disabled"
                  }
              ],
              "apiserverProfile": {
                  "visibility": "Public"
              },
              "ingressProfiles": [
                  {
                      "name": "default",
                      "visibility": "Public"
                  }
              ]
          }
      }
      ```
    </details>

Not yet

Preview extension-only change. We will perform comprehensive testing before releasing this extension to users. 

### Additional notes
This PR is opened against ARO-RP's master branch. We may desire to isolate MIWI CLI changes to a dedicated branch, in order to afford us the ability to perform changes to our extension code pointing against the "stable" (v20231122) API, or to another future in-development API, but as we do not currently have any plans for parallel API efforts, I've chosen to merge these changes into our trunk branch.

### DIFF
--- a/python/az/aro/azext_aro/_actions.py
+++ b/python/az/aro/azext_aro/_actions.py
@@ -1,0 +1,29 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the Apache License 2.0.
+
+import argparse
+
+from azext_aro.vendored_sdks.azure.mgmt.redhatopenshift.v2024_08_12_preview.models import PlatformWorkloadIdentity
+from azure.cli.core.azclierror import CLIError
+
+
+# pylint:disable=protected-access
+# pylint:disable=too-few-public-methods
+class AROPlatformWorkloadIdentityAddAction(argparse._AppendAction):
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        try:
+            if len(values) != 2:
+                msg = f"{option_string} requires 2 values in format: `OPERATOR_NAME RESOURCE_ID`"
+                raise argparse.ArgumentError(self, msg)
+
+            operator_name, resource_id = values
+            parsed = PlatformWorkloadIdentity(
+                operator_name=operator_name,
+                resource_id=resource_id
+            )
+
+            super().__call__(parser, namespace, parsed, option_string)
+
+        except ValueError as e:
+            raise CLIError(f"usage error: {option_string} NAME ID") from e
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -21,6 +21,7 @@ from azext_aro._validators import validate_outbound_type
 from azext_aro._validators import validate_load_balancer_managed_outbound_ip_count
 from azext_aro._validators import validate_enable_managed_identity
 from azext_aro._validators import validate_platform_workload_identities
+from azext_aro._validators import validate_cluster_identity
 from azure.cli.core.commands.parameters import name_type
 from azure.cli.core.commands.parameters import get_enum_type, get_three_state_flag
 from azure.cli.core.commands.parameters import resource_group_name_type
@@ -135,6 +136,10 @@ def load_arguments(self, _):
                    options_list=['--assign-platform-workload-identity', '--assign-platform-wi'],
                    validator=validate_platform_workload_identities,
                    action=AROPlatformWorkloadIdentityAddAction, nargs='+')
+        c.argument('mi_user_assigned', arg_group='Identity',
+                   options_list=['--mi-user-assigned', '--assign-cluster-identity'],
+                   validator=validate_cluster_identity,
+                   help='Set the user managed identity on the cluster.')
 
     with self.argument_context('aro update') as c:
         c.argument('client_secret',
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -18,6 +18,7 @@ from azext_aro._validators import validate_refresh_cluster_credentials
 from azext_aro._validators import validate_version_format
 from azext_aro._validators import validate_outbound_type
 from azext_aro._validators import validate_load_balancer_managed_outbound_ip_count
+from azext_aro._validators import validate_enable_managed_identity
 from azure.cli.core.commands.parameters import name_type
 from azure.cli.core.commands.parameters import get_enum_type, get_three_state_flag
 from azure.cli.core.commands.parameters import resource_group_name_type
@@ -122,6 +123,11 @@ def load_arguments(self, _):
                    help='The desired number of IPv4 outbound IPs created and managed by Azure for the cluster public load balancer.',  # pylint: disable=line-too-long
                    validator=validate_load_balancer_managed_outbound_ip_count,
                    options_list=['--load-balancer-managed-outbound-ip-count', '--lb-ip-count'])
+
+        c.argument('enable_managed_identity', arg_group='Identity', arg_type=get_three_state_flag(),
+                   options_list=['--enable-managed-identity', '--enable-mi'],
+                   validator=validate_enable_managed_identity,
+                   help='Enable managed identity for this cluster.', is_preview=True)
 
     with self.argument_context('aro update') as c:
         c.argument('client_secret',
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the Apache License 2.0.
 
+from azext_aro._actions import AROPlatformWorkloadIdentityAddAction
 from azext_aro._validators import validate_cidr
 from azext_aro._validators import validate_client_id
 from azext_aro._validators import validate_client_secret
@@ -19,6 +20,7 @@ from azext_aro._validators import validate_version_format
 from azext_aro._validators import validate_outbound_type
 from azext_aro._validators import validate_load_balancer_managed_outbound_ip_count
 from azext_aro._validators import validate_enable_managed_identity
+from azext_aro._validators import validate_platform_workload_identities
 from azure.cli.core.commands.parameters import name_type
 from azure.cli.core.commands.parameters import get_enum_type, get_three_state_flag
 from azure.cli.core.commands.parameters import resource_group_name_type
@@ -128,6 +130,11 @@ def load_arguments(self, _):
                    options_list=['--enable-managed-identity', '--enable-mi'],
                    validator=validate_enable_managed_identity,
                    help='Enable managed identity for this cluster.', is_preview=True)
+        c.argument('platform_workload_identities', arg_group='Identity',
+                   help='Assign a platform workload identity used within the cluster', is_preview=True,
+                   options_list=['--assign-platform-workload-identity', '--assign-platform-wi'],
+                   validator=validate_platform_workload_identities,
+                   action=AROPlatformWorkloadIdentityAddAction, nargs='+')
 
     with self.argument_context('aro update') as c:
         c.argument('client_secret',
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -2,30 +2,35 @@
 # Licensed under the Apache License 2.0.
 
 from azext_aro._actions import AROPlatformWorkloadIdentityAddAction
-from azext_aro._validators import validate_cidr
-from azext_aro._validators import validate_client_id
-from azext_aro._validators import validate_client_secret
-from azext_aro._validators import validate_cluster_resource_group
-from azext_aro._validators import validate_disk_encryption_set
-from azext_aro._validators import validate_domain
-from azext_aro._validators import validate_pull_secret
-from azext_aro._validators import validate_subnet
-from azext_aro._validators import validate_visibility
-from azext_aro._validators import validate_vnet
-from azext_aro._validators import validate_vnet_resource_group_name
-from azext_aro._validators import validate_worker_count
-from azext_aro._validators import validate_worker_vm_disk_size_gb
-from azext_aro._validators import validate_refresh_cluster_credentials
-from azext_aro._validators import validate_version_format
-from azext_aro._validators import validate_outbound_type
-from azext_aro._validators import validate_load_balancer_managed_outbound_ip_count
-from azext_aro._validators import validate_enable_managed_identity
-from azext_aro._validators import validate_platform_workload_identities
-from azext_aro._validators import validate_cluster_identity
-from azure.cli.core.commands.parameters import name_type
-from azure.cli.core.commands.parameters import get_enum_type, get_three_state_flag
-from azure.cli.core.commands.parameters import resource_group_name_type
-from azure.cli.core.commands.parameters import tags_type
+from azext_aro._validators import (
+    validate_cidr,
+    validate_client_id,
+    validate_client_secret,
+    validate_cluster_resource_group,
+    validate_disk_encryption_set,
+    validate_domain,
+    validate_pull_secret,
+    validate_subnet,
+    validate_visibility,
+    validate_vnet,
+    validate_vnet_resource_group_name,
+    validate_worker_count,
+    validate_worker_vm_disk_size_gb,
+    validate_refresh_cluster_credentials,
+    validate_version_format,
+    validate_outbound_type,
+    validate_load_balancer_managed_outbound_ip_count,
+    validate_enable_managed_identity,
+    validate_platform_workload_identities,
+    validate_cluster_identity,
+)
+from azure.cli.core.commands.parameters import (
+    name_type,
+    get_enum_type,
+    get_three_state_flag,
+    resource_group_name_type,
+    tags_type
+)
 from azure.cli.core.commands.validators import get_default_location_from_resource_group
 
 
--- a/python/az/aro/azext_aro/_validators.py
+++ b/python/az/aro/azext_aro/_validators.py
@@ -311,6 +311,9 @@ def validate_enable_managed_identity(namespace):
     if not namespace.platform_workload_identities:
         raise RequiredArgumentMissingError('Enabling managed identity requires platform workload identities to be provided')  # pylint: disable=line-too-long
 
+    if not namespace.mi_user_assigned:
+        raise RequiredArgumentMissingError('Enabling managed identity requires cluster identity to be provided')
+
 
 def validate_platform_workload_identities(cmd, namespace):
     if namespace.platform_workload_identities is None:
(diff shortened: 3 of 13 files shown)


python/az/aro/azext_aro/_validators.py
```diff
@@ -283,3 +289,74 @@ def validate_load_balancer_managed_outbound_ip_count(namespace):
...
+    if not namespace.enable_managed_identity:
+        return
+
+    if namespace.client_id is not None:
+        raise InvalidArgumentValueError('Must not specify --client-id when --enable-managed-identity is True')
+
+    if namespace.client_secret is not None:
+        raise InvalidArgumentValueError('Must not specify --client-secret when --enable-managed-identity is True')
+
+    if namespace.version is None or not re.match(r'^[4-9]{1}\.[0-9]{1,2}\.[0-9]{1,2}$', namespace.version):
```
kimorris27, line 304: The `if namespace.version is None` part of this conditional makes it so that the cx is going to have to explicitly pass a version even after the default install version is >= 4.14, but I don't think we have a way around enforcing this at the CLI level since the default install version is an "RP-side" thing - am I understanding correctly?

We could technically not enforce it at the CLI level and in doing so allow the request to hit the RP and fail from there, but I like this way you've done it so that a faulty request will fail as early as possible where possible.

python/az/aro/azext_aro/_validators.py
```diff
@@ -283,3 +289,74 @@ def validate_load_balancer_managed_outbound_ip_count(namespace):
...
+        if not identity_resource_id_is_valid(identity.resource_id):
+            raise InvalidArgumentValueError(f"Resource {identity.resource_id} used for platform workload identity {identity.name} is not a valid userAssignedIdentity")  # pylint: disable=line-too-long
+
+
+def validate_cluster_identity(cmd, namespace):
+    if namespace.mi_user_assigned is None:
+        return
+
+    if not namespace.enable_managed_identity:
+        raise RequiredArgumentMissingError('Must set --enable-managed-identity when providing platform workload identities')  # pylint: disable=line-too-long
```
kimorris27, line 339:
```suggestion
        raise RequiredArgumentMissingError('Must set --enable-managed-identity when providing a cluster identity')  # pylint: disable=line-too-long
```

Suggesting this change so that the message aligns with the parameter name.

python/az/aro/azext_aro/custom.py
```diff
@@ -183,10 +184,32 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
...
+        oc.platform_workload_identity_profile = openshiftcluster.PlatformWorkloadIdentityProfile(
+            platform_workload_identities=platform_workload_identities
+        )
+
+        oc.identity = openshiftcluster.Identity(
+            type='UserAssigned',
+            user_assigned_identities={mi_user_assigned: {}}
+        )
+
+        # TODO - add cluster identity here
```
kimorris27, line 201: Re: the comment on line 201, haven't you already added the cluster identity on line 198?

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -918,3 +942,218 @@ def test_validate_load_balancer_managed_outbound_ip_count(test_description, name
...
+    ),
+    (
+        "Should raise InvalidArgumentValueError when version is invalid",
+        Mock(enable_managed_identity=True,
+             client_id=None, client_secret=None,
+             version="a"),
+        InvalidArgumentValueError, 'Enabling managed identity requires --version >= 4.14.z'
+    ),
+    (
+        "Should raise InvalidArgumentValueError when version <= 4.14.0",
```
kimorris27, line 985:
```suggestion
        "Should raise InvalidArgumentValueError when version < 4.14.0",
```

Suggesting this change since 4.14.0 should be allowed.

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -918,3 +942,218 @@ def test_validate_load_balancer_managed_outbound_ip_count(test_description, name
...
+)
+def test_validate_enable_managed_identity(test_description, namespace, expected_exception, expected_exception_message):
+    if expected_exception is None:
+        validate_enable_managed_identity(namespace)
+    else:
+        with pytest.raises(expected_exception, match=expected_exception_message):
+            validate_enable_managed_identity(namespace)
+
+
+test_validate_platform_workload_identities_data = [
```
kimorris27, line 1033: This list could use a `Should not raise any exception when valid` case where the identities are passed as resource IDs.

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -918,3 +942,218 @@ def test_validate_load_balancer_managed_outbound_ip_count(test_description, name
...
+    else:
+        with pytest.raises(expected_exception):
+            validate_platform_workload_identities(cmd, namespace)
+
+    if expected_identities is not None:
+        for expected, actual in zip(expected_identities, namespace.platform_workload_identities):
+            assert (expected.resource_id == actual.resource_id)
+
+
+test_validate_cluster_identity_data = [
```
kimorris27, line 1102: This list could use a `Should not raise any exception when valid` case where the identity is passed as a resource ID.

kimorris27 (changes requested):
LGTM overall! I suggested some small nitpicky changes.

python/az/aro/azext_aro/_validators.py
tsatam: I think we'll need to push a dedicated CLI change after 4.14+ is the default, yeah. I don't see a better way around handling this since the CLI has no knowledge of what ARO's "default" install version is. 

I'm actually fine with removing our CLI-side validation if we want to just handle this in API static validation within the RP. I think that might be a better option.

python/az/aro/azext_aro/_validators.py
tsatam: Good catch, thanks!

python/az/aro/azext_aro/custom.py
tsatam: Yes. I likely re-introduced the comment with a bad rebase. I'll remove it.

tsatam, line 201:
```suggestion
```

kimorris27 (approved):

kimorris27 (dismissed):

python/az/aro/azext_aro/_actions.py
```diff
@@ -0,0 +1,29 @@
...
+
+import argparse
+
+from azext_aro.vendored_sdks.azure.mgmt.redhatopenshift.v2024_08_12_preview.models import PlatformWorkloadIdentity
+from azure.cli.core.azclierror import CLIError
+
+
+# pylint:disable=protected-access
+# pylint:disable=too-few-public-methods
+class AROPlatformWorkloadIdentityAddAction(argparse._AppendAction):
```
bitoku, line 12: Was using validator not enough?
It looks to me that we can achieve the same with a validator and some formalizations after the parse.

I think we shouldn't do `protected-access` because it might be broken some day.

python/az/aro/azext_aro/_params.py
```diff
@@ -18,6 +19,9 @@
 from azext_aro._validators import validate_version_format
 from azext_aro._validators import validate_outbound_type
 from azext_aro._validators import validate_load_balancer_managed_outbound_ip_count
+from azext_aro._validators import validate_enable_managed_identity
```
bitoku, line 22: nit: we can use `import (...)` just as `python/az/aro/azext_aro/tests/latest/unit/test_validators.py`

python/az/aro/azext_aro/_validators.py
```diff
@@ -10,7 +10,8 @@
 from azure.cli.core.commands.client_factory import get_mgmt_service_client, get_subscription_id
 from azure.cli.core.profiles import ResourceType
 from azure.cli.core.azclierror import CLIInternalError, \
-    InvalidArgumentValueError, RequiredArgumentMissingError
+    InvalidArgumentValueError, RequiredArgumentMissingError, \
```
bitoku, line 13: nit: we can use `import (...)` just as `python/az/aro/azext_aro/tests/latest/unit/test_validators.py`

python/az/aro/azext_aro/_validators.py
```diff
@@ -283,3 +289,74 @@ def validate_load_balancer_managed_outbound_ip_count(namespace):
...
+    return resource_id(
+        subscription=get_subscription_id(cmd.cli_ctx),
+        resource_group=namespace.resource_group_name,
+        namespace='Microsoft.ManagedIdentity',
+        type='userAssignedIdentities',
+        name=name,
+    )
+
+
+def identity_resource_id_is_valid(rid):
```
bitoku, line 359: nit: I like this because it starts with `is` and it is clearer that it returns bool.
```suggestion
def is_valid_identity_resource_id(rid):
```

python/az/aro/azext_aro/_validators.py
```diff
@@ -283,3 +289,74 @@ def validate_load_balancer_managed_outbound_ip_count(namespace):
...
+def validate_platform_workload_identities(cmd, namespace):
+    if namespace.platform_workload_identities is None:
+        return
+
+    if not namespace.enable_managed_identity:
+        raise RequiredArgumentMissingError('Must set --enable-managed-identity when providing platform workload identities')  # pylint: disable=line-too-long
+
+    for identity in namespace.platform_workload_identities:
+        if not is_valid_resource_id(identity.resource_id):
+            identity.resource_id = identity_name_to_resource_id(
```
bitoku, line 327: It seems it does more than validation.
I think it would be better that we stick to validation in validate functions, and if we need formalization, do somewhere else, otherwise it will be confusing.

btw is it safe to assume that `identity.resource_id` is a MI name when it is not valid?

python/az/aro/azext_aro/_validators.py
```diff
@@ -283,3 +289,74 @@ def validate_load_balancer_managed_outbound_ip_count(namespace):
...
+
+def validate_cluster_identity(cmd, namespace):
+    if namespace.mi_user_assigned is None:
+        return
+
+    if not namespace.enable_managed_identity:
+        raise RequiredArgumentMissingError('Must set --enable-managed-identity when providing a cluster identity')  # pylint: disable=line-too-long
+
+    if not is_valid_resource_id(namespace.mi_user_assigned):
+        namespace.mi_user_assigned = identity_name_to_resource_id(
```
bitoku, line 342: It seems it does more than validation.
I think it would be better that we stick to validation in validate functions, and if we need formalization, do somewhere else, otherwise it will be confusing.

btw is it safe to assume that `identity.mi_user_assigned` is a MI name when it is not valid?

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -3,13 +3,19 @@
 
 from unittest.mock import Mock, patch
 from azext_aro._validators import (
-    validate_cidr, validate_client_id, validate_client_secret, validate_cluster_resource_group, validate_outbound_type,
-    validate_disk_encryption_set, validate_domain, validate_pull_secret, validate_subnet, validate_subnets,
-    validate_visibility, validate_vnet_resource_group_name, validate_worker_count, validate_worker_vm_disk_size_gb, validate_refresh_cluster_credentials,
-    validate_load_balancer_managed_outbound_ip_count
+    validate_cidr, validate_client_id, validate_client_secret,
```
bitoku, line 6: nit: I would make them one by one line. This would make diff cleaner.

```suggestion
    validate_cidr,
    validate_client_id,
    validate_client_secret,
```

bitoku (commented):
Sorry for a lot of nits.
LGTM for the business logic itself, but I'd like you to take a look at non-nit reviews.

python/az/aro/azext_aro/_actions.py
tsatam: This implementation follows the Azure CLI recommendations for implementing "action"-based parameters like `--assign-platform-workload-identity`: https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#collection-management-with-actions 

It might be possible to do this with a custom implementation, but given that this code will eventually be upstreamed to the Azure CLI I'd prefer to just follow their conventions.

python/az/aro/azext_aro/_validators.py
tsatam: I did this to keep in line with our "validation" performed on vnets/subnets (that does the same name -> resourceid conversion): https://github.com/Azure/ARO-RP/blob/tsatam/ARO-4398-wimi-cli-p1-base-api-functionality/python/az/aro/azext_aro/_validators.py#L239

I agree that this is confusing, it tripped me up when implementing this. Perhaps we move these conversions to the Actions implementation used above? 

This is something I'd like to tackle in an overall refactor, but perhaps not within this PR itself.

tsatam: As for our name / resource ID handling here, we are not yet validating that the identity resource even exists yet. 

Ultimately this will be handled by API-side validation, and we will introduce checks for the identity existing further along in MIWI efforts, once we introduce validation of the passed-in identities and their roleassignments.

python/az/aro/azext_aro/_validators.py
tsatam: Same response as https://github.com/Azure/ARO-RP/pull/3624#discussion_r1657599264

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
tsatam: I made this change globally across our imports for consistency's sake.

python/az/aro/azext_aro/_params.py
tsatam: I made this change globally across our imports for consistency's sake.

python/az/aro/azext_aro/_validators.py
tsatam: I made this change globally across our imports for consistency's sake.

python/az/aro/azext_aro/_validators.py
tsatam: Done

python/az/aro/azext_aro/_actions.py
bitoku: I didn't know there's a doc about it. thanks!

python/az/aro/azext_aro/_validators.py
bitoku: thanks! fair enough

> Perhaps we move these conversions to the Actions implementation used above?
It would be better I guess.

I was worried about `identity.resource_id` conversion because if there's just a typo, it could be interpreted as a resource name.
However, on second thoughts, it doesn't create anything, and even though there's a typo, it'll end up just getting an error, so it isn't a big deal.

bitoku (dismissed):

SudoBrendan (approved):
This all seems OK to me - if you want to move some ocp version validation to RP frontend that would be fine but I wouldn't call it required.

kimorris27 (dismissed):

kimorris27 (approved):
The current E2E failure is an issue with the pipeline and not with these code changes.

python/az/aro/azext_aro/custom.py
```diff
@@ -183,10 +198,30 @@ def aro_create(cmd,  # pylint: disable=too-many-locals
...
+        oc.platform_workload_identity_profile = openshiftcluster.PlatformWorkloadIdentityProfile(
+            platform_workload_identities=platform_workload_identities
+        )
+
+        oc.identity = openshiftcluster.Identity(
+            type='UserAssigned',
+            user_assigned_identities={mi_user_assigned: {}}
+        )
+
+        # TODO - perform client-side validation of required identity permissions
```
SudoBrendan, line 215: q: is this required for this PR, or something for the future?

SudoBrendan (approved):
Seems fine to me, but to be sure - is the TODO required before merge?

python/az/aro/azext_aro/custom.py
tsatam: It's for the future.

tsatam:
Merging this PR now to unblock #3709

PR #3630, opened 2024-06-12
reduce the amount of package names in the logs

hawkowl:
ported from #3422

Reduces the number of redundant package names in the logs

Test plan for issue:

Unit tests

N/A

### DIFF
--- a/pkg/util/steps/condition_test.go
+++ b/pkg/util/steps/condition_test.go
@@ -9,21 +9,29 @@ import (
 	"testing"
 )
 
+type teststruct struct{}
+
 // functionnames that will be used in the conditionFunction below
 // All the keys of map timeoutConditionErrors
-func attachNSGs(context.Context) (bool, error)                             { return false, nil }
-func apiServersReady(context.Context) (bool, error)                        { return false, nil }
-func minimumWorkerNodesReady(context.Context) (bool, error)                { return false, nil }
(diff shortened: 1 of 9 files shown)


pkg/util/steps/condition_test.go
```diff
@@ -9,21 +9,29 @@ import (
 	"testing"
 )
 
+type teststruct struct{}
```
bitoku, line 12: This is introduced because the names of methods have "-fm" as suffix differently from functions, isn't it?
It would be good to have the comment here mentioning that.

pkg/util/steps/runner.go
```diff
@@ -21,7 +21,7 @@ import (
 
 // FriendlyName returns a "friendly" stringified name of the given func.
```
bitoku, line 22: This is not a part of your change, but it's good time to have more specific comment here.

pkg/util/steps/condition_test.go
hawkowl: done

pkg/util/steps/runner.go
hawkowl: done

jaitaiwan (approved):
LGTM, I'm glad these changes are being made. When I looked at this code previously I mused that it seemed unnecessarily verbose as well.

PR #3631, opened 2024-06-12
Emit Cluster Feature Metrics for Cluster Operations

s-fairchild:

[ARO-7839](https://issues.redhat.com/browse/ARO-7839)

Fixes

Emit metrics for all RP operation types, creating, updating, adminupdating, and deleting.

### Test plan for issue:

Covered by unit tests.

No

### DIFF
--- a/pkg/backend/metrics.go
+++ b/pkg/backend/metrics.go
@@ -1,0 +1,216 @@
+package backend
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/util/dns"
+	utillog "github.com/Azure/ARO-RP/pkg/util/log"
+)
+
+func (ocb *openShiftClusterBackend) emitMetrics(log *logrus.Entry, doc *api.OpenShiftClusterDocument, operationType, provisioningState api.ProvisioningState, backendErr error) map[string]string {
+	dimensions := map[string]string{}
+
+	ocb.gatherOperationMetrics(operationType, provisioningState, backendErr, dimensions)
+	ocb.gatherCorrelationID(doc, dimensions)
+	ocb.gatherMiscMetrics(doc, dimensions)
+	ocb.gatherAuthMetrics(doc, dimensions)
+	ocb.gatherNetworkMetrics(doc, dimensions)
+	ocb.gatherNodeMetrics(doc, dimensions)
+
+	ocb.logMetricDimensions(log, operationType, dimensions)
+	ocb.m.EmitGauge(ocb.getMetricName(operationType), metricValue, dimensions)
+
+	// dimensions is returned here for testing purposes
+	return dimensions
+}
+
+func (ocb *openShiftClusterBackend) getMetricName(operationType api.ProvisioningState) string {
+	return fmt.Sprintf("%s.%s", metricPackage, operationType)
+}
+
+func (ocb *openShiftClusterBackend) getResultType(backendErr error) utillog.ResultType {
+	var resultType utillog.ResultType
+	err, ok := backendErr.(*api.CloudError)
+	if ok {
+		resultType = utillog.MapStatusCodeToResultType(err.StatusCode)
+	}
+	return resultType
+}
+
+func (ocb *openShiftClusterBackend) logMetricDimensions(log *logrus.Entry, operationType api.ProvisioningState, dimensions map[string]string) {
+	for metric, value := range dimensions {
+		log.Info(fmt.Sprintf("%s.%s: %s = %s", metricPackage, operationType, metric, value))
+	}
+}
+
+func (m *openShiftClusterBackend) gatherCorrelationID(doc *api.OpenShiftClusterDocument, dimensions map[string]string) {
+	if doc.CorrelationData != nil {
+		dimensions[correlationDataIdMetricName] = doc.CorrelationData.CorrelationID
+		dimensions[correlationDataClientRequestIdMetricName] = doc.CorrelationData.ClientRequestID
+		dimensions[correlationDataRequestIdMetricName] = doc.CorrelationData.RequestID
+	} else {
+		dimensions[correlationDataIdMetricName] = empty
+		dimensions[correlationDataClientRequestIdMetricName] = empty
+		dimensions[correlationDataRequestIdMetricName] = empty
+	}
+}
+
+func (ocb *openShiftClusterBackend) gatherOperationMetrics(operationType, provisioningState api.ProvisioningState, backendErr error, dimensions map[string]string) {
+	dimensions[operationTypeMetricName] = operationType.String()
+	dimensions[provisioningStateMetricName] = provisioningState.String()
+	dimensions[resultTypeMetricName] = string(ocb.getResultType(backendErr))
+}
+
+func (ocb *openShiftClusterBackend) gatherMiscMetrics(doc *api.OpenShiftClusterDocument, dimensions map[string]string) {
+	dimensions[subscriptionIdMetricName] = ocb.env.SubscriptionID()
+	dimensions[resourceIdMetricName] = doc.ResourceID
+	if doc.OpenShiftCluster != nil {
+		dimensions[clusterNameMetricName] = doc.OpenShiftCluster.Name
+		dimensions[locationMetricName] = doc.OpenShiftCluster.Location
+		dimensions[ocpVersionMetricName] = doc.OpenShiftCluster.Properties.ClusterProfile.Version
+		dimensions[rpVersionMetricName] = doc.OpenShiftCluster.Properties.ProvisionedBy
+		dimensions[resourecGroupMetricName] = doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID
+
+		for flag, feature := range doc.OpenShiftCluster.Properties.OperatorFlags {
+			dimensions[fmt.Sprintf("%s-%s", operatorFlagsMetricName, flag)] = feature
+		}
+	}
+
+	dimensions[asyncOperationsIdMetricName] = doc.AsyncOperationID
+
+	if doc.OpenShiftCluster.Properties.WorkerProfiles != nil {
+		dimensions[workerProfileCountMetricName] = strconv.FormatInt(int64(len(doc.OpenShiftCluster.Properties.WorkerProfiles)), 10)
+	}
+
+	if doc.OpenShiftCluster.Tags != nil {
+		dimensions[tagsMetricName] = enabled
+	} else {
+		dimensions[tagsMetricName] = disabled
+	}
+}
+
+func (ocb *openShiftClusterBackend) gatherNodeMetrics(doc *api.OpenShiftClusterDocument, dimensions map[string]string) {
+	if doc.OpenShiftCluster.Properties.MasterProfile.DiskEncryptionSetID != "" {
+		dimensions[masterProfileEncryptionSetIdMetricName] = enabled
+	} else {
+		dimensions[masterProfileEncryptionSetIdMetricName] = disabled
+	}
+
+	mp := doc.OpenShiftCluster.Properties.MasterProfile
+	dimensions[masterProfileVmSizeMetricName] = string(mp.VMSize)
+
+	if doc.OpenShiftCluster.Properties.MasterProfile.EncryptionAtHost == api.EncryptionAtHostEnabled {
+		dimensions[masterEncryptionAtHostMetricName] = string(api.EncryptionAtHostEnabled)
+	} else if doc.OpenShiftCluster.Properties.MasterProfile.EncryptionAtHost == api.EncryptionAtHostDisabled {
+		dimensions[masterEncryptionAtHostMetricName] = string(api.EncryptionAtHostDisabled)
+	} else {
+		dimensions[masterEncryptionAtHostMetricName] = unknown
+	}
+
+	if len(doc.OpenShiftCluster.Properties.WorkerProfiles) > 0 {
+		wp := doc.OpenShiftCluster.Properties.WorkerProfiles[0]
+		dimensions[workerVmSizeMetricName] = string(wp.VMSize)
+		dimensions[workerVmDiskSizeMetricName] = strconv.FormatInt(int64(wp.DiskSizeGB), 10)
+
+		dimensions[workerVmSizeMetricName] = string(wp.VMSize)
+		dimensions[workerVmDiskSizeMetricName] = strconv.FormatInt(int64(wp.DiskSizeGB), 10)
+
+		if wp.EncryptionAtHost == api.EncryptionAtHostEnabled {
+			dimensions[workerEncryptionAtHostMetricName] = string(api.EncryptionAtHostEnabled)
+		} else if wp.EncryptionAtHost == api.EncryptionAtHostDisabled {
+			dimensions[workerEncryptionAtHostMetricName] = string(api.EncryptionAtHostDisabled)
+		} else {
+			dimensions[workerEncryptionAtHostMetricName] = unknown
+		}
+	}
+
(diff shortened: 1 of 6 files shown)


pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+	} else {
+		dimensions["fips"] = unknown
+	}
+}
+
+func (ocb *openShiftClusterBackend) gatherAuthMetrics(doc *api.OpenShiftClusterDocument, dimensions map[string]string) {
+	if doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile != nil {
+		dimensions["workloadidentity"] = enabled
+	} else if doc.OpenShiftCluster.Properties.ServicePrincipalProfile != nil {
+		dimensions["serviceprincipal"] = disabled
```
tsatam, line 139: this should be enabled, right?

tsatam: I'm also wondering if we should instead have a single dimension here named something like `clusteridentity` that is set to either `serviceprincipal` or `managed`?

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+		dimensions["masterprofile.encryptionathost"] = unknown
+	}
+
+	for _, p := range doc.OpenShiftCluster.Properties.WorkerProfiles {
+		if p.VMSize != api.VMSizeStandardD4sV3 {
+			dimensions["customvmsize"] = string(p.VMSize)
+		}
+
+		if p.DiskSizeGB > 128 {
+			dimensions["customisksizegb"] = strconv.FormatInt(int64(p.DiskSizeGB), 10)
```
tsatam, line 114:
```suggestion
			dimensions["customdisksizegb"] = strconv.FormatInt(int64(p.DiskSizeGB), 10)
```

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+		dimensions["tags"] = disabled
+	}
+}
+
+func (ocb *openShiftClusterBackend) gatherNodeMetrics(doc *api.OpenShiftClusterDocument, dimensions map[string]string) {
+	if doc.OpenShiftCluster.Properties.MasterProfile.DiskEncryptionSetID != "" {
+		dimensions["masterprofile.diskencryptionsetid"] = enabled
+	} else {
+		dimensions["masterprofile.diskencryptionsetid"] = disabled
+	}
```
tsatam, line 98: IMO we don't need a dimension for this - we can just utilize whether or not encryptionathost is enabled for this purpose.

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+	dimensions["location"] = doc.OpenShiftCluster.Location
+	dimensions["resourcegroup"] = doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID
+	dimensions["name"] = doc.OpenShiftCluster.Name
+	dimensions["ocpversion"] = doc.OpenShiftCluster.Properties.ClusterProfile.Version
+	dimensions["workercount"] = strconv.FormatInt(int64(len(doc.OpenShiftCluster.Properties.WorkerProfiles)), 10)
+	dimensions["rpversion"] = doc.OpenShiftCluster.Properties.ProvisionedBy
+
+	for flag, feature := range doc.OpenShiftCluster.Properties.OperatorFlags {
+		dimensions[fmt.Sprintf("operator-%s", flag)] = feature
+	}
```
tsatam, line 84: Do we need this? I don't think users have any ability to influence changes to the default operator flags we put down, so these values will be the same for all cluster installs.

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if doc.OpenShiftCluster.Tags != nil {
+		dimensions["tags"] = enabled
+	} else {
+		dimensions["tags"] = disabled
+	}
+}
```
tsatam, line 91: Is this a useful dimension for us? We don't have any behavioral changes based on tags. I could see this being useful for us to e.g. identify clusters whose installs failed due to a policy requiring tags, but those are usually case-by-case situations where common metrics wouldn't help I don't think.

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+		dimensions["preconfigurednsg"] = string(api.PreconfiguredNSGDisabled)
+	} else {
+		dimensions["preconfigurednsg"] = unknown
+	}
+
+	if doc.OpenShiftCluster.Properties.FeatureProfile.GatewayEnabled {
+		dimensions["gatewayenabled"] = enabled
+	} else {
+		dimensions["gatewayenabled"] = disabled
+	}
```
tsatam, line 202: Do we need this? The default state for all new cluster installs is enabled.

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+		dimensions["masterprofile.encryptionathost"] = string(api.EncryptionAtHostEnabled)
+	} else if doc.OpenShiftCluster.Properties.MasterProfile.EncryptionAtHost == api.EncryptionAtHostDisabled {
+		dimensions["masterprofile.encryptionathost"] = string(api.EncryptionAtHostDisabled)
+	} else {
+		dimensions["masterprofile.encryptionathost"] = unknown
+	}
+
+	for _, p := range doc.OpenShiftCluster.Properties.WorkerProfiles {
+		if p.VMSize != api.VMSizeStandardD4sV3 {
+			dimensions["customvmsize"] = string(p.VMSize)
```
tsatam, line 110: IMO we should just emit the VM size to a generic `workervmsize` dimension, and not worry about whether it's the "default" or not here. 

We should also emit the VM size chosen for the control plane as a dimension here too.

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if doc.OpenShiftCluster.Properties.NetworkProfile.ServiceCIDR != "172.30.0.0/16" {
+		dimensions["servicecidr"] = custom
+	} else {
+		dimensions["servicecidr"] = defaultSet
+	}
+
+	if doc.OpenShiftCluster.Properties.ClusterProfile.Domain != "" {
+		dimensions["domain"] = enabled
+	}
```
tsatam, line 184: `domain` is always set to a value during cluster installations, either a fully qualified domain provided by the customer, or a random ID for managed domains. 

We may want to include logic similar to https://github.com/Azure/ARO-RP/blob/6997415b87d29fa466963f7c50a71c02122045da/pkg/util/dns/dns.go#L195 to identify whether the domain is managed or not.

tsatam (commented):
I have some questions around the specific dimensions we're emitting - I want to make sure that each individual dimension we include on these metrics is explicitly useful to us so we're not introducing noise into our observability.

pkg/backend/metrics.go
swiencki: This is a common Cluster Install Failure, it would be good to capture the rate at which it happens

tsatam: My understanding might have been incorrect - looking through our code it looks like it is possible to either set or omit diskencryptionsetid when encryption at host is enabled.

pkg/backend/metrics.go
s-fairchild: It maybe helpful for when we do adminupdates, which will also emit these metrics.  
This gets called for creating, adminupdate, update, and deleting.

pkg/backend/metrics.go
s-fairchild: We want to capture the use of all user provided input during cluster creation.  
So this isn't so much of a, will this help us see a problem, but rather are users using things? What things?

pkg/backend/metrics.go
kimorris27: It makes sense to me to include it, since these metrics are being emitted on all operations and not just on new cluster creations.

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+	metricInstall = metricPackage + ".installphase"
+	enabled       = "Enabled"
+	disabled      = "Disabled"
+	custom        = "Custom"
+	defaultSet    = "Default"
+	unknown       = "unknown"
+	empty         = "empty"
+)
+
+func (ocb *openShiftClusterBackend) emitMetrics(log *logrus.Entry, doc *api.OpenShiftClusterDocument, operationType, provisioningState api.ProvisioningState, backendErr error) {
```
kimorris27, line 27: I recommend unit testing this function using mock emitters. An example from `pkg/monitor`: https://github.com/Azure/ARO-RP/blob/master/pkg/monitor/cluster/certificateexpirationstatuses_test.go.

kimorris27 (changes requested):
Good work! I recommended adding unit tests for `emitMetrics` and left a few other smaller question/concerns.

pkg/backend/metrics.go
tsatam: I missed that part, that invalidates a lot of my feedback so far. I'll go through and fixup suggestions.

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if doc.OpenShiftCluster.Properties.MasterProfile.EncryptionAtHost == api.EncryptionAtHostEnabled {
+		dimensions["masterprofile.encryptionathost"] = string(api.EncryptionAtHostEnabled)
+	} else if doc.OpenShiftCluster.Properties.MasterProfile.EncryptionAtHost == api.EncryptionAtHostDisabled {
+		dimensions["masterprofile.encryptionathost"] = string(api.EncryptionAtHostDisabled)
+	} else {
+		dimensions["masterprofile.encryptionathost"] = unknown
+	}
+
+	for _, p := range doc.OpenShiftCluster.Properties.WorkerProfiles {
```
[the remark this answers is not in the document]
tsatam: I agree that this should be handled differently (clusters post-install will have a 1:1 mapping between WorkerProfiles here and MachineSets AFAICT) - we might want to either omit data for worker profiles entirely, or emit some sort of "summary" of the status of a given field across all worker profiles.

pkg/backend/metrics.go
s-fairchild: Thank you.

pkg/backend/metrics.go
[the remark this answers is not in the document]
s-fairchild: I was originally thinking about using the indexer to separate them. WorkerProfile 0, 1, 2 would have their own distinct metrics. I overlooked that this is in fact overwriting the previous profile.

We still want to emit the profile size so it's available in the metrics, even though it's in the machinetsets.

I'll go back to using the index value, any objections?

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (m *openShiftClusterBackend) gatherCorrelationID(doc *api.OpenShiftClusterDocument, dimensions map[string]string) {
+	if doc.CorrelationData != nil {
+		dimensions["correlationid"] = doc.CorrelationData.CorrelationID
+		dimensions["client_requestid"] = doc.CorrelationData.ClientRequestID
+		dimensions["requestid"] = doc.CorrelationData.RequestID
+		dimensions["async_operationsid"] = doc.AsyncOperationID
+	} else {
+		dimensions["correlationid"] = empty
+		dimensions["client_requestid"] = empty
```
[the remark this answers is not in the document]
s-fairchild: Overlooked it, thanks.

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+		dimensions["async_operationsid"] = empty
```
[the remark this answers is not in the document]
s-fairchild: Good point, I moved async_operationsid to the misc function.

pkg/backend/metrics.go
s-fairchild: I update this to simply set the vmsize for workers and masters.

pkg/backend/metrics.go
s-fairchild: It is getting unit tested by [openshiftcluster_test.go](https://github.com/Azure/ARO-RP/blob/7caa4a1bc9fb359d966a725a299d0cdff22af8d0/pkg/backend/openshiftcluster_test.go).
Test coverage is about 66%, it could be better, but it's not worth modifying those tests further.
I can add some unit tests specifically for `emitMetrics`.

pkg/backend/metrics.go
```diff
@@ -0,0 +1,203 @@
...
+	disabled      = "Disabled"
+	custom        = "Custom"
+	defaultSet    = "Default"
+	unknown       = "unknown"
+	empty         = "empty"
+)
+
+func (ocb *openShiftClusterBackend) emitMetrics(log *logrus.Entry, doc *api.OpenShiftClusterDocument, operationType, provisioningState api.ProvisioningState, backendErr error) {
+	dimensions := map[string]string{
+		"operationtype":     operationType.String(),
```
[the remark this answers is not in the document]
s-fairchild: I was thinking the same, a lot of work, but worthwhile. I added the metric names as constants.

pkg/backend/metrics.go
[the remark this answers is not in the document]
s-fairchild: Seeing as how all the worker profiles should contain the same values, I simplified it by checking if the array is greater than 0, and grabbing the first worker profiles data.

pkg/backend/metrics.go
s-fairchild: Thanks, I updated this to check if it's a managed domain or not.

pkg/backend/openshiftcluster.go
```diff
@@ -148,11 +148,14 @@ func (ocb *openShiftClusterBackend) handle(ctx context.Context, log *logrus.Entr
 		// backend worker to pick up next install phase
 		doc, err = ocb.dbOpenShiftClusters.Get(ctx, strings.ToLower(doc.OpenShiftCluster.ID))
 		if err != nil {
+			ocb.emitMetrics(log, doc, api.ProvisioningStateFailed, api.ProvisioningStateCreating, err)
```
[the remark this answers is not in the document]
s-fairchild: Thanks for your review. I agree, calling this from endLease makes the most sense.
I've moved this function call to endLease, and addressed your other concerns. I'll add some unit tests specifically for metrics next week.

pkg/backend/metrics.go
s-fairchild: That makes more sense, updated with your suggestion.

pkg/backend/metrics.go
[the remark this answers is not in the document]
kimorris27: I did some digging, and as far as I can tell, there's only one worker profile and it's only validated and used at install time. See here: https://github.com/Azure/ARO-RP/blob/6997415b87d29fa466963f7c50a71c02122045da/pkg/api/v20240812preview/openshiftcluster_validatestatic.go#L131

So emitting only the first worker profile's data works for me.

pkg/backend/metrics.go
```diff
@@ -0,0 +1,222 @@
...
+)
+
+func (ocb *openShiftClusterBackend) emitMetrics(log *logrus.Entry, doc *api.OpenShiftClusterDocument, operationType, provisioningState api.ProvisioningState, backendErr error) error {
+	dimensions := map[string]string{}
+
+	ocb.gatherOperationMetrics(operationType, provisioningState, backendErr, dimensions)
+	ocb.gatherCorrelationID(doc, dimensions)
+	ocb.gatherMiscMetrics(doc, dimensions)
+	ocb.gatherAuthMetrics(doc, dimensions)
+	err := ocb.gatherNetworkMetrics(doc, dimensions)
```
hawkowl, line 24: We should wrap the error here to make it clear that it is the metrics that is returning, *or* log the error but do not return it and continue along anyway.

kimorris27 (dismissed):

pkg/backend/metrics.go
s-fairchild: I've added a warn log here, and a helper function to log warn messages for empty metric values.
This will help to remove duplicated `if` blocks and `log.Warnf()` calls.

kimorris27 (approved):

cadenmarchese (approved):
lgtm

PR #3633, opened 2024-06-12
EnableAutomaticOSUpgrade for dev-proxy-vmss and remove the no longer needed SA params

nwnt:

Fixes: ARO-8414

S360/Security Waves flagged all our VMSSes that they need to enable Automatic OS Upgrade.  We will request an exception for RP and Gateway VMSSes, and we will fix only dev-proxy by turning on the option.  This PR also removes the no longer needed storage account parameters for the previously deleted boot diagnostics SA so that the deployment can continue.

### Test plan for issue:

Will test in INT with E2E.

N/A

The full INT deployment should be enough.

### DIFF
--- a/pkg/deploy/assets/env-development.json
+++ b/pkg/deploy/assets/env-development.json
@@ -280,7 +280,10 @@
             },
             "properties": {
                 "upgradePolicy": {
-                    "mode": "Rolling"
+                    "mode": "Rolling",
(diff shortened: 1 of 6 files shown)


weinong (approved):
LGTM

weinong:
can we also close https://issues.redhat.com/browse/ARO-4269?

hlipsig (approved):

PR #3634, opened 2024-06-13
Initial ansible proof of concept

ventifus:

This PR leverages Ansible to automate cluster creation in predefined configurations. These configurations are intended to exercise ARO features in a repeatable manner. A formal proposal is under development, see

https://docs.google.com/document/d/1AIcP1jMQS-9hsI_bS6Qhr3IoqtD97LqKfw6c9PNXD5g

### Test plan for issue:

Then run

```
$ make cluster
```

This will first build the required aro-ansible container image if it doesn't exist, then it will execute `ansible-playbook` in the container.

There are several variables implemented to control aspects of the playbook execution. These can be combined with each other as needed.

To personalize the resulting resource groups, set the CLUSTERPREFIX as desired. This variable defaults to your current shell's `$USER`.

```
$ make cluster CLUSTERPREFIX=ocpbugs35300
```

The default region used is `eastus`. Set the REGION parameter to choose a different region:

```
$ make cluster REGION=centraluseuap
```

To choose one or more cluster configurations, set the CLUSTERPATTERN parameter to a wildcard string that matches the cluster scenarios you wish to test:

```
$ make cluster CLUSTERPATTERN=udr
```

To clean up at the end of the run, set CLEANUP to True. This will delete the cluster, the resource group, then the Entra Service Principal and Application.

```
$ make cluster CLEANUP=True
```

- `basic`: Simplest cluster, nothing fancy
- `private`: Simple cluster with apiserver and ingress visibility set to private.
- `enc`: Encryption-at-host enabled
- `udr`: UserDefinedRouting with a blackhole Route Table
- `byok`: Disk encryption using bring-your-own-key
- `sre-shared-cluster` Incomplete; intended to reproducibly recreate the shared cluster

Private clusters such as `private` and `udr` will cause the creation of a jumphost to access the cluster API. Your local SSH public key will be passed to the jumphost, then ansible will use your corresponding private key to tunnel SSH through it. If your local SSH configuration differs from defaults, the Makefile supports two variables to tweak things:

```
SSH_CONFIG_DIR := $(HOME)/.ssh/
SSH_KEY_BASENAME := id_rsa
```

Yes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -346,7 +346,7 @@ install-go-tools:
 
 .PHONY: ansible-image
 ansible-image:
-	docker image exists aro-ansible:$(VERSION) || docker build . -f Dockerfile.ansible --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) --no-cache=$(NO_CACHE) --tag aro-ansible:$(VERSION)
+	podman $(PODMAN_REMOTE_ARGS) image exists aro-ansible:$(VERSION) || podman $(PODMAN_REMOTE_ARGS) build . -f Dockerfile.ansible --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) --no-cache=$(NO_CACHE) --tag aro-ansible:$(VERSION)
 
 LOCATION := eastus
 CLUSTERPREFIX := $(USER)
@@ -366,12 +366,12 @@ cluster: ansible-image cluster-deploy cluster-cleanup
 	@true
 .PHONY: cluster-deploy
 cluster-deploy:
-	docker run --rm -it -v $${AZURE_CONFIG_DIR:-~/.azure}:/opt/app-root/src/.azure$(PODMAN_VOLUME_OVERLAY) -v ./ansible:/ansible$(PODMAN_VOLUME_OVERLAY) -v $(SSH_CONFIG_DIR):/root/.ssh$(PODMAN_VOLUME_OVERLAY) aro-ansible:$(VERSION) -i hosts.yaml $(CLUSTERFILTER) -e location=$(LOCATION) -e CLUSTERPREFIX=$(CLUSTERPREFIX) -e CLEANUP=$(CLEANUP) -e SSH_KEY_BASENAME=$(SSH_KEY_BASENAME) $(SKIP_VERBOSE) deploy.playbook.yaml
+	podman $(PODMAN_REMOTE_ARGS) run --rm -it -v $${AZURE_CONFIG_DIR:-~/.azure}:/opt/app-root/src/.azure$(PODMAN_VOLUME_OVERLAY) -v ./ansible:/ansible$(PODMAN_VOLUME_OVERLAY) -v $(SSH_CONFIG_DIR):/root/.ssh$(PODMAN_VOLUME_OVERLAY) aro-ansible:$(VERSION) -i hosts.yaml $(CLUSTERFILTER) -e location=$(LOCATION) -e CLUSTERPREFIX=$(CLUSTERPREFIX) -e CLEANUP=$(CLEANUP) -e SSH_KEY_BASENAME=$(SSH_KEY_BASENAME) $(SKIP_VERBOSE) deploy.playbook.yaml
 .PHONY: cluster-cleanup
 cluster-cleanup:
 	@if [ "${CLEANUP}" == "True" ]; \
 	then \
-		docker run --rm -it -v $${AZURE_CONFIG_DIR:-~/.azure}:/opt/app-root/src/.azure$(PODMAN_VOLUME_OVERLAY) -v ./ansible:/ansible$(PODMAN_VOLUME_OVERLAY) -v $(SSH_CONFIG_DIR):/root/.ssh$(PODMAN_VOLUME_OVERLAY) aro-ansible:$(VERSION) -i hosts.yaml $(CLUSTERFILTER) -e location=$(LOCATION) -e CLUSTERPREFIX=$(CLUSTERPREFIX) -e CLEANUP=$(CLEANUP) -e SSH_KEY_BASENAME=$(SSH_KEY_BASENAME) $(SKIP_VERBOSE) cleanup.playbook.yaml; \
+		podman $(PODMAN_REMOTE_ARGS) run --rm -it -v $${AZURE_CONFIG_DIR:-~/.azure}:/opt/app-root/src/.azure$(PODMAN_VOLUME_OVERLAY) -v ./ansible:/ansible$(PODMAN_VOLUME_OVERLAY) -v $(SSH_CONFIG_DIR):/root/.ssh$(PODMAN_VOLUME_OVERLAY) aro-ansible:$(VERSION) -i hosts.yaml $(CLUSTERFILTER) -e location=$(LOCATION) -e CLUSTERPREFIX=$(CLUSTERPREFIX) -e CLEANUP=$(CLEANUP) -e SSH_KEY_BASENAME=$(SSH_KEY_BASENAME) $(SKIP_VERBOSE) cleanup.playbook.yaml; \
 	fi
 
 .PHONY: lint-ansible
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ ARO_IMAGE_BASE = ${RP_IMAGE_ACR}.azurecr.io/aro
 E2E_FLAGS ?= -test.v --ginkgo.v --ginkgo.timeout 180m --ginkgo.flake-attempts=2 --ginkgo.junit-report=e2e-report.xml
 GO_FLAGS ?= -tags=containers_image_openpgp,exclude_graphdriver_btrfs,exclude_graphdriver_devicemapper
 NO_CACHE ?= true
+PODMAN_VOLUME_OVERLAY=$(shell if [[ $$(getenforce) == "Enforcing" ]]; then echo ":O"; else echo ""; fi 2>/dev/null)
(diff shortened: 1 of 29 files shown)


Makefile
```diff
@@ -285,4 +286,23 @@ vendor:
 install-go-tools:
 	go install ${GOTESTSUM}
 
-.PHONY: admin.kubeconfig aks.kubeconfig aro az ci-rp ci-clean clean client deploy dev-config.yaml discoverycache fix-macos-vendor generate image-aro-multistage image-fluentbit image-proxy init-contrib lint-go runlocal-rp proxy publish-image-aro-multistage publish-image-fluentbit publish-image-proxy secrets secrets-update e2e.test tunnel test-e2e test-go test-python vendor build-all validate-go unit-test-go coverage-go validate-fips install-go-tools
+ansible-image:
+	docker image exists aro-ansible:$(VERSION) || docker build . -f Dockerfile.ansible --build-arg REGISTRY=$(REGISTRY) --build-arg VERSION=$(VERSION) --no-cache=$(NO_CACHE) --tag aro-ansible:$(VERSION)
```
s-fairchild, line 290: If we want to recreate the aro-ansible container during development, does this mean we will have to manually delete the image? `podman image rm aro-ansiable:whatever-version`

SudoBrendan: Good call on this, +1, We should leverage proper build context and docker caching instead of using `docker image exists`.

This means that when we rebuild without changes, then we hit the cache 100% as long as our docker build context, mostly managed by Dockerfile.ansible.dockerignore (and missing in this PR) is stood up correctly. It also means if we rebuild _with_ changes, we correctly rebuild the image as desired.

SudoBrendan:
Adding a hold because this is a POC. FWIW, I really love the idea. Andrew, could you please put up some cross-region meetings to go over this proposal and get feedback and/or buy-in from the team on the idea so we can either close or move forward with it?

Makefile
ventifus: On my machine, rebuilding the container image takes 5m40s even if all layers end up being the same, mostly due to re-running `pip` to install `ansible` and `azure-cli`. This proved frustrating when all I wanted to do was to run the existing image again. Is there a way of making `docker build` know it's going to be a no-op because nothing (local) has changed?

I'll add an empty `dockerignore` (only the `/ansible` directory gets copied into the container so there's nothing to ignore right now).

pepedocs:
I don't know why my comment got lost. But I'm adding it here.

I have to do the following in my case because overlay is always Enforcing. I have both podman and docker installed, but I'm using docker for this makefile. Because of this the docker volume mode becomes `O` , which AFAICR there is no such thing in docker.

Diff of my changes. after these changes everything worked.
```
**@@ -6,6 +6,8 @@ E2E_FLAGS ?= -test.v --ginkgo.v --ginkgo.timeout 180m --ginkgo.flake-attempts=2
 GO_FLAGS ?= -tags=containers_image_openpgp,exclude_graphdriver_btrfs,exclude_graphdriver_devicemapper
 NO_CACHE ?= true
 PODMAN_VOLUME_OVERLAY=$(shell if [[ $$(getenforce) == "Enforcing" ]]; then echo ":O"; else echo ""; fi 2>/dev/null)
+$(info    PODMAN_VOLUME_OVERLAY is $(PODMAN_VOLUME_OVERLAY))
+PODMAN_VOLUME_OVERLAY=
**
```

Makefile
s-fairchild: A manual way that comes to mind, is looping over the files in the image with bash conditional expression `-nt` to determine if they have been modified after the image was built. [Bash Conditional Expressions](https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html), although something native to docker builld or podman builbah would be ideal.

ansible/tasks/create_aro_cluster.yaml
```diff
@@ -0,0 +1,134 @@
...
+#       subnet_id: "{{ master_subnet_state.state.id }}"
+- name: create_aro_cluster | Check if cluster already exists
+  ansible.builtin.command:
+    argv: ["az", "aro", "show", "--name={{ name }}", "--resource-group={{ resource_group }}", "-o=yaml"]
+  failed_when: "'provisioningState: Succeeded' not in aro_existing_cluster.stdout"
+  delegate_to: localhost
+  register: aro_existing_cluster
+  ignore_errors: true
+- name: create_aro_cluster | Create aro cluster
+  when: aro_existing_cluster is not success
```
sankur-codes, line 50: This condition wouldn't be enough to verify if the cluster already exists. According to this, if we have a cluster  already existing with `provisioningState: Failed`, it will still initiate the cluster creation which will eventually fail. 

I think if we get any result to the `az aro show` command for the cluster, we could consider the cluster to be present ?

sankur-codes (commented):
I haven't yet tried using it but have a few nitpicks & suggestion

Makefile
sankur-codes: I don't see much of the changes to be happening in the docker file or image over a period of time. Would it be a good idea to build a version of the image and make it readily available in a image registry (or as a tar file) so there's not much time wasted on rebuilding it ?

ventifus:
> I don't know why my comment got lost. But I'm adding it here.
> 
>     1. I have to do the following in my case because overlay is always Enforcing. I have both podman and docker installed, but I'm using docker for this makefile. Because of this the docker volume mode becomes `O` , which AFAICR there is no such thing in docker.

This needs improvement. I added `:O` because on my system the container user didn't have permissions on the volumes' files and I didn't want to relabel the source files with `:Z`. I need an approach that will work with whatever container tool is in use, I know of at least three: docker/moby, podman and colima. 

> 
> 
> Diff of my changes. after these changes everything worked.
> 
> ```
> **@@ -6,6 +6,8 @@ E2E_FLAGS ?= -test.v --ginkgo.v --ginkgo.timeout 180m --ginkgo.flake-attempts=2
>  GO_FLAGS ?= -tags=containers_image_openpgp,exclude_graphdriver_btrfs,exclude_graphdriver_devicemapper
>  NO_CACHE ?= true
>  PODMAN_VOLUME_OVERLAY=$(shell if [[ $$(getenforce) == "Enforcing" ]]; then echo ":O"; else echo ""; fi 2>/dev/null)
> +$(info    PODMAN_VOLUME_OVERLAY is $(PODMAN_VOLUME_OVERLAY))
> +PODMAN_VOLUME_OVERLAY=
> **
> ```
> 
>     2. The udr install did not exit although the failed cluster installation was successful. I think that is because of the check that doesn't satisfy because it was a failed cluster installation test.

I don't know, what step did it fail on? Some tasks have retry loops so it might take a while for it to truly fail.

ansible/tasks/create_aro_cluster.yaml
ventifus: This is somewhat intentional, since I haven't decided what to do if we encounter a non-successful cluster here. Delete it and recreate it? Fail out? What if it's not in a terminal state?

ansible/hosts.yaml
```diff
@@ -0,0 +1,206 @@
...
+          channel: stable-4.14
+          version: 4.14.33
+          admin_acks:
+            - {"data":{"ack-4.13-kube-1.27-api-removals-in-4.14":"true"}}
+        - from: "4.14"
+          channel: stable-4.15
+          version: 4.15.22
+        - from: "4.15"
+          channel: stable-4.16
+          version: latest
```
yjst2012, line 180: the stable-4.16 is empty right now, also an admin-ack is required, if really need to proceed like me, can go with candidate-4.16 instead:
```
        - from: "4.15"
          channel: candidate-4.16
          version: 4.16.5
          admin_acks:
            - {"data":{"ack-4.15-kube-1.29-api-removals-in-4.16":"true"}}

```

ansible/hosts.yaml
```diff
@@ -0,0 +1,206 @@
...
+    resource_group: "{{ CLUSTERPREFIX }}-{{ inventory_hostname }}-{{ location }}"
+    network_prefix_cidr: 10.0.0.0/22
+    master_cidr: 10.0.0.0/23
+    worker_cidr: 10.0.2.0/23
+    apiserver_visibility: Private
+    ingress_visibility: Private
+    outbound_type: UserDefinedRouting
+
+upgrade_clusters:
+  hosts:
```
yjst2012, line 158: I'm using this as the currently default version is 4.13.40:
```
    upgrade-413to416:
      name: aro
      version: 4.13.40
      upgrade:
        # https://access.redhat.com/labs/ocpupgradegraph/update_path?channel=stable-4.13&arch=x86_64&is_show_hot_fix=false&current_ocp_version=4.12.25&target_ocp_version=4.15.22
        - from: "4.13"
          channel: stable-4.13
          version: 4.13.45
          admin_acks:
            - {"data":{"ack-4.12-kube-1.26-api-removals-in-4.13":"true"}}
          allow-not-recommended: true
        - from: "4.13"
          channel: stable-4.14
          version: 4.14.33
          admin_acks:
            - {"data":{"ack-4.13-kube-1.27-api-removals-in-4.14":"true"}}
        - from: "4.14"
          channel: stable-4.15
          version: 4.15.22
        - from: "4.15"
          channel: candidate-4.16
          version: 4.16.5
          admin_acks:
            - {"data":{"ack-4.15-kube-1.29-api-removals-in-4.16":"true"}}
```

yjst2012 (commented):
very useful script, saved a lot of time and effort, thank you

ventifus:
Moved to https://github.com/Azure/ARO-RP/pull/3843

PR #3638, opened 2024-06-18
use track2 SDK in updateAPIIPEarly

bitoku:

Fixes https://issues.redhat.com/browse/ARO-7316

This has the same change as https://github.com/Azure/ARO-RP/pull/3579 , but changed tenant ID.
This will fix the failure e2e in canary.

### Test plan for issue:

cluster creation in local dev.
cluster creation in canary.

N/A

cluster creation success


tiguelu (approved):
LGTM. This is reapplying the changes in ipaddresses files that were already reviewed in #3636, plus the bugfix for the credentials to use the cluster's tenant ID instead of the RP tenant.

PR #3642, opened 2024-06-24
[ARO-9169] Remove disused clustermanager frontend code

hawkowl:
Fixes https://issues.redhat.com/browse/ARO-9169

Removes the ClusterManager frontend code that is non-functional and won't be switched on.


mociarain:
Can you give some guidance on how to review this? I'm not super sure how to test this or what removing all this code means. If there's any supporting docs or anything I'd appreciate a link :)

kimorris27 (approved):
LGTM!

@mociarain , my understanding is that this code is related to some API endpoints for functionality that never ended up being used. And because Azure API docs are generated from code, these unused APIs are documented publicly... so every once in a while, we get an alert where a customer is trying to do something with one of these unsupported endpoints. It looks like this is the first in a series of PRs that will remove all of this cluster manager stuff.

The Jira epic description supports my understanding, which is why I'm comfortable approving.

PR #3649, opened 2024-06-26
Add mirroring of ose-tools-rhel8 for 4.15

razo7:
Add mirroring of ose-tools-rhel8 for 4.15.
ose-tools image is used for troubleshooting scenarios in the wiki, as it comes with the oc client and a few handy tools for investigations.

BTW this week OCP 4.16 will be released and we might want to add 4.16 image of ose-tools-rhel8.


bitoku (approved):

tiguelu (approved):
LGTM

PR #3650, opened 2024-06-26
Exclude platform identities from permissions denial

yithian:

Fixes [ARO-6830](https://issues.redhat.com/browse/ARO-6830)

Add platform workload identities to the list of service principals excluded from the permissions denial so that those identities can manage Azure resources in the cluster's resource group

### Test plan for issue:

Test plan is to PUT/PATCH a PlatformWorkloadIdentityProfile on an existing cluster, PUCM/adminUpdate it, and ensure that the deny assignment is updated correctly. Deny assignment isn't present in dev though so it'll have to be tested in canary.

Docs don't need to be updated, this is part of enabling MI/WI

Very simple logic. The main question mark is what kind type of principal an identity should have when added to the `ExcludePrincipals` list

### DIFF
--- a/pkg/cluster/deploybaseresources_additional.go
+++ b/pkg/cluster/deploybaseresources_additional.go
@@ -20,7 +20,22 @@ import (
 )
 
 func (m *manager) denyAssignment() *arm.Resource {
-	return &arm.Resource{
+	excludePrincipals := []mgmtauthorization.Principal{}
+	if m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile != nil && m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil {
+		for _, identity := range m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
+			excludePrincipals = append(excludePrincipals, mgmtauthorization.Principal{
+				ID:   &identity.ObjectID,
+				Type: to.StringPtr(string(mgmtauthorization.ServicePrincipal)),
+			})
+		}
+	} else {
+		excludePrincipals = append(excludePrincipals, mgmtauthorization.Principal{
+			ID:   &m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID,
+			Type: to.StringPtr(string(mgmtauthorization.ServicePrincipal)),
+		})
+	}
+
+	resource := &arm.Resource{
 		Resource: &mgmtauthorization.DenyAssignment{
 			Name: to.StringPtr("[guid(resourceGroup().id, 'ARO cluster resource group deny assignment')]"),
 			Type: to.StringPtr("Microsoft.Authorization/denyAssignments"),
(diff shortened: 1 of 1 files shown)


pkg/cluster/deploybaseresources_additional.go
```diff
@@ -20,7 +20,23 @@ import (
...
+		excludePrincipals = append(excludePrincipals, mgmtauthorization.Principal{
+			ID:   &m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID,
+			Type: to.StringPtr("ServicePrincipal"),
+		})
+	}
+	if m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile != nil {
+		for _, identity := range m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
+			excludePrincipals = append(excludePrincipals, mgmtauthorization.Principal{
+				ID:   &identity.ObjectID,
+				Type: to.StringPtr(string(mgmtauthorization.ServicePrincipal)),
```
cadenmarchese, line 34: Is there reasoning behind having the `Type` here different than the one we have on line 27?

cadenmarchese (commented):
Mostly lgtm! We should do some manual testing - at a minimum we should make sure that nothing changes with a regular prod cluster install, but we should also ensure that if there are any identities, they are added properly. Your test plan should be roughly accurate.

pkg/cluster/deploybaseresources_additional.go
yithian: My thinking was that the `mgmtauthorization.Principal` for the `ServicePrincipal` would stay the same (aside from indentation) since this patch isn't really touching it. But since the `mgmtauthorization.Principal`s for identities were net new, I would reference the `mgmtauthorization.PrincipalType` const. Pretty much just a "I'm not really touching this part other than moving it" thing

It's a pretty fine hair to split. I can match them up if you'd like

pkg/cluster/deploybaseresources_additional.go
```diff
@@ -20,7 +20,23 @@ import (
...
 func (m *manager) denyAssignment() *arm.Resource {
-	return &arm.Resource{
+	excludePrincipals := []mgmtauthorization.Principal{}
+	if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile != nil {
+		excludePrincipals = append(excludePrincipals, mgmtauthorization.Principal{
+			ID:   &m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID,
+			Type: to.StringPtr("ServicePrincipal"),
+		})
+	}
+	if m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile != nil {
```
rajdeepc2792, line 30: Can we keep the above conditions as `if-else`, checking for MIWI in the similar as we do:-
`if m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile != nil && m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil` and keeping the `&m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID` in the else?

I suggest this as we had no nil check for ServicePrincipalProfile before and IMO we should consider non MIWI cluster using above check as a Service Principal cluster. It would also fail the cluster installation here that way if ServicePrincipalProfile is really nil.

pkg/cluster/deploybaseresources_additional.go
rajdeepc2792: IMO, it is better to get them matched. I am not sure though which one's to use `mgmtauthorization.ServicePrincipal` or `ServicePrincipal`.

pkg/cluster/deploybaseresources_additional.go
yithian: That makes sense. I've amended the PR with this

pkg/cluster/deploybaseresources_additional.go
yithian: I've standardized on `mgmtauthorization.ServicePrincipal` in the amended PR

yithian:
Forgot PRs to the RP (hah!) can't be from a fork. Reopened this as #3650 3653

PR #3652, opened 2024-06-26
[ARO-4562] Migrate DNS to track2 SDK replace dns in cluster.

jonathan34c:

Fixes

Update DNS SDK to track2 SDK and add update in the cluster and DNS. Will work on deployment and replace the old DNS in the next PR.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
-	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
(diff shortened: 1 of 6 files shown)


jaitaiwan (approved):
LGTM. I don't see any errors in tests which is good. It might be good for me to come back later and check references to the manager to make sure there aren't any breakages for things that don't have tests somehow.

mociarain (dismissed):
LGTM

mociarain:
Just ran a rebase to see if that would unstick the E2E stuff. In the pipelines it complained about permissions so my working idea is that the new E2E stuff needs to be included.

jonathan34c:
Rebase and all checks passed

mociarain (dismissed):

pkg/cluster/cluster.go
```diff
@@ -157,6 +158,12 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 		return nil, err
 	}
 
+	options := _env.Environment().EnvironmentCredentialOptions()
+	spTokenCredential, err := azidentity.NewEnvironmentCredential(options)
+	if err != nil {
+		return nil, err
+	}
+
```
bitoku, line 166: `localFPAuthorizer` should be compatible with `fpCredRPTenant`.
I don't think we need to define a new credential.

jonathan34c: I think ```localFPAuthorizer``` uses autorest.Authorizer which is not an track2 sdk. They are doing simular jobs but since we are migrating to track2. we should stick with track2 SDK authorizer (azidentity)

jonathan34c: thanks @bitoku  ! i think it's working.

mociarain (approved):

bitoku (approved):

PR #3653, opened 2024-06-27
Exclude platform identities from permissions denial

yithian:

Fixes [ARO-6830](https://issues.redhat.com/browse/ARO-6830)

Add platform workload identities to the list of service principals excluded from the permissions denial so that those identities can manage Azure resources in the cluster's resource group

### Test plan for issue:

Test plan is to PUT/PATCH a PlatformWorkloadIdentityProfile on an existing cluster, PUCM/adminUpdate it, and ensure that the deny assignment is updated correctly. Deny assignment isn't present in dev though so it'll have to be tested in canary.

Docs don't need to be updated, this is part of enabling MI/WI

Very simple logic. The main question mark is what kind type of principal an identity should have when added to the `ExcludePrincipals` list

### DIFF
--- a/pkg/cluster/denyassignment.go
+++ b/pkg/cluster/denyassignment.go
@@ -17,8 +17,8 @@ func (m *manager) createOrUpdateDenyAssignment(ctx context.Context) error {
 	}
 
 	// needed for AdminUpdate so it would not block other steps
-	if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID == "" {
-		m.log.Print("skipping createOrUpdateDenyAssignment: SPObjectID is empty")
+	if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil && m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile == nil {
+		m.log.Print("skipping createOrUpdateDenyAssignment: ServicePrincipalProfile and PlatformWorkloadIdentityProfile are empty")
 		return nil
 	}
 
--- a/pkg/cluster/denyassignment.go
+++ b/pkg/cluster/denyassignment.go
@@ -17,7 +17,7 @@ func (m *manager) createOrUpdateDenyAssignment(ctx context.Context) error {
 	}
 
 	// needed for AdminUpdate so it would not block other steps
-	if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil && m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile == nil {
+	if !m.doc.OpenShiftCluster.UsesWorkloadIdentity() && m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID == "" {
 		m.log.Print("skipping createOrUpdateDenyAssignment: ServicePrincipalProfile and PlatformWorkloadIdentityProfile are empty")
 		return nil
 	}
--- a/pkg/cluster/denyassignment.go
+++ b/pkg/cluster/denyassignment.go
@@ -26,6 +26,11 @@ func (m *manager) createOrUpdateDenyAssignment(ctx context.Context) error {
 			}
 		}
 	} else {
+		if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil {
+			m.log.Print("skipping createOrUpdateDenyAssignment: ServicePrincipalProfile is empty")
+			return nil
+		}
+
 		if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID == "" {
 			m.log.Print("skipping createOrUpdateDenyAssignment: SPObjectID is empty")
 			return nil
--- a/pkg/cluster/denyassignment.go
+++ b/pkg/cluster/denyassignment.go
@@ -5,6 +5,7 @@ package cluster
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/util/arm"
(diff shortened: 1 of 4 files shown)


pkg/cluster/denyassignment_test.go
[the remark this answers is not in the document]
yithian: That makes sense

It'll require some refactoring of this test but it shouldn't be too hard

yithian:
I've run into a snag testing this in canary: neither `ServicePrincipalProfile` nor `PlatformWorkloadIdentityProfile` are flagged as mutable in the v20240812preview `OpenShiftClusterProperties`, so those fields can't be removed or added to a cluster

In order to test that the deny assignment gets updated, I'll need to set those fields to be mutable in the preview API so that I can remove the service principal profile and add a platform workload identity profile to a test cluster. Obviously we don't want that behavior to go to prod though, so I've put a hold on this PR during testing so that I can be sure to revert the mutability changes before the PR is merged

pkg/cluster/denyassignment_test.go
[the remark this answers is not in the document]
yithian: I ended up creating a unit test for `denyAssignment()` directly rather than through `createOrUpdateDenyAssignment`

pkg/cluster/denyassignment_test.go
```diff
@@ -34,6 +34,14 @@ func TestCreateOrUpdateDenyAssignment(t *testing.T) {
 						SPObjectID: fakeClusterSPObjectId,
 					},
 				},
+				Identity: &api.Identity{
```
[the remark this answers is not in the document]
yithian: I split this doc into two different test cases to assert that they're both valid individually

yithian:
I haven't been able to figure out how to change from a `ServicePrincipalProfile` to a `PlatformWorkloadIdentityProfile` on a cluster in canary, even after tagging those resources as mutable in the API

The unit tests for `denyAssignment()` and `createOrUpdateDenyAssignment()` indicate that the new code is working as expected so I'm going to suggest that we merge these changes in and wait until we have a way to directly provision a MIWI-style cluster for final verification (rather than trying to convert a vanilla cluster to a MIWI one)

pkg/cluster/denyassignment.go
```diff
@@ -17,8 +17,8 @@ func (m *manager) createOrUpdateDenyAssignment(ctx context.Context) error {
 	}
 
 	// needed for AdminUpdate so it would not block other steps
-	if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID == "" {
-		m.log.Print("skipping createOrUpdateDenyAssignment: SPObjectID is empty")
+	if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil && m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile == nil {
```
cadenmarchese, line 20: Can we move over to the shared function instead?
https://github.com/Azure/ARO-RP/blob/master/pkg/api/openshiftcluster.go#L31-L33

yithian: Gladly!

pkg/cluster/denyassignment.go
```diff
@@ -17,8 +17,8 @@ func (m *manager) createOrUpdateDenyAssignment(ctx context.Context) error {
 	}
 
 	// needed for AdminUpdate so it would not block other steps
-	if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID == "" {
-		m.log.Print("skipping createOrUpdateDenyAssignment: SPObjectID is empty")
+	if m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
```
rajdeepc2792, line 20: Is this correct, sorry if I am missing any other discussion.
IMO, the deny assignment will be only skipped whenever the case is similar to:-
!m.doc.OpenShiftCluster.UsesWorkloadIdentity() && `m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID == ""`

pkg/cluster/deploybaseresources_additional.go
```diff
@@ -20,7 +20,22 @@ import (
 )
 
 func (m *manager) denyAssignment() *arm.Resource {
-	return &arm.Resource{
+	excludePrincipals := []mgmtauthorization.Principal{}
+	if m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile != nil && m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil {
```
rajdeepc2792, line 24: Can use `m.doc.OpenShiftCluster.UsesWorkloadIdentity()`?

pkg/cluster/denyassignment.go
yithian: :facepalm: You're right, thanks for catching this

pkg/cluster/deploybaseresources_additional.go
yithian: Yup, thanks for catching this

pkg/cluster/denyassignment.go
```diff
@@ -17,8 +17,8 @@ func (m *manager) createOrUpdateDenyAssignment(ctx context.Context) error {
 	}
 
 	// needed for AdminUpdate so it would not block other steps
-	if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID == "" {
-		m.log.Print("skipping createOrUpdateDenyAssignment: SPObjectID is empty")
+	if !m.doc.OpenShiftCluster.UsesWorkloadIdentity() && m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID == "" {
```
kimorris27, line 20: The implication from the comment above seems to be that if you set an exclusion rule with an empty object ID it will error out and block further admin update steps. So with that in mind, I think this condition should be something like this:

```
if (cluster is WI and any platform WI has an empty object ID) or (cluster is SP and object ID is empty)
```

I can't think of an exact situation where a PWI would have an empty object ID and we'd want to do an admin update despite that, but I don't think it hurts to be safe.

What do you think? If you choose to make this change, it'd also be nice to see unit test cases that cover the different scenarios in which we skip creating/updating the deny assignment.

kimorris27 (commented):
LGTM broadly speaking. I think there's one more case where we could skip creating/updating the deny assignment.

pkg/cluster/denyassignment.go
yithian: If my memory is working, I had to put this in to avoid a nil pointer dereference...?

I've updated the logic to check for those conditions and included unit tests for it

kimorris27 (dismissed):

kimorris27 (approved):

pkg/cluster/denyassignment.go
```diff
@@ -17,9 +18,18 @@ func (m *manager) createOrUpdateDenyAssignment(ctx context.Context) error {
...
-		return nil
+	if m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		for _, i := range m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
+			if i.ObjectID == "" {
+				m.log.Print(fmt.Sprintf("skipping createOrUpdateDenyAssignment: ObjectID for identity %s is empty", i.OperatorName))
+				return nil
+			}
+		}
+	} else {
+		if m.doc.OpenShiftCluster.Properties.ServicePrincipalProfile.SPObjectID == "" {
```
tsatam, line 29: Will there be a risk of nil pointer dereference here? (when both PlatformWorkloadIdentityProfile and ServicePrincipalProfile are nil?)

tsatam (changes requested):
Change mostly looks good to me, this would not appear to impact existing service principal cluster installation/updates.

One small suggestion to guard us against any potential nil pointer dereferences - even if we do not expect to see any here, we should guard ourselves from this happening in case of any bugs elsewhere.

pkg/cluster/denyassignment.go
yithian: There _shouldn't_ be, static validation should prevent a scenario where both PlatformWorkloadIdentityProfile and ServicePrincipalProfile are nil https://github.com/Azure/ARO-RP/blob/master/pkg/api/v20240812preview/openshiftcluster_validatestatic.go#L488-L490

Would you prefer to explicitly check for a nil pointer first?

yithian: I added a check (with associated unit test) to prevent a nil pointer dereference in this case if the static validation doesn't catch this scenario

kimorris27 (approved):

rajdeepc2792 (approved):

cadenmarchese (approved):

tsatam (approved):

PR #3655, opened 2024-06-28
Add tenant ID to internal apis for CMSI usage

niontive:

Part of issue [ARO 8606](https://issues.redhat.com/browse/ARO-8608)

- It's possible for MSI dataplane to not return tenant ID in request for managed identity. We're expected to use the header from ARM, so persist this.
- 
### Test plan for issue:

Unit test

No


niontive:
Using #3514 as inspiration

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
-var errMissingIdentityURL error = fmt.Errorf("identityURL not provided but required for workload identity cluster")
+var errMissingIdentityParmeter error = fmt.Errorf("identity parameter not provided but required for workload identity cluster")
```
kimorris27, line 28: Nit: missing an "a" in "parameter" in the variable name.

kimorris27 (commented):
LGTM! My only comment is the tiniest nit.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -96,9 +98,16 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 		}
 	}
 
-	err = validateIdentityUrl(doc.OpenShiftCluster, identityURL, isCreate)
-	if err != nil {
-		return nil, err
+	// Don't persist identity parameters in non-wimi clusters
+	if doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil || doc.OpenShiftCluster.Identity != nil {
+		if isCreate {
```
cadenmarchese, line 103: Do we need to validate and persist identity URL in cases of update is well? My understanding is that we will still want to persist identity URL and tenant ID even in cases where it's not a new cluster creation, such as for an update, where we need to refresh client IDs for existing identities, or federate new identities. But I'm assuming that the secret URL/identity URL changes.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -96,9 +98,16 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 		}
 	}
 
-	err = validateIdentityUrl(doc.OpenShiftCluster, identityURL, isCreate)
-	if err != nil {
-		return nil, err
+	// Don't persist identity parameters in non-wimi clusters
+	if doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil || doc.OpenShiftCluster.Identity != nil {
```
rajdeepc2792, line 102: To be consistent, the current way of determining the WIMI cluster is:-
`doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil && doc.OpenShiftCluster.PlatformWorkloadIdentityProfile != nil` all other are considered to non-miwi.

pkg/frontend/openshiftcluster_putorpatch.go
rajdeepc2792: If we don't have onboarding a new MSI at the time of update under scope, the identityURL and tenantID is not expected in that case. The credentials fetched via MSI Dataplane which will be persisted in KV will be the trusted source for the refreshing of the credentials.
The identityURL and TenantID will only be validated and persisted here(at frontend) in case of new MSI onboarding which is only scoped for the cluster creation I assume.

niontive: Here's my understanding:

- We won't support updating the cluster MSI. Any attempt to remove, replace or assign additional should return an error to the user.
- Thus, we only need identity URL and tenant ID to fetch the credentials object and store in key vault during cluster creation.

This is in line with Rajdeep's points.

To address Caden's points:
- For any updates to the federated identities, we can just get the credential object from key vault and generate a token to perform federation to new identities.
- I'm not sure what you mean by "refresh client IDs" -> if a customer wants to update the federated identities, this isn't affected by the identity URL/tenant ID persistence. Any new identities we get can be federated by the cluster MSI, and we should have the backing credentials object in key vault.

pkg/frontend/openshiftcluster_putorpatch.go
niontive: Not checking `doc.OpenshiftCluster.Identity` to determine MIWI cluster seems pretty problematic - for example, let's say a customer specifies the workload identities but forgets to specify the cluster MSI. The above check would pass even though the cluster doesn't have a cluster MSI.

I'll write a separate function to determine if an `OpenShiftCluster` is MIWI or not so we have a single source of truth.

pkg/frontend/openshiftcluster_putorpatch.go
cadenmarchese: thanks for the clarification 👍

pkg/frontend/openshiftcluster_putorpatch.go
cadenmarchese: @niontive @rajdeepc2792 we already have a shared function in progress, and the logic is currently `doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil || doc.OpenShiftCluster.Identity != nil`. The idea with || is that we want to account for cases where the cluster doc is not exactly as expected.

niontive: @cadenmarchese I'll revert my change for a shared function and use `doc.OpenShiftCluster.Properties.ServicePrincipalProfile == nil || doc.OpenShiftCluster.Identity != nil`. I'll add a TODO to use as shared function once it's implemented so we don't forget.

kimorris27:
E2E failed with what appear to be flakes. I'm going to investigate a bit and see if I can card up some tickets to address them, but in the meantime, I'll trigger E2E again.

kimorris27:
At least one of the flakes seems to be related to some PIP configs that got into a Failed provisioning state somehow. I also saw HTTP connections that were dropped between the E2E agent and the cluster API. I couldn't find anything in ARM logs to explain what I saw, and the Azure resources from the last E2E run are already gone, so I think we'll have to let this one go.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -44,19 +44,21 @@ func (f *frontend) putOrPatchOpenShiftCluster(w http.ResponseWriter, r *http.Req
...
 		return err
 	})
 
 	frontendOperationResultLog(log, r.Method, err)
 	reply(log, w, header, b, err)
 }
 
-func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.Entry, body []byte, correlationData *api.CorrelationData, systemData *api.SystemData, path, originalPath, method, referer string, header *http.Header, converter api.OpenShiftClusterConverter, staticValidator api.OpenShiftClusterStaticValidator, subId, resourceProviderNamespace string, apiVersion string, identityURL string) ([]byte, error) {
+// TODO - refactor this function to reduce the number of parameters
+func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.Entry, body []byte, correlationData *api.CorrelationData, systemData *api.SystemData, path, originalPath, method, referer string, header *http.Header, converter api.OpenShiftClusterConverter, staticValidator api.OpenShiftClusterStaticValidator, subId, resourceProviderNamespace string, apiVersion string, identityURL string, identityTenantID string) ([]byte, error) {
```
mociarain, line 61: I feel we often leave these types of comments with the best intentions but then never loop back on them.

From just jumping into this I would think we need a little struct here but maybe I'm missing something? I get not adding it here to reduce the PR complexity but could we have a branch off this with the refactor so we know it's all going to go together?

Happy to pair on this too if that helps :)

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -96,9 +98,18 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 		}
 	}
 
-	err = validateIdentityUrl(doc.OpenShiftCluster, identityURL, isCreate)
-	if err != nil {
-		return nil, err
+	if isCreate {
+		// Persist identity URL and tenant ID only for managed/workload identity cluster create
+		// We don't support updating cluster managed identity after cluster creation
+		// TODO - use a common function to check if the cluster is a managed/workload identity cluster
```
mociarain, line 104: Is there a card or something to capture this for MIWI? (@kimorris27 maybe?)

mociarain (commented):
LGTM, just looking for some clarity on the TODOs

pkg/api/openshiftcluster.go
```diff
@@ -808,4 +808,5 @@ type Identity struct {
 	Type                   string                 `json:"type,omitempty"`
 	UserAssignedIdentities UserAssignedIdentities `json:"userAssignedIdentities,omitempty"`
 	IdentityURL            string                 `json:"identityURL,omitempty" mutable:"true"`
+	TenantID               string                 `json:"tenantId,omitempty" mutable:"true"`
```
cadenmarchese, line 811: Should this be a pointer field?
```suggestion
	TenantID               *string                 `json:"tenantId,omitempty" mutable:"true"`
```

pkg/frontend/openshiftcluster_putorpatch.go
kimorris27: Yep, we've got https://issues.redhat.com/browse/ARO-7856 for this.

pkg/api/openshiftcluster.go
kimorris27: It's a regular string field in the AKS RP, so I think we should do the same thing as them to make sure we don't run into issues with the Swagger API spec validation later on.

niontive: I say we should keep this a string. Reasoning:
1. Identity URL is a string, and TenantID is just another MSI header we get from ARM
2. Pointers increase the likelihood of panics via nil pointers.

pkg/frontend/openshiftcluster_putorpatch.go
niontive: I don't have bandwidth at the moment to tackle this refactor. On top of that, I'm not the best person to work on this since I don't own the frontend code.

I'll defer to @bennerv and @hlipsig for this.

kimorris27 (approved):

cadenmarchese (approved):

PR #3660, opened 2024-07-03
Add controller to disable CPMS on ARO clusters

tsatam:

Fixes [ARO-8836](https://issues.redhat.com/browse/ARO-8836)

- [x] Updates openshift/api dependency to release-4.12 (to bring in CPMS struct)
- [x] Adds a new operatorflag to enable/disable CPMSDeactivator controller
- [x] Adds new CPMSDeactivator controller
- [x] Updates SRE RBAC to allow managing CPMS resource

### Test plan for issue:

- [x] Local 4.12 cluster installs w/ new operator and observes no change in behavior
- [x] Local 4.13+ cluster installs w/ new operator and disables CPMS
- [x] Local <=4.11 cluster (before the introduction of the CPMS resource) can be updated to new operator and observes no change in behavior

No 

- Tests performed above 

### Additional notes
We should add an E2E test that ensures CPMS is disabled. I did not do it within this PR as it would require another dependency bump+revendor for openshift/client-go.


tsatam (commented):
Feedback from sync PR review session

go.mod
```diff
@@ -274,7 +274,7 @@ require (
 	k8s.io/apiserver v0.26.2 // indirect
 	k8s.io/component-base v0.26.2 // indirect
 	k8s.io/gengo v0.0.0-20211129171323-c02415ce4185 // indirect
-	k8s.io/klog/v2 v2.120.1 // indirect
+	k8s.io/klog/v2 v2.130.1 // indirect
```
hlipsig, line 277: Non blocker, but why did this need to be bumped?

pkg/operator/flags.go
```diff
@@ -10,6 +10,7 @@ const (
 	AzureSubnetsServiceEndpointManaged = "aro.azuresubnets.serviceendpoint.managed"
 	BannerEnabled                      = "aro.banner.enabled"
 	CheckerEnabled                     = "aro.checker.enabled"
+	CPMSDeactivatorEnabled             = "aro.cpmsdeactivator.enabled"
```
hlipsig, line 13: Does this turn machinesets off or on? Name is confusing can we make it clearer.

hlipsig (changes requested):
Let's talk about the controller name before merging.

go.mod
tsatam: I needed to bring in openshift/api release-4.12 or newer to have access to the ControlPlaneMachineSet struct.

hlipsig: Thanks for the clarification. Not that keeping dependencies current is at all bad, but I didn't see the exact cause immediately in my review so I was just curious.

pkg/operator/flags.go
tsatam: Happy for better suggestions for the name. 

Enabling the controller disables the machineset, which is inherently confusing.

hlipsig (approved):

cmd/aro/operator.go
```diff
@@ -226,6 +233,17 @@ func operator(ctx context.Context, log *logrus.Entry) error {
 			client)).SetupWithManager(mgr); err != nil {
 			return fmt.Errorf("unable to create controller %s: %v", cloudproviderconfig.ControllerName, err)
 		}
+
+		// only register CPMS controller on clusters that support the CRD
+		if err := discovery.ServerSupportsVersion(discoverycli, machinev1.GroupVersion); err == nil {
+			if err = (cpms.NewReconciler(
+				log.WithField("controller", cpms.ControllerName),
+				client)).SetupWithManager(mgr); err != nil {
+				return fmt.Errorf("unable to create controller %s: %v", cpms.ControllerName, err)
```
s-fairchild, line 242: Not really a big deal, but ultimately we want `err` as a string, may as well do it directly IMO.  

```suggestion
				return fmt.Errorf("unable to create controller %s: %s", cpms.ControllerName, err.String())
```

cmd/aro/operator.go
```diff
@@ -226,6 +233,17 @@ func operator(ctx context.Context, log *logrus.Entry) error {
...
+			}
+		} else {
+			log.Infof("Cluster does not support machinev1 API, will not set up CPMS controller: %v", err)
```
s-fairchild, line 245:
```suggestion
			log.Infof("Cluster does not support machinev1 API, will not set up CPMS controller: %s", err.String())
```

pkg/operator/controllers/cpms/cpms_controller.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	if cpms.Spec.State == machinev1.ControlPlaneMachineSetStateInactive {
+		r.Log.Info("CPMS is inactive, nothing to do")
+		return ctrl.Result{}, nil
+	}
+
+	// disable CPMS by deleting it
+	// https://docs.openshift.com/container-platform/4.12/machine_management/control_plane_machine_management/cpmso-disabling.html
+	return ctrl.Result{}, r.Client.Delete(ctx, cpms)
```
s-fairchild, line 81: Should we validate cpms is in an inactive state after deleting it?
I see we do that check beforehand, the next time the controller reconciles it will do the check.  
I'm just concerned about possible scenarios where cpms is getting recreated by something/someone, and our controller will just keep deleting it without raising a red flag.

Idea, maybe we can keep a tally somewhere of how many times this controller has deleted cpms, and if it's more than X(1?) generate an error.

Per the docs:
> Verification
Check the control plane machine set custom resource state. A result of Inactive indicates that the removal and replacement process is successful. A ControlPlaneMachineSet CR exists but is not activated.

s-fairchild (commented):
Great work here, just a few minor suggestions, and thoughts on verifying directly after deleting cpms.

pkg/operator/controllers/cpms/cpms_controller.go
tsatam: In the interest of keeping this controller in-line with controller recommendations (do one thing per reconcile loop), and the fact that this would introduce a dependency on this controller waiting for CPMSO to roll out a new Inactive CPMS, I'd rather not introduce a "wait" into the controller itself. 

The controller will reconcile against the newly-created inactive CPMS, and log that. IMO if we want to gather any data/metrics around CPMS being active/inactive, we should do so in a monitor rather than in this operator controller.

cmd/aro/operator.go
tsatam: This is an interesting suggestion, but it's not in line with the conventions we have been following throughout the codebase (e.g. https://github.com/Azure/ARO-RP/blob/e711e610a0ad352ac11d0e72328170d759638242/pkg/operator/controllers/workaround/workaround_controller.go#L72). I think if we want to do this, it should be done in a dedicated "style-fixup" PR.

pkg/operator/controllers/cpms/cpms_controller.go
```diff
@@ -24,33 +24,34 @@ import (
 // Licensed under the Apache License 2.0.
 
 const (
-	DeactivatorControllerName = "CPMSDeactivatorController"
-	SingletonCPMSName         = "cluster"
-	SingletonCPMSNamespace    = "openshift-machine-api"
+	ControllerName         = "CPMSDeactivatorController"
```
tsatam, line 27: rename this

cmd/aro/operator.go
SudoBrendan: a lot of our code uses `%v`, but it is rare that `%v` or `%s` are best here, we should use `%w` unless we care for some reason that the Error type we return is distinct from `err` itself: https://stackoverflow.com/questions/61283248/format-errors-in-go-s-v-or-w

`%s` vs `%v` is also discussed in that thread - the biggest concern is if `err` is nil, which shouldn't happen...?

pkg/operator/controllers/cpms/cpms_controller.go
```diff
@@ -0,0 +1,100 @@
...
+	return &Reconciler{
+		AROController: base.AROController{
+			Log:    log,
+			Client: client,
+			Name:   ControllerName,
+		},
+	}
+}
+
+// CPMS reconciler will disable the cluster CPMS if `aro.cpms.enabled` is false or missing.
```
SudoBrendan, line 46: I think this comment is the perfect place in all this code to _not_ use `CPMS` and explicity state it as `Control Plane Machine Set` so we codify the meaning we are after very clearly without having to read the whole function and flow of the operator/controller.

pkg/operator/flags.go
```diff
@@ -10,6 +10,7 @@ const (
 	AzureSubnetsServiceEndpointManaged = "aro.azuresubnets.serviceendpoint.managed"
 	BannerEnabled                      = "aro.banner.enabled"
 	CheckerEnabled                     = "aro.checker.enabled"
+	CPMSEnabled                        = "aro.cpms.enabled"
```
SudoBrendan, line 13: nit: this may become a confusing name in a world where we _do_ have OCP CPMS enabled en-masse for the fleet. We should probably use a name like `CPMSDisablerEnabled`

pkg/operator/flags.go
```diff
@@ -47,6 +48,7 @@ func DefaultOperatorFlags() map[string]string {
 		AzureSubnetsServiceEndpointManaged: FlagTrue,
 		BannerEnabled:                      FlagFalse,
 		CheckerEnabled:                     FlagTrue,
+		CPMSEnabled:                        FlagFalse,
```
SudoBrendan, line 51: nit: same naming issues as above.

SudoBrendan (changes requested):
Overall LGTM - ping me in slack when the new names are here it might clarify my comments up too.

pkg/operator/controllers/cpms/cpms_controller.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/Azure/ARO-RP/pkg/operator"
+	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+	"github.com/Azure/ARO-RP/pkg/operator/controllers/base"
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+const (
+	ControllerName         = "CPMSController"
```
SudoBrendan, line 27: nit: similar to my comments below, I think `CPMSDisablerController` is a better name, more future-proof to when we use CPMS across the fleet (if that happens)

pkg/operator/flags.go
tsatam: I had that before and we decided to move away from that name due to it being confusing the other way round: https://github.com/Azure/ARO-RP/pull/3660/commits/c44d92065a0917ba667ec7f830453e746f226738

SudoBrendan (changes requested):

SudoBrendan (approved):
Overall LGTM - feel free to address and/or close any remaining comments.

pkg/operator/controllers/cpms/cpms_controller.go
tsatam: We plan to use this controller to operationalize CPMS once we're ready for it.

pkg/operator/controllers/cpms/cpms_controller.go
```diff
@@ -0,0 +1,100 @@
...
+func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	instance, err := r.GetCluster(ctx)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if instance.Spec.OperatorFlags.GetSimpleBoolean(operator.CPMSEnabled) {
+		r.Log.Infof("Flag %s is true, will not deactivate CPMS", operator.CPMSEnabled)
+		return reconcile.Result{}, nil
+	}
```
SudoBrendan, line 56: nit: might be worth a comment here stating that this controller is unique in that `enabled: true` means that CPMS sticks around, and `enabled: false` means that we will deactivate it. I think it is highly clear once you see it - but it is also distinct.

SudoBrendan: alternatively, I see the comment in the header of this function, doing some `**false**` formatting or something may draw attention to the uniqueness.

cmd/aro/operator.go
s-fairchild: Thanks for the link. I forgot there was an error formatter, that would be ideal.

tsatam: The entirety of cmd/aro/operator.go is a prime candidate for some `%w` replacement on error wrapping - I think this should be done in a follow-up PR across _all_ of our controller instantiation.

kimorris27 (approved):
LGTM, but it would be nice to see an E2E test in a follow-up PR.

hlipsig (approved):
Great discussions. LGTM as is and the improvements we can card up for the future.

PR #3666, opened 2024-07-04
[ARO-8862] local-rp dev docs should mention registering the sub

mrWinston:

Fixes: https://issues.redhat.com/browse/ARO-8862

Using a local RP to create a cluster requires registering the subscription to the RP. Following the local development RP docs however yields in this error when trying to spin up a new cluster:

```
(InvalidSubscriptionState) Request is not allowed in unregistered subscription '...'.
```

This PR makes sure that registering the subscription to the local RP is explicitly mentioned.

### DIFF
--- a/docs/deploy-development-rp.md
+++ b/docs/deploy-development-rp.md
@@ -93,29 +93,47 @@
    make runlocal-rp
    ```
 
-1. To create a cluster, EITHER follow the instructions in [Create, access, and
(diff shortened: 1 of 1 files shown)


mrWinston:
@microsoft-github-policy-service agree company="RedHat"

rolandmkunkel (approved):

tiguelu (approved):

PR #3667, opened 2024-07-04
Smoke test automation for some basic checks

bitoku:

Fixes https://issues.redhat.com/browse/ARO-7343, https://issues.redhat.com/browse/ARO-7344, https://issues.redhat.com/browse/ARO-7351.

This PR automates some of smoke tests which we always run when we release the new OCP version installer or gap analysis.
With this change, we can run smoke tests by running `E2E_LABEL=smoke make test-e2e`.

### Test plan for issue:

Run locally to check the functionality.
Run e2e to confirm that this test doesn't interfere the existing e2e tests. 

Run locally.
Run e2e on CI.

### DIFF
--- a/test/e2e/ocp_cluster.go
+++ b/test/e2e/ocp_cluster.go
@@ -1,0 +1,81 @@
+package e2e
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"slices"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Cluster Operators", Label(smoke), func() {
+	It("should be all available", func(ctx context.Context) {
+		By("checking that all cluster operators are available")
+		cos, err := clients.ConfigClient.ConfigV1().ClusterOperators().List(ctx, metav1.ListOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		for _, co := range cos.Items {
+			available := false
+			for _, condition := range co.Status.Conditions {
+				if condition.Type == configv1.OperatorAvailable {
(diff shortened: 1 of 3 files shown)


test/e2e/setup.go
```diff
@@ -118,6 +122,12 @@ func skipIfNotHiveManagedCluster(adminAPICluster *admin.OpenShiftCluster) {
 	}
 }
 
+func skipUnlessSmokeTestEnabled() {
+	if os.Getenv("ARO_E2E_SMOKE_TEST") == "" {
+		Skip("skipping smoke tests")
+	}
+}
+
```
mociarain, line 130: Rather than skipping these tests based on a env var what if we just ran the e2e tests with a `--label-filter="!smoke"`. Equally, we could label all the tests as `E2E` and run with `--label-filter="E2E"` but that seems like overkill.

What do you think?

bitoku: nice, it would be simpler.

https://onsi.github.io/ginkgo/#spec-labels

test/e2e/ocp_cluster.go
```diff
@@ -0,0 +1,81 @@
...
+		for _, co := range cos.Items {
+			available := false
+			for _, condition := range co.Status.Conditions {
+				if condition.Type == configv1.OperatorAvailable {
+					Expect(condition.Status).To(Equal(configv1.ConditionTrue), "operator %s is not available", co.Name)
+					available = true
+					break
+				}
+			}
+			Expect(available).To(BeTrue(), "operator %s is not available", co.Name)
```
kimorris27, line 34: Isn't this duplicating the `Expect` check that's being done on line 29? It seems like we could remove the `available` variable and all references to it.

kimorris27 (commented):
LGTM except for one spot where it seems to me that we're duplicating a particular check.

test/e2e/ocp_cluster.go
kimorris27: Oh, I see what you mean. I don't think that's thinking too much.

With that said, I found it a bit confusing that there are two `Expect`s, since we're logically checking one thing: whether or not the operator is available.

What do you think of replacing lines 29-30 with `available = (condition.Status == configv1.ConditionTrue)`? This is totally a matter of opinion, so the code is also fine as-is if you think it's more readable that way.

kimorris27 (dismissed):

mociarain (approved):
LGTM

PR #3668, opened 2024-07-05
[ARO-7788] Don't overwrite default installer version in local dev

mrWinston:
Fixes https://issues.redhat.com/browse/ARO-7788

Currently, the dev cluster creation tooling always makes sure that the default openshift version together with the default installer image is written to the cosmos db before creating the cluster. This however makes it impossible to use a custom installer image with the default openshift version, as any modifications to the installer image are overwritten.

This PR changes this behaviour so that the default version and installer image is only written when it is not already present.

### Test plan for issue:

- testing was performed according to instructions in jira ticket.

- No
- 

- this doesn't affect production.

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -461,7 +461,7 @@ func (c *Cluster) createCluster(ctx context.Context, vnetResourceGroup, clusterN
 			return err
 		}
 
-		err = c.insertDefaultVersionIntoCosmosdb(ctx)
+		err = c.ensureDefaultVersionInCosmosdb(ctx)
 		if err != nil {
 			return err
 		}
@@ -484,7 +484,7 @@ func (c *Cluster) createCluster(ctx context.Context, vnetResourceGroup, clusterN
 	return c.openshiftclusters.CreateOrUpdateAndWait(ctx, vnetResourceGroup, clusterName, ocExt)
 }
 
-var localClient *http.Client = &http.Client{
+var insecureLocalClient *http.Client = &http.Client{
 	Transport: &http.Transport{
 		TLSClientConfig: &tls.Config{
 			InsecureSkipVerify: true,
@@ -518,7 +518,7 @@ func (c *Cluster) registerSubscription(ctx context.Context) error {
 
 	req.Header.Set("Content-Type", "application/json")
 
-	resp, err := localClient.Do(req)
+	resp, err := insecureLocalClient.Do(req)
 	if err != nil {
 		return err
 	}
@@ -526,7 +526,8 @@ func (c *Cluster) registerSubscription(ctx context.Context) error {
 	return resp.Body.Close()
 }
 
-// getVersionsInCosmosDB
+// getVersionsInCosmosDB connects to the local RP endpoint and queries the
+// available OpenShiftVersions
 func getVersionsInCosmosDB(ctx context.Context) ([]*api.OpenShiftVersion, error) {
 	type getVersionResponse struct {
 		Value []*api.OpenShiftVersion `json:"value"`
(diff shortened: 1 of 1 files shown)


pkg/util/cluster/cluster.go
```diff
@@ -484,6 +484,16 @@ func (c *Cluster) createCluster(ctx context.Context, vnetResourceGroup, clusterN
...
 
+var insecureLocalClient *http.Client = &http.Client{
+	Transport: &http.Transport{
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
+		},
+	},
+}
+
+var localDefaultURL string = "https://localhost:8443"
```
mociarain, line 495: What `var` and not `const`?

pkg/util/cluster/cluster.go
```diff
@@ -501,30 +511,68 @@ func (c *Cluster) registerSubscription(ctx context.Context) error {
...
 	if err != nil {
 		return err
 	}
 
 	return resp.Body.Close()
 }
 
-func (c *Cluster) insertDefaultVersionIntoCosmosdb(ctx context.Context) error {
+// getVersionsInCosmosDB connects to the local RP endpoint and queries the
+// available OpenShiftVersions
```
mociarain, line 530: Pure curiosity:

I'm still new to go and one of the things I find very odd is the lack of recommended line lengths. Here you seem to have tightened your comments to ~80 but longer in the code. Is there a reason for this/some convention I'm unaware of?

pkg/util/cluster/cluster.go
```diff
@@ -501,30 +511,68 @@ func (c *Cluster) registerSubscription(ctx context.Context) error {
...
+}
+
+// ensureDefaultVersionInCosmosdb puts a default openshiftversion into the
+// cosmos DB IF it doesn't already contain an entry for the default version. It
+// is hardcoded to use the local-RP endpoint `https://localhost:8443`
+//
+// It returns without an error when a default version is already present or a
+// default version was successfully put into the db.
+func (c *Cluster) ensureDefaultVersionInCosmosdb(ctx context.Context) error {
+	// first, we make sure that we don't overwrite any existing entry for the default version
```
mociarain, line 562: Do we need this comment?

pkg/util/cluster/cluster.go
```diff
@@ -501,30 +511,68 @@ func (c *Cluster) registerSubscription(ctx context.Context) error {
...
+	versionsInDB, err := getVersionsInCosmosDB(ctx)
+	if err != nil {
+		return fmt.Errorf("couldn't query versions in cosmosdb: %w", err)
+	}
+
+	for _, versionFromDB := range versionsInDB {
+		// we already have an entry for the default version
```
mociarain, line 569: Do we need this comment?

mociarain (commented):
Nice change. Just a few questions

pkg/util/cluster/cluster.go
mrWinston: Good point, `const` would indeed be the better choice here.

pkg/util/cluster/cluster.go
mrWinston: No, i don't think there's a convention regarding that. I wrap a line when it feels like it's getting too long, and for godoc comments i just like the look of them them being wrapped :)

pkg/util/cluster/cluster.go
mociarain: No worries. I'm borderline anti-comment so I challenge them often. Been bitten by stale comments too many times in the past :D

mociarain (approved):

tiguelu (approved):

tiguelu (changes requested):
@mrWinston Sorry, E2E cannot run in the PR because the source branch is in a fork of the ARO-RP. Unfortunately, we have this restriction for security reasons, so you will need to create a new PR using a branch created directly in the ARO-RP repo.

I will close this PR.

Sorry for that, and thank you for fixing this bug!

PR #3670, opened 2024-07-07
Initial start of RP int automation

razo7:

Relates to ARO-7755

The PR introduces an initial creation of an ADO pipeline to automate the full ARO-RP int-like deployment:

- `rp-full-dev-setup.yml` splits the list of commands from https://github.com/Azure/ARO-RP/blob/master/docs/deploy-full-rp-service-in-dev.md#deploying-an-int-like-development-rp into ADO pipeline steps where some are simply bash scripts and some are new [templates](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops&pivots=templates-includes) (under `.pipeline/templates/rp-dev/...`) which could be reused for other pipelines. 
- Parameters file, `rp-dev-params.yml`, with the parameters that a user might want to tweak
- Variables file, `rp-dev-vars`, with all the variables that should remain constant. 
- Bash functions for getting and copying the right images to the ACR under `hack/devtools/rp-dev-helper.sh`
- Support a new env var for Azure resource names prefix - #3681 

This new pipeline is just an initial start that will be used for creating ADO Onebranch pipelines and afterward for local debugging.

### Test plan for issue:
None

New documentation will be created in a follow up PR for how to interact and modify the new pipeline.


.pipelines/rp-int-deploy.yml
```diff
@@ -0,0 +1,88 @@
+trigger: none
```
tiguelu, line 1: Should we rename this file to something like `rp-full-dev-setup.yaml` to avoid confusion? Otherwise, it sounds like we are deploying INT, for which there are pipelines in the ADO repo.

If it that makes sense, I would avoid references to int for that same reason in the file content. Same for the name of the rp-int-deployment directory under templates.

Maybe we should rename too the files rp-params.yml and rp-vars.yml since there are other pipelines in this folder deploying RPs but they won't be using these files.

.pipelines/rp-int-deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+        displayName: ⚙️ Fully deploy ARO RP integreation
+      - script: |
+          source ./hack/devtools/deploy-shared-env.sh
+          deploy_oic_for_dedicated_rp
+        name: create-sa-role
+        displayName: ⚙️ Create storage account and role assignment
+## few more steps would be added
+
+# cleanup
+      - template: ./templates/template-clean-subscription.yml
```
tiguelu, line 80: Isn't this the template we use in the pipeline that cleans dev clusters in the RH tenant? If so, why is it being used here?

.pipelines/rp-int-deploy.yml
cadenmarchese: Agreed, we can use "int-like" if we have to, but we usually just refer to it as a full RP environment so rp-full-dev-setup.yaml makes sense.

.pipelines/rp-int-deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+jobs:
+  - job: RP_int_deployment
+    timeoutInMinutes: 180
+    pool:
+      name: 1es-aro-ci-pool
+
+    steps:
+# setup
+      - template: ./templates/template-checkout.yml
+      - template: ./templates/template-az-cli-login.yml # does it have az CLI alraedy?
```
cadenmarchese, line 26: It should already have az, hopefully. I checked over our pipelines (such as e2e) and we use `az` there, and it doesn't appear to be explicitly installed by the pipeline.

.pipelines/rp-int-deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+      - template: ./templates/rp-int-deployment/template-download-secrets.yml
+      - template: ./templates/rp-int-deployment/template-generate-rp-dev-config.yml
+        parameters:
+          location: ${{ parameters.location }}
+          resourceGroup: ${{ parameters.resourceGroup }}
+          databaseAccountName: ${{ parameters.databaseAccountName }}
+          keyvaultPrefix: ${{ parameters.keyvaultPrefix }}
+          aroImage: ${{ parameters.aroImage }}
+          fluentbitImage: ${{ parameters.fluentbitImage }}
+      - script: make deploy # use different target which will succeed and won't fail
```
cadenmarchese, line 38: We should have a predeploy pipeline / "bootstrap" pipeline around somewhere in ev2/ARO-Pipelines that predeploys resources such that make deploy doesn't fail. It's for production use and I'm not sure exactly where it lives, but it might be worth tracking down for the possibility of re-use (or at least figuring out what resources need to be pre-created and how) if you don't want to go down the path of retrying / looping `make deploy` knowing that it will fail in the middle.

.pipelines/rp-params.yml
```diff
@@ -0,0 +1,40 @@
...
+    default: $(UNIQUE_IDENTIFIER)-aro-${{ parameters.location }}
+  - name: keyvaultPrefix
+    type: string
+    default: $(UNIQUE_IDENTIFIER)-aro-${{ parameters.location }}
+  - name: dst_acr_name
+    type: string
+    default: $(UNIQUE_IDENTIFIER)aro
+  - name: aroImage
+    type: string
+    default: ${{ parameters.dst_acr_name }}.azurecr.io/aro:$(git rev-parse --short=7 HEAD) # do we have git?
```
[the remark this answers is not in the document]
razo7: In theory yes, but in practice I came to realize that [VERSION](https://github.com/Azure/ARO-RP/blob/master/Makefile#L105C75-L105C82) isn't neccrially the commit. When the git commit also includes a git tag then [VERSION gets the git tag value instead of the git commit](https://github.com/Azure/ARO-RP/blob/master/Makefile#L27-L31), and this causes some issues when the aro image tag isn't aligned with the deployment version

.pipelines/rp-params.yml
```diff
@@ -0,0 +1,40 @@
...
+  - name: fluentbitImageTag
```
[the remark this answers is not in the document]
razo7: Sounds better and algins better with the code validation.

.pipelines/rp-params.yml
```diff
@@ -0,0 +1,40 @@
+parameters:
+  - name: gitCommit
```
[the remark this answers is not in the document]
razo7: I guess you are right, but I think @cadenmarchese had a request/concern about it in the epic. I am fine with only gitBranch as long as we don't miss any important functionality/usability

.pipelines/rp-int-deploy.yml
razo7: I though it could be used for cleaning all the resources the pipeline is currently creating. But I guess it should be done in a different way..

.pipelines/rp-int-deploy.yml
razo7: I assume it comes from `registry.access.redhat.com/ubi8/toolbox:8.8`. Yesterday @tony-schndr @rolandmkunkel were wondering about it as well.

.pipelines/rp-int-deploy.yml
razo7: Sounds interesting

.pipelines/rp-params.yml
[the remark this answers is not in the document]
cadenmarchese: I think branch works just the same!

.pipelines/rp-params.yml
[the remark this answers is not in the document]
razo7: I have added new functions and scripts to get that and set it nicely for importing and copying images by digest for the fluenbit image.

.pipelines/templates/rp-int-deployment/template-push-aro-fluenbit-to-acr.yml
```diff
@@ -0,0 +1,15 @@
+parameters:
+  fluentbitImageTag: ""
+  fluentbitImage: ""
+steps:
+  - script: |
+      make publish-image-aro-multistage
```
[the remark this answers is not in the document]
razo7: >  I see it's run in a previous step to deploy OCP images, but if there's a dependency, would it be better to make the login an explicit step instead of inserting it within a prior step?

I am separating it into a different template, and in a follow up I will try to add a dependency by 
> dependsOn: acr-login
  condition: failed()

where `condition` will receive the output of the `acr-login` template.

UlrichSchlueter (dismissed):
LGTM

Let's get this going

UlrichSchlueter:
trying to get the checks green, happy to merge without e2e going green

.pipelines/rp-full-dev-setup.yml
```diff
@@ -0,0 +1,81 @@
...
+          resourceGroup: ${{ parameters.resourceGroup }}
+          databaseAccountName: ${{ parameters.databaseAccountName }}
+          keyvaultPrefix: ${{ parameters.keyvaultPrefix }}
+          aroImage: ${{ parameters.aroImage }}
+      - script: make deploy # use different target which will succeed and won't fail
+        name: semi-deploy-aro-rp-dev
+        displayName: ⚙️ Run pre-deplyments tasks prior to AKS
+      - script: source ./hack/devtools/deploy-shared-env.sh && deploy_vpn_for_dedicated_rp
+        name: deploy-vpn-gateway
+        displayName: ⚙️ Deploy VPN Gateway
```
SudoBrendan, line 44: q: what's the expected way for a developer to get the certificates to connect to the VPN resource?

.pipelines/templates/rp-dev/rp-dev-vars.yml
```diff
@@ -0,0 +1,8 @@
+variables:
+  AZURE_UNIQUE_PREFIX: zzz # dummy value
+  NO_CACHE: false
```
SudoBrendan, line 3: we likely want this to be true in a pipeline run

.pipelines/templates/rp-dev/rp-dev-vars.yml
```diff
@@ -0,0 +1,8 @@
...
+  AZURE_EXTENSION_DEV_SOURCES: "$(pwd)/python"
```
SudoBrendan, line 4: Is there a reason for us to use the ARO-RP dev extension for deploying an int-like environment, or can we get away with production `az aro`? I don't think any commands for setting up an int-like rely on the `az aro` commandlet?

.pipelines/templates/rp-dev/rp-dev-vars.yml
```diff
@@ -0,0 +1,8 @@
...
+  ARO_INSTALL_VIA_HIVE: true
+  ARO_ADOPT_BY_HIVE: true
+  ARO_SKIP_PKI_TESTS: true
```
SudoBrendan, line 7: I think we can eliminate this, since it's only used for E2E test suite runs, which don't happen as part of deployment, correct?

.pipelines/templates/rp-dev/template-download-secrets.yml
```diff
@@ -0,0 +1,23 @@
...
+steps:
+  - script: |
+      # Export the SECRET_SA_ACCOUNT_NAME environment variable
+      export SECRET_SA_ACCOUNT_NAME=${{ parameters.secretSA }}
+      # Run the make secrets command
+      make secrets
+
+      # Define the expected directory and file names
+      expected_dir="secrets"
+      expected_env_file="env"
```
SudoBrendan, line 13: I don't think this file will exist after a checkout of master.

.pipelines/templates/rp-dev/rp-dev-vars.yml
razo7: I agree, it can be discarded. I hope we don't miss something that happens automatically.

Having said that this automation doesn't include the OCP cluster creation, but if/when it does, then I believe the above extension will be required as it is used in https://github.com/Azure/ARO-RP/blob/master/docs/deploy-full-rp-service-in-dev.md#deploy-a-cluster step 8 (`az aro create \`).

.pipelines/templates/rp-dev/rp-dev-vars.yml
razo7: I think that's accurate. Would it make sense to discard it as well in https://github.com/Azure/ARO-RP/blob/master/env.example#L6?

.pipelines/templates/rp-dev/template-download-secrets.yml
razo7: 1. IIUC the `make secrets` will create a new directory with some secrets/configuration files using the shred RP. The file that we are talking about (path= `secrets/env`) is supposed to be generated by the above target. Thus I don't expect it to be on the _**master**_ branch.
2. The file will remain after _git checkout BRANCH_ due to https://github.com/Azure/ARO-RP/blob/master/.gitignore#L23

.pipelines/rp-full-dev-setup.yml
razo7: The developer may use his Azure account to log in and search the certificates in the Azure subscription, but it might be trickier than that.

.pipelines/rp-full-dev-setup.yml
```diff
@@ -0,0 +1,94 @@
...
+
+      # 8) Final preperation before deploying RP
+      - template: ./templates/rp-dev/template-prepare-rp-deploy.yml
+        parameters:
+          rpResourceGroup: ${{ parameters.resourceGroup }}
+          rpVmssName: rp-vmss-${{ parameters.gitCommit }}
+          gwyResourceGroup: $(AZURE_UNIQUE_PREFIX)-gwy-${{ parameters.location }}
+          gwyVmssName: gateway-vmss-${{ parameters.gitCommit }}
+
+      # 9) Fully delpoy RP
```
tony-schndr, line 80: typo `deploy`

.pipelines/templates/rp-dev/template-generate-rp-dev-config.yml
```diff
@@ -0,0 +1,44 @@
...
+  - script: |
+      set -e
+      # Export environment variables from parameters
+      export LOCATION=${{ parameters.location }} RESOURCEGROUP=${{ parameters.resourceGroup }} DATABASE_ACCOUNT_NAME=${{ parameters.databaseAccountName }}
+      export KEYVAULT_PREFIX=${{ parameters.keyvaultPrefix }} ARO_IMAGE=${{ parameters.aroImage }}
+      # Source environment variables from the secrets file
+      source secrets/env
+
+      # Generate SSH key
+      ssh-keygen -t rsa -N "" -f ~/.ssh/id_rsa
```
tony-schndr, line 36: I think this key is being used to ssh to RP/Gateway right?  I don't see this being exported for SRE use outside the pipeline so we wouldn't have ssh access to the RP/Gateway.  We could add a new ssh pub/priv key to `rharosecretsdev` and use that here instead, that would give everyone access to ssh to the RP/Gateway VMs.

tony-schndr (commented):
Nice work here!

.pipelines/templates/rp-dev/template-generate-rp-dev-config.yml
razo7: Gotcha sounds like you just resolved @SudoBrendan [question](https://github.com/Azure/ARO-RP/pull/3670#discussion_r1678286758) :)
IIUC then your suggestion is something that should be done with shared RP creation? If so, then it isn't something we should do in this PR

.pipelines/rp-full-dev-setup.yml
```diff
@@ -0,0 +1,94 @@
...
+          mirrorTag: ${{ parameters.ocpVersionTag }}
+          dst_acr_name: ${{ parameters.dst_acr_name }}
+
+      # 8) Final preperation before deploying RP
+      - template: ./templates/rp-dev/template-prepare-rp-deploy.yml
+        parameters:
+          rpResourceGroup: ${{ parameters.resourceGroup }}
+          rpVmssName: rp-vmss-${{ parameters.gitCommit }}
+          gwyResourceGroup: $(AZURE_UNIQUE_PREFIX)-gwy-${{ parameters.location }}
+          gwyVmssName: gateway-vmss-${{ parameters.gitCommit }}
```
tiguelu, line 78: There are two references to gitCommit here (which was removed from parameters in favour of gitBranch).

.pipelines/templates/rp-dev/template-generate-rp-dev-config.yml
tony-schndr: @tiguelu is correct.  We can do this in a few steps:
1. Make sure you have the latest secrets `SECRET_SA_ACCOUNT_NAME=rharosecretsdev make secrets`
2. Outside of the pipeline create a SSH priv/pub key (one time only or any time to renew / rotate the keys)
3. Move the public / private key to `secrets/`
4. Update `secrets/env` file with:
```
export SSH_PRIVATE_KEY=secrets/<whatever the Priv key file path is>
export SSH_PUBLIC_KEY=secrets/<whatever the pub key file path is>
```
5. Run `SECRET_SA_ACCOUNT_NAME=rharosecretsdev make secrets-update`, this will update the storage blob, overwriting the previous version so use caution :).

You should be able to remove `ssh-keygen ...` from the pipeline and it should just work, `make dev-config.yaml` should see the `SSH_PUBLIC_KEY` environment variable and

hack/devtools/rp-dev-helper.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash -e
+######## Helper file to run full RP dev either locally or using Azure DevOps Pipelines ########
+
+# Function to extract the image tag - (FUNCTION_NAME)
+extract_image_tag() {
```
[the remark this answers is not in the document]
razo7: >Scraping the tags from const.go with grep would break if those functions or the patterns get changed, renamed or removed.

Indeed, it would require some maintenance and extra care for this changes in const.go. I am not sure how different it is then a possible change to Makefile targets that the suggested pipeline will use (e.g., `deploy` and `secrets`).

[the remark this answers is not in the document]
razo7: > For the sake of simplicity, lets do this as a follow-up task. I will raise a jira to review this, if we don't have one already.

https://issues.redhat.com/browse/ARO-9238 👍🏻

.pipelines/templates/rp-dev/template-generate-rp-dev-config.yml
razo7: SGTM I will test in a follow-up, thank you @tony-schndr!

tiguelu (approved):

PR #3676, opened 2024-07-10
Run Deploy Target with DEPLOY_COMMIT

razo7:

Fixes a bug when running `make deploy` as part of [Deploying an int-like Development RP](https://github.com/Azure/ARO-RP/blob/master/docs/deploy-full-rp-service-in-dev.md#deploying-an-int-like-development-rp) and it fails on pulling unknown aro image error (USER=razo) as follows.

`Writing manifest to image destination ++ docker pull razoaro.azurecr.io/aro:v20240614.00 Trying to pull razoaro.azurecr.io/aro:v20240614.00... Error: initializing source docker://razoaro.azurecr.io/aro:v20240614.00: reading manifest v20240614.00 in razoaro.azurecr.io/aro: manifest unknown: manifest tagged by "v20240614.00" is not found '. More information on troubleshooting is available at https://aka.ms/VMExtensionCSELinuxTroubleshoot.`

It was also mentioned on [Slack thread](https://redhat-internal.slack.com/archives/C02ULBRS68M/p1718714122181379).

- The PR is checking **FULL_RP_DEV** env value to decide which commit value to use when running the RP and GWY deployment. When the env is set use always use COMMIT, otherwise use VERSION which can be COMMIT/TAG.

- When the current commit also includes a git tag (e.g., [commit](https://github.com/Azure/ARO-RP/commit/c87a7bed3afaa5e8264dad8e3577b93edc3f75b6) `c87a7bed3` has a tag `v20240614.00`), then following an int-like RP deployment we set the **VERSION** var to be **TAG** and not **COMMIT**.  The _env-int_ file sets the ARO image based on the commit, and later when we push it to the ACR then it uses this env var _ARO_IMAGE_.
Therefore, on `make deploy` we normally use the commit as the version that would match the pushed aro image (and probably other places), and when we don't then we try to pull a missing image from the user ACR and fail on pulling an unknown aro image.

References:

- TAG - https://github.com/Azure/ARO-RP/blob/master/Makefile#L2
- When TAG isn't empty set VERSION as TAG - https://github.com/Azure/ARO-RP/blob/master/Makefile#L24-L28
- ARO_IMAGE - https://github.com/Azure/ARO-RP/blob/master/env-int.example#L9
- Deploy target - https://github.com/Azure/ARO-RP/blob/master/Makefile#L102

### Test plan for issue:
None

None

I have applied it manually, and it resolved the issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,13 @@ else
 	REGISTRY = $(RP_IMAGE_ACR)
 endif
 
+# (workaround) use git commit sha without dirty suffix for FULL RP Dev automation
+ifeq ($(FULL_RP_DEV),)
+	DEPLOY_COMMIT = $(VERSION)
+else
+	DEPLOY_COMMIT = $(shell git rev-parse --short=7 HEAD)
+endif
+
 # prod images
 ARO_IMAGE ?= $(ARO_IMAGE_BASE):$(VERSION)
 GATEKEEPER_IMAGE ?= ${REGISTRY}/gatekeeper:$(GATEKEEPER_VERSION)
@@ -92,7 +99,7 @@ client: generate $(GOIMPORTS)
 # override COMMIT.
 .PHONY: deploy
 deploy:
-	go run -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=$(VERSION)" ./cmd/aro deploy dev-config.yaml ${LOCATION}
+	go run -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=$(DEPLOY_COMMIT)" ./cmd/aro deploy dev-config.yaml ${LOCATION}
 
 .PHONY: dev-config.yaml
 dev-config.yaml:


razo7:
Since the **VERSION** of Makefile is also used in different places, then this current fix could be a potential workaround and not a full fix.  
I also don't know whether it will affect any other jobs/workflows.

Other solutions to consider could be:

- Modify the `env-int.example` file to set ARO_IMAGE _tag_ similarly to how it is done in the Makefile
- Modify the Makefile always to use the commit over a tag for VERSION var.

Makefile
```diff
@@ -102,7 +102,7 @@ ci-clean:
 # override COMMIT.
 .PHONY: deploy
 deploy:
-	go run -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=$(VERSION)" ./cmd/aro deploy dev-config.yaml ${LOCATION}
+	go run -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=$(COMMIT)" ./cmd/aro deploy dev-config.yaml ${LOCATION}
```
tiguelu, line 105: I think we should not do that because in several parts of our prod pipelines we reference `make deploy`, so we're using this in prod with the TAG and we would likely break things.

razo7: I added a condition on FULL_RP_DEV which will be set whenever we wish to always use COMMIT. It would be useful for the full RP automation #3764

jaitaiwan (changes requested):
LGTM but I think this PR will also need to include documentation updates (deploy shared rp md or something like that) because it won't be clear initially that this functionality is available from the outset.

PR #3677, opened 2024-07-10
CIF: OSProvisioningTimedOut

rhamitarora:

[Fixes](https://issues.redhat.com/browse/ARO-5419)


bitoku (dismissed):

pkg/hive/failure/reasons.go
```diff
@@ -36,3 +37,12 @@ var AzureInvalidTemplateDeployment = InstallFailingReason{
 		regexp.MustCompile(`"code":\w?"InvalidTemplateDeployment"`),
 	},
 }
+
+var AzureOSProvisioningTimedOut = InstallFailingReason{
+	Name:    "AzureOSProvisioningTimedOut",
+	Reason:  "AzureOSProvisioningTimedOut",
+	Message: "OS Provisioning for VM, didn't finished in the allotted time. Please check provisioning state later.",
+	SearchRegexes: []*regexp.Regexp{
+		regexp.MustCompile(`"code\\":\W\\"OSProvisioningTimedOut\\"`),
```
tsatam, line 46: Same feedback as in https://github.com/Azure/ARO-RP/pull/3691#pullrequestreview-2178564677, we should ensure that this regex pattern matches against the real logs emitted by the installer in a Hive cluster.

tsatam (changes requested):
I'd suggest testing this change by applying the Hive ConfigMap to a dev/full-service Hive cluster and attempting to reproduce these installation failures.

pkg/hive/failure/reasons.go
```diff
@@ -36,3 +37,12 @@ var AzureInvalidTemplateDeployment = InstallFailingReason{
 		regexp.MustCompile(`"code":\w?"InvalidTemplateDeployment"`),
 	},
 }
+
+var AzureOSProvisioningTimedOut = InstallFailingReason{
+	Name:    "AzureOSProvisioningTimedOut",
+	Reason:  "AzureOSProvisioningTimedOut",
+	Message: "OS Provisioning for VM, didn't finished in the allotted time. Please check provisioning state later.",
```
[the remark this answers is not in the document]
rhamitarora: Error updated

pkg/hive/failure/reasons.go
rhamitarora: Regex format updated, added 3 different test cases to match multiple regex strings.

bitoku (approved):

pkg/hive/failure/handler.go
```diff
@@ -50,6 +50,16 @@ func HandleProvisionFailed(ctx context.Context, cd *hivev1.ClusterDeployment, co
 			AzureInvalidTemplateDeployment.Message,
 			*armError,
 		)
+	case AzureOSProvisioningTimedOut.Reason:
```
SudoBrendan, line 53: while I think that it's good that we now have a well-defined error for this case, we should also add code modifications in the RP to handle this specific error, because it should not be returned to customers. OS provisioning timeouts are not something customers can fix, but may be something we wish to be alerted on.

SudoBrendan (changes requested):
I like the code changes so far, we should add logic to the RP code to not return this specific error to customers - it likely should result in a 500 error to the customer, and create some kind of data point for SREs to action.

mociarain:
(Sorry @SudoBrendan, I shouldn't have pinged you for this one)

pkg/hive/failure/handler.go
SudoBrendan: Another alternative feature we may choose to implement with approval from BU and buy-in from SREs is to extend the timeout duration of cluster installs for this specific case if we find it results in a higher install success rate.

SudoBrendan (changes requested):
my previous review was accidentally dismissed - see comments above.

PR #3680, opened 2024-07-11
Update controllergen commands to work with 0.9.0

hawkowl:
fixes a local build failure

Right now we rely on controller-gen 0.5.0, which is old and gives the wrong version when we run with `go run tool@version`. So, we should update it to 0.9.0 (which is current).

### Test plan for issue:
CI/E2E

N/A

E2E


bitoku:
> fixes a local build failure

What error did you get?

hawkowl:
@bitoku Running it with the real v0.5.0 outside of vendoring causes it to change the version number to the correct v0.5.0, and using the v0.9.0 that it reports causes another diff of newlines. This is because the v0.5.0 we're using misreports itself as v0.9.0 as per the `go.mod` before overrides.

mociarain (approved):
LGTM

PR #3681, opened 2024-07-11
Optionally Use USER Environment Variable for Azure Resources

razo7:

Mitigate the Azure resource name prefix dependency with the `USER` env var. Instead of always using that value for creating Azure resources we can now look for`AZURE_PREFIX` and fallback to `USER` (default behavior) in case the new env var is missing. This change would resolve any issue of having a common USER env var value (e.g., as `root`), which could lead to a clash of some resources in Azure (e.g., Azure Container Registry name).

1. Fetch the USER env var once instead of multiple times for faster DevConfig func execution time
2. Use an Azure unique prefix for the Azure resources that ARO-RP is using/creating. When `AZURE_PREFIX` env var is not set, then use the USER env var value.
3. Modify configuration files, and documentation files such as _env.example_ and _docs/deploy-full-rp-service-in-dev.md_.

### Test plan for issue:
None

Yes, any documentation that uses the USER env var:

1. docs/deploy-development-rp.md
2. docs/deploy-full-rp-service-in-dev.md
4. docs/disk-encryption-set.md
5. docs/operators.md
6. docs/prepare-a-shared-rp-development-environment.md
7. docs/unit-testing-for-monitoring-metrics.md

I tested it manually as part of https://github.com/Azure/ARO-RP/blob/master/docs/deploy-full-rp-service-in-dev.md#deploying-an-int-like-development-rp


bitoku (changes requested):
I don't think this change is thorough.
I tried searching with `$USER` in whole codebase, and there are some other usages.
I'm not sure if we need to change those files, and testing all of them would need a lot of effort. 

At least I think we need to change https://github.com/Azure/ARO-RP/blob/master/docs/deploy-full-rp-service-in-dev.md#deploying-an-int-like-development-rp accordingly.

razo7:
>I don't think this change is thorough.
I tried searching with $USER in whole codebase, and there are some other usages.

I agree, some references are relevant and some are just for stating the user name (maybe it was mentioned as for the Azure prefix, but I am not certain though). I have applied the needed changes wherever I was confident that this kind of change would be applicable (docs as well). In case the PR is still incomplete let me know @bitoku

docs/deploy-full-rp-service-in-dev.md
```diff
@@ -49,9 +49,9 @@
     What to change in `env-int` file:
 
     * if using a public key separate from `~/.ssh/id_rsa.pub` (for ssh access to RP and Gateway vmss instances), source it with `export SSH_PUBLIC_KEY=~/.ssh/id_separate.pub`
-    * don't try to change `$USER` prefix used there
+    - Modify `AZURE_PREFIX` environment variable for a different `$AZUREPREFIX` prefix of the created Azure resources.
```
bitoku, line 52: What's the difference between those?
Having took a glance, I don't think it needs two variables.

If you are concerning about this part `export AZUREPREFIX="${AZURE_PREFIX:-$USER}"`, it can be `export AZURE_PREFIX="${AZURE_PREFIX:-$USER}"`

razo7: I agree

bitoku (dismissed):
LGTM but I want some more eyes on it to make sure it won't break anything.

LiniSusan:
@razo7 : LGTM but just one question: if we execute `SECRET_SA_ACCOUNT_NAME=rharosecretsdev make secrets` as per https://github.com/LiniSusan/ARO-RP/blob/master/docs/deploy-development-rp.md, how is it going to affect with the changes? we are downloading the secrets folder  `az storage blob download -n secrets.tar.gz -c secrets -f secrets.tar.gz --account-name ${SECRET_SA_ACCOUNT_NAME} >/dev/null` in make command. So after sourcing the env file, will that make any conflict?

razo7:
>if we execute `SECRET_SA_ACCOUNT_NAME=rharosecretsdev make secrets` as per [`master`/docs/deploy-development-rp.md](https://github.com/Azure/ARO-RP/blob/master/docs/deploy-development-rp.md?rgh-link-date=2024-07-29T07%3A40%3A46Z), how is it going to affect with the changes? we are downloading the secrets folder `az storage blob download -n secrets.tar.gz -c secrets -f secrets.tar.gz --account-name ${SECRET_SA_ACCOUNT_NAME} >/dev/null` in make command. So after sourcing the env file, will that make any conflict?

AFAIU I don't see any conflict here. The `secrets` target is unrelated to this change. The changes here will do any change only when AZURE_PREFIX env var is present. When it is empty the value of $USER will be used (e.g., `AZURE_PREFIX="${AZURE_PREFIX:-$USER}"`)

hack/local-monitor-testing/deploy_MDM_VM.sh
```diff
@@ -10,11 +10,13 @@ MDMFRONTENDURL=https://global.ppe.microsoftmetrics.com/
...
 MDMSOURCEROLE=rp
 MDMSOURCEROLEINSTANCE=$HOSTNAME
+# use unique prefix for Azure resources when it is set, otherwise use your user's name
+export AZURE_PREFIX="${AZURE_PREFIX:-$USER}"
 
 echo "Using:"
 
 echo "Resourcegroup = $RESOURCEGROUP"
-echo "User          = $USER"
+echo "AzureUniquePrefix = $AZURE_PREFIX"
```
tiguelu, line 19: nit: indentation for the other echo commands were using the same left padding.

hack/local-monitor-testing/startMDMNetwork.sh
```diff
@@ -27,13 +27,15 @@ function killProcesses() {
...
 BASE=$( git rev-parse --show-toplevel)
 SOCKETFILE="$BASE/cmd/aro/mdm_statsd.socket"
+# use unique prefix for Azure resources when it is set, otherwise use your user's name
+export AZURE_PREFIX="${AZURE_PREFIX:-$USER}"
 
 echo "Using:"
 
 echo "Resourcegroup = $RESOURCEGROUP"
-echo "User          = $USER"
+echo "AzureUniquePrefix          = $AZURE_PREFIX"
```
tiguelu, line 36: nit: left padding in the echo commands.

docs/deploy-full-rp-service-in-dev.md
```diff
@@ -49,9 +49,9 @@
     What to change in `env-int` file:
 
     * if using a public key separate from `~/.ssh/id_rsa.pub` (for ssh access to RP and Gateway vmss instances), source it with `export SSH_PUBLIC_KEY=~/.ssh/id_separate.pub`
-    * don't try to change `$USER` prefix used there
+    - Modify `AZURE_PREFIX` environment variable for a different prefix of the created Azure resources.
```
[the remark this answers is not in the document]
razo7: Yes. Suggesting the user to set AZURE_PREFIX if she/he wants to. However, exporting this var and sourcing env-int would give the same result.

> export AZURE_PREFIX=ortest
source env-int`

hack/local-monitor-testing/deploy_MDM_VM.sh
razo7: I do not fully understand what you mean by left padding. Let me know if the new commit helped resolve that.
(Here and other places)

docs/deploy-development-rp.md
```diff
@@ -70,7 +70,7 @@
    ```bash
    az deployment group create \
      -g "$RESOURCEGROUP" \
-     -n "databases-development-$USER" \
+     -n "databases-development-$AZURE_PREFIX" \
```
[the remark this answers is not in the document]
razo7: When `env.example` or its modified replica is not sourced then you are right.

razo7 (commented):
Thanks for the input 👍🏻

hack/local-monitor-testing/deploy_MDM_VM.sh
tiguelu: Just meant the spacing that kept the output aligned (all the values printed in the same column), but just a nitpick.

tiguelu (approved):

PR #3684, opened 2024-07-12
Update local_dev_env.sh and instructions

niontive:

Fixes


hack/devtools/local_dev_env.sh
kimorris27, line 1: Today I learned that this script exists. Seems pretty handy. Good call to incorporate the mock MSI stuff here.

kimorris27 (approved):

hack/devtools/msi.sh
```diff
@@ -3,20 +3,17 @@
...
 
-# Extract the client ID and tenant ID from the result
-clientID=$(echo $result | jq -r .appId)
-tenantID=$(echo $result | jq -r .tenant)
-
-# Extract the certificate information from the result
-certFilePath=$(echo $result | jq -r '.fileWithCertAndPrivateKey')
-base64EncodedCert=$(base64 -w 0 $certFilePath)
-rm $certFilePath
+sp=$(create_mock_msi)
```
gouthamMN, line 11: Doesn't want to hold you from merging this since it's for Dev. But you can add err check or sp is not nil condition.

hack/devtools/local_dev_env.sh
```diff
@@ -39,13 +39,42 @@ ask_to_create_default_env_config() {
...
+}
+
+get_mock_msi_cert() {
+    certFilePath=$(echo "$1" | jq -r '.fileWithCertAndPrivateKey')
+    base64EncodedCert=$(base64 -w 0 $certFilePath)
+    rm $certFilePath
+    echo $base64EncodedCert
+}
+
 create_env_file() {
```
gouthamMN, line 63: Since we are creating Mock MSI under default env file, should we add a prompt to see if the user is planning to create a MIWI ARO RP or Regular RP for Dev?

hack/devtools/local_dev_env.sh
```diff
@@ -39,13 +39,42 @@ ask_to_create_default_env_config() {
     fi
 }
 
+# We use a service principal and certificate as the mock MSI object
+create_mock_msi() {
+    appName="mock-msi-$(openssl rand -base64 9 | tr -dc 'a-zA-Z0-9' | head -c 6)"
+    az ad sp create-for-rbac --name $appName --create-cert --output json
```
gouthamMN, line 45: Don't we have to echo the command output in order to return the SP value?

cadenmarchese (approved):

hack/devtools/local_dev_env.sh
niontive: I'll add this when we merge to master

hack/devtools/msi.sh
niontive: I'll add this when we merge to master.

PR #3685, opened 2024-07-12
Create sp for each e2e run

mociarain:
[Ticket](https://issues.redhat.com/browse/ARO-8854)

This PR:
- reverts the contents of https://github.com/Azure/ARO-RP/pull/3498/files so we can run E2E in the new tenant.
- makes the deletion flow more explicit and attempts to clean up everything at the end.

### DIFF
--- a/docs/deploy-development-rp.md
+++ b/docs/deploy-development-rp.md
@@ -127,7 +127,6 @@
    
       ```bash
       CLUSTER=<cluster-name> go run ./hack/cluster delete
-      CLUSTER=<cluster-name> go run ./hack/cluster deleteapp
       ```
 
       By default, a public cluster will be created. In order to create a private cluster, set the `PRIVATE_CLUSTER` environment variable to `true` prior to creation. Internet access from the cluster can also be restricted by setting the `NO_INTERNET` environment variable to `true`.
--- a/docs/deploy-development-rp.md
+++ b/docs/deploy-development-rp.md
@@ -119,9 +119,6 @@
    1. use the create utility:
 
       ```bash
-      # Create the application to run the cluster as and load it
-      CLUSTER=<cluster-name> go run ./hack/cluster createapp
-      source clusterapp.env
       # Create the cluster
       CLUSTER=<cluster-name> go run ./hack/cluster create
       ```
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -140,13 +140,13 @@ type AppDetails struct {
 }
 
 func (c *Cluster) createApp(ctx context.Context, clusterName string) (applicationDetails AppDetails, err error) {
-	c.log.Infof("creating AAD application")
+	c.log.Infof("Creating AAD application")
 	appID, appSecret, err := c.createApplication(ctx, "aro-"+clusterName)
 	if err != nil {
 		return AppDetails{}, err
 	}
 
-	c.log.Infof("creating service principal")
+	c.log.Infof("Creating service principal")
 	spID, err := c.createServicePrincipal(ctx, appID)
 	if err != nil {
 		return AppDetails{}, err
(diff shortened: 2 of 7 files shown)


hawkowl:
Don't we need the unique SPs for the production E2E?

mociarain:
> Don't we need the unique SPs for the production E2E?

Not in the new tenant :sunglasses:

pkg/util/cluster/cluster.go
```diff
@@ -714,13 +728,28 @@ func (c *Cluster) deleteCluster(ctx context.Context, resourceGroup, clusterName
 	return nil
 }
 
-func (c *Cluster) deleteClusterResourceGroup(ctx context.Context, resourceGroup string) error {
+func (c *Cluster) ensureResourceGroupDeleted(ctx context.Context, resourceGroupName string) error {
```
mociarain, line 731: This was added due to some racey deletions and some stuff not getting cleaned up correctly in the new tenant. AFAIK in the past we were letting the VnetRG get cleaned up via the `purge:true` flag (https://github.com/Azure/ARO-RP/pull/3513/files @tsatam can you confirm this). This just makes all the deletion a bit more explicit but we still have the purge script as a fall back

pkg/util/cluster/cluster.go
```diff
@@ -133,31 +134,26 @@ func New(log *logrus.Entry, environment env.Core, ci bool) (*Cluster, error) {
 	return c, nil
 }
 
-func (c *Cluster) CreateApp(ctx context.Context, clusterName string) error {
-	c.log.Infof("creating AAD application")
+type AppDetails struct {
+	applicationId     string
+	applicationSecret string
+	SPId              string
+}
```
bitoku, line 141: I think all these (struct and fields) can be private.

pkg/util/cluster/cluster.go
```diff
@@ -714,13 +716,28 @@ func (c *Cluster) deleteCluster(ctx context.Context, resourceGroup, clusterName
...
+	c.log.Printf("deleting resource group %s", resourceGroupName)
+	timeoutCtx, cancel := context.WithTimeout(ctx, 10*time.Minute)
+	defer cancel()
+
+	return wait.PollImmediateUntil(5*time.Second, func() (bool, error) {
+		_, err := c.groups.Get(ctx, resourceGroupName)
+		if azureerrors.ResourceGroupNotFound(err) {
+			c.log.Infof("finished deleting resource group %s", resourceGroupName)
+			return true, nil
+		}
```
bitoku, line 729: We should catch other errors here although it doesn't have to fail by the error.

docs/deploy-development-rp.md
```diff
@@ -119,9 +119,6 @@
    1. use the create utility:
 
       ```bash
-      # Create the application to run the cluster as and load it
-      CLUSTER=<cluster-name> go run ./hack/cluster createapp
```
tiguelu, line 123: Do we want to keep providing a custom SP in local dev as an option instead of forcibly auto-creating one?

We could still use `az` with dev extension, although we would have to create it in a few more steps, we wouldn't use the smaller workers SKU and it would be easier to forget to delete the app (there would be no file around to remind us we have a lingering app/SP)

pkg/util/cluster/cluster.go
mociarain: Good point. I'll add it to the error message

docs/deploy-development-rp.md
tiguelu: Talked to Maitiu and we are carding it as a follow-up to allow to merge this PR as it is, since we want to get the change reverted quickly and this card can be also a good-first-item:
https://issues.redhat.com/browse/ARO-9320

tiguelu (approved):

PR #3686, opened 2024-07-12
Preflight update and tests

carvalhe:

Currently we have preflight set up for create operations but also need it for update. These changes allow for both put/patch operations on preflight. This PR also cleans up the way we validate for preflight and adds new checks for it as well. There was a previous PR for this here: https://github.com/Azure/ARO-RP/pull/3365 but it was getting messy, and needed to move off the forked branch so that E2E tests would work correctly. 

### Test plan for issue:
Created additional unit tests that are similar to put/patch tests. Kusto logs also show preflight success and failures with more details than before. Preflight is also marked for testing and won't hit live until another PR in the future sets that value.

### DIFF
--- a/pkg/frontend/openshiftcluster_preflightvalidation_test.go
+++ b/pkg/frontend/openshiftcluster_preflightvalidation_test.go
@@ -6,96 +6,123 @@ package frontend
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
+	"strings"
 	"testing"
-
-	"github.com/Azure/go-autorest/autorest/to"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/metrics/noop"
+	"github.com/Azure/ARO-RP/pkg/util/version"
 	testdatabase "github.com/Azure/ARO-RP/test/database"
 )
 
 func TestPreflightValidation(t *testing.T) {
 	ctx := context.Background()
 	mockSubID := "00000000-0000-0000-0000-000000000000"
+	apiVersion := "2020-04-30"
+	clusterId := "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resourceGroup/providers/Microsoft.RedHatOpenShift/openShiftClusters/resourceName"
+	location := "eastus"
+	defaultProfile := "default"
+	resourceGroup := "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/resourceGroupTest"
+	netProfile := "10.128.0.0/14"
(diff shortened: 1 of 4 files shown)


cadenmarchese (approved):
lgtm!

pkg/frontend/openshiftcluster_preflightvalidation_test.go
```diff
@@ -6,96 +6,342 @@ package frontend
...
+		api.ExampleOpenShiftClusterDocument().OpenShiftCluster.Type,
+		location, defaultProfile, resourceGroup, api.EncryptionAtHostEnabled, version.DefaultInstallStream.Version.String(),
+		mockSubID, mockSubID, netProfile, netProfile, api.VMSizeStandardD32sV3, masterSub,
+		api.EncryptionAtHostEnabled, encryptionSet,
+		api.ExampleOpenShiftClusterDocument().OpenShiftCluster.Properties.WorkerProfiles[0].Name, api.VMSizeStandardD32sV3,
+		api.ExampleOpenShiftClusterDocument().OpenShiftCluster.Properties.WorkerProfiles[0].DiskSizeGB,
+		api.EncryptionAtHostEnabled, workerSub,
+		api.ExampleOpenShiftClusterDocument().OpenShiftCluster.Properties.WorkerProfiles[0].Count,
+		encryptionSet, api.VisibilityPublic, defaultProfile, api.VisibilityPublic,
+		api.ExampleOpenShiftClusterDocument().OpenShiftCluster.Properties.IngressProfiles[0].IP))
```
SudoBrendan, line 95: nit: 28 ordered parameters is going to be hard to maintain over time - if it's easy enough, can we convert this to a [golang string template](https://pkg.go.dev/text/template#pkg-overview), and then populate it with named parameters rather than ordered ones?

pkg/frontend/openshiftcluster_preflightvalidation_test.go
```diff
@@ -6,96 +6,342 @@ package frontend
...
+			fixture: func(f *testdatabase.Fixture) {
+				f.AddSubscriptionDocuments(api.ExampleSubscriptionDocument())
+			},
+			changeFeed: defaultVersionChangeFeed,
+			preflightRequest: func() *api.PreflightRequest {
+				return &api.PreflightRequest{
+					Resources: []json.RawMessage{
+						[]byte(fmt.Sprintf(`
+							{
+								"apiVersion": "%s",
```
SudoBrendan, line 215: nit: same as above - now thinking this should be wrapped in a function call, where you pass it a map of variables.

SudoBrendan (approved):
Thanks for the continued work here! I don't have anything blocking, but I think updating the SprintF to a template execution could be warranted here if desired.

PR #3689, opened 2024-07-12
create platform identity

gouthamMN:

Fixes


niontive (approved):

hack/devtools/platform_msi.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+# This script creates a platform Identities to use for local development
+# The script reads the operators file and creates platform identities for each operator
+
+prompt_for_operators_file() {
```
cadenmarchese, line 6: Should this be 
```suggestion
prompt_for_platform_workload_identity_role_set() {
```
instead, or what is the operators file in this case, just a list of operators only?

It might require a little jq magic, but I'm thinking we could use a platformworkloadidentityroleset instead since we already have that defined (see description of https://github.com/Azure/ARO-RP/pull/3617)

What do others think?

kimorris27 (commented):
LGTM except for Caden's open question and one other question: how do you plan to add the role assignments? In a follow-up PR that extends this same script?

hack/devtools/platform_msi.sh
gouthamMN: @cadenmarchese  So, if I understand correctly, what you're suggesting is, to ask the user to set env var PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS and read the values from this env var instead of file?

hack/devtools/platform_msi.sh
```diff
@@ -0,0 +1,52 @@
...
+    local operatorsFilePath
+    read -p "Please provide the absolute path of operators file including the file name?" operatorsFilePath
```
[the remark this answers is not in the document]
kimorris27: Cool, that addresses my question about the role assignments and works for me. Did you see my code suggestion to change that question mark to a colon though?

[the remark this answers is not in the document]
gouthamMN: Yes, I will fix it once I have @cadenmarchese replies to my comment about using env var. If we have to use env var instead of file, then that line will be removed.

hack/devtools/platform_msi.sh
cadenmarchese: @gouthamMN I'd be open to that idea or Kipp's idea, whichever we prefer. The nice thing about the PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS is that it also contains the role assignments that need to be done, so if you took it in as an environment variable you could extend the functionality of the script to also assign the roles.

hack/devtools/platform_msi.sh
```diff
@@ -0,0 +1,79 @@
...
+    name=$(echo $result | jq -r .name)
+
+    echo "Client ID: $clientID"
+    echo "Principal ID: $principalId"
+    echo "Resource ID: $resourceId"
+    echo "Name: $name"
+    echo ""
+
+    if [[ "${operatorName}" == "MachineApiOperator" || "${operatorName}" == "NetworkOperator" \
+        || "${operatorName}" == "AzureFilesStorageOperator" || "${operatorName}" == "ServiceOperator" ]]; then
```
cadenmarchese, line 47: I don't know if this `if` statement should be included:
- The operator names in the platform workload identity role set env var should be allowed to change, this hardcodes them
- There are 8 total roles that need to be assigned to 8 identities, and there are only 4 here - don't we want to assign roles to all identities?

hack/devtools/platform_msi.sh
```diff
@@ -1,23 +1,35 @@
...
 
 # This script creates a platform Identities to use for local development
-# The script reads the operators file and creates platform identities for each operator
+# The script reads the env var PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS and creates platform identities for each operator
 
-prompt_for_operators_file() {
-    local operatorsFilePath
-    read -p "Please provide the absolute path of operators file including the file name?" operatorsFilePath
+get_platform_workloadIdentity_role_sets() {
+    local platformWorkloadIdentityRole
```
kimorris27, line 7: Nit: should this be `platformWorkloadIdentityRoles` to match the name of the variable used later in the function?

hack/devtools/platform_msi.sh
kimorris27: >There are 8 total roles that need to be assigned to 8 identities, and there are only 4 here - don't we want to assign roles to all identities?

This script only does the role assignments that are scoped to the vnet RG. The others will be done by the RP via the cluster MSI.

>The operator names in the platform workload identity role set env var should be allowed to change, this hardcodes them

The script has to somehow know which role assignments go at the vnet RG scope. I don't see a better way than hard-coding them.

Also, this script is a temporary solution to accelerate local dev MIWI cluster creation. It'll be superseded by the automatic CLI stuff in a few months. So this doesn't need to be perfect or future-proof. Hard-coding the 4.14 operator names is fine IMO.

kimorris27 (commented):
Apologies for not catching this and mentioning it in my earlier review, but we're still missing the federated credential role assignment on the cluster MSI. I'd be cool with doing that in its own PR through, since it's an extension to this other stuff.

hack/devtools/platform_msi.sh
rajdeepc2792: Seeing Kipp's comment I agree that we are okay to use the hardcoded values in IF.
I believe it can be either way, assigning the desired role over the Network RG even when not needed shouldn't cause any problems other than anything that is identified during testing. So to be safe it is a good idea here to hardcode only the identities that will need interaction with Network RG.

hack/devtools/platform_msi.sh
```diff
@@ -0,0 +1,79 @@
...
+        || "${operatorName}" == "AzureFilesStorageOperator" || "${operatorName}" == "ServiceOperator" ]]; then
+
+        assign_role_to_platform_identity "${principalId}" "${roleDefinitionId}"
+    fi
+}
+
+setup_platform_identity() {
+    local platformWorkloadIdentityRoles=$(get_platform_workloadIdentity_role_sets)
+
+    echo "INFO: Creating platform identities under RG ($RESOURCEGROUP) and Sub Id ($AZURE_SUBSCRIPTION_ID)"
```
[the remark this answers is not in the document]
rajdeepc2792: Extending the conversation here, curious to know how the created managed identities will be assigned to the cluster creation process? And if we are going to use the same managed identities for all the dev cluster creations, the federated credentials for the managed identities are limited to 10. That is if federated credentials are not deleted the MIWI cluster creation will be limited to 10.

hack/devtools/platform_msi.sh
cadenmarchese: I see, thank you for the clarifications!

hack/devtools/platform_msi.sh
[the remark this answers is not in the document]
cadenmarchese: My understanding just from past assumptions and by reading this PR is that the identities will be in the network resource group. They won't be re-used in dev but we can get around the limitation by creating additional resource groups (since the identities need to have unique names)

Maybe we should modify the script to name the identities $CLUSTER-$operatorName or similar?

hack/devtools/platform_msi.sh
gouthamMN: fixed

hack/devtools/platform_msi.sh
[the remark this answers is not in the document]
gouthamMN: fixed

gouthamMN:
> Apologies for not catching this and mentioning it in my earlier review, but we're still missing the federated credential role assignment on the cluster MSI. I'd be cool with doing that in its own PR through, since it's an extension to this other stuff.

Let's do it in a separate PR.

kimorris27 (approved):

PR #3690, opened 2024-07-15
Try and get the NewMonitor/Frontend/etc mess under control

hawkowl:
This makes the extra db params for NewMonitor/Frontend/etc needed for MIMO much smaller by wrapping the required databases in an object that is able to contain them and be passed through instead.


jaitaiwan (approved):
LGTM

kimorris27 (approved):

PR #3693, opened 2024-07-15
fix dev env panic in e2e

bitoku:

Fixes N/A

When I ran e2e to the local dev cluster without encryption, I got this error and it panics.

```
  STEP: getting the test cluster resource @ 07/15/24 13:05:49.935
  STEP: checking that disk encryption at rest is enabled for masters @ 07/15/24 13:05:50.148
  [PANICKED] in [It] - /usr/local/Cellar/go/1.22.2/libexec/src/runtime/panic.go:261 @ 07/15/24 13:05:50.148
• [PANICKED] [0.429 seconds]
Disk encryption at rest [It] must be enabled with customer managed key for the cluster and each disk must have it enabled
/Users/atokubi/ARO-RP/test/e2e/disk_encryption.go:70

  [PANICKED] Test Panicked
  In [It] at: /usr/local/Cellar/go/1.22.2/libexec/src/runtime/panic.go:261 @ 07/15/24 13:05:50.148

  runtime error: invalid memory address or nil pointer dereference

  Full Stack Trace
    github.com/Azure/ARO-RP/test/e2e.init.func19.1({0x25a3cc788, 0xc000994270})
    	/Users/atokubi/ARO-RP/test/e2e/disk_encryption.go:85 +0x313
```

The problem is that it tried to dereference `DiskEncryptionSetID` which is nil and it failed.
This PR adds nil check before that not to panic if it is nil.

### Test plan for issue:

Run e2e to a local dev cluster.

e2e bug fix N/A

Run e2e in CI.

### DIFF
--- a/test/e2e/disk_encryption.go
+++ b/test/e2e/disk_encryption.go
@@ -82,13 +82,15 @@ var _ = Describe("Disk encryption at rest", func() {
 		By("checking that disk encryption at rest is enabled for masters")
 		Expect(oc.OpenShiftClusterProperties).To(Not(BeNil()))
 		Expect(oc.OpenShiftClusterProperties.MasterProfile).To(Not(BeNil()))
-		Expect(*(*oc.OpenShiftClusterProperties.MasterProfile).DiskEncryptionSetID).NotTo(BeEmpty())
+		Expect(oc.OpenShiftClusterProperties.MasterProfile.DiskEncryptionSetID).NotTo(BeNil())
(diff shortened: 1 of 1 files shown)


kimorris27:
I started E2E again because I think the failure was unrelated to your change.

LGTM if E2E passes!

kimorris27 (approved):
Thanks for the effort put into improving our E2E!

mociarain (approved):

PR #2961, opened 2023-06-14
pkg/cluster: Use Action for fixupClusterSPObjectID

mbarnes:
### Which issue this PR addresses:

[[ARO-1919] AAD: Migrate ARO-RP Azure AD Graph -> Microsoft Graph](https://issues.redhat.com/browse/ARO-1919)

### What this PR does / why we need it:

Minor follow-up to https://github.com/Azure/ARO-RP/pull/1970 where @bennerv requested this in his review.

The `fixupClusterSPObjectID` step in PUCM uses only the [MS Graph client](https://pkg.go.dev/github.com/microsoftgraph/msgraph-sdk-go#GraphServiceClient) (no ARM calls), and the client already handles HTTP 401 Unauthorized errors by refreshing its authorization token. So there's no need to use the [AuthorizationRetryingAction](https://github.com/Azure/ARO-RP/blob/1ee42feb78ae64ca694bf832be3ee17bde8cbea7/pkg/util/steps/refreshing.go#L21-L24) wrapper.
    
The 401 Unauthorized error handling logic for the MS Graph client is in the [github.com/microsoft/kiota-http-go](https://github.com/microsoft/kiota-http-go/blob/main/nethttp_request_adapter.go#L173-L198) module.
   

### Test plan for issue:

- Create an ARO cluster.
- Run `az aro update --refresh-credentials` on it.
- Verify it completes successfully.

### Is there any documentation that needs to be updated for this PR?

No, this is just an internal change.

PR #1970, opened 2022-02-16
Migrate RP from Azure AD Graph to Microsoft Graph

mbarnes:
### Which issue this PR addresses:

[[ARO-1919] AAD: Migrate ARO-RP Azure AD Graph -> Microsoft Graph](https://issues.redhat.com/browse/ARO-1919)

### What this PR does / why we need it:

Microsoft has [announced](https://techcommunity.microsoft.com/t5/microsoft-entra-azure-ad-blog/azure-ad-change-management-simplified/ba-p/2967456) the impending end-of-support for Azure Active Directory Graph API:

> We will retire the Azure AD Graph API any time after June 30th, 2023.

We need to migrate the RP code to the [Microsoft Graph SDK for Go](https://pkg.go.dev/github.com/microsoftgraph/msgraph-sdk-go).

### Test plan for issue:

Pass all* GitHub checks, especially E2E tests.

\* Except perhaps CodeQL, see **additional notes** below.

### Is there any documentation that needs to be updated for this PR?

None that I'm aware of.  This is an internal RP change and should not alter its behavior.

### Additional notes: ###

CodeQL currently fails due to the vendoring of [Microsoft Graph SDK for Go](https://pkg.go.dev/github.com/microsoftgraph/msgraph-sdk-go), which, due to its enormous API surface, causes memory consumption during CodeQL analysis to double, putting it well beyond the 8 GiB memory limit of standard GitHub Action runners.  I'm working on moving CodeQL analysis to an Azure Pipeline, where we can use a virtual machine with adequate RAM (see the "pipelines" commit).

PR #3694, opened 2024-07-15
pkg/cluster: Use Action for fixupClusterSPObjectID

ehvs:
Picking up from @mbarnes  https://github.com/Azure/ARO-RP/pull/2961
[[ARO-1919] AAD: Migrate ARO-RP Azure AD Graph -> Microsoft Graph](https://issues.redhat.com/browse/ARO-1919)

Minor follow-up to #1970 where @bennerv requested this in his review.

The fixupClusterSPObjectID step in PUCM uses only the MS Graph client (no ARM calls), and the client already handles HTTP 401 Unauthorized errors by refreshing its authorization token. So there's no need to use the AuthorizationRetryingAction wrapper.

The 401 Unauthorized error handling logic for the MS Graph client is in the github.com/microsoft/kiota-http-go module.

### Test plan for issue:
Create an ARO cluster.
Run az aro update --refresh-credentials on it.
Verify it completes successfully.
No, this is just an internal change.

### DIFF
--- a/pkg/cluster/adminupdate_test.go
+++ b/pkg/cluster/adminupdate_test.go
@@ -36,7 +36,7 @@ func TestAdminUpdateSteps(t *testing.T) {
 		"[Action initializeKubernetesClients]",
 		"[Action ensureBillingRecord]",
 		"[Action ensureDefaults]",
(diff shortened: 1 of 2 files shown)


mociarain (approved):
LGTM

Note: We should close the old PR when this is merged

PR #3701, opened 2024-07-17
Add panic recovery within endLease

s-fairchild:
This is to ensure EndLease always gets called, including cases when a panic is encountered.

[ARO-9241](https://issues.redhat.com/browse/ARO-9241)

Fixes

In the event that a panic occurs within `endLease()`, we want to ensure that `ocb.dbOpenShiftClusters.EndLease()` is always called.

### Test plan for issue:

Run a local RP and induce a panic during a cluster operation.

No.

Local testing.

### DIFF
--- a/pkg/backend/openshiftcluster.go
+++ b/pkg/backend/openshiftcluster.go
@@ -288,10 +288,15 @@ func (ocb *openShiftClusterBackend) updateAsyncOperation(ctx context.Context, lo
 	return nil
 }
 
-func (ocb *openShiftClusterBackend) endLease(ctx context.Context, log *logrus.Entry, stop func(), doc *api.OpenShiftClusterDocument, operationType, provisioningState api.ProvisioningState, backendErr error) error {
+func (ocb *openShiftClusterBackend) endLease(ctx context.Context, log *logrus.Entry, stop func(), doc *api.OpenShiftClusterDocument, operationType, provisioningState api.ProvisioningState, backendErr error) (err error) {
 	var adminUpdateError *string
 	var failedProvisioningState api.ProvisioningState
 	initialProvisioningState := doc.OpenShiftCluster.Properties.ProvisioningState
+
+	defer func() {
+		recover.Panic(log)
+		_, err = ocb.dbOpenShiftClusters.EndLease(ctx, doc.Key, provisioningState, failedProvisioningState, adminUpdateError)
+	}()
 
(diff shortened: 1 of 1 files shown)


pkg/backend/openshiftcluster.go
```diff
@@ -292,6 +292,12 @@ func (ocb *openShiftClusterBackend) endLease(ctx context.Context, log *logrus.En
 	var adminUpdateError *string
 	var failedProvisioningState api.ProvisioningState
 	initialProvisioningState := doc.OpenShiftCluster.Properties.ProvisioningState
+	var err error
+
+	defer func() {
+		recover.Panic(log)
+		_, err = ocb.dbOpenShiftClusters.EndLease(ctx, doc.Key, provisioningState, failedProvisioningState, adminUpdateError)
```
SudoBrendan, line 299: I don't think this works as expected - `err` is swallowed in the `defer` here.

Here's a playground to show what I mean: https://go.dev/play/p/3Efb6cUNo8U

SudoBrendan (changes requested):

pkg/backend/openshiftcluster.go
SudoBrendan: found the fix - use a named return result, in the playground, changing the function header to this works:

func sevenIsTooScary() (err error)

in the `defer`, we are then modifying the named `err` value.

tl;dr - remove `var err error` and change the `endLease` header to return `(err error)` instead of `error`.

SudoBrendan (changes requested):

bitoku (approved):

pkg/backend/openshiftcluster.go
bitoku: @kimorris27 
This might help
https://news.ycombinator.com/item?id=14668955
> Yep, and one cannot simulate that with local variables. In Go "return v" copies v into the return location before calling the deferred code. If that location is not named, the deferred function has no way to change it

kimorris27: Yeah, that does help, thank you! I learned some new things about Go today.

kimorris27 (approved):

mociarain (approved):

rajdeepc2792 (approved):

rhamitarora (approved):

tsatam (approved):

SudoBrendan (approved):

PR #3704, opened 2024-07-17
Fix Managed Upgrade Operator reconciler

slintes:

Fixes TODO

On OCP 4.15, the MUO reconciler is running into an endless loop, constantly updating the `trusted-ca-bundle` ConfigMap.
Reason is that since OCP 4.15 the controller, which inserts the certificates into that ConfigMap, is also adding a new annotation: `"openshift.io/owning-component": "Networking / cluster-network-operator"`.
So we need to copy that over to the new ConfigMap in case of updates.

### Test plan for issue:

The issue was discovered by very flake MUO deletion test on OCP 4.15. With this change, the test succeeds.
Also the corresponding unit test was extended for the new annotation.

No, this just is a not user facing implementation detail.

Trusting the e2e test.

### DIFF
--- a/pkg/util/dynamichelper/dynamichelper.go
+++ b/pkg/util/dynamichelper/dynamichelper.go
@@ -275,6 +275,8 @@ func merge(old, new kruntime.Object) (kruntime.Object, bool, string, error) {
 			if ext {
 				new.Data["ca-bundle.crt"] = caBundle
 			}
+			// since OCP 4.15 this annotation is added to the trusted-ca-bundle ConfigMap by the ConfigMap's controller
+			copyAnnotation(&new.ObjectMeta, &old.ObjectMeta, "openshift.io/owning-component")
 		}
 
 	case *machinev1beta1.MachineHealthCheck:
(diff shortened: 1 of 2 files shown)


tsatam (changes requested):
Change LGTM, can we get an additional test case added to https://github.com/Azure/ARO-RP/blob/master/pkg/util/dynamichelper/dynamichelper_test.go to cover for this?

slintes:
@microsoft-github-policy-service agree [company="Red Hat"]

slintes:

tsatam:
Manual CI run: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=98412498&view=results
Manual E2E run: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=98412429&view=results

bitoku (commented):
Why don't we just add that annotation from the beginning?

tsatam (approved):
Changes look good to me, manually-run CI/E2E pipelines are green. I think this is good to merge.

bitoku:
I approve the change, but I wanted to figure out other way because dynamic helper is used in some functions, and would rather not add adhoc change in the function.

PR #3705, opened 2024-07-18
Fix missing database refactor on preflight validation in frontend

bennerv:
Fixes CI failures on https://github.com/Azure/ARO-RP/actions/runs/9992357721/job/27617159964

PR #3706, opened 2024-07-18
[Dev Tooling] generateSubnets should check if subnets are already taken

mrWinston:
This PR fixes errors like:
```
Subnet 'maschulz-master' is not valid because its IP address range overlaps with that of an existing subnet in virtual network 'dev-vnet'
```
when creating a cluster using `go run ./hack/cluster create`.

Instead of just randomly generating the subnet cidrs, `generateSubnets` now first retrieves all subnets in the `dev-vnet` and makes sure that the subnet cidrs don't overlap with any of the preexisting ones.

- No

- This change doesn't affect production.

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"net"
 	"net/http"
 	"os"
 	"strings"
@@ -345,22 +346,103 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 	return nil
 }
 
+// ipRangesContainCIDR checks, weather any of the ipRanges overlap with the cidr string. In case cidr isn't valid, false is returned.
+func ipRangesContainCIDR(ipRanges []*net.IPNet, cidr string) bool {
+	_, cidrNet, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return false
+	}
+
+	for _, snet := range ipRanges {
+		if snet.Contains(cidrNet.IP) || cidrNet.Contains(snet.IP) {
+			return true
+		}
+	}
+	return false
+}
+
+// GetIpRangesFromSubnet converts a given azure subnet to a list if IPNets.
+// Because an az subnet can cover multiple ipranges, we need to return a slice
+// instead of just a single ip range. This function never errors. If something
+// goes wrong, it instead returns an empty list.
+func GetIpRangesFromSubnet(subnet mgmtnetwork.Subnet) []*net.IPNet {
+	ipRanges := []*net.IPNet{}
+	if subnet.AddressPrefix != nil {
+		_, ipRange, err := net.ParseCIDR(*subnet.AddressPrefix)
+		if err == nil {
+			ipRanges = append(ipRanges, ipRange)
+		}
+	}
+
+	if subnet.AddressPrefixes == nil {
+		return ipRanges
+	}
+
+	for _, snetPrefix := range *subnet.AddressPrefixes {
+		_, ipRange, err := net.ParseCIDR(snetPrefix)
+		if err == nil {
+			ipRanges = append(ipRanges, ipRange)
+		}
+	}
+	return ipRanges
+}
+
+
+// getAllDevSubnets queries azure to retrieve all subnets assigned the vnet
+// `dev-vnet` in the current resource group
+func (c *Cluster) getAllDevSubnets() ([]mgmtnetwork.Subnet, error) {
+	allSubnets := []mgmtnetwork.Subnet{}
+	availSnetResults, err := c.subnets.List(context.TODO(), c.env.ResourceGroup(), "dev-vnet")
+	if err != nil {
+		return allSubnets, err
+	}
+	allSubnets = append(allSubnets, availSnetResults.Values()...)
+	for availSnetResults.NotDone() {
+		err = availSnetResults.NextWithContext(context.TODO())
+		if err != nil {
+			break
+		}
+		allSubnets = append(allSubnets, availSnetResults.Values()...)
+	}
+
+	return allSubnets, nil
+}
+
+
 func (c *Cluster) generateSubnets() (vnetPrefix string, masterSubnet string, workerSubnet string) {
 	// pick a random 23 in range [10.3.0.0, 10.127.255.0]
 	// 10.0.0.0/16 is used by dev-vnet to host CI
 	// 10.1.0.0/24 is used by rp-vnet to host Proxy VM
 	// 10.2.0.0/24 is used by dev-vpn-vnet to host VirtualNetworkGateway
-	var x, y int
-	// Local Dev clusters are limited to /16 dev-vnet
-	if !c.ci {
-		x, y = 0, 2*rand.Intn(128)
-	} else {
-		x, y = rand.Intn((124))+3, 2*rand.Intn(128)
+
+	allSubnets, err := c.getAllDevSubnets()
+	if err != nil {
+		c.log.Warnf("Error getting existing subnets. Continuing regardless: %v", err)
 	}
-	vnetPrefix = fmt.Sprintf("10.%d.%d.0/23", x, y)
-	masterSubnet = fmt.Sprintf("10.%d.%d.0/24", x, y)
-	workerSubnet = fmt.Sprintf("10.%d.%d.0/24", x, y+1)
-	return
+
+	ipRanges := []*net.IPNet{}
+	for _, snet := range allSubnets {
+		ipRanges = append(ipRanges, GetIpRangesFromSubnet(snet)...)
+	}
+
+	for i := 1; i < 100; i++ {
+		var x, y int
+		// Local Dev clusters are limited to /16 dev-vnet
+		if !c.ci {
+			x, y = 0, 2*rand.Intn(128)
+		} else {
+			x, y = rand.Intn((124))+3, 2*rand.Intn(128)
+		}
+		c.log.Infof("Generate Subnet try: %d\n", i)
+		vnetPrefix = fmt.Sprintf("10.%d.%d.0/23", x, y)
+		masterSubnet = fmt.Sprintf("10.%d.%d.0/24", x, y)
+		workerSubnet = fmt.Sprintf("10.%d.%d.0/24", x, y+1)
+		if !ipRangesContainCIDR(ipRanges, workerSubnet) && !ipRangesContainCIDR(ipRanges, masterSubnet) {
+			return vnetPrefix, masterSubnet, workerSubnet
+		}
+	}
+  
+	return vnetPrefix, masterSubnet, workerSubnet
 }
 
 func (c *Cluster) Delete(ctx context.Context, vnetResourceGroup, clusterName string) error {
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -67,7 +67,6 @@ type Cluster struct {
 	vaultsClient         armkeyvault.VaultsClient
 }
 
-
 const GenerateSubnetMaxTries = 100
 
 func New(log *logrus.Entry, environment env.Core, ci bool) (*Cluster, error) {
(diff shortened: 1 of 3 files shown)


pkg/util/cluster/cluster.go
```diff
@@ -356,22 +357,103 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
...
+
+	return allSubnets, nil
+}
+
+
 func (c *Cluster) generateSubnets() (vnetPrefix string, masterSubnet string, workerSubnet string) {
 	// pick a random 23 in range [10.3.0.0, 10.127.255.0]
 	// 10.0.0.0/16 is used by dev-vnet to host CI
 	// 10.1.0.0/24 is used by rp-vnet to host Proxy VM
 	// 10.2.0.0/24 is used by dev-vpn-vnet to host VirtualNetworkGateway
```
tsatam, line 427: Can we update this comment to indicate the changes here?

tsatam (changes requested):
Changes LGTM.

Current state of master is broken, you'll need to rebase once #3705 is merged. 

Make sure to run `make generate` and fix the linter errors to pass the other failing checks as well.

pkg/util/cluster/cluster.go
```diff
@@ -356,22 +357,103 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
...
+		} else {
+			x, y = rand.Intn((124))+3, 2*rand.Intn(128)
+		}
+		c.log.Infof("Generate Subnet try: %d\n", i)
+		vnetPrefix = fmt.Sprintf("10.%d.%d.0/23", x, y)
+		masterSubnet = fmt.Sprintf("10.%d.%d.0/24", x, y)
+		workerSubnet = fmt.Sprintf("10.%d.%d.0/24", x, y+1)
+		if !ipRangesContainCIDR(ipRanges, workerSubnet) && !ipRangesContainCIDR(ipRanges, masterSubnet) {
+			return vnetPrefix, masterSubnet, workerSubnet
+		}
```
bitoku, line 453: If we change this part, we can stop random try approach and check vacant subnets instead.

mrWinston: What do you mean exactly? Like iterating through all possible subnets in order instead of generating them randomly?
At the end, we'd still need to check for each subnet individually if it overlaps and the random approach would probably find an empty slot faster.

pkg/util/cluster/cluster.go
```diff
@@ -356,22 +357,103 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 	return nil
 }
 
+// ipRangesContainCIDR checks, weather any of the ipRanges overlap with the cidr string. In case cidr isn't valid, false is returned.
+func ipRangesContainCIDR(ipRanges []*net.IPNet, cidr string) bool {
+	_, cidrNet, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return false
+	}
```
AldoFusterTurpin, line 365: I am not sure if this is a good idea as in _generateSubnets()_ we actually return a "good" result when _ipRangesContainCIDR_ returns false, so maybe this could bring wrong values when returning "false" from _generateSubnets()_ due to an error and not due to a real non overlapping condition. I think it would be great to return an error from this function (if any)

Something like:
```go
// ipRangesContainCIDR checks, weather any of the ipRanges overlap with the cidr string
// and an error, if any.
func ipRangesContainCIDR(ipRanges []*net.IPNet, cidr string) (bool, error) {
	_, cidrNet, err := net.ParseCIDR(cidr)
	if err != nil {
		return false, err
	}

	for _, snet := range ipRanges {
		if snet.Contains(cidrNet.IP) || cidrNet.Contains(snet.IP) {
			return true, nil
		}
	}
	return false, nil
}
```
 and then in _generateSubnets_ change the if condition (check next comment).

pkg/util/cluster/cluster.go
```diff
@@ -356,22 +357,103 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
...
+	return ipRanges
+}
+
+
+// getAllDevSubnets queries azure to retrieve all subnets assigned the vnet
+// `dev-vnet` in the current resource group
+func (c *Cluster) getAllDevSubnets() ([]mgmtnetwork.Subnet, error) {
+	allSubnets := []mgmtnetwork.Subnet{}
+	availSnetResults, err := c.subnets.List(context.TODO(), c.env.ResourceGroup(), "dev-vnet")
```
AldoFusterTurpin, line 406: Super small nit: I think context.TODO() is usually used to indicate that we are not sure what context to use and hence we create one empty but using TODO to indicate that it probably needs to be revisited in the future. 

I think that in our case, context.Background() is a bit better as it indicates that it is a plain new context. Just a semantic difference that has no impact on the result 🙂

pkg/util/cluster/cluster.go
```diff
@@ -356,22 +357,103 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
...
+	if err != nil {
+		return allSubnets, err
+	}
+	allSubnets = append(allSubnets, availSnetResults.Values()...)
+	for availSnetResults.NotDone() {
+		err = availSnetResults.NextWithContext(context.TODO())
```
AldoFusterTurpin, line 412: Super small nit: I think context.TODO() is usually used to indicate that we are not sure what context to use and hence we create one empty but using TODO to indicate that it probably needs to be revisited in the future. 

I think that in our case, context.Background() is a bit better as it indicates that it is a plain new context. Just a semantic difference that has no impact on the result 🙂

pkg/util/cluster/cluster.go
AldoFusterTurpin: IMHO the random approach is better.

AldoFusterTurpin, line 453: Taking into account the consideration I mentioned about the _ipRangesContainCIDR_, this could be changed to:

```go
workerSubnetOverlaps, err := ipRangesContainCIDR(ipRanges, workerSubnet)
if err != nil || workerSubnetOverlaps {
    continue
}
masterSubnetSubnetOverlaps, err := ipRangesContainCIDR(ipRanges, masterSubnet)
if err != nil || masterSubnetSubnetOverlaps {
    continue
}

return vnetPrefix, masterSubnet, workerSubnet
```
which is a bit more explicit and can not produce the false positive when there is an error in _ipRangesContainCIDR_

🙂

AldoFusterTurpin: We could go even one step further and return an error from _generateSubnets()_ when after trying 100 times it was not possible to get valid results. 

Right now we are returning the last values of vnetPrefix, masterSubnet, workerSubnet that we retrieved. It is very unlikely that this happens but we could be returning the wrong values without explicitly stating that. If we want to overcome that, we can simply return an error value outside the loop and check the error from the outside, like:
```go
func (c *Cluster) generateSubnets() (vnetPrefix string, masterSubnet string, workerSubnet string, err error) {
	allSubnets, err := c.getAllDevSubnets()
	if err != nil {
		c.log.Warnf("Error getting existing subnets. Continuing regardless: %v", err)
	}

	ipRanges := []*net.IPNet{}
	for _, snet := range allSubnets {
		ipRanges = append(ipRanges, GetIpRangesFromSubnet(snet)...)
	}

	nTries := 100
	for i := 1; i < nTries; i++ {
		var x, y int
		if !c.ci {
			x, y = 0, 2*rand.Intn(128)
		} else {
			x, y = rand.Intn(124)+3, 2*rand.Intn(128)
		}
		c.log.Infof("Generate Subnet try: %d\n", i)
		vnetPrefix = fmt.Sprintf("10.%d.%d.0/23", x, y)
		masterSubnet = fmt.Sprintf("10.%d.%d.0/24", x, y)
		workerSubnet = fmt.Sprintf("10.%d.%d.0/24", x, y+1)

		workerSubnetOverlaps, err := ipRangesContainCIDR(ipRanges, workerSubnet)
		if err != nil || workerSubnetOverlaps {
			continue
		}

		masterSubnetSubnetOverlaps, err := ipRangesContainCIDR(ipRanges, masterSubnet)
		if err != nil || masterSubnetSubnetOverlaps {
			continue
		}

		return vnetPrefix, masterSubnet, workerSubnet, nil
	}

	return "", "", "", fmt.Errorf("was not able to generate master and worker subnets after %v tries", nTries)
}
```

This has the benefit that clearly returns the zeroed default values of the strings, which  is good practice in Go when returning an error, instead of the last value retrieved. 
Maybe not needed in this case but just an idea.
🙂

AldoFusterTurpin (dismissed):
@mrWinston Thank you for this. Looks good. I just left some suggestions in case you find them useful. Thanks!

pkg/util/cluster/cluster.go
bitoku: >  Like iterating through all possible subnets in order instead of generating them randomly?
Yes.

I won't strongly oppose to the random approach, but when the vnet is congested, this approach might fail even though there are available subnets.
According to the code, if we take in-order approach, it iterates only 128 times at most, so it won't be that slow.
If you're concerning the early several subnets are always checked, we can pick the starting subnets randomly and then iterate afterward in order.

Anyway, this is just nit.

pkg/util/cluster/cluster.go
mrWinston: ah, yes, that's a really good point!

pkg/util/cluster/cluster.go
mrWinston: +1 to adding `error` to the return values of `generateSubnets`

mrWinston: I'll leave it then as is for now. If we encounter any problems with the random approach, we can still refactor it to be more structured.

tsatam (approved):

pkg/util/cluster/cluster.go
```diff
@@ -345,22 +352,111 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
...
+			break
+		}
+		allSubnets = append(allSubnets, availSnetResults.Values()...)
+	}
+
+	return allSubnets, nil
+}
+
+func (c *Cluster) generateSubnets() (vnetPrefix string, masterSubnet string, workerSubnet string, err error) {
+	// pick a random 23 in range [10.3.0.0, 10.127.255.0], making sure it doesn't
```
SudoBrendan, line 417: idea: I wonder if this randomness is actually what shoots us in the foot. What if instead, we got very structured with this.

What if, given that both masters and workers are /24 nets, we adopted a CIDR convention rather than shooting in the dark x many times? I'm thinking something like the 10.3.0.0/ space is entirely dedicated to dev clusters, and we take that a slice at a time:

```
10.3.0.0/24
10.3.1.0/24
10.3.2.0/24
10.3.3.0/24
10.3.4.0/24
10.3.5.0/24
10.3.6.0/24
...
```

Now, to find the next "open" subnet, all you have to do is iterate over this CIDR convention, and see if it shows up in the VNet or not, and it supports a scale I think we can easily handle (255 subnets, or 127 clusters per region). Thoughts?

SudoBrendan (approved):
Ideas aside, this worked for me on the first try locally, so huge +1 considering this is a critical blocker on local testing, thank you so much for this.

pkg/util/cluster/cluster.go
tsatam: It was discussed here as a potential follow-up improvement: https://github.com/Azure/ARO-RP/pull/3706#discussion_r1684208163

pkg/util/cluster/cluster.go
```diff
@@ -345,22 +352,111 @@ func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName str
 	return nil
 }
 
-func (c *Cluster) generateSubnets() (vnetPrefix string, masterSubnet string, workerSubnet string) {
-	// pick a random 23 in range [10.3.0.0, 10.127.255.0]
+// ipRangesContainCIDR checks, weather any of the ipRanges overlap with the cidr string. In case cidr isn't valid, false is returned.
```
kevinobriendotca, line 355: nit: weather should be whether

jhoreman (approved):
Looks great, thanks for this!

shubhadapaithankar (approved):
LGTM

mociarain (approved):

PR #3708, opened 2024-07-19
refactor package pki to be unit testeable

AldoFusterTurpin:

Makes the package pki unit testeable decoupling the HTTP calls with the business logic and removes the unused "fake" unit test we are skipping in the CI. It also removes the unused state that was maintained just for the sake of the previous unit test (technically not a unit test as was touching HTPP and was flaky).

We still need to decide what needs to be tested and if the unit test that we will create is enough.

The main idea is to change from: 
- function which performs HTTP call + business logic 
to 
- a public function that makes the HTTP call + another function that is unit testeable that uses the previous result.

I think caMap, mu, getCaCertPoolFromMap and setCaCertPoolInMap can be deleted.

**Slack [thread](https://redhat-internal.slack.com/archives/C02ULBRS68M/p1707324423230289) with context.**

I have also refactored _newPipeline_ that now calls _getClientOptions_ which contains the logic to build the clientOptions in one function as IMHO the current usage of the initialised cp to nil is not straightforward to reason about. i,e: we are calling _NewClientOptions(nil)_ implicitly when options == nil and issuerUrlTemplate == "" || caName == "", something that now is explicit and clear.

I have also introduced an early return in the functions to make things a bit clearer.

We need to stop having a flaky hidden integration test as part of our regular unit tests. Unit tests should never touch the external world (I/O, netowrking, etc.). This PR also removes the unnecessary global state that was created just to be used in the test of the HTTP Get calls.

### Test plan for issue:

I have removed the "fake" unit test that existed in pkg/util/pki/pki.go because it was calling the HTTP endpoint and checking that the result was not nil. It was not even testing explicitly any result of business logic after performing the HTTP calls, so I think at this point it does not bring much value. On top of that, we are most of the time skipping this test in the CI (using the env var) because it is almost always failing. 

One last note: I have refactored the code to be unit-testable. If we really want to test the objects without the external HTTP calls we can do that in the near future, but a different kind of test would be more valuable.

N/A


AldoFusterTurpin:
For reference: this was previously from my fork: https://github.com/Azure/ARO-RP/pull/3389

pkg/util/pki/fetchdata.go
```diff
@@ -0,0 +1,33 @@
...
+// returns the ame certs
+func FetchDataFromGetIssuerPki(url string) (*RootCAs, error) {
+	response, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+
+	defer response.Body.Close()
+
+	// Read in certs from endpoint
```
mociarain, line 24: Do we need this comment?

pkg/util/pki/pki_test.go
```diff
@@ -1,61 +0,0 @@
...
-	"crypto/tls"
-	"crypto/x509"
-	"fmt"
-	"net/http"
-	"os"
-	"testing"
-)
-
-func TestGetTlsConfig(t *testing.T) {
-	if os.Getenv("ARO_SKIP_PKI_TESTS") != "" {
```
mociarain, line 16: Should we remove this env var as part of this PR?

mociarain (commented):
Nice change. Just 2 small questions

pkg/util/pki/fetchdata.go
AldoFusterTurpin: It was there, but we don't need it. Removed. Thanks.

pkg/util/pki/pki_test.go
AldoFusterTurpin: Yes. Done. Thanks.

mociarain (approved):
LGTM

anshulvermapatel (approved):
Thanks for this.

PR #3711, opened 2024-07-20
Added reconcile methods to AROReconciler interface

ehvs:

Picking up from @mbarnes , rebasing and making some minor tweaks to adjust solution.

This is loosely related to the epic [[ARO-1627] Propagate Errors and Statuses of ARO Controllers to ARO Cluster Operator](https://issues.redhat.com/browse/ARO-1627).  It's an enhancement to the `AROController` base type that was developed for that epic.

All of ARO's controllers can be disabled through operator flags in the cluster object, and so every `Reconcile` function has a common preamble: fetch the cluster object, check the appropriate operator flag, and short-circuit reconciliation if the flag is set to disabled.

This pull request moves that preamble into `AROController` for consistent handling of disabled controllers, and allows individual controllers to focus exclusively on their business logic.  It also introduces a common interface for controllers to simplify registration during startup.

### Test plan for issue:

Add unit tests for the `Reconcile` logic in `AROController`, otherwise rely on existing unit and E2E tests.

I verified the controllers using `AROController` start up by running the operator locally with `-loglevel=debug` and saw lots of messages like
```
pkg/operator/controllers/base/aro_controller.go:53 base.(*AROController).Reconcile() running
```

No, this is strictly an internal refactoring.

### Additional notes

For reviewers, I suggest reviewing one commit at a time.  I structured the commits as an incremental progression.

Also please note the commit messages, especially in the last commit where I tried to leave a detailed explanation.

### DIFF
--- a/pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
+++ b/pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
@@ -27,7 +27,8 @@ import (
 )
 
 const (
-	ControllerName = "CloudProviderConfig"
+	controllerName    = "CloudProviderConfig"
+	controllerEnabled = operator.CloudProviderConfigEnabled // calling from flags.go due to removal on #3327
 )
 
 var cloudProviderConfigName = types.NamespacedName{Name: "cloud-provider-config", Namespace: "openshift-config"}
@@ -133,31 +134,23 @@ type CloudProviderConfigReconciler struct {
 	base.AROController
 }
 
-func NewReconciler(Log *logrus.Entry, client client.Client) *CloudProviderConfigReconciler {
-	return &CloudProviderConfigReconciler{
+func NewReconciler(log *logrus.Entry, client client.Client) *CloudProviderConfigReconciler {
+	r := &CloudProviderConfigReconciler{
 		AROController: base.AROController{
-			Log:    Log,
-			Client: client,
-			Name:   ControllerName,
+			Log:         log.WithField("controller", controllerName),
+			Client:      client,
+			Name:        controllerName,
+			EnabledFlag: controllerEnabled,
 		},
 	}
+	r.Reconciler = r
+	return r
 }
 
 // Reconcile makes sure that the cloud-provider-config is healthy
-func (r *CloudProviderConfigReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+func (r *CloudProviderConfigReconciler) ReconcileEnabled(ctx context.Context, request ctrl.Request, instance *arov1alpha1.Cluster) (ctrl.Result, error) {
 	r.Log.Debug("reconcile ConfigMap openshift-config/cloud-provider-config")
 
-	instance, err := r.GetCluster(ctx)
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-
-	if !instance.Spec.OperatorFlags.GetSimpleBoolean(operator.CloudProviderConfigEnabled) {
-		r.Log.Debug("controller is disabled")
-		return reconcile.Result{}, nil
-	}
-
-	r.Log.Debug("running")
 	return reconcile.Result{}, r.updateCloudProviderConfig(ctx)
 }
 
@@ -176,7 +169,7 @@ func (r *CloudProviderConfigReconciler) SetupWithManager(mgr ctrl.Manager) error
 			&handler.EnqueueRequestForObject{},
 			builder.WithPredicates(cloudProviderConfigPredicate),
 		).
-		Named(ControllerName).
+		Named(r.GetName()).
 		Complete(r)
 }
 
--- a/pkg/operator/controllers/dnsmasq/machineconfig_controller.go
+++ b/pkg/operator/controllers/dnsmasq/machineconfig_controller.go
@@ -16,12 +16,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/Azure/ARO-RP/pkg/operator"
+	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 	"github.com/Azure/ARO-RP/pkg/operator/controllers/base"
 	"github.com/Azure/ARO-RP/pkg/util/dynamichelper"
 )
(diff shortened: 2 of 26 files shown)


pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -133,31 +134,23 @@ type CloudProviderConfigReconciler struct {
...
+			EnabledFlag: controllerEnabled,
 		},
 	}
+	r.Reconciler = r
+	return r
 }
 
 // Reconcile makes sure that the cloud-provider-config is healthy
-func (r *CloudProviderConfigReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+func (r *CloudProviderConfigReconciler) ReconcileEnabled(ctx context.Context, request ctrl.Request, instance *arov1alpha1.Cluster) (ctrl.Result, error) {
```
stevekuznetsov, line 151: nit: `...Reconciler.ReconcileEnabled()` stutters, consider `...Reconciler.Enabled()`

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
```diff
@@ -133,31 +134,23 @@ type CloudProviderConfigReconciler struct {
...
-			Log:    Log,
-			Client: client,
-			Name:   ControllerName,
+			Log:         log.WithField("controller", controllerName),
+			Client:      client,
+			Name:        controllerName,
+			EnabledFlag: controllerEnabled,
 		},
 	}
+	r.Reconciler = r
```
stevekuznetsov, line 146: This recursive self-reference may indicate that a better type factoring exists. Is it possible to write a type that allows for composable components - which may share common references to shared input parameters - that is not self-referential?

pkg/operator/controllers/dnsmasq/machineconfig_controller.go
```diff
@@ -16,12 +16,13 @@ import (
...
 
 	"github.com/Azure/ARO-RP/pkg/operator"
+	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 	"github.com/Azure/ARO-RP/pkg/operator/controllers/base"
 	"github.com/Azure/ARO-RP/pkg/util/dynamichelper"
 )
 
 const (
-	MachineConfigControllerName = "DnsmasqMachineConfig"
+	machineConfigControllerName = "DnsmasqMachineConfig"
```
stevekuznetsov, line 25: nit: super small thing, and may be 100% personal preference, but I find it's easier to separate entirely mechanical changes like changing all the names to be private into a separate commit - it keeps the core logic changes that need good scrutiny in a small change-set that's easy to review

pkg/operator/controllers/dnsmasq/machineconfigpool_controller.go
```diff
@@ -30,35 +31,29 @@ type MachineConfigPoolReconciler struct {
...
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-
-	if !instance.Spec.OperatorFlags.GetSimpleBoolean(operator.DnsmasqEnabled) {
-		r.Log.Debug("controller is disabled")
-		return reconcile.Result{}, nil
-	}
+func (r *MachineConfigPoolReconciler) ReconcileEnabled(ctx context.Context, request ctrl.Request, instance *arov1alpha1.Cluster) (ctrl.Result, error) {
+	var err error
```
stevekuznetsov, line 50: nit: where possible, don't allow short-lived values to escape their smallest possible scope. For instance, something like

```go
if err := something(); err != nil {
}
```

Can help to reduce mistakes - one global `err` that gets mutated over and over again won't be caught by the compiler if, for instance, one of the individual mutations fails to check `err != nil` once more.

pkg/operator/controllers/genevalogging/genevalogging_controller.go
```diff
@@ -41,14 +42,17 @@ type Reconciler struct {
 }
 
 func NewReconciler(log *logrus.Entry, client client.Client, dh dynamichelper.Interface) *Reconciler {
-	return &Reconciler{
+	r := &Reconciler{
 		AROController: base.AROController{
-			Log:    log,
-			Client: client,
-			Name:   ControllerName,
+			Log:         log.WithField("controller", controllerName),
```
stevekuznetsov, line 47: If we pass the controller's name into the `AROController` instance and we expect every time a new `AROController` is instantiated, the logger must be updated with a `"controller"` field, perhaps we can move that into an `NewAROController()` constructor.

pkg/operator/controllers/imageconfig/image_controller.go
```diff
@@ -26,42 +26,35 @@ import (
 )
 
 const (
-	ControllerName = "ImageConfig"
+	controllerName = "ImageConfig"
 	// Kubernetes object name
 	imageConfigResource = "cluster"
+	controllerEnabled   = operator.ImageConfigEnabled
```
stevekuznetsov, line 32: Question for me - when would we choose to have the indirection here (having the package-level constant) versus not? It looks like the value is only used once.

pkg/operator/controllers/cloudproviderconfig/cloudproviderconfig_controller.go
mbarnes: This is a way of emulating dynamic method dispatching in Go, since it's not native to the language.  (e.g. when `AROController` calls `c.Reconciler.ReconcileEnabled` we want to call `CloudProviderConfig`'s implementation)  But 20+ years of using object-oriented programming languages has me sometimes stuck in that mode of thinking.  Maybe there's a better way to organize in Go.

stevekuznetsov: Yes, I see the intent - it's very far from idiomatic, though. Composition of small atoms would feel much better.

pkg/util/mocks/env/core.go
SudoBrendan, line 1: flagging this as potentially needing a rebase on master to pick up the .bingo changes. Dismiss if untrue (but it probably doesn't hurt to do it anyways).

SudoBrendan (changes requested):

PR #3713, opened 2024-07-23
Add scripts for creating and persisting mock cluster MSI

niontive:
### Which issue this PR addresses:

Fixes https://issues.redhat.com/browse/ARO-8986

### What this PR does / why we need it:

Create a service principal for local development so we can mock cluster MSI

### Test plan for issue:

Local testing

### Is there any documentation that needs to be updated for this PR?

Yes

PR #3714, opened 2024-07-23
Add middleware for mock MSI headers

niontive:

Fixes https://issues.redhat.com/browse/ARO-9193

Middleware to mock MSI headers. For tenant ID, this should be set by developer. See this PR: https://github.com/Azure/ARO-RP/pull/3713/files

### Test plan for issue:

Local/unit tests


pkg/frontend/frontend.go
```diff
@@ -231,8 +231,14 @@ func (f *frontend) chiAuthenticatedRoutes(router chi.Router) {
 				r.With(f.apiVersionMiddleware.ValidateAPIVersion).Route("/", func(r chi.Router) {
 					r.Delete("/", f.deleteOpenShiftCluster)
 					r.Get("/", f.getOpenShiftCluster)
-					r.Patch("/", f.putOrPatchOpenShiftCluster)
-					r.Put("/", f.putOrPatchOpenShiftCluster)
+
+					if f.env.IsLocalDevelopmentMode() {
+						r.With(middleware.MockMSIMiddleware).Patch("/", f.putOrPatchOpenShiftCluster)
```
rajdeepc2792, line 236: What is the idea here to use the middleware for a Patch request? Since we are assuming that the Cluster MSI/Identity object will not be changed. 
As per MSI RP it is considered successful no operation, so it should be fine to use the middleware here.
But can we just skip this middleware part for Patch, as because not getting a identityURL for an empty Identity object is also a successful no operation and valid here?

rajdeepc2792: I also want to raise an additional point here around patch request, that might be related here.
How to handle the case where cx patches the identity object such that it removes the msi from the object(doesn't delete the identity though)(This request will return from ARO RP as Identity object changed which is immutable), but cx later on patches with exact same identity object as was present previously?

The old MSI Certificate will no longer work as it will be deleted when cx patched the identity object first time. Should we somehow refresh credential via a patch request or just ignore the whole scenario as an unsupported cluster patching?

cadenmarchese: My understanding is that the Cluster MSI headers are still returned on a PATCH, not just a CREATE, so we want to mock both instead of just create. I'm not sure on your second point, though...

niontive: 1. The headers are part of PATCH operation: https://eng.ms/docs/products/arm/rbac/managed_identities/msionboardingsteps
2. Based on our previous architecture discussions, we agreed that we shouldn't support removing/modifying the cluster MSI. If we want to support that in the future, we can.

kimorris27: >I also want to raise an additional point here around patch request, that might be related here.
How to handle the case where cx patches the identity object such that it removes the msi from the object(doesn't delete the identity though)(This request will return an error from ARO RP as Identity object got changed, which is immutable), but cx later on patches with exact same identity object as was present previously?
>
>The old MSI Certificate will no longer work as it will be deleted when cx patched the identity object first time. Should we somehow refresh credential via the second patch request or just ignore the whole scenario as an unsupported cluster patching?

I think this is currently impossible since the `Identity` is immutable, so we can just ignore the whole scenario unless we later decide to support removing/modifying the MSI.

kimorris27 (approved):

kimorris27 (dismissed):

cadenmarchese (dismissed):

pkg/frontend/middleware/msi_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	// Create a new HTTP request
+	req := httptest.NewRequest(http.MethodGet, "http://example.com", nil)
+	// Create a response recorder to capture the response
+	rr := httptest.NewRecorder()
+
+	// Serve the HTTP request using the wrapped handler
+	handler.ServeHTTP(rr, req)
+
+	// Check the response headers
```
mociarain, line 35: We can remove all of these comments

mociarain (changes requested):
Other than the comments LGTM

rhamitarora (dismissed):

kimorris27 (approved):

pkg/frontend/frontend.go
rajdeepc2792: Yes we can mark the cluster unsupported if the MSI is tried to be removed from the cluster object as part of PUT request.

@kimorris27, yes the flow will fail as part of ARO-RP but the MSI-RP would consider it deleted and then there will be a inconsistency between MIRP and ARO-RP. Apologies, IDK why I didn't attach the ref earlier:- https://msazure.visualstudio.com/One/_git/AzureUX-ARMWiki?path=/src/rbac/managed_identities/MSIOnboardingUserAssigned.md&version=GBmaster&line=1&_a=preview&anchor=client-error-when-more-than-1-identity-is-configured-on-the-resource

PR #3717, opened 2024-07-23
Add load balancer smoke test

bitoku:

Fixes https://issues.redhat.com/browse/ARO-7350

This PR adds the test checking whether it can create a load balancer and the service is reachable.

### Test plan for issue:

N/A

This is a test. not production feature.

### DIFF
--- a/test/e2e/ocp_loadbalancer.go
+++ b/test/e2e/ocp_loadbalancer.go
@@ -1,0 +1,83 @@
+package e2e
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+var _ = Describe("Load Balancer", Label(smoke), Ordered, func() {
+	var objs []unstructured.Unstructured
+	BeforeAll(func(ctx context.Context) {
+		// Initialize objs to avoid objs being counted multiple times when retrying.
+		objs = []unstructured.Unstructured{}
+		By("creating a load balancer")
+		lb, err := staticResources.Open("static_resources/loadbalancer.yaml")
+		Expect(err).NotTo(HaveOccurred())
+		defer func() {
+			_ = lb.Close()
+		}()
+
+		dec := yaml.NewYAMLOrJSONDecoder(lb, 4096)
+		// It can't load multiple objects from a single file, so we need to loop through the file and load them one by one.
+		for {
+			var obj unstructured.Unstructured
+			err := dec.Decode(&obj)
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			Expect(err).NotTo(HaveOccurred())
+			objs = append(objs, obj)
+		}
+
+		for _, obj := range objs {
+			By(fmt.Sprintf("creating %s/%s", obj.GetNamespace(), obj.GetName()))
+			cli, err := clients.Dynamic.GetClient(&obj)
+			Expect(err).NotTo(HaveOccurred())
+			CreateK8sObjectWithRetry(ctx, cli.Create, &obj, metav1.CreateOptions{})
+		}
+	})
+
+	It("should create a service", func(ctx context.Context) {
+		var ip string
+		By("waiting for the service to be created and get an IP address")
+		Eventually(func(g Gomega, ctx context.Context) {
+			svc, err := clients.Kubernetes.CoreV1().Services("default").Get(ctx, "test-lb", metav1.GetOptions{})
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(svc.Status.LoadBalancer.Ingress).NotTo(BeEmpty())
+			ip = svc.Status.LoadBalancer.Ingress[0].IP
+		}, DefaultEventuallyTimeout, 10*time.Second, ctx).Should(Succeed())
+
+		By("checking the service is reachable and it returns a 200 OK")
+		Eventually(func(g Gomega, ctx context.Context) {
+			client := http.Client{Timeout: 10 * time.Second}
+			req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("http://%s", ip), nil)
+			g.Expect(err).NotTo(HaveOccurred())
+			resp, err := client.Do(req)
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(resp.StatusCode).To(Equal(http.StatusOK))
+		}, DefaultEventuallyTimeout, 10*time.Second, ctx).Should(Succeed())
+	})
+
+	AfterAll(func(ctx context.Context) {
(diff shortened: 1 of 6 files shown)


kimorris27 (commented):
LGTM if E2E passes.

test/util/dynamic/dynamic.go
```diff
@@ -0,0 +1,91 @@
...
+	Create(ctx context.Context, obj *unstructured.Unstructured, options metav1.CreateOptions) (*unstructured.Unstructured, error)
+	Delete(ctx context.Context, name string, options metav1.DeleteOptions) error
+}
+
+type resourceClient struct {
+	client dynamic.ResourceInterface
+}
+
+func (d *resourceClient) Get(ctx context.Context, name string, options metav1.GetOptions) (*unstructured.Unstructured, error) {
+	// Get the object without the subresources argument to align with methods of other clients.
```
mociarain, line 79: I don't understand this comment. What other clients are you referring to?

test/util/dynamic/dynamic.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func NewDynamicClient(kubeConfig *rest.Config) (Client, error) {
+	// Create a dynamic client
+	cli, err := dynamic.NewForConfig(kubeConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create a discovery mapper
```
mociarain, line 36: Nit: I think we should remove this comment

test/util/dynamic/dynamic.go
```diff
@@ -0,0 +1,91 @@
...
+	GetClient(obj *unstructured.Unstructured) (ResourceClient, error)
+}
+
+type client struct {
+	dynamic.Interface
+	mapping meta.RESTMapper
+}
+
+func NewDynamicClient(kubeConfig *rest.Config) (Client, error) {
+	// Create a dynamic client
```
mociarain, line 30: Nit: I think we should remove this comment

test/e2e/setup.go
```diff
@@ -49,13 +50,17 @@ import (
...
 	"github.com/Azure/ARO-RP/pkg/util/version"
+	"github.com/Azure/ARO-RP/test/util/dynamic"
 	"github.com/Azure/ARO-RP/test/util/kubeadminkubeconfig"
 )
 
 const (
 	smoke = "smoke"
 )
 
+//go:embed static_resources
```
mociarain, line 61: TIL about embedded stuff in go. I found this useful btw: https://www.scaler.com/topics/golang/golang-embed/

test/e2e/ocp_loadbalancer.go
```diff
@@ -0,0 +1,83 @@
...
+		objs = []unstructured.Unstructured{}
+		By("creating a load balancer")
+		lb, err := staticResources.Open("static_resources/loadbalancer.yaml")
+		Expect(err).NotTo(HaveOccurred())
+		defer func() {
+			_ = lb.Close()
+		}()
+
+		dec := yaml.NewYAMLOrJSONDecoder(lb, 4096)
+		// Load all objects in the yaml file
```
mociarain, line 35: Nit: I think we should remove this comment

mociarain (changes requested):
Few small things but this is great

test/util/dynamic/dynamic.go
bitoku: https://pkg.go.dev/k8s.io/client-go/dynamic#ResourceInterface
This is dynamic client interface.

For example,
https://pkg.go.dev/k8s.io/client-go@v0.30.3/kubernetes/typed/core/v1#ServiceInterface
This is service client interface.

Because of the difference, it doesn't fit to these types.
https://github.com/Azure/ARO-RP/blob/e5339b26b2fecb25993b533efc7f1fab40bb255f/test/e2e/helpers.go#L29-L32

test/e2e/ocp_loadbalancer.go
bitoku: I'll keep this one, but will make it clear.

test/util/dynamic/dynamic.go
bitoku: but the comments are a bit too redundant. I'll put the explanation somewhere else and remove those comments.

test/e2e/setup.go
bitoku: It's used in guardrail as well.

https://github.com/Azure/ARO-RP/blob/40e7987ef8d88112e081a48e288f1dec41c627ec/pkg/operator/controllers/guardrails/guardrails_config.go#L71-L78

mociarain (dismissed):
LGTM

test/util/dynamic/dynamic.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if obj.GetNamespace() != "" {
+		return &resourceClient{d.Resource(gvr.Resource).Namespace(obj.GetNamespace())}, nil
+	}
+	return &resourceClient{d.Resource(gvr.Resource)}, nil
+}
+
+// ResourceClient is an interface that can be used for *K8sObjectWithRetry helper functions.
+// The default dynamic ResourceInterface can't be used for *K8sObjectWithRetry
+// because it has different method signatures from other clients interface.
```
stevekuznetsov, line 67: nit: it would be clearer to mention that the reason for this diff is thata dynamic clients support actions over any subresource, whereas the typed clients break those out into extensions.

test/util/dynamic/dynamic.go
```diff
@@ -0,0 +1,89 @@
...
+	return d.mapping.RESTMapping(gvk.GroupKind(), gvk.Version)
+}
+
+// GetClient returns a non-namespaced or namespaced ResourceClient depending on a given object.
+func (d *client) GetClient(obj *unstructured.Unstructured) (ResourceClient, error) {
+	gvr, err := d.getGVR(obj)
+	if err != nil {
+		return nil, err
+	}
+	if obj.GetNamespace() != "" {
```
stevekuznetsov, line 59: nit: if we want to be perfectly correct, we need to use discovery to determine if the GVR is namespaced or not - a user passing in a namespaced object but forgetting to set a namespace will cause a subtle error here

test/util/dynamic/dynamic.go
```diff
@@ -0,0 +1,89 @@
...
+
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(kubeConfig)
+	if err != nil {
+		return nil, err
+	}
+	groupResources, err := restmapper.GetAPIGroupResources(discoveryClient)
+	if err != nil {
+		return nil, err
+	}
+	mapper := restmapper.NewDiscoveryRESTMapper(groupResources)
```
stevekuznetsov, line 43: If at all possible, use a deferred discovery mapper - otherwise, this will eagerly send a request storm to grab all discovery information at once.

mociarain:
Just as a note please don't force push your response to comments. This makes diff'ing the changes very difficult and re-reviews harded

mociarain (approved):
Nice change

kimorris27 (approved):

kimorris27:
> Just as a note please don't force push your response to comments. This makes diff'ing the changes very difficult and re-reviews harder

@mociarain, I had trouble diff'ing the changes for my re-review as well, but I don't think force pushing alone is the issue.

@bitoku, IIUC you amended the 2nd commit and then rebased and force pushed. So I would just recommend keeping in mind that rebasing/force pushing and amending commits each have their purposes and are fine when done separately but that it can get confusing for reviewers when the two are combined. :slightly_smiling_face:

bitoku:
@kimorris27 Thanks! I was wondering why `compare` doesn't work properly.

PR #3718, opened 2024-07-23
Update MDSD, MDM, Fluentbit base images

s-fairchild:

[ARO-5049](https://issues.redhat.com/browse/ARO-5049)

Fixes

Bump MDSD, MDM, Fluentbit base images.

No


jaitaiwan (approved):
LGTM

pkg/util/version/const.go
```diff
@@ -40,19 +40,21 @@ var DefaultInstallStream = Stream{
 
 // FluentbitImage contains the location of the Fluentbit container image
 func FluentbitImage(acrDomain string) string {
-	return acrDomain + "/fluentbit:1.9.10-cm20240301@sha256:5a6a6987a1e8d4223b7e64524117cb294acbd7a0b10f813f298d4f632efe3c4f"
+	// Latest stable release is 2.0.20240628
```
cadenmarchese, line 43: Should we remove this comment? It seems like it would have to be updated a lot if we left it in, plus it looks like we're now going to be pulling 1.9.10 instead of 2.0.2.

mociarain (approved):

pkg/util/version/const.go
mociarain: I agree with this but merging it as is to get a release rolling

PR #3719, opened 2024-07-24
Drop some unneccessary dependencies by moving to `bingo` for tooling

hawkowl:

Rather than depending on these deps in the main go.mod, we can use `go run` to run them without adding their dependency trees and potential dependency conflicts in. This is most notable in tools like controller-gen.

### Test plan for issue:

E2E, CI

N/A

Shouldn't affect production code.


stevekuznetsov (commented):
Have we considered using [`bingo`](https://github.com/bwplotka/bingo)? It's an industry standard at this point, the DX is fairly excellent. Keeping static references to dependencies used in tooling and support keeps the main `go.mod` clean but has all the upside of full-fledged version management, etc.

hawkowl:
@stevekuznetsov Oooh, good idea. Let me poke it.

bitoku:
The idea is nice, but I think it can be split into smaller PRs like per package.

stevekuznetsov:
@hawkowl this is looking great! Are we waiting on anything else to merge? I don't have approval rights.

yjst2012 (dismissed):
This is great, thanks for putting this together, LGTM

tsatam (dismissed):
LGTM. Tested this change locally and all interactions worked as well. Thank you for this!

Dockerfile.ci-rp
```diff
@@ -45,18 +41,19 @@ COPY hack hack
 COPY cmd cmd
 COPY pkg pkg
 COPY test test
+COPY .bingo .bingo
+COPY Makefile Makefile
```
SudoBrendan, line 45: This is the one change I disagree with in this PR. With our move to containerization, one of our primary objectives was to remove complexity and "back and forth" logic with `make`. When we run `make ci-rp` in master today, commands go in exactly one direction:

1. `make ci-rp`
2. Dockerfile
    - Shell commands and environment configuration in-file, which can be read clearly top-to-bottom

To author that change, it took me something like *8hr* to parse and understand exactly what `Makefile` does. Having done that, I believe that `Makefile` is actually the most complicated piece of software we maintain today.

This PR is now re-introducing that complexity. Let's look at an example to show what I mean. To parse and/or debug why golang linting failed in their `make ci-rp` run, a developer will need to review things in this order:

1. `make ci-rp` executes a `podman build`
2. `Dockerfile.ci-rp`
    - executes shell commands, which includes additional `make` targets
        - Back in the `Makefile`, we look at `make lint-go`
            - Confusion alert: Makefile sets a bunch of golang envvars at the top that will override our docker env config for golang configuration, so this is already unclear and easy to miss
            - `lint-go` target points us to a shell script
                - we're redirected to `/hack/lint-go.sh`
                    - This file contains `golangci-lint run --verbose` wrapped in an `if` block.


OK - so I just debugged only one of these 3 changes, and it took me into *5 different files* and *2 different forms of shell environment variable configuration*. What is removed in this change is: `RUN golangci-lint run --verbose` in the Dockerfile.

I would really prefer from here on out that `Makefile` is never required for `Dockerfile.ci-rp` to build. Undoing this I think adds a lot of unnecessary obfuscation I would rather avoid moving forward with the development and release of our product.

SudoBrendan (dismissed):
There is so much that I love about this PR - thank you for working to get us out of `vendor` Amber :)

The one change I would suggest is eliminating `COPY Makefile Makefile` and all downstream dependencies of that in `Dockerfile.ci-rp`

Dockerfile.ci-rp
hawkowl: This is somewhat on purpose -- we used to have individual steps broken out in dockerfiles and the like, and when we needed to add a new goflags argument/etc we had to change it in a bunch of places, and the proliferation of places that we can set this means that things like the no-cgo thing was incorrectly removed in some places and not others in the past.

I also did already remove the lint-go indirection, because it was now pointless :)

I would like there to be _one_ place for all of the goflags/etc -- perhaps we need to make a golang.env or something with all of them? I would also like the "building our binary" part to be deduplicated somewhere, and I don't know anywhere better than a makefile for that. Perhaps we isolate the build portions into a separate makefile?

SudoBrendan: We agree - we need a single place to define all our envs/versions/dependencies/etc, and I think that given that we run containers in production, container builds are the perfect place to house all that information in a single place. Containerfiles give us a single API that works across all our various local dev environments including MacOS v Linux, eliminates diffs in tooling (e.g. golang/python/npm/linter versions), is easily read top to bottom without any form of `goto`, optimizes caching, makes artifacts sharable, and probably other benefits I'm not thinking of over `make` configuring shell and calling local bins. I think for cases that are _not_ containerized (I can't think of [m]any), it would be best to `podman cp` artifacts out of containers used to build.

I know this is a huge change compared to how we build today - but I think given the example above (and others not mentioned that are similar in Makefile), it makes sense to shift. I'm also open to feedback from the team at large because this change will affect how everyone develops and tests code locally. I can say anecdotally that I have driven transformations like this in 3 other orgs, and container builds eliminate entire classes of problems (e.g. the notorious slack/email threads of "ok PR 1234 has merged - everyone in the team please now go bump your local golang version" or the fact that [prepare your dev environment](https://github.com/Azure/ARO-RP/blob/master/docs/prepare-your-dev-environment.md) even exists at all outside "install podman. code.").

Dockerfile.ci-rp
```diff
@@ -32,10 +32,6 @@ WORKDIR /app
 ENV GOPATH=/root/go
 ENV GOFLAGS="-tags=containers_image_openpgp,exclude_graphdriver_btrfs,exclude_graphdriver_devicemapper"
 
-# Install golangci-lint and verify
-RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/local/bin v1.56.2 && \
-    golangci-lint --version || (echo "golangci-lint not found" && exit 1)
```
SudoBrendan, line 37: Related to the discussion on container install methods - this concerns me. I chose this route specifically because golangci-lint's own documentation says that `go install` is unreliable: https://golangci-lint.run/welcome/install/#install-from-source - their reasoning:

```
go install/go get installation isn't recommended because of the following points:

    Some users use -u flag for go get, which upgrades our dependencies. Resulting configuration wasn't tested and isn't guaranteed to work.
    [go.mod](https://github.com/golangci/golangci-lint/blob/master/go.mod) replacement directive doesn't apply. It means a user will be using patched version of golangci-lint if we use such replacements.
    We've encountered a lot of issues with Go modules hashes.
    It allows installation from master branch which can't be considered stable.
    It's slower than binary installation.
```

Dockerfile.ci-rp
SudoBrendan: ok - after a review of https://github.com/Azure/ARO-RP/pull/3558 which introduced containerization, I'm gonna say this change is actually a hard no-go for me. `make ci-rp` does not work as written.

The first error you'll get locally is:

```
[2/2] STEP 17/23: COPY Makefile Makefile
Error: building at STEP "COPY Makefile Makefile": no items matching glob "/home/bbergen/go/src/github.com/Azure/ARO-RP/Makefile" copied (1 filtered out using /home/bbergen/go/src/github.com/Azure/ARO-RP/Dockerfile.ci-rp.dockerignore): no such file or directory
make: *** [Makefile:101: ci-rp] Error 125
```

So we add `Makefile` to Dockerfile.ci-rp.dockerignore to fix, rebuild, and see:

```
[2/2] STEP 19/23: RUN make lint-go
/root/go/bin/golangci-lint-v1.59.1 run --verbose
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
/bin/bash: /root/go/bin/golangci-lint-v1.59.1: No such file or directory
make: *** [Makefile:310: lint-go] Error 127
Error: building at STEP "RUN make lint-go": while running runtime: exit status 2
make: *** [Makefile:101: ci-rp] Error 2
```

We 100% should not `COPY .git .git` into the Dockerfile, this will invalidate the build cache 100% of the time. Make has to go.

SudoBrendan: This is the Dockerfile.ci-rp diff I was able to get building locally following that convention:

1. Add a new section at the top to `COPY .bingo .bingo` and then run through the bingo install process using the `Variables.mk` file you added
2. Configure the PATH to include `$GOPATH/bin`
3. Use the bins (the only change required here was `gotestsum` being a direct bin exec rather than `go run`)


```diff
diff --git a/Dockerfile.ci-rp b/Dockerfile.ci-rp
index b4bbcc0d1..f5228a80f 100644
--- a/Dockerfile.ci-rp
+++ b/Dockerfile.ci-rp
@@ -32,9 +32,14 @@ WORKDIR /app
 ENV GOPATH=/root/go
 ENV GOFLAGS="-tags=containers_image_openpgp,exclude_graphdriver_btrfs,exclude_graphdriver_devicemapper"
 
-# Install golangci-lint and verify
-RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/local/bin v1.56.2 && \
-    golangci-lint --version || (echo "golangci-lint not found" && exit 1)
+# install golang build tooling via bingo
+COPY .bingo .bingo
+RUN source .bingo/variables.env \
+  # install bingo itself
+  && make -f .bingo/Variables.mk ${BINGO} \
+  # and all the tools it manages
+  && ${BINGO} get -l
+ENV PATH=$PATH:${GOPATH}/bin/
 
 # Copy dependencies and source files
 COPY go.mod go.sum ./
@@ -56,7 +61,7 @@ RUN go build -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=${A
 RUN go test ./test/e2e/... -tags e2e,codec.safe -c -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=${ARO_VERSION}" -o e2e.test
 
 # Additional tests
-RUN go run gotest.tools/gotestsum@v1.11.0 --format pkgname --junitfile report.xml -- -coverprofile=cover.out ./...
+RUN gotestsum --format pkgname --junitfile report.xml -- -coverprofile=cover.out ./...
 RUN hack/fips/validate-fips.sh ./aro
 
 ###############################################################################
```

Dockerfile.ci-rp
SudoBrendan: if we're ok with accepting this risk for all the other benefits we get with bingo, I can be onboard, just wanted to call this out.

hawkowl: https://github.com/bwplotka/bingo?tab=readme-ov-file#real-life-examples has some information about it -- basically, bingo side-steps the issues that it doesn't recommend it because of.

Plus, the module pinning that bingo uses is better than downloading binaries, in terms of security, imo.

SudoBrendan (dismissed):

jaitaiwan (approved):
LGTM, one comment from Brendan I'd like to see addressed before merge

.github/workflows/golint.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     - name: Run golangci-lint
       uses: golangci/golangci-lint-action@v6
       with:
-        version: v1.56.2
+        version: v1.59.1
```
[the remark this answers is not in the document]
hawkowl: belatedly: this is to match with the bingo version, as well as the newer versions people would get from downloading it when setting up now

PR #3720, opened 2024-07-24
fix repeated reconciliation in node controller

bitoku:

Fixes N/A

In a part of other work, I found when I scale machinesets up and down, node controller starts to complain:
```
time="2024-07-24T17:16:45Z" level=error msg="Node \"sre-shared-cluster-ltnmx-worker-eastus1-cl8zj\" not found" func="node.(*Reconciler).Reconcile()" file="pkg/operator/controllers/node/node_controller.go:67" controller=Node
```
This happens because the node is deleted before the request is handled.
Current implementation returns nil if the node is not found and retries, which causes the repeated reconciliation.

### Test plan for issue:

Test in local dev cluster.

N/A, bug fix

Try scaling machinesets and check whether there is node controller error logs in aro-operator logs.

### DIFF
--- a/pkg/operator/controllers/node/node_controller.go
+++ b/pkg/operator/controllers/node/node_controller.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


stevekuznetsov (approved):

mociarain (approved):

kimorris27 (approved):
I always wondered why this controller seems to log `node not found` messages so often. Good catch!

tsatam (approved):

AldoFusterTurpin (approved):
LGTM. Thanks!

PR #3683, opened 2024-07-11
shared utility function to check if a cluster is workload identity

slawande2:
### Which issue this PR addresses:
[ARO-7856](https://issues.redhat.com/browse/ARO-7856)

### What this PR does / why we need it:
Throughout the RP codebase and installer wrapper, we are using if platformworkloadidentityprofile != nil or ServicePrincipalProfile to indicate whether or not a cluster is workload identity or not. This PR includes a shared utility function to check if a function is a Workload Identity or Service Principal cluster, which can be imported in all parts of the codebase.

### Test plan for issue:
Passed all the unit test cases.

PR #3721, opened 2024-07-24
move iswimi function to openshiftcluster.go

slawande2:

[ARO-7856](https://issues.redhat.com/browse/ARO-7856)

This is a follow-up PR to https://github.com/Azure/ARO-RP/pull/3683 
In this PR, I've moved the shared utility function to pkg/api/openshiftcluster.go in order to obtain a richer domain where functions are closer where the data is used and to avoid having an anemic Domain Model. This moves the business logic closer to where the data lives in. 

### Test plan for issue:

Passed all unit tests.


pkg/api/openshiftcluster_test.go
```diff
@@ -49,3 +52,41 @@ func TestIsTerminal(t *testing.T) {
 		})
 	}
 }
+
+func TestIsWimi(t *testing.T) {
+	tests := []*struct {
```
rajdeepc2792, line 57: Can you please also add 2 more test cases here, although this is not going to happen, but to cover all the cases for non-miwi?:-
1. Both PlatformWorkloadIdentityProfile and ServicePrincipalProfile are nil. want - false
2. Both PlatformWorkloadIdentityProfile and ServicePrincipalProfile are set. want - false

pkg/api/openshiftcluster.go
```diff
@@ -27,6 +27,11 @@ type OpenShiftCluster struct {
 	Lock sync.Mutex `json:"-"`
 }
 
+// IsWimi checks whether a cluster is a Workload Identity cluster or Service Principal cluster
+func (oc *OpenShiftCluster) IsWimi() bool {
```
stevekuznetsov, line 31: What is "Wimi" - could we spell out "WorkloadIdenitity..." for readability please?

rajdeepc2792: @slawande2 , IMO you can also just use `IsWorkloadIdentity` instead. Reference of usage:- https://docs.openshift.com/container-platform/4.15/release_notes/ocp-4-15-release-notes.html#ocp-4-15-console-supports-azure-sts-detection

slawande2: done, thank you for the suggestion.

pkg/api/openshiftcluster_test.go
slawande2: Done, thank you for the suggestion.

pkg/api/openshiftcluster.go
slawande2: done, thank you for the suggestion.

AldoFusterTurpin (dismissed):
LGTM! Great. Thank you.

pkg/api/openshiftcluster.go
```diff
@@ -27,6 +27,11 @@ type OpenShiftCluster struct {
 	Lock sync.Mutex `json:"-"`
 }
 
+// IsWimi checks whether a cluster is a Workload Identity cluster or Service Principal cluster
```
bitoku, line 30:
```suggestion
// IsWorkloadIdentity checks whether a cluster is a Workload Identity cluster or Service Principal cluster
```

pkg/api/openshiftcluster.go
```diff
@@ -27,6 +27,11 @@ type OpenShiftCluster struct {
...
+func (oc *OpenShiftCluster) IsWorkloadIdentity() bool {
```
bitoku, line 31: nit: This is probably a matter of preference, but I like `IsWimi` better than `IsWorkloadIdentity` because "the cluster is wimi" sounds more natural than "the cluster is workload identity".
If it's only me, please ignore this comment.

bitoku: sorry I missed the previous discussion.

pkg/api/openshiftcluster.go
stevekuznetsov: Ideally, turn on `golangci-lint` and get this type of review automated!

pkg/api/openshiftcluster.go
stevekuznetsov: `UsesWorkloadIdentity` might be better if we're worried about reading it like that - my comment was moreso that asking everyone to go learn more acronyms is not ideal!

AldoFusterTurpin: Agree. UsesWorkloadIdentity is probably the best option. 🙂

slawande2: UsesWorkloadIdentity sounds good to me! Thank you for the suggestion.

kimorris27 (approved):

mociarain (approved):
Great change. Ty for doing this

PR #3723, opened 2024-07-25
Drop SyncSets check

mociarain:

Fixes: https://issues.redhat.com/browse/ARO-7787

This PR drops the sync sets check from `IsClusterDeploymentReady`.

Monitoring of SyncSets will come as part of it's own monitoring in the future and since we have no SyncSets at present there is no temporary gap.

### Test plan for issue:

Tested locally using the dev Hive


tsatam (approved):
Change and motivation LGTM. 

I'm curious if we should still keep an explicit SyncSets ready check at the end of our install process, just to make sure we don't have any WIP before we hand the cluster over to the user, but that can be introduced whenever we actually start using SyncSets.

kimorris27 (approved):

SrinivasAtmakuri (approved):
LGTM!

PR #3725, opened 2024-07-26
[ARO-7345] smoke test automation for aro-operator log 

bitoku:

Fixes https://issues.redhat.com/browse/ARO-7345

This PR adds smoke test that confirms there is no error log in aro-operator logs.
This is labelled as "smoke", so it doesn't run automatically in CI.

### Test plan for issue:

N/A

This is a part of tests.

### DIFF
--- a/test/e2e/operator.go
+++ b/test/e2e/operator.go
@@ -41,8 +41,12 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/ready"
 )
 
+const (
+	aroOperatorNamespace = "openshift-azure-operator"
+)
+
 func updatedObjects(ctx context.Context, nsFilter string) ([]string, error) {
-	listFunc := clients.Kubernetes.CoreV1().Pods("openshift-azure-operator").List
+	listFunc := clients.Kubernetes.CoreV1().Pods(aroOperatorNamespace).List
 	pods := ListK8sObjectWithRetry(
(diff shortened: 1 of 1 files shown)


kimorris27 (commented):
The change and the motivation make sense, but I'm not sure if I really think the operator is this stable... let's see if E2E passes the new test 😅

mociarain:
> let's see if E2E passes the new test 😅

So E2E won't run the smoke tests as is: https://github.com/azure/ARO-RP/blob/09a56b1c466c54d10fb9132e18ff3dd78ce0533b/Makefile#L6


Having said that I agree with your concerns @kimorris27. If there's some error in the aro-operator logs do we care? I think the answer is "it depends" but given the scope of where we use the smoke tests I'm happy to ship this as is. Thoughts?

mociarain (approved):

kimorris27:
>So E2E won't run the smoke tests as is: https://github.com/azure/ARO-RP/blob/09a56b1c466c54d10fb9132e18ff3dd78ce0533b/Makefile#L6
>
>Having said that I agree with your concerns @kimorris27. If there's some error in the aro-operator logs do we care? I think the answer is "it depends" but given the scope of where we use the smoke tests I'm happy to ship this as is. Thoughts?

Oh, that's right; I forgot. That makes sense to me. We can tweak the test later on if this causes too many "flakes" in the smoke tests.

kimorris27 (approved):

PR #3729, opened 2024-07-29
Hive SyncSet GenevaActions

SrinivasAtmakuri:
Implements:

1. Geneva Action to list selectorsyncsets for a given hive instance
2. Geneva Action to retrieve an individual selectorsyncset on a given hive instance
3. Geneva Action to list syncsets for a cluster defined by resource id in a hive instance
4. Geneva Action to retrieve an individual syncsets on a given hive instance

How to call:

List selector/syncsets - /admin/hivesyncset?namespace={namespace}?isSyncset
Get a selector/syncset   - /admin/hivesyncset/syncsetname/{syncsetname}?namespace={namespace}?isSyncSet

Notes:
1. namespace should be "" for selectorsyncsets
2. namespace cannot be "" for syncsets
3. isSyncSet can be true/false, setting it changes from using default selectorsyncset to syncset

Fixes

### Test plan for issue:

Testing is underway., below is the plan.,

1. Connect to shared AKS cluster's hive
2. Create a cluster and Syncset in CD namespace and SelectorSyncSet with hive
3. Run admin GA using curl via localrp


pkg/frontend/frontend.go
```diff
@@ -124,6 +125,7 @@ func NewFrontend(ctx context.Context,
 	clusterm metrics.Emitter,
 	aead encryption.AEAD,
 	hiveClusterManager hive.ClusterManager,
+	hiveSyncSetManager hive.SyncSetManager,
```
mociarain, line 128: Would it make more sense to add this using the builder pattern or something similar:https://medium.com/@reetas/clean-ways-of-adding-new-optional-fields-to-a-golang-struct-99ae2fe9719d?

What do you think?

pkg/frontend/admin_hive_syncset_get.go
```diff
@@ -0,0 +1,65 @@
...
+	adminReply(log, w, nil, b, err)
+}
+
+func (f *frontend) _getAdminHiveSyncSet(ctx context.Context, namespace string, syncsetname string, isSyncSet bool) ([]byte, error) {
+	ssType := reflect.TypeOf(hivev1.SelectorSyncSet{})
+	if isSyncSet {
+		ssType = reflect.TypeOf(hivev1.SyncSet{})
+	}
+	if isSyncSet && namespace == "" {
+		return nil, api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden, "", "namespace cannot be null for getting a syncset")
```
pepedocs, line 43: Is there any reason why Forbidden (403) status is use here rather than Bad Request (400)?

pkg/frontend/admin_hive_syncset_get_test.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	for _, tt := range []*test{
+		{
+			name:           "selectorSyncSets are not namespaced",
+			namespace:      "aro-" + fakeUUID,
+			syncsetname:    "syncsetTest",
+			isSyncSet:      false,
+			mocks:          func(tt *test, s *mock_hive.MockSyncSetManager) {},
+			wantStatusCode: 403,
```
pepedocs, line 53: Same here and lines below, we should use 400 for bad request.

pkg/frontend/admin_hive_syncset_list.go
```diff
@@ -0,0 +1,60 @@
...
+		api.WriteCloudError(w, cloudErr)
+		return
+	}
+
+	adminReply(log, w, nil, b, err)
+}
+
+func (f *frontend) _listAdminHiveSyncSet(ctx context.Context, namespace string, label string, isSyncSet bool) ([]byte, error) {
+	// defaults to listing selectorSyncSets for an AKS instance
+	ssType := reflect.TypeOf(hivev1.SelectorSyncSet{})
```
pepedocs, line 37: Looks like this is repeated in a couple of places, I think we should take this out from these specific functions `_list/get...` and call it outside so we don't duplicate this code. And we should call `_list*SyncSet, _list*SelectorSyncSet` explicitly to set this up in the long-run. For example, modifying code for SyncSet only will not cause more if statements. This also promotes short and simple functions.

pkg/frontend/admin_hive_syncset_list_test.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	for _, tt := range []*test{
+		{
+			name:           "selectorSyncSets are not namespaced",
+			namespace:      "aro-" + fakeUUID,
+			syncsetname:    "syncsetTest",
+			isSyncSet:      false,
+			mocks:          func(tt *test, s *mock_hive.MockSyncSetManager) {},
+			wantStatusCode: 403,
```
pepedocs, line 75: Same comment as previous, I think this should be 400.

pepedocs (changes requested):
I have added my comments mostly for enhancement. However I'd like to request changes for the usage of HTTP 403 Forbidden status as I've specified. HTTP 400 Bad Request should be the right response for client requests that lack parameters or wrong parameters.

bitoku:
It would be better to refactor like https://github.com/Azure/ARO-RP/pull/3665/files first to focus on the actual change.

bitoku (commented):
We want e2e tests for this feature.

SrinivasAtmakuri:
> I have added my comments mostly for enhancement. However I'd like to request changes for the usage of HTTP 403 Forbidden status as I've specified. HTTP 400 Bad Request should be the right response for client requests that lack parameters or wrong parameters.

Resolved, made the changes.

pkg/frontend/frontend.go
SrinivasAtmakuri: the rational for creating a new interface is bcoz the syncsets are not cluster scoped and hive instance scoped, so its cleaner to have a new interface which implements syncset functions.

pkg/hive/generate.go
SudoBrendan, line 1: flagging this as needing a rebase on master to pick up the .bingo changes.

SudoBrendan (changes requested):

pkg/frontend/admin_hive_syncset_get.go
```diff
@@ -0,0 +1,68 @@
...
+
+	if cloudErr, ok := err.(*api.CloudError); ok {
+		api.WriteCloudError(w, cloudErr)
+		return
+	}
+
+	adminReply(log, w, nil, b, err)
+}
+
+func (f *frontend) _getAdminHiveSyncSet(ctx context.Context, namespace string, syncsetname string, isSyncSet bool) ([]byte, error) {
```
pepedocs, line 36: Recommend: We should not somewhere that we are using the term SyncSet for both SelectorSyncSet and plain Syncset to avoid confusion.

pepedocs (commented):
UPDATE: I'm currently testing this code. So far I have found bugs in it that needs to be fixed. I will sumbit a new PR later - https://issues.redhat.com/browse/ARO-14581

sankur-codes:
New PR has been raised with bug fixes and necessary changes.

PR #3730, opened 2024-07-29
[ARO-7860] Fix local E2E tests failing due to pki timeout

mrWinston:

Fixes ARO-7860

This PR disables fetching the cert pool from azure for applens when running the RP locally. This fixes the following 4 E2E test failures:

```
Summarizing 4 Failures:
  [FAIL] [Admin API] Delete managed resource action [It] should be possible to delete managed cluster resources
  test/e2e/adminapi_delete_managedresource.go:123
  [FAIL] [Admin API] Delete managed resource action [It] should NOT be possible to delete the private link service in the cluster's managed resource group
  test/e2e/adminapi_delete_managedresource.go:116
  [FAIL] [Admin API] VM redeploy action [It] must trigger a selected VM to redeploy
  test/e2e/adminapi_redeployvm.go:53
  [FAIL] [Admin API] List Azure resources action [It] must list Azure resources for a cluster
  test/e2e/adminapi_resources.go:80
```

- No

- This change only affects the local dev environment, hence it doesn't affect production.

### DIFF
--- a/pkg/env/dev.go
+++ b/pkg/env/dev.go
@@ -41,6 +41,9 @@ func newDev(ctx context.Context, log *logrus.Entry, component ServiceComponent) 
 		d.features[feature] = true
 	}
 
+  // Disable applens in local dev environment: ARO-7860
+  d.Environment().PkiIssuerUrlTemplate = ""
+
 	d.prod.clusterGenevaLoggingAccount = version.DevClusterGenevaLoggingAccount
 	d.prod.clusterGenevaLoggingConfigVersion = version.DevClusterGenevaLoggingConfigVersion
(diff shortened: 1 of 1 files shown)


pkg/env/dev.go
```diff
@@ -41,6 +41,9 @@ func newDev(ctx context.Context, log *logrus.Entry, component ServiceComponent)
 		d.features[feature] = true
 	}
 
+	// Disable applens in local dev environment: ARO-7860
+	d.Environment().PkiIssuerUrlTemplate = ""
```
bitoku, line 45: I think this should be here where the environment is created.
https://github.com/Azure/ARO-RP/blob/4c59ba73f586552561859631fc3fc648a4b8af23/pkg/util/instancemetadata/dev.go#L13

PR #3732, opened 2024-07-29
Configure gateway container to log to /var/log/aro-gateway/gateway.log

ventifus:

Fixes [ARO-8920](https://issues.redhat.com/browse/ARO-8920)

Podman writes the container's stdout to the log file `ctr.log`. This is done by podman itself, not by a process within the container and so any changes to the path must be done via podman configuration. 

This new configuration relocates the log file created by podman into /var/log/aro-gateway and also sets a 20GB upper limit on its size.

Because podman is now managing the size of the log file, a logrotate configuration is no longer needed.


pepedocs (commented):
I'm a bit confused, Jira/Slack talked about podman ctr.log file, however from this PR alone, I don't see any usage of podman?

mociarain:
FYI you'll need to move this to a branch off the main repo. Forks are no longer workable with the current MSFT restrictions

PR #3736, opened 2024-07-30
[MIMO] Move cluster certificate functionality to ClientHelper

hawkowl:
Splitout of MIMO M1 work

Retrofits clienthelper support into some functions that MIMO tasks require

### Test plan for issue:

CI, E2E

N/A

E2E should cover it, since it applies during install as well

### DIFF
--- a/pkg/cluster/apply.go
+++ b/pkg/cluster/apply.go
@@ -1,0 +1,54 @@
+package cluster
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/pem"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/Azure/ARO-RP/pkg/env"
+	"github.com/Azure/ARO-RP/pkg/util/clienthelper"
+	utilpem "github.com/Azure/ARO-RP/pkg/util/pem"
+)
+
+func EnsureTLSSecretFromKeyvault(ctx context.Context, env env.Interface, ch clienthelper.Interface, target types.NamespacedName, certificateName string) error {
+	bundle, err := env.ClusterKeyvault().GetSecret(ctx, certificateName)
+	if err != nil {
+		return err
+	}
+
+	key, certs, err := utilpem.Parse([]byte(*bundle.Value))
+	if err != nil {
+		return err
+	}
+
+	b, err := x509.MarshalPKCS8PrivateKey(key)
+	if err != nil {
+		return err
+	}
+
+	var cb []byte
+	for _, cert := range certs {
+		cb = append(cb, pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})...)
+	}
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      target.Name,
+			Namespace: target.Namespace,
+		},
+		Data: map[string][]byte{
+			corev1.TLSCertKey:       cb,
+			corev1.TLSPrivateKeyKey: pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: b}),
+		},
+		Type: corev1.SecretTypeTLS,
+	}
+
+	return ch.Ensure(ctx, secret)
+}
--- a/pkg/cluster/apply.go
+++ b/pkg/cluster/apply.go
@@ -12,13 +12,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
-	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/util/clienthelper"
+	"github.com/Azure/ARO-RP/pkg/util/keyvault"
 	utilpem "github.com/Azure/ARO-RP/pkg/util/pem"
 )
 
-func EnsureTLSSecretFromKeyvault(ctx context.Context, env env.Interface, ch clienthelper.Interface, target types.NamespacedName, certificateName string) error {
-	bundle, err := env.ClusterKeyvault().GetSecret(ctx, certificateName)
+func EnsureTLSSecretFromKeyvault(ctx context.Context, kv keyvault.Manager, ch clienthelper.Writer, target types.NamespacedName, certificateName string) error {
+	bundle, err := kv.GetSecret(ctx, certificateName)
 	if err != nil {
 		return err
 	}
@@ -38,6 +38,8 @@ func EnsureTLSSecretFromKeyvault(ctx context.Context, env env.Interface, ch clie
 		cb = append(cb, pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})...)
 	}
 
+	privateKey := pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: b})
+
 	secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      target.Name,
@@ -45,7 +47,7 @@ func EnsureTLSSecretFromKeyvault(ctx context.Context, env env.Interface, ch clie
 		},
 		Data: map[string][]byte{
 			corev1.TLSCertKey:       cb,
-			corev1.TLSPrivateKeyKey: pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: b}),
+			corev1.TLSPrivateKeyKey: privateKey,
 		},
 		Type: corev1.SecretTypeTLS,
 	}
--- a/pkg/cluster/install.go
+++ b/pkg/cluster/install.go
@@ -546,7 +546,7 @@ func (m *manager) initializeKubernetesClients(ctx context.Context) error {
 // initializeKubernetesClients initializes clients which are used
 // once the cluster is up later on in the install process.
 func (m *manager) initializeOperatorDeployer(ctx context.Context) (err error) {
-	m.aroOperatorDeployer, err = deploy.New(m.log, m.env, m.doc.OpenShiftCluster, m.arocli, m.ch.Client(), m.extensionscli, m.kubernetescli, m.operatorcli)
+	m.aroOperatorDeployer, err = deploy.New(m.log, m.env, m.doc.OpenShiftCluster, m.arocli, m.ch, m.extensionscli, m.kubernetescli, m.operatorcli)
 	return
 }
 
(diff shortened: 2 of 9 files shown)


mociarain (commented):
A few things:
- `ClientHelper.New` is never used. I think we can kill it without trouble.
- Can we collapse the `New` stuff in this so we have private constructors and a public `GetClient` that builds or creates a client as needed. Basically, should this be a singleton?
- General ask to add some more descriptive naming. Can we say what kind of client it is? It think `k8sClient` but I'm not sure. Maybe a better name for `clientHelper` is `AROK8sClient` i.e. our wrapped version of the k8s client.

pkg/cluster/install.go
```diff
@@ -534,16 +535,18 @@ func (m *manager) initializeKubernetesClients(ctx context.Context) error {
...
+
+	m.ch = clienthelper.NewWithClient(m.log, client)
 	return err
 }
 
 // initializeKubernetesClients initializes clients which are used
 // once the cluster is up later on in the install process.
 func (m *manager) initializeOperatorDeployer(ctx context.Context) (err error) {
-	m.aroOperatorDeployer, err = deploy.New(m.log, m.env, m.doc.OpenShiftCluster, m.arocli, m.client, m.extensionscli, m.kubernetescli, m.operatorcli)
+	m.aroOperatorDeployer, err = deploy.New(m.log, m.env, m.doc.OpenShiftCluster, m.arocli, m.ch.Client(), m.extensionscli, m.kubernetescli, m.operatorcli)
```
stevekuznetsov, line 549: Wrapping the client into the helper only to ask the helper to give the client right back is awkward. Either the client helper closes over the client, or hold on to both values, IMO. The in-between seems like an incorrect abstraction that closes over nothing.

pkg/cluster/apply.go
```diff
@@ -0,0 +1,54 @@
...
+			Namespace: target.Namespace,
+		},
+		Data: map[string][]byte{
+			corev1.TLSCertKey:       cb,
+			corev1.TLSPrivateKeyKey: pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: b}),
+		},
+		Type: corev1.SecretTypeTLS,
+	}
+
+	return ch.Ensure(ctx, secret)
```
AldoFusterTurpin, line 53: As previously mentioned, as we just use the Ensure method of _ch_, maybe we could define here an interface with just this method to stick to the Interface Segregation Principle.

pkg/cluster/install.go
AldoFusterTurpin: I agree with that. My question: is there a good reason to do so here that we are not aware of ?  

I would simply make the client public instead of providing a getter as this indirect access doesn't bring much value (specially when we have multiple stuff that is called "Client", methods, struct fields, etc) IMHO.

AldoFusterTurpin (commented):
Thank you @hawkowl for this, I like that! 🙂 
I just left some comments in case any make sense to you, but not a blocker.

pkg/cluster/apply.go
```diff
@@ -0,0 +1,54 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/Azure/ARO-RP/pkg/env"
+	"github.com/Azure/ARO-RP/pkg/util/clienthelper"
+	utilpem "github.com/Azure/ARO-RP/pkg/util/pem"
+)
+
+func EnsureTLSSecretFromKeyvault(ctx context.Context, env env.Interface, ch clienthelper.Interface, target types.NamespacedName, certificateName string) error {
```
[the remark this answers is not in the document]
hawkowl: fair, done

pkg/cluster/install.go
hawkowl: I thought about it a bit, and I've made it close over the whole client instead, since that was pretty easy. This means that the places that we currently use a controller-runtime client, we can upgrade it to the clienthelper fairly easily.

pkg/cluster/apply.go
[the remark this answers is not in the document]
hawkowl: I've altered the code to have Reader and Writer interfaces, similar to the controller-runtime client. This should let us define things a bit nicer if we want to restrict down the interface. I'm not too eager to narrow down the interfaces more than this, though.

I think that making single-method interfaces (especially defining them in consumers), is overly restricting and can cause a lot of work if any of those interfaces need to change in the future. Every time I've gone to use it, I've wished I hadn't later on. More granular interfaces from the provider can mitigate some of this duplication, but I'm also not sure how often we can get away with having super-granular interfaces. A Reader/Writer should hopefully do for now.

I also don't think it makes code easier to test when we have verified fakes (e.g. the controller-runtime client). It _does_ make it easier to deal with when we're writing mocks for interfaces, since the mocks (and what the code might call) are therefore smaller, but I firmly believe that mocks are an anti-pattern and that even big interfaces like `env.Interface` would be a lot more tolerable if designed with a proper fake in parallel.

pkg/util/clienthelper/clienthelper.go
```diff
@@ -34,35 +33,34 @@ import (
 	_ "github.com/Azure/ARO-RP/pkg/util/scheme"
 )
 
-type Interface interface {
-	EnsureDeleted(ctx context.Context, gvk schema.GroupVersionKind, key types.NamespacedName) error
+type Writer interface {
+	client.Writer
+	// Ensure applies self-contained objects to a Kubernetes API, merging
```
stevekuznetsov, line 38: I know you're not changing this in the PR, so this is not a blocker here, but we should use server-side apply instead of whatever client-side logic exists, as it's guaranteed to be a better implementation :)

hawkowl: I would _love_ to use server-side apply, but the fake implementation of controller-runtime doesn't yet support it :(

Refs: 
- https://github.com/kubernetes-sigs/controller-runtime/issues/2341
- https://github.com/kubernetes/kubernetes/pull/125560 -- the groundwork does look like it's beginning to be there :)
- https://github.com/kubernetes-sigs/controller-runtime/pull/2702 hi @troy0820 fancy seeing you there :D

troy0820: _No comment, just waving at_ @hawkowl 👋🏽

stevekuznetsov: Yes, the controller-runtime fakes are horrendous and IMO moving away from them as soon as possible is also a great boon to the project :)

SSA is very complex and faking it is likely not going to happen - not even just in some soon time-frame, but perhaps ever. Client fakes have fallen out of favor and everyone has a horror story of working around the mess that is envtest without gaining any confidence in their application functioning in the real world against a real API server. Even past that, the concerns Alvaro has in Troy's PR are a mountain of work that the upstream SIGs are not interested in taking up or sponsoring, from my understanding.

This is neither here nor there, but you might want to consider not putting SSA support in the controller-runtime fakes as a precondition to using this (amazingly useful and beneficial) technology in your production deployments.
 
See an example of how to test controllers or similar using static data in, static data out here: https://github.com/openshift/hypershift/blob/main/control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller_test.go

troy0820: The issue beyond the requisite work are the apply types that are not scaffold with crd/operator-sdk/kubebuilder.  That work is a side effect of the client (controller-runtime) not implementing that whereas client-go has that necessary logic.  

The interface client being used here is the mechanism where that can't be utilized. I have no context around what this is trying to accomplish but if it is indeed using corev1 types where a `kubernetes.Interface` can be used, those applyConfigurations are available to fulfil the server side apply.

The faking of the clients are tracked with a "tracker" and trying to track a server side apply in any client would be trying to hold state for things it wouldn't worry about because of the client's responsibility.  Having a test suite where that can happen seems outside of the scope of what the client/fakes could do

stevekuznetsov: I have worked with the `controller-runtime` ecosystem and upstream Kubernetes client-sets for many years, so I'm pretty familiar with the structure. I don't think I communicated the "static data in, static data out" link well - my suggestion is not that some client helper hold on to apply configurations, nor is it to somehow teach the `controller-runtime` client how to apply. 

In case we get too far in the weeds - I just want to re-iterate that the comments here are not review comments for this PR, just a general comment. I'm not suggesting any changes are made right now.

---

What I *am* suggesting, though, is that operator reconciliation loops are written as pure functions - either take some set of cluster state as input, or allow dependency injection to take producers of cluster state as input, and produce state as output. Apply configurations have a wonderful property in that they encode your intent in static data - so, testing your lower-case `r` `reconcile()` is straightforward, and no mocks of any persuasion are necessary.

The top-level `Reconcile()` might look like:

```go
func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
    // assemble the data necessary by using listers, etc
    input := r.getInput()
    
    // transform input state into intended state
    intent, err := r.reconcile(input)
    
    // realize intended state onto the cluster
    return r.execute(intent)
}
```

Where `r.execute(intent)` takes apply configurations and uses a `kubernetes.Clientset` to call `Apply`.

--- 

In my experience, relying heavily on mocks leads many teams to stumble into deeply frustrating issues, often many years down the line. Examples I've seen:

- implementing highly-performant caching using metadata-only watches requires 100% of tests to be rewritten, as the fakes don't know how to propagate data between the metadata fake and the "real" fake
- simplifying controller code to reduce maintenance burden and eliminate incorrect client-side patching by using server-side apply means all unit tests fail
- fragile assumptions about the number of times your reconciliation is called and when, in what order, etc, lead integration tests to be a poor substitute for end-to-end testing against actual API servers, leading to test duplication
- fakes make many highly surprising decisions, taking valuable mental overhead and productivty away from the team, like:
  - faking out field selectors (usually handled in the server) with indices, which requires test code to be written to create indices
  - objects with deletion timestamps but no finalizers cause some internal panic even though it's totally valid for your controller to see one in production
  - fake clients need subresources registered manually
  
---

Teams that spend time writing unit tests for unit-level concerns and do not reach to mocks, in my experience, move faster and get a higher RoI on their tests. End-to-end tests and chaos tests validate production modalities that integration tests attempt to, but cannot.

AldoFusterTurpin: @stevekuznetsov I completely agree with that. 🥇 

Even though I am by far the one less experienced in this conversation about this domain, I have seen similar problems in other domains that made me resonate a lot with your comment while reading it.

Sometimes we focus too much on trying to reproduce the state of the world where our app will run, and we forget that we should primarily test what our app does, and what our business logic is. It is almost always much simpler and has a greater return on investment to just have our business logic in pure functions that just expect input data and produce output data when talking about unit tests. We then forget about how this data will be gathered and what will be done with the result, and we can simply ensure that domainLogicFn(x) == expectedOutput.

This is a common concern that applies to a lot of different domains, but it is pretty common in k8s because most of the time the business logic gets mixed/loosed in the effort to understand/simulate the "state" where all that business logic will happen (due to the stateful nature of k8s controllers). So I just want to reaffirm that avoiding complex mocks when possible with pure functions is a great idea and not just in this context 🙂

jaitaiwan (approved):
LGTM!

PR #3739, opened 2024-07-30
Fix NotFound error in DNS isn't handled properly

bitoku:

Fixes N/A

In a part of other task, I found it failed to delete a cluster when dns records are already deleted.
This was caused by the update of the sdk to track2.
Since track2, autorest has been deprecated, we needed to change it to ResponseError.

cf. https://github.com/Azure/go-autorest
https://github.com/Azure/azure-sdk-for-go/blob/e302bb60ef995455b0eaabf590a75c8396e32834/sdk/azcore/internal/exported/response_error.go#L112

I found https://pkg.go.dev/github.com/Azure/azure-sdk-for-go-extensions/pkg/errors#IsNotFoundErr this repository but we can't use it, because the error we get is `NotFound`, but it checks if the error is `ResourceNotFound`.
Probably this discrepancy might come from the Azure API server, but anyway we can't use it.

### Test plan for issue:

Delete a dev cluster with dns records already deleted.
Unit test (it couldn't find the error when we updated the sdk though)

N/A

Delete a cluster with dns records already deleted.

### DIFF
--- a/pkg/util/dns/dns.go
+++ b/pkg/util/dns/dns.go
@@ -12,12 +12,12 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	sdkdns "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/dns/armdns"
-	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/env"
 	"github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/armdns"
+	azerrors "github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/errors"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/util/dns/dns.go
```diff
@@ -138,8 +139,9 @@ func (m *manager) Delete(ctx context.Context, oc *api.OpenShiftCluster) error {
...
 
 	rs, err := m.recordsets.Get(ctx, m.env.ResourceGroup(), m.env.Domain(), "api."+prefix, sdkdns.RecordTypeA, nil)
-	if detailedErr, ok := err.(autorest.DetailedError); ok &&
-		detailedErr.StatusCode == http.StatusNotFound {
+	var azErr *azcore.ResponseError
+	if errors.As(err, &azErr) &&
+		azErr.StatusCode == http.StatusNotFound {
 		return nil
 	}
 	if err != nil {
```
mociarain, line 147:
```suggestion
	if errors.As(err, &azErr) && azErr.StatusCode == http.StatusNotFound {
		return nil
	} else if err != nil {
```

mociarain (commented):
Small change in syntax but LGTM

bitoku:
I force-pushed this time again, but you can compare it with the previous one from `Compare` button right above.

bitoku:
I found another problem in DNS manager. I'll make it hold to fix all of them.

mociarain (approved):

AldoFusterTurpin (approved):
LGTM. Thanks.

PR #3740, opened 2024-07-30
change load balancer image in smoke test

bitoku:

Fixes https://issues.redhat.com/browse/ARO-7350

This is a followup of #3717 .
Apparently the image `nginx` is not compliant to MSFT security advisory.
https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=99367257&view=logs&j=10f66467-59e5-53bc-0c8a-42cc89770de4&t=22254a09-0937-5b6f-413f-542ad48300e2&l=28
I followed the warning and modified tests accordingly.

### Test plan for issue:

Run tests locally.

N/A

Check CI status.

### DIFF
--- a/test/e2e/static_resources/loadbalancer.yaml
+++ b/test/e2e/static_resources/loadbalancer.yaml
@@ -15,7 +15,8 @@ spec:
     spec:
       containers:
         - name: nginx
-          image: nginx
+          image: mcr.microsoft.com/cbl-mariner/base/nginx:1
+          command: ["/bin/sh", "-c", "mkdir -p /usr/share/nginx/html && echo 'hello world' > /usr/share/nginx/html/index.html && nginx -g 'daemon off;'"]
           ports:
             - containerPort: 80
               name: http-web-svc


test/e2e/ocp_loadbalancer.go
```diff
@@ -61,14 +61,15 @@ var _ = Describe("Load Balancer", Label(smoke), Ordered, func() {
...
 		Eventually(func(g Gomega, ctx context.Context) {
 			client := http.Client{Timeout: 10 * time.Second}
 			req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("http://%s", ip), nil)
 			g.Expect(err).NotTo(HaveOccurred())
 			resp, err := client.Do(req)
 			g.Expect(err).NotTo(HaveOccurred())
-			g.Expect(resp.StatusCode).To(Equal(http.StatusOK))
+			// Since we use mcr.microsoft.com/cbl-mariner/base/nginx image, 404 NotFound is expected.
+			g.Expect(resp.StatusCode).To(Equal(http.StatusNotFound))
 		}, DefaultEventuallyTimeout, 10*time.Second, ctx).Should(Succeed())
```
AldoFusterTurpin, line 73: Sorry if this is a stupid question (or if I am missing some context): does it make sense to test that we get an HTTP 404 when doing that HTTP call?  We could get a 404 for different reasons and maybe we are getting a 404 response but not due to the reason we think it is, even the test/assertion is "green".

AldoFusterTurpin (commented):
Just left a question. Thank you!

test/e2e/ocp_loadbalancer.go
AldoFusterTurpin: True! Also, what would happen is that _client.Do(req)_ would return an error and we would fail before checking the statuscode (and resp would be nil). 👍

tiguelu (approved):

PR #3743, opened 2024-07-31
ARO-4373 fix oidcstorageaccount variable validation error in aro-monitor

rajdeepc2792:

Fixes deployment failures related to aro-monitor service
IcM:- https://portal.microsofticm.com/imp/v5/incidents/details/527703987/summary

Moves the validation check for variable OIDC_STORAGE_ACCOUNT_NAME from the prod env object setup(that aro-monitor also uses) to rp.go codebase.
Doesn't change the functionality for the RP service, just removal of validation from a general codebase.

### Test plan for issue:
Must pass local cluster creation, e2e and canary.

aro-rp was running previously so there's no impact on the aro-rp service.
aro-monitor doesn't need OIDC_STORAGE_ACCOUNT_NAME variable, which was previously checked but not after this PR.

### DIFF
--- a/cmd/aro/rp.go
+++ b/cmd/aro/rp.go
@@ -52,6 +52,7 @@ func rp(ctx context.Context, log, audit *logrus.Entry) error {
 	if _env.IsLocalDevelopmentMode() {
 		keys = []string{
 			"PULL_SECRET",
+			env.OIDCStorageAccountName,
 		}
 	} else {
 		keys = []string{
@@ -61,6 +62,7 @@ func rp(ctx context.Context, log, audit *logrus.Entry) error {
 			"CLUSTER_MDM_NAMESPACE",
 			"MDM_ACCOUNT",
 			"MDM_NAMESPACE",
+			env.OIDCStorageAccountName,
 		}
 
 		if _, found := os.LookupEnv("PULL_SECRET"); found {
--- a/pkg/env/prod.go
+++ b/pkg/env/prod.go
@@ -205,10 +205,6 @@ func newProd(ctx context.Context, log *logrus.Entry, component ServiceComponent)
 		p.gatewayDomains = append(p.gatewayDomains, p.acrDomain, acrDataDomain)
 	}
 
-	if err := ValidateVars(OIDCStorageAccountName); err != nil {
-		return nil, err
-	}
-
 	p.ARMHelper, err = newARMHelper(ctx, log, p)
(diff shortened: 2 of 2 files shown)


tsatam (approved):

cmd/aro/rp.go
```diff
@@ -52,6 +52,7 @@ func rp(ctx context.Context, log, audit *logrus.Entry) error {
 	if _env.IsLocalDevelopmentMode() {
```
cadenmarchese, line 52: Is it needed in development?

cmd/aro/rp.go
```diff
@@ -61,6 +62,7 @@ func rp(ctx context.Context, log, audit *logrus.Entry) error {
 			"CLUSTER_MDM_NAMESPACE",
 			"MDM_ACCOUNT",
 			"MDM_NAMESPACE",
+			env.OIDCStorageAccountName,
```
cadenmarchese, line 65: Why not use `OIDC_STORAGE_ACCOUNT_NAME` to match the other vars in this slice?

kimorris27 (approved):

cmd/aro/rp.go
rajdeepc2792: Yes it is needed as there will a need to create a bloburl which will contain the storage account name.

cmd/aro/rp.go
rajdeepc2792: Since the `OIDC_STORAGE_ACCOUNT_NAME ` is used at multiple places:-
Validation and Fetching env variable. Preferred to have a single source of truth in a const.

s-fairchild (approved):

cmd/aro/rp.go
```diff
@@ -52,6 +52,7 @@ func rp(ctx context.Context, log, audit *logrus.Entry) error {
 	if _env.IsLocalDevelopmentMode() {
 		keys = []string{
 			"PULL_SECRET",
+			env.OIDCStorageAccountName,
```
gouthamMN, line 55: Nit: Why not initialize `keys ` list with `env.OIDCStorageAccountName` outside `if ` condition since it's used in both `if ` & `else ` part

tsatam: I agree with this refactor, but for the sake of fixing the broken deployment I suggest we tackle this in a follow-up PR.

PR #3744, opened 2024-07-31
Ensure identity headers are persisted in requests to create MIWI clusters

kimorris27:

This issue came up while I was working on https://issues.redhat.com/browse/ARO-4360

- Updates the unit tests for `putOrPatchOpenShiftCluster` to use the latest API version
- Fixes a nil pointer dereference in the converter for the new preview API
- Shifts the code to persist the identity headers further down in the code to ensure that the headers don't get lost during the repeated conversions between external and internal API types

### Test plan for issue:

New unit test case

N/A

MIWI feature will be tested extensively when development is done


kimorris27:
Opened from my fork by mistake. Closing and will open a new one so I can run E2E.

PR #3745, opened 2024-07-31
Containerized portal as it runs in prod, locally

SudoBrendan:

Fixes [ARO-9500](https://issues.redhat.com/browse/ARO-9500)

You can now test portal changes locally the same way it's run in prod - via a container run process on `localhost:8444`

### Test plan for issue:

I need confirmation from MacOS users that this works for them - I updated some Portal server configs like we did for Shubhada's PR for RP (e.g. directly use `:8444` instead of `localhost:8444` so it works with podman machine setups).

I have validated on Linux that I can run:
1. `make runlocal-portal`
3. navigate to `https://localhost:8444` to see the portal in action.

Yes, it's in the PR.

This is still an unused container image - it is useful now in local development only and is only bringing conformity between prod and local dev.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,33 @@ docker-secrets: aks.kubeconfig
 	docker secret rm --ignore proxy.crt
 	docker secret create proxy.crt ./secrets/proxy.crt
 
+.PHONY: runlocal-portal
+runlocal-portal: ci-rp docker-secrets
+	podman run \
+		--name aro-portal \
+		--rm \
+		--cap-drop net_raw \
+		-e RP_MODE \
+		-e AZURE_SUBSCRIPTION_ID \
+		-e AZURE_TENANT_ID \
+		-e LOCATION \
+		-e RESOURCEGROUP \
+		-e AZURE_PORTAL_CLIENT_ID \
+		-e AZURE_PORTAL_ELEVATED_GROUP_IDS \
+		-e AZURE_PORTAL_ACCESS_GROUP_IDS \
+		-e AZURE_RP_CLIENT_SECRET \
+		-e AZURE_RP_CLIENT_ID \
+		-e KEYVAULT_PREFIX \
+		-e DATABASE_ACCOUNT_NAME \
+		-e DATABASE_NAME \
+		-e NO_NPM=1 \
+		--secret proxy-client.key,target=/app/secrets/proxy-client.key \
+		--secret proxy-client.crt,target=/app/secrets/proxy-client.crt \
+		--secret proxy.crt,target=/app/secrets/proxy.crt \
+		-p 127.0.0.1:8444:8444 \
+		-p 127.0.0.1:2222:2222 \
+		$(RP_IMAGE_LOCAL) portal
+
 # Target to run the local RP
 .PHONY: runlocal-rp
 runlocal-rp: ci-rp docker-secrets
@@ -132,7 +159,6 @@ runlocal-rp: ci-rp docker-secrets
 		--secret proxy.crt,target=/app/secrets/proxy.crt \
 		$(RP_IMAGE_LOCAL) rp
 
-		
 .PHONY: az
 az: pyenv
 	. pyenv/bin/activate && \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -67,6 +67,47 @@ build-all:
 aro: check-release generate
 	go build -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=$(VERSION)" ./cmd/aro
 
+
+# Target to create docker secrets
+.PHONY: docker-secrets
+docker-secrets: aks.kubeconfig
+	docker secret rm --ignore aks.kubeconfig
+	docker secret create aks.kubeconfig ./aks.kubeconfig
+
```
SudoBrendan, line 76: todo: remove this

Makefile
```diff
@@ -67,6 +67,47 @@ build-all:
...
+	docker secret rm --ignore proxy-client.key
+	docker secret create proxy-client.key ./secrets/proxy-client.key
+
+	docker secret rm --ignore proxy-client.crt
+	docker secret create proxy-client.crt ./secrets/proxy-client.crt
+
+	docker secret rm --ignore proxy.crt
+	docker secret create proxy.crt ./secrets/proxy.crt
```
bitoku, line 84: IIRC we aren't supposed to use docker command any more.

https://www.redhat.com/sysadmin/podman-kubernetes-secrets
podman also has secret command. Doesn't it work?

SudoBrendan: We could move this to strictly say `podman` - it's been assumed for a while that all ARO devs are using `podman-docker` library to effectively `alias docker=podman` and not use the `docker` binary - `podman` is what I was testing with locally under the hood here.

I can update if we've decided to move away from the `docker` alias - but `docker` is ubiquitous throughout the Makefile today. If we want to do that, I think it should be done all at once and in a different PR maybe?

mociarain: FWIW +1 for being explicit with `podman` and +1 for a separate PR to convert the lot

bitoku: If we concern downstream impacts, we can do one by one starting with this PR.
IMHO I don't like the way to do all at once because we can't determine what breaks what.

And also `runlocal-portal` uses podman, so we're good to use podman here as well, I guess.

tsatam (approved):
LGTM. I tested this out (manually by running the make targets' implementation directly against other portal PRs) and the portal worked fine for all our use cases.

SudoBrendan:
Adding `hold` because I'm getting reports this doesn't work with podman machine / MacOs

Makefile
```diff
@@ -67,6 +67,44 @@ build-all:
...
+
+.PHONY: runlocal-portal
+runlocal-portal: ci-rp docker-secrets
+	podman run \
```
kevinobriendotca, line 84: We should set the container name in the podman run command.  The RP is aro-rp, maybe aro-portal for the portal?

Makefile
```diff
@@ -67,6 +67,44 @@ build-all:
...
+		-e AZURE_RP_CLIENT_SECRET \
+		-e AZURE_RP_CLIENT_ID \
+		-e KEYVAULT_PREFIX \
+		-e DATABASE_ACCOUNT_NAME \
+		-e DATABASE_NAME \
+		-e NO_NPM=1 \
+		--secret proxy-client.key,target=/app/secrets/proxy-client.key \
+		--secret proxy-client.crt,target=/app/secrets/proxy-client.crt \
+		--secret proxy.crt,target=/app/secrets/proxy.crt \
+		-p 8444:8444 \
```
kevinobriendotca, line 104: We should probably restrict the listeners to localhost, same as the aro-rp container.

-p 127.0.0.1:8444:8444
-p 127.0.0.1:2222:2222

SudoBrendan: I think* this breaks authentication redirects, I know I tried something similar to this and it didn't work. Do you have something that does?

shubhadapaithankar (dismissed):
LGTM
I tested your PR, and it's working great on my Mac.

shubhadapaithankar (dismissed):

kevinobriendotca (approved):

Makefile
SudoBrendan: I created a follow-up PR for this (and some other things I've noticed we need) https://github.com/Azure/ARO-RP/pull/3775

shubhadapaithankar (approved):

PR #3747, opened 2024-08-01
Add PVC smoke tests

bitoku:

Fixes https://issues.redhat.com/browse/ARO-7347

This PR adds tests for the default PVCs.

### Test plan for issue:

N/A

This is not a production feature.

### DIFF
--- a/test/e2e/helpers.go
+++ b/test/e2e/helpers.go
@@ -5,11 +5,15 @@ package e2e
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/yaml"
 
 	projectv1 "github.com/openshift/api/project/v1"
 	projectclient "github.com/openshift/client-go/project/clientset/versioned"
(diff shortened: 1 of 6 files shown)


test/e2e/ocp_persistent_volume.go
```diff
@@ -0,0 +1,64 @@
...
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+var _ = Describe("Persistent Volume", Label(smoke), func() {
+	DescribeTable("should provision PVCs", func(ctx context.Context, pvcName string) {
+		const namespace = "default"
+		manifest := fmt.Sprintf("static_resources/pvc-%s.yaml", pvcName)
+		podName := fmt.Sprintf("bb-%s", pvcName)
+		By(fmt.Sprintf("Creating a pod with %s", pvcName))
+		pod, err := loadResourcesFromYaml(manifest)
```
AldoFusterTurpin, line 24: If the previous comment is useful, this block would change to
```go
f, err := staticResources.Open(manifestPath)
if err != nil {
}
defer func() {
	err := f.Close()
}()
pod, err := loadResourcesFromReader(f)
```
changing _manifest_ to _manifestPath_.

On top of that, I think it would be a good idea to check the error of f.Close()

test/e2e/helpers.go
AldoFusterTurpin, line 1: Would it make sense to not just test that pvc.Status.Phase is ClaimBound and pv.Status.Phase is VolumeBound, but also check that there exists a link between that exact PVC with that pod? (maybe the answer is no, I am just thinking out loud)
🙂

AldoFusterTurpin (approved):
LGTM. Thank you! 

I just left a question about the pod + pvc testing and a suggestion in one function, none of that blocking the PR.

test/e2e/helpers.go
bitoku: The link is already checked implicitly because the PV is fetched according to the PVC's volume name.

test/e2e/helpers.go
```diff
@@ -224,3 +230,73 @@ func (p Project) VerifyProjectIsDeleted(ctx context.Context) {
...
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			return nil, err
+		}
+		objs = append(objs, obj)
+	}
+	return objs, nil
+}
```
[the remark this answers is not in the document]
bitoku: make sense 😃

SudoBrendan (commented):
Thanks for this PR @bitoku, the code is really elegant, and I think we could actually refactor a lot of our current tests to use the pattern of "apply straight-up yaml from this file" rather than authoring golang structs manually and applying (e.g. our StatefulSet test and I'm sure others too).

I don't have concerns with the code, but I do wonder what kind of value this test adds to our codebase validation. To get at what I mean here - what would be an ARO SRE action item to take if this test suite fails on a PR? Would it be:

1. Modifying RP Code?
2. Submitting an OCPBUGS report?

My guess is it's 2, right? Would we be better off trusting OCP conformance testing on this case, considering that ARO does nothing special with the StorageClasses in the cluster by default compared to OCP? Or is there RP functionality I'm unaware of as it relates to modifying the setup of default StorageClasses in clusters (maybe a new y-version thing I just missed)?

SudoBrendan (approved):

tiguelu (approved):

PR #3757, opened 2024-08-05
Lay groundwork for cluster MSI usage

kimorris27:

https://issues.redhat.com/browse/ARO-4360

This PR lays the groundwork for later usage of cluster MSI by:

- Instantiating an MSI dataplane client
- Creating and persisting an MSI certificate during bootstrapping and deleting the certificate during cluster deletion
- Setting up a feature flag that the RP uses to know whether to use the mock MSI dataplane (since INT doesn't have a prod MSI RP, checking `IsLocalDevelopmentMode()` is not sufficient)

### Test plan for issue:

- Added unit tests for new functions
- Added in some code (removed before opening PR) to create federated credentials on the platform workload identities, which validated that the MSI dataplane stub is working, the certificates are being handled correctly, etc.

I don't think so, but let me know if I missed anything.

MIWI will be tested extensively before the feature is released.


cadenmarchese (commented):
quick first pass, looking good! Just a few small questions.

pkg/cluster/clustermsi.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/msi-dataplane/pkg/dataplane"
+	"github.com/Azure/msi-dataplane/pkg/store"
+
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/armmsi"
+)
+
+const (
+	msiCertValidityDays = 90
```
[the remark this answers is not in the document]
kimorris27: I like to give a name to this kind of magic number as a form of documentation-as-code. If we ever need to change this validity period, the person doing it won't have to read as much code to understand how to make the change.

pkg/deploy/assets/cluster-development-predeploy.json
```diff
@@ -104,6 +104,7 @@
                 },
                 "accessPolicies": [],
                 "enabledForDiskEncryption": true,
+                "enableRbacAuthorization": false,
```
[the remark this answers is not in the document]
kimorris27: That's for a different storage account. There are three key vaults that get deployed in dev: the MSI one and two others. The other two use the legacy key vault access policies instead of Entra ID RBAC, which is why I made this attribute configurable through the `keyVault()` function and only set it to true for the MSI storage account.

kimorris27:
Stuck a hold label on this because I remembered that I had to comment out the `validateResources` step during cluster bootstrapping to test the cluster MSI stuff since dynamic validation for MSI clusters hasn't merged yet.

pkg/cluster/clustermsi.go
```diff
@@ -0,0 +1,172 @@
...
+func (m *manager) clusterMsiResourceId() (*arm.ResourceID, error) {
+	var clusterMsi *arm.ResourceID
+	if m.doc.OpenShiftCluster.Identity != nil && m.doc.OpenShiftCluster.Identity.UserAssignedIdentities != nil {
+		for msiResourceId := range m.doc.OpenShiftCluster.Identity.UserAssignedIdentities {
+			var err error
+			clusterMsi, err = arm.ParseResourceID(msiResourceId)
+			if err != nil {
+				return nil, err
+			}
+		}
```
gouthamMN, line 164: So, we always pick the last non errored out MSI. Can we pick the first non-errored out MSI instead to save some cycles? Lmk, if this would change the actual behavior for any reason.

pkg/cluster/delete.go
```diff
@@ -356,6 +357,27 @@ func (m *manager) deleteGateway(ctx context.Context) error {
...
+		return nil
+	}
+
+	secretName, err := m.clusterMsiSecretName()
+	if err != nil {
+		return err
+	}
+
+	err = m.clusterMsiKeyVaultStore.DeleteCredentialsObject(ctx, secretName)
+	if azcoreErr, ok := err.(*azcore.ResponseError); !ok || azcoreErr.StatusCode != 404 {
```
gouthamMN, line 374: Should we check if `err != nil` here first?

pkg/cluster/cluster.go
```diff
@@ -5,13 +5,21 @@ package cluster
...
 	"net/http"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
+	"github.com/Azure/go-autorest/autorest/to"
```
niontive, line 19: `go-autorest` is out of support: https://github.com/Azure/go-autorest

It looks like we're using `to` get string ptrs. Could we just write a super simple function to do this so we don't take a dependency on deprecated code?

niontive: This is a nit comment

pkg/cluster/clustermsi.go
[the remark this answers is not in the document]
niontive: +1 to documenting magic numbers

pkg/cluster/clustermsi.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/Azure/msi-dataplane/pkg/store"
+
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/armmsi"
+)
+
+const (
+	msiCertValidityDays = 90
+)
+
+// ensureClusterMsiCertificate leverages the MSI dataplane module to create a new MSI
```
niontive, line 26: Nit: `module to fetch the backing certificate..`

pkg/cluster/clustermsi.go
```diff
@@ -0,0 +1,172 @@
...
+		TenantID:    m.doc.OpenShiftCluster.Identity.TenantID,
+	}
+
+	msiCredObj, err := m.msiDataplane.GetUserAssignedIdentities(ctx, uaMsiRequest)
+	if err != nil {
+		return err
+	}
+
+	now := time.Now()
+	expirationDate := now.AddDate(0, 0, msiCertValidityDays)
```
niontive, line 60: You can get this date from the credentials object - use `not_after` in nested credentials object

pkg/cluster/clustermsi.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// clusterMsiResourceId returns the resource ID of the cluster MSI or an error
+// if it encounters an issue while grabbing the resource ID from the cluster
+// doc. It is written under the assumption that there is only one cluster MSI
+// and will have to be refactored if we ever use more than one.
+func (m *manager) clusterMsiResourceId() (*arm.ResourceID, error) {
+	var clusterMsi *arm.ResourceID
+	if m.doc.OpenShiftCluster.Identity != nil && m.doc.OpenShiftCluster.Identity.UserAssignedIdentities != nil {
+		for msiResourceId := range m.doc.OpenShiftCluster.Identity.UserAssignedIdentities {
```
gouthamMN, line 158: So, we always pick the last non errored out MSI. Can we pick the first non-errored out MSI instead to save some cycles? Lmk, if this would change the actual behavior for any reason.

pkg/cluster/clustermsi.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return err
+	}
+
+	certs, key, err := azidentity.ParseCertificates(decodedCertData, nil)
+	if err != nil {
+		return err
+	}
+
+	cred, err := azidentity.NewClientCertificateCredential(tenantId, clientId, certs, key, m.env.Environment().ClientCertificateCredentialOptions())
```
niontive, line 118: You can just use this guy: https://github.com/Azure/msi-dataplane/blob/94e135accf5ed734b30612be73fdb8537d8ae2ce/pkg/dataplane/identity.go#L43

pkg/cluster/delete.go
kimorris27: I did thorough unit test cases here, and it works correctly when the error is non-nil and not an `azcore.ResponseError`.

But now that you've asked this question, I see that even though it works correctly it reads unintuitively. I'll think of a way to simplify it so that it reads better.

pkg/cluster/cluster.go
kimorris27: I'll consider this out of scope for this PR since it would have to be changed all throughout the RP and not just here.

pkg/cluster/clustermsi.go
kimorris27: Static validation enforces the presence of one and only one MSI. I used a for loop instead of doing `m.doc.OpenShiftCluster.Identity.UserAssignedIdentities[0]` because I thought it was an elegant way to handle the case where `m.doc.OpenShiftCluster.Identity.UserAssignedIdentities` is empty (which would only happen if this function got called on a non-WI cluster by mistake, but still).

I also noted in the doc comment for this function that it'll need to be refactored if we ever use more than one MSI.

Making the change you suggested wouldn't change the actual behavior, but it wouldn't save any cycles either.

pkg/cluster/clustermsi.go
kimorris27: Are you sure? How do we know that it will be 90 days (since this value was agreed upon during the design phase)? Also, I think the stub would return a nil `NotAfter`, so we'd have to add a conditional to set this "manually" when using the stub.

pkg/cluster/clustermsi.go
kimorris27: I tried using it, but I got stuck :/

The key vault storage client only accepts the `CredentialsObject` type, so after getting it from the key vault, I have to make the `UserAssignedIdentities` object myself. But there's no constructor function and the `cloud` field is package-private, so I can't set it. This won't work in FairFax.

Am I understanding correctly? If so, I think the dataplane module either needs a `UserAssignedIdentities` constructor function or some other tweak.

pkg/cluster/delete.go
kimorris27: I pushed an update here. Let me know if this didn't help!

pkg/cluster/clustermsi.go
kimorris27: I pushed an update here. Let me know if I didn't correctly capture your intended meaning!

pkg/cluster/cluster.go
bitoku: It's not that I want you to fix this in this PR, but I think changing little by little is better than changing all at once because it will narrow the scope and we don't have to be bothered by conflicts.

Also https://pkg.go.dev/k8s.io/utils/ptr#To is already available in the codebase, so we can replace it now.

kimorris27: I went ahead and switched to that function you pointed out since it's already available and easy enough 👍🏻

bitoku (commented):
I haven't used it but there's a fake client for armmsi.
https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/resourcemanager/msi/armmsi#fakes
With this, you don't have to make mock client perhaps.

I don't know if it can be used here, but just in case you aren't aware.

pkg/cluster/clustermsi.go
bitoku: What if NotAfter is earlier than 90days?
Can't this happen?

niontive: 1. MSI team has a doc on this: https://eng.ms/docs/products/arm/rbac/managed_identities/msionboardingcertificaterotation. 

`day 90: The certificate expires in Azure AD and can no longer be used to get tokens. This is the not_after field in the credential response.`

2. What if the MSI team makes a change and says cert expiration is now at 45 days? Given the above doc, it's better to rely on a value received from them than to hardcode something on our end.

3. Re stub -> we can just set `NotAfter` to some arbitrary date in the future for testing (ex: one minute from now).

pkg/cluster/clustermsi.go
niontive: I'll fix this in `msi-dataplane` and will post an update.

niontive: PR to fix this: https://github.com/Azure/msi-dataplane/pull/20/files

pkg/cluster/clustermsi.go
kimorris27: Oh, I see. Then I'll use the `NotAfter` returned by the MSI RP.

pkg/cluster/cluster.go
```diff
@@ -281,5 +287,110 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
...
+		switch strings.ToUpper(_env.Environment().Name) {
+		case dataplane.AzurePublicCloud:
+			cloud = dataplane.AzurePublicCloud
+		case dataplane.AzureUSGovCloud:
+			cloud = dataplane.AzureUSGovCloud
+		}
+
+		if cloud == "" {
+			return nil, errors.New("could not determine which Azure Cloud to use to instantiate MSI dataplane client")
+		}
```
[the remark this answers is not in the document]
kimorris27: No, I don't think that will help. `_env.Environment()` already gives us the correct `AROEnvironment`.

The reason I had to put in this custom logic is because the cloud names are all uppercase in the [MSI dataplane module](https://github.com/Azure/msi-dataplane/blob/4e8e11fa5d231732524fbcfa65d96b98081eae7c/pkg/dataplane/constants.go#L7) but they're in camel case when you get them from the `AROEnvironment`.

I guess I could pull this out into a helper function so that you don't have to care about it as you're trying to read through and understand at a high-level how the cluster manager is being initialized.

pkg/cluster/cluster.go
```diff
@@ -281,5 +287,110 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 		hiveClusterManager:                hiveClusterManager,
 		now:                               func() time.Time { return time.Now() },
 		openShiftClusterDocumentVersioner: new(openShiftClusterDocumentVersionerService),
-	}, nil
+	}
+
+	if doc.OpenShiftCluster.UsesWorkloadIdentity() {
```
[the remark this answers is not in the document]
kimorris27: I don't think this whole block necessarily needs to be separated out into another function, since it's specific to the initialization of the cluster manager, which is what this function is for. If your concern is that it's long and distracting from the high-level idea, I think that will be addressed once I've pulled a few of the more complex pieces into helper functions based on some of your other suggestions.

pkg/cluster/delete_test.go
```diff
@@ -49,25 +53,25 @@ func TestDeleteNic(t *testing.T) {
 			name: "nic is in succeeded provisioning state",
 			mocks: func(networkInterfaces *mock_network.MockInterfacesClient) {
 				nic.InterfacePropertiesFormat.ProvisioningState = mgmtnetwork.Succeeded
-				networkInterfaces.EXPECT().Get(gomock.Any(), clusterRG, nicName, "").Return(nic, nil)
-				networkInterfaces.EXPECT().DeleteAndWait(gomock.Any(), clusterRG, nicName).Return(nil)
+				networkInterfaces.EXPECT().Get(deprecratedgomock.Any(), clusterRG, nicName, "").Return(nic, nil)
```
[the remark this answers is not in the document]
kimorris27: Wait, never mind... I tried again and I was reminded of why I didn't do it in the first place: these other mock clients need the `deprecatedgomock` because they are generated by us, and they currently use the deprecated one instead of the Uber one.

I agree that we should switch to the Uber module, but I think changing the way the mocks are generated is out of scope for this PR.

pkg/cluster/cluster.go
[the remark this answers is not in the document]
bitoku: Sorry, yes you are right.
It sounds good to make a new helper function!

pkg/cluster/delete.go
```diff
@@ -343,7 +344,7 @@ func (m *manager) deleteGateway(ctx context.Context) error {
 		return nil
 	}
 
-	// https://docs.microsoft.com/en-us/azure/cosmos-db/change-feed-design-patterns#deletes
+	// http!s://docs.microsoft.com/en-us/azure/cosmos-db/change-feed-design-patterns#deletes
```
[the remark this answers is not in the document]
kimorris27: Fixed :)

pkg/cluster/delete.go
```diff
@@ -356,6 +357,29 @@ func (m *manager) deleteGateway(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	err = m.clusterMsiKeyVaultStore.DeleteCredentialsObject(ctx, secretName)
+	azcoreErr, ok := err.(*azcore.ResponseError)
+	isNotFoundErr := ok && azcoreErr.StatusCode == 404
+	if err != nil && !isNotFoundErr {
+		return err
+	}
```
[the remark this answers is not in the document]
kimorris27: Good suggestion. I found that our codebase actually has a util function for doing exactly what you suggested, so I was able to use that here.

pkg/cluster/cluster.go
```diff
@@ -281,5 +287,110 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
...
+		}
+
+		clusterMsiKeyVaultName := os.Getenv(env.KeyvaultPrefix) + env.ClusterMsiKeyVaultSuffix
+		if _env.IsLocalDevelopmentMode() {
+			prefix := os.Getenv("RESOURCEGROUP")
+			if len(prefix) > 10 {
+				prefix = prefix[:10]
+			}
+			clusterMsiKeyVaultName = prefix + generator.SharedMSIKeyVaultNameSuffix
+		}
```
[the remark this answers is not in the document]
kimorris27: Done!

pkg/cluster/clustermsi.go
kimorris27: Fixed!

pkg/cluster/clustermsi.go
kimorris27: Yep, it worked well this time! All fixed.

pkg/cluster/install.go
```diff
@@ -289,22 +289,34 @@ func setFieldCreatedByHive(createdByHive bool) database.OpenShiftClusterDocument
...
+	}
 
-		// TODO: this relies on an authorizer that isn't exposed in the manager
-		// struct, so we'll rebuild the fpAuthorizer and use the error catching
-		// to advance
-		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.clusterSPObjectID),
+	if m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		s = append(s,
+			steps.Action(m.ensureClusterMsiCertificate),
+			steps.Action(m.initializeClusterMsiClients),
```
rajdeepc2792, line 307: This step will need to be performed before the dynamic validation, as there will be a need to populate the cliendid and objectid of the platform identities. Also via the MSI Certificate there will be a need to populate the objectid for cluster MSI. objectid will be needed to use the checkaccess api. 

The cluster MSI dynamic validation will contain:-
- Checking if the cluster MSI has sufficient permission to read platformidentities and read/update their corresponding federated credentials.

Also, the msi clients initialization will be required for the update flow, will that be covered in this story or somewhere else.

pkg/api/openshiftcluster.go
```diff
@@ -32,6 +35,29 @@ func (oc *OpenShiftCluster) UsesWorkloadIdentity() bool {
 	return oc.Properties.PlatformWorkloadIdentityProfile != nil && oc.Properties.ServicePrincipalProfile == nil
 }
 
+// ClusterMsiResourceId returns the resource ID of the cluster MSI or an error
+// if it encounters an issue while grabbing the resource ID from the cluster
+// doc. It is written under the assumption that there is only one cluster MSI
+// and will have to be refactored if we ever use more than one.
+func (oc *OpenShiftCluster) ClusterMsiResourceId() (*arm.ResourceID, error) {
```
rajdeepc2792, line 42: This relies on the assumption that the UserAssignedIdentities will have a single identity always. I know we are validating it as part of static validation but do you want to make sure we return the clusterMsi on the first iteration itself, or add an additional check here?

pkg/cluster/cluster.go
```diff
@@ -281,5 +287,110 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
...
+			cloud = dataplane.AzureUSGovCloud
+		}
+
+		if cloud == "" {
+			return nil, errors.New("could not determine which Azure Cloud to use to instantiate MSI dataplane client")
+		}
+
+		msiDataplaneClientOptions := clientOptions.ClientOptions
+
+		if _env.FeatureIsSet(env.FeatureUseMockMsiRp) {
```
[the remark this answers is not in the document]
rajdeepc2792: Can we also move the validatevars to here:- https://github.com/Azure/ARO-RP/blob/master/cmd/aro/rp.go#L73
So that we validate these vars when RP comes up in dev rather than for every cluster creation?

pkg/api/openshiftcluster.go
kimorris27: Goutham had asked a similar question, and I responded: https://github.com/Azure/ARO-RP/pull/3757#discussion_r1707873400

But since this has come up twice, there's probably value in changing this function to explicitly check for one cluster MSI instead of using a loop. I'll make the change.

pkg/cluster/cluster.go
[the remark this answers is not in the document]
kimorris27: >Can we also move the validatevars to here:- https://github.com/Azure/ARO-RP/blob/master/cmd/aro/rp.go#L73
So that we validate these vars when RP comes up in dev rather than for every cluster creation?

I'd rather leave it where it is because if we validate the vars on RP startup we'd be forcing people who don't need to create MIWI clusters to set env vars they aren't going to use. Since setting the env vars correctly is non-trivial, I find it easy to imagine that people would end up setting them to nonsensical values just to get the pre-startup validation to work, which would be an annoying extra step to start a local dev or INT RP.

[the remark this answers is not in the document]
kimorris27: >I would make a new method in env that returns the mocked ClientOptions, and use it here.

I'll still be making this change though.

pkg/cluster/install.go
kimorris27: Gotcha, I'll fix the order of the steps.

>Also, the msi clients initialization will be required for the update flow, will that be covered in this story or somewhere else?

I figured it could be handled as needed in another PR that actually uses the cluster MSI clients during cluster updates.

kimorris27:
> I haven't used it but there's a fake client for armmsi. https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/resourcemanager/msi/armmsi#fakes With this, you don't have to make mock client perhaps.
> 
> I don't know if it can be used here, but just in case you aren't aware.

@gouthamMN, you might find this useful for the federated identity credential work.

pkg/api/openshiftcluster.go
kimorris27: Done!

pkg/cluster/install.go
kimorris27: Done!

pkg/deploy/devconfig.go
```diff
@@ -189,26 +190,27 @@ func DevConfig(_env env.Core) (*Config, error) {
 				"DisableReadinessDelay",
 				"EnableOCMEndpoints",
 				"RequireOIDCStorageWebEndpoint",
+				"UseMockMsiRp",
```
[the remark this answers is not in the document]
kimorris27: Done!

pkg/cluster/cluster.go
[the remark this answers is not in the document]
kimorris27: Done with the `ClientOptions` change 👍🏻

rajdeepc2792:
> I haven't used it but there's a fake client for armmsi. https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/resourcemanager/msi/armmsi#fakes With this, you don't have to make mock client perhaps.
> 
> I don't know if it can be used here, but just in case you aren't aware.

Yes there are fakes for all(mostly) the clients under https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/resourcemanager/ , and I believe we should start relying on them.
Ref:- https://github.com/Azure/azure-sdk-for-go/issues/16613#issuecomment-1883940233

pkg/api/openshiftcluster.go
```diff
@@ -32,6 +35,29 @@ func (oc *OpenShiftCluster) UsesWorkloadIdentity() bool {
...
+			var err error
+			clusterMsi, err = arm.ParseResourceID(msiResourceId)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	if clusterMsi == nil {
+		return nil, errors.New("could not find cluster MSI in cluster doc or unexpectedly found more than one cluster MSI")
```
cadenmarchese, line 55: Would it make sense to separate out these cases in code (and in your error message)?
1) Could not find cluster MSI in cluster doc
2) Unexpectedly found more than one cluster MSI

pkg/cluster/install.go
```diff
@@ -288,23 +288,39 @@ func setFieldCreatedByHive(createdByHive bool) database.OpenShiftClusterDocument
...
-	s := []steps.Step{
-		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.validateResources),
+	s := []steps.Step{}
+
+	if m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		s = append(s,
+			steps.Action(m.ensureClusterMsiCertificate),
+			steps.Action(m.initializeClusterMsiClients),
+		)
+	}
```
cadenmarchese, line 298: Not entirely related to your PR, but since you're introducing `if` blocks during bootstrap, should `createOIDC` be included in one as well?

I know order matters, but it seems like we have some wi/mi steps in if blocks now, and others in line with the rest.

pkg/deploy/devconfig.go
```diff
@@ -8,7 +8,7 @@ import (
 	"encoding/pem"
 	"os"
 
-	"github.com/Azure/go-autorest/autorest/to"
+	"k8s.io/utils/ptr"
```
tsatam, line 11: Was there a specific reason this was changed? I'm fine with the move, but I want to make sure I'm not missing any potential behavioral differences that necessitated the move.

kimorris27: Nic pointed out in his review that `github.com/Azure/go-autorest/autorest/to` is deprecated, so I went ahead and switched to a different "to pointer" function in the files I touched in this PR. There are no changes in behavior.

pkg/cluster/install.go
kimorris27: It looks like the reason that `createOIDC` doesn't need to be in an `if` statement is because it does nothing and returns `nil` when called for a CSP cluster, which I agree is a bit confusing now that you point it out. I think refactoring that (if we want to) would be out of scope for this PR though.

kimorris27: To be clear, when I say "refactor", I mean:

1. Modify `createOIDC` such that it returns an error when called on a CSP cluster
2. Update the bootstrap steps such that `createOIDC` only gets called for MIWI clusters

If I were to decide how to refactor it myself, this would make the most sense to me, because this way if we accidentally call a MIWI step on a CSP cluster as we're developing locally, we get an error that informs us we shouldn't have called that step.

kimorris27: With that in mind, I think I'll update the two MIWI steps I've added in this PR to error out if called on CSP clusters.

kimorris27: On second thought, I think I'm overcomplicating things and I should just leave my new code how it is. Let me know if you have other thoughts though.

pkg/api/openshiftcluster.go
kimorris27: Agreed and done 👍🏻

pkg/util/mocks/env/env.go
SudoBrendan, line 1: flagging this as potentially needing a rebase on master to pick up the .bingo changes. Dismiss if untrue (but it probably doesn't hurt to do it anyways).

SudoBrendan (changes requested):

pkg/util/azureclient/azuresdk/azsecrets/client.go
```diff
@@ -0,0 +1,39 @@
+package azsecrets
```
rajdeepc2792, line 1: Same as above:-
It seems the functions for the client are not used till now anywhere, but would it be good to have a mocks generated for it now?

pkg/util/azureclient/azuresdk/azsecrets/client.go
```diff
@@ -0,0 +1,39 @@
...
+	SetSecret(ctx context.Context, name string, parameters azsecrets.SetSecretParameters, options *azsecrets.SetSecretOptions) (azsecrets.SetSecretResponse, error)
+}
+
+type client struct {
+	*azsecrets.Client
+}
+
+var _ Client = &client{}
+
+func NewClient(vaultURL string, credential azcore.TokenCredential, options *arm.ClientOptions) (Client, error) {
```
rajdeepc2792, line 31: Same as above:-
It's the exact implementation that was asked to be changed here:- https://github.com/Azure/ARO-RP/pull/3619#discussion_r1736492379
That is not returning the interface type.

rajdeepc2792 (changes requested):
Functionality LGTM, added some suggestions around mocks and refactoring.

pkg/cluster/delete.go
```diff
@@ -458,6 +480,14 @@ func (m *manager) Delete(ctx context.Context) error {
...
 	}
 
+	if m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		m.log.Printf("deleting cluster MSI certificate")
+		err = m.deleteClusterMsiCertificate(ctx)
+		if err != nil {
+			return err
+		}
+	}
+
```
tsatam, line 490: I'm assuming it likely doesn't matter, but are we sure that deleting the cluster MSI certificate before deleting the resources is safe to do, e.g. in a situation where the resource deletion fails incredibly early and a user continues to use the cluster as-is? 

Is there any explicit reason why we wouldn't want to delete the cluster MSI certificate after deleting the resources, or should we just move this deletion to after resource deletion just to be safe?

kimorris27: That makes sense to me. I've made this change.

pkg/util/azureclient/azuresdk/armmsi/federated_identity_credentials.go
```diff
@@ -0,0 +1,37 @@
+package armmsi
```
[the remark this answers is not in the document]
kimorris27: Like we discussed in standup, I encountered some issues generating mocks that will require some nontrivial changes to fix, so I'll be adding the mocks in a follow-up PR.

pkg/util/azureclient/azuresdk/azsecrets/client.go
kimorris27: I'll plan to include the mocks for this client as well in my follow-up PR. I didn't end up needing them in unit tests for this PR because of the limited extent to which I used this client.

pkg/util/azureclient/azuresdk/azsecrets/client.go
kimorris27: I've made this change 👍🏻

pkg/cluster/delete.go
```diff
@@ -356,6 +357,27 @@ func (m *manager) deleteGateway(ctx context.Context) error {
 	return nil
 }
 
+func (m *manager) deleteClusterMsiCertificate(ctx context.Context) error {
+	// The cluster MSI may have been deleted prior to cluster deletion. If that's the case
+	// we will have already deleted the certificate.
+	if m.doc.OpenShiftCluster.Identity == nil || m.doc.OpenShiftCluster.Identity.UserAssignedIdentities == nil || len(m.doc.OpenShiftCluster.Identity.UserAssignedIdentities) == 0 {
```
jaitaiwan, line 363: So this is an example of the previous logic I said might want to be functionized. Especially so it's easier to mentally parse.

jaitaiwan (approved):
Few small nits but LGTM overall

pkg/cluster/clustermsi.go
```diff
@@ -0,0 +1,145 @@
...
+// vault. It does not concern itself with whether an existing certificate is valid
+// or not; that can be left to the certificate refresher component.
+func (m *manager) ensureClusterMsiCertificate(ctx context.Context) error {
+	secretName, err := m.clusterMsiSecretName()
+	if err != nil {
+		return err
+	}
+
+	_, err = m.clusterMsiKeyVaultStore.GetCredentialsObject(ctx, secretName)
+	if err == nil {
```
cadenmarchese, line 37: I assume since we're not storing a var that the purpose of line 36 is to just check that the credentials object exists. If that's the case, do we need line 37 and 38 or should we just proceed to line 39/40?

pkg/cluster/clustermsi.go
```diff
@@ -0,0 +1,145 @@
...
+		return err
+	}
+
+	now := time.Now()
+
+	var expirationDate time.Time
+	if m.env.FeatureIsSet(env.FeatureUseMockMsiRp) {
+		expirationDate = now.AddDate(0, 0, mockMsiCertValidityDays)
+	} else {
+		if msiCredObj.CredentialsObject.ExplicitIdentities == nil || len(msiCredObj.CredentialsObject.ExplicitIdentities) == 0 || msiCredObj.CredentialsObject.ExplicitIdentities[0] == nil || msiCredObj.CredentialsObject.ExplicitIdentities[0].NotAfter == nil {
```
cadenmarchese, line 65: Similarly to @jaitaiwan 's comment, this is pretty long and could be a separate function, but it's just a matter of taste I think.

pkg/cluster/clustermsi.go
```diff
@@ -0,0 +1,145 @@
...
+func (m *manager) initializeClusterMsiClients(ctx context.Context) error {
+	secretName, err := m.clusterMsiSecretName()
+	if err != nil {
+		return err
+	}
+
+	kvSecret, err := m.clusterMsiKeyVaultStore.GetCredentialsObject(ctx, secretName)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
kimorris27: In `install.go`'s bootstrap steps, we call the certificate-ensuring step right before this client initialization step. I think if we ensured here, we'd be blurring the logical lines between the two steps to the point that they might as well be one step, but that would result in a pretty large function.

pkg/cluster/clustermsi.go
kimorris27: No, because if the credentials object already exists, we want to exit early instead of creating a new one and trying to replace the existing one.

pkg/cluster/clustermsi.go
kimorris27: I think I'm gonna skip making a function for this one if that's cool, since I don't foresee us needing to check a `CredentialsObject`'s `NotAfter` outside of this function.

env.example
```diff
@@ -9,4 +9,12 @@ export CLUSTER_NAME="${USER}-aro-cluster"
...
 
-. secrets/env
\ No newline at end of file
+# You'll need these to create MIWI clusters with your local RP, but you can comment them
+# out or remove them from your env file if you're only going to be creating service principal
+# clusters.
+export MOCK_MSI_CLIENT_ID="replace_with_value_output_by_hack/devtools/msi.sh"
+export MOCK_MSI_CERT="replace_with_value_output_by_hack/devtools/msi.sh"
+export MOCK_MSI_TENANT_ID="replace_with_value_output_by_hack/devtools/msi.sh"
+export PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS="replace_with_value_output_by_hack/devtools/msi.sh"
```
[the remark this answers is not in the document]
kimorris27: I could see the value in automation to set these env vars, but I consider it out-of-scope for this PR. I snuck in this small, unrelated change here because someone had mentioned to me in Slack that it's difficult to know you need to set these env vars if you miss the corresponding steps [here](https://github.com/Azure/ARO-RP/blob/master/docs/deploy-development-rp.md#prepare-your-environment).

pkg/cluster/delete.go
kimorris27: Closing this thread out to address this idea in your other thread about it.

pkg/api/openshiftcluster.go
```diff
@@ -32,6 +35,25 @@ func (oc *OpenShiftCluster) UsesWorkloadIdentity() bool {
 	return oc.Properties.PlatformWorkloadIdentityProfile != nil && oc.Properties.ServicePrincipalProfile == nil
 }
 
+// ClusterMsiResourceId returns the resource ID of the cluster MSI or an error
+// if it encounters an issue while grabbing the resource ID from the cluster
+// doc. It is written under the assumption that there is only one cluster MSI
+// and will have to be refactored if we ever use more than one.
+func (oc *OpenShiftCluster) ClusterMsiResourceId() (*arm.ResourceID, error) {
+	if oc.Identity == nil || oc.Identity.UserAssignedIdentities == nil || len(oc.Identity.UserAssignedIdentities) == 0 {
```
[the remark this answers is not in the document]
kimorris27: I considered different ways of going about this issue, and I decided on having the function return a boolean rather than a list of identities, since at least for now we expect the cluster doc to only have one identity. Let me know if that works from your perspective.

cadenmarchese (dismissed):
All my comments have been addressed. thanks!

pkg/cluster/cluster.go
```diff
@@ -289,17 +278,56 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
...
+		}
+
+		authenticatorPolicy := dataplane.NewAuthenticatorPolicy(fpCredRPTenant, _env.MsiRpEndpoint())
+		msiDataplane, err := dataplane.NewClient(cloud, authenticatorPolicy, msiDataplaneClientOptions)
+		if err != nil {
+			return nil, err
+		}
+
+		clusterMsiKeyVaultName := _env.ClusterMsiKeyVaultName()
+		clusterMsiKeyVaultURL := fmt.Sprintf("https://%s.vault.azure.net", clusterMsiKeyVaultName)
```
rajdeepc2792, line 322: Can we use `_env.Environment().KeyVaultDNSSuffix` here instead of `vault.azure.net`??

rajdeepc2792 (changes requested):
Other than the keyvault prefix comment everything else LGTM!!
Please disregard the comment if the const KeyvaultDNSSuffix isn't the correct usage here.

pkg/cluster/cluster.go
kimorris27: Done, and good point. I didn't realize there were different suffixes for different envs, so we might've hit issues in FairFax later on if you hadn't caught this.

rajdeepc2792 (dismissed):

cadenmarchese (approved):

cadenmarchese (approved):

bitoku (approved):

tsatam (approved):

pkg/cluster/clustermsi.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// Note that we are assuming that all of the platform MIs are in the same subscription.
+	pwiResourceId, err := arm.ParseResourceID(m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities[0].ResourceID)
+	if err != nil {
+		return err
+	}
+
+	subId := pwiResourceId.SubscriptionID
```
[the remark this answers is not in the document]
kimorris27: I guess I didn't think about it that much, since we're assuming we'll get the same ID from either place. Since I'll be opening a follow-up PR anyway, how about I switch to using the subscription doc there?

pkg/cluster/clustermsi_test.go
```diff
@@ -0,0 +1,310 @@
...
+			},
+			wantResult: fmt.Sprintf("%s-%s", mockGuid, miName),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			controller := gomock.NewController(t)
+			defer controller.Finish()
+
```
tsatam, line 296: Is this controller unused?

PR #3761, opened 2024-08-07
ARO-6416 generate a secret for the operator from workload identity

yithian:

Fixes https://issues.redhat.com/browse/ARO-6416

Use the provided identity information to generate a secret for the operator

### Test plan for issue:

This will be hard to test until we have a way to provision a cluster that uses workload identities. Until then, unit tests are the best we can do

n/a

As above, this will be very challenging to test until there's a way to provision a workload identity cluster. The new code is gated behind a check to see if a cluster has identity information provided, so at the very least, it shouldn't impact standard clusters

### DIFF
--- a/pkg/cluster/arooperator_test.go
+++ b/pkg/cluster/arooperator_test.go
@@ -141,12 +141,60 @@ func TestInstallAROOperator(t *testing.T) {
 		wantErr string
 	}{
 		{
-			name: "install success",
+			name: "install success with service principal",
 			doc: &api.OpenShiftClusterDocument{
 				Key: strings.ToLower(key),
 				OpenShiftCluster: &api.OpenShiftCluster{
 					ID: key,
 					Properties: api.OpenShiftClusterProperties{
+						ServicePrincipalProfile: &api.ServicePrincipalProfile{},
+						IngressProfiles: []api.IngressProfile{
+							{
+								Visibility: api.VisibilityPublic,
+								Name:       "default",
+							},
+						},
+						ProvisioningState: api.ProvisioningStateAdminUpdating,
+						ClusterProfile: api.ClusterProfile{
+							Version: "4.8.18",
+						},
+						NetworkProfile: api.NetworkProfile{
+							APIServerPrivateEndpointIP: "1.2.3.4",
+						},
+					},
+				},
+			},
+			mocks: func(dep *mock_deploy.MockOperator) {
+				dep.EXPECT().
+					Install(gomock.Any()).
+					Return(nil)
+			},
+		},
+		{
+			name: "install success with identity",
+			doc: &api.OpenShiftClusterDocument{
+				Key: strings.ToLower(key),
+				OpenShiftCluster: &api.OpenShiftCluster{
+					ID: key,
+					Identity: &api.Identity{
+						UserAssignedIdentities: api.UserAssignedIdentities{
+							"fakeIdentity": api.ClusterUserAssignedIdentity{
+								ClientID:    "fake",
+								PrincipalID: "alsoFake",
+							},
+						},
+					},
+					Properties: api.OpenShiftClusterProperties{
+						PlatformWorkloadIdentityProfile: &api.PlatformWorkloadIdentityProfile{
+							PlatformWorkloadIdentities: []api.PlatformWorkloadIdentity{
+								{
+									OperatorName: "openshift-azure-operator",
+									ObjectID:     "00000000-0000-0000-0000-000000000000",
+									ClientID:     "11111111-1111-1111-1111-111111111111",
+									ResourceID:   "/subscriptions/22222222-2222-2222-2222-222222222222/resourceGroups/something/providers/Microsoft.ManagedIdentity/userAssignedIdentities/identity-name",
+								},
+							},
+						},
 						IngressProfiles: []api.IngressProfile{
 							{
 								Visibility: api.VisibilityPublic,
--- a/pkg/cluster/arooperator_test.go
+++ b/pkg/cluster/arooperator_test.go
@@ -141,60 +141,12 @@ func TestInstallAROOperator(t *testing.T) {
 		wantErr string
 	}{
 		{
-			name: "install success with service principal",
+			name: "install success",
 			doc: &api.OpenShiftClusterDocument{
 				Key: strings.ToLower(key),
 				OpenShiftCluster: &api.OpenShiftCluster{
 					ID: key,
 					Properties: api.OpenShiftClusterProperties{
-						ServicePrincipalProfile: &api.ServicePrincipalProfile{},
-						IngressProfiles: []api.IngressProfile{
-							{
-								Visibility: api.VisibilityPublic,
-								Name:       "default",
-							},
-						},
-						ProvisioningState: api.ProvisioningStateAdminUpdating,
-						ClusterProfile: api.ClusterProfile{
-							Version: "4.8.18",
-						},
-						NetworkProfile: api.NetworkProfile{
-							APIServerPrivateEndpointIP: "1.2.3.4",
-						},
-					},
-				},
-			},
-			mocks: func(dep *mock_deploy.MockOperator) {
-				dep.EXPECT().
-					Install(gomock.Any()).
-					Return(nil)
-			},
-		},
-		{
-			name: "install success with identity",
-			doc: &api.OpenShiftClusterDocument{
-				Key: strings.ToLower(key),
-				OpenShiftCluster: &api.OpenShiftCluster{
-					ID: key,
-					Identity: &api.Identity{
-						UserAssignedIdentities: api.UserAssignedIdentities{
-							"fakeIdentity": api.ClusterUserAssignedIdentity{
-								ClientID:    "fake",
-								PrincipalID: "alsoFake",
-							},
-						},
-					},
-					Properties: api.OpenShiftClusterProperties{
-						PlatformWorkloadIdentityProfile: &api.PlatformWorkloadIdentityProfile{
-							PlatformWorkloadIdentities: []api.PlatformWorkloadIdentity{
-								{
-									OperatorName: "openshift-azure-operator",
-									ObjectID:     "00000000-0000-0000-0000-000000000000",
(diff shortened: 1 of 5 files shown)


pkg/operator/const.go
```diff
@@ -9,4 +9,8 @@ const (
 
 	Namespace  = "openshift-azure-operator"
 	SecretName = "cluster"
+
+	OperatorIdentityName       = "ServiceOperator"
```
bitoku, line 13: Do we have a naming rule for the identity name?
`ServiceOperator` doesn't sound the identity for `aro-operator` for me.

pkg/operator/deploy/deploy.go
```diff
@@ -196,6 +196,33 @@ func (o *operator) resources(ctx context.Context) ([]kruntime.Object, error) {
 	}
 
 	// then dynamic resources
+	if o.oc.UsesWorkloadIdentity() {
+		var operatorIdentity *api.PlatformWorkloadIdentity // use a pointer to make it easy to check if we found an identity below
+		for _, i := range o.oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
+			if i.OperatorName == pkgoperator.OperatorIdentityName {
+				operatorIdentity = &i
```
bitoku, line 203: Should we break here?

```suggestion
				operatorIdentity = &i
				break
```

bitoku (commented):
Do we have to merge this first?
I think we want to wait for the WIMI implementation if we can.

pkg/operator/deploy/deploy.go
```diff
@@ -196,6 +196,33 @@ func (o *operator) resources(ctx context.Context) ([]kruntime.Object, error) {
...
+		results = append(results, &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: pkgoperator.Namespace,
+				Name:      pkgoperator.OperatorIdentitySecretName,
+			},
+			StringData: map[string]string{
+				"azure_client_id":            operatorIdentity.ClientID,
+				"azure_tenant_id":            o.env.TenantID(),
+				"azure_region":               o.oc.Location,
+				"azure_subscription_id":      o.env.SubscriptionID(),
```
kimorris27, line 220: If you get the tenant ID and sub ID from `o.env`, that data is coming from the RP's instance metadata and isn't correct for the platform managed identities. To get the correct values for the ARO operator's managed identity, you should parse the MI's resource ID using [`arm.ParseResourceID`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/arm#ParseResourceID).

pkg/cluster/arooperator_test.go
```diff
@@ -141,12 +141,60 @@ func TestInstallAROOperator(t *testing.T) {
...
+								ClientID:    "fake",
+								PrincipalID: "alsoFake",
+							},
+						},
+					},
+					Properties: api.OpenShiftClusterProperties{
+						PlatformWorkloadIdentityProfile: &api.PlatformWorkloadIdentityProfile{
+							PlatformWorkloadIdentities: []api.PlatformWorkloadIdentity{
+								{
+									OperatorName: "openshift-azure-operator",
```
kimorris27, line 191: This test case should fail since the operator name isn't correct (should be `"ServiceOperator"`)... but there's also not much value in me pointing this out given that the unit tests in this file mock the functionality of the operator deployer, which is what this PR actually changes. I think the changes to this file don't add value and can be removed.

With that said, I see why you didn't unit test in `pkg/operator/deploy` though - that `resources` function is tricky to write unit tests for.

To make it easier to unit test the functionality you added in, you could break everything within the `if o.oc.UsesWorkloadIdentity()` out into a helper function that returns `(&corev1.Secret, error)` and then do unit tests for the helper function.

kimorris27 (changes requested):
Mostly LGTM, but I pointed out one thing to fix and a way to improve the unit tests.

pkg/operator/deploy/deploy.go
cadenmarchese: The resource ID contains the subscription ID but not the tenant ID. For the tenant ID, do we need to pull that from the subscription doc, or does it make sense to use `o.oc.Identity.TenantID` (which is populated by MSI-RP)?

It seems like we should use the subscription doc.

kimorris27: Yeah, Alex and I discussed this in Slack and he brought it to my attention that the resource ID doesn't include the tenant ID. I recommended using the Azure SDK to `GET` the MI by resource ID and then pull the tenant ID from the response, but I guess using the subscription doc works too since we're assuming that all MIs are in the same tenant as the ARO cluster sub.

pkg/cluster/arooperator_test.go
yithian: Dang, I misunderstood how the install tests worked (clearly!)

I pulled the secret generation code into a helper function and wrote up a unit test for that like you suggested

pkg/operator/deploy/deploy.go
yithian: Thanks for the explanations! I've updated this to use the subscription doc instead of `o.env`

kimorris27: I should've mentioned this earlier, and maybe I'm being overly cautious, but I would still take the subscription ID from the MI's resource ID to be super safe. The MI might be in the same tenant but a different sub, and I don't see why we couldn't support that.

kimorris27 (changes requested):
I made one last suggestion re: the subscription ID value to use in the Secret, and there's still an outstanding lint error. LGTM otherwise!

pkg/operator/deploy/deploy.go
rajdeepc2792: Pointing more around how the workload identity client is created, the subscriptionID is not needed for creating the client. Ref:- https://azure.github.io/azure-workload-identity/docs/quick-start.html
SubscriptionID is provided as part of secret to point out the scope where those Managed Identities will be used and not where that MI reside, this is similar to the region.
So, I believe there isn't a need to take the subscriptionID from the MI and rather use the subscriptionID from the cluster doc.
PS- Trying to find better source to point that out.

rajdeepc2792: It is similar to the usage here:- https://github.com/openshift/machine-api-provider-azure/blob/main/pkg/cloud/azure/actuators/machine_scope.go#L454

kimorris27: Oh, I see. Thank you for pointing that out. Then I agree that the cluster doc is the best place to get the subscription ID from.

yithian: Thanks for the discussion!

Is the subscription doc is accessible from [an operator variable|https://github.com/Azure/ARO-RP/blob/a1c4de75a48b/pkg/operator/deploy/deploy.go#L65-L76]? I'm not seeing an obvious path, and and the operator variable is all this function currently has access to

If there isn't a chain of fields to use to get the subscription doc, would using [arm.ParseResourceID()|https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/arm#ParseResourceID] with `operator.oc.ID` be acceptable?

yithian:
I nabbed the linting error (thanks for catching that) and added a question about how to access the SubscriptionID if the subscription doc isn't easily available

pkg/operator/deploy/deploy.go
rajdeepc2792: Apologies, I don't get the question exactly.
IMO, the existing way in PR to pass the subscription doc and getting the subscriptionID is correctly done and there's no change needed there.
@kimorris27 please confirm the same.

kimorris27: >If there isn't a chain of fields to use to get the subscription doc, would using [arm.ParseResourceID()](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/arm#ParseResourceID) with operator.oc.ID be acceptable?

This would technically work too, but I agree with Rajdeep that the way you have it right now looks good. The work to add the subscription doc as a field on the `operator` type is already done.

yithian: Sorry, I meant cluster doc, not subscription doc in that comment

Regardless, it sounds like things are good now, though

kimorris27 (dismissed):

cadenmarchese (approved):
lgtm

pkg/operator/deploy/deploy_test.go
```diff
@@ -582,3 +583,66 @@ func TestTestEnsureUpgradeAnnotation(t *testing.T) {
...
+
+func TestGenerateOperatorIdentitySecret(t *testing.T) {
+	tests := []struct {
+		Name           string
+		Operator       *operator
+		ExpectedSecret *corev1.Secret
+		ExpectError    bool
+	}{
+		{
+			Name: "valid cluser operator",
```
SudoBrendan, line 595: nit: spelling on `cluster`

pkg/operator/const.go
```diff
@@ -9,4 +9,8 @@ const (
...
+	OperatorIdentitySecretName = "azure-cloud-credentials"
+	OperatorTokenFile          = "/var/run/secrets/openshift/serviceaccount/token"
```
SudoBrendan, line 15: I just happened to have reviewed https://github.com/Azure/ARO-RP/pull/3776 - is there any way this const could be used in both places? It seems like these two PRs are very much related - one, generating the tokens, the other, mounting them.

SudoBrendan (approved):
Ultimately this code LGTM, if we can figure out how to get consts in place for things like the token file between this PR and 3776, all the better!

Thanks for your hard work on this!

pkg/operator/deploy/deploy_test.go
yithian: Whoops, thanks!

pkg/operator/const.go
yithian: I'll rebase #3776 onto this PR and pass `OperatorTokenFile` in to the template

kimorris27 (approved):

cadenmarchese (approved):

PR #3764, opened 2024-08-08
Containerized Full RP Dev Automation

razo7:
The PR automates the [full RP dev int-like](https://github.com/Azure/ARO-RP/blob/master/docs/deploy-full-rp-service-in-dev.md#deploying-an-int-like-development-rp) procedure using dockerfile and Makefile. 

### Which issue this PR addresses:

Fixes ARO-9327

### Procedure Steps and Timing:
On average the full automation takes 65 minutes in _eastus_ and below are the procedure steps and timing (following the [design](https://docs.google.com/document/d/1JiakpGL3n4Y77dLRs95NV7qbHIF_HCmD312X0F2z5C8/edit#heading=h.fbts9k7ez097)):

0. Verify full RP succeeded ~ 3 seconds
1. Download shared secrets ~ 5 seconds
2. Setup dev-config.yaml ~ 30 seconds
3. Pre-deploy resources ~ 4 minutes
4. add_hive ~ 32 minutes (dev-vpn 20 minutes, aks-dev 12 minutes)
5. Install Hive ~1 minute
6. mirror_images 14 minutes
7. prepare_RP_deployment  ~1 minute
8. fully_deploy_resources ~ 13 minutes

### What are these bash scripts?

- `rp_dev_helper.sh` bash script has functions that any SRE could use regardless of the Full RP automation for checking and using Azure resources.
- `full_rp_funcs.sh` has functions that correspond to the designed steps of Full RP creation and are triggered with `full_rp_deploy.sh` inside your local container.

### What this PR does / why we need it:
The PR adds the capability of quickly and easily provisioning INT-like development environments without manual steps using a local containerized process.

How to run the ARO RP automation?

1. Login to Azure
2. Run `AZURE_PREFIX=aaa RP_LOCATION=eastus SKIP_DEPLOYMENTS=true make full-rp-dev`

Comments:

- We copy local Azure credentials to the container, so make sure they are at `${HOME}/.azure`.
You can run `az account show --query state -o tsv` to check if your login to Azure CLI is successful. Look for`Enabled` in the output.
- Use **AZURE_PREFIX** with unique characters (otherwise there could be a collision between other developers). See also https://github.com/Azure/ARO-RP/pull/3681#event-13792085488 for _AZURE_PREFIX_ motivation.
- Running the automation will be skipped in case the RP and GWY VMSSs succeed and their final deployment succeeds. 
- (optional) Set **RP_LOCATION** var to your preferred Azure location (by default it is `eastus`).
[...]

PR #3765, opened 2024-08-08
Refactor Hive Directory

razo7:

1. Group the Hive files from `hack` directory to `hack/hive` directory.
2. Add main function, and use `hack/utils.sh` functions
3. Hive installation was timed out using the `hack/hive-dev-install.sh` script, so we remove the redundant timeout waiting. 

> .
> .
> .
> clusterrolebinding.rbac.authorization.k8s.io/hive-operator-rolebinding created
> deployment.apps/hive-operator created
> deployment.apps/hive-operator condition met
> Error from server (NotFound): deployments.apps "hive-controllers" not found

- Simplify Hive scripts and creation and also use common functions. 

- Aligning with [production Hive installation](https://dev.azure.com/msazure/AzureRedHatOpenShift/_git/ARO-Pipelines/pullrequest/9992450) by not waiting for the _hive-controllers_ deployment to be **Available** and the pod to be **Ready**.  It's the responsibility of the operator to manage its resources, and it is an overkill that even causes the Hive installation to fail on waiting. Rerun the script was succeeded
- Fixing the Hive installation timeout issue would help #3764 automation.

### Test plan for issue:
Not needed

Yes, due to the change of files location.


hawkowl (changes requested):
I don't think we want to remove it entirely, but increase it to some amount that Hive will _surely_ be operational by. If we do want to just trust in the operator, then we should instead put in some output of the script that indicates how to tell when Hive is operational and you can use it (e.g. "check readiness via kubectl x y z to monitor rollout").

tsatam:
These waits/timeouts were intentionally added in order to solve for issues we were facing when deploying Hive into ARO production clusters, where the deployment itself was "healthy" but the Hive operator was degraded due to various issues (namely missing OpenShift-specific CRDs in the AKS clusters we run Hive in). 

Without these checks, the script would exit "successfully" and thus our deployment would be marked as successful, despite Hive being degraded. 

That being said, I don't necessarily think we need to retain _this_ specific solution to that problem, and there's probably a better long-term solution (e.g. release pipeline takes Hive health metrics into account). 

For now I think we could make this change if truly desired, since this script isn't itself used for production Hive deployments. I just backported the changes we made to the production deployment script here.

hack/hive/hive-dev-install.sh
```diff
@@ -0,0 +1,89 @@
...
+	else
+		$KUBECTL create namespace $HIVE_OPERATOR_NS
+	fi
+
+	log "Hive is ready to be installed"
+	$KUBECTL apply -f ./hack/hive/hive-config/crds
+	echo "$PULL_SECRET" > /tmp/.tmp-secret
+	# Using dry-run allows updates to work seamlessly
+	$KUBECTL create secret generic hive-global-pull-secret --from-file=.dockerconfigjson=/tmp/.tmp-secret --type=kubernetes.io/dockerconfigjson --namespace $HIVE_OPERATOR_NS -o yaml --dry-run=client | $KUBECTL apply -f - 2>/dev/null
+	rm -f /tmp/.tmp-secret
```
razo7, line 44: Not used in production, so it might be possible to remove it from here as well.

hack/hive/hive-config/generate.go
```diff
@@ -3,4 +3,4 @@ package hiveconfig
 // Copyright (c) Microsoft Corporation.
 // Licensed under the Apache License 2.0.
 
-//go:generate go run ../genhiveconfig ./hive-additional-install-log-regexes.yaml
+//go:generate go run ../../genhiveconfig ./hive-additional-install-log-regexes.yaml
```
razo7, line 6: Probably moving the `genhiveconfig` directory under `hack/hive/` would be a good idea for grouping Azure resources.

hack/hive/hive-config/generate.go
SudoBrendan, line 1: flagging this as potentially needing a rebase on master to pick up the .bingo changes. Dismiss if untrue (but it probably doesn't hurt to do it anyways).

SudoBrendan (changes requested):

hack/hive/hive-config/generate.go
razo7: I believe you are referring to #3719. **Done**

tiguelu:
> I don't think we want to remove it entirely, but increase it to some amount that Hive will _surely_ be operational by. If we do want to just trust in the operator, then we should instead put in some output of the script that indicates how to tell when Hive is operational and you can use it (e.g. "check readiness via kubectl x y z to monitor rollout").

Thank you Amber. A log line has been added to print commands to check readiness.

tiguelu (approved):
All requested changes have been addressed. LGTM.

PR #3767, opened 2024-08-12
Consume go otel audit lib

gouthamMN:

Fixes


go.mod
```diff
@@ -1603,3 +1611,6 @@ replace (
 	go.mongodb.org/mongo-driver => go.mongodb.org/mongo-driver v1.9.4
 	google.golang.org/grpc => google.golang.org/grpc v1.56.3
 )
+
+// go-otel-audit pins
```
stevekuznetsov, line 1615: Please explain why this is necessary - it's clear that the `replace` directive is pinning :)

pkg/frontend/fixetcd_test.go
```diff
@@ -515,7 +515,7 @@ func TestFixEtcd(t *testing.T) {
 			})
 
 			f, err := NewFrontend(ctx,
-				ti.audit,
+				ti.audit, ti.otelAudit,
```
stevekuznetsov, line 518: Likely best in these cases to be consistent with where the newlines are

pkg/frontend/middleware/log.go
```diff
@@ -141,6 +143,54 @@ func (l LogMiddleware) Log(h http.Handler) http.Handler {
 			},
 		})
 
+		auditMsg := msgs.Msg{Type: msgs.ControlPlane}
+		auditRec := audit.GetAuditRecord()
+
+		callerIpAddress, err := msgs.ParseAddr(strings.Split(r.RemoteAddr, ":")[0])
```
stevekuznetsov, line 149: Please handle `strings.Split` not finding a `:`

pkg/frontend/middleware/log.go
```diff
@@ -141,6 +143,54 @@ func (l LogMiddleware) Log(h http.Handler) http.Handler {
...
+		if err != nil {
+			log.Printf("Error parsing remote address: %s, error: %v", r.RemoteAddr, err)
```
stevekuznetsov, line 151: If this is not recoverable, you need to return or exit the process.

PR #3771, opened 2024-08-15
ARO-9570: Add a controller to the ARO operator to lay down etc hosts machine config

lranjbar:

Fixes ARO-9570

This PR adds a MachineConfigs for the ARO operator to lay down:

99-master-aro-etc-hosts-gateway-domains
99-worker-aro-etc-hosts-gateway-domains

During the UDR installation incident we have gone back and forth on how resolve the root issue. The root cause is that the gateway domains are not populated for OpenShift images to be pulled and installed. These domains are IPs so we can pass them into `/etc/hosts` to resolve them before other DNS resolution options are available.

### Test plan for issue:

I've some minimal smoke testing on the operator with this PR.

### DIFF
--- a/pkg/operator/controllers/etchosts/machineconfig_controller.go
+++ b/pkg/operator/controllers/etchosts/machineconfig_controller.go
@@ -1,0 +1,124 @@
+package etchosts
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"regexp"
+
+	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/sirupsen/logrus"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	kruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/Azure/ARO-RP/pkg/operator"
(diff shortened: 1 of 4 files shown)


pkg/operator/controllers/etchosts/machineconfig_controller.go
```diff
@@ -0,0 +1,124 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/Azure/ARO-RP/pkg/operator"
+	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+	"github.com/Azure/ARO-RP/pkg/operator/controllers/base"
+	"github.com/Azure/ARO-RP/pkg/util/dynamichelper"
+)
+
+const (
+	EtcHostsMachineConfigControllerName = "DnsmasqMachineConfig"
```
kimorris27, line 26: I'm not super up-to-speed with the full incident response plan, but I guess this name choice indicates that this controller is intended to completely replace the existing dnsmasq MachineConfig controller?

If so, does it need the additional controllers for watching the ARO cluster CR and the MCPs (like the existing controllers in `pkg/operator/controllers/dnsmasq`)?

kimorris27 (commented):
LGTM broadly speaking, but I asked a few questions about parts I'm uncertain about. It would also be nice to have E2E tests, though they could probably go in a separate PR if we're in a rush.

Also, I don't think we can run E2E unless you reopen your PR using a branch directly off of `Azure/ARO-RP`. I'll copy my review comments over to the new PR once you're opened it to keep things organized in one place.

SudoBrendan:
> No pipelines are associated with this pull request.

@lranjbar yea - kipp is right - can you move this to a branch under ARO-RP so we can run our pipelines? This is a result of security hardening... you can run `make init-contrib` to set up some local hooks on branch naming strategies, generally, `<GITHUB_USERNAME>/<JIRA_ID><other-optional-text>`.

PR #3772, opened 2024-08-16
fix e2e case

yjst2012:

Fixes https://issues.redhat.com/browse/ARO-9684

fix the problem discovered during 4.16 E2E tests on private clusters

### Test plan for issue:

E2E CI

NA

E2E pass

### DIFF
--- a/test/e2e/adminapi_redeployvm.go
+++ b/test/e2e/adminapi_redeployvm.go
@@ -41,6 +41,12 @@ var _ = Describe("[Admin API] VM redeploy action", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(vms).NotTo(BeEmpty())
 		vm := vms[0]
+		for _, possibleVM := range vms {
+			if strings.Contains(*possibleVM.Name, "-master-") || strings.Contains(*possibleVM.Name, "-worker-") {
+				vm = possibleVM
+				break
+			}
+		}
 		log.Infof("selected vm: %s", *vm.Name)
 
 		By("saving the current uptime")
@@ -72,7 +78,7 @@ var _ = Describe("[Admin API] VM redeploy action", func() {
 		By("getting system uptime again and making sure it is newer")
 		newUptime, err := getNodeUptime(Default, ctx, *vm.Name)
 		Expect(err).NotTo(HaveOccurred())
-		Expect(newUptime).To(BeTemporally(">", oldUptime))
+		Expect(newUptime).To(BeTemporally(">=", oldUptime))
 	})
 })
 
--- a/test/e2e/adminapi_serialconsole.go
+++ b/test/e2e/adminapi_serialconsole.go
@@ -34,6 +34,7 @@ var _ = Describe("[Admin API] VM serial console action", func() {
 		for _, possibleVM := range vms {
 			if strings.Contains(*possibleVM.Name, "-master-") {
 				vm = *possibleVM.Name
+				break
 			}
 		}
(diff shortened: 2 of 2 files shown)


test/e2e/adminapi_redeployvm.go
```diff
@@ -41,6 +41,12 @@ var _ = Describe("[Admin API] VM redeploy action", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(vms).NotTo(BeEmpty())
 		vm := vms[0]
+		for _, possibleVM := range vms {
+			if strings.Contains(*possibleVM.Name, "-master-") || strings.Contains(*possibleVM.Name, "-worker-") {
+				vm = possibleVM
+				break
+			}
+		}
```
bitoku, line 49: I think the normal cluster has only masters and workers.
In what case do we need this check?

Also I thought we should be able to redeploy any type of VM in the cluster.
Is it not expected to redeploy other than master or worker?

test/e2e/adminapi_redeployvm.go
```diff
@@ -72,7 +78,7 @@ var _ = Describe("[Admin API] VM redeploy action", func() {
 		By("getting system uptime again and making sure it is newer")
 		newUptime, err := getNodeUptime(Default, ctx, *vm.Name)
 		Expect(err).NotTo(HaveOccurred())
-		Expect(newUptime).To(BeTemporally(">", oldUptime))
+		Expect(newUptime).To(BeTemporally(">=", oldUptime))
```
bitoku, line 81: I think it should be ">" because we expect uptime will be different, don't we?

test/e2e/adminapi_redeployvm.go
yjst2012: well, during my test, I put a jumphost in the master subnet, which is probably not very common and it was then picked by the selection and redeployed, error msg:

```
  [FAILED] Timed out after 600.000s.
  The function passed to Eventually failed at /home/jeff/go/src/redhat/ARO-RP/test/e2e/helpers.go:45 with:
  Unexpected error:
      <*errors.StatusError | 0xc000b29cc0>: 
      pods "jeff-jumphost-1-uptime-1" not found
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "pods \"jeff-jumphost-1-uptime-1\" not found",
              Reason: "NotFound",
              Details: {
                  Name: "jeff-jumphost-1-uptime-1",
                  Group: "",
                  Kind: "pods",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 404,
          },
      }
  occurred

```

test/e2e/adminapi_redeployvm.go
yjst2012: hmm, ">" failed my test, but ">=" can pass with no issue, I didn't dig too much at that time as there are a lot of pending issues, and now the env has been cleaned up 🤷

kimorris27 (commented):
I see the idea behind this change, but I don't think it's logically correct or necessary. Under normal conditions, the deny assignment will prevent cxs from creating jump hosts or other VMs under the managed/cluster RG.

I'd say that this issue was simply an unfortunate interaction between the way you chose to test and the logic of this particular E2E test. I won't push back if others want to approve and merge this though, since I don't see any harm being done.

PR #3775, opened 2024-08-16
Makefile Podman Refactors

SudoBrendan:

Adds flexibility and correctness to Makefile's `podman` calls. Each commit is a unique fix.

- Strictly calls `podman` instead of `docker` to prevent confusion for local cases
- `NO_CACHE=true` actually works now for ci-rp, and we still correctly tag our builder image layers while only building once
- Standardizes image names and tags + the Makefile variable names used to define these things
- Sets up future changes for "remote" podman sockets (see [run podman](https://github.com/Azure/ARO-RP/blob/a387947960ba60f83ee9473d6c38b6c5f30b4076/hack/e2e/run-rp-and-e2e.sh#L107) and [run selenium](https://github.com/Azure/ARO-RP/blob/a387947960ba60f83ee9473d6c38b6c5f30b4076/hack/e2e/run-rp-and-e2e.sh#L145-L146) in our E2E script to see where this is going - where `make ci-rp` and/or `make runlocal-rp` can execute without extra shell scripts like these do today by setting `PODMAN_REMOTE_ARGS="-r --url tcp://localhost:8888"` and everything "just works" the same as local)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,23 +67,23 @@ build-all:
 aro: check-release generate
 	go build -ldflags "-X github.com/Azure/ARO-RP/pkg/util/version.GitCommit=$(VERSION)" ./cmd/aro
 
-# Target to create docker secrets
-.PHONY: docker-secrets
-docker-secrets: aks.kubeconfig
-	docker secret rm --ignore aks.kubeconfig
-	docker secret create aks.kubeconfig ./aks.kubeconfig
+# Target to create podman secrets
+.PHONY: podman-secrets
+podman-secrets: aks.kubeconfig
+	podman secret rm --ignore aks.kubeconfig
+	podman secret create aks.kubeconfig ./aks.kubeconfig
 
-	docker secret rm --ignore proxy-client.key
-	docker secret create proxy-client.key ./secrets/proxy-client.key
+	podman secret rm --ignore proxy-client.key
+	podman secret create proxy-client.key ./secrets/proxy-client.key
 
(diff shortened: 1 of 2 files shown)


shubhadapaithankar (approved):
I can confirm that this PR works perfectly on macOS without any issues.
All changes looks good to me .

Makefile
```diff
@@ -42,9 +43,13 @@ else
...
 endif
 
-RP_IMAGE_LOCAL ?= aro
-ARO_PORTAL_BUILD_IMAGE ?= $(RP_IMAGE_LOCAL)-portal-build
-ARO_BUILDER_IMAGE ?= $(RP_IMAGE_LOCAL)-builder
+# local images
+LOCAL_ARO_RP_IMAGE ?= aro
+LOCAL_ARO_PORTAL_BUILD_IMAGE ?= $(LOCAL_ARO_RP_IMAGE)-portal-build
+LOCAL_ARO_RP_BUILD_IMAGE ?= $(LOCAL_ARO_RP_IMAGE)-build
+LOCAL_AZ_EXT_ARO_IMAGE ?= azext-aro
```
SudoBrendan, line 50: @shubhadapaithankar I think this has implications on CI, so we will need to resolve what envvars we use and what order these PRs merge in.

tsatam (approved):

cadenmarchese (approved):

shubhadapaithankar (approved):

PR #3776, opened 2024-08-16
ARO-4639 update the operator master deployment to support workload identity

yithian:

Fixes ARO-4639

This causes the spec for the operator master deployment to mount the service account token as a volume, and maps the path to the environment variable expected by Azure to support workload identities

### Test plan for issue:

Unit tests confirm that the generated yaml is valid and will be applied to a test cluster to confirm that it's a valid OpenShift resource as well

No, part of larger project of adding support for MIWI

As with all MIWI support work, it will be extremely challenging to test until we have a way to provision a MIWI cluster. Applying the updated generated resource to a test cluster will probably be the best we can do until then


kimorris27 (commented):
LGTM as long as CI and E2E pass!

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
```diff
@@ -79,4 +90,14 @@ spec:
...
+      {{ if .UsesWorkloadIdentity }}
+      volumes:
+        - name: bound-sa-token
+          projected:
+          defaultMode: 420
+          sources:
+          - serviceAccountToken:
+              audience: openshift
+              expirationSeconds: 3600
+              path: token
```
SudoBrendan, line 102: YAML indentation error? we want all of this under the `projected` block, yea? ref https://kubernetes.io/docs/concepts/storage/projected-volumes/#example-configuration-secret-downwardapi-configmap

```suggestion
            defaultMode: 420
            sources:
            - serviceAccountToken:
                audience: openshift
                expirationSeconds: 3600
                path: token
```

SudoBrendan (changes requested):
I think there's a YAML syntax error on whitespace - dismiss if this is incorrect :)

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
SudoBrendan: I'm actually confused as to how CI/E2E passed with this - maybe I am wrong...

yithian: Maybe the conversion to json noticed the `:` after `projectd` and Just Did The Right Thing :tm:?

Either way, I've adjusted the indentation, even if only to make it more immediately apparent what's going on :laughing:

SudoBrendan (approved):

yithian:
Putting a hold on this until #3761 is merged, as this was rebased onto that to pull in the token file location/mountpoint

yithian:
#3761 is merged, removing my hold

pkg/operator/deploy/deploy.go
```diff
@@ -160,12 +162,19 @@ func (o *operator) createDeploymentData(ctx context.Context) (deploymentData, er
...
 		IsLocalDevelopment:           o.env.IsLocalDevelopmentMode(),
 		Image:                        image,
 		SupportsPodSecurityAdmission: usePodSecurityAdmission,
 		Version:                      version,
-	}, nil
+	}
+
+	if o.oc.UsesWorkloadIdentity() {
+		data.UsesWorkloadIdentity = o.oc.UsesWorkloadIdentity()
+		data.TokenVolumeMountPath = pkgoperator.OperatorTokenFile
```
bitoku, line 174: If we want to place the token at `OperatorTokenFile` (= `/var/run/secrets/openshift/serviceaccount/token`), the mount path should be the directory of `OperatorTokenFile`.

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
```diff
@@ -79,4 +90,14 @@ spec:
...
         operator: Exists
-
+      {{ if .UsesWorkloadIdentity }}
+      volumes:
+      - name: bound-sa-token
+        projected:
+          defaultMode: 420
+          sources:
+          - serviceAccountToken:
+              audience: openshift
```
bitoku, line 100: This is just a question, but is this required?
Apparently this field is optional, so I'm curious why we need it.

pkg/operator/deploy/deploy_test.go
```diff
@@ -646,3 +681,50 @@ func TestGenerateOperatorIdentitySecret(t *testing.T) {
 		})
 	}
 }
+
+func TestTemplateManifests(t *testing.T) {
+	tests := []struct {
+		Name           string
+		DeploymentData deploymentData
+		ExpectError    bool
```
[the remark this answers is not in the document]
yithian: Removed!

pkg/operator/deploy/deploy.go
yithian: Thanks for catching that :facepalm: 

Fixed

pkg/operator/deploy/staticresources/master/deployment.yaml.tmpl
yithian: My understanding is that the value of the `audience` field needs to match what's in the token itself, otherwise it will be rejected

image

@cadenmarchese is this accurate? Do we need the `audience` field?

cadenmarchese: According to @bennerv it is required. Linking the thread with explanation: https://redhat-external.slack.com/archives/C06FRQKUMFF/p1723242231734469

> OpenShift hardcodes all it's operators & deployments to mount a service account token for the audience: openshift instead of using azure.workload.identity/use: true label on the pod spec.
This is because the cluster version operator doesn't know (or doesn't care) if you're using workload identity when it lays down the manifests for the operators.
So the audience on the federated credentials for these identities is openshift.  if you end up using the azure.workload.identity/use label on the ARO operator for workload identity enabled clusters, you'll have a different audience in the federated credentials.  It'll be api://AzureADTokenExchange

cadenmarchese (approved):
makes sense to me!

rajdeepc2792 (approved):

kimorris27 (approved):
The only E2E failure is a known flake.

PR #3778, opened 2024-08-19
ARO-9263: Add ACR token issue date and check validity

edisonLcardenas:

Jira https://issues.redhat.com/browse/ARO-8472
       https://issues.redhat.com/browse/ARO-9263

- Add issue date in registry profile for ACR Token. 
- Check the issue date if still valid or needs to be rotated.

### Test plan for issue:

Unit tests. 
Manual validation.

Part of MIMO M1

Unit tests.


pkg/util/mocks/env/core.go
SudoBrendan, line 1: I'm unsure - but this PR may need a rebase on master to get the .bingo changes. Dismiss if untrue (but it likely doesn't hurt to do it anyways).

SudoBrendan (dismissed):

pkg/mimo/tasks/cluster/acrtoken_checker.go
```diff
@@ -0,0 +1,65 @@
...
+		case daysInterval >= daysShouldRotate:
+			return mimo.TerminalError(fmt.Errorf("%d days have passed since azure container registry (acr) token was issued, please rotate the token now", daysInterval))
+		default:
+			th.SetResultMessage("azure container registry (acr) token is valid")
+		}
+	} else {
+		return mimo.TerminalError(errors.New("no registry profile detected"))
+	}
+
+	return nil
```
jaitaiwan, line 64: Makes no sense to have `return nil` at the bottom. It's unreachable code. I would remove the else block entirely and move the return from the else block to the bottom of the function. You might be able to reorder the if statements so you return earlier and don't need nested code branching.

pkg/mimo/tasks/cluster/acrtoken_checker.go
```diff
@@ -0,0 +1,65 @@
...
+
+	registryProfiles := th.GetOpenShiftClusterProperties().RegistryProfiles
+	rp := manager.GetRegistryProfileFromSlice(registryProfiles)
+	if rp != nil {
+		var now = time.Now().UTC()
+		issueDate := rp.IssueDate
+
+		if issueDate == nil {
+			return mimo.TerminalError(errors.New("no issue date detected"))
+		}
```
jaitaiwan, line 48: If we don't have an issue date we've probably got a token that will never expire. Shouldn't we remove it anyway?

pkg/mimo/tasks/cluster/acrtoken_checker.go
```diff
@@ -0,0 +1,65 @@
...
+
+		daysInterval := int32(now.Sub(issueDate.Time).Hours() / 24)
```
jaitaiwan, line 50: I feel like we should just be making use of golang's inbuilt duration funcs for this. Something like this:

```
now := time.Now().UTC()
lastValidIssue := now.Sub(daysShouldRotate * time.Day)
lastValidRotation := now.Sub(daysValid * time.Day)

if issueDate.Time.Before(lastValidIssue) {
  return mimo.TerminalError(fmt.Errorf("%d days have passed since azure container registry (acr) token was issued, please rotate the token now", daysInterval))
}

if issueDate.Time.Before(lastValidRotation) {
  return mimo.TerminalError(fmt.Errorf("azure container registry (acr) token is not valid, %d days have passed", daysInterval))
}
```

pkg/mimo/tasks/cluster/acrtoken_checker.go
```diff
@@ -0,0 +1,65 @@
...
+	manager, err := acrtoken.NewManager(env, localFpAuthorizer)
+	if err != nil {
+		return err
+	}
+
+	registryProfiles := th.GetOpenShiftClusterProperties().RegistryProfiles
+	rp := manager.GetRegistryProfileFromSlice(registryProfiles)
+	if rp != nil {
+		var now = time.Now().UTC()
+		issueDate := rp.IssueDate
```
jaitaiwan, line 44: Unless I'm missing something there's no IssueDate on the Profile.

jaitaiwan (approved):
A couple things to improve and a couple of nits. Overall good work!

PR #3781, opened 2024-08-20
ARO-9570: Add a controller to the ARO operator to lay down etchosts

lranjbar:

Fixes ARO-9570

This PR adds a MachineConfigs for the ARO operator to lay down:

99-master-aro-etc-hosts-gateway-domains
99-worker-aro-etc-hosts-gateway-domains

During the UDR installation incident we have gone back and forth on how resolve the root issue. The root cause is that the gateway domains are not populated for OpenShift images to be pulled and installed. These domains are IPs so we can pass them into `/etc/hosts` to resolve them before other DNS resolution options are available.

### Test plan for issue:

I've some minimal smoke testing on the operator with this PR.

### DIFF
--- a/pkg/operator/controllers/etchosts/cluster_controller.go
+++ b/pkg/operator/controllers/etchosts/cluster_controller.go
@@ -1,0 +1,131 @@
+package etchosts
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+
+	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/sirupsen/logrus"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/Azure/ARO-RP/pkg/operator"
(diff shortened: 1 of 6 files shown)


bitoku:
we have to change here as well to add this operator to aro-operator, don't we?

https://github.com/Azure/ARO-RP/blob/1b6e0e758436f53c2d75c4ae2e1c27223bf8deaf/cmd/aro/operator.go#L96-L101

pkg/operator/controllers/etchosts/cluster_controller.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	r.Log.Debug("running")
+
+	mc := &mcv1.MachineConfig{}
+	mcp := &mcv1.MachineConfigPool{}
+
+	// If 99-master-aro-etc-hosts-gateway-domains doesn't exist, create it
+	err = r.Client.Get(ctx, types.NamespacedName{Namespace: "openshift-machine-api", Name: "99-master-aro-etc-hosts-gateway-domains"}, mc)
+	if kerrors.IsNotFound(err) {
```
cadenmarchese, line 70: Regarding the error handling here. It looks like we will drop any error from line 69 that isn't a `kerrors.IsNotFound` error. Should we make changes to handle other errors on the `Get`?

Here's an example of where we use IsNotFound, but then ensure to catch any other errors too:
https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/autosizednodes/autosizednodes_controller.go#L79-L91

pkg/operator/controllers/etchosts/cluster_controller.go
```diff
@@ -0,0 +1,131 @@
...
+		if err != nil {
+			r.Log.Error(err)
+			r.SetDegraded(ctx, err)
+			return reconcile.Result{}, err
+		}
+	}
+
+	// If 99-worker-aro-etc-hosts-gateway-domains doesn't exist, create it
+	err = r.Client.Get(ctx, types.NamespacedName{Namespace: "openshift-machine-api", Name: "99-worker-aro-etc-hosts-gateway-domains"}, mc)
+	if kerrors.IsNotFound(err) {
```
cadenmarchese, line 91: Same comment as above.

mociarain (approved):

cadenmarchese (approved):
My previous comments aren't blocking, so I'll go ahead and approve. thanks!

PR #3782, opened 2024-08-20
Allow prerelease OCP versions and version strings with metadata during cluster installation 

tsatam:

Part of [ARO-7404](https://issues.redhat.com/browse/ARO-7404)

- Updates our cluster version validation to allow versions with prerelease suffixes (e.g. `4.14.0-0.nightly-2024-01-01-000000`), as well as versions with a metadata suffix (e.g. `4.14.0+installerref-123456`). This is required for environments such as CI that will be validating nightly and ad-hoc builds of OpenShift release artifacts and the ARO installer. 
- Updates our `disableUpdates` step run during installation to not propagate any metadata suffixes (as outlined above) into the ClusterVersion resource. 

### Test plan for issue:

- [x] Unit tests were added to ensure the new behavior
- [x] Change has been manually tested in a development RP environment with prerelease and metadata versions explicitly added, and cluster installation is successful

No

This change should not impact production, as we would need to explictly add a prerelease version or version string containing metadata to the production RP environment in order to utilize this functionality. The validation in place still ensures that a requested version is actually available before proceeding.


SudoBrendan (approved):

cadenmarchese (approved):

shubhadapaithankar (approved):
LGTM

PR #3786, opened 2024-08-21
ARO-9382 prevent updating existing platform identities

yithian:

Fixes ARO-9382

This adds a check to v20240812preview static validation that raises an error if either the name or resource ID of an existing platform identity

### Test plan for issue:

Unit tests

This is part of the general managed identity / workload identity feature development and this change should be reflected in the final documentation

Unit tests for now


pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -446,6 +446,18 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message)
 	}
 
+	if current.UsesWorkloadIdentity() {
+		for i := range current.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
```
tsatam, line 450: If I'm understanding this correctly, this code will reject an update that changes the order of platform workload identities in the list but otherwise keeps the identity order the same. 

I think we might want to permit such modifications - while the [CLI implementation](https://github.com/Azure/ARO-RP/blob/880835249791f94060b467e5ab58467a8deaecf0/python/az/aro/azext_aro/custom.py#L482) should preserve the existing order, users using ARM/Bicep or Terraform might not guarantee the order of identities here, and as far as I am aware the order of this list should not matter.

tsatam (changes requested):
LGTM, just one suggestion to support updates that change the order of platform workload identities in the array but otherwise preserve the same name -> resource ID mappings.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
yithian: It's slightly concerning that the order of an ordered list isn't reliable but I've updated the logic to look for a matching `OperatorName` or `ResourceID` in the update data and if found, see if the identity's `ResourceID` or `OperatorName` (respectively) is different

cadenmarchese: I agree with Tanmay, there could be situations where an ARM call has the identities in a different order than the current order in the cluster doc, which would still be valid. thanks @yithian for making it order-agnostic.

kimorris27: I don't think the fix you pushed will account for a situation where a cx tries to change the `ResourceID` and the `OperatorName` in a single API call. Here's my idea for how to implement the validation to properly account for this situation:

1. Construct two maps where the keys are operator names and the values are PWI structs - one map for the old cluster doc and one map for the new cluster doc
2. For all keys in new doc that are also in old doc, ensure that all four fields in the structs are equal

I _think_ that would do it, but additional unit test cases would verify for us.

kimorris27 (changes requested):
I recommended a change to the validation to make sure we account for all possible scenarios.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -446,6 +446,29 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
...
 
+	if current.UsesWorkloadIdentity() {
+		currentIdentities := map[string]PlatformWorkloadIdentity{}
+		for _, i := range current.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
+			currentIdentities[i.OperatorName] = i
+		}
+
+		updateIdentities := map[string]PlatformWorkloadIdentity{}
+		for _, i := range oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
+			updateIdentities[i.OperatorName] = i
```
rajdeepc2792, line 457: I know we handle the duplicate PlatformWorkloadIdentities at CLI side, if we want to handle the duplication error in the provided list of identities, this would be the best place to handle it.
@tsatam @cadenmarchese , do you agree handling it here?

cadenmarchese: +1 on handling this here (checking for duplicates in OperatorName). I've added it to the JIRA.

rajdeepc2792: I think this still needs handling here:-

```
_ , duplicate := updateIdentities[i.OperatorName]
if duplicate {
				return api.NewCloudError(http.StatusBadRequest, api.[SOME-CLOUD-ERROR-FOR-NO-DUPLICATE-OPERATOR-IDENTITIES], "properties.platformWorkloadIdentityProfile.platformWorkloadIdentities", "<SOME MESSAGE FOR DUPLICATE OPERATOR IDENTITIES>")
			}
```

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -455,6 +479,9 @@ func (sv openShiftClusterStaticValidator) validatePlatformWorkloadIdentityProfil
 		return nil
 	}
 
+	// collect operator names to check for duplicates
+	operatorNames := []string{}
```
rajdeepc2792, line 483: Can we handle the duplication check using a map?

yithian: I suppose so? Would that accomplish something different?

rajdeepc2792: No, nothing from functionality pov, but I believe from complexity pov.

rajdeepc2792 (changes requested):
LGTM from functional and unit tests pov, added a comment to use a different data structure for duplication check.

rajdeepc2792 (approved):
LGTM!!

slawande2 (commented):
Just seeking clarification for my understanding- why is UsesWorkloadIdentity moved from pkg/api/openshiftcluster.go to pkg/api/v20240812preview/openshiftcluster.go ?

kimorris27 (approved):

tsatam:
> Just seeking clarification for my understanding- why is UsesWorkloadIdentity moved from pkg/api/openshiftcluster.go to pkg/api/v20240812preview/openshiftcluster.go ?

It was copied, not moved. We maintain separate OpenShiftCluster struct definitions for each individual APIversion we support, so we'd need to maintain duplicate functions for each struct as well.

tsatam (approved):

PR #3793, opened 2024-08-23
fix: Ensure RP run with local version tag

jaitaiwan:
Fixes 3792


jaitaiwan:
Closing out as Shubada's PR was used instead

PR #3797, opened 2024-08-27
upgrade go version to 1.22

gouthamMN:

Fixes: https://issues.redhat.com/browse/ARO-9866

 This PR upgrades go version from `1.21` to `1.22` and pulls the Golang container image `mcr.microsoft.com/oss/go/microsoft/golang:1.22.6-fips-cbl-mariner2.0` from Microsoft container registry.

### Test plan for issue:
1. Ensure E2E is success.
2. Build and Deploy ARO-RP image with new go version in INT.

No

- Testing successfully in INT


Makefile
```diff
@@ -20,7 +20,7 @@ GATEKEEPER_VERSION = v3.15.1
 GOTESTSUM = gotest.tools/gotestsum@v1.11.0
 
 # Golang version go mod tidy compatibility
-GOLANG_VERSION ?= 1.21
```
stevekuznetsov, line 23: If we can, let's remove this? Recent Go versions are explicit about the version and toolchain in `go.mod` itself.

.golangci.yml
```diff
@@ -23,6 +23,15 @@ issues:
       text: "SA1019: codec.BasicHandle is deprecated"
     # This is set to false to disregard the default EXC0011 exclusion
     # which shadows all the ST* checks. Refer https://github.com/golangci/golangci-lint/issues/2281
+    - linters:
+        - govet
+      text: "printf: non-constant format string"
+    - linters:
+        - staticcheck
```
gouthamMN, line 30: These lint errors are showing up due to the new `golangci-lint version 1.60.3` upgrade. We can fix these in the future.

Makefile
```diff
@@ -66,7 +63,7 @@ endif
 
 .PHONY: build-all
 build-all:
-	go build ./...
+	go build -buildvcs=false ./...
```
gouthamMN, line 66: setting buildvcs to false, as this was resulting in [ci-go/ci-from-docker](https://github.com/Azure/ARO-RP/actions/runs/10617514593/job/29430527499?pr=3797) check failure with error `error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.`

Makefile
gouthamMN: addressed

.golangci.yml
stevekuznetsov: In my previous experience, whoever's doing the bump does the work to munge these as they come up.

gouthamMN: Addressed

pkg/api/admin/openshiftcluster_validatestatic.go
```diff
@@ -26,7 +26,7 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
 	err := immutable.Validate("", oc, current)
 	if err != nil {
 		err := err.(*immutable.ValidationError)
-		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message)
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message) //nolint:govet
```
stevekuznetsov, line 29: Could you please explain what the `govet` error is here and why it's not (easily) fixable?

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
```diff
@@ -39,7 +39,7 @@ func (sv platformWorkloadIdentityRoleSetStaticValidator) validate(new *PlatformW
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "properties.openShiftVersion", "Must be provided")
 	}
 
-	if new.Properties.PlatformWorkloadIdentityRoles == nil || len(new.Properties.PlatformWorkloadIdentityRoles) == 0 {
+	if len(new.Properties.PlatformWorkloadIdentityRoles) == 0 || new.Properties.PlatformWorkloadIdentityRoles == nil {
```
stevekuznetsov, line 42: Perhaps omit one of these two checks - as far as I can tell there's no possible way with `[]struct{}` to distinguish between the zero-value being `nil` and the zero-value being an empty slice.

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
```diff
@@ -74,7 +74,7 @@ func (sv platformWorkloadIdentityRoleSetStaticValidator) validateDelta(new, curr
 	err := immutable.Validate("", new, current)
 	if err != nil {
 		err := err.(*immutable.ValidationError)
-		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message)
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message) //nolint:govet
```
stevekuznetsov, line 77: Please try to explain why these had to be ignored.

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
```diff
@@ -446,13 +446,13 @@ func (e *etcdrenew) isEtcdRevised(ctx context.Context) (bool, error) {
 	isAtRevision := true
 	rawEtcd, err := e.k.KubeGet(ctx, "etcd.operator.openshift.io", "", "cluster")
 	if err != nil {
-		e.log.Warnf(err.Error())
+		e.log.Warnf("%s", err.Error())
```
stevekuznetsov, line 449: Instead of doing the no-op formatting, prefer

```go
e.log.Warn(err.Error())
```

pkg/operator/controllers/subnets/subnets_controller.go
```diff
@@ -142,7 +142,7 @@ func (r *reconcileManager) reconcileSubnets(ctx context.Context) error {
 	}
 
 	if len(combinedErrors) > 0 {
-		return fmt.Errorf(strings.Join(combinedErrors, "\n"))
+		return fmt.Errorf("%s", strings.Join(combinedErrors, "\n"))
```
stevekuznetsov, line 145: Instead of doing a no-op format, use

```go
return errors.New(strings.Join(combinedErrors, "\n"))
```

Or use a utility for error aggregation - the k8s libraries have some, etc.

pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
```diff
@@ -775,7 +775,7 @@ func TestEnsureGlobalPullSecret(t *testing.T) {
 			utilerror.AssertErrorMessage(t, err, tt.wantError)
 
 			if !reflect.DeepEqual(s, tt.wantSecret) {
-				t.Fatalf(cmp.Diff(s, tt.wantSecret))
+				t.Fatalf("%s", cmp.Diff(s, tt.wantSecret))
```
stevekuznetsov, line 778: In all cases where the format functions had no format strings, simply use the non-formatted version. Here, for instance:

```go
t.Fatalf(cmp.Diff(s, tt.wantSecret))
```

Or, of course, actually add some context:

```go
t.Fatalf("invalid secret: %s", cmp.Diff(s, tt.wantSecret))
```

pkg/validate/dynamic/dynamic.go
```diff
@@ -49,7 +48,10 @@ var (
 	errMsgInvalidVNetLocation               = "The vnet location '%s' must match the cluster location '%s'."
 )
 
-const minimumSubnetMaskSize int = 27
+const (
+	minimumSubnetMaskSize        int = 27
+	errMsgNSGNotProperlyAttached     = "When the enable-preconfigured-nsg option is specified, both the master and worker subnets should have network security groups (NSG) attached to them before starting the cluster installation."
```
stevekuznetsov, line 53: I'm not sure I understand this diff - could you instead just change the entire `var` block above to be a `const` block? Does code ever re-assign any of those strings?

pkg/api/admin/openshiftcluster_validatestatic.go
gouthamMN: govet error: `::printf: non-constant format string in call to github.com/Azure/ARO-RP/pkg/api.NewCloudError (govet)` the `NewCloudError()` constructor uses `fmt.Sprintf()` and the `err.Message` field passed is of non-constant format string. I'm open to change them to something like `api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, "Error Message: %s", err.Message)` if that's better.

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
gouthamMN: Explained [here](https://github.com/Azure/ARO-RP/pull/3797#discussion_r1739404938)

pkg/validate/dynamic/dynamic.go
gouthamMN: I think all other fields under `var` block has `'%s'` except `errMsgNSGNotProperlyAttached`. So, making those fields const could result error/issues during string formatting.

pkg/api/admin/openshiftcluster_validatestatic.go
stevekuznetsov: Hm. The constructor has the signature:

```go
func NewCloudError(statusCode int, code, target, message string, a ...interface{}) *CloudError {}
```

It then goes on to do `fmt.Sprintf(message, a...)` - judging by the amount of `//nolint:govet` you had to add, mis-using this constructor seems really common. I agree that using it correctly by passing some static format string would be workable, but you may also investigate how often it's used "correctly" (where someone _actually_ passes in a format string and args). If it's not common, you might save time by changing the signature to 

```go
func NewCloudError(statusCode int, code, target, message string) *CloudError {}
```

Let the caller do the formatting as necessary and simply inline the message in the constructor.

pkg/validate/dynamic/dynamic.go
stevekuznetsov: What errors specifically are you seeing? This might be a case of a linter adding rules that don't have utility. Constant strings with format specifiers are the preference.

pkg/api/admin/platformworkloadidentityroleset_validatestatic.go
gouthamMN: addressed

pkg/frontend/admin_openshiftcluster_etcdcertificaterenew.go
gouthamMN: Addressed

pkg/operator/controllers/subnets/subnets_controller.go
gouthamMN: Addressed

pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
gouthamMN: Addressed

pkg/api/admin/openshiftcluster_validatestatic.go
gouthamMN: There're are around 598 instances of these (correct usage).

pkg/validate/dynamic/dynamic.go
gouthamMN: My bad, it's doesn't throw any error :) fixed it.

niontive (approved):

pkg/api/admin/openshiftcluster_validatestatic.go
```diff
@@ -26,7 +26,7 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
 	err := immutable.Validate("", oc, current)
 	if err != nil {
 		err := err.(*immutable.ValidationError)
-		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message)
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message) // nolint: govet
```
stevekuznetsov, line 29: Missed one?

gouthamMN: updated

stevekuznetsov (approved):

PR #3802, opened 2024-08-28
MI/WI - Generate secrets for platform identities

tsatam:

Fixes [ARO-9924](https://issues.redhat.com/browse/ARO-9924)

- Adds a `SecretLocation` coordinate to `PlatformWorkloadIdentityRole`, that holds a NamespacedName reference to the default Secret location for each platform workload identity. These coordinates are intended for use internally, and will not be exposed on the public ARO API. 
- Adds a function `generatePlatformWorkloadIdentitySecrets` to the cluster manager for eventual use during ARO cluster installation (to generate secrets to pass to the installer) and during ARO updates (to directly update secrets on-cluster). Note that the current implementation as-is is unused anywhere within the codebase, and will be used in eventual work items [ARO-4518 (create)](https://issues.redhat.com/browse/ARO-4518) and [ARO-4371 (update)](https://issues.redhat.com/browse/ARO-4371). 

### Test plan for issue:

- [x] Unit tests were added for this direct functionality and they pass

No

This implementation as-is goes unused in any production contexts. It will eventually be utilized in MIWI cluster installation and update flows (see above).


pkg/api/admin/platformworkloadidentityroleset.go
```diff
@@ -43,4 +43,13 @@ type PlatformWorkloadIdentityRole struct {
...
 	ServiceAccounts []string `json:"serviceAccounts,omitempty" mutable:"true" validate:"required"`
+
+	// SecretLocation represents the location of the in-cluster secret containing credentials for the platform workload identity.
+	SecretLocation SecretLocation `json:"secretLocation,omitempty"`
+}
+
+// SecretLocation represents the location of the in-cluster secret containing credentials for the platform workload identity.
+type SecretLocation struct {
+	Namespace string `json:"namespace,omitempty"`
+	Name      string `json:"name,omitempty"`
```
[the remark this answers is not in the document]
tsatam: I'll add them, I don't think it would harm anything to include them as we expect these fields to always be present for every identity, and we should allow ourselves the ability to change them in the future just in case.

rajdeepc2792:
LGTM, just a comment on additional manifests needed.
There are 2 more manifests:-
`
apiVersion: config.openshift.io/v1
kind: Authentication
metadata:
  name: cluster
spec:
  serviceAccountIssuer: <OIDC ISSUER URL>
---
apiVersion: v1
stringData:
  azure_tenant_id: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
kind: Secret
metadata:
  name: azure-credentials
  namespace: openshift-cloud-credential-operator
type: Opaque
`

Can/Should we handle these two at the RP side and in the same PR?

tsatam:
> LGTM, just a comment on additional manifests needed. There are 2 more manifests:-
> 
> ```
> apiVersion: config.openshift.io/v1
> kind: Authentication
> metadata:
>   name: cluster
> spec:
>   serviceAccountIssuer: <OIDC ISSUER URL>
> ---
> apiVersion: v1
> stringData:
>   azure_tenant_id: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
> kind: Secret
> metadata:
>   name: azure-credentials
>   namespace: openshift-cloud-credential-operator
> type: Opaque
> ```
> 
> Can/Should we handle these two at the RP side and in the same PR?

For the Authentication resource, if it's a static, single resource as it appears to be, I think it can be done with a similar pattern to what's done here but it doesn't need to happen directly in this PR. Do we need to update this resource during cluster update as well (can the OIDC issuer URL change)? 

For the CCO secret - I thought `cloud-credential-operator` would be included as a required platform workload identity and therefore have an entry in the roleset, meaning this PR would generate that secret with coordinates to a dedicated identity, is that not the case? Running `oc adm release extract --from=${RELEASE_IMAGE} --credentials-requests --to=./scratch/credentials-requests --cloud azure` gives me a CredentialsRequest resource for CCO so I assumed it'd get a dedicated identity.

rajdeepc2792:
> Do we need to update this resource during cluster update as well (can the OIDC issuer URL change)?
No OIDC issuer URL update is not in scope currently, but the ensure function for the same can help here?

> For the CCO secret - I thought `cloud-credential-operator` would be included as a required platform workload identity and therefore have an entry in the roleset, meaning this PR would generate that secret with coordinates to a dedicated identity, is that not the case? Running `oc adm release extract --from=${RELEASE_IMAGE} --credentials-requests --cloud azure` gives me a CredentialsRequest resource for CCO so I assumed it'd get a dedicated identity.
Tried executing `oc adm release extract --cloud=azure --credentials-requests quay.io/openshift-release-dev/ocp-release:4.14.0-rc.6-x86_64` and it doesn't return the credential requests for CCO. The secret for CCO will not be part of the dedicated required Identities and secret will need to be created explicitly, as CCO doesn't require to access Azure resources, but it uses the TenantID for some of it's operations, probably related to setting up the webhook.

pkg/cluster/workloadidentityresources.go
```diff
@@ -0,0 +1,109 @@
...
+					"azure_federated_token_file": azureFederatedTokenFileLocation,
+				},
+			})
+		}
+	}
+
+	return secrets, nil
+}
+
+func (m *manager) generateCloudCredentialOperatorSecret() (*corev1.Secret, error) {
```
rajdeepc2792, line 80: What's the idea of using `UsesWorkloadIdentity` in generatePlatformWorkloadIdentitySecrets but not in generateAuthenticationConfig and generateCloudCredentialOperatorSecret?
Is it because the other 2 will always be called via generateWorkloadIdentityResources?

tsatam: This is the intent, yes. I forgot to remove the check within `generatePlatformWorkloadIdentitySecrets`, but otherwise, these functions are effectively private (though we can't really make them inaccessible to the entire rest of the `cluster` package) so they shouldn't be called outside of `generateWorkloadIdentityResources`.

rajdeepc2792:
LGTM, since it is dependent on https://github.com/Azure/ARO-RP/pull/3619, whenever it is merged this can be rebased and unblocked.

pkg/cluster/workloadidentityresources.go
```diff
@@ -0,0 +1,105 @@
...
+			Name:      "azure-cloud-credentials",
+		},
+		Type: corev1.SecretTypeOpaque,
+		StringData: map[string]string{
+			"azure_tenant_id": tenantId,
+		},
+	}, nil
+}
+
+func (m *manager) generateAuthenticationConfig() (*configv1.Authentication, error) {
```
cadenmarchese, line 91: Should we also be creating the pod identity webhook as part of this PR?
```
apiVersion: v1
stringData:
  azure_tenant_id: <tenant ID>
kind: Secret
metadata:
  name: azure-credentials
  namespace: openshift-cloud-credential-operator
type: Opaque
```

rajdeepc2792: It should be this one:- https://github.com/Azure/ARO-RP/pull/3802/files#diff-1a67ca285a59aa054f6939e0c471cb6d0534cf8c3417be5db84564dd29e7b00bR76
But correctly pointed out the name should be `azure-credentials`

pkg/cluster/workloadidentityresources.go
```diff
@@ -0,0 +1,109 @@
...
+	return secrets, nil
+}
+
+func (m *manager) generateCloudCredentialOperatorSecret() (*corev1.Secret, error) {
+	tenantId := m.subscriptionDoc.Subscription.Properties.TenantID
+
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "openshift-cloud-credential-operator",
+			Name:      "azure-cloud-credentials",
```
rajdeepc2792, line 82: The name should be `azure-credentials`.

pkg/cluster/workloadidentityresources.go
tsatam: Pushed up a fix.

pkg/util/azureclient/azuresdk/armauthorization/generate.go
```diff
@@ -0,0 +1,8 @@
+package armauthorization
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+//go:generate rm -rf ../../../../util/mocks/$GOPACKAGE
+//go:generate go run ../../../../../vendor/github.com/golang/mock/mockgen -destination=../../../../util/mocks/azureclient/azuresdk/$GOPACKAGE/$GOPACKAGE.go github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/$GOPACKAGE RoleDefinitionsClient
+//go:generate go run ../../../../../vendor/golang.org/x/tools/cmd/goimports -local=github.com/Azure/ARO-RP -e -w ../../../../util/mocks/azureclient/azuresdk/$GOPACKAGE/$GOPACKAGE.go
```
SudoBrendan, line 8: Needs rebase on .bingo

SudoBrendan (changes requested):

pkg/util/azureclient/azuresdk/armauthorization/generate.go
tsatam: was part of #3619 which was included in this PR, I've now rebased on latest master since that has merged and this should go away.

rajdeepc2792 (approved):

pkg/cluster/workloadidentityresources.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+
+	return secrets, nil
+}
+
+func (m *manager) generateCloudCredentialOperatorSecret() (*corev1.Secret, error) {
+	tenantId := m.subscriptionDoc.Subscription.Properties.TenantID
+
+	return &corev1.Secret{
```
rajdeepc2792, line 88: Can we add and make sure the typeMeta is set when returning the object?
Maybe:-
```
TypeMeta: metav1.TypeMeta{
					Kind:       "Secret",
					APIVersion: corev1.SchemeGroupVersion.String(),
				},
```

pkg/cluster/workloadidentityresources.go
```diff
@@ -0,0 +1,114 @@
...
+	}, nil
+}
+
+func (m *manager) generateAuthenticationConfig() (*configv1.Authentication, error) {
+	oidcIssuer := m.doc.OpenShiftCluster.Properties.ClusterProfile.OIDCIssuer
+	if oidcIssuer == nil {
+		return nil, fmt.Errorf("oidcIssuer not present in clusterdoc")
+	}
+
+	return &configv1.Authentication{
```
rajdeepc2792, line 106: Same as above, can we set the TypeMeta here?

pkg/cluster/workloadidentityresources.go
```diff
@@ -0,0 +1,114 @@
...
+	subscriptionId := m.subscriptionDoc.ID
+	tenantId := m.subscriptionDoc.Subscription.Properties.TenantID
+	region := m.doc.OpenShiftCluster.Location
+
+	roles := m.platformWorkloadIdentityRolesByVersion.GetPlatformWorkloadIdentityRolesByRoleName()
+
+	secrets := []*corev1.Secret{}
+	for _, identity := range m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
+		if role, ok := roles[identity.OperatorName]; ok {
+			secrets = append(secrets, &corev1.Secret{
```
rajdeepc2792, line 65: Same as other place, can we set TypeMeta?

rajdeepc2792 (changes requested):
Asked for changes related to TypeMeta when returning manifests.

pkg/cluster/workloadidentityresources.go
tsatam: I'll do it for now, but I think there ought to be an elegant way to have our JSON/YAML serializer set the TypeMeta on the resources when writing them. This TypeMeta is not necessary when doing e.g. cluster updates, as K8S clients will automatically inject the TypeMeta on the resulting HTTP request bodies before sending them.

rajdeepc2792: Yup, can we have them generated based on isCreate flag somehow? In a way that the TypeMeta is only set for creation time manifests. When going through the installer code there are multiple places where it's just placed when returning the objects, eg:- https://github.com/openshift/installer/blob/release-4.14/pkg/asset/machines/baremetal/hosts.go#L39

rajdeepc2792: Apologies for un resolving it, wanted to make it visible.

tsatam: I don't think it's necessarily harmful to set it when doing updates, so I'll leave it alone. If I find that it causes problems later when implementing update, I can make it conditional then (or just remove TypeMeta before applying the resources).

rajdeepc2792 (approved):

PR #3811, opened 2024-08-31
Fix slow tests in /pkg/cluster

SudoBrendan:

Fixes https://issues.redhat.com/browse/ARO-9968

Generate smaller OIDC keys for unit tests, which removes all OIDC related tests from the "slowest" in our suite.

- significantly increases unit test performance by moving from 4096 -> 256 bit keys (10s -> 0.01s)
- preserves 4096 bit keys for all non-testing scenarios

### Test plan for issue:

make unit-test-go

N/A

env.prod has been hard-coded with the expected 4096 bit size.

### DIFF
--- a/pkg/env/prod.go
+++ b/pkg/env/prod.go
@@ -271,6 +271,10 @@ func (p *prod) OIDCEndpoint() string {
 	return fmt.Sprintf("https://%s/", os.Getenv("OIDC_AFD_ENDPOINT"))
 }
 
+func (p *prod) OIDCKeyBitSize() int {
+	return 4096
+}
+
 func (p *prod) AROOperatorImage() string {
 	return fmt.Sprintf("%s/aro:%s", p.acrDomain, version.GitCommit)
 }
--- a/pkg/cluster/deploybaseresources_test.go
+++ b/pkg/cluster/deploybaseresources_test.go
@@ -1421,6 +1421,7 @@ func TestCreateOIDC(t *testing.T) {
 			},
 		},
 	}
+	testOIDCKeyBitSize := 256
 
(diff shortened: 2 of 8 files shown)


pkg/env/prod.go
```diff
@@ -271,6 +271,10 @@ func (p *prod) OIDCEndpoint() string {
 	return fmt.Sprintf("https://%s/", os.Getenv("OIDC_AFD_ENDPOINT"))
 }
 
+func (p *prod) OIDCKeyBitSize() int {
```
SudoBrendan, line 274: This might be considered "abuse" of the env.Interface because the value is hard-coded - at the same time it felt like the easiest plug-and-play solution to get a [steps.Action call configured properly](https://github.com/Azure/ARO-RP/blob/710eb2bdf0dc02772066b48a5111000367f48e0e/pkg/cluster/install.go#L301). Please LMK if there are alternative ways I'm not thinking of to inject this configuration :)

tsatam (approved):

pkg/env/prod.go
rajdeepc2792: Since it is needed to be tested and passed from https://github.com/Azure/ARO-RP/blob/d6f776e05f72c5558815a1bb57c7e5974dde27ec/pkg/cluster/deploybaseresources_test.go, I too think `env` can be the easiest way to configure the value. Better than inserting an if case for the localDev.

rajdeepc2792 (approved):
LGTM, good to know, 4096 -> 256 improves the test performance that much!! TY!

kevinobriendotca (approved):

PR #3816, opened 2024-09-03
Separate Makefile targets for local vs containerized

SudoBrendan:

Fixes feedback/concerns raised by the wider ARO team.

Loki is committed to enhancing our build and release process by moving local testing and validation closer to a production-like state. While we believe this approach will effectively address current issues, we respect individual preferences and will not mandate its use. Loki will continue to elevate our CI and containerized builds, ensuring top-tier local execution. Our CI/CD roadmap will continue to focus on containerized builds and validation for both PR validation and production builds, because our current architecture runs containers in production as defined in these build processes.

TL;DR - you now need to opt into the new, containerized build process locally. There is a clear distinction and naming consistency between local binary and containerized build/execution. Use whatever you feel is best locally.

Provides flexibility in local development - you can now opt into containerized build/run commands. Old commands are 100% reverted.

### Test plan for issue:

Only affects local.

Yes, included in the PR.

N/A

### DIFF
--- a/.pipelines/ci.yml
+++ b/.pipelines/ci.yml
@@ -175,7 +175,7 @@ stages:
 
             # run build
             export PODMAN_REMOTE_ARGS="-r --url=tcp://localhost:8888"
-            make azext-aro
+            make ci-azext-aro
 
             # Stop podman
             kill_podman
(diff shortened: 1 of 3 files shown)


tsatam (dismissed):

shubhadapaithankar (dismissed):

kevinobriendotca (approved):

jhoreman (approved):
Approved, thanks for this.

kevinobriendotca (approved):

tsatam (approved):

PR #3818, opened 2024-09-04
ARO-10003: Add managed flag to etchosts controller

lranjbar:

Fixes ARO-10003

Adding a managed flag to the etchosts controller so that we can remove these machine configs in the future if needed.

### Test plan for issue:

- Run operator unit tests
- Run operator e2e tests
- Manually test using the operator testing procedure outlined in the [operator docs](https://github.com/Azure/ARO-RP/blob/master/docs/operators.md#how-to-run-a-custom-operator-image)

N/A

N/A

### DIFF
--- a/pkg/operator/controllers/etchosts/cluster_controller.go
+++ b/pkg/operator/controllers/etchosts/cluster_controller.go
@@ -9,6 +9,7 @@ import (
 	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"github.com/sirupsen/logrus"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -29,9 +30,27 @@ const (
 	ClusterControllerName = "EtcHostsCluster"
 )
 
+var (
+	etchostsMasterMCMetadata = &mcv1.MachineConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "99-master-aro-etc-hosts-gateway-domains",
+		},
+		TypeMeta: metav1.TypeMeta{
+			Kind: "MachineConfig",
+		},
+	}
+	etchostsWorkerMCMetadata = &mcv1.MachineConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "99-worker-aro-etc-hosts-gateway-domains",
+		},
+		TypeMeta: metav1.TypeMeta{
+			Kind: "MachineConfig",
+		},
+	}
+)
+
 type EtcHostsClusterReconciler struct {
 	base.AROController
-
 	dh dynamichelper.Interface
 }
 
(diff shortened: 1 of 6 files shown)


pkg/operator/flags.go
```diff
@@ -35,6 +35,7 @@ const (
 	GuardrailsDeployManaged            = "aro.guardrails.deploy.managed"
 	CloudProviderConfigEnabled         = "aro.cloudproviderconfig.enabled"
 	EtcHostsEnabled                    = "aro.etchosts.enabled"
+	EtcHostsManaged                    = "aro.etchosts.managed"
```
bitoku, line 38: This is just a suggestion, but how about leaving comments how it will work when we enable them.

bitoku (commented):
Why do we have two similar controllers?

and this is a nit, but it doesn't care about this scenario.
> enabled:true and managed:anything other than true or false, it performs no action

Since we don't validate the value in `GetSimpleBoolean`, I don't think we have to care about it though.

pkg/operator/controllers/etchosts/cluster_controller.go
```diff
@@ -68,43 +89,47 @@ func (r *EtcHostsClusterReconciler) Reconcile(ctx context.Context, request ctrl.
...
+			}
 			return reconcile.Result{}, err
 		}
-		err = reconcileMachineConfigs(ctx, instance, "master", r.dh, *mcp)
 		if err != nil {
 			r.Log.Error(err)
 			r.SetDegraded(ctx, err)
 			return reconcile.Result{}, err
 		}
+		return reconcile.Result{}, nil
```
tsatam, line 109: IMO we should requeue here

```suggestion
		return reconcile.Result{Requeue: true}, nil
```

pkg/operator/controllers/etchosts/cluster_controller.go
```diff
@@ -68,43 +89,47 @@ func (r *EtcHostsClusterReconciler) Reconcile(ctx context.Context, request ctrl.
...
+			}
 			return reconcile.Result{}, err
 		}
-		err = reconcileMachineConfigs(ctx, instance, "worker", r.dh, *mcp)
 		if err != nil {
 			r.Log.Error(err)
 			r.SetDegraded(ctx, err)
 			return reconcile.Result{}, err
 		}
+		return reconcile.Result{}, nil
```
tsatam, line 132: I think we should requeue here too, just to ensure we complete one no-op Reconcile loop and clear any outstanding status conditions.

pkg/operator/flags.go
```diff
@@ -74,5 +75,6 @@ func DefaultOperatorFlags() map[string]string {
 		GuardrailsDeployManaged:            FlagFalse,
 		CloudProviderConfigEnabled:         FlagTrue,
 		EtcHostsEnabled:                    FlagFalse,
+		EtcHostsManaged:                    FlagFalse,
```
tsatam, line 78: This should be set to True, as this default only applies to new clusters (existing clusters will "default" to False for all newly added flags by virtue of returning nil).

```suggestion
		EtcHostsManaged:                    FlagTrue,
```

tsatam (changes requested):
Mostly ready to go, just a few fixups

tsatam (dismissed):
LGTM - ready to merge on green checks

bitoku:
Thanks @lranjbar.
> Machine config controller reconciles on the machine configs already existing. However if the config is not there it wont create it. The cluster controller will reconcile on changes to the cluster object, and create the configs if they aren't there. The machine config controller, will only reconcile on existing configs.

I was curious about why we have to separate them. I think we don't need to separate them because dynamic helper `Ensure`  creates the resource if we don't have it.
https://github.com/Azure/ARO-RP/blob/486b50838b4d35d1c7bece1e7621e42e4f2c79c0/pkg/util/dynamichelper/dynamichelper.go#L140-L155

This could be out of scope for this PR, so you don't have to refactor it right now though.

bitoku:
I think it should have either e2e tests or unit tests for this controllers (ideally both), but it doesn't seem to have it.
For example this bug should have been caught in CI.
https://github.com/Azure/ARO-RP/pull/3818#discussion_r1745144858

The change will affect all production clusters, so we should be more careful.

The CPMS operator e2e test will help you write the e2e test. 
https://github.com/Azure/ARO-RP/blob/f9c7eb720e1d5a04fdb0ce0b726800bb9c33c165/test/e2e/operator.go#L742

pkg/operator/controllers/etchosts/cluster_controller.go
SudoBrendan: marking unresolved - assuming this got force-pushed over after `Accept Suggestion` was clicked ;)

SudoBrendan (dismissed):
mentioned in slack - re-request review after the requeue is fixed :)

hlipsig (approved):

bitoku (commented):
Mostly LGTM, but I left some comments about the parts looking like redundant

SudoBrendan (approved):
LGTM

hlipsig (approved):

hlipsig:
Looks good.

PR #3819, opened 2024-09-04
Add run-rp docs

SudoBrendan:

Follow up to split between `runlocal-rp` and `run-rp` - these docs outline how to use the new feature of containerized builds and RP execution

- adds documentation on build/run dependencies of containerized build
- adds instructions on how to use 'make run-rp'

Docs

### Test plan for issue:

N/A - docs

Yes, this PR

N/A - docs

### DIFF
--- a/docs/deploy-development-rp.md
+++ b/docs/deploy-development-rp.md
@@ -104,9 +104,19 @@ mock a cluster MSI. This script will also create the platform identities, platfo
 
 1. Run the RP
 
-   ```bash
-   make runlocal-rp
-   ```
+    Option 1: using local compilation and binaries (requires local `go`/build dependencies/etc):
+    ```bash
+    make runlocal-rp
+    ```
+
+    Option 2: using containerized build and run (requires local `podman` and `openvpn`):
+    ```bash
+    # establish a VPN connection to the shared dev environment Hive cluster
+    sudo openvpn secrets/vpn-${LOCATION}.ovpn &
+
+    # build/run the RP as a container
+    make run-rp
+    ```
 
 1. To create a cluster, use one of the following methods:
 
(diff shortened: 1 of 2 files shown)


shubhadapaithankar (approved):

docs/deploy-development-rp.md
```diff
@@ -104,9 +104,19 @@ mock a cluster MSI. This script will also create the platform identities, platfo
...
-   ```
+    Option 1: using local compilation and binaries (requires local `go`/build dependencies/etc):
+    ```bash
+    make runlocal-rp
+    ```
+
+    Option 2: using containerized build and run (requires local `podman` and `openvpn`):
+    ```bash
+    # establish a VPN connection to the shared dev environment Hive cluster
+    sudo openvpn secrets/vpn-${LOCATION}.ovpn &
```
ehvs, line 115: From my testing it must be `vpn-aks-$LOCATION.ovpn`

ehvs (commented):
Perhaps the containerized mode could use either its own page or its own sessions on how setup it, with a shorter set of commands that is required only to succeed `make run-rp`.

docs/deploy-development-rp.md
SudoBrendan: There should be variations of both `secrets/vpn-${LOCATION}.ovpn` and `secrets/vpn-aks-${LOCATION}.ovpn` for every dev region. I've always used the non-aks one locally :thinking: and have tested this in `westeurope` with success... do we have diverging dev environments maybe?

hlipsig: @ehvs Can you follow up on this documentation task in a new PR? Maybe we can reconcile the difference, but I don't think we should continue to block the PR.

hlipsig (approved):

PR #3821, opened 2024-09-05
WIP: Create VSCode Dev Container

SudoBrendan:
- Add a new stage to ci-rp to include all local build tools
- Add config for vscode Dev Containers extension

Improvements for local dev. Makes it so `docs/prepare-your-dev-environment.md` is basically unused.

Implements [VSCode Dev Containers](https://code.visualstudio.com/docs/devcontainers/containers) so that local dev environments are in sync with what CI runs (golang versions, other tools and versions, etc)

### Test plan for issue:

1. If you don't have it already, install [VSCode](https://code.visualstudio.com/docs)
2. Set up the Dev Containers Extension
```
code --install-extension ms-vscode-remote.remote-containers
```
3. [optional] If you have some "standard" VSCode Extensions you also want in your Dev Containers, you can set those up in File > Preferences > Settings > Extensions > Dev Containers (e.g. I'm a VIM normie, so I added `vscodevim.vim` / etc)
4. In the bottom left corner of the IDE, a symbol like `><` now appears, click it, which opens a dropdown menu - and select Open in Dev Container
5. The ci-rp base dev image is built, and then you are launched into the standard dev environment with all required tools installed. You can do things like `make generate` etc without additional local setup (and all it took was `podman` and `podman-docker`!)

Yea - will follow up with this before merge (still WIP).

N/A - local only

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,51 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/docker-existing-dockerfile
+{
+	"name": "aro-rp dev",
+	"build": {
+		// Sets the run context to one level up instead of the .devcontainer folder.
+		"context": "..",
+		// Update the 'dockerFile' property if you aren't using the standard 'Dockerfile' filename.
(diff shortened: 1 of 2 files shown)


jaitaiwan:
LGTM so far

PR #3826, opened 2024-09-09
Encryption At Host Dynamic Validation

SudoBrendan:

Fixes https://issues.redhat.com/browse/ARO-3211

- Frontend will now validate that Encryption At Host is enabled on customer's subscription before installing clusters - this will fix a Cluster Install Failure issue
- In memory of Swetha. Thanks for your amazing contributions to our team, I'm glad we can finally get your work across the line with this one.

### Test plan for issue:

Unit tests. I will perform a manual validation of this, but that will require some effort (a subscription without the feature enabled).

N/A fixes a bug

Unit tests and manual tests.


SudoBrendan:
Adding `hold` because apparently our FPSP permissions were flagged for being not in use for a substantial time, so we need to redo that effort and rollout the permissions before this can merge.

PR #3837, opened 2024-09-11
ARO-9990: Update etchosts controller to use ForceReconcilation flag

lranjbar:

Fixes ARO-9990

This updates the new etchosts controller to wait until the cluster is updating to reconcile the machine configs. This is needed to roll out the controller to the fleet without causing a fleet wide rebooting action.

### Test plan for issue:

- New tests added to the controllers
- Functional operator testing

N/A

- New tests added to the controllers
- Functional operator testing

### DIFF
--- a/pkg/operator/controllers/etchosts/machineconfig_controller.go
+++ b/pkg/operator/controllers/etchosts/machineconfig_controller.go
@@ -64,10 +64,21 @@ func (r *EtcHostsMachineConfigReconciler) Reconcile(ctx context.Context, request
 		return reconcile.Result{}, nil
 	}
 
+	allowReconcile, err := r.AllowRebootCausingReconciliation(ctx, instance)
+	if err != nil {
+		r.Log.Error(err)
+		r.SetDegraded(ctx, err)
+		return reconcile.Result{}, err
+	}
+
 	// EtchostsManaged = false, remove machine configs
-	if !instance.Spec.OperatorFlags.GetSimpleBoolean(operator.EtcHostsManaged) {
+	if !instance.Spec.OperatorFlags.GetSimpleBoolean(operator.EtcHostsManaged) && allowReconcile {
 		r.Log.Debug("etchosts managed is false, removing machine configs")
 		err = r.removeMachineConfig(ctx, etchostsMasterMCMetadata)
+		if kerrors.IsNotFound(err) {
+			r.ClearDegraded(ctx)
+			return reconcile.Result{}, nil
+		}
 		if err != nil {
 			r.Log.Error(err)
 			r.SetDegraded(ctx, err)
@@ -75,6 +86,10 @@ func (r *EtcHostsMachineConfigReconciler) Reconcile(ctx context.Context, request
 		}
 
 		err = r.removeMachineConfig(ctx, etchostsWorkerMCMetadata)
+		if kerrors.IsNotFound(err) {
+			r.ClearDegraded(ctx)
+			return reconcile.Result{}, nil
+		}
 		if err != nil {
 			r.Log.Error(err)
 			r.SetDegraded(ctx, err)
@@ -111,7 +126,7 @@ func (r *EtcHostsMachineConfigReconciler) Reconcile(ctx context.Context, request
 		return reconcile.Result{}, nil
 	}
 
-	err = reconcileMachineConfigs(ctx, instance, role, r.dh, *mcp)
+	err = reconcileMachineConfigs(ctx, instance, role, r.dh, allowReconcile, *mcp)
 	if err != nil {
 		r.Log.Error(err)
 		r.SetDegraded(ctx, err)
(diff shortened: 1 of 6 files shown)


test/e2e/operator.go
```diff
@@ -891,7 +892,7 @@ var _ = Describe("ARO Operator - etchosts", func() {
 			}).WithContext(ctx).WithTimeout(DefaultEventuallyTimeout).Should(Succeed())
 		}
 
-		if !instance.Spec.OperatorFlags.GetSimpleBoolean(etchostsManaged) {
+		if !instance.Spec.OperatorFlags.GetSimpleBoolean(etchostsManaged) && instance.Spec.OperatorFlags.GetSimpleBoolean(forceReconciliation) {
```
bitoku, line 895: Can we reach this condition?
The default value of `etchostsManaged` is true. If we don't change the value elsewhere, we won't reach here.
Also it runs just once in each e2e test run, so only one of those conditions can be run, and others can't.

I think we should split this test cases into multiple test cases according to operator flags and test each behaviour in each test case not using conditional branches.

SudoBrendan: I think I prefer the test as-is, because it is future-proof to our E2E test expanding out from the rigid structure it is in today (e.g. us performing permutation tests on various clusters with different features set in each one).

SudoBrendan (approved):
LGTM - if you can remove the unnecessary Client from that function header, all the better. Thanks Lisa, great work!

SudoBrendan (approved):

pkg/operator/controllers/etchosts/machineconfig_controller.go
```diff
@@ -141,7 +156,7 @@ func (r *EtcHostsMachineConfigReconciler) SetupWithManager(mgr ctrl.Manager) err
 		Complete(r)
 }
 
-func reconcileMachineConfigs(ctx context.Context, instance *arov1alpha1.Cluster, role string, dh dynamichelper.Interface, mcps ...mcv1.MachineConfigPool) error {
+func reconcileMachineConfigs(ctx context.Context, instance *arov1alpha1.Cluster, role string, dh dynamichelper.Interface, c client.Client, allowReconcile bool, mcps ...mcv1.MachineConfigPool) error {
```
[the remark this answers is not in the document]
bitoku: > allowReconcile only matters on calls to Ensure

It's true,

> I think if we wanted to make this more clear, we could have some follow-up work to move Ensure into Reconcile (probably for every controller - not just this one), with all other functions building up a resources object to pass into Ensure - how does that sound?

 but I don't understand why we need to move `Ensure` to `Reconcile`.
I just think that the logic from `r.AllowRebootCausingReconciliation(ctx, instance)` to `Ensure` is not used if `allowReconcile` is false (sorry if this understanding is wrong).
If it stops reconciling just after `r.AllowRebootCausingReconciliation(ctx, instance)`, it will be more readable because we don't have to follow the lines until `Ensure`, and also there will be (very) slight improvement in performance.

Also `AllowRebootCausingReconciliation` is used only in dnsmasq and etchosts.
And the dnsmasq has the reason not to stop earlier (because it has to create some things), so I don't think we have the convention we should follow here.

> We can card this up but to my knowledge this allowReconcile approach is going to be EOL with SyncSets. Eventually this will move into syncsets.

If so, we might not necessarily have to change the code, but I'd say just changing the code here is doable and worth doing.

test/e2e/operator.go
bitoku: Can't we change operator flags before running each test case?
I think we can because we can change cluster operator flags by running oc edit command.
If we can change the flags then, we can set them before each test, and we can test all scenarios.

Each test case should focus on one scenario, and this test case doesn't follow the principle.
Otherwise the purpose of the test cases will be unclear.
If we had to wait for the change in how to run E2E, we can rewrite this test case like below:

```go
It("must not reconcile machine configs when enabled == false", func(ctx context.Context) {
  if instance.Spec.OperatorFlags.GetSimpleBoolean(etchostsEnabled) {
    Skip("skipping because enabled == true")
  }
// ...
}
It("must reconcile machine configs when managed == true and ...", func(ctx context.Context) {
  if !instance.Spec.OperatorFlags.GetSimpleBoolean(etchostsManaged) || 
     !instance.Spec.OperatorFlags.GetSimpleBoolean(forceReconciliation) {
    Skip("skipping because managed == false || forceReconciliation == false")
  }
}
// ...
It("must remove machine configs when managed == false and ...", func(ctx context.Context) {
  if instance.Spec.OperatorFlags.GetSimpleBoolean(etchostsManaged) || 
     !instance.Spec.OperatorFlags.GetSimpleBoolean(forceReconciliation) {
      Skip("skipping because managed == true || forceReconciliation == false")
  }
}
```

jhoreman (approved):
This looks good, thanks for this!

shubhadapaithankar (approved):

hlipsig:
e2e is passed. Merging.

pkg/operator/controllers/etchosts/machineconfig_controller.go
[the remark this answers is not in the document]
bitoku: > You seem to had a fundamental mistunderstanding of what allowReconcile is doing. This must be used to stop from Ensure or Create being called to have the desired effect.

That's what I understand.
That's why I'm asking you why these codes are needed to run even though they don't Ensure or Create.

https://github.com/Azure/ARO-RP/blob/3762a791fe70e5ac439d03d289b31a3dacdddf86/pkg/operator/controllers/etchosts/machineconfig_controller.go#L160-L178

test/e2e/operator.go
bitoku: Which question is No? You mean we can't change operator flags before running each test case?

Also I want you to explain which part is wrong to get better understanding.

PR #3841, opened 2024-09-13
ARO-4518 pass custom manifests(MIWI) to hive cluster deployment as secret

rajdeepc2792:

Fixes (https://issues.redhat.com/browse/ARO-4518)

Call generatePlatformWorkloadIdentitySecrets() for generating the Platform Identity Secret for a MIWI cluster.

Pass all the Secrets as a hive cluster secret and pass it the name of hive secret to  cluster deployment's `Spec.Provisioning.ManifestsSecretRef`

These secrets will be managed in ARO-Installer-Wrapper

Additional Handling of CredentialsSecretRef is required as it is only used for the Non-MIWI clusters, but since it is used to pass the cluster and subscription doc to wrapper it is still neded.

### Test plan for issue:
- [x] Unit Tests
- [x] CI
- [x] CI e2e
- [x] Local cluster installation using the installer-wrapper PR https://github.com/openshift/installer-aro-wrapper/pull/166

No

- For non-miwi clusters, tested the the cluster creation flow.
- Custom Manifests creation is tested successfully in combination with the installer-wrapper PR  https://github.com/openshift/installer-aro-wrapper/pull/166

### DIFF
--- a/pkg/hive/install.go
+++ b/pkg/hive/install.go
@@ -57,7 +57,7 @@ func (c *clusterManager) Install(ctx context.Context, sub *api.SubscriptionDocum
 		return err
 	}
 
-	clusterManifestsSecret, err := clusterManifestsSecret(doc.OpenShiftCluster.Properties.HiveProfile.Namespace, customManifests)
+	manifestsSecret, err := clusterManifestsSecret(doc.OpenShiftCluster.Properties.HiveProfile.Namespace, customManifests)
 	if err != nil {
 		return err
 	}
@@ -82,7 +82,7 @@ func (c *clusterManager) Install(ctx context.Context, sub *api.SubscriptionDocum
 
 	resources := []kruntime.Object{
 		azureCredentialSecret,
-		clusterManifestsSecret,
+		manifestsSecret,
 		envSecret(doc.OpenShiftCluster.Properties.HiveProfile.Namespace, c.env.IsLocalDevelopmentMode()),
 		psSecret,
 		installConfigCM(doc.OpenShiftCluster.Properties.HiveProfile.Namespace, doc.OpenShiftCluster.Location),
--- a/pkg/hive/install.go
+++ b/pkg/hive/install.go
@@ -51,8 +51,13 @@ func makeEnvSecret(name string) corev1.EnvVar {
 	}
 }
 
-func (c *clusterManager) Install(ctx context.Context, sub *api.SubscriptionDocument, doc *api.OpenShiftClusterDocument, version *api.OpenShiftVersion) error {
-	sppSecret, err := servicePrincipalSecretForInstall(doc.OpenShiftCluster, sub, c.env.IsLocalDevelopmentMode())
+func (c *clusterManager) Install(ctx context.Context, sub *api.SubscriptionDocument, doc *api.OpenShiftClusterDocument, version *api.OpenShiftVersion, customManifests map[string]kruntime.Object) error {
+	azureCredentialSecret, err := azureCredentialSecretForInstall(doc.OpenShiftCluster, sub, c.env.IsLocalDevelopmentMode())
+	if err != nil {
+		return err
+	}
+
+	clusterManifestsSecret, err := clusterManifestsSecret(doc.OpenShiftCluster.Properties.HiveProfile.Namespace, customManifests)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


pkg/hive/install.go
```diff
@@ -76,7 +81,8 @@ func (c *clusterManager) Install(ctx context.Context, sub *api.SubscriptionDocum
 	}
 
 	resources := []kruntime.Object{
-		sppSecret,
+		azureCredentialSecret,
+		clusterManifestsSecret,
```
cadenmarchese, line 85: Is there an opportunity to combine these two so we only pass one secrets object to Hive?

cadenmarchese (commented):
just a quick pass on the draft PR. thank you!

pkg/hive/install.go
rajdeepc2792: No, it has to be separate secrets as these both secrets go in their dedicated volume for the installer pod.
azureCredentialSecret -> `/.azure`
clusterManifestsSecret -> `/manifests`

Also, the usage for both secrets are different, so it is kept as per what Hive appears.

pkg/hive/install.go
```diff
@@ -51,8 +51,13 @@ func makeEnvSecret(name string) corev1.EnvVar {
 	}
 }
 
-func (c *clusterManager) Install(ctx context.Context, sub *api.SubscriptionDocument, doc *api.OpenShiftClusterDocument, version *api.OpenShiftVersion) error {
-	sppSecret, err := servicePrincipalSecretForInstall(doc.OpenShiftCluster, sub, c.env.IsLocalDevelopmentMode())
+func (c *clusterManager) Install(ctx context.Context, sub *api.SubscriptionDocument, doc *api.OpenShiftClusterDocument, version *api.OpenShiftVersion, customManifests map[string]kruntime.Object) error {
+	azureCredentialSecret, err := azureCredentialSecretForInstall(doc.OpenShiftCluster, sub, c.env.IsLocalDevelopmentMode())
```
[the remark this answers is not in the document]
rajdeepc2792: Changed the name as the new secret that is `azureCredentialSecret` would also be used wi/mi cluster to transfer the oc and subscription doc. Installer-wrapper uses it here https://github.com/openshift/installer-aro-wrapper/blob/main/cmd/aro/create.go#L150.

pkg/hive/resources.go
```diff
@@ -98,17 +102,53 @@ func aroServiceKubeconfigSecret(namespace string, kubeConfig []byte) *corev1.Sec
...
+		if err != nil {
+			return nil, err
+		}
+		secret.Data["osServicePrincipal.json"] = clusterSPBytes
+	}
+
+	return secret, nil
+}
+
+func clusterManifestsSecret(namespace string, customManifests map[string]kruntime.Object) (*corev1.Secret, error) {
```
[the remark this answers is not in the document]
rajdeepc2792: Instead of changing the function name, changed the var name:- https://github.com/Azure/ARO-RP/pull/3841/files#diff-f83dd6c00cfc7b417ce96e6f7bfbe58ce118101ea90630f21d658abf3e96b129R60

`generateClusterManifestsSecret` name is similar to `https://github.com/Azure/ARO-RP/pull/3841/files#diff-71a1dd610bf07a9e4ece9f782af2bad11e6f2d0e03df442019088f8c47265497R281` where the actual generation is taking place and keeping it similar to similar type of secret creations in the file.

pkg/hive/resources_test.go
```diff
@@ -18,3 +23,92 @@ func TestInstallConfigMap(t *testing.T) {
...
+					Properties: &api.SubscriptionProperties{
+						TenantID: "tenantid",
+					},
+				},
+				ID: "fakeID",
+			}
+			r, _ := clusterAzureSecret("testNamespace", tt.oc, &subDoc)
+
+			_, ok := r.Data["osServicePrincipal.json"]
+			if tt.wantSP != ok {
```
bitoku, line 106: nit: You can use assert pkg.

https://github.com/Azure/ARO-RP/blob/5370de9fd11365073e09f6a4774298a40daa8378/pkg/cluster/networkprofile_test.go#L294-L295

bitoku (approved):

pkg/hive/resources_test.go
tsatam: While I'm personally in favor of using the assert pkg here, I don't think we have a consistent pattern for these assertions across our codebase's test cases, so I'm fine either way with changing this or letting it merge as-is.

tsatam (approved):
LGTM - excited to see this working!

PR #3842, opened 2024-09-16
Update azure sdk to a newer (yet still deprecated) API Version

jaitaiwan:
When trying to perform the update/creation of a shared RP environment, a user will get an error from the Azure API stating that the mimimum supported API version is 2020-01-01. The current version of the containerregistry and networking clients were below that and required updating.

> [!NOTE]
> The PR is broken up commit-wise so that you can ignore commits that add vendoring, and focus on commits that change actual files we care about.


jaitaiwan:
Closing out in favour of an upgrade to track 2 already in progress by @atokobi and I will be working on the containerruntime client.

PR #3843, opened 2024-09-16
Initial ansible proof of concept

ventifus:
**NOTE** Deprecated in favor of https://github.com/openshift/aro-e2e/pull/3

This PR leverages Ansible to automate cluster creation in predefined configurations. These configurations are intended to exercise ARO features in a repeatable manner. A formal proposal is under development, see

https://docs.google.com/document/d/1AIcP1jMQS-9hsI_bS6Qhr3IoqtD97LqKfw6c9PNXD5g

### Test plan for issue:

Then run

```
$ make cluster
```

This will first build the required aro-ansible container image if it doesn't exist, then it will execute `ansible-playbook` in the container.

There are several variables implemented to control aspects of the playbook execution. These can be combined with each other as needed.

To personalize the resulting resource groups, set the `CLUSTERPREFIX` as desired. This variable defaults to your current shell's `$USER`.

```
$ make cluster CLUSTERPREFIX=ocpbugs35300
```

The default region used is `eastus`. Set the `REGION` parameter to choose a different region:

```
$ make cluster REGION=centraluseuap
```

To choose one or more cluster configurations, set the `CLUSTERPATTERN` parameter to a wildcard string that matches the cluster scenarios you wish to test:

```
$ make cluster CLUSTERPATTERN=udr
```

To clean up at the end of the run, set `CLEANUP` to True. This will delete the cluster, the resource group, then the Entra Service Principal and Application.

```
$ make cluster CLEANUP=True
```

- `basic`: Simplest cluster, nothing fancy
- `private`: Simple cluster with apiserver and ingress visibility set to private.
- `enc`: Encryption-at-host enabled
- `udr`: UserDefinedRouting with a blackhole Route Table
- `byok`: Disk encryption using bring-your-own-key

Private clusters such as `private` and `udr` will cause the creation of a jumphost to access the cluster API. Your local SSH public key will be passed to the jumphost, then ansible will use your corresponding private key to tunnel SSH through it. If your local SSH configuration differs from defaults, the Makefile supports two variables to tweak things:

```
SSH_CONFIG_DIR := $(HOME)/.ssh/
SSH_KEY_BASENAME := id_rsa
```

Yes

### DIFF
--- a/Dockerfile.ansible
+++ b/Dockerfile.ansible
@@ -1,0 +1,56 @@
+ARG REGISTRY
+ARG VERSION
+
+###############################################################################
+# ansible is the base Python image with ansible and azure-cli
+###############################################################################
+FROM ${REGISTRY}/ubi9/python-311:1-66 AS ansible
+# Versions
+# pipx https://pypi.org/project/pipx/#history
+# azure-cli https://pypi.org/project/azure-cli/#history
+# ansible https://pypi.org/project/ansible/#history
+# ansible.azcollection https://galaxy.ansible.com/ui/repo/published/azure/azcollection/
+ARG PIPX_VERSION=1.7.1 \
+    ANSIBLE_VERSION=10.4.0 \
+    AZURE_CLI_VERSION=2.64.0
+# ANSIBLE_AZCOLLECTION_VERSION=2.7.0 # ansible now ships with this
+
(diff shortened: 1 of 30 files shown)


ventifus:
Moved to https://github.com/openshift/aro-e2e/pull/3

PR #3709, opened 2024-07-19
WI/MI CLI Phase 1 - Base Update Functionality

tsatam:
### Which issue this PR addresses:

Fixes [ARO-6451](https://issues.redhat.com/browse/ARO-6451)

### What this PR does / why we need it:

Implements the ability to set additional platform workload identities on a managed identity-enabled cluster via `az aro update`. 

### Test plan for issue:

- [X] Unit tests for validating the `--assign-platform-workload-identity` argument have been updated to account for update scenarios
- [X] `az aro update` against a MIWI cluster results in an API call being made to the RP with the necessary fields populated
  - Example invocation: 
    ```sh
    az aro update \
      --subscription "${SUBSCRIPTION_ID}" \
      --resource-group "${CLUSTER_RESOURCE_GROUP}" \
      --name azcli-testcluster \
      --assign-platform-workload-identity bar qwert \
      --assign-platform-workload-identity baz zxcvb
    ```
  - Against cluster w/ document:
    <details>
      <summary>Open</summary>
      
      ```json
      {
          "id": "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${CLUSTER_RESOURCE_GROUP}/providers/Microsoft.RedHatOpenShift/openShiftClusters/azcli-testcluster",
          "location": "eastus",
          "identity": {
              "type": "UserAssigned",
              "userAssignedIdentities": {
                  "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${CLUSTER_RESOURCE_GROUP}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/baz": {}
              }
          },
          "properties": {
              "clusterProfile": {
                  "pullSecret": "",
                  "domain": "fv0w9cvw",
                  "version": "4.14.16",
                  "resourceGroupId": "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/aro-fv0w9cvw",
                  "fipsValidatedModules": "Disabled"
              },
              "platformWorkloadIdentityProfile": {
                  "platformWorkloadIdentities": [
                      {
                          "operatorName": "foo",
                          "resourceId": "/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${CLUSTER_RESOURCE_GROUP}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/asdf"
                      },
                      {
                          "operatorName": "bar",
[...]

PR #3844, opened 2024-09-16
CLI Phase 1 - add UpgradeableTo field to update functionality

slawande2:

[ARO-9607](https://issues.redhat.com/browse/ARO-9607)

Add support for Cx to provide the value of the OpenShift version they want to upgrade to using UpgradeableTo field during update such that the backend can validate platform workload identities for the current and upgradeableTo minor versions and federate the upgradeableTo version identities.

### Test plan for issue:
 - [x] Unit tests for validating the --upgradeable-to argument have been added to account for update scenarios
 - [ ] az aro update against a MIWI cluster results in an API call being made to the RP with the necessary fields populated

Not yet

Preview extension-only change. We will perform comprehensive testing before releasing this extension to users.

### DIFF
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -160,6 +160,9 @@ def load_arguments(self, _):
                    options_list=['--assign-platform-workload-identity', '--assign-platform-wi'],
                    validator=validate_platform_workload_identities(isCreate=False),
                    action=AROPlatformWorkloadIdentityAddAction, nargs='+')
+        c.argument('upgradeable_to', arg_group='Identity', options_list=['--upgradeable_to'],
+        help='OpenShift version to upgrade/update to.', is_preview=True,
+        validator=validate_version_format)
 
     with self.argument_context('aro get-admin-kubeconfig') as c:
         c.argument('file',
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -23,6 +23,7 @@ from azext_aro._validators import (
     validate_enable_managed_identity,
     validate_platform_workload_identities,
     validate_cluster_identity,
+    validate_upgradeable_to_format,
 )
 from azure.cli.core.commands.parameters import (
     name_type,
@@ -160,9 +161,9 @@ def load_arguments(self, _):
                    options_list=['--assign-platform-workload-identity', '--assign-platform-wi'],
                    validator=validate_platform_workload_identities(isCreate=False),
                    action=AROPlatformWorkloadIdentityAddAction, nargs='+')
-        c.argument('upgradeable_to', arg_group='Identity', options_list=['--upgradeable_to'],
+        c.argument('upgradeable_to', arg_group='Identity', options_list=['--upgradeable-to'],
         help='OpenShift version to upgrade/update to.', is_preview=True,
-        validator=validate_version_format)
+        validator=validate_upgradeable_to_format)
 
     with self.argument_context('aro get-admin-kubeconfig') as c:
         c.argument('file',
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -162,8 +162,8 @@ def load_arguments(self, _):
                    validator=validate_platform_workload_identities(isCreate=False),
                    action=AROPlatformWorkloadIdentityAddAction, nargs='+')
         c.argument('upgradeable_to', arg_group='Identity', options_list=['--upgradeable-to'],
-        help='OpenShift version to upgrade/update to.', is_preview=True,
-        validator=validate_upgradeable_to_format)
+                   help='OpenShift version to upgrade to.', is_preview=True,
+                   validator=validate_upgradeable_to_format)
 
     with self.argument_context('aro get-admin-kubeconfig') as c:
         c.argument('file',
--- a/python/az/aro/azext_aro/_validators.py
+++ b/python/az/aro/azext_aro/_validators.py
@@ -41,6 +41,7 @@ def validate_cidr(key):
 def validate_client_id(namespace):
     if namespace.client_id is None:
         return
+    raise RequiredArgumentMissingError('--client-id must be not set with --upgradeable-to.') # pylint: disable=line-too-long
     if namespace.enable_managed_identity is True:
         raise MutuallyExclusiveArgumentError('Must not specify --client-id when --enable-managed-identity is True')  # pylint: disable=line-too-long
     if namespace.platform_workload_identities is not None:
@@ -59,6 +60,7 @@ def validate_client_secret(isCreate):
     def _validate_client_secret(namespace):
         if namespace.client_secret is None:
             return
+        raise RequiredArgumentMissingError('--client-secret must be not set with --upgradeable-to.') # pylint: disable=line-too-long
         if namespace.enable_managed_identity is True:
             raise MutuallyExclusiveArgumentError('Must not specify --client-secret when --enable-managed-identity is True')  # pylint: disable=line-too-long
         if namespace.platform_workload_identities is not None:
@@ -292,8 +294,6 @@ def validate_upgradeable_to_format(namespace):
         return
     if not re.match(r'^[4-9]{1}\.[0-9]{1,2}\.[0-9]{1,2}$', namespace.upgradeable_to):
         raise InvalidArgumentValueError('--upgradeable-to is invalid')
-    if namespace.client_secret is not None or namespace.client_id is not None:
-        raise RequiredArgumentMissingError('--client-id and --client-secret must be not set with --upgradeable-to.')  # pylint: disable=line-too-long
 
 def validate_load_balancer_managed_outbound_ip_count(namespace):
     if namespace.load_balancer_managed_outbound_ip_count is None:
--- a/python/az/aro/azext_aro/_validators.py
+++ b/python/az/aro/azext_aro/_validators.py
@@ -287,6 +287,13 @@ def validate_version_format(namespace):
     if namespace.version is not None and not re.match(r'^[4-9]{1}\.[0-9]{1,2}\.[0-9]{1,2}$', namespace.version):
         raise InvalidArgumentValueError('--version is invalid')
 
+def validate_upgradeable_to_format(namespace):
+    if not namespace.upgradeable_to:
+        return
+    if not re.match(r'^[4-9]{1}\.[0-9]{1,2}\.[0-9]{1,2}$', namespace.upgradeable_to):
+        raise InvalidArgumentValueError('--upgradeable-to is invalid')
+    if namespace.client_secret is not None or namespace.client_id is not None:
+        raise RequiredArgumentMissingError('--client-id and --client-secret must be not set with --upgradeable-to.')  # pylint: disable=line-too-long
 
 def validate_load_balancer_managed_outbound_ip_count(namespace):
     if namespace.load_balancer_managed_outbound_ip_count is None:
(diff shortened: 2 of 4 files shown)


tsatam (changes requested):
Looks good! We just need to make sure we add unit tests for this and complete functional testing before merge.

python/az/aro/azext_aro/_params.py
```diff
@@ -160,6 +161,9 @@ def load_arguments(self, _):
                    options_list=['--assign-platform-workload-identity', '--assign-platform-wi'],
                    validator=validate_platform_workload_identities(isCreate=False),
                    action=AROPlatformWorkloadIdentityAddAction, nargs='+')
+        c.argument('upgradeable_to', arg_group='Identity', options_list=['--upgradeable-to'],
+        help='OpenShift version to upgrade/update to.', is_preview=True,
+        validator=validate_upgradeable_to_format)
```
rajdeepc2792, line 166: I believe the indentation is incorrect here.

python/az/aro/azext_aro/custom.py
```diff
@@ -479,20 +480,24 @@ def aro_update(cmd,
...
 
-        oc_update.platform_workload_identity_profile = openshiftcluster.PlatformWorkloadIdentityProfile(
-            platform_workload_identities=list(pwis.values())
-        )
+            for i in platform_workload_identities:
+                pwis[i.operator_name] = i
+
+            oc_update.platform_workload_identity_profile.platform_workload_identities=list(pwis.values())
+        
+        oc_update.platform_workload_identity_profile.upgradeable_to = upgradeable_to
```
rajdeepc2792, line 500: I think, we shouldn't be overwriting the upgradeable_to if it is `None`?

python/az/aro/azext_aro/_validators.py
```diff
@@ -287,6 +287,13 @@ def validate_version_format(namespace):
     if namespace.version is not None and not re.match(r'^[4-9]{1}\.[0-9]{1,2}\.[0-9]{1,2}$', namespace.version):
         raise InvalidArgumentValueError('--version is invalid')
 
+def validate_upgradeable_to_format(namespace):
+    if not namespace.upgradeable_to:
+        return
+    if not re.match(r'^[4-9]{1}\.[0-9]{1,2}\.[0-9]{1,2}$', namespace.upgradeable_to):
+        raise InvalidArgumentValueError('--upgradeable-to is invalid')
+    if namespace.client_secret is not None or namespace.client_id is not None:
```
rajdeepc2792, line 295: Should/Can we add this check in client_secret and client_id validations rather than adding here? Suggesting this because for other MIWI fields that validation is perfromed in the client_secret and client_id.

python/az/aro/azext_aro/custom.py
tsatam: The oc_update object is a patch object and is otherwise-empty, so we're overwriting None with None in this case.

rajdeepc2792 (changes requested):
<s>Added some change request comments regarding the functionality.</s>
There are no comments now that impacts functionality, just some suggestions.

python/az/aro/azext_aro/custom.py
rajdeepc2792: So, in that case, would be impacting the `upgradeable_to` in the doc or not? Just concerned, as the platform_workload_identity_profile is pointer type. Checking and confirming here.

rajdeepc2792: The field is being handled correctly at the backend, so keeping it none won't be causing the overwrite. Thanks!!

python/az/aro/azext_aro/_params.py
tsatam: this doesn't appear to break functionality at all, but I agree with fixing this before merge.

python/az/aro/azext_aro/_validators.py
tsatam: I agree with this suggestion, this makes more sense than what I proposed previously (current implementation in this PR right now).

python/az/aro/azext_aro/_validators.py
kimorris27, line 1: It looks to me like we should also disallow the combination of `refresh-credentials` with `upgradeable-to`.

kimorris27 (commented):
LGTM for the most part! I left one suggestion for additional validation, but I'd appreciate if someone else would verify that it's needed.

shubhadapaithankar (dismissed):
LGTM

python/az/aro/azext_aro/_validators.py
rajdeepc2792: I agree with Kipp's argument here, should we also disallow `refresh-credentials` with `enable_managed_identity` and `platform_workload_identities`?
I would wait for @tsatam backing on this though.

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -1276,3 +1277,51 @@ def test_validate_cluster_identity(test_description, namespace, expected_excepti
...
+    ),
+
+    (
+        "should raise RequiredArgumentMissingError Exception because namespace.client_id is present",
+        Mock(upgradeable_to="4.14.5", client_id="client_id_456", client_secret=None),
+        RequiredArgumentMissingError, '--client-id must be not set with --upgradeable-to.'
+    ),
+
+    (
+        "should raise RequiredArgumentMissingError Exception because namespace.client_secret is present",
```
cadenmarchese, line 1296: Just for completeness, should we add a test case for when --client-id, --client-secret, and --upgradeable-to are all present?

python/az/aro/azext_aro/_validators.py
kimorris27: I don't think we need to worry about `enable_managed_identity`. It's not possible to specify it together with `refresh_credentials`, since the former can only be passed to `az aro create` and the latter can only be passed to `az aro update`.

As for the other two suggestions in this thread: 
1. Disallow `refresh_cluster_credentials` together with `upgradeable_to`
2. Disallow `refresh_cluster_credentials` together with `platform_workload_identities`

@slawande2 could definitely do 1, but 2 is technically out of scope for this PR; we missed it when https://github.com/Azure/ARO-RP/pull/3709 was in review. Maybe it's worth putting 2 in its own little PR?

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -171,15 +172,27 @@ def test_validate_client_id(test_description, namespace, expected_exception):
...
     ),
     (
         "should not raise any exception when isCreate is false and all arguments valid",
         False,
-        Mock(client_secret="123", platform_workload_identities=None),
+        Mock(upgradeable_to=None, client_secret="123", platform_workload_identities=None),
         None
     ),
+    (
+        "should raise any exception when isCreate is true and upgradeable_to, client_id and client_secret are present",
```
kimorris27, line 185: Nit: 
```suggestion
        "should raise MutuallyExclusiveArgumentError exception when isCreate is true and upgradeable_to, client_id and client_secret are present",
```

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -171,15 +172,27 @@ def test_validate_client_id(test_description, namespace, expected_exception):
...
+        True,
+        Mock(upgradeable_to="4.14.2", client_id="12345678123456781234567812345678", client_secret="123", platform_workload_identities=None),
+        RequiredArgumentMissingError
+    ),
+    (
+        "should raise any exception when isCreate is false and upgradeable_to, client_id and client_secret are present",
```
kimorris27, line 191: Nit:
```suggestion
        "should raise MutuallyExclusiveArgumentError exception when isCreate is false and upgradeable_to, client_id and client_secret are present",
```

kimorris27 (changes requested):
Looking good, but I requested a few changes to semantics. There's also a bit more additional validation we can do.

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -817,9 +817,15 @@ def test_validate_worker_vm_disk_size_gb(test_description, namespace, expected_e
     ),
     (
         "should not raise any Exception because namespace.client_secret is None and namespace.client_id is None",
-        Mock(client_secret=None, client_id=None),
+        Mock(upgradeable_to=None, client_secret=None, client_id=None),
         None
-    )
+    ), 
+    (
+        "should raise RequiredArgumentMissingError Exception because namespace.upgradeable_to is not None",
```
kimorris27, line 824: Another tiny nit:
```suggestion
        "should raise MutuallyExclusiveArgumentError exception because namespace.upgradeable_to is not None",
```

kimorris27 (approved):
My most recent nitpick is non-blocking since it's clear from reading the code that it's a run-of-the-mill typo 🙂

shubhadapaithankar (dismissed):

kimorris27 (dismissed):

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -804,9 +817,15 @@ def test_validate_worker_vm_disk_size_gb(test_description, namespace, expected_e
     ),
     (
         "should not raise any Exception because namespace.client_secret is None and namespace.client_id is None",
-        Mock(client_secret=None, client_id=None),
+        Mock(upgradeable_to=None, client_secret=None, client_id=None),
         None
-    )
+    ),
```
cadenmarchese, line 822:
```suggestion
    ),
```

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
```diff
@@ -804,9 +817,15 @@ def test_validate_worker_vm_disk_size_gb(test_description, namespace, expected_e
...
-        Mock(client_secret=None, client_id=None),
+        Mock(upgradeable_to=None, client_secret=None, client_id=None),
         None
-    )
+    ), 
+    (
+        "should raise MutuallyExclusiveArgumentError exception because namespace.upgradeable_to is not None",
+        Mock(upgradeable_to="4.14.2", client_id=None, client_secret=None),
+        MutuallyExclusiveArgumentError
+    ),
```
cadenmarchese, line 827:
```suggestion
    ),
```

cadenmarchese (commented):
There's still an outstanding issue with the python lint - I think that this should fix it.
```
Running flake8 on extensions...
ERROR: /__w/1/go/src/github.com/Azure/ARO-RP/python/az/aro/azext_aro/tests/latest/unit/test_validators.py:822:7: W291 trailing whitespace
/__w/1/go/src/github.com/Azure/ARO-RP/python/az/aro/azext_aro/tests/latest/unit/test_validators.py:827:7: W291 trailing whitespace
/__w/1/go/src/github.com/Azure/ARO-RP/python/az/aro/build/lib/azext_aro/tests/latest/unit/test_validators.py:822:7: W291 trailing whitespace
/__w/1/go/src/github.com/Azure/ARO-RP/python/az/aro/build/lib/azext_aro/tests/latest/unit/test_validators.py:827:7: W291 trailing whitespace
4     W291 trailing whitespace
```

python/az/aro/azext_aro/tests/latest/unit/test_validators.py
slawande2: Thank you, Caden. Don't know why it wasn't caught on my machine.

cadenmarchese (approved):

shubhadapaithankar (approved):

kimorris27 (approved):

PR #3845, opened 2024-09-16
[SFI] Fix CVE-2024-41110

nwnt:

Fixes ARO-9396

The current version of docker has a vulnerability.  It needs to be upgraded to its latest patch version.

### Test plan for issue:

E2E

N/A

With E2E passes, it should be ok.


jaitaiwan (approved):
LGTM. Pending E2E and CI

PR #3847, opened 2024-09-17
Generate Federated Identity Credentials for MIWI Cluster

gouthamMN:

Fixes: https://issues.redhat.com/browse/ARO-4375

This PR perform the followings:
1. Add/Create Role Assignments on each customer provided Platform Workload Identities over Managed Resource Group and Vnet Resource Group as mentioned in [MIWI Test Matrix doc](https://docs.google.com/document/d/1Q0Xw5QVKkVEjTcX_VfX43GaujFDip-nrSM4TUCblvhs/edit#heading=h.iwzh6ck2qjrx) by reading the respective role definitions from `PlatformWorkloadIdentityRoleSetProperties.PlatformWorkloadIdentityRole.RoleDefinitionID`.
3. Create `FederatedIdentityCredentials` for each customer provided Platform Workload Identities during cluster install/create for each Service accounts in `PlatformWorkloadIdentityRoleSetProperties.PlatformWorkloadIdentityRole.ServiceAccounts`.
4. Delete/Cleanup `FederatedIdentityCredentials` for each customer provided Platform Workload Identities that were created during cluster delete.

### Test plan for issue:
- New Unit tests are added.
- e2e

No

- If the e2e pass without errors


pkg/cluster/deploybaseresources.go
```diff
@@ -444,6 +454,60 @@ func (m *manager) setMasterSubnetPolicies(ctx context.Context) error {
...
+	platformWorkloadIdentities := m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities
+
+	for _, identity := range platformWorkloadIdentities {
+		identityResourceId, err := azure.ParseResourceID(identity.ResourceID)
+		if err != nil {
+			return err
+		}
+
+		platformWIRole, exists := platformWIRolesByRoleName[identity.OperatorName]
+		if !exists {
```
rajdeepc2792, line 478: There can be some managed identities in  platformIdentities list(`m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities`) that won't be present in `m.platformWorkloadIdentityRolesByVersion.GetPlatformWorkloadIdentityRolesByRoleName()`. 

The `platformWIRolesByRoleName` should be used to confirm what identities are need to be federated, thus here whatever `exists` only those should be federated.

pkg/cluster/delete.go
```diff
@@ -356,6 +358,73 @@ func (m *manager) deleteGateway(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+
+		federatedIdentityCredentialResourceName, err := m.getPlatformWorkloadIdentityFedertedCredName(identity)
+		if err != nil {
+			return err
+		}
+
+		_, err = m.clusterMsiFederatedIdentityCredentials.Get(
```
rajdeepc2792, line 399: Similar to creation of federated credentials, only get and delete the federated credentials present in `platformWIRolesByRoleName`.

pkg/cluster/deploybaseresources_additional.go
```diff
@@ -92,6 +93,81 @@ func (m *manager) clusterServicePrincipalRBAC() *arm.Resource {
...
+		return nil, nil
+	}
+
+	resources := []*arm.Resource{}
+	platformWIRolesByRoleName := m.platformWorkloadIdentityRolesByVersion.GetPlatformWorkloadIdentityRolesByRoleName()
+	platformWorkloadIdentities := m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities
+
+	for _, identity := range platformWorkloadIdentities {
+		role, exists := platformWIRolesByRoleName[identity.OperatorName]
+		if !exists {
```
rajdeepc2792, line 138: Similar to federated creation, only need to perform RBAC for identities those exists in `platformWIRolesByRoleName`.

pkg/cluster/deploybaseresources_additional.go
```diff
@@ -92,6 +93,81 @@ func (m *manager) clusterServicePrincipalRBAC() *arm.Resource {
...
+					Message: "OperatorName " + identity.OperatorName + " not found in platform workload identity roles",
+				},
+			}
+		}
+
+		roleID := stringutils.LastTokenByte(role.RoleDefinitionID, '/')
+		resources = append(resources, m.workloadIdentityResourceGroupRBAC(roleID, identity.ObjectID, managedRG))
+
+		if role.OperatorName != "StorageOperator" {
+			resources = append(resources, m.workloadIdentityResourceGroupRBAC(roleID, identity.ObjectID, vnetRG))
```
rajdeepc2792, line 152: Why are we handling the RBAC for vnetRG? The RBAC is only needed for the managedRG, FPSP won't have permissions for role assignment on vnetRG. And for managedRG, all the identities existing in `platformWIRolesByRoleName` will have a role assignment performed for their corresponding roles.

pkg/cluster/install.go
```diff
@@ -289,34 +289,58 @@ func setFieldCreatedByHive(createdByHive bool) database.OpenShiftClusterDocument
...
+	s = append(s,
 		steps.Action(m.ensureResourceGroup),
 		steps.Action(m.ensureServiceEndpoints),
 		steps.Action(m.setMasterSubnetPolicies),
 		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.deployBaseResourceTemplate),
+	)
+
+	if m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		s = append(s,
+			steps.Action(m.generateFederatedIdentityCredentials),
```
rajdeepc2792, line 332: Not sure if the `FederatedIdentityCredentials` ensure need to be part of this PR or not. But for update flow, the credentials will need to be ensure such that they exists and with correct values of issuer, audience etc.

rajdeepc2792 (dismissed):
Thanks for the PR!!
Performed a quick review of the PR, added some comments for the federated credentials creation/deletion and RBAC part.
The review would be easier to do after the MSI PR is merged.

pkg/cluster/deploybaseresources_additional.go
gouthamMN: Ah, so if I understand correctly, we don't throw an error instead ` if !exists { continue }` ?

pkg/cluster/deploybaseresources.go
gouthamMN: so, if I understand correctly, we don't throw an error instead `if !exists { continue }` ?

pkg/cluster/deploybaseresources_additional.go
rajdeepc2792: Yes exactly, platformWIRolesByRoleName is just to make sure we federate and RBAC only for the desired platformIdentities.

pkg/cluster/deploybaseresources_additional.go
rajdeepc2792: Yes, the VNET is to be handled by customer, and that check will already done till this point as part of dynamic validation.

hack/devtools/local_dev_env.sh
```diff
@@ -189,10 +189,8 @@ create_platform_identity_and_assign_role() {
     echo "Name: $name"
     echo ""
 
-    # The following operators require a role assignment since they need access to customer BYO virtual network
-    # RP will be responsible for other role assignments
-    if [[ "${operatorName}" == "MachineApiOperator" || "${operatorName}" == "NetworkOperator" \
-        || "${operatorName}" == "AzureFilesStorageOperator" || "${operatorName}" == "ServiceOperator" ]]; then
+    # Storage Operator don't require access to customer BYO virtual network
+    if [[ "${operatorName}" != "StorageOperator" ]]; then
```
niontive, line 193: What about image registry operator? or cloud controller manager?

pkg/cluster/delete.go
```diff
@@ -484,13 +531,17 @@ func (m *manager) Delete(ctx context.Context) error {
...
 	}
-
 	if m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
 		m.log.Printf("deleting cluster MSI certificate")
 		err = m.deleteClusterMsiCertificate(ctx)
 		if err != nil {
 			return err
 		}
+
+		err = m.deleteFederatedCredentials(ctx)
```
niontive, line 541: Nit: we can call this "deleteIdentityFederations"

pkg/cluster/workloadidentityresources.go
```diff
@@ -124,3 +126,15 @@ func (m *manager) generateAuthenticationConfig() (*configv1.Authentication, erro
...
+func (m *manager) getPlatformWorkloadIdentityFedertedCredName(identity api.PlatformWorkloadIdentity) (string, error) {
+	if !m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		return "", fmt.Errorf("getPlatformWorkloadIdentityFedertedCredName called for a CSP cluster")
+	}
+
+	identityResourceId, err := azure.ParseResourceID(identity.ResourceID)
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%s-%s", m.doc.ID, identityResourceId.ResourceName), nil
```
niontive, line 139: Any justification for this naming convention?

Want to make sure it fits within the requirements for Azure resource naming. For example, how long is `doc.ID`?

pkg/cluster/deploybaseresources.go
```diff
@@ -444,6 +454,60 @@ func (m *manager) setMasterSubnetPolicies(ctx context.Context) error {
 	return err
 }
 
+func (m *manager) generateFederatedIdentityCredentials(ctx context.Context) error {
```
niontive, line 457: Nit: I'd call this `federateIdentityCredentials` - we aren't "generating" anything per say; we are "federating" the identities.

pkg/cluster/deploybaseresources.go
```diff
@@ -444,6 +454,60 @@ func (m *manager) setMasterSubnetPolicies(ctx context.Context) error {
...
+	if !m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		return nil
+	}
+
+	if m.doc.OpenShiftCluster.Properties.ClusterProfile.OIDCIssuer == nil {
+		return fmt.Errorf("OIDCIssuer is nil for the cluster with ID: %s", m.doc.OpenShiftCluster.ID)
+	}
+
+	Issuer := to.StringPtr((string)(*m.doc.OpenShiftCluster.Properties.ClusterProfile.OIDCIssuer))
```
niontive, line 466: nit: `issuer := ...`

pkg/cluster/deploybaseresources.go
```diff
@@ -444,6 +454,60 @@ func (m *manager) setMasterSubnetPolicies(ctx context.Context) error {
...
+		}
+
+		for _, sa := range platformWIRole.ServiceAccounts {
+			_, err := m.clusterMsiFederatedIdentityCredentials.CreateOrUpdate(
+				ctx,
+				identityResourceId.ResourceGroup,
+				identityResourceId.ResourceName,
+				federatedIdentityCredentialResourceName,
+				armmsi.FederatedIdentityCredential{
+					Properties: &armmsi.FederatedIdentityCredentialProperties{
```
niontive, line 494: See this as an example for setting properties: https://azure.github.io/azure-workload-identity/docs/quick-start.html#4-create-an-aad-application-or-user-assigned-managed-identity-and-grant-permissions-to-access-the-secret

pkg/cluster/delete.go
gouthamMN: updated

pkg/cluster/deploybaseresources.go
gouthamMN: updated

pkg/cluster/deploybaseresources_additional.go
gouthamMN: updated

pkg/cluster/workloadidentityresources.go
rajdeepc2792: https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation-considerations#errors gives an idea about the naming.
Also when creating federated at portal:-
`The name of a federated identity credentials must be within 3 and 120 characters, only contains letters (A-Z, a-z), numbers, hyphens and dashes and must start with a number or letter.`

hack/devtools/local_dev_env.sh
gouthamMN: @cadenmarchese mentioned except `StorageOperator` all other operators would need access to Vnet as per https://issues.redhat.com/browse/OCPBUGS-28762

pkg/cluster/deploybaseresources.go
gouthamMN: Anything specific you want me to look for? If you're talking about the way we set `Audiences: openshift`,[ here is the slack conversation details](https://redhat-external.slack.com/archives/C06FRQKUMFF/p1723242301968319?thread_ts=1723242231.734469&cid=C06FRQKUMFF) why we set it to `openshift` instaed of `"api://AzureADTokenExchange"`

pkg/cluster/deploybaseresources.go
gouthamMN: updated

pkg/cluster/deploybaseresources.go
gouthamMN: updated

pkg/cluster/delete.go
gouthamMN: updated

pkg/cluster/install.go
gouthamMN: @cadenmarchese - can you confirm this? I didn't that in the [JIRA item](https://issues.redhat.com/browse/ARO-4375).

pkg/cluster/workloadidentityresources.go
gouthamMN: Given that we are constructing a name with `m.doc.ID` and `identityResourceId.ResourceName`. I think we could assume all those conditions are satisfied and no need to add logic to protect against that?

pkg/cluster/deploybaseresources_additional.go
```diff
@@ -92,6 +92,65 @@ func (m *manager) clusterServicePrincipalRBAC() *arm.Resource {
...
+	r, err := m.platoformWorkloadIdentityRBAC(managedRG)
+	if err != nil {
+		return nil, err
+	}
+
+	resources = append(resources, r...)
+	return resources, nil
+}
+
+func (m *manager) platoformWorkloadIdentityRBAC(managedRG string) ([]*arm.Resource, error) {
```
cadenmarchese, line 120: nit - typo:
```suggestion
func (m *manager) platformWorkloadIdentityRBAC(managedRG string) ([]*arm.Resource, error) {
```

pkg/cluster/install.go
cadenmarchese: @gouthamMN @rajdeepc2792 I am tempted to say that the update flow / ensure flow makes more sense under https://issues.redhat.com/browse/ARO-4371 rather than here. I have added it there. According to my understanding, we just need to ensure that federated credentials for all identities 1) exist 2) have correct issuer 3) have correct audience 4) have correct service account (?). That would cover update case as well as the create case.

hack/devtools/local_dev_env.sh
rajdeepc2792: For the GA, since CLI isn't creating the identities, we also assume that the customer running the create cluster won't be having the permission to assign RBAC over the vnet.
Thus it is being assumed that the provided identities to the CLI during cluster create/update will have the RBAC already performed.

cadenmarchese (changes requested):
Now that https://github.com/Azure/ARO-RP/pull/3757 has merged, can we rebase here and change this PR to target `master`?

gouthamMN:
> Now that #3757 has merged, can we rebase here and change this PR to target `master`?

Updated

pkg/cluster/deploybaseresources_additional.go
gouthamMN: updated

pkg/cluster/delete.go
```diff
@@ -377,6 +378,58 @@ func (m *manager) deleteClusterMsiCertificate(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+
+		federatedIdentityCredentialResourceName, err := m.getPlatformWorkloadIdentityFederatedCredName(identity)
+		if err != nil {
+			return err
+		}
+
+		_, err = m.clusterMsiFederatedIdentityCredentials.Get(
```
kimorris27, line 404: I tried testing the PR locally, but I get a nil pointer dereference on this line because the fed creds client isn't initialized. I think we'll need to:

1. Make sure the cluster MSI clients are initialized before trying to use them during deletion
2. Additionally, account for the case where the cluster MSI is missing and we're therefore unable to delete the fed creds. We probably don't want to block cluster deletion on this but rather log a warning message saying we did what we could (i.e.: `Unable to delete platform workload identities' federated credentials because cluster MSI has already been deleted; continuing with rest of cluster deletion`) and then move forward.

@cadenmarchese, what do you think about #2? Does what I suggested sound good?

pkg/cluster/deploybaseresources.go
```diff
@@ -444,6 +454,60 @@ func (m *manager) setMasterSubnetPolicies(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+
+		for _, sa := range platformWIRole.ServiceAccounts {
+			_, err := m.clusterMsiFederatedIdentityCredentials.CreateOrUpdate(
+				ctx,
+				identityResourceId.ResourceGroup,
+				identityResourceId.ResourceName,
+				federatedIdentityCredentialResourceName,
```
kimorris27, line 492: This looks to me like it's going to overwrite the same fed cred repeatedly rather than creating multiple fed creds when a platform MI has more than one SA. Maybe the SA name should be incorporated into the fed cred name?

kimorris27 (changes requested):
I tried this out locally and ran into some blocking issues. I'll revisit the PR and do a full review once these initial comments are addressed.

pkg/cluster/delete.go
gouthamMN: @kimorris27 - for point 1, should we error out if the Fed Cred client is `nil`? Also, is this required, as we already error out during `initializeClusterMsiClients` as [here](https://github.com/Azure/ARO-RP/blob/84816de0e43d611efdfbfa1aef1bf7f40f4bedf8/pkg/cluster/clustermsi.go#L121) if the client is not created.

pkg/cluster/deploybaseresources.go
gouthamMN: I think it shouldn't overwrite, [federate-multiple-kubernetes-service-accounts-with-a-single-identity](https://azure.github.io/azure-workload-identity/docs/faq.html#how-to-federate-multiple-kubernetes-service-accounts-with-a-single-identity)

pkg/cluster/deploybaseresources.go
```diff
@@ -444,6 +454,60 @@ func (m *manager) setMasterSubnetPolicies(ctx context.Context) error {
 	return err
 }
 
+func (m *manager) federateIdentityCredentials(ctx context.Context) error {
+	if !m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		return nil
+	}
+
+	if m.doc.OpenShiftCluster.Properties.ClusterProfile.OIDCIssuer == nil {
+		return fmt.Errorf("OIDCIssuer is nil for the cluster with ID: %s", m.doc.OpenShiftCluster.ID)
```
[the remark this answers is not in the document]
gouthamMN: updated

pkg/cluster/delete.go
kimorris27: That function currently isn't getting called at all during deletion, but to your point, I think we will need to initialize the fed creds client differently during deletion, i.e. create it if possible and log a warning and continue with cluster deletion if not possible.

When an RP backend goroutine picks up a cluster whose state has changed to `Deleting`, it goes straight to the `Delete` function without following any of the other "steps", as you can see [here](https://github.com/Azure/ARO-RP/blob/84816de0e43d611efdfbfa1aef1bf7f40f4bedf8/pkg/backend/openshiftcluster.go#L186). So I think the logic for the fed creds client initialization that I described in the previous paragraph might need to be incorporated here.

pkg/cluster/deploybaseresources.go
```diff
@@ -200,6 +202,15 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
 		m.networkInternalLoadBalancer(azureRegion),
 	}
 
+	if m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
```
kimorris27, line 205: In addition to adding the new WI resources, we need to skip `clusterServicePrincipalRBAC` (see a few lines above this). It caused another nil pointer dereference in my local testing.

pkg/cluster/deploybaseresources.go
kimorris27: You can definitely create multiple fed creds using the same MI, but you're going to need to give each fed cred resource a unique name.

kimorris27 (changes requested):
I commented out `validateResources` to get past the issue we were discussing in Slack and further my local testing, but I hit an error with the base resource template deployment that I think you'll need to reproduce and fix.

```
Message: Deployment failed.
Exception Details:	(None) {"code":"InvalidTemplate","message":"Deployment template validation failed: 'The template resource '[concat(aro-kimorris, '/Microsoft.Authorization/', guid(resourceId('Microsoft.Resources/resourceGroups', aro-kimorris), , 'ef318e2a-8334-4a05-9e4a-295a196c6a6e'))]' at line '1' and column '8651' is not valid: The language expression 'concat(aro-kimorris, '/Microsoft.Authorization/', guid(resourceId('Microsoft.Resources/resourceGroups', aro-kimorris), , 'ef318e2a-8334-4a05-9e4a-295a196c6a6e'))' is not valid: the string character 'k' at position '11' is not expected.. Please see https://aka.ms/arm-functions for usage details.'.","target":null,"details":null,"innererror":null,"additionalInfo":[{"info":{"lineNumber":1,"linePosition":8651,"path":"properties.template.resources[8]"},"type":"TemplateViolation"}]}
```

pkg/cluster/delete.go
gouthamMN: @kimorris27 - Was there any reason we didn't initialize the Fed Cred client when creating a [New cluster manager](https://github.com/Azure/ARO-RP/blob/84816de0e43d611efdfbfa1aef1bf7f40f4bedf8/pkg/cluster/cluster.go#L316)? Just wanted to understand this before making the changes?

pkg/cluster/deploybaseresources.go
gouthamMN: Given that we are limited to 120 characters, I will let @cadenmarchese @rajdeepc2792 and @kimorris27 to suggest the naming convention to be maintained :) before I make any changes. My suggestion would look something like below
```
sa := strings.Split(ServiceAccount, ":")
saSecondPart := sa[1] //second part of service account

return fmt.Sprintf("%s-%s", m.doc.ID, concat(saSecondPart, identityResourceId.ResourceName)[:83])
// 36 chars from doc.ID 
// +1 from "-"
// Remaining 83 comes from first 83 chars of concat result of saSecondPart & ResourceName
```

gouthamMN:
> I commented out `validateResources` to get past the issue we were discussing in Slack and further my local testing, but I hit an error with the base resource template deployment that I think you'll need to reproduce and fix.
> 
> ```
> Message: Deployment failed.
> Exception Details:	(None) {"code":"InvalidTemplate","message":"Deployment template validation failed: 'The template resource '[concat(aro-kimorris, '/Microsoft.Authorization/', guid(resourceId('Microsoft.Resources/resourceGroups', aro-kimorris), , 'ef318e2a-8334-4a05-9e4a-295a196c6a6e'))]' at line '1' and column '8651' is not valid: The language expression 'concat(aro-kimorris, '/Microsoft.Authorization/', guid(resourceId('Microsoft.Resources/resourceGroups', aro-kimorris), , 'ef318e2a-8334-4a05-9e4a-295a196c6a6e'))' is not valid: the string character 'k' at position '11' is not expected.. Please see https://aka.ms/arm-functions for usage details.'.","target":null,"details":null,"innererror":null,"additionalInfo":[{"info":{"lineNumber":1,"linePosition":8651,"path":"properties.template.resources[8]"},"type":"TemplateViolation"}]}
> ```

Can you share the complete template that's throwing this error. What I noticed from the above error is that the object ID is empty in your deployment.

pkg/cluster/delete.go
cadenmarchese: @kimorris27 as long as the customer can clean up resources on their own, I don't think we need to block deletion on failure to clean up federated credentials, so your suggestion sounds good to me.

pkg/cluster/deploybaseresources.go
cadenmarchese: What's the rationale behind using the doc ID instead of the cluster name / resource name?

stevekuznetsov: Is `ServiceAccount` a k8s SA? e.g. `system:serviceaccount:namespace:name`?

cadenmarchese: @stevekuznetsov yes, but in this case it's in namespace:name format only: https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/RP-Config?path=/deploy/prod-config.yaml&version=GBmaster&line=2964&lineEnd=2965&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents 

So an example would be `openshift-cloud-controller-manager:cloud-controller-manager`

pkg/cluster/delete.go
gouthamMN: Addressed both point 1 & 2.

pkg/cluster/delete_test.go
```diff
@@ -486,3 +490,207 @@ func TestDeleteClusterMsiCertificate(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		err = pir.PopulatePlatformWorkloadIdentityRolesByVersion(ctx, tt.doc.OpenShiftCluster, dbPlatformWorkloadIdentityRoleSets)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		t.Run(tt.name, func(t *testing.T) {
+			controller := gomock.NewController(t)
```
tsatam, line 682: Is this controller unused?

pkg/cluster/deploybaseresources_additional.go
```diff
@@ -92,6 +92,65 @@ func (m *manager) clusterServicePrincipalRBAC() *arm.Resource {
...
+func (m *manager) ensureWorkloadIdentityRBAC() ([]*arm.Resource, error) {
+	if !m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		return nil, nil
+	}
+
+	clusterMSIResourceId, err := m.doc.OpenShiftCluster.ClusterMsiResourceId()
+	if err != nil {
+		return nil, err
+	}
+	clusterMSI := m.doc.OpenShiftCluster.Identity.UserAssignedIdentities[clusterMSIResourceId.String()]
```
tsatam, line 104: nit: this might fall victim to case sensitivity issues. `ClusterMsiResourceId` returns a formatted `arm.ResourceId` which coerces the original string into the "correct" casing. 

We may end up in a situation where a customer-provided clusterMSI id with incorrect casing (e.g. `resourcegroups` instead of `resourceGroups`) causes line 104 to fail to get the identity from the map.

pkg/cluster/deploybaseresources_additional.go
```diff
@@ -92,6 +92,72 @@ func (m *manager) clusterServicePrincipalRBAC() *arm.Resource {
...
+	if strings.TrimSpace(clusterMSI.PrincipalID) == "" {
+		return nil, fmt.Errorf("cluster MSI principal ID '%s' is invalid for clusterMSIResourceId %s", clusterMSI.PrincipalID, clusterMSIResourceId.String())
+	}
+
+	resources := []*arm.Resource{}
+	managedRG := stringutils.LastTokenByte(m.doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID, '/')
+
+	resources = append(resources, m.workloadIdentityResourceGroupRBAC(rbac.RoleAzureRedHatOpenShiftFederatedCredentialRole, clusterMSI.PrincipalID, managedRG))
```
rajdeepc2792, line 112: There is no need to assign rbac.RoleAzureRedHatOpenShiftFederatedCredentialRole on the managed RG. This role will be provided on Cluster MIS with scope PlatformIdentities by customer.

tsatam (commented):
Mostly LGTM - there's some minor nits and fixups that don't necessarily have to be addressed, and one concern about handling case-insensitivity for resourceIDs for managed identities in maps.

pkg/cluster/deploybaseresources.go
stevekuznetsov: OK, please consider that the name of the SA is allowed to be 246 characters long - account for this in your algorithm. Consider base32(sha224(inputs)) as a simple way to derive a short name deterministically without having cascades of length adjustment code.

gouthamMN: @stevekuznetsov - In the case where we aren't able to clean up the federated Identity during deletion process as [mentioned here](https://github.com/Azure/ARO-RP/pull/3847#discussion_r1779115235), how will the customer be able to find the Hashed name [base32(sha224(inputs))] to clean up identities on their own? Should the customer list all the federated identities under a given User Assigned Identity and delete them?
cc - @cadenmarchese @kimorris27 @rajdeepc2792

pkg/cluster/deploybaseresources_additional.go
gouthamMN: @rajdeepc2792 - Do you mean `rbac.RoleAzureRedHatOpenShiftFederatedCredentialRole` will be added by customer to Cluster MSI and we don't have to perform this?
cc - @cadenmarchese @kimorris27 @niontive

pkg/cluster/deploybaseresources.go
```diff
@@ -209,8 +206,16 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
 		}
 
 		resources = append(resources, r...)
+	} else {
+		resources = append(resources, m.clusterServicePrincipalRBAC())
 	}
 
+	resources = append(
```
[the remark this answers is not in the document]
gouthamMN: I thought of maintaining the same old order to ensure nothing breaks. I was also not sure if we had to have the RBAC performed on the Cluster MSI/SP before deploying these network resources. But from the ARM resource template I don't think these depends on Cluster MSI/SP RBAC.

pkg/cluster/deploybaseresources_additional.go
gouthamMN: updated

pkg/cluster/delete_test.go
gouthamMN: updated

pkg/cluster/deploybaseresources.go
gouthamMN: updated

pkg/cluster/deploybaseresources_additional.go
gouthamMN: updated

cadenmarchese (dismissed):

pkg/cluster/delete.go
```diff
@@ -377,6 +378,77 @@ func (m *manager) deleteClusterMsiCertificate(ctx context.Context) error {
...
+	if m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile == nil {
+		return nil
+	}
+
+	if m.clusterMsiFederatedIdentityCredentials == nil {
+		m.log.Warning("cluster MSI federated identity credentials client is nil, trying to initialize")
+		err := m.initializeClusterMsiClients(ctx)
+		if err != nil {
+			m.log.Errorf("cluster MSI federated identity credentials client initialization failed with error: %v", err)
+			return nil
```
niontive, line 391: Why are returning nil here? I'd expect this to return `err`, and whoever is calling `deleteIdentityFederations` can decide what to do with that error. Example: log the error with "unable to delete due to client init failure, will continue...".

pkg/cluster/workloadidentityresources.go
```diff
@@ -124,3 +130,37 @@ func (m *manager) generateAuthenticationConfig() (*configv1.Authentication, erro
...
+	clusterResourceId, err := azure.ParseResourceID(m.doc.OpenShiftCluster.ID)
+	if err != nil {
+		return "", err
+	}
+
+	clusterResourceKey := fmt.Sprintf("%s-%s-%s", clusterResourceId.SubscriptionID, clusterResourceId.ResourceGroup, clusterResourceId.ResourceName)
+	name := fmt.Sprintf("%s-%s-%s", clusterResourceKey, sa, identityResourceId.ResourceName)
+	// the base-36 encoded string of a SHA-224 hash will typically be around 43 to 44 characters long.
+	hash := sha256.Sum224([]byte(name))
+	encodedName := (&big.Int{}).SetBytes(hash[:]).Text(36)
```
niontive, line 158: Nit: please use constants for hardcoded numbers

pkg/cluster/workloadidentityresources.go
```diff
@@ -124,3 +130,37 @@ func (m *manager) generateAuthenticationConfig() (*configv1.Authentication, erro
...
+	identityResourceId, err := azure.ParseResourceID(identity.ResourceID)
+	if err != nil {
+		return "", err
+	}
+	clusterResourceId, err := azure.ParseResourceID(m.doc.OpenShiftCluster.ID)
+	if err != nil {
+		return "", err
+	}
+
+	clusterResourceKey := fmt.Sprintf("%s-%s-%s", clusterResourceId.SubscriptionID, clusterResourceId.ResourceGroup, clusterResourceId.ResourceName)
```
niontive, line 154: Why do we need this complicated calculation?

If you want to be safe, just do:
- hash the SA name w/ SHA1
- hash the identity resource URI w/ SHA1
- return `sa-hash`-`resourceURI-hash`

This way we can perform additional service account federations if needed in the future.

Re collisions: https://stackoverflow.com/questions/4014090/is-it-safe-to-ignore-the-possibility-of-sha-collisions-in-practice

pkg/cluster/delete.go
gouthamMN: As per the previous discussions [here](https://github.com/Azure/ARO-RP/pull/3847#discussion_r1779260847) we don't want to block the resource cleanup process if we aren't able to delete the Fed cred Identity. As of now, we return `error` when we get error parsing a resource ID of the Identity. 

Today this function is called only during resource cleanup from `Delete()` func under `delete.go`

pkg/cluster/workloadidentityresources.go
gouthamMN: updated

kimorris27 (changes requested):
I tried testing once more, and there seems to be a different syntax error in the base resource deployment template. We'll have to test and debug with a local dev RP.

```
{"code":"InvalidTemplate","message":"Deployment template validation failed: 'The template resource '['redacted-guid']' at line '1' and column '8651' is not valid: Evaluation result of language expression 'guid(resourceGroup().id, 'redacted-guid')' is type 'String', expected type is 'Boolean'.. Please see https://aka.ms/arm-functions for usage details.'.","target":null,"details":null,"innererror":null,"additionalInfo":[{"info":{"lineNumber":1,"linePosition":8651,"path":"properties.template.resources[8]"},"type":"TemplateViolation"}]}
```

gouthamMN:
> I tried testing once more, and there seems to be a different syntax error in the base resource deployment template. We'll have to test and debug with a local dev RP.
> 
> ```
> {"code":"InvalidTemplate","message":"Deployment template validation failed: 'The template resource '['redacted-guid']' at line '1' and column '8651' is not valid: Evaluation result of language expression 'guid(resourceGroup().id, 'redacted-guid')' is type 'String', expected type is 'Boolean'.. Please see https://aka.ms/arm-functions for usage details.'.","target":null,"details":null,"innererror":null,"additionalInfo":[{"info":{"lineNumber":1,"linePosition":8651,"path":"properties.template.resources[8]"},"type":"TemplateViolation"}]}
> ```

fixed now

pkg/cluster/delete.go
```diff
@@ -377,6 +378,77 @@ func (m *manager) deleteClusterMsiCertificate(ctx context.Context) error {
...
+
+		platformWIRole, exists := platformWIRolesByRoleName[identity.OperatorName]
+		if !exists {
+			continue
+		}
+
+		for _, sa := range platformWIRole.ServiceAccounts {
+			federatedIdentityCredentialResourceName, err := m.getPlatformWorkloadIdentityFederatedCredName(sa, identity)
+			if err != nil {
+				m.log.Errorf("failed to get federated identity credential name for %s: %v", identity.ResourceID, err)
```
kimorris27, line 417: This is an error on the RP side and not related to the cluster MSI having been deleted by the cx, so I think we should error out here rather than logging the error and continuing.

pkg/cluster/delete.go
```diff
@@ -377,6 +378,77 @@ func (m *manager) deleteClusterMsiCertificate(ctx context.Context) error {
...
+				identityResourceId.ResourceGroup,
+				identityResourceId.ResourceName,
+				federatedIdentityCredentialResourceName,
+				&armmsi.FederatedIdentityCredentialsClientGetOptions{},
+			)
+
+			// If the federated identity credentials do not exist, we can skip the deletion
+			if err != nil {
+				continue
+			}
```
kimorris27, line 432: This is an error on the RP side and not related to the cluster MSI having been deleted by the cx, so I think we should error out here rather than logging the error and continuing.

Additionally, we can get more granular than `if err != nil` here:

1. If error is a 404, skip deletion and continue
2. If error is non-nil and anything other than a 404, error out

pkg/cluster/delete.go
```diff
@@ -377,6 +378,77 @@ func (m *manager) deleteClusterMsiCertificate(ctx context.Context) error {
...
+				identityResourceId.ResourceGroup,
+				identityResourceId.ResourceName,
+				federatedIdentityCredentialResourceName,
+				&armmsi.FederatedIdentityCredentialsClientDeleteOptions{},
+			)
+
+			// As long as the customer can clean up resources on their own, we don't block deletion on failure to clean up federated credentials
+			if err != nil {
+				m.log.Errorf("failed to delete federated identity credentials for %s: %v", identity.ResourceID, err)
+			}
```
kimorris27, line 445: This is an error on the RP side and not related to the cluster MSI having been deleted by the cx, so I think we should error out here rather than logging the error and continuing.

Also, couldn't we skip the `GET` call right above this and simply rework this such that we continue if the `DELETE` results in a 404? We should be able to get the same result with one fewer API call.

pkg/cluster/deploybaseresources_additional.go
kimorris27: What about the case where there should be an entry in `platformWIRolesByRoleName` but we don't find it? Maybe we should error out in that case.

test/util/azure/msi/armmsi.go
kimorris27, line 1: I don't think we need to manually maintain a fake client here. We should be able to get rid of this and use an autogenerated client once https://github.com/Azure/ARO-RP/pull/3899 merges.

pkg/cluster/delete.go
kimorris27: We could encounter an error here for one of two reasons:

1. The cx deleted their cluster MSI; we definitely want to log the error and return nil in this case
2. There's some other RP-side error; we probably want to error out in this case

kimorris27 (changes requested):
I pushed a few small changes after trying this out locally, since @gouthamMN 's still blocked from running a local dev RP. The main change I pushed was moving fed cred deletion to be before cluster MSI cert deletion because we need to use the MSI to delete the fed creds.

I think we need to revisit the discussion about whether to error out on fed cred deletion failures though. I went back and looked at @cadenmarchese 's comment [here](https://github.com/Azure/ARO-RP/pull/3847#discussion_r1783151290), and I think you all may have misunderstood what I said earlier in the conversation.

I wasn't suggesting we log the error and continue for _any_ error that occurs during the process of trying to delete the federated credentials; I was only suggesting we do that in the case where we can't initialize the federated credentials client because the cx deleted their MSI. In other cases where the error is the service's fault (ex: the MI RP returns an internal server error while we're trying to delete a fed cred), I think we should still error out.

I left my thoughts here rather than directly making the changes so that others can confirm that that's the route we want to go. Once we have a consensus, I'm happy to make the changes and test out locally if needed.

tsatam (changes requested):
This PR should be rebased on latest master, as the merge of this PR will introduce some duplicated resources into `pkg/cluster/deploybaseresources.go` that will need to be manually removed.

test/util/azure/msi/armmsi.go
gouthamMN: Wasn't this already decided [here under 3757](https://github.com/Azure/ARO-RP/pull/3757#issuecomment-2284517423) to use arm fake clients?

pkg/cluster/delete.go
gouthamMN: updated

pkg/cluster/delete.go
gouthamMN: updated

pkg/cluster/delete.go
gouthamMN: updated

gouthamMN:
> This PR should be rebased on latest master, as the merge of this PR will introduce some duplicated resources into `pkg/cluster/deploybaseresources.go` that will need to be manually removed.

rebased

pkg/cluster/deploybaseresources_additional.go
gouthamMN: @rajdeepc2792

test/util/azure/msi/armmsi.go
kimorris27: Apparently I gave conflicting suggestions in different places... I'm sorry about that 🙈

At the time I made that comment on #3757, I didn't realize that using the fake client would require us to "manually" write some code to create the client - I was thinking we could just import a fake client that's ready to use without requiring that additional maintenance effort.

With that in mind, I'm now thinking that it'd be better to stick with the generated clients. Those have their own pros and cons, but it seems to me that the major pro is less maintenance effort.

I'd like for others to weigh in with opinions though. @bitoku maybe?

pkg/cluster/delete.go
```diff
@@ -438,10 +424,15 @@ func (m *manager) deleteFederatedCredentials(ctx context.Context) error {
 				federatedIdentityCredentialResourceName,
 				&armmsi.FederatedIdentityCredentialsClientDeleteOptions{},
 			)
-
-			// As long as the customer can clean up resources on their own, we don't block deletion on failure to clean up federated credentials
 			if err != nil {
-				m.log.Errorf("failed to delete federated identity credentials for %s: %v", identity.ResourceID, err)
+				cloudErr := err.(*api.CloudError)
+
+				if cloudErr.StatusCode != http.StatusNotFound {
```
kimorris27, line 430: Should this be `==` instead of `!=`?

kimorris27 (commented):
One last quick question before we merge with the understanding that some of the other outstanding feedback can be addressed in a follow-up PR.

pkg/cluster/delete.go
gouthamMN: That's right, not sure where my brain was at that time :)

gouthamMN:
> One last quick question before we merge with the understanding that some of the other outstanding feedback can be addressed in a follow-up PR.

Yup, I think once we come to a conclusion about how the Fed Cred Deletion we should work that under a new JIRA item.

kimorris27 (approved):
LGTM with the understanding that we can finalize the mock client issue and the fed cred deletion issue in a follow-up conversation/PR.

test/util/azure/msi/armmsi.go
gouthamMN: As per the discussion, this will be addressed in the new PR under a new JIRA once we come to final conclusion.

pkg/cluster/deploybaseresources_additional.go
gouthamMN: As per the discussion, this will be addressed in the new PR under a new JIRA once we come to final conclusion.

tsatam (approved):

cadenmarchese (approved):

PR #3849, opened 2024-09-18
Idea: Quick RP compile build target

SudoBrendan:

Fixes


SudoBrendan:
cc @bitoku and @jaitaiwan

jaitaiwan:
LGTM, just needs doco around making sure you've got a build of portal avail as well. Likely needs something similar for portal I guess.

PR #3823, opened 2024-09-06
Clean up the remainder of `go run` uses of external tools

hawkowl:
### What this PR does / why we need it:

Cleans up the last bits where we're using go run for external code, or go run ./hack/<some tool with import> like gojq.

### Test plan for issue:

CI, E2E

### Is there any documentation that needs to be updated for this PR?
N/A

### How do you know this will function as expected in production? 
N/A

PR #3850, opened 2024-09-19
Fix cleaning the gratis subscription

hawkowl:
Fixes subscription cleaning that was broken in #3719

### DIFF
--- a/.pipelines/clean-subscription.yml
+++ b/.pipelines/clean-subscription.yml
@@ -22,11 +22,6 @@ jobs:
 
     steps:
       - template: ./templates/template-checkout.yml
-      - script: |
-          set -xe
-          sudo dnf install -y jq
-        displayName: Setup (Container)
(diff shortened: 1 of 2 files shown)


bitoku:
https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=103609613&view=logs&j=a4f1910f-c367-5697-edcd-724d81cde23b&t=bc103e47-9b88-5c56-fcad-f3f2f0b2ed91

Validated the code with this pipeline.

bitoku (approved):

kimorris27 (approved):
I'm curious why it broke in the first place, but the fix looks good 👍🏻 

Thanks for the validation @bitoku!

bitoku:
@kimorris27 
The only cause I can think of is https://github.com/Azure/ARO-RP/pull/3823

PR #3851, opened 2024-09-19
Adding role assignment for the devops managed identity

anshulvermapatel:
While updating OpenShift version in CosmosDB, the pipeline does not authorise to access CosmosDB. This PR will create role assignments when RP will be deployed so that when a OpenShift version is added/removed it has access to CosmosDB. 

We will need to edit the rp-config as well.

Fixes - ARO-9920

### Test plan for issue:
Will be tested in INT env

### DIFF
--- a/pkg/deploy/config.go
+++ b/pkg/deploy/config.go
@@ -74,6 +74,7 @@ type Configuration struct {
 	GlobalResourceGroupName            *string                `json:"globalResourceGroupName,omitempty" value:"required"`
 	GlobalResourceGroupLocation        *string                `json:"globalResourceGroupLocation,omitempty" value:"required"`
 	GlobalSubscriptionID               *string                `json:"globalSubscriptionId,omitempty" value:"required"`
+	GlobalDevopsManagedIdentity        *string                `json:"globalDevopsManagedIdentity,omitempty" value:"required"`
 	KeyvaultDNSSuffix                  *string                `json:"keyvaultDNSSuffix,omitempty" value:"required"`
 	KeyvaultPrefix                     *string                `json:"keyvaultPrefix,omitempty" value:"required"`
 	MDMFrontendURL                     *string                `json:"mdmFrontendUrl,omitempty" value:"required"`
--- a/pkg/deploy/assets/rp-development.json
+++ b/pkg/deploy/assets/rp-development.json
@@ -97,6 +97,20 @@
             ]
         },
         {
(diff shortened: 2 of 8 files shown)


SudoBrendan (dismissed):

pkg/deploy/config.go
```diff
@@ -74,6 +74,7 @@ type Configuration struct {
 	GlobalResourceGroupName            *string                `json:"globalResourceGroupName,omitempty" value:"required"`
 	GlobalResourceGroupLocation        *string                `json:"globalResourceGroupLocation,omitempty" value:"required"`
 	GlobalSubscriptionID               *string                `json:"globalSubscriptionId,omitempty" value:"required"`
+	GlobalDevopsManagedIdenetity       *string                `json:"globalDevopsManagedIdenetity,omitempty" value:"required"`
```
bennerv, line 77: nit: spelling

```suggestion
	GlobalDevopsManagedIdentity       *string                `json:"globalDevopsManagedIdentity,omitempty" value:"required"`

Comment (anshulvermapatel) [reply pkg/deploy/config.go]: Corrected.

File: pkg/deploy/config.go [hunk 2]
```diff
@@ -74,6 +74,7 @@ type Configuration struct {
 	GlobalResourceGroupName            *string                `json:"globalResourceGroupName,omitempty" value:"required"`
 	GlobalResourceGroupLocation        *string                `json:"globalResourceGroupLocation,omitempty" value:"required"`
 	GlobalSubscriptionID               *string                `json:"globalSubscriptionId,omitempty" value:"required"`
+	GlobalDevopsManagedIdentity        *string                `json:"globalDevopsManagedIdentity,omitempty" value:"required"`
```
tiguelu, line 77: I also left a comment in PR 10887667 for RP-Config with this question, but cross-posting it here for visibility.

This new config which is being added to RP-Config is already present in RP-Config as `ev2Configuration.deploymentSettings.managedIdentityName`. The new config is duplicating that existing value. 

If we change one config in the future, we will need to remember to change the duplicated one too. Is there a technical or design reason not to use the existing config to avoid that duplication?

tiguelu (commented):
Just left one question regarding the new config. In any case, before merging this PR we need to either change the config here or update the PR for RP-Config (which has a typo).

pkg/deploy/config.go
anshulvermapatel: Hey, sorry for the delay. So actually, the field is indeed there in the rp-config however, it is in the ev2Configuration section which is used only for ev2 stuff and we don't even have it in our model/struct in our go code and that is the reason, I created another field in order to keep things segregated.

niontive (approved):

PR #3857, opened 2024-09-23
Move clientOptions to environment

bitoku:

Fixes N/A
This is needed for https://issues.redhat.com/browse/ARO-4665

This PR moves clientOptions to environment so that we can reuse the option from anywhere.

### Test plan for issue:

ci, e2e

N/A

cluster creation in prod.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -5,11 +5,8 @@ package cluster
 
 import (
 	"context"
-	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


kimorris27:
This change is already included in https://github.com/Azure/ARO-RP/pull/3757, so I think we can close this PR once that one merges.

bitoku:
Thanks! I thought I'd seen this change somewhere, but couldn't remember.

PR #3859, opened 2024-09-23
SDK2: Remove deprecated network sdk from e2e.

bitoku:

This PR removes deprecated network sdk from e2e.

### Test plan for issue:

e2e

N/A

Mostly the changes are nothing to do with prod.
Only monitor nsg has a small change because of the subnets client change.
We can validate the monitor nsg function by checking these metrics.
```
	MetricPreconfiguredNSGEnabled  = "monitor.preconfigurednsg.enabled"
	MetricFailedNSGMonitorCreation = "monitor.preconfigurednsg.failedmonitorcreation"
	MetricInvalidDenyRule          = "monitor.preconfigurednsg.invaliddenyrule"
	MetricSubnetAccessForbidden    = "monitor.preconfigurednsg.subnetaccessforbidden"
```

### DIFF
--- a/pkg/util/azureclient/azuresdk/armnetwork/subnets.go
+++ b/pkg/util/azureclient/azuresdk/armnetwork/subnets.go
@@ -7,24 +7,23 @@ import (
 	"context"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
-	sdknetwork "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v2"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v2"
 
 	"github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/azcore"
 )
 
 // SubnetsClient is a minimal interface for azure-sdk-for-go subnets client
 type SubnetsClient interface {
-	Get(ctx context.Context, resourceGroupName, virtualNetworkName, subnetName string, options *sdknetwork.SubnetsClientGetOptions) (sdknetwork.SubnetsClientGetResponse, error)
+	Get(ctx context.Context, resourceGroupName, virtualNetworkName, subnetName string, options *armnetwork.SubnetsClientGetOptions) (armnetwork.SubnetsClientGetResponse, error)
+	SubnetsClientAddons
 }
 
 type subnetsClient struct {
-	*sdknetwork.SubnetsClient
+	*armnetwork.SubnetsClient
 }
 
-var _ SubnetsClient = (*subnetsClient)(nil)
-
 func NewSubnetsClient(subscriptionID string, credential azcore.TokenCredential, options *arm.ClientOptions) (SubnetsClient, error) {
-	client, err := sdknetwork.NewSubnetsClient(subscriptionID, credential, options)
+	client, err := armnetwork.NewSubnetsClient(subscriptionID, credential, options)
 
-	return subnetsClient{client}, err
+	return &subnetsClient{client}, err
 }
--- a/test/e2e/adminapi_resources.go
+++ b/test/e2e/adminapi_resources.go
@@ -59,17 +59,16 @@ var _ = Describe("[Admin API] List Azure resources action", func() {
 			subnets[strings.ToLower(*p.SubnetID)] = struct{}{}
 		}
 
-		vnet, err := clients.VirtualNetworks.Get(ctx, r.ResourceGroup, r.ResourceName, "")
+		vnet, err := clients.VirtualNetworks.Get(ctx, r.ResourceGroup, r.ResourceName, nil)
 		Expect(err).NotTo(HaveOccurred())
 
-		for _, subnet := range *vnet.Subnets {
+		for _, subnet := range vnet.Properties.Subnets {
 			if _, ok := subnets[strings.ToLower(*subnet.ID)]; !ok {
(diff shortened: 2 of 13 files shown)


kimorris27 (dismissed):

pkg/util/azureclient/azuresdk/armnetwork/subnets.go
```diff
@@ -7,24 +7,23 @@ import (
...
+	Get(ctx context.Context, resourceGroupName, virtualNetworkName, subnetName string, options *armnetwork.SubnetsClientGetOptions) (armnetwork.SubnetsClientGetResponse, error)
+	SubnetsClientAddons
 }
 
 type subnetsClient struct {
-	*sdknetwork.SubnetsClient
+	*armnetwork.SubnetsClient
 }
 
-var _ SubnetsClient = (*subnetsClient)(nil)
```
kimorris27, line 24: Not a concern I'd block merging over, but why'd you remove this line and the corresponding line for the vnet client?

kimorris27 (approved):

test/e2e/operator.go
```diff
@@ -448,9 +454,9 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
 		for subnet, correctNSG := range subnetsToReconcile {
 			By(fmt.Sprintf("waiting for the subnet %q to be reconciled so it includes the original cluster NSG", subnet))
 			Eventually(func(g Gomega, ctx context.Context) {
-				s, err := clients.Subnet.Get(ctx, resourceGroup, vnetName, subnet, "")
+				s, err := clients.Subnet.Get(ctx, resourceGroup, vnetName, subnet, nil)
 				g.Expect(err).NotTo(HaveOccurred())
```
tiguelu, line 458: Subnet field not extracted from response.
```suggestion
				resp, err := clients.Subnet.Get(ctx, resourceGroup, vnetName, subnet, nil)
				g.Expect(err).NotTo(HaveOccurred())
				s := resp.Subnet
```

test/e2e/adminapi_resources.go
```diff
@@ -59,17 +59,16 @@ var _ = Describe("[Admin API] List Azure resources action", func() {
 			subnets[strings.ToLower(*p.SubnetID)] = struct{}{}
 		}
 
-		vnet, err := clients.VirtualNetworks.Get(ctx, r.ResourceGroup, r.ResourceName, "")
+		vnet, err := clients.VirtualNetworks.Get(ctx, r.ResourceGroup, r.ResourceName, nil)
 		Expect(err).NotTo(HaveOccurred())
```
tiguelu, line 63: Response from Get is not being processed.
```suggestion
		resp, err := clients.VirtualNetworks.Get(ctx, r.ResourceGroup, r.ResourceName, nil)
		Expect(err).NotTo(HaveOccurred())
		vnet := resp.VirtualNetwork
```

tiguelu (commented):
Thank you for putting this together Ayato. 

E2E is passing in the PR, but I think some responses from the new clients are not being processed (see comments below). Can you confirm?

pkg/util/azureclient/azuresdk/armnetwork/subnets.go
bitoku: This idiom is to check if the struct implements the interface in the compile time.
This is usually used in libraries where there are some exposed structs which are not used within the same libraries.

In our case, the struct is assigned to some variables (or struct fields) whose type is the interface, so the implementation is validated at that point, so we don't have to check here.

https://go.dev/doc/effective_go#blank_implements
> By convention, such declarations are only used when there are no static conversions already present in the code, which is a rare event.

test/e2e/operator.go
```diff
@@ -344,41 +344,43 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
 		// This is expensive but will prevent flakes.
 		By("gathering existing subnet NSGs")
 		for subnet := range subnetsToReconcile {
-			subnetObject, err := clients.Subnet.Get(ctx, resourceGroup, vnetName, subnet, "")
+			subnetObject, err := clients.Subnet.Get(ctx, resourceGroup, vnetName, subnet, nil)
 			Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
bitoku: We don't have to assign the response to `resp` to get `Subnet` every time.

The first return value of Get is SubnetsClientGetResponse.
https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork#SubnetsClientGetResponse
and it can transparently reference to the fields of Subnet like this.
```
subnetsToReconcile[subnet] = subnetObject.Properties.NetworkSecurityGroup.ID
```
So we don't have to use `resp` just to get the `Subnet` value.
However I used `resp` in some places where I have to reassign the value to `Subnet` type, otherwise I don't use `resp` and use it directly as if it is a `Subnet` type.

test/e2e/operator.go
bitoku: ditto

test/e2e/adminapi_resources.go
bitoku: ditto

test/e2e/operator.go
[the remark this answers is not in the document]
tiguelu: Thank you for the clarification.

tiguelu (approved):

PR #3698, opened 2024-07-16
Update Go to 1.21

niontive:
### Which issue this PR addresses:

Fixes: https://issues.redhat.com/browse/ARO-8998

### What this PR does / why we need it:
Update to Go 1.21 since 1.20 is out of support.

### Test plan for issue:

Unit tests. Will also deploy to INT

### Is there any documentation that needs to be updated for this PR?

Yes - see PR

PR #3862, opened 2024-09-24
Cluster MSI follow-up PR

kimorris27:

https://issues.redhat.com/browse/ARO-4360

The main purpose of this PR is to remove our dependency on https://github.com/golang/mock and completely switch over to using https://github.com/uber-go/mock instead for two reasons:

1. It needs to be done eventually anyway
2. It happens to fix an issue that I ran into generating mocks for some new Azure clients I added in my previous cluster MSI PR (the golang version doesn't seem to be up-to-date enough to handle generics properly in certain contexts)

I also added in a few small changes in response to previous PR feedback. If you follow the commit history, it's hopefully easy to pick out which of the numerous changed files you need to care about as a reviewer (as opposed to the ones where the only change is the mock repo referenced). Notably, I upgrade Go from 1.21.11 to 1.21.13 to satisfy Uber mockgen's dependency on Go >= 1.21.12.

### Test plan for issue:

CI and E2E

I don't think so, but let me know if you know of any.

E2E during the next release :D


kimorris27:
E2E failed because our arointsvc registry doesn't have the Go 1.21.13 image. Looks like I'll need to follow what was done here to fix that: https://github.com/Azure/ARO-RP/pull/3698#issuecomment-2233560573

pkg/util/azureclient/azuresdk/armcosmos/generate.go
```diff
@@ -0,0 +1,8 @@
+package armcosmos
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+//go:generate rm -rf ../../../../../pkg/util/mocks/azureclient/azuresdk/$GOPACKAGE
+//go:generate mockgen -destination=../../../mocks/azureclient/azuresdk/$GOPACKAGE/$GOPACKAGE.go github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/$GOPACKAGE DatabaseAccountsClient
+//go:generate goimports -local=github.com/Azure/ARO-RP -e -w ../../../mocks/azureclient/azuresdk/$GOPACKAGE/$GOPACKAGE.go
```
tsatam, line 8: If we weren't generating mocks for these two clients, that means we weren't using mock clients in our tests for RP functionality using these clients, correct? (assuming there even are tests). 

If that's the case, should we even bother generating mocks that go unused?

tsatam (approved):
I've reviewed the changes and they look good, but I'd suggest perhaps splitting the "replace golang/mock with uber-go/mock" part out into a separate PR to get more focused review from the wider team, only if it is easy enough to do so.

kimorris27:
Closing this to split into two more granular PRs: one that moves us to the Uber mock fork and one that has the follow-up stuff from the previous cluster MSI PR.

pkg/util/azureclient/azuresdk/armcosmos/generate.go
kimorris27: So there were mocks present, but if you remove them and `make generate`, they don't get replaced. I'm guessing that when they were added, the committer forgot to commit their changes to the `generate.go` files.

This came to my attention as I was moving us over to the Uber fork because it caused errors while I was trying to get builds and unit tests working, so I codified the generation properly.

PR #3866, opened 2024-09-26
fix: Ensure dns disabled for networks

jaitaiwan:
Fixes a latency issue with outbound connections caused by the creation of dedicated podman networks.

Fix a regression in outbound connection latency.

### Test plan for issue:
RP deploy

### Full details
Podman creates the default bridge network with dns_enabled to false. Subsequent calls to podman network create then create bridge networks where dns_enabled is set to true causing containers on said network to make use of podman's dnsutil plugin. This results in latency issues for gateway connections.

This change ensures that networks created use similar settings to that of the default bridge network which has no latency issues.

Fixes

### DIFF
--- a/pkg/deploy/generator/scripts/util-system.sh
+++ b/pkg/deploy/generator/scripts/util-system.sh
@@ -329,6 +329,7 @@ create_podman_networks() {
         log "Creating podman network \"$n\" with subnet \"${nets[$n]}\""
         podman network \
             create \
+            --disable-dns \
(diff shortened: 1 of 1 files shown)


bitoku:
@jaitaiwan Is this PR still valid?

jaitaiwan:
Still relevant. We'll need this going forward when we start moving towards aks deployments as this will help to simulate the environment if I remember correctly and without dns-disabled we'll continue to see the same issues we saw in a previous incident.

PR #3867, opened 2024-09-27
add an update step to replace digicert with onecert

yithian:

Fixes ARO-10309

This allows upgrades to 4.16 to proceed. These upgrades were being blocked by the SHA-1 signing algorithm used by DigiCert, which is incompatible with 4.16

### Test plan for issue:

I'm not sure how to test this. It will need to be deployed somewhere with access to an affected cluster

No, it's mitigation for an incident

The new step is limited to affected clusters only


pkg/cluster/replacedigicert.go
```diff
@@ -0,0 +1,47 @@
...
+	"strings"
+
+	"github.com/Azure/ARO-RP/pkg/util/keyvault"
+)
+
+// if the cluster is using a managed domain and has a DigiCert-issued
+// certificate, replace the certificate with one issued by OneCert. This
+// ensures that clusters upgrading to 4.16 aren't blocked due to the SHA-1
+// signing algorithm in use by DigiCert
+func (m *manager) replaceDigicert(ctx context.Context) error {
```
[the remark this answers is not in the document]
tsatam: I'd even go as far as to suggest this function should be called something like "ensureCertificateIssuer" or something generic, to go in line with your suggestion above to abstract the specifics around what issuer is being managed here.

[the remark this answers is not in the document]
yithian: Due to [how `steps.actionFunction` is defined](https://github.com/Azure/ARO-RP/blob/master/pkg/util/steps/action.go#L16), I don't think the high-level step to convert from DigiCert to OneCert can be a generic `ensureCertificateIssuer`-style function with parameters

I can extract the logic for that into a generic util function though

tsatam (commented):
Looks good, will approve after the PR is tested in an applicable environment to ensure it behaves as expected.

fahlmant (commented):
LGTM! Thanks!

yithian:
I flipped the cert policy issuer on my test cluster back to DigiCert after pushing my latest RP changes to canary. After running a PUCM, the (DigiCert-issued) certificate was replaced with one from OneCert

cadenmarchese (commented):
There are a number of utility functions related to managed domains under https://github.com/Azure/ARO-RP/blob/master/pkg/util/dns/dns.go - can we use any of them in this PR?

pkg/cluster/correct_cert_issuer.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/Azure/ARO-RP/pkg/util/keyvault"
+)
+
+// if the cluster isn't using a managed domain and has a DigiCert-issued
+// certificate, replace the certificate with one issued by OneCert. This
+// ensures that clusters upgrading to 4.16 aren't blocked due to the SHA-1
+// signing algorithm in use by DigiCert
+func (m *manager) correctCertificateIssuer(ctx context.Context) error {
+	if !strings.Contains(m.doc.OpenShiftCluster.Properties.ClusterProfile.Domain, ".") {
+		apiCertName := m.doc.ID + "-apiserver"
```
kimorris27, line 20: Nit (and tangentially related to your changes): we could turn this into a "constant" by making a util method on the cluster manager type that returns the API cert's Secret name. This would ensure that it's correct across the different files that need to know and use the Secret name. Same applies to the ingress cert's Secret name.

pkg/cluster/correct_cert_issuer.go
```diff
@@ -0,0 +1,68 @@
...
+		apiHostname := strings.Split(strings.TrimPrefix(m.doc.OpenShiftCluster.Properties.APIServerProfile.URL, "https://"), ":")[0]
+		err := m.ensureCertificateIssuer(ctx, apiCertName, apiHostname, "OneCertV2-PublicCA")
```
kimorris27, line 22: Nit: the OneCert issuer name could be a constant since we use it in more than one place.

pkg/cluster/correct_cert_issuer.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	clusterKeyvault := m.env.ClusterKeyvault()
+
+	bundle, err := clusterKeyvault.GetCertificate(ctx, certificateName)
+	if err != nil {
+		return err
+	}
+
+	if *bundle.Policy.IssuerParameters.Name != issuerName {
```
kimorris27, line 50: Nit: I'd be pretty surprised if we found any nil pointers in a non-error result of `GetCertificate`, but I'd still check them all to be super safe.

fahlmant:
@yithian What's left to do here before merge?

yithian:
> @yithian What's left to do here before merge?

@fahlmant I need to write up the util function (and related tests) that Kipp described [here](https://github.com/Azure/ARO-RP/pull/3867#discussion_r1802066128)

I have fixes for the other comments staged up locally but I've been fully dedicated to PUCMs for the past several days

fahlmant:
This was supposed to go into the PUMC. I'm concerned this is getting kicked down the road. Is it possible to get this updated this week?

yithian:
Yes, this should be done this week

kimorris27 (approved):
LGTM with the context that it's been tested successfully 🚀

fahlmant (dismissed):
lgtm again :)

fahlmant (approved):

kimorris27 (approved):

cadenmarchese (approved):

PR #3856, opened 2024-09-20
RP E2E Test Execution

shubhadapaithankar:
### Which issue this PR addresses:

This PR addresses inefficiencies in our CI and E2E testing process. The current setup is slow and lacks full containerization and Hive integration for cluster management. By using Docker Compose and integrating Hive, this PR reduces runtime, simplifies orchestration, and improves testing reliability.

Fixes: [ARO-9501](https://issues.redhat.com/browse/ARO-9501)

### What this PR does / why we need it:

This PR introduces an E2E testing stage using Docker Compose to orchestrate the RP, Selenium, and E2E test services. The pipeline pulls the required RP image from Azure Container Registry (ACR) and ensures the correct environment is set up for the E2E tests. It helps validate the end-to-end functionality of the system and ensures that tests are running using containerized RP images.

- **Why we need it:** The E2E stage ensures that the entire system is tested in a environment, using containerized components (including RP) and validating workflows through Selenium for browser-based testing.
-  **Key benefits:** This change helps automate the test execution process for the complete system, ensuring that every aspect of the application functions correctly in a containerized environment before deployment.

#### Key Points:

- [ ] **Improved Efficiency:** This new E2E pipeline reduces the runtime for both CI and E2E stages by more than half compared to the current process.
- [ ] **Hive Integration:** We’re now using Hive in the pipeline to create and manage clusters. Hive allows us to propagate required credentials to the installer, as outlined in [Hive’s AWS provisioning guide](https://github.com/openshift/hive/blob/master/docs/aws-sts-provisioning.md) (we apply similar functionality for Azure). This is a major step forward for MIWI, as it enhances our cluster setup, ensuring smoother and more reliable testing.
- [ ] **Deployment of Hive Operator:** Deploying the Hive Operator in this process is another major step forward for MIWI, ensuring better control and integration of clusters.
- [ ] **Fully Dockerized Process:** All services are run via Docker Compose, allowing us to orchestrate the setup with a single command, making the process simpler and more streamlined.
- [ ] **Image Management:** Images for components like rp and e2e are created during the CI containerized stage and then pulled directly for use in the E2E stage.

### Test plan for issue:

1. E2E tests are executed within the pipeline's new E2E stage using Docker Compose.
2. The tests are triggered automatically after the Containerized stage, ensuring seamless orchestration.

### Is there any documentation that needs to be updated for this PR?
N/A

### How do you know this will function as expected in production? 
N/A

PR #3874, opened 2024-10-01
Remove pipeline chaining from our E2E pipeline

hawkowl:

Back when we had automatically triggering ADO pipelines (which we don't have because "require comment by contributor" is locked on at the org level) it was convenient to have the CI run and then the e2e trigger off a successful CI run to reduce pointless E2E runs. Now that we have to do the `/azp run` command (which triggers CI + E2E), we can end up with cases where the E2E will be manually triggered but also secondarily triggered from the successful CI run for the same commit (an example from this week):

image

This removes this e2e duplication, which fills up the E2E queue pointlessly and never ends up assigned to the GitHub check anyway.

I've also disabled overriding the YAML in the e2e.yml file in the UI, so it will use these values.

### Test plan for issue:

Is CI/E2E based

### DIFF
--- a/.pipelines/ci.yml
+++ b/.pipelines/ci.yml
@@ -11,6 +11,8 @@ trigger:
     include:
       - v2*
 
+# PR triggers are overridden in the ADO UI
+
 resources:
   containers:
     - container: golang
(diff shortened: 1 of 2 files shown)


tsatam (approved):
I'm fine with this as a temporary fix for the duplicated E2E pipeline execution, but long term we'll be moving the E2E stage to be directly within the CI pipeline (and use the aro artifacts built during CI for E2E), the WIP PR is here: #3856

PR #3875, opened 2024-10-01
Don't run long-running regression tests on PRs

hawkowl:

This runs regression tests on CI runs (i.e. tags and master) but does not run them during PRs, which cuts 20-30 minutes from the runtime.

### Test plan for issue:

E2E changes, hopefully just works according to the docs

Comments are added in the relevant code

E2E tests, doesn't touch production


hawkowl:
E2E was 85m down from 113 :)

tsatam (approved):

SrinivasAtmakuri (approved):

PR #3881, opened 2024-10-03
Hotfix - MSI RP endpoint env var

kimorris27:

No Jira

https://github.com/Azure/ARO-RP/pull/3757 introduced a new RP-Config parameter, but I missed a few places where we need to edit our code to make sure the parameter makes it all the way from the config to the running RP.

We found the issue when someone's full service dev RP was failing to start because it found this parameter was missing during startup.

### Test plan for issue:

We deployed this change to canary and made sure the RP started up okay.

N/A

Tested in canary.


tsatam (approved):
Biased LGTM since I helped out on this :)

PR #3882, opened 2024-10-03
Migrate VPN and rp service to Docker Compose

shubhadapaithankar:

Fixes : [ARO-9501](https://issues.redhat.com/browse/ARO-9501)

This PR migrates the `VPN`, `RP,` and `Portal services` to Docker Compose, simplifying service management by containerizing them. The `docker-compose.yml` now orchestrates all three services. This migration makes the environment setup and service orchestration more efficient and maintainable, improving local development and CI/CD workflows.

Key Updates:

- [ ] VPN Service: Managed via docker-compose.yml with an associated `Dockerfile.vpn`, allowing for easier configuration of the VPN through Docker Compose.
- [ ] RP Service: Uses `Dockerfile.ci-rp` and is defined in `docker-compose.yml`, with the health checks in place to ensure proper operation.
- [ ] Portal Service: Migrated to Docker Compose using `docker-compose.yml`, with secrets mounted, environment variables passed, and the necessary portal command defined to run the service correctly.
- [ ] All services are fully containerized and share consistent configuration between local and pipeline environments.

Other Important Highlights:

- Updated Azure DevOps pipeline to ensure correct variable definitions (REGISTRY, VERSION, LOCAL_ARO_RP_IMAGE, etc.).
- Refactored the template to handle Docker authentication and image tagging/pushing with the correct use of VERSION.
- Updated the Docker Compose file to consistently use pipeline variables for image names, tags, and build arguments.
- Simplified Makefile by removing redundant variable definitions and ensuring compatibility with pipeline-defined variables.

### Test plan for issue:

- The VPN service was tested using Docker Compose.
- Verified that the VPN connects successfully with the provided OpenVPN configuration.
- The RP service was also tested in Docker Compose, ensuring it runs with proper health checks.
- Health checks confirm that both the VPN and RP services are operational

N/A

### DIFF
--- a/.pipelines/templates/template-acr-push.yml
+++ b/.pipelines/templates/template-acr-push.yml
@@ -20,6 +20,23 @@ steps:
       scriptLocation: 'inlineScript'
       inlineScript: |
         set -xe
+
+        # Update packages and install docker.io and docker-compose-plugin
+        echo "Installing Docker and Docker Compose Plugin..."
+        sudo apt-get update
+        sudo apt-get install -y ca-certificates curl gnupg
+        sudo install -m 0755 -d /etc/apt/keyrings
+        curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo tee /etc/apt/keyrings/docker.asc
+        sudo chmod a+r /etc/apt/keyrings/docker.asc
+        echo \
+          "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/ubuntu \
+          $(. /etc/os-release && echo \"$VERSION_CODENAME\") stable" | \
+          sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+        sudo apt-get update
+        sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
+        sudo systemctl start docker
+        sudo systemctl enable docker
+
         echo "Authenticating to Azure and ACR..."
         ACR_FQDN="${{ parameters.acrFQDN }}"
         REGISTRY_NAME=$(echo $ACR_FQDN | cut -d'.' -f1)
@@ -32,34 +49,37 @@ steps:
         docker images
 
         # Ensure the image is available locally before tagging the build ID
-        IMAGE_NAME="${{ parameters.repository }}:$(Build.BuildId)"
+        IMAGE_NAME="${ACR_FQDN}/${{ parameters.repository }}:$(VERSION)"  # Use full image name
+        echo "Checking for image $IMAGE_NAME..."
         if [[ "$(docker images -q $IMAGE_NAME 2> /dev/null)" == "" ]]; then
           echo "Error: Image $IMAGE_NAME not found. Exiting."
           exit 1
         fi
 
         # Ensure the image is available locally before tagging 'latest'
-        IMAGE_LATEST="${{ parameters.repository }}:latest"
+        IMAGE_LATEST="${ACR_FQDN}/${{ parameters.repository }}:latest"
+        echo "Checking for image $IMAGE_LATEST..."
         if [[ "$(docker images -q $IMAGE_LATEST 2> /dev/null)" == "" ]]; then
-          echo "Warning: Image $IMAGE_LATEST not found. Skipping latest tag."
+          echo "Warning: Image $IMAGE_LATEST not found. Skipping 'latest' tag."
           SKIP_LATEST=true
         else
+          echo "Image $IMAGE_LATEST found. Proceeding with 'latest' tag."
           SKIP_LATEST=false
         fi
 
         # Tag the image with the ACR repository for the build ID
         echo "Tagging image with build ID..."
-        docker tag $IMAGE_NAME ${ACR_FQDN}/${{ parameters.repository }}:$(Build.BuildId)
+        docker tag $IMAGE_NAME ${ACR_FQDN}/${{ parameters.repository }}:$(VERSION)
 
         # If the latest image exists, tag it as well
         if [ "$SKIP_LATEST" == "false" ]; then
           echo "Tagging image with 'latest'..."
-          docker tag $IMAGE_LATEST ${ACR_FQDN}/${{ parameters.repository }}:latest
+          docker tag $IMAGE_NAME ${ACR_FQDN}/${{ parameters.repository }}:latest
         fi
 
         # Push the Docker image to ACR with build ID
         echo "Pushing image with build ID to ACR..."
-        docker push ${ACR_FQDN}/${{ parameters.repository }}:$(Build.BuildId)
+        docker push ${ACR_FQDN}/${{ parameters.repository }}:$(VERSION)
 
         # Optionally push the image as 'latest'
         if [ "${{ parameters.pushLatest }}" == "true" ] && [ "$SKIP_LATEST" == "false" ]; then
--- a/.pipelines/templates/template-acr-push.yml
+++ b/.pipelines/templates/template-acr-push.yml
@@ -11,7 +11,6 @@ parameters:
     default: false
 
 steps:
-  # Authenticate to ACR and push the image
   - task: AzureCLI@2
     displayName: 'Authenticate to Azure and Push Docker Image'
     inputs:
@@ -21,7 +20,7 @@ steps:
       inlineScript: |
         set -xe
 
-        # Update packages and install docker.io and docker-compose-plugin
+        # Install Docker dependencies
         echo "Installing Docker and Docker Compose Plugin..."
         sudo apt-get update
         sudo apt-get install -y ca-certificates curl gnupg
@@ -37,6 +36,7 @@ steps:
         sudo systemctl start docker
         sudo systemctl enable docker
 
+        # Authenticate to Azure and ACR
         echo "Authenticating to Azure and ACR..."
         ACR_FQDN="${{ parameters.acrFQDN }}"
         REGISTRY_NAME=$(echo $ACR_FQDN | cut -d'.' -f1)
@@ -48,12 +48,17 @@ steps:
         echo "Listing Docker images..."
         docker images
 
-        # Ensure the image is available locally before tagging the build ID
-        IMAGE_NAME="${ACR_FQDN}/${{ parameters.repository }}:$(VERSION)"  # Use full image name
-        echo "Checking for image $IMAGE_NAME..."
-        if [[ "$(docker images -q $IMAGE_NAME 2> /dev/null)" == "" ]]; then
-          echo "Error: Image $IMAGE_NAME not found. Exiting."
+        # Check if the image is already tagged with the full repository path
+        IMAGE_NAME="${ACR_FQDN}/${{ parameters.repository }}:$(VERSION)"
+        LOCAL_IMAGE="${{ parameters.repository }}:$(VERSION)"
+
+        echo "Checking for image $LOCAL_IMAGE..."
+        if [[ "$(docker images -q $LOCAL_IMAGE 2> /dev/null)" == "" ]]; then
+          echo "Error: Image $LOCAL_IMAGE not found. Exiting."
           exit 1
+        else
+          echo "Image found. Retagging with full repository path..."
+          docker tag $LOCAL_IMAGE $IMAGE_NAME
         fi
 
         # Ensure the image is available locally before tagging 'latest'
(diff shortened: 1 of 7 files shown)


.pipelines/templates/template-acr-push.yml
```diff
@@ -20,6 +20,23 @@ steps:
...
+        curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo tee /etc/apt/keyrings/docker.asc
+        sudo chmod a+r /etc/apt/keyrings/docker.asc
+        echo \
+          "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/ubuntu \
+          $(. /etc/os-release && echo \"$VERSION_CODENAME\") stable" | \
+          sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+        sudo apt-get update
+        sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
+        sudo systemctl start docker
+        sudo systemctl enable docker
```
SudoBrendan, line 38: Is this shell equivalent to the [Docker Install ADO Task](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/docker-installer-v0?view=azure-pipelines) or is it different?

Makefile
```diff
@@ -374,7 +374,6 @@ DOCKER_BUILD_CI_ARGS ?=
 
 # Image names that will be found in the local podman image registry after build
 # (tags are always VERSION).
-LOCAL_ARO_RP_IMAGE ?= aro
```
SudoBrendan, line 377: removing this means we need to notify all devs that new envvars are required to build - right? Should we update env.example?

Makefile
```diff
@@ -397,30 +396,29 @@ ci-clean:
 	podman $(PODMAN_REMOTE_ARGS) \
 	    image prune --all --filter="label=aro-*=true"
 
+.PHONY: version
+version:
+	echo "RP $(LOCAL_ARO_RP_IMAGE):$(VERSION)"
+
```
SudoBrendan, line 402: do we need this, or can it be removed?

shubhadapaithankar: I have removed the echo "RP $(LOCAL_ARO_RP_IMAGE):$(VERSION)"

.pipelines/templates/template-acr-push.yml
shubhadapaithankar: The shell script is different from the DockerInstaller ADO task. The script manually installs Docker and Docker Compose from Docker’s official repository, providing more control over the process.

Makefile
```diff
@@ -486,56 +480,15 @@ run-portal: ci-rp podman-secrets
...
-		-e MOCK_MSI_TENANT_ID \
-		-e MOCK_MSI_CLIENT_ID \
-		-e MOCK_MSI_CERT \
-		--secret aks.kubeconfig,target=/app/secrets/aks.kubeconfig \
-		--secret proxy-client.key,target=/app/secrets/proxy-client.key \
-		--secret proxy-client.crt,target=/app/secrets/proxy-client.crt \
-		--secret proxy.crt,target=/app/secrets/proxy.crt \
-		$(LOCAL_ARO_RP_IMAGE):$(VERSION) rp
+	docker compose rm -sf rp
+	docker compose up rp
```
SudoBrendan, line 484: syntax error: we need the target name :)

Dockerfile.vpn
```diff
@@ -0,0 +1,8 @@
+# Use a Microsoft-approved image
+FROM mcr.microsoft.com/azure-cli:2.61.0 AS base
```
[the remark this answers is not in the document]
shubhadapaithankar: Updated the `env.example`

Makefile
shubhadapaithankar: Updated the env.example

Makefile
shubhadapaithankar: sorry , My bad 
I fixed it .

Dockerfile.vpn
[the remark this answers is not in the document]
shubhadapaithankar: Thanks for the suggestion!

I did try several slimmer images (including Mariner), but I ran into issues with missing dependencies needed to run the VPN properly. Unfortunately, none of them worked without breaking something.

Also, when we used the Red Hat registry image in the pipeline, we got security warnings from Microsoft, so we had to avoid that.

For now, the current image works without issues or security concerns, but I’m open to exploring better options if we find something.

Makefile
```diff
@@ -399,28 +399,23 @@ ci-clean:
...
-		-t $(LOCAL_ARO_RP_BUILD_IMAGE):$(VERSION)
+		-t ${LOCAL_ARO_RP_BUILD_IMAGE}:${VERSION}
 
-	docker build . $(DOCKER_BUILD_CI_ARGS) \
-		-f Dockerfile.ci-rp \
-		--ulimit=nofile=4096:4096 \
-		--build-arg REGISTRY=$(REGISTRY) \
-		--build-arg ARO_VERSION=$(VERSION) \
-		-t $(LOCAL_ARO_RP_IMAGE):$(VERSION)
+	docker compose build rp
```
kevinobriendotca, line 411: I had to add REGISTRY=${REGISTRY} before the docker compose command to get this to work locally.

kevinobriendotca (approved):
Tested docker compose build locally and everything works as expected.

slawande2 (dismissed):
Tested locally with Shubhada. LGTM!

slawande2 (dismissed):

env.example
```diff
@@ -17,4 +17,11 @@ export MOCK_MSI_CERT="replace_with_value_output_by_hack/devtools/msi.sh"
 export MOCK_MSI_TENANT_ID="replace_with_value_output_by_hack/devtools/msi.sh"
 export PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS="replace_with_value_output_by_hack/devtools/msi.sh"
 
+# you will need this to run-rp , vpn and ci-rp using Docker compose
+export REGISTRY=registry.access.redhat.com
+export LOCAL_ARO_RP_IMAGE=localhost/aro
+export VERSION=latest
+export TAG=latest
+export LOCAL_VPN_IMAGE=arosvcdev.azurecr.io/vpn
```
SudoBrendan, line 25: nit: `docker compose` for me locally attempts to optimize here, and pull the image before building. This results in a warning that auth to the ACR failed. Definitely not a blocker, but I wonder of making this `LOCAL_VPN_IMAGE=vpn` and the `LOCAL_ARO_RP_IMAGE=aro` for simplicity makes sense or not?

Makefile
```diff
@@ -456,86 +451,22 @@ podman-secrets: aks.kubeconfig
...
+
+.PHONY: vpn
+vpn:
+	docker compose build vpn
+
+.PHONY: run-vpn
+run-vpn:
+	docker compose rm -sf vpn
+	docker compose up vpn
+	docker compose logs --follow vpn
```
SudoBrendan, line 472: A simple question with no right answers, but I want to ask: is there value in us running VPN containerized and standalone, or should we just rely on `docker compose` to manage this (ideally hidden) component? That is to say, in theory, if I do `make run-rp` with this change, the VPN is built and/or instantiated without me needing to `make run-vpn` in addition to that because the compose configuration defines `vpn` as a dependency of `rp`; your code in the compose "just works" in a single command.

All this to say: should we simplify this code and not include specific targets in the Makefile for `vpn` images, and leave it up to compose instead?

I guess I could see value in a re-build, if we have modified the VPN image in some way...

docker-compose.yml
```diff
@@ -0,0 +1,153 @@
+services:
+  vpn:
+    image: ${LOCAL_VPN_IMAGE}:${VERSION}
+    build:
+      context: .
+      dockerfile: Dockerfile.vpn
+    container_name: vpn
+    privileged: true
```
SudoBrendan, line 8: Locally, `make run-rp` fails for me as it fails to connect to the Hive cluster in my region. I think this is because the VPN is being instantiated in it's own Net namespace on the Linux kernel. Changing this to `hostNetwork: true` fixes the problem for me locally:

```suggestion
    privileged: true
    network_mode: host  # Use host network to set VPN for all containers running on the host
```

docker-compose.yml
```diff
@@ -0,0 +1,153 @@
...
+      timeout: 30s
+      retries: 3
+    restart: on-failure:3
+
+  portal:
+    image: ${LOCAL_ARO_RP_IMAGE}:${VERSION}
+    container_name: aro-portal
+    depends_on:
+      rp:
+        condition: service_started
```
SudoBrendan, line 109: I think we need this to be `service_healthy`?

SudoBrendan (changes requested):
This PR is yet another delight! I cannot express how much joy there is in doing a single command to stand up our entire stack locally: `docker compose up`. Wow, this is a GAME CHANGER! No more "run commands A, B, C, D, E, F, G in that order, and :pray:". This solution "just works".

There's a few fixes I pointed out that need to made, other than that Looks Incredible To Me :1st_place_medal:

Also - one question for the team. When I do `make run-portal` and go to `https:127.0.0.1:8444/` I do hit the portal auth, login with SSO and get a redirect - but I ultimately land at a `forbidden` page - is that just me or a product of this change, or something else entirely?

Makefile
shubhadapaithankar: yeah, good point! I'll comment out the VPN-specific Makefile targets and rely on Docker Compose to manage the VPN as a dependency of the rp service. This should simplify the workflow. We can revisit if we find the need for explicit control later. Thanks!

docker-compose.yml
shubhadapaithankar: Changed portal service dependency condition to` service_healthy` to ensure the` rp`
  service is fully ready before starting the portal, enhancing reliability

Makefile
```diff
@@ -456,86 +451,25 @@ podman-secrets: aks.kubeconfig
...
+
+#.PHONY: vpn
+#vpn:
+#	docker compose build vpn
+
+#.PHONY: run-vpn
+#run-vpn:
+#	docker compose rm -sf vpn
+#	docker compose up vpn
+#	docker compose logs --follow vpn
```
SudoBrendan, line 475: nit: let's just delete instead of merging in comments.

docker-compose.yml
shubhadapaithankar: Updated VPN service configuration to use host network in `docker-compose.yml`

Makefile
shubhadapaithankar: Deleted VPN stuff from the Makefile

env.example
shubhadapaithankar: Updated the env file with  variables `LOCAL_VPN_IMAGE=vpn `and the `LOCAL_ARO_RP_IMAGE=aro `for simplicity

SudoBrendan (approved):
The PR now works as written, so it gets a :heavy_check_mark: from me! One more nit if we want.

docker-compose.yml
```diff
@@ -0,0 +1,154 @@
...
+
+secrets:
+  proxy-client-key:
+    file: ./secrets/proxy-client.key
+  proxy-client-crt:
+    file: ./secrets/proxy-client.crt
+  proxy-crt:
+    file: ./secrets/proxy.crt
+  hive-kubeconfig:
+    file: ./aks.kubeconfig
```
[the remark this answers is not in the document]
shubhadapaithankar: If we remove the secrets section, Docker Compose will indeed throw an error like:

`service "rp" refers to undefined secret proxy-client-key: invalid compose project`

The top-level secrets section is what tells Docker Compose where to find the actual files for the secrets. Without it, Docker doesn’t know where to look.

slawande2 (approved):

docker-compose.yml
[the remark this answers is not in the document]
SudoBrendan: my suggestion is the opposite direction - let's use compose, rather than https://github.com/Azure/ARO-RP/blob/d64782c410f1155a79047849b5d1fb8900242c28/Makefile#L444-L456 - is there a reason we couldn't delete this part of the makefile now?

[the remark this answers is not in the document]
SudoBrendan: this also makes me realize we've removed the `aks.kubeconfig` dependency (which was in the podman-secrets target, and now needs to be on the `run-rp` target)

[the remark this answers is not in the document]
shubhadapaithankar: Thanks for pointing this, I agree to remove it from Makefile 
- The `podman-secrets` target has been removed as Docker Compose now fully manages the secrets.

Makefile
```diff
@@ -441,101 +435,14 @@ ifeq ($(shell uname -s),Darwin)
...
-		-e ARO_ADOPT_BY_HIVE="true" \
-		-e MOCK_MSI_TENANT_ID \
-		-e MOCK_MSI_CLIENT_ID \
-		-e MOCK_MSI_CERT \
-		--secret aks.kubeconfig,target=/app/secrets/aks.kubeconfig \
-		--secret proxy-client.key,target=/app/secrets/proxy-client.key \
-		--secret proxy-client.crt,target=/app/secrets/proxy-client.crt \
-		--secret proxy.crt,target=/app/secrets/proxy.crt \
-		$(LOCAL_ARO_RP_IMAGE):$(VERSION) rp
+run-rp:
```
SudoBrendan, line 445: since we removed `podman-secrets`, which relied on `aks.kubeconfig` - can we add `aks.kubeconfig` as a dependent target of `run-rp`?

shubhadapaithankar: I completely agree with you. I've added `aks.kubeconfig` as a dependency for the` run-rp` target

docker-compose.yml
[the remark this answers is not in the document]
shubhadapaithankar: I've added aks.kubeconfig as a dependency for the run-rp target

SudoBrendan (dismissed):

slawande2 (dismissed):

slawande2 (approved):

PR #3884, opened 2024-10-04
SDK2: Replace SDK to new one in util/cluster

bitoku:

This replaces old SDK usages to the new one.
This is used for hack script to create a cluster and e2e cluster creation.

### Test plan for issue:

ci, e2e

no

not a production feature.


tsatam (approved):
LGTM if we get passing E2E

PR #3885, opened 2024-10-04
SDK2: Ensure service endpoints track2

bitoku:

Fixes https://issues.redhat.com/browse/ARO-4665 and https://issues.redhat.com/browse/ARO-7316

This PR replaces old SDK to the new one.

### Test plan for issue:

ci, e2e

no

Creating a cluster can verify the change.
It can be checked in e2e.


pkg/cluster/ensureendpoints_test.go
```diff
@@ -19,88 +23,438 @@ var (
...
+func TestAddEndpointsToSubnets(t *testing.T) {
+	for _, tt := range []struct {
+		name           string
+		subnet         *armnetwork.Subnet
+		newEndpoints   []string
+		expectedSubnet *armnetwork.Subnet
+		expectedResult bool
+	}{
+		{
+			name: "addEndpointsToSubnet should  when the subnet contains all new endpoints in succeeded state",
```
kimorris27, line 282: Nit:

```suggestion
			name: "addEndpointsToSubnet should do nothing when the subnet contains all new endpoints in succeeded state",
```

kimorris27 (approved):
LGTM except for a tiny nit.

kimorris27:
One question though: how does this address the cluster install failure?

bitoku:
@kimorris27 

We have seen some install failures about authorization in this step.
It should be caused by the AAD propagation delay.

We can wrap the step with `AuthorizationRetryingAction`, but it will retry the whole step again.
Because we don't make steps idempotent, retrying the whole step might cause different issues.

By updating the old SDK to the new one, we can use built-in retry strategy in the new SDK.
This will retry just the failed API, so it's a safer way than `AuthorizationRetryingAction`.

Aside from the CIF, we should update old SDKs anyway.

kimorris27 (approved):
> @kimorris27
> 
> We have seen some install failures about authorization in this step. It should be caused by the AAD propagation delay.
> 
> We can wrap the step with `AuthorizationRetryingAction`, but it will retry the whole step again. Because we don't make steps idempotent, retrying the whole step might cause different issues.
> 
> By updating the old SDK to the new one, we can use built-in retry strategy in the new SDK. This will retry just the failed API, so it's a safer way than `AuthorizationRetryingAction`.
> 
> Aside from the CIF, we should update old SDKs anyway.

Oh, cool. I didn't realize the new SDK clients had that built-in retry strategy. That makes sense!

pkg/cluster/ensureendpoints.go
```diff
@@ -36,3 +64,49 @@ func (m *manager) getSubnetIds() ([]string, error) {
...
 	return subnets, nil
 }
+
+// addEndpointsToSubnet adds the endpoints (that either are missing in subnet
+// or aren't in succeeded state in the subnet) to the subnet and returns the updated subnet
+func addEndpointsToSubnet(endpoints []string, subnet *armnetwork.Subnet) (subnetChanged bool) {
+	for _, endpoint := range endpoints {
+		endpointFound, serviceEndpointPtr := subnetContainsEndpoint(subnet, endpoint)
+
+		if !endpointFound || *serviceEndpointPtr.ProvisioningState != armnetwork.ProvisioningStateSucceeded {
```
dkeohane, line 74: Can ProvisioningState ever be in an error state? We wouldn't want to add it then I'm guessing

dkeohane (commented):
Left a quick review, didn't look through unit tests added.

pkg/cluster/cluster.go
```diff
@@ -98,11 +98,12 @@ type manager struct {
 	fpPrivateEndpoints       network.PrivateEndpointsClient // TODO: use armFPPrivateEndpoints instead.
 	armFPPrivateEndpoints    armnetwork.PrivateEndpointsClient
 	armRPPrivateLinkServices armnetwork.PrivateLinkServicesClient
+	armSubnets               armnetwork.SubnetsClient
 	userAssignedIdentities   armmsi.UserAssignedIdentitiesClient
 
 	dns     dns.Manager
 	storage storage.Manager
-	subnet  subnet.Manager
+	subnet  subnet.Manager // TODO: use armSubnet instead.
```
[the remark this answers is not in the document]
bitoku: It should be removed when we get rid of all usages of old SDKs.
It's a part of https://issues.redhat.com/browse/ARO-4665 so I don't think we card it up.

pkg/cluster/ensureendpoints.go
bitoku: This is a copy/paste from the old code.
Also this function adds endpoints that are not available in the subnet, so it's supposed to be ok.

jaitaiwan (approved):
LGTM. 1 small nit

kimorris27 (approved):

UlrichSchlueter (approved):
Code LGTM

Has this been tested against a private cluster ?

bitoku:
not yet. I'll test it in local dev rp.

UlrichSchlueter:
@bitoku never mind, I misread one of the comments

bitoku:
FYI, it worked for a private cluster as well.

PR #3886, opened 2024-10-04
MacOS `make generate` fix - regenerate certain mocks using source mode instead of reflect mode

kimorris27:

No Jira

I found after updating my OS to Sonoma 14.7, which included updating the Command Line Tools for Xcode from 15.3 to 16.0, that `make generate` was failing for certain packages and was also outputting some linker warnings. I can't explain exactly why it works, but I found that having `mockgen` use source mode instead of reflect mode for the failing packages fixed things. The linker warnings about a deprecated `-ld_classic`, which were noted in the [CLT for Xcode 16.0 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-16-release-notes#Linking), also went away, which is interesting to note.

I could try to dig into the reasoning for this change further, but I don't think it'd be a great use of time.

I've encountered multiple issues related to reflect mode in my work recently, which makes me wonder if we should switch to using source mode throughout the repository.

### Test plan for issue:

Tested locally to confirm that it fixes `make generate` on my Mac and unit test still work, and then ran `make ci-rp` to make sure it didn't break anything in the containerized process either.

N/A

Not a production change; only affects generation of mock interfaces that are used in unit tests.


kimorris27:
Skipping E2E since this PR only impacts unit tests.

tsatam (approved):
LGTM, I tested `make generate` on Linux and it works fine and doesn't create a diff, our CI checks back that up too. 

I'd be incredibly disappointed if somehow changing our mock generation impacts production flows in any way, but out of an abundance of caution I say we should wait for green E2E on this before merge.

PR #3888, opened 2024-10-08
removed flowcontrol.apiserver.k8s.io/v1beta2 for 4.16

yjst2012:
K8s1.29 (and OCP4.16) has deprecated flowcontrol.apiserver.k8s.io/v1beta2 
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-29
the pr is to address this via "make discoverycache"

Fixes ARO-10647

removed flowcontrol.apiserver.k8s.io/v1beta2 which is deprecated by k8s 1.29 and OCP4.16

### Test plan for issue:
existing unit and E2E tests.

NA

passing unit and E2E tests.


pkg/util/dynamichelper/discovery/cache_fallback_discovery_client_test.go
```diff
@@ -37,7 +37,7 @@ func TestVersion(t *testing.T) {
 
 	assetsVersion := strings.TrimSuffix(string(b), "\n")
 	// NOTE: This is checking for the version of the oldest supported minor stream.
```
bitoku, line 39: We should change this comment probably?
According to the comment, the version should be the oldest supported minor stream, which is 4.13.

bitoku (dismissed):
I'm not sure if we should remove the deprecated API even though there are some old clusters in prod.
I think it shouldn't be a big deal because IIUC the cache is only used as a fallback when the cluster isn't responsive.

pkg/util/dynamichelper/discovery/cache_fallback_discovery_client_test.go
yjst2012: Thanks @bitoku , I've removed that comment as it is inaccurate the version is just a string read from a test file that has whatever people put there.

as to if we should remove the deprecated API, I'm not very sure as well, maybe a good question for TL @hlipsig @bennerv what do you think🤔

hawkowl:
I agree, we don't need to remove this from the cache. Since it's just a cache and that gets invalidated, and since we don't have any functions that create/read/update those objects, it'll never affect us.

PR #3889, opened 2024-10-08
SDK2: Remove old SDK from Dynamic/ValidateQuota

bitoku:

This PR replaces the old SDK to the new SDK2.

It affects Dynamic pkg (used in validateResources step) and ValidateQuota.go (used in openshiftcluster_putorpatch.go).

### Test plan for issue:

e2e, ci
e2e creates a cluster, which verifies the validateResource step and ValidateQuota.

na

Cluster creation will verify the change


kimorris27 (dismissed):

pkg/validate/dynamic/dynamic.go
```diff
@@ -597,7 +612,7 @@ func (dv *dynamic) validateCIDRRanges(ctx context.Context, subnets []Subnet, add
 func (dv *dynamic) validateVnetLocation(ctx context.Context, vnetr azure.Resource, location string) error {
 	dv.log.Print("validateVnetLocation")
 
-	vnet, err := dv.virtualNetworks.Get(ctx, vnetr.ResourceGroup, vnetr.ResourceName, "")
+	vnet, err := dv.virtualNetworks.Get(ctx, vnetr.ResourceGroup, vnetr.ResourceName, nil)
```
jaitaiwan, line 615: I think we should make it explicit that we're deliberately not providing options here... e.g.:

```
options := nil
vnet, err := dv.virtualNetworks.Get(ctx, vnetr.ResourceGroup, vnetr.ResourceName, options)
```

I think that these options may end up being relevant in the future

jaitaiwan (changes requested):
Mostly LGTM. I think there's some convention questions as well as a question around why the use of v2 vs v6 (the current version in the repo)

pkg/frontend/quota_test.go
```diff
@@ -7,14 +7,14 @@ import (
 	"context"
 	"testing"
 
+	sdknetwork "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v2"
```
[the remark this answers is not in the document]
bitoku: I use `armnetwork` for the client of armnetwork and use `sdknetwork` for usages of SDK.
These are sometimes used at the same time, and to keep them consistent, I think we should use `sdknetwork` and `armnetwork` in this way every time.
You can see this usage in here as well.
https://github.com/Azure/ARO-RP/blob/69378fb634f2bd861d6d35e1aad8ed58f63dc410/pkg/util/cluster/cluster.go

pkg/util/azureclient/azuresdk/armnetwork/usage.go
```diff
@@ -0,0 +1,27 @@
+package armnetwork
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v2"
```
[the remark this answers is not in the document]
bitoku: Because v2 was already being used when I started.
I don't object to using v6, but we have to replace all of them at once because we need to change clients first.
We can do it after we merge current PRs.

pkg/validate/dynamic/dynamic.go
bitoku: For a few reasons, I'd like to leave it as it is:
1. I think the current way is better, because in terms of explicitness, there's no difference IMHO, and the current way is simpler than the suggested way.
2. I've done some migrations and most of them didn't use this argument. If we change them, it will make a lot of redundant lines.
3. This PR aims to migrate from old SDK to new SDK. Since the old one didn't do in that way, changing the way is out of scope.

jaitaiwan (changes requested):
Thanks to hawkowl we realised there are fakes etc available so we no longer have need for generating our own mocks. See pkg/util/liveconfig/hive_test.go in my PR https://github.com/Azure/ARO-RP/pull/3889/files

bitoku:
it's better to keep the code as it is because the PR aims to replace the usage of mgmt clients, and changes of the way of testing (or change the way of listing) is out of scope.
Also this is just a migration, so it's dangerous to change the test code at the same time though it's not necessary.
We should keep the test code the same as much as possible otherwise we can't ensure that the behaviour hasn't been changed.

pkg/validate/dynamic/loadbalancerprofile_test.go
```diff
@@ -8,13 +8,14 @@ import (
 	"strconv"
 	"testing"
 
-	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
+	sdknetwork "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v2"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/sirupsen/logrus"
 	"go.uber.org/mock/gomock"
+	"k8s.io/utils/ptr"
```
jaitaiwan, line 15: Why are we using this lib instead of just going with `to.Ptr`?

jaitaiwan (dismissed):
LGTM; 1 question

kimorris27 (approved):

pkg/validate/dynamic/loadbalancerprofile_test.go
bitoku: Because it's already deprecated.
https://github.com/Azure/go-autorest

jaitaiwan: The `to` functionality moved into the core sdk: `github.com/Azure/azure-sdk-for-go/sdk/azcore/to`

bitoku: There are already usages of `ptr.To` in other places. We should discuss which should be used, but anyway we need to change the code from `to.Ptr*` to `to.Ptr` or `ptr.To`.
I don't have any preference, but I don't think it's a blocker for this PR.

bitoku: https://redhat-internal.slack.com/archives/C02ULBRS68M/p1731940533915429

jaitaiwan (approved):
Changes look reasonable. LGTM

PR #3891, opened 2024-10-08
Fix Managed Identity Cluster creation dynamic validation flow

rajdeepc2792:
Jira - https://issues.redhat.com/browse/ARO-11049

Fixes the Managed Identity Cluster Creation flow until before the Federated Credential Creation

The PR resolves the bugs causing issues for the MIWI cluster creation during the dynamic validation flow.

### Test plan for issue:
Updated Unit Test cases
Tested the local cluster creation flow till the federated identity creation

No

Tested in the local environment
e2e testing will be done for the flow after the feature is being tested.

### DIFF
--- a/pkg/util/azureclient/azuresdk/armauthorization/roledefinitions.go
+++ b/pkg/util/azureclient/azuresdk/armauthorization/roledefinitions.go
@@ -5,6 +5,7 @@ package armauthorization
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
@@ -17,13 +18,20 @@ type RoleDefinitionsClient interface {
 
 type ArmRoleDefinitionsClient struct {
 	*armauthorization.RoleDefinitionsClient
+	subscriptionID string
 }
 
 var _ RoleDefinitionsClient = &ArmRoleDefinitionsClient{}
 
-func NewArmRoleDefinitionsClient(credential azcore.TokenCredential, options *arm.ClientOptions) (*ArmRoleDefinitionsClient, error) {
+func NewArmRoleDefinitionsClient(credential azcore.TokenCredential, subscriptionID string, options *arm.ClientOptions) (*ArmRoleDefinitionsClient, error) {
 	client, err := armauthorization.NewRoleDefinitionsClient(credential, options)
 	return &ArmRoleDefinitionsClient{
 		RoleDefinitionsClient: client,
+		subscriptionID:        subscriptionID,
 	}, err
+}
+
+func (client ArmRoleDefinitionsClient) GetByID(ctx context.Context, roleDefinitionID string, options *armauthorization.RoleDefinitionsClientGetByIDOptions) (armauthorization.RoleDefinitionsClientGetByIDResponse, error) {
+	roleID := fmt.Sprintf("/subscriptions/%s%s", client.subscriptionID, roleDefinitionID)
+	return client.RoleDefinitionsClient.GetByID(ctx, roleID, options)
 }
--- a/pkg/validate/dynamic/platformworkloadidentityprofile_test.go
+++ b/pkg/validate/dynamic/platformworkloadidentityprofile_test.go
@@ -201,7 +201,7 @@ func TestValidatePlatformWorkloadIdentityProfile(t *testing.T) {
 				},
 			},
 			mocks: func(roleDefinitions *mock_armauthorization.MockRoleDefinitionsClient) {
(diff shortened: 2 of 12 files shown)


pkg/util/azureclient/azuresdk/armauthorization/roledefinitions.go
```diff
@@ -17,13 +18,20 @@ type RoleDefinitionsClient interface {
...
+func NewArmRoleDefinitionsClient(credential azcore.TokenCredential, subscriptionID string, options *arm.ClientOptions) (*ArmRoleDefinitionsClient, error) {
 	client, err := armauthorization.NewRoleDefinitionsClient(credential, options)
 	return &ArmRoleDefinitionsClient{
 		RoleDefinitionsClient: client,
+		subscriptionID:        subscriptionID,
 	}, err
 }
+
+func (client ArmRoleDefinitionsClient) GetByID(ctx context.Context, roleDefinitionID string, options *armauthorization.RoleDefinitionsClientGetByIDOptions) (armauthorization.RoleDefinitionsClientGetByIDResponse, error) {
+	roleID := fmt.Sprintf("/subscriptions/%s%s", client.subscriptionID, roleDefinitionID)
```
kimorris27, line 35: I recommend reworking this so that callers of this function only have to pass in the GUID portion of the role definition ID. Even if you were to explain it in a doc comment, I see potential for confusing errors if the caller doesn't understand that they have to pass `/providers/Microsoft.Authorization/roleDefinitions/xxx...` and that the leading slash at the front of the string is mandatory.

I had to trace through a few layers of function calls to understand what was going on here myself, but that wouldn't be necessary if the logic related to formatting the role definition ID was encapsulated within the role definition client.

kimorris27 (changes requested):
LGTM, and I tested it locally too! I just have one relatively unimportant recommendation.

pkg/validate/dynamic/platformworkloadidentityprofile_test.go
```diff
@@ -201,7 +201,7 @@ func TestValidatePlatformWorkloadIdentityProfile(t *testing.T) {
 				},
 			},
 			mocks: func(roleDefinitions *mock_armauthorization.MockRoleDefinitionsClient) {
-				roleDefinitions.EXPECT().GetByID(ctx, rbac.RoleAzureRedHatOpenShiftFederatedCredentialRole, &sdkauthorization.RoleDefinitionsClientGetByIDOptions{}).Return(msiRequiredPermissions, nil)
+				roleDefinitions.EXPECT().GetByID(ctx, fmt.Sprintf("/providers/Microsoft.Authorization/roleDefinitions/%s", rbac.RoleAzureRedHatOpenShiftFederatedCredentialRole), &sdkauthorization.RoleDefinitionsClientGetByIDOptions{}).Return(msiRequiredPermissions, nil)
```
cadenmarchese, line 204: Total nitpick from me, but I think this deserves a var since it's used so many times.
```suggestion
				roleDefinitions.EXPECT().GetByID(ctx, path, rbac.RoleAzureRedHatOpenShiftFederatedCredentialRole), &sdkauthorization.RoleDefinitionsClientGetByIDOptions{}).Return(msiRequiredPermissions, nil)
```
Or something like that.

pkg/validate/dynamic/platformworkloadidentityprofile_test.go
```diff
@@ -232,7 +232,7 @@ func TestValidatePlatformWorkloadIdentityProfile(t *testing.T) {
...
```
cadenmarchese, line 235: Same comment as above.

pkg/util/azureclient/azuresdk/armauthorization/roledefinitions.go
rajdeepc2792: Updated the usage here to only take the roleDefinitionID, please confirm.

pkg/validate/dynamic/platformworkloadidentityprofile_test.go
rajdeepc2792: Changing the GetByID function got rid of the whole `fmt.Sprintf` here.

pkg/validate/dynamic/platformworkloadidentityprofile_test.go
rajdeepc2792: Changing the GetByID function got rid of the whole `fmt.Sprintf` here.

rajdeepc2792:
Not making the change for this in the PR:- https://github.com/Azure/ARO-RP/commit/82f41dc62913704102947734dd7e7fb07aca8048
This can go in after the changes are confirmed.

kimorris27 (approved):

tsatam (approved):

cadenmarchese (approved):

PR #3892, opened 2024-10-09
bump grpc to 1.65.0 wrt OCPBUGS-38376

yjst2012:

fixes https://issues.redhat.com/browse/ARO-11060

the fix for OCPBUGS-38376 is to upgrade google.golang.org/grpc to v1.65.0, and ARO_RP is still on a lower version of grpc, so bump it too

### Test plan for issue:
existing unit and e2e tests

NA

existing unit and e2e tests


yjst2012:
looks like grpc bump has been done in another PR, withdrawing this.

PR #3898, opened 2024-10-10
Update CLI integration test

bitoku:

Fixes https://issues.redhat.com/browse/ARO-8293

This PR implements a scenario-based integration test for all subcommands.
This will increase the test coverage in ARO-RP repo, and will reduce the test execution time in upstream because it doesn't create a cluster for each subcommand.

I'm going to open a PR in upstream after this PR is merged.

### Test plan for issue:

Run `azdev test`.
https://github.com/Azure/ARO-RP/blob/master/docs/az-aro-python-development.md#tests

na

not production feature.

### DIFF
--- a/python/az/aro/azext_aro/tests/latest/integration/test_aro_scenario.py
+++ b/python/az/aro/azext_aro/tests/latest/integration/test_aro_scenario.py
@@ -17,8 +17,8 @@ logger = get_logger(__name__)
 
 class AroScenarioTests(ScenarioTest):
     @AllowLargeResponse()
-    @ResourceGroupPreparer(random_name_length=28, name_prefix='clitestaro_create', location='eastus')
-    @AROClusterServicePrincipalPreparer(name_prefix='clitestaro_create')
+    @ResourceGroupPreparer(random_name_length=28, name_prefix='cli_test_aro', location='eastus')
+    @AROClusterServicePrincipalPreparer(name_prefix='cli_test_aro')
     def test_aro_public_cluster(self, resource_group):
         from azure.mgmt.core.tools import resource_id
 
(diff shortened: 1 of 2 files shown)


python/az/aro/azext_aro/tests/latest/integration/test_aro_scenario.py
```diff
@@ -2,42 +2,93 @@
...
+        temp_kubeconfig_path = self.create_random_name('kubeconfig', 24) + '.tmp'
+
         self.kwargs.update({
-            'name': 'test1'
+            'name': name,
+            'resource_group': resource_group,
+            'subscription': subscription,
+            'master_subnet': master_subnet,
+            'worker_subnet': worker_subnet,
+            'master_ip_range': '10.{}.{}.0/24'.format(randint(0, 127), randint(0, 255)),
```
tsatam, line 39: We don't need these IP ranges to be random if we're making the vnet these are defined on, right?

tsatam (dismissed):
Looks good, there's a couple non-blocking questions/followups that can either be handled here or in follow-up efforts.

python/az/aro/azext_aro/tests/latest/integration/test_aro_scenario.py
bitoku: That's true, but the upstream code is written like this, and I can't find any downside to write in this way, so I'll keep it as it is.
https://github.com/Azure/azure-cli/blob/8de4e1305646c0bc931bd50f29377761c5fd324f/src/azure-cli/azure/cli/command_modules/aro/tests/latest/test_aro_commands.py#L34-L44

tsatam (approved):

PR #3899, opened 2024-10-10
Quickfix - restore inadvertently overwritten mocks

kimorris27:

No Jira; just addresses a small issue I encountered in my work.

The commands specified in `generate.go` in the `armmsi` package prior to this PR clobbered the contents of one of the `mockgen` commands, resulting in mocks for only one of the two interfaces being written to the file. This PR has `mockgen` write each of the interfaces to separate files to restore the mocks that were overwritten.

### Test plan for issue:

Ran `make generate`

N/A

N/A


bitoku (approved):
nit: 
You can use the old way (`mocks/azureclient/azuresdk/$GOPACKAGE/$GOPACKAGE.go`) if you specify the clients in a single line just like this, I guess.
https://github.com/Azure/ARO-RP/blob/d50158cd9cb9ffaa7cc9225e105f1345826cb268/pkg/util/azureclient/azuresdk/armnetwork/generate.go#L7

tsatam (approved):
change looks good, checks look good, :shipit:

PR #3900, opened 2024-10-10
Quickfix - codify solution to another `make generate` issue on MacOS

kimorris27:

No Jira; addresses a small issue I encountered in my own work.

This ensures that https://github.com/uber-go/mock/issues/185 is always fixed automatically for MacOS users.

I'm not sure why, but I found in the course of my work that I had to run the `codesign` command again even though I figured it was a one time fix. It must be related to the fact that `make install-tools` is rerun each time you `make generate`. So I figured I'd codify the fix.

### Test plan for issue:

Only relevant for MacOS users. Run `make generate` and make sure you don't run into https://github.com/uber-go/mock/issues/185.

N/A

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ discoverycache:
 	$(MAKE) generate
 
 .PHONY: generate
-generate: install-tools fix-macos-mockgen
+generate: install-tools
 	go generate ./...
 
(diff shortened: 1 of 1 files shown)


bitoku (dismissed):
nit:
We may want to change `install-tools` target instead, because it's a kind of part of the installation for macOS users, probably?

PR #3902, opened 2024-10-14
Fix subscription ID of armRPPrivateLinkServices client

bitoku:

Fixes N/A

This PR fixes InvalidAuthenticationTokenTenant in Prod.

The old client used `_env.SubscriptionID()` (see L262 and L264 on the left pane in https://github.com/Azure/ARO-RP/pull/3616/files#diff-f4933755bd63c9e160a2068e8ccb4048e8df1b43f7f4544dcbc7ae2650a07478), but the new client uses `r.SubscriptionID`, which caused the subscription mismatch and caused the InvalidAuthenticationTokenTenant error.
This PR fixes the subscription for `armRPPrivateLinkServices` and `armFPPrivateEndpoints`.

### Test plan for issue:

e2e, e2e in prod

N/A

Cluster creation in prod

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -205,7 +205,7 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


UlrichSchlueter (approved):
LGTM

PR #3903, opened 2024-10-14
Correct convention for OIDC issuer urls

slawande2:

[ARO-10948](https://issues.redhat.com/browse/ARO-10948)

Currently, an OIDC issuer URL looks like: `https://eastus.oic.aro.azure-test.net/oic-e97c0f2c-38b6-4741-afb5-cbfc20902d80`
Currently, the convention is `<Oic>-<document ID>`
Instead, it should follow the below convention:
`https://<region>.oic.aro.azure.com/<tenant-id>/<guid>` 

### Test plan for issue:
Passed all unit tests

not yet

### DIFF
--- a/pkg/cluster/delete.go
+++ b/pkg/cluster/delete.go
@@ -451,7 +451,7 @@ func (m *manager) Delete(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		err = oidcbuilder.DeleteOidcFolder(ctx, env.OIDCBlobDirectoryPrefix+m.doc.ID, azBlobClient)
+		err = oidcbuilder.DeleteOidcFolder(ctx, fmt.Sprintf("%s/%s", m.subscriptionDoc.Subscription.Properties.TenantID, m.doc.ID), azBlobClient)
 		if err != nil {
 			return err
 		}
--- a/pkg/cluster/delete.go
+++ b/pkg/cluster/delete.go
@@ -448,11 +448,10 @@ func (m *manager) Delete(ctx context.Context) error {
 		m.log.Printf("deleting OIDC configuration")
 		blobContainerURL := oidcbuilder.GenerateBlobContainerURL(m.env)
 		azBlobClient, err := m.rpBlob.GetAZBlobClient(blobContainerURL, &azblob.ClientOptions{})
-		OIDCIssuerPath := m.subscriptionDoc.Subscription.Properties.TenantID
 		if err != nil {
 			return err
 		}
-		err = oidcbuilder.DeleteOidcFolder(ctx, fmt.Sprintf("%s/%s", OIDCIssuerPath, m.doc.ID), azBlobClient)
+		err = oidcbuilder.DeleteOidcFolder(ctx, oidcbuilder.GetBlobName(m.subscriptionDoc.Subscription.Properties.TenantID, m.doc.ID), azBlobClient)
 		if err != nil {
 			return err
 		}
--- a/pkg/cluster/delete.go
+++ b/pkg/cluster/delete.go
@@ -448,10 +448,11 @@ func (m *manager) Delete(ctx context.Context) error {
 		m.log.Printf("deleting OIDC configuration")
 		blobContainerURL := oidcbuilder.GenerateBlobContainerURL(m.env)
 		azBlobClient, err := m.rpBlob.GetAZBlobClient(blobContainerURL, &azblob.ClientOptions{})
+		OIDCIssuerPath := m.subscriptionDoc.Subscription.Properties.TenantID
 		if err != nil {
 			return err
 		}
-		err = oidcbuilder.DeleteOidcFolder(ctx, fmt.Sprintf("%s/%s", m.subscriptionDoc.Subscription.Properties.TenantID, m.doc.ID), azBlobClient)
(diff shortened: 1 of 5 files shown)


jaitaiwan (dismissed):
LGTM

pkg/cluster/delete.go
```diff
@@ -451,7 +451,7 @@ func (m *manager) Delete(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		err = oidcbuilder.DeleteOidcFolder(ctx, env.OIDCBlobDirectoryPrefix+m.doc.ID, azBlobClient)
```
kimorris27, line 454: Nit: `OIDCBlobDirectoryPrefix` can be removed from `pkg/env/env.go` since it's only references are removed in this PR.

kimorris27 (changes requested):
LGTM, but I do have one small request that will help keep the codebase clean.

pkg/cluster/delete.go
```diff
@@ -451,7 +451,7 @@ func (m *manager) Delete(ctx context.Context) error {
...
+		err = oidcbuilder.DeleteOidcFolder(ctx, fmt.Sprintf("%s/%s", m.subscriptionDoc.Subscription.Properties.TenantID, m.doc.ID), azBlobClient)
```
bitoku, line 454: I'd suggest making a new variable to represent the OICD issuer path and using it, instead of making the path ad-hoc.
It can be reused easily and less error-prone.

pkg/cluster/deploybaseresources_test.go
```diff
@@ -1407,11 +1407,21 @@ func TestCreateOIDC(t *testing.T) {
...
+				},
+			},
+		},
+	}
 	storageWebEndpointForDev := oidcStorageAccountName + ".web." + azureclient.PublicCloud.StorageEndpointSuffix
 	resourceID := "/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/resourceGroup/providers/Microsoft.RedHatOpenShift/openShiftClusters/resourceName"
 	blobContainerURL := fmt.Sprintf("https://%s.blob.%s/%s", oidcStorageAccountName, azureclient.PublicCloud.StorageEndpointSuffix, oidcbuilder.WebContainer)
-	prodOIDCIssuer := fmt.Sprintf("https://%s/%s%s", afdEndpoint, env.OIDCBlobDirectoryPrefix, clusterID)
-	devOIDCIssuer := fmt.Sprintf("https://%s/%s%s", storageWebEndpointForDev, env.OIDCBlobDirectoryPrefix, clusterID)
+	prodOIDCIssuer := fmt.Sprintf("https://%s/%s%s", afdEndpoint, m.subscriptionDoc.Subscription.Properties.TenantID, clusterID)
```
bitoku, line 1423: I'm curious why it is `"https://%s/%s%s"` not `"https://%s/%s/%s"`

pkg/cluster/delete.go
kimorris27: @slawande2, I saw the change you made in response to this. I think @bitoku was saying that you could make the whole thing (i.e. `fmt.Sprintf("%s/%s", m.subscriptionDoc.Subscription.Properties.TenantID, m.doc.ID)`) a constant.

You can't use the `const` keyword though, since the value depends on the live state of the cluster manager object. I would make a utility method on the cluster manager that returns `fmt.Sprintf("%s/%s", m.subscriptionDoc.Subscription.Properties.TenantID, m.doc.ID)` and use that.

pkg/cluster/delete.go
```diff
@@ -448,10 +448,11 @@ func (m *manager) Delete(ctx context.Context) error {
 		m.log.Printf("deleting OIDC configuration")
 		blobContainerURL := oidcbuilder.GenerateBlobContainerURL(m.env)
 		azBlobClient, err := m.rpBlob.GetAZBlobClient(blobContainerURL, &azblob.ClientOptions{})
+		OIDCIssuerPath := m.subscriptionDoc.Subscription.Properties.TenantID
```
niontive, line 451: nit: I'd just call this local variable `tenantID`.

pkg/cluster/delete.go
```diff
@@ -448,10 +448,11 @@ func (m *manager) Delete(ctx context.Context) error {
...
 		if err != nil {
 			return err
 		}
-		err = oidcbuilder.DeleteOidcFolder(ctx, env.OIDCBlobDirectoryPrefix+m.doc.ID, azBlobClient)
+		err = oidcbuilder.DeleteOidcFolder(ctx, fmt.Sprintf("%s/%s", OIDCIssuerPath, m.doc.ID), azBlobClient)
```
niontive, line 455: Since we use `fmt.Sprintf("%s/%s", OIDCIssuerPath, m.doc.ID)` in multiple places, write a function in `oidcbuilder.go` to call this. 

Something like:
```
func GetBlobName(tenantID string, docID string) string {
   return fmt.Sprintf("%s/%s", tenantID, docID)
}
```

pkg/cluster/delete.go
kimorris27: Resolving this thread since the issue's been addressed over in https://github.com/Azure/ARO-RP/pull/3903#discussion_r1803490866

kimorris27 (approved):
LGTM! IMO this can merge without E2E.

bitoku (approved):

shubhadapaithankar (approved):
LGTM

PR #3911, opened 2024-10-21
[ARO-1885] Implement OperatorFlagsMergeStrategy

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-1885

Allows us to 'reset' feature flags to their defaults by providing ones that we wish to layer on top of said defaults.

Broken out of #3571 because I needed it for some tests. Based off work by @SrinivasAtmakuri in another PR.

### Test plan for issue:

Unit tests

SOPs for operator flag working might be needed? 

Unit tests, plus it won't do anything in production unless we make it do so

### DIFF
--- a/pkg/api/admin/operatorflags_test.go
+++ b/pkg/api/admin/operatorflags_test.go
@@ -1,5 +1,8 @@
 package admin
 
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
 import (
 	"testing"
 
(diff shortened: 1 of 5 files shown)


SrinivasAtmakuri (approved):

pkg/api/admin/operatorflags_test.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+func TestOperatorFlagsMergeStrategy(t *testing.T) {
+	tests := []struct {
+		name    string
+		oc      *api.OpenShiftCluster
+		wantOc  *api.OpenShiftCluster
```
rhamitarora, line 17: The variable `wantoc` is declared, but its usage is not shown anywhere.

bitoku (commented):
I think we shouldn't change the struct, but we should change only API because OperatorFlagsMergeStrategy is something only used for this API ad-hoc, and essentially not an attribute of the OpenShiftCluster struct.

hawkowl:
@bitoku so that struct is the admin API only. The canonical object for OpenShiftCluster is just in api/, api/admin/ and all the others are the converted external API versions. 

I also did try pulling that struct out and I had some problems getting a test working, so I think this is the easiest way forward.

bitoku (approved):

PR #3913, opened 2024-10-21
Use correct error type when handling federated identity credential deletion

tsatam:

Fixes no Jira

When deleting a federated identity credential, we were incorrectly casting the error to a api.CloudError type, however the error returned by the FederatedIdentityCredentialsClient is an `azcore.ResponseError` type. This PR updates the type used when typecasting the error, and additionally adds a run-time type check to gracefully handle errors not of the expected type as "unknown errors" 

(note - these types of unknown errors will fail cluster deletion, while we generally want identifiably "Not Found" errors to continue deletion as that is the desired state). 

### Test plan for issue:

- [ ] Local cluster deletion no longer panics

No

Feature not yet in production

### DIFF
--- a/pkg/cluster/delete.go
+++ b/pkg/cluster/delete.go
@@ -6,6 +6,7 @@ package cluster
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"sort"
@@ -426,8 +427,8 @@ func (m *manager) deleteFederatedCredentials(ctx context.Context) error {
 				&armmsi.FederatedIdentityCredentialsClientDeleteOptions{},
 			)
 			if err != nil {
-				cloudErr, ok := err.(*azcore.ResponseError)
-				if ok && cloudErr.StatusCode == http.StatusNotFound {
+				var respErr *azcore.ResponseError
+				if errors.As(err, &respErr); respErr.StatusCode == http.StatusNotFound {
 					m.log.Errorf("federated identity credentials not found for %s: %v", identity.ResourceID, err.Error())
 					continue
 				} else {


pkg/cluster/delete.go
```diff
@@ -425,13 +426,12 @@ func (m *manager) deleteFederatedCredentials(ctx context.Context) error {
 				&armmsi.FederatedIdentityCredentialsClientDeleteOptions{},
 			)
 			if err != nil {
-				cloudErr := err.(*api.CloudError)
-
-				if cloudErr.StatusCode == http.StatusNotFound {
-					m.log.Errorf("federated identity credentials not found for %s: %v", identity.ResourceID, cloudErr.Error())
+				cloudErr, ok := err.(*azcore.ResponseError)
```
stevekuznetsov, line 429: Are you able to use `errors.As()`? You don't have to worry about a panic and can handle wrapped errors as well.

tsatam: I'll do this, thanks for the suggestion!

cadenmarchese (approved):
thanks for fixing - I also faced this issue.

pkg/cluster/delete.go
```diff
@@ -425,13 +427,12 @@ func (m *manager) deleteFederatedCredentials(ctx context.Context) error {
 				&armmsi.FederatedIdentityCredentialsClientDeleteOptions{},
 			)
 			if err != nil {
-				cloudErr := err.(*api.CloudError)
-
-				if cloudErr.StatusCode == http.StatusNotFound {
-					m.log.Errorf("federated identity credentials not found for %s: %v", identity.ResourceID, cloudErr.Error())
+				var respErr *azcore.ResponseError
+				if errors.As(err, &respErr); respErr.StatusCode == http.StatusNotFound {
```
bitoku, line 431: You can use this.
https://github.com/Azure/ARO-RP/blob/2143cb81353867aa5ff21d15dfcc4f51cc8c16c2/pkg/util/azureclient/azuresdk/errors/errors.go#L14-L17

cadenmarchese:
Going to get this one in as its blocking mi/wi cluster deletion. We can address @bitoku 's comments in a followup PR if needed.

pkg/cluster/delete.go
tsatam: https://github.com/Azure/ARO-RP/pull/3915

PR #3914, opened 2024-10-22
Tolerate failures to delete federated credentials during deletion

cadenmarchese:

Fixes no JIRA

Our federated credential built-in role doesn't yet have permissions to clean up federated credentials, so currently wi/mi installs are failing to clean them up. For the time being, we need to tolerate failures to clean up federated credentials rather than fail the whole delete.

### Test plan for issue:

Tested this locally on deletion of a wi/mi cluster. Deletion succeeded.


tsatam (approved):
LGTM for now, but IMO we should have follow-up work to identify and fix deletion failures.

PR #3916, opened 2024-10-22
Update role definition names and keys used for MIWI

bennerv:
Role definition names after feedback and more [explicit role names](https://docs.google.com/spreadsheets/d/1D2o6Wi54yq75SLfIkP5wIMRamwre7pVDO9eJpCXRhXI/edit?gid=0#gid=0).

Role definition names after feedback and more explicit role names.

### Test plan for issue:

N/A


tsatam (dismissed):
LGTM

cadenmarchese (changes requested):
Can we also update the roleset in pkg/util? 

https://github.com/Azure/ARO-RP/blob/master/pkg/util/rolesets/const.go

bennerv:
>Can we also update the roleset in pkg/util?

Why is it defined in two places?  (actually three, RP config too).

kimorris27:
> > Can we also update the roleset in pkg/util?
> 
> Why is it defined in two places? (actually three, RP config too).

It looks like the `pkg/util/rolesets/const.go` one is only referenced by the dev cluster hack: https://github.com/Azure/ARO-RP/blob/3b6426c8c155e1ec76e09ab47ef1f410412fa7b1/pkg/util/cluster/cluster.go#L702

The hack could probably read it from the env var to knock it down from three places to two.

kimorris27 (approved):

PR #3920, opened 2024-10-23
Fix CheckAccess group expansion bug

gouthamMN:

Fixes: https://issues.redhat.com/browse/ARO-10904

1) This PR consumes [checkaccess-v2-go-sdk](https://github.com/Azure/checkaccess-v2-go-sdk) to use the `CheckAccess V2` API to authorize data plane request and removes the local code that was used to call `CheckAccess v2` API.
2) This also fixes the `group expansion` bug described in the JIRA link above by passing the right `ClaimName` that suffice the following conditions.
```
1. groups claim presented, _claim_names not presented. Pass groups claim to checkAccess API in payload.
2. groups not presented, _claim_names presented. Pass _claim_names to checkAccess.
3. groups not presented, _claim_names not presented. Identity is not in any group, no need to pass anything.
4. groups and _claim_names both presented. Not a valid scenario. 
```

### Test plan for issue:
- New Unit tests and existing E2E.

No

- All Unit tests and existing E2E must pass.


pkg/validate/dynamic/dynamic.go
```diff
@@ -435,41 +435,61 @@ func (dv *dynamic) validateActionsByOID(ctx context.Context, r *azure.Resource,
...
+
+	tokenClaims, err := token.ExtractClaims(t.Token)
+	if err != nil {
+		c.dv.log.Error("Unable to parse the token oid claim: ", err)
+		return err
+	}
+	c.subjectAttributes.ObjectId = tokenClaims.ObjectId
+
+	if tokenClaims.ClaimNames != nil && len(tokenClaims.Groups) == 0 {
+		c.subjectAttributes.ClaimName = client.GroupExpansion
```
stevekuznetsov, line 472: I was expecting to see the helpers for parsing the JWT, extracting groups, walking the graph API when necessary, etc, in the SDK, otherwise every client will need this logic to correctly call it, right?

pkg/validate/dynamic/generate.go
```diff
@@ -9,4 +9,6 @@ package dynamic
 
 //go:generate rm -rf ../../../pkg/util/mocks/$GOPACKAGE
 //go:generate mockgen -destination=../../../pkg/util/mocks/$GOPACKAGE/$GOPACKAGE.go -source=dynamic.go
+//go:generate mockgen -destination=../../../pkg/util/mocks/checkaccess/checkaccess.go github.com/Azure/checkaccess-v2-go-sdk/client RemotePDPClient
 //go:generate goimports -local=github.com/Azure/ARO-RP -e -w ../../../pkg/util/mocks/$GOPACKAGE/$GOPACKAGE.go
+//go:generate goimports -local=github.com/Azure/ARO-RP -e -w ../../../pkg/util/mocks/checkaccess/checkaccess.go
```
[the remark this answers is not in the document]
gouthamMN: We had a discussion on this over teams.

pkg/validate/dynamic/dynamic.go
gouthamMN: Moved this logic to SDK

fahlmant (dismissed):
lgtm

pkg/cluster/cluster.go
```diff
@@ -169,10 +169,11 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 	if err != nil {
 		return nil, err
 	}
-	fpspID, err := token.GetObjectId(t.Token)
+	tokenClaims, err := token.ExtractClaims(t.Token)
 	if err != nil {
 		return nil, err
 	}
+	fpspID := tokenClaims.ObjectId
```
bennerv, line 176: Wondering if we should name the parameter to specifically call it out as the instantiation in the customer's tenant, not our own.

bennerv (approved):
lgtm, mostly nits that i think we should clean up not necessarily introduced in the PR.

pkg/validate/dynamic/dynamic.go
```diff
@@ -446,31 +446,45 @@ type closure struct {
 	resource *azure.Resource
 	actions  []string
 	oid      *string
+	jwtToken *string
+}
+
+func (c *closure) checkAccessAuthReqToken() error {
+	scope := c.dv.azEnv.ResourceManagerEndpoint + "/.default"
```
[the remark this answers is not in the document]
gouthamMN: I removed the usage of `azEnv` wherever possible but wasn't able to remove it from `dynamic` struct as it was referenced in other places too.

hlipsig (approved):

hlipsig:
Agree with Ben's comments, but these can be followed on in smaller PRs.

PR #3928, opened 2024-10-29
ARO-9716 during miwi cluster install remove azure file csi storage class

rajdeepc2792:
https://issues.redhat.com/browse/ARO-9716

Since shared access key is disabled for Cluster & Image Registry Storage Account in case of Workload Identity Cluster, we cannot support default File CSI Storage class. Since there's a chance the file csi driver can use those storage accounts.
ADR:- https://docs.google.com/document/d/150MJvfTSgvY8LWp_iRh9-En1jIxvMPf3NlKTzfn36Bw/edit

### Test plan for issue:
[x] Unit Tests
[x] Test by MIWI Cluster Creation
[x] CI
[x] e2e

No.

Tested by creating MIWI cluster locally

### DIFF
--- a/pkg/cluster/install.go
+++ b/pkg/cluster/install.go
@@ -426,6 +426,7 @@ func (m *manager) Install(ctx context.Context) error {
 			steps.Action(m.configureIngressCertificate),
 			steps.Condition(m.ingressControllerReady, 30*time.Minute, true),
 			steps.Action(m.configureDefaultStorageClass),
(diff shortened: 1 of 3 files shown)


tsatam (approved):
LGTM - I'd like to see the RemoveBootstrap steps extracted to a function, similar to `bootstrap()` and for the MIWI-vs-CSP logic to be reflected in that function, so that it is immediately apparent which steps happen in which contexts.

That can be a follow-up refactor, though.

PR #3930, opened 2024-10-30
disallow duplicate operator identity resource IDs

yithian:

Fixes ARO-11689

This adds a step to static validation that checks to see if operator identities use the same resource ID as any other operator identities and causes the cluster installation to fail if so. Each operator should have its own identity.

### Test plan for issue:

Unit test added

No, this is already documented as a requirement

Unit tests for other static validation cases have already been added and accepted, this just expands on what's already there

### DIFF
--- a/pkg/api/v20240812preview/openshiftcluster_validatestatic.go
+++ b/pkg/api/v20240812preview/openshiftcluster_validatestatic.go
@@ -469,7 +469,14 @@ func (sv openShiftClusterStaticValidator) validatePlatformWorkloadIdentityProfil
 	}
 
 	// Validate the PlatformWorkloadIdentities
+	foundIdentityResourceIDs := map[string]string{}
+
 	for name, p := range pwip.PlatformWorkloadIdentities {
+		if _, present := foundIdentityResourceIDs[strings.ToLower(p.ResourceID)]; present {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, fmt.Sprintf("%s.PlatformWorkloadIdentities", path), "ResourceID %s used by multiple identities.", p.ResourceID)
+		}
+		foundIdentityResourceIDs[strings.ToLower(p.ResourceID)] = ""
+
 		resource, err := azcorearm.ParseResourceID(p.ResourceID)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -469,7 +470,14 @@ func (sv openShiftClusterStaticValidator) validatePlatformWorkloadIdentityProfil
 	}
 
 	// Validate the PlatformWorkloadIdentities
+	foundIdentityResourceIDs := []string{}
```
kimorris27, line 473: Nit: This could be a map instead of a slice so that we can check for duplicates in linear time instead of quadratic time, but pragmatically speaking it doesn't matter given the small number of platform workload identities.

kimorris27 (dismissed):

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -469,7 +470,14 @@ func (sv openShiftClusterStaticValidator) validatePlatformWorkloadIdentityProfil
...
+
 	for name, p := range pwip.PlatformWorkloadIdentities {
+		if slices.Contains(foundIdentityResourceIDs, p.ResourceID) {
```
rajdeepc2792, line 476: I believe, since we have seen resourceIDs allow the case insensitivity, so we should check the map/list considering the same, i.e. `resourceID` is equivalent to `ResourceID`. Disallow the resourceIDs if the case differs but they are still same.

kimorris27: Not directly related to this PR, but on the topic of case sensitivity, how do we handle that for the operator names during dynamic validation? I peeked at the code out of curiosity and didn't find any case sensitivity special handling.

yithian: I amended the commit to be case insensitive (and switched to a map instead of a slice)

kimorris27 (commented):
LGTM after the lint failures are addressed!

kimorris27 (approved):

tsatam (approved):

cadenmarchese (approved):

PR #3931, opened 2024-10-30
Update cosmosdb auth used by update_role_sets.go

cadenmarchese:

Fixes No Jira

Similarly to https://github.com/Azure/ARO-RP/pull/3648/files, update update_role_sets to not use master key / AEAD auth and use database.NewDatabaseClientFromEnv instead.

### Test plan for issue:

Tested it with a local RP + go run ./cmd/aro update-role-sets to confirm the role sets are updated in cosmosdb.

### DIFF
--- a/cmd/aro/update_role_sets.go
+++ b/cmd/aro/update_role_sets.go
@@ -8,14 +8,12 @@ import (
 	"fmt"
 	"os"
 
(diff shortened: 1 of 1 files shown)


kimorris27 (approved):
Works for me locally as well!

bennerv (approved):

PR #3933, opened 2024-10-31
Fix AdminAPI representation of PreconfiguredNSG status

tsatam:

Fixes no Jira

Ensures that the status of the cluster document's `networkProfile.preconfiguredNSG` property is included on Admin API representations of the document. 

### Test plan for issue:

- [x] Ensure AdminAPI representation of clusterdoc includes this property in local env

No

Testing done above

### DIFF
--- a/pkg/api/admin/openshiftcluster_convert.go
+++ b/pkg/api/admin/openshiftcluster_convert.go
@@ -54,6 +54,7 @@ func (c openShiftClusterConverter) ToExternal(oc *api.OpenShiftCluster) interfac
 				APIServerPrivateEndpointIP: oc.Properties.NetworkProfile.APIServerPrivateEndpointIP,
 				GatewayPrivateEndpointIP:   oc.Properties.NetworkProfile.GatewayPrivateEndpointIP,
 				GatewayPrivateLinkID:       oc.Properties.NetworkProfile.GatewayPrivateLinkID,
+				PreconfiguredNSG:           PreconfiguredNSG(oc.Properties.NetworkProfile.PreconfiguredNSG),
 			},
 			MasterProfile: MasterProfile{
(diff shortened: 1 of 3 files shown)


bennerv (approved):

s-fairchild (dismissed):

jhoreman (dismissed):

pkg/api/admin/openshiftcluster_convert.go
```diff
@@ -179,6 +179,7 @@ func (c openShiftClusterConverter) ToExternal(oc *api.OpenShiftCluster) interfac
 	}
 
 	if oc.Identity != nil {
+		out.Identity = &ManagedServiceIdentity{}
```
stevekuznetsov, line 182: You should check if `out.Identity == nil` or this will clobber anything that is already there. This would have been a 100% panic if `out.Identity` were always `nil`. Is it?

jhoreman (approved):

pkg/api/admin/openshiftcluster_convert.go
tsatam: The `out` object is created from scratch earlier up in this function (L18-75), so we don't need to worry about clobbering any existing Identity properties, `out.Identity` is always `nil` at this point.

stevekuznetsov: So are you saying this code would always panic? Your original comment made it seem like it was normally fine and only having a panic in MIWI

tsatam: The if statement this line is contained in, `if oc.Identity != nil`, would only be true for MIWI clusters, but the code within that if block would always panic since `out.Identity` was always nil.

cadenmarchese (approved):

PR #3937, opened 2024-11-01
specify json output for a couple az commands that were missing it

yithian:

n/a

A couple az commands in the `create_miwi_env_file` function in hack/devtools/local_dev_env.sh script didn't specify to use json output, so changing the default output type (eg: to table) breaks things

### Test plan for issue:

Manually run

No, this is a bugfix of a dev script

It will not go to production, it's a dev script

### DIFF
--- a/hack/devtools/local_dev_env.sh
+++ b/hack/devtools/local_dev_env.sh
@@ -129,7 +129,7 @@ get_mock_msi_tenantID() {
 }
 
 get_mock_msi_objectID() {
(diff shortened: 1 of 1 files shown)


kimorris27 (commented):
Thanks for this fix! I took a quick look through the rest of the file and found one more that could be fixed at the same time: the `az deployment group list` on line 302.

yithian:
> Thanks for this fix! I took a quick look through the rest of the file and found one more that could be fixed at the same time: the `az deployment group list` on line 302.

Thanks, added!

kimorris27 (approved):

PR #3938, opened 2024-11-04
Fix test flake in preview API for MIWI

hawkowl:

Fixes a test failing on #3571 

Depending on map key ordering, this test can fail because it will log the capitalised version or vice versa. Now it will return the lowercased version always in the error, which is the same as other parts of the code.

### Test plan for issue:

Is a unit test

N/A

N/A


jaitaiwan (commented):
Even though this was merged, LGTM

PR #3941, opened 2024-11-06
Make PreconfiguredNSG mutable

mociarain:
Hotfix.

Details are here: https://redhat-internal.slack.com/archives/C07UY4A7HM1/p1730882723923239


tsatam (approved):
LGTM. We tested this against a localdev cluster and it performed the desired change.

PR #3946, opened 2024-11-07
ARO-12034 Fix CheckAccessV2 usage for MIWI dynamic validation

rajdeepc2792:

Fixes https://issues.redhat.com/browse/ARO-12034

After https://github.com/Azure/ARO-RP/pull/3920 got merged, the MIWI cluster dynamic validation fails as the access token is mandatory to use checkAccessV2.
But since Platform Workload Identities are used via Federated Credentials or OIDC way, the access token for them can only be fetched after KubeAPI Server is up. So, for the fix, when validating Platform Workload Identity, we won't be using the access token and also won't be using the Group Expansion.

Related Thread:- https://redhat-external.slack.com/archives/C03F6AA3HDH/p1730225417576639

### Test plan for issue:
[x] Unit Test Cases
[x] Local Cluster Creation
[x] CI
[x] E2E

Update the Customer Facing documentation for MIWI, such that the customer should only perform the role assignment directly on the platform workload identity and shouldn't be adding the identities to group for the role assignment.

Testing in local and canary.

### DIFF
--- a/pkg/cluster/validate.go
+++ b/pkg/cluster/validate.go
@@ -6,11 +6,16 @@ package cluster
 import (
 	"context"
 
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/azcore"
 	"github.com/Azure/ARO-RP/pkg/validate"
 )
 
 func (m *manager) validateResources(ctx context.Context) error {
+	var clusterMSICredential azcore.TokenCredential
+	if m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		clusterMSICredential = m.userAssignedIdentities.GetClusterMSICredential()
+	}
 	return validate.NewOpenShiftClusterDynamicValidator(
-		m.log, m.env, m.doc.OpenShiftCluster, m.subscriptionDoc, m.fpAuthorizer, m.armRoleDefinitions, m.platformWorkloadIdentityRolesByVersion,
+		m.log, m.env, m.doc.OpenShiftCluster, m.subscriptionDoc, m.fpAuthorizer, m.armRoleDefinitions, m.platformWorkloadIdentityRolesByVersion, clusterMSICredential,
 	).Dynamic(ctx)
 }
--- a/pkg/validate/dynamic/platformworkloadidentityprofile.go
+++ b/pkg/validate/dynamic/platformworkloadidentityprofile.go
@@ -58,11 +58,6 @@ func (dv *dynamic) ValidatePlatformWorkloadIdentityProfile(ctx context.Context, 
 			"properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities", "There's a mismatch between the required and expected set of platform workload identities for the requested OpenShift minor version '%s'. The required platform workload identities are '%v'", v, requiredOperatorIdentities)
 	}
 
-	err := dv.validateClusterMSI(ctx, oc, roleDefinitions)
-	if err != nil {
-		return err
-	}
-
 	for _, role := range platformWorkloadIdentityRolesByRoleName {
 		roleDefinitionID := stringutils.LastTokenByte(role.RoleDefinitionID, '/')
 		actions, err := getActionsForRoleDefinition(ctx, roleDefinitionID, roleDefinitions)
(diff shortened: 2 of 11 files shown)


pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -76,21 +71,22 @@ func (dv *dynamic) ValidatePlatformWorkloadIdentityProfile(ctx context.Context,
...
+	dv.log.Print("ValidateClusterUserAssignedIdentity")
+
+	for resourceID := range oc.Identity.UserAssignedIdentities {
 		_, err := azure.ParseResourceID(resourceID)
 		if err != nil {
 			return err
 		}
-
-		return dv.validateClusterMSIPermissions(ctx, identity.PrincipalID, oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities, roleDefinitions)
+		return dv.validateClusterMSIPermissions(ctx, oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities, roleDefinitions)
```
kimorris27, line 82: How are we validating the cluster MSI if we parse the MSI resource ID and then do nothing with it?

kimorris27 (commented):
>So, for the fix, when validating Platform Workload Identity, we won't be using the access token and also won't be using the Group Expansion.

I'm a little confused. My reading of the new code is that changes are being made not to the platform identity validation but to the cluster MSI validation, and that the cluster MSI is no longer being validated. I find check access v2 confusing in general though. Could you help me understand it a bit better?

pkg/validate/dynamic/platformworkloadidentityprofile.go
rajdeepc2792: Ohh thanks, previously we needed the principalID so we were iterating over the oc.Identity.UserAssignedIdentites. It is no more needed. We can call the validateClusterMSIPermissions directly.


> My reading of the new code is that changes are being made not to the platform identity validation but to the cluster MSI validation, and that the cluster MSI is no longer being validated. I find check access v2 confusing in general though. Could you help me understand it a bit better?

So for the Cluster MSI we have the access token(i.e. azcore.TokenCredential), so we just use the exact similar CheckAccessV2 flow  as used by FPSP and CSP. Getting OID and AccessToken from the azcore.TokenCredential

But since we don't get credential for the Platform Workload Identity, we just pass OID adding an else case for creating AuthorizationRequest only based on OID and not using claims from the access token here:-

kimorris27: Got it, thank you for humoring my questions!

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -103,7 +87,7 @@ func (dv *dynamic) validateClusterMSIPermissions(ctx context.Context, oid string
 			return err
 		}
 
-		err = dv.validateActionsByOID(ctx, &pid, actions, &oid)
+		err = dv.validateActionsByOID(ctx, &pid, actions, nil)
```
bitoku, line 90: We should change the comment for validateActionsByOID.
https://github.com/Azure/ARO-RP/blob/6c56ef7d153972d00a72a7f93afedfc74e0554c1/pkg/validate/dynamic/dynamic.go#L431
Because oid is nil though it's supposed to be called for MIWI cluster.

pkg/validate/openshiftcluster_validatedynamic.go
```diff
@@ -175,7 +178,24 @@ func (dv *openShiftClusterDynamicValidator) Dynamic(ctx context.Context) error {
...
+			dv.fpAuthorizer,
+			nil,
+			dynamic.AuthorizerClusterUserAssignedIdentity,
+			dv.clusterMSICredential,
+			pdpClient,
+		)
+		err = cmsiDynamic.ValidateClusterUserAssignedIdentity(ctx, dv.oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities, dv.roleDefinitions)
+		if err != nil {
+			return err
+		}
```
bitoku, line 196: It seems to be only used to run `ValidateClusterUserAssignedIdentity`.
If `spDynamic` has `dv.clusterMSICredential`, can `spDynamic` run `ValidateClusterUserAssignedIdentity`, and can we remove `cmsiDynamic`?

rajdeepc2792: We are performing dv.clusterMSICredential as part of `cmsiDynamic` because we need  dv.clusterMSICredential to get the EntraID access token and oid for the jwttoken present in dv.clusterMSICredential.
But all the validations done by `spDynamic` are without the jwttoken, it has been broken up into two different dynamic objects to separate their way of creating the pdpClient's authorization request.

I would keep them separated as they are validations for 2 different type of identities, one is Cluster MSI(work with access token) and other being all the Platform Workload Identities(oid need to be passed, no jwttoken exists).

pkg/validate/dynamic/platformworkloadidentityprofile.go
rajdeepc2792: Yes will change it, it is still need to be non-nil for MIWI cluster though, only Cluster MSI in MIWI cluster will have oid as nil.

cadenmarchese (approved):
tested and working for me. thanks!

pkg/cluster/validate.go
```diff
@@ -6,11 +6,16 @@ package cluster
...
 	"context"
 
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/azcore"
 	"github.com/Azure/ARO-RP/pkg/validate"
 )
 
 func (m *manager) validateResources(ctx context.Context) error {
+	var clusterMSICredential azcore.TokenCredential
+	if m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		clusterMSICredential = m.userAssignedIdentities.GetClusterMSICredential()
```
kimorris27, line 16: I don't think this will work during an `az aro update` since the cluster MSI clients get initialized after dynamic validation. You'll want to move `m.ensureClusterMsiCertificate` and `m.initializeClusterMsiClients` to be before `m.validateResources`.

kimorris27 (changes requested):
LGTM except that there's one spot where it looks to me like it won't work during an `az aro update`. The fix is easy though.

pkg/cluster/validate.go
tsatam: I need a change in parallel to move MSI client instantiation before dynamic validation anyways, in order to validate existing federated identity credentials on new/existing identities. 

@rajdeepc2792 want me to push that commit onto your branch?

tsatam: https://github.com/Azure/ARO-RP/tree/tsatam/ARO-11854-handle-noncluster-federated-credentials it's literally the only change I have made on this branch so far.

rajdeepc2792: I have used the top commit from @tsatam  fork/branch as it is.

kimorris27 (approved):

cadenmarchese (approved):

PR #3947, opened 2024-11-08
ARO-3258: propagate errors of ARO PullSecret controller to ARO operator

lbossis:
Fixes https://issues.redhat.com/browse/ARO-3258

It is needed to propagate errors of ARO PullSecret controller to ARO operator in order to see those errors at the Operator level and to see all the controller errors in one place.

### Test plan for issue:
[x] Unit Test Cases
[x] Local Cluster Creation
[] CI
[] E2E

### Are there unit tests 
Yes, https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/pullsecret/pullsecret_controller_test.go

Running all unit tests locally and verifying tests output

N/A

### DIFF
--- a/pkg/operator/controllers/pullsecret/pullsecret_controller.go
+++ b/pkg/operator/controllers/pullsecret/pullsecret_controller.go
@@ -114,6 +114,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	}
 
 	err = r.Client.Update(ctx, instance)
+	if err == nil {
+		r.ClearConditions(ctx)
+	} else {
+		r.SetDegraded(ctx, err)
+	}
 	return reconcile.Result{}, err
 }
 
--- a/pkg/operator/controllers/pullsecret/pullsecret_controller.go
+++ b/pkg/operator/controllers/pullsecret/pullsecret_controller.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/Azure/ARO-RP/pkg/operator"
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
+	"github.com/Azure/ARO-RP/pkg/operator/controllers/base"
 	"github.com/Azure/ARO-RP/pkg/operator/predicates"
 	"github.com/Azure/ARO-RP/pkg/util/pullsecret"
 )
@@ -43,15 +44,16 @@ var rhKeys = []string{"registry.redhat.io", "cloud.openshift.com", "registry.con
 
 // Reconciler reconciles a Cluster object
 type Reconciler struct {
-	log *logrus.Entry
-
-	client client.Client
+	base.AROController
 }
 
 func NewReconciler(log *logrus.Entry, client client.Client) *Reconciler {
 	return &Reconciler{
-		log:    log,
-		client: client,
+		AROController: base.AROController{
+			Log:    log,
+			Client: client,
+			Name:   ControllerName,
+		},
 	}
 }
 
@@ -65,21 +67,22 @@ func NewReconciler(log *logrus.Entry, client client.Client) *Reconciler {
 //   - If the pull Secret object (which is not owned by the Cluster object)
 //     changes, we'll see the pull Secret object requested.
 func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
-	instance := &arov1alpha1.Cluster{}
-	err := r.client.Get(ctx, types.NamespacedName{Name: arov1alpha1.SingletonClusterName}, instance)
+	instance, err := r.GetCluster(ctx)
 	if err != nil {
+		r.Log.Error(err)
 		return reconcile.Result{}, err
 	}
 
 	if !instance.Spec.OperatorFlags.GetSimpleBoolean(operator.PullSecretEnabled) {
-		r.log.Debug("controller is disabled")
+		r.Log.Debug("controller is disabled")
 		return reconcile.Result{}, nil
 	}
 
-	r.log.Debug("running")
+	r.Log.Debug("running")
 	userSecret := &corev1.Secret{}
-	err = r.client.Get(ctx, pullSecretName, userSecret)
+	err = r.Client.Get(ctx, pullSecretName, userSecret)
 	if err != nil && !kerrors.IsNotFound(err) {
+		r.Log.Error(err)
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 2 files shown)


lbossis:

cadenmarchese (commented):
first pass looks good, just a few suggestions. thanks!

pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
```diff
@@ -285,11 +287,11 @@ func TestPullSecretReconciler(t *testing.T) {
 			ctx := context.Background()
 
 			clientFake := ctrlfake.NewClientBuilder().WithObjects(tt.instance).WithObjects(tt.secrets...).Build()
+			assert.NotNil(t, clientFake)
```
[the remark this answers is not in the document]
lbossis: It should be helpful to manifest nil object occurrence whenever that object is supposed to be not nil and thus to render test case failure early on.

pkg/operator/controllers/pullsecret/pullsecret_controller.go
```diff
@@ -65,36 +67,39 @@ func NewReconciler(log *logrus.Entry, client client.Client) *Reconciler {
...
+		err = r.Client.Get(ctx, types.NamespacedName{Namespace: operator.Namespace, Name: operator.SecretName}, operatorSecret)
 		if err != nil {
+			r.Log.Error(err)
 			return reconcile.Result{}, err
 		}
 
 		// fix pull secret if its broken to have at least the ARO pull secret
 		userSecret, err = r.ensureGlobalPullSecret(ctx, operatorSecret, userSecret)
 		if err != nil {
+			r.Log.Error(err)
```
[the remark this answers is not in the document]
cadenmarchese: Resolving as we've agreed in the JIRA to not set degraded as part of this PR as it could impact user's ability to rotate their pull secret.

fahlmant (dismissed):
lgtm

slawande2 (dismissed):

pkg/operator/controllers/pullsecret/pullsecret_controller.go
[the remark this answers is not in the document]
tsatam: We can safely SetDegraded here without impacting any functionality. The controller will still continue to execute as it would normally, and the controller Degraded status will (as of right now) not be cascaded to the ARO Cluster Operator itself, so this controller being degraded will not e.g. block cluster upgrades.

pkg/operator/controllers/pullsecret/pullsecret_controller.go
```diff
@@ -104,10 +109,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	// - list of Red Hat pull-secret keys in status.
 	instance.Status.RedHatKeysPresent, err = r.parseRedHatKeys(userSecret)
 	if err != nil {
+		r.Log.Error(err)
 		return reconcile.Result{}, err
 	}
 
-	err = r.client.Update(ctx, instance)
+	err = r.Client.Update(ctx, instance)
 	return reconcile.Result{}, err
```
tsatam, line 117: We should validate if err is nil or not here, and:
- if err is nil, call the ClearConditions function to reset the controller back to Available: True, Progressing: False, Degraded: False
- if err is not nil, call SetDegraded to set the degraded status.

pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
```diff
@@ -285,11 +286,11 @@ func TestPullSecretReconciler(t *testing.T) {
 			ctx := context.Background()
 
 			clientFake := ctrlfake.NewClientBuilder().WithObjects(tt.instance).WithObjects(tt.secrets...).Build()
```
tsatam, line 288: We should also add validation in these tests to ensure that the conditions on the controller are what we expect (e.g. Degraded:True when an error occurs). You can see an example of this in the dnsmasq machine config controller tests: https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/dnsmasq/machineconfig_controller_test.go#L36

tsatam (dismissed):
Few fixups and additions to our test cases, otherwise LGTM!

pkg/operator/controllers/pullsecret/pullsecret_controller.go
lbossis: This comment has been addressed

pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
lbossis: This comment has been addressed

kimorris27: I think some work is still needed here. Even though we now validate that the conditions are what we expect, there are no test cases covering the cases where the controller becomes degraded.

So, for example, we could add a test case where the reconciler fails to update the ARO Cluster resource toward the end of the `Reconcile` function, and this test case could then verify that the degraded condition is set properly.

I looked into ways to get that `Update` function to return an error, and I happened to come across an article written by a Red Hatter that looks like it'll help: https://www.bater.io/posts/2020/06/throwing-errors-with-controller-runtime-fake/

kimorris27 (changes requested):
The changes to the reconciler LGTM, but I agree with Tanmay about the additional case where we can set the controller's `Degraded` status. I also think we could use some more unit test coverage and went into more detail in an existing comment thread.

pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
```diff
@@ -333,10 +333,16 @@ func TestPullSecretReconciler(t *testing.T) {
...
+			is_valid_key := false
+			key := types.NamespacedName{Namespace: "openshift-config", Name: "invalid-pull-secret"}
+			pullSecretReconcilerTestHelper(t, r, ctx, s, key, is_valid_key)
+			key = types.NamespacedName{Namespace: "invalid-openshift-config", Name: "pull-secret"}
+			pullSecretReconcilerTestHelper(t, r, ctx, s, key, is_valid_key)
+			key = types.NamespacedName{Namespace: "invalid-openshift-config", Name: "invalid-pull-secret"}
+			pullSecretReconcilerTestHelper(t, r, ctx, s, key, is_valid_key)
+			is_valid_key = true
+			key = types.NamespacedName{Namespace: "openshift-config", Name: "pull-secret"}
+			pullSecretReconcilerTestHelper(t, r, ctx, s, key, is_valid_key)
```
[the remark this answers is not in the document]
lbossis: This code has been removed. A negative test case will be implemented in another branch for another Jira ARO-RP 13079

kimorris27 (approved):

cadenmarchese (approved):

PR #3948, opened 2024-11-08
Realign CredentialsRequest to match current ARO operator role definition

fahlmant:

Partially fixes [ARO-12070](https://issues.redhat.com/browse/ARO-12070)

Resets ARO Operator CredentialsRequest to the base role definition we have now for MiWi. Prepares it for any future permissions we need to add.

### Test plan for issue:

MiWi clusters are already using this definition

n/a

MiWi clusters are already using this definition

### DIFF
--- a/pkg/operator/deploy/staticresources/credentialsrequest.yaml
+++ b/pkg/operator/deploy/staticresources/credentialsrequest.yaml
@@ -15,8 +15,14 @@ spec:
   providerSpec:
     apiVersion: cloudcredential.openshift.io/v1
     kind: AzureProviderSpec
-    roleBindings:
-    - role: Contributor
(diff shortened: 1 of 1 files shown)


bennerv (approved):
lgtm.  

Note that this might get out of sync over time because CVO doesn't own it and won't update it on minor version rollout.

cadenmarchese (approved):
thanks!

fahlmant:
@bennerv True, since this is just a record of the definition, it could get out of sync. My proposal is to match what ROSA does: any changes must be committed to this repo via a PR before the Azure role definition gets updated.

fahlmant:
Closing in favor of #3953

PR #3950, opened 2024-11-11
MIWI API converter fixes

kimorris27:

No Jira

Fixes two issues with the API converters:

1. Ensures all cluster MSI / `Identity` fields are included in conversions
2. Ensures that all platform workload identities are included when converting from internal -> admin

### Test plan for issue:

I tested both the admin and preview APIs with my local dev RP. I'll also be testing in canary.

N/A

Will be tested in canary.

### DIFF
--- a/pkg/api/admin/openshiftcluster_convert.go
+++ b/pkg/api/admin/openshiftcluster_convert.go
@@ -198,9 +198,9 @@ func (c openShiftClusterConverter) ToExternal(oc *api.OpenShiftCluster) interfac
 
 		for name, pwi := range oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
 			out.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities[name] = PlatformWorkloadIdentity{
-				ResourceID:  pwi.ResourceID,
-				ClientID:    pwi.ClientID,
-				PrincipalID: pwi.PrincipalID,
+				ResourceID: pwi.ResourceID,
+				ClientID:   pwi.ClientID,
+				ObjectID:   pwi.ObjectID,
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


cadenmarchese (approved):

pkg/api/admin/openshiftcluster_convert.go
```diff
@@ -299,13 +301,16 @@ func (c openShiftClusterConverter) ToInternal(_oc interface{}, out *api.OpenShif
...
 		out.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities = make(map[string]api.PlatformWorkloadIdentity, len(oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities))
 
-		for k := range oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
-			if entry, ok := out.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities[k]; ok {
-				entry.ClientID = oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities[k].ClientID
-				entry.ObjectID = oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities[k].ObjectID
-				entry.ResourceID = oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities[k].ResourceID
-
-				out.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities[k] = entry
+		for name, pwi := range oc.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities {
```
rajdeepc2792, line 304: Why are we adding clientID/ObjectID when it is not present in the out and otherwise only ResourceID? Why can't we add ClientID and ObjectID for everything?

kimorris27: Sorry, I went ahead and merged to get this to canary ASAP. I copied what Tanmay had already done in the preview API converter. My understanding is that this prevents the client ID and object ID from being overwritten during a PUT/PATCH if they're already present in an existing cluster doc.

PR #3951, opened 2024-11-11
Don't scan generated code for CodeQL

hawkowl:
We shouldn't scan this, because it's useless and doesn't give us useful information + takes longer.


hawkowl:
@tsatam Because it does a build and checks the compiled binary, that all works. The go config in there is basically because it needs to exist for the matrix, it doesn't effectively do anything.

yjst2012 (approved):
looks good to me, just not sure if we want to ignore the vendor folder as well as Tanmay suggested?

hawkowl:
@yjst2012 You can't ignore dirs for compiled code, so it wouldn't do anything. The existing ones there don't really do much either :)

PR #3959, opened 2024-11-15
Fail to validate MIWI cluster create/updates if platform workload identities contain unexpected federated credentials

tsatam:

Fixes [ARO-11854](https://issues.redhat.com/browse/ARO-11854)

- [X] Fails cluster dynamic validation if any federated credentials on any platform workload identities are unexpected

The determination of an "unexpected" federated credential is determined by name - if a FIC has a name that wouldn't match the deterministic name generated by the ARO service, we will consider it unexpected and error out. 

### Test plan for issue:

- [X] Unit tests were added and/or updated to handle this behavior
- [ ] Manual attempts to create and update a MIWI cluster with extra FIC fails as expected

Not yet. We will eventually need public documentation to cover our requirements for identities used in MIWI clusters. 

Feature is not yet in production.


pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -29,6 +41,45 @@ func (dv *dynamic) ValidatePlatformWorkloadIdentityProfile(ctx context.Context,
...
+					)
+				}
+
+				if len(federatedCredential.Properties.Audiences) != 1 ||
+					*federatedCredential.Properties.Audiences[0] != expectedAudience ||
+					*federatedCredential.Properties.Issuer != string(*oc.Properties.ClusterProfile.OIDCIssuer) {
+					return api.NewCloudError(
+						http.StatusBadRequest,
+						api.CloudErrorCodePlatformWorkloadIdentityContainsInvalidFederatedCredential,
+						fmt.Sprintf("properties.platformWorkloadIdentityProfile.platformWorkloadIdentities.%s.resourceId", k),
```
rajdeepc2792, line 75: Can we also add role/issuer to the error message here to be clear of all the details used to validate?

pkg/validate/dynamic/platformworkloadidentityprofile_test.go
```diff
@@ -254,6 +258,10 @@ func TestValidatePlatformWorkloadIdentityProfile(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	msiResourceID := resourceGroupID + "/providers/Microsoft.ManagedIdentity/userAssignedIdentities/miwi-msi-resource"
+	clusterResourceId, _ := azure.ParseResourceID(clusterID)
+	platformIdentity1ResourceId, _ := azure.ParseResourceID(platformIdentity1)
+	expectedPlatformIdentity1FederatedCredName := platformworkloadidentity.GetPlatformWorkloadIdentityFederatedCredName(clusterResourceId, platformIdentity1ResourceId, platformIdentity1SAName)
```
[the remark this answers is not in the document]
tsatam: I still want to gather opinions on whether we should search credentials by name vs iss/aud before we accept and merge this PR, if we decide to go back to using names we can just cleanly revert the commit(s) that change us to the other behavior. 

I personally prefer having the name generation function split out, it's a fairly complex piece of logic but has no dependencies on clients or other services, so it's a good candidate for more robust unit tests (though I'm guilty of not writing those on this PR yet, I'd rather add those tests in than move the function back). We've also needed a good amount of discussion to come up with the current naming scheme, so I'd prefer to keep that naming scheme in an easy-to-find location in the code.

pkg/validate/dynamic/platformworkloadidentityprofile.go
tsatam: `k` is the role (OperatorName) here. We can include the issuer on the error message, and I think it's worth having these error messages reviewed in detail. I'll bring this up in our next sync meeting.

pkg/validate/dynamic/platformworkloadidentityprofile.go
```diff
@@ -29,6 +41,45 @@ func (dv *dynamic) ValidatePlatformWorkloadIdentityProfile(ctx context.Context,
...
+						"Unexpected federated credential '%s' found on platform workload identity '%s' used for role '%s'. Please ensure only federated credentials provisioned by the ARO service for this cluster are present.",
+						*federatedCredential.Name,
+						pwi.ResourceID,
+						k,
+					)
+				}
+			}
```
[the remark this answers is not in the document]
tsatam: Good call-out, I'll add that.

fahlmant:
Other than outstanding comments, lgtm

pkg/validate/dynamic/platformworkloadidentityprofile.go
tsatam: Actually, given the issues with nil pointer dereferences above, I'm in favor of not including the issuer. It could be nil and conditionally handling the nil pointer dereference would introduce too much complexity into the error generation process. We return the full resource ID of the identity and the name of the offending FIC here, which I think should be sufficient for customers to identify what's wrong.

tsatam:
We might want to include the same determination of "our" federated identity credentials into the cluster deletion process, in order to delete all FICs the ARO service may have put down for that cluster. Merging this PR as-is would introduce a potential issue with our existing delete implementation not cleaning up FICs on an identity.

pkg/validate/dynamic/platformworkloadidentityprofile_test.go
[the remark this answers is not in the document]
rajdeepc2792: IMO, iss/aud is a better approach than blocking the operations completely due to the change in the naming convention.

pkg/validate/dynamic/platformworkloadidentityprofile.go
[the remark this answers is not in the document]
tsatam: handled in latest changes

pkg/cluster/delete.go
```diff
@@ -405,30 +405,41 @@ func (m *manager) deleteFederatedCredentials(ctx context.Context) error {
...
+					identityResourceId.ResourceName,
+					*federatedCredential.Name,
+					&armmsi.FederatedIdentityCredentialsClientDeleteOptions{},
+				)
+				if err != nil {
+					if azuresdkerrors.IsNotFoundError(err) {
+						m.log.Errorf("federated identity credentials not found for %s: %v", identity.ResourceID, err.Error())
+						continue
+					} else {
+						m.log.Infof("failed to delete federated identity credentials for %s: %v", identity.ResourceID, err.Error())
```
rajdeepc2792, line 441: Not sure why we are consoling error when the identity is not found and other errors as info, I feel this should be reversed. Not found would be normal, otherwise it should be consoled out as error??

pkg/cluster/delete_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"testing"
 
+	sdkmsi "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/msi/armmsi"
```
rajdeepc2792, line 13: NIT: can't we use `armmsi` directly?

pkg/cluster/delete.go
tsatam: I agree, this ought to be flipped. I can include that change in this PR.

pkg/validate/dynamic/platformworkloadidentityprofile_test.go
```diff
@@ -318,6 +327,7 @@ func TestValidatePlatformWorkloadIdentityProfile(t *testing.T) {
 			name:                  "Success - Validation for the OC doc for PlatformWorkloadIdentityProfile",
```
rajdeepc2792, line 327: Can I please ask to add a test case for erroring out on failed `federatedIdentityCredentials.List`?

rajdeepc2792 (commented):
Changes LGTM, added comments on possible unit test cases addition. Thanks!!

rajdeepc2792 (approved):

pkg/cluster/delete.go
```diff
@@ -405,30 +405,41 @@ func (m *manager) deleteFederatedCredentials(ctx context.Context) error {
...
-					m.log.Infof("failed to delete federated identity credentials for %s: %v", identity.ResourceID, err.Error())
+		for _, federatedCredential := range federatedCredentials {
+			switch {
+			case federatedCredential == nil,
+				federatedCredential.Name == nil,
+				federatedCredential.Properties == nil,
+				len(federatedCredential.Properties.Audiences) != 1,
+				*federatedCredential.Properties.Audiences[0] != "openshift",
+				federatedCredential.Properties.Issuer == nil,
+				*federatedCredential.Properties.Issuer != string(*m.doc.OpenShiftCluster.Properties.ClusterProfile.OIDCIssuer):
```
rajdeepc2792, line 426: Faced an issue here while testing this part, if the cluster install fails before the OIDC Issuer is generated, and there already exists some federated credentials for an identity, the deletion flow is facing panic.

rajdeepc2792 (changes requested):
Taking back the approval as found an issue around OIDCIssuer handling.

pkg/cluster/delete.go
tsatam: Nice catch. 

I think in that scenario, we can assume we never generated any FICs for the cluster and just bail out here, right?

tsatam: Fixed in latest commit.

kimorris27 (approved):

pkg/cluster/delete.go
```diff
@@ -405,30 +405,41 @@ func (m *manager) deleteFederatedCredentials(ctx context.Context) error {
...
-		if !exists {
-			continue
+		federatedCredentials, err := m.clusterMsiFederatedIdentityCredentials.List(
+			ctx,
+			identityResourceId.ResourceGroup,
+			identityResourceId.ResourceName,
+			&armmsi.FederatedIdentityCredentialsClientListOptions{},
+		)
+		if err != nil {
+			return err
```
rajdeepc2792, line 415: Another potential bug, when a Cx during creation provides a platform identity with a missing role assignment on MSI to read the platform identity fails during populating the clientid/objectid.
But during deletion, it fails here too, this causes a blocker to delete.
Can we instead just console the error and allow the delete flow to continue??

> m.log.Errorf("failed to get federated identity credentials for %s: %v", identity.ResourceID, err.Error())
> continue

tsatam: Do you have specifics on what shape the error would take? I think the solution here would be to bail out and stop trying to delete any federated credentials, if we run into an authorization error when attempting to either list or delete credentials, but I want to make sure this manifests as an authorization error specifically before implementing that.

rajdeepc2792: Since you are bailing out here based on if OIDC populated or not, it won't be blocked for the example I provided above, but can happen in case the MSI's permission were revoked after cluster deletion or a new platform identity is added with MSI permission missing and then deletion is ran.

Below is an example of the error:

ERRO[2024-11-21T10:34:20-05:00]pkg/cluster/delete.go:415 cluster.(*manager).deleteFederatedCredentials() failed to get federated identity credentials for /subscriptions/<REDACTED>/resourceGroups/rajdchau-miwi/providers/Microsoft.ManagedIdentity/userAssignedIdentities/aro-NetworkOperator: GET https://management.azure.com/subscriptions/<REDACTED>/resourceGroups/rajdchau-miwi/providers/Microsoft.ManagedIdentity/userAssignedIdentities/aro-NetworkOperator/federatedIdentityCredentials
--------------------------------------------------------------------------------
RESPONSE 403: 403 Forbidden
ERROR CODE: AuthorizationFailed
--------------------------------------------------------------------------------
{
  "error": {
    "code": "AuthorizationFailed",
    "message": "The client '<REDACTED>' with object id '<REDACTED>' does not have authorization to perform action 'Microsoft.ManagedIdentity/userAssignedIdentities/federatedIdentityCredentials/read' over scope '/subscriptions/<REDACTED>/resourceGroups/rajdchau-miwi/providers/Microsoft.ManagedIdentity/userAssignedIdentities/aro-NetworkOperator' or the scope is invalid. If access was recently granted, please refresh your credentials."
  }
}
--------------------------------------------------------------------------------

tsatam: Thanks. I'll implement a non-failing short-circuit here if there is an auth error.

tsatam: or actually, I guess to stay in line with error handling during deletion, we should log and continue regardless of what the error is.

tsatam: Handled in latest commit.

rajdeepc2792 (approved):

kimorris27 (approved):

cadenmarchese (approved):

PR #3906, opened 2024-10-17
Jaitaiwan/aro 11234

jaitaiwan:
### Which issue this PR addresses:

Fixes

PR #3964, opened 2024-11-19
Fix AKS connectivity 

kimorris27:

No Jira

This PR reintroduces the `serverFqdn` parameter used to get AKS admin credentials that we removed in #3906. I noticed it needed to be added back in when I found that my local dev RP was trying to connect to AKS through a private endpoint.

### Test plan for issue:

Ran my local dev RP and confirmed that it can connect to AKS again with this fix.

N/A

Canary E2E during next release.


tsatam (approved):
LGTM, merge on green

shubhadapaithankar (approved):

PR #3967, opened 2024-11-20
Update NetworkType in cluster doc if cluster has migrated from SDN to OVN

fahlmant:
https://issues.redhat.com/browse/ARO-12232

Clusters that are using SDN will need to migrate to OVN before upgrading to OpenShift 4.17.z
This change checks any cluster that is still marked as using SDN in its cluster doc, and updates it if the on-cluster networkType has been updated to use OVN

### Test plan for issue:

Run local RP, install SDN Cluster, run migration to OVN, check cluster doc, run Update/AdminUpdate, check cluster doc

Maybe?

By testing local RP


kimorris27 (commented):
One of the lint/validation checks is unhappy about some whitespace, but LGTM otherwise!

Question for my understanding: what is this property in the cluster doc even used for? I didn't find many references to it in the rest of the codebase.

fahlmant:
@kimorris27 The only thing I know it's used for is install config. However, given that it's part of the clusterdoc for the life of the cluster, it seemed best to make sure it gets updated to be accurate in case we ever do need to use it down the line.

fahlmant:

kimorris27 (approved):
This is good to merge without E2E IMO.

PR #3971, opened 2024-11-21
Skip service principal enrichment for MIWI clusters

kimorris27:

https://issues.redhat.com/browse/ARO-11680

This PR modifies cluster enrichment such that the service principal enricher is skipped for MIWI clusters. I didn't add any additional MIWI-specific enrichment because it's not necessary; unlike in the service principal case, MIWI identities/federated credentials can't be rotated without going through the RP.

### Test plan for issue:

1. Ran `az aro list` against local dev RP and observed an error saying that the SP enricher failed
2. Made code changes and restarted dev RP
3. Ran `az aro list` again and observed the error was no longer present, meaning SP enrichment had been skipped

N/A

E2E

### DIFF
--- a/pkg/util/clusterdata/clusterdata.go
+++ b/pkg/util/clusterdata/clusterdata.go
@@ -98,10 +98,17 @@ func (p ParallelEnricher) enrichOne(ctx context.Context, log *logrus.Entry, oc *
 		return
 	}
 
-	errors := make(chan error, len(p.enrichers))
+	numEnrichers := len(p.enrichers)
+
+	// We skip the service principal enricher for workload identity clusters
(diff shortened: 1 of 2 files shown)


tsatam (approved):
Implementation looks good to me as-is, but I'm curious if we want to consider a simpler approach where we just make the service principal enricher exit early if the cluster is a MIWI cluster, and not have to worry about maintaining the list and number of enrichers differently for CSP vs MIWI clusters: https://github.com/Azure/ARO-RP/blob/master/pkg/util/clusterdata/service_principal.go

rajdeepc2792 (approved):

cadenmarchese (approved):

PR #3974, opened 2024-11-22
Dynamic image tagging using PR-specific variables in Azure DevOps pip…

shubhadapaithankar:

This PR addresses the need for improved image management in the pipeline by introducing PR-specific tagging for RP images. Currently, it can be challenging to trace images back to specific pull requests, which complicates troubleshooting and version management. This update standardizes image tags, supports artifact caching, and enables caching for dependent images, improving both efficiency and traceability.

Fixes: [9503](https://issues.redhat.com/browse/ARO-9503)

This PR adds several key enhancements to the CI pipeline:

- PR-Specific Tagging: RP images are pushed to the development ACR instance with PR-specific tags `(pr-${PR_NUMBER}-${GIT_COMMIT_SHA})`. This improves traceability by linking images directly to the pull requests that generated them.
- The pipeline improvements introduced in this PR support more efficient builds, help with version control, and allow easier tracking of images and artifacts related to specific PRs.
- The PR-specific tagging also simplifies troubleshooting, as we can directly identify which PR generated a particular image.

**Key benefits:**

- Enhanced Troubleshooting: PR-specific tags make it easier to trace images back to the relevant pull requests, simplifying debugging.
- Consistent Image Management: Standardizing tags across images and dependencies helps maintain consistency, allowing SREs to manage images effectively across environments.

### Test plan for issue:
Verify PR-specific tagging `(pr-${PR_NUMBER}-${GIT_COMMIT_SHA})` in dev ACR, confirm artifact caching reuses previous builds

N/A

N/A

### DIFF
--- a/.pipelines/ci.yml
+++ b/.pipelines/ci.yml
@@ -30,7 +30,7 @@ variables:
   - name: PR_TAG
     value: $[ coalesce(format('pr-{0}-{1}', variables['System.PullRequest.PullRequestId'], variables['System.PullRequest.SourceCommitId']), '') ]
   - name: MASTER_TAG
-    value: $[ coalesce(format('master-{0}', variables['Build.SourceVersion']), '') ]
+    value: $[ format('master-{0}', variables['Build.SourceVersion']) ]
   - name: TAG
     value: $[ coalesce(variables['PR_TAG'], variables['MASTER_TAG']) ]
   - name: ARO_IMAGE
--- a/.pipelines/ci.yml
+++ b/.pipelines/ci.yml
@@ -25,12 +25,12 @@ variables:
     value: arosvcdev.azurecr.io/vpn
   - name: LOCAL_E2E_IMAGE
     value: arosvcdev.azurecr.io/e2e
-  - name: TAG
-    value: $(Build.BuildId)
   - name: VERSION
-    value: $(Build.BuildId)
+    value: $(TAG)
+  - name: TAG
+    value: pr-$(System.PullRequest.PullRequestId)-$(System.PullRequest.SourceCommitId)
   - name: ARO_IMAGE
-    value: arosvcdev.azurecr.io/aro:$(Build.BuildId)
+    value: arosvcdev.azurecr.io/aro:$(TAG)
   - name: ARO_SELENIUM_HOSTNAME
     value: localhost
   - name: E2E_LABEL
@@ -48,7 +48,7 @@ stages:
           # Build and test the Az ARO Extension
           - script: |
               set -xe
-              DOCKER_BUILD_CI_ARGS="--load" make ci-azext-aro VERSION=$(VERSION)
+              DOCKER_BUILD_CI_ARGS="--load" make ci-azext-aro VERSION=$(TAG)
             displayName: 🛠 Build & Test Az ARO Extension
 
           # Push the image to ACR
@@ -56,6 +56,7 @@ stages:
             parameters:
               acrFQDN: 'arosvcdev.azurecr.io'
               repository: 'azext-aro'
+              tag: $(TAG)
               pushLatest: true
 
       - job: Build_And_Test_RP_And_Portal
@@ -67,7 +68,7 @@ stages:
           # Build and test RP and Portal
           - script: |
               set -xe
-              DOCKER_BUILD_CI_ARGS="--load" make ci-rp VERSION=$(VERSION)
+              DOCKER_BUILD_CI_ARGS="--load" make ci-rp VERSION=$(TAG)
             displayName: 🛠 Build & Test RP and Portal
 
           # Publish test results
@@ -91,6 +92,7 @@ stages:
             parameters:
               acrFQDN: 'arosvcdev.azurecr.io'
               repository: 'aro'
+              tag: $(TAG)
               pushLatest: true
 
       - job: Build_And_Push_E2E_Image
@@ -102,7 +104,7 @@ stages:
           # Build the E2E image
           - script: |
               set -xe
-              DOCKER_BUILD_CI_ARGS="--load" make aro-e2e VERSION=$(VERSION)
+              DOCKER_BUILD_CI_ARGS="--load" make aro-e2e VERSION=$(TAG)
             displayName: 🛠 Build the E2E image
 
           # Push the E2E image to ACR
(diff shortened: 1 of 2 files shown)


slawande2 (dismissed):

chreed-rh (approved):
LGTM

rhamitarora (approved):

kevinobriendotca (dismissed):

.pipelines/ci.yml
```diff
@@ -25,12 +25,12 @@ variables:
     value: arosvcdev.azurecr.io/vpn
   - name: LOCAL_E2E_IMAGE
     value: arosvcdev.azurecr.io/e2e
-  - name: TAG
-    value: $(Build.BuildId)
   - name: VERSION
-    value: $(Build.BuildId)
+    value: $(TAG)
+  - name: TAG
+    value: pr-$(System.PullRequest.PullRequestId)-$(System.PullRequest.SourceCommitId)
```
tsatam, line 31: This pipeline also runs against the master branch directly, do we want to have a different tag for those builds? I expect that `System.PullRequest.PullRequestId` will not be set during those builds and will either cause issues or undesired names (e.g. `pr-null-asdhfjkl` or equivalent)

tsatam (changes requested):
One outstanding issue about the naming convention for master branch builds, otherwise LGTM

.pipelines/templates/template-acr-push.yml
```diff
@@ -48,44 +50,25 @@ steps:
...
-        IMAGE_LATEST="${ACR_FQDN}/${{ parameters.repository }}:latest"
-        echo "Checking for image $IMAGE_LATEST..."
-        if [[ "$(docker images -q $IMAGE_LATEST 2> /dev/null)" == "" ]]; then
-          echo "Warning: Image $IMAGE_LATEST not found. Skipping 'latest' tag."
-          SKIP_LATEST=true
+          echo "Error: Image $IMAGE_NAME not found locally. Exiting."
+          exit 1
         else
-          echo "Image $IMAGE_LATEST found. Proceeding with 'latest' tag."
-          SKIP_LATEST=false
```
kimorris27, line 80: I don't see any problem with removing this, but why are we removing it? Just asking to understand better.

shubhadapaithankar: We removed the extra checks to simplify the pipeline:

- The earlier steps already ensure the image is built, so checking again isn’t needed.
- The `pushLatest` parameter controls whether the image is tagged and pushed as `latest`, making the extra logic unnecessary.
- This makes the pipeline cleaner without affecting reliability or functionality.

tsatam (approved):

chreed-rh (approved):
LGTM

.pipelines/ci.yml
```diff
@@ -25,12 +25,12 @@ variables:
     value: arosvcdev.azurecr.io/vpn
   - name: LOCAL_E2E_IMAGE
     value: arosvcdev.azurecr.io/e2e
-  - name: TAG
-    value: $(Build.BuildId)
   - name: VERSION
-    value: $(Build.BuildId)
+    value: $(TAG)
+  - name: TAG
+    value: $[ coalesce('pr-$(System.PullRequest.PullRequestId)-$(System.PullRequest.SourceCommitId)', 'master-$(Build.SourceVersion)') ]
```
kimorris27, line 31: I don't think this usage of `coalesce` is going to work how you intended. Check out these docs: 

- https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#coalesce
- https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#runtime-expression-syntax

Description of `coalesce`:

> Evaluates the parameters in order (left to right), and returns the first value that doesn't equal null or empty-string.

The second doc discusses different ways to interpolate variable values into the pipeline file. Based on what's being used here, I think on master we would end up tagging the image with the literal string `pr-$(System.PullRequest.PullRequestId)-$(System.PullRequest.SourceCommitId)`.

With that said, it might take some experimenting to figure out exactly what will work.

kimorris27 (changes requested):

.pipelines/ci.yml
shubhadapaithankar: Thanks for pointing this out! I've updated the pipeline to ensure proper runtime evaluation of the TAG variable. The logic now uses separate variables for PR and master builds `(PR_TAG and MASTER_TAG) `and combines them with `coalesce` to dynamically construct the TAG. This resolves the issue of invalid literal strings and ensures consistent naming `(pr-[PR_ID]-[COMMIT_HASH] `for PRs and `master-[COMMIT_HASH]` for master builds).

.pipelines/ci.yml
```diff
@@ -25,12 +25,16 @@ variables:
...
-    value: $(Build.BuildId)
   - name: VERSION
-    value: $(Build.BuildId)
+    value: $(TAG)
+  - name: PR_TAG
+    value: $[ format('pr-{0}-{1}', variables['System.PullRequest.PullRequestId'], variables['System.PullRequest.SourceCommitId']) ]
+  - name: MASTER_TAG
+    value: $[ format('master-{0}', variables['Build.SourceVersion']) ]
+  - name: TAG
+    value: $[ coalesce(variables['PR_TAG'], variables['MASTER_TAG']) ]
```
kimorris27, line 35: I think this still won't work. Let's assume we're on master, meaning the `System.PullRequest` attributes are null. `PR_TAG` will be set to `"pr--"`, and then `TAG` will try:

```
coalesce("pr--", "master-whatever-Build.SourceVersion-is")
```

which will return `"pr--"`.

I think one option would be to conditionally assign a value to `PR_TAG` such that it only gets a non-empty value when we're on a pull request branch. That way `coalesce(variables['PR_TAG'], variables['MASTER_TAG'])` will work how we want it to. It looks like there are some things worth trying here: https://stackoverflow.com/questions/57532138/can-conditional-variable-assignment-be-done-in-azure-pipelines

kimorris27 (changes requested):

.pipelines/ci.yml
shubhadapaithankar: I have added the code that  TAG generation to avoid "pr--" by ensuring PR_TAG is assigned only for PR builds. Implements conditional logic to validate System.PullRequest attributes, ensuring coalesce correctly falls back to MASTER_TAG for master builds, addressing concerns about invalid tags.

kimorris27 (approved):

kimorris27:
Not 100% sure that the tagging logic will work correctly for the master branch, but it looks good for the PR branch image tags. We have to merge it to try it out on master, and it won't cause any harm if it doesn't work - we can just debug with small follow-up PRs.

PR #3980, opened 2024-11-28
[ARO-12792] Fix CIF - ServerError - error status code received from the API

bitoku:

Fixes https://issues.redhat.com/browse/ARO-12795

This PR adds error handling from MSGraph, so that it can return meaningful error message to users.
When the application is deleted just before the cluster creation, it gets `error status code received from the API`.
By adding the error handling, it returns 400 error with the error message from MSGraph.

Reproducible code is in https://github.com/Azure/ARO-RP/tree/temp/aro-12792 . Running the hack script will reproduce the issue.

### Test plan for issue:

Running the hack script in https://github.com/Azure/ARO-RP/tree/temp/aro-12792.
unit test

N/A

Running the hack script in https://github.com/Azure/ARO-RP/tree/temp/aro-12792.

### DIFF
--- a/pkg/util/steps/runner.go
+++ b/pkg/util/steps/runner.go
@@ -5,6 +5,7 @@ package steps
 
 import (
 	"context"
+	"errors"
 	"net/http"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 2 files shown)


jaitaiwan (approved):
LGTM

kimorris27 (approved):
LGTM, and I was able to locally reproduce the issue and observe that your fix worked. Thanks for making it easy to test by providing a branch for it.

PR #3981, opened 2024-12-02
[ARO-12518] Map other errors' http status codes to correct result type.

pepedocs:
Fixes https://issues.redhat.com/browse/ARO-12518

**What**
Map other known backend errors returned by azure to their correct result type when logging them. 

**Why**
Currently, these known errors are mapped to "InternalServerError" regardless of their nature (e.g. 5xx, 4xx) and logged as an InternalServerError. This breaks our CIF SLO reporting as it depends on parsing these logs.

**Example UserErrors that are mapped to InternalServerError**
```
network.SubnetsClient#CreateOrUpdate: Failure sending request: StatusCode=403 – Original Error: Code="LinkedAuthorizationFailed" Message= (backend)

ClientSecretCredential authentication failed. FromClientSecret(): http call(https://login.microsoftonline.com/xxx/oauth2/v2.0/token)(POST) error: reply status code was 400:
```

### Test plan for issue:

**Unit Test**
- Ensure the following
  - ResponseError and DetailedError types' StatusCodes are utilized when mapping to error result type

**Manual Test**
 - Mocked  `network.SubnetsClient#CreateOrUpdate: `  to use an invalid subnet so it will return a ResponseError and checked if it was logged as UserError.
 
**Integration Tests**
- Please suggest a method if you think it's necessary.

- None

- Tests were satisfied

### DIFF
--- a/pkg/backend/openshiftcluster.go
+++ b/pkg/backend/openshiftcluster.go
@@ -5,12 +5,15 @@ package backend
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
 	"sync/atomic"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/sirupsen/logrus"
@@ -351,11 +354,25 @@ func (ocb *openShiftClusterBackend) asyncOperationResultLog(log *logrus.Entry, i
 			"properties.servicePrincipalProfile", "The Azure Red Hat Openshift resource provider service principal has been removed from your tenant. To restore, please unregister and then re-register the Azure Red Hat OpenShift resource provider.")
 	}
 
-	err, ok := backendErr.(*api.CloudError)
-	if ok {
-		resultType := utillog.MapStatusCodeToResultType(err.StatusCode)
-		log = log.WithField("resultType", resultType)
+	var statusCode int
+	var cloudErr *api.CloudError
+	var detailedErr = autorest.DetailedError{}
+	var respErr *azcore.ResponseError
 
+	// Get the HTTP status code and map it to a result type
+	if errors.As(backendErr, &cloudErr) {
+		statusCode = cloudErr.StatusCode
+	} else if errors.As(backendErr, &detailedErr) {
+		if detailedErr.Response != nil {
+			statusCode = detailedErr.Response.StatusCode
+		}
+	} else if errors.As(backendErr, &respErr) {
+		statusCode = respErr.StatusCode
+	}
+
+	if statusCode > 0 {
+		resultType := utillog.MapStatusCodeToResultType(statusCode)
+		log = log.WithField("resultType", resultType)
 		if resultType == utillog.SuccessResultType {
 			log.Info("long running operation succeeded")
 			return
--- a/pkg/backend/openshiftcluster_test.go
+++ b/pkg/backend/openshiftcluster_test.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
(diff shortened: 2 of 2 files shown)


pepedocs:
@microsoft-github-policy-service agree

@microsoft-github-policy-service agree company="Microsoft"

bitoku:
What if our RP code has a bug and we get a 4XX azure error because of that bug? Will it be counted as User Error?

rhamitarora:
Consider adding logging for cases where **statusCode** <= 0 to debug any **unmapped** or **unexpected** errors.

pepedocs:
> What if our RP code has a bug and we get a 4XX azure error because of that bug? Will it be counted as User Error?

Yes. I understand the concern here, but this is a very unlikely case (bug). I do not think we should optimize on low probability cases.

Again I do understand the concern here, and the risk should be mitigated by adding more debug information. I believe this information should not be in this part of this code, rather in the caller code to keep it simple and separate the concerns. 

>I think all azure errors should be handled one by one meticulously, otherwise we will be less aware of our bugs.

I don't think this is even possible. For example, the "LinkAuthorizationFailed" error is a 4xx error, which implies a client error. But this could be caused by, wrong user params/inputs/data OR a bug in RP. The former is easy to detect, the latter is almost impossible, not to mention there are many of these azure errors. Unless if there's a way that I do not know (yet), meticulously checking for these errors not practical because a bug can happen anywhere. The good news is this that it's fair to say that this bug is very unlikely to happen.

Note that right now, RP code is logging all of these non-CloudError UserErrors (true positive) as InternalServerError, and this is a bug that is a very likely one - which is getting fixed here.

pepedocs:
> Consider adding logging for cases where **statusCode** <= 0 to debug any **unmapped** or **unexpected** errors.
@rhamitarora 
I didn't want to introduce new logic to keep it simple. The previous code do not care about this. And this task is simply for enhancing the logging for SLO reporting purposes only. Maybe on another task we can handle <0 errors?

bitoku:
For example "AuthorizationFailed" error, we retry when the error occurs, and after a few retry failures, it returns 400 user error.
If we return an error even though the retryable error occurs only once, it's our bug and should be counted as Server Error.
I'm concerned this kind of error might be overseen by this change.

But I understand what you mean. It's impossible to confirm all errors are on our side or user side.

By the way, this is just a change for metrics. Are you going to change the error handling in `reply` as well?
https://github.com/Azure/ARO-RP/blob/78feafb7f3980b7a807e6a81d0a84c453246f5ee/pkg/frontend/frontend.go#L441-L454

pepedocs:
> If we return an error even though the retryable error occurs only once, it's our bug and should be counted as Server Error.

I totally agree with this. I think though that this is a bigger issue in the code IMHO. At the least, I don't see any consistent and dedicated error-handling pattern anywhere.

>By the way, this is just a change for metrics. Are you going to change the error handling in reply as well?

Thanks for asking. The Jira only asked for updating the logs and I have not looked at the reply - if you mean the message returned to the user. I will have a look and see what I can do here.

bitoku (dismissed):

rhamitarora (dismissed):

pkg/backend/openshiftcluster.go
```diff
@@ -351,11 +353,22 @@ func (ocb *openShiftClusterBackend) asyncOperationResultLog(log *logrus.Entry, i
...
 	}
 
-	err, ok := backendErr.(*api.CloudError)
-	if ok {
-		resultType := utillog.MapStatusCodeToResultType(err.StatusCode)
-		log = log.WithField("resultType", resultType)
+	var statusCode int
 
+	// Get the HTTP status code and map it to a result type
+	if err, ok := backendErr.(*api.CloudError); ok {
```
bitoku, line 359: Sorry I oversaw these error handlings.
If there's no reason to use type assertion, please use `errors.Is` or `errors.As` instead?
https://issues.redhat.com/browse/ARO-7922

pepedocs: @bitoku No worries, there isn't :smile: , it's just old ways. Will update this.

bitoku (approved):

PR #3982, opened 2024-12-02
Add FPSP to MIWI deny assignment ExcludePrincipals

kimorris27:

https://issues.redhat.com/browse/ARO-13246

This PR adds the FPSP to the MIWI deny assignment's `ExcludePrincipals`. See the Jira for more context on why we're doing this.

### Test plan for issue:

Augmented unit tests; will soon test in canary

N/A

MIWI feature testing


cadenmarchese (approved):
Since canary testing is time sensitive, I think we can proceed with merging this without e2e as it will be immediately tested in canary and our CI e2e doesn't have coverage for deny assignment changes since it creates a dev cluster.

fahlmant (approved):

PR #3985, opened 2024-12-02
Add disable-infra annotation to hive clusterdeployments

tsatam:

Fixes [ARO-11607](https://issues.redhat.com/browse/ARO-11607)

Adds the `hive.openshift.io/infra-disabled` annotation to our Hive ClusterDeployments (for both new installs as well as adopted clusters), set to `true`. This is required in order to disable the hibernation controller, which does not currently work for ARO MIWI clusters. 

> [!NOTE]
> This PR currently disables the hibernation controller for all cluster installations (CSP and MIWI) - we do not use this functionality in ARO's implementation of Hive today. 

See corresponding change in Hive to add this annotation for more details: https://github.com/openshift/hive/pull/2501 

### Test plan for issue:

- [x] E2E test was added to ensure the annotation is set
- [x] E2E cluster creation ensures the cluster is installed successfully with this annotation set

No

E2E tests outlined above

### DIFF
--- a/pkg/hive/install.go
+++ b/pkg/hive/install.go
@@ -203,9 +203,9 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 			Name:      ClusterDeploymentName,
 			Namespace: doc.OpenShiftCluster.Properties.HiveProfile.Namespace,
 			Labels: map[string]string{
-				"hive.openshift.io/cluster-platform": "azure",
-				"hive.openshift.io/cluster-region":   doc.OpenShiftCluster.Location,
-				createdByHiveLabelKey:                "true",
+				hiveClusterPlatformLabel: "azure",
+				hiveClusterRegionLabel:   doc.OpenShiftCluster.Location,
+				createdByHiveLabelKey:    "true",
 			},
 			Annotations: map[string]string{
 				// https://github.com/openshift/hive/pull/2157
@@ -217,7 +217,7 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 
 				// https://github.com/openshift/hive/pull/2501
 				// Disable hibernation controller
-				"hive.openshift.io/infra-disabled": "true",
+				hiveInfraDisabledAnnotation: "true",
 			},
 		},
 		Spec: hivev1.ClusterDeploymentSpec{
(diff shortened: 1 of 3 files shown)


pkg/hive/resources.go
```diff
@@ -193,6 +193,11 @@ func adoptedClusterDeployment(namespace, clusterName, clusterID, infraID, locati
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      ClusterDeploymentName,
 			Namespace: namespace,
+			Annotations: map[string]string{
+				// https://github.com/openshift/hive/pull/2501
+				// Disable hibernation controller
```
cadenmarchese, line 198: Tiniest nit, but it's not clear looking at the code or the linked PR why we're disabling it.
```suggestion
				// Disable hibernation controller as it is not used as part of ARO's Hive implementation 
```

cadenmarchese (approved):

pkg/hive/resources.go
```diff
@@ -193,6 +193,11 @@ func adoptedClusterDeployment(namespace, clusterName, clusterID, infraID, locati
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      ClusterDeploymentName,
 			Namespace: namespace,
+			Annotations: map[string]string{
+				// https://github.com/openshift/hive/pull/2501
+				// Disable hibernation controller as it is not used as part of ARO's Hive implementation 
+				"hive.openshift.io/infra-disabled": "true",
```
kimorris27, line 199: Super nit: we could make a `const` for this in `pkg/hive` since it's used in two places.

kimorris27 (commented):
LGTM. I only left my nit because there's a formatting issue that needs to be fixed anyway.

pkg/hive/resources.go
tsatam: addressed in latest commit

pkg/hive/install.go
```diff
@@ -214,6 +214,10 @@ func (c *clusterManager) clusterDeploymentForInstall(doc *api.OpenShiftClusterDo
 
 				// TODO: remove until we use a version of hive at minimal install
 				"hive.openshift.io/cli-domain-from-installer-image": "true",
+
+				// https://github.com/openshift/hive/pull/2501
+				// Disable hibernation controller
+				"hive.openshift.io/infra-disabled": "true",
```
kimorris27, line 220: One more super nit since you humored me on the first one: you could use `hiveInfraDisabledAnnotation` here too. 🙇🏻

tsatam: that's what I get for making this change live during PR review session and not fully paying attention. I'll fix this.

cadenmarchese (approved):

kimorris27 (approved):

PR #3986, opened 2024-12-02
Enable Hive E2E tests in CI

tsatam:

Fixes no Jira

With the merge of #3978, the Hive tests now function correctly in CI (tested manually, see https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=109487069&view=logs&j=db690490-d817-5143-da35-29bc37ac10d9&t=7ab7f38e-816f-5a76-939c-0ca727e8db32 for an example).

This PR re-enables these tests. Note that they are still skipped when not in the development context, e.g. during release E2E.

In this PR, we also update the "adoptedClusterDeployment" definition, which happens to also be put down in the Hive cluster when an ARO cluster is Updated or AdminUpdated, to set  various additional properties:
- Hive labels (`cluster-platform` and `cluster-region`)
- `Spec.Platform.Azure.BaseDomainResourceGroupName`
- `Spec.ControlPlaneConfig.APIURLOverride`

These properties are populated on the ClusterDeployment for clusters installed via Hive, but the adopted CD was setting these values back to the empty string, causing the Hive E2E tests to sometimes fail (if the Hive test happened to run after an E2E test that performed an Update or AdminUpdate). 

### Test plan for issue:

- [X] E2E on this PR passes and the Hive tests are included

No

Non-production change

### DIFF
--- a/pkg/hive/resources.go
+++ b/pkg/hive/resources.go
@@ -6,6 +6,7 @@ package hive
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	hivev1azure "github.com/openshift/hive/apis/hive/v1/azure"
@@ -69,6 +70,7 @@ func (hr *clusterManager) resources(sub *api.SubscriptionDocument, doc *api.Open
 		doc.OpenShiftCluster.Properties.ClusterProfile.ResourceGroupID,
 		doc.OpenShiftCluster.Location,
 		doc.OpenShiftCluster.Properties.NetworkProfile.APIServerPrivateEndpointIP,
+		doc.OpenShiftCluster.Properties.ClusterProfile.Domain,
 	)
 	err := utillog.EnrichHiveWithCorrelationData(cd, doc.CorrelationData)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/hive/resources.go
```diff
@@ -207,7 +208,7 @@ func adoptedClusterDeployment(namespace, clusterName, clusterID, infraID, locati
 			},
 			Platform: hivev1.Platform{
 				Azure: &hivev1azure.Platform{
-					BaseDomainResourceGroupName: "",
+					BaseDomainResourceGroupName: resourceGroupID,
```
kimorris27, line 211: I guess if E2E passes then my question is answered, but are we sure that the ID goes here given that the parameter is attribute is call `...ResourceGroupName`?

I checked the doc comment in the `Platform` struct definition, but it didn't clarify.

tsatam: It's what we've been putting down on the "install" ClusterDeployment: https://github.com/Azure/ARO-RP/blob/d6d73be4c054f07e2eb440e3e3207de5dfa300c5/pkg/hive/install.go#L224. The fix here is to just update the "adopt" ClusterDeployment (which we apparently put down on Update/AdminUpdate) to do the same.

tsatam: I meant to put this context in the PR description but forgot.

kimorris27 (commented):
E2E failed here: https://github.com/Azure/ARO-RP/blob/d9a1538949bf93bfe1aa9497b147d6f2bc075535/test/e2e/hive.go#L75

Maybe the `APIURLOverride` is getting clobbered on cluster updates and we need to add it to the adopt `ClusterDeployment` like you did for the resource group ID?

tsatam:
> E2E failed here:
> 
> https://github.com/Azure/ARO-RP/blob/d9a1538949bf93bfe1aa9497b147d6f2bc075535/test/e2e/hive.go#L75
> 
> Maybe the `APIURLOverride` is getting clobbered on cluster updates and we need to add it to the adopt `ClusterDeployment` like you did for the resource group ID?

yeah, I just pushed this up, I'll kick off another run.

kimorris27 (approved):

PR #3997, opened 2024-12-09
improve federated cred naming

slawande2:

[ARO-13251](https://issues.redhat.com/browse/ARO-13251)
Fixes

The customer should have a clear idea of what federated credentials are associated with a given cluster.
Currently, we use subscription ID + service account name + hash.
We want to move the naming to cluster name + service account name (without the namespace)  + hash instead, so it’s clear what cluster a federated credential is associated with.

### Test plan for issue:

Updated unit tests
Tested by creating a MIWI cluster

Not yet

### DIFF
--- a/pkg/util/platformworkloadidentity/federatedcredentials.go
+++ b/pkg/util/platformworkloadidentity/federatedcredentials.go
@@ -18,7 +18,7 @@ const (
 )
 
 func GetPlatformWorkloadIdentityFederatedCredName(clusterResourceId, identityResourceId azure.Resource, serviceAccountName string) string {
-	clusterResourceKey := fmt.Sprintf("%s-%s", serviceAccountName, clusterResourceId.ResourceName)
+	clusterResourceKey := fmt.Sprintf("%s_%s", serviceAccountName, clusterResourceId.ResourceName)
 	name := fmt.Sprintf("%s-%s-%s", clusterResourceKey, serviceAccountName, identityResourceId.ResourceName)
 	// the base-36 encoded string of a SHA-224 hash will typically be around 43 to 44 characters long.
 	hash := sha256.Sum224([]byte(name))
--- a/pkg/util/platformworkloadidentity/federatedcredentials.go
+++ b/pkg/util/platformworkloadidentity/federatedcredentials.go
@@ -18,7 +18,7 @@ const (
 )
 
 func GetPlatformWorkloadIdentityFederatedCredName(clusterResourceId, identityResourceId azure.Resource, serviceAccountName string) string {
-	clusterResourceKey := fmt.Sprintf("%s-%s-%s", clusterResourceId.SubscriptionID, clusterResourceId.ResourceGroup, clusterResourceId.ResourceName)
+	clusterResourceKey := fmt.Sprintf("%s-%s", serviceAccountName, clusterResourceId.ResourceName)
 	name := fmt.Sprintf("%s-%s-%s", clusterResourceKey, serviceAccountName, identityResourceId.ResourceName)
 	// the base-36 encoded string of a SHA-224 hash will typically be around 43 to 44 characters long.
 	hash := sha256.Sum224([]byte(name))
--- a/pkg/util/platformworkloadidentity/federatedcredentials.go
+++ b/pkg/util/platformworkloadidentity/federatedcredentials.go
@@ -7,7 +7,6 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"math/big"
-
 	"strings"
 
 	"github.com/Azure/go-autorest/autorest/azure"
@@ -20,7 +19,8 @@ const (
 )
 
 func GetPlatformWorkloadIdentityFederatedCredName(clusterResourceId, identityResourceId azure.Resource, serviceAccountName string) string {
-	sanitizedServiceAccountName := strings.Join((strings.Split(strings.ReplaceAll(serviceAccountName, ":", "-"), "-"))[2:], "-")
+	parts := strings.Split(serviceAccountName, ":")
+	sanitizedServiceAccountName := parts[len(parts)-1]
 	clusterResourceKey := fmt.Sprintf("%s_%s", clusterResourceId.ResourceName, sanitizedServiceAccountName)
 	name := fmt.Sprintf("%s-%s-%s", clusterResourceKey, sanitizedServiceAccountName, identityResourceId.ResourceName)
 	// the base-36 encoded string of a SHA-224 hash will typically be around 43 to 44 characters long.
@@ -29,7 +29,7 @@ func GetPlatformWorkloadIdentityFederatedCredName(clusterResourceId, identityRes
 	remainingChars := maxFederatedCredNameLength - len(encodedName) - numberOfDelimiters
 
 	if remainingChars < len(clusterResourceKey) {
-		return fmt.Sprintf("%s_%s", clusterResourceKey, encodedName)[:remainingChars]
+		return fmt.Sprintf("%s_%s", clusterResourceKey[:remainingChars], encodedName)
 	}
 
 	return fmt.Sprintf("%s_%s", clusterResourceKey, encodedName)
--- a/pkg/util/platformworkloadidentity/federatedcredentials.go
+++ b/pkg/util/platformworkloadidentity/federatedcredentials.go
@@ -20,10 +20,8 @@ const (
 )
 
 func GetPlatformWorkloadIdentityFederatedCredName(clusterResourceId, identityResourceId azure.Resource, serviceAccountName string) string {
-	sanitizedServiceAccountName := strings.ReplaceAll(serviceAccountName, ":", "-")
-	parts := strings.Split(sanitizedServiceAccountName, "-")
-	sanitizedServiceAccountName = strings.Join(parts[2:], "-")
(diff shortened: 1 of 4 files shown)


tsatam (approved):
Implementation LGTM, but now that I'm seeing it I have 1 concern with the new naming format - service account names take the form `system:serviceaccount:${NAMESPACE}:${WORKLOAD_NAME}`, those two parameters often have dashes within them (e.g. `system:serviceaccount:openshift-cloud-controller-manager:cloud-controller-manager`). 

If we suffix `-${CLUSTER_NAME}` to that, we'll end up with e.g. `system:serviceaccount:openshift-cloud-controller-manager:cloud-controller-manager-aro-cluster` or equivalent. 

Perhaps we should use a different separator between the SA name and the cluster name? For example `--`, or `_`? According to https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation-considerations, we're limited to alphanumeric, dash, and underscore characters.

pkg/util/platformworkloadidentity/federatedcredentials.go
```diff
@@ -18,7 +18,7 @@ const (
 )
 
 func GetPlatformWorkloadIdentityFederatedCredName(clusterResourceId, identityResourceId azure.Resource, serviceAccountName string) string {
-	clusterResourceKey := fmt.Sprintf("%s-%s-%s", clusterResourceId.SubscriptionID, clusterResourceId.ResourceGroup, clusterResourceId.ResourceName)
+	clusterResourceKey := fmt.Sprintf("%s_%s", serviceAccountName, clusterResourceId.ResourceName)
```
tsatam, line 21: We'll need to replace any occurrences of `:` in the serviceAccountName as well, `:` is not a valid character for federated credential names. We may want to consider parsing the name and trimming off the first two sections, as these service account names will generally start with `system:serviceaccount:` which is redundant for the purposes of naming the fedcred.

tsatam (dismissed):

pkg/util/platformworkloadidentity/federatedcredentials.go
```diff
@@ -18,15 +20,18 @@ const (
...
+	clusterResourceKey := fmt.Sprintf("%s_%s", sanitizedServiceAccountName, clusterResourceId.ResourceName)
+	name := fmt.Sprintf("%s-%s-%s", clusterResourceKey, sanitizedServiceAccountName, identityResourceId.ResourceName)
 	// the base-36 encoded string of a SHA-224 hash will typically be around 43 to 44 characters long.
 	hash := sha256.Sum224([]byte(name))
 	encodedName := (&big.Int{}).SetBytes(hash[:]).Text(base36Encoding)
 	remainingChars := maxFederatedCredNameLength - len(encodedName) - numberOfDelimiters
 
 	if remainingChars < len(clusterResourceKey) {
-		return fmt.Sprintf("%s-%s", clusterResourceKey[:remainingChars], encodedName)
+		return fmt.Sprintf("%s-%s", clusterResourceKey, encodedName)[:remainingChars]
```
cadenmarchese, line 34: Let's use underscore, and truncate the encodedName rather than the whole string:
```suggestion
		return fmt.Sprintf("%s_%s", clusterResourceKey, encodedName[:remainingChars])
```

pkg/util/platformworkloadidentity/federatedcredentials.go
```diff
@@ -18,15 +20,18 @@ const (
...
 	}
 
 	return fmt.Sprintf("%s-%s", clusterResourceKey, encodedName)
```
cadenmarchese, line 37: Similar to previous comment, let's use an underscore:
```suggestion
	return fmt.Sprintf("%s_%s", clusterResourceKey, encodedName)
```

pkg/util/platformworkloadidentity/federatedcredentials.go
```diff
@@ -18,15 +20,18 @@ const (
 )
 
 func GetPlatformWorkloadIdentityFederatedCredName(clusterResourceId, identityResourceId azure.Resource, serviceAccountName string) string {
-	clusterResourceKey := fmt.Sprintf("%s-%s-%s", clusterResourceId.SubscriptionID, clusterResourceId.ResourceGroup, clusterResourceId.ResourceName)
-	name := fmt.Sprintf("%s-%s-%s", clusterResourceKey, serviceAccountName, identityResourceId.ResourceName)
+	sanitizedServiceAccountName := strings.ReplaceAll(serviceAccountName, ":", "-")
+	parts := strings.Split(sanitizedServiceAccountName, "-")
```
cadenmarchese, line 24: The unit tests are failing because the tests themselves don't supply a valid service account name for you to sanitize. For example, on line 670 of pkg/cluster/workloadidentityresources_test.go:
```
serviceAccount: "openshift-cloud-controller-manager:cloud-controller-manager",
```
Should be this:
```
serviceAccount: "system:serviceaccount:openshift-cloud-controller-manager:cloud-controller-manager",
```
The same is true on pkg/validate/dynamic/platformworkloadidentityprofile_test.go line 427. This:
```
expectedPlatformIdentity1ExtraFederatedCredName := platformworkloadidentity.GetPlatformWorkloadIdentityFederatedCredName(clusterResourceId, platformIdentity1ResourceId, "something else")
```
Should be a valid service account name like this:
```
expectedPlatformIdentity1ExtraFederatedCredName := platformworkloadidentity.GetPlatformWorkloadIdentityFederatedCredName(clusterResourceId, platformIdentity1ResourceId, "system:serviceaccount:something:else")
```
Hopefully this fixes the unit test issues.

cadenmarchese (changes requested):
Left comments on how to fix the unit test issues. thanks for your efforts here!

shubhadapaithankar (dismissed):
LGTM

pkg/util/platformworkloadidentity/federatedcredentials.go
```diff
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"math/big"
```
cadenmarchese, line 10:
```suggestion
```

pkg/util/platformworkloadidentity/federatedcredentials.go
```diff
@@ -18,16 +20,17 @@ const (
 )
 
 func GetPlatformWorkloadIdentityFederatedCredName(clusterResourceId, identityResourceId azure.Resource, serviceAccountName string) string {
-	clusterResourceKey := fmt.Sprintf("%s-%s-%s", clusterResourceId.SubscriptionID, clusterResourceId.ResourceGroup, clusterResourceId.ResourceName)
-	name := fmt.Sprintf("%s-%s-%s", clusterResourceKey, serviceAccountName, identityResourceId.ResourceName)
+	sanitizedServiceAccountName := strings.Join((strings.Split(strings.ReplaceAll(serviceAccountName, ":", "-"), "-"))[2:], "-")
```
cadenmarchese, line 23: Consider using strings.TrimPrefix() instead, since the service account name will always start with `system:serviceaccount:`.

Consider also dropping the namespace from the federated credential name. You could do this by splitting the service account name string like you do now, but keep only the last element.

pkg/util/platformworkloadidentity/federatedcredentials.go
```diff
@@ -18,16 +20,17 @@ const (
...
+	clusterResourceKey := fmt.Sprintf("%s_%s", clusterResourceId.ResourceName, sanitizedServiceAccountName)
+	name := fmt.Sprintf("%s-%s-%s", clusterResourceKey, sanitizedServiceAccountName, identityResourceId.ResourceName)
 	// the base-36 encoded string of a SHA-224 hash will typically be around 43 to 44 characters long.
 	hash := sha256.Sum224([]byte(name))
 	encodedName := (&big.Int{}).SetBytes(hash[:]).Text(base36Encoding)
 	remainingChars := maxFederatedCredNameLength - len(encodedName) - numberOfDelimiters
 
 	if remainingChars < len(clusterResourceKey) {
-		return fmt.Sprintf("%s-%s", clusterResourceKey[:remainingChars], encodedName)
+		return fmt.Sprintf("%s_%s", clusterResourceKey, encodedName)[:remainingChars]
```
cadenmarchese, line 32: Instead of trimming the whole resulting string, we only want to trim the clusterResourceKey (which if trimmed will be the service account name).
```suggestion
		return fmt.Sprintf("%s_%s", clusterResourceKey[:remainingChars], encodedName)
```

pkg/util/platformworkloadidentity/federatedcredentials_test.go
```diff
@@ -29,7 +29,7 @@ func TestGetPlatformWorkloadIdentityFederatedCredName(t *testing.T) {
 	})
 
 	t.Run("has expected key as prefix", func(t *testing.T) {
-		wantPrefix := fmt.Sprintf("%s-%s-%s", subscriptionId, resourceGroup, clusterName)
+		wantPrefix := fmt.Sprintf("%s_%s", clusterName, strings.Join((strings.Split(strings.ReplaceAll(saName, ":", "-"), "-"))[2:], "-"))
```
cadenmarchese, line 32: Consider being explicit here rather than duplicating the strings logic.
```suggestion
		wantPrefix := fmt.Sprintf("%s_%s", clusterName, "openshift-cloud-controller-manager")
```

cadenmarchese (commented):
Thanks again, here are our notes from code review.

cadenmarchese (approved):

shubhadapaithankar (approved):

PR #3998, opened 2024-12-09
Add openshift-ovn-kubernetes NS to permitted NS list in admin actions

tsatam:

Fixes [ARO-13390](https://issues.redhat.com/browse/ARO-13390)

Adds the `openshift-ovn-kubernetes` namespace to the list of OpenShift namespaces permitted for SREs to access using admin API endpoints. 

### Test plan for issue:

- [X] Simple unit test was added to ensure this namespace is returned by the list 

No

We will have to see if it works in production or not, we cannot replicate the full SRE admin environment locally without extreme effort.

### DIFF
--- a/pkg/util/namespace/namespace.go
+++ b/pkg/util/namespace/namespace.go
@@ -72,6 +72,7 @@ func IsOpenShiftNamespace(ns string) bool {
 		"openshift-operator-lifecycle-manager":             {},
 		"openshift-operators":                              {},
 		"openshift-ovirt-infra":                            {},
+		"openshift-ovn-kubernetes":                         {},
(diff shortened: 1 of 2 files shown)


cadenmarchese (approved):

yjst2012 (approved):
Thanks for putting this together

rajdeepc2792 (approved):

kimorris27 (approved):

kimorris27:
E2E has failed with `Timed out waiting for condition` during the cluster install twice in a row. I'm going to try it one more time, and then if it fails again we may need to dig deeper.

PR #3999, opened 2024-12-10
Fix: Ensure valid image tags for PR and master builds.

shubhadapaithankar:

Fixes

### DIFF
--- a/.pipelines/ci.yml
+++ b/.pipelines/ci.yml
@@ -28,11 +28,11 @@ variables:
   - name: VERSION
     value: $(TAG)
   - name: PR_TAG
-    value: $[ coalesce(format('pr-{0}-{1}', variables['System.PullRequest.PullRequestId'], variables['System.PullRequest.SourceCommitId']), '') ]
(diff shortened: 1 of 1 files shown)


.pipelines/ci.yml
```diff
@@ -41,19 +37,56 @@ variables:
...
+    jobs:
+      - job: Set_Tag_Variables
+        displayName: "Set TAG Variables"
+        pool:
+          name: 1es-aro-ci-pool
+        steps:
+          - checkout: self
+
+          - script: |
+              echo "Setting dynamic variables..."
```
tsatam, line 51: Can we not just have this stage output a single `TAG` variable? Do we care in any downstream steps if it's a PR vs branch?

tsatam (changes requested):

tsatam (approved):

.pipelines/ci.yml
shubhadapaithankar: Thanks for the suggestion, Tanmay! I've simplified the logic to output a single TAG variable, removing the need for separate PR_TAG and MASTER_TAG. The downstream steps now use this unified TAG directly, ensuring consistent tagging across all stages.

PR #4000, opened 2024-12-10
Remove montioring controller to allow setting persistence

fahlmant:

Fixes https://issues.redhat.com/browse/ARO-13325

Enables configuration of persistence for the monitoring stack
Test plan for issue:

    Create cluster
    Deploy ARO operator with changes
    Try and edit monitoring configmap
    Verify config is not removed
    Verify ARO operator logs related to removing config are absent.

Probably yes
How do you know this will function as expected in production?

Testing


hlipsig (changes requested):
I think it'd be better to actually add this to a feature flag which defaults to false similar to the allowReconcile one. Then after we've trialed this for a while do the deletion.

PR #4001, opened 2024-12-10
Rename resource endpoint from PlatformWorkloadIdentityRoleSet to PlatformWorkloadIdentityRoleSets

cadenmarchese:

No Jira

Fixes an issue in our swagger where an endpoint is not plural


bennerv (commented):
Need updates in `pkg/frontend/frontend.go` as well.  For updated path.

bennerv (approved):
lgtm

PR #4005, opened 2024-12-12
improve OCP version definition in local RP

Tof1973:

NA

When creating local RP, it was not obvious to create the OCP release.
Also, for DEV we do not need to specify the sha for the image, and this eases the add of a release in the DB.

So I propose this PR.

### Test plan for issue:

NA

NA

NA


Tof1973:
@microsoft-github-policy-service agree company="RedHat"

PR #4007, opened 2024-12-12
Extend pull secret controller unit test to cover Degraded status cond…

lbossis:
https://issues.redhat.com/browse/ARO-13079

Extend pull secret controller unit tests with negative unit test case so that we know the controller's Reconcile function sets the appropriate controller state in response to function failure, which is expected to cause controller to enter a Degraded state. Potentially, an approach of Reconcile function failures on-demand for client.Client attempts to Update k8s objects can be re-used for other controllers.
 
### Test plan for issue:
[x] Unit Test Cases
[] Local Cluster Creation
[] CI
[] E2E

### Are there unit tests?
Yes, https://github.com/Azure/ARO-RP/blob/master/pkg/operator/controllers/pullsecret/pullsecret_controller_test.go

Running all unit tests locally and verifying their output

N/A

### DIFF
--- a/pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
+++ b/pkg/operator/controllers/pullsecret/pullsecret_controller_test.go
@@ -6,8 +6,10 @@ package pullsecret
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"reflect"
 	"testing"
+	"time"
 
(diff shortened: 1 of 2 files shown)


kimorris27 (changes requested):
Mostly LGTM! I suggested a change to a log statement and gave a suggestion on how to fix the lint error.

PR #4011, opened 2024-12-13
Remove unused ClusterManager API structs

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-9171

Removes the invalid ClusterManager API structs + generation code + regenerates the clients.

### Test plan for issue:

Unit test/E2E.

N/A?

Should need to pass ARM review.


cadenmarchese:
thanks! We're partially duplicating this change in https://github.com/Azure/ARO-RP/pull/4014/files for the purposes of getting it reviewed by ARM (and probably breaking changes team too), so I'll put a hold here. The only question I have is why we're removing the resource types from _all_ API versions instead of just subsequent ones. I am not sure that the ARM team would be OK with such a change.

hawkowl:
@cadenmarchese Since they were never wired up and we're implementing things differently, plus they don't conform with the ARM guidelines, it's easier to remove them. We could perhaps just not upload previous APIs if it's too much of an issue, but we don't want to keep all of the generation code around just for it.

tsatam (approved):
LGTM - there are downstream consequences from this change that we'll have to work through, but none of them should block the change being made to ARO-RP itself.

PR #4013, opened 2024-12-13
20241111.00-hotfix - Ensure CI is passing / confirm release payload health

cadenmarchese:
* Update roundtripper.go

Fixes:
Remove RequestURL supplementary data to enhance security of log content.

* Update roundtripper.go

Realized I could easily change to only the base domain. Made that change instead of deleting the field from logging.

Fixes


cadenmarchese:
Closing this PR as we've received a healthy signal here and the release is progressing (other than e2e which is a known issue at the moment)

PR #4017, opened 2024-12-16
Default `MonitoringEnabled` flag to false

fahlmant:

Fixes [ARO-13325](https://issues.redhat.com/browse/ARO-13325)

Defaults to having the monitoring controller in the ARO operator to off. This will allow ARO cluster administrators to enable persistence for the core monitoring stack. 

### Test plan for issue:

Installed cluster, ran operator locally with flag set.

Externally yes

Tested on development cluster

### DIFF
--- a/pkg/operator/flags.go
+++ b/pkg/operator/flags.go
@@ -61,7 +61,7 @@ func DefaultOperatorFlags() map[string]string {
 		MachineSetEnabled:                  FlagTrue,
 		MachineHealthCheckEnabled:          FlagTrue,
 		MachineHealthCheckManaged:          FlagTrue,
-		MonitoringEnabled:                  FlagTrue,
+		MonitoringEnabled:                  FlagFalse,
 		NodeDrainerEnabled:                 FlagTrue,
 		PullSecretEnabled:                  FlagTrue,
 		PullSecretManaged:                  FlagTrue,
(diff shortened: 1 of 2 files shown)


test/e2e/operator.go
```diff
@@ -179,45 +176,6 @@ var _ = Describe("ARO Operator - Geneva Logging", func() {
 	})
 })
 
-var _ = Describe("ARO Operator - Cluster Monitoring ConfigMap", func() {
-	It("must not have persistent volume set", func(ctx context.Context) {
```
bennerv, line 183: can we condition this rather than just remove it?  Condition it on the operator flag.

pkg/operator/flags.go
```diff
@@ -61,7 +61,7 @@ func DefaultOperatorFlags() map[string]string {
 		MachineSetEnabled:                  FlagTrue,
 		MachineHealthCheckEnabled:          FlagTrue,
 		MachineHealthCheckManaged:          FlagTrue,
-		MonitoringEnabled:                  FlagTrue,
+		MonitoringEnabled:                  FlagFalse,
```
bennerv, line 64: This name sucks.  I wonder if we should keep the string but rename the var.  Your choice if you want to leave this as is.

hlipsig (approved):

PR #4020, opened 2024-12-17
Increase ClusterDeployment readiness timeout in CI

kimorris27:

No Jira

About 95% of recent CI cluster creations have failed on the step that waits for the Hive ClusterDeployment to be ready. After investigating, I found that the issue seems to be that it takes a bit longer than 5 minutes for the `Unreachable` status condition on the CD to get into the desired state. Trying to increase this timeout in CI to see if CI starts passing consistently. Then once CI passes consistently, maybe we need to try to figure out why it started taking longer for the CD to be ready.

### Test plan for issue:

CI

N/A

N/A

### DIFF
--- a/pkg/hive/manager.go
+++ b/pkg/hive/manager.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strings"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


pkg/hive/manager.go
```diff
@@ -210,6 +211,20 @@ func (hr *clusterManager) IsClusterInstallationComplete(ctx context.Context, doc
...
+// This allows CI to be resilient to temporary VPN failures that occur while the RP polls the Hive ClusterDeployment.
+// This function can be removed if the VPN gateway is replaced with some less flaky networking solution.
+func (hr *clusterManager) handleClusterDeploymentGetError(err error) error {
+	if hr.env.IsCI() && err != nil && strings.Contains(err.Error(), "http2: client connection lost") {
+		return nil
+	}
+
+	return err
+}
+
```
kimorris27, line 227: Thought: since this error would be transient even if it were to occur in prod, maybe we can skip checking to see if we're in CI?

cadenmarchese (approved):

tsatam (approved):
Looks good as-is, though I agree with dropping the CI check as well if you choose to do that.

pkg/hive/manager.go
kimorris27: Dropped the CI check here 👍🏻

kimorris27:
Closing this in favor of a new PR that will address the situation in a different way.

PR #4022, opened 2024-12-18
aro-12538 - Migrating subnet client in purge.go to use arm.network

ehvs:
https://issues.redhat.com/browse/ARO-12538

Besides the change to arm.network, some functions from arm.network like 'CreateOrUpdateAndWait' and 'Get' requires passing the exact name of multiple arguments rather than using the ID.

To adhere with the new implementation.

### Test plan for issue:

Tested locally and manually triggering the pipeline:
https://msazure.visualstudio.com/5d69ab04-7ded-49dc-84d5-bbbcac4add8d/_apis/build/builds/110655017/logs/13

No. Tech debt.

### DIFF
--- a/pkg/util/purge/resourcegroups.go
+++ b/pkg/util/purge/resourcegroups.go
@@ -96,7 +96,7 @@ func (rc *ResourceCleaner) cleanNetworking(ctx context.Context, resourceGroup mg
 				return err
 			}
 
-			rc.log.Debugf("Removing security group from subnet: %s/%s/%s", *resourceGroup.Name, *networkSecGroup.Name, *subnet.Name)
+			rc.log.Printf("Dettaching NSG from subnet: %s/%s/%s", *resourceGroup.Name, *networkSecGroup.Name, *subnet.Name)
 
 			if !rc.dryRun {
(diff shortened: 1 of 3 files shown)


pkg/util/purge/resourcegroups.go
```diff
@@ -62,35 +64,54 @@ func (rc *ResourceCleaner) cleanResourceGroup(ctx context.Context, resourceGroup
...
+			if err != nil {
+				return err
+			}
+
+			subnetRGName, err := apisubnet.SplitRG(*nsgSubnet.ID)
+			if err != nil {
+				return err
+			}
+
+			subnet, err := rc.subnet.Get(ctx, subnetRGName, vnetName.ResourceName, subnetName, nil)
```
bitoku, line 94: You don't need to parse subnet ID one by one.
You can parse it like this.
https://github.com/Azure/ARO-RP/blob/88f29dee662395c5a34490030c7e15d3477aec3c/pkg/cluster/ensureendpoints.go#L33-L37

ehvs: Implemented, thank you!

bitoku (approved):

PR #4026, opened 2024-12-19
Log Hive CD on install failures

tsatam:

Fixes [ARO-13781](https://issues.redhat.com/browse/ARO-13781)

- Logs Hive CD on install failures

### Test plan for issue:

- [X] Unit tests were added/updated to handle this

No

If we see Hive CDs in RP logs when an install fails, we'll know


kimorris27 (commented):
LGTM! Can revisit with official approval once the other PR merges.

fahlmant (approved):
lgtm

PR #4029, opened 2024-12-20
Fix so that we mirror :latest

hawkowl:
Make sure we mirror :latest so that images with shifting tags get updated correctly.


cadenmarchese (approved):

kimorris27 (approved):

kimorris27 (changes requested):
LGTM, but I think you'll need to reopen the PR using a branch off of the trunk repo so we can run CI.

PR #4030, opened 2024-12-20
Refactor operations endpoint to always return all supported operations

bennerv:

Fixes a bug with our operations endpoint.  

Our operations API does not adhere to spec.  

> The set of operations specified in the response payload must not vary with each API version of the operations API. The Resource Provider service should always return all the operations that are supported across all the API versions of its resource types [1](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/proxy-api-reference.md#exposing-available-operations)

### Test plan for issue:

unit tests


shubhadapaithankar (approved):
LGTM

tsatam (approved):

PR #4033, opened 2024-12-26
Migrate from subnet.Manager to use armnetwork.SubnetsClient

ehvs:
 Card https://issues.redhat.com/browse/ARO-12537 

Migrates the outdated client Manager for the subnet to use armnetwork.

### Test plan for issue:

Unit tests and e2e

No. Tech debt

### DIFF
--- a/pkg/operator/controllers/subnets/subnet_controller_test.go
+++ b/pkg/operator/controllers/subnets/subnet_controller_test.go
@@ -143,7 +143,6 @@ func TestReconcileManager(t *testing.T) {
 
 				subnetObjectMaster := getValidSubnet()
 				mock.EXPECT().Get(gomock.Any(), vnetResourceGroup, vnetName, subnetNameMaster, nil).Return(armnetwork.SubnetsClientGetResponse{Subnet: *subnetObjectMaster}, nil).MaxTimes(2)
-				//mock.EXPECT().Get(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(0)
 				subnetObjectWorker := getValidSubnet()
 				subnetObjectWorker.Properties.NetworkSecurityGroup.ID = to.StringPtr(nsgv1NodeResourceId)
 				mock.EXPECT().Get(gomock.Any(), vnetResourceGroup, vnetName, subnetNameWorker, nil).Return(armnetwork.SubnetsClientGetResponse{Subnet: *subnetObjectWorker}, nil).MaxTimes(2)
--- a/pkg/operator/controllers/subnets/subnet_controller_test.go
+++ b/pkg/operator/controllers/subnets/subnet_controller_test.go
@@ -10,13 +10,13 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/Azure/ARO-RP/pkg/util/pointerutils"
 	armnetwork "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v2"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/sirupsen/logrus"
 	"go.uber.org/mock/gomock"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	"github.com/Azure/ARO-RP/pkg/api"
@@ -78,7 +78,7 @@ func getValidSubnet() *armnetwork.Subnet {
 	for _, endpoint := range api.SubnetsEndpoints {
 		s.Properties.ServiceEndpoints = append(s.Properties.ServiceEndpoints, &armnetwork.ServiceEndpointPropertiesFormat{
 			Service:           to.StringPtr(endpoint),
-			ProvisioningState: ptr.To(armnetwork.ProvisioningStateSucceeded),
+			ProvisioningState: pointerutils.ToPtr(armnetwork.ProvisioningStateSucceeded),
 		})
 	}
 	return s
@@ -405,7 +405,7 @@ func TestReconcileManager(t *testing.T) {
 				subnetObjectMasterUpdate := getValidSubnet()
 				subnetObjectMasterUpdate.Properties.NetworkSecurityGroup.ID = to.StringPtr(nsgv2ResourceId)
 				for i := range subnetObjectMasterUpdate.Properties.ServiceEndpoints {
-					subnetObjectMasterUpdate.Properties.ServiceEndpoints[i].Locations = []*string{ptr.To("*")}
+					subnetObjectMasterUpdate.Properties.ServiceEndpoints[i].Locations = []*string{pointerutils.ToPtr("*")}
 					subnetObjectMasterUpdate.Properties.ServiceEndpoints[i].ProvisioningState = nil
 				}
 				mock.EXPECT().CreateOrUpdateAndWait(gomock.Any(), vnetResourceGroup, vnetName, subnetNameWorker, *subnetObjectMasterUpdate, nil).Return(nil)
@@ -420,7 +420,7 @@ func TestReconcileManager(t *testing.T) {
 
 				subnetObjectWorkerUpdate.Properties.NetworkSecurityGroup.ID = to.StringPtr(nsgv2ResourceId)
 				for i := range subnetObjectWorkerUpdate.Properties.ServiceEndpoints {
-					subnetObjectWorkerUpdate.Properties.ServiceEndpoints[i].Locations = []*string{ptr.To("*")}
+					subnetObjectWorkerUpdate.Properties.ServiceEndpoints[i].Locations = []*string{pointerutils.ToPtr("*")}
 					subnetObjectWorkerUpdate.Properties.ServiceEndpoints[i].ProvisioningState = nil
 				}
 				mock.EXPECT().CreateOrUpdateAndWait(gomock.Any(), vnetResourceGroup, vnetName, subnetNameWorker, *subnetObjectWorkerUpdate, nil).Return(nil)
--- a/pkg/operator/controllers/subnets/subnet_controller_test.go
+++ b/pkg/operator/controllers/subnets/subnet_controller_test.go
@@ -10,7 +10,6 @@ import (
 	"strconv"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


ehvs:
@tiguelu @bitoku This change seems to impact the subnet_nsg.go and subnet_serviceendpoint.go as per messages from golint. Should I address those in the PR to comply?

ehvs:
CI failed, there are unit tests failing locally. WIP

ehvs:
Working on local PUCM tests.

ehvs:
Tested using local ARO operator, and it worked.

pkg/operator/controllers/subnets/subnet_controller_test.go
```diff
@@ -10,18 +10,20 @@ import (
...
 	"testing"
 
-	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
+	armnetwork "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v2"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/sirupsen/logrus"
 	"go.uber.org/mock/gomock"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
```
tiguelu, line 19: Can we use our own implementation of `ptr.To`? 

Reference: https://redhat-internal.slack.com/archives/C02ULBRS68M/p1731940533915429

tiguelu (changes requested):

pkg/operator/controllers/subnets/subnet_controller_test.go
ehvs: Done.

pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -42,18 +48,18 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
 	}
 
 	// if the NSG is assigned && it's the correct NSG - do nothing
-	if subnetObject.SubnetPropertiesFormat.NetworkSecurityGroup != nil && strings.EqualFold(*subnetObject.NetworkSecurityGroup.ID, correctNSGResourceID) {
+	if subnetObject.Properties.NetworkSecurityGroup != nil && strings.EqualFold(*subnetObject.Properties.NetworkSecurityGroup.ID, correctNSGResourceID) {
```
Tof1973, line 51: Wondering if we should not consider subnetObject.Properties.NetworkSecurityGroup.ID pointer nil to protect our code, even if that probably should not happen...

pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -24,15 +25,20 @@ const (
...
+	subnetObject, err := r.subnets.Get(ctx, subnetID.ResourceGroupName, subnetID.Parent.Name, subnetID.Name, nil)
 	if err != nil {
 		if azureerrors.IsNotFoundError(err) {
 			r.log.Infof("Subnet %s not found, skipping", s.ResourceID)
 			return nil
 		}
 		return err
 	}
-	if subnetObject.SubnetPropertiesFormat == nil {
+	if subnetObject.Properties == nil {
```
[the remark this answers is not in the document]
bdematte: I don't think this can happen without throwing an error (and then, exiting the function in the previous block).

pkg/operator/controllers/subnets/subnet_nsg.go
ehvs: Discussed it today in PR Review meeting, will look into implementing!

pkg/operator/controllers/subnets/subnets_controller.go
```diff
@@ -92,13 +93,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
-	azRefreshAuthorizer, err := clusterauthorizer.NewAzRefreshableAuthorizer(r.log, &azEnv, r.client)
+	credential, err := azidentity.NewDefaultAzureCredential(azEnv.DefaultAzureCredentialOptions())
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
-	authorizer, err := azRefreshAuthorizer.NewRefreshableAuthorizerToken(ctx)
+	options := azEnv.ArmClientOptions()
+
+	subnetGroupsClient, err := armnetwork.NewSubnetsClient(resource.SubscriptionID, credential, options)
```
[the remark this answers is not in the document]
ehvs: Good catch. This is a convention when using the  `armnetwork.<resource>GroupsClient`, for this case, would be sufficient to be `subnetClient` since there is no `subnetGroupsClient` : https://pkg.go.dev/github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/armnetwork#NewSubnetsClient

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
```diff
@@ -21,46 +22,49 @@ func (r *reconcileManager) ensureSubnetServiceEndpoints(ctx context.Context, s s
...
-					Locations: &[]string{"*"},
+					Locations: []*string{to.StringPtr("*")},
 				})
 				changed = true
 			}
 		}
 
 		if changed {
-			err = r.subnets.CreateOrUpdate(ctx, s.ResourceID, subnetObject)
+			err = r.subnets.CreateOrUpdateAndWait(ctx, subnetID.ResourceGroupName, subnetID.Parent.Name, subnetID.Name, subnetObject.Subnet, nil)
```
shubhadapaithankar, line 67: Could we consider validating that subnetID.Parent.Name, subnetID.Name, and subnetID.ResourceGroupName are non-empty before invoking CreateOrUpdateAndWait? Just to avoid any unexpected ARM errors from malformed IDs.

pkg/operator/controllers/subnets/subnet_nsg.go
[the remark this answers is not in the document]
fahlmant: Still good to double check IMO. That check helps explicitly avoid nil pointers

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
```diff
@@ -21,46 +22,49 @@ func (r *reconcileManager) ensureSubnetServiceEndpoints(ctx context.Context, s s
...
 		if err != nil {
 			if azureerrors.IsNotFoundError(err) {
 				r.log.Infof("Subnet %s not found, skipping. err: %v", s.ResourceID, err)
 				return nil
 			}
 			return err
 		}
-
-		if subnetObject == nil { // just in case
+		if r.subnets == nil { // just in case
```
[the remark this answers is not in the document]
ehvs: The new client is not a pointer anymore. So in my recent change I completely deleted this line.

pkg/operator/controllers/subnets/subnet_nsg.go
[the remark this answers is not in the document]
ehvs: In further inspection, the nil validation is not actually validating if the subnet ID or name are empty. It will actually pass if any field inside subnetObject is filled. Eg: subnetObject.NetworkSecurityGroup.ID or subnetObject.ID or else. Which dismiss the point of a validation. So I am choosing to remove completely in favor of specific checks when needed.

pkg/operator/controllers/subnets/subnet_serviceendpoint.go
ehvs: Hi @shubhadapaithankar , i was looking into this, and although we could add some manual validation for non-empty, taking a look on the whole repo does not seem a wide-adopted practice.
That said, it is a good point still.
Looking at the latest version of the `parseResourceID` which is the source from where these split fields come from. there are already error checks covered on it: https://github.com/Azure/azure-sdk-for-go/blob/10288adce4ede29da6301416454499a9fdad8db8/sdk/azcore/arm/internal/resource/resource_identifier.go#L64 
Opened a thread here too: https://redhat-internal.slack.com/archives/C02ULBRS68M/p1750341011301219

pkg/operator/controllers/subnets/subnet_nsg.go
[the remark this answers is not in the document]
tiguelu: On closer look, Hevs showed me the new subnetObject is not a pointer any more, it only was in the old implementation. So we can resolve this thread.

pkg/operator/controllers/subnets/subnet_controller_test.go
```diff
@@ -116,11 +118,11 @@ func TestReconcileManager(t *testing.T) {
...
 
 				subnetObjectMaster := getValidSubnet()
-				mock.EXPECT().Get(gomock.Any(), subnetResourceIdMaster).Return(subnetObjectMaster, nil).MaxTimes(2)
+				mock.EXPECT().Get(gomock.Any(), vnetResourceGroup, vnetName, subnetNameMaster, nil).Return(armnetwork.SubnetsClientGetResponse{Subnet: *subnetObjectMaster}, nil).MaxTimes(2)
 
 				subnetObjectWorker := getValidSubnet()
-				subnetObjectWorker.NetworkSecurityGroup.ID = pointerutils.ToPtr(nsgv1NodeResourceId)
-				mock.EXPECT().Get(gomock.Any(), subnetResourceIdWorker).Return(subnetObjectWorker, nil).MaxTimes(2)
+				subnetObjectWorker.Properties.NetworkSecurityGroup.ID = pointerutils.ToPtr(nsgv1NodeResourceId)
+				mock.EXPECT().Get(gomock.Any(), vnetResourceGroup, vnetName, subnetObjectWorker, nil).Return(armnetwork.SubnetsClientGetResponse{Subnet: *subnetObjectWorker}, nil).MaxTimes(2)
```
tiguelu:
```suggestion
				mock.EXPECT().Get(gomock.Any(), vnetResourceGroup, vnetName, subnetNameWorker, nil).Return(armnetwork.SubnetsClientGetResponse{Subnet: *subnetObjectWorker}, nil).MaxTimes(2)
```

Although it isn't making this test case to fail (because it's a disabled case maybe?), it looks like that param needs a change for the test to remain valid should anything else change in the future.

ehvs: Thanks for raising that. Will commit to update it.

PR #4034, opened 2025-01-01
[ARO-11484] Fix fixetcd GA

bitoku:

Fixes https://issues.redhat.com/browse/ARO-11484

This PR fixes fixetcd GA.

* Fix the label selector
* Change the image to ubi9 because it got glibc error with ubi8.
* Change the job to a pod. It only runs one pod, so there's no reason to use a job. Also its watcher returns a job object not a pod object.

Also I added e2e, but it takes so long and it can't run in parallel, so I make it `regression test`.
It doesn't run by default in CI.

E2E test is intended to ensure the master replacement SOP is valid because I couldn't reproduce the etcd issue with 100% possiblity.
I think it's enough solution until we find the reliable reproducible scenario.

### Test plan for issue:

e2e

no

e2e, master replacement & fixetcd GA

### DIFF
--- a/test/e2e/etcd.go
+++ b/test/e2e/etcd.go
@@ -28,7 +28,7 @@ var _ = Describe("Master replacement", Label(regressiontest), func() {
 	BeforeEach(skipIfNotInDevelopmentEnv)
 
 	It("should fix etcd automatically", Serial, func(ctx context.Context) {
-		By("Disable reconciliation")
+		By("Disabling reconciliation")
 		dep, err := clients.Kubernetes.AppsV1().Deployments("openshift-cluster-version").Get(ctx, "cluster-version-operator", metav1.GetOptions{})
 		Expect(err).NotTo(HaveOccurred())
 		dep.Spec.Replicas = pointerutils.ToPtr(int32(0))
@@ -64,7 +64,7 @@ var _ = Describe("Master replacement", Label(regressiontest), func() {
 			g.Expect(machines.Items).To(HaveLen(2))
 		}, 10*time.Minute, 10*time.Second, ctx).Should(Succeed())
 
-		By("Revert deployments") // cluster-version-operator reconciles etcd-operator.
+		By("Reverting deployments") // cluster-version-operator reconciles etcd-operator.
 		dep, err = clients.Kubernetes.AppsV1().Deployments("openshift-cluster-version").Get(ctx, "cluster-version-operator", metav1.GetOptions{})
 		Expect(err).NotTo(HaveOccurred())
 		dep.Spec.Replicas = pointerutils.ToPtr(int32(1))
@@ -80,7 +80,7 @@ var _ = Describe("Master replacement", Label(regressiontest), func() {
 		_, err = clients.MachineAPI.MachineV1beta1().Machines("openshift-machine-api").Create(ctx, &machine, metav1.CreateOptions{})
 		Expect(err).NotTo(HaveOccurred())
 
-		By("Waiting for the machine to be created")
+		By("Waiting for the machine to be created and its node to be ready")
 		Eventually(func(g Gomega, ctx context.Context) {
 			node, err := clients.Kubernetes.CoreV1().Nodes().Get(ctx, machine.Name, metav1.GetOptions{})
 			g.Expect(err).NotTo(HaveOccurred())
--- a/test/e2e/etcd.go
+++ b/test/e2e/etcd.go
@@ -1,0 +1,130 @@
+package e2e
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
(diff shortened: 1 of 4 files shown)


test/e2e/etcd.go
```diff
@@ -0,0 +1,130 @@
...
+			g.Expect(err).NotTo(HaveOccurred())
+		}, 5*time.Minute, 10*time.Second, ctx).Should(Succeed())
+
+		By("Running etcd recovery API")
+		resp, err := adminRequest(ctx, http.MethodPost, "/admin"+clusterResourceID+"/etcdrecovery", nil, true, nil, nil)
+		Expect(err).NotTo(HaveOccurred())
+		// The master replacement doesn't always break the etcd.
+		// It returns 200 and fixes it if broken, and it returns 400 if not broken.
+		// If it gets either of them, we can say that the etcd is fixed (or not broken).
+		Expect(resp.StatusCode).To(Or(Equal(http.StatusOK), Equal(http.StatusBadRequest)))
```
kimorris27, line 107: This makes sense, but if we're going to allow the test to pass in the case where etcd isn't broken, we're not really testing the etcd recovery code. Do you think it'd be better to do something like what I put below?

```
Eventually(
    breakEtcdByReplacingMasterMachine()
    Expect(etcdToBeBroken())
)

runEtcdRecovery()
```

kimorris27 (changes requested):
Mostly LGTM, and it was thoughtful to add an E2E test. I made some small suggestions and have one other thing to point out.

There's a part of the user story that I don't see addressed in the PR: `I want to add a conditional check for when the node IP address remains the same, and delete the existing etcd Pod if it's in a crashloop`. Is that part of the story still needed?

bitoku:
@kimorris27 Thank you for taking a look!

> There's a part of the user story that I don't see addressed in the PR: I want to add a conditional check for when the node IP address remains the same, and delete the existing etcd Pod if it's in a crashloop. Is that part of the story still needed?

I don't think it's actually needed.
During the test, I got the error many times when the pod's IP address is unchanged. I think we need additional change for the state of IP address.
It seems when there's a change in etcd, etcd-operator automatically redeploys all etcd pods.

You might be able to reproduce it by running e2e. When you get 200 from the fixetcd api, the etcd pod should be CrashLoopBackoff.

test/e2e/etcd.go
```diff
@@ -0,0 +1,130 @@
...
+		By("Waiting for the machine to be deleted")
+		Eventually(func(g Gomega, ctx context.Context) {
+			machines, err := clients.MachineAPI.MachineV1beta1().Machines("openshift-machine-api").
+				List(ctx, metav1.ListOptions{LabelSelector: masterMachineLabel})
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(machines.Items).To(HaveLen(2))
+		}, 10*time.Minute, 10*time.Second, ctx).Should(Succeed())
+
+		By("Revert deployments") // cluster-version-operator reconciles etcd-operator.
+		dep, err = clients.Kubernetes.AppsV1().Deployments("openshift-cluster-version").Get(ctx, "cluster-version-operator", metav1.GetOptions{})
```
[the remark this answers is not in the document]
bitoku: No. As commented, cluster-version-operator automatically reverts the etcd-operator.

test/e2e/etcd.go
bitoku: That would be better, but I'm not sure why it's broken and how many times of the replacement needed.
Also this e2e is rather testing if the master replacement SOP works than testing fixetcd API itself. I'd say this e2e is useful enough to test the SOP.

And for personal reason, I already left the ARO team, I can't make the time to investigate and fix like that.

kimorris27: If the intent is mainly to test the master replacement SOP, I think it's fine as-is.

kimorris27 (changes requested):
LGTM given the responses to my original comments. The only issue I see now is that it looks like some unit tests in `pkg/frontend` need to be updated to reflect the changes to `fixetcd.go`, but maybe someone else will need to pick up this work first?

kimorris27 (dismissed):

s-fairchild:
Have you tested this with a dev cluster?

komidore64:
rebased

s-fairchild (dismissed):
LGTM. Since we have E2E tests, we can merge without holding for manual testing.

kimorris27:
> LGTM. Since we have E2E tests, we can merge without holding for manual testing.

Even though we have E2E testing, it still needs to be run manually per Ayato's original description: 

>Also I added e2e, but it takes so long and it can't run in parallel, so I make it regression test.
It doesn't run by default in CI.

I think it would be good for someone to run it once more before merge.

s-fairchild:
> I think it would be good for someone to run it once more before merge.

I agree, my only concern is it's difficult to induce the problem (etcd member list IP mismatch).  
Originally to achieve this I added a new NIC to a master node, then deleted the old NIC, causing a new IP address to be associated with the master. I can try this method again.

komidore64:
Since the PR was closed, is there any opposition to me deleting this branch?

I will assume there is no opposition of I don't see any comments by end-of-day 9 June 2025. Thanks!

PR #4037, opened 2025-01-03
Fix MIWI Cluster update flow to add new openshift operator identity

rajdeepc2792:

Fixes https://issues.redhat.com/browse/ARO-13916
https://issues.redhat.com/browse/ARO-13924

When adding a new OpenShift Identity to an existing cluster, the federated identity and secret in the cluster must be created.
During testing 2 issues were found, this PR fixes them:
1. ObjectID and ClientID are not populated during update for the new operator identity
2. For secret creation, the namespace must exist, so create the namespaces before secret creation.

### Test plan for issue:
[x] Unit Tests
[x] Test the cluster update in local

No

Feature not available in production.


pkg/cluster/install.go
```diff
@@ -220,28 +220,23 @@ func (m *manager) Update(ctx context.Context) error {
...
-		)
-	}
-
-	s = append(s, steps.AuthorizationRetryingAction(m.fpAuthorizer, m.validateResources))
-
-	if m.doc.OpenShiftCluster.UsesWorkloadIdentity() {
-		s = append(s,
 			steps.AuthorizationRetryingAction(m.fpAuthorizer, m.clusterIdentityIDs),
 			steps.AuthorizationRetryingAction(m.fpAuthorizer, m.platformWorkloadIdentityIDs),
-			steps.Action(m.federateIdentityCredentials),
```
tsatam, line 232: Does this step just get skipped entirely on update now? Don't we need it somewhere?

rajdeepc2792: Good catch, I messed it up. Fixing it.

rajdeepc2792: Done

pkg/cluster/install.go
```diff
@@ -220,28 +220,25 @@ func (m *manager) Update(ctx context.Context) error {
...
 			// Since ServicePrincipalProfile is now a pointer and our converters re-build the struct,
 			// our update path needs to enrich the doc with SPObjectID since it was overwritten by our API on put/patch.
 			steps.AuthorizationRetryingAction(m.fpAuthorizer, m.fixupClusterSPObjectID),
-
-			// CSP credentials rotation flow steps
-			steps.Action(m.createOrUpdateClusterServicePrincipalRBAC),
 		)
 	}
 
+	s = append(s, steps.AuthorizationRetryingAction(m.fpAuthorizer, m.validateResources))
```
kimorris27, line 234: Why does this need to be moved down here? I'm thinking that now we could end up with situations where a cx passes a valid MI resource ID, and we either persist its object/client IDs to the cluster doc (or error out with lack of permission trying to do so) before we make it to dynamic validation, which would give the cx a more helpful error message.

cadenmarchese (approved):
thanks for the fix!

pkg/cluster/install.go
rajdeepc2792: Yes that's correct, this is needed only for the new MIs added for the new operators. If the object/client id isn't populated the dynamic validation can't be performed for the new MI as obejct ID is needed in the checkaccessv2 request subject.

kimorris27: Should we take steps to avoid persisting the IDs to the cluster doc if dynamic validation fails then?

rajdeepc2792: Do you mean the resourceID for the new identity or the objectID/ClientID? Why do we want to avoid persisting ID in case the Identity doesn't have the role assignment? The step followed by the Cx would be to perform the role assignment on the provided MI and perform the `az aro update` again if dynamic validation fails.
This is similar to the flow during the cluster creation.

kimorris27: I agree that those would be the correct steps for the cx to move forward, but it doesn't seem right to persist changes to the cluster doc on a request that returns an HTTP 400: https://httpwg.org/specs/rfc9110.html#status.400

The key here for me is:

> the server cannot or will not process the request

Persisting the new MI's metadata and then returning a 400 while we wait for the cx to do the role assignment technically works, but it doesn't cleanly follow the standard HTTP semantics and instead leaves the cluster doc in an intermediary, unresolved state.

rajdeepc2792: Tried to solve the issue by proposing an alternative to perform the doc update only after the dynamic validation. This moves both CMSI and Platform Identities object/client ID populating after dynamic validation in bootstrap as well as update.
Please start the review starting commit: https://github.com/Azure/ARO-RP/pull/4037/commits/4fdfd8573f37193ddcb199ceeaef0daa71a2fd27

rajdeepc2792: Updated the install and update flow, such that the platform workload identities with client ID and object ID aren't updated in the cluster doc until the dynamic validation is performed. Dynamic Validation uses platform workload identities as a field in cluster manager object and not from the cluster doc.

tsatam (approved):

kimorris27 (approved):

PR #4041, opened 2025-01-09
Don't block cluster installation on disableSamples step

kimorris27:

Current prod cluster installation failure for a particular cluster.

The RP is failing to install a particular prod cluster because the sample operator object is not available until about an hour after the `disableSamples` step fails. This is a potential temporary way to help this customer install their cluster while we work on a full solution.

### Test plan for issue:

Updated unit tests.

N/A

Updated unit tests + live attempt on the currently failing cluster 😄

### DIFF
--- a/pkg/cluster/samples.go
+++ b/pkg/cluster/samples.go
@@ -13,14 +13,14 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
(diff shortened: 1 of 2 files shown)


kimorris27:
Update: this would probably not help fix the install failure because even if we allow the RP to progress past this step, the cluster version will be blocked on becoming Ready by the samples operator's degraded/progressing status.

PR #4044, opened 2025-01-13
Revert readdition of network_mode to rp container

kimorris27:

No Jira

Fixes az aro CLI -> local containerized RP connectivity on MacOS by removing `network_mode: host` from docker-compose.yml for the RP container.

### Test plan for issue:

Tested locally and confirmed that it works well for me.

N/A

N/A

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -84,7 +84,6 @@ services:
           soft: 4096
           hard: 4096
     container_name: aro-rp
-    network_mode: host
     depends_on:
       vpn:
         condition: service_healthy


geowa4:
Do we need to change some URLs to `host.containers.internal` to match this networking configuration?

kimorris27:
> Do we need to change some URLs to `host.containers.internal` to match this networking configuration?

Which URLs are you thinking we'd need to change?

geowa4:
> > Do we need to change some URLs to `host.containers.internal` to match this networking configuration?
> 
> 
> 
> @geowa4, which URLs are you thinking we'd need to change?

If anything is referencing localhost and not calling something in the same container, I would expect it to fail. If there are no such calls, then this change is safe. I'm not sure if any exist but I recall they did at some point.

kimorris27:
> > > Do we need to change some URLs to `host.containers.internal` to match this networking configuration?
> > 
> > 
> > @geowa4, which URLs are you thinking we'd need to change?
> 
> If anything is referencing localhost and not calling something in the same container, I would expect it to fail. If there are no such calls, then this change is safe. I'm not sure if any exist but I recall they did at some point.

Gotcha. I can't think of any such calls.

shubhadapaithankar (approved):
LGTM

tsatam (approved):

cadenmarchese (approved):

PR #4045, opened 2025-01-13
Update disableSamples to create samples config resource if it does not exist

tsatam:

Fixes [ARO-14389](https://issues.redhat.com/browse/ARO-14389)

The current implementation of disableSamples fails and retries if the config resource does not exist. This resource is created by the samples operator at the end of its bootstrapping process. However, in some circumstances, this process can take a long time to finish, and as a result this resource will not exist before our disableSamples implementation gives up and exits. 

This PR updates this process to instead, create the config resource if it does not exist, following the recommendations outlined in https://docs.openshift.com/container-platform/4.17/openshift_images/configuring-samples-operator.html#samples-operator-restricted-network-install-with-access. 

Further improvement ideas (not implemented in this PR):
- Should we implement an explicit condition check for ensuring the samples CO is healthy after performing this change, rather than failing implicitly on the "clusterVersionReady" condition (which depends on all COs being healthy)? 

### Test plan for issue:

- [X] Unit tests were updated to handle this case
- ~[ ] Attempt a cluster install with this fix and a reproducer for the failure mode here (e.g. custom routing that redirects traffic for registry.redhat.io to a location that times out requests)~ we were not able to create a reproducer. 

No

Above testing/validation


hlipsig:
LGTM in addition to automated e2e I'd like a couple of by hand tests executed in different regions to ensure we didn't miss anything.

pkg/cluster/samples.go
```diff
@@ -27,7 +28,26 @@ func (m *manager) disableSamples(ctx context.Context) error {
...
 		func() error {
 			c, err := m.samplescli.SamplesV1().Configs().Get(ctx, "cluster", metav1.GetOptions{})
-			if err != nil {
+
+			// if config cr is missing, we create it with managementState: Removed to
+			// immediately disable the samples operator without waiting for it to
+			// bootstrap and create the resource
+			//
+			// https://docs.openshift.com/container-platform/4.17/openshift_images/configuring-samples-operator.html#samples-operator-restricted-network-install-with-access
+			if errors.IsNotFound(err) {
```
hawkowl, line 37: Should we make sure that if this fails with a conflict (e.g. it gets bootstrapped, maybe?) then we leave it alone? Or retry, maybe by using clienthelper's Ensure or something?

tsatam: We already have been retrying on conflicts: https://github.com/Azure/ARO-RP/blob/23600d9a78319c232e7a9e90b39f32264a03f750/pkg/cluster/samples.go#L24.

fahlmant (approved):
lgtm

pkg/cluster/samples.go
hawkowl: Oh, I guess the whole thing being wrapped in it would do it as well. I'm a derp :)

yjst2012 (approved):
LGTM, one minor thing, is it worth pulling the disableSamples step a bit forward, eg, before waiting for the workers to be ready?

```
+			steps.Action(m.disableSamples),
			steps.Condition(m.minimumWorkerNodesReady, 30*time.Minute, true),
```

PR #4046, opened 2025-01-13
Fix MIWI admin update

kimorris27:

https://issues.redhat.com/browse/ARO-12033

This PR fixes several small issues with admin updates on MIWI clusters. Each one of my original three commits to this branch fixes one issue.

### Test plan for issue:

- Tested all four different types of admin updates against a dev cluster and made sure that the update succeeded and no fields in the cluster doc were clobbered/erased
- After each of the admin updates runs, did an admin GET and verified that all MIWI-specific fields were present in cluster doc and available via the admin API (except for the identity URL and the bound service account signing key, which I don't think we need to be there)
- Updated unit tests where applicable

Note that the Jira says we should add unit tests to ensure that none of the MIWI fields are clobbered/erased. It can't really be effectively unit tested. We could add it to MIWI E2E tests though - it would be very easy to check there.

N/A

Will be tested in canary before the MIWI feature goes live.

### DIFF
--- a/pkg/api/admin/openshiftcluster.go
+++ b/pkg/api/admin/openshiftcluster.go
@@ -451,10 +451,10 @@ type PlatformWorkloadIdentity struct {
 // UserAssignedIdentity stores information about a user-assigned managed identity in a predefined format required by Microsoft's Managed Identity team.
 type UserAssignedIdentity struct {
 	// The ClientID of the ClusterUserAssignedIdentity resource
-	ClientID string `json:"clientId,omitempty"`
+	ClientID string `json:"clientId,omitempty" swagger:"readOnly"`
 
 	// The PrincipalID of the ClusterUserAssignedIdentity resource
-	PrincipalID string `json:"principalId,omitempty"`
+	PrincipalID string `json:"principalId,omitempty" swagger:"readOnly"`
 }
 
 // The ManagedServiceIdentity type.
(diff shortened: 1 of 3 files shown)


tsatam (approved):
Change LGTM as-is, but some follow-up questions/actions:

> - After each of the admin updates runs, did an admin GET and verified that all MIWI-specific fields were present in cluster doc and available via the admin API (except for the identity URL and the bound service account signing key, which I don't think we need to be there)

Are the identity URL and boundSASigningKey still present on the clusterdoc in Cosmos and just not displayed on the admin API, or are they actually wiped from the database? Even if we don't _technically_ need these values after a successful install, I think we should try to preserve them in case there's something in the future that requires their use. 


> Note that the Jira says we should add unit tests to ensure that none of the MIWI fields are clobbered/erased. It can't really be effectively unit tested. We could add it to MIWI E2E tests though - it would be very easy to check there.

They're a little cumbersome to write (apologies), but the frontend's putorpatch test would be the place to add unit tests: https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/openshiftcluster_putorpatch_test.go. These are our "these put/patch payloads result in this database result and responses" unit tests as far as I'm aware. I'd be fine with these tests added in a follow-up PR though.

kimorris27:
>Are the identity URL and boundSASigningKey still present on the clusterdoc in Cosmos and just not displayed on the admin API, or are they actually wiped from the database?

They're preserved in CosmosDB but not included in the admin API response.

> the frontend's putorpatch test would be the place to add unit tests

Oh, right. I can fit those unit tests into this PR!

rajdeepc2792 (approved):

cadenmarchese (approved):

PR #4047, opened 2025-01-14
*: update to use the new msi-dataplane library

stevekuznetsov:
go.mod: bump to new Azure/msi-dataplane

---

*: update to use the new msi-dataplane library

---


stevekuznetsov:
Update to use https://github.com/Azure/msi-dataplane/pull/34

pkg/cluster/cluster.go
```diff
@@ -315,36 +314,36 @@ func New(ctx context.Context, log *logrus.Entry, _env env.Interface, db database
 			return nil, err
 		}
 
-		msiDataplaneClientOptions, err := _env.MsiDataplaneClientOptions(msiResourceId)
-		if err != nil {
-			return nil, err
-		}
+		var msiDataplane dataplane.ClientFactory
+		if _env.FeatureIsSet(env.FeatureUseMockMsiRp) {
+			msiDataplane = _env.MockMSIResponses(msiResourceId)
```
stevekuznetsov, line 319: We need to mock out the MSI dataplane, so we can just provide a different implementation for the interface. The previous stub approach of injecting data through an HTTP transport was fragile - it, for one, did not make any distinction between which method was being called!

pkg/cluster/clustermsi.go
```diff
@@ -39,25 +40,29 @@ func (m *manager) ensureClusterMsiCertificate(ctx context.Context) error {
 		return err
 	}
 
-	_, err = m.clusterMsiKeyVaultStore.GetCredentialsObject(ctx, secretName)
-	if err == nil {
-		return nil
-	} else if azcoreErr, ok := err.(*azcore.ResponseError); !ok || azcoreErr.StatusCode != http.StatusNotFound {
-		return err
-	}
+	// TODO: why are we fetching from the keyvault but ignoring the output?
```
stevekuznetsov, line 43: Kipp, could you speak to this?

pkg/util/keyvault/keyvault.go
```diff
@@ -41,6 +41,7 @@ type Manager interface {
 	GetCertificatePolicy(ctx context.Context, certificateName string) (azkeyvault.CertificatePolicy, error)
 	GetCertificateSecret(context.Context, string) (*rsa.PrivateKey, []*x509.Certificate, error)
 	GetSecret(context.Context, string) (azkeyvault.SecretBundle, error)
+	DeleteSecret(ctx context.Context, secretName string) (azkeyvault.DeletedSecretBundle, error)
```
stevekuznetsov, line 44: I didn't fully understand all the different levels of wrapping and mocking going on here since I'm new to the repo, so please let me know if this is not the correct way to go about this. I needed a KeyVault client + mock since `msi-dataplane` lets you bring your own now.

pkg/util/msidataplane/generate.go
```diff
@@ -0,0 +1,14 @@
+package msidataplane
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+//go:generate rm -rf ../mocks/$GOPACKAGE
+//go:generate mockgen -destination=../mocks/$GOPACKAGE/client_factory.go github.com/Azure/msi-dataplane/pkg/dataplane ClientFactory
+//go:generate goimports -local=github.com/Azure/ARO-RP -e -w ../mocks/$GOPACKAGE/client_factory.go
+
+// mockgen unfortunately walks the type aliases in the dataplane package, and tries to import their
```
stevekuznetsov, line 10: This is an unfortunate limitation of mockgen, so we incur some manual work to update the mock for the client in the future if it changes the interface.

pkg/cluster/delete_test.go
```diff
@@ -12,14 +12,12 @@ import (
 
 	sdkmsi "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/msi/armmsi"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v2"
-	"github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azsecrets"
```
stevekuznetsov, line 15: I don't fully understand why the `msi-dataplane` repo used to use a client based on this `azsecrets` package, whereas this repo uses the clients based on the `keyvault` one.

pkg/cluster/clustermsi.go
```diff
@@ -165,13 +191,16 @@ func (m *manager) clusterIdentityIDs(ctx context.Context) error {
...
+		DelegatedResources: &[]string{clusterMsiResourceId.String()},
+	}
+
+	client, err := m.msiDataplane.NewClient(m.doc.OpenShiftCluster.Identity.IdentityURL)
+	if err != nil {
+		return err
 	}
 
-	msiCredObj, err := m.msiDataplane.GetUserAssignedIdentities(ctx, uaMsiRequest)
+	msiCredObj, err := client.GetUserAssignedIdentitiesCredentials(ctx, uaMsiRequest)
```
stevekuznetsov, line 203: Kipp, the names of every client method and the names of the underlying types are up in the air - I couldn't get over how un-helpful `NestedCredentialObject` was, and I hope I've improved the situation, but please feel free to give your preference for any of these names. Field names and their JSON encodings are of course locked down, but the top-level type names and method names are not.

pkg/cluster/clustermsi.go
```diff
@@ -39,25 +40,29 @@ func (m *manager) ensureClusterMsiCertificate(ctx context.Context) error {
...
 
 	clusterMsiResourceId, err := m.doc.OpenShiftCluster.ClusterMsiResourceId()
 	if err != nil {
 		return err
 	}
 
-	uaMsiRequest := dataplane.UserAssignedMSIRequest{
-		IdentityURL: m.doc.OpenShiftCluster.Identity.IdentityURL,
-		ResourceIDs: []string{clusterMsiResourceId.String()},
-		TenantID:    m.doc.OpenShiftCluster.Identity.TenantID,
```
kimorris27, line 57: Where are we passing the tenant ID along to the MSI dataplane now that we're no longer doing it here? I can't find any meaningful references to it.

pkg/cluster/clustermsi.go
kimorris27: The names you've got are clear to me 👍🏻

pkg/cluster/delete_test.go
kimorris27: This repo is still in the process of migrating away from the deprecated clients under `github.com/Azure/azure-sdk-for-go/services`, so you'll find a mixture of the clients from the `services` directory and the clients from the `sdk` directory.

pkg/env/prod.go
```diff
@@ -399,52 +398,87 @@ func (p *prod) MsiRpEndpoint() string {
...
+
+type mockClient struct {
+	aadHost       string
+	msiResourceId string
+}
+
+var _ dataplane.Client = (*mockClient)(nil)
+
+func (m *mockClient) DeleteSystemAssignedIdentity(ctx context.Context) error {
+	panic("not yet implemented")
```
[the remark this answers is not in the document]
kimorris27: Panicking seems fine to me since ARO classic doesn't plan to use system-assigned identities.

pkg/util/msidataplane/generate.go
kimorris27: Oof, this is going to be an issue because it's causing the `generate-check` to fail in CI... maybe we should look at adding an exception to the check for `pkg/util/mocks/msidataplane`.

pkg/util/keyvault/keyvault.go
kimorris27: I think the purpose of this `Manager` interface is to serve as a value-added wrapper around the `BaseClient` for managing cluster certificates (API server and ingress). You could have used the SDK client directly, but this works too.

kimorris27 (commented):
LGTM for the most part! My only concern is that I don't see where the tenant ID is getting passed along to the MSI dataplane.

pkg/cluster/clustermsi.go
stevekuznetsov: This was only being used to default some values in the MSI RP response in case they were not set, I removed that logic as I could not find any reason for it to exist.

pkg/util/msidataplane/generate.go
stevekuznetsov: ACK, let me figure that one out.

pkg/cluster/clustermsi.go
kimorris27: Commenting here to close the loop publicly: I think this means we can remove the tenant ID from our API. I'll start a team-level conversation about it.

stevekuznetsov:
I'll fixup the unit test later today.

cadenmarchese (approved):

PR #4048, opened 2025-01-15
Add disable samples manifest to install

tsatam:

Fixes [ARO-14389](https://issues.redhat.com/browse/ARO-14389)

Extends the implementation of #4045 to additionally pass the disabling samples config as an install manifest to Hive/the installer. This ensures that the cluster is born with the Samples operator disabled, rather than waiting until the RP reaches its post-install step to do so.

The post-install step is still retained to absolutely ensure the operator is disabled after install when it needs to be. 

### Test plan for issue:

- [x] E2E still passes
- [X] Manually confirmed on a local installation that the necessary value is added to the manifests secret in the Hive installation cluster:
  ```
  $ kubectl get secrets -n ${CLUSTER_NS} cluster-manifests-secret -o jsonpath='{.data.cluster-config-samples\.yaml}' | base64 -d
  apiVersion: samples.operator.openshift.io/v1
  kind: Config
  metadata:
    creationTimestamp: null
    name: cluster
  spec:
    managementState: Removed
  status: {}
  ```

No

Above testing/validation

### DIFF
--- a/pkg/cluster/install.go
+++ b/pkg/cluster/install.go
@@ -6,6 +6,7 @@ package cluster
 import (
 	"context"
 	"fmt"
+	"maps"
 	"time"
 
 	"github.com/coreos/go-semver/semver"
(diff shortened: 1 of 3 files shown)


kimorris27 (approved):

shubhadapaithankar (approved):
LGTM

PR #4050, opened 2025-01-16
ARO-12200 include expected values for MIWI arguments in az aro create usage

yithian:

Fixes [ARO-12200](https://issues.redhat.com/browse/ARO-12200)

This updates the usage for `az aro create` to show the expected values for the `--assign-platform-workload-identity`/`--assign-platform-wi` and `--mi-user-assigned`/`--assign-cluster-identity` arguments. This should make it easier for users to provide the expected input

### Test plan for issue:

Ran `az aro create -h` after configuring `az` to use my local files as the aro extension

No, this is updating documentation to match expectations

### DIFF
--- a/python/az/aro/azext_aro/_params.py
+++ b/python/az/aro/azext_aro/_params.py
@@ -138,7 +138,7 @@ def load_arguments(self, _):
                    options_list=['--enable-managed-identity', '--enable-mi'],
                    validator=validate_enable_managed_identity)
         c.argument('platform_workload_identities', arg_group='Identity', is_preview=True,
-                   help='Assign a platform workload identity used within the cluster. Requires two values: the use inside the cluster and the user assigned identity.',
+                   help='Assign a platform workload identity used within the cluster. Requires two values: the user inside the cluster and the user assigned identity.',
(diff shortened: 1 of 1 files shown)


python/az/aro/azext_aro/_params.py
```diff
@@ -138,12 +138,12 @@ def load_arguments(self, _):
                    options_list=['--enable-managed-identity', '--enable-mi'],
                    validator=validate_enable_managed_identity)
         c.argument('platform_workload_identities', arg_group='Identity', is_preview=True,
-                   help='Assign a platform workload identity used within the cluster',
+                   help='Assign a platform workload identity used within the cluster. Requires two values: the use inside the cluster and the user assigned identity.',
```
fahlmant, line 141:
```suggestion
                   help='Assign a platform workload identity used within the cluster. Requires two values: the user inside the cluster and the user assigned identity.',
```

Just for clarification, is the user inside the cluster a ServiceAccount or a User? Same for the Azure side; what's the Azure terminology for the object?

fahlmant (approved):
lgtm

rhamitarora (approved):

rajdeepc2792 (approved):

PR #4049, opened 2025-01-15
Fix bugs for MIWI e2e tests

rajdeepc2792:
### Which issue this PR addresses:

Fixes
https://issues.redhat.com/browse/ARO-14420
https://issues.redhat.com/browse/ARO-14421
https://issues.redhat.com/browse/ARO-14422

### What this PR does / why we need it:

Fixes issues identified as part of MIWI e2e run https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=112549139&view=results
### Test plan for issue:
e2e tests

### Is there any documentation that needs to be updated for this PR?
No

### How do you know this will function as expected in production? 
e2e must pass for MIWI

PR #4056, opened 2025-01-22
[ARO-13400] Old E2E pipeline uses MiWi clusters

mrWinston:

Fixes ARO-6153
Fixes ARO-13400
Fixes ARO-6196

This PR is the amalgamation of most miwi e2e efforts from @bitoku and Me. Roughly speaking, it implements the following:
- Support miwi cluster creation in [hack/cluster/cluster.go](https://github.com/Azure/ARO-RP/blob/db9f2241faf333ec890ecc19dc8fd57d2819fb56/hack/cluster/cluster.go)
- Modify [old e2e pipeline](https://github.com/Azure/ARO-RP/blob/db9f2241faf333ec890ecc19dc8fd57d2819fb56/.pipelines/e2e.yml) to create and test a miwi cluster
- Fix E2E test failures that came up after running miwi e2e tests the first time

After this PR is merged, it is possible to run e2e tests against a miwi cluster on a PR by triggering the old pipeline with this comment:
```
/azp run e2e
```


.pipelines/e2e.yml
```diff
@@ -86,14 +111,20 @@ jobs:
...
 
+          update_role_sets
+
           run_rp
           validate_rp_running
 
           register_sub
 
+          # cluster is cleaned up in a later step. Keep it around for now to enable must-gather collection on error
+          export E2E_DELETE_CLUSTER=false
```
rajdeepc2792, line 127: Should we note or create a task to change it to true whenever we are done with collecting errors.

go.sum
```diff
@@ -601,6 +605,8 @@ github.com/rogpeppe/go-internal v1.12.0/go.mod h1:E+RYuTGaKKdloAfM02xzb0FW3Paa99
 github.com/russross/blackfriday v1.6.0 h1:KqfZb0pUVN2lYqZUYRddxF4OR8ZMURnJIG5Y3VRLtww=
 github.com/russross/blackfriday v1.6.0/go.mod h1:ti0ldHuxg49ri4ksnFxlkCfN+hvslNlmVHqNRXXJNAY=
 github.com/russross/blackfriday/v2 v2.1.0/go.mod h1:+Rmxgy9KzJVeS9/2gXHxylqXiyQDYRxCVz55jmeOWTM=
+github.com/sagikazarmark/locafero v0.6.0 h1:ON7AQg37yzcRPU69mt7gwhFEBwxI6P9T4Qu3N51bwOk=
```
hlipsig, line 608: Is this a library we're already using?

hlipsig (changes requested):
I noticed several new libraries being imported some of which are not in fully supported states. Do we already use these libraries elsewhere? Do we know the history behind these selections?

.pipelines/e2e.yml
```diff
@@ -86,14 +111,20 @@ jobs:
           run_selenium
           validate_selenium_running
 
+          update_role_sets
+
```
cadenmarchese, line 115: Do we need to update the role sets twice? (see lines 119-120 as well)

hack/devtools/local_dev_env.sh
```diff
@@ -11,7 +10,7 @@ set -o pipefail
 
 PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS='[
     {
-        "openShiftVersion": "4.14",
+        "openShiftVersion": "4.15",
```
cadenmarchese, line 13: I would suggest populating this file with the full 4.14 and 4.15 role sets in case someone sources this script and wants to create a 4.14 cluster instead.

hack/devtools/local_dev_env.sh
```diff
@@ -159,6 +158,13 @@ get_platform_workloadIdentity_role_sets() {
 assign_role_to_identity() {
     local objectId=$1
     local roleId=$2
+    
+    # if CLUSTER_RESOURCEGROUP is unset, it should probably be set to the cluster name
+    if [[ "${CLUSTER_RESOURCEGROUP}" == "" ]]; then
```
cadenmarchese, line 163: Should we add an `if os.Getenv("CI")` to this if case as well? My thinking is that if it's CI e2e, we want cluster and resource group name to match. If not CI, we want the user to specify both.

hack/devtools/local_dev_env.sh
```diff
@@ -251,20 +257,15 @@ create_miwi_env_file() {
     mockCert=$(get_mock_msi_cert "$mockMSI")
     mockObjectID=$(get_mock_msi_objectID "$mockClientID")
 
-    setup_platform_identity
```
cadenmarchese, line 254: Why remove this step?

cadenmarchese (commented):
I left some comments on the pipeline / bash components here. Thanks for your efforts!

rajdeepc2792:
Just to confirm, this PR merge won't be impacting the Prod release pipeline, or in other words will the MIWI e2e cluster creation will run as part of Prod release pipeline? We don't want it to be running in the Prod until the preview API endpoints are available there.

mrWinston:
@rajdeepc2792 
> Just to confirm, this PR merge won't be impacting the Prod release pipeline, or in other words will the MIWI e2e cluster creation will run as part of Prod release pipeline?
This PR shouldn't have an impact on the prod pipeline. Using a miwi cluster for the E2E tests needs to be explicitly enabled in the pipeline.

.pipelines/e2e.yml
mrWinston: The comment is referring to the e2e cluster, not the variable.
If this job (`Execute Tests`) fails, the pipeline tries to collect a must gather in the subsequent job. Regardless of weather the tests failed or succeeded, the cleanup job is always run.
This means that we should always keep the cluster around after the e2e tests finished (by exporting `E2E_DELETE_CLUSTER=false`), as it's cleaned up anyway. And if we don't, then the pipeline won't be able to collect the must gather in case of e2e test failure.

Does that make sense?

go.sum
mrWinston: I added [viper](https://github.com/spf13/viper) as a dependency to help managing the different envioronment variables as a config struct ( see [ClusterConfig](https://github.com/Azure/ARO-RP/blob/db9f2241faf333ec890ecc19dc8fd57d2819fb56/pkg/util/cluster/cluster.go#L57) definition and viper usage in [NewClusterConfigFromEnv](https://github.com/Azure/ARO-RP/blob/db9f2241faf333ec890ecc19dc8fd57d2819fb56/pkg/util/cluster/cluster.go#L121))

All the addiitonal dependencies are pulled in as dependencies of viper itself.

.pipelines/e2e.yml
mrWinston: Probably not, good catch :)

hack/devtools/local_dev_env.sh
mrWinston: Good idea. I'll just need to make sure that all scripts are able to select the correct roleset based on the version. Right now, i think we only ever expect a single roleset to be present.

hack/devtools/local_dev_env.sh
mrWinston: Thinking about this for a bit, i think the most straightforward way would be to just remove this check here and instead explicitly set `$CLUSTER_RESOURCEGROUP` to the cluster name in [run-rp-and-e2e.sh](https://github.com/Azure/ARO-RP/blob/db9f2241faf333ec890ecc19dc8fd57d2819fb56/hack/e2e/run-rp-and-e2e.sh#L15).

From a general design standpoint, i don't really like that it's not at all clear what setting `CI=true` does. So i'm more in favor of making those settings explicit.

hack/devtools/local_dev_env.sh
mrWinston: This step is handled in the go code [here](https://github.com/Azure/ARO-RP/blob/db9f2241faf333ec890ecc19dc8fd57d2819fb56/pkg/util/cluster/cluster.go#L352) and is currently run as part of the cluster creation.

pkg/cluster/install.go
```diff
@@ -272,6 +272,7 @@ func (m *manager) Update(ctx context.Context) error {
 		steps.Action(m.fixUserAdminKubeconfig),
 		steps.Action(m.reconcileLoadBalancerProfile),
 		steps.Action(m.reconcileSoftwareDefinedNetwork),
+		steps.Action(m.ensureCredentialsRequest),
```
kimorris27, line 275: Why do we need to do this unconditionally when we were previously doing it only for service principal clusters?

pkg/util/cluster/cluster.go
```diff
@@ -194,36 +298,215 @@ func (c *Cluster) createApp(ctx context.Context, clusterName string) (applicatio
...
+				)
+
+				// Ignore if the role assignment already exists
+				if detailedError, ok := err.(autorest.DetailedError); ok {
+					if detailedError.StatusCode == http.StatusConflict {
+						err = nil
+					}
+				}
+
+				// TODO: tighten this error check
```
kimorris27, line 333: If we're not going to do this in this PR, let's write a Jira for it and include the Jira ticket number in this comment.

pkg/util/cluster/cluster.go
```diff
@@ -194,36 +298,215 @@ func (c *Cluster) createApp(ctx context.Context, clusterName string) (applicatio
 	return appDetails{appID, appSecret, spID}, nil
 }
 
-func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName string, osClusterVersion string) error {
-	clusterGet, err := c.openshiftclusters.Get(ctx, vnetResourceGroup, clusterName)
+// Missing conf:
+// appDetails
+func (c *Cluster) SetupClassicRoleAssignments(ctx context.Context, diskEncryptionSetID string, clusterServicePrincipalID string) error {
```
kimorris27, line 303:
```suggestion
func (c *Cluster) SetupServicePrincipalRoleAssignments(ctx context.Context, diskEncryptionSetID string, clusterServicePrincipalID string) error {
```

Suggesting this change for clarity, since CSP and MIWI clusters are both "ARO classic" clusters.

pkg/util/cluster/cluster.go
```diff
@@ -194,36 +298,215 @@ func (c *Cluster) createApp(ctx context.Context, clusterName string) (applicatio
 	return appDetails{appID, appSecret, spID}, nil
 }
 
-func (c *Cluster) Create(ctx context.Context, vnetResourceGroup, clusterName string, osClusterVersion string) error {
-	clusterGet, err := c.openshiftclusters.Get(ctx, vnetResourceGroup, clusterName)
+// Missing conf:
+// appDetails
```
kimorris27, line 302: Is this comment needed?

pkg/util/cluster/cluster.go
```diff
@@ -194,36 +298,215 @@ func (c *Cluster) createApp(ctx context.Context, clusterName string) (applicatio
...
+			return err
+		}
+
+		if wi.OperatorName != "aro-Cluster" {
+			c.workloadIdentities[wi.OperatorName] = api.PlatformWorkloadIdentity{
+				ResourceID: *resp.ID,
+			}
+		}
+		// TODO Remove this once builtin roles for disk-csi-driver has
+		// `Microsoft.Compute/diskEncryptionSets/read` permission
```
kimorris27, line 423: Let's track this with a Jira if we aren't already.

pkg/util/cluster/cluster.go
```diff
@@ -194,36 +298,215 @@ func (c *Cluster) createApp(ctx context.Context, clusterName string) (applicatio
...
 	}
 
-	fpSPId := os.Getenv("AZURE_FP_SERVICE_PRINCIPAL_ID")
-	if fpSPId == "" {
-		return fmt.Errorf("fp service principal id is not found")
-	}
-
-	appDetails, err := c.createApp(ctx, clusterName)
+	c.log.Info("Creating app")
+	appDetails, err := c.createApp(ctx, c.Config.ClusterName)
```
kimorris27, line 495: Why do we create this app unconditionally even though only service principal clusters need it?

pkg/util/cluster/cluster.go
```diff
@@ -730,6 +1019,9 @@ func (c *Cluster) fixupNSGs(ctx context.Context, vnetResourceGroup, clusterName
 }
 
 func (c *Cluster) deleteRoleAssignments(ctx context.Context, vnetResourceGroup, clusterName string) error {
+	if c.Config.UseWorkloadIdentity {
+		c.log.Print("Skipping deletion of classic role assignmnets")
```
kimorris27, line 1023:
```suggestion
		c.log.Print("Skipping deletion of service principal role assignments")
```

See context in a different comment.

pkg/util/cluster/cluster_test.go
```diff
@@ -0,0 +1,140 @@
...
+			if gotErr != nil {
+				if !tt.wantErr {
+					t.Errorf("GetPlatformWIRoles() failed: %v", gotErr)
+				}
+				return
+			}
+			if tt.wantErr {
+				t.Fatal("GetPlatformWIRoles() succeeded unexpectedly")
+			}
+			// TODO: update the condition below to compare got with tt.want.
```
kimorris27, line 134: This comment seems like it can be removed.

pkg/validate/dynamic/dynamic.go
```diff
@@ -493,6 +493,9 @@ func (c *closure) checkAccessAuthReqToken() error {
 // usingCheckAccessV2 uses the new RBAC checkAccessV2 API
 func (c closure) usingCheckAccessV2() (result bool, err error) {
 	c.dv.log.Info("validateActions with CheckAccessV2")
+	if c.dv.checkAccessSubjectInfoCred == nil {
```
kimorris27, line 496: What does this change do? I'm curious because it caught me by surprise in this PR.

kimorris27 (changes requested):
Nice work! I have a few questions and suggestions.

pkg/cluster/install.go
mrWinston: That's a very good question. This change was part of several changes done by @rajdeepc2792 to fix the failing e2e tests for miwi cluster: https://github.com/Azure/ARO-RP/pull/4049

@rajdeepc2792  Do you still know what the reason for this change was?

pkg/util/cluster/cluster.go
mrWinston: To be honest, i don't really know what this comment is about. I just moved this code from `cluster.create()` into it's own method to clean it up a bit, but didn't touch the actual logic (here on the master branch: https://github.com/Azure/ARO-RP/blob/master/pkg/util/cluster/cluster.go#L352). 

And looking at git blame, it seems like this comment has been present in the code for the 5 years :) 
I'd be in favor of just removing it. What do you think?

kimorris27: Oh, gotcha. Removing the comment with no other action works for me then 👍🏻

pkg/util/cluster/cluster.go
mrWinston: I initially kept this step cause this is used as input for the predeploy arm template:
```go
	parameters := map[string]*arm.ParametersParameter{
		"clusterName":               {Value: c.Config.ClusterName},
		"ci":                        {Value: c.Config.IsCI},
		"clusterServicePrincipalId": {Value: appDetails.SPId},
		"fpServicePrincipalId":      {Value: c.Config.FPServicePrincipalID},
		"vnetAddressPrefix":         {Value: addressPrefix},
		"masterAddressPrefix":       {Value: masterSubnet},
		"workerAddressPrefix":       {Value: workerSubnet},
		"kvName":                    {Value: kvName},
	}
```

But i just looked into this a bit more, and found that the parameter isn't even used in the [arm template](https://github.com/Azure/ARO-RP/blob/master/pkg/deploy/assets/cluster-development-predeploy.json). I'll go ahead and remove the app definition!

pkg/validate/dynamic/dynamic.go
mrWinston: to be honest, i'm not entirely sure. This is part of the changes implemented by ayato before i took the work over.
This might have been a quick and dirty workaround for https://issues.redhat.com/browse/ARO-12034 before https://github.com/Azure/ARO-RP/pull/3946 got merged. 

Let me temporarily remove this and trigger another pipeline run to see if this is still needed.

mrWinston: The e2e tests are still green with the change removed.  Seems like the change isn't needed anymore. Good catch!

pkg/util/cluster/cluster.go
mrWinston: Created [ARO-15117](https://issues.redhat.com/browse/ARO-15117) to keep track of this and added the ticket number to the comment.

hack/devtools/local_dev_env.sh
mrWinston: Done!

go.sum
rajdeepc2792: It was done to fix this e2e failure: https://issues.redhat.com/browse/ARO-14420
Reasoning is in comments.

pkg/cluster/install.go
rajdeepc2792: It was done to fix this e2e failure: https://issues.redhat.com/browse/ARO-14420 
Reasoning is in comments.

.pipelines/e2e.yml
```diff
@@ -58,6 +58,20 @@ jobs:
         displayName: Enable regression tests in CI
         condition: in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI')
 
+      - script: |
```
cadenmarchese, line 61: Is it easy for us to rename this pipeline/azp run target to something like /azp run e2e-wimi or something similar rather than `/azp run e2e`? We can handle it in a future PR as well, if that's easier.

This was raised as something that could cause confusion for contributors who want to re-run CI e2e as normal but end up kicking off a MIWI e2e instead.

cadenmarchese (commented):
This LGTM, I posed a few questions, but am otherwise good to approve this on next rebase.

pkg/cluster/install.go
kimorris27: Makes sense! Thanks for providing that context.

hack/devtools/platform_workload_identity_role_sets.json
```diff
@@ -0,0 +1,144 @@
...
+                "operatorName": "aro-operator",
+                "roleDefinitionName": "Azure Red Hat OpenShift Service Operator",
+                "roleDefinitionId": "/providers/Microsoft.Authorization/roleDefinitions/4436bae4-7702-4c84-919b-c4069ff25ee2",
+                "serviceAccounts": ["system:serviceaccount:openshift-azure-operator:aro-operator-master"],
+                "secretLocation": { "namespace": "openshift-azure-operator", "name": "azure-cloud-credentials" }
+            }
+        ]
+    },
+    {
+        "openShiftVersion": "4.14",
```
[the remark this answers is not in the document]
kimorris27: Since these are JSON manifests representing static data, I think duplicating the JSON for each version is probably the correct approach here. As a developer trying to read and understand the role sets for a given version, I wouldn't want to have to, for example, read the role set for 4.14 and then additionally parse through the incremental changes added in each version between 4.14 and the version I care about.

kimorris27 (commented):
LGTM pending a rebase and responses to Caden's feedback.

.pipelines/e2e.yml
mrWinston: i'm looking into that. currently i'm unsure how to rename a pipeline.

mrWinston: Renaming the pipeline isn't possible from this PR, but instead has to happen in the azure pipelines ui. I don't have permission to do so, but i'll find someone who does and get this renamed!

pkg/cluster/deploybaseresources.go
```diff
@@ -209,6 +209,12 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
 		}
 
 		resources = append(resources, r...)
+
+		storageBlobContributorRBAC, err := m.fpspStorageBlobContributorRBAC(clusterStorageAccountName, m.fpServicePrincipalID)
```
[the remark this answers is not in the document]
mrWinston: No, this was just moved a couple of lines upwards so there aren't two `if m.doc.OpenShiftCluster.UsesWorkloadIdentity(){...}` statements. Functionally, it is identical.

pkg/cluster/deploybaseresources.go
```diff
@@ -231,24 +237,16 @@ func (m *manager) deployBaseResourceTemplate(ctx context.Context) error {
 		)
 	}
 
+	if !m.env.FeatureIsSet(env.FeatureDisableDenyAssignments) {
```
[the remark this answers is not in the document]
mrWinston: Just stylistically it made more sense to first have all elements added to `resources` before creating the `arm.Template` with it. But functionally, it's the same

.pipelines/e2e.yml
mrWinston: I created a new ticket to address the renaming:
https://issues.redhat.com/browse/ARO-15569

Renaming the Pipeline can happen independently from this PR though, so we should be good to move forward

kimorris27 (commented):
LGTM, thank you for the hard work on this over a long period of time! I will defer to @cadenmarchese for official approval and merge.

mrWinston:
Build error should now be fixed, i made a small error when merging  the master branch. Restarting the pipelines now.

PR #4070, opened 2025-01-30
Implementation of MISE middleware

sankur-codes:

Fixes ARO-5946. Updated PR for #3555 

This PR integrates MISE authentication for our RP to improve security while keeping existing certificate based authentication as a fallback.

### Test plan for issue:

Unit tests are included. Testing has been done on dev and Canary.

No

Been tested on INT and Canary as well with expected results.


pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -25,27 +25,33 @@ enable_services() {
...
     local -n role="$2"
     local -n conf_file="$3"
+    local -n ipaddress="$4"
     log "starting"
     log "Configuring aro-gateway service"
 
     local -r aro_gateway_conf_filename='/etc/sysconfig/aro-gateway'
+    local -r add_conf_file="PODMAN_NETWORK='podman'
+IPADDRESS='$ipaddress'
+ROLE='${role,,}'"
```
s-fairchild, line 40: What is the purpose of this?

Uppercase variables are meant to communicate that a variable is used externally. `role` is used internal to this function.

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -84,26 +92,32 @@ WantedBy=multi-user.target
...
     local -n role="$2"
     local -n conf_file="$3"
+    local -n ipaddress="$4"
     log "starting"
     log "Configuring aro-rp service"
 
     local -r aro_rp_conf_filename='/etc/sysconfig/aro-rp'
+    local -r add_conf_file="PODMAN_NETWORK='podman'
+IPADDRESS='$ipaddress'
+ROLE='${role,,}'"
```
s-fairchild, line 108: Same as above.

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -183,14 +202,16 @@ DATABASE_ACCOUNT_NAME='$DATABASEACCOUNTNAME'
 KEYVAULT_PREFIX='$KEYVAULTPREFIX'
 MDM_ACCOUNT='$RPMDMACCOUNT'
 MDM_NAMESPACE=BBM
-RPIMAGE='$image'"
+RPIMAGE='$image'
+PODMAN_NETWORK='podman'
+IPADDRESS='$ipaddress'"
```
s-fairchild, line 207: Same as above

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -251,14 +276,16 @@ MDM_ACCOUNT='$RPMDMACCOUNT'
 MDM_NAMESPACE=Portal
 PORTAL_HOSTNAME='$LOCATION.admin.$RPPARENTDOMAINNAME'
 OTEL_AUDIT_QUEUE_SIZE='$OTELAUDITQUEUESIZE'
-RPIMAGE='$image'"
+RPIMAGE='$image'
+PODMAN_NETWORK='podman'
+IPADDRESS='$ipaddress'"
```
s-fairchild, line 281: Same as above.

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -286,17 +315,232 @@ ExecStart=/usr/bin/podman run \
...
+    log "starting"
+    log "Configuring aro-otel-collector service"
+
+    # shellcheck disable=SC2034
+    local -r aro_otel_collector_service_conf_filename='/etc/sysconfig/aro-otel-collector'
+    # shellcheck disable=SC2034
+    local -r aro_otel_collector_service_conf_file="GOMEMLIMIT=1000MiB
+OTELIMAGE='$image'
+PODMAN_NETWORK='podman'
+IPADDRESS='$ipaddress'"
```
s-fairchild, line 464: Same as above.

pkg/deploy/devconfig.go
```diff
@@ -171,6 +172,13 @@ func DevConfig(_env env.Core) (*Config, error) {
 			MDMFrontendURL:              pointerutils.ToPtr("https://global.ppe.microsoftmetrics.com/"),
 			MDSDEnvironment:             pointerutils.ToPtr(version.DevGenevaLoggingEnvironment),
 			MsiRpEndpoint:               pointerutils.ToPtr("https://iamaplaceholder.com"),
+			MiseValidAudiences: []string{
+				"https://management.core.windows.net/",
+				_env.Environment().ResourceManagerEndpoint,
+			},
+			MiseValidAppIDs: []string{
+				"2187cde1-7e28-4645-9104-19edfa500053",
```
pepedocs, line 180: Can we make a more descriptive variable for this? Who's app id is this?

pkg/deploy/generator/resources_rp.go
```diff
@@ -410,6 +411,18 @@ func (g *generator) rpVMSS() *arm.Resource {
...
 	}
 
+	// convert array variables to string using ARM string() function to be passed via customScript later
+	for _, variable := range []string{
+		"miseValidAudiences",
+		"miseValidAppIDs",
+	} {
+		parts = append(parts,
+			fmt.Sprintf("'%s=$(base64 -d <<<'''", strings.ToUpper(variable)),
+			fmt.Sprintf("base64(string(parameters('%s')))", variable),
```
pepedocs, line 421: If this variable is manually input, there's a chance a space character is added. To ensure there is no extra character, I think it's best to trim this variable first.

pepedocs (changes requested):
Everything looks fine to me. Some nitipicks. But I would be more confident to see a unitest for the authenticate part, including tests for the error logs and errors we expect to see when there are errors . Reason for this is that auth is a critical part of rp and ensuring right errors returned are also important in debugging.

pkg/frontend/ready_get.go
```diff
@@ -24,9 +24,18 @@ func (f *frontend) checkReady() bool {
 	_, okOcpVersions := f.lastOcpVersionsChangefeed.Load().(time.Time)
 	_, okPlatformWorkloadIdentityRoleSets := f.lastPlatformWorkloadIdentityRoleSetsChangefeed.Load().(time.Time)
 
+	var miseAuthReady, armAuthReady, authReady bool
+	if f.authMiddleware.EnableMISE {
+		miseAuthReady = f.env.MISEAuthorizer().IsReady()
+	}
+	// skip ARM Authorizer is MISE is Enforcing
+	if !f.authMiddleware.EnforceMISE {
```
geowa4, line 32: It looks like this should be an `else`? If we do that, we might be able to get rid of `miseAuthReady` and `armAuthReady`.

pkg/util/miseadapter/miseadapter.go
```diff
@@ -0,0 +1,194 @@
...
+
+	req.Header.Add("Authorization", input.AuthorizationHeader)
+	req.Header.Add("Original-Uri", input.OriginalUri)
+	req.Header.Add("Original-Method", input.OriginalMethod)
+	req.Header.Add("X-Forwarded-For", input.OriginalIPAddress)
+
+	if input.ReturnAllActorClaims {
+		req.Header.Add("Return-All-Actor-Token-Claims", "1")
+	} else {
+		for _, val := range input.ActorClaimsToReturn {
```
geowa4, line 108: I didn't see this being used in this PR in any code but tests. Is that a miss, do we intend on adding it later (if so, which ticket?), or do we only include this here for compliance to a spec?

pkg/util/miseadapter/miseadapter.go
```diff
@@ -0,0 +1,194 @@
...
+	res := Result{
+		StatusCode:    response.StatusCode,
+		SubjectClaims: map[string][]string{},
+		ActorClaims:   map[string][]string{},
+	}
+
+	subjectTokenClaimPrefix := "Subject-Token-Claim-"
+	actorTokenClaimPrefix := "Actor-Token-Claim-"
+	encodedSubjectTokenClaimPrefix := "Subject-Token-Encoded-Claim-"
+	encodedActorTokenClaimPrefix := "Actor-Token-Encoded-Claim-"
```
geowa4, line 136: These should be constants.

pkg/env/prod.go
```diff
@@ -249,20 +251,36 @@ func newProd(ctx context.Context, log *logrus.Entry, component ServiceComponent)
...
+		p.miseAuthorizer = miseadapter.NewAuthorizer(os.Getenv("MISE_ADDRESS"), p.log)
+	}
+
+	if !p.FeatureIsSet(FeatureEnforceMISE) {
+		if !p.FeatureIsSet(FeatureEnableDevelopmentAuthorizer) {
+			p.armClientAuthorizer = clientauthorizer.NewARM(p.log, p.Core)
+		} else {
+			armClientAuthorizer, err := clientauthorizer.NewSubjectNameAndIssuer(
+				p.log,
+				"/etc/aro-rp/arm-ca-bundle.pem",
```
geowa4, line 275: I know this predates your change, but this string should be a constant.

pkg/deploy/devconfig.go
wanghaoran1988: +1, could we make this a configuration variable?

pkg/deploy/generator/scripts/util-services.sh
sankur-codes: Replied above

pkg/deploy/generator/scripts/util-services.sh
sankur-codes: Replied above

pkg/deploy/generator/scripts/util-services.sh
sankur-codes: Replied above

pkg/deploy/generator/scripts/util-services.sh
sankur-codes: Replied above

pkg/deploy/generator/scripts/util-services.sh
sankur-codes: We could get rid of the initial assignment and directly reference the positional arguments like 

```
IPADDRESS='$4'
ROLE='$2'"
```

but that would not have much of an improvement over readability of the code.

pkg/deploy/devconfig.go
sankur-codes: Variable name `MiseValidAppIDs` signifies list of Azure AD App IDs which are authorised to send request for authentication to MISE. As of now the app id is of ARO-RP but it can contain any other app id for a service we want to authenticate for. So keeping the variable name in reference to MISE rather than to the services whose app id will be added as a valid app id.

pkg/deploy/generator/resources_rp.go
sankur-codes: These variables are not being input manually but rather are being fetched from the configurations in RP-Config repository.

pkg/frontend/ready_get.go
sankur-codes: We are rolling out MISE in two phases. This is the first phase with a fallback mechanism to ensure if there is any authentication failure via MISE, we fall back to TLS cert authentication for surety. In short, till we are not enforcing MISE, we will authenticate via TLS as well.

pkg/util/miseadapter/miseadapter.go
sankur-codes: Currently, our primary concern is determining whether a request is authenticated. Therefore, we are not including the header in the `input` or passing any actor claims in the request to MISE. Retrieving all actor claims would be inefficient, as it would result in unnecessarily large responses, impacting performance.  

Additionally, since granular authorization based on client roles or other claims is not a requirement at this stage, we are not fetching specific actor claims either. This code block serves as a placeholder for potential future implementations, should the need for more detailed authorization arise in later phases.

fahlmant:
I still seem some comments, mostly about refactoring things into constants or improving variable names. Is it possible to get those addressed before merging?

pkg/deploy/devconfig.go
wanghaoran1988: Could you add a comment in the code to describe this to make it more readable, where the id comes from, what it stands for .etc.

pkg/deploy/generator/scripts/rpVMSS.sh
```diff
@@ -121,6 +132,25 @@ main() {
...
+        ["rp_config"]="aro_rp_conf_file"
+        ["fluentbit"]="fluentbit_conf_file"
+        ["mdsd"]="mdsd_config_version"
+        ["static_ip_address"]="static_ip_addresses"
+    )
+
+    # shellcheck disable=SC2034
+    # use default podman network with range 10.88.0.0/16
+    local -rA static_ip_addresses=(
+        ["rp"]="10.88.0.2"
```
wanghaoran1988, line 147: Could you point me why we need static ip? Also I see the gateway and RP is using the same IP, is this expected?

sankur-codes: Static IP addresses are being assigned to containers to use that in the Geneva Monitor to identify the MISE endpoint in a generic way across all VMs in VMSS. 
As we are only monitoring MISE endpoints in this way and  Gateway & RP services are deployed in different VMSS, we have kept their IPs as same.

pkg/deploy/devconfig.go
sankur-codes: The id used here is a dummy id for dev. The actual list of IDs comes from our repository which maintains configurations for RP. However, I can include a comment describing what this variable stands for.

pkg/deploy/generator/scripts/rpVMSS.sh
wanghaoran1988: Thanks, could you also add comment in the code, so others can understand why's this?

pkg/frontend/middleware/authenticated.go
```diff
@@ -7,27 +7,55 @@ import (
...
+
+	EnableMISE  bool
+	EnforceMISE bool
+
 	AdminAuth clientauthorizer.ClientAuthorizer
 	ArmAuth   clientauthorizer.ClientAuthorizer
+	MiseAuth  miseadapter.MISEAdapter
 }
 
 func (a AuthMiddleware) Authenticate(h http.Handler) http.Handler {
```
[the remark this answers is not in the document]
sankur-codes: Unit tests added

pkg/deploy/generator/scripts/util-services.sh
s-fairchild: > but that would not have much of an improvement over readability of the code.  
```bash
    local -n image="$1"
    local -n role="$2"
    local -n conf_file="$3"
    local -n ipaddress="$4"
```
Right, the numbered variables are already assigned to named varaibles, I think it's overly complicated to re-name them again with a new set of variables, when `ipaddress` and `role` are never directly used.

My recommendation is to use the first assignment variables, there's no need to copy them again.

pkg/util/miseadapter/miseadapter.go
geowa4: Ok I see that returning all would be good and like that the code as it is now seeks to return an empty set of claims for efficiency.

geowa4 (commented):
Hey Ankur, went through this in our code review meeting. Left some comments for you.

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -84,26 +92,32 @@ WantedBy=multi-user.target
...
     local -r aro_rp_service_filename='/etc/systemd/system/aro-rp.service'
     # shellcheck disable=SC2034
-    local -r aro_rp_service_file="[Unit]
+    local -r aro_rp_service_file='[Unit]
 After=network-online.target
 Wants=network-online.target
 
 [Service]
-EnvironmentFile=${aro_rp_conf_filename}
+EnvironmentFile=/etc/sysconfig/aro-rp
```
s-fairchild, line 120: I understand why you made this change, as the file is wrapped in single quotes now to prevent expansion of those in the `EnvironmentFile`. I don't really like hard coding the value either though... it's already set in the constant which should be repeated. I don't have a good idea for fixing this at the moment though.

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -74,7 +82,7 @@ StartLimitInterval=0
 
 [Install]
 WantedBy=multi-user.target
-    "
+    '
```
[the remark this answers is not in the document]
s-fairchild: It is. This is hard to figure out, viewing it in the diff view, but the intent is to use environment variables from the systemd `EnvironmentFile`. The single quotes stop the script from trying to expand them (and failing, as they are unset).

pkg/deploy/generator/scripts/util-services.sh
s-fairchild: Thanks for explaining. My bad, I read this quickly and didn't realize they are being assigned to those values in the `EnvironmentFile`, not in the current shell context.

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -286,17 +315,232 @@ ExecStart=/usr/bin/podman run \
   -v /run/systemd/journal:/run/systemd/journal \
   -v /var/etw:/var/etw:z \
   -v /var/run/mdsd/asa:/var/run/mdsd/asa:z \
-  $image \
+  $RPIMAGE \
```
s-fairchild, line 318:
```suggestion
  ${RPIMAGE} \
```
Can we use some consistency with brace expansion in the same quote block?
Using braces is probably best, as it should evaluate to "" if it's unset (although I'm not aware of exactly how systemd handles unset variables).

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -286,17 +315,232 @@ ExecStart=/usr/bin/podman run \
...
+# 1) image - nameref, string; MISE container image
+# 2) ipaddress - nameref, string; static ip of podman network to be attached
+configure_service_aro_mise() {
+    local -n image="$1"
+    local -n ipaddress="$2"
+    log "starting"
+    log "Configuring aro-mise service"
+
+    LOGININSTANCE="https://login.microsoftonline.com"
+    if [[ $AZURECLOUDNAME == "AzureUSGovernment" ]]; then
```
s-fairchild, line 340: Can you add `AzureUSGovernment` as a constant (read only var `declare -r`) in `util-common.sh`?
We may want to check the cloud instance in other places, and it would be good to set it as a read only variable now.

pkg/deploy/generator/scripts/util-services.sh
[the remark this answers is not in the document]
s-fairchild: Can you add a comment above these systemd file contents explaining that the variables are to be expanded at systemd start time (by systemd, not this script)?

s-fairchild (changes requested):
Overall the bash script changes look good. Appologies for the intial confusion, the diff view in github hid the variable assignment for where the file contents were being created, leading to my confusion thinking that you were duplicating the variables.
As I saw someone else was also confused by the addition of single quotes, some comments are probably best. It is confusing to see variables that are supposed to expand at bootstrapping, and those that should expand later.

I made some small suggestions, and found one error.

Have you deploy a full service dev rp using this branch? If not, please do. It is very difficult to catch problems that break any of these services if they don't cause the `aro-rp` service to fail to start (that's what our health probe checks).

pkg/deploy/generator/scripts/util-services.sh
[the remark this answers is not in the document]
s-fairchild: > What's the output of systemctl status for these services?
The output is the expanded variable value. They get parsed by `systemd-exec`, which is reflected in `systemctl status`.

pkg/util/miseadapter/authorizer.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	result, err := m.client.ValidateRequest(ctx, i, m.log)
+	if err != nil {
+		return false, err
+	}
+
+	if result.StatusCode != http.StatusOK {
+		m.log.Errorf("mise authentication failed with %d: %s", result.StatusCode, result.ErrorDescription)
+		return false, nil
```
wanghaoran1988, line 50: Why we are return nil not the error?

pkg/util/miseadapter/authorizer.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	resp, respErr := m.client.httpClient.Do(req)
+	if respErr != nil {
+		m.log.Errorf("mise readyz failed, %s", err)
+		return false
+	}
+
+	defer resp.Body.Close()
+
+	return resp.StatusCode == http.StatusOK
```
wanghaoran1988, line 73: What if the status code is not 200? let's print out a log?

pkg/frontend/ready_get.go
wanghaoran1988: How about this:
```
func (f *frontend) checkReady() bool {
	if !f.env.FeatureIsSet(env.FeatureDisableReadinessDelay) &&
		time.Since(f.startTime) < 2*time.Minute {
		return false
	}

	_, okOcpVersions := f.lastOcpVersionsChangefeed.Load().(time.Time)
	_, okPlatformWorkloadIdentityRoleSets := f.lastPlatformWorkloadIdentityRoleSetsChangefeed.Load().(time.Time)

	authReady := f.checkAuthReady()

	return okOcpVersions && okPlatformWorkloadIdentityRoleSets &&
		f.ready.Load().(bool) &&
		authReady &&
		f.env.AdminClientAuthorizer().IsReady()
}

func (f *frontend) checkAuthReady() bool {
	var authReady bool

	if f.authMiddleware.EnableMISE && f.authMiddleware.EnforceMISE {
		authReady = f.env.MISEAuthorizer().IsReady()
	} else {
		authReady = f.env.ArmClientAuthorizer().IsReady()
	}

	return authReady
}
```

pkg/util/miseadapter/miseadapter.go
brizrobbo: With regards to "This code block serves as a placeholder for potential future implementations, should the need for more detailed authorization arise in later phases.", isn't that code currently unecessary? I think it is better to put that in when we actually do need it? We have be burned in the past where we have tried to future-proof and then it never got implemented (or the strategy changed slightly) and we were left with tech debt which we needed to clean-up. WDYT @geowa4 ?

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -84,26 +92,32 @@ WantedBy=multi-user.target
...
 configure_service_aro_rp() {
     local -n image="$1"
     local -n role="$2"
     local -n conf_file="$3"
+    local -n ipaddress="$4"
     log "starting"
     log "Configuring aro-rp service"
 
     local -r aro_rp_conf_filename='/etc/sysconfig/aro-rp'
+    local -r add_conf_file="PODMAN_NETWORK='podman'
```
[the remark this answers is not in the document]
sankur-codes: We have 3 env vars. The ending double quotes are on line 107.

pkg/frontend/ready_get.go
sankur-codes: @geowa4 @wanghaoran1988 Thanks for the suggestions. I understand that this might seem a bit confusing—we are checking the readiness of both MISE and ARM (TLS) endpoints and proceeding only if both are ready. However, exclusivity is necessary here.  

The reason is that if the MISE endpoint is ready but indicates that a request is not authenticated, we still log it for reference (with dashboards to monitor failure rates and logs for deeper insights) and proceed with authentication via TLS. This ensures that the request is genuinely invalid before taking further action.  

If we notice a mismatch—where a request fails authentication via MISE but succeeds via TLS—we can investigate the discrepancy without causing any issues for the customer. Therefore, ensuring both endpoints are ready is a critical first step in the first phase of MISE implementation.

pkg/deploy/generator/scripts/util-services.sh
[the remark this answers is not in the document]
sankur-codes: Thanks for the explanation @s-fairchild. Sure I can add a comment over these.

pkg/frontend/middleware/authenticated_test.go
```diff
@@ -0,0 +1,124 @@
...
+			name:         "Admin authentication failure",
+			apiVersion:   admin.APIVersion,
+			isAdminAuth:  false,
+			expectStatus: http.StatusForbidden,
+		},
+		{
+			name:         "ARM authentication success via MISE",
+			enableMISE:   true,
+			isMiseAuth:   true,
+			expectStatus: http.StatusOK,
```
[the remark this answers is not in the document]
sankur-codes: The nest two cases are exactly this where MISE is enabled but fails and we have TLS failing or passing(as TLS will be there always).

pkg/deploy/generator/scripts/util-services.sh
sankur-codes: The only way I see is to have more variables in the EnvironmentFile having values for the path of these `sysconfig` files for each service and address them in this service file. However, current approach seems to be less complicated than this. WDYT ?

pkg/deploy/generator/scripts/util-services.sh
sankur-codes: Thanks for highlighting this. If it is unset, system just keep it as it is and does not evaluate or expand it. So `${RPIMAGE}` remains `${RPIMAGE}`. For consistency, I have removed the curly braces for service files as that is the standard way

pkg/util/miseadapter/miseadapter.go
geowa4: Yeah, I appreciate the use of `[]string` to return an empty set of claims (hooray efficiency), but if there's a large time gap between now and when this block is used, the API could have changed and we will have a bug to find.

pkg/deploy/generator/scripts/util-services.sh
s-fairchild: I think this is fine, it's not worth the effort and complication to use the same file path variable.

pkg/deploy/generator/scripts/util-services.sh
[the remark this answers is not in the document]
s-fairchild: Ah, yes of course. Thanks.
The github diff makes it harder to see these things.

pkg/deploy/generator/scripts/util-services.sh
sankur-codes: Done. added this is constant and addressed here accordingly.

pkg/util/miseadapter/authorizer.go
sankur-codes: Two reasons. One is that we are already logging the error before returning and another is that we want this to go ahead for authentication via TLS method even if MISE authentication fails. We might investigate on the failure later and segregate it as valid or invalid.

pkg/util/miseadapter/miseadapter.go
sankur-codes: I see we can work on this by creating a JIRA ticket and mentioning about either utilizing this section of code in next phase while we Enforce MISE or get rid of it completely. Getting this code out includes removing it references i other files where this is being parsed from the response received from MISE. 
This way we have a record so this granular access control section doesn't get missed while we also make sure it is removed if not used later !
@geowa4 @brizrobbo WDYT ?

pkg/frontend/middleware/authenticated_test.go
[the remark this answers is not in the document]
geowa4: I'm thinking of the below. Is this not a valid case? 
```
	enableMISE:   true,
	isMiseAuth:   true,
	expectStatus: http.StatusForbidden,
```

pkg/frontend/ready_get.go
geowa4: I'm focusing on this code above (simplified for this comment field)
```
if f.authMiddleware.EnableMISE {...}
if !f.authMiddleware.EnforceMISE {...}
```
This is logically equivalent to the below, which is what I'm advocating. 
```
if f.authMiddleware.EnableMISE {...}
else {...}
```

pkg/util/miseadapter/miseadapter.go
geowa4: +1 If we have an impending action item I can get behind deferring the decision to keep/remove. For the future, keeping our PRs to only the minimally required paths reduces complexity, increases readability, and minimizes our surface area to attack, bugs, and generally getting stale.

pkg/deploy/generator/scripts/util-services.sh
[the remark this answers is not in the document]
geowa4: This didn't match my experience with systemd but it looks like it is supported. Nice.

>Basic environment variable substitution is supported. Use "${FOO}" as part of a word, or as a word of its own, on the command line, in which case it will be erased and replaced by the exact value of the environment variable (if any) including all whitespace it contains, always resulting in exactly a single argument. Use "$FOO" as a separate word on the command line, in which case it will be replaced by the value of the environment variable split at whitespace, resulting in zero or more arguments. For this type of expansion, quotes are respected when splitting into words, and afterwards removed.

https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#Command%20lines

geowa4 (commented):
The big things I commented before, like the systemd unit stuff (TIL), seem good. Thanks for the discussion. I left a couple small follow-up comments, but I don't see any blocking issues.

pkg/util/miseadapter/authorizer.go
```diff
@@ -0,0 +1,79 @@
...
+	IsAuthorized(ctx context.Context, r *http.Request) (bool, error)
+	IsReady() bool
+}
+
+type miseAdapter struct {
+	client *Client
+	log    *logrus.Entry
+}
+
+func NewAuthorizer(miseAddress string, log *logrus.Entry) MISEAdapter {
```
pepedocs, line 26: At first I thought this was authorizing something, but it looks like it's just authenticating using the credentials retrieved from the Authorization header. Correct me if I'm wrong here. 

The term authenticated and authorized have been loosely used here, it doesn't tell immediately whether it's actually authorizing something, therefore this can be confusing.

What is the reason for not simply calling this Authenticator?

pepedocs (changes requested):
Sorry for the late review and thanks for the additional changes previously requested. But I want to clarify the Authorizer part, should that be named Authenticator and use the authenticate term consistently? It did not seem to look like it's actually authorizing something, and this can be confusing as both terms are used loosely.

pkg/util/miseadapter/authorizer.go
sankur-codes: Here the main reason is that, I am simply following the ARM and Admin authorizer implemented before the mise one. Also, the capability of this authorizer is not limited to authentication but rather role based authorizing using actor claims as well which can be leveraged.

pkg/frontend/ready_get.go
geowa4: I misread Enforce and Enable MISE. My bad.

pkg/frontend/middleware/authenticated_test.go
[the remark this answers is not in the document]
geowa4: @sankur-codes and I chatted about this. I'm good with the test scenarios we have. `expectStatus` is what the mise component returns to its caller.

pkg/util/miseadapter/authorizer.go
pepedocs: I understand the prior work here that makes this a little bit confusing.

@sankur-codes I think we should create a Jira for this and iterate later. 

As a good practice, the code should make it clear whether it is authorizing or authenticating only or both. It's a good practice because it enhances readability, it tells developers where and what the code is actually doing. For example, if I want to find and understand the authorization methods of this service, then I can find it easily and focus on it. In this Authorizer's case, if not careful enough the fact that it's not authorizing and only authentication can be missed easily.

pepedocs (dismissed):
Hi @sankur-codes /LGTM

Could you please create a Jira for iterating the Authorizer code, making it more clearer and separate concerns with respect to authz and authn?

wanghaoran1988 (approved):

s-fairchild (dismissed):
LGTM.  
Please rebase, squashing your fix commits. If it's all good to go into one commit, allowing Github to do this when merging also works.

pkg/deploy/devconfig.go
```diff
@@ -189,6 +198,7 @@ func DevConfig(_env env.Core) (*Config, error) {
 				"EnableOCMEndpoints",
 				"RequireOIDCStorageWebEndpoint",
 				"UseMockMsiRp",
+				"EnableMISE",
```
tsatam, line 201: Do we actually want this feature on for full-service RPs (where this devconfig gets used)? 

If it works fine in that context, I say keep it in, but if it doesn't work at all or causes other issues we may want to disable this for full-service.

pkg/env/zz_generated_feature_enumer.go
```diff
@@ -7,11 +7,11 @@ import (
 	"strings"
 )
 
-const _FeatureName = "FeatureDisableDenyAssignmentsFeatureDisableSignedCertificatesFeatureEnableDevelopmentAuthorizerFeatureRequireD2sV3WorkersFeatureDisableReadinessDelayFeatureEnableOCMEndpointsFeatureRequireOIDCStorageWebEndpointFeatureUseMockMsiRp"
+const _FeatureName = "FeatureDisableDenyAssignmentsFeatureDisableSignedCertificatesFeatureEnableDevelopmentAuthorizerFeatureRequireD2sV3WorkersFeatureDisableReadinessDelayFeatureEnableOCMEndpointsFeatureRequireOIDCStorageWebEndpointFeatureUseMockMsiRpFeatureEnableMISEFeatureEnforceMISE"
```
tsatam, line 10: I may have caused a required rebase here by merging another PR that changed RP feature flags, apologies.

pkg/util/miseadapter/authorizer.go
tsatam: +1 to handling this change later.

pkg/env/zz_generated_feature_enumer.go
sankur-codes: I have done the rebase but now lost all the approvals. Will follow up on these.

pepedocs (approved):

pkg/deploy/devconfig.go
sankur-codes: For full-service RP there is no involvement of ARM so MISE will not be needed here. However, we have added test detail for the config to pass validation. Ideally this should work, but I am yet to create a full-service RP setup to test this practically.

s-fairchild:
I placed the hold label on this for now.  
We need to create the branch for RP Release 268 before this is merged.
This should be merged mid next week.

pkg/deploy/devconfig.go
sankur-codes: @tsatam While creating this, the VMs had MISE and OTEL coming up as expected but rp.portal were failing due other unrelated issues. I have removed the feature flag and added instruction for mirroring MISE image from int for letting full-rp-service setup work as it is for now.

s-fairchild (approved):
LGTM. The release branch for these changes must be based on https://github.com/Azure/ARO-RP/releases/tag/v20250325.00.

tsatam (approved):

PR #4071, opened 2025-01-30
ARO-14652: Allow the ability to specify custom master/worker VM sizes in release E2E pipeline

Tof1973:

Fixes https://issues.redhat.com/browse/ARO-14652

Allow user to specify custom master and/or worker Vm Size instead of default one when creating a cluster through E2E pipeline or hack script.

SOP for ```Testing new instance types``` (docs/adding-new-instance-types.md) as ben updated to use new variables.

I have created cluster following the SOP changes, and we need to run the E2E pipeline with it.


Tof1973:
E2E pipeline has been run to test with those new parameters [here](https://dev.azure.com/msazure/AzureRedHatOpenShift/_build/results?buildId=113872762&view=logs&j=9729bfe2-2fc6-5a01-2383-649af09dff52&t=9bb48e38-c2fa-5fd7-dd36-0f00d672daf6&l=115)
During the test we have seen:
`$ az aro show -n v4-e2e-V113872762-westcentralus-prod -g v4-e2e-V113872762-westcentralus-prod | grep -i vmsize
    "vmSize": "Standard_D8s_v4"
      "vmSize": "Standard_D4s_v4"`

ARO-Pipeline changes are implemented in [here](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/ARO-Pipelines/pullrequest/11658739)

pkg/util/version/const.go
```diff
@@ -26,16 +26,20 @@ const (
 var GitCommit = "unknown"
 
 type Stream struct {
-	Version  *Version `json:"version"`
-	PullSpec string   `json:"-"`
+	Version      *Version   `json:"version"`
+	PullSpec     string     `json:"-"`
+	MasterVmSize api.VMSize `json:"masterVmSize"`
```
tsatam, line 31: I think we're better off splitting the VM sizes out of this Stream struct, and leaving the struct focused on OCP version information. We can just make the "default" values static within test/e2e/setup.go and hack/cluster/cluster.go, or have them both reference a const value elsewhere.

tsatam (changes requested):

pkg/util/version/const.go
Tof1973: moves default size to util.cluster

tsatam (approved):

pkg/util/cluster/cluster.go
```diff
@@ -579,8 +584,10 @@ func (c *Cluster) createCluster(ctx context.Context, vnetResourceGroup, clusterN
 		if err != nil {
 			return err
 		}
-
-		oc.Properties.WorkerProfiles[0].VMSize = api.VMSizeStandardD2sV3
+		// In LocalDev mode, if workerVmSize is not default one, then it means user requested a specific one we need to keep.
```
kimorris27, line 587: Nit: I'd rework this comment to explain what the conditional is doing rather than explaining what it's _not_ doing.

kimorris27 (changes requested):
Kind of a nitpicky request, but I suggested reworking a comment because I'm having trouble understanding what the code directly below it is doing.

pkg/util/cluster/cluster.go
kimorris27: I have a more concrete suggestion because it clicked after I walked away for a bit:

```suggestion
		// If we're in local dev mode and the user has not overridden the default VM size, use a smaller size for cost-saving purposes
```

kimorris27 (approved):

kimorris27 (approved):

Tof1973:
Test re-done with latest changes: https://dev.azure.com/msazure/AzureRedHatOpenShift/_build/results?buildId=114152007&view=logs&j=9729bfe2-2fc6-5a01-2383-649af09dff52&t=9bb48e38-c2fa-5fd7-dd36-0f00d672daf6
And we see custom VMs created

Tof1973:
Test re-done with latest changes: https://dev.azure.com/msazure/AzureRedHatOpenShift/_build/results?buildId=114152007&view=logs&j=9729bfe2-2fc6-5a01-2383-649af09dff52&t=9bb48e38-c2fa-5fd7-dd36-0f00d672daf6
And we see custom VMs created:
```
$ az aro show -n v4-e2e-V114152007-westcentralus-prod -g v4-e2e-V114152007-westcentralus-prod | jq .masterProfile.vmSize,.workerProfiles[].vmSize
"Standard_D8s_v4"
"Standard_D4s_v4"
```

pkg/util/cluster/cluster.go
```diff
@@ -69,6 +69,8 @@ type Cluster struct {
 
 const GenerateSubnetMaxTries = 100
 const localDefaultURL string = "https://localhost:8443"
+const DefaultMasterVmSize = api.VMSizeStandardD8sV3
+const DefaultWorkerVmSize = api.VMSizeStandardD4sV3
```
tiguelu, line 73: In the original code, the VM sizes were not specified in the openshift doc and the SKU of  the VMs relied on whichever mechanism the RP has to determine it (not sure where).

By adding these definitions in this file and always setting the field explicitly in the doc , aren't we duplicating logic to define the default VMs?

Tof1973: AFAIK, the default was already defined in that file in the createCluster function, so I guess I'm just moving the logic in const, don't I ? Not sure which openshift doc your are referencing...

tiguelu: Sorry, I read the diff too quickly and thought there was no previous specification of size, but there was. Thanks for the clarification.

Also, for the record, we checked the code and confirmed that this is enforced in the [client code](https://github.com/Azure/ARO-RP/blob/tof1973/ARO-14652/python/az/aro/azext_aro/custom.py#L177) not the RP, which makes sense. So, the definitions for defaults need to be duplicated in the dev/e2e cluster setup logic because this provisioning method doesn't use the az client.

tiguelu (approved):

kimorris27 (approved):

PR #4072, opened 2025-01-30
Bother bennerv instead of mjudeikis when failing codeql analysis cron

### LINKED ISSUE #4051: Blast from the past :D
Image

The analyze job failed, and I somehow keep getting notified :D either fix the job or remove me please :)

p.s Hey!

bennerv:

Fixes #4051

Apparently the last person to modify the `cron` on a scheduled job gets [notified about failures](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule).  So instead of bothering someone else who no longer is here, we'll bother myself :old-man-yells-at-ben: 🤷 

### Test plan for issue:

Watch the thing notify me when it fails

nope

N/a

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -8,7 +8,7 @@ on:
       - master
   pull_request:
   schedule:
-    - cron: '20 0 * * 6'
+    - cron: '30 0 * * 6'
 
 permissions:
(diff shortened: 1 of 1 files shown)


tsatam (approved):
LGTM :old-man-yells-at-ben:

PR #4074, opened 2025-01-31
Enable $web and disable anonymous blob access on rpversions container

tsatam:

Fixes [ARO-14928](https://issues.redhat.com/browse/ARO-14928)

Updates our insertion of the deployed RP version into the rpversion storageaccount to use the static website hosting feature's `$web` container. This will allow us to disable anonymous blob access on the storage account generally, while still allowing access to content within the `$web` container. We enable static web hosting on the storage account before writing the version file to the container, as it is not enabled by default, and depends on a few configuration parameters that were not originally set on these SAs. 

In order to do this, we also force an upgrade of the storage account to StorageV2 if necessary, as V2 is required for serving static content. We also update the minimum TLS version on the SA to 1.2. These changes are made to the applied ARM template. 

We also replace our blobsClient used in this process with the newer azure-sdk-for-go's azblob client, allowing us to drop the older vendored blobclient entirely (this was its last usage in the RP codebase). 

This will require corresponding changes to our consumption of this content, as the URL will change from a `.blob.core.windows.net` URL to `.web.core.windows.net`. 

### Test plan for issue:

- [x] Deploy this change in a non-production environment and ensure it:
  - [x] Can be deployed over the existing configuration successfully
  - [x] Sets the Storage Account Kind to `StorageV2`
  - [x] Sets the minimum TLS version to 1.2
  - [x] Enables static web hosting on the SA
  - [x] Disables anonymous blob access on the SA
  - [x] Inserts the version information successfully into the `$web` container
  - [x] The version information within the `$web` container can be accessed anonymously. 

The entire `saveversion` function could benefit from some dedicated unit tests, I will try to add these in a follow-up effort. 

No

Above testing


rajdeepc2792:
Changes lgtm, haven't tested by running a local deploy.

tsatam:
Tested this deployment change in an int deployment: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=114252528&view=results.

pkg/deploy/generator/resources_rp.go
```diff
@@ -1518,36 +1518,13 @@ func (g *generator) rpACRRBAC() []*arm.Resource {
...
-				},
-			},
-			APIVersion: azureclient.APIVersion("Microsoft.Storage"),
-			DependsOn: []string{
-				"[resourceId('Microsoft.Storage/storageAccounts', parameters('rpVersionStorageAccountName'))]",
-			},
-		},
-		{
-			Resource: &mgmtstorage.BlobContainer{
-				Name: to.StringPtr("[concat(parameters('rpVersionStorageAccountName'), '/default/ocpversions')]"),
```
bennerv, line 1541: I'm fairly confident we're not using OCPVersions at all, but can you confirm?  I'm pretty certain they're stale.

pkg/deploy/saveversion.go
```diff
@@ -4,25 +4,27 @@ package deploy
...
+	"fmt"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
 	mgmtstorage "github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2021-09-01/storage"
-	azstorage "github.com/Azure/azure-sdk-for-go/storage"
 	"github.com/Azure/go-autorest/autorest/date"
+
+	"github.com/Azure/ARO-RP/pkg/util/azureclient/azuresdk/azblob"
+	"github.com/Azure/ARO-RP/pkg/util/pointerutils"
```
bennerv, line 16: nit: this package is duplicated.  We now have `k8s.io/utils/ptr` which performs the exact same functionality.

jhoreman:
I think we have some doc updates to go along with this. Will reach out on Slack.

pkg/deploy/generator/resources_rp.go
bennerv: For context/history (if i'm remembering correctly), we used this back when we had a single version install target and didn't support multiple versions.

tsatam: I checked our prod container and the newest OCP version it had was a 4.11.z version, so I hope we still aren't using it for anything

pkg/deploy/saveversion.go
tsatam: We had an internal discussion on this a while back (pinging you in the thread) - between K8S's, Autorest's, and our own, we decided to stick to ours since its implementation is literally just `func ToPtr[T any](t T) *T { return &t }` and doesn't tightly couple us to any 3rd party deps for the functionality.

pkg/deploy/generator/resources_rp.go
tsatam: It's also worth noting that the absence of these blobcontainer resources doesn't result in existing containers being deleted, they're just left alone (verified in int). If we do want to delete these resources we'll likely need to add an explicit fixup step during deployment.

hlipsig (approved):
My only other thought is regarding the minimum TLS version. Since we must support TLSv3, is there any reason the changes here won't work with v3 as well? I'm assuming not since all Azure services are supporting v3.

pkg/deploy/assets/rp-production-global.json
```diff
@@ -101,40 +101,16 @@
             "sku": {
                 "name": "Standard_LRS"
             },
+            "kind": "StorageV2",
```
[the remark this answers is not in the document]
tsatam: I think the required code change should be fairly self-explanatory, and would also be dependent on whatever the reason for changing/extending this would be. I also don't think there's actually any compelling reason for us to need any kind other than `StorageV2` in the future, but unforeseen circumstances might require that. 

Regarding TLS, this property is just setting the "minimum" supported TLS version, the highest we can even set this to is 1.2 currently anyways (https://learn.microsoft.com/en-us/azure/storage/common/transport-layer-security-configure-minimum-version?tabs=azure-cli#configure-the-minimum-tls-version-for-a-storage-account).

jhoreman (approved):

PR #4083, opened 2025-02-05
Makefile: add openshift-goimports, remove old verify

stevekuznetsov:
Makefile: add openshift-goimports, remove old verify

The openshift-goimports formatter is superior to the old strategy of
verifying import groups since:

- it is more flexible, allowing configuration flags to be passed that
  specify which groups need to exist and what order they should be in
- it has an automatic formatter built in, so no humans will ever have to
  make manual changes to fix things
- it runs in microseconds or milliseconds, so adding it to a pre-commit
  hook has little to no downside

---

*: format imports

---


cmd/aro/operator.go
```diff
@@ -8,13 +8,15 @@ import (
...
 
-	machinev1 "github.com/openshift/api/machine/v1"
 	"github.com/sirupsen/logrus"
+
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 
+	machinev1 "github.com/openshift/api/machine/v1"
```
stevekuznetsov, line 18: By default it has broken these out into separate import groups since they are under a different org. We can configure the tool not to, but before I spent time on that I figured we might as well see if folks are ok with this approach. Everything is fully automated end-to-end so it's low-touch from here on out.

hawkowl: If this behaviour is OpenShift's standard, then I think that's a good reason to keep the defaults.

kimorris27 (approved):

PR #4084, opened 2025-02-05
fix: add logic to skip subnet update under certain conditions

scottd018:

Fixes https://issues.redhat.com/browse/ARO-15009

This commit addresses the need for the subnet/write permissions.  Currently, in all scenarios, the subnet/write permission is used via the subnet.CreateOrUpdate method.  This change adds logic to determine if the values are already set to Disabled and skips running the subnet.CreateOrUpdate method if it is already set appropriately.  This will save the need to make an extra API call and also allow users to configure things like Azure policy where subnet/write permissions are not needed.

### Test plan for issue:

Currently awaiting access for the Dev ARO-RP shared environment.  Once I get access I can run a full E2E test and comment with the results.

No documentation needs to be updated, as the end user experience looks identical.

No change to the existing logic is introduced.  We are only introducing a skip of the API call to create or update a subnet if it does not currently match a desired stated of "Disabled".

### DIFF
--- a/pkg/cluster/deploybaseresources.go
+++ b/pkg/cluster/deploybaseresources.go
@@ -432,10 +432,27 @@ func (m *manager) setMasterSubnetPolicies(ctx context.Context) error {
 		s.SubnetPropertiesFormat = &mgmtnetwork.SubnetPropertiesFormat{}
 	}
 
+	// we need to track whether or not we need to send an update to the AzureRM API based on whether
+	// or not our private endpoint network policies or private link service network policies
+	// already match a desired condition of 'Disabled' or not.
+	var needsUpdate bool
+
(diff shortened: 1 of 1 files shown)


hlipsig:
I've kicked off CI/e2e. Looking to get a few more reviews and go over this on the weekly review meeting next week, but tentatively LGTM.

kimorris27:
@scottd018 , these changes LGTM, but you'll need to close your current PR and reopen it from a branch off the trunk repo rather than your fork. We can't run our CI Azure pipelines against PRs from forks. If you don't have permission to push a branch to the trunk repo, an ARO team lead should be able to help.

scottd018:
Closing in favor of https://github.com/Azure/ARO-RP/pull/4087 so that CI may properly run

PR #4085, opened 2025-02-05
Update all to.Ptr equivalents to our internal one

bennerv:

Consistent usage of a library rather than using 3 different ones. 

The team agreed that they want to use the library we've written prior to the k8s generic bumps and what not.  If we start using functionality that's present in k8s.io/utils/ptr or azcore, then we will pivot.  

https://redhat-internal.slack.com/archives/C02ULBRS68M/p1738697814649279?thread_ts=1731940533.915429&cid=C02ULBRS68M

### Test plan for issue:

e2e & CI


tsatam (approved):

bennerv:
e2e failed check is the old one. Going to merge as a result.

PR #4087, opened 2025-02-06
[ARO-15009] fix: add logic to skip subnet update under certain conditions

scottd018:

Fixes https://issues.redhat.com/browse/ARO-15009

This commit addresses the need for the subnet/write permissions.  Currently, in all scenarios, the subnet/write permission is used via the subnet.CreateOrUpdate method.  This change adds logic to determine if the values are already set to Disabled and skips running the subnet.CreateOrUpdate method if it is already set appropriately.  This will save the need to make an extra API call and also allow users to configure things like Azure policy where subnet/write permissions are not needed.

Closed https://github.com/Azure/ARO-RP/pull/4084 in favor of this PR.

### Test plan for issue:

Currently awaiting access for the Dev ARO-RP shared environment.  Once I get access I can run a full E2E test and comment with the results.

No documentation needs to be updated, as the end user experience looks identical.

No change to the existing logic is introduced.  We are only introducing a skip of the API call to create or update a subnet if it does not currently match a desired stated of "Disabled".

### DIFF
--- a/pkg/cluster/deploybaseresources_test.go
+++ b/pkg/cluster/deploybaseresources_test.go
@@ -501,7 +501,18 @@ func TestSetMasterSubnetPolicies(t *testing.T) {
 			gatewayEnabled: true,
 		},
 		{
-			name: "ok, skipCreateOrUpdate",
+			name: "ok, skipCreateOrUpdate, !gatewayEnabled",
+			mocks: func(subnet *mock_subnet.MockManager) {
+				subnet.EXPECT().Get(ctx, "subnetID").Return(&mgmtnetwork.Subnet{
+					SubnetPropertiesFormat: &mgmtnetwork.SubnetPropertiesFormat{
+						PrivateLinkServiceNetworkPolicies: to.StringPtr("Disabled"),
+					},
+				}, nil)
+				subnet.EXPECT().CreateOrUpdate(gomock.Any(), gomock.Any(), gomock.Any()).Times(0)
+			},
+		},
+		{
+			name: "ok, skipCreateOrUpdate, gatewayEnabled",
 			mocks: func(subnet *mock_subnet.MockManager) {
 				subnet.EXPECT().Get(ctx, "subnetID").Return(&mgmtnetwork.Subnet{
 					SubnetPropertiesFormat: &mgmtnetwork.SubnetPropertiesFormat{


kimorris27 (changes requested):
The change LGTM, but one thing I missed from my original review - you could add a unit test case validating that `CreateOrUpdate` is not called when those two attributes both already have the correct values.

scottd018:
> The change LGTM, but one thing I missed from my original review - you could add a unit test case validating that `CreateOrUpdate` is not called when those two attributes both already have the correct values.

Absolutely can do that.  I'll get a new commit added and pushed.  I thought about doing it originally, but hadn't seen any test cases for the specific method (maybe I missed them...let me know if there's some specific place you want them otherwise I can add it to the `_test.go` file like I'm used to doing).

kimorris27:
> > The change LGTM, but one thing I missed from my original review - you could add a unit test case validating that `CreateOrUpdate` is not called when those two attributes both already have the correct values.
> 
> Absolutely can do that. I'll get a new commit added and pushed. I thought about doing it originally, but hadn't seen any test cases for the specific method (maybe I missed them...let me know if there's some specific place you want them otherwise I can add it to the `_test.go` file like I'm used to doing).

The unit tests for the function you tweaked are here: https://github.com/Azure/ARO-RP/blob/fb723cfe08127192ca43ef2fac0e9daa02bc51b0/pkg/cluster/deploybaseresources_test.go#L470

scottd018:
> > > The change LGTM, but one thing I missed from my original review - you could add a unit test case validating that `CreateOrUpdate` is not called when those two attributes both already have the correct values.
> > 
> > 
> > Absolutely can do that. I'll get a new commit added and pushed. I thought about doing it originally, but hadn't seen any test cases for the specific method (maybe I missed them...let me know if there's some specific place you want them otherwise I can add it to the `_test.go` file like I'm used to doing).
> 
> The unit tests for the function you tweaked are here:
> 
> https://github.com/Azure/ARO-RP/blob/fb723cfe08127192ca43ef2fac0e9daa02bc51b0/pkg/cluster/deploybaseresources_test.go#L470

@kimorris27 Take a look at that.  I added a couple of tests that seem to align with the others that currently exist.  Let me know if you need any other changes.  Thanks!

kimorris27 (approved):

PR #4089, opened 2025-02-07
Add RHCOS image mirroring script back

hawkowl:
This moves the RHCOS script from the wrapper back to ARO-RP, and makes it able to do arbitrary versions.


hawkowl:
Since HCP is mirroring these automatically now to the same account, we don't actually need this.

PR #4090, opened 2025-02-07
bump golang-ci-lint

stevekuznetsov:
Makefile: bump golangci-lint

---

test/e2e: fixup bad Ginkgo assertions

ginkgo-linter: wrong error assertion. Consider using `Eventually(func(ctx context.Context) error {
	return project.VerifyProjectIsReady(ctx)
}).WithContext(ctx).WithTimeout(DefaultEventuallyTimeout).Should(Succeed())` instead (ginkgolinter)

---

pkg/api: remove unnecessary nil checks

S1009: should omit nil check; len() for []string is defined as zero (gosimple)

---

pkg/util: fix ineffectual compiler directive

SA9009: ineffectual compiler directive due to extraneous space:
"//  go:generate mockgen -destination=../mocks/$GOPACKAGE/client.go
github.com/Azure/msi-dataplane/pkg/dataplane Client" (staticcheck)

---

*: fix non-constant format strings

printf: non-constant format string in call to (*github.com/sirupsen/logrus.Entry).Warnf (govet)

---

cmd/migrate: add a program to refactor NewCloudError

---

*: run the program to refactor NewCloudError

---

pkg/api: refactor NewCloudError to take a static message

---

Revert "cmd/migrate: add a program to refactor NewCloudError"

This reverts commit 8aa74d1f5d1f006d840e025bf025ce728cced0c4.

---


pkg/api/error.go
```diff
@@ -113,11 +113,17 @@ const (
 
 // NewCloudError returns a new CloudError
 func NewCloudError(statusCode int, code, target, message string, a ...interface{}) *CloudError {
+	var msg string
```
stevekuznetsov, line 116: I reviewed some of the thousands of errors that were getting flagged here, and they were all cases where only one argument was passed and it was expected to be interpreted verbatim.

pkg/operator/controllers/checkers/internetchecker/checker.go
```diff
@@ -66,7 +66,7 @@ func (r *checker) Check(URLs []string) error {
 	}
 	if len(errsAll) != 0 {
 		// TODO: Consider replacing with multi error wrapping with Go 1.20: https://github.com/golang/go/issues/53435#issuecomment-1320343377
-		return fmt.Errorf(strings.Join(errsAll, "\n"))
+		return fmt.Errorf("%s", strings.Join(errsAll, "\n"))
```
tsatam, line 69: We should fix this with the recommendation in the TODO, but that can be done in a follow-up.

pkg/operator/controllers/subnets/subnets_controller.go
```diff
@@ -144,7 +144,7 @@ func (r *reconcileManager) reconcileSubnets(ctx context.Context) error {
 	}
 
 	if len(combinedErrors) > 0 {
-		return fmt.Errorf(strings.Join(combinedErrors, "\n"))
+		return fmt.Errorf("%s", strings.Join(combinedErrors, "\n"))
```
tsatam, line 147: We should replace this with errors.Join in a follow-up effort.

tsatam (approved):
LGTM, merge on green CI

pkg/api/error.go
weinong: what did linter complain?

weinong (dismissed):
LGTM

stevekuznetsov:
We need to write a Go AST-based refactoring tool to actually change the call-sites for `NewCloudError` - I'll do this next week.

pkg/api/error.go
stevekuznetsov: @weinong 

```
printf: non-constant format string in call to (*github.com/sirupsen/logrus.Entry).Warnf (govet)
```

When the first argument to `fmt.Sprintf` is not a constant string, if that value ends up having e.g. `%` literals in it, the output of `fmt.Sprintf()` will be unknown. Please see the new commits I pushed for how i solved this problem

tsatam (dismissed):
LGTM, though I'll be waiting on green checks to fully confirm.

pkg/operator/controllers/checkers/internetchecker/checker.go
stevekuznetsov: Tracked in https://issues.redhat.com/browse/ARO-15179

tsatam (approved):

PR #4091, opened 2025-02-10
[ARO-9308] Add MDSD Certificate rotation tasks to MIMO

hawkowl:

Fixes [ARO-9308](https://issues.redhat.com/browse/ARO-9308)

Adds MDSD certificate rotation to MIMO. Rewrites the code to use ClientHelper. Common underlying implementation between MIMO and AdminUpdate is used, and tests are added for the MIMO usage.

### Test plan for issue:

Unit tests + E2E

N/A

Unit tests :)


pkg/cluster/mdsd.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/Azure/ARO-RP/pkg/env"
+	pkgoperator "github.com/Azure/ARO-RP/pkg/operator"
+	"github.com/Azure/ARO-RP/pkg/util/clienthelper"
+	utilpem "github.com/Azure/ARO-RP/pkg/util/pem"
+)
+
+// "Rotate" the MDSD certificates in the cluster. The copy that is rotated from
+// comes from RP-Config, and is only able to be updated when the RP is updated.
+func RenewMDSDCertificate(ctx context.Context, log *logrus.Entry, _env env.Interface, ch clienthelper.Interface) error {
```
kimorris27, line 23: Not a request for change, just a question for my understanding: what was the motivation to refactor this to use the `clienthelper`?

kimorris27 (commented):
Looks good! In the test plan you mentioned E2E. Should we expect an E2E test for this in a follow-up PR?

pkg/cluster/mdsd.go
hawkowl: Consistency with the rest of MIMO + better testability (imo).

hawkowl:
@kimorris27 I'm not sure that I can write an e2e test for this specific case, since it require the env to change (which is a bit hard, although I could probably come up with a way to do it). I'm mostly relying on the e2e to test the codepath works, since I'm fairly sure we do an admin update and that will run it.

kimorris27:
> @kimorris27 I'm not sure that I can write an e2e test for this specific case, since it require the env to change (which is a bit hard, although I could probably come up with a way to do it). I'm mostly relying on the e2e to test the codepath works, since I'm fairly sure we do an admin update and that will run it.

Oh, gotcha... I was not thinking about the fact that E2E runs an admin update. Makes sense.

kimorris27 (approved):

PR #4096, opened 2025-02-12
[ARO-15124] Merge the implementations of dynamichelper and clienthelper's Merge function

hawkowl:

Part of ARO-15124

Reduce a bunch of code duplication. I'm not sure why I didn't do this in the first place.

### Test plan for issue:

E2E and unit tests should cover it.

N/A

Should be a functional no-op, so the E2E should demonstrate that.


kimorris27:
E2E (probably) flaked with a Hive installation failure. Rerunning it.

kimorris27 (approved):

PR #4097, opened 2025-02-12
Send Otel Audit Message 

gouthamMN:

Fixes: https://issues.redhat.com/browse/ARO-14985
As per SFI SM2.3.1 Action item RP code need to be instrument with OTel Audit.

### Test plan for issue:
- unit tests
- Validate that the logs are showing in Jarvis as mentioned [here](https://eng.ms/docs/products/geneva/collect/instrument/opentelemetryaudit/dotnet/validation)

None

By successfully testing in Canary


pkg/portal/middleware/aad_test.go
```diff
@@ -66,7 +66,7 @@ func (o *noopOauther) Exchange(context.Context, string, ...oauth2.AuthCodeOption
 }
 
 func TestNewAAD(t *testing.T) {
-	_, err := NewAAD(nil, nil, nil, nil, "", nil, "", nil, nil, nil, nil, nil)
+	_, err := NewAAD(nil, nil, nil, nil, nil, "", nil, "", nil, nil, nil, nil, nil)
```
stevekuznetsov, line 69: Somewhere between the 12th and 13th `nil` it is likely best to refactor this to take a `struct` or functional options, if these are actually optional.

gouthamMN: I'm not sure what's the reason behind why the way it's written that way. The refactor of this need to be handled under a separate JIRA ticket.

pkg/frontend/middleware/log.go
```diff
@@ -153,12 +158,19 @@ func (l LogMiddleware) Log(h http.Handler) http.Handler {
...
+				otelAuditMsg.Record.OperationResultDescription = fmt.Sprintf("Status code: %d", statusCode)
 			}
 
 			if r.URL.Path == "/healthz/ready" {
 				return
 			}
 
+			audit.Validate(&otelAuditMsg.Record)
+			if err := l.OutelAuditClient.Send(r.Context(), otelAuditMsg); err != nil {
+				log.Printf("Frontend - Error sending audit message: %v", err)
```
stevekuznetsov, line 171: Why `printf` - can we error or warn?

pkg/portal/middleware/log.go
```diff
@@ -115,6 +118,13 @@ func Log(env env.Core, auditLog, baseLog *logrus.Entry) func(http.Handler) http.
 				resultType := audit.ResultTypeSuccess
 				if statusCode >= http.StatusBadRequest {
 					resultType = audit.ResultTypeFail
+					otelAuditMsg.Record.OperationResult = msgs.Failure
```
stevekuznetsov, line 121: This seems duplicated from `frontend/middleware/log.go` - why? Can we DRY it out? Would we ever want these two to differ?

pkg/util/log/audit/otel_audit.go
```diff
@@ -0,0 +1,186 @@
...
+	auditLogQueueSize = 4000
+	Unknown           = "Unknown"
+)
+
+type Client interface {
+	Send(ctx context.Context, msg msgs.Msg) error
+}
+
+type Audit struct {
+	Client *audit.Client
```
stevekuznetsov, line 28: Why do we need `Audit` if it just embed a client? Why not use a client itself?

pkg/util/log/audit/otel_audit.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &Audit{
+		Client: client,
+	}, nil
+}
+
+func (a *Audit) Send(ctx context.Context, msg msgs.Msg) error {
```
stevekuznetsov, line 57: If `Audit` embeds `Send`, this method is superfluous

pkg/util/log/audit/otel_audit.go
```diff
@@ -0,0 +1,186 @@
...
+	"strings"
+
+	"github.com/microsoft/go-otel-audit/audit"
+	"github.com/microsoft/go-otel-audit/audit/base"
+	"github.com/microsoft/go-otel-audit/audit/conn"
+	"github.com/microsoft/go-otel-audit/audit/msgs"
+)
+
+const (
+	auditLogQueueSize = 4000
```
stevekuznetsov, line 19: Where does this value come from? What is the memory footprint implication? Why isn't this tunable?

pkg/util/log/audit/otel_audit.go
```diff
@@ -0,0 +1,186 @@
...
+		return conn.NewNoOP(), nil
+	}
+
+	client, err := audit.New(newNoOpConn)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Audit{
+		Client: client,
```
stevekuznetsov, line 80: What's the point of bothering to create this client if we overwrite `Send` to do nothing?

pkg/util/log/audit/otel_audit.go
```diff
@@ -0,0 +1,186 @@
...
+	}, nil
+}
+
+func (a *MockAudit) Send(ctx context.Context, msg msgs.Msg) error {
+	return nil
+}
+
+func GetOperationType(method string) msgs.OperationType {
```
stevekuznetsov, line 88: Does this need to be exported?

pkg/util/log/audit/otel_audit.go
```diff
@@ -0,0 +1,186 @@
...
+
+		for _, resource := range resources {
+			if err := resource.Validate(); err != nil {
+				resource.Name = Unknown
+			}
+		}
+	}
+}
+
+func isTestEnv() bool {
```
stevekuznetsov, line 180: Tests should explicitly create a mock when they need one, adding this to the production codepath is a bad idea.

pkg/util/log/audit/otel_audit.go
```diff
@@ -0,0 +1,186 @@
...
+	return flag.Lookup("test.v") != nil
+}
+
+func isLocalDevelopmentMode() bool {
```
stevekuznetsov, line 184: Why not pass this in as a parameter or otherwise make it obvious from `_env`? Implicit and un-traceable dependencies on env vars without proper parameter passing through the call stack make for difficult maintenance.

test/util/log/audit.go
```diff
@@ -24,6 +24,10 @@ func NewAudit() (*test.Hook, *logrus.Entry) {
 	return h, logrus.NewEntry(logger)
 }
 
+func NewOtelAuditClient() (audit.Client, error) {
```
stevekuznetsov, line 27: Why does this signature 
a) return an error
b) do anything other than return the mock

stevekuznetsov: (this is the appropriate place to write a mock implementation of the interface to be used during testing)

pkg/util/log/audit/otel_audit.go
gouthamMN: `send()` function under `Client` interface is overridden by `MockAudit` for testing purposes.

pkg/util/log/audit/otel_audit.go
gouthamMN: `send()` function is overridden by `MockAudit` for testing purposes.

pkg/util/log/audit/otel_audit.go
gouthamMN: This defines the Max buffer size of 4KB and for audit log purpose this should be more than enough.

pkg/util/log/audit/otel_audit.go
gouthamMN: For testing purpose

pkg/util/log/audit/otel_audit.go
gouthamMN: This is called by both `portal` and `frontend` middleware when formatting the otel Audit record.

pkg/util/log/audit/otel_audit.go
stevekuznetsov: Mock audit can implement the interface itself, there is no need for the secondary structure.

pkg/portal/middleware/log.go
stevekuznetsov: Pretty much everything in this file, yes.

test/util/log/audit.go
gouthamMN: addressed

pkg/util/log/audit/otel_audit.go
gouthamMN: addressed

pkg/util/log/audit/otel_audit.go
stevekuznetsov:
```go
type Client interface {
	Send(ctx context.Context, msg msgs.Msg) error
}

type fakeClient struct {}

func (f *fakeClient) Send(ctx context.Context, msg msgs.Msg) error {
	return nil
}


func DoSomething(c Client) {}

func examples{
    real, err := audit.New(...)
    DoSomething(real)
    
    fake := &fakeClient{}
    DoSomething(fake)
}
```

pkg/util/log/audit/otel_audit.go
stevekuznetsov: Did you mean to push another commit? I still see an environment variable access.

pkg/util/log/audit/otel_audit.go
stevekuznetsov: Yes, a parameter passed through config file or flag into the RP. This value must be big enough to store sufficiently high number of logs such that none are lost, right? That depends on the rate of creation for logs as well as the rate at which they are read from the queue. We don't know those rates a priori and the rates may be different in different environments, so making this tunable by the operators is a good idea.

pkg/util/log/audit/otel_audit.go
gouthamMN: addressed

pkg/util/log/audit/otel_audit.go
gouthamMN: We need this method to send the msg to audit server

pkg/util/log/audit/otel_audit.go
stevekuznetsov: @gouthamMN did you mean to push another commit? We can entirely remove `type Audit struct` - you only need to pass around implementations of `type Client interface`

pkg/portal/middleware/log.go
stevekuznetsov: (https://www.diffchecker.com/CnmjpW5z/)

cmd/aro/portal.go
```diff
@@ -44,7 +46,9 @@ func portal(ctx context.Context, log *logrus.Entry, audit *logrus.Entry) error {
 	err = env.ValidateVars(
 		"AZURE_PORTAL_CLIENT_ID",
 		"AZURE_PORTAL_ACCESS_GROUP_IDS",
-		"AZURE_PORTAL_ELEVATED_GROUP_IDS")
+		"AZURE_PORTAL_ELEVATED_GROUP_IDS",
+		env.OtelAuditQueueSize,
```
stevekuznetsov, line 50: @tony-schndr suggested that this come in through RP-config. We can add a per-env field and default to 4k if unset.

pkg/util/log/audit/otel_audit.go
```diff
@@ -0,0 +1,185 @@
...
+func NewOtelAuditClient(auditLogQueueSize int, isDevEnv bool) (Client, error) {
+	if isDevEnv {
+		return initializeNoOpOtelAuditClient()
+	}
+
+	return initializeOtelAuditClient(auditLogQueueSize)
+}
+
+// https://eng.ms/docs/products/geneva/collect/instrument/opentelemetryaudit/golang/linux/installation
+func initializeOtelAuditClient(auditLogQueueSize int) (Client, error) {
```
stevekuznetsov, line 37:
```go
func initializeOtelAuditClient(auditLogQueueSize int) (Client, error) {
    return audit.New(
        func() (conn.Audit, error) {
		return conn.NewDomainSocket()
	},
	audit.WithAuditOptions(
	    base.WithSettings(
		base.Settings{
		    QueueSize: auditLogQueueSize,
		},
	    ),
	),
    )
}
```

pkg/util/log/audit/otel_audit.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
+}
+
+// initializeNoOpOtelAuditClient creates a new no-op audit client.
+// NoOP is a no-op connection to the remote audit server used during E2E testing or development environment.
+func initializeNoOpOtelAuditClient() (Client, error) {
```
stevekuznetsov, line 52:
```go
func initializeNoOpOtelAuditClient(auditLogQueueSize int) (Client, error) {
    return audit.New(
        func() (conn.Audit, error) {
		return conn.NewNoOP(), nil
	},
    )
}
```

pkg/util/log/audit/otel_audit.go
```diff
@@ -0,0 +1,185 @@
...
+		Type:   msgs.ControlPlane,
+		Record: record,
+	}
+
+	return msg
+}
+
+// Validate ensures that all required fields in the Record are set to default values if they are empty or invalid.
+// It modifies the Record in place to ensure it meets the expected structure and data requirements.
+func Validate(r *msgs.Record) {
```
stevekuznetsov, line 112: Maybe this is really `EnsureDefaults` or something? I would expect validation to return errors. Are there any fields in this that should not be set to defaults (e.g. we really expect them to be set before this is called)?

cmd/aro/portal.go
tony-schndr: I'm not sure about the validation, but the value will ultimately be sourced from https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/RP-Config

gouthamMN: Yes, these values come from RP-Config. I have updated the code to set the default value to `4000` when env is not set and for Dev this will always be `zero` as we create noop conn.

tony-schndr: I don't think this validation is right.  ValidateVars checks for environment variables, if not set then it will throw an error.  Looks like this is trying to pass the actual value.
https://github.com/Azure/ARO-RP/blob/2be29ec9ccd44d4c1ee92121659a509eca9c5e12/pkg/env/env.go#L138

pkg/util/log/audit/otel_audit.go
gouthamMN: addressed

pkg/util/log/audit/otel_audit.go
gouthamMN: addressed

pkg/util/log/audit/otel_audit.go
gouthamMN: addressed

cmd/aro/portal.go
gouthamMN: addressed

pkg/util/log/audit/otel_audit.go
gouthamMN: addressed

pkg/util/log/audit/otel_audit.go
gouthamMN: addressed

pkg/util/log/audit/otel_audit.go
gouthamMN: addressed

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -248,6 +250,7 @@ KEYVAULT_PREFIX='$KEYVAULTPREFIX'
 MDM_ACCOUNT='$RPMDMACCOUNT'
 MDM_NAMESPACE=Portal
 PORTAL_HOSTNAME='$LOCATION.admin.$RPPARENTDOMAINNAME'
+OTEL_AUDIT_QUEUE_SIZE=4000
```
tony-schndr, line 253: I think the value of queue size should be moved to https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/RP-Config

When the RP is deployed the config.yaml will be read in.

gouthamMN: addressed

stevekuznetsov (approved):

hlipsig (approved):

PR #4099, opened 2025-02-12
ARO-13921 Add cluster subnets' network security groups to List Cluster Azure Resources output if cluster uses preconfiguredNSG

aasserzo:

Fixes [ARO-13921](https://issues.redhat.com/browse/ARO-13921)

Currently the List Cluster Azure Resources admin endpoint lists all resources within the managed resource group.
BYO NSGs, that some customers might have configured, belong to different resource groups, and are therefore omitted from this endpoint's output.
This change checks whether a subnet has an NSG that belongs to a different RG, and appends it to the Cluster Resources list.

### Test plan for issue:

The test mocks two subnets, one containing a managed NSG, the other containing a BYO NSG, then runs [WriteToStream()](https://github.com/Azure/ARO-RP/blob/d813341a11fda7da4567f116c88d7dab9900e3a3/pkg/frontend/adminactions/resources_list.go#L209).
That, in turn, calls [appendAzureNetworkResources()](https://github.com/Azure/ARO-RP/blob/d813341a11fda7da4567f116c88d7dab9900e3a3/pkg/frontend/adminactions/resources_list.go#L116). 
The expected result is for both of them to be present in the resulting list.

No documentation exists for this functionality.

TBD

### DIFF
--- a/pkg/frontend/adminactions/resources_list.go
+++ b/pkg/frontend/adminactions/resources_list.go
@@ -159,6 +159,24 @@ func (a *azureActions) appendAzureNetworkResources(ctx context.Context, armResou
 					Resource: rt.RouteTable,
 				})
 			}
+			//Due to BYO NSGs not belonging to the managed resource group, we would like to list them as well
+			if snet.Properties.NetworkSecurityGroup.ID != nil {
+				r, err := azure.ParseResourceID(*snet.Properties.NetworkSecurityGroup.ID)
+				if err != nil {
+					a.log.Warnf("skipping NSG '%s' due to ID parse error: %s", *snet.Properties.NetworkSecurityGroup.ID, err)
+					continue
+				}
+				if r.ResourceGroup != stringutils.LastTokenByte(a.oc.Properties.ClusterProfile.ResourceGroupID, '/') {
+					nsg, err := a.securityGroups.Get(ctx, r.ResourceGroup, r.ResourceName, nil)
+					if err != nil {
+						a.log.Warnf("skipping NSG '%s' due to Get error: %s", *snet.Properties.NetworkSecurityGroup.ID, err)
+						continue
+					}
+					armResources = append(armResources, arm.Resource{
+						Resource: nsg.SecurityGroup,
+					})
+				}
+			}
 		}
 	}
 
--- a/pkg/frontend/adminactions/resources_list.go
+++ b/pkg/frontend/adminactions/resources_list.go
@@ -131,55 +131,57 @@ func (a *azureActions) appendAzureNetworkResources(ctx context.Context, armResou
 	armResources = append(armResources, arm.Resource{
 		Resource: vnet.VirtualNetwork,
 	})
-	if vnet.Properties.Subnets != nil {
-		for _, snet := range vnet.Properties.Subnets {
-			//we already have the VNet resource, filtering subnets instead of fetching them individually with a SubnetClient
-			interestingSubnet := (*snet.ID == a.oc.Properties.MasterProfile.SubnetID)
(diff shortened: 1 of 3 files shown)


aasserzo:

pkg/frontend/adminactions/resources_list.go
```diff
@@ -159,6 +159,24 @@ func (a *azureActions) appendAzureNetworkResources(ctx context.Context, armResou
 					Resource: rt.RouteTable,
 				})
 			}
+			//Due to BYO NSGs not belonging to the managed resource group, we would like to list them as well
+			if snet.Properties.NetworkSecurityGroup.ID != nil {
```
mrWinston, line 163: Instead of nesting this further, i would suggest you swap this if-statement around like so:

```go
if snet.Properties.NetworkSecurityGroup.ID == nil {
    continue // don't do anything when it's nil
}
// do the rest that's currently inside the if-statement.
r, err := azure.ParseResourceID(*snet.Properties.NetworkSecurityGroup.ID)
....
```

pkg/frontend/adminactions/resources_list.go
```diff
@@ -159,6 +159,24 @@ func (a *azureActions) appendAzureNetworkResources(ctx context.Context, armResou
...
+				r, err := azure.ParseResourceID(*snet.Properties.NetworkSecurityGroup.ID)
+				if err != nil {
+					a.log.Warnf("skipping NSG '%s' due to ID parse error: %s", *snet.Properties.NetworkSecurityGroup.ID, err)
+					continue
+				}
+				if r.ResourceGroup != stringutils.LastTokenByte(a.oc.Properties.ClusterProfile.ResourceGroupID, '/') {
```
mrWinston, line 169: Same here: Consider swapping the if-statement around to have less nesting.

pkg/frontend/adminactions/resources_list.go
```diff
@@ -159,6 +159,24 @@ func (a *azureActions) appendAzureNetworkResources(ctx context.Context, armResou
 					Resource: rt.RouteTable,
 				})
 			}
+			//Due to BYO NSGs not belonging to the managed resource group, we would like to list them as well
+			if snet.Properties.NetworkSecurityGroup.ID != nil {
+				r, err := azure.ParseResourceID(*snet.Properties.NetworkSecurityGroup.ID)
```
mrWinston, line 164: I'd consider renaming `r` to something more descriptive like `networkSecurityGroup` or just `nsg`.

pkg/frontend/adminactions/resources_list.go
mrWinston, line 163: Are we sure, that `snet.Properties.NetworkSecurityGroup` won't be nil? Or do we need to add a separate check for this as well?

mrWinston (commented):
Looking really good! I mostly have minor comments regarding style. the only important thing for me would be the nil check for `snet.Properties.NetworkSecurityGroup`.

pkg/frontend/adminactions/resources_list.go
aasserzo: I think the ID can only be nil if NetworkSecurityGroup is nil, and I will check for this instead, as well as add a test to cover this.

aasserzo: checking for NetworkSecurityGroup instead of NetworkSecurityGroup.ID now, also added an Subnet with no NSG property to validate this

mrWinston (approved):

tsatam (approved):

kimorris27 (approved):

PR #4100, opened 2025-02-12
Refactor aro-dnsmasq-pre.sh to not overwrite /etc/resolv.conf

ventifus:

Fixes ARO-15180
Companion installer PR https://github.com/openshift/installer-aro-wrapper/pull/255

Derived from the method in https://github.com/openshift/machine-config-operator/blob/master/templates/common/gcp/files/usr-local-bin-update-dns-server.yaml

We've been overwriting `/etc/resolv.conf`. NetworkManager owns this file and if NetworkManager needs to refresh it we will lose our changes. Instead, create a NetworkManager drop-in `/etc/NetworkManager/conf.d/dns-servers.conf` with the node's IP.

No, but the change needs to be socialized amongst ARO SRE since it affects how nameservers are managed.

Testing has been done with an extant UDR+bad dns cluster to ensure there are no external DNS dependencies. Nodes boot and scale correctly.

### DIFF
--- a/pkg/operator/controllers/dnsmasq/scripts/aro-dnsmasq-pre.sh.gotmpl
+++ b/pkg/operator/controllers/dnsmasq/scripts/aro-dnsmasq-pre.sh.gotmpl
@@ -15,7 +15,7 @@ if [ "$NODEIP" != "" ]; then
     SEARCHDOMAIN=$(awk '/^search/ { print $2; }' /etc/resolv.conf.dnsmasq)
     /bin/chmod 0744 /etc/resolv.conf.dnsmasq
 
-    cat <<EOF | /bin/tee /etc/NetworkManager/conf.d/dns-servers.conf
+    cat <<EOF | /bin/tee /etc/NetworkManager/conf.d/aro-dns.conf
 # Added by dnsmasq.service
 [global-dns]
 searches=$SEARCHDOMAIN
--- a/pkg/operator/controllers/dnsmasq/scripts/aro-dnsmasq-pre.sh.gotmpl
+++ b/pkg/operator/controllers/dnsmasq/scripts/aro-dnsmasq-pre.sh.gotmpl
@@ -12,10 +12,14 @@ NODEIP=$(/sbin/ip --json route get 168.63.129.16 | /bin/jq -r ".[].prefsrc")
 
 if [ "$NODEIP" != "" ]; then
     /bin/cp -Z /etc/resolv.conf /etc/resolv.conf.dnsmasq
+    SEARCHDOMAIN=$(awk '/^search/ { print $2; }' /etc/resolv.conf.dnsmasq)
     /bin/chmod 0744 /etc/resolv.conf.dnsmasq
 
     cat <<EOF | /bin/tee /etc/NetworkManager/conf.d/dns-servers.conf
 # Added by dnsmasq.service
+[global-dns]
+searches=$SEARCHDOMAIN
+
 [global-dns-domain-*]
 servers=$NODEIP
 EOF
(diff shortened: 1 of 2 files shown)


hawkowl:
@tsatam When the Operator is first installed, it is set to allow all reconciliations, and then that is switched to only on upgrades at the end of the install process. So, this will apply to new clusters (at the cost of a reboot + install time, so we should also update the installer wrapper).

ventifus:
I've tested this now in a UDR + misconfigured DNS cluster (vnet dns = 172.16.0.0). I set `aro.dnsmasq.enabled: "false"` and manually edited `99-master-aro-dns` and `99-worker-aro-dns` to have the new content. 

After all nodes roll out they have the following config, which looks good

```
sh-5.1# cat /etc/resolv.conf
# Generated by NetworkManager
nameserver 10.0.0.7
```
```
sh-5.1# cat /etc/resolv.conf.dnsmasq
# Generated by NetworkManager
search reddog.microsoft.com
nameserver 172.16.0.0
```
```
sh-5.1# cat /etc/NetworkManager/conf.d/dns-servers.conf
# Added by dnsmasq.service
[global-dns-domain-*]
servers=10.0.0.7
```

I made sure all the cluster operators were healthy, and worker machinesets can scale up.

ventifus:
A concern I have with this is we're losing `search`. Depending on pod configuration, this can be exposed to user workloads so they may depend on it. I'll see if that's something we can preserve.

ventifus:
Ok I've fixed the search domain by adding a `[global-dns]` section.

```
sh-5.1# cat /etc/resolv.conf
# Generated by NetworkManager
search reddog.microsoft.com
nameserver 10.0.2.7
```
```
sh-5.1# cat /etc/resolv.conf.dnsmasq
# Generated by NetworkManager
search reddog.microsoft.com
nameserver 172.16.0.0
```
```
sh-5.1# cat /etc/NetworkManager/conf.d/dns-servers.conf
# Added by dnsmasq.service
[global-dns]
searches=reddog.microsoft.com

[global-dns-domain-*]
servers=10.0.2.7
```

ventifus:
Companion installer PR https://github.com/openshift/installer-aro-wrapper/pull/255

pkg/operator/controllers/dnsmasq/scripts/aro-dnsmasq-pre.sh.gotmpl
```diff
@@ -8,49 +8,25 @@ set -euo pipefail
...
-#nameservers as separate lines
-echo "$NAMESERVER_RAW" | while read -r line
-do
-    echo "nameserver $line" >> $TMPNETRESOLV
-done
-# device IPs are returned in address/mask format
-echo "$LOCAL_IPS_RAW" | while read -r line
-do
-    echo "nameserver $line" | cut -d'/' -f 1 >> $TMPSELFRESOLV
-done
```
zaneb, line 47: Unless you were having trouble with the code to retrieve the search domains and IP addresses, I'd be tempted to keep it the same.

ventifus: We do have problems with the existing code. It tries to make guesses about which network interface to use based on if the interface br-ex exists. We've seen a number of instances where this fails, particularly if the service startup order changes.

ventifus: @zaneb Another approach to finding the search domain
```
for DEV in $(nmcli --fields device,state,type --terse device | awk 'BEGIN {FS=":"} ; {if ($2 == "connected") { print $1 }}'); do nmcli dev sho $DEV | awk 'BEGIN {FS=":\\s*"}; { if ($1 ~ /DOMAIN/ && $2 ~ /.+/) { print $2} }'; done | sort -u
```

zaneb: Fair enough, non-determinism is definitely not what you want here :smile:

Looping over all interfaces doesn't look that bad though.

tsatam (approved):

cadenmarchese (approved):

PR #4102, opened 2025-02-14
Improve over-reconciliation of Hive Deployments

hawkowl:

Fixes ARO-15124

Makes admin-updates not over-reconcile the Hive deployment (e.g. erasing the version tags)

### Test plan for issue:

I added unit tests, hopefully E2E checks it doesn't break anything? 

N/A

Not sure how we test Hive-touching changes.

### DIFF
--- a/pkg/util/clienthelper/clienthelper_test.go
+++ b/pkg/util/clienthelper/clienthelper_test.go
@@ -698,7 +698,7 @@ func TestMerge(t *testing.T) {
 			wantEmptyDiff: true,
 		},
 		{
-			name: "Hive ClusterDeployment no changes",
+			name: "Hive ClusterDeployment changes",
 			old: &hivev1.ClusterDeployment{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: map[string]string{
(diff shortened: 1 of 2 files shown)


pkg/util/clienthelper/clienthelper_test.go
```diff
@@ -637,6 +640,124 @@ func TestMerge(t *testing.T) {
...
+				},
+				Status: hivev1.ClusterDeploymentStatus{
+					APIURL: "example",
+				},
+			},
+			wantChanged:   false,
+			wantEmptyDiff: true,
+		},
+		{
+			name: "Hive ClusterDeployment no changes",
```
kimorris27, line 701: Nit:

```suggestion
			name: "Hive ClusterDeployment changes",
```

kimorris27 (commented):
It looks to me like it'll get tested via this code path that installs touch: https://github.com/Azure/ARO-RP/blob/master/pkg/hive/install.go#L89-L113 (maybe not perfectly, but it'll at least make sure it doesn't break things?)

tsatam (approved):

PR #4111, opened 2025-02-20
ARO-13789 Add support for deleting LB probes

dmichellis:

Fixes

### DIFF
--- a/pkg/frontend/adminactions/delete_managedresource.go
+++ b/pkg/frontend/adminactions/delete_managedresource.go
@@ -17,8 +17,10 @@ import (
 )
 
 var (
-	frontendIPConfigurationPattern = `(?i)^/subscriptions/(.+)/resourceGroups/(.+)/providers/Microsoft\.Network/loadBalancers/(.+)/frontendIPConfigurations/([^/]+)$`
-	denyList                       = []string{
+	frontendIPConfigurationPattern        = `(?i)^/subscriptions/(.+)/resourceGroups/(.+)/providers/Microsoft\.Network/loadBalancers/(.+)/frontendIPConfigurations/([^/]+)$`
+	loadbalancerProbeConfigurationPattern = `(?i)^/subscriptions/(.+)/resourceGroups/(.+)/providers/Microsoft\.Network/loadBalancers/(.+)/probes/([^/]+)$`
+
+	denyList = []string{
 		`(?i)^/subscriptions/(.+)/resourceGroups/(.+)/providers/Microsoft\.Network/privateLinkServices/([^/]+)$`,
 		`(?i)^/subscriptions/(.+)/resourceGroups/(.+)/providers/Microsoft\.Network/privateEndpoints/([^/]+)$`,
 		`(?i)^/subscriptions/(.+)/resourceGroups/(.+)/providers/Microsoft\.Storage/(.+)$`,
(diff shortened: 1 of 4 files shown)


pkg/frontend/adminactions/delete_managedresource.go
```diff
@@ -71,3 +79,22 @@ func (a *azureActions) deleteFrontendIPConfiguration(ctx context.Context, resour
 
 	return a.loadBalancers.CreateOrUpdateAndWait(ctx, rg, lbName, lb)
 }
+
+func (a *azureActions) deleteLoadbalancerProbeConfiguration(ctx context.Context, resourceID string) error {
+	// For further details on why this was added, please see JIRA ticket [ARO-13789](https://issues.redhat.com/browse/ARO-13789)
+	idParts := strings.Split(resourceID, "/")
```
aasserzo, line 85: Maybe add a check for nil/empty vars here?

kimorris27: The Azure SDK for Go has a function that will make this easier and safer: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/arm#ParseResourceID

pkg/util/loadbalancer/loadbalancer.go
```diff
@@ -24,6 +24,24 @@ func RemoveFrontendIPConfiguration(lb *mgmtnetwork.LoadBalancer, resourceID stri
 	lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations = &newFrontendIPConfig
 	return nil
 }
+func RemoveLoadbalancerProbeConfiguration(lb *mgmtnetwork.LoadBalancer, resourceID string) error {
+	newProbeConfiguration := make([]mgmtnetwork.Probe, 0)
+	for _, probe := range *lb.LoadBalancerPropertiesFormat.Probes {
```
kimorris27, line 29: Nit: Although it's very unlikely for this to be nil, I think it would be worth checking to avoid panics.

pkg/util/loadbalancer/loadbalancer_test.go
```diff
@@ -59,12 +59,136 @@ var originalLB = mgmtnetwork.LoadBalancer{
...
+			probeConfig: "/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/clusterRG/providers/Microsoft.Network/loadBalancers/infraID/probes/probe-1",
+			currentLB:   originalLB,
+			expectedLB:  expectedLB,
+			expectedErr: "",
+		},
+		{
+			name:        "Fail to remove probe-2 because it still has references",
+			probeConfig: "/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/clusterRG/providers/Microsoft.Network/loadBalancers/infraID/probes/probe-2",
+			currentLB:   originalLB,
+			expectedLB:  expectedLB,
```
kimorris27, line 180: We would actually want this to be the original LB - if we try and fail to delete probe 2, both probes should still be left on the load balancer.

The reason the test cases pass as-is is because the `originalLB`'s probes are stored in a pointer, so when we specify `currentLB: originalLB` in both test cases, the same underlying slice of probes is being used in both test cases. Thus, the first test case removes probe 1, and this change carries over to the second test case. You can validate that this is the case by replacing `originalLB` with a hard-coded load balancer struct in both test cases - the second test case fails.

I recommend hard-coding the Azure SDK for Go resource structs in each test case to avoid these types of issues. I know it's verbose, but it avoids confusing bugs and makes each test case clearer.

kimorris27 (changes requested):
I found a bug in the unit tests and made some other small recommendations, but the core functionality LGTM 👍🏻

PR #4116, opened 2025-02-25
CWP - Optional checks with clusterDomain

LiniSusan:

Fixes
Making api and api-int values as optional if clusterDomain is given. 
Also added few entries to identify the missing noProxy values.

This is an enhancement for existing CWP monitor for better Cx experience.

### DIFF
--- a/pkg/monitor/cluster/clusterwideproxystatus.go
+++ b/pkg/monitor/cluster/clusterwideproxystatus.go
@@ -161,45 +161,54 @@ func (mon *Monitor) emitCWPStatus(ctx context.Context) error {
 			mon.log.Errorf("Error in getting cluster information: %v", err)
 			return err
 		}
-		clusterDomain := clusterdetails.Spec.Domain
-		clusterDomaincheck := noProxyMap[clusterDomain] || noProxyMap["."+clusterDomain]
-		if !(noProxyMap[".apps."+clusterDomain] || clusterDomaincheck) {
-			missing_no_proxy_list = append(missing_no_proxy_list, clusterDomain)
-		}
 		for _, gatewayDomain := range clusterdetails.Spec.GatewayDomains {
 			gatewayDomain = strings.ToLower(gatewayDomain)
 			if !noProxyMap[gatewayDomain] {
 				missing_no_proxy_list = append(missing_no_proxy_list, gatewayDomain)
 			}
 		}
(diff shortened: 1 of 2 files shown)


hlipsig (commented):
The first three checks are nearly identical to each other, and the last three checks are nearly identical to each other. Can we refactor these to use generics instead of duplicating the code?

pkg/monitor/cluster/clusterwideproxystatus.go
```diff
@@ -162,7 +162,8 @@ func (mon *Monitor) emitCWPStatus(ctx context.Context) error {
 			return err
 		}
 		clusterDomain := clusterdetails.Spec.Domain
-		if !(noProxyMap[clusterDomain] || noProxyMap[".apps."+clusterDomain] || noProxyMap["."+clusterDomain]) {
+		clusterDomaincheck := noProxyMap[clusterDomain] || noProxyMap["."+clusterDomain]
+		if !(noProxyMap[".apps."+clusterDomain] || clusterDomaincheck) {
```
sankur-codes, line 166: Instead of adding the `clusterDomaincheck` individually, we should take it out as one if condition and then put the things that are to be executed in its absence inside it. Something like : 

```
if !(clusterDomaincheck) {
      //Perform other checks
}
```

sankur-codes (changes requested):
I would still recommend to abstract the master and worker address prefix check as much as possible into a helper function because those are quite repetitive and finally add a few test cases to support the new cluster domain check.

LiniSusan:
> The first three checks are nearly identical to each other, and the last three checks are nearly identical to each other. Can we refactor these to use generics instead of duplicating the code?

@hlipsig : even though first 3 checks looks identical, we are checking different parameters. 
for the last 3 checks I have refactored with following comment:
As per our OpenShift and ARO documentation, we expect customers to add api.clusterdomain, api-int.clusterdomain, and .apps.clusterdomain.
However, for existing customers with CWP enabled, they have only included clusterDomain in their noProxy list, and these clusters are functioning as expected.
Our SRE testing has also not identified any functionality issues with this configuration.
Therefore, we will make this check optional if the clusterDomain is already included in the list.
This check is not aligned with our documentation, but we are implemented it this way for the sake of code optimization.

rhamitarora (approved):

sankur-codes (approved):
LGTM

PR #4118, opened 2025-02-25
clustermsi: use the standard keyvault format

stevekuznetsov:
go.mod: bump Azure/msi-dataplane to v0.4.1

---

clustermsi: use the standard keyvault format

There are a number of producers and consumers of MSI data from KeyVault,
so the msi-dataplane library now exposes a canonical formatter to ensure
that the data in KeyVault can be relied on by all.

---


carvalhe (approved):
LGTM

pkg/cluster/clustermsi.go
```diff
@@ -62,40 +54,12 @@ func (m *manager) ensureClusterMsiCertificate(ctx context.Context) error {
 		return err
 	}
 
-	now := time.Now()
-
-	var expirationDate time.Time
-	if m.env.FeatureIsSet(env.FeatureUseMockMsiRp) {
```
stevekuznetsov, line 68: I didn't quite understand why we handled this logic after getting data from the MSI dataplane in mock mode, since we control the mock entirely. It's simple enough to just do the right think in the mock response.

pkg/cluster/clustermsi.go
```diff
@@ -62,40 +54,12 @@ func (m *manager) ensureClusterMsiCertificate(ctx context.Context) error {
...
-
-		// The OpenAPI spec for the MI RP specifies that NotAfter will be "in the format 2017-03-01T14:11:00Z".
-		expirationDate, err = time.Parse(time.RFC3339, *identity.NotAfter)
-		if err != nil {
-			return err
-		}
-	}
-
-	raw, err := json.Marshal(msiCredObj)
+	_, parameters, err := dataplane.FormatManagedIdentityCredentialsForStorage(m.doc.ID, *msiCredObj)
```
stevekuznetsov, line 57: TODO for me - need another PR to expose the name (the `_` output of this func) for cases where we need to look up the name based on the ID, without the rest of the msi credentials.

pkg/env/prod.go
```diff
@@ -556,12 +561,12 @@ func (m *mockClient) GetUserAssignedIdentitiesCredentials(ctx context.Context, r
 				ObjectID:                   pointerutils.ToPtr(os.Getenv("MOCK_MSI_OBJECT_ID")),
 				ResourceID:                 pointerutils.ToPtr(m.msiResourceId),
 				AuthenticationEndpoint:     pointerutils.ToPtr(m.aadHost),
-				CannotRenewAfter:           &placeholder,
+				CannotRenewAfter:           pointerutils.ToPtr(now.AddDate(0, 0, mockMsiCertValidityDays*5).Format(time.RFC3339)),
```
stevekuznetsov, line 564: I was thinking ... since we mock the msi client out entirely in testing, this is only tested in e2e. Oh well.

pkg/cluster/clustermsi.go
stevekuznetsov: Fixed in https://github.com/Azure/msi-dataplane/pull/45

kimorris27:
Ticket added to backlog for new flake: https://issues.redhat.com/browse/ARO-15531

kimorris27 (approved):

PR #4122, opened 2025-02-27
Disable and remove use of local/shared key auth for SAs

fahlmant:
https://issues.redhat.com/browse/ARO-14659

As part of security wave, all SAs need to disable local/shared key auth.

### DIFF
--- a/pkg/deploy/generator/resources_rp.go
+++ b/pkg/deploy/generator/resources_rp.go
@@ -1555,18 +1555,11 @@ func (g *generator) rpVersionStorageAccount() []*arm.Resource {
 			map[string]*string{},
 		),
 		rbac.ResourceRoleAssignmentWithName(
-			rbac.RoleStorageAccountContributor,
-			"parameters('globalDevopsServicePrincipalId')",
-			resourceTypeStorageAccount,
-			storageAccountName,
-			fmt.Sprintf("concat(%s, '/Microsoft.Authorization/', guid(resourceId('%s', %s)))", storageAccountName, resourceTypeStorageAccount, storageAccountName),
-		),
-		rbac.ResourceRoleAssignmentWithName(
 			rbac.RoleStorageBlobDataContributor,
 			"parameters('globalDevopsServicePrincipalId')",
-			resourceTypeStorageAccount,
-			storageAccountName,
-			fmt.Sprintf("concat(%s, '/Microsoft.Authorization/', guid(%s))", storageAccountName, storageAccountName),
+			resourceTypeBlobContainer,
+			"$web",
+			fmt.Sprintf("concat(%s, '/Microsoft.Authorization/', guid(resourceId('%s', %s)))", storageAccountName, resourceTypeStorageAccount, storageAccountName),
 		),
 	}
 }
--- a/pkg/deploy/generator/resources_rp.go
+++ b/pkg/deploy/generator/resources_rp.go
@@ -1562,14 +1562,15 @@ func (g *generator) rpVersionStorageAccount() []*arm.Resource {
 			fmt.Sprintf("concat(%s, '/Microsoft.Authorization/', guid(resourceId('%s', %s)))", storageAccountName, resourceTypeStorageAccount, storageAccountName),
 		),
 		g.storageAccountBlobContainer(
-			fmt.Sprintf("[resourceId('Microsoft.Storage/storageAccounts/blobServices/containers', %s, 'default', '$web')]", storageAccountName),
+			fmt.Sprintf("concat(%s, '/default', '/$web')", storageAccountName),
+			storageAccountName,
 			&mgmtstorage.ContainerProperties{},
 		),
 		rbac.ResourceRoleAssignmentWithNameAndNestedResources(
 			rbac.RoleStorageBlobDataContributor,
 			"parameters('globalDevopsServicePrincipalId')",
 			fmt.Sprintf("%s/%s", resourceTypeStorageAccount, resourceTypeBlobContainer),
-			[]string{storageAccountName, "'default'", "'$web'"},
+			[]string{storageAccountName, "'default'", "'containers/$web'"},
 			fmt.Sprintf("concat(%s, '/default/$web/Microsoft.Authorization/', guid(%s))", storageAccountName, storageAccountName),
 		),
 	}
--- a/pkg/deploy/generator/resources_rp.go
+++ b/pkg/deploy/generator/resources_rp.go
@@ -1555,11 +1555,18 @@ func (g *generator) rpVersionStorageAccount() []*arm.Resource {
 			map[string]*string{},
 		),
 		rbac.ResourceRoleAssignmentWithName(
+			rbac.RoleStorageAccountContributor,
+			"parameters('globalDevopsServicePrincipalId')",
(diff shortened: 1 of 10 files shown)


pkg/deploy/saveversion.go
```diff
@@ -6,36 +6,21 @@ package deploy
...
 	"github.com/Azure/ARO-RP/pkg/util/pointerutils"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
 // SaveVersion for current location in shared storage account for environment
-func (d *deployer) SaveVersion(ctx context.Context) error {
+func (d *deployer) SaveVersion(ctx context.Context, tokenCredential azcore.TokenCredential) error {
 	d.log.Printf("saving RP version %s deployed in %s to storage account %s", d.version, d.config.Location, *d.config.Configuration.RPVersionStorageAccountName)
-	t := time.Now().UTC().Truncate(time.Second)
-	res, err := d.globalaccounts.ListAccountSAS(
```
tsatam, line 23: Not an immediate priority but we should do this before merge - as far as I can tell this is the `only` usage of the (track1) globalaccounts client in the entire deployer, we should be good to remove it from the deployer struct entirely.

kimorris27: Looks like this is done now IIUC.

kimorris27 (commented):
Do you plan to test this with a [full service dev RP](https://github.com/Azure/ARO-RP/blob/master/docs/deploy-full-rp-service-in-dev.md) before merge? I'm not sure if it's worth the effort.

pkg/deploy/generator/resources_rp.go
```diff
@@ -1535,5 +1536,12 @@ func (g *generator) rpVersionStorageAccount() []*arm.Resource {
 			storageAccountName,
 			fmt.Sprintf("concat(%s, '/Microsoft.Authorization/', guid(resourceId('%s', %s)))", storageAccountName, resourceTypeStorageAccount, storageAccountName),
 		),
+		rbac.ResourceRoleAssignmentWithName(
+			rbac.RoleStorageBlobDataContributor,
+			"parameters('globalDevopsServicePrincipalId')",
+			resourceTypeStorageAccount,
+			storageAccountName,
+			fmt.Sprintf("concat(%s, '/Microsoft.Authorization/', guid(resourceId('%s', %s)))", storageAccountName, resourceTypeStorageAccount, storageAccountName),
```
bennerv, line 1544: said it in DM, but you're going to have a naming conflict with the role assignment name above.

fahlmant:
@kimorris27 No, I ran a INT Release pipeline against the PR to ensure that the version upload works as expected, so I agree a full deployment test is overkill.

pkg/deploy/saveversion.go
```diff
@@ -6,36 +6,21 @@ package deploy
...
-	if err != nil {
-		return err
-	}
 
-	d.log.Infof("instantiating blobs client using SAS token")
-	sasUrl := fmt.Sprintf("https://%s.blob.%s/?%s", *d.config.Configuration.RPVersionStorageAccountName, d.env.Environment().StorageEndpointSuffix, *res.AccountSasToken)
-	blobsClient, err := azblob.NewBlobsClientUsingSAS(sasUrl, d.env.Environment().ArmClientOptions())
+	d.log.Infof("instantiating blobs client using SAS token for ensure static web content is enabled")
+	serviceUrl := fmt.Sprintf("https://%s.blob.%s", *d.config.Configuration.RPVersionStorageAccountName, d.env.Environment().StorageEndpointSuffix)
+	blobsClient, err := azblob.NewBlobsClientUsingEntra(serviceUrl, tokenCredential, d.env.Environment().ArmClientOptions())
```
bennerv, line 23: Maybe instead of constructing the blobsClient in the `SaveVersion` function, we should construct it in the deployer similar to what we do for `clusterKeyvault`, from a consistency perspective.  Then we don't have to pass a token everywhere.  

Thoughts?  Non-blocking, just a comment.

bennerv: Same would apply for uploadBlobsClient below as well.

pkg/deploy/saveversion.go
```diff
@@ -50,9 +35,17 @@ func (d *deployer) SaveVersion(ctx context.Context) error {
...
 	}
 
+	d.log.Infof("instantiating blobs client using SAS token to upload content")
+	containerUrl := fmt.Sprintf("https://%s.blob.%s/%s", *d.config.Configuration.RPVersionStorageAccountName, d.env.Environment().StorageEndpointSuffix, "$web")
+	uploadBlobsClient, err := azblob.NewBlobsClientUsingEntra(containerUrl, tokenCredential, d.env.Environment().ArmClientOptions())
+	if err != nil {
+		d.log.Errorf("failure to instantiate blobs client using SAS: %v", err)
+		return err
+	}
+
```
bennerv, line 45: why is there a separate blobs client.  Can we not use the previous one?  Is there something special about the new service URL containing the $web container, even though we reference it in `upload buffer`

pkg/deploy/generator/resources_rp.go
```diff
@@ -1535,5 +1536,12 @@ func (g *generator) rpVersionStorageAccount() []*arm.Resource {
 			storageAccountName,
 			fmt.Sprintf("concat(%s, '/Microsoft.Authorization/', guid(resourceId('%s', %s)))", storageAccountName, resourceTypeStorageAccount, storageAccountName),
 		),
+		rbac.ResourceRoleAssignmentWithName(
+			rbac.RoleStorageBlobDataContributor,
+			"parameters('globalDevopsServicePrincipalId')",
+			resourceTypeStorageAccount,
+			storageAccountName,
```
bennerv, line 1543: Do we want to set the role at the scope of the $web container instead of the whole storage account?

```
az role assignment create \
    --role "Storage Blob Data Contributor" \
    --assignee <email> \
    --scope "/subscriptions/<subscription-id>/resourceGroups/<resource-group-name>/providers/Microsoft.Storage/storageAccounts/<storage-account-name>/blobServices/default/containers/<container-name>"
```

https://learn.microsoft.com/en-us/azure/storage/blobs/assign-azure-role-data-access?tabs=azure-cli#assign-an-azure-role

tsatam (approved):

tsatam (approved):

pkg/util/azureclient/azuresdk/azblob/blobs.go
```diff
@@ -9,22 +9,22 @@ import (
...
 	DeleteBlob(ctx context.Context, containerName string, blobName string, o *azblob.DeleteBlobOptions) (azblob.DeleteBlobResponse, error)
+	ServiceClient() *service.Client
 	BlobsClientAddons
 }
 
 type blobsClient struct {
 	*azblob.Client
 }
 
-var _ BlobsClient = &blobsClient{}
```
bennerv, line 26: Add this back plz :)

bennerv: https://go.dev/doc/faq#guarantee_satisfies_interface

bennerv (commented):
lgtm, just one comment

bennerv (approved):

PR #4123, opened 2025-02-28
fix(admin): preserve PreconfiguredNSG in OpenShiftCluster conversion

shubhadapaithankar:

This commit fixes a regression in the conversion logic for OpenShiftCluster objects where the **PreconfiguredNSG** field in the **NetworkProfile** was not correctly preserved or defaulted when converting from the internal `(api.OpenShiftCluster)` to the external `(admin.OpenShiftCluster)`
representation.

- Previously, if the **PreconfiguredNSG** field was missing from the internal object, the external representation did not default to the expected value (**Disabled**), leading to discrepancies in behavior. 
- if the field was explicitly set to **Enabled** internally, that value wasn’t properly carried over.

**Fixes** :  [ARO-12509](https://issues.redhat.com/browse/ARO-12509)

1. In **pkg/api/admin/openshiftcluster_convert.go** 
   - Update the mapping for the NetworkProfile field so that:
     ```go
     PreconfiguredNSG: PreconfiguredNSG(oc.Properties.NetworkProfile.PreconfiguredNSG),
     ```
     is used to set the external object's value.
   - This change ensures that:
     - When the internal object’s `PreconfiguredNSG` is set to `Enabled`, the external object reflects `Enabled`.
     - When the field is missing in the internal object, it defaults correctly to `Disabled`.

2. In the tests (located in **pkg/frontend/openshiftcluster_putorpatch_test.go**):
   - **TestConversion_PreconfiguredNSG** This test checks that:
     - **PreconfiguredNSG** is preserved as Enabled when explicitly set.
     -  **PreconfiguredNSG** defaults to Disabled when missing.
   - **TestRegression_PreconfiguredNSGField** ensures that this conversion logic does not regress in future changes.

3. Enforcing Immutability in `pkg/api/admin/openshiftcluster_validatestatic.go`
-  PreconfiguredNSG should not be mutable, so we ensured that validation prevents its modification.

4.  Test Case Fixes & Validations
- Updated tests in `pkg/frontend/openshiftcluster_putorpatch_test.go`
   
**TestConversion_PreconfiguredNSG:** 
- Ensures **PreconfiguredNSG is preserved as Enabled** when explicitly set.
- Ensures **PreconfiguredNSG defaults to Disabled** when missing.

**TestRegression_PreconfiguredNSGField**
- Ensures this conversion logic does not **regress in future changes.**

5. **Fix for TestAdminListOpenShiftCluster in admin_openshiftcluster_list_test.go**
- Fixed a mismatch in expected vs actual responses for PreconfiguredNSG.
- This issue caused test failures in the **pipeline**, which have now been resolved.

### Test plan for issue:

- Running:
```
go test -v -count=1 ./pkg/frontend -run=PreconfiguredNSG
=== RUN   TestConversion_PreconfiguredNSG
=== RUN   TestConversion_PreconfiguredNSG/preconfiguredNSG_is_correctly_preserved_as_Enabled
=== RUN   TestConversion_PreconfiguredNSG/preconfiguredNSG_defaults_to_Disabled_when_missing
--- PASS: TestConversion_PreconfiguredNSG (0.00s)
    --- PASS: TestConversion_PreconfiguredNSG/preconfiguredNSG_is_correctly_preserved_as_Enabled (0.00s)
    --- PASS: TestConversion_PreconfiguredNSG/preconfiguredNSG_defaults_to_Disabled_when_missing (0.00s)
=== RUN   TestRegression_PreconfiguredNSGField
--- PASS: TestRegression_PreconfiguredNSGField (0.00s)
PASS
ok      github.com/Azure/ARO-RP/pkg/frontend    1.143s
```
Verifying Fix for TestAdminListOpenShiftCluster
```
go test -v -count=1 ./pkg/frontend -run=TestAdminListOpenShiftCluster
time="2025-03-05T08:44:12-08:00" level=error msg="listener closed"
--- PASS: TestAdminListOpenShiftCluster (0.01s)
    --- PASS: TestAdminListOpenShiftCluster/clusters_exists_in_db (0.00s)
    --- PASS: TestAdminListOpenShiftCluster/no_clusters_found_in_db (0.00s)
    --- PASS: TestAdminListOpenShiftCluster/internal_error_while_iterating_list (0.00s)
PASS
```

### DIFF
--- a/pkg/api/admin/openshiftcluster_validatestatic.go
+++ b/pkg/api/admin/openshiftcluster_validatestatic.go
@@ -23,21 +23,16 @@ func (sv openShiftClusterStaticValidator) Static(_oc interface{}, _current *api.
 }
 
 func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCluster) error {
-	// Make a deep copy of the objects to modify without affecting originals
-	ocCopy := *oc
-	currentCopy := *current
-	ocCopy.Properties = oc.Properties
-	currentCopy.Properties = current.Properties
-	ocCopy.Properties.NetworkProfile.PreconfiguredNSG = currentCopy.Properties.NetworkProfile.PreconfiguredNSG
-
-	// Run immutability validation, excluding only PreconfiguredNSG
-	err := immutable.Validate("", &ocCopy, &currentCopy)
+	oc.Properties.NetworkProfile.PreconfiguredNSG = current.Properties.NetworkProfile.PreconfiguredNSG
+	// Run immutability validation
+	err := immutable.Validate("", oc, current)
 	if err != nil {
 		if validationErr, ok := err.(*immutable.ValidationError); ok {
 			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, validationErr.Target, validationErr.Message)
 		}
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, "", err.Error())
 	}
+
 	return validateMaintenanceTask(oc.Properties.MaintenanceTask)
 }
 
--- a/pkg/api/admin/openshiftcluster_validatestatic.go
+++ b/pkg/api/admin/openshiftcluster_validatestatic.go
@@ -24,7 +24,6 @@ func (sv openShiftClusterStaticValidator) Static(_oc interface{}, _current *api.
 
 func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCluster) error {
 	oc.Properties.NetworkProfile.PreconfiguredNSG = current.Properties.NetworkProfile.PreconfiguredNSG
-	// Run immutability validation
 	err := immutable.Validate("", oc, current)
 	if err != nil {
 		if validationErr, ok := err.(*immutable.ValidationError); ok {
--- a/pkg/api/admin/openshiftcluster_validatestatic.go
+++ b/pkg/api/admin/openshiftcluster_validatestatic.go
@@ -23,10 +23,20 @@ func (sv openShiftClusterStaticValidator) Static(_oc interface{}, _current *api.
 }
 
 func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCluster) error {
-	err := immutable.Validate("", oc, current)
+	// Make a copy of the objects to modify without affecting originals
+	ocCopy := *oc
+	currentCopy := *current
+
+	// Allow PreconfiguredNSG to change by resetting it before validation
+	ocCopy.Properties.NetworkProfile.PreconfiguredNSG = currentCopy.Properties.NetworkProfile.PreconfiguredNSG
+
+	// Run immutability validation, excluding PreconfiguredNSG
+	err := immutable.Validate("", &ocCopy, &currentCopy)
 	if err != nil {
-		err := err.(*immutable.ValidationError)
-		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, err.Target, err.Message)
+		if validationErr, ok := err.(*immutable.ValidationError); ok {
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, validationErr.Target, validationErr.Message)
+		}
+		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodePropertyChangeNotAllowed, "", err.Error())
 	}
 
 	return validateMaintenanceTask(oc.Properties.MaintenanceTask)
--- a/pkg/api/admin/openshiftcluster_validatestatic.go
+++ b/pkg/api/admin/openshiftcluster_validatestatic.go
@@ -28,11 +28,9 @@ func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCl
 	currentCopy := *current
 	ocCopy.Properties = oc.Properties
 	currentCopy.Properties = current.Properties
-	ocCopy.Properties.NetworkProfile = oc.Properties.NetworkProfile
-	currentCopy.Properties.NetworkProfile = current.Properties.NetworkProfile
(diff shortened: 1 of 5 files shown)


slawande2 (approved):
LGTM

chreed-rh:
LGTM

chreed-rh (dismissed):

pkg/api/admin/openshiftcluster_validatestatic.go
```diff
@@ -23,10 +23,20 @@ func (sv openShiftClusterStaticValidator) Static(_oc interface{}, _current *api.
 }
 
 func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCluster) error {
-	err := immutable.Validate("", oc, current)
+	// Make a copy of the objects to modify without affecting originals
+	ocCopy := *oc
+	currentCopy := *current
```
geowa4, line 28: Does this copy or dereference?

shubhadapaithankar: This dereferences the pointer and creates a shallow copy of the struct.

shubhadapaithankar: Thanks for pointing. We first used `ocCopy := *oc` and `currentCopy := *current`, which only copies the top-level fields. But nested fields (like pointers) still point to the original object. To fix this, I updated the code to fully copy the necessary fields, so changes to `ocCopy` won’t affect the original data. Now, the validation works without modifying the input."

chreed-rh (dismissed):
LGTM

slawande2 (dismissed):

RickARodriguez:
Reviewed code, everything look good to me.

RickARodriguez (approved):
Reviewed the code and everything appears to be good.

pkg/api/admin/openshiftcluster_validatestatic.go
```diff
@@ -23,12 +23,23 @@ func (sv openShiftClusterStaticValidator) Static(_oc interface{}, _current *api.
 }
 
 func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCluster) error {
-	err := immutable.Validate("", oc, current)
+	// Make a deep copy of the objects to modify without affecting originals
+	ocCopy := *oc
```
cadenmarchese, line 27: Can you give me some background on why we have changed this to validate copies rather than the original objects?

pkg/api/admin/openshiftcluster_validatestatic.go
```diff
@@ -23,12 +23,23 @@ func (sv openShiftClusterStaticValidator) Static(_oc interface{}, _current *api.
...
+	currentCopy := *current
+	ocCopy.Properties = oc.Properties
+	currentCopy.Properties = current.Properties
+	ocCopy.Properties.NetworkProfile = oc.Properties.NetworkProfile
+	currentCopy.Properties.NetworkProfile = current.Properties.NetworkProfile
+	ocCopy.Properties.NetworkProfile.PreconfiguredNSG = currentCopy.Properties.NetworkProfile.PreconfiguredNSG
```
cadenmarchese, line 33: We're only preparing NetworkProfile for comparison - what about other fields in the cluster document (cluster profile, etc..)

pkg/api/admin/openshiftcluster_validatestatic.go
shubhadapaithankar: We changed this to use copies to ensure i`mmutability validation `runs without modifying the original input objects. Previously, validation could unintentionally alter the `oc `or `current objects,` which might have unexpected side effects. By working on copies, we ensure the validation logic is non-destructive while preserving the original data integrity.

pkg/api/admin/openshiftcluster_validatestatic.go
shubhadapaithankar: Thanks for pointing this out! I've updated the code to ensure all properties are copied before validation, not just NetworkProfile. This should make immutability validation more consistent. ✅

pkg/api/admin/openshiftcluster_validatestatic.go
tsatam: Our previous implementation already took that into effect, you can see it here: https://github.com/Azure/ARO-RP/blob/master/pkg/api/util/immutable/immutable.go

We did not have issues with this validateDelta function causing the property to change previously, so I'd rather keep the implementation the way it was before.

shubhadapaithankar: Thanks for your feedback! We had to change this because the original `immutable.Validate()` was modifying the input objects while running validation. This caused issues where `PreconfiguredNSG` was mistakenly treated as immutable, even when it should be allowed to change.

I saw multiple test failures like this in pipeline and locally:
`400: PropertyChangeNotAllowed: properties.networkProfile.preconfiguredNSG: 
Changing property 'properties.networkProfile.preconfiguredNSG' is not allowed.
`
By creating copies before validation, we ensure that **immutability validation does not modify the original data** while still enforcing immutability correctly for all other fields.

This fixes the test failures while keeping all other fields properly validated.

tsatam: If `immutable.Validate()` is modifying objects passed in to it, we should fix that IMO. I can take a look and see what's going on, but it is not supposed to be modifying objects.

shubhadapaithankar: We have updated the fix to remove deep copies and instead reset` PreconfiguredNSG` before running validation.
This ensures immutability validation works correctly without modifying input objects.
The test failures are now resolved while keeping the original validation logic intact.

pkg/api/admin/openshiftcluster_validatestatic.go
```diff
@@ -23,10 +23,14 @@ func (sv openShiftClusterStaticValidator) Static(_oc interface{}, _current *api.
 }
 
 func (sv openShiftClusterStaticValidator) validateDelta(oc, current *OpenShiftCluster) error {
+	oc.Properties.NetworkProfile.PreconfiguredNSG = current.Properties.NetworkProfile.PreconfiguredNSG
+	// Run immutability validation
```
tsatam, line 27: nit: unnecessary comment

shubhadapaithankar: removed unnecessary comment

tsatam (approved):

PR #4128, opened 2025-03-03
Enhance pod.containerstatuses metric to include lastTerminationState of failed pods

yadneshk:

Fixes [ARO-7793](https://issues.redhat.com/browse/ARO-7793)

Add the reason for last termination state of pods stuck in waiting state as a dimension in pod.containerstatuses metric. The value of this field could further be used to segregate pod failures such as `OOMKilled`.

### Test plan for issue:
- Unit tests
- CI/CD 
- Canary

### DIFF
--- a/pkg/monitor/cluster/podconditions.go
+++ b/pkg/monitor/cluster/podconditions.go
@@ -108,23 +108,30 @@ func (mon *Monitor) _emitPodContainerStatuses(ps *corev1.PodList) {
 				continue
 			}
 
-			mon.emitGauge("pod.containerstatuses", 1, map[string]string{
-				"name":          p.Name,
-				"namespace":     p.Namespace,
-				"nodeName":      p.Spec.NodeName,
-				"containername": cs.Name,
-				"reason":        cs.State.Waiting.Reason,
-			})
+			containerStatus := map[string]string{
+				"name":                 p.Name,
+				"namespace":            p.Namespace,
+				"nodeName":             p.Spec.NodeName,
+				"containername":        cs.Name,
+				"reason":               cs.State.Waiting.Reason,
+				"lastTerminationState": "",
+			}
+
+			if cs.LastTerminationState.Terminated != nil {
+				containerStatus["lastTerminationState"] = cs.LastTerminationState.Terminated.Reason
+			}
+
+			mon.emitGauge("pod.containerstatuses", 1, containerStatus)
 
 			if mon.hourlyRun {
-				mon.log.WithFields(logrus.Fields{
-					"metric":        "pod.containerstatuses",
-					"name":          p.Name,
-					"namespace":     p.Namespace,
-					"containername": cs.Name,
-					"reason":        cs.State.Waiting.Reason,
-					"message":       cs.State.Waiting.Message,
-				}).Print()
+				logFields := logrus.Fields{
+					"metric":  "pod.containerstatuses",
+					"message": cs.State.Waiting.Message,
+				}
+				for label, labelVal := range containerStatus {
+					logFields[label] = labelVal
+				}
+				mon.log.WithFields(logFields).Print()
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


LiniSusan:
LGTM, 
@yadneshk :  one question: Could you please confirm whether the hourlyRun check is required here or not?

pkg/monitor/cluster/podconditions.go
```diff
@@ -160,3 +161,36 @@ func (mon *Monitor) _emitPodContainerRestartCounter(ps *corev1.PodList) {
...
+		for _, cntrStatus := range p.Status.ContainerStatuses {
+			if cntrStatus.State.Terminated == nil {
+				continue
+			}
+			if cntrStatus.State.Terminated.Reason == "OOMKilled" {
+				mon.emitGauge("pod.oomkilled", 1, map[string]string{
+					"name":          p.Name,
+					"namespace":     p.Namespace,
+					"nodeName":      p.Spec.NodeName,
+					"containername": cntrStatus.Name,
```
wanghaoran1988, line 181: Do you think it make sense if we add the pod name as a dimension?

yadneshk: @wanghaoran1988  `p.Name` represents pod name which is already added to the metric.

pkg/monitor/cluster/podconditions.go
```diff
@@ -160,3 +161,36 @@ func (mon *Monitor) _emitPodContainerRestartCounter(ps *corev1.PodList) {
...
+					"reason":        cntrStatus.State.Terminated.Reason,
```
kimorris27, line 182: Isn't including this info redundant since it will always be `OOMKilled`?

yadneshk: @kimorris27 yeah you are right! I'll exclude the reason field from the metric

pepedocs: +1 

On the other hand, I have seen this common pattern in RP monitoring code, where the intention of a dedicated metric is only to indeicate a boolean value (1 or 0) and in some cases like this it will always be 1 or 0.

From a metric type's perspective, this type of metric is close to a Gauge, but it's not a necessary one as it doesn't really represent changing values. 

I think that this should be a dimension or a label for an existing metric?

kimorris27: Nice suggestion, @pepedocs! The new implementation makes sense to me.

kimorris27 (dismissed):
LGTM, but I think it would benefit for the PR to have an additional approval from someone more knowledgeable about the monitor before merging.

pkg/monitor/cluster/podconditions.go
yadneshk: @kimorris27 thanks reviewing the latest changes.
@pepedocs I have implemented your suggestions by including the last termination state of all failed pods which would not only capture oomkilled pods but also everyother termination reasons. Thanks for the suggestion btw!

yadneshk: Also I am testing these changes tomorrow in Canary

yadneshk: Screenshot 2025-03-24 161645
Folks, here's some output from my testing in geneva. The second field is the last terminated state.

pepedocs (approved):
Looks good to me @yadneshk .

kimorris27 (approved):

pkg/monitor/cluster/podconditions.go
```diff
@@ -108,23 +108,29 @@ func (mon *Monitor) _emitPodContainerStatuses(ps *corev1.PodList) {
...
 				"name":          p.Name,
 				"namespace":     p.Namespace,
 				"nodeName":      p.Spec.NodeName,
 				"containername": cs.Name,
 				"reason":        cs.State.Waiting.Reason,
-			})
+			}
+
+			if cs.LastTerminationState.Terminated != nil {
+				containerStatus["lastTerminatedState"] = cs.LastTerminationState.Terminated.Reason
```
bizz001, line 120: Can monitors and dashboards properly segregate data by this metric if it is not present in all emitted metrics? Would a default value help?

kimorris27 (approved):

kimorris27 (approved):

PR #4131, opened 2025-03-04
Bump dependencies (March '25)

hawkowl:

Fixes some dependabot alerts.

Bumps dependencies that govulncheck and dependabot flagged. Also, replace some specific pins with excludes to avoid holding us on older versions or introducing dependencies into the tree which may no longer be held transiently.

### Test plan for issue:

CI, E2E

N/A

E2E :)


fahlmant (approved):
lgtm

PR #4137, opened 2025-03-07
CLI - Do not fail ensuring resource permissions if workerprofiles or workerprofilesstatus contain profiles with invalid subnets

tsatam:

Fixes [ARO-15596](https://issues.redhat.com/browse/ARO-15596)

Currently, our CLI, on cluster creation as well as update, attempts to ensure the cluster identity (CSP only currently) has the required permissions over all cluster resources. It does this by retrieving each subnet present on the cluster's masterProfile, workerProfiles, and workerProfilesStatus, and ensuring permissions over each subnet itself, and any route tables, NAT Gateways, and NSGs (preconfiguredNSG only) attached to each subnet. 

It is possible for these subnets to be invalid, in the following scenarios:
- A cluster is created with a specific worker subnet provided, but then all worker nodes on the cluster are replaced with new nodes pointing to a new subnet, and the original worker subnet is deleted
- A cluster has an invalid machineset, pointing to a subnet that does not exist

The resource validation will fail, immediately outputting an error corresponding to the subnet being invalid, whether the subnet itself is not found, or the resource group it belongs to is not found. 

This PR tolerates these NotFound errors and continues execution, skipping the handling of any such subnets. 

We still continue to ensure such subnets actually exist during cluster creation. 

This PR also includes an unrelated fix for a MIWI-specific regression that prevented providing the `--client-id` and `--client-secret` arguments unless explicitly specifying `--enable-managed-identity false`, in order to facilitate testing this change alongside the already-present MIWI functionality. 

### Test plan for issue:

- [X] Manually created a cluster with the invalid scenario outlined above, and ensured that this change allows cluster updates to succeed when they would have otherwise failed (on ARO-RP master's CLI extension code)

No

This change is not yet in production, we will need to upstream this change to Azure CLI's ARO module once we confirm it here.

### DIFF
--- a/python/az/aro/azext_aro/_validators.py
+++ b/python/az/aro/azext_aro/_validators.py
@@ -42,7 +42,7 @@ def validate_client_id(isCreate):
     def _validate_client_id(namespace):
         if namespace.client_id is None:
             return
-        if namespace.enable_managed_identity is True:
+        if hasattr(namespace, 'enable_managed_identity') and namespace.enable_managed_identity is True:
             raise MutuallyExclusiveArgumentError('Must not specify --client-id when --enable-managed-identity is True')  # pylint: disable=line-too-long
         if namespace.platform_workload_identities is not None:
             raise MutuallyExclusiveArgumentError('Must not specify --client-id when --assign-platform-workload-identity is used')  # pylint: disable=line-too-long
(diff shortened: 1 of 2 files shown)


kimorris27 (commented):
Two questions:

1. For the `enable-managed-identity` fix, would it make sense to add unit test cases?
2. For the worker profile fix, are we sure this is the fix we want? My takeaway from reading the Jira ticket was that we want to use worker profiles statuses if they are non-nil and use worker profiles otherwise. If we merge this PR as-is, it seems like it would result in cluster creation requests that include invalid subnets making it to the RP rather than being stopped by the CLI.

tsatam:
> 1. For the `enable-managed-identity` fix, would it make sense to add unit test cases?

I tried to, but unfortunately the mock namespace we use in the unit tests for validation do not allow me to reproduce the "missing properties" I see when using the CLI for real. 


> 2. For the worker profile fix, are we sure this is the fix we want? My takeaway from reading the Jira ticket was that we want to use worker profiles statuses if they are non-nil and use worker profiles otherwise. If we merge this PR as-is, it seems like it would result in cluster creation requests that include invalid subnets making it to the RP rather than being stopped by the CLI.

There is a separate CLI-side validation flow that runs during cluster creation (see the `dyn_validate_subnet_and_route_tables` function in `_dynamic_validators.py`, which is called for both the inputted master and worker subnets), the only invocation of the changed code here is during `ensure_resource_permissions`.

kimorris27 (approved):

shubhadapaithankar (approved):
LGTM

PR #4121, opened 2025-02-26
*: migrate more keyvault & certificates to track2

stevekuznetsov:
https://issues.redhat.com/browse/ARO-4669

PR #4139, opened 2025-03-11
ARO-4669: Swapping wrong parameters in function calls

tiguelu:

Fixes #4121
https://issues.redhat.com/browse/ARO-4669

The original PR used two arguments (for cert name and issuer name) in the wrong functions, and they had to be swapped.

The bug is not causing issues in local RP, which is why it wasn't picked up by our PR E2E. However, it prevented cluster creation in Canary.

### Test plan for issue:

A pipeline for Canary was run deploying this test branch and we confirmed we could now create clusters.

n/a

Test pipeline was run in Canary.


tsatam:
Note: the failed e2e check here is the dedicated MIWI E2E check - E2E passed for CSP cluster creation within the CI pipeline.

PR #4144, opened 2025-03-17
Remove vendoring

hawkowl:

Removes vendoring, so that we can use submodules for things like pkg/api and parts of pkg/util. For full rationale, see SD-TDR-0049.

### Test plan for issue:

Unit tests, e2e

N/A

N/A


wanghaoran1988:
The vendor directory is not removed, is this expected?

wanghaoran1988 (approved):

PR #4145, opened 2025-03-17
Draft: [ARO-13355] [POC] Reuse User Assigned Identities across e2e runs

mrWinston:

Fixes [ARO-13355](https://issues.redhat.com/browse/ARO-13355)

This PR provides a proof of concept for identity reuse according to the design document. It implements an Identity pool represented by an azure resource group. 

### Testing locally

These instruction assume that you already have your RP and local environment set up to create MiWi clusters. Also, the steps assume that your local RP is Already running.

```sh
# export variable for the identity pool resource group
export PLATFORM_WORKLOAD_IDENTITY_POOL_RESOURCEGROUP="shared-identities"
# export variable for default timeout of claimed identities
export PLATFORM_WORKLOAD_IDENTITY_POOL_CLAIM_DURATION_HOURS=4

# create the resourcegroup 
az group create --location "$LOCATION" --resource-group "$PLATFORM_WORKLOAD_IDENTITY_POOL_RESOURCEGROUP"

# create the cluster
go run ./hack/cluster create

# after creation, check identities in the identity pool rg
# identities should have tags to correlate them to the newly created cluster
az identity list --resource-group $PLATFORM_WORKLOAD_IDENTITY_POOL_RESOURCEGROUP 
...
{
    "clientId": "....",
    "id": "/....",
    "location": "eastus",
    "name": "id-...",
    "principalId": "...",
    "resourceGroup": "shared-identities",
    "systemData": null,
    "tags": {
      "CLAIMED_BY_CLUSTER": "<CLUSTER NAME>",
      "CLAIMED_BY_RG": "<CLUSTER RG>",
      "CLAIMED_UNTIL": "2025-03-17T14:49:55+00:00"
    },
  },
...


# Delete the cluster again
go run ./hack/cluster delete

# After deletion, the identities should still be there, but with it's tags empty
az identity list --resource-group $PLATFORM_WORKLOAD_IDENTITY_POOL_RESOURCEGROUP 

...
{
...
    "resourceGroup": "shared-identities",
    "systemData": null,
    "tags": {
      "CLAIMED_BY_CLUSTER": "",
      "CLAIMED_BY_RG": "",
      "CLAIMED_UNTIL": ""
    },
  },
...

# Creating a cluster again, should cause these identities to be reused again.

```


pkg/util/cluster/managedidentitypool.go
```diff
@@ -0,0 +1,207 @@
...
+
+	parsedClaimedUntil, err := time.Parse(time.RFC3339, *claimedUntil)
+	if err != nil {
+		return false
+	}
+
+	return parsedClaimedUntil.After(time.Now())
+}
+
+func (pool *ManagedIdentityPool) GetAllIdentitiesInPool(ctx context.Context) ([]*armmsi.Identity, error) {
```
tsatam, line 67: probably a future enhancement idea, but our usage of this function's output is either to  iterate over the identity list until we have the desired number of identities, at which point we exit the loop early; or to free identities belonging to a single cluster. 

We may want to consider passing the responsibility of handling the pager response to the ClaimIdentities function here, so that it can avoid having to iterate the entire identity list response if it grabs the required identities from the pool early; and to update the FreeAllIdentitiesOfCluster function to specifically target the identities present in the clusterdoc rather than iterating over all identities. 

This would only be necessary if we anticipate the list of identities to grow beyond the default page size (I think it's 1000)?

pkg/util/cluster/cluster.go
```diff
@@ -394,22 +405,39 @@ func (c *Cluster) SetupWorkloadIdentity(ctx context.Context, vnetResourceGroup s
...
-		c.log.Infof("creating WI: %s", wi.OperatorName)
-		resp, err := c.msiClient.CreateOrUpdate(ctx, vnetResourceGroup, wi.OperatorName, armmsi.Identity{
-			Location: to.StringPtr(c.Config.Location),
-		}, nil)
+	newIdentities := []*armmsi.Identity{}
+	if c.Config.IdentityPoolResourceGroup != "" {
+		c.log.Info("Claiming workload identities")
+		pool := NewManagedIdentityPool(c.msiClient, c.Config.IdentityPoolResourceGroup)
+		timeout := time.Duration(c.Config.IdentityPoolClaimDurationHours) * time.Hour
+		identities, err := pool.ClaimIdentities(ctx, len(platformWorkloadIdentityRoles), c.Config.VnetResourceGroup, c.Config.ClusterName, timeout)
```
tsatam, line 413: Do we need to do `len(platformWorkloadIdentityRoles) + 1` here, to account for the cluster MSI?

tsatam (commented):
Took another look at this PR during code review, left a comment about the number of identities necessary (current implementation seems to be off by 1).

pkg/util/cluster/cluster.go
```diff
@@ -392,22 +403,39 @@ func (c *Cluster) SetupWorkloadIdentity(ctx context.Context, vnetResourceGroup s
...
 		if err != nil {
 			return err
 		}
+		newIdentities = append(newIdentities, identities...)
+	} else {
+		for _, wi := range platformWorkloadIdentityRoles {
+			c.log.Infof("creating WI: %s", wi.OperatorName)
+			resp, err := c.msiClient.CreateOrUpdate(ctx, vnetResourceGroup, wi.OperatorName, armmsi.Identity{
```
shubhadapaithankar, line 419: Nice fallback here for creating a new identity if none are available from the pool. Would it help to log this explicitly (e.g. “Creating new identity for role %s”) to distinguish from reused ones during debugging?

PR #4146, opened 2025-03-17
[ARO-15514] feat: move subnet validation logic to subnet level

scottd018:
As part of ARO-15514 to descope permissions, move the validation logic for validating subnet permissions down the subnet level.  This removes the need for the cluster identities to have visibility to all subnets associated with the ARO VNET, as some ARO users will inter-mix ARO subnets with their other application subnets.

Additionally, adjust VNET/subnet tests to account for the split between the new logic.

Fixes https://issues.redhat.com/browse/ARO-15514

This PR moves the validation logic for the cluster identities for the subnet actions from the VNET level down to the subnet level.  This is to address a tighter least privilege implementation for the cluster identities that run the ARO service within the cluster.  Prior to this change, subnet permissions were required at the VNET level, which means that ARO could potentially have access to non-ARO subnets which could be risky to ARO users who are worried about ARO accessing/modifying subnets outside of its own scope.

### Test plan for issue:

Added unit tests and ran scripted E2E tests with a local development RP with these changes.

No documentation needs to be updated.  The current documentation for permissions will satisfy these changes.

In Azure, permissions are inherited from parent > child.  Up until now, we have required subnet permissions be set at the VNET level.  Because subnets are a child of the parent VNET, we can guarantee that each implementation in production already has these subnet-level permissions set.

### DIFF
--- a/pkg/validate/dynamic/dynamic.go
+++ b/pkg/validate/dynamic/dynamic.go
@@ -36,26 +36,28 @@ import (
 )
 
 var (
-	errMsgNSGAttached                       = "The provided subnet '%s' is invalid: must not have a network security group attached."
-	errMsgOriginalNSGNotAttached            = "The provided subnet '%s' is invalid: must have network security group '%s' attached."
-	errMsgNSGNotAttached                    = "The provided subnet '%s' is invalid: must have a network security group attached."
-	errMsgNSGNotProperlyAttached            = "When the enable-preconfigured-nsg option is specified, both the master and worker subnets should have network security groups (NSG) attached to them before starting the cluster installation."
-	errMsgSPHasNoRequiredPermissionsOnNSG   = "The %s service principal (Application ID: %s) does not have Network Contributor role on network security group '%s'. This is required when the enable-preconfigured-nsg option is specified."
-	errMsgWIHasNoRequiredPermissionsOnNSG   = "The %s platform managed identity does not have required permissions on network security group '%s'. This is required when the enable-preconfigured-nsg option is specified."
-	errMsgSubnetNotFound                    = "The provided subnet '%s' could not be found."
-	errMsgSubnetNotInSucceededState         = "The provided subnet '%s' is not in a Succeeded state"
-	errMsgSubnetInvalidSize                 = "The provided subnet '%s' is invalid: must be /27 or larger."
-	errMsgSPHasNoRequiredPermissionsOnVNet  = "The %s service principal (Application ID: %s) does not have Network Contributor role on vnet '%s'."
-	errMsgWIHasNoRequiredPermissionsOnVNet  = "The %s platform managed identity does not have required permissions on vnet '%s'."
-	errMsgVnetNotFound                      = "The vnet '%s' could not be found."
-	errMsgSPHasNoRequiredPermissionsOnRT    = "The %s service principal does not have Network Contributor role on route table '%s'."
-	errMsgWIHasNoRequiredPermissionsOnRT    = "The %s platform managed identity does not have required permissions on route table '%s'."
-	errMsgRTNotFound                        = "The route table '%s' could not be found."
-	errMsgSPHasNoRequiredPermissionsOnNatGW = "The %s service principal does not have Network Contributor role on nat gateway '%s'."
-	errMsgWIHasNoRequiredPermissionsOnNatGW = "The %s platform managed identity does not have required permissions on nat gateway '%s'."
-	errMsgNatGWNotFound                     = "The nat gateway '%s' could not be found."
-	errMsgCIDROverlaps                      = "The provided CIDRs must not overlap: '%s'."
-	errMsgInvalidVNetLocation               = "The vnet location '%s' must match the cluster location '%s'."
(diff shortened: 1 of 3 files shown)


pkg/validate/dynamic/dynamic.go
```diff
@@ -305,6 +311,95 @@ func (dv *dynamic) validateVnetPermissions(ctx context.Context, vnet azure.Resou
...
+		switch detailedErr.StatusCode {
+		case http.StatusNotFound:
+			return api.NewCloudError(
+				http.StatusBadRequest,
+				api.CloudErrorCodeInvalidLinkedSubnet,
+				"",
+				fmt.Sprintf(
+					errMsgSubnetNotFound,
+					subnetr.String(),
+				))
```
tsatam, line 390: I think we will need to "tolerate" missing subnets here and either silently continue, or do an explicit check beforehand on each subnet to ensure it exists before attempting to validate it - it is possible for a cluster to end up with missing or invalid subnets in these coordinates, simply by creating a MachineSet on the cluster that points to a nonexistent subnet (the MachineSet will fail to provision any machines, but that subnet will get propagated in the list of worker subnets evaluated here).

We've had a recent example of this happening on a production cluster, and have had to develop some fixes for the various scenarios this can cause (e.g. #4137). See this Jira epic for more context: https://issues.redhat.com/browse/ARO-15595

tsatam: ...though looking at the existing dynamic validation code more closely, we're already subject to validation "failures" in this scenario anyways, so we can consider not handling this issue in this PR if it proves to be too complex, and tackling the problem within the validation as a whole later on.

kimorris27: I agree that this issue could be handled across all of dynamic validation in a different PR.

kimorris27 (commented):
LGTM pending some testing that we're working on in both local dev and canary 👍🏻

pkg/validate/dynamic/dynamic.go
scottd018: Yeah, agreed.  If we are already subject to these failures, I think it makes the most sense to track separately, especially since this will branch out beyond just subnet object (e.g. route tables, nat gateways).

pkg/validate/dynamic/dynamic.go
```diff
@@ -305,6 +311,95 @@ func (dv *dynamic) validateVnetPermissions(ctx context.Context, vnet azure.Resou
...
+	subnetr, err := azure.ParseResourceID(s.ID)
+	if err != nil {
+		return err
+	}
+
+	// we need to explicitly set the resource type to virtualNetworks/{vnetID}/subnets, as the
+	// ParseResourceID function does not properly handle parsing child resources (e.g.
+	// VNET = parent, subnet = child) and gives the incorrect resource type, effectively
+	// giving the incorrect resource ID which causes the validateActions method to fail.
+	subnetr.ResourceType = fmt.Sprintf("%s/%s/%s", vnetr.ResourceType, vnetr.ResourceName, "subnets")
```
scottd018, line 336: @kimorris27 @tsatam are there any plans to move from the autorest/azure package to the azure-sdk-for-go?  I noticed this package is in public archive currently and I did find this "bug" in the package.  Looks like there is some heavy integration with it and the `azure.Resource` object, so I'm guessing would be a relatively large refactor.  Was just curious if this was noted at all, as this was the one piece of this PR that was interesting to solve.

kimorris27: I did a quick search in our backlog and didn't come up with anything. I think it may eventually be addressed together with the efforts to migrate from the track 1 -> track 2 SDK clients if I had to guess.

cadenmarchese:
PR looks good to me! I've attached the `hold` label until we can verify with bicep in canary, which should be unblocked as soon as we have the new Machine API role definition applied in all prod sectors.

pkg/validate/dynamic/dynamic.go
scottd018: Yeah, I think the challenge here is that the `validateActions` method expects the `azure.Resource` type as its input.  So, from a quick look, we could use the new `ResourceType` object (which comes from an internal package at https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/arm/internal/resource/resource_type.go#L28-L42) from the `ParseResourceType` function and convert it to the expected object, however the types are different enough there would still be a fair amount of manual parsing.

I think it's fair to leave it the way it is, and see what the track1 -> track2 SDK clients brings as @kimorris27 mentions above.  Happy to do the refactor in this PR for the subnets if anyone disagrees though!

tsatam: nope, I think the necessary work you outlined to make the change is sufficient to move that to its own PR, so I'm fine with this as-is for now. 

I created a subtask for this in the track1->track2 epic: https://issues.redhat.com/browse/ARO-15876

tsatam (approved):

cadenmarchese (approved):
Changes were tested in canary with both a cluster service principal install (CLI) and a MIWI install (bicep) so I'll remove the `hold` label here. lgtm!

kimorris27 (approved):

PR #4156, opened 2025-03-21
Cluster-Wide Proxy monitor changes for enhancement and better Cx experience

LiniSusan:
https://issues.redhat.com/browse/ARO-15873 

Fixes

SRE observed that geneva is not able to parse some special characters thus not able to view the metrics and it is affecting the monitor. So we changed special chars that we used in our messages
Also during the scrutiny we observed that Cx are not adding the the full gatewayDomains in the noProxy list, only a part of it is added to noProxy, so inorder to avoid false alerts and better Cx experience, did some adjustments.

Note: Creating it as draft because expecting peer review for the changes.

### DIFF
--- a/pkg/monitor/cluster/clusterwideproxystatus.go
+++ b/pkg/monitor/cluster/clusterwideproxystatus.go
@@ -44,7 +44,7 @@ func (mon *Monitor) emitCWPStatus(ctx context.Context) error {
 		})
 	} else {
 		// Create the noProxy map for efficient lookups
-		no_proxy_list := strings.Split(proxyConfig.Spec.NoProxy, ",")
+		no_proxy_list := strings.Split(proxyConfig.Status.NoProxy, ",")
 		noProxyMap := make(map[string]bool)
 		var missing_no_proxy_list []string
 		for _, proxy := range no_proxy_list {
@@ -105,7 +105,7 @@ func (mon *Monitor) emitCWPStatus(ctx context.Context) error {
 
 		if res.Properties.AddressPrefix != nil {
 			if !noProxyMap[*res.Properties.AddressPrefix] {
-				missing_no_proxy_list = append(missing_no_proxy_list, "machineCIDR:"+*res.Properties.AddressPrefix)
+				missing_no_proxy_list = append(missing_no_proxy_list, *res.Properties.AddressPrefix)
 			}
 		}
 
@@ -146,12 +146,12 @@ func (mon *Monitor) emitCWPStatus(ctx context.Context) error {
 		}
 		for _, network := range networkConfig.Spec.ClusterNetwork {
 			if !noProxyMap[network.CIDR] {
-				missing_no_proxy_list = append(missing_no_proxy_list, "PodCIDR:"+network.CIDR)
+				missing_no_proxy_list = append(missing_no_proxy_list, network.CIDR)
 			}
 		}
 		for _, network := range networkConfig.Spec.ServiceNetwork {
 			if !noProxyMap[network] {
-				missing_no_proxy_list = append(missing_no_proxy_list, "ServiceCIDR:"+network)
+				missing_no_proxy_list = append(missing_no_proxy_list, network)
 			}
 		}
 
--- a/pkg/monitor/cluster/clusterwideproxystatus.go
+++ b/pkg/monitor/cluster/clusterwideproxystatus.go
@@ -45,9 +45,17 @@ func (mon *Monitor) emitCWPStatus(ctx context.Context) error {
 	} else {
 		// Create the noProxy map for efficient lookups
 		no_proxy_list := strings.Split(proxyConfig.Status.NoProxy, ",")
+		if proxyConfig.Status.NoProxy == "" {
+			mon.emitGauge(cwp, 1, map[string]string{
+				"status":  strconv.FormatBool(true),
+				"Message": "CWP NOT Enabled Successfully. Check the network operator status or review the noProxy items ",
+			})
+			no_proxy_list = strings.Split(proxyConfig.Spec.NoProxy, ",")
+		}
 		noProxyMap := make(map[string]bool)
 		var missing_no_proxy_list []string
 		for _, proxy := range no_proxy_list {
+			proxy = strings.TrimSpace(proxy)
 			noProxyMap[proxy] = true
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/monitor/cluster/clusterwideproxystatus.go
```diff
@@ -44,7 +44,14 @@ func (mon *Monitor) emitCWPStatus(ctx context.Context) error {
 		})
 	} else {
 		// Create the noProxy map for efficient lookups
-		no_proxy_list := strings.Split(proxyConfig.Spec.NoProxy, ",")
+		no_proxy_list := strings.Split(proxyConfig.Status.NoProxy, ",")
```
pepedocs, line 47: I suggest we sanitize NoProxy first before using it. Remove trailing spaces and trailing ",". If there's a typo such as a trailing "," or " ", assuming this is not invalidated yet somewhere, then this will cause the NoProxy to return a non-empty string value, and it may also cause the no_proxy_list array to have an extra item after splitting.

pkg/monitor/cluster/clusterwideproxystatus.go
```diff
@@ -164,7 +171,24 @@ func (mon *Monitor) emitCWPStatus(ctx context.Context) error {
 		for _, gatewayDomain := range clusterdetails.Spec.GatewayDomains {
 			gatewayDomain = strings.ToLower(gatewayDomain)
 			if !noProxyMap[gatewayDomain] {
-				missing_no_proxy_list = append(missing_no_proxy_list, gatewayDomain)
+				parts := strings.Split(gatewayDomain, ".")
+				domainfound := false
+				// Loop until there are at least 3 parts
+				for len(parts) >= 2 {
```
pepedocs, line 177: Shouldn't this be >= 3, to align with the above comment (at least 3 parts)?

pkg/monitor/cluster/clusterwideproxystatus.go
LiniSusan: Good point @pepedocs : Will update it.

pkg/monitor/cluster/clusterwideproxystatus.go
LiniSusan: Yeah, initially I started with 3 and later changed to 2 for testing. I think 2 is fine, its still an open item to discuss and finalize.

pkg/monitor/cluster/clusterwideproxystatus.go
LiniSusan: @pepedocs : a blank space or typo error like ,, or , ,  will result in error from the network operator stating invalid proxy.
ex: `"The configuration is invalid for proxy 'cluster' (invalid noProxy: ). Use 'oc edit proxy.config.openshift.io cluster' to fix."`

So from the monitor perspective, I dont think we can do anything from here, but for trailing and leading space issues with the noProxy item is fixed by : strings.TrimSpace(proxy).

LiniSusan: @pepedocs , Could you please review the change and let me know your thoughts.

kimorris27 (commented):
>Also during the scrutiny we observed that Cx are not adding the the full gatewayDomains in the noProxy list, only a part of it is added to noProxy, so inorder to avoid false alerts and better Cx experience, did some adjustments.

Where is this addressed in the PR? My understanding from the email you sent out is that you intended to address it by checking the proxy resource's status rather than it's spec, but I don't think I precisely understand the difference that that will make in the outcome.

In general, the changes make sense at a high level.

LiniSusan:
@kimorris27 : Thanks for reviewing the changes.
For the above mentioned question, following is the code change:
```
	remainingParts := strings.Join(parts[1:], ".")
					// If remaining parts are found in the noProxyMap, stop checking further
					if noProxyMap[remainingParts] {
						domainfound = true
						break
					}
					// Otherwise, remove the first part and continue the check
					parts = parts[1:]
				}
				// If no valid domain was found, add the original domain to the missing list
				if !domainfound {
					missing_no_proxy_list = append(missing_no_proxy_list, gatewayDomain)
				}
```
Here what we are doing is, suppose the gateway domain is a.b.c.x.y.com and the Cx has added only x.y.com (we expect to add the full gateway domain a.b.c.x.y.com), current adjustment is: if a.b.c.x.y.com is not present then will check for b.c.x.y.com, if that is also not present then c.x.y.com and so on up to last 2 parts, in between if we see the item in between then will stop checking for it. If nothing is present then will add to missing list.

pepedocs (approved):
/LGTM

rhamitarora (dismissed):

kimorris27:
> @kimorris27 : Thanks for reviewing the changes. For the above mentioned question, following is the code change:
> 
> ```
> 	remainingParts := strings.Join(parts[1:], ".")
> 					// If remaining parts are found in the noProxyMap, stop checking further
> 					if noProxyMap[remainingParts] {
> 						domainfound = true
> 						break
> 					}
> 					// Otherwise, remove the first part and continue the check
> 					parts = parts[1:]
> 				}
> 				// If no valid domain was found, add the original domain to the missing list
> 				if !domainfound {
> 					missing_no_proxy_list = append(missing_no_proxy_list, gatewayDomain)
> 				}
> ```
> 
> Here what we are doing is, suppose the gateway domain is a.b.c.x.y.com and the Cx has added only x.y.com (we expect to add the full gateway domain a.b.c.x.y.com), current adjustment is: if a.b.c.x.y.com is not present then will check for b.c.x.y.com, if that is also not present then c.x.y.com and so on up to last 2 parts, in between if we see the item in between then will stop checking for it. If nothing is present then will add to missing list.

Oh, I understand now. "Full gateway domains" didn't mean the full list of gateway domains - it meant the full name for each entry in the list. That's the thing I was missing. Makes sense then 👍🏻

kimorris27 (dismissed):

pkg/monitor/cluster/clusterwideproxystatus.go
hlipsig: We should just make sure the comments and the loop count actually matches because that's confusing.

hlipsig (changes requested):
Just a minor nit. Otherwise it looks good. Thanks for taking ownership of the alert review and improvements.

pkg/monitor/cluster/clusterwideproxystatus.go
LiniSusan: @hlipsig : sure thing, will go with 3 itself for the time being. If there is a necessity to reduce it, will do so at that time.

kimorris27 (approved):

PR #4158, opened 2025-03-21
Skip restarting already-unhealthy VMSS instances during predeploy

tsatam:

Fixes no Jira

During our predeploy step, we rotate secrets and restart all existing RP VMSS instances to update the secrets, then wait for them to be healthy before proceeding with the deployment. However, if there is a situation where the old VMSS instances will be permanently unhealthy, this will stall the deployment.

This PR adds an explicit check before restarting the VMSS instance to see if it is currently healthy - if it is not, it will skp the restart. 

### Test plan for issue:

- [X] Unit tests were updated

No

Tests above

### DIFF
--- a/pkg/deploy/predeploy.go
+++ b/pkg/deploy/predeploy.go
@@ -546,6 +546,13 @@ func (d *deployer) restartOldScaleset(ctx context.Context, vmssName string, lbHe
 	}
 
 	for _, vm := range scalesetVMs {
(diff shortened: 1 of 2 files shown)


tsatam:
Closing this for now - we may want to adopt this kind of strategy in the future as a time-saving measure, but it is not currently necessary, as simply restarting a deployment that fails due to the VMSS instances being unhealthy is sufficient to bypass the issue.

PR #4161, opened 2025-03-24
return actionable error when invalid identity is passed

slawande2:
[ARO-15923](https://issues.redhat.com/browse/ARO-15923) and [ARO-15931](https://issues.redhat.com/browse/ARO-15931)

This PR is created to return an actionable error when an invalid platform workload identity is provided during MIWI cluster installation, instead of an Internal Server Error.  

### Test plan for issue:

Intentionally failing the cluster creation by providing an invalid identity in the create command to get an InvalidPlatformWorkloadIdentity error. 

No

### DIFF
--- a/pkg/cluster/platformworkloadidentities.go
+++ b/pkg/cluster/platformworkloadidentities.go
@@ -44,8 +44,8 @@ func (m *manager) platformWorkloadIdentityIDs(ctx context.Context) error {
 
 		identityDetails, err := m.userAssignedIdentities.Get(ctx, resourceId.ResourceGroupName, resourceId.Name, &armmsi.UserAssignedIdentitiesClientGetOptions{})
 		if err != nil {
-			if azureerrors.IsNotFoundError(err) {
-				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidPlatformWorkloadIdentity, operatorName, fmt.Sprintf("platform workload identity '%s' not found", resourceId.Name))
+			if azureerrors.IsInvalidIdentityUserError(err) {
+				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidPlatformWorkloadIdentity, operatorName, fmt.Sprintf("platform workload identity '%s' is invalid", resourceId.Name))
 			} else {
 				return fmt.Errorf("error occured when retrieving platform workload identity '%s' details: %w", resourceId.Name, err)
 			}
--- a/pkg/cluster/platformworkloadidentities.go
+++ b/pkg/cluster/platformworkloadidentities.go
@@ -46,7 +46,7 @@ func (m *manager) platformWorkloadIdentityIDs(ctx context.Context) error {
 		if err != nil {
 			if azureerrors.Is4xxError(err) {
 				m.log.Error(err)
-				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidPlatformWorkloadIdentity, fmt.Sprintf(`.platformWorkloadIdentityProfile.platformWorkloadIdentities["%s"]`, operatorName), fmt.Sprintf("platform workload identity '%s' is invalid", operatorName))
+				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidPlatformWorkloadIdentity, fmt.Sprintf(`.properties.platformWorkloadIdentityProfile.platformWorkloadIdentities["%s"]`, operatorName), fmt.Sprintf("platform workload identity '%s' is invalid", operatorName))
 			} else {
 				return fmt.Errorf("error occured when retrieving platform workload identity '%s' details: %w", operatorName, err)
 			}
--- a/pkg/cluster/platformworkloadidentities.go
+++ b/pkg/cluster/platformworkloadidentities.go
@@ -43,7 +43,7 @@ func (m *manager) platformWorkloadIdentityIDs(ctx context.Context) error {
 
 		identityDetails, err := m.userAssignedIdentities.Get(ctx, resourceId.ResourceGroupName, resourceId.Name, &armmsi.UserAssignedIdentitiesClientGetOptions{})
 		if err != nil {
-			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidPlatformWorkloadIdentity, operatorName, fmt.Sprintf("error occured when retrieving platform workload identity '%s'", operatorName))
+			return api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, operatorName, fmt.Sprintf("platform workload identity '%s' not found", operatorName))
 		}
 
 		updatedIdentities[operatorName] = api.PlatformWorkloadIdentity{
--- a/pkg/cluster/platformworkloadidentities.go
+++ b/pkg/cluster/platformworkloadidentities.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/msi/armmsi"
 
 	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/util/azureerrors"
 )
 
 func (m *manager) persistPlatformWorkloadIdentityIDs(ctx context.Context) (err error) {
@@ -43,7 +44,11 @@ func (m *manager) platformWorkloadIdentityIDs(ctx context.Context) error {
 
 		identityDetails, err := m.userAssignedIdentities.Get(ctx, resourceId.ResourceGroupName, resourceId.Name, &armmsi.UserAssignedIdentitiesClientGetOptions{})
 		if err != nil {
-			return api.NewCloudError(http.StatusNotFound, api.CloudErrorCodeResourceNotFound, operatorName, fmt.Sprintf("platform workload identity '%s' not found", operatorName))
+			if azureerrors.IsNotFoundError(err) {
+				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidPlatformWorkloadIdentity, operatorName, fmt.Sprintf("platform workload identity '%s' not found", resourceId.Name))
+			} else {
+				return fmt.Errorf("error occured when retrieving platform workload identity '%s' details: %w", resourceId.Name, err)
+			}
 		}
 
 		updatedIdentities[operatorName] = api.PlatformWorkloadIdentity{
--- a/pkg/cluster/platformworkloadidentities.go
+++ b/pkg/cluster/platformworkloadidentities.go
@@ -6,6 +6,7 @@ package cluster
 import (
 	"context"
 	"fmt"
+	"net/http"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/msi/armmsi"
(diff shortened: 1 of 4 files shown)


wanghaoran1988 (approved):

pkg/cluster/platformworkloadidentities.go
```diff
@@ -42,7 +43,7 @@ func (m *manager) platformWorkloadIdentityIDs(ctx context.Context) error {
 
 		identityDetails, err := m.userAssignedIdentities.Get(ctx, resourceId.ResourceGroupName, resourceId.Name, &armmsi.UserAssignedIdentitiesClientGetOptions{})
 		if err != nil {
-			return fmt.Errorf("error occured when retrieving platform workload identity '%s' details: %w", operatorName, err)
+			return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidPlatformWorkloadIdentity, operatorName, fmt.Sprintf("error occured when retrieving platform workload identity '%s'", operatorName))
```
pepedocs, line 46: Is this always caused by a bad request (400)? Otherwise, can we infer the specific error (e.g. 50*, 40*) from the returned error?

pepedocs (changes requested):
Can you please check my comments? @slawande

pkg/cluster/platformworkloadidentities.go
tsatam: We will likely need to enumerate specific types of error cases on the get all here in which we return the BadRequest error, and default to the internal server error otherwise. Off the top of my head, I can think of the following:
- We get a `Not Found` error (404)
- We get some sort of auth error (401/403)

slawande2: Should we make this a 404: ResourceNotFound error instead of InvalidPlatformWorkloadIdentity ?

pepedocs: @slawande2 You should be able to infer that from the returned error. Please have a look at the doc/code of the MSI client being used here and check for the type of the error returned.

I believe this is the client, please verify though - https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/msi/armmsi/userassignedidentities_client.go#L178

The main reason why I'm recommending this change is because if we don't parse the error/handle right away where it exactly happened (in the caller), there is a high chance that this error will be handled above the call stack, and this would result to error handling everywhere in RP code (bad), worst is when this error has nothing to do with the handler. You can see this in RP code if you inspect it, in my CIF work, I saw a couple of this pattern already and it makes the error handling difficult to find and reduces code readability as well.

tsatam: I don't think we should return a "404 not found" in response to a cluster create/update request if a passed-in identity doesn't exist, as the cluster itself still exists. Without thinking too hard about it, I think we should just make such an error return a Bad Request (400) for now, and we can revisit the specific error code later if we need something more specific. 

Agree with @pepedocs that we should handle all known situations where they occur, so here for all error responses from the MSI client.

tsatam: Discussed this during code review, what we think we should do here is something akin to the following:

```go
if err != nil {
  if azureerrors.IsNotFoundError(err) {
    return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidPlatformWorkloadIdentity, operatorName, fmt.Sprintf("platform workload identity '%s' not found", operatorName))
  }
  else {
    return fmt.Errorf("error occured when retrieving platform workload identity '%s' details: %w", operatorName, err)
  }
}
```

We've omitted handling auth errors from the MSI client for now, as we cannot reproduce this right away and do not know what such errors would look like. We can revisit this later in follow-up work if needed.

pepedocs: @tsatam I'm ok with the snippet above for now.

However, another suggestion would be replacing the `IsNotFoundError` to `IsUserError` to check for 4**'s. I know that 4**s can be caused by RP itself, but I am confident that this type of error will have a very low probability relative to real UserErrors. So if it happens that it's an RP error and we returned it to the user, which is a bug, this is still better than returning a `500 InternalServerError` for all errors that is not `IsNotFoundError`, which will impact the user negatively. Another advantage of this is our CIF reporting will correctly classify user errors vs server errors for most of the time. WDYT?

pepedocs: @slawande2 How about the suggestion above? ^^^ .

cadenmarchese: @pepedocs why create a new function (something like `azureerrors.IsIdentityNotFoundError` or similar) to check for 4**s when `IsNotFoundError` already does that? Or am I misunderstanding your request?

pepedocs: @cadenmarchese I thought `IsNotFoundError` is just for HTTP 404 (Resource Not Found), I didn't look at where this is implemented, if it's checking for all 4**s then @slawande2 please ignore the previous comment. Otherwise I meant `IsUserError` as all 4**s.

pepedocs: @cadenmarchese `IsNotFoundError` looks like it's just checking for 404 specifically. Please check.

https://github.com/Azure/ARO-RP/blob/master/pkg/util/azureclient/azuresdk/errors/errors.go#L16

pkg/util/azureerrors/error.go
```diff
@@ -149,3 +149,17 @@ func ResourceGroupNotFound(err error) bool {
 	}
 	return false
 }
+
+func IsInvalidIdentityUserError(err error) bool {
+	var detailedErr autorest.DetailedError
```
bennerv, line 154: Let's use `azcore` instead of autorest.  Then you won't have to cast the int below.  

`azcore.ResponseError`

bennerv: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore#ResponseError

pkg/cluster/platformworkloadidentities.go
```diff
@@ -42,7 +44,11 @@ func (m *manager) platformWorkloadIdentityIDs(ctx context.Context) error {
 
 		identityDetails, err := m.userAssignedIdentities.Get(ctx, resourceId.ResourceGroupName, resourceId.Name, &armmsi.UserAssignedIdentitiesClientGetOptions{})
 		if err != nil {
-			return fmt.Errorf("error occured when retrieving platform workload identity '%s' details: %w", operatorName, err)
+			if azureerrors.IsInvalidIdentityUserError(err) {
+				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidPlatformWorkloadIdentity, operatorName, fmt.Sprintf("platform workload identity '%s' is invalid", resourceId.Name))
```
bennerv, line 48: The one thing we're losing here is visibility into the actual error message that's here.  

Should we minimize the errors we're checking for, or log out the error message internally?  If we limit it to 404s on the identity, we know we're hitting it, but if all of a sudden we start getting 418s we're going to lose it in the response.

bennerv: I know we should / could assume any 400 is going to be a user error, but we could hit an issue internally where we're forming a bad GET request to the user identity, which is an RP error rather than a customer error.  Hence me raising it.

tsatam: > we could hit an issue internally where we're forming a bad GET request to the user identity

this is why I'm not fully comfortable with just saying "any 4xx error here is a UserError". I'd rather handle the specific, obvious cases we know about explicitly as UserErrors here, and log the rest and return an internal server error, and we can then add future enhancements for any new surprises we find out in the field later. 

Perhaps we should just always log the error here before determining how to handle it?

pepedocs: @bennerv 

>but we could hit an issue internally where we're forming a bad GET request to the user identity, which is an RP error rather than a customer error. Hence me raising it.

After doing some CIF works, my perspective on this became a little bit more pragmatic. Firstly, I agree that we shouldn't lose the specific error message. In reality, however, I've seen more UserErrors (4**) caused by users, anecdotally 99% of the time, the rest are mostly MSFT service issues, and very seldom due to an RP bug. Because of this observation. Let's say it's 80/20, I believe it's more practical to return a UserError by default because there is relatively little chance that they're actually ServerErrors such as caused by RP or MSFT service, these servers are programmed to not fail although they're not perfect. On the other hand if ServerErrors are returned by default, there is a high chance that there are actually UserErrors, users are prone to make mistakes anyways and this was very evident in our CIF dashboards in the past. One of the stories I did was properly converting the ServerErrors that was skewing our reporting to UserErrors, there were a lot of them.

In the end sending ServerError when it's not and UserError when it's not both upsets a customer, I just believe that the former is very less likely. Hence I recommended just returning a generic UserError plus the original error message. WDYT?

pepedocs: > @pepedocs I think what you've suggested makes sense, but it is a deviation from the current strategy we've generally been using to handle errors in the RP.
> 
> For the sake of not blocking this individual PR on such a shift, I'd suggest we let this PR merge as-is, since this is a fairly high-priority change, and tackle our error handling more generally in a separate, focused effort, in which we get more eyes and hands on the desired outcome and corresponding changes.

Agree. @slawande2 Could you please create a spike jira to investigate/continue this discussion?

pepedocs (dismissed):
I have suggested a [strategy](https://github.com/Azure/ARO-RP/pull/4161#discussion_r2024015607) on which default error should be returned (servererror or usererror) with the reasons in one of my comments. However, this deviates from the current strategy so for the sake of merging these changes as it's high priority, I'm approving it. I've also noted to create a Jira ticket for further discussion cc @slawande2

pkg/cluster/platformworkloadidentities.go
bennerv: Can we at least log out the error message internally so if we see something that's not a 404 or common error (it's actually us causing the issue) we know what is going on?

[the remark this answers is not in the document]
bennerv: ^ i didn't check this myself, so please confirm the path.  It should match the validation we're already doing in other places that use `.properties. ....`

stevekuznetsov: Yeah there is no justification for hiding the error by not logging at the very least. Your on-call and SRE will thank you.

pkg/util/azureerrors/error.go
```diff
@@ -149,3 +149,11 @@ func ResourceGroupNotFound(err error) bool {
 	}
 	return false
 }
+
+func IsInvalidIdentityUserError(err error) bool {
```
stevekuznetsov, line 153: n.b. this is poorly named - perhaps `Is4xxError()` - any `azcore.ResponseError` from any call can be passed here. Only in one specific context of this exported function being used will it have anything to do with identities.

cadenmarchese (approved):
lgtm

cadenmarchese (commented):
One more thing - the path needs to be the path of the internal representation, not the external one. my fault :)

cadenmarchese (approved):

pkg/cluster/platformworkloadidentities.go
```diff
@@ -42,7 +44,12 @@ func (m *manager) platformWorkloadIdentityIDs(ctx context.Context) error {
 
 		identityDetails, err := m.userAssignedIdentities.Get(ctx, resourceId.ResourceGroupName, resourceId.Name, &armmsi.UserAssignedIdentitiesClientGetOptions{})
 		if err != nil {
-			return fmt.Errorf("error occured when retrieving platform workload identity '%s' details: %w", operatorName, err)
+			if azureerrors.Is4xxError(err) {
+				m.log.Error(err)
+				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidPlatformWorkloadIdentity, fmt.Sprintf(`.properties.platformWorkloadIdentityProfile.platformWorkloadIdentities["%s"]`, operatorName), fmt.Sprintf("platform workload identity '%s' is invalid", operatorName))
+			} else {
```
bennerv, line 50: nit: this else isn't needed.  Since you're returning from within the if statement.

PR #4167, opened 2025-03-27
[ARO-13386] Add UT coverage for admin_openshiftcluster_serialconsole.go

aasserzo:

Fixes https://issues.redhat.com/browse/ARO-13386
The original description of the ticket asks for E2E coverage for any VM console log ingestion code paths that aren't covered in E2E.
Upon some investigation, this is not required. However, I did notice that admin_openshiftcluster_serialconsole_test.go is not covered by UT. 

The PR adds UT coverage for admin_openshiftcluster_serialconsole_test.go

### Test plan for issue:

NA

Negative

NA

### DIFF
--- a/pkg/frontend/admin_openshiftcluster_serialconsole_test.go
+++ b/pkg/frontend/admin_openshiftcluster_serialconsole_test.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/go-chi/chi/v5"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
@@ -73,6 +74,9 @@ func TestGetAdminOpenShiftClusterSerialConsole(t *testing.T) {
 			fixture:    databaseFixture,
 			mocks: func(mockActions *mock_adminactions.MockAzureActions) {
 				mockActions.EXPECT().
+					ResourceGroupHasVM(gomock.Any(), "master-0").
+					Return(true, nil)
+				mockActions.EXPECT().
 					VMSerialConsole(gomock.Any(), gomock.Any(), "master-0", gomock.Any()).
 					DoAndReturn(func(ctx context.Context, log *logrus.Entry, vmName string, writer io.Writer) error {
 						_, err := writer.Write([]byte("console output"))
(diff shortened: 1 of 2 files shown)


pkg/frontend/admin_openshiftcluster_serialconsole_test.go
```diff
@@ -0,0 +1,142 @@
...
+		resourceID     string
+		fixture        func(*testdatabase.Fixture)
+		mocks          func(*mock_adminactions.MockAzureActions)
+		wantStatusCode int
+		wantError      string
+	}{
+		{
+			name:       "valid request returns console output",
+			vmName:     "master-0",
+			resourceID: strings.ToLower(testdatabase.GetResourcePath(mockSubID, "resourceName")),
```
mociarain, line 71: Thought for a future PR:

I wonder could we neaten up all the URL Parameter Keys to a single constants file? Maybe this could even be injected into the Frontend.

pkg/frontend/admin_openshiftcluster_serialconsole_test.go
```diff
@@ -0,0 +1,142 @@
...
+			}
+
+			f, err := NewFrontend(context.Background(), ti.auditLog, ti.log, ti.otelAudit, ti.env, ti.dbGroup, api.APIs, &noop.Noop{}, &noop.Noop{}, nil, nil, nil, nil, func(*logrus.Entry, env.Interface, *api.OpenShiftCluster, *api.SubscriptionDocument) (adminactions.AzureActions, error) {
+				return a, nil
+			}, nil, nil)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			w := httptest.NewRecorder()
```
mociarain, line 125: Nit: I assume you named this `w` to leave `r` free for the Request but I would rather they both had clearer names. i.e. `recorder` and `request` or `rec` and `req` I guess but I have a personal preference for longer clearer names.

mociarain (commented):
Overall this is good. Just one question and I left one idea for something you might find interesting in future 
Looking at the `admin_opensgiftcluster_serialconsole.go` code I see a few other error cases we could test for. 

Why did you decide to not test these?

pkg/frontend/admin_openshiftcluster_serialconsole_test.go
tsatam: A Jira card similar to the one that drove this PR would be fine.

pkg/frontend/admin_openshiftcluster_serialconsole_test.go
```diff
@@ -0,0 +1,159 @@
...
+						CloudErrorBody: &api.CloudErrorBody{
+							Message: "invalid VM name",
+						},
+					})
+			},
+			wantStatusCode: http.StatusBadRequest,
+			wantError:      "invalid VM name",
+		},
+		{
+			name:           "missing vm name parameter returns error",
```
mociarain, line 104: This is interesting. We collapse the missing and invalid cases into one. Looking at the code the `Get` will return an empty string if the key isn't there. 

IMO we should separate this out into a distinct error i.e. a 404 VM name NOT_FOUND.

mociarain (changes requested):

pkg/frontend/admin_openshiftcluster_serialconsole_test.go
aasserzo: Added a check for empty vmName URL parameter and non existent VM
Amended tests

tiguelu (approved):
LGTM

PR #4170, opened 2025-03-28
Making fix for D2s check in full-rp setup

sankur-codes:

Add the missed check for D2s VM validation in full-rp dev setup

There was a recent change to the VM instance type in the dev environment. A follow-up [PR](https://github.com/Azure/ARO-RP/pull/4151) was raised to fix the validation error for this update, but a minor adjustment for the full-rp dev setup was missed. 

### Test plan for issue:

Not Required

No

Not meant for production. Only dev setup.


tsatam:
Thanks for fixing my mistake! Merged.

PR #4171, opened 2025-03-28
fix: GHA prune-tags

komidore64:

ARO-14879

Fixes the "Prune tags" GitHub Workflow. Command substitution does not work in a
`run` step.

### Test plan for issue:

n/a

No.

This change does not affect the RP running in production.


wanghaoran1988 (approved):

komidore64:
related to #4117.

cadenmarchese (approved):

PR #4177, opened 2025-04-01
[ARO-9395] FixSSH refactoring for CPMS enablement

kevinobriendotca:

Fixes https://issues.redhat.com/browse/ARO-9395

Running a CPMS update of a cluster breaks SSH connectivity to the master nodes.  This PR refactors the FixSSH admin update step to add the master NICs to the correct SSH backend pools after the update.   In addition, we are adding the capability to find and delete any orphaned NICs from the previous master nodes.

There are some differences in how the NICs for a public vs private cluster are updated by the machine API.  For public clusters, the new NICs are not added to the SSH backend pools at all.  For the private clusters, all 3 new NICs are added to the ssh-0 backend address pool.  The refactoring covers both cases.

### Test plan for issue:

The changes have been tested against private and public dev clusters.  This branch will be tested in canary against prod public and private clusters as well.

make test-go is passing all tests.

N/A

Production testing in the canary environment.

### DIFF
--- a/pkg/cluster/fixssh.go
+++ b/pkg/cluster/fixssh.go
@@ -10,7 +10,7 @@ import (
 	"slices"
 	"strings"
 
-	armnetwork "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
 	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
 	"github.com/Azure/go-autorest/autorest/to"
 
@@ -19,7 +19,7 @@ import (
 )
 
 var (
-	masterNICRegex               = regexp.MustCompile(`.*(master).*([0-2])\-nic`)
+	masterNICRegex               = regexp.MustCompile(`.*(master).*([0-2])-nic`)
 	interfacesCreateOrUpdateOpts = &armnetwork.InterfacesClientBeginCreateOrUpdateOptions{ResumeToken: ""}
 )
 
@@ -153,12 +153,12 @@ func (m *manager) updateILBBackendPools(ipc armnetwork.InterfaceIPConfiguration,
 	// Check for NICs that are in the wrong SSH backend pool and remove them.
 	// This covers the case for the bad NIC backend pool placements for CPMS updates to a private cluster
 	ipc.Properties.LoadBalancerBackendAddressPools = slices.DeleteFunc(ipc.Properties.LoadBalancerBackendAddressPools, func(backendPool *armnetwork.BackendAddressPool) bool {
-		delete := *backendPool.ID != *sshBackendPool.ID && strings.Contains(*backendPool.ID, "ssh-")
-		if delete {
+		remove := *backendPool.ID != *sshBackendPool.ID && strings.Contains(*backendPool.ID, "ssh-")
+		if remove {
 			m.log.Infof("Removing NIC %s from Internal Load Balancer API Address Pool %s", nicName, *backendPool.ID)
 			updated = true
 		}
-		return delete
+		return remove
 	})
 
 	if !slices.ContainsFunc(ipc.Properties.LoadBalancerBackendAddressPools, func(backendPool *armnetwork.BackendAddressPool) bool {
--- a/pkg/cluster/fixssh.go
+++ b/pkg/cluster/fixssh.go
@@ -6,13 +6,21 @@ package cluster
 import (
 	"context"
 	"fmt"
+	"regexp"
+	"slices"
 	"strings"
 
+	armnetwork "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
 	mgmtnetwork "github.com/Azure/azure-sdk-for-go/services/network/mgmt/2020-08-01/network"
 	"github.com/Azure/go-autorest/autorest/to"
 
 	"github.com/Azure/ARO-RP/pkg/api"
 	"github.com/Azure/ARO-RP/pkg/util/stringutils"
+)
+
+var (
+	masterNICRegex               = regexp.MustCompile(`.*(master).*([0-2])\-nic`)
+	interfacesCreateOrUpdateOpts = &armnetwork.InterfacesClientBeginCreateOrUpdateOptions{ResumeToken: ""}
 )
 
 func (m *manager) fixSSH(ctx context.Context) error {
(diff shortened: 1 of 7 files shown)


pkg/util/mocks/azureclient/azuresdk/armnetwork/armnetwork.go
```diff
@@ -13,7 +13,9 @@ import (
 	context "context"
 	reflect "reflect"
 
+	runtime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	armnetwork "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v2"
+	armnetwork0 "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
```
chreed-rh, line 18: Why are we loading in two version of the armnetwork library?

pkg/cluster/fixssh_test.go
```diff
@@ -6,21 +6,32 @@ package cluster
...
+	"net/http"
+	"strings"
 	"testing"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	azfake "github.com/Azure/azure-sdk-for-go/sdk/azcore/fake"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
+	fake_armnetwork "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6/fake"
```
chreed-rh, line 18: does the fake_armnetwork section refer to what that library will be referenced as in code?

chreed-rh (commented):
Looks extensive, nothing that leaps out at me red flag wise. Would be a good one to go through code review wise IMO

pkg/cluster/fixssh.go
```diff
@@ -35,230 +43,171 @@ func (m *manager) fixSSH(ctx context.Context) error {
...
 
-func (m *manager) checkAndUpdateLB(ctx context.Context, resourceGroup string, lbName string) (lb mgmtnetwork.LoadBalancer, err error) {
-	lb, err = m.loadBalancers.Get(ctx, resourceGroup, lbName, "")
-	if err != nil {
-		return lb, err
-	}
+func (m *manager) checkAndUpdateNICsInResourceGroup(ctx context.Context, resourceGroup string, infraID string, lb *mgmtnetwork.LoadBalancer) (err error) {
+	masterSubnetID := m.doc.OpenShiftCluster.Properties.MasterProfile.SubnetID
+	pagerOpts := &armnetwork.InterfacesClientListOptions{}
+	pager := m.armInterfaces.NewListPager(resourceGroup, pagerOpts)
```
bennerv, line 61: We have traditionally constructed wrappers around clients which will fetch all the resources rather than paging through them.  

I don't know if we're doing this today still, but can/should we add an Addon to the interfaces client we have today? 

Something like [this](https://github.com/Azure/ARO-RP/blob/765fc6e6b708d40663fc12f8e203948c46126ea2/pkg/util/azureclient/azuresdk/armnetwork/privatelinkservices_addons.go#L18-L29)

kevinobriendotca: 10-4.  I'll refactor that bit and add the wrapper as per the example you linked to.

pkg/util/mocks/azureclient/azuresdk/armnetwork/armnetwork.go
kevinobriendotca: I did test upgrading the armnetwork package from v2 to v6 everywhere in the code base and didn't hit any issues, but figured that is a change that might be better in a separate PR.  I'm pretty sure it's fully backward compatible.

The main reason for using the newer armnetwork v6 package is to get the fake server implementations, which make it easier to mock the InterfacesClient interface.  The fake servers allow us to return custom responses to API requests with the same behaviour as the real Azure APIs.

pkg/cluster/fixssh_test.go
kevinobriendotca: Yes, that's an import alias to the fake package name, which is just "fake".  Using an alias makes it clear which fake package we're referring to.

pkg/util/azureclient/azuresdk/armnetwork/interfaces_addons.go
```diff
@@ -32,3 +33,17 @@ func (c *interfacesClient) DeleteAndWait(ctx context.Context, resourceGroupName
 	_, err = poller.PollUntilDone(ctx, nil)
 	return err
 }
+
+func (c *interfacesClient) List(ctx context.Context, resourceGroupName string, options *armnetwork.InterfacesClientListOptions) (result []*armnetwork.Interface, err error) {
```
shubhadapaithankar, line 37: Love having this convenience List method for aggregating pages! Could we add a unit test for it—especially to cover an empty result set and an induced NextPage error—so we’re confident error propagation and empty lists behave as expected?

kevinobriendotca: I ended up bumping the armnetwork version to v6 so I could use the fake server implementation to mock the NewListPager calls and create unit tests for the new List.

pkg/containerinstall/install_test.go
```diff
@@ -26,7 +26,7 @@ import (
 	testlog "github.com/Azure/ARO-RP/test/util/log"
 )
 
-const TEST_PULLSPEC = "registry.access.redhat.com/openshift-release-dev/golang-builder--partner-share:rhel-9-golang-1.22-openshift-4.19"
+const TEST_PULLSPEC = "registry.redhat.io/ubi9/ubi-minimal:latest"
```
kimorris27: Why did you need to change this?

pkg/cluster/fixssh.go
```diff
@@ -35,230 +44,169 @@ func (m *manager) fixSSH(ctx context.Context) error {
...
 
-	if updateILBPool {
-		m.log.Printf("Adding NIC %s to Internal Load Balancer API Address Pool %s", nicName, ilbBackendPoolID)
-		*(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools = append(*(*nic.IPConfigurations)[0].LoadBalancerBackendAddressPools, mgmtnetwork.BackendAddressPool{
-			ID: &ilbBackendPoolID,
-		})
+	if !slices.ContainsFunc(ipc.Properties.LoadBalancerBackendAddressPools, func(backendPool *armnetwork.BackendAddressPool) bool {
+		return *backendPool.ID == *sshBackendPool.ID
+	}) {
+		m.log.Infof("Adding NIC %s to Internal Load Balancer API Address Pool %s", nicName, sshBackendPoolID)
```
kimorris27:
```suggestion
		m.log.Infof("Adding NIC %s to Internal Load Balancer SSH Address Pool %s", nicName, sshBackendPoolID)
```

pkg/cluster/fixssh.go
```diff
@@ -35,230 +44,169 @@ func (m *manager) fixSSH(ctx context.Context) error {
...
-			return err
-		}
+			ilbBackendPoolsUpdated = m.updateILBBackendPools(*ipc, infraID, *nic.Name, *lb.ID)
 
-		if m.doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType == api.OutboundTypeUserDefinedRouting {
-			return nil
-		}
+			if m.doc.OpenShiftCluster.Properties.NetworkProfile.OutboundType == api.OutboundTypeUserDefinedRouting {
+				m.log.Infof("Updating UDR Cluster Network Interface %s", *nic.Name)
+				err := m.armInterfaces.CreateOrUpdateAndWait(ctx, resourceGroup, *nic.Name, *nic, interfacesCreateOrUpdateOpts)
```
kimorris27, line 106: What does this update do and why does it need to be done only for UDR clusters?

pkg/cluster/fixssh.go
```diff
@@ -35,230 +44,169 @@ func (m *manager) fixSSH(ctx context.Context) error {
...
-			return err
+			elbBackendPoolsUpdated = m.updateELBBackendPools(*ipc, infraID, *nic.Name, *elb.ID)
 		}
 
-		err = m.updateELBAddressPool(ctx, &nic, nicName, &elb, resourceGroup, infraID)
-		if err != nil {
-			return err
+		if ilbBackendPoolsUpdated || elbBackendPoolsUpdated {
+			m.log.Infof("Updating Network Interface %s", *nic.Name)
+			err = m.armInterfaces.CreateOrUpdateAndWait(ctx, resourceGroup, *nic.Name, *nic, interfacesCreateOrUpdateOpts)
```
kimorris27, line 128: What is this update for?

kimorris27 (commented):
Looks great for the most part! I just asked a few questions to clarify my understanding.

pkg/containerinstall/install_test.go
kevinobriendotca: There was a previous change that broke that test because the image didn't exist in that registry.  That was a temporary fix to get the tests passing again for me.

That fix was merged here: https://github.com/Azure/ARO-RP/pull/4227

pkg/cluster/fixssh.go
kevinobriendotca: It looks like there is actually a condition missing here.  We should be checking if the backend pools were updated for that NIC before attempting the update unnecessarily.  I'll add the check for ilbBackendPoolsUpdated.

The broken SSH connectivity problem manifests itself a little differently for private clusters: all 3 NICs for the control plane nodes are placed in the ssh-0 backend pool after a CPMS update.  This puts them back into ssh-0, ssh-1 and ssh-2.

That update call will fix the SSH backend pool assignments for the internal LB and the continue is to stop executing before making any changes to the external LB pools.  The SSH backend pools are still needed for the internal SSH connectivity via a VPN gateway.

pkg/cluster/fixssh.go
kevinobriendotca: This update call is what actually updates the NIC configuration with the correct SSH backend pool assignments.  The pool assignments are modified in the IPConfiguration, then the NIC is updated with the new IPConfiguration.

pkg/cluster/fixssh.go
kimorris27: Gotcha. A cluster can be private without using UDR though, so is this still the correct condition to check?

kevinobriendotca: Good catch. I updated the check for a private cluster to check for both API and ingress visibility being private.

pkg/util/mocks/azureclient/azuresdk/armnetwork/armnetwork.go
jaitaiwan: Because we're already using deprecated apis that could stop working at a moment's notice (which is what happened to us end of last year) I would be okay with just moving straight to v6 - in fact I'd appreciate seeing it done. I just wanted to mark this as I know the updated versions do this, but mark my support for it.

jaitaiwan (dismissed):
I don't have context for FixSSH specific logic but track2 migration work in this LGTM

PR #4180, opened 2025-04-04
hotfix: correct certificate issuer

cadenmarchese:

Fixes (no JIRA)

The issuer and certificate name were swapped in the `CreateCertificate` method, causing the certificate to be created with the wrong parameters.

### Test plan for issue:

- CI and e2e
- Unit test adjustments

### DIFF
--- a/pkg/cluster/correct_cert_issuer.go
+++ b/pkg/cluster/correct_cert_issuer.go
@@ -78,7 +78,7 @@ func (m *manager) ensureCertificateIssuer(ctx context.Context, certificateName, 
 			return err
 		}
 
-		_, err = clusterKeyvault.CreateCertificate(ctx, issuerName, azcertificates.SignedCertificateParameters(certificateName, dnsName, azcertificates.EkuServerAuth), nil)
+		_, err = clusterKeyvault.CreateCertificate(ctx, certificateName, azcertificates.SignedCertificateParameters(issuerName, dnsName, azcertificates.EkuServerAuth), nil)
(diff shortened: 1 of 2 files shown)


pkg/cluster/correct_cert_issuer.go
```diff
@@ -78,7 +78,7 @@ func (m *manager) ensureCertificateIssuer(ctx context.Context, certificateName,
 			return err
 		}
 
-		_, err = clusterKeyvault.CreateCertificate(ctx, issuerName, azcertificates.SignedCertificateParameters(certificateName, dnsName, azcertificates.EkuServerAuth), nil)
+		_, err = clusterKeyvault.CreateCertificate(ctx, certificateName, azcertificates.SignedCertificateParameters(issuerName, dnsName, azcertificates.EkuServerAuth), nil)
```
tsatam, line 81: I think we need to move this call outside the `if *bundle.Policy.IssuerParameters.Name != issuerName` conditional, and wrap it in its own conditional that validates the certificate itself. Subsequent invocations of this function may assume that the certificate is already good because the policy is updated, which will not be the case for the impacted certificates due to this issue.

tsatam (changes requested):

pkg/cluster/correct_cert_issuer.go
tsatam: This is not actually necessary - the `bundle` in the existing conditional comes from the cert itself, not the policy.

PR #4184, opened 2025-04-08
ARO-16202 retry bootstrap steps if an authentication error is returned

yithian:

Fixes ARO-16202

If there's a delay between when the first party service principal is granted the necessary roles and the RP tries to perform these actions, an authorization error is returned. Instead of failing (and returning a relatively unhelpful error message to the customer), these changes cause the steps to be retried instead.

### Test plan for issue:

We can't really reproduce this issue so we'll need to watch to see if these errors continue popping up or not. If so, it's likely they'll be on a different step of the install process which will then need to be updated to retry also

No; this is a bugfix

The functions used are identical except for retrying if given an authorization failure

### DIFF
--- a/pkg/cluster/install.go
+++ b/pkg/cluster/install.go
@@ -411,10 +411,10 @@ func (m *manager) bootstrap() []steps.Step {
 
 	s = append(s,
 		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.attachNSGs),
-		steps.Action(m.updateAPIIPEarly),
-		steps.Action(m.createOrUpdateRouterIPEarly),
-		steps.Action(m.ensureGatewayCreate),
-		steps.Action(m.createAPIServerPrivateEndpoint),
+		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.updateAPIIPEarly),
+		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.createOrUpdateRouterIPEarly),
+		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.ensureGatewayCreate),
+		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.createAPIServerPrivateEndpoint),
 		steps.Action(m.createCertificates),
 	)
 


pkg/cluster/install.go
```diff
@@ -411,11 +411,11 @@ func (m *manager) bootstrap() []steps.Step {
 
 	s = append(s,
 		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.attachNSGs),
-		steps.Action(m.updateAPIIPEarly),
-		steps.Action(m.createOrUpdateRouterIPEarly),
-		steps.Action(m.ensureGatewayCreate),
-		steps.Action(m.createAPIServerPrivateEndpoint),
-		steps.Action(m.createCertificates),
+		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.updateAPIIPEarly),
+		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.createOrUpdateRouterIPEarly),
```
cadenmarchese, line 415: Makes sense to me. I like the idea of using AuthorizationRetryingAction to subsequent steps as well, to avoid the potential issue of us hitting the auth failure later on down the line.

pkg/cluster/install.go
```diff
@@ -411,11 +411,11 @@ func (m *manager) bootstrap() []steps.Step {
...
+		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.ensureGatewayCreate),
+		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.createAPIServerPrivateEndpoint),
+		steps.AuthorizationRetryingAction(m.fpAuthorizer, m.createCertificates),
```
kimorris27, line 418: I think we should skip making this step an `AuthorizationRetryingAction`. This step creates API and ingress certificates in a KV that we manage in the AME tenant, so it's not related to any RBAC that customers are responsible for setting up. Retrying won't help if this RBAC is missing.

See https://github.com/Azure/ARO-RP/blob/master/pkg/env/prod.go#L184-L195 for reference.

kimorris27 (changes requested):
LGTM except that I think one step we modified is better as-is.

pkg/cluster/install.go
cadenmarchese: Also, I confirmed that we've hit auth failures further down the line (during `m.createAPIServerPrivateEndpoint` specifically as well)

pkg/cluster/install.go
yithian: Updated!

cadenmarchese (approved):

PR #4185, opened 2025-04-08
add Makefile target for logging into shared MIWI cluster

slawande2:

This PR adds the Makefile target for logging into sre-shared-miwi-cluster

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -319,6 +319,12 @@ shared-cluster-login:
 		-u kubeadmin \
 		-p $(shell az aro list-credentials -g sre-shared-cluster -n sre-shared-cluster  -ojson --query "kubeadminPassword")
 
+.PHONY: shared-miwi-cluster-login
+shared-miwi-cluster-login:
+	@oc login $(shell az aro show -g sre-shared-miwi-cluster -n sre-shared-miwi-cluster -ojson --query apiserverProfile.url) \
+		-u kubeadmin \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -319,6 +319,12 @@ shared-cluster-login:
...
 		-p $(shell az aro list-credentials -g sre-shared-cluster -n sre-shared-cluster  -ojson --query "kubeadminPassword")
 
+.PHONY: shared-miwi-cluster-login
+shared-miwi-cluster-login:
+	@oc login $(shell az aro show -g sre-shared-miwi-cluster -n sre-shared-miwi-cluster -ojson --query apiserverProfile.url) \
+		-u kubeadmin \
+		-p $(shell az aro list-credentials -g sre-shared-miwi-cluster -n sre-shared-miwi-cluster  -ojson --query "kubeadminPassword")
+
 .PHONY: shared-cluster-create
 shared-cluster-create:
```
tsatam, line 329: Maybe orthogonal to this change (can be a follow-up item), but should we create equivalent `shared-miwi-cluster-create` and `shared-miwi-cluster-delete` targets to simplify creation/deletion of the cluster if necessary?

tsatam (approved):
LGTM, one suggestion for follow-up work but that doesn't need to be in this PR.

kimorris27 (approved):

PR #4187, opened 2025-04-11
[ARO-11589] consider failed operation that can retry as not an error, but as if inProgress

Tof1973:

Fixes ARO-11589

In some cases, we are facing Failure that have status `FailedCanRetry`. In those cases, it is better to consider it as `inProgress` to continue trying until eventual timeout, instead of just exiting with an error.

### Test plan for issue:

I did my best to try to face those `FailedCanRetry`, but was not able...

NA

In best cases, we will have situation where we have `FailedCanRetry` that will be retied and complete after some time, and in worst cases, we will reach the 15 timeout and have install failure arriving later, but we will have re-tried...

### DIFF
--- a/pkg/util/azureclient/azuresdk/azcertificates/helpers.go
+++ b/pkg/util/azureclient/azuresdk/azcertificates/helpers.go
@@ -141,7 +141,9 @@ func checkOperation(op azcertificates.CertificateOperation, log *logrus.Entry) (
 			log.Warningf("certificateOperation FailedCanRetry %s: Error %v", *op.Status, op.Error)
 			return false, nil
 		}
+		// in case the failure is not retriable, then fallthrough and continue the switch to next (default) case
 		fallthrough
+		// no code after fallthrough
 
 	default:
 		if op.StatusDetails != nil {
--- a/pkg/util/azureclient/azuresdk/azcertificates/helpers.go
+++ b/pkg/util/azureclient/azuresdk/azcertificates/helpers.go
@@ -127,6 +127,13 @@ func checkOperation(op azcertificates.CertificateOperation) (bool, error) {
 	case "completed":
 		return true, nil
 
+	case "failed":
+		// consider failed operation that can retry as not an error, but as if inProgress
+		if op.Error != nil && strings.Contains(op.Error.Error(), "[Status:FailedCanRetry]") {
+			return false, nil
+		}
+		fallthrough
+
 	default:
 		if op.StatusDetails != nil {
 			return false, fmt.Errorf("certificateOperation %s (%s): Error %w", *op.Status, *op.StatusDetails, op.Error)
--- a/pkg/util/azureclient/azuresdk/azcertificates/helpers.go
+++ b/pkg/util/azureclient/azuresdk/azcertificates/helpers.go
@@ -118,7 +118,7 @@ func WaitForCertificateOperation(parent context.Context, log *logrus.Entry, oper
 }
 
 var errorInfoContains = func(e *azcertificates.ErrorInfo, substr string) bool {
-	return e != nil && strings.Contains(e.Error(), "[Status:FailedCanRetry]")
+	return e != nil && strings.Contains(e.Error(), substr)
 }
 
 func checkOperation(op azcertificates.CertificateOperation, log *logrus.Entry) (bool, error) {
--- a/pkg/util/azureclient/azuresdk/azcertificates/helpers.go
+++ b/pkg/util/azureclient/azuresdk/azcertificates/helpers.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azcertificates"
+	"github.com/sirupsen/logrus"
 
 	"github.com/Azure/ARO-RP/pkg/util/pointerutils"
 )
@@ -101,7 +102,7 @@ func IsCertificateNotFoundError(err error) bool {
 
 // WaitForCertificateOperation wraps the certificates client to poll for an operation to finish,
 // as the Track 2 client still does not support runtime.Poller.
-func WaitForCertificateOperation(parent context.Context, operation func(ctx context.Context) (azcertificates.CertificateOperation, error)) error {
+func WaitForCertificateOperation(parent context.Context, log *logrus.Entry, operation func(ctx context.Context) (azcertificates.CertificateOperation, error)) error {
 	ctx, cancel := context.WithTimeout(parent, 15*time.Minute)
 	defer cancel()
 
@@ -111,12 +112,12 @@ func WaitForCertificateOperation(parent context.Context, operation func(ctx cont
 			return false, err
 		}
 
-		return checkOperation(op)
+		return checkOperation(op, log)
 	}, ctx.Done())
 	return err
 }
 
-func checkOperation(op azcertificates.CertificateOperation) (bool, error) {
+func checkOperation(op azcertificates.CertificateOperation, log *logrus.Entry) (bool, error) {
 	if op.Status == nil {
 		return false, fmt.Errorf("operation status is nil")
 	}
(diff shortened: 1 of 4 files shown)


pkg/util/azureclient/azuresdk/azcertificates/helpers.go
```diff
@@ -128,6 +128,13 @@ func checkOperation(op azcertificates.CertificateOperation) (bool, error) {
 	case "completed":
 		return true, nil
 
+	case "failed":
+		// consider failed operation that can retry as not an error, but as if inProgress
+		if op.Error != nil && strings.Contains(op.Error.Error(), "[Status:FailedCanRetry]") {
```
s-fairchild, line 133: Is this error logged elsewhere? If not, we should still log it here so we don't lose it, even though we aren't returning the error.

Tof1973: I do not think those errors are logged elsewhere, but I do not think we can log anything from here, as we do not have any manager/logger object here, can we ?

Tof1973: I do not think those errors are logged elsewhere, but I do not think we can log anything from here, as we do not have any manager/logger object here, can we ?

s-fairchild: In it's current state there is no logrus object to use.  
However, there is a logrus entrypoint embedded in the `m *manager` struct that calls [`WaitForCertificateOperation`](https://github.com/Azure/ARO-RP/blob/fe1dfac081f198dd5518e0c838cb517c8f80962e/pkg/cluster/tls.go#L65).
You could pass `m.log` to [`WaitForCertificateOperation`](https://github.com/Azure/ARO-RP/blob/fe1dfac081f198dd5518e0c838cb517c8f80962e/pkg/cluster/tls.go#L65), then pass it to `checkOperation` to allow logging from here.  

Logging the value of `*op.Status` just [before the case statement](https://github.com/Azure/ARO-RP/blob/3b18abb4e3a7fc4465a3bcf6463a4529a6f7ef82/pkg/util/azureclient/azuresdk/azcertificates/helpers.go#L123) would be simplest.  
As for the amount of overhead added, I'm unsure.

Tof1973: Added...

s-fairchild (changes requested):
Thank you for adding some logging here.  

> I did my best to try to face those FailedCanRetry, but was not able...

I think we can test this with a unit test?  

You will need to use gomock calls for the `m.env.ClusterCertificates().CreateCertificate()` call, then add unit tests that set `*op.Status` to all four of the case options.  
For `case "failed"` test, set `op.Error == fmt.Error("[Status:FailedCanRetry]")`.

pkg/util/azureclient/azuresdk/azcertificates/helpers_test.go
```diff
@@ -0,0 +1,74 @@
...
+			op:                azcertificates.CertificateOperation{Status: ptr("failed")},
+			errorInfoContains: errorInfoContainsTrue,
+			expectedResult:    false,
+			expectedError:     nil,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			errorInfoContains = tc.errorInfoContains
```
shubhadapaithankar, line 64: Maybe we could pass `errorInfoContains` as a parameter to checkOperation instead of overriding it here?

Tof1973: Hummm,
honestly, I do not like the idea so much, as it will change checkOperation interface just for the need of the check, and not sure what it will bring. Why having the function as parameter would be better ? What is the issue of overriding the variable function just for the purpose of the test ?

pkg/util/azureclient/azuresdk/azcertificates/helpers.go
```diff
@@ -128,6 +133,17 @@ func checkOperation(op azcertificates.CertificateOperation) (bool, error) {
 	case "completed":
 		return true, nil
 
+	case "failed":
+		// consider failed operation that can retry as not an error, but as if inProgress
+		if errorInfoContains(op.Error, "[Status:FailedCanRetry]") {
```
mociarain, line 138: Is there a link we can add in here to the docs/code where this status is generated? Just so we can check the codes in future.

pkg/util/azureclient/azuresdk/azcertificates/helpers.go
```diff
@@ -128,6 +133,17 @@ func checkOperation(op azcertificates.CertificateOperation) (bool, error) {
...
+			if op.StatusDetails != nil {
+				log.Warningf("certificateOperation FailedCanRetry %s (%s): Error %v", *op.Status, *op.StatusDetails, op.Error)
+			}
+			log.Warningf("certificateOperation FailedCanRetry %s: Error %v", *op.Status, op.Error)
+			return false, nil
+		}
+		fallthrough
```
mociarain, line 145: `fallthrough` makes me uncomfortable. I would like to see an explicit `else` instead.

pkg/util/azureclient/azuresdk/azcertificates/helpers.go
```diff
@@ -112,12 +113,16 @@ func WaitForCertificateOperation(parent context.Context, operation func(ctx cont
...
 
-		return checkOperation(op)
+		return checkOperation(op, log)
 	}, ctx.Done())
 	return err
 }
 
-func checkOperation(op azcertificates.CertificateOperation) (bool, error) {
+var errorInfoContains = func(e *azcertificates.ErrorInfo, substr string) bool {
+	return e != nil && strings.Contains(e.Error(), "[Status:FailedCanRetry]")
```
mociarain, line 122:
```suggestion
	return e != nil && strings.Contains(e.Error(), substr)
```

mociarain: I assume

mociarain (changes requested):
Few small things

pkg/util/azureclient/azuresdk/azcertificates/helpers_test.go
shubhadapaithankar: makes sense to me. I agree, better to avoid changing the function signature just for the test.

pkg/util/azureclient/azuresdk/azcertificates/helpers.go
Tof1973: AFAIK, there is no doc/code that we can reference for this status code, as this is purely in Azure code and only based on observation on our side in this US.

pkg/util/azureclient/azuresdk/azcertificates/helpers.go
Tof1973: I guess a else will implies duplication of code, and fallthrough is appropriate here.
Nevertheless, I can understand this is not common, and have added some comments to explain the logic to avoid confision...

tiguelu (changes requested):
Thank you for coding this retry. I just added some minor comments.

pkg/util/azureclient/azuresdk/azcertificates/helpers_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azcertificates"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+)
+
+func errorInfoContainsTrue(e *azcertificates.ErrorInfo, substr string) bool  { return true }
+func errorInfoContainsFalse(e *azcertificates.ErrorInfo, substr string) bool { return false }
+
+func ptr(s string) *string {
```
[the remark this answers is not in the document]
Tof1973: done

pkg/util/azureclient/azuresdk/azcertificates/helpers.go
Tof1973: done

pkg/util/azureclient/azuresdk/azcertificates/helpers_test.go
Tof1973: The issue is that ErrorInfo struct has data field which is private, so we cannot create an object of this type outside of azcertificates package, so AFAIK, it is not possible to implement Error() outside azcertificates, as we would not be able to create an object of type ErrorInfo.

tiguelu (approved):
Thank you for adding this retry logic @Tof1973

shubhadapaithankar (approved):
LGTM

PR #4196, opened 2025-04-16
Adding known targets

sankur-codes:

Adds a target "help" to list all/frequently used targets.

With alot of new targets introduced, its difficult to keep a track of what each one is used for. This will make it abit easier to figure out. 

### Test plan for issue:

NA

NA

NA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,44 @@ else
 endif
 endif
 
+.PHONY: help
+help:
+	@echo "Available targets:"
+	@echo "  aks.kubeconfig        - Get admin AKS kubeconfig"
+	@echo "  aro                   - Build the aro binary"
+	@echo "  build                 - Build binaries"
+	@echo "  build-all             - Build all binaries"
+	@echo "  check-release         - Check if the VERSION is a valid tag for release builds"
+	@echo "  ci-azext-aro          - Build the ci-azext-aro Docker image"
+	@echo "  ci-clean              - Clean up container images and buildah containers"
+	@echo "  ci-rp                 - Build the ci-rp Docker image"
+	@echo "  e2e.test              - Build the end-to-end test binary"
+	@echo "  e2etools              - Build end-to-end test tools"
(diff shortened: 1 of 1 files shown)


sankur-codes:
Simple addition and this is how the output looks like. 
image

Makefile
```diff
@@ -58,6 +58,44 @@ else
 endif
 endif
 
+.PHONY: help
+help:
+	@echo "Available targets:"
+	@echo "  aks.kubeconfig        - Get admin AKS kubeconfig"
+	@echo "  aro                   - Build the aro binary"
+	@echo "  build                 - Build binaries"
+	@echo "  build-all             - Build all binaries"
```
kevinobriendotca, line 67: A bit more information about the difference in what binaries are built with the build vs build-all targets would be helpful here.

shubhadapaithankar (approved):

Makefile
```diff
@@ -47,9 +47,12 @@ endif
 ARO_IMAGE ?= $(ARO_IMAGE_BASE):$(VERSION)
 GATEKEEPER_IMAGE ?= ${REGISTRY}/gatekeeper:$(GATEKEEPER_VERSION)
 
+
+help:  ## Show help message
+	@awk 'BEGIN {FS = ": .*##"; printf "\nUsage:\n  make \033[36m\033[0m\n"} /^[$$()% 0-9a-zA-Z_-]+(\\:[$$()% 0-9a-zA-Z_-]+)*:.*?##/ { gsub(/\\:/,":", $$1); printf "  \033[36m%-25s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
```
kevinobriendotca, line 52: It looks like the output from "make help" is missing the targets with "." in the name, ie admin.kubeconfig.

Updating the regexes to "0-9a-zA-Z._-" seems to fix it.

PR #4203, opened 2025-04-18
fix(prune-tags): correctly store command step output

komidore64:
I was incorrectly storing the `gather_tags` step output. Each bit of
stored output needs to be assigned to a key for recall later.

ref: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-output-parameter

Related to ARO-14879

This time (third time's the charm?) the GHA should output non-production git
tags that are older than 60 days, and prune them from the repo.

### Test plan for issue:

Feel free to run `.github/list-prune-tags.sh` by hand to see what the script
returns for pruning.

No.

This GitHub Workflow and Actions do not affect production.

### DIFF
--- a/.github/workflows/prune-tags.yml
+++ b/.github/workflows/prune-tags.yml
@@ -24,4 +24,4 @@ jobs:
         run: echo "TAGS=$(.github/list-prune-tags.sh -n ${{ github.ref_name }})" >> $GITHUB_OUTPUT
 
       - name: Prune tags
-        run: git push origin --delete ${{ steps.gather_tags.output.TAGS }}
+        run: git push origin --delete ${{ steps.gather_tags.outputs.TAGS }}
--- a/.github/workflows/prune-tags.yml
+++ b/.github/workflows/prune-tags.yml
@@ -18,8 +18,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
       - name: Gather tags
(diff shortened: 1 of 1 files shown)


.github/workflows/prune-tags.yml
```diff
@@ -18,8 +18,10 @@ jobs:
...
         uses: actions/checkout@v4
+
       - name: Gather tags
         id: gather_tags
-        run: .github/list-prune-tags.sh -n ${{ github.ref_name }} >> $GITHUB_OUTPUT
+        run: echo "TAGS=$(.github/list-prune-tags.sh -n ${{ github.ref_name }})" >> $GITHUB_OUTPUT
+
       - name: Prune tags
-        run: git push origin --delete ${{ steps.gather_tags.output }}
+        run: git push origin --delete ${{ steps.gather_tags.output.TAGS }}
```
kimorris27, line 27: Based on the workflow commands doc you linked to:

```suggestion
        run: git push origin --delete ${{ steps.gather_tags.outputs.TAGS }}
```

komidore64: thank you for the catch!

kimorris27 (approved):

cadenmarchese (approved):

komidore64:
ty

PR #4207, opened 2025-04-22
[ARO-16955] Move encryption at host validation to frontend skuvalidation

hawkowl:

Fixes ARO-16955

Moves the encryption-at-host validation to the SKUValidation in the front end. This probably makes more sense, since I don't really understand why we'd do it during dynamicvalidation. It also allows us to remove the VMSku stuff from pkg/env.

### Test plan for issue:

Tests, e2e

N/A

tests, hopefully :)


jaitaiwan (dismissed):
LGTM

tsatam (dismissed):

tsatam (approved):

jhoreman (approved):

PR #4209, opened 2025-04-22
Updating RP Deployment Code to avoid conflicts with DNS for staging environment

mhsreddy:

Fixes: https://issues.redhat.com/browse/ARO-14736 and https://issues.redhat.com/browse/ARO-14736

This is to update the RP Deployment code to avoid naming conflicts with DNS for the new Staging regions rollout.

### Test plan for issue:

Tested the DNS entries creation in Full RP Dev Setup.  Also tested the same in INT and Canary region (eastus2euap).

None.

By testing this INT, Stage and Canary before rolling out to productions regions.


cadenmarchese (approved):
makes sense - thanks for testing in canary :)

PR #4210, opened 2025-04-23
MIWI CLI: automated identity deletion

kimorris27:

https://issues.redhat.com/browse/ARO-6443

This PR adds automated identity deletion for MIWI clusters. Some key things to note:

- There is a single `--delete-identities` flag that signals to the CLI that the user wants to delete both the cluster MSI and the platform identities
- The PR includes validation of the CLI user's permission to delete identities prior to proceeding with cluster deletion
- If the user doesn't want their MIWI cluster's identities to be automatically deleted, the CLI will now print commands they can use to delete the identities
- I feel like the CLI's dynamic validation could use some refactoring, but it's out of scope for this PR (more context in one of my commit messages)

### Test plan for issue:

Unit tests for newly-added dynamic validation plus the following locally executed ad-hoc tests:

- Created and deleted CSP cluster to confirm there were no regressions
- Created and deleted MIWI cluster both with and without `--delete-identities` and validated the behavior in each case
- Logged into Azure via CLI as a service principal that lacked identity delete permission and attempted to delete a MIWI cluster with `--delete-identities`; confirmed that deletion was blocked by lack of permission and the CLI printed messages explaining the missing permissions

When we release the automated CLI features, we'll probably want to add a "how to delete a managed identity cluster" page that explains how the `--delete-identities` flag works.

The testing I described above.


python/az/aro/azext_aro/_params.py
```diff
@@ -174,3 +175,10 @@ def load_arguments(self, _):
         c.argument('file',
                    help='Path to the file where kubeconfig should be saved. Default: kubeconfig in local directory',
                    options_list=['--file', '-f'])
+
+    with self.argument_context('aro delete') as c:
+        c.argument('delete_identities',
```
tsatam, line 180: nit: we should add the `is_preview=True` property on this argument, in case we cut a new release of the preview extension with this new functionality.

tsatam (changes requested):
Implementation LGTM, one suggestion to keep this new identity argument as a preview for now.

python/az/aro/azext_aro/_params.py
kimorris27: Done 👍🏻

tsatam (approved):

PR #4211, opened 2025-04-24
[ARO-9171] Remove clustermanager code from pkg/database and pkg/api

hawkowl:

Fixes ARO-9171

Removes the last clustermanager code from pkg/database and pkg/api, as well as creation of the databases. Since the databases had no data in them (the APIs were never used), we can safely just leave the extant ones.

### Test plan for issue:

CI, E2E

N/A

Removing of unused code, should be fine.


tsatam (approved):
LGTM. I think we'll need a separate, dedicated release pipeline created to actually delete the existing database containers, since removing the ARM resource from our deployment won't delete the resource automatically, but that shouldn't be in or block this PR.

PR #4222, opened 2025-05-06
Refresh MSI certs on update, adminUpdate if eligible for rotation

cadenmarchese:

Fixes https://issues.redhat.com/browse/ARO-17110

- MSI certs are required to refresh federated identity credentials on MIWI clusters and have a valid lifespan of 90 days.
- We are working on a credential refresher which will run in in an AKS cluster and refresh these automatically before expiration, but as a stop-gap we also want to refresh them in case they expire or are eligible for a refresh and an update / adminUpdate is ran, so that subsequent update steps that require valid MSI certs don't fail.

### Test plan for issue:

- [x] local dev
  - Create, update, delete, with expired and eligible MSI certs to confirm they were refreshed
- [x] Canary CRUD smoke test
- [x] Unit testing / CI / e2e

### DIFF
--- a/pkg/cluster/clustermsi.go
+++ b/pkg/cluster/clustermsi.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/msi-dataplane/pkg/dataplane"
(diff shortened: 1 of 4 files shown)


kimorris27 (commented):
LGTM pending result of the canary testing 👍🏻

pkg/cluster/clustermsi.go
```diff
@@ -24,13 +26,17 @@ var (
 
 // ensureClusterMsiCertificate leverages the MSI dataplane module to fetch the MSI's
 // backing certificate (if needed) and store the certificate in the cluster MSI key
-// vault. It does not concern itself with whether an existing certificate is valid
-// or not; that can be left to the certificate refresher component.
+// vault. If the certificate stored in keyvault is invalid, it will request and persist
```
kimorris27, line 29: Nit:

```suggestion
// vault. If the certificate stored in keyvault is eligible for renewal, this function will request and persist
```

kimorris27 (changes requested):
Recommending one correction to a comment, but LGTM otherwise 👍🏻

kimorris27 (approved):

PR #4223, opened 2025-05-07
Impliment Monitoring for Cluster Authentication Type

slawande2:
[ARO-15682](https://issues.redhat.com/browse/ARO-15682)

This PR creates a monitor for cluster authentication type, either CSP or MI, So that metrics are emitted which report the authentication type in use by every cluster and we have an accurate, live count of clusters using both authentication types.

### Test plan for issue:

Unit tests have been added

### DIFF
--- a/pkg/monitor/cluster/clusterauthenticationtype.go
+++ b/pkg/monitor/cluster/clusterauthenticationtype.go
@@ -5,34 +5,21 @@ package cluster
 
 import (
 	"context"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	operatorv1 "github.com/openshift/api/operator/v1"
 )
 
 const (
-	authenticationTypeMetricsTopic = "cluster.cloudCredentialsMode"
+	authenticationTypeMetricsTopic = "cluster.AuthenticationType"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/monitor/cluster/clusterauthenticationtype.go
```diff
@@ -0,0 +1,38 @@
...
+func (mon *Monitor) emitClusterAuthenticationType(ctx context.Context) error {
+	cloudCredentialObject, err := mon.operatorcli.OperatorV1().CloudCredentials().Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		mon.log.Errorf("Error in getting the cluster authentication type: %v", err)
+		return err
+	}
+
+	if cloudCredentialObject.Spec.CredentialsMode == operatorv1.CloudCredentialsModeManual {
+		mon.emitGauge(authenticationTypeMetricsTopic, 1, map[string]string{
+			"type": "managedIdentity",
```
wanghaoran1988, line 27: Could you help me understand why we can't just use the value "manual" here from the cloudCredentialObject spec?

slawande2: @wanghaoran1988 it was discussed in code review to use the cluster doc instead. I will mark this PR as draft for now.

PR #4225, opened 2025-05-14
cmd: allow loading and comparing RA-style config

stevekuznetsov:
A new initiative is moving ARO RP-Config to an Ev2 RA-style, which allows us to be much more succinct and makes new region build-out easier. This commit adds the capacity to load the new configuration file, if one is provided, and to error out if any difference is found. This behavior is configurable to allow a pipeline to be forced to run while a diff exists if there's an emergency.

A new command-line tool was added, as well, that diffs a directory of classic configurations with a directory of new-style ones, to ensure that there are no changes.

### DIFF
--- a/cmd/compareconfig.go
+++ b/cmd/compareconfig.go
@@ -1,0 +1,206 @@
+package main
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/yaml"
+
+	"github.com/Azure/ARO-RP/pkg/deploy"
+	"github.com/Azure/ARO-RP/pkg/util/cmp"
+)
+
+func main() {
+	opts := DefaultOptions()
+	cmd := &cobra.Command{
+		Use:          "overview",
+		Short:        "overview",
+		Long:         "overview",
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return compare(cmd.Context(), opts)
+		},
+	}
+	if err := BindOptions(opts, cmd); err != nil {
+		slog.Error("failed to bind options", "error", err)
+		os.Exit(1)
+	}
+
+	if err := cmd.Execute(); err != nil {
+		slog.Error("failed to compare configs", "error", err)
+		os.Exit(1)
+	}
+}
+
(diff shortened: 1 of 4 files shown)


kimorris27 (commented):
LGTM, but a few questions since I'm not that familiar with the new EV2 RA stuff:

- Is maintaining the deploy configuration in two formats/places a temporary measure while we work on this migration? I'm guessing yes but am double checking for sake of understanding.
- If the answer to the above is yes, is the main point of this work simply to ensure that we don't introduce regressions in the configuration mid-migration?
- Have you tested `cmd/compareconfig.go` locally to validate it?

stevekuznetsov:
> Is maintaining the deploy configuration in two formats/places a temporary measure while we work on this migration? 

Yes. RP-Config and `aro deploy` are extremely complex. I want to do this nice and slow without any sudden movements ;)

> is the main point of this work simply to ensure that we don't introduce regressions in the configuration mid-migration?

I'd word it as ensuring that the two configs are identical, but yes.

> Have you tested cmd/compareconfig.go locally to validate it?

Yep, the configs for every cloud/env/region are 100% identical as far as `deploy.GetConfig()` is concerned.

kimorris27 (approved):

tsatam (approved):

cmd/compareconfig.go
```diff
@@ -0,0 +1,206 @@
...
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/yaml"
+
+	"github.com/Azure/ARO-RP/pkg/deploy"
+	"github.com/Azure/ARO-RP/pkg/util/cmp"
+)
+
+func main() {
```
tsatam, line 22: nit: it might make sense to put this command in our `./hack` folder - as much as I don't like that becoming the kitchen sink, we may want to reserve the commands in `cmd` for our mission-critical ones. 

Not a huge deal either way though.

cmd/compareconfig.go
```diff
@@ -0,0 +1,206 @@
...
+	validated, err := opts.Validate()
+	if err != nil {
+		return err
+	}
+	completed, err := validated.Complete()
+	if err != nil {
+		return err
+	}
+	if diff := cmp.Diff(completed.ClassicConfig, completed.RAConfig); diff != "" {
+		fmt.Println(diff)
```
jharrington22, line 54: @stevekuznetsov you want to print this diff?

stevekuznetsov: Yeah, whoever is running this will need it to figure out what's wrong :)

stevekuznetsov:
After chatting with James to clarify all the bits here, going to pull this in to start getting CI into RP-Config.

PR #4228, opened 2025-05-19
[ARO-17460] Support zonal IPs in cluster creation

hawkowl:

Fixes ARO-17460

Initial pass at provisioning the PIPs with zonal support for zonal redundancy.

### Test plan for issue:

Will require manual testing as well as CI.

Yes, likely customer facing documentation.

Manual testing should suffice, and then the cluster install success rates.


hawkowl:
Manually verified that it deploys zone-redundant PIPs/FIPs.

PR #4233, opened 2025-05-20
Add dimensions to node metrics

tsatam:

Fixes [ARO-17473](https://issues.redhat.com/browse/ARO-17473)

Adds the following dimensions to our `node.conditions` metric:
- `role` - the node/machine role (e.g. `master`, `worker`, or `infra`)
- `machineset` - if present, the machineset containing the machine backing this node (will be set to `""` for control plane nodes)

This PR also adds the `role` dimension to the `node.kubelet.version` metric - we could add the machineset dimension as well but I did not immediately see a use-case for it. 

### Test plan for issue:

- [X] Unit tests were added/refactored for this monitor component
- [ ] E2E tests continue to pass
- [ ] Local run of the monitor against a testcluster 

No

- Above testing 
- The dimensions here are mostly additive - while this PR does refactor how the existing `spotInstance` dimension is retrieved, the new implementation should retrieve the value from the same location on the machine spec as before.

### DIFF
--- a/pkg/monitor/cluster/nodeconditions.go
+++ b/pkg/monitor/cluster/nodeconditions.go
@@ -12,8 +12,13 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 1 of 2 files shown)


ehvs (commented):
LGTM

PR #4234, opened 2025-05-21
Fix nil pointer dereference when hive ClusterMetadata is nil

ventifus:

Fixes ARO-17093

We observed a customer encountered a this panic in clienthelper.Merge. The "new" Hive ClusterDeployment.Spec.ClusterMetadata was nil. If the new ClusterMetadata is nil, we now preserve the entirety of the old ClusterMetadata.

### Test plan for issue:

Added unit tests that replicate the panic. These tests now pass.

No, bug fix.

Unit tests and e2e


kimorris27 (approved):

hlipsig (approved):

hlipsig:
Triggered CI again because the pipeline for the first run was cancelled.

PR #4236, opened 2025-05-21
Fix bounds error validating network profile when IngressProfiles is empty

ventifus:

Fixes ARO-17423

Some API versions check the visibility of IngressProfiles[0]. When IngressProfiles is empty this fails with a boundsError. Now we skip this validation when IngressProfiles is empty, because it will fail validation later with "There should be exactly one ingress profile."

### Test plan for issue:

Added unit tests to replicate the issue.

No, bug fix

Unit tests + e2e


pkg/api/v20250725/openshiftcluster_validatestatic.go
kimorris27, line 1: It looks like v20250725 is missing the new unit test cases.

kimorris27 (commented):
LGTM except that I think the unit test cases need to be added to the v20250725 API version.

kimorris27:
The previous E2E run failed with a zonal allocation error :/

PR #4237, opened 2025-05-21
Add the support for multiple role definition assignment per Platform Managed Identity

rajdeepc2792:

https://issues.redhat.com/browse/ARO-17517

There might be scenarios when a Platform Managed Identity might need multiple Role Definition assignment, currently the 
PWIRS is an array and can support addition of a roleDefinition for an operator. Ref: https://redhat-external.slack.com/archives/C06FRQKUMFF/p1743696979015069?thread_ts=1743696623.882759&cid=C06FRQKUMFF

Also, there might be scenarios, when same roleDefinition will be needed for different Managed Identities.

This PR adds the functionality to allow the role assignment as part of Create, Update and Deletion flow.

### Test plan for issue:

- [x ] Check MIWI cluster create flow with multiple roleDefinitions for a Managed Identity
- [x ] Check Update flow with multiple roleDefinitions for a Managed Identity
- [x ] Check Update(with UpgradeableTo) flow with multiple roleDefinitions for a Managed Identity
- [x ] Check Delete flow with multiple roleDefinitions for a Managed Identity
- [] Add Unit Test cases

NA

Tested the scenarios as per the test plan in local dev. Shouldn't impact the existing functionality.


pkg/cluster/deploybaseresources.go
```diff
@@ -519,34 +519,36 @@ func (m *manager) federateIdentityCredentials(ctx context.Context) error {
...
 		if !exists {
 			return platformworkloadidentity.GetPlatformWorkloadIdentityMismatchError(m.doc.OpenShiftCluster, platformWIRolesByRoleName)
 		}
 
-		for _, sa := range platformWIRole.ServiceAccounts {
-			federatedIdentityCredentialResourceName, err := m.getPlatformWorkloadIdentityFederatedCredName(sa, identity)
-			if err != nil {
-				return err
-			}
+		if len(platformWIRoles) > 0 {
```
kimorris27, line 527: I suggest removing this check for len > 0 because:

- `GetPlatformWorkloadIdentityRolesByRoleName`'s implementation guarantees that any slice present in the map will have a nonzero length.
- While it's true that checking the len theoretically prevents an out of bounds exception, it also introduces the possibility for this code to silently continue on without returning an error if the implementation of `GetPlatformWorkloadIdentityRolesByRoleName` ever changes such that the invariant in the first point no longer holds. This would be worse than an out of bounds exception because it would be much harder to find and debug the problem.

pkg/cluster/workloadidentityresources.go
```diff
@@ -85,50 +85,52 @@ func (m *manager) generatePlatformWorkloadIdentitySecretsAndNamespaces(isCreate
...
-		namespaces = append(namespaces, &corev1.Namespace{
-			TypeMeta: metav1.TypeMeta{
-				APIVersion: corev1.SchemeGroupVersion.Identifier(),
-				Kind:       "Namespace",
-			},
-			ObjectMeta: metav1.ObjectMeta{
-				Name: role.SecretLocation.Namespace,
-			},
-		})
+		if len(roles) > 0 {
```
kimorris27, line 104: I suggest removing this check for len > 0 because:

- `GetPlatformWorkloadIdentityRolesByRoleName`'s implementation guarantees that any slice present in the map will have a nonzero length.
- While it's true that checking the len theoretically prevents an out of bounds exception, it also introduces the possibility for this code to silently continue on without returning an error if the implementation of `GetPlatformWorkloadIdentityRolesByRoleName` ever changes such that the invariant in the first point no longer holds. This would be worse than an out of bounds exception because it would be much harder to find and debug the problem.

kimorris27 (dismissed):
LGTM, but suggesting one small change that would make a big difference in debugging effort level required for a potential future issue.

pkg/cluster/deploybaseresources.go
kevinobriendotca: @kimorris27 Do you mean remove the > 0 check in favour of checking for == 0 and logging a message or throwing an error?

Should the length check here be == 1 since we're hard coding the reference to the first entry in platformWIRoles?

Also, would it be useful to know if platformWIRoles was empty here and log a message? If a workload identity has no roles at this point, should that be an error?

kimorris27: Oops, I did mean the > 0 check.

We don't want a length check here, because we don't expect there to be exactly 1 entry. It's just that within this particular conditional we need to grab some data from the first entry that should actually be the same in all entries (tl;dr - it's a quirk of the MIWI API design).

rajdeepc2792: Created a new PR for the same: https://github.com/Azure/ARO-RP/pull/4452
This PR was merged so creating a separate PR to address the resolution.

PR #4240, opened 2025-05-22
Fluentbit latest image

rhamitarora:

Fixes

### DIFF
--- a/Dockerfile.fluentbit
+++ b/Dockerfile.fluentbit
@@ -1,5 +1,5 @@
 ARG MARINER_VERSION
-FROM mcr.microsoft.com/cbl-mariner/base/core:2.0.${MARINER_VERSION}-amd64 as builder
+FROM mcr.microsoft.com/cbl-mariner/base/core:2.0.${MARINER_VERSION}-amd64 AS builder
 RUN tdnf repolist --refresh
 RUN tdnf update -yq
 RUN tdnf install -yq ca-certificates flex build-essential openssl-devel systemd-devel cmake wget tar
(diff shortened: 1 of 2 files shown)


Dockerfile.fluentbit
```diff
@@ -1,132 +1,192 @@
 ARG MARINER_VERSION
-FROM mcr.microsoft.com/cbl-mariner/base/core:2.0.${MARINER_VERSION}-amd64 as builder
-RUN tdnf repolist --refresh
-RUN tdnf update -yq
-RUN tdnf install -yq ca-certificates flex build-essential openssl-devel systemd-devel cmake wget tar
+FROM mcr.microsoft.com/cbl-mariner/base/core:2.0.${MARINER_VERSION}-amd64 AS builder
+RUN tdnf repolist --refresh && tdnf update -yq
+RUN tdnf install -yq ca-certificates flex build-essential systemd-devel cmake wget tar c-ares-devel sqlite-devel openssl-devel zstd-devel nghttp2-devel
```
kimorris27, line 4: Are the new packages that have been added here dependencies of the new Fluentbit version?

rhamitarora: Yes, the newly added packages are required dependencies for the Fluentbit 4.0+ version. These changes align with the upstream requirements introduced in the newer release. We have already tested the updated Fluentbit image with these dependencies on a production server for approximately 15 days to a month without any issues.

rhamitarora:
> One other question. Reading over the Jira, it sounds like this task's goal was not necessarily to carry out the Fluentbit upgrade but to consider different ways to improve the upgrade process. In the end though, we ended up choosing to stick with our current upgrade process, so in this PR we're doing the upgrade as usual. Is that right?

You're right — the initial goal of the Jira was to explore and evaluate potential improvements to the Fluent Bit upgrade process. However, after considering the options, we decided to proceed with our current upgrade approach.

This PR specifically addresses the first part of the work: updating the existing Dockerfile to build the new Fluent Bit image (4.0+), including all required dependencies. The next step—rolling out this new image across all existing clusters—is being tracked separately under [ARO-14270](https://issues.redhat.com/browse/ARO-14270).

kimorris27 (approved):

PR #4243, opened 2025-05-26
Containerized development RP

aasserzo:

The idea is to streamline our cumbersome dev environment setup process.
I know efforts have been made to run a containerized RP for development and testing, but I couldn't find concrete documentation/usages.

This should help us quickly get started with a new development environment, using podman-compose

### Test plan for issue:

This works on my Fedora to run 'make runlocal-rp' within a container using podman-compose.
Please test this on Mac/Ubuntu and/or whatever other OS you are running and LMK.

There is containerized-dev-environment.md in this PR

NA


kimorris27 (commented):
We already have a `make` target for running a containerized local dev RP + VPN (`make run-rp`), and it's mentioned in the local dev setup docs as an option: https://github.com/Azure/ARO-RP/blob/master/docs/deploy-development-rp.md#run-the-rp-and-create-a-cluster

Note that it incorrectly says that you should run `openvpn` "manually" - `make run-rp` actually runs the VPN for you, so we should fix that part of the doc.

Sorry you put work into making your own, but do you want to try the existing one and see if it works for you?

aasserzo:
> We already have a `make` target for running a containerized local dev RP + VPN (`make run-rp`), and it's mentioned in the local dev setup docs as an option: https://github.com/Azure/ARO-RP/blob/master/docs/deploy-development-rp.md#run-the-rp-and-create-a-cluster
> 
> Note that it incorrectly says that you should run `openvpn` "manually" - `make run-rp` actually runs the VPN for you, so we should fix that part of the doc.
> 
> Sorry you put work into making your own, but do you want to try the existing one and see if it works for you?

@kimorris27 from what I can tell, `make run-rp` builds your code and runs the RP as a service, which makes sense for testing/pipelines. The container I propose in this PR provides an actual development environment. Your local repo is mounted into the workspace, requirements/tools are installed. I think using the existing "ci-rp" Dockerfile can be very useful, but does not replace your local dev environment, which is what I aim to do here. WDYT?

kimorris27:
> > We already have a `make` target for running a containerized local dev RP + VPN (`make run-rp`), and it's mentioned in the local dev setup docs as an option: https://github.com/Azure/ARO-RP/blob/master/docs/deploy-development-rp.md#run-the-rp-and-create-a-cluster
> > Note that it incorrectly says that you should run `openvpn` "manually" - `make run-rp` actually runs the VPN for you, so we should fix that part of the doc.
> > Sorry you put work into making your own, but do you want to try the existing one and see if it works for you?
> 
> @kimorris27 from what I can tell, `make run-rp` builds your code and runs the RP as a service, which makes sense for testing/pipelines. The container I propose in this PR provides an actual development environment. Your local repo is mounted into the workspace, requirements/tools are installed. I think using the existing "ci-rp" Dockerfile can be very useful, but does not replace your local dev environment, which is what I aim to do here. I've seen many issues with people trying to run wrong python/golang versions, using all kinds of platform-specific tools for Fedora/ubuntu/mac. I think the use of a dev container like this can solve that whole struggle. WDYT?

Oh, I see. I understand the distinction between what's in this PR vs. what we already have now.

I know this is not what everyone does, but I'm in the habit of using `make run-rp` for local development purposes. I like working that way because I know that I'm building and running the RP the same way we do elsewhere, and so I most likely won't find any surprises when I push my work and CI runs against it. It eliminates Go version problems since the RP runs in a container, and for Python version problems, I think the virtual environment setup we have in our instructions is a more lightweight way to solve those same problems.

With that in mind, I'm not against merging this, but I don't think I'd use it.

Dockerfile.dev-env
```diff
@@ -0,0 +1,64 @@
...
+
+# Install Python 3.10 using pyenv
+RUN curl https://pyenv.run | bash
+ENV PATH="/root/.pyenv/bin:${PATH}"
+RUN echo 'eval "$(pyenv init --path)"' >> ~/.bashrc && \
+    echo 'eval "$(pyenv init -)"' >> ~/.bashrc && \
+    echo 'eval "$(pyenv virtualenv-init -)"' >> ~/.bashrc
+
+# Install Python 3.10 and set up pip
+RUN pyenv install 3.10.0 && \
```
mrWinston: I would not specify the python minor release, so that we are always up to date on the bug and security fixes for python:

```
pyenv install 3.10
```

see [pyenv doc](https://github.com/pyenv/pyenv?tab=readme-ov-file#prefix-auto-resolution-to-the-latest-version)

docs/containerized-dev-environment.md
```diff
@@ -0,0 +1,117 @@
...
+    podman-compose logs aro-dev-env
+    ```
+
+6.  **Enter the container shell:**
+    To interact with the environment inside the container (e.g., run other commands, debug).
+
+    ```bash
+    podman-compose exec aro-dev-env bash
+    ```
+    *Note: The entrypoint script has already initialized pyenv, created and activated a Python virtual environment at `/root/pyenv`, and sourced environment variables from `/workspace/env`. You do not need to run `make pyenv` or manually activate the venv.*
```
mrWinston: This isn't the case unfortunately. The `ENTRYPOINT` doesn't get executed when you `exec` into the container:
```sh
# run container directly
$> podman-compose run aro-dev-env bash
#> python --version
Python 3.10.0

# in a separate terminal, while the container is still running, exec into the created contaienr
$> podman exec -it $(podman ps -l -q) bash 
#> python --version
Python 3.13.3
```

What do you think about converting the `dev-container-entrypoint.sh` to a `.bashrc` file instead and then use `/bin/bash` as the image's entrypoint?

Dockerfile.dev-env
```diff
@@ -0,0 +1,64 @@
...
+# Install Go 1.22.9
+RUN wget https://go.dev/dl/go1.22.9.linux-amd64.tar.gz && \
+    tar -C /usr/local -xzf go1.22.9.linux-amd64.tar.gz && \
+    rm go1.22.9.linux-amd64.tar.gz
+
+# Set up Go environment
+ENV PATH="/usr/local/go/bin:${PATH}"
+ENV GOPATH="/go"
+ENV PATH="${GOPATH}/bin:${PATH}"
+
```
mrWinston: I'm voting to add the go language server `gopls` to the image as well so i can get completion and everything in neovim from the dev container.

```sh
go install golang.org/x/tools/gopls@latest
```

docs/containerized-dev-environment.md
```diff
@@ -0,0 +1,117 @@
...
+The following files, located at the project root, are used for this setup:
+
+- `Dockerfile.dev-env`: Defines the container image with necessary dependencies and tools.
+- `dev-container-entrypoint.sh`: Script executed when the container starts to set up the environment (pyenv, venv, source env).
+- `docker-compose.yml`: Contains the definition for the `aro-dev-env` service.
+
+## Prerequisites
+
+1.  Podman installed on your host system ([https://podman.io/docs/installation](https://podman.io/docs/installation)).
+2.  Podman Compose installed on your host system.
```
bdematte: Makefile is already using compose but from "docker compose". 
We should align it and/or make it adaptive to what is available on the system when running to ease the use. 
Requiring both to be installed seems overkill and source of mistakes to me.

Dockerfile.dev-env
```diff
@@ -0,0 +1,64 @@
...
+    podman \
+    podman-docker \
+    git \
+    curl \
+    wget \
+    python3-pip \
+    which \
+    && dnf clean all
+
+# Install Go 1.22.9
```
bdematte: This is not working for Mac users (and potentially others depending on HW I guess).
Digging locally, best is probably to install go with dnf so that it is figuring out the right flavor to install.
Testing locally, it solves the issue.

Dockerfile.dev-env
aasserzo: Good point, changing

Dockerfile.dev-env
aasserzo: Currently this setup relied on your code being on your local disk, and it's mounted into the container. I don't think gopls on the container would be useful, since you're editing locally.

docs/containerized-dev-environment.md
aasserzo: I believe you could use either podman or docker (-compose). Are you saying this should be reflected in the doc?

Dockerfile.dev-env
aasserzo: Can you expand on this? What isn't working on Mac?

docs/containerized-dev-environment.md
aasserzo: Changed the doc

docs/containerized-dev-environment.md
aasserzo: I removed python setup altogether for now

Dockerfile.dev-env
mrWinston: likely because macs are arm-based, and you're explicitly installing the `amd64` version of golang.

docs/containerized-dev-environment.md
mrWinston: I'm curious why you removed python from the image. Having a separated python env inside the container would be a good way to avoid the issues that come with setting up old python versions on your local machine.

aasserzo: I believe make targets that run pyenv create their own python and pyenv setup either way

docs/containerized-dev-environment.md
```diff
@@ -0,0 +1,117 @@
...
+
+- `Dockerfile.dev-env`: Defines the container image with necessary dependencies and tools.
+- `dev-container-entrypoint.sh`: Script executed when the container starts to set up the environment (Go version, source env).
+- `docker-compose.yml`: Contains the definition for the `aro-dev-env` service.
+
+## Prerequisites
+
+1.  Docker installed on your host system ([https://docs.docker.com/get-docker/](https://docs.docker.com/get-docker/)).
+2.  Docker Compose plugin installed.
+3.  Azure CLI installed on your host system.
```
alcasim, line 27: As pre-requisite, for an entire "brand new" dev environment, I also needed to run this "prepare your environment" step, where RP database and MIWI setup is performed:

https://github.com/Azure/ARO-RP/blob/master/docs/deploy-development-rp.md#prepare-your-environment

Also, I already ran the entire [prepare your dev environment](https://github.com/Azure/ARO-RP/blob/master/docs/prepare-your-dev-environment.md) guide before trying this, so probably is also needed.

Dockerfile.dev-env
```diff
@@ -0,0 +1,54 @@
...
+RUN GO_VERSION=$(awk '/^go / {print $2}' /tmp/go.mod) && \
+    wget -q "https://go.dev/dl/go${GO_VERSION}.linux-amd64.tar.gz" -O /tmp/go.tar.gz && \
+    rm -rf /usr/local/go && \
+    tar -C /usr/local -xzf /tmp/go.tar.gz && \
+    rm /tmp/go.tar.gz && \
+    rm /tmp/go.mod
+ENV PATH="/usr/local/go/bin:${PATH}"
+
+# Install bingo and Go tools
+RUN /usr/local/go/bin/go install github.com/bwplotka/bingo@latest
```
alcasim, line 53: I think we also need `mock` here... `make test-go` fails because of it.

docs/containerized-dev-environment.md
```diff
@@ -0,0 +1,117 @@
...
+    docker compose exec aro-dev-env bash
+    ```
+    *Note: The entrypoint script has already ensured the correct Go version is installed and sourced environment variables from `/workspace/env`.*
+
+7.  **Run other development commands (Inside container shell):**
+    From inside the container, you can run project-specific `make` commands or scripts that expect the Go environment to be set up.
+
+    ```bash
+    # Example: Run tests
+    make test
```
alcasim: I think this should be `make test-go`

docs/containerized-dev-environment.md
aasserzo: Good point, I'll add these prerequisites

Dockerfile.dev-env
aasserzo: It fails regardless of whether I run it locally or in the container. I think this is a bingo problem more than a Dockerfile problem.
I will look into that, but for now resolving this comment, as it's related to a separate issue.

docker-compose.yml
```diff
@@ -195,6 +195,23 @@ services:
...
       ]
 
+  aro-dev-env:
+    build:
+      context: .
+      dockerfile: Dockerfile.dev-env
+    image: aro-rp_aro-dev
+    volumes:
+      - .:/workspace:Z
+      - ~/.azure:/root/.azure:ro,Z
```
tiguelu, line 205: Although it will run inside the container, wouldn't it be preferred if the RP doesn't run as root?

docs/containerized-dev-environment.md
```diff
@@ -0,0 +1,118 @@
+# Containerized Development Environment
+
+This document describes how to set up and use a containerized development environment for ARO-RP using Docker Compose.
+
+## Files for this setup
+
+The following files, located at the project root, are used for this setup:
+
+- `Dockerfile.dev-env`: Defines the container image with necessary dependencies and tools.
+- `dev-container-entrypoint.sh`: Script executed when the container starts to set up the environment (Go version, source env).
```
tiguelu: Wasn't this script removed?

docs/containerized-dev-environment.md
```diff
@@ -0,0 +1,118 @@
...
+    docker compose logs aro-dev-env
+    ```
+
+6.  **Enter the container shell:**
+    To interact with the environment inside the container (e.g., run other commands, debug).
+
+    ```bash
+    docker compose exec aro-dev-env bash
+    ```
+    *Note: The entrypoint script has already ensured the correct Go version is installed and sourced environment variables from `/workspace/env`.*
```
tiguelu: GO version is set in the Dockrefile and the env is sourced in the docker-compose.yml command.

tiguelu (changes requested):

docker-compose.yml
aasserzo: It would complicate setup quite a bit to run as non-root, and I don't think the benefits outweigh that

docs/containerized-dev-environment.md
aasserzo: Yes, this is outdated, will remove.

docs/containerized-dev-environment.md
aasserzo: Indeed, removing.

docker-compose.yml
tiguelu: IMO, the benefits would to avoid environmental drift (we don't run the RP as root anywhere else) and avoid running an app as root, even if it's in a container it's a setup we try to avoid.

However, being a dev container is not a blocker for merging and can be done as a follow-up. Resolving this conversation.

tiguelu (approved):

PR #4245, opened 2025-05-27
Bump to Go 1.23

stevekuznetsov:
*: automated Go version bumps

We can update all of the build images and Go version with a couple
commands:

sed -i 's/go 1.22.9/go 1.23.6/g' $( git grep -l 22 )
sed -i 's/partner-share:rhel-9-golang-1.22/partner-share:rhel-9-golang-1.23/g' $( git grep -l partner-share:rhel-9-golang-1.22 )

---

.bingo: bump golangci-lint to v2

It's best practice to use the most recent version of the linter with
newer versions of Go, as it's common to have older linters fail to parse
new Go syntax and OOM.

The v2 linter changed how they architect their configuration file, and
provided a `golangci-lint migrate` tool to move previous coinfigurations
over, which was used to do the updates here. No semantic differences are
expected.

---

.golangci.yml: bump Go version

It's not clear to me why this tool needs a separate source of truth for
the version of Go to be used - we should look into whether or not we can
drop this value entirely, I assume the tool can just look up the version
from go.mod.

---

Makefile: add a target to run the auto-fixer

The --fix argument to the linter will automatically update the AST where
possible to make the codebase conformant.

---

*: run make lint-go-fix

All of these changes are auto-generated.

---

validation: manually fix up complex validation

This validation was running afoul of the DeMorgan's law lint rule, and
for good reason - it's really complex, and there was even an error in
the boolean expression. It's just as easy to encode the intent here by
counting, so we prefer that.

---


stevekuznetsov:
Last commit is motivated by https://redhat-external.slack.com/archives/C03F6AA3HDH/p1748232892097209

stevekuznetsov:
[e2e failure](https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=125448133&view=logs&j=db690490-d817-5143-da35-29bc37ac10d9&t=7ab7f38e-816f-5a76-939c-0ca727e8db32&l=143) seems unrelated?

```
Attaching to run-e2e
Gracefully stopping... (press Ctrl+C again to force)
dependency failed to start: container aro-rp is unhealthy
```

Makefile
```diff
@@ -307,6 +307,10 @@ unit-test-go-coverpkg: $(GOTESTSUM)
 lint-go:
 	$(GOLANGCI_LINT) run --verbose
 
+.PHONY: lint-go-fix
+lint-go-fix:
```
cadenmarchese: Can we spin off this lint-go-fix target (and generated results) into their own PR? It's a lot of changes to review that go beyond just updating the Go version.

cadenmarchese (commented):
All of the version bumps LGTM. There are some changes here that are not necessarily related to the Go / golangci bump:
- gci
- new makefile target (and associated generated code changes)
Although I support both changes, I think having separate PRs and reviews for those would be preferred. 

The e2e failures are unrelated. I will kick CI again when we have that fixed!

Makefile
stevekuznetsov: Yes, but please don't review it. Not only can you not change the output but it's auto-generated based on the lint rules.

stevekuznetsov:
Broken out into this chain: https://github.com/Azure/ARO-RP/pull/4246 https://github.com/Azure/ARO-RP/pull/4247 https://github.com/Azure/ARO-RP/pull/4248 https://github.com/Azure/ARO-RP/pull/4249

PR #4246, opened 2025-05-29
Makefile: correctly depend on tools

stevekuznetsov:
The entire point of having per-tool Makefile targets is so that each top-level target that requires some tool can ensure that the tools it uses are installed as necessary. Having a top-level `install-tools` target is fine, if you need to do it, but using it exclusively loses the whole point of incremental installs and revalidations from Make.


cadenmarchese (approved):

stevekuznetsov:
Going to bypass E2E for expediency since this categorically cannot impact any functionality in the product.

PR #4247, opened 2025-05-29
Makefile: add a target to run the auto-fixer

stevekuznetsov:
Makefile: add a target to run the auto-fixer

The --fix argument to the linter will automatically update the AST where
possible to make the codebase conformant.

Signed-off-by: Steve Kuznetsov <stekuznetsov@microsoft.com>

---

PR #4248, opened 2025-05-29
bump to Go 1.23, golangci-lint to v2

stevekuznetsov:
.bingo: bump golangci-lint to v2

It's best practice to use the most recent version of the linter with
newer versions of Go, as it's common to have older linters fail to parse
new Go syntax and OOM.

The v2 linter changed how they architect their configuration file, and
provided a `golangci-lint migrate` tool to move previous coinfigurations
over, which was used to do the updates here. No semantic differences are
expected.

---

*: automated fixups from linter

This commit was generated with `make lint-go-fix`.

---

validation: manually fix up complex validation

This validation was running afoul of the DeMorgan's law lint rule, and
for good reason - it's really complex, and there was even an error in
the boolean expression. It's just as easy to encode the intent here by
counting, so we prefer that.

---


stevekuznetsov:
Depends on #4247

cadenmarchese (approved):

PR #4255, opened 2025-06-02
ARO-4556: make e2e admin update test concurrency-safe (retry on 409, …

shubhadapaithankar:
Fixes ADO [ARO-4556  ](https://issues.redhat.com/browse/ARO-4556)
(Concurrent e2e “admin update” calls were racing against each other, causing intermittent failures.)

1. **Make the e2e test safe from races**  
   - Before sending any PATCH …?admin=true, the test now waits until the cluster’s state is “Succeeded.”  
   - When the test issues the PATCH, it may get back a 409 Conflict if the cluster is already updating. In that case, it sleeps 5 seconds, re-reads the cluster (for a fresh ETag), and tries again—up to three times.  
   - If it still sees 409 after three attempts, the test explicitly fails with a clear message.  
   - Once a PATCH succeeds, the test immediately checks that `ProvisioningState == AdminUpdating` (and that `LastProvisioningState == Succeeded`), then waits until the cluster goes back to `Succeeded`, and finally confirms that `LastAdminUpdateError == ""`.  
   - As a result, two parallel test runners can’t both skip over “AdminUpdating.” One will see 409 and retry, so the test always observes the intermediate state.

2. **Prevent two concurrent admin-update calls in the RP code**  
   - In the RP handler for `PATCH …/openshiftclusters/{name}?admin=true`, we now acquire a lock per cluster ID before changing its state.  
   - We read the cluster’s current state:
     - If it is already “AdminUpdating,” we return HTTP 409 Conflict immediately.  
     - Otherwise, we set `ProvisioningState = AdminUpdating`, persist it, release the lock, start the background reconcile, and return HTTP 200.  
   - This guarantees that if two real clients—or two tests—call “admin update” at the same time, the first flips the state to “AdminUpdating” and the second immediately receives 409. Only after the first update finishes (and state goes back to `Succeeded`) can another call succeed.

These two changes together ensure:

- We have **eliminated the e2e test conflict** (no more “Succeeded → Succeeded” races).  
- We have **added runtime protection** so any concurrent admin-update call is either rejected or runs sequentially in a guaranteed state flow.

How we know this works:

- The modified e2e test always sees:
  1. Cluster state “Succeeded.”  
  2. PATCH (either 200 OK on first try or 409 → wait → 200).  
  3. Cluster state “AdminUpdating.”  
  4. Cluster state “Succeeded” again.  
  5. `LastAdminUpdateError == ""`.  

- If two PATCH requests arrive at exactly the same time, the first one sets the state to “AdminUpdating,” and the second immediately gets a 409 Conflict. When the first finishes, a new PATCH can succeed.

These two changes together 

- [ ] We’ve identified and eliminated test-side races.  
- [ ]  We’ve added runtime robustness so that any concurrent admin-update call is either rejected (409) or runs sequentially in a guaranteed state flow.

### DIFF
--- a/test/e2e/adminapi_cluster_update.go
+++ b/test/e2e/adminapi_cluster_update.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -19,6 +20,8 @@ var _ = Describe("[Admin API] Cluster admin update action", Serial, func() {
 
 	It("must run cluster update operation on a cluster", func(ctx context.Context) {
 		var oc = &admin.OpenShiftCluster{}
+		var resp *http.Response
+		var err error
 
 		// Wait for the cluster to be in a succeeded state before continuing
 		Eventually(func(g Gomega, ctx context.Context) {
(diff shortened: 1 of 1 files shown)


kimorris27:
This was a complicated issue...

After looking at the error logs in the Jira carefully, I don't think this PR is needed. Let me know what you think of these notes I put together outlining why.

### Explanation of the E2E failure:

- The failure occurred on 10/24/2023. Here's the line that the failure occurred on, on the commit used at the time: https://github.com/Azure/ARO-RP/blob/9ef2e0bc0ccf4975a033fec44960e93e830d8ce9/test/e2e/adminapi_cluster_update.go#L29
- The acceptance criteria given in the Jira is misleading. Note from the error logs that the cluster was actually in `Succeeded` at a point in time when the code wanted it to be in `AdminUpdating` (the opposite of what the acceptance criteria states).
	- Note verbiage in failure message and compare to the line of code linked above:
	  ```
	  [FAILED] Expected
          <admin.ProvisioningState>: Succeeded
	  to equal
          <admin.ProvisioningState>: AdminUpdating
      ```
- Additional details worth noting:
	- This E2E test we're looking at is the only one that performs an admin update, so even if E2E had been performing tests concurrently or in parallel, I don't think it would have been possible for E2E to admin update a cluster that's already in `AdminUpdating`.
	- Even if E2E *did* try to do an admin update on a cluster already in `AdminUpdating`, it would have errored out on [line 25](https://github.com/Azure/ARO-RP/blob/9ef2e0bc0ccf4975a033fec44960e93e830d8ce9/test/e2e/adminapi_cluster_update.go#L25) before making it to line 29. [`validateTerminalProvisioningState`](https://github.com/Azure/ARO-RP/blob/9ef2e0bc0ccf4975a033fec44960e93e830d8ce9/pkg/frontend/openshiftcluster_putorpatch.go#L96) in the frontend would have returned an error, which would have caused the test to error out before making it to the provisioning state check on line 29.
 
### Implications:

- I don't know how to explain the fact that the cluster was in `Succeeded` immediately after the admin PATCH returned a 200. My guess is that this failure occurred while work on admin update maintenance tasks was still in progress and was a true failure that was misinterpreted as a flake.
	- [PR history showing the work I'm referring to that was being done around this time](https://github.com/Azure/ARO-RP/pulls?q=is%3Apr+is%3Aclosed+maintenance+author%3Aniontive)
- What we do know is that it wasn't the case that concurrent admin update attempts were the cause of the failure.

### Conclusion

I'm thinking the error in the Jira was a true test failure that was misinterpreted as a flake. Either way, this PR won't address it. I'd suggest closing the Jira as unneeded unless we've been seeing reoccurrences of the same failure.

PR #4256, opened 2025-06-02
Set the identity and description to "Unkonw" when empty

gouthamMN:
Right now, the Otel Audit logs error `Portal - Error sending audit message: identity is required: validation error` whenever an identity or description is not set. We need to set these values to default "unkown" value when these values are not set or available from the incoming request.
Here is the [reference](https://github.com/microsoft/go-otel-audit/blob/cdfeea399bd85d6af31e8178c59a78e838e0aabc/audit/msgs/msgs.go#L468) to code that's handling validation from go-otel-audit

Set the values to default "unkown" value when these values are not set or available from the incoming request.

### Test plan for issue:
- existing E2E and unit tests

No


jhoreman (approved):
Makes sense, thanks for the quick turnaround

kimorris27 (approved):

PR #2314, opened 2022-08-08
Cleanup hack/clean.go

facchettos:
### Which issue this PR addresses:
some cleanup

PR #4262, opened 2025-06-06
Add a cleanup script to purge old subnets

hawkowl:

Something I revived as a part of [ARO-17865](https://issues.redhat.com/browse/ARO-17865)

Adds a script we can clean up subnets with to remove old NSGs/etc.

### DIFF
--- a/hack/subnet-clean/clean.go
+++ b/hack/subnet-clean/clean.go
@@ -1,0 +1,182 @@
+package main
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"flag"
+	"os"
+	"strings"
+	"time"
+
+	mgmtfeatures "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-07-01/features"
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/env"
+	utillog "github.com/Azure/ARO-RP/pkg/util/log"
+	"github.com/Azure/ARO-RP/pkg/util/purge"
+)
+
+// denylist exists as belt and braces protection for important RGs, even though
+// they may already have the persist=true tag set, especially if it is easy to
+// accidentally redeploy the RG without the persist=true tag set.
+var denylist = []string{
+	"v4-eastus",
+	"v4-australiasoutheast",
+	"v4-westeurope",
+	"management-westeurope",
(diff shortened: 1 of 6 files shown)


edisonLcardenas (approved):

hack/subnet-clean/clean.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	if time.Since(createdAt) < s.ttl {
+		log.Infof("Group %s is still less than TTL. SKIP.", *resourceGroup.Name)
+		return false
+	}
+
+	// TODO(mj): Fix this!
+	if contains(denylist, *resourceGroup.Name) {
+		return false
+	}
```
mociarain, line 179: This seems sus

hack/subnet-clean/clean.go
```diff
@@ -0,0 +1,182 @@
...
+	"v4-eastus",
+	"v4-australiasoutheast",
+	"v4-westeurope",
+	"management-westeurope",
+	"management-eastus",
+	"management-australiasoutheast",
+	"images",
+	"secrets",
+	"dns",
+}
```
mociarain, line 34: This list exists in clean.go now but maybe it didn't when this was first written. Can we update this to use that same list instead of duplicating it? Maybe it's time for a common utils in the 'hack' dir?

mociarain: In fact, a lot of this is now in 'hack/clean/clean.go'. Maybe this can be just folded into that?

mociarain (changes requested):
This looks good but I think we can DRY it out a little ;)

Also do we want some reference to this from the docs?

hack/subnet-clean/clean.go
edisonLcardenas: What does the TODO need to fix? 🤔

hawkowl: That's a good question! I copied it from clean.go :)

hack/subnet-clean/clean.go
hawkowl: Possibly -- I brought this back because one of our environments was buckling under the number of subnets, I'm not sure if we need it long-term, but since I've needed it a few times...

hack/subnet-clean/clean.go
mociarain: This is the PR where it was added for reference but there's no more information in there: https://github.com/Azure/ARO-RP/pull/2314

PR #4265, opened 2025-06-09
update validate_enable_managed_identity function to no longer check the --version argument

slawande2:
[ARO-4400](https://issues.redhat.com/browse/ARO-4400)

This is a small PR that updates `_validators.py`'s `validate_enable_managed_identity` function to no longer check the `–version` argument
During preview development, ARO still supported some OCP versions that were not compatible with MIWI. By the time the stable release of MIWI will roll out, ARO will only support MIWI-compatible OCP versions (4.14.38 or newer, 4.15.35 or newer, any 4.16 or newer y version).

I tested for platform workload identity replacement using `az aro update --assign-platform-workload-identity` and was able to replace a few identities.  

no


cadenmarchese (approved):
thanks for the cleanup!

kimorris27 (approved):

PR #4266, opened 2025-06-09
Prevent panic in Hive installer on retry

shubhadapaithankar:
**Description**
Sometimes Hive install will hit a transient error (e.g. a CosmosDB DNS hiccup), then on retry it panics because the service-principal secret’s `Data` map was never initialized. That panic crashes the entire RP.

This PR makes two small changes:

1. **Catch any panic** in runHiveInstaller (using a defer/recover) so it becomes a normal error return instead of a crash.
3. **Ensure the SP secret’s** `Data` **map is always non-nil** before writing to it, so we never hit a nil-pointer panic on retry.

With these in place, retries will fail cleanly and give you proper error logs instead of taking down the RP.

Fixes : [ARO-16309](https://issues.redhat.com/browse/ARO-16309)

- Wraps the Hive installer call in panic recovery so the RP won’t crash on unexpected errors.
- Initializes the SP secret’s `Data` map to avoid nil-map panics when Hive updates existing secrets on a retry.

### DIFF
--- a/pkg/backend/openshiftcluster.go
+++ b/pkg/backend/openshiftcluster.go
@@ -48,6 +48,8 @@ func newOpenShiftClusterBackend(b *backend) *openShiftClusterBackend {
 // succeeded in dequeuing anything - if this is false, the caller should sleep
 // before calling again
 func (ocb *openShiftClusterBackend) try(ctx context.Context, monitorDeleteWaitTimeSec int) (bool, error) {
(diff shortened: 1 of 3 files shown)


hawkowl (changes requested):
I think that https://github.com/Azure/ARO-RP/pull/4234 fixed the underlying error, and I think the Data code will be a no-op. I'm also not sure about catching panics elsewhere, I'll elaborate in-channel.

PR #4267, opened 2025-06-10
HiveManager NIL fix

rhamitarora:

[ARO-17801](https://issues.redhat.com/browse/ARO-17801)

This PR includes the addition of missing Hive-related variables required for the monitor container.
It also updates the `clustersync` monitor variable names to align with the column names in the service logs table.

### Test plan for issue:
Updated unit test cases and validated dashboard data by creating test clusters in the Canary environment.

Updated unit test cases and validated dashboard data by creating test clusters in the Canary environment.


pkg/hive/manager.go
```diff
@@ -87,7 +87,7 @@ func NewFromEnvCLusterManager(ctx context.Context, log *logrus.Entry, env env.In
 		return nil, err
 	}
 	if !adoptByHive && !installViaHive {
-		log.Infof("hive is disabled, skipping creation of ClusterManager")
+		log.Infof("hive is disabled, skipping creation of ClusterManager adoptByHive: %t installViaHive: %t", adoptByHive, installViaHive)
```
kimorris27: Logging the values of `adoptByHive` and `installViaHive` is unnecessary, so this change could be reverted. If we make it to this line of code, we already know both values.

pkg/monitor/worker.go
```diff
@@ -284,7 +284,15 @@ func (mon *monitor) workOne(ctx context.Context, log *logrus.Entry, doc *api.Ope
...
-	hiveClusterManager, _ := hive.NewFromConfigClusterManager(log, mon.env, hiveRestConfig)
+	hiveClusterManager, err := hive.NewFromEnvCLusterManager(ctx, log, mon.env)
+	if err != nil {
+		log.Error(err)
+		return
+	}
+
+	if hiveClusterManager == nil {
+		log.Info("skipping: no hive cluster manager")
+	}
```
kimorris27, line 295: What is this for? Since this block doesn't include an early return, it's not clear to me what we're skipping.

pkg/monitor/cluster/clustersync.go
```diff
@@ -13,57 +13,57 @@ import (
...
 func (mon *Monitor) emitClusterSync(ctx context.Context) error {
 	if mon.hiveClusterManager == nil {
-		// TODO(hive): remove this once we have Hive everywhere.
 		mon.log.Info("skipping: no hive cluster manager")
 		return nil
 	}
 
 	clusterSync, err := mon.hiveClusterManager.GetClusterSync(ctx, mon.doc)
 	if err != nil {
+		mon.log.Errorf("Error in getting the clustersync data %v", err)
```
kimorris27: A colon would be nice here to delineate our custom messaging from the underlying error message: 
```suggestion
		mon.log.Errorf("Error in getting the clustersync data: %v", err)
```

pkg/monitor/cluster/clustersync.go
```diff
@@ -13,57 +13,57 @@ import (
...
-			return nil
-		} else {
+	if clusterSync != nil {
+		if clusterSync.Status.SyncSets != nil {
 			for _, s := range clusterSync.Status.SyncSets {
 				mon.emitGauge("hive.clustersync", 1, map[string]string{
-					"syncType": "SyncSets",
-					"name":     s.Name,
-					"result":   string(s.Result),
+					"metric":  "hive.clustersync",
```
kimorris27: What do we need to add this dimension for if this is already the name of the metric we're emitting?

kimorris27 (commented):
I just left a few small suggestions and questions.

pkg/monitor/worker.go
rhamitarora: This is for logging purpose.

pkg/hive/manager.go
rhamitarora: done

pkg/monitor/cluster/clustersync.go
rhamitarora: done

pkg/monitor/cluster/clustersync.go
rhamitarora: done

pkg/monitor/cluster/clustersync.go
```diff
@@ -13,57 +14,54 @@ import (
...
 					mon.log.WithFields(logrus.Fields{
-						"syncType":           "SelectorSyncSets",
 						"name":               s.Name,
-						"result":             string(s.Result),
+						"status":             string(s.Result),
+						"type":               "SelectorSyncSets",
 						"firstSuccessTime":   timeToString(s.FirstSuccessTime),
 						"lastTransitionTime": timeToString(&s.LastTransitionTime),
-						"failureMessage":     s.FailureMessage,
+						"message":            cleanString(s.FailureMessage),
```
sankur-codes: LGTM. 
Only thing needed to be verified with a Canary deployment is that the message is being populated in required column of the logs table. It did not appear with other fields before this change.

pkg/monitor/cluster/clustersync.go
```diff
@@ -78,3 +76,8 @@ func timeToString(t *metav1.Time) string {
 	}
 	return t.String()
 }
+
+func cleanString(s string) string {
+	// remove special character and new line except _ and -
+	return regexp.MustCompile(`[^\w\-\n]`).ReplaceAllString(s, "")
```
pepedocs: MustCompile is better used in global vars because it panics when a regex is invalid, when a panic happens it will be detected right away during testing/development. Although I can see that the regex here is not invalid so there is low risk. I think a better use here is Compile where it returns an error and that error must be checked.

rhamitarora: done

pkg/monitor/cluster/clustersync.go
rhamitarora: column name updated from `message` to `response`

PR #4269, opened 2025-06-11
[ARO-17938] Check Gateway VMSS Instance Health Before Stopping Gateway Service

kevinobriendotca:

Fixes https://issues.redhat.com/browse/ARO-17938

Adding check for gateway VMSS instance health before running gateway stop service command.

### Test plan for issue:

Unit tests and E2E passing.

N/A

### DIFF
--- a/pkg/deploy/upgrade_gateway.go
+++ b/pkg/deploy/upgrade_gateway.go
@@ -79,20 +79,29 @@ func (d *deployer) gatewayRemoveOldScaleset(ctx context.Context, vmssName string
 
 	d.log.Printf("stopping scaleset %s", vmssName)
 	errors := make(chan error, len(scalesetVMs))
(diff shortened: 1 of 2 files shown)


pkg/deploy/upgrade_gateway.go
```diff
@@ -80,19 +80,26 @@ func (d *deployer) gatewayRemoveOldScaleset(ctx context.Context, vmssName string
 	d.log.Printf("stopping scaleset %s", vmssName)
 	errors := make(chan error, len(scalesetVMs))
 	for _, vm := range scalesetVMs {
```
tsatam, line 82: outside the scope of your PR, but: why are we performing these actions in parallel? I think we would want to limit the scope of undesired results by performing these actions in sequence (e.g. if operations fail against VMSS instance 0, we should _not_ proceed with VMSS instance 1). 

I think we should un-parallelize this step, but this can be a follow-up change.

PR #4271, opened 2025-06-11
Restore original NSG on Azure subnets and annotate reconciliation timestamp

shubhadapaithankar:
Restore original NSG on Azure subnets and annotate reconciliation timestamp

Fixes: [ARO-4848](https://issues.redhat.com/browse/ARO-4848)

- When the ARO Operator sees an out-of-band change to the subnet’s NSG (as the e2e test does), it now:

1. Retrieves the “correct” original NSG ID for master & worker subnets
3. Re-applies that NSG via CreateOrUpdate
5. Annotates the Cluster CR with `aro.openshift.io/subnet-reconciled-timestamp (RFC1123)`

- The e2e “Azure Subnet Reconciler” test waits on two conditions inside an Eventually:

1. That the subnet’s` .Properties.NetworkSecurityGroup.ID `equals the original cluster NSG
2. That our reconciliation timestamp annotation exists

- By restoring the NSG and writing the timestamp in one atomic reconcile pass, the test will now see both succeed well before the 5 minute timeout.
- Removed duplicate AnnotationTimestamp constants across the subnets package to resolve compiler redeclaration errors.

### DIFF
--- a/test/e2e/operator.go
+++ b/test/e2e/operator.go
@@ -442,15 +442,25 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
 			pods := ListK8sObjectWithRetry(
 				ctx,
 				clients.Kubernetes.CoreV1().Pods(aroOperatorNamespace).List,
-				metav1.ListOptions{LabelSelector: "app=openshift-azure-operator"},
+				metav1.ListOptions{
+					LabelSelector: "app=openshift-azure-operator",
+				},
 			)
 			for _, pod := range pods.Items {
-				log, err := clients.Kubernetes.CoreV1().
+				logBytes, err := clients.Kubernetes.CoreV1().
 					Pods(aroOperatorNamespace).
-					GetLogs(pod.Name, &corev1.PodLogOptions{SinceSeconds: pointerutils.ToPtr(int64(60))}).
+					GetLogs(
+						pod.Name,
+						&corev1.PodLogOptions{
+							SinceSeconds: pointerutils.ToPtr(int64(300)),
+						},
+					).
 					DoRaw(ctx)
 				Expect(err).NotTo(HaveOccurred())
-				fmt.Fprintf(GinkgoWriter, "=== operator logs from pod %s ===\n%s\n", pod.Name, string(log))
+				fmt.Fprintf(GinkgoWriter,
+					"=== operator logs from pod %s (last 5m) ===\n%s\n",
+					pod.Name, string(logBytes),
+				)
 			}
 
 		}


slawande2 (approved):

pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -53,18 +54,18 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
...
 	return r.updateReconcileSubnetAnnotation(ctx)
 }
 
+// updateReconcileSubnetAnnotation writes the current time into the cluster annotation.
 func (r *reconcileManager) updateReconcileSubnetAnnotation(ctx context.Context) error {
 	if r.instance.Annotations == nil {
 		r.instance.Annotations = make(map[string]string)
 	}
-	r.instance.Annotations[AnnotationTimestamp] = time.Now().Format(time.RFC1123)
+	r.instance.Annotations[AnnotationTimestamp] = time.Now().UTC().Format(time.RFC1123)
```
cadenmarchese, line 81: As discussed in a call, merging this wouldn't do any harm, however unless we're sure that using `time.Now().Format(time.RFC1123)` instead of `time.Now().UTC().Format(time.RFC1123)` is responsible for the flakes, this one might need some further investigation, since this is the only effective code change in this PR outside of changing the label used by the e2e test to indicate completion, which should be a transparent change.

shubhadapaithankar: I’ve pushed a small instrumentation commit that:

- Logs exactly what key/value we’re writing for aro.openshift.io/subnet-reconciled-timestamp
- Immediately time.Parse(time.RFC1123, …) it and warns if it fails

This will let us verify in the operator logs that we’re emitting a valid RFC1123 GMT timestamp before looking deeper

pepedocs (commented):
Hi @shubhadapaithankar Are we sure that this is not caused by a timeout issue? 

https://github.com/Azure/ARO-RP/blob/master/test/e2e/operator.go#L439

AFAICT, the subnet recon will be eventually consistent here, even though the test waited for the machineset CR change, right? So it may take more time than the default timeout.

OTOH, I don't mind more logs are added here to check for specific e2e failure. If we are not sure yet whether the timestamp is the cause of the issue, I would discourage adding code for that.

pkg/operator/controllers/subnets/subnet_nsg.go
shubhadapaithankar: I’ve added the instrumentation and confirmed in the operator logs that we’re writing exactly  and verified in the pipeline 
`aro.openshift.io/subnet-reconciled-timestamp="Wed, 18 Jun 2025 21:59:09 GMT"`
That key and value parse cleanly as RFC1123/GMT, so the annotation check is now fully satisfied alongside the NSG restore.

pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -53,18 +54,30 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
...
 	if r.instance.Annotations == nil {
 		r.instance.Annotations = make(map[string]string)
 	}
-	r.instance.Annotations[AnnotationTimestamp] = time.Now().Format(time.RFC1123)
-	return r.client.Update(ctx, r.instance)
+	// Generate RFC1123 GMT timestamp for the e2e test’s annotation check
+	ts := time.Now().UTC().Format(time.RFC1123)
+	r.log.Infof("Annotating Cluster CR with %s=%q", AnnotationTimestamp, ts)
+
+	if _, err := time.Parse(time.RFC1123, ts); err != nil {
```
cadenmarchese: Given that we're instantiating the time above on line 70, and the only change here is to now parse it and check for errors parsing it, I am not sure what the difference in behavior is here from what it is currently.

pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -43,7 +44,7 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
 
 	// if the NSG is assigned && it's the correct NSG - do nothing
```
cadenmarchese: This comment is no longer true, since we are now reconciling even if the correct NSG matches the current one, but per my other comment, I am not sure we need this change.

test/e2e/operator.go
```diff
@@ -431,12 +431,15 @@ var _ = Describe("ARO Operator - Azure Subnet Reconciler", func() {
...
+
 				g.Expect(*s.Properties.NetworkSecurityGroup.ID).To(Equal(*correctNSG))
 
 				co, err := clients.AROClusters.AroV1alpha1().Clusters().Get(ctx, "cluster", metav1.GetOptions{})
 				g.Expect(err).NotTo(HaveOccurred())
+				// 🚀 DEBUG: dump exactly what annotations we saw
+				By(fmt.Sprintf("Annotations on cluster: %v", co.Annotations))
 				g.Expect(co.Annotations).To(Satisfy(subnetReconciliationAnnotationExists))
-			}).WithContext(ctx).WithTimeout(DefaultEventuallyTimeout).Should(Succeed())
+			}).WithContext(ctx).WithTimeout(8 * time.Minute).Should(Succeed()) // give subnet reconciler extra headroom to avoid flakes
```
cadenmarchese: What is the justification around 8 versus 5 minutes, and do we have any evidence that supports the theory that operator reconciliation is taking longer than 5 minutes?

pkg/operator/controllers/subnets/subnet_nsg.go
```diff
@@ -43,7 +44,7 @@ func (r *reconcileManager) ensureSubnetNSG(ctx context.Context, s subnet.Subnet)
...
 	if subnetObject.NetworkSecurityGroup != nil && strings.EqualFold(*subnetObject.NetworkSecurityGroup.ID, correctNSGResourceID) {
-		return nil
+		return r.updateReconcileSubnetAnnotation(ctx)
```
cadenmarchese: This shouldn't be necessary, as in the context of the test, we are intentionally changing the NSG, and so this code path is not executed during the course of the test.

cadenmarchese (changes requested):
As discussed in code review, the only effective change here is to bump the wait timeout from 5 to 8 minutes. It's possible that the original issue was due to the operator not reconciling quickly enough, but I'm concerned that the timeout bump is a band-aid on an issue not fully understood yet.

IMO, we should further investigate before making any code changes to the operator or the test.

pkg/operator/controllers/subnets/subnet_nsg.go
shubhadapaithankar: Done—this now returns nil when the NSG is already correct, so we only annotate on real fixes.

test/e2e/operator.go
shubhadapaithankar: we’ve dropped the 8 min override and put the default back.

pkg/operator/controllers/subnets/subnet_nsg.go
shubhadapaithankar: We removed that instrumentation entirely, so there’s nothing left to parse

pkg/operator/controllers/subnets/subnet_nsg.go
shubhadapaithankar: We reverted the always-annotate branch, so that comment no longer applies.

PR #4272, opened 2025-06-12
ARO-17124 allow changing cluster identity

yithian:

Fixes ARO-17124

This PR adds functionality to the RP and the CLI that allows the customer to change the identity assigned to the cluster

### Test plan for issue:

Included a unit test for the updated putorpatch code and manually tested against a localdev RP with a dev cluster. After running the update via the CLI I verified in CosmosDB that the cluster's identity has been updated

Maybe? This may need to be documented in customer-facing documentation for ARO

This has been tested in localdev and will be tested again in canary


python/az/aro/azext_aro/custom.py
```diff
@@ -542,6 +546,12 @@ def aro_update(cmd,
             if client_id is not None:
                 oc_update.service_principal_profile.client_id = client_id
 
+    if mi_user_assigned is not None:
```
kimorris27, line 554: To catch the error at the CLI if a cx passes this argument to a CSP cluster update, we could do something similar to: https://github.com/Azure/ARO-RP/blob/c8ed71db9892e125622ddd3cc348e960f0b02478/python/az/aro/azext_aro/custom.py#L527-L530

pkg/api/admin/openshiftcluster.go
```diff
@@ -474,13 +474,13 @@ type ManagedServiceIdentity struct {
...
-	PrincipalID string `json:"principalId,omitempty" swagger:"readOnly"`
+	PrincipalID string `json:"principalId,omitempty" swagger:"readOnly" mutable:"true"`
 
 	// The TenantID provided by the MSI RP
-	TenantID string `json:"tenantId,omitempty" swagger:"readOnly"`
+	TenantID string `json:"tenantId,omitempty" swagger:"readOnly" mutable:"true"`
 
 	// A map of user assigned identities attached to the cluster, specified in a type required by Microsoft's Managed Identity team.
-	UserAssignedIdentities map[string]UserAssignedIdentity `json:"userAssignedIdentities,omitempty"`
+	UserAssignedIdentities map[string]UserAssignedIdentity `json:"userAssignedIdentities,omitempty" mutable:"true"`
```
kimorris27: I think we can skip making these attributes mutable on the admin API.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
```diff
@@ -516,10 +516,6 @@ func (sv openShiftClusterStaticValidator) validatePlatformIdentities(oc *OpenShi
 		return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "identity", "Cluster identity and platform workload identities require each other.")
 	}
 
-	if clusterIdentityPresent && len(oc.Identity.UserAssignedIdentities) != 1 {
```
kimorris27: What's the intention behind removing this validation?

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -279,23 +309,31 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
...
+			return nil, err
+		}
+	} else {
+		setUpdateProvisioningState(doc, putOrPatchClusterParameters.apiVersion)
+	}
+
+	// Persist identity URL and tenant ID for managed/workload identity clusters
+	// We do support updating cluster managed identity after cluster creation
+	if doc.OpenShiftCluster.UsesWorkloadIdentity() {
+		// PATCHes can have an empty identityURL
```
kimorris27: If we're going to call this out in a comment, we might as well be extra clear:

```suggestion
		// A PATCH will only have a non-empty identityURL if the PATCH includes a new cluster identity
```

Same suggestion for the tenant ID comment a few lines down.

kimorris27 (commented):
Looks great! Just a few small suggestions and questions.

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
yithian: When testing the CLI changes, I was getting an error from this validation, I think because [the number of UserAssignedIdentities is briefly 0](https://github.com/Azure/ARO-RP/pull/4272/files#diff-c0dceb05e2a59300f2e11380f3601ea682171f3f4837bb769689fc79e78b1a70R231). I'll re-add the validation locally and confirm that the CLI is still tripping over it, since I ran into that issue before making some other changes to the RP logic

cadenmarchese: Can we confirm as part of this change that it's not possible to null out the identity fields using an update? I'm also concerned that we may need this validation (or a new version of it) to ensure that there's always a cluster identity present.

pkg/frontend/openshiftcluster_putorpatch.go
```diff
@@ -208,6 +209,35 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
...
+
+			// if the patch parameters contain a new identity, remove the old
+			// identity from the cluster doc so that the new one will be the
+			// only identity
+			if tempPatchParameters.Identity != nil && len(tempPatchParameters.Identity.UserAssignedIdentities) > 0 {
+				doc.OpenShiftCluster.Identity.UserAssignedIdentities = map[string]api.UserAssignedIdentity{}
+				for k, v := range tempPatchParameters.Identity.UserAssignedIdentities {
+					doc.OpenShiftCluster.Identity.UserAssignedIdentities[k] = api.UserAssignedIdentity{
+						ClientID:    v.ClientID,
+						PrincipalID: v.PrincipalID,
```
cadenmarchese: Won't these fields always be empty until the rest of the update flow is ran? If so, would it make more sense to make that clearer for the reader:
```suggestion
						// client and principal ID are updated as part of the update/adminUpdate flow, and not via put or patch
						ClientID:    "",
						PrincipalID: "".
```
Or something like that

pkg/api/v20240812preview/openshiftcluster_validatestatic.go
yithian: I was mistaken in my testing, I've re-added this validation

yithian:
@cadenmarchese @kimorris27 I think I've addressed your feedback and concerns and have manually confirmed that these changes successfully update the identity for the cluster. Would you mind re-reviewing?

pkg/cluster/clustermsi.go
```diff
@@ -26,19 +26,48 @@ var (
...
+			if m.isEligibleForRenewal(existingMsiCertificate) {
+				needsNewCert = true
 			}
 		}
 	} else if !azureerrors.IsNotFoundError(err) {
 		return err
+	} else {
+		// ie: if there's an error getting the secret but the error isn't a IsNotFoundError
+		needsNewCert = true
+	}
```
cadenmarchese: We didn't handle this case before, but we may want to return the error here (and unit test if possible) rather than swallow it. At the very least we should log the error before proceeding.

pkg/cluster/clustermsi_test.go
```diff
@@ -220,6 +244,31 @@ func TestEnsureClusterMsiCertificate(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+			msiDataplaneStub: func(client *mock_msidataplane.MockClient) {
+				client.EXPECT().GetUserAssignedIdentitiesCredentials(gomock.Any(), gomock.Any()).Return(alternateCredentialsObject, nil)
+			},
+			kvClientMocks: func(kvclient *mock_azsecrets.MockClient) {
+				kvclient.EXPECT().GetSecret(gomock.Any(), secretName, "", nil).Return(azsecrets.GetSecretResponse{}, secretNotFoundError).Times(1)
```
cadenmarchese: When replacing an identity, a secret should be in KV already, so there should be no 404. From my read of the test, I would expect this instead, indicating that we got a secret and also set a secret:
```
				kvclient.EXPECT().GetSecret(gomock.Any(), secretName, "", nil).Return(getSecretResponse, nil).Times(1)
				kvclient.EXPECT().SetSecret(gomock.Any(), secretName, gomock.Any(), nil).Return(azsecrets.SetSecretResponse{}, nil).Times(1)
```

kimorris27 (commented):
I gave another review, and I have nothing to add except that I agree with both of Caden's suggestions.

yithian:
Canary testing was successful

```
[achvatal@xoth ~]$ az aro list -g achvtest2 -o json | jq -r ".[0].identity.userAssignedIdentities"
{
  "/subscriptions/60bf318d-6914-4105-a3b5-d0d2c10388c8/resourceGroups/achvtest2/providers/Microsoft.ManagedIdentity/userAssignedIdentities/aro-Cluster": {
    "clientId": "4fd93196-c9c4-4a19-961f-9ec019275912",
    "principalId": "10f9dae8-9a58-49e9-8478-b63bcb6e66ed"
  }
}
[achvatal@xoth ~]$ az aro update -g achvtest2 -n achvtest --assign-cluster-identity /subscriptions/60bf318d-6914-4105-a3b5-d0d2c10388c8/resourcegroups/achvtest2/providers/Microsoft.ManagedIdentity/userAssignedIdentities/aro-Cluster2
Argument '--assign-cluster-identity' is in preview and under development. Reference and support levels: https://aka.ms/CLI_refstatus
Location     Name      ProvisioningState    ResourceGroup
-----------  --------  -------------------  ---------------
eastus2euap  achvtest  Succeeded            achvtest2
[achvatal@xoth ~]$ az aro list -g achvtest2 -o json | jq -r ".[0].identity.userAssignedIdentities"
{
  "/subscriptions/60bf318d-6914-4105-a3b5-d0d2c10388c8/resourcegroups/achvtest2/providers/Microsoft.ManagedIdentity/userAssignedIdentities/aro-Cluster2": {
    "clientId": "154b4393-1557-4d06-8287-681f61456fc2",
    "principalId": "f9f35efb-91e5-4326-b421-42fce12486c2"
  }
}```

shubhadapaithankar (approved):
lgtm

PR #4277, opened 2025-06-17
fix rp-development template with missing param

Tof1973:

During investigation of [ARO-14166](https://issues.redhat.com/browse/ARO-14166) we saw that rp-development template was broken with a missing defined parameter which was used in the template.

This PR will fix the broken PR adding the param in the template and also pass the param from deploy-shared-env functions

### Test plan for issue:

Run the deploy_rp_dev with new param

No, this does not change way deploy_rp_dev function is called

Not used in production.

### DIFF
--- a/pkg/deploy/generator/templates_rp.go
+++ b/pkg/deploy/generator/templates_rp.go
@@ -29,6 +29,7 @@ func (g *generator) rpTemplate() *arm.Template {
 		"databaseAccountName",
 		"fpServicePrincipalId",
 		"rpServicePrincipalId",
+		"globalDevopsServicePrincipalId",
 	}
 	if g.production {
 		params = append(params,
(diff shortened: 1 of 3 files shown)


shubhadapaithankar (approved):
LGTM

pkg/deploy/generator/templates_rp.go
```diff
@@ -29,6 +29,7 @@ func (g *generator) rpTemplate() *arm.Template {
 		"databaseAccountName",
 		"fpServicePrincipalId",
 		"rpServicePrincipalId",
+		"globalDevopsServicePrincipalId",
```
tiguelu, line 32: `globalDevopsServicePrincipalId` is appended later to `params` in line 61 when this code runs in Production, which raises two questions for me:

1. What is the effect of adding the param twice to the list?
2. Now that this param is added for the dev template, could it be removed from the prod specific appends and left as a global param?

Tof1973: Indeed you are right. This param was already included in the production list, and it is required in all cases.
So, as it is now added to global list, I removed it now from production appends, as it is useless.
Note, that this template_rp.go is used by the generator to create development and production code and files generated for production did not change with this PR (as you can see in list of changed files), that is why I mention this is not used in Production.
BTW, as you can see, removing globalDevopsServicePrincipalId from production appends, did not change anything to generated files, so it was just useless...

tiguelu (approved):
LGTM

PR #4280, opened 2025-06-19
upgrade to gatekeeper v3.19.2

yjst2012:

Fixes ARO-18152

to address CVE in gatekeeper v3.19.1

### Test plan for issue:

tested on our shared cluster

NA

tested on a prod cluster (except for the gatekeeper image source)


cadenmarchese (approved):
makes sense.

PR #4281, opened 2025-06-19
[ARO-17795] Replace multiple pointer utils with the current one we're using anyway

hawkowl:

Nominally something raised as a part of [ARO-17795] that I decided to fix everywhere.

Replaces disparate pointerutils/to.Ptr code with consistent usage from the non-deprecated MSFT SDK we're using anyway.

### Test plan for issue:

Unit tests, E2E

N/A

Test coverage, type checking


jaitaiwan (approved):
LGTM Pending Tests

jaitaiwan (approved):
LGTM pending pipeline

PR #4287, opened 2025-06-20
fix: skip leftover cluster check in release E2E

shubhadapaithankar:
Fixes
This PR updates the \`setup()\` logic in \`setup.go\` to only run the leftover cluster check if:

- We're running in CI (\`conf.IsCI\`)
- And it's a local development environment (\`_env.IsLocalDevelopmentMode()\`)

In release E2E runs, the leftover cluster check was causing a panic due to a nil pointer dereference. This fix ensures the check only runs in environments where it's relevant.

### How to Test

- Run E2E in release pipeline — the leftover cluster check should be skipped.
- Run E2E in CI — the check should still run as usual.

### DIFF
--- a/test/e2e/setup.go
+++ b/test/e2e/setup.go
@@ -504,7 +504,8 @@ func setup(ctx context.Context) error {
 	azOCClient := redhatopenshift20240812preview.NewOpenShiftClustersClient(
 		_env.Environment(), _env.SubscriptionID(), authAdapter)
 
-	if conf.IsCI {
+	// Only check for leftover clusters in local dev CI, not in release E2E
+	if _env.IsLocalDevelopmentMode() && conf.IsCI {
 		const (
 			maxRetries  = 10
 			waitBetween = 30 * time.Second


test/e2e/setup.go
```diff
@@ -504,7 +504,8 @@ func setup(ctx context.Context) error {
 	azOCClient := redhatopenshift20240812preview.NewOpenShiftClustersClient(
 		_env.Environment(), _env.SubscriptionID(), authAdapter)
 
-	if conf.IsCI {
+	// Only check for leftover clusters in local dev CI, not in release E2E
+	if _env.IsLocalDevelopmentMode() && conf.IsCI {
```
tsatam: nit: `conf` should have its own IsLocalDevelopmentMode() definition, that we should use just to be consistent with where we're pulling `conf.IsCI` from: https://github.com/Azure/ARO-RP/blob/45ff164e78f0fd410dd3ad378119831fab0a6d1b/pkg/util/cluster/cluster.go#L81

tsatam (dismissed):
LGTM, one nit/suggestion but it doesn't need to block merge, fix it if you'd like.

slawande2 (dismissed):

test/e2e/setup.go
shubhadapaithankar: Good catch — updated to use conf.IsLocalDevelopmentMode() as suggested. Thanks!

cadenmarchese (approved):
lgtm

PR #4289, opened 2025-06-25
ARO-9309: Geneva Action to Renew MDSD Certificate

edisonLcardenas:

Create geneva action for renewing mdsd certificate. - https://issues.redhat.com/browse/ARO-9309 

Geneva action ["Renew MDSD Certificate" ](https://msazure.visualstudio.com/One/_git/Compute-ARO-GenevaActions/pullrequest/12527787) calls the logic on this PR to renew mdsd certificates.
Basically, calls MIMO underneath rather than the SRE having to have the UUID for the MIMO task handy.

### Test plan for issue:

Unit test on admin_mimo_manifest_create_mdsdcertificaterenew_test.go: Passed
Local Geneva Action [Testing](https://issues.redhat.com/browse/ARO-9309): Passed 

Yes, added this to [ARO.wiki](https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/802215/-Not-yet-published-Renew-MDSD-Certificate).

Tested this using local RP and geneva action in a Windows VM. Test result is pinned on the comments found on the jira above.


pkg/frontend/admin_mimo_manifest_create.go
```diff
@@ -61,9 +56,19 @@ func (f *frontend) _putAdminMaintManifestCreate(ctx context.Context, r *http.Req
 	}
 
 	var ext *admin.MaintenanceManifest
-	err = json.Unmarshal(body, &ext)
-	if err != nil {
-		return nil, api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidRequestContent, "", "The request content could not be deserialized: "+err.Error())
+
+	if maintenanceTaskID != "" {
+		ext = &admin.MaintenanceManifest{}
```
pepedocs: Why not specify MaintenanceTaskID as param here?

pkg/frontend/admin_mimo_manifest_create_mdsdcertificaterenew.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+	"github.com/Azure/ARO-RP/pkg/mimo"
+)
+
+func (f *frontend) putAdminMaintManifestMdsdCertificateRenew(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
```
pepedocs: I have seen this use case everywhere, but this looks like this can panic if log is nil.

I think it should be something like the following;

logIface := ctx.Value(middleware.ContextKeyLog)
log, ok := logIface.(*logrus.Entry)
if !ok {
}

pepedocs (changes requested):
Please have a look at my comments and make changes as necessary.

pkg/frontend/admin_mimo_manifest_create.go
edisonLcardenas: > Why not specify MaintenanceTaskID as param here?
It is actually a parameter, but I pass the argument  through pkg/frontend/admin_mimo_manifest_create_mdsdcertificaterenew.go.

pepedocs: I meant like the following so you don't need the next line;
```
&admin.MaintenanceManifest{
  MaintenanceTaskID: maintenanceTaskID
}
```

pepedocs (commented):
/lgtm for me but I'd like for another person to review this esp the mimo part.

pepedocs (approved):

PR #4290, opened 2025-06-25
Skip leftover-cluster cleanup in release E2E and handle ARM 404 gracefully

shubhadapaithankar:
Fixes a panicking E2E in our release pipeline when the “leftover cluster” resource group doesn’t exist (404). In release  runs we shouldn’t even try to clean up old clusters, and in PR runs we need to treat 404 as “no cluster to delete” rather than crashing the test.

- Skip the leftover-cluster cleanup entirely in non-PR (release) E2E runs by gating on a real PR trigger `(BUILD_REASON=="PullRequest" or SYSTEM_PULLREQUEST_PULLREQUESTID).
`

- Replace the brittle `strings.Contains(err.Error(), "not found")` check with an` errors.As(err, *azcore.ResponseError) + StatusCode==404, so genuine 404s` simply break out of the retry loop instead of panicking.
- 

### Test plan for issue:

- PR build: trainer creates a fake leftover cluster → retry logic runs → sees 404 → logs and proceeds.
- Release build: no PR ID present → cleanup block is skipped → fresh cluster is always created.

-->

### DIFF
--- a/test/e2e/setup.go
+++ b/test/e2e/setup.go
@@ -509,12 +509,10 @@ func setup(ctx context.Context) error {
 
 	// Determine if this is a PR build:
 	azPR := os.Getenv("SYSTEM_PULLREQUEST_PULLREQUESTID")
-	buildReason := os.Getenv("BUILD_REASON")
-	sourceBranch := os.Getenv("BUILD_SOURCEBRANCH")
-	isPR := azPR != "" || buildReason == "PullRequest" || strings.HasPrefix(sourceBranch, "refs/pull/")
+	log.Infof("PR build? SYSTEM_PULLREQUEST_PULLREQUESTID=%q", azPR)
 
 	// Only run leftover-cluster logic on PR-based E2E (skip in release builds)
-	if conf.IsLocalDevelopmentMode() && conf.IsCI && isPR {
+	if conf.IsLocalDevelopmentMode() && conf.IsCI && azPR != "" {
 		const (
 			maxRetries  = 10
 			waitBetween = 30 * time.Second


test/e2e/setup.go
```diff
@@ -502,47 +504,61 @@ func setup(ctx context.Context) error {
...
 	authAdapter := azidext.NewTokenCredentialAdapter(tokenCred, scopes)
 	azOCClient := redhatopenshift20240812preview.NewOpenShiftClustersClient(
-		_env.Environment(), _env.SubscriptionID(), authAdapter)
+		_env.Environment(), _env.SubscriptionID(), authAdapter,
+	)
 
-	// Only check for leftover clusters in local dev CI, not in release E2E
-	if conf.IsLocalDevelopmentMode() && conf.IsCI {
+	// Only run leftover-cluster logic on PR-based E2E (skip in release builds)
+	if conf.IsLocalDevelopmentMode() && conf.IsCI && os.Getenv("SYSTEM_PULLREQUEST_PULLREQUESTID") != "" {
```
cadenmarchese:
```
os.Getenv("SYSTEM_PULLREQUEST_PULLREQUESTID")
```
Where / how is this env var being set?

shubhadapaithankar: SYSTEM_PULLREQUEST_PULLREQUESTID is a built-in Azure Pipelines variable. Whenever a PR kicks off a build, Azure DevOps automatically sets System.PullRequest.PullRequestId (exposed in the agent as SYSTEM_PULLREQUEST_PULLREQUESTID) to PR number. we don’t need to define it.

Here’s the docs:
https://learn.microsoft.com/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#system-pullrequest-pullrequestid

cadenmarchese: Looking at those docs, that var appears to only be set if the build ran because of a [Git PR affected by a branch policy](https://learn.microsoft.com/en-us/azure/devops/repos/git/branch-policies?view=azure-devops#build-validation), so it seems like there'd be some additional work to configure this.

Looking at e2e for this PR confirms that it isn't being set: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=128419144&view=logs&j=f8772640-e625-5559-7f97-12eb89cfbf13&t=258a7ff9-252e-50fd-3747-bb12c2c202fe&l=157

Is there another way we could determine if it's a PR build?

shubhadapaithankar: I’ve updated the logic to use` BUILD_REASON` and `BUILD_SOURCEBRANCH` to detect PR builds, so leftover cleanup only runs on PRs. We now always call `cluster.Create()`, so release E2E will spin up a new cluster and won’t hit the 404 panic

test/e2e/setup.go
```diff
@@ -502,47 +504,66 @@ func setup(ctx context.Context) error {
...
+
+	// Determine if this is a PR build:
+	buildReason := os.Getenv("BUILD_REASON")
+	sourceBranch := os.Getenv("BUILD_SOURCEBRANCH")
+	isPR := buildReason == "PullRequest" || strings.HasPrefix(sourceBranch, "refs/pull/")
 
-	// Only check for leftover clusters in local dev CI, not in release E2E
-	if conf.IsLocalDevelopmentMode() && conf.IsCI {
+	// Only run leftover-cluster logic on PR-based E2E (skip in release builds)
+	if conf.IsLocalDevelopmentMode() && conf.IsCI && isPR {
```
cadenmarchese: In the e2e run for this change, I see the following log message:
```
run-e2e  | time="2025-06-26T21:35:15Z" level=info msg="Skipping leftover-cluster cleanup because this is not a PR build" func="e2e.setup()" file="test/e2e/setup.go:557"
```
This indicates to me that `isPR` is getting set to `false` here, when we need it to be `true`.

shubhadapaithankar: Hey Caden – I’ve updated our PR to detect PR builds by checking

```
SYSTEM_PULLREQUEST_PULLREQUESTID
BUILD_REASON == "PullRequest"
BUILD_SOURCEBRANCH starts with refs/pull/
```

That logic now drives isPR, so on a real PR build it will be true and we’ll run the leftover‐cluster cleanup. I’ve kicked off a fresh E2E run against this branch

cadenmarchese:
I think we're OK to close this PR now, as the issue we were hitting in the Role Sets pipeline was the result of the e2e binary not including [this](https://github.com/Azure/ARO-RP/pull/4287) change. Since that PR seems to have fixed the leftover check issue, this one isn't needed.

Long term, we should make sure that our ev2 pipelines always use the same RP tag for both deploy and e2e stages, rather than pulling e2e images from the RP versions storage account. That issue can be handled separately, since it's an ARO-Pipelines change.

Thank you for your assistance in debugging this!

PR #4297, opened 2025-07-01
Adhere to Microsoft/Azure security baseline in the portal SSH server

ventifus:

Fixes ARO-18937

The Microsoft security baseline advises
```
Appropriate ciphers should be used for SSH. (Ciphers aes128-ctr,aes192-ctr,aes256-ctr)
(113)
```

### Test plan for issue:

Manual testing of the portal in Canary is needed. E2E tests were attempted, but did not work and a spike is needed to see if it is possible to end-to-end test SSH to a cluster via the portal.

The problem is well understood. Testing will be performed prior to the next release.

### DIFF
--- a/pkg/portal/ssh/ssh.go
+++ b/pkg/portal/ssh/ssh.go
@@ -211,32 +211,23 @@ func (s *SSH) internalServerError(w http.ResponseWriter, err error) {
 	http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 }
 
-// as of 30 Jun 2025 / go 1.24.4, this server supports the following algorithms
+// as of 30 Jun 2025 / go 1.24.4 / PR , this server supports the following algorithms
 //
 // $ nmap --script ssh2-enum-algos localhost -p 2222
-// Starting Nmap 7.92 ( https://nmap.org ) at 2025-06-30 16:17 PDT
+// Starting Nmap 7.92 ( https://nmap.org ) at 2025-08-21 09:08 PDT
 // Nmap scan report for localhost (127.0.0.1)
-// Host is up (0.00021s latency).
-// Other addresses for localhost (not scanned): ::1
-
-// PORT     STATE SERVICE
-// Starting Nmap 7.92 ( https://nmap.org ) at 2025-07-01 09:55 PDT
-// Nmap scan report for localhost (127.0.0.1)
-// Host is up (0.00021s latency).
+// Host is up (0.00020s latency).
 // Other addresses for localhost (not scanned): ::1
 
 // PORT     STATE SERVICE
 // 2222/tcp open  EtherNetIP-1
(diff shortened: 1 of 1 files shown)


kimorris27 (approved):

rajdeepc2792 (approved):

shubhadapaithankar (dismissed):

pkg/portal/ssh/proxy_test.go
```diff
@@ -90,7 +90,13 @@ func fakeServer(clientKey *rsa.PublicKey) (*listener.Listener, error) {
...
 			return nil, nil
 		},
-		Config: cryptossh.Config{},
+		Config: cryptossh.Config{
+			Ciphers: []string{
+				cryptossh.CipherAES128CTR,
+				cryptossh.CipherAES192CTR,
+				cryptossh.CipherAES256CTR,
+			},
+		},
```
s-fairchild, line 97: Since we are using the same values in tests, and in `pkg/portal/ssh/ssh.go`, can we create a function that returns our `Ciphers` string array? This way we can call the get crypto ciphers helper function in both places.

Or a const `[]string{}` that can be used by both, I'm unsure which is ideal.

pkg/portal/ssh/ssh.go
```diff
@@ -78,7 +78,17 @@ func New(env env.Core,
...
-		baseServerConfig: &cryptossh.ServerConfig{},
+		baseServerConfig: &cryptossh.ServerConfig{
+			Config: cryptossh.Config{
+				// Per security baseline requirements,
+				// https://learn.microsoft.com/en-us/azure/governance/policy/samples/guest-configuration-baseline-linux
+				Ciphers: []string{
+					cryptossh.CipherAES128CTR,
+					cryptossh.CipherAES192CTR,
+					cryptossh.CipherAES256CTR,
+				},
```
s-fairchild:
```suggestion
		                },
				KeyExchanges: []string{
					cryptossh.KeyAlgoED25519,
					cryptossh.KeyAlgoRSA,
					cryptossh.KeyAlgoRSASHA256,
					cryptossh.KeyAlgoRSASHA512,
					cryptossh.KeyAlgoECDSA256,
					cryptossh.KeyAlgoECDSA384,
					cryptossh.KeyAlgoECDSA521,
				},
```

Should we also set the KexAlgorithms here too? I think it will still try and use the default curve25519-sha256 otherwise.

kimorris27 (commented):
LGTM, but do we have a test plan?

s-fairchild (dismissed):
LGTM. Thanks for the additional changes.

s-fairchild:
> LGTM, but do we have a test plan?

I think an E2E test would cover this?  
ssh from `aro-portal` to the E2E cluster.  

I don't think we spin up the portal during E2E tests though..

PR #4299, opened 2025-07-03
[ARO-18931] Implement zone-redundancy migration code for internal frontend IPs

hawkowl:

Fixes #ARO-18931, mostly

Adds the code to migrate internal LB frontend IPs to be zonally redundant. It adds it as a special admin update "strategy", rather than taking place in a normal PUCM.

### Test plan for issue:

Unit tests. Unsure if there's a viable e2e method for this.

Likely a deployment plan once we decide.

Unit tests should cover the code, but we will need to do some more manual testing.

### DIFF
--- a/pkg/api/admin/openshiftcluster.go
+++ b/pkg/api/admin/openshiftcluster.go
@@ -136,6 +136,10 @@ var validMaintenanceTasks = []MaintenanceTask{
 	MaintenanceTaskRenewCerts,
 	MaintenanceTaskSyncClusterObject,
 	MaintenanceTaskMigrateLoadBalancer,
+	// internal maintenance state signals
+	MaintenanceTaskPending,
+	MaintenanceTaskNone,
+	MaintenanceTaskCustomerActionNeeded,
 }
 
 // Operator feature flags


pkg/cluster/ipaddresses.go
```diff
@@ -220,8 +220,18 @@ func (m *manager) populateDatabaseIntIP(ctx context.Context) error {
 		return err
 	}
 
+	frontendIPs := map[string]string{}
+	for _, fip := range lb.Properties.FrontendIPConfigurations {
+		frontendIPs[*fip.Name] = *fip.Properties.PrivateIPAddress
```
edisonLcardenas: Issue: The code dereferences pointers without null checks.
Risk: This can cause panics if fip.Name or fip.Properties.PrivateIPAddress is nil.

pkg/cluster/loadbalancerinternal.go
```diff
@@ -0,0 +1,159 @@
...
+	lbZones := []*string{}
+	for _, z := range controlPlaneZones {
+		lbZones = append(lbZones, pointerutils.ToPtr(z))
+	}
+
+	// Update the document with the internal LB zones
+	updatedDoc, err := m.db.PatchWithLease(ctx, m.doc.Key, func(oscd *api.OpenShiftClusterDocument) error {
+		oscd.OpenShiftCluster.Properties.NetworkProfile.LoadBalancerProfile.Zones = controlPlaneZones
+		return nil
+	})
```
edisonLcardenas: Issue: If the load balancer update fails after this database patch, there's no rollback mechanism. This could leave the cluster in an inconsistent state.

hawkowl: Good catch, I thought I already fixed this lol

pkg/api/admin/openshiftcluster_validatestatic.go
```diff
@@ -43,6 +43,7 @@ func validateMaintenanceTask(task MaintenanceTask) error {
 		task != MaintenanceTaskPending &&
 		task != MaintenanceTaskNone &&
 		task != MaintenanceTaskSyncClusterObject &&
+		task != MaintenanceTaskMigrateLoadBalancer &&
```
wanghaoran1988: This can be refactored like this?
```
func validateMaintenanceTask(task MaintenanceTask) error {
    allowedTasks := []MaintenanceTask{
        MaintenanceTaskPending,
        MaintenanceTaskNone,
        MaintenanceTaskSyncClusterObject,
        MaintenanceTaskMigrateLoadBalancer,
        MaintenanceTaskCustomerActionNeeded,
    }

    for _, t := range allowedTasks {
        if t == task {
            return nil
        }
    }
}
```

pkg/api/openshiftcluster.go
```diff
@@ -241,10 +241,11 @@ const (
...
 
-	MaintenanceTaskEverything        MaintenanceTask = "Everything"
-	MaintenanceTaskOperator          MaintenanceTask = "OperatorUpdate"
-	MaintenanceTaskRenewCerts        MaintenanceTask = "CertificatesRenewal"
-	MaintenanceTaskSyncClusterObject MaintenanceTask = "SyncClusterObject"
+	MaintenanceTaskEverything          MaintenanceTask = "Everything"
+	MaintenanceTaskOperator            MaintenanceTask = "OperatorUpdate"
+	MaintenanceTaskRenewCerts          MaintenanceTask = "CertificatesRenewal"
+	MaintenanceTaskSyncClusterObject   MaintenanceTask = "SyncClusterObject"
+	MaintenanceTaskMigrateLoadBalancer MaintenanceTask = "MigrateLoadBalancer"
```
wanghaoran1988, line 248: These are duplicates with the ones in pkg/api/admin/openshiftcluster.go, can we move to a central place so can be referenced in other places?

pkg/cluster/loadbalancerinternal.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	lbZones := []*string{}
+	for _, z := range controlPlaneZones {
+		lbZones = append(lbZones, pointerutils.ToPtr(z))
+	}
+
+	ilbBackendPoolID := fmt.Sprintf("%s/backendAddressPools/%s", *lb.ID, infraID)
+	if m.doc.OpenShiftCluster.Properties.ArchitectureVersion == api.ArchitectureVersionV1 {
+		ilbBackendPoolID = ilbBackendPoolID + "-internal-controlplane-v4"
```
wanghaoran1988, line 72: Will this exceed the maximum character limit of 80?

pkg/api/openshiftcluster.go
hawkowl: The different versions (`pkg/api/<version>` as well as `pkg/api/admin`) are deliberately duplicated so that our "external" APIs don't automatically reflect our internal one, even though most of the time it's passthru.

pkg/cluster/loadbalancerinternal.go
hawkowl: I don't think it should, considering the name should already exist in this case

pkg/api/admin/openshiftcluster_validatestatic.go
hawkowl: Fixed (slightly differently)

PR #4300, opened 2025-07-03
[ARO-19061] Clean up python to use not-removed functions to fix CI

hawkowl:

Fixes https://issues.redhat.com/browse/ARO-19061

Fixes up our Python client to use not-removed functions and updates the autorest so that everything works. Also removes most of the clients that we don't use.

### Test plan for issue:

Unit tests

N/A

N/A


Makefile
```diff
@@ -160,7 +168,7 @@ image-aro-multistage:
 
 .PHONY: image-autorest
 image-autorest:
-	docker build --platform=linux/amd64 --network=host --no-cache --build-arg AUTOREST_VERSION="${AUTOREST_VERSION}" --build-arg REGISTRY=$(REGISTRY) --build-arg BUILDER_REGISTRY=$(BUILDER_REGISTRY) -f Dockerfile.autorest -t ${AUTOREST_IMAGE} .
+	docker build --platform=linux/amd64 --network=host --no-cache --build-arg AUTOREST_VERSION="${AUTOREST_VERSION}" --build-arg REGISTRY=$(REGISTRY) -f Dockerfile.autorest -t ${AUTOREST_IMAGE} .
```
ehvs, line 168: Why are we removing BUILDER_REGISTRY=$(BUILDER_REGISTRY) here?

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/Azure/go-autorest/autorest v0.11.29
 	github.com/Azure/go-autorest/autorest/adal v0.9.23
 	github.com/Azure/go-autorest/autorest/date v0.3.0
+	github.com/Azure/go-autorest/autorest/to v0.4.0
```
ehvs, line 26: Since we are deprecating autorest/to in favor of pointerutils on .golangci.yml , can we use the same for this?

Makefile
hawkowl: We don't use it in the dockerfile (and locally it breaks saying it references itself?)

go.mod
hawkowl: This is autorest generated code which I'd rather not mess with, so I just added an exception for it in the linter.

PR #4302, opened 2025-07-07
Make OIDC Issuer URL readOnly & update PutOrPatch unit test cases

rajdeepc2792:

Fixes https://issues.redhat.com/browse/ARO-17783

- Make OIDCIssuer to ReadOnly as it is blocking the updates done via ARM Templates
- Update Put/Patch Test Cases to use the static validator

### Test plan for issue:
- [x] Update Unit Test Cases
- [x] Perform a PUT Request that doesn't pass OIDC Issuer URL
- [x] Perform a PATCH Request that doesn't pass OIDC Issuer URL as unchanged
- [x] PATCH/PUT Request with unchanged OIDC Issuer URL would still fail with error, it is similar behavior as other readOnly fields

None

Tested in Local

### DIFF
--- a/pkg/frontend/openshiftcluster_putorpatch_test.go
+++ b/pkg/frontend/openshiftcluster_putorpatch_test.go
@@ -1062,6 +1062,38 @@ func TestPutorPatchOpenShiftClusterUpdatePut(t *testing.T) {
 				return response
 			},
 		},
+		{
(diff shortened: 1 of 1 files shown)


pepedocs (approved):
\lgtm

kimorris27 (approved):

kimorris27:
Stuck a hold label on this to make sure it doesn't get merged before local dev testing is done.

PR #4305, opened 2025-07-09
turn on guardrails by default

yjst2012:

Fixes ARO-19693

turn on guardrails by default as per [discussion](https://redhat-internal.slack.com/archives/C02ULBRS68M/p1751928306847779)

### Test plan for issue:

tested on some clusters

customer facing doc will be updated by ARO-16213

it's been tested on some clusters


pkg/operator/flags.go
```diff
@@ -72,11 +83,19 @@ func DefaultOperatorFlags() map[string]string {
...
+		GuardrailsEnabled:                  FlagTrue,
+		GuardrailsDeployManaged:            FlagTrue,
 		CloudProviderConfigEnabled:         FlagTrue,
 		ForceReconciliation:                FlagFalse,
 		EtcHostsEnabled:                    FlagTrue,
 		EtcHostsManaged:                    FlagTrue,
+
+		// Guardrails policies switches
+		GuardrailsPolicyMachineDenyManaged:           FlagTrue,
+		GuardrailsPolicyMachineDenyEnforcement:       GuardrailsPolicyDeny,
```
ArrisLee, line 95: so we are going to ONLY enable the machine deny policy for the next release?

ArrisLee: nwm, found the description in the JIRA

ArrisLee (approved):
LGTM

pkg/operator/flags.go
yjst2012: yes, correct

PR #4311, opened 2025-07-15
[ARO-19967] Fix E2E Release Pipeline skipping cluster creation

mrWinston:

Fixes ARO-19967

- Release E2E Pipeline is failing because cluster creation is skipped.
- This PR moves the cluster creation logic in the e2e setup so that it's not only run in local development mode.

### Test plan for issue:

- E2E Tests

- No, this restores expected behavior

- N/A


tiguelu (approved):
LGTM. Waiting on E2E check to complete.

PR #4317, opened 2025-07-17
LB probe deletion in GA "deleted managed resource"

alcasim:

Fixes [ARO-13789](https://issues.redhat.com/browse/ARO-13789)

Admin actions trying to delete Load Balancer health probes fail as they can't be deleted with the `DeleteByIDAndWait` call. This PR addresses this problem, adding some logic to prevent deleting health probes used by a backend.

### Test plan for issue:

Added unit tests

It's an extra case over an existing function. Local cluster tests worked as expected.


pkg/frontend/adminactions/delete_managedresource.go
```diff
@@ -52,6 +53,14 @@ func (a *azureActions) ResourceDeleteAndWait(ctx context.Context, resourceID str
 		return a.deleteFrontendIPConfiguration(ctx, resourceID)
 	}
 
+	re = regexp.MustCompile(healthProbePattern)
+	// HealthProbes cannot be deleted with DeleteByIDAndWait either.
+	matches := re.FindStringSubmatch(resourceID)
```
ehvs: Can we give `matches` a more descriptive name?

pkg/frontend/adminactions/delete_managedresource.go
```diff
@@ -52,6 +53,14 @@ func (a *azureActions) ResourceDeleteAndWait(ctx context.Context, resourceID str
...
+	if len(matches) > 0 {
+		loadBalancerName := matches[3]
+		return a.deleteHealthProbe(ctx, resourceID, loadBalancerName)
+	}
+
 	return a.resources.DeleteByIDAndWait(ctx, resourceID, apiVersion)
```
ehvs, line 58: I see 'resources' has been used before, but can we make more descriptive to on this file?

pkg/util/loadbalancer/loadbalancer.go
```diff
@@ -28,3 +31,20 @@ func RemoveFrontendIPConfiguration(lb *armnetwork.LoadBalancer, resourceID strin
...
 }
+
+func RemoveHealthProbe(lb *armnetwork.LoadBalancer, resourceID string) error {
+	newProbes := make([]*armnetwork.Probe, 0)
+
+	// Iterate over probes, build a list without the targeted resourceID, if possible
+	for _, probe := range lb.Properties.Probes {
+		if strings.EqualFold(*probe.ID, resourceID) {
+			if probe.Properties.LoadBalancingRules != nil {
+				return api.NewCloudError(http.StatusBadRequest, api.CloudErrorCodeInvalidParameter, "", fmt.Sprintf("probe %s is used by load balancing rules, remove the referencing load balancing rules before removing the probe", resourceID))
```
ehvs: Small nit: Start sentences with caps letter "Probe", and perhaps use the full name of the resource:"Health Probe %s ...   removing the Health Probe"

pkg/frontend/adminactions/delete_managedresource.go
```diff
@@ -52,6 +53,14 @@ func (a *azureActions) ResourceDeleteAndWait(ctx context.Context, resourceID str
 		return a.deleteFrontendIPConfiguration(ctx, resourceID)
 	}
 
+	re = regexp.MustCompile(healthProbePattern)
```
mrWinston: Mhhh, This is the first time i'm looking at exactly this file, and honestly, i would be in favor replacing the manual regex approach and replace it with the actual azure-sdk way of parsing the resource ID and getting its parts. `github.com/Azure/go-autorest/autorest/azure.ParseResourceID()` is officially deprecated and actually doesn't support parsing subresources. The replacement for this is `github.com/Azure/azure-sdk-for-go/sdk/azcore/arm.ParseResourceID()`.So instead of the regex things, you

mrWinston: Mhhh, This is the first time i'm looking at exactly this file, and honestly, i would be in favor replacing the manual regex approach and replace it with the actual azure-sdk way of parsing the resource ID and getting its parts. `github.com/Azure/go-autorest/autorest/azure.ParseResourceID()` is officially deprecated and actually doesn't support parsing subresources. The replacement for this is `github.com/Azure/azure-sdk-for-go/sdk/azcore/arm.ParseResourceID()`.So instead of the regex things, you can do something like:

```go
idParts, err := arm.ParseResourceID(resourceID)
```

mrWinston: Mhhh, This is the first time i'm looking at exactly this file, and honestly, i would be in favor replacing the manual regex approach and replace it with the actual azure-sdk way of parsing the resource ID and getting its parts. `github.com/Azure/go-autorest/autorest/azure.ParseResourceID()` is officially deprecated and actually doesn't support parsing subresources. The replacement for this is `github.com/Azure/azure-sdk-for-go/sdk/azcore/arm.ParseResourceID()`.So instead of the regex things, you can do something like:

```go
idParts, err := arm.ParseResourceID(resourceID)

if idParts.ResourceType == "Microsoft.Network/loadBalancers/probes" {
	return a.deleteHealthProbe(ctx, resourceID, idParts.Parent.Name)
}
```

Also, you wouldn't need to pass the Load Balancer Name to `deleteHealthProbe` when using `arm.ParseResourceID()`, cause it also extracts the correctly handles parsing of the parent resource name.

pkg/frontend/adminactions/delete_managedresource.go
alcasim: hey, I'm not sure about the `resources` you mean... is it about the contents of `matches`? I reckon that I can add a comment about what's inside (just the group matches from the regex),

pkg/util/loadbalancer/loadbalancer.go
```diff
@@ -28,3 +31,20 @@ func RemoveFrontendIPConfiguration(lb *armnetwork.LoadBalancer, resourceID strin
 func isFrontendIPConfigReferenced(fipConfig *armnetwork.FrontendIPConfiguration) bool {
 	return fipConfig.Properties.LoadBalancingRules != nil || fipConfig.Properties.InboundNatPools != nil || fipConfig.Properties.InboundNatRules != nil || fipConfig.Properties.OutboundRules != nil
 }
+
+func RemoveHealthProbe(lb *armnetwork.LoadBalancer, resourceID string) error {
+	newProbes := make([]*armnetwork.Probe, 0)
+
+	// Iterate over probes, build a list without the targeted resourceID, if possible
+	for _, probe := range lb.Properties.Probes {
+		if strings.EqualFold(*probe.ID, resourceID) {
```
mrWinston: we might need a nil-check here, incase `probe.ID` is actually unset. I don't know why that would be the case, but I find dereferencing pointers without explicitly checking for null always very scary.

pkg/util/loadbalancer/loadbalancer_test.go
```diff
@@ -127,3 +129,163 @@ func TestRemoveLoadBalancerFrontendIPConfiguration(t *testing.T) {
...
+									},
+									{
+										ID: pointerutils.ToPtr("ae3506385907e44eba9ef9bf76eac973-TCP-443"),
+									},
+								},
+							},
+						},
+					},
+				},
+				Name:     pointerutils.ToPtr(infraID),
```
mrWinston, line 272: I'm in favor of using `&infraID` instead of `pointerutils.ToPtr`. This keeps the code more idiomatic and easier to understand, imho.

pkg/frontend/adminactions/delete_managedresource.go
alcasim: I was tempted to change how the entire logic worked, as I wasn't happy with the regex approach either; but given than the opinion is to do the things right, I'll just do that. Thanks for the tips! :)

pkg/frontend/adminactions/delete_managedresource.go
alcasim: I ended up removing matches all together

pkg/util/loadbalancer/loadbalancer.go
```diff
@@ -28,3 +31,24 @@ func RemoveFrontendIPConfiguration(lb *armnetwork.LoadBalancer, resourceID strin
 func isFrontendIPConfigReferenced(fipConfig *armnetwork.FrontendIPConfiguration) bool {
 	return fipConfig.Properties.LoadBalancingRules != nil || fipConfig.Properties.InboundNatPools != nil || fipConfig.Properties.InboundNatRules != nil || fipConfig.Properties.OutboundRules != nil
 }
+
+func RemoveHealthProbe(lb *armnetwork.LoadBalancer, resourceID string) error {
+	newProbes := make([]*armnetwork.Probe, 0)
+
+	if lb.Properties.Probes != nil {
```
mrWinston: Sorry, i could have been clearer what I meant with the null check.
Checking `if lb.Properties.Probes != nil` here doesn't actually do anything. That is becuase the `for`-loop that follows handles the nil value gracefully and just doesn't go into the loop at all. so the `range` keyword already has that nil-check builtin.

What I meant to suggest is to add a nil-check to
```
if strings.EqualFold(*probe.ID, resourceID) {
```
so it looks like:

```
if probe.ID != nil && strings.EqualFold(*probe.ID, resourceID) {
...
```

mrWinston (changes requested):
Added a clarification on my part.

pkg/util/loadbalancer/loadbalancer.go
alcasim: That should be fixed now (I added the same check to `RemoveFrontendIPConfiguration` for consistency too).

PR #4326, opened 2025-07-24
removing vpn dependency for E2E tests that run from local development…

alcasim:
… on non-hive clusters

Fixes https://issues.redhat.com/browse/ARO-20079

This PR skips Hive E2E client initialization and some Hive E2E tests, when the tests are locally run on non-Hive clusters.

### Test plan for issue:

I ran the E2E tests locally for both Hive and non-Hive clusters, assessing that when the non-Hive E2E tests are executed, the expected tests are run and no VPN is needed to run them.

This change affects E2E tests, under specific conditions (local runs) so it should be a noop, from production's perspective.


tiguelu (approved):
LGTM. Waiting on E2E to confirm PR CI runs all the Hive tests.

PR #4330, opened 2025-07-31
[ARO-20197] Reduce memory usage of the Monitor

hawkowl:

Fixes ARO-20197

Reduces the memory usage of the monitor by avoiding querying more than we need, and by querying in smaller batches.

### Test plan for issue:
Unit tests, E2E, INT?

N/A

Unsure :)

### DIFF
--- a/pkg/monitor/cluster/cluster.go
+++ b/pkg/monitor/cluster/cluster.go
@@ -35,6 +35,7 @@ import (
 	"github.com/Azure/ARO-RP/pkg/monitor/monitoring"
 	arov1alpha1 "github.com/Azure/ARO-RP/pkg/operator/apis/aro.openshift.io/v1alpha1"
 	aroclient "github.com/Azure/ARO-RP/pkg/operator/clientset/versioned"
+	"github.com/Azure/ARO-RP/pkg/util/clienthelper"
 	"github.com/Azure/ARO-RP/pkg/util/steps"
 )
 
@@ -58,7 +59,7 @@ type Monitor struct {
 	env         env.Interface
 	tenantID    string
 
-	ocpclientset  client.Client
+	ocpclientset  clienthelper.Interface
 	hiveclientset client.Client
 
 	// access below only via the helper functions in cache.go
@@ -73,6 +74,10 @@ type Monitor struct {
 	wg                 *sync.WaitGroup
 	hiveClusterManager hive.ClusterManager
 	doc                *api.OpenShiftClusterDocument
+
+	// Namespaces that are OpenShift or ARO managed that we want to monitor
+	namespacesToMonitor []string
+	queryLimit          int
 }
 
 func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftCluster, doc *api.OpenShiftClusterDocument, env env.Interface, tenantID string, m metrics.Emitter, hiveRestConfig *rest.Config, hourlyRun bool, wg *sync.WaitGroup, hiveClusterManager hive.ClusterManager) (*Monitor, error) {
@@ -142,21 +147,23 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
 		oc:   oc,
 		dims: dims,
 
-		restconfig:         restConfig,
-		cli:                cli,
-		configcli:          configcli,
-		operatorcli:        operatorcli,
-		maocli:             maocli,
-		mcocli:             mcocli,
-		arocli:             arocli,
-		env:                env,
-		tenantID:           tenantID,
-		m:                  m,
-		ocpclientset:       ocpclientset,
-		hiveclientset:      hiveclientset,
-		wg:                 wg,
-		hiveClusterManager: hiveClusterManager,
-		doc:                doc,
+		restconfig:          restConfig,
+		cli:                 cli,
+		configcli:           configcli,
+		operatorcli:         operatorcli,
+		maocli:              maocli,
+		mcocli:              mcocli,
+		arocli:              arocli,
+		env:                 env,
+		tenantID:            tenantID,
+		m:                   m,
+		ocpclientset:        clienthelper.NewWithClient(log, ocpclientset),
+		hiveclientset:       hiveclientset,
+		wg:                  wg,
+		hiveClusterManager:  hiveClusterManager,
+		doc:                 doc,
+		namespacesToMonitor: []string{},
+		queryLimit:          50,
 	}, nil
 }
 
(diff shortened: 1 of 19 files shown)


pkg/monitor/cluster/managednamespaces.go
```diff
@@ -0,0 +1,43 @@
...
+	l := &corev1.NamespaceList{}
+
+	for {
+		err := mon.ocpclientset.List(ctx, l, client.Continue(cont), client.Limit(mon.queryLimit))
+		if err != nil {
+			return fmt.Errorf("error in list operation: %w", err)
+		}
+
+		for _, i := range l.Items {
+			if namespace.IsOpenShiftNamespace(i.GetName()) {
```
aasserzo, line 30: Is the list in IsOpenShiftNamespace() exhaustive? Should we also monitor 'kube-' namespaces? Are there potentially namespaces created by other operators, that aren't in the list? I wonder if there is a 'smarter' way to filter out customer workloads than a static list.

pkg/monitor/cluster/cluster.go
```diff
@@ -142,21 +149,23 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
...
+		arocli:              arocli,
+		env:                 env,
+		tenantID:            tenantID,
+		m:                   m,
+		ocpclientset:        clienthelper.NewWithClient(log, ocpclientset),
+		hiveclientset:       hiveclientset,
+		wg:                  wg,
+		hiveClusterManager:  hiveClusterManager,
+		doc:                 doc,
+		namespacesToMonitor: []string{},
```
aasserzo, line 167: Can this value be tested/benchmarked outside of production somehow?

pkg/monitor/cluster/deploymentstatuses.go
```diff
@@ -5,46 +5,44 @@ package cluster
...
-		cont = ds.Continue
-		if cont == "" {
-			break
+			cont = l.Continue
+			if cont == "" {
+				break
+			}
 		}
 	}
-	mon.emitGauge("deployment.count", count, nil)
```
aasserzo, line 48: What is the impact of removing these count gauges? Are there monitors that rely on them and need to be removed? Are they important and require an alternative?

pkg/monitor/cluster/managednamespaces.go
hawkowl: For this, we only care about core OCP namespaces (e.g. not openshift-gitops, for example, which is a customer workload). The kube- namespaces should also be stuff like `openshift-kube-apiserver`, not plain `kube-system` (iirc).

pkg/monitor/cluster/deploymentstatuses.go
hawkowl: I don't think we ever used them, but this was before we had access to prometheus -- so we couldn't find out if a customer had a billion pods without asking them. Now, we can look at the cluster Prometheus instead.

pkg/monitor/cluster/cluster.go
```diff
@@ -142,21 +149,23 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
...
+		queryLimit:          50,
```
aasserzo, line 168: Can this value be tested/benchmarked outside of production somehow?

pkg/monitor/cluster/deploymentstatuses.go
bizz001: I am also a bit concerned that we are removing various Gauges

pkg/monitor/cluster/daemonsetstatuses.go
```diff
@@ -5,48 +5,44 @@ package cluster
...
+		for {
+			err := mon.ocpclientset.List(ctx, l, client.InNamespace(ns), client.Continue(cont), client.Limit(mon.queryLimit))
+			if err != nil {
+				return fmt.Errorf("error in list operation: %w", err)
 			}
 
-			if ds.Status.DesiredNumberScheduled == ds.Status.NumberAvailable {
-				continue
+			for _, ds := range l.Items {
+				if ds.Status.DesiredNumberScheduled == ds.Status.NumberAvailable {
```
Tof1973, line 29: I'm wondering if ds. Status could be nil, and did not find evidence of the opposite, but not sure. In case it could, it should be "protected".

pkg/monitor/cluster/deploymentstatuses.go
```diff
@@ -5,46 +5,44 @@ package cluster
...
+		for {
+			err := mon.ocpclientset.List(ctx, l, client.InNamespace(ns), client.Continue(cont), client.Limit(mon.queryLimit))
+			if err != nil {
+				return fmt.Errorf("error in list operation: %w", err)
 			}
 
-			if d.Status.Replicas == d.Status.AvailableReplicas {
-				continue
+			for _, d := range l.Items {
+				if d.Status.Replicas == d.Status.AvailableReplicas {
```
Tof1973, line 29: Same comment as above.

pkg/monitor/cluster/jobconditions.go
```diff
@@ -19,47 +19,43 @@ var jobConditionsExpected = map[batchv1.JobConditionType]corev1.ConditionStatus{
...
+		for {
+			err := mon.ocpclientset.List(ctx, l, client.InNamespace(ns), client.Continue(cont), client.Limit(mon.queryLimit))
+			if err != nil {
+				return fmt.Errorf("error in list operation: %w", err)
 			}
 
-			for _, cond := range job.Status.Conditions {
-				if cond.Status == jobConditionsExpected[cond.Type] {
+			for _, job := range l.Items {
+				if job.Status.Active > 0 {
```
Tof1973, line 34: Same than above

pkg/monitor/cluster/daemonsetstatuses.go
mrWinston: ds.Status can not be `nil` because it is not a pointer, but the full struct.

mrWinston: See https://pkg.go.dev/k8s.io/api/apps/v1#DaemonSet and https://pkg.go.dev/k8s.io/api/apps/v1#DaemonSetStatus for the docs.

pkg/monitor/cluster/deploymentstatuses.go
mrWinston: Same comment as above :) `deployment.Status` is not a pointer, which is why it can not be nil: https://pkg.go.dev/k8s.io/api/apps/v1#Deployment

pkg/monitor/cluster/jobconditions.go
mrWinston: The same is true for `job.Status` as well: https://pkg.go.dev/k8s.io/api/batch/v1#Job

pkg/monitor/cluster/deploymentstatuses.go
mrWinston: I checked in geneva and couldn't find any monitors that rely on the removed metrics. I don't believe this should be a problem. Even if this breaks a dashboard or two, i'd still argue the benefit of not needed to query ever single namespace in the cluster outweighs that downside.

mrWinston (dismissed):
looks good to me!

yithian (approved):
Seems sane to me

PR #4332, opened 2025-08-01
[ARO-20197] Move e2e hive check to use liveconfig instead

hawkowl:

Fixes blocker for [ARO-20197]

Fixes prod E2E by not requiring the admin API in the before checks

### Test plan for issue:

It is tests :)

N/A

N/A


hawkowl:
ref changes in https://github.com/Azure/ARO-RP/pull/4326

PR #4333, opened 2025-08-01
ARO-20185 - use armnetwork subnets client in delete.go

aasserzo:

Fixes https://issues.redhat.com/browse/ARO-20185

Tech debt

### Test plan for issue:
Covered in UT and E2E

No

Cluster deletion is covered in E2E

### DIFF
--- a/pkg/cluster/delete_test.go
+++ b/pkg/cluster/delete_test.go
@@ -335,7 +335,7 @@ func TestDisconnectSecurityGroup(t *testing.T) {
 				subnets.EXPECT().Get(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), nil).Times(0)
 				subnets.EXPECT().CreateOrUpdateAndWait(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), nil).Times(0)
 			},
-			wantErr: "400: InvalidLinkedSubnet: invalid-subnet-id: Invalid subnet resource ID format. For more details, please refer to https://docs.microsoft.com/azure/azure-resource-manager/management/resource-name-rules",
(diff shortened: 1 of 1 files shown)


pkg/cluster/delete.go
```diff
@@ -87,14 +88,30 @@ func (m *manager) disconnectSecurityGroup(ctx context.Context, resourceID string
 		// Note: subnet only has value in the ID field,
 		// so we have to make another API request to get full subnet struct
 		// TODO: there is probably an undesirable race condition here - check if etags can help.
-		s, err := m.subnet.Get(ctx, *subnet.ID)
+		r, err := arm.ParseResourceID(*subnet.ID)
+		if err != nil {
+			return &api.CloudError{
```
mociarain, line 93: Can we add a test for this error?

pkg/cluster/delete.go
```diff
@@ -87,14 +88,30 @@ func (m *manager) disconnectSecurityGroup(ctx context.Context, resourceID string
 		// Note: subnet only has value in the ID field,
 		// so we have to make another API request to get full subnet struct
 		// TODO: there is probably an undesirable race condition here - check if etags can help.
```
mociarain, line 90: Do you know if this comment is still relevant? (Not a blocker)

mociarain (changes requested):
Small comment

pkg/cluster/delete.go
aasserzo: Great point, I added a test and improved the error

pkg/cluster/delete.go
aasserzo: Yes, it is relevant and my changes don't resolve the race condition.
As in the TODO, Etags can be used, but I'd have that done in a separate PR

alcasim (approved):

PR #4334, opened 2025-08-04
[ARO-20197-2] Run the cluster metrics steps in parallel

hawkowl:
Part of ARO-20197

- adds parallelism to the each-cluster metrics collection
- shifts the Hive-specific monitors into its own monitor struct
- cleans up a lot of the Hive restconfig/clustermanager usage so that it consistently uses one client rather than making 2 different sets for each cluster every run
- adds a metric for how long the wall-clock took (as well as debug logs)
- adds unit tests for the changed cluster code.
- removes the cache code which is racy when ran in parallel
- paginates uses of node reading
- modernisation of touched code to use clienthelper

### Test plan for issue:

Unit tests, manual testing

N/A

Run locally, but we should test in INT first


shubhadapaithankar (approved):

pkg/database/extensions_test.go
```diff
@@ -24,7 +24,7 @@ type testStruct struct {
 func TestExtensions(t *testing.T) {
 	ctx := context.Background()
 
-	aead, err := encryption.NewXChaCha20Poly1305(ctx, []byte("\x63\xb5\x59\xf0\x43\x34\x79\x49\x68\x46\xab\x8b\xce\xdb\xc1\x2d\x7a\x0b\x14\x86\x7e\x1a\xb2\xd7\x3a\x92\x4e\x98\x6c\x5e\xcb\xe1"))
+	aead, err := encryption.NewXChaCha20Poly1305(ctx, []byte("\x63\xb5\x59\xf0\x43\x34\x79\x49\x68\x46\xab\x8b\xce\xdb\xc1\x2d\x7a\x0b\x14\x86\x7e\x1a\xb2\xd7\x3a\x92\x4e\x98\x6c\x5e\xcb\xe1"), "")
```
jaitaiwan: Looks like a magic bytestring. Would be good to see it as a named constant or variable

pkg/util/encryption/multi.go
```diff
@@ -52,41 +53,58 @@ func NewMulti(ctx context.Context, serviceKeyvault azsecrets.Client, secretName,
...
 		}
 	}
 
-	return nil, err
+	errStrings := ""
+	for _, err := range errs {
+		errStrings = errStrings + fmt.Sprintf("\n\t* %s", err)
+	}
+
+	return nil, fmt.Errorf("no openers succeeded:%s", errStrings)
```
jaitaiwan: Prefer if we use `var ErrorNoOpeners = errors.New("no openers succeeded")` and make use of `errors.Join` to wrap the errors. Just allows for better testing + reuse.

pkg/util/encryption/multi_test.go
```diff
@@ -46,15 +47,16 @@ func TestOpen(t *testing.T) {
 				firstOpener.EXPECT().Open(mockInput).Return(nil, errors.New("fake error from the first opener"))
 				secondOpener.EXPECT().Open(mockInput).Return(nil, errors.New("fake error from the second opener"))
 			},
-			wantErr: "fake error from the second opener",
+			wantErr: "no openers succeeded:\n\t* first: fake error from the first opener\n\t* second: fake error from the second opener",
```
jaitaiwan: So this is where we could use `errors.Is` for better tests

jaitaiwan (changes requested):

pkg/monitor/cluster/cluster.go
```diff
@@ -254,13 +254,18 @@ func (mon *Monitor) Monitor(ctx context.Context) (errs []error) {
...
+			mon.log.Debugf("running %s", collectorName)
 			innerErr := f(ctx)
 			if innerErr != nil {
+				mon.log.Debugf("failed to run %s: %s", collectorName, innerErr.Error())
 				// emit metrics collection failures and collect the err, but
 				// don't stop running other metric collections
-				mon.emitFailureToGatherMetric(steps.ShortName(f), innerErr)
-				errChan <- fmt.Errorf("failure running cluster collector '%s': %w", steps.ShortName(f), innerErr)
+				mon.emitFailureToGatherMetric(collectorName, innerErr)
+				errChan <- fmt.Errorf("failure running cluster collector '%s': %w", collectorName, innerErr)
```
jaitaiwan, line 270: Again would prefer named errors with `errors.Join` to wrap.

jaitaiwan (changes requested):

pkg/monitor/cluster/arooperatorheartbeat.go
```diff
@@ -5,21 +5,29 @@ package cluster
...
-		"aro-operator-worker": false}
+		"aro-operator-worker": false,
+	}
 
-	aroDeployments, err := mon.listARODeployments(ctx)
+	l := &appsv1.DeploymentList{}
+	err := mon.ocpclientset.List(ctx, l, client.InNamespace(pkgoperator.Namespace))
 	if err != nil {
-		return err
+		return fmt.Errorf("failed listing ARO Operator deployments: %w", err)
```
jaitaiwan, line 27: Prefer named errors and wrapping using `errors.Join`

pkg/monitor/cluster/clusterversionconditions.go
```diff
@@ -17,9 +20,10 @@ var clusterVersionConditionsExpected = map[configv1.ClusterStatusConditionType]c
 }
 
 func (mon *Monitor) emitClusterVersionConditions(ctx context.Context) error {
-	cv, err := mon.getClusterVersion(ctx)
+	cv := &configv1.ClusterVersion{}
+	err := mon.ocpclientset.Get(ctx, types.NamespacedName{Name: "version"}, cv)
 	if err != nil {
-		return err
+		return fmt.Errorf("failure fetching ClusterVersion: %w", err)
```
jaitaiwan, line 26: Prefer named errors and wrapping w/ `errors.Join`

pkg/database/extensions_test.go
hawkowl: rebased this PR without this commit

pkg/util/encryption/multi_test.go
hawkowl: rebased this PR without this commit

pkg/util/encryption/multi.go
hawkowl: rebased this PR without this commit

jaitaiwan:
As per slack, we can do follow up for `errors.Join` but I think comment on the importance of the len() is necessary to prevent future oopsies issues.

jaitaiwan (approved):
LGTM. Improvements to errors can be done in future iterations.

PR #4338, opened 2025-08-08
ARO-19806 part 1 - remove MGMT network client consumers

aasserzo:
### Dependencies

This PR will remain a draft until the following PRs are merged:
[ARO-20185 - use armnetwork subnets client in delete.go](https://github.com/Azure/ARO-RP/pull/4333)
[ARO-20202 - get rid of deprecated mgmt network cilent in deploybaseresources.go and deploybaseresources_additional.go](https://github.com/Azure/ARO-RP/pull/4335)

Fixes https://issues.redhat.com/browse/ARO-19806
This is one of 3 PRs that address the final cleanup of the MGMT network SDK imports and usages.

Tech debt - moving away from MGMT network SDK to the newer ARM network SDK

### Test plan for issue:

Covered in UT and E2E

No documentation updates needed, this is tech debt.

Will be covered by E2E testing.

### DIFF
--- a/pkg/cluster/delete.go
+++ b/pkg/cluster/delete.go
@@ -14,6 +14,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/wait"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/msi/armmsi"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
 	mgmtfeatures "github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-07-01/features"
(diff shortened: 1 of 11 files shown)


pkg/util/arm/marshal.go
```diff
@@ -10,12 +10,37 @@ import (
...
+	armnetwork "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
 )
 
 const (
 	track2sdkPkgPathPrefix = "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager"
 )
 
+// ARM resource types that should include Type and Location fields when marshaling to JSON
+// Add new types here as needed.
+var armResourceTypes = map[reflect.Type]bool{
```
mrWinston: I feel really strongly that we should not include such a list in the code. This is a nightmare to maintain. Let's instead look for a way to identify these types directly in `MarshalJSON()`.

What exactly is the difference between these types and others when marshalled?

pkg/util/arm/marshal.go
```diff
@@ -42,6 +67,17 @@ func (r *Resource) MarshalJSON() ([]byte, error) {
 		if r.DependsOn != nil {
 			dataMap["dependsOn"] = r.DependsOn
 		}
+		resourceType := reflect.TypeOf(r.Resource)
+		elemType := resourceType
+		if resourceType.Kind() == reflect.Ptr {
+			elemType = resourceType.Elem()
+		}
+		if (armResourceTypes[resourceType] || armResourceTypes[elemType]) && r.Type != "" {
```
mrWinston: I did some debugging to understand the logic here a bit better and whenever this block is entered `dataMap["type"]` is nil and vice versa. (whenever `dataMap["type"]==nil`, this block was entered.).

Would it make sense to replace this cross checking against `armResourceTypes` with a simple nil-check?
```
if dataMap["type"] == nil && r.Type != ""
```

pkg/util/arm/marshal.go
aasserzo: Agreed, and removed. This logic is unnecessary if we don't include these fields in the resource to begin with.

pkg/util/arm/marshal.go
aasserzo: This is also removed

pkg/deploy/generator/resources_rp.go
```diff
@@ -1338,16 +1326,14 @@ func (g *generator) rpCosmosDBTriggers(databaseName, containerName, triggerID, t
 					TriggerType:      &triggerType,
 				},
 			},
-			Name:     pointerutils.ToPtr("[concat(parameters('databaseAccountName'), '/', " + databaseName + ", '/" + containerName + "/" + triggerID + "')]"),
-			Type:     pointerutils.ToPtr("Microsoft.DocumentDB/databaseAccounts/sqlDatabases/containers/triggers"),
-			Location: pointerutils.ToPtr("[resourceGroup().location]"),
+			Name: pointerutils.ToPtr("[concat(parameters('databaseAccountName'), '/', " + databaseName + ", '/" + containerName + "/" + triggerID + "')]"),
+			Type: pointerutils.ToPtr("Microsoft.DocumentDB/databaseAccounts/sqlDatabases/containers/triggers"),
```
mrWinston: Why was the `Location` field removed here? Do we know if this has any effect on the provisioning of the databases?

aasserzo: I believe this is a mistake on my part. I shall fix

PR #4339, opened 2025-08-08
ARO-19806 part 2 - use ARM network types in generators

aasserzo:
### Dependencies

This PR will remain a draft until the following PRs are merged:
[ARO-20185 - use armnetwork subnets client in delete.go](https://github.com/Azure/ARO-RP/pull/4333)
[ARO-20202 - get rid of deprecated mgmt network cilent in deploybaseresources.go and deploybaseresources_additional.go](https://github.com/Azure/ARO-RP/pull/4335)
[ARO-19806 part 1 - remove MGMT network client consumers](https://github.com/Azure/ARO-RP/pull/4338)

### Which issue this PR addresses:

Fixes https://issues.redhat.com/browse/ARO-19806
This is one of 3 PRs that address the final cleanup of the MGMT network SDK imports and usages.

### What this PR does / why we need it:

Tech debt - moving away from MGMT network SDK to the newer ARM network SDK

### Test plan for issue:

Covered in UT and E2E

### Is there any documentation that needs to be updated for this PR?

No documentation updates needed, this is tech debt.

### How do you know this will function as expected in production? 

Will be covered by E2E testing.

PR #4340, opened 2025-08-08
ARO-19806 part 3 - cleanup old mgmt/network package

aasserzo:
### Dependencies

This PR will remain a draft until the following PRs are merged:
[ARO-20185 - use armnetwork subnets client in delete.go](https://github.com/Azure/ARO-RP/pull/4333)
[ARO-20202 - get rid of deprecated mgmt network cilent in deploybaseresources.go and deploybaseresources_additional.go](https://github.com/Azure/ARO-RP/pull/4335)
[ARO-19806 part 1 - remove MGMT network client consumers](https://github.com/Azure/ARO-RP/pull/4338)
[ARO-19806 part 2 - use ARM network types in generators](https://github.com/Azure/ARO-RP/pull/4339)

Fixes https://issues.redhat.com/browse/ARO-19806
This is one of 3 PRs that address the final cleanup of the MGMT network SDK imports and usages.

Tech debt - moving away from MGMT network SDK to the newer ARM network SDK

### Test plan for issue:

Covered in UT and E2E

No documentation updates needed, this is tech debt.

Will be covered by E2E testing.


pkg/cluster/delete.go
```diff
@@ -100,7 +100,7 @@ func (m *manager) disconnectSecurityGroup(ctx context.Context, resourceID string
 			}
 		}
 
-		subnetResp, err := m.armSubnets.Get(ctx, r.ResourceGroupName, r.Parent.Name, r.Name, nil)
+		subnetResponse, err := m.armSubnets.Get(ctx, r.ResourceGroupName, r.Parent.Name, r.Name, nil)
```
mociarain, line 103: ❤️

pkg/util/subnet/cluster_subnets.go
```diff
@@ -22,6 +22,11 @@ const (
 	machineSetsNamespace = "openshift-machine-api"
 )
 
+type Subnet struct {
```
mociarain: I dislike that we're duplicating models but we have no concept of a common models or anything so I don't know what else we should do. Any ideas?

(Not blocking)

aasserzo: Good point. I've consolidated on using the Subnet struct from pkg/api/util/subnet everywhere, removing the duplicate definition. LMK if this looks better to you.

aasserzo:
> > > @aasserzo Just to clarify: This PR removes all those old client implementations because because they have been replace with the new SDK? And the PRs that introduced the new sdks did not remove the now unused implementations, right?
> > > I'm trying to wrap my head around the fact that we can just remove so many files without it impacting anything.
> > 
> > 
> > @mrWinston yes, this is cleanup after previous PRs, where the old client was gradually made obsolete.
> 
> Nice! Do we also need to run `go mod tidy` to get rid of now unused dependencies?

I didn't remove dependencies on any external packages, so go mod tidy doesn't change anything.

PR #4343, opened 2025-08-12
ARO-17016 - Retrevial of effective route tables

chreed-rh:

https://issues.redhat.com/browse/ARO-17016

Fixes

Will allow to retrieve effective route tables for a particular cluster through Geneva actions

### Test plan for issue:

There are unit test included. To test, execute make local-rp, and then the following in another terminal

curl -X GET -k "https://localhost:8443/admin/subscriptions/<sub-id>/resourceGroups/<rg>/providers/Microsoft.RedHatOpenShift/openShiftClusters/<cluster-name>/effectiveroutingtables?subid=<sub-id>&rgn=<rg>&nic=<nic-id>"

None that I'm aware of at this time.

Replicated environment where curl was executed from Microsoft VM, and also with C# code that replicated Geneva Actions front end.

### DIFF
--- a/pkg/frontend/admin_openshiftcluster_get_effectiveroutetable.go
+++ b/pkg/frontend/admin_openshiftcluster_get_effectiveroutetable.go
@@ -4,74 +4,22 @@ package frontend
 // Licensed under the Apache License 2.0.
 
 import (
-	"context"
 	"net/http"
 	"path/filepath"
-	"strings"
 
-	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
-	"github.com/sirupsen/logrus"
 	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+	"github.com/sirupsen/logrus"
 )
 
 func (f *frontend) getAdminOpenshiftClusterEffectiveRouteTable(w http.ResponseWriter, r *http.Request) {
-    ctx := r.Context()
-    log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
-    r.URL.Path = filepath.Dir(r.URL.Path)
+	ctx := r.Context()
+	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
+	r.URL.Path = filepath.Dir(r.URL.Path)
 
-    effectiveRouteTableColleciton, err := f._getAdminOpenshiftClusterEffectiveRouteTable(ctx, r, log)
+	e, err := f._getOpenshiftClusterEffectiveRouteTable(ctx, w, r, log)
+	if err != nil {
+		log.Fatalf("Unable to get effective route table: %v", err)
+	}
 
-    if err != nil {
-        log.Fatalf("Unable to get effective route table: %v", err)   
-    }
-
-    convertedEffectiveRouteTableCollection, err := effectiveRouteTableColleciton.EffectiveRouteListResult.MarshalJSON()
-
-    adminReply(log, w, nil, convertedEffectiveRouteTableCollection, err)
-}
-
-func (f *frontend) _getAdminOpenshiftClusterEffectiveRouteTable(ctx context.Context, r *http.Request, log *logrus.Entry) (armnetwork.InterfacesClientGetEffectiveRouteTableResponse, error) {
-    subID := r.URL.Query().Get("subscriptionID")
-    rg := r.URL.Query().Get("resourceGroupName")
-    nicName := r.URL.Query().Get("nic")
-
-    resourceID := strings.TrimPrefix(r.URL.Path, "/admin")
-
-    dbOpenShiftClusters, err := f.dbGroup.OpenShiftClusters()
-
-    if err != nil {
-        //return nil, api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
-        log.Fatalf("failed to load cluster from cosmosDB: %v", err)
-    }
-
-    doc, err := dbOpenShiftClusters.Get(ctx, resourceID)
-
-    subscriptionDoc, err := f.getSubscriptionDocument(ctx, doc.Key)
-    if err != nil {
-        //return nil, err
-        log.Fatalf("failed to retrieve subscription document: %v", err)
-    }
-
-    credential, err := f.env.FPNewClientCertificateCredential(subscriptionDoc.Subscription.Properties.TenantID, nil)
-    if err != nil {
-        //return nil, err
-        log.Fatalf("failed to create client: %v", err)
-    }
-
-    clientFactory, err := armnetwork.NewClientFactory(subID, credential, nil)
-    if err != nil {
-        log.Fatalf("failed to create client: %v", err)
-    }
-
-    poller, err := clientFactory.NewInterfacesClient().BeginGetEffectiveRouteTable(context.Background(), rg, nicName, nil)
-    if err != nil {
-        log.Fatalf("failed to finish the request: %v", err)
-    }
-
-    res, err := poller.PollUntilDone(ctx, nil)
-    if err != nil {
-        log.Fatalf("failed to pull the result: %v", err)
-    }
-
-    return res, err
+	adminReply(log, w, nil, e, err)
 }
--- a/pkg/frontend/admin_openshiftcluster_get_effectiveroutetable.go
+++ b/pkg/frontend/admin_openshiftcluster_get_effectiveroutetable.go
@@ -1,0 +1,77 @@
+package frontend
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"net/http"
+	"path/filepath"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
+	"github.com/sirupsen/logrus"
+	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+)
+
+func (f *frontend) getAdminOpenshiftClusterEffectiveRouteTable(w http.ResponseWriter, r *http.Request) {
+    ctx := r.Context()
+    log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
+    r.URL.Path = filepath.Dir(r.URL.Path)
+
+    effectiveRouteTableColleciton, err := f._getAdminOpenshiftClusterEffectiveRouteTable(ctx, r, log)
+
+    if err != nil {
+        log.Fatalf("Unable to get effective route table: %v", err)   
+    }
+
+    convertedEffectiveRouteTableCollection, err := effectiveRouteTableColleciton.EffectiveRouteListResult.MarshalJSON()
+
+    adminReply(log, w, nil, convertedEffectiveRouteTableCollection, err)
+}
+
+func (f *frontend) _getAdminOpenshiftClusterEffectiveRouteTable(ctx context.Context, r *http.Request, log *logrus.Entry) (armnetwork.InterfacesClientGetEffectiveRouteTableResponse, error) {
+    subID := r.URL.Query().Get("subscriptionID")
+    rg := r.URL.Query().Get("resourceGroupName")
+    nicName := r.URL.Query().Get("nic")
+
+    resourceID := strings.TrimPrefix(r.URL.Path, "/admin")
+
+    dbOpenShiftClusters, err := f.dbGroup.OpenShiftClusters()
+
+    if err != nil {
+        //return nil, api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
+        log.Fatalf("failed to load cluster from cosmosDB: %v", err)
+    }
+
+    doc, err := dbOpenShiftClusters.Get(ctx, resourceID)
+
+    subscriptionDoc, err := f.getSubscriptionDocument(ctx, doc.Key)
+    if err != nil {
+        //return nil, err
+        log.Fatalf("failed to retrieve subscription document: %v", err)
+    }
+
+    credential, err := f.env.FPNewClientCertificateCredential(subscriptionDoc.Subscription.Properties.TenantID, nil)
+    if err != nil {
+        //return nil, err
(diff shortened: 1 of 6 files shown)


pkg/frontend/admin_openshiftcluster_get_effectiveroutetable.go
```diff
@@ -0,0 +1,26 @@
...
+
+func (f *frontend) getAdminOpenshiftClusterEffectiveRouteTable(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	log := ctx.Value(middleware.ContextKeyLog).(*logrus.Entry)
+	r.URL.Path = filepath.Dir(r.URL.Path)
+
+	e, err := f._getOpenshiftClusterEffectiveRouteTable(ctx, w, r, log)
+	if err != nil {
+		log.Fatalf("Unable to get effective route table: %v", err)
+	}
```
kimorris27: We can skip this logging step. `adminReply` will make sure the error gets logged.

pkg/frontend/frontend.go
```diff
@@ -324,6 +324,8 @@ func (f *frontend) chiAuthenticatedRoutes(router chi.Router) {
 				// Top nodes metrics endpoint
 				r.Get("/top/nodes", f.getAdminTopNodes)
 
+				r.Get("/effectiveroutingtables", f.getAdminOpenshiftClusterEffectiveRouteTable)
```
kimorris27:
```suggestion
				r.Get("/effectiveroutetable", f.getAdminOpenshiftClusterEffectiveRouteTable)
```

Suggesting this change for alignment with the verbiage used elsewhere throughout the code.

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
```diff
@@ -0,0 +1,69 @@
...
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
+)
+
+func (f *frontend) _getOpenshiftClusterEffectiveRouteTable(ctx context.Context, w http.ResponseWriter, r *http.Request, log *logrus.Entry) ([]byte, error) {
+	subID := r.URL.Query().Get("subid")
+	rg := r.URL.Query().Get("rgn")
```
kimorris27: There's no need to include these two things as URL parameters. You can grab both of them from the cluster doc.

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
```diff
@@ -0,0 +1,69 @@
...
+	resourceID := strings.TrimPrefix(r.URL.Path, "/admin")
+
+	dbOpenShiftClusters, err := f.dbGroup.OpenShiftClusters()
+
+	if err != nil {
+		//return nil, api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
+		log.Fatalf("failed to load cluster from cosmosDB: %v", err)
+	}
+
+	doc, _ := dbOpenShiftClusters.Get(ctx, resourceID)
```
kimorris27: We'll definitely want to handle any error encountered here. For reference: https://github.com/Azure/ARO-RP/blob/f9e8426e34da55b422003f4f24f7023fb95d6689/pkg/frontend/admin_openshiftcluster_kubernetesobjects.go#L56-L62

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
```diff
@@ -0,0 +1,69 @@
...
+	nicName := r.URL.Query().Get("nic")
+
+	resourceID := strings.TrimPrefix(r.URL.Path, "/admin")
+
+	dbOpenShiftClusters, err := f.dbGroup.OpenShiftClusters()
+
+	if err != nil {
+		//return nil, api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
+		log.Fatalf("failed to load cluster from cosmosDB: %v", err)
+	}
```
kimorris27:
```suggestion
	if err != nil {
		return nil, api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
	}
```

I'm assuming this log statement was leftover from local testing.

Also, just pointing out for clarity's sake that an error here doesn't indicate we failed to get the cluster from CosmosDB - it indicates that we failed to get the OpenShiftClusters database client.

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
```diff
@@ -0,0 +1,69 @@
...
+
+	doc, _ := dbOpenShiftClusters.Get(ctx, resourceID)
+
+	subscriptionDoc, err := f.getSubscriptionDocument(ctx, doc.Key)
+
+	if err != nil {
+		//return nil, err
+		log.Fatalf("failed to retrieve subscription document: %v", err)
```
kimorris27:
```suggestion
		return nil, err
```

Similar to other suggestions; I'm assuming the log statement if leftover from local testing.

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
```diff
@@ -0,0 +1,69 @@
...
+
+	if err != nil {
+		log.Fatalf("failed to finish the request: %v", err)
+	}
+
+	res, err := poller.PollUntilDone(ctx, nil)
+
+	if err != nil {
+		log.Fatalf("failed to pull the result: %v", err)
+	}
```
kimorris27: There are opportunities here to better align with existing patterns in the codebase:

1. Rather than directly instantiating Azure SDK clients in places where we need them, we have a util package where we keep client wrappers. See the preexisting `InterfacesClient` wrapper: https://github.com/Azure/ARO-RP/blob/master/pkg/util/azureclient/azuresdk/armnetwork/interfaces.go
2. For the frontend and admin actions specifically, we have an `AzureActions` interface type for encapsulating any Azure SDK clients and functionality. See https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/adminactions/azureactions.go#L27 for that interface and https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/admin_openshiftcluster_redeployvm.go for an example of an admin API endpoint that uses it. Using this existing scaffolding will eliminate the one-off credential grabbing and client instantiating currently done here.

The `azureActions` struct type (implements `AzureActions`) already has an `InterfacesClient`, so that should make it easy to switch over.

kimorris27 (changes requested):
Nice work taking advantage of the existing SDK function for this. I know you tried a number of different approaches before we found out about that function 😅 

I suggested some changes to make sure we're handling errors appropriately and leveraging existing patterns in the codebase.

pkg/frontend/admin_openshiftcluster_get_effectiveroutetable.go
chreed-rh: Change has been implemented

pkg/frontend/frontend.go
chreed-rh: Change has been implemented

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
chreed-rh: Change has been implemented

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
chreed-rh: Change has been implemented

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
chreed-rh: That is correct. Leftover code that snuck through the cracks. Change has been implemented

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
chreed-rh: Change has been implemented

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
chreed-rh: Change has been implemented. I beleive I did it originally because the attempt to use the existing wrappers had an issue trying to get an instance that could be used for this action. Also it required a fair number of refactoring, and I didn't want to refactor a fair number of files without making sure this would work to begin with.

pkg/frontend/admin_openshiftcluster_get_effectiveroutetable.go
```diff
@@ -0,0 +1,32 @@
...
+	r.URL.Path = filepath.Dir(r.URL.Path)
+
+	e, err := f._getOpenshiftClusterEffectiveRouteTable(ctx, r)
+	if err != nil {
+		log.Errorf("Unable to get effective route table: %v", err)
+		api.WriteError(w, http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", "Failed to retrieve effective route table.")
+		return
+	}
+
+	adminReply(log, w, nil, e, err)
```
kimorris27: You can pass `err` straight to `adminReply` and let it do all of the handling. A reference example: https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/admin_openshiftcluster_kubernetesobjects.go#L25-L33

pkg/frontend/admin_openshiftcluster_get_effectiveroutetable.go
```diff
@@ -0,0 +1,32 @@
...
+	"net/http"
+	"path/filepath"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/Azure/ARO-RP/pkg/api"
+	"github.com/Azure/ARO-RP/pkg/frontend/middleware"
+)
+
+// getAdminOpenshiftClusterEffectiveRouteTable is the refactored admin version using ARO-RP utilities
```
kimorris27: We can remove this before merge.

pkg/frontend/frontend.go
```diff
@@ -324,6 +324,8 @@ func (f *frontend) chiAuthenticatedRoutes(router chi.Router) {
 				// Top nodes metrics endpoint
 				r.Get("/top/nodes", f.getAdminTopNodes)
 
+				r.Get("/effectiveroutingtable", f.getAdminOpenshiftClusterEffectiveRouteTable)
```
kimorris27: Nit: I had suggested `effectiveroutetable` and not `effectiveroutingtable`. It's a nitpicky difference, but I imagine it'll make the code just a bit easier to search and read if we're fully consistent in our word choice.

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// Create ARM network interfaces client using ARO-RP utility
+	interfacesClient, err := utilarmnetwork.NewInterfacesClient(resource.SubscriptionID, credential, nil)
+	if err != nil {
+		return nil, api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "",
+			fmt.Sprintf("Failed to create network interfaces client: %v", err))
+	}
+
+	// Call GetEffectiveRouteTable using the ARO-RP utility pattern
+	result, err := interfacesClient.GetEffectiveRouteTableAndWait(ctx, resource.ResourceGroup, nicName, nil)
```
kimorris27: You'll want to grab the resource group from the cluster doc's `ClusterResourceGroup` attribute rather than using the cluster's resource ID. The cluster's resource ID gives us the resource group containing the cluster, but we need the managed resource group's ID (that's where the NIC will be).

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
kimorris27: It looks like you implemented 1 but not 2 (unless you're still working on it as I comment?).

Since the `azureActions` struct already has an `InterfacesClient` wired into it, I guess I really mentioned 1 more for explanatory purposes.

pkg/frontend/frontend.go
chreed-rh: Totally agree. That was an oversight on my part

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
kimorris27, line 1: I ought to have caught this sooner in the review process, but we can move the contents of this file to `pkg/frontend/admin_openshiftcluster_get_effectiveroutetable.go`. The convention for the files in `pkg/frontend` is to use one file per endpoint, and this file essentially contains a helper function for the admin endpoint. Compare to https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/admin_openshiftcluster_delete_managedresource.go, for example.

pkg/frontend/admin_openshiftcluster_get_effectiveroutetable_test.go
kimorris27, line 1: All of these functions test the dependencies of the new functionality without testing the new functionality itself.

See https://github.com/Azure/ARO-RP/blob/master/pkg/frontend/admin_openshiftcluster_kubernetesobjects_test.go for an example of how to leverage existing testing infrastructure to instantiate a frontend instance and make requests to it at the new endpoint you've added.

kimorris27 (changes requested):
I added one organizational request and a suggestion about how the unit tests might be redone.

Another observation is that we could add an E2E test for this new admin endpoint, but we could put that in its own PR if you wanted to.

pkg/frontend/openshiftcluster_get_effectiveroutetable.go
chreed-rh: Sounds good. The original reason why I split it in that matter was due to using another submission as a template, where it had an admin, and a non admin file like how I had it before. It was a bit more involved so maybe it wasn't the best example to run with.

kimorris27 (approved):

cadenmarchese (approved):
The admin action and test look good. My nitpick is that we should add it to docs in the usual spot: https://github.com/Azure/ARO-RP/blob/master/docs/deploy-development-rp.md#make-admin-action-api-calls-to-a-running-local-rp

PR #4344, opened 2025-08-12
Ignore GlobalDevopsManagedIdentity RP-Config attribute in full service dev

kimorris27:

https://issues.redhat.com/browse/ARO-20532

https://github.com/Azure/ARO-RP/pull/3851 introduced a new `required` RP-Config attribute even though full service dev doesn't have it. This PR removes the `required` tag and adds code to skip any of the role assignments that would be done to the global devops MSI in a full service dev context.

I also snuck in an update to the AKS version used in full service dev because 1.29 is too old to be supported now. Note that the aks-development.json file is not generated and I edited it manually.

### Test plan for issue:

Prior to this fix, my full service dev deployment errored out saying that I was missing the attribute, and after fixing that, it errored out indicating the MSI's SP ID was missing from the ARM template. `make deploy` progresses past these issues with the changes in this PR (though there are various other reasons that full service dev is still broken right now).

No

Not a production change.

### DIFF
--- a/pkg/deploy/assets/rp-development.json
+++ b/pkg/deploy/assets/rp-development.json
@@ -12,7 +12,8 @@
             "type": "string"
         },
         "globalDevopsServicePrincipalId": {
-            "type": "string"
+            "type": "string",
+            "defaultValue": ""
         },
         "rpServicePrincipalId": {
             "type": "string"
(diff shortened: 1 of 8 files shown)


fahlmant (approved):

kimorris27:
Merging the PR since I finally got a successful INT deployment.

PR #4349, opened 2025-08-15
Use GetIssuers for admin api ca bundle

kimorris27:

https://issues.redhat.com/browse/ARO-20210

This PR switches us over to using the GetIssuers API to grab the admin API CA bundle in prod environments. That way we can remove the hard-coded CA bundle from RP-Config to address a Microsoft security/compliance item.

Note that the RP-Config attribute is still needed for full service dev. I wanted to test my branch in full service dev as well, but the env has so many issues right now that I had to let it go.

### Test plan for issue:

Tested by deploying my branch to INT and running a List clusters Geneva Action; everything worked 👍🏻 

N/A

Tested in INT

### DIFF
--- a/pkg/env/prod.go
+++ b/pkg/env/prod.go
@@ -285,19 +285,34 @@ func (p *prod) InitializeAuthorizers() error {
 		}
 	}
 
-	var issuerPkiUrls []string
-	for _, ca := range p.Environment().PkiCaNames {
-		issuerPkiUrls = append(issuerPkiUrls, fmt.Sprintf(p.Environment().PkiIssuerUrlTemplate, ca))
-	}
+	var adminCertPool *x509.CertPool
 
-	rootCAs, err := pki.FetchDataFromGetIssuerPkiUrls(issuerPkiUrls)
-	if err != nil {
-		return err
-	}
+	if !p.FeatureIsSet(FeatureEnableDevelopmentAuthorizer) {
+		var issuerPkiUrls []string
+		for _, ca := range p.Environment().PkiCaNames {
+			issuerPkiUrls = append(issuerPkiUrls, fmt.Sprintf(p.Environment().PkiIssuerUrlTemplate, ca))
+		}
 
-	adminCertPool, err := pki.BuildCertPoolFromCAData(rootCAs)
-	if err != nil {
-		return err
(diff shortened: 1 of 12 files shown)


pkg/util/azureclient/environments.go
```diff
@@ -18,6 +18,15 @@ import (
 	utilgraph "github.com/Azure/ARO-RP/pkg/util/graph"
 )
 
+// Because these values are the same between Public and GovCloud
+var (
+	pkiIssuerUrlTemplate = "https://issuer.pki.azure.com/dsms/issuercertificates?getissuersv3&caName=%s"
+	pkiCaNames           = []string{
+		"ame",  // AMEv1
```
cadenmarchese: Looking at https://eng.ms/docs/products/onecert-certificates-key-vault-and-dsms/key-vault-dsms/reference/ca-details?tabs=ame-pki, it seems that fetching CAs for AMEv1 is not going to be necessary come some point in 2026 as they are moving everything to AMEv2.

If this understanding is correct, is it worth leaving a TODO or code comment stating that the ame caName is no longer going to be necessary at that point?

cadenmarchese (commented):
LGTM, just one idea.

pkg/util/azureclient/environments.go
kimorris27: Makes sense, and done! 👍🏻

pkg/env/prod.go
```diff
@@ -273,16 +285,47 @@ func (p *prod) InitializeAuthorizers() error {
 		}
 	}
 
+	var adminCertPool *x509.CertPool
+
+	if !p.FeatureIsSet(FeatureEnableDevelopmentAuthorizer) {
```
yithian: I'm not sure I follow what's happening here. Can you add a comment to describe it?

kimorris27: Just pushed a comment. Let me know if that does the trick.

yithian: Thanks, that's super helpful! I would have guessed the prod/dev blocks were reversed :sweat_smile:

kimorris27: You make a good point about the order of the blocks being confusing because of the negation in the `if` statement. I flipped the blocks and removed the comment. Hopefully that's even better than having the comments.

PR #4352, opened 2025-08-20
[ARO-20197] Reduce memory usage in subscriptions cache as well as correct logging and metrics being unqueryable

hawkowl:

More on ARO-20197, fixes https://issues.redhat.com/browse/ARO-20734 

- Cuts down on memory usage by limiting what we need for the Subscriptions changefeed.
- Introduces paging for the changefeeds so that we're not making huge JSON blobs.
- Adds CosmosDB metrics to the monitor
- Adds more dimensions to the system monitors (e.g. the RP version)
- Adds a "system" column to log outputs so that the monitor db logging doesn't get lost inamongst the other services, since we currently cannot tell what belongs to what

### Test plan for issue:

Unit tests, manual testing, E2E

N/A

Local testing, E2E

### DIFF
--- a/cmd/aro/deploy.go
+++ b/cmd/aro/deploy.go
@@ -21,7 +21,7 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/version"
 )
 
-func deploy(ctx context.Context, log *logrus.Entry) error {
+func deploy(ctx context.Context, _log *logrus.Entry) error {
 	// TODO(mjudeikis): Remove this hack in public once we moved to EV2
 	// We are not able to use MSI in public cloud CI as we would need
 	// to have dedicated node pool with MSI where we can controll which jobs are running
@@ -34,7 +34,7 @@ func deploy(ctx context.Context, log *logrus.Entry) error {
 	var tokenCredential azcore.TokenCredential
 	if os.Getenv("AZURE_EV2") != "" { // running in EV2 - use MSI
 		var err error
-		_env, err = env.NewCore(ctx, log, env.COMPONENT_DEPLOY)
+		_env, err = env.NewCore(ctx, _log, env.COMPONENT_DEPLOY)
 		if err != nil {
 			return err
 		}
@@ -54,7 +54,7 @@ func deploy(ctx context.Context, log *logrus.Entry) error {
 			return err
 		}
 
-		_env, err = env.NewCoreForCI(ctx, log, env.COMPONENT_DEPLOY)
+		_env, err = env.NewCoreForCI(ctx, _log, env.COMPONENT_DEPLOY)
 		if err != nil {
 			return err
 		}
@@ -67,6 +67,8 @@ func deploy(ctx context.Context, log *logrus.Entry) error {
 	env := _env
 
 	deployVersion, location := version.GitCommit, flag.Arg(2)
+
+	log := _env.Logger()
 
 	log.Printf("deploying version %s to location %s", deployVersion, location)
 
@@ -107,7 +109,7 @@ func deploy(ctx context.Context, log *logrus.Entry) error {
 		}
 	}
 
-	deployer, err := pkgdeploy.New(ctx, log, env, config, deployVersion, tokenCredential)
+	deployer, err := pkgdeploy.New(ctx, env, config, deployVersion, tokenCredential)
 	if err != nil {
 		return err
 	}
--- a/cmd/aro/deploy.go
+++ b/cmd/aro/deploy.go
@@ -34,7 +34,7 @@ func deploy(ctx context.Context, _log *logrus.Entry) error {
 	var tokenCredential azcore.TokenCredential
 	if os.Getenv("AZURE_EV2") != "" { // running in EV2 - use MSI
 		var err error
-		_env, err = env.NewCore(ctx, _log, env.COMPONENT_DEPLOY)
+		_env, err = env.NewCore(ctx, _log, env.SERVICE_DEPLOY)
 		if err != nil {
 			return err
 		}
@@ -54,7 +54,7 @@ func deploy(ctx context.Context, _log *logrus.Entry) error {
 			return err
 		}
 
-		_env, err = env.NewCoreForCI(ctx, _log, env.COMPONENT_DEPLOY)
+		_env, err = env.NewCoreForCI(ctx, _log, env.SERVICE_DEPLOY)
 		if err != nil {
 			return err
 		}
--- a/cmd/aro/deploy.go
+++ b/cmd/aro/deploy.go
@@ -54,7 +54,7 @@ func deploy(ctx context.Context, log *logrus.Entry) error {
 			return err
 		}
 
-		_env, err = env.NewCoreForCI(ctx, log)
+		_env, err = env.NewCoreForCI(ctx, log, env.COMPONENT_DEPLOY)
 		if err != nil {
 			return err
 		}
--- a/pkg/env/core.go
+++ b/pkg/env/core.go
@@ -33,6 +33,7 @@ const (
 	COMPONENT_TOOLING             ServiceComponent = "TOOLING"
 	COMPONENT_MIMO_SCHEDULER      ServiceComponent = "MIMO_SCHEDULER"
 	COMPONENT_MIMO_ACTUATOR       ServiceComponent = "MIMO_ACTUATOR"
+	COMPONENT_E2E                 ServiceComponent = "E2E"
 )
 
 // Core collects basic configuration information which is expected to be
@@ -46,8 +47,9 @@ type Core interface {
 	NewLiveConfigManager(context.Context) (liveconfig.Manager, error)
 	instancemetadata.InstanceMetadata
 
-	Component() string
+	Service() string
 	Logger() *logrus.Entry
+	LoggerForComponent(string) *logrus.Entry
 }
 
 type core struct {
@@ -56,8 +58,8 @@ type core struct {
 	isLocalDevelopmentMode bool
 	isCI                   bool
 
-	component    ServiceComponent
-	componentLog *logrus.Entry
+	service    ServiceComponent
+	serviceLog *logrus.Entry
 
 	msiAuthorizers map[string]autorest.Authorizer
 }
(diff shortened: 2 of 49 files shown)


kimorris27 (approved):

kimorris27 (approved):

jaitaiwan (approved):

jaitaiwan (changes requested):
Requested changes to improve operational experience and avoid potential race condition

pkg/monitor/worker.go
```diff
@@ -322,10 +323,16 @@ func (mon *monitor) workOne(ctx context.Context, log *logrus.Entry, doc *api.Ope
...
+func execute(ctx context.Context, log *logrus.Entry, done chan<- bool, wg *sync.WaitGroup, monitors []monitoring.Monitor) {
 	for _, monitor := range monitors {
 		wg.Add(1)
-		go monitor.Monitor(ctx)
+		go func() {
+			errs := monitor.Monitor(ctx)
+			err := errors.Join(errs...)
+			if err != nil {
+				log.Error(err)
+			}
```
jaitaiwan, line 333: wg.Done is never called in the goroutine. `defer wg.Done()` should be added to the top of the goroutine.

pkg/monitor/worker.go
```diff
@@ -27,6 +28,9 @@ import (
 // nsgMonitoringFrequency is used for initializing NSG monitoring ticker
 var nsgMonitoringFrequency = 10 * time.Minute
 
+// changefeedBatchSize is how many items in the changefeed to fetch in each page
+const changefeedBatchSize = 50
```
jaitaiwan, line 32: Would be beneficial to have this as an environment variable so it's configurable without a code redeployment.

pkg/monitor/worker.go
hawkowl: the Monitor created at https://github.com/Azure/ARO-RP/pull/4352/files/3b4a5e4cdc6d499edb4d4e3bf5fb4e4d4f96cc9a..eb2fe0e8e20ab6862f03a5af12c16c0172a2ef7e#diff-fa2cef49d9d4dd21a31223298da75f2580147cbd8addde62391f8942aa889f2aL315 calls it here: https://github.com/Azure/ARO-RP/blob/eb2fe0e8e20ab6862f03a5af12c16c0172a2ef7e/pkg/monitor/cluster/cluster.go#L201 (or in equiv places)

It's a bit non-obvious though, so maybe this could do some refactoring.

pkg/monitor/worker.go
hawkowl: We do have to retag/rebuild/redeploy VMSSes from scratch at the moment, so I don't know if it would give us a great deal of changeability. But, I'll see if I can put something in.

cmd/aro/deploy.go
```diff
@@ -68,6 +68,8 @@ func deploy(ctx context.Context, log *logrus.Entry) error {
 
 	deployVersion, location := version.GitCommit, flag.Arg(2)
 
+	log := _env.Logger()
```
wanghaoran1988, line 71: A bit confused here, why we need a new log instance here?

hawkowl: Every time you make a new logger in logrus, it's a new copy -- just cuts down on the number of log instances we have floating about by creating one that we use a few times.

pkg/env/core.go
```diff
@@ -45,8 +46,9 @@ type Core interface {
 	NewLiveConfigManager(context.Context) (liveconfig.Manager, error)
 	instancemetadata.InstanceMetadata
 
-	Component() string
+	Service() string
```
wanghaoran1988, line 51: What's the difference between service and component, I see some places are using component, some are using service.

pkg/monitor/cluster/cluster.go
```diff
@@ -212,35 +211,50 @@ func (mon *Monitor) Monitor(ctx context.Context) (errs []error) {
...
+			// don't stop running other metric collections
+			mon.emitMonitorCollectorError(collectorName)
+			return e
+		} else {
+			timeToComplete := time.Since(innerNow).Seconds()
+			mon.emitMonitorCollectionTiming(collectorName, timeToComplete)
+			mon.log.Debugf("successfully ran cluster collector '%s' in %2f sec", collectorName, timeToComplete)
+		}
+		return innerErr
+	}
```
wanghaoran1988:
```suggestion
func (mon *Monitor) timeCall(ctx context.Context, collector func(context.Context) error) error {
	startTime := time.Now()
	collectorName := steps.ShortName(collector)

	mon.log.Debugf("running %s", collectorName)
	err := collector(ctx)

	if err != nil {
		wrappedErr := fmt.Errorf("failure running cluster collector '%s': %w", collectorName, err)
		mon.log.Debug(wrappedErr.Error())
		mon.emitMonitorCollectorError(collectorName)
		return wrappedErr
	}

	duration := time.Since(startTime).Seconds()
	mon.emitMonitorCollectionTiming(collectorName, duration)
	mon.log.Debugf("successfully ran cluster collector '%s' in %.2f sec", collectorName, duration)

	return nil
}
```

pkg/monitor/worker.go
```diff
@@ -127,10 +134,32 @@ func (mon *monitor) changefeed(ctx context.Context, baseLog *logrus.Entry, stop
 				break
 			}
 
+			baseLog.Debugf("subscriptions changefeed page was %d docs", subs.Count)
+
 			mon.mu.Lock()
 
 			for _, sub := range subs.SubscriptionDocuments {
-				mon.subs[sub.ID] = sub
+				id := strings.ToLower(sub.ID)
```
wanghaoran1988, line 141: Could you add a comment why we need lower case the subscription ID?

pkg/monitor/worker.go
hawkowl: I've cleaned this up.

pkg/env/core.go
hawkowl: Cleaned this up, added some comments around the logging component function to make it clearer

cadenmarchese (dismissed):
Looks like we're hitting the same e2e failure on multiple attempts:
```
run-e2e  | Summarizing 1 Failure:
run-e2e  |   [FAIL] Monitor [It] must run and must not return any errors
run-e2e  |   /app/test/e2e/monitor.go:28
```
Since other PRs haven't had this issue, I am betting that there's something in the PR we might need to adjust before merge.

pkg/metrics/statsd/statsd.go
```diff
@@ -84,8 +128,8 @@ func (s *statsd) emitMetric(m *metric) {
 	if m.dimensions == nil {
 		m.dimensions = map[string]string{}
 	}
-	m.dimensions["location"] = s.env.Location()
-	m.dimensions["hostname"] = s.env.Hostname()
+
+	maps.Copy(m.dimensions, s.extraDimensions)
 	m.timestamp = s.now()
 
 	s.ch <- m
```
jaitaiwan, line 133: I know this is not a part of the changes, but sending to a channel before there's a reader is asking for a deadlock

pkg/metrics/statsd/statsd.go
```diff
@@ -29,19 +31,61 @@ type statsd struct {
...
+
+	if s.account == "" {
+		s.account = "*"
+	}
+
+	if s.namespace == "" {
+		s.namespace = "*"
+	}
+
+	go s.run()
```
jaitaiwan: Having the run started implicitly inside of the new is introducing potential side effects - I'd prefer it at all possible for the goroutine to be managed with the component that has the lifetime of the metricsEmitter

pkg/monitor/azure/nsg/nsg.go
```diff
@@ -221,5 +216,9 @@ func (n *NSGMonitor) Monitor(ctx context.Context) []error {
 			}
 		}
 	}
-	return errors
+
+	// emit a metric with how long we took
+	n.emitter.EmitFloat("monitor.nsg.duration", time.Since(now).Seconds(), n.dims)
+
+	return errors.Join(errs...)
```
jaitaiwan, line 223: errors.Join is more for errors that result from errors, aka linear chains that you can follow to see root cause etc.

pkg/monitor/cluster/cluster.go
```diff
@@ -197,12 +193,30 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
 	return mon, nil
 }
 
+func (mon *Monitor) timeCall(ctx context.Context, f func(context.Context) error) error {
+	innerNow := time.Now()
+	collectorName := steps.ShortName(f)
+	mon.log.Debugf("running %s", collectorName)
+	innerErr := f(ctx)
+	if innerErr != nil {
+		e := fmt.Errorf("failure running cluster collector '%s': %w", collectorName, innerErr)
```
jaitaiwan: Still a good opportunity for a ErrFailureRunningClusterF variable.

pkg/monitor/worker.go
jaitaiwan: It's normalization for database

jaitaiwan (changes requested):
Looks mostly good. One major concern and most are a light concern or nit from there.

pkg/metrics/statsd/statsd.go
hawkowl: fixed

pkg/monitor/azure/nsg/nsg.go
hawkowl: errors.Join implements Unwrap() []error, not Unwrap() error as returned by uses of fmt.Errorf("%w"). It seems like Join() is more for "these errors happened", rather than a strictly linear wrapped chain. Unfortunately linear wrapping with sentinel errors requires more boilerplate, heh

pkg/monitor/cluster/cluster.go
hawkowl: I actually did this and then merge  conflicted it out, oops, fixed

pkg/metrics/statsd/statsd.go
hawkowl: s.ch is a buffered channel so it should only block if the buffer is full (which I don't think can happen here?)

jaitaiwan: You're right - I missed it being buffered

pkg/monitor/azure/nsg/nsg.go
jaitaiwan: I missed that in it's implementation. That is WEIRD

pkg/monitor/worker.go
jaitaiwan: Assuming we're not handling this, this time around

jaitaiwan (approved):

PR #4358, opened 2025-08-28
Alcasim/aro 11484 - Fix ETCD Geneva action

alcasim:

Fixes https://issues.redhat.com/browse/ARO-11484

It fixes the fix ETCD Geneva action.

- Jobs have been replaced by pods with `Never` as restart policy.
- The action now waits until the pods are completed (label selector fix + loop the watcher until desired event arrives).
- The backup folder name in the node ends with the date and time, to prevent the action failure in the rare case that the same etcd node fails more than once and running the action again is required

This PR is heavily based on Ayato's excellent work on PR: https://github.com/Azure/ARO-RP/pull/4034

I decided to not include the E2E test from the previous PR due to existing problems with E2E testing.

The fix works for etcd nodes that keep the previous IP address when they restart.

### Test plan for issue:

Mostly done on local development:

- oc debug/chroot into a node
- remove a wal file from etcd and restart the container from crictl
- etcd is now in CrashLoopBackOff in that node
- Run the action

The action makes the etcd operator to delete the crashing pod, recreate it and join the quorum. The operator re-creates the remaining etcd pods as a re-deployment is forced as part of the procedure.

https://msazure.visualstudio.com/AzureRedHatOpenShift/_wiki/wikis/ARO.wiki/837850/Etcd-Recovery-Single-Master-Etcd-Recovery-Standard-Operating-Procedure to reflect the change in the backup folder naming.

the action was already following the [procedure](https://docs.redhat.com/en/documentation/openshift_container_platform/4.15/html/backup_and_restore/control-plane-backup-and-restore#replacing-unhealthy-etcd-member) to replace an unhealthy etcd member, so the correct result is expected in production.


pkg/frontend/fixetcd.go
```diff
@@ -194,72 +193,55 @@ func getPeerPods(pods []corev1.Pod, de *degradedEtcd, cluster string) (string, e
 	return peerPods, nil
 }
 
-func newJobFixPeers(cluster, peerPods, deNode string) *unstructured.Unstructured {
-	const jobNameFixPeers = jobName + "fix-peers"
+func newPodFixPeers(cluster, peerPods, deNode string) *unstructured.Unstructured {
```
mrWinston: the argument `cluster` seems to be unused now. can you remove it?

pkg/frontend/fixetcd.go
```diff
@@ -194,72 +193,55 @@ func getPeerPods(pods []corev1.Pod, de *degradedEtcd, cluster string) (string, e
...
+	const podNameFixPeers = genericPodName + "fix-peers"
 	// Frontend kubeactions expects an unstructured type
-	jobFixPeers := &unstructured.Unstructured{
+	podFixPeers := &unstructured.Unstructured{
```
mrWinston: Just a nitpick, cause it's been that way before:
Instead of creating the pod definition as `unstructured.Unstructured` you can create it as a `&corev1.Pod` and then convert it to `unstructured.Unstructured`. This way we would have full editor support for the pod definition.

You could do the conversion like so:

```go
podFixPeers := &corev1.Pod{} // add neccessary fields
podUnstructuredContent, err := kruntime.DefaultUnstructuredConverter.ToUnstructured(podFixPeers)

if err != nil {} // handle err

return &unstructured.Unstructured{
		Object: podUnstructuredContent,
	}
```
Doing it like this should also make sure that the `metadata` is set correctly in the resulting object so we wouldn't need to use the helper functions to set them later.

pkg/frontend/fixetcd.go
alcasim: good catch! done.

pkg/frontend/fixetcd.go
alcasim: yes I also prefer doing it in the way you propose. I didn't want to change too much of the original code, but it's a good opportunity to do it.

pkg/frontend/fixetcd.go
```diff
@@ -303,25 +295,25 @@ func fixPeers(ctx context.Context, log *logrus.Entry, de *degradedEtcd, pods *co
...
+	containerLogs, err := waitForPodSucceed(ctx, log, watcher, podFixPeers, kubeActions)
 	if err != nil {
 		return containerLogs, err
 	}
 
-	log.Infof("Deleting %s now", jobFixPeers.GetName())
+	log.Infof("Deleting %s now", podFixPeers.GetName())
 	propPolicy := metav1.DeletePropagationBackground
-	err = kubeActions.KubeDelete(ctx, "Job", namespaceEtcds, jobFixPeers.GetName(), true, &propPolicy)
+	err = kubeActions.KubeDelete(ctx, "Pod", namespaceEtcds, podFixPeers.GetName(), true, &propPolicy)
```
mrWinston: Nitpick: consider using `podFixPeers.GetKind()` and `podFixPeers.GetNamespace()` as args to `KubeDelete`, just so it's obviously coupled to `podFixPeers`

pkg/frontend/fixetcd.go
```diff
@@ -194,89 +193,79 @@ func getPeerPods(pods []corev1.Pod, de *degradedEtcd, cluster string) (string, e
...
+		return nil, err
+	}
 
-	return jobFixPeers
+	return &unstructured.Unstructured{
+		Object: unstructuredPod,
+	}, nil
 }
 
 // fixPeers creates a job that ssh's into the failing pod's peer pods, and deletes the failing pod from it's member's list
```
mrWinston: This comment still mentions that a job is created.

pkg/frontend/adminactions/kubeactions.go
```diff
@@ -157,8 +158,8 @@ func (k *kubeActions) KubeWatch(ctx context.Context, o *unstructured.Unstructure
 	}
 
 	listOpts := metav1.ListOptions{
-		Limit:         1000, // just in case
-		LabelSelector: o.GetLabels()[labelKey],
+		LabelSelector: fmt.Sprintf("%v=%v", labelKey, o.GetLabels()[labelKey]),
+		Watch:         true,
```
dkeohane, line 162: Any concerns with removing the limit? even if it was only there 'just in case'

pkg/frontend/fixetcd.go
```diff
@@ -492,141 +484,162 @@ func createPrivilegedServiceAccount(ctx context.Context, log *logrus.Entry, name
 //
 // If backups already exists the job is cowardly and refuses to overwrite them
 func backupEtcdData(ctx context.Context, log *logrus.Entry, cluster, node string, kubeActions adminactions.KubeActions) ([]byte, error) {
-	jobDataBackup := createBackupEtcdDataJob(cluster, node)
+	podDataBackup, err := createBackupEtcdDataPod(node)
```
dkeohane, line 497: minor: Given you've removed cluster from use here, is it being used in the function anymore? If not, can probably remove from this function definition

pkg/frontend/fixetcd.go
```diff
@@ -492,141 +484,162 @@ func createPrivilegedServiceAccount(ctx context.Context, log *logrus.Entry, name
...
-	case event := <-watcher.ResultChan():
-		pod := event.Object.(*corev1.Pod)
-
-		switch pod.Status.Phase {
-		case corev1.PodSucceeded:
-			log.Infof("Job %s completed with %s", pod.GetName(), pod.Status.Message)
-		case corev1.PodFailed:
-			log.Infof("Job %s reached phase %s with message: %s", pod.GetName(), pod.Status.Phase, pod.Status.Message)
-			waitErr = fmt.Errorf("pod %s event %s received with message %s", pod.Name, pod.Status.Phase, pod.Status.Message)
+outer:
```
mrWinston: I mostly advise against using loop labels to break out of nested loops, as i find they make the code more confusing. But when using it in a `for{ select{...} }` block, i find them tolerable :) 

Alternatively, you could move this whole block to a separate function that just returns an error, as the block doesn't seem to have any other dependencies to what comes after it.

pkg/frontend/fixetcd.go
```diff
@@ -492,141 +484,162 @@ func createPrivilegedServiceAccount(ctx context.Context, log *logrus.Entry, name
...
+		case <-ctx.Done():
+			waitErr = fmt.Errorf("context was cancelled while waiting for %s because %s", o.GetName(), ctx.Err())
+			break outer
 		}
-	case <-ctx.Done():
-		waitErr = fmt.Errorf("context was cancelled while waiting for %s because %s", o.GetName(), ctx.Err())
 	}
 
-	// get container name
-	cxName := o.UnstructuredContent()["spec"].(map[string]interface{})["template"].(map[string]interface{})["spec"].(map[string]interface{})["containers"].([]corev1.Container)[0].Name
```
mrWinston, line 538: oof, I'm really happy you replaced that with something more structured 👍🏻

mrWinston (changes requested):
Looking really good! I would consider the comment about the comment the only thing that needs to be addressed to get this merged. The rest is optional nitpicking :) Thanks for your effort!

pkg/frontend/adminactions/kubeactions.go
alcasim: Limit shouldn't apply in Watch, as far as I know, as we would get a list of events as they happen, instead of calling the API and retrieving a long list that may require pagination.

pkg/frontend/fixetcd.go
alcasim: good catch! I'm removing it.

pkg/frontend/adminactions/kubeactions.go
mrWinston: yep, the kube apidocs explicitly state:

>This field [limit] is not supported if watch is true.

https://kubernetes.io/docs/reference/kubernetes-api/common-parameters/common-parameters/#limit

pkg/frontend/fixetcd.go
alcasim: I moved the loop into a separate function (I actually renamed the existing function so the name honors what it does). You can check this change in the latest commit.

pkg/frontend/fixetcd.go
alcasim: Done :)

pkg/frontend/fixetcd.go
alcasim: I missed that, good catch! done.

pkg/frontend/fixetcd.go
```diff
@@ -491,142 +483,174 @@ func createPrivilegedServiceAccount(ctx context.Context, log *logrus.Entry, name
...
-			log.Infof("Job %s completed with %s", pod.GetName(), pod.Status.Message)
-		case corev1.PodFailed:
-			log.Infof("Job %s reached phase %s with message: %s", pod.GetName(), pod.Status.Phase, pod.Status.Message)
-			waitErr = fmt.Errorf("pod %s event %s received with message %s", pod.Name, pod.Status.Phase, pod.Status.Message)
-		}
-	case <-ctx.Done():
-		waitErr = fmt.Errorf("context was cancelled while waiting for %s because %s", o.GetName(), ctx.Err())
+	err := waitForPodSucceed(ctx, log, watcher)
+
+	if err != nil {
```
mrWinston, line 570: Shouldn't this method try to get the podlogs even when `waitForPodSucceeded` returned an error? I think that's what `var waitErr error` was used for, to save the error and return it together with the logs at the end.

pkg/frontend/fixetcd.go
```diff
@@ -491,142 +493,182 @@ func createPrivilegedServiceAccount(ctx context.Context, log *logrus.Entry, name
...
-		case corev1.PodSucceeded:
-			log.Infof("Job %s completed with %s", pod.GetName(), pod.Status.Message)
-		case corev1.PodFailed:
-			log.Infof("Job %s reached phase %s with message: %s", pod.GetName(), pod.Status.Phase, pod.Status.Message)
-			waitErr = fmt.Errorf("pod %s event %s received with message %s", pod.Name, pod.Status.Phase, pod.Status.Message)
+	err := waitForPodSucceed(ctx, log, watcher)
+
+	if err != nil {
+		var failedErr *podFailedError
+		if !errors.As(err, &failedErr) {
```
mrWinston, line 572: Yes, i love that :)

mrWinston (approved):
LGTM!

PR #4362, opened 2025-09-03
Fix creation of UDR clusters

hawkowl:

Fixes ARO-20979

Fixes creation of UDR clusters by moving the zones param into the general properties. This is generally where we should keep it anyway.

### Test plan for issue:

Unit tests

N/A, doesn't have a matched exposed public API

It's simpler than what was there before :)

### DIFF
--- a/pkg/api/openshiftcluster.go
+++ b/pkg/api/openshiftcluster.go
@@ -140,6 +140,10 @@ type OpenShiftClusterProperties struct {
 	OperatorFlags   OperatorFlags `json:"operatorFlags,omitempty"`
 	OperatorVersion string        `json:"operatorVersion,omitempty"`
 
+	// Zones are the availability zones this cluster occupies
+	// Only new clusters have this set
+	Zones []string `json:"zones,omitempty"`
+
 	CreatedAt time.Time `json:"createdAt,omitempty"`
 
 	// CreatedBy is the RP version (Git commit hash) that created this cluster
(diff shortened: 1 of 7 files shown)


pkg/api/openshiftcluster.go
```diff
@@ -369,8 +373,6 @@ type LoadBalancerProfile struct {
 	OutboundIPs []OutboundIP `json:"outboundIps,omitempty"`
 	// The desired outbound IP Prefix resources for the cluster load balancer.
 	OutboundIPPrefixes []OutboundIPPrefix `json:"outboundIpPrefixes,omitempty"`
-	// The desired availability zones for the load balancer frontend IPs/RP-created PIPs
-	Zones []string `json:"outboundIpAvailabilityZones,omitempty"`
```
cadenmarchese, line 373: Can we accomplish this fix without moving `Zones` from LoadBalancerProfile to openShiftClusterProfile? While this field doesn't appear to be duplicated into our public APIs / swagger files, I'm not sure it makes sense to move it, since we're treating Zones as a property of the load balancer object.

cadenmarchese (commented):
I think I'm missing some context on the issue itself + the fix. Moving a field from one part of the API to another shouldn't change anything on its own - is there a pointer issue present?

pkg/api/openshiftcluster.go
hawkowl: Right now the only thing that uses it is the load balancer zones, but there's some future works regarding moving this into the RP to centralise that decision making (for weird regions like EUAP) rather than whack-a-moling special cases in various places. This just moves that before we actually use the field that's problematic here.

hawkowl:
@fahlmant https://github.com/Azure/ARO-RP/blob/master/pkg/api/defaults.go#L57-L63 is the cuplrit -- it seems like we made this a pointer because it seems like the intent is that this should only exist for clusters where we control the external routing, and me putting the internal LB zones in there seems to be overloading it somewhat.

pkg/api/openshiftcluster.go
cadenmarchese: Ok, if it's going to be used elsewhere in the future I think moving it makes sense.

fahlmant (approved):

PR #4304, opened 2025-07-08
[ARO-20049] Update deps for July 25

hawkowl:
### Which issue this PR addresses:
https://issues.redhat.com/browse/ARO-20049

### What this PR does / why we need it:

Updates golang dependencies

### Test plan for issue:

CI, E2E

### Is there any documentation that needs to be updated for this PR?

N/A

### How do you know this will function as expected in production? 

E2E

PR #4364, opened 2025-09-03
Bump golangci-lint to v2.2.1 in github actions

cadenmarchese:

No JIRA

https://github.com/Azure/ARO-RP/pull/4304 bumped the golangci-lint version to v2.2.1 - we need to do the same in github actions so that PR CI completes successfully.

### Test plan for issue:

- Green CI :)


kimorris27 (approved):
LGTM - golangci-lint check is already green.

PR #4342, opened 2025-08-11
Enabled Dv6 VMs for 4.19+

fahlmant:
### Which issue this PR addresses:

Fixes

PR #4369, opened 2025-09-08
[ARO-21041] Add lint/checking to pkg/api/ and fix a pkg/util/version reference

hawkowl:

Fixes ARO-21041

https://github.com/Azure/ARO-RP/pull/4342 added a dependency on the ARO-RP proper in pkg/api/ which means it can't be used outside of ARO-RP. This should have caused some sort of lint error, but it appears as if golangci-lint doesn't go into subpackages. This PR:

- moves parts of pkg/util/version/ that are needed by pkg/api into pkg/api/util/version
- makes it an interface instead of letting things get inside the struct
- adds linting to pkg/api/ and fixes lint errors
- reduces the object instantiation count in pkg/api/validate/vm.go
- runs go mod tidy on pkg/api which brings in a missing import

### Test plan for issue:

Unit test, e2e

N/A

Should be a functional no-op


pkg/monitor/cluster/clusterversions.go
```diff
@@ -72,16 +72,20 @@ func (mon *Monitor) prefetchClusterVersion(ctx context.Context) error {
 	}
 
 	av := actualVersion(cv)
-	mon.clusterActualVersion, err = version.ParseVersion(av)
+	avobj, err := version.ParseVersion(av)
 	if err != nil {
 		mon.log.Errorf("failure parsing ClusterVersion: %s", err.Error())
+	} else {
```
fahlmant, line 78: This else is not needed

kimorris27 (approved):

PR #4372, opened 2025-09-08
E2E for FixEtcd Geneva action

alcasim:

Fixes https://issues.redhat.com/browse/ARO-11486

It adds a new E2E Regression test that:

1. Temporary disables reconciliation
2. Checks if there are guardrails preventing machines to be deleted, and disables them if necessary
3. Deletes first master machine
4. Enables operators
5. Recreates Machine
6. Waits for new Etcd pod
7. Runs the fix
8. Waits until operators recover from degraded
9. Enables back guardrails if necessary

As with the [PR that fixed the action](https://github.com/Azure/ARO-RP/pull/4358), this is heavily based on https://github.com/Azure/ARO-RP/pull/4034

### Test plan for issue:

I ran this E2E test on a local cluster, I used the following commands:

```
make e2e-test
./e2e.test -test.v --ginkgo.vv --ginkgo.timeout 180m --ginkgo.flake-attempts=2 --ginkgo.junit-report=e2e-report.xml --ginkgo.focus="Master replacement"
```

No

This PR does not modify production itself, just the tests we run before deploying

### DIFF
--- a/test/e2e/fix_etcd.go
+++ b/test/e2e/fix_etcd.go
@@ -1,0 +1,202 @@
+package e2e
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/machine/v1beta1"
+
+	"github.com/Azure/ARO-RP/pkg/util/pointerutils"
+)
+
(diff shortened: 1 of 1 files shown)


test/e2e/fix_etcd.go
```diff
@@ -0,0 +1,202 @@
...
+	It("should fix etcd automatically", Serial, func(ctx context.Context) {
+		By("Disabling reconciliation")
+		dep, err := clients.Kubernetes.AppsV1().Deployments("openshift-cluster-version").Get(ctx, "cluster-version-operator", metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		dep.Spec.Replicas = pointerutils.ToPtr(int32(0))
+		_, err = clients.Kubernetes.AppsV1().Deployments("openshift-cluster-version").Update(ctx, dep, metav1.UpdateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		dep, err = clients.Kubernetes.AppsV1().Deployments("openshift-etcd-operator").Get(ctx, "etcd-operator", metav1.GetOptions{})
+		Expect(err).NotTo(HaveOccurred())
```
ehvs, line 54: Could we have clear error messages for clarity if the step has failed?

alcasim: This is handled by the E2E suite already, when a test fails it fully displays the error message returned by the application. What we do here is to validate that `err` is `nil`, if it's not, the test stops and shows the reason why it stopped.

PR #4373, opened 2025-09-09
Get FirstPartyUsage Tagged IP for RP VMSS and GWY VMSS

preetisht:

https://issues.redhat.com/browse/ARO-20272 

- VMSS IP Tags for both RP and Gateway
- Unified Configuration (vmssIpTags, vmssIpTagsDisabledRegions)
- ARM Conditional Logic to apply empty ipTag for disabled regions
- Edge Case Handling (empty arrays, missing config)
- Configuration picked up from RP-Config

Fixes

https://issues.redhat.com/browse/ARO-16025

This is a part of Microsoft Security Wave 5 fix. All our PublicIP's for both MSIT and AME environment for ARO Classic shoudl have FirstPartyUsage tags . 

This PR implements IP tags support for Azure Red Hat OpenShift Virtual Machine Scale Sets (VMSS) to comply with Microsoft Security Wave 5 requirements. The implementation adds unified configuration parameters (vmssIpTags, vmssIpTagsDisabledRegions) that enable region-specific IP tagging for both RP and Gateway VMSS public IPs, with ARM template conditional logic that gracefully handles disabled regions by deploying empty IP tags arrays. The solution has ARM conditionals logic and includes  edge case handling for missing or empty configuration scenarios.

### Test plan for issue:

- [x] make test.go passes all the test.
- [x] Deployed in Canary eastus2euap. Post deployment verify if the new VMSS instances both RP and GWY have proper tagged IP's. 
- [x] Create a  cluster in Canary and check if the cluster is sending metrics to geneva.
- [x] Log into both RP-VMSS and GWY-VMSS instance and check if some of the critical endpoints are accessable .  

- A doc will be created on how to deploy in regions where we do not FirstPartyUsage tags has not yet been provided to us by Microsoft . 
- We will create a TSG which will be a part of another PR which will be related to monitoring of our Public IP's which does not  have a FirstPartyUsage Tag .

- We will do an extensive tests in Canary region and updated all our test cases and results in JIRA . 
- [x] All the artifacts are uploaded and JIRA commets updated .

### DIFF
--- a/pkg/deploy/assets/rp-production.json
+++ b/pkg/deploy/assets/rp-production.json
@@ -215,6 +215,14 @@
             "type": "bool",
             "defaultValue": true
         },
+        "vmssIpTags": {
+            "type": "array",
+            "defaultValue": []
+        },
+        "vmssIpTagsDisabledRegions": {
+            "type": "array",
+            "defaultValue": []
+        },
         "vmssName": {
             "type": "string"
         }
@@ -438,7 +446,10 @@
                                                 },
                                                 "primary": true,
                                                 "publicIPAddressConfiguration": {
-                                                    "name": "rp-vmss-pip"
+                                                    "name": "rp-vmss-pip",
+                                                    "properties": {
+                                                        "ipTags": "[if(or(contains(parameters('vmssIpTagsDisabledRegions'), resourceGroup().location), equals(length(parameters('vmssIpTags')), 0)), createArray(), createArray(createObject('ipTagType', parameters('vmssIpTags')[0].type, 'tag', parameters('vmssIpTags')[0].value)))]"
+                                                    }
                                                 },
                                                 "loadBalancerBackendAddressPools": [
                                                     {
--- a/pkg/deploy/deploy_rp.go
+++ b/pkg/deploy/deploy_rp.go
@@ -76,6 +76,12 @@ func (d *deployer) DeployRP(ctx context.Context) error {
 	parameters.Parameters["vmssName"] = &arm.ParametersParameter{
 		Value: d.version,
 	}
+	parameters.Parameters["vmssIpTags"] = &arm.ParametersParameter{
+		Value: d.config.Configuration.VmssIpTags,
+	}
+	parameters.Parameters["vmssIpTagsDisabledRegions"] = &arm.ParametersParameter{
+		Value: d.config.Configuration.VmssIpTagsDisabledRegions,
+	}
 	parameters.Parameters["keyvaultDNSSuffix"] = &arm.ParametersParameter{
 		Value: d.env.Environment().KeyVaultDNSSuffix,
 	}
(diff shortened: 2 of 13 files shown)


wanghaoran1988 (approved):

pkg/deploy/deploy_rp.go
```diff
@@ -76,6 +76,12 @@ func (d *deployer) DeployRP(ctx context.Context) error {
 	parameters.Parameters["vmssName"] = &arm.ParametersParameter{
 		Value: d.version,
 	}
+	parameters.Parameters["vmssIpTags"] = &arm.ParametersParameter{
+		Value: d.config.Configuration.VmssIpTags,
```
tuxerrante, line 80: Could we normalize nil slices to empty arrays before assigning this parameter value? `getParameters` typically omits nil fields so template defaults apply, but explicitly setting a nil slice can serialize as `null`, which may break ARM expressions that call `length()` or index into `vmssIpTags`.

pkg/deploy/generator/templates.go
```diff
@@ -53,6 +53,8 @@ func (g *generator) templateFixup(t *arm.Template) ([]byte, error) {
 	b = bytes.ReplaceAll(b, []byte(`"throughput": `+strconv.Itoa(cosmosDbGatewayProvisionedThroughputHack)), []byte(`"throughput": "[parameters('cosmosDB').gatewayProvisionedThroughput]"`))
 	// pickZones doesn't work for regions that don't have zones.  We have created param nonZonalRegions in both rp and gateway and set default values to include all those regions.  It cannot be passed in-line to contains function, has to be created as an array in a parameter :(
 	b = bytes.ReplaceAll(b, []byte(`"zones": []`), []byte(`"zones": "[if(contains(parameters('nonZonalRegions'),toLower(replace(resourceGroup().location, ' ', ''))),'',pickZones('Microsoft.Network', 'publicIPAddresses', resourceGroup().location, 3))]"`))
+	// IP tags conditional logic: if region is disabled OR no IP tags configured, use empty array, otherwise use configured tags
+	b = bytes.ReplaceAll(b, []byte(`"ipTags": []`), []byte(`"ipTags": "[if(or(contains(parameters('vmssIpTagsDisabledRegions'), resourceGroup().location), equals(length(parameters('vmssIpTags')), 0)), createArray(), createArray(createObject('ipTagType', parameters('vmssIpTags')[0].type, 'tag', parameters('vmssIpTags')[0].value)))]"`))
```
tuxerrante, line 57: Would you consider normalizing `resourceGroup().location` here (for example similar to the non-zonal logic) before checking `vmssIpTagsDisabledRegions`? Exact-string comparison can be brittle and may leave tags enabled in regions that were intended to be disabled.

pkg/deploy/assets/rp-production.json
```diff
@@ -438,7 +446,10 @@
                                                 },
                                                 "primary": true,
                                                 "publicIPAddressConfiguration": {
-                                                    "name": "rp-vmss-pip"
+                                                    "name": "rp-vmss-pip",
+                                                    "properties": {
+                                                        "ipTags": "[if(or(contains(parameters('vmssIpTagsDisabledRegions'), resourceGroup().location), equals(length(parameters('vmssIpTags')), 0)), createArray(), createArray(createObject('ipTagType', parameters('vmssIpTags')[0].type, 'tag', parameters('vmssIpTags')[0].value)))]"
```
tuxerrante, line 451: Could we clarify whether only one IP tag is intentionally supported? This expression always reads index `[0]`, so additional configured tags would be ignored silently.

PR #4380, opened 2025-09-10
[ARO-21049] Unit testing the monitor run loop

mrWinston:

Fixes ARO-21049

Monitor barely has any unit testing, as it heavily depends on `changefeeds` from CosmosDB. This PR uses a very basic implementation of `changefeeds` in the fake client to add some base unit tests:

- Cache: most of the functions here have a unit test now (this didn't really needed `changefeed`).
- worker: the `changefeed` func is tested with some basic cases, as the current fake implementation doesn't work well with updates and deletes.
- monitor: it basically tests that all valid added clusters receive a "visit" from monitor.

In order to reduce wait times during tests, and to mock monitors (which are tested independently), we have modified the main monitor struct to allow injecting different monitors and tickers from the tests.

Note, this PR depends on https://github.com/bennerv/go-cosmosdb/pull/29 

### Test plan for issue:

This PR is solely focused on adding new tests. These tests have been run locally and from the CI

No, this PR is tech debt clean up.

We kept the same values for monitors and tickers when the code doesn't run from the tests, so this should be a no-op change from production perspective.

### DIFF
--- a/pkg/monitor/cache_test.go
+++ b/pkg/monitor/cache_test.go
@@ -17,7 +17,7 @@ type cacheTestOperation int
 const (
 	Upsert cacheTestOperation = iota
 	Delete
-	Noop
+	NoOp
 )
 
 func TestUpsertAndDelete(t *testing.T) {
@@ -166,7 +166,7 @@ func TestUpsertAndDelete(t *testing.T) {
 		},
 		{
 			name:      "test fixDoc - remove bucket ownership",
-			action:    Noop,
+			action:    NoOp,
 			clusterID: "cluster-3",
 			bucket:    2,
 			state:     api.ProvisioningStateSucceeded,
@@ -202,7 +202,7 @@ func TestUpsertAndDelete(t *testing.T) {
 				testMon.upsertDoc(doc)
 			case Delete:
 				testMon.deleteDoc(doc)
-			case Noop:
+			case NoOp:
 				// Do nothing, we don't need to call any func for the test to run
 			default:
 				t.Fatalf("unknown test action")


pkg/monitor/cache.go
```diff
@@ -64,9 +61,6 @@ func (mon *monitor) fixDoc(doc *api.OpenShiftClusterDocument) {
 	} else if ours && v.stop == nil {
 		ch := make(chan struct{})
 		v.stop = ch
-
-		delay := time.Duration(rand.Intn(60)) * time.Second
-
-		go mon.worker(ch, delay, doc.ID)
+		go mon.worker(ch, doc.ID)
```
bizz001, line 64: why are we removing the delay here?

pkg/monitor/monitor_test.go
```diff
@@ -0,0 +1,201 @@
...
+		go func() {
+			err := mon.Run(ctx)
+			if err != nil && err != context.DeadlineExceeded {
+				t.Logf("Unexpected error: %v", err)
+			}
+			wg.Done()
+		}()
+	}
+
+	time.Sleep(5 * time.Second)
```
bizz001, line 76: why do we need this sleep?

pkg/monitor/worker.go
```diff
@@ -112,6 +109,8 @@ func (mon *monitor) changefeed(ctx context.Context, baseLog *logrus.Entry, stop
 				ps := doc.OpenShiftCluster.Properties.ProvisioningState
 				fps := doc.OpenShiftCluster.Properties.FailedProvisioningState
 
+				mon.baseLog.Infof("Got cluster in changefeed: %s", doc.ID)
```
bizz001: this has the potential to flood the logs. I would remove it.

pkg/monitor/cache_test.go
```diff
@@ -0,0 +1,247 @@
...
+			},
+		},
+	}
+
+	// Execute operations in sequence on the same monitor
+	for _, op := range operations {
+		t.Run(op.name, func(t *testing.T) {
+			doc := createMockClusterDoc(op.clusterID, op.bucket, op.state)
+
+			switch op.action {
```
bizz001, line 200: the comment above those functions all mention this:  "Caller must hold mon.mu.Lock." Do we not need that here?

pkg/monitor/worker.go
alcasim: good catch, we added that get visibility on what was going on with the worker when it received docs from the changefeed, but it escaped my efforts to remove that logging before the final version. I will remove it.

pkg/monitor/monitor_test.go
alcasim: just to wait for `changefeed` to run once and let the above clusters to be added into the workers, before adding a last cluster. At the end we test that this cluster got checked by the monitors as well.

The idea is to ensure that no matter when the cluster is added, it should be monitored.

pkg/monitor/cache.go
alcasim: delay is now part of the `monitor` struct. I moved it there so we can manipulate it to run smaller delays in tests. Originally it was just defined inside the func, and therefore not reachable when we tried to run tests, we were forced to wait a random amount of time, which is not good for test speed.

pkg/monitor/cache_test.go
alcasim: it's a good observation. Those functions are indeed intended to be used by holders of the lock, when the entire monitor logic is running. However, from the unit test point, we're mostly interested to see that the function itself performs what it should do. By running it without locks from the monitor we just test that, without looking into handling concurrency, that happens in the `changefeed` goroutine. However, we can add the lock here if considered necessary.

jaitaiwan: I think we should be testing how we expect callers to implement. If we expect callers to lock then we should lock in the test.

alcasim: Ok, I will add the locks and a small concurrent test. I'm still not 100% convinced that we should add the locks as part of these functions tests, as it's responsibility of other functions (that we also test/try to test), but I'm not against adding it either.

About the concurrent test, it's a simple one just to "show" how it would look like, I would either reconvert the current test into a concurrent one, or add more cases to the concurrent test, suggestions are welcome :)

jaitaiwan: LGTM

pkg/monitor/cache_test.go
```diff
@@ -0,0 +1,325 @@
...
+	testMon := env.CreateTestMonitor("test-cache")
+	// Set owned buckets for the entire test sequence
+	ownedBuckets := []int{1, 2, 5}
+	for _, bucket := range ownedBuckets {
+		testMon.buckets[bucket] = struct{}{}
+	}
+
+	type operation struct {
+		name      string
+		action    string // "upsert", "delete", "fixDoc", "validate"
```
mociarain: Should this be an enum? Probably outside the scope of this PR but it's a thought.

pkg/monitor/master.go
```diff
@@ -31,6 +31,7 @@ func (mon *monitor) master(ctx context.Context) error {
 	if !mon.isMaster {
 		return nil
 	}
+	mon.baseLog.Info("I Am master")
```
mociarain:
```suggestion
	mon.baseLog.Info("I am master")
```

alcasim: I think I'm going to better remove this log line... we added it to figure things out but it's not needed

pkg/monitor/cache_test.go
alcasim: Added an enum

pkg/monitor/cache_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+	"github.com/Azure/ARO-RP/pkg/api"
+)
+
+type cacheTestOperation int
+
+const (
+	Upsert cacheTestOperation = iota
+	Delete
+	Noop
```
mociarain:
```suggestion
	NoOp
```

If I understood this correctly, this is clearer imo. WDYT?

PR #4388, opened 2025-09-16
Add red-hat-managed: true tag to install-config

fahlmant:

Fixes https://issues.redhat.com/browse/ARO-21174

Adds `red-hat-managed: true` tag to the install config for new clusters. This will propagate to the on-cluster `infrastructure` CR, which will then be used by OpenShift to add this tag to all Azure resources that are part of the cluster.

### DIFF
--- a/pkg/hive/install.go
+++ b/pkg/hive/install.go
@@ -35,6 +35,8 @@ const (
 platform:
   azure:
     region: "%s"
+    userTags:
+      red-hat-managed: "true"
 `
 )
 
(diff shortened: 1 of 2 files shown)


cadenmarchese (approved):
I have never personally tested modifying the install-config.yaml this way, however if it is working and you prefer to do it in ARO-RP I think it's fine.

fahlmant:
@cadenmarchese Unfortunately, it hasn't been working for me. Looks like I will need to put it in the wrapper.

PR #4389, opened 2025-09-17
[ARO-21064] Initial pass on bucketing for MIMO Actuator

hawkowl:

Part of ARO-21064

A first pass on bucketing for the MIMO Actuator.

Rather than implement the method that the Monitor does (which is rather complicated and would require another cosmosdb db/table), just do a naive approach which should roughly balance the cluster by the randomness of the bucket it's put in. There's almost certainly a better way (since it uses the VM number for it, which means us scaling up and deleting an old one would have weird results and we'd have an unbalanced third one), but this should be enough to build on for now -- a future approach similar to the monitor (but supporting several services) or reusing some common algorithm to autobalance in case of outages like the monitor does would be preferable.

### Test plan for issue:

Unit tests

Maybe?

Testing in INT when it gets to it

### DIFF
--- a/pkg/mimo/actuator/service.go
+++ b/pkg/mimo/actuator/service.go
@@ -326,6 +326,9 @@ out:
 func DetermineBuckets(env env.Core, hostnameFunc func() (string, error)) []int {
 	_log := env.Logger()
 
+	// We have a VMSS with 3 VMs in prod
+	vmCount := 3
+
 	b := []int{}
 	if !env.IsLocalDevelopmentMode() {
 		name, err := hostnameFunc()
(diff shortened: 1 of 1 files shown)


pkg/mimo/actuator/service.go
```diff
@@ -307,3 +320,48 @@ out:
...
+					if num > 3 || num <= 0 {
+						// Rather than guess, we fall back to all buckets. This
+						// means that a VMSS replacement of -4 might have some
+						// weird behaviour, but because we get a lock on the
+						// OpenShiftClusterObject before we do anything to the
+						// cluster, it should be fine.
+						_log.Warningf("vmss number is %d, currently only handles 3 partitions (vm numbers 1-3), falling back to all", num)
+					} else {
+						for i := num - 1; i < bucket.Buckets; i = i + 3 {
+							b = append(b, i)
```
wanghaoran1988: Let's do this and add a comment? 
```
// We have 3 VM instance in the VM Set, this is split all the openshift object into 3 pieces by the bucket id.
step := 3
for i := num - 1; i < bucket.Buckets; i += step {
    b = append(b, i)
}
```

hawkowl: I updated this and added some more comments.

PR #4390, opened 2025-09-17
Run ensureClusterMsiCertificate during MaintenanceTaskRenewCerts

cadenmarchese:

Fixes No Jira

- During a certificate renewal we should also be checking MIWI clusters for MSI certs in KV that may be eligible for renewal as they only last 90 days. We currently already do this on full update or adminUpdate - let's add it for certificate renewal task as well.

### Test plan for issue:

- Unit testing should be sufficient.

No

### DIFF
--- a/pkg/cluster/adminupdate_test.go
+++ b/pkg/cluster/adminupdate_test.go
@@ -85,6 +85,20 @@ func TestAdminUpdateSteps(t *testing.T) {
 		"[Action renewMDSDCertificate]",
 	}
 
+	managedIdentityCertificateRenewalSteps := []string{
+		"[Action startVMs]",
+		"[Condition apiServersReady, timeout 30m0s]",
+		"[Action populateDatabaseIntIP]",
+		"[Action correctCertificateIssuer]",
+		"[Action fixMCSCert]",
+		"[Action fixMCSUserData]",
+		"[Action configureAPIServerCertificate]",
+		"[Action configureIngressCertificate]",
+		"[Action initializeOperatorDeployer]",
+		"[Action renewMDSDCertificate]",
+		"[Action ensureClusterMsiCertificate]",
+	}
+
 	operatorUpdateSteps := []string{
 		"[Action startVMs]",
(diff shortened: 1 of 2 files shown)


pkg/cluster/adminupdate_test.go
```diff
@@ -247,7 +247,7 @@ func TestAdminUpdateSteps(t *testing.T) {
 				return doc, true
 			},
 			shouldRunSteps: utilgenerics.ConcatMultipleSlices(
-				zerothStepsManagedIdentity, generalFixesSteps, certificateRenewalSteps,
+				zerothStepsManagedIdentity, generalFixesSteps, append(certificateRenewalSteps, "[Action ensureClusterMsiCertificate]"),
```
tsatam: tiny suggestion to move this string next to the other steps defined at the top of the test file, and name it something like `managedIdentityClusterCertificateRenewalSteps` to make the intent here clear.

cadenmarchese: @tsatam done, thanks!

pkg/cluster/adminupdate_test.go
```diff
@@ -85,6 +85,20 @@ func TestAdminUpdateSteps(t *testing.T) {
...
+		"[Action populateDatabaseIntIP]",
+		"[Action correctCertificateIssuer]",
+		"[Action fixMCSCert]",
+		"[Action fixMCSUserData]",
+		"[Action configureAPIServerCertificate]",
+		"[Action configureIngressCertificate]",
+		"[Action initializeOperatorDeployer]",
+		"[Action renewMDSDCertificate]",
+		"[Action ensureClusterMsiCertificate]",
+	}
```
yithian: `managedIdentityCertificateRenewalSteps` is almost identical to `certificateRenewalSteps`, is there any way to deduplicate it?

```suggestion
	managedIdentityCertificateRenewalSteps := append(
	    certificateRenewalSteps,
		"[Action ensureClusterMsiCertificate]",
		)
```

PR #4396, opened 2025-09-22
[ARO-20791] add network security perimeter to some resources

mrWinston:

Fixes: https://issues.redhat.com/browse/ARO-21136

- Add Network Security Perimeters to CosmosDBs, Keyvaults and StoageAccounts having public access enabled
- required by security wave 5
- NSPs are in "Learning" mode, meaning they don't perform any filtering but enable us to monitor connections to these resources

Resources affected:
- aro-<location>-gwy
- aro-<location>-por
- aro-<location>-svc
- aro-<location>-cls
- aro-<location>-cosmosdb

### Test plan for issue:

- [ ] Deploy and Test in Int
- [ ] Deploy and Test in Canary


go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/keyvault/armkeyvault v1.5.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/msi/armmsi v1.3.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6 v6.2.0
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v7 v7.0.0
```
aasserzo, line 18: Do we need both versions?

aasserzo:
LGTM

go.mod
mrWinston: I don't know. I only switched to v7 in the files that i've touched so I don't start introducing changes in places completely unrelated to this PR.
I did a quick search through our codebase and there are still ~70 files that import the v6 version. Updating it everywhere would make reviewing this PR much harder.

mociarain:
I think you can just push an empty commit to make this check run. 

```
git commit --amend --no-edit;
git push --force-with-lease
```

pkg/deploy/generator/resources.go
```diff
@@ -44,6 +44,55 @@ func (g *generator) dnsZone(name string) *arm.Resource {
...
+			fmt.Sprintf("[resourceId('Microsoft.Network/networkSecurityPerimeters', '%s')]", nspName),
+		},
+	}
+}
+
+func (g *generator) networkSecurityPerimeterAssociation(nspName string, associationName string, targetResourceId string) *arm.Resource {
+	return &arm.Resource{
+		Resource: &armnetwork.NspAssociation{
+			Properties: &armnetwork.NspAssociationProperties{
+				AccessMode: pointerutils.ToPtr(armnetwork.AssociationAccessModeLearning),
```
alcasim, line 77: minor: if some day we want to move this to enforce or audit... would it be worth to have it as a func parameter? I like how clean looks right now, so it's not a deal breaker.

pkg/deploy/generator/resources.go
```diff
@@ -44,6 +44,55 @@ func (g *generator) dnsZone(name string) *arm.Resource {
...
+			Location:   pointerutils.ToPtr("[resourceGroup().location]"),
+			Properties: &armnetwork.SecurityPerimeterProperties{},
+			Name:       &name,
+			Type:       pointerutils.ToPtr("Microsoft.Network/networkSecurityPerimeters"),
+		},
+		APIVersion: azureclient.APIVersion("Microsoft.Network/networkSecurityPerimeters"),
+	}
+}
+
+// networkSecurityPerimeterProfile creates a new nsp profile with the hardcoded name `default`.
```
alcasim, line 59: reading the code, this func creates a resource with name `nspname/default` which is not exactly "hardcoded default".

pkg/deploy/generator/resources_rp.go
```diff
@@ -952,6 +981,12 @@ func (g *generator) rpCosmosDB() []*arm.Resource {
 		rs = append(rs, g.CosmosDBDataContributorRoleAssignment("'ARO'", "globalDevops"))
 	}
 
+	cosmosNSP := g.networkSecurityPerimeter("cosmos-nsp")
+	cosmosNSPProfile := g.networkSecurityPerimeterProfile("cosmos-nsp")
+	cosmosNSPAssociation := g.networkSecurityPerimeterAssociation("cosmos-nsp", "cosmos-nsp-association", "[resourceId('Microsoft.DocumentDB/databaseAccounts', parameters('databaseAccountName'))]")
+
+	rs = append(rs, cosmosNSP, cosmosNSPProfile, cosmosNSPAssociation)
```
alcasim, line 994: According to this https://learn.microsoft.com/en-us/azure/private-link/network-security-perimeter-concepts#onboarded-private-link-resources Cosmos DB is not GA, but on "Public Preview"... will it work in all of our regions?

PR #4397, opened 2025-09-23
Fix `make az` and `make image-autorest` Makefile targets

cadenmarchese:

Fixes No Jira

- `make az` sets dev_sources to the repository root which doesn't work if you're trying to use the local CLI (it needs to point to ./python)
- `make image-autorest` fails due to recursive variables that don't appear to be needed.
- This PR is meant to fix both issues.

### Test plan for issue:

- Tested the Makefile targets

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,8 +32,7 @@ else
 	VERSION = $(TAG)
 endif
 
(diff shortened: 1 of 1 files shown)


yithian (commented):
lgtm

PR #4400, opened 2025-09-25
Proposal For ADR-255

hlipsig:

This PR introduces a Proof of Concept to demonstrate how ADR 255 would work.  What tests are required for it, and documentation. This code was generated by Claude and should not be merged as is. 

Proof of concept.
### Test plan for issue:
Local testing and claude added unit and e2e tests. 

Developer documentation and processes beyond what are in this PR would need to be updated if we choose to go the route of this proof of concept.

Can test in Canary, but ultimate if we go this route we approve the test accounts for the AFEC flag and we'll know in short order.


hlipsig:
Comments from Amber. We can fix the needs to not go through ACR with her WIP PR, which will allow us to simplify this, and also makes our mirroring better. Once that's over the line, we can rely on tags such as "nightly" to determine if a version can be installed during validation of the input.

PR #4403, opened 2025-09-29
Use secret tags to determine MSI certificate validity

cadenmarchese:

Fixes

- When the msi-dataplane returns a secret value that ARO-RP can persist to MSI keyvault, [it also sets some tags](https://github.com/Azure/msi-dataplane/blob/main/pkg/dataplane/keyvault.go#L92-L93) on the secret (the useful ones are `cannot_renew_after` and `renew_after`)
-[ These tags are then used by Credential Refresher to determine if a cert can be refreshed or not](https://msazure.visualstudio.com/AzureRedHatOpenShift/_git/Credential-Refresher?path=/internal/refresher/refresher.go&version=GBmain&line=186&lineEnd=210&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents). For now we don't have the Credential Refresher and we're refreshing certs in RP, but we are using hardcoded date values which isn't reliable as the MSI-RP can always decide to change those.
- The ARO-RP should adopt the same approach as the Credential Refresher when refreshing, and use the tags that the msi-dataplane added to the secrets instead.

Also:

- Refactor the function ensureClusterMsiCertificate because it's getting too hard to read.
- Adjust federated credential deletion to not ensureClusterMsiCertificate if the PWIP is empty, and remove the unit test cases that were testing this functionality.

### Test plan for issue:

- Unit testing - MSI interactions cannot be tested in dev.


pkg/cluster/clustermsi.go
```diff
@@ -102,11 +79,74 @@ func (m *manager) ensureClusterMsiCertificate(ctx context.Context) error {
...
+	if len(keyvaultCredentials.ExplicitIdentities) == 0 {
+		return true, nil
+	}
+
+	// Check if the secret is for a different identity (e.g., after a cluster update).
+	clusterMsiResourceId, err := m.doc.OpenShiftCluster.ClusterMsiResourceId()
+	if err != nil {
+		return false, err
+	}
+	if *keyvaultCredentials.ExplicitIdentities[0].ResourceID != clusterMsiResourceId.String() {
```
kimorris27: Nit: I'd check this pointer for nil to be safe.

pkg/cluster/clustermsi.go
```diff
@@ -102,11 +79,74 @@ func (m *manager) ensureClusterMsiCertificate(ctx context.Context) error {
...
+	// In the future, certificate refreshing will be handled by the Certificate Refresher. For now, handle it here.
+	refreshNeeded, err := m.needsRefresh(cert, now)
+	if err != nil {
+		return false, err
+	}
+	if refreshNeeded {
+		return true, nil
+	}
+
+	return false, nil
```
kimorris27: To simplify:

```suggestion
	return m.needsRefresh(cert, now)
```

pkg/cluster/clustermsi.go
```diff
@@ -102,11 +79,74 @@ func (m *manager) ensureClusterMsiCertificate(ctx context.Context) error {
...
+
+		parsedTime, err := time.Parse(time.RFC3339, *valuePtr)
+		if err != nil {
+			return false, fmt.Errorf("invalid time format for tag %s: %w", tagKey, err)
+		}
+
+		*timeVarPtr = parsedTime
+	}
+
+	inRenewalWindow := !renewAfter.After(now) && !now.After(cannotRenewAfter)
```
kimorris27: For readability:

```suggestion
	inRenewalWindow := now.After(renewAfter) && now.Before(cannotRenewAfter)
```

pkg/cluster/clustermsi_test.go
```diff
@@ -290,71 +239,189 @@ func TestEnsureClusterMsiCertificate(t *testing.T) {
...
+						dataplane.CannotRenewAfterKeyVaultTag: pointerutils.ToPtr(now.Format(time.RFC3339)),
 					},
 				},
+			},
+			currentTime: now,
+			wantBool:    true,
+			wantErr:     false,
+		},
+		{
+			name:        "tags are nil",
```
kimorris27:
```suggestion
			name:        "error - tags are nil",
```

pkg/cluster/clustermsi_test.go
```diff
@@ -290,71 +239,189 @@ func TestEnsureClusterMsiCertificate(t *testing.T) {
...
+			wantBool:    false,
+			wantErr:     false,
 		},
-	}
-
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			secret := azsecrets.GetSecretResponse{
+		{
+			name: "exactly at renewal time",
```
kimorris27:
```suggestion
			name: "success - exactly at renewal time",
```

pkg/cluster/clustermsi_test.go
```diff
@@ -290,71 +239,189 @@ func TestEnsureClusterMsiCertificate(t *testing.T) {
...
+						dataplane.CannotRenewAfterKeyVaultTag: pointerutils.ToPtr(expireTime),
+					},
+				},
+			},
+			currentTime: now,
+			wantBool:    true,
+			wantErr:     false,
+		},
+		{
+			name: "exactly at expiration time",
```
kimorris27:
```suggestion
			name: "success - exactly at expiration time",
```

pkg/cluster/delete.go
```diff
@@ -395,8 +395,15 @@ func (m *manager) deleteFederatedCredentials(ctx context.Context) error {
 		return nil
 	}
 
+	// If there's nothing in the platform workload identity profile, there's nothing to do.
+	if m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile == nil ||
+		len(m.doc.OpenShiftCluster.Properties.PlatformWorkloadIdentityProfile.PlatformWorkloadIdentities) == 0 {
+		m.log.Info("no platform workload identities found, skipping federated credential deletion")
+		return nil
+	}
+
```
kimorris27: We can skip this since we already check `!m.doc.OpenShiftCluster.UsesWorkloadIdentity()` a few lines above here.

pkg/cluster/clustermsi.go
cadenmarchese: Yeah, this is hard to read. But your suggestion causes the "exactly at expiration time" and "exactly at renewal time" test cases to fail, because it changes the logic to require that the time is at some point after to be renewed.

This scenario is basically impossible though, so we can adopt your logic and get rid of those unit test cases if you think that's better.

kimorris27: I'm more concerned with readability, so my vote would be to remove those unit test cases. I agree that it's close enough to impossible for us to try renewing at one of those exact times.

jaitaiwan: Before we go ahead with this, can we comment somewhere why we're removing the exactly case? I'd hate for us to hit that case when this is being trigged by the scheduler and it takes longer to figure out why we made this choice.

pkg/cluster/clustermsi.go
```diff
@@ -102,11 +79,67 @@ func (m *manager) ensureClusterMsiCertificate(ctx context.Context) error {
...
+		return true, nil
+	}
+
+	// Check if the secret is for a different identity (e.g., after a cluster update).
+	clusterMsiResourceId, err := m.doc.OpenShiftCluster.ClusterMsiResourceId()
+	if err != nil {
+		return false, err
+	}
+	if keyvaultCredentials.ExplicitIdentities[0].ResourceID == nil ||
+		*keyvaultCredentials.ExplicitIdentities[0].ResourceID != clusterMsiResourceId.String() {
```
edisonLcardenas, line 102: I think if the ResourceID is nil, the second part of the OR condition will panic when dereferencing *keyvaultCredentials.ExplicitIdentities[0].ResourceID. Probably do:

resourceID := keyvaultCredentials.ExplicitIdentities[0].ResourceID
if resourceID == nil || *resourceID != clusterMsiResourceId.String() {
    return true, nil
}

pkg/cluster/clustermsi.go
```diff
@@ -102,11 +79,67 @@ func (m *manager) ensureClusterMsiCertificate(ctx context.Context) error {
...
+	tagsToParse := map[string]*time.Time{
+		dataplane.RenewAfterKeyVaultTag:       &renewAfter,
+		dataplane.CannotRenewAfterKeyVaultTag: &cannotRenewAfter,
+	}
+
+	for tagKey, timeVarPtr := range tagsToParse {
+		valuePtr, ok := item.Tags[tagKey]
+		if !ok || valuePtr == nil {
+			return false, fmt.Errorf("missing or invalid tag: %s", tagKey)
+		}
```
edisonLcardenas, line 130: aren't there any legitimate scenarios wherein missing tags are fine?

cadenmarchese: Not for an MSI cert, no - the msi-dataplane should always be persisting them, if they are not persisted, it would be a bug.

pkg/cluster/clustermsi.go
cadenmarchese: I actually think we're good here, according to https://go.dev/ref/spec#Logical_operators, the second operand is only evaluated if the first one is not true. That means that we'd only check `*keyvaultCredentials.ExplicitIdentities[0].ResourceID != clusterMsiResourceId.String()` if `keyvaultCredentials.ExplicitIdentities[0].ResourceID == nil` is false, so we should be safe from nil pointer.

pkg/cluster/clustermsi.go
cadenmarchese: I added a comment to this line indicating that we need to be within the window to renew (and can't be exactly at time, although this should be nearly impossible)

kimorris27 (approved):

kevinobriendotca (approved):
LGTM

PR #4407, opened 2025-10-01
Fix ACR authentication issue for ubi9/toolbox in CI pipelines

shubhadapaithankar:
**Root Cause**: Azure DevOps tries to pull container images during pipeline initialization, before any `task: Docker@2` login steps execute.

**Solution**: Changed container resource definitions to use public Red Hat registry (`registry.access.redhat.com/ubi9/toolbox`) which doesn't require authentication and is the original source for these base images.

**Files Changed**:
- `.pipelines/e2e.yml` - Updated container image reference (line 24)
- `.pipelines/rp-full-dev-setup.yml` - Updated container image reference (line 12)

### Test plan for issue:
- E2E pipelines should initialize successfully without authentication errors
- Container initialization step should complete without Docker pull failures  
- No functional changes - same image content, different registry source

No. This is a pipeline infrastructure fix that uses the original public image sources instead of internal mirrors.

### DIFF
--- a/.pipelines/e2e.yml
+++ b/.pipelines/e2e.yml
@@ -22,6 +22,7 @@ resources:
   containers:
     - container: container
       image: arointsvc.azurecr.io/ubi9/toolbox
+      endpoint: arointsvc
       options: --user=0 --privileged -v /dev/shm:/dev/shm --device /dev/net/tun --name vpn
 
 # Azure DevOps Pipeline running e2e tests
--- a/.pipelines/rp-full-dev-setup.yml
+++ b/.pipelines/rp-full-dev-setup.yml
@@ -10,6 +10,7 @@ resources:
   containers:
     - container: container
       image: arointsvc.azurecr.io/ubi9/toolbox
+      endpoint: arointsvc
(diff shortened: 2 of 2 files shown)


slawande2 (approved):
Thanks for the quick fix!

tsatam (dismissed):

.pipelines/e2e.yml
```diff
@@ -21,7 +21,7 @@ pr:
 resources:
   containers:
     - container: container
-      image: arointsvc.azurecr.io/ubi9/toolbox
+      image: registry.access.redhat.com/ubi9/toolbox
       options: --user=0 --privileged -v /dev/shm:/dev/shm --device /dev/net/tun --name vpn
```
cadenmarchese, line 26: Assuming that the image exists in ACR, referencing the service connection should also fix this issue and avoid the external registry:
```
  containers:
    - container: container
      image: arointsvc.azurecr.io/ubi9/toolbox
      endpoint: arointsvc
      options: --user=0 --privileged -v /dev/shm:/dev/shm --device /dev/net/tun --name vpn
```

shubhadapaithankar: Perfect! 🎯 That's exactly the right solution - thank you!

I've updated the PR with your fix:

PR #4410, opened 2025-10-03
use-same-e2e-code-for-miwi-and-csp

slawande2:
[ARO-21864](https://issues.redhat.com/browse/ARO-21864)
Fixes

The MIWI e2e pipeline is currently failing. This PR uses the same CSP e2e code and incorporates MIWI-specific requirements to enable a functional e2e pipeline for MIWI.

no

### DIFF
--- a/.pipelines/e2e.yml
+++ b/.pipelines/e2e.yml
@@ -1,3 +1,5 @@
+# Azure DevOps Pipeline running CI for MIWI
+
 trigger:
   branches:
     include:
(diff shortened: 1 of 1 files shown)


.pipelines/e2e.yml
```diff
@@ -18,190 +20,299 @@ pr:
...
+              export ARO_ADOPT_BY_HIVE="true"
+              export HIVE_KUBE_CONFIG_PATH="$(realpath ./aks.kubeconfig)"
+              
+              . ./env
+              . secrets/env
+              . ./hack/e2e/run-rp-and-e2e.sh
+              az acr login --name arosvcdev
+              az acr login --name arointsvc
+
+              run_vpn
```
cadenmarchese: Not sure if this is required since we're now using the `docker compose` method.

.pipelines/e2e.yml
```diff
@@ -18,190 +20,299 @@ pr:
...
+              export SKIP_MIWI_ROLE_ASSIGNMENT=true
+              . ./hack/devtools/local_dev_env.sh
+              create_miwi_env_file
+            displayName: Setup (Azure)
+            
+          # Run the E2E test suite
+          - bash: |
+              export USE_WI=true
+              export ARO_INSTALL_VIA_HIVE="true"
+              export ARO_ADOPT_BY_HIVE="true"
```
cadenmarchese: This shouldn't be needed as the `docker compose` method should already be installing via Hive.

.pipelines/e2e.yml
```diff
@@ -18,190 +20,299 @@ pr:
...
+              az account set -s $AZURE_SUBSCRIPTION_ID
+              SECRET_SA_ACCOUNT_NAME=$(SECRET_SA_ACCOUNT_NAME) make secrets
+              . secrets/env
+              . ./hack/e2e/run-rp-and-e2e.sh
+              deploy_e2e_db
+
+              export SKIP_MIWI_ROLE_ASSIGNMENT=true
+              . ./hack/devtools/local_dev_env.sh
+              create_miwi_env_file
+            displayName: Setup (Azure)
```
cadenmarchese: For lines 234-238, that should not be required as we are already doing those steps above.

For lines 240-242, those lines are required, however I think it would make more sense to put them in the following `script` instead (right before we execute the e2e using `docker compose`)

.pipelines/e2e.yml
```diff
@@ -18,190 +20,299 @@ pr:
...
+              . ./hack/e2e/run-rp-and-e2e.sh
+              az acr login --name arosvcdev
+              az acr login --name arointsvc
+
+              run_vpn
+
+              update_role_sets
+
+              run_rp
+              validate_rp_running
```
cadenmarchese: This is being handled by docker compose, so no need to include it.

.pipelines/e2e.yml
```diff
@@ -18,190 +20,299 @@ pr:
...
-          . ./hack/e2e/run-rp-and-e2e.sh
-
-          export KUBECONFIG=admin.kubeconfig
-
-          # retrieve the oc cli
-          wget -nv https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/$(OpenShiftCLIVersion)/openshift-client-linux-$(OpenShiftCLIVersion).tar.gz
-          tar xf openshift-client-linux-$(OpenShiftCLIVersion).tar.gz
+      TAG: $[ stageDependencies.Set_Tag_Stage.Set_Tag_Variable.outputs['SetTag.TAG'] ]
+    jobs:
+      - job: Run_E2E_Tests
```
cadenmarchese: My recommended approach to start would be to just copy/paste the existing CSP e2e verbatim: https://github.com/Azure/ARO-RP/blob/master/.pipelines/ci.yml#L174-L292

Then add the steps required for MIWI to the step where we run the docker compose (but before we run it):

- export USE_WI=true
- export SKIP_MIWI_ROLE_ASSIGNMENT=true
- create_miwi_env_file
- update_role_sets
- `. ./env`
- any others I'm forgetting :)

.pipelines/e2e.yml
```diff
@@ -18,190 +20,214 @@ pr:
...
+                echo "Kubeconfig retrieved successfully."
+              else
+                echo "Failed to retrieve Kubeconfig."
+                exit 1
+              fi
+            displayName: Get Kubeconfig for AKS Cluster
+
+          # Run the E2E test suite
+          - bash: |
+              export CI=true
```
cadenmarchese: We seem to already be doing this on line 141-143 here. I see it's set as a pipeline variable here. We should check to see if we can avoid re-setting this here.

.pipelines/ci.yml
```diff
@@ -248,7 +248,131 @@ stages:
...
+              . ./hack/devtools/local_dev_env.sh
+              create_miwi_env_file
+              echo "created miwi env file successfully"
+              . ./env
+              az acr login --name arosvcdev
+              az acr login --name arointsvc
+              deploy_e2e_db
+              register_sub
+              docker compose up --quiet-pull --exit-code-from e2e e2e
+              docker compose logs e2e
```
shubhadapaithankar: Move docker compose logs e2e to the "Log Service Output" step (line 378) or remove it - it's redundant since you log it at line 383.

.pipelines/ci.yml
```diff
@@ -248,7 +248,131 @@ stages:
...
+              export SKIP_MIWI_ROLE_ASSIGNMENT=true
+              . ./hack/devtools/local_dev_env.sh
+              create_miwi_env_file
+              echo "created miwi env file successfully"
+              . ./env
+              az acr login --name arosvcdev
+              az acr login --name arointsvc
+              deploy_e2e_db
+              register_sub
+              docker compose up --quiet-pull --exit-code-from e2e e2e
```
shubhadapaithankar, line 384: Add docker compose build vpn before the docker compose up command (like CSP does at line 241).

.pipelines/ci.yml
```diff
@@ -248,7 +248,131 @@ stages:
...
+              export CI=true
+              export USE_WI=true
+              . ./env
+              . secrets/env
+              . ./hack/e2e/run-rp-and-e2e.sh
+              export SKIP_MIWI_ROLE_ASSIGNMENT=true
+              . ./hack/devtools/local_dev_env.sh
+              create_miwi_env_file
+              echo "created miwi env file successfully"
+              . ./env
```
shubhadapaithankar, line 377: Remove the duplicate . ./env - you already sourced it at line 353.

slawande2: This will be needed to source the `PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS` and MOCK MSI env variables that are set in the `create_miwi_env_file` step.

.pipelines/ci.yml
```diff
@@ -248,7 +248,131 @@ stages:
...
+                echo "Kubeconfig retrieved successfully."
+              else
+                echo "Failed to retrieve Kubeconfig."
+                exit 1
+              fi
+            displayName: Get Kubeconfig for AKS Cluster
+
+          # Run the E2E test suite
+          - bash: |
+              export CI=true
```
cadenmarchese: We already set this env var on lines 323-325, so I don't think we need to set it again.

.pipelines/ci.yml
```diff
@@ -248,7 +248,131 @@ stages:
...
+              export USE_WI=true
```
cadenmarchese: Should we try setting this in a dedicated stage like we do the `CI=true`? Something like:
```
          - bash: |
              echo "##vso[task.setvariable variable=USE_WI]true"
            displayName: Set USE_WI=true
```
It would make it more explicit that we're using WI in the pipeline.

.pipelines/ci.yml
```diff
@@ -248,7 +248,131 @@ stages:
...
+              . ./env
```
cadenmarchese: You can remove this line, as the `env` file doesn't exist yet at the point where this is ran. See the log message: https://msazure.visualstudio.com/AzureRedHatOpenShift/_build/results?buildId=140297269&view=logs&j=b80e3ff7-0805-55f4-2a94-911601e325b1&t=0e82bb6e-8fa6-5c62-5689-21c8e19655c1&l=12

.pipelines/ci.yml
```diff
@@ -248,7 +248,131 @@ stages:
...
+              . ./hack/devtools/local_dev_env.sh
+              create_miwi_env_file
+              echo "created miwi env file successfully"
+              . ./env
+              az acr login --name arosvcdev
+              az acr login --name arointsvc
+              deploy_e2e_db
+              register_sub
+              docker compose build vpn
+              docker compose up --quiet-pull --exit-code-from e2e e2e
```
cadenmarchese, line 384: We're still getting a cluster service principal here, and I think it might be because `docker compose` is not passing in our local environment variables into the container that it's starting up. We're setting them correctly _in the pipeline_, but they never make it to the RP running inside a container when we do `docker compose up`. Maybe @shubhadapaithankar is able to confirm this.

To fix this we will probably need to make some changes to the docker-compose.yml file, specifically the `services.e2e.environment` section. This environment section should also contain the env vars we need to create a MIWI cluster.

It's also possible that our vars are getting passed in and there's something else going on, but let's start with the docker compose.

.pipelines/e2e.yml
cadenmarchese: This comment is left over from an old review, I'm resolving it as you've deleted that file.

.pipelines/ci.yml
cadenmarchese: We can remove the first one (line 353). The one here is required 👍

.pipelines/ci.yml
shubhadapaithankar: Confirmed! The MIWI environment variables need to be added to the **`docker-compose.yml`** file.

**File:** `docker-compose.yml`

**Location:** In the `services.e2e.environment` section (around line 176)

Add the following MIWI-related environment variables:

```yaml
    environment:
      <<: *common-env
      ARO_SELENIUM_HOSTNAME:
      CLUSTER:
      E2E_DELETE_CLUSTER:
      E2E_LABEL:
      MOCK_MSI_CERT:
      MOCK_MSI_CLIENT_ID:
      MOCK_MSI_OBJECT_ID:
      MOCK_MSI_TENANT_ID:
      OS_CLUSTER_VERSION:
      MASTER_VM_SIZE:
      WORKER_VM_SIZE:
      PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS:
      PORTAL_HOSTNAME: "https://localhost:8444"
      RP_BASE_URL: "https://localhost:8443"
```
These variables are already configured in the `rp` service (lines 110-114) and need to be added to the `e2e` service as well to ensure they're properly passed into the e2e container for MIWI cluster creation.

.pipelines/ci.yml
```diff
@@ -262,6 +264,144 @@ stages:
 
           # Collect must-gather logs
           - bash: |
+              . secrets/env
+              go build ./hack/db
```
cadenmarchese, line 273: nit - can we leave a comment for future contributors who may be wondering why we need to explicitly build this?

.pipelines/ci.yml
```diff
@@ -262,6 +264,144 @@ stages:
...
+            displayName: Set CI=true
+
+          # Override the E2E label for IndividualCI/BatchedCI (i.e. not manually
+          # ran/PR jobs) to run all non-smoke tasks (default is !smoke&&!regressiontest)
+          - bash: |
+              echo "##vso[task.setvariable variable=E2E_LABEL]!smoke"
+            displayName: Enable regression tests in CI
+            condition: in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI')
+
+          # Get Kubeconfig for AKS Cluster with corrected Key Vault configuration
```
cadenmarchese: What is the corrected key vault configuration?

.pipelines/ci.yml
```diff
@@ -262,6 +264,144 @@ stages:
...
+
+          # Run the E2E test suite
+          - bash: |
+              echo "##vso[task.setvariable variable=USE_WI]true"
+            displayName: Set USE_WI=true
+          - bash: |
+              . secrets/env
+              . ./hack/e2e/run-rp-and-e2e.sh
+              echo "##vso[task.setvariable variable=CLUSTER]${CLUSTER}"
+              echo "###vso[task.setvariable variable=DATABASE_NAME]{$DATABASE_NAME}"
```
cadenmarchese: Would it make more sense to have these as separate pipeline steps like you did for setting `USE_WI`? I'm in favor for keeping the bash scripts as minimal as possible and relying as much as we can on the Azure Pipelines tooling.

.pipelines/ci.yml
```diff
@@ -262,6 +264,144 @@ stages:
...
+              export SKIP_MIWI_ROLE_ASSIGNMENT=true
```
cadenmarchese, line 373: It would be helpful to leave a comment on why we need this env var for a MIWI pipeline specifically.

.pipelines/ci.yml
```diff
@@ -262,6 +264,144 @@ stages:
...
+              . ./hack/devtools/local_dev_env.sh
+              create_miwi_env_file
+              echo "created miwi env file successfully"
```
cadenmarchese: I think this log line would log even if the `create_miwi_env_file` was unsuccessful, if we need this log line, we should ensure that it's only ran if we get a successful exit code from `create_miwi_env_file`. Or for simplicity's sake we could also just remove this log line.

pkg/util/cluster/cluster.go
```diff
@@ -1017,6 +1048,73 @@ func (c *Cluster) ensureDefaultVersionInCosmosdb(ctx context.Context) error {
...
+				c.log.Infof("ensureDefaultRoleSetInCosmosdb: PlatformWorkloadIdentityRoleSet for version %s already in DB; skipping PUT", defaultVersion.Version.MinorVersion())
+				return nil
+			}
+		}
+	}
+
+	c.log.Infof("building default payload for OpenShift version %s", defaultVersion.Version.MinorVersion())
+
+	var roleSets []api.PlatformWorkloadIdentityRoleSetProperties
+	if err := json.Unmarshal([]byte(c.Config.WorkloadIdentityRoles), &roleSets); err != nil {
```
cadenmarchese, line 1083: Regarding your most recent CSP failure, it looks like you are trying to run this function on CSP runs as well (which is good), but the `c.Config.WorkloadIdentityRoles` is only populated if this is a MIWI cluster because it gets its value from the `PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS` env var. Since we don't set that env var on a CSP run, we get the `failed to unmarshal platform workload identity role sets from config: unexpected end of JSON input` error.

I know we talked about always populating PWIRS no matter what type of cluster is being installed, but at this point I'm most interested in simplicity/debuggability since we're increasing pipeline complexity. We would have to re-do some of the MIWI setup steps in the CSP run to get `PLATFORM_WORKLOAD_IDENTITY_ROLE_SETS` env var set which wouldn't make sense to the reader. As a result I think we should just revert to the way it was before, where we only set PWIRS if it's a MIWI run.

.pipelines/ci.yml
slawande2: I'll add a log line before the `create_miwi_env_file` something like "creating MIWI env file", I think that will be helpful

.pipelines/ci.yml
slawande2: updated the comment to just `Get Kubeconfig for AKS cluster`

.pipelines/ci.yml
```diff
@@ -231,6 +231,11 @@ stages:
             displayName: Get Kubeconfig for AKS Cluster
 
           # Run the E2E test suite
+          - bash: |
+              . ./hack/e2e/run-rp-and-e2e.sh
+              echo "##vso[task.setvariable variable=CLUSTER]${CLUSTER}"
+              echo "##vso[task.setvariable variable=DATABASE_NAME]{$DATABASE_NAME}"
```
shubhadapaithankar: Can we update the DATABASE_NAME variable export syntax on both CSP (Line 237) and MIWI (Line 364)?
Current:
`echo "##vso[task.setvariable variable=DATABASE_NAME]{$DATABASE_NAME}"`
Suggested:
`echo "##vso[task.setvariable variable=DATABASE_NAME]${DATABASE_NAME}"`

 The current syntax {$DATABASE_NAME} has the $ and { in the wrong order for bash variable expansion. In bash, {$VARIABLE} treats the opening brace as a literal character and only expands `$VARIABLE}`, which can lead to unexpected results. The correct syntax is ${VARIABLE}.

.pipelines/ci.yml
```diff
@@ -248,7 +253,147 @@ stages:
...
+
+          # Run the E2E test suite
+          - bash: |
+              echo "##vso[task.setvariable variable=USE_WI]true"
+            displayName: Set USE_WI=true
+          - bash: |
+              . secrets/env
+              . ./hack/e2e/run-rp-and-e2e.sh
+              echo "##vso[task.setvariable variable=CLUSTER]${CLUSTER}"
+              echo "##vso[task.setvariable variable=DATABASE_NAME]{$DATABASE_NAME}"
```
shubhadapaithankar: Can we update the DATABASE_NAME variable export syntax  MIWI (Line 364)?
Current:
echo "##vso[task.setvariable variable=DATABASE_NAME]{$DATABASE_NAME}"
Suggested:
echo "##vso[task.setvariable variable=DATABASE_NAME]${DATABASE_NAME}"

The current syntax {$DATABASE_NAME} has the $ and { in the wrong order for bash variable expansion. In bash, {$VARIABLE} treats the opening brace as a literal character and only expands `$VARIABLE}`, which can lead to unexpected results. The correct syntax is ${VARIABLE}.

.pipelines/ci.yml
slawande2: great catch! I'll update it.

.pipelines/ci.yml
```diff
@@ -248,7 +253,147 @@ stages:
...
+              docker compose logs e2e
+            displayName: Log Service Output
+            condition: always()
+
+          # Collect must-gather logs
+          - bash: |
+              . secrets/env
+              go build ./hack/db
+              . ./hack/get-admin-kubeconfig.sh /subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$CLUSTER/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER > admin.kubeconfig
+              export KUBECONFIG=admin.kubeconfig
```
shubhadapaithankar: Same changes needed for the MIWI must-gather step:
 **Current:**
`. ./hack/get-admin-kubeconfig.sh /subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$CLUSTER/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER > admin.kubeconfig`
export KUBECONFIG=admin.kubeconfig

**Suggested**:
`./hack/get-admin-kubeconfig.sh /subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$CLUSTER/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER > admin.kubeconfig
export KUBECONFIG=$(pwd)/admin.kubeconfig`

Same as above -we need to execute (not source) the script for proper stdout redirection, and use an absolute path for reliable kubeconfig resolution.

.pipelines/ci.yml
```diff
@@ -248,7 +253,147 @@ stages:
...
+              docker compose logs vpn
+              docker compose logs selenium
+              docker compose logs rp
+              docker compose logs portal
+              docker compose logs e2e
+            displayName: Log Service Output
+            condition: always()
+
+          # Collect must-gather logs
+          - bash: |
```
shubhadapaithankar, line 269: Can we add a comment similar to the one you wrote for MIWI (lines 410-411) to explain why we're building the db binary?
```
'# Collect must-gather logs
- bash: |
    # Build the db binary needed by get-admin-kubeconfig.sh to query CosmosDB for cluster credentials
    # This step runs conditionally on failure, so the binary may not have been built earlier
    . secrets/env
    go build ./hack/db
    ./hack/get-admin-kubeconfig.sh /subscriptions/$AZURE_SUBSCRIPTION_ID/resourceGroups/$CLUSTER/providers/Microsoft.RedHatOpenShift/openShiftClusters/$CLUSTER > admin.kubeconfig
    export KUBECONFIG=$(pwd)/admin.kubeconfig
    wget -nv https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/$(OpenShiftCLIVersion)/openshift-client-linux-$(OpenShiftCLIVersion).tar.gz
    tar xf openshift-client-linux-$(OpenShiftCLIVersion).tar.gz
    ./oc adm must-gather --image arointsvc.azurecr.io/openshift4/ose-must-gather:latest
    tar cf must-gather.tar.gz must-gather.local.*
  displayName: Collect must-gather
  condition: Failed()'
```
  
  
  This step only runs when tests fail (condition: Failed()), and it's not immediately obvious why we need to build the db binary here. The comment helps future contributors understand:
What the db binary does (queries CosmosDB for credentials)
Why we build it on-demand (the step runs conditionally, so it may not exist)
The dependency relationship between get-admin-kubeconfig.sh and the db tool

shubhadapaithankar (approved):
LGTM

cadenmarchese (approved):

kimorris27 (approved):

PR #4414, opened 2025-10-10
Tuxerrante/gh issue 647

### LINKED ISSUE #647: Need ability to tell if cluster has pullSecret set
For debugging purposes, we need easy ability to find out if cluster in question has pullsecret set or not.

This could be either metric or redacted admin mode:
```
if len(oc.Properties.ClusterProfile.PullSecret) > 0 {
		out.Properties.ClusterProfile.PullSecret = "##REDACTED##"
	}
```

tuxerrante:
Fixes #647 

This PR provides a new `redact` package from where is possible to retrieve a full copy of an [OpenShiftCluster struct](https://github.com/Azure/ARO-RP/blob/master/pkg/api/openshiftcluster.go), without any sensitive data, including Properties.ClusterProfile.PullSecret.

### Test plan for issue:
- A unit test is provided in the package which verifies that known secrets do change their value after the copy is made.
- I'm still not aware of how to run integration tests

No, it is a debugging helper, as far as I've understood.

Is not meant to be run in a production cluster.

### DIFF
--- a/pkg/util/redact/pullsecret.go
+++ b/pkg/util/redact/pullsecret.go
@@ -9,8 +9,6 @@ import (
 const (
 	// RedactedString is the token used to replace sensitive data.
 	RedactedString = "##REDACTED##"
-	// MaxSecretSizeBytes aims to defend from excessive memory allocation.
-	MaxSecretSizeBytes = 10 * 1024 * 1024 // 10MB
 )
 
 // RedactPullSecret creates a deep c of the OpenshiftCluster with the pull secret redacted.


hawkowl:
The related issue is open since May, 5 years ago :) I'm actually not sure why we have GitHub issues open, since we use the internal JIRA and the GH issues haven't been used for a long, long time. 

We're (speaking non-authoritatively, as an IC on the Red Hat side) not really set up for external contributions, unfortunately, and have some difficult limitations in that regard (e.g. accessing most CI is in Azure systems which aren't public). It's less of a community-oriented open source project like k8s and more us "working in the open".

I'll ask one of our team leadership to look into making sure the GitHub issues are closed to make that clearer to folks. Without the internal context a lot of the ones that are even still valid (which most aren't...) are probably impossible to implement in a way that we could take. Sorry about that.

PR #4420, opened 2025-10-15
Replace SHA-1 with SHA-256 for certificate thumbprint calculation and…

tuxerrante:
Fixes #4419 

[SHA1](https://pkg.go.dev/crypto/sha1@go1.25.2) should not be used anymore.

It’s not a compile-time/API break, but it is a behavioural (runtime) breaking change for any client that depends on the previous SHA‑1 thumbprint value.

**Breaking** for callers that:
- Compare Thumbprint() to previously persisted strings (DB, files).
- Use Thumbprint() to identify certificates across systems (fingerprint matching).
- Send the thumbprint to other services/clients that expect SHA‑1.

### Test plan for issue:
Unit tests were created, since missing.

No published docs referenced the kind of digest algorithm used.

This package seems used [only internally by monitoring functions](https://github.com/search?q=repo%3AAzure%2FARO-RP+cert.Thumbprint&type=code).
More verifications are needed for possible calls coming from outside.

### DIFF
--- a/pkg/util/cert/cert.go
+++ b/pkg/util/cert/cert.go
@@ -5,6 +5,7 @@ package cert
 
 import (
 	"crypto/sha1"
+	"crypto/sha256"
 	"crypto/x509"
 	"fmt"
 	"time"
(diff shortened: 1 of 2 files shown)


hawkowl:
We don't match on the thumbprint in any way that is security-dependent, since we only use it for matching across other systems as a shorthand to identify expiring certificates. So, this would break that case.

PR #4422, opened 2025-10-16
Adding retry for MISE authentication

sankur-codes:

Fixes [ARO-18123](https://issues.redhat.com/browse/ARO-18123)

This PR adds a linear backoff retry mechanism for MISE authentication to minimize false negatives. Some authentication attempts that initially failed through MISE (due to context timeouts or connection resets) later succeeded via the fallback TLS method. Introducing retries helps mitigate such transient errors and improves overall reliability.

### Test plan for issue:

Test cases added. 

No

Not changing the functional logic for auth just adding a loop for retry.

### DIFF
--- a/pkg/frontend/middleware/authenticated_test.go
+++ b/pkg/frontend/middleware/authenticated_test.go
@@ -89,6 +89,46 @@ func TestAuthenticate(t *testing.T) {
 			isArmAuth:    false,
 			expectStatus: http.StatusForbidden,
 		},
+		{
+			name:         "ARM authentication MISE enforced - success",
+			enableMISE:   true,
+			enforceMISE:  true,
+			isMiseAuth:   true,
+			isArmAuth:    false,
+			expectStatus: http.StatusOK,
+		},
+		{
+			name:         "ARM authentication MISE enforced - failure, no TLS fallback",
+			enableMISE:   true,
+			enforceMISE:  true,
+			isMiseAuth:   false,
+			isArmAuth:    true, // TLS would succeed but should be ignored
+			expectStatus: http.StatusForbidden,
+		},
+		{
+			name:         "ARM authentication MISE enforced - both would fail",
+			enableMISE:   true,
+			enforceMISE:  true,
+			isMiseAuth:   false,
+			isArmAuth:    false,
+			expectStatus: http.StatusForbidden,
+		},
+		{
+			name:         "ARM authentication MISE disabled, TLS success",
+			enableMISE:   false,
+			enforceMISE:  false,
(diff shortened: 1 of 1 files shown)


pkg/util/miseadapter/authorizer.go
```diff
@@ -40,17 +40,50 @@ func (m *miseAdapter) IsAuthorized(ctx context.Context, r *http.Request) (bool,
...
+
+		if result.StatusCode == http.StatusOK {
+			if attempt > 0 {
+				m.log.Infof("mise authentication succeeded on attempt %d/%d", attempt+1, maxRetries)
+			}
+			return true, nil
+		}
+
+		// Don't retry on non-transient errors (4xx client errors except for specific cases)
+		if result.StatusCode >= 400 && result.StatusCode < 500 && result.StatusCode != http.StatusRequestTimeout && result.StatusCode != http.StatusTooManyRequests {
```
cadenmarchese, line 75: We already have some pretty robust logic for retrying 4xx errors in pkg/util/steps/refreshing.go. Should we try to re-use that or at least move the retrying logic over there?

sankur-codes: From what I see in refreshing.go, the retries account for role propagation delays, extending up to about 10 minutes. In contrast, MISE auth retries are limited to ~300 ms and primarily resend the same request, while refreshing.go handles both auth refresh and retry. We could consider a generic implementation in pkg/util/retry/retry.go for both, but I’d argue that the current retry logic is distinct for each use case and remains clearer when kept separate ?

pkg/util/miseadapter/authorizer.go
```diff
@@ -40,17 +40,50 @@ func (m *miseAdapter) IsAuthorized(ctx context.Context, r *http.Request) (bool,
...
-	if err != nil {
-		return false, err
-	}
+	const maxRetries = 3
+	const retryDelayMs = 100
+
+	var result Result
+	var err error
+
+	for attempt := 0; attempt < maxRetries; attempt++ {
```
pepedocs, line 49: If `attempt` starts at 1 then some of the lines/code here can be omitted.

sankur-codes: Yes it does. We could get the `attempt+1` to `attempt` but still have to have `attempt-1` for sleep calculation instead. I just kept it to be conventional with a start with `0` for it to be cleaner.

shubhadapaithankar (approved):
LGTM

shubhadapaithankar (approved):

wanghaoran1988 (approved):

PR #4424, opened 2025-10-16
fix: parse authentication errors during installation for retries

komidore64:
Fixes: ARO-17752

Co-authored-by: cadenmarchese <caden.marchese@gmail.com>

Fixes: ARO-17752

This change accounts for the parsing of "track2" SDK errors that are returned
from Azure during installation when some resources take a little bit of time to
populate. Logic in `pkg/util/steps/refreshing.go`
([link](https://github.com/Azure/ARO-RP/blob/master/pkg/util/steps/refreshing.go#L80-L84))
checks for some specific errors to know that it can retry before giving up.

### Test plan for issue:

This is considerably difficult to test, but I have added some new entries to
unit tests (`make unit-test-go`).

Not that I'm aware.

No logic has been replaced. Only additional logic has been added.

### DIFF
--- a/pkg/util/azureerrors/error.go
+++ b/pkg/util/azureerrors/error.go
@@ -17,22 +17,29 @@ import (
 )
 
 const (
-	CodeInvalidTemplateDeployment = "InvalidTemplateDeployment"
+	CODE_AUTHFAILED       = "AuthorizationFailed"
+	CODE_DEPLOYACTIVE     = "DeploymentActive"
+	CODE_DEPLOYFAILED     = "DeploymentFailed"
+	CODE_FORBIDDEN        = "Forbidden"
+	CODE_INVALIDTEMPL     = "InvalidTemplateDeployment"
+	CODE_LINKEDAUTHFAILED = "LinkedAuthorizationFailed"
+	CODE_RGNOTFOUND       = "ResourceGroupNotFound"
 )
 
 // HasAuthorizationFailedError returns true it the error is, or contains, an
 // AuthorizationFailed error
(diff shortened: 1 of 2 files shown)


pkg/util/azureerrors/error.go
```diff
@@ -47,10 +54,10 @@ func deploymentFailedDueToAuthError(err error, authCode string) bool {
 	}
 
 	if serviceErr, ok := err.(*azure.ServiceError); ok &&
-		serviceErr.Code == "DeploymentFailed" {
+		serviceErr.Code == CODE_DEPLOYFAILED {
 		for _, d := range serviceErr.Details {
 			if code, ok := d["code"].(string); ok &&
-				code == "Forbidden" {
+				code == CODE_FORBIDDEN {
```
rajdeepc2792, line 60: I have personally not seen any issues due to case mismatch, but is there a need to handle the equality as case insensitive?

rajdeepc2792 (approved):
A comment on case matching, otherwise LGTM.

pkg/util/azureerrors/error.go
komidore64: I have not seen that either, but I guess it couldn't hurt. @cadenmarchese what is your opinion? make as few changes as possible, or add a few little improvements (such as case-insensitivity) here and there?

kimorris27: I'd be content merging this without the case insensitive checks. This code is battle-tested, and the only change is that we're introducing constants for the same previously hard-coded values.

kimorris27 (approved):

PR #4425, opened 2025-10-17
adding local cosmosdb to run monitor changefeed tests

alcasim:

https://issues.redhat.com/browse/ARO-21855

Fixes

It uses the cosmosdb emulator to run the changefeed tests on Monitor. This only works in local development at the moment, due to disk space constraints on the CI runners (a separate ticket would address this).

In order to make the local emulator work, this PR:

- Adds new options in the Mafike to start the local cosmosdb, run unit tests with local cosmosdb and remove the cosmosdb emulator container.
- Adds a new helper to setup the local database schema (limited to Monitor functionality, as a first approach to run tests for Monitor)
- Sets the unit tests for Monitor depending on if it runs locally with the `USE_LOCAL_COSMOS_FOR_TEST` set up, loading the emulator client and initializing the schema. If this variable is not set (like in the CI runs), the tests will still use the fake client.
- Adds new tests that leverage on the emulator. These tests will only run if a client for the emulator is created, so the CI won't run them.

### Test plan for issue:

this Pr implement tests

Probably we need to update the monitor.md file to reflect this.

it won't change production code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -427,7 +427,18 @@ endif
 .PHONY: start-local-cosmosdb
 start-local-cosmosdb:
 	docker run --detach --publish 8081:8081 --publish 10250-10255:10250-10255 --name local-cosmosdb mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:latest
-	while ! curl -s -k -o /dev/null https://localhost:8081/_explorer/index.html; do sleep 1; done
+	@echo "Waiting for CosmosDB emulator to be ready..."
+	@timeout=180; \
+	elapsed=0; \
+	while ! curl -s -k -o /dev/null https://localhost:8081/_explorer/index.html; do \
+		if [ $$elapsed -ge $$timeout ]; then \
+			echo "ERROR: CosmosDB emulator failed to start after $$timeout seconds"; \
+			exit 1; \
+		fi; \
+		sleep 1; \
+		elapsed=$$((elapsed + 1)); \
+	done; \
+	echo "CosmosDB emulator is ready"
 
 .PHONY: stop-and-delete-local-cosmosdb
 stop-and-delete-local-cosmosdb:
--- a/Makefile
+++ b/Makefile
@@ -312,9 +312,17 @@ validate-fips: $(BINGO)
 	hack/fips/validate-fips.sh ./aro
 
 .PHONY: unit-test-go
-unit-test-go: $(GOTESTSUM) start-local-cosmosdb
-	$(GOTESTSUM) --format pkgname --junitfile report.xml -- -coverprofile=cover.out ./... || ($(MAKE) stop-and-delete-local-cosmosdb && exit 1)
-	$(MAKE) stop-and-delete-local-cosmosdb
+unit-test-go: $(GOTESTSUM) 
+	$(GOTESTSUM) --format pkgname --junitfile report.xml -- -coverprofile=cover.out ./...
+
+# .PHONY: unit-test-go-local-cosmosdb
+# unit-test-go-local-cosmosdb: $(GOTESTSUM) start-local-cosmosdb
+# 	LOCAL_COSMOS_FOR_TEST_HOST=127.0.0.1 $(GOTESTSUM) --format pkgname --junitfile report.xml -- -coverprofile=cover.out ./... || ($(MAKE) stop-and-delete-local-cosmosdb && exit 1)
+# 	$(MAKE) stop-and-delete-local-cosmosdb
+.PHONY: unit-test-go-local-cosmosdb
+unit-test-go-local-cosmosdb:
+	USE_LOCAL_COSMOS_FOR_TEST="true" $(GOTESTSUM) --format pkgname --junitfile report.xml -- -coverprofile=cover.out ./...
+
 
 .PHONY: unit-test-go-coverpkg
 unit-test-go-coverpkg: $(GOTESTSUM)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -411,6 +412,16 @@ ifeq ($(shell uname -s),Darwin)
 	codesign -f -s - ${GOPATH}/bin/mockgen
 endif
 
+.PHONY: start-local-cosmosdb
+start-local-cosmosdb:
+	docker run --detach --publish 8081:8081 --publish 10250-10255:10250-10255 --name local-cosmosdb mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:latest
```
aasserzo, line 433: Maybe we should pin a version instead of latest, and bump as needed? In case a new emulator breaks our tests.

Makefile
```diff
@@ -411,6 +412,16 @@ ifeq ($(shell uname -s),Darwin)
...
+	while ! curl -s -k -o /dev/null https://localhost:8081/_explorer/index.html; do sleep 1; done
```
aasserzo: Correct me if I'm wrong, but this might run forever. Add a counter?

Makefile
```diff
@@ -411,6 +412,16 @@ ifeq ($(shell uname -s),Darwin)
...
+
+.PHONY: stop-and-delete-local-cosmosdb
+stop-and-delete-local-cosmosdb:
+	docker stop local-cosmosdb
+	docker rm local-cosmosdb
```
aasserzo, line 450: Are there cases where we would like to not delete the local DB?

pkg/monitor/local_cosmos.go
```diff
@@ -0,0 +1,237 @@
...
+func localCosmosNewClient(_env env.Core, m metrics.Emitter, aead encryption.AEAD) (cosmosdb.DatabaseClient, error) {
+	logrusEntry := _env.LoggerForComponent("database")
+
+	masterKey := "C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==" // To-Do: move this outside the code
+	dbAuthorizer, err := cosmosdb.NewMasterKeyAuthorizer(masterKey)
+	if err != nil {
+		return nil, err
+	}
+
+	h, err := database.NewJSONHandle(aead)
```
aasserzo: I prefer meaningful names.

pkg/monitor/local_cosmos.go
```diff
@@ -0,0 +1,237 @@
...
+			MaxIdleConnsPerHost: 20,
+			// Skip TLS verification for local emulator with self-signed cert
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		},
+		Timeout: 30 * time.Second,
+	}
+
+	databaseHostname := "127.0.0.1:8081"
+
+	return cosmosdb.NewDatabaseClient(logrusEntry, c, h, databaseHostname, dbAuthorizer), nil
```
aasserzo: This might return an error.

Makefile
alcasim: they don't seem to tag the emulator :grimacing: so here be dragons for now

Makefile
alcasim: Usually we want to start the tests from a clean sleet, and we don't pre-populate a lot of stuff, so I think that keeping the local container may induce to errors or confusion.

pkg/monitor/local_cosmos.go
alcasim: this is a copy-pasta from the original client creation, but agreed, naming can be improved.

pkg/monitor/local_cosmos.go
alcasim: oh yes, there are several places that returns an error and it's not being handled so far, they'll need to be sorted out before the final version :+1:

Makefile
aasserzo: We might be able to pull by digest, if those are kept even though they aren't tagged.

Makefile
aasserzo: This was never there in the first place, so it's just a potential improvement, but we can save the existing image for debug. But yea, since this was never implemented anyway, it's just a minor thing

pkg/monitor/local_cosmos.go
alcasim: quickly checked this... that func only returns a pointer to the client. Client's struct is populated directly with data provided to the function, and nothing else happens [there](https://github.com/Azure/ARO-RP/blob/master/pkg/database/cosmosdb/zz_generated_database.go#L63-L73) :)

docs/monitoring.md
```diff
@@ -50,3 +50,46 @@ monitoring from inside the cluster as well as a complementary near-term goal.
...
+## Running monitor locally and capturing metrics
+
+Monitor can be run in the local environment with the `make runlocal-monitor`
+command, after the your local environment is ready as stated in
+[Prepare Your Development Environment](https://github.com/Azure/ARO-RP/blob/master/docs/prepare-your-dev-environment.md).
+It will start monitoring itself and any cluster that has been created
+by a local RP, so if you want to monitor a cluster, it needs to be created by
+a local RP beforehand.
+
+If you are interested to capture the metrics that monitor sends out,
```
ehvs: _To make easier the flow of reading would be good to divide by concise session_

### Pre-requisite
Local cluster with RP, add docs link

###  Capturing the metrics from the monitor
1. < add socat command >
2. start the monitor,  `make runlocal-monitor`
[...explanation..]

ehvs (changes requested):

PR #4436, opened 2025-10-24
[ARO-12449] Add mimo intro and TSG doc

ArrisLee:

https://issues.redhat.com/browse/ARO-7736

Adding a new doc for MIMO regarding infra, components, metrics and troubleshooting

### Test plan for issue:

N/A

Yes, it is a doc

### DIFF
--- a/docs/mimo/intro-troubleshooting.md
+++ b/docs/mimo/intro-troubleshooting.md
@@ -937,7 +937,7 @@ Once the new manifest completes successfully, you can delete the old failed one:
 
 For additional MIMO information, see the following documentation in this folder:
 
-### For SREs and Operators:
+### For SREs
 - **[MIMO Overview (README.md)](./README.md)** - High-level overview of MIMO architecture, components, and concepts. Read this first to understand the fundamentals of how MIMO works.
 - **[Admin API Reference (admin-api.md)](./admin-api.md)** - Technical reference for Admin API endpoints. Note that SREs should use PowerShell scripts instead of calling these APIs directly.
 - **[Actuator Documentation (actuator.md)](./actuator.md)** - Detailed explanation of the Actuator component, including a flowchart showing how tasks are processed from the queue.


docs/mimo/intro-troubleshooting.md
```diff
@@ -0,0 +1,951 @@
...
+
+**Note:** `{resourceId}` is the full ARM resource ID for an ARO cluster:
+```
+/subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/Microsoft.RedHatOpenShift/openShiftClusters/{clusterName}
+```
+
+⚠️ **Important Limitations:**
+- **Global queue queries** (e.g., viewing all queued manifests across all clusters) are NOT available via Admin API
+- **Single manifest detail queries** (e.g., getting specific manifest by ID) are NOT available via Admin API
+- For these operations, use internal tooling: **Geneva Actions**, **Admin Portal**, **Kusto/DGrep**, or **monitoring dashboards**
```
kimorris27, line 342: For doing these two types of queries that aren't supported by the admin API, I don't see how Geneva Actions, the Admin Portal, or Kusto/DGrep would help. There are a few points in the doc where the TSG directs you to do these types of queries, so I think we should be more specific about how to do them (even if we need to them in a suboptimal way right now since the ideal tooling isn't in place yet).

mociarain (changes requested):
This is fantastic work and I think the detail is excellent. I do think though it's not laid out in a way that makes it easy for people to find what they want quickly.

Some ideas:
- The Admin API stuff can me move to the already existing AdminAPI file and referenced.
- The entire first section belongs in the README. This is really great overview and helped me a lot. I think it should go front and center and then referenced as needed.
- All the content on alerting rules belongs in it's own file IMO

For the rest I don't know if this belongs in the RP. AFAIK we keep our TSG and this type of on-call information in the Wiki (that's migrating atm but still). Would this stuff be better in there so someone on-call will find it?

PR #4441, opened 2025-10-29
disable automatic vmss repairs before cleanup and add retry

bizz001:

Fixes https://issues.redhat.com/browse/ARO-22263

Disables VMSS repair before removing old VMSS in order to prevent race conditions with the Azure automatic repair.

### Test plan for issue:

Tested for regression by deploying to Canary.

No.

Deploy logs show the steps.

### DIFF
--- a/pkg/deploy/deploy_test.go
+++ b/pkg/deploy/deploy_test.go
@@ -413,7 +413,7 @@ func TestDisableAutomaticRepairsOnVMSS(t *testing.T) {
 					if update.VirtualMachineScaleSetUpdateProperties == nil {
 						t.Fatalf("expected VirtualMachineScaleSetUpdateProperties to be set")
 					}
-					policy := update.VirtualMachineScaleSetUpdateProperties.AutomaticRepairsPolicy
+					policy := update.AutomaticRepairsPolicy
 					if policy == nil || policy.Enabled == nil {
 						t.Fatalf("expected AutomaticRepairsPolicy.Enabled to be set")
 					}


pkg/deploy/error.go
```diff
@@ -18,3 +20,19 @@ func isDeploymentNotFoundError(err error) bool {
 	}
 	return false
 }
+
+func isOperationPreemptedError(err error) bool {
```
ehvs, line 26: Are `requestErr`  the same as `serviceErr`? Do we want to make some distinction here? Would the difference between them warrant any different action? If not, looks good as is.

pkg/deploy/deploy.go
```diff
@@ -266,3 +269,61 @@ func (d *deployer) checkForKnownError(serviceErr *azure.ServiceError, deployAtte
 
 	return "", nil
 }
+
+// disableAutomaticRepairsOnVMSS disables automatic repairs on a VMSS to prevent
+// race conditions during teardown when stopping services causes health probe failures.
+// This is a best-effort operation - we log errors as warnings but don't fail the
+// teardown, allowing cleanup to proceed even if we can't disable repairs.
+func (d *deployer) disableAutomaticRepairsOnVMSS(ctx context.Context, resourceGroupName, vmssName string) error {
```
alcasim, line 277: not directly related to the current pipeline failures, however.... I wonder if it's possible to add some tests for this func and `runCommandWithRetry` in `deploy_test.go` ...

pkg/deploy/error.go
```diff
@@ -18,3 +20,19 @@ func isDeploymentNotFoundError(err error) bool {
...
+	if detailedErr, ok := err.(autorest.DetailedError); ok {
+		if requestErr, ok := detailedErr.Original.(*azure.RequestError); ok &&
+			requestErr.ServiceError != nil &&
+			requestErr.ServiceError.Code == "OperationPreempted" {
```
mociarain: Can we move 'OperationPreempted' into an enum and give a link to where it came from if possible

pkg/deploy/error.go
bizz001: I don't think we need to make a distinction. We just need the error code.

pkg/deploy/deploy.go
bizz001: tests added.

pkg/deploy/error.go
bizz001: done. As for the link, I could not find specific docs mentioning it, but that is the code we saw in the logs when the issue happened.

PR #4442, opened 2025-10-29
Removal of SRE-built MHC alert

ehvs:
https://issues.redhat.com/browse/ARO-22359

Upon disabling the monitor Geneva, this also needs removal of the alert rule from in-cluster.

### Test plan for issue:
Added Unit Tests

- [updated] Managed Feature Flag is false: ensure MHC is deleted
- [new] Managed Feature Flag is true: ensure Prometheus Alert is deleted
- [updated] Managed Feature Flag is false: MHC fails to delete, an error is returned

### DIFF
--- a/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
+++ b/pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
@@ -89,6 +89,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 
 	var resources []kruntime.Object
 
+	if instance.Spec.OperatorFlags.GetSimpleBoolean(operator.MachineHealthCheckManaged) {
+		err = r.dh.EnsureDeleted(ctx, "PrometheusRule", "openshift-machine-api", "mhc-remediation-alert")
+		if err != nil {
+			r.Log.Error(err)
+			r.SetDegraded(ctx, err)
+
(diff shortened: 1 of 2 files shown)


pkg/operator/controllers/machinehealthcheck/machinehealthcheck_controller.go
```diff
@@ -86,21 +83,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 		}
 
-		err = r.dh.EnsureDeleted(ctx, "PrometheusRule", "openshift-machine-api", "mhc-remediation-alert")
-		if err != nil {
-			r.Log.Error(err)
-			r.SetDegraded(ctx, err)
-
-			return reconcile.Result{RequeueAfter: time.Hour}, err
-		}
-
```
kimorris27, line 96: If we're going to remove this, do we have another plan for ensuring the PrometheusRule is deleted from clusters created before this operator version?

ehvs: Good point Kipp. I've just pushed an update with a test and a condition in the controller for that.
One thing to mention here, is that my implementation focus is to cover when we have MHC on Managed state now when at Unmanaged. But Im open to ideas, if we want to cover both.

kimorris27: The way you've done it makes sense to me 👍🏻

kimorris27 (approved):

PR #4444, opened 2025-10-30
modified operator.go to add machineconfigselector for 4.18 validatingadmissionPolicy

vinaybommana:
### Which issue this PR addresses:
#itn-2025-00278 relevant JIRA: ARO-22392
resolving the e2e test fail due validatingAdmissionPolicy which is enforced from 4.18 

validated the created manifest is working by testing it manually on 4.18 and 4.16 clusters

PR #4445, opened 2025-10-30
[ARO-22392] modified operator.go to add machineconfigselector for 4.18 validatingadmissionPolicy

hawkowl:
Replaces #4444 

#Which issue this PR addresses:

`#itn-2025-00278` relevant JIRA: [ARO-22392](https://issues.redhat.com/browse/ARO-22392)
resolving the e2e test fail due validatingAdmissionPolicy which is enforced from 4.18

validated the created manifest is working by testing it manually on 4.18 and 4.16 clusters

### DIFF
--- a/test/e2e/operator.go
+++ b/test/e2e/operator.go
@@ -599,7 +599,7 @@ var _ = Describe("ARO Operator - dnsmasq", func() {
 			Name: mcpName,
 		},
 		Spec: mcv1.MachineConfigPoolSpec{
-			// OCP 4.18.26+ ValidatingAdmissionPolicy requires custom MCPs to inherit from worker pool
+			// OCP 4.18+ ValidatingAdmissionPolicy requires custom MCPs to inherit from worker pool
 			// Using matchExpressions to include both "worker" and custom role
 			MachineConfigSelector: &metav1.LabelSelector{
 				MatchExpressions: []metav1.LabelSelectorRequirement{


test/e2e/operator.go
```diff
@@ -598,7 +598,24 @@ var _ = Describe("ARO Operator - dnsmasq", func() {
...
 			Name: mcpName,
 		},
-		Spec: mcv1.MachineConfigPoolSpec{},
+		Spec: mcv1.MachineConfigPoolSpec{
+			// OCP 4.18+ ValidatingAdmissionPolicy requires custom MCPs to inherit from worker pool
+			// Using matchExpressions to include both "worker" and custom role
+			MachineConfigSelector: &metav1.LabelSelector{
+				MatchExpressions: []metav1.LabelSelectorRequirement{
+					{
+						Key:      "machineconfiguration.openshift.io/role",
```
ehvs, line 607: We could have the key as a constant, instead of hard-coding this string.
Similar to what we have implemented [here](https://github.com/openshift/node-observability-operator/blob/f39afbdfb474c87297c6d559f117a7735cc77bb3/pkg/operator/controller/machineconfig/mcp.go#L88).

Same for line 615.

vinaybommana: will be merging this for, don't want another 2hrs for E2E to pass

mociarain: I agree with Hevs. Let's do this as a follow-up and I'll merge this now to get the rollout going

PR #4450, opened 2025-10-31
e2e - check for nil `oc` before accessing ServicePrincipalProfile

cadenmarchese:

Fixes ARO-22414

If getting `oc` returns a nil `oc` and an error, we just append the error and continue through the function rather than exiting. As a result, the subsequent line to check oc.Properties.ServicePrincipalProfile causes a nil pointer panic in every case where the cluster could not be found. Instead of panic, we should handle this case and only execute the "delete service principal" functionality if there's actually a service principal to delete.

### Test plan for issue:

- Unit and e2e

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -738,10 +738,12 @@ func (c *Cluster) Delete(ctx context.Context, vnetResourceGroup, clusterName str
 			c.log.Errorf("CI E2E cluster %s not found in resource group %s", clusterName, vnetResourceGroup)
 			errs = append(errs, err)
 		}
-		if oc.Properties.ServicePrincipalProfile != nil {
-			errs = append(errs,
-				c.deleteApplication(ctx, oc.Properties.ServicePrincipalProfile.ClientID),
-			)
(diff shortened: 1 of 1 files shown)


pkg/util/cluster/cluster.go
```diff
@@ -738,10 +738,12 @@ func (c *Cluster) Delete(ctx context.Context, vnetResourceGroup, clusterName str
 			c.log.Errorf("CI E2E cluster %s not found in resource group %s", clusterName, vnetResourceGroup)
 			errs = append(errs, err)
 		}
-		if oc.Properties.ServicePrincipalProfile != nil {
-			errs = append(errs,
-				c.deleteApplication(ctx, oc.Properties.ServicePrincipalProfile.ClientID),
-			)
+		if oc != nil {
```
komidore64, line 741: question: Would it make sense to short-circuit and return an error if `oc` is indeed `nil`? 

What should happen if `oc == nil`?

komidore64: I should have read your description closer. You clearly answered my question. Sorry for wasting time!

PR #4454, opened 2025-11-04
ARO-21854 - Allow developers to standup a local CosmosDB instance

aasserzo:

https://issues.redhat.com/browse/ARO-21854

Fixes

This PR will allow devs to standup a local cosmosDB emulator for development and testing purposes.

### Test plan for issue:

TBD

Docs within the repo were updated to provide usage instructions.

Will not run in production.

### DIFF
--- a/hack/cosmosdb-emulator/init-db-structure.sh
+++ b/hack/cosmosdb-emulator/init-db-structure.sh
@@ -12,8 +12,8 @@ ARM_TEMPLATE="./pkg/deploy/assets/databases-development.json"
 echo "Initializing Cosmos DB database structure..."
 
 # Check if the emulator is running
-if ! curl -k -s https://${EMULATOR_ENDPOINT}/_explorer/emulator.pem > /dev/null 2>&1; then
-    echo "ERROR: Cosmos DB emulator is not running at https://${EMULATOR_ENDPOINT}"
+if ! curl -s http://localhost:1234 > /dev/null 2>&1; then
+    echo "ERROR: Cosmos DB emulator is not running"
     echo "Please start the emulator first with: make start-cosmos-emulator"
     exit 1
 fi
--- a/hack/cosmosdb-emulator/start-cosmos-emulator.sh
+++ b/hack/cosmosdb-emulator/start-cosmos-emulator.sh
@@ -5,13 +5,12 @@ set -e
 
 CONTAINER_NAME="${COSMOSDB_EMULATOR_CONTAINER_NAME:-aro-cosmosdb-emulator}"
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-$(command -v podman || command -v docker)}"
-COSMOSDB_EMULATOR_IMAGE="${COSMOSDB_EMULATOR_IMAGE:-mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:latest}"
+COSMOSDB_EMULATOR_IMAGE="${COSMOSDB_EMULATOR_IMAGE:-mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:vnext-preview}"
 EMULATOR_ENDPOINT="${COSMOSDB_EMULATOR_ENDPOINT:-localhost:8081}"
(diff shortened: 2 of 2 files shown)


env.example
```diff
@@ -27,5 +27,14 @@ export TAG=latest
 export LOCAL_VPN_IMAGE=vpn
 export E2E_LABEL='!smoke&&!regressiontest'
 
+# Local Cosmos DB Emulator Configuration
+# export USE_COSMOS_DB_EMULATOR=true
```
alcasim, line 31: nitpick: we should default the example to false, so new joiners or people rebuilding their local environment don't have an unexpected change of behavior.

hack/cosmosdb-emulator/setup.go
```diff
@@ -0,0 +1,246 @@
...
+			if err := createTrigger(ctx, dbClient, dbName, resource); err != nil {
+				log.Errorf("Error creating trigger: %v", err)
+				hasErrors = true
+			}
+			time.Sleep(500 * time.Millisecond)
+		}
+	}
+
+	if hasErrors {
+		log.Fatal("Database setup completed with errors")
```
alcasim, line 105: nitpick: shouldn't we exit here with an exit code 1?

hack/cosmosdb-emulator/setup.go
```diff
@@ -0,0 +1,246 @@
...
+
+	templatePath := "pkg/deploy/assets/databases-development.json"
+	template, err := readARMTemplate(templatePath)
+	if err != nil {
+		log.Fatalf("Failed to read ARM template: %v", err)
+	}
+
+	dbClient, err := database.NewLocalDatabaseClient(log, &noop.Noop{}, nil)
+	if err != nil {
+		log.Fatalf("Failed to create database client: %v", err)
```
alcasim, line 48: if we fail her, or in the previous if, we should not continue executing.

aasserzo: We don't. log.Fatal() not only logs but also exits the program.

hack/cosmosdb-emulator/setup.go
aasserzo: log.Fatal() not only logs but also exits the program.

env.example
aasserzo: It's commented out, uncommenting enables the use of the emulator. There is no use case where we would like this set to false

hack/cosmosdb-emulator/setup.go
alcasim: never review PRs after an on-call 🤦

hack/cosmosdb-emulator/setup.go
alcasim: never review PRs after an on-call 🤦

docs/local-cosmosdb-emulator.md
```diff
@@ -0,0 +1,100 @@
...
+|----------|-------------|---------|
+| `COSMOSDB_EMULATOR_ENDPOINT` | Emulator endpoint | No default, must be set |
+| `COSMOSDB_EMULATOR_KEY` | Authentication key | No default, must be set |
+
+### Optional Environment Variables
+
+| Variable | Description | Default |
+|----------|-------------|---------|
+| `COSMOSDB_EMULATOR_CONTAINER_NAME` | Container name | `aro-cosmosdb-emulator` |
+| `COSMOSDB_EMULATOR_DATABASE_NAME` | Database name | `ARO` |
```
alcasim: we may want to get rid of this. When setting up the local environment, one of the generated env vars in the `secrets/env` file is `DATABASE_NAME`. By defaults points to `$USER`

So... we may just want to see if `DATABASE_NAME` exists, and stops if it doesn't... as it should be part of the local environment setup

aasserzo: Yeah, I'd figured this out from testing. Pushing new commits soon that will remove this as well.

aasserzo:
[The task](https://issues.redhat.com/browse/ARO-21854) is now obsolete due to technical limitations in the cosmosdb emulator that prevent us from using it.

PR #4458, opened 2025-11-07
Add CNV Prometheus metrics monitor

machadovilaca:

Fixes https://issues.redhat.com/browse/ARO-22503

This PR adds a cluster monitor exporter for Prometheus metrics. As of this PR, it queries the cluster's Prometheus instance for `kubevirt_vmi_info` metrics and emits them as custom metrics with all the same labels.

### Test plan for issue:

Added comprehensive unit tests and manually verified that it works correctly on a live cluster.

No documentation update is needed for this PR

error handling, logging, and failure isolation

### DIFF
--- a/pkg/monitor/cluster/cluster.go
+++ b/pkg/monitor/cluster/cluster.go
@@ -187,6 +187,7 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
 		mon.emitIngressAndAPIServerCertificateExpiry,
 		mon.emitEtcdCertificateExpiry,
 		mon.emitPrometheusAlerts, // at the end for now because it's the slowest/least reliable
+		mon.emitPrometheusMetrics,
 		mon.emitCWPStatus,
 		mon.emitClusterAuthenticationType,
 	}
--- a/pkg/monitor/cluster/prometheusmetrics.go
+++ b/pkg/monitor/cluster/prometheusmetrics.go
@@ -1,0 +1,153 @@
+package cluster
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the Apache License 2.0.
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+
+	"github.com/Azure/ARO-RP/pkg/util/portforward"
+)
+
+type prometheusMetrics struct {
+	mon                *Monitor
+	client             *http.Client
+	prometheusQueryURL string
+}
+
+type prometheusQueryResponse struct {
+	Status string `json:"status"`
+	Data   struct {
+		ResultType string             `json:"resultType"`
+		Result     []prometheusResult `json:"result"`
+	} `json:"data"`
+}
+
+type prometheusResult struct {
+	Metric map[string]string `json:"metric"`
+	Value  []any             `json:"value"`
+}
+
+const prometheusQueryURL = "https://prometheus-k8s.openshift-monitoring.svc:9091/api/v1/query?query=%s"
+
+func (mon *Monitor) emitPrometheusMetrics(ctx context.Context) error {
+	mon.log.Debugf("running emitPrometheusMetrics")
+	pm := &prometheusMetrics{
+		mon:                mon,
+		prometheusQueryURL: prometheusQueryURL,
+	}
+
+	err := pm.connectToPrometheus(ctx)
+	if err != nil {
+		return err
+	}
+
+	return pm.emitCNVMetrics(ctx)
+}
+
+func (pm *prometheusMetrics) connectToPrometheus(ctx context.Context) error {
+	var err error
+
+	for i := range 2 {
+		pm.client = &http.Client{
+			Transport: &http.Transport{
+				DialContext: func(ctx context.Context, network, address string) (net.Conn, error) {
+					_, port, err := net.SplitHostPort(address)
+					if err != nil {
+						return nil, err
+					}
+
+					return pm.dialPrometheus(ctx, i, port)
+				},
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: true,
+				},
+				DisableKeepAlives: true,
+			},
+		}
+
+		_, err = pm.queryPrometheus(ctx, fmt.Sprintf("prometheus_build_info{pod='prometheus-k8s-%d'}", i))
+		if err == nil {
+			pm.mon.log.Debugf("successfully queried prometheus-k8s-%d", i)
+			return nil
+		}
+		pm.mon.log.Debugf("failed to query prometheus-k8s-%d: %+v", i, err)
+	}
+
+	return fmt.Errorf("failed to connect to any prometheus pod: %w", err)
+}
+
+func (pm *prometheusMetrics) dialPrometheus(ctx context.Context, podIndex int, port string) (net.Conn, error) {
+	podName := fmt.Sprintf("prometheus-k8s-%d", podIndex)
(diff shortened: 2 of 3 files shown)


pkg/monitor/cluster/cluster.go
```diff
@@ -187,6 +187,7 @@ func NewMonitor(log *logrus.Entry, restConfig *rest.Config, oc *api.OpenShiftClu
 		mon.emitIngressAndAPIServerCertificateExpiry,
 		mon.emitEtcdCertificateExpiry,
 		mon.emitPrometheusAlerts, // at the end for now because it's the slowest/least reliable
+		mon.emitPrometheusMetrics,
```
stevekuznetsov, line 190: based on the comment for the one above, do we want to place this elsewhere? does the ordering matter? will this not run if something earlier in the order fails?

pkg/monitor/cluster/prometheusmetrics.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/Azure/ARO-RP/pkg/util/portforward"
+)
+
+type prometheusMetrics struct {
+	mon                *Monitor
+	client             *http.Client
+	prometheusQueryURL string
+}
+
+type prometheusQueryResponse struct {
```
stevekuznetsov, line 26: I would prefer you import their types and use their SDK

pkg/monitor/cluster/prometheusmetrics.go
```diff
@@ -0,0 +1,153 @@
...
+		Result     []prometheusResult `json:"result"`
+	} `json:"data"`
+}
+
+type prometheusResult struct {
+	Metric map[string]string `json:"metric"`
+	Value  []any             `json:"value"`
+}
+
+const prometheusQueryURL = "https://prometheus-k8s.openshift-monitoring.svc:9091/api/v1/query?query=%s"
```
stevekuznetsov, line 39: this doesn't look like something that should be `const`, and certainly should not have format-strings or query params in it. please use the Go stdlib to create query params and format the `url.URL` as a string

pkg/monitor/cluster/prometheusmetrics.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (mon *Monitor) emitPrometheusMetrics(ctx context.Context) error {
+	mon.log.Debugf("running emitPrometheusMetrics")
```
stevekuznetsov, line 42: this is a low-value log, please omit it

pkg/monitor/cluster/prometheusmetrics.go
```diff
@@ -0,0 +1,153 @@
...
+		return err
+	}
+
+	return pm.emitCNVMetrics(ctx)
+}
+
+func (pm *prometheusMetrics) connectToPrometheus(ctx context.Context) error {
+	var err error
+
+	for i := range 2 {
```
stevekuznetsov, line 59: why 2? what does the retry behavior look like for a `Monitor` in general?

pkg/monitor/cluster/prometheusmetrics.go
```diff
@@ -0,0 +1,153 @@
...
+		pm.client = &http.Client{
+			Transport: &http.Transport{
+				DialContext: func(ctx context.Context, network, address string) (net.Conn, error) {
+					_, port, err := net.SplitHostPort(address)
+					if err != nil {
+						return nil, err
```
stevekuznetsov, line 65: general comment: please add some context to errors before returning

pkg/monitor/cluster/prometheusmetrics.go
```diff
@@ -0,0 +1,153 @@
...
+					}
+
+					return pm.dialPrometheus(ctx, i, port)
+				},
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: true,
```
stevekuznetsov, line 71: why? at minimum add some input config for this to turn off tls in testing, but certainly not for prod

pkg/monitor/cluster/prometheusmetrics.go
```diff
@@ -0,0 +1,153 @@
...
+				},
+				DisableKeepAlives: true,
+			},
+		}
+
+		_, err = pm.queryPrometheus(ctx, fmt.Sprintf("prometheus_build_info{pod='prometheus-k8s-%d'}", i))
```
stevekuznetsov, line 77: why are we throwing away the result?
why do we choose one specific prometheus pod?

pkg/monitor/cluster/prometheusmetrics.go
```diff
@@ -0,0 +1,153 @@
...
+func (pm *prometheusMetrics) queryPrometheus(ctx context.Context, query string) ([]prometheusResult, error) {
+	queryURL := fmt.Sprintf(pm.prometheusQueryURL, url.QueryEscape(query))
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, queryURL, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	token := pm.mon.restconfig.BearerToken
+	if token == "" && pm.mon.restconfig.BearerTokenFile != "" {
+		tokenBytes, err := os.ReadFile(pm.mon.restconfig.BearerTokenFile)
```
stevekuznetsov, line 102: how often does the token change? seems like you should be loading it once at startup (or rotating it) but not during every single query

pkg/monitor/cluster/prometheusmetrics.go
```diff
@@ -0,0 +1,153 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("load bearer token file: %w", err)
+		}
+		token = string(tokenBytes)
+	}
+
+	if token != "" {
```
stevekuznetsov, line 109: if token _is_ empty, isn't this an error?

pkg/monitor/cluster/prometheusmetrics.go
```diff
@@ -0,0 +1,153 @@
...
+		req.Header.Set("Authorization", "Bearer "+token)
+	}
+
+	resp, err := pm.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
```
stevekuznetsov, line 117: handle your errors

pkg/monitor/cluster/prometheusmetrics.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	if queryResp.Status != "success" {
+		return nil, fmt.Errorf("prometheus query failed with status %s: %+v", queryResp.Status, queryResp.Data)
+	}
+
+	return queryResp.Data.Result, nil
+}
+
+func (pm *prometheusMetrics) emitCNVMetrics(ctx context.Context) error {
```
stevekuznetsov, line 140: can we use the meta-apis to list all kubevirt-related metrics, and, for each, figure out what kind they are (gauge, counter, etc), then generically query & emit all of them without having to list them here?

pkg/monitor/cluster/cluster.go
mociarain: It does matter. IIRC there is a window of 60s so I would put this above that last one. (There are plans to improve this in the works)

hlipsig: It does matter. Putting it at the end ensure the previous checks are guaranteed to finish. Especially on large clusters prometheus checks can time out. This is why I suggested a different approach via email.

pkg/monitor/cluster/prometheusmetrics.go
hlipsig, line 71: Why are we skipping verification?

pkg/monitor/cluster/cluster.go
hlipsig: We should put this below authentication type. That check cannot time out.

PR #4459, opened 2025-11-10
[ARO-22548] Add belt-and-braces checks to ensure that non-managed API certs aren't rotated by MIMO

hawkowl:

Fixes ARO-22548

Adds belt-and-braces checks to ensure that we don't try and rotate a non-managed certificate, in case the task gets run for some reason.

### Test plan for issue:

Unit tests

N/A

Unit tests

### DIFF
--- a/pkg/mimo/steps/cluster/tls.go
+++ b/pkg/mimo/steps/cluster/tls.go
@@ -17,7 +17,7 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/mimo"
 )
 
-func isManagedDomain(tc mimo.TaskContext) (string, error) {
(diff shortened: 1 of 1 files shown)


pkg/mimo/steps/cluster/tls.go
```diff
@@ -17,12 +17,37 @@ import (
 	"github.com/Azure/ARO-RP/pkg/util/mimo"
 )
 
+func isManagedDomain(tc mimo.TaskContext) (string, error) {
```
wanghaoran1988:
```suggestion
func managedDomain(tc mimo.TaskContext) (string, error) {
```

PR #4464, opened 2025-11-18
Preetisht/aro 20272 vmss ip tag

preetisht:

https://issues.redhat.com/browse/ARO-20272 

- VMSS IP Tags for both RP and Gateway
- ARM Conditional Logic to apply empty ipTag for disabled regions
- Configuration picked up from RP-Config

Fixes

https://issues.redhat.com/browse/ARO-16025

This is a part of Microsoft Security Wave 5 fix. All our PublicIP's for both MSIT and AME environment for ARO Classic shoudl have FirstPartyUsage tags . 

This PR implements IP tags support for Azure Red Hat OpenShift Virtual Machine Scale Sets (VMSS) to comply with Microsoft Security Wave 5 requirements. The implementation adds unified configuration parameters (vmssIpTags, vmssIpTagsDisabledRegions) that enable region-specific IP tagging for both RP and Gateway VMSS public IPs, with ARM template conditional logic that gracefully handles disabled regions by deploying empty IP tags arrays. The solution has ARM conditionals logic and includes  edge case handling for missing or empty configuration scenarios.

### Test plan for issue:

- [x] make test.go passes all the test.
- [x] Deployed in Canary eastus2euap. Post deployment verify if the new VMSS instances both RP and GWY have proper tagged IP's. 
- [x] Create a  cluster in Canary and check if the cluster is sending metrics to geneva.
- [x] Re-deployed in Canary eastus2euap with a different tag ( created after must modifying a comment ). Post deployment verify if the new VMSS instances both RP and GWY have proper tagged IP's. 
- [x] Create a  cluster in Canary and check if the cluster is sending metrics to geneva.
- [x] Reverted back canary to last known good release and it also went fine . 

- A doc will be created on how to deploy in regions where we do not FirstPartyUsage tags has not yet been provided to us by Microsoft . 
- We will create a TSG which will be a part of another PR which will be related to monitoring of our Public IP's which does not  have a FirstPartyUsage Tag .

- We will do an extensive tests in Canary region and updated all our test cases and results in JIRA . 
- [x] All the artifacts are uploaded and JIRA commets updated .

### DIFF
--- a/pkg/deploy/config.go
+++ b/pkg/deploy/config.go
@@ -101,8 +101,10 @@ type Configuration struct {
 	SubscriptionResourceGroupLocation  *string                `json:"subscriptionResourceGroupLocation,omitempty" value:"required"`
 	VMSize                             *string                `json:"vmSize,omitempty" value:"required"`
 	VMSSCleanupEnabled                 *bool                  `json:"vmssCleanupEnabled,omitempty"`
-	VmssIpTags                         []IPTag                `json:"vmssIpTags,omitempty"`
-	VmssIpTagsDisabledRegions          []string               `json:"vmssIpTagsDisabledRegions,omitempty"`
+	RPVmssIpTags                       []IPTag                `json:"rpVmssIpTags,omitempty"`
+	RPVmssIpTagsDisabledRegions        []string               `json:"rpVmssIpTagsDisabledRegions,omitempty"`
+	GwyVmssIpTags                      []IPTag                `json:"gwyVmssIpTags,omitempty"`
+	GwyVmssIpTagsDisabledRegions       []string               `json:"gwyVmssIpTagsDisabledRegions,omitempty"`
 	OIDCStorageAccountName             *string                `json:"oidcStorageAccountName,omitempty" value:"required"`
 	OtelAuditQueueSize                 *string                `json:"otelAuditQueueSize,omitempty" value:"required"`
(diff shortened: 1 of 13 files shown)


pkg/deploy/config.go
```diff
@@ -101,6 +101,10 @@ type Configuration struct {
 	SubscriptionResourceGroupLocation  *string                `json:"subscriptionResourceGroupLocation,omitempty" value:"required"`
 	VMSize                             *string                `json:"vmSize,omitempty" value:"required"`
 	VMSSCleanupEnabled                 *bool                  `json:"vmssCleanupEnabled,omitempty"`
+	RPVmssIpTags                       []IPTag                `json:"rpVmssIpTags,omitempty"`
+	RPVmssIpTagsDisabledRegions        []string               `json:"rpVmssIpTagsDisabledRegions,omitempty"`
```
mociarain, line 105: I've never considered it until now but could we do better than strings for regions i.e. enums. Outside the scope of this PR but I came across this: https://gist.github.com/ausfestivus/04e55c7d80229069bf3bc75870630ec8

What do you think?

mociarain (commented):
LGTM. I left a comment on an idea but it's not blocking

pkg/deploy/config.go
preetisht: @mociarain , will definitely consider, for now this is anyways on hold .

PR #4466, opened 2025-11-19
Getting default flag for versions

sankur-codes:

This PR gets the value for "default" key for openshift versions

This will make it easier to know which version is set as default out of all the openshift versions in cosmosdb. This is needed as we do not have direct access to production database.

### Test plan for issue:

There are tests in place for the change.

NA

Testing in INT and Canary.


tiguelu (approved):
Approved. This was deployed to Canary and it displayed the default version in the GA output.

PR #4467, opened 2025-11-19
Metric to expose configured network MTU

rhamitarora:

[ARO-14232](https://issues.redhat.com/browse/ARO-14232)

This PR adds a new monitoring capability to track cluster network MTU (Maximum Transmission Unit) configuration in ARO clusters.

**Changes:**
- Added `emitNetworkMTU()` function in `networkmtu.go` to collect MTU values from OpenShift Network API
- Added comprehensive unit tests covering different network types (OpenShiftSDN, OVNKubernetes)
- Emits `network.mtu` metric with MTU value and network type labels
- Fetches MTU from `configv1.Network` resource using `mon.configcli.ConfigV1().Networks().Get()`

This enhancement provides visibility into cluster network MTU settings, enabling better monitoring and troubleshooting of network-related issues in ARO environments.

### Test plan for issue:
Test cases added.

Will create TSG

This enables dashboard creation to display cluster MTU configuration

### DIFF
--- a/pkg/monitor/cluster/networkmtu_test.go
+++ b/pkg/monitor/cluster/networkmtu_test.go
@@ -5,6 +5,7 @@ package cluster
 
 import (
 	"context"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -23,6 +24,11 @@ import (
 func TestEmitNetworkMTU(t *testing.T) {
 	ctx := context.Background()
 
+	const (
+		networkType       string = "OVNKubernetes"
+		clusterNetworkMTU int    = 1340
+	)
+
 	tests := []struct {
 		name           string
 		oc             *api.OpenShiftCluster
@@ -30,7 +36,7 @@ func TestEmitNetworkMTU(t *testing.T) {
 		expectedMetric metricExpectation
 	}{
 		{
-			name: "MTU 1500 cluster with OpenShiftSDN",
+			name: "MTU cluster test",
 			oc: &api.OpenShiftCluster{
 				Properties: api.OpenShiftClusterProperties{
 					NetworkProfile: api.NetworkProfile{
@@ -43,47 +49,18 @@ func TestEmitNetworkMTU(t *testing.T) {
 					Name: "cluster",
 				},
 				Spec: configv1.NetworkSpec{
-					NetworkType: "OpenShiftSDN",
+					NetworkType: networkType,
 				},
 				Status: configv1.NetworkStatus{
-					ClusterNetworkMTU: 1500,
+					ClusterNetworkMTU: clusterNetworkMTU,
 				},
 			},
 			expectedMetric: metricExpectation{
 				name:  "network.mtu",
 				value: 1,
 				labels: map[string]string{
-					"mtu":          "1500",
-					"network_type": "OpenShiftSDN",
-				},
-			},
-		},
-		{
-			name: "MTU 1400 cluster with OVNKubernetes",
-			oc: &api.OpenShiftCluster{
-				Properties: api.OpenShiftClusterProperties{
-					NetworkProfile: api.NetworkProfile{
-						MTUSize: api.MTU1500,
-					},
-				},
-			},
-			networkConfig: &configv1.Network{
-				ObjectMeta: metav1.ObjectMeta{
-					Name: "cluster",
-				},
-				Spec: configv1.NetworkSpec{
-					NetworkType: "OVNKubernetes",
-				},
-				Status: configv1.NetworkStatus{
-					ClusterNetworkMTU: 1400,
-				},
-			},
-			expectedMetric: metricExpectation{
-				name:  "network.mtu",
-				value: 1,
-				labels: map[string]string{
-					"mtu":          "1400",
-					"network_type": "OVNKubernetes",
+					"mtu":          strconv.Itoa(clusterNetworkMTU),
+					"network_type": networkType,
 				},
 			},
 		},


pkg/monitor/cluster/networkmtu.go
```diff
@@ -0,0 +1,38 @@
...
+func (mon *Monitor) emitNetworkMTU(ctx context.Context) error {
+	networkConfig, err := mon.configcli.ConfigV1().Networks().Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	// Get the MTU value from network status
+	mtuString := strconv.Itoa(networkConfig.Status.ClusterNetworkMTU)
+
+	// Emit metric for configured MTU size
```
mociarain: Nit: You don't need these 2 comments. This [blog](https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/) explains is better than I will but in short comments should explain the "why" and never the "how"

pkg/monitor/cluster/networkmtu_test.go
```diff
@@ -0,0 +1,155 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			controller := gomock.NewController(t)
+			m := mock_metrics.NewMockEmitter(controller)
+			_, log := testlog.New()
+
+			// Create fake config client with network config
```
mociarain: Same nit as above

pkg/monitor/cluster/networkmtu_test.go
```diff
@@ -0,0 +1,155 @@
...
+			configcli := configfake.NewSimpleClientset(tt.networkConfig)
+
+			mon := &Monitor{
+				oc:        tt.oc,
+				configcli: configcli,
+				m:         m,
+				log:       log,
+			}
+
+			// Set expectation for the MTU metric
```
mociarain: Same nit as above

pkg/monitor/cluster/networkmtu_test.go
```diff
@@ -0,0 +1,155 @@
...
+				t.Errorf("emitNetworkMTU() error = %v", err)
+			}
+		})
+	}
+}
+
+func TestEmitNetworkMTUError(t *testing.T) {
+	ctx := context.Background()
+
+	// Test case where network config is not found
```
mociarain: Same nit as above

pkg/monitor/cluster/networkmtu_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	oc := &api.OpenShiftCluster{
+		Properties: api.OpenShiftClusterProperties{
+			NetworkProfile: api.NetworkProfile{
+				MTUSize: api.MTU1500,
+			},
+		},
+	}
+
+	// Empty client with no network config
```
mociarain: Same nit as above

pkg/monitor/cluster/networkmtu_test.go
```diff
@@ -0,0 +1,155 @@
...
+		oc:        oc,
+		configcli: configcli,
+		m:         m,
+		log:       log,
+	}
+
+	err := mon.emitNetworkMTU(ctx)
+	if err == nil {
+		t.Error("expected error when network config is not found, got nil")
+	}
```
mociarain, line 121: Is it worth asserting what the error is?

pkg/monitor/cluster/networkmtu_test.go
```diff
@@ -0,0 +1,121 @@
...
+				name:  "network.mtu",
+				value: 1,
+				labels: map[string]string{
+					"mtu":          "1500",
+					"network_type": "OpenShiftSDN",
+				},
+			},
+		},
+		{
+			name: "MTU 1400 cluster with OVNKubernetes",
```
mociarain: It's not clear to me what value this second test adds after the first one. Do we expect OVN to have some difference to the outcome?

pkg/monitor/cluster/networkmtu_test.go
mociarain: I think this test had value. Making sure we raise and error when we want to is a good idea. 

Why did you remove it?

pkg/monitor/cluster/networkmtu_test.go
rhamitarora: The second test validates different network types emit different MTU values (1400 vs 1500) and correct network_type labels, ensuring our monitoring works across both OpenShiftSDN and OVNKubernetes network implementations.

pkg/monitor/cluster/networkmtu_test.go
rhamitarora: This would help catch scenarios where the function fails for unexpected reasons (like permission errors, network issues, etc.) versus the intended test case of missing network configuration.
Added again.

pkg/monitor/cluster/networkmtu_test.go
mociarain: So, I maintain that in both tests fundamentally test "string in == string out". In fact, I think using specific strings when they don't actually "matter" can give us a false sense of security. 

I was playing around with this, and I did a small dummy PR to show what I mean: https://github.com/Azure/ARO-RP/pull/4493

Let me know what you think

rhamitarora:
> Also, you seem to be rebasing and force pushing after each review which gives no diffs and makes future reviews harder. For such a small PR it's not real an issue but in general I wouldn't suggest fwiw

I was doing `git commit --amend --no-eidt` to maintain single commit.

pkg/monitor/cluster/networkmtu_test.go
rhamitarora: Agreed with comment, which simplified and focused, single test case that tests the core functionality.

PR #4469, opened 2025-11-19
Hive upgrade artifact cache

shubhadapaithankar:

Related: [ARO-22319](https://issues.redhat.com/browse/ARO-22319)
Supports: [ARO-22319](https://issues.redhat.com/browse/ARO-22319)

Adds ARM/Bicep templates for ACR artifact cache to enable Hive upgrades in shared dev environments (v4-eastus, v4-westeurope, v4-australiaeast) and E2E. Hive team moved to new Quay repository requiring artifact cache instead of mirroring pipeline. Updates Hive script to pull from artifact cache ACR (`arolocaldevsvc` for dev, `arosvcdev` for E2E) and bumps to production version `f84d11f6765b20de5a6c66998f2114b6855e94e0`.

### Test plan for issue:

Manual deployment:
1. Deploy `arolocaldevsvc` ACR with artifact cache templates (mirrors existing `ocpciarotest` setup)
2. Configure credentials and cache rules for new Hive repo
3. Grant AKS clusters ACR pull access via role assignment
4. Deploy Hive with `HIVE_ACR_REGISTRY=arolocaldevsvc.azurecr.io`
5. Verify pods pull image: `redhat-services-prod/crt-redhat-acm-tenant/hive-operator/hive:f84d11f6765b20de5a6c66998f2114b6855e94e0`

Repeat for E2E using `arosvcdev` ACR.

No. Infrastructure templates only.

Dev/E2E only. Follows production pattern (ARO-20992) and existing artifact cache setup (`ocpciarotest`). Standard Azure monitoring and K8s pod status verification.


kimorris27 (changes requested):
The new templates and the changes to `hive-generate-config.sh` look good, but we're missing a few things:

- ARM/Bicep to deploy the ACR itself. Some things to keep in mind:
    - Since we have three shared development environments in different regions, using a single ACR name may result in conflicts upon deployment attempts. It may be easiest to use one ACR per region. I'm thinking the ACR would go in the `v4-$LOCATION` resource group.
    - We might want to consider the full service dev env as well. The full service dev env already includes its own ACR instance. I seem to recall the full service dev deployment process being broken though, so I'd say not to worry about it if it's still broken.
- We'll want to incorporate these templates into our existing deployment procedures and documentation as needed. Here's where the other functions used to deploy templates for the full service dev env live: https://github.com/Azure/ARO-RP/blob/master/hack/devtools/deploy-shared-env.sh

hack/devtools/deploy-shared-env.sh
```diff
@@ -83,6 +83,79 @@ deploy_aks_dev() {
             "sshRSAPublicKey=$(<secrets/proxy_id_rsa.pub)" >/dev/null
 }
 
+deploy_hive_acr_dev() {
+    echo "########## Deploying Hive ACR in RG $RESOURCEGROUP ##########"
+    local acr_name="arolocaldev${LOCATION}"
```
kimorris27: Nit: We could add an environment variable for the ACR name to `secrets/env` so that it doesn't have to be redefined in each of these functions.

hack/devtools/deploy-shared-env.sh
```diff
@@ -83,6 +83,79 @@ deploy_aks_dev() {
...
+    az deployment group create \
+        -g "$RESOURCEGROUP" \
+        -n hive-aks-acr-pull-role \
+        --template-file pkg/deploy/assets/aks-acr-pull-role.json \
+        --parameters \
+            "aksClusterName=$aks_cluster" \
+            "acrName=$acr_name" \
+            "acrResourceGroup=$RESOURCEGROUP" >/dev/null
+    echo "########## AKS cluster $aks_cluster granted pull access to $acr_name ##########"
+}
```
kimorris27, line 118: I didn't think about this until seeing all of the functions, but how about we group the AKS role assignment, credential set, and cache rules into a single Bicep/ARM template and then put that deployment plus the ACR deployment into a single function? That way the dependencies between the resources are captured in the code without needing as much documentation.

I would go further and say that the ACR resource could go in the same template, but I see that you were able to reuse an existing template for that, so I think it's fine in its own template.

hack/devtools/deploy-shared-env.sh
```diff
@@ -83,6 +83,79 @@ deploy_aks_dev() {
...
+    echo "########## Created ACR: $acr_name ##########"
+}
+
+deploy_hive_artifact_cache_credentials() {
+    echo "########## Deploying Hive artifact cache credentials in RG $RESOURCEGROUP ##########"
+    local acr_name="arolocaldev${LOCATION}"
+    
+    if [ -z "$HIVE_PULL_USERNAME" ] || [ -z "$HIVE_PULL_PASSWORD" ]; then
+        echo "ERROR: HIVE_PULL_USERNAME and HIVE_PULL_PASSWORD must be set"
+        echo "Contact Kipp/Adam for Hive pull secret credentials"
```
kimorris27: Instead of putting individual engineers' names in the code, let's put information about the Quay robot account on the team wiki for people to find. I'm not aware of an existing page for documenting these types of robot accounts, but we can ask the team in Slack.

kimorris27 (changes requested):
Thanks for the changes so far. I left a few more suggestions, and we're still missing updates to the shared dev env deployment docs: https://github.com/Azure/ARO-RP/blob/master/docs/prepare-a-shared-rp-development-environment.md

hack/devtools/deploy-shared-env.sh
shubhadapaithankar: Done! Now using `${HIVE_ACR_NAME:-arolocaldev${LOCATION}}` in both functions so it can be set in secrets/env.

hack/devtools/deploy-shared-env.sh
shubhadapaithankar: @kimorris27 Done! Consolidated into a single `hive-acr-cache-and-access.bicep` template that combines credential set, cache rules, and AKS role assignment. Dependencies are now handled in the template itself.

Reduced from 4 functions to 2:
- `deploy_hive_acr_dev()` - Creates ACR using existing ci-development.json
- `deploy_hive_acr_cache_and_access()` - Deploys everything else

This eliminates manual dependency management and reduces code from 170 to 85 lines.

hack/devtools/deploy-shared-env.sh
shubhadapaithankar: Done! Removed specific names and changed to "See team wiki for Hive pull secret credentials". We can create a wiki page for robot account credentials.

kimorris27 (changes requested):
The code changes look good! It's just missing the documentation update now.

It'd also be good to see it deployed and working in at least one of the three regional shared dev environments before merge.

PR #4470, opened 2025-11-19
Validate version before running static validation

cadenmarchese:

Fixes #itn-2025-00300

- Version validation populates the version if not specified by the user, so we need to run it before we do static validation so that static validation can determine if the SKU is valid or not.

- Better error handling when:
1. no default ocp version is set in frond end cache
2. no default ocp version can be retrieved from the sync (every 10 seconds) in version update change feed
3. better error explanation to the user/cu to suggest them using `--verison` param when there is no default ocp version available

### Test plan for issue:

- [ ] Unit tests
- [x] test in local RP, OR test in canary

### DIFF
--- a/pkg/frontend/openshiftcluster_putorpatch.go
+++ b/pkg/frontend/openshiftcluster_putorpatch.go
@@ -252,10 +252,11 @@ func (f *frontend) _putOrPatchOpenShiftCluster(ctx context.Context, log *logrus.
 	if isCreate {
 		putOrPatchClusterParameters.converter.ToInternal(ext, doc.OpenShiftCluster)
 
-		err = f.validateInstallVersion(ctx, doc.OpenShiftCluster)
+		err = f.setDefaultVersionIfEmpty(doc.OpenShiftCluster)
 		if err != nil {
 			return nil, err
 		}
+		ext = putOrPatchClusterParameters.converter.ToExternal(doc.OpenShiftCluster)
 
 		err = f.ValidateNewCluster(ctx, subscription, doc.OpenShiftCluster, putOrPatchClusterParameters.staticValidator, ext, putOrPatchClusterParameters.path)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/frontend/changefeed.go
```diff
@@ -80,17 +80,42 @@ func (f *frontend) updateOcpVersions(docs []*api.OpenShiftVersionDocument) {
...
 			}
 		}
 	}
+
+	// After processing all documents, check if we have a valid default version
+	if f.baseLog != nil {
+		if len(f.enabledOcpVersions) > 0 && f.defaultOcpVersion == "" {
+			f.baseLog.Errorf("No default OpenShift version is set in CosmosDB. %d enabled version(s) available but none marked as default. Clusters created without --version parameter will fail.", len(f.enabledOcpVersions))
+		} else if hadCachedDefaultVersionBefore && !foundDefaultInUpdate && f.defaultOcpVersion == "" {
+			f.baseLog.Error("Default OpenShift version was removed and no replacement was set. Clusters created without --version parameter will fail.")
```
mociarain, line 116: Note: All of this makes me think this whole thing should be in a transaction but looking at the code I don't think we have that ability with our current DBClient... Am I right in thinking this?

Tof1973: Not sure what you mean by "be in a transaction", but we are already protected for external access/changes by using ocpVersionsMu mutex which is locked during this function, and which is used by others that access versions. Does it reply to your concern?
In all cases, I guess we should not address this refactor (if any) within this hotfix.

tiguelu: I don´t know if we have capability to do transactions with the DB client, but this code isn´t reading or writing to CosmosDB The error message that says a defailt wasn´t found in CosmosDB can be misleading. The default version is really coming from a cache from the DB stored in the frontend.

Like @Tof1973 said, the mutex on OCP version should be the mechanism that prevents this cache from being updated in the middle of other logic and makes it "transactional".

PR #4475, opened 2025-11-21
Use 4.16.0 version in VMSizeIsValidForVersion if we cannot parse version

Tof1973:

Fixes #itn-2025-00300

Use a 4.16.0 version when version cannot be get.
This is temporary solution, until we find proper fix

### DIFF
--- a/pkg/api/validate/vm.go
+++ b/pkg/api/validate/vm.go
@@ -388,7 +388,10 @@ func VMSizeIsValidForVersion(vmSize api.VMSize, requireD2sWorkers, isMaster bool
 
 	clusterVersion, err := version.ParseVersion(v)
 	if err != nil {
-		return false
+		// return false
+		// As temporary fix for itn-2025-00300, we will use version 4.16 for the following check if version cannot be determined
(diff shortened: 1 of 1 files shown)


pkg/api/validate/vm.go
```diff
@@ -388,7 +388,10 @@ func VMSizeIsValidForVersion(vmSize api.VMSize, requireD2sWorkers, isMaster bool
 
 	clusterVersion, err := version.ParseVersion(v)
 	if err != nil {
-		return false
+		// return false
+		// As temporary fix for itn-2025-00300, we will use version 4.16 for the following check if version cannot be determined
```
hlipsig, line 392: 4.16 is in EUS, why not 4.17? Is there a higher rate of install failures on 4.17 compared to 4.16?

mociarain (changes requested):
Can you give more context? If we can't parse a Version we should fail IMO. This is a process we own entirely so setting defaults smells to me

Tof1973:
Proper fix with https://github.com/Azure/ARO-RP/pull/4470

PR #4476, opened 2025-11-21
Improve logging of e2e cluster cleanup

cadenmarchese:

Fixes (no jira)

- Currently, when we delete an e2e cluster in CI or ev2, we append all deletion errors together, check for "not found" and log if the string matches, otherwise we don't log anything other than successes.
- If there's a "not found" error from something other than the cluster resource or somewhere else in the deletion flow, this is treated as a "cluster not found" error when the "not found" error could have come from elsewhere.
- There are many steps to e2e cluster cleanup, and so if a cluster fails to clean up, we should add logging to clearly understand what resource or step specifically failed to clean up rather than just silently failing.

### Test plan for issue:

- CI e2e should complete and show the logging.

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -735,42 +735,90 @@ func (c *Cluster) Delete(ctx context.Context, vnetResourceGroup, clusterName str
 		oc, err := c.openshiftclusters.Get(ctx, vnetResourceGroup, clusterName)
 		clusterResourceGroup := fmt.Sprintf("aro-%s", clusterName)
 		if err != nil {
-			c.log.Errorf("CI E2E cluster %s not found in resource group %s", clusterName, vnetResourceGroup)
-			errs = append(errs, err)
+			if azureerrors.IsNotFoundError(err) {
+				c.log.Infof("Cluster %s not found in resource group %s, assuming already deleted", clusterName, vnetResourceGroup)
+			} else {
+				c.log.Errorf("Failed to get cluster %s in resource group %s: %v", clusterName, vnetResourceGroup, err)
+				errs = append(errs, fmt.Errorf("failed to get cluster: %w", err))
+			}
 		}
 		if oc != nil {
 			if oc.Properties.ServicePrincipalProfile != nil {
-				errs = append(errs,
-					c.deleteApplication(ctx, oc.Properties.ServicePrincipalProfile.ClientID),
-				)
+				if err := c.deleteApplication(ctx, oc.Properties.ServicePrincipalProfile.ClientID); err != nil {
+					c.log.Errorf("Failed to delete application: %v", err)
+					errs = append(errs, fmt.Errorf("failed to delete application: %w", err))
+				}
 			}
 		}
 
-		errs = append(errs,
-			c.deleteCluster(ctx, vnetResourceGroup, clusterName),
-			c.deleteWimiRoleAssignments(ctx, vnetResourceGroup),
-			c.deleteWI(ctx, vnetResourceGroup),
-			c.ensureResourceGroupDeleted(ctx, clusterResourceGroup),
-			c.deleteResourceGroup(ctx, vnetResourceGroup),
-		)
+		if err := c.deleteCluster(ctx, vnetResourceGroup, clusterName); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/util/cluster/cluster.go
```diff
@@ -735,42 +735,89 @@ func (c *Cluster) Delete(ctx context.Context, vnetResourceGroup, clusterName str
 		oc, err := c.openshiftclusters.Get(ctx, vnetResourceGroup, clusterName)
 		clusterResourceGroup := fmt.Sprintf("aro-%s", clusterName)
 		if err != nil {
-			c.log.Errorf("CI E2E cluster %s not found in resource group %s", clusterName, vnetResourceGroup)
-			errs = append(errs, err)
+			if azureerrors.IsNotFoundError(err) {
+				c.log.Infof("Cluster %s not found in resource group %s, assuming already deleted", clusterName, vnetResourceGroup)
+			} else {
+				c.log.Errorf("Failed to get cluster %s in resource group %s: %v", clusterName, vnetResourceGroup, err)
+				errs = append(errs, fmt.Errorf("failed to get cluster: %w", err))
```
kimorris27, line 742: If we pass the error along and eventually log it elsewhere, aren't we duplicating the log by logging it here? Same question applies to most of the other log statements added in this file.

pkg/util/cluster/cluster.go
```diff
@@ -735,42 +735,89 @@ func (c *Cluster) Delete(ctx context.Context, vnetResourceGroup, clusterName str
...
+		}
+
+		if err := c.deleteVnetResources(ctx, vnetResourceGroup, "dev-vnet", clusterName); err != nil {
+			errs = append(errs, fmt.Errorf("failed to delete VNet resources: %w", err))
+		}
 	}
 
-	c.log.Info("done")
+	if len(errs) > 0 {
+		c.log.Errorf("Delete completed with %d error(s)", len(errs))
```
kimorris27:
```suggestion
		c.log.Errorf("Delete failed with %d error(s)", len(errs))
```

Nit: "Delete completed" threw me off when I first read it 🙂

pkg/util/cluster/cluster.go
```diff
@@ -1239,8 +1286,13 @@ func (c *Cluster) deleteWimiRoleAssignments(ctx context.Context, vnetResourceGro
 func (c *Cluster) deleteCluster(ctx context.Context, resourceGroup, clusterName string) error {
 	c.log.Printf("deleting cluster %s", clusterName)
 	if err := c.openshiftclusters.DeleteAndWait(ctx, resourceGroup, clusterName); err != nil {
+		if azureerrors.IsNotFoundError(err) {
```
kimorris27: I noticed you added these `IsNotFoundError` checks to some of the `deleteXXX` functions and not others. Is that because the other functions' delete calls are already idempotent?

pkg/util/cluster/cluster.go
cadenmarchese: Are you referring specifically to the `failed to get cluster` string we're adding to the error here and other places?

We could do a simpler errs = append(errs, err), only concern is it's less clear from the message which deletion function that the error is coming from without searching the code. For example, the deleteWimiRoleAssignments function can just fail with a `error listing role assignments for service principal` which on its own isn't super clear.

pkg/util/cluster/cluster.go
cadenmarchese: No, actually this was an oversight. It might make sense to use the IsNotFoundError in other places as well.

pkg/util/cluster/cluster.go
kimorris27: I think I see now - the idea is that including the log statement on line 741 will make it easier to find the source of the problem because that log line will include the precise location in the code where the error happened (?).

pkg/util/cluster/cluster.go
cadenmarchese: Hm, after looking it over, I think it's fine to leave the deleteXXX functions as-is and not log a special string if the resource is not found. I don't think it provides a benefit, but let me know if you disagree.

I did keep the `IsNotFoundError` checking on the overall cluster resource, since that probably makes sense to keep, but removed it in other places.

PR #4480, opened 2025-11-24
Add CNV VMI statuses collector

machadovilaca:

Fixes https://issues.redhat.com/browse/ARO-22503

This PR adds a cluster monitor exporter for CNV VMI statuses.

### Test plan for issue:

Added comprehensive unit tests and manually verified that it works correctly on a live cluster.

No documentation update is needed for this PR

error handling

### DIFF
--- a/pkg/monitor/cluster/cnvvmistatuses.go
+++ b/pkg/monitor/cluster/cnvvmistatuses.go
@@ -7,8 +7,9 @@ import (
 	"context"
 	"strings"
 
-	"k8s.io/apimachinery/pkg/api/meta"
 	k6tv1 "kubevirt.io/api/core/v1"
+
+	"k8s.io/apimachinery/pkg/api/meta"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )


mociarain:
> Create OpenShift CNV VM telemetry to enable the creation of monthly consumption reports.

So these metrics will end up in our Geneva monitoring stack within the Azure "walled garden" IIRC. Do we know how we plan to get them out of there for the creation of these reports?

hlipsig:
The metrics will be turned into a Dashboard that Business Stakeholders within Red Hat and Microsoft will use. No automated report generation needs to occur. They'll just pull an export down once a month or week or so.

go.mod
```diff
@@ -228,6 +229,7 @@ require (
 	github.com/opencontainers/runtime-tools v0.9.1-0.20250523060157-0ea5ed0382a2 // indirect
 	github.com/opencontainers/selinux v1.12.0 // indirect
 	github.com/openshift/custom-resource-status v1.1.3-0.20220503160415-f2fdb4999d87 // indirect
+	github.com/pborman/uuid v1.2.0 // indirect
```
hlipsig: This package does not seem to be maintained and the version of the Upstream is behind what was the last version from the Google Repo. We need to ensure everything we import can meet our requirements for FedRAMP Moderate and High Compliance. Is there a different package we can use, does the one we already import not suit?

go.mod
```diff
@@ -784,3 +788,5 @@ replace github.com/docker/distribution v2.8.3+incompatible => github.com/docker/
 
 // ARO-RP sub-packages
 replace github.com/Azure/ARO-RP/pkg/api => ./pkg/api
+
+replace kubevirt.io/containerized-data-importer-api => kubevirt.io/containerized-data-importer-api v1.55.0
```
hlipsig: How often will we need to update this API version? Not that it's blocking in any way, but we might want to validate it'll get updated via dependabot or something.

go.sum
```diff
@@ -293,6 +293,7 @@ github.com/google/pprof v0.0.0-20250630185457-6e76a2b096b5 h1:xhMrHhTJ6zxu3gA4en
 github.com/google/pprof v0.0.0-20250630185457-6e76a2b096b5/go.mod h1:5hDyRhoBCxViHszMt12TnOpEI4VVi+U8Gm9iphldiMA=
 github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510 h1:El6M4kTTCOh6aBiKaUGG7oYTSPP8MxqL4YI3kZKwcP4=
 github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510/go.mod h1:pupxD2MaaD3pAXIBCelhxNneeOaAeabZDe5s4K6zSpQ=
+github.com/google/uuid v1.0.0/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
```
hlipsig: I recognize this is needed for the other version of UUID to work, but I'm concerned about bringing in something that has been archived to the project for the same security and compliance reasons.

go.mod
hawkowl: Looks like kubevirt changed it in v1.2.0: https://github.com/kubevirt/kubevirt/commit/77f0f91de96d9290a4a8e71e4af35c3ae06c49a0

machadovilaca: updated to the latest compatible version, 1.5

go.sum
machadovilaca: removed

go.mod
machadovilaca: we need to fix the version because some incompatibilities with the k8s client on the original version

```
kubevirt.io/containerized-data-importer-api@v1.63.1/pkg/apis/core/v1beta1/types.go:82:19: undefined: corev1.VolumeResourceRequirements

kubevirt.io/containerized-data-importer-api@v1.63.1/pkg/apis/core/v1beta1/types.go:105:24: undefined: corev1.TypedObjectReference

kubevirt.io/containerized-data-importer-api@v1.63.1/pkg/apis/core/v1beta1/zz_generated.deepcopy.go:1837:17: undefined: v1.TypedObjectReference

FAIL    github.com/Azure/ARO-RP/pkg/monitor/cluster [build failed]
```

PR #4481, opened 2025-11-24
Add environment field in the logs to filter logs based on the environment in Dgrep and Kusto

mhsreddy:
**Which issue this PR addresses:**
Fixes: https://issues.redhat.com/browse/ARO-14708

**What this PR does / why we need it:**
This PR is to add new environment field in the logs to filter the logs in Kusto or Dgrep. This change enables us to use the production Logging Account in Geneva and filter the logs easily between Production and Stage using the new Environment Field that are injecting.

**Test plan for issue:**
Tested this in INT, Staging and Canary Environments.

**Is there any documentation that needs to be updated for this PR?**
N/A

**How do you know this will function as expected in production?**
The Logs should contain the Environment field in DGREP and Kusto for ARORPLogs, AROClusterLogs


hlipsig (approved):
Changes LGTM @mociarain will need to confirm since he requested changes.

pkg/deploy/generator/scripts/util-services.sh
```diff
@@ -331,6 +338,7 @@ ExecStart=/usr/bin/podman run \
   -e PORTAL_HOSTNAME \
   -e OTEL_AUDIT_QUEUE_SIZE \
   -e ARO_LOG_LEVEL \
+  -e ENIVRONMETN_TYPE \
```
kimorris27: Should this be `ENVIRONMENT`? If not, is there a typo? ("T" and "N" are swapped at end of word "ENVIRONMENT")

pkg/operator/deploy/deploy.go
```diff
@@ -366,6 +366,14 @@ func (o *operator) clusterObject() (*arov1alpha1.Cluster, error) {
 		},
 	}
 
+	// Set aro.environment operator flag for MONITORING_ENVIRONMENT in cluster logs
```
kimorris27: Should this be `ENVIRONMENT`?

pkg/deploy/generator/scripts/util-services.sh
mhsreddy: Yes Its typo. Will fix this.

pkg/operator/deploy/deploy.go
mhsreddy: Yep this should ENVIRONMENT.

PR #4483, opened 2025-11-24
[ARO-22853] Update Control Plane Machine Naming Convention Checks To Work With CPMS

kevinobriendotca:

Fixes https://issues.redhat.com/browse/ARO-22853

Makes the resize VM API resource compatible with machines that have been created via CPMS.

### Test plan for issue:

Added unit test for machine name validation for machines created via CPMS.

N/A

E2E

### DIFF
--- a/pkg/frontend/admin_openshiftcluster_vmresize.go
+++ b/pkg/frontend/admin_openshiftcluster_vmresize.go
@@ -75,7 +75,7 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
 			fmt.Sprintf("failed to decode machine object for %s, %s", vmName, err.Error()))
 	}
 
-	isControlPlaneMachine, err := utilmachine.IsMasterRole(machine)
+	isControlPlaneMachine, err := utilmachine.HasMasterRole(machine)
 	if err != nil {
 		return api.NewCloudError(http.StatusInternalServerError, api.CloudErrorCodeInternalServerError, "", err.Error())
 	}
--- a/pkg/operator/controllers/machine/machine.go
+++ b/pkg/operator/controllers/machine/machine.go
@@ -84,7 +84,7 @@ func (r *Reconciler) checkMachines(ctx context.Context) (errs []error) {
 	}
 
 	for _, machine := range machines.Items {
-		isMaster, err := utilmachine.IsMasterRole(&machine)
+		isMaster, err := utilmachine.HasMasterRole(&machine)
 		if err != nil {
 			errs = append(errs, err)
 			continue
--- a/pkg/util/machine/machine.go
+++ b/pkg/util/machine/machine.go
@@ -9,7 +9,7 @@ import (
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 )
 
-func IsMasterRole(m *machinev1beta1.Machine) (bool, error) {
+func HasMasterRole(m *machinev1beta1.Machine) (bool, error) {
 	role, ok := m.Labels["machine.openshift.io/cluster-api-machine-role"]
 	if !ok {
 		return false, fmt.Errorf("machine %s: cluster-api-machine-role label not found", m.Name)
--- a/pkg/util/machine/machine_test.go
+++ b/pkg/util/machine/machine_test.go
@@ -53,13 +53,13 @@ func TestIsMasterRole(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := IsMasterRole(tt.args.m)
+			got, err := HasMasterRole(tt.args.m)
 			if (err != nil) != tt.wantErr {
-				t.Errorf("IsMasterRole() error = %v, wantErr %v", err, tt.wantErr)
+				t.Errorf("HasMasterRole() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
 			if got != tt.want {
-				t.Errorf("IsMasterRole() got = %v, want %v", got, tt.want)
+				t.Errorf("HasMasterRole() got = %v, want %v", got, tt.want)
 			}
 		})
 	}


pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -55,20 +62,12 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
...
 }
 
-// A bland check, to validate if the node is master by checking vmName ends with pattern "-master-[0-9]"
-// return false, when the node is either not a master node or not adhering to the standard naming convention.
-// return true, if regexp satisfies
+// Check if VM name follows standard control plane node naming convention
+// vmName is expected to end with pattern "-master-[0-9]" or "-master-[0-9A-Za-z]{5}-[0-9]" for nodes created via CPMS
 func nodeIsMaster(vmName string) bool {
-	r := regexp.MustCompile(`.*-master-[0-9]{1,}$`)
+	r := regexp.MustCompile(`.*-master-([0-9A-Za-z]{5}-)?[0-9]$`)
```
kevinobriendotca: This change removes support for multi-digit suffix for the control plane machine names, ie -master-123.

There is another spot in the code where we only allow a single digit on the end of the machine name, so this aligns those two validation checks: https://github.com/Azure/ARO-RP/blob/71764a97456e3bdd0e99bd8e598e0ce2e9660c96/pkg/frontend/fixetcd.go#L190.

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -55,20 +62,12 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
...
-
 	return action.VMResize(ctx, vmName, vmSize)
 }
 
-// A bland check, to validate if the node is master by checking vmName ends with pattern "-master-[0-9]"
-// return false, when the node is either not a master node or not adhering to the standard naming convention.
-// return true, if regexp satisfies
+// Check if VM name follows standard control plane node naming convention
+// vmName is expected to end with pattern "-master-[0-9]" or "-master-[0-9A-Za-z]{5}-[0-9]" for nodes created via CPMS
 func nodeIsMaster(vmName string) bool {
```
ehvs: Based on the function name "nodeIsMaster" perhaps could use a validation from nodeconditions.go or machineconditions.go?


Or if we really want to check regex for the name, use a function name more aligned with the function purpose perhaps since the call seems to be coming from Azure VM not k8s. Eg vmValidMasterNameConvention

pkg/frontend/admin_openshiftcluster_vmresize.go
```diff
@@ -32,6 +32,13 @@ func (f *frontend) _postAdminOpenShiftClusterVMResize(log *logrus.Entry, ctx con
 	resourceType := chi.URLParam(r, "resourceType")
 	resourceGroupName := chi.URLParam(r, "resourceGroupName")
 
+	if !nodeIsMaster(vmName) {
+		return api.NewCloudError(http.StatusForbidden, api.CloudErrorCodeForbidden, "",
+			fmt.Sprintf(
+				`"The vmName '%s' provided cannot be resized. It is either not a master node or not adhering to the standard naming convention."`,
```
ehvs: This aligns with my other comment, I think if we care that is a proper `master k8s node` object the validation must be on that, and the error message specific on the problem found.

pkg/frontend/admin_openshiftcluster_vmresize.go
kevinobriendotca: I was trying to go light on the refactoring, but yeah, not a fan of that function name either.  I'll update it to isValidControlPlaneVMName.

And yes, ideally we'd be able to inspect the machine or node object to determine whether or not this is a control plane machine, but since we're calling this from the ResizeControlPlaneVMs GA operation and the validation for checking the k8s machine object to confirm we're dealing with a control plane machine is already done there, I'm OK with this extra check being against the VM name string.

There is a way to get the machine object by calling frontend.kubeActionsFactory using the cluster doc that is returned from the f.prepareAdminActions call that we are currently throwing away, then kubeActionsFactory.KubeGet the machine, check role, etc. but seems like overkill for this case.

Also, I feel like there is a case to do away with the name validation here entirely since the ResizeVMOperation GA has no name or machine type / role validation and could be used for infra or worker nodes, but this blocks that from happening.  I know we're not obligated to help customers with the infra or worker nodes, but I can see a special request coming to SREs to do that.

pkg/frontend/admin_openshiftcluster_vmresize.go
kevinobriendotca: See my other response.  And just to be clear, this isn't new, just moved to earlier in the execution to avoid some API / DB calls if the name doesn't pass validation.

pkg/frontend/admin_openshiftcluster_vmresize.go
kevinobriendotca: I ended up pulling the isMasterRole function out of operator/controllers/machine/machine.go and into a util function.

I reverted the change to pkg/frontend/fixetcd.go, so that file is still using a regex to check if it's a control plane machine or not.

pkg/frontend/admin_openshiftcluster_vmresize.go
ehvs: It seems to be only failing on the imports now, but the core changes LGTM.

pkg/frontend/admin_openshiftcluster_vmresize.go
ehvs: thank you Kevin, the core changes LGTM. 
Quick question, since we dont have Canary available now, are we confident that the changes done wont break current functionality?

kevinobriendotca: I just pushed the fixes for the vmresize tests.  We should be able to test this in Canary.  I was under the impression that Canary was excluded from the current CCOA.

ehvs (approved):
LGTM

PR #4530, opened 2025-12-18
MIWI E2E - give FPSP Network Contributor over vnet

kimorris27:

https://issues.redhat.com/browse/ARO-23278

CSP E2E already gives the FPSP Network Contributor over the vnet, but MIWI E2E wasn't for some reason. I noticed while working on another task that the only reason MIWI E2E works today is because the FPSP has Owner at subscription scope over the Classic E2E subscription. We should remove that Owner role assignment and rely on properly scoped role assignments.

### Test plan for issue:

First, to validate our understanding of the issue:

- [x] Remove FPSP's Owner assignment on Classic E2E sub
- [x] Run an E2E pipeline; observe that CSP E2E passes, while MIWI E2E fails during dynamic validation

To test my code changes:

- [x] Remove FPSP's Owner assignment on Classic E2E sub
- [x] Deploy these changes to canary; observe that E2E passes for both cluster types

N/A

Tested in canary.

### DIFF
--- a/pkg/util/cluster/cluster.go
+++ b/pkg/util/cluster/cluster.go
@@ -337,7 +337,7 @@ func (c *Cluster) createApp(ctx context.Context, clusterName string) (applicatio
 	return appDetails{appID, appSecret, spID}, nil
 }
 
-func (c *Cluster) SetupServicePrincipalRoleAssignments(ctx context.Context, diskEncryptionSetID string, clusterServicePrincipalID string) error {
+func (c *Cluster) SetupServicePrincipalRoleAssignments(ctx context.Context, diskEncryptionSetID string, principalIDs []string) error {
 	c.log.Info("creating role assignments")
 
 	for _, scope := range []struct{ resource, role string }{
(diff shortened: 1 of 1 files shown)


pkg/util/cluster/cluster.go
```diff
@@ -627,10 +627,21 @@ func (c *Cluster) Create(ctx context.Context) error {
...
+	if !c.Config.UseWorkloadIdentity {
+		c.log.Info("creating cluster service principal and FPSP role assignments")
+		principalIds = append(principalIds, appDetails.SPId)
 	} else {
-		c.log.Info("creating Classic role assignments")
-		c.SetupServicePrincipalRoleAssignments(ctx, diskEncryptionSetID, appDetails.SPId)
+		c.log.Info("creating FPSP role assignments")
 	}
+
+	c.SetupServicePrincipalRoleAssignments(ctx, diskEncryptionSetID, principalIds)
```
hawkowl: Shouldn't we check this error?

hawkowl (dismissed):
question

pkg/util/cluster/cluster.go
kimorris27: Good catch, thank you! It's fixed 👍🏻

cadenmarchese (approved):
lgtm

PR #4538, opened 2026-01-06
Disable default outbound access on master/worker subnets for hack/e2e clusters

tsatam:

Fixes N/A

Disables defaultOutboundAccess on the master and worker subnets created during pkg/util/cluster's cluster creation (used for E2E and localdev contexts). This is done in order to match the upcoming change to default this value to false on new subnets: https://azure.microsoft.com/en-us/updates?id=default-outbound-access-for-vms-in-azure-will-be-retired-transition-to-a-new-method-of-internet-access

Note that in order to enable the presence of the `defaultOutboundAccess` field on subnets resources, this PR updates our Microsoft.Network apiversion used for ARM template generation to 2021-05-01. This will have implications for all ARM templates we apply, including during RP deployments and cluster creation steps performed by the RP. 

### Test plan for issue:

- [x] E2E passes 

No

E2E-only change

### DIFF
--- a/pkg/util/azureclient/apiversions.go
+++ b/pkg/util/azureclient/apiversions.go
@@ -27,7 +27,7 @@ var apiVersions = map[string]string{
 	"microsoft.keyvault":                       "2019-09-01",
 	"microsoft.keyvault/vaults/accesspolicies": "2021-10-01",
 	"microsoft.managedidentity":                "2018-11-30",
-	"microsoft.network":                        "2020-08-01",
+	"microsoft.network":                        "2021-05-01",
 	"microsoft.network/dnszones":               "2018-05-01",
 	"microsoft.network/privatednszones":        "2018-09-01",
 	"microsoft.storage":                        "2021-09-01",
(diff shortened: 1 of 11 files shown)


pkg/util/azureclient/apiversions.go
```diff
@@ -27,7 +27,7 @@ var apiVersions = map[string]string{
 	"microsoft.keyvault":                       "2019-09-01",
 	"microsoft.keyvault/vaults/accesspolicies": "2021-10-01",
 	"microsoft.managedidentity":                "2018-11-30",
-	"microsoft.network":                        "2020-08-01",
+	"microsoft.network":                        "2021-05-01",
```
tuxerrante, line 30: Could we confirm this global `microsoft.network` API-version bump is intended for all callers (not only hack/e2e templates)? Admin/resource operations resolve this same map entry, so it can change behavior beyond the subnet default-outbound-access scope. Reference call site: https://github.com/Azure/ARO-RP/blob/201c1edc89c0dbb299bda341dcd6ebc21373fb6f/pkg/frontend/adminactions/delete_managedresource.go#L39

